Issue #1511, opened 2020-01-16
Fix rescan_block_device() to update disk_nsectors.


dianpopa:
The `rescan_block_device` function (which can only be triggered after the guest has booted up) does not update `disk_nsectors` field. This field is used in the `process_queue` so it is vital to the good functioning of the block's transport requests.
The change fixing this bug should also make sure this duplicated warning ([here](https://github.com/firecracker-microvm/firecracker/blob/refactoring/epoll_handler/src/devices/src/virtio/block.rs#L509) and [here](https://github.com/firecracker-microvm/firecracker/blob/refactoring/epoll_handler/src/vmm/src/controller.rs#L88)) gets deduplicated and is used both when `Block::new` is created and also at `rescan_block_device`.

ioanachirca:
Rescanning a block device can be done through a PATCH /drives API call.
More details can be found at `docs/api_requests/patch-block.md`.

PR #1542, opened 2020-01-23
Remove RescanBlockDevice from the /actions API

ioanachirca:
## Reason for This PR

The RescanBlockDevice /actions API call is redundant, the same
functionality can be achieved using a PATCH /drives API call.

This resolves issue #1511  because through the latter call,
the `ndisk_sectors` field that is used in the processing of the disk's
transport requests is updated correctly.

## Description of Changes

- Remove the RescanBlockDevice API call and related tests.
- Update the docs to include an example of using the PATCH /drives call 
to rescan a block device.
- Remove duplicated warning code about disk_size not being a multiple of SECTOR_SIZE.

## License Acceptance

## PR Checklist

- [x] The required doc changes are included in this PR. Docs in scope are all `*.md` files
      located in the `docs/` directory.
      code is included in this PR.
- [x] The API changes are reflected in `firecracker/swagger.yaml`.
- [x] The changes in this PR have user impact and have been added to the `CHANGELOG.md`   file.

### DIFF
--- a/src/api_server/src/request/actions.rs
+++ b/src/api_server/src/request/actions.rs
@@ -1,18 +1,17 @@
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use serde_json::Value;
-
 use super::super::VmmAction;
 use logger::{Metric, METRICS};
-use request::{Body, Error, ParsedRequest, StatusCode};
+#[cfg(target_arch = "aarch64")]
+use request::StatusCode;
+use request::{Body, Error, ParsedRequest};
 
 // The names of the members from this enum must precisely correspond (as a string) to the possible
 // values of "action_type" from the json request body. This is useful to get a strongly typed
 // struct from the Serde deserialization process.
 #[derive(Debug, Deserialize, Serialize)]
 enum ActionType {
-    BlockDeviceRescan,
     FlushMetrics,
     InstanceStart,
     SendCtrlAltDel,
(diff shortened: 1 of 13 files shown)


src/api_server/src/request/actions.rs
```diff
@@ -30,27 +29,8 @@ struct ActionBody {
...
-                None => Err(Error::Generic(
-                    StatusCode::BadRequest,
-                    "Payload is required for block device rescan.".to_string(),
-                )),
-            }
-        }
         ActionType::FlushMetrics | ActionType::InstanceStart | ActionType::SendCtrlAltDel => {
-            // Neither FlushMetrics nor InstanceStart should have a payload.
+            // Neither of these actions should have a payload.
             if action_body.payload.is_some() {
```
acatangiu: Please remove `payload` field altogether from the `ActionBody` struct. It was only used with `BlockDeviceRescan` and was complicating the code because of all the necessary checking for/against its presence. Let's take this opportunity to clean that up as well.

src/vmm/src/lib.rs
```diff
@@ -1546,14 +1546,7 @@ impl Vmm {
...
                     .map_err(|_| DriveError::BlockDeviceUpdateFailed)?;
-                if new_size % virtio::block::SECTOR_SIZE != 0 {
-                    warn!(
-                        "Disk size {} is not a multiple of sector size {}; \
-                         the remainder will not be visible to the guest.",
-                        new_size,
-                        virtio::block::SECTOR_SIZE
-                    );
-                }
+                virtio::block::check_disk_size(new_size);
```
dianpopa: I d place the `check_disk_size` function inside the `build_config_space` function since these two are tightly coupled. Every time you look at size you will also end up modifying the config, right? In  this scenario the vmm calls into the device manager which ultimately calls into the device for rebuilding config space.

ioanachirca: I moved the warning from the `check_disk_size` function inside `build_config_space`.

src/api_server/src/request/actions.rs
ioanachirca: As it would require more changes to the API to get rid of `ActionBody` completely, I only removed the `payload` field.

PR #1500, opened 2020-01-12
jailer: remove hardcoded api socket location

KarthikNedunchezhiyan:
Signed-off-by: karthik nedunchezhiyan <karthik1705.n@gmail.com>

## Reason for This PR

`--api-sock` value was constant in jailer and not allowed to modify by the user.

## Description of Changes

Removed constant `--api-sock` from the jailer. Now user can pass `--api-sock` value as an extra argument (after --) to the jailer.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria. Where there are two options, keep one.]`
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [ ] All commits in this PR are signed (`git commit -s`).
- [ ] Either this PR is linked to an issue, or, the reason for this PR is
      clearly provided.
- [ ] The description of changes is clear and encompassing.
- [ ] Either no docs need to be updated as part of this PR, or, the required
      doc changes are included in this PR. Docs in scope are all `*.md` files
      located either in the repository root, or in the `docs/` directory.
- [ ] Either no code has been touched, or, code-level documentation for touched
      code is included in this PR.
- [ ] Either no API changes are included in this PR, or, the API changes are
      reflected in `firecracker/swagger.yaml`.
- [ ] Either the changes in this PR have no user impact, or, the changes in
      this PR have user impact and have been added to the `CHANGELOG.md` file.
- [ ] Either no new `unsafe` code has been added, or, the newly added `unsafe`
      code is unavoidable and properly documented.

PR #1548, opened 2020-01-25
Fix regression in jailer

### LINKED ISSUE #1551: CI: Exercise default param scenarios
Test default parameter scenarios during integration testing for the jailer/firecracker. This would have avoided regression from #1548.

dianpopa:
## Reason for This PR

Fix for [regression](https://github.com/firecracker-microvm/firecracker/pull/1500/commits/1e5e6ac45fb0b4cdbadbf11d9355876f5a09089e).
When run only with the required parameter the  jailer would crash.
Fixes #1551.

## Description of Changes

* Added a new function which constructs the required folder hierarchy for firecracker to be able to run. Momentarily these are `dev`, `dev/net` and `run.` 
* The function also gives the necessary permissions.
* First two commits are nit's. The third one is the one fixing the regression.
* The 4th, 5th, 6th and 7th are integration tests related commits.

## License Acceptance

## PR Checklist

- [x] The reason for this PR is
      clearly provided.

### DIFF
--- a/tests/integration_tests/security/test_jail.py
+++ b/tests/integration_tests/security/test_jail.py
@@ -2,6 +2,25 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests that verify the jailer's behavior."""
 import os
+import stat
+
+# These are the permissions that all files/dirs inside the jailer have.
+REG_PERMS = stat.S_IRUSR | stat.S_IWUSR | \
+            stat.S_IXUSR | stat.S_IRGRP | stat.S_IXGRP | \
+            stat.S_IROTH | stat.S_IXOTH
+DIR_STATS = stat.S_IFDIR | REG_PERMS
+FILE_STATS = stat.S_IFREG | REG_PERMS
+# These are the stats of the devices created by tha jailer.
+CHAR_STATS = stat.S_IFCHR | stat.S_IRUSR | stat.S_IWUSR
+
+
+def check_stats(filepath, stats, uid, gid):
+    """Assert on uid, gid and expected stats for the given path."""
+    st = os.stat(filepath)
+
+    assert st.st_gid == gid
+    assert st.st_uid == uid
+    assert st.st_mode ^ stats == 0
 
 
 def test_default_chroot(test_microvm_with_ssh):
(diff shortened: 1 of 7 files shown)


KarthikNedunchezhiyan:
Ohh, i missed this case. It went good as we run integ test with /api.socket alone. We need to add a testcase with default parameter for jailer 🙂

dianpopa:
> This has been missed in our integ test as we tested with only /api.socket . We need to add a integ test case for the jailer with default parameter 🙂.

Indeed.

dianpopa:
I added integration tests for the regression. Please review.

tests/integration_tests/security/test_jail.py
```diff
@@ -79,3 +80,15 @@ def test_default_chroot_hierarchy(test_microvm_with_initrd):
                 CHAR_STATS, test_microvm.jailer.uid, test_microvm.jailer.gid)
     check_stats(os.path.join(test_microvm.jailer.chroot_path(),
                              "firecracker"), FILE_STATS, 0, 0)
+
+
+def test_arbitrary_usocket_location(test_microvm_with_initrd):
```
aghecenco: Does this need to be `test_microvm_with_initrd`?

tests/integration_tests/security/test_jail.py
```diff
@@ -79,3 +80,15 @@ def test_default_chroot_hierarchy(test_microvm_with_initrd):
...
+    """Test arbitrary location scenario for the api socket."""
+    test_microvm = test_microvm_with_initrd
+    test_microvm.jailer.extra_args = {'api-sock': 'api.socket'}
```
aghecenco: Can we use a `tempfile` path instead of `api.socket`?

dianpopa: Hm it is a bit tricky since we would need to create the tempfile inside the chroot but the chroot is not created only after the jailer was spawned...

tests/integration_tests/security/test_jail.py
aghecenco: I get it now, it is kinda confusing to see `initrd` here without knowing this secret :smiley:

tests/integration_tests/security/test_jail.py
aghecenco: OK, not worth the extra complexity, `api.socket` will do

PR #1494, opened 2020-01-08
Migrate Metrics and the Serial device to EventManager (polly)

acatangiu:
## Reason for This PR

Another step in migrating Firecracker to the new `polly` event manager.

Fixes #1522 
Fixes #1523 

### Metrics

Both `LOGGER` and `METRICS` are Firecracker specific, not `Vmm` specific, and thus should not be tied to the `Vmm`.
Logger has already been taken out of `Vmm`, now take the metrics periodic logging logic out of the `Vmm` as well.

A `PeriodicMetrics` object is defined in the firecracker crate, which uses `EventManager` to drive periodic reporting of `METRICS`.

### Serial

The serial device no longer has to be injected with input from some external component. It can now be created with any input object that is `dyn std::io::Read + std::os::unix::io::AsRawFd`.
Because the serial now implements the `EventHandler` trait, it can internally handle all input.

Removed the `EpollDispatch::Stdin` previously handled by the `Vmm`.

The builder can now build a `Serial` device with any input and/or output sources.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

PR #1550, opened 2020-01-27
Migrate the Vmm-exit and Api events to EventManager (polly)

### LINKED ISSUE #1524: Migrate the Vcpu-Exit event to using EventManager (polly)

### LINKED ISSUE #1546: Migrate the Vmm ControlAction event to using EventManager (polly)

acatangiu:
## Reason for This PR

Fixes #1524 
Fixes #1546 

### Exit event

The `Vmm` object registers its exit event to `EventManager` and provides a `handle_read` routine for it.
On EPOLLIN/read events, the `Vmm` will call its `stop()` routine and exit the process.
`Vmm::run_event_loop` becomes a static function since it no longer needs a `self: &Vmm`, this brings us one step closer to removing it altogether.
Since `Vmm` implements `EventHandler`, it now has to be wrapped in an `Arc<Mutex>`.

### Api events

The RPC `PrebootApiController` now also has a method to drive itself all the way to the `Vmm` being built.
The same logic was previously in the `firecracker` crate, but through some refactoring, it's now generic enough to live in `rpc_interface.rs`.

The `ApiServerAdapter` registers itself to `EventManager` and dispatches runtime api messages to the `RuntimeApiController`.

The legacy `EpollContext` is now cascaded under `EventManager` instead of the other way around. `EpollContext` currently only handles `DeviceHandler` events for `net` and `vsock`. Once `net` and `vsock` are also migrated to using `EventManager`, `EpollContext` can be removed.

`EventManager` is no longer a member of `VmmController`, it needs to live at the top level since it now contains all other Firecracker components.

### Bonus: decouple Vmm from io::Stdin

Created `vmm::CustomData` trait that defines two functions which will be called on microVm boot and on microVm teardown respectivelly.

The Vmm now holds an Option<Box<dyn CustomData>> and calls the init and stop functions implemented by the actual custom object on `start_vcpus()` and on `stop()` respectivelly.

The Firecracker-specific Vmm is given a `SerialStdin` which implements the `CustomData` trait and which sets the terminal in raw mode on boot and back to canonic mode on Vmm stop.

## License Acceptance

### DIFF
--- a/src/polly/src/event_manager.rs
+++ b/src/polly/src/event_manager.rs
@@ -63,7 +63,7 @@ impl EventHandlerData {
 
 /// A trait to express the ability to respond to I/O event readiness
 /// using callbacks.
-pub trait EventHandler: Send {
+pub trait EventHandler {
     /// Handle a read event (EPOLLIN).
     fn handle_read(&mut self, _source: Pollable) -> Vec<PollableOp> {
         vec![]
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -12,7 +12,6 @@ use std::sync::{Arc, Mutex};
 use super::{EpollContext, Vmm};
 
 use super::Error;
-use device_manager;
 #[cfg(target_arch = "x86_64")]
 use device_manager::legacy::PortIODeviceManager;
 use device_manager::mmio::MMIODeviceManager;
@@ -22,6 +21,7 @@ use devices::virtio::MmioTransport;
 use memory_model::{GuestAddress, GuestMemory, GuestMemoryError};
 use polly::event_manager::{Error as EventManagerError, EventManager};
 use utils::eventfd::EventFd;
+use utils::terminal::Terminal;
 use utils::time::TimestampUs;
 use vmm_config;
 use vmm_config::boot_source::BootConfig;
@@ -29,6 +29,7 @@ use vmm_config::drive::BlockDeviceConfigs;
 use vmm_config::net::NetworkInterfaceConfigs;
 use vmm_config::vsock::VsockDeviceConfig;
 use vstate::{KvmContext, Vcpu, VcpuConfig, Vm};
+use {device_manager, VmmEventsObserver};
 
 /// Errors associated with starting the instance.
 #[derive(Debug)]
@@ -169,8 +170,14 @@ impl Display for StartMicrovmError {
     }
 }
 
-// Wrapper over io::Stdin that implements `ReadableFd`.
+// Wrapper over io::Stdin that implements `Serial::ReadableFd` and `vmm::VmmEventsObserver`.
 struct SerialStdin(io::Stdin);
+impl SerialStdin {
+    /// Returns a `SerialStdin` wrapper over `io::stdin`.
+    pub fn get() -> Self {
+        SerialStdin(io::stdin())
+    }
+}
 impl io::Read for SerialStdin {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         self.0.read(buf)
@@ -182,6 +189,21 @@ impl AsRawFd for SerialStdin {
     }
 }
 impl devices::legacy::ReadableFd for SerialStdin {}
+impl VmmEventsObserver for SerialStdin {
+    fn on_vmm_boot(&mut self) -> std::result::Result<(), utils::errno::Error> {
+        // Set raw mode for stdin.
+        self.0.lock().set_raw_mode().map_err(|e| {
+            warn!("Cannot set raw mode for the terminal. {:?}", e);
+            e
+        })
+    }
+    fn on_vmm_stop(&mut self) -> std::result::Result<(), utils::errno::Error> {
+        self.0.lock().set_canon_mode().map_err(|e| {
+            warn!("Cannot set canonical mode for the terminal. {:?}", e);
+            e
+        })
+    }
+}
 
 /// Builds and starts a microVM based on the current Firecracker VmResources configuration.
 ///
@@ -217,7 +239,7 @@ pub fn build_microvm(
     {
         Some(setup_serial_device(
             event_manager,
-            Box::new(SerialStdin(io::stdin())),
+            Box::new(SerialStdin::get()),
             Box::new(io::stdout()),
         )?)
     } else {
@@ -277,7 +299,7 @@ pub fn build_microvm(
     }
 
     let mut vmm = Vmm {
-        stdin_handle: io::stdin(),
+        events_observer: Some(Box::new(SerialStdin::get())),
         guest_memory,
         kernel_cmdline,
         vcpus_handles: Vec::new(),
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -214,7 +214,7 @@ pub fn build_microvm(
     epoll_context: &mut EpollContext,
     event_manager: &mut EventManager,
     seccomp_level: u32,
-) -> std::result::Result<Vmm, StartMicrovmError> {
+) -> std::result::Result<Arc<Mutex<Vmm>>, StartMicrovmError> {
     let boot_config = vm_resources
         .boot_source()
         .ok_or(StartMicrovmError::MissingKernelConfig)?;
@@ -255,11 +255,22 @@ pub fn build_microvm(
         (arch::IRQ_BASE, arch::IRQ_MAX),
     );
 
+    let exit_evt = EventFd::new(libc::EFD_NONBLOCK)
+        .map_err(Error::EventFd)
+        .map_err(StartMicrovmError::Internal)?;
+
     #[cfg(target_arch = "x86_64")]
     // Safe to unwrap 'serial_device' as it's always 'Some' on x86_64.
-    let mut pio_device_manager = PortIODeviceManager::new(serial_device.unwrap())
-        .map_err(Error::CreateLegacyDevice)
-        .map_err(StartMicrovmError::Internal)?;
+    // x86_64 uses the i8042 reset event as the Vmm exit event.
+    let mut pio_device_manager = PortIODeviceManager::new(
+        serial_device.unwrap(),
+        exit_evt
+            .try_clone()
+            .map_err(Error::EventFd)
+            .map_err(StartMicrovmError::Internal)?,
+    )
+    .map_err(Error::CreateLegacyDevice)
+    .map_err(StartMicrovmError::Internal)?;
 
     let vcpus;
     // For x86_64 we need to create the interrupt controller before calling `KVM_CREATE_VCPUS`
@@ -303,7 +314,7 @@ pub fn build_microvm(
         guest_memory,
         kernel_cmdline,
         vcpus_handles: Vec::new(),
-        exit_evt: None,
+        exit_evt,
         vm,
         mmio_device_manager,
         #[cfg(target_arch = "x86_64")]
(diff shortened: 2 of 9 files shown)


src/vmm/src/device_manager/mmio.rs
```diff
@@ -276,25 +276,19 @@ impl MMIODeviceManager {
...
             .map_err(Error::RegisterIrqFd)?;
 
+        let raw_io_device = serial.clone();
+
         self.bus
-            .insert(bus_device, self.mmio_base, MMIO_LEN)
+            .insert(serial, self.mmio_base, MMIO_LEN)
             .map_err(|err| Error::BusError(err))?;
 
         // Register the RawIOHandler trait.
```
andreeaflorescu: With the current implementation the `RawIOHandler` should not be needed any more. We should revert to the implementation of the Serial to the state before the patch that introduced `RawIOHandler`.

src/polly/src/event_manager.rs
```diff
@@ -284,21 +284,14 @@ impl EventManager {
...
 
-    /// Register a new event handler.
-    /// handler.init() will specify the pollable and event set.
-    ///
     /// Returns the handler wrapped in an Arc<Mutex>.
-    pub fn register<T: EventHandler + 'static>(&mut self, handler: T) -> Result<Arc<Mutex<T>>> {
-        let wrapped_handler = Arc::new(Mutex::new(handler));
-        self.register_protected(wrapped_handler.clone())?;
-        Ok(wrapped_handler)
+    pub fn wrap_handler<T: EventHandler + 'static>(handler: T) -> Arc<Mutex<T>> {
```
andreeaflorescu: Why does this function belong to the `EventManager`?

src/devices/src/legacy/serial.rs
```diff
@@ -346,52 +372,100 @@ mod tests {
...
-            serial_out.buf.lock().unwrap().as_slice(),
-            &[b'a', b'b', b'c']
+            serial_out.internal.lock().unwrap().write_buf.as_slice(),
+            &RAW_INPUT_BUF
         );
     }
 
     #[test]
-    fn serial_input() {
+    fn serial_raw_input() {
```
andreeaflorescu: nit: tests name typically start with `test_` so that you can easily differentiate them from helper functions.

src/firecracker/src/metrics.rs
```diff
@@ -9,41 +9,50 @@ use polly::event_manager::EventHandler;
...
-    fn log_metrics() {
+    fn log_metrics(&mut self) {
         // Please note that, if LOGGER has no output file configured yet, it will write to
         // stdout, so logging will interfere with console output.
         if let Err(e) = LOGGER.log_metrics() {
             METRICS.logger.missed_metrics_count.inc();
             error!("Failed to log metrics: {}", e);
         }
+
+        // Only used in tests, but has virtually no cost in production.
```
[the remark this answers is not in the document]
acatangiu: error[E0658]: attributes on expressions are experimental

this is a simple counter with just increments on every flush (every 60s) so it's no runtimecost. Its purpose is to enable testing, without it we would have to drop the unittest..

Let's move this conversation to #1494

src/devices/src/legacy/serial.rs
```diff
@@ -67,10 +83,15 @@ pub struct Serial {
     baud_divisor: u16,
     in_buffer: VecDeque<u8>,
     out: Option<Box<dyn io::Write + Send>>,
+    input: Option<SizedReadableFd>,
 }
 
 impl Serial {
-    fn new(interrupt_evt: EventFd, out: Option<Box<dyn io::Write + Send>>) -> Serial {
+    fn new(
```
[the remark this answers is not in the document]
acatangiu: I agree, but there's a compromise we have to make here in order to move the `input` and `output` to setters instead of having specialized constructors:
- with setters we introduce usage complexity around registering the Serial device with EventManager. If one would create the Serial device, then register it with EventManager without having also called `.set_input()`, it would not get any input events. Calling `.set_input()` later would not also register the input with EventManager as that needs to happen externally.

I think having specialized constructors provides a better UX, maybe we can work on the naming.

Also, let's move this conversation to #1494

src/vmm/src/lib.rs
```diff
@@ -487,16 +447,16 @@ impl Vmm {
...
-    fn get_serial_device(&self) -> Option<Arc<Mutex<dyn RawIOHandler>>> {
-        Some(self.pio_device_manager.stdio_serial.clone())
-    }
-
-    #[cfg(target_arch = "aarch64")]
-    fn get_serial_device(&self) -> Option<&Arc<Mutex<dyn RawIOHandler>>> {
-        self.mmio_device_manager
-            .get_raw_io_device(DeviceType::Serial)
-    }
+    //    #[cfg(target_arch = "x86_64")]
```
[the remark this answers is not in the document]
acatangiu: Fixed in #1494

src/vmm/src/device_manager/mmio.rs
acatangiu: Done in #1494

src/polly/src/event_manager.rs
acatangiu: Fixed in #1494

src/devices/src/legacy/serial.rs
acatangiu: Fixed in #1494

src/vmm/src/builder.rs
```diff
@@ -169,8 +170,14 @@ impl Display for StartMicrovmError {
     }
 }
 
-// Wrapper over io::Stdin that implements `ReadableFd`.
+// Wrapper over io::Stdin that implements `Serial::ReadableFd` and `vmm::CustomData`.
 struct SerialStdin(io::Stdin);
+impl SerialStdin {
+    /// Returns a `SerialStdin` wrapper over `io::stdin`.
+    pub fn get() -> Self {
```
alexandruag: `new` seems a more appropriate name here.

src/vmm/src/builder.rs
```diff
@@ -182,6 +189,21 @@ impl AsRawFd for SerialStdin {
     }
 }
 impl devices::legacy::ReadableFd for SerialStdin {}
+impl CustomData for SerialStdin {
```
alexandruag: `CustomData` is not a very suggestive name, but I don't have a better idea for now. Maybe we can find something else by the time we merge this into master.

src/vmm/src/controller.rs
```diff
@@ -57,21 +58,23 @@ impl VmmController {
...
+        self.vmm.lock().unwrap().stop(exit_code)
     }
 
     /// Creates a new `VmmController`.
     pub fn new(
         epoll_context: EpollContext,
         event_manager: EventManager,
         vm_resources: VmResources,
-        vmm: Vmm,
+        vmm: Arc<Mutex<Vmm>>,
```
andreeaflorescu: As we discussed offline, I think we shouldn't wrap the whole Vmm object in an Arc<Mutex>. We can change this in a future PR by either deciding on dropping Arc<Mutex> for all implementors of `EpollHandler` or we can just lock parts of the Vmm in an Arc<Mutex>.

src/arch/src/aarch64/gic.rs
```diff
@@ -19,7 +19,7 @@ pub enum Error {
 type Result<T> = result::Result<T, Error>;
 
 /// Trait for GIC devices.
-pub trait GICDevice {
+pub trait GICDevice: Send {
```
alexandruag: It looks like the `Send` bound is only required because `EventHandler` has a `Send` bound as well (in `polly/src/event_manager.rs:66`).  I think we can remove that, and by doing so we no longer have to add this one, nor the `unsafe impl`s below.

acatangiu: Good catch! Since the Handler is wrapped in a Mutex it doesn't have to be Send itself.
I will remove the `Send` bound from `EventHandler` and that should remove the need to mark `GICDevice` as `Send` too.

src/vmm/src/builder.rs
acatangiu: Idiomatic naming implies that `new` returns a **new** instance of an object.

In this scenario we're in a bit of a gray area:
technically this function does provide a **new** wrapper over the Stdin handle, but the wrapper is used as the stdin-handle itself which is a globally unique object, also the wrapper has no state other than the global handle.

I chose `SerialStdin::get()` to hint at the fact that this is a unique global object and not an independent instantiation. I propose we keep it as `get` or find another name better suited than `new`.

src/vmm/src/builder.rs
acatangiu: +1

src/vmm/src/controller.rs
acatangiu: I would like to avoid wrapping the Vmm in an Arc<Mutex> as well, but as of now I have no better alternative.

We can refactor this later if we come up with a better solution. Right now, I think removing `EpollContext` is the highest priority.

andreeaflorescu: I can add an issue and we can think about a better solution after we finish with the decoupling. What do you say?

src/vmm/src/builder.rs
andreeaflorescu: Same here. I also don't have a better name in mind :(

src/vmm/src/controller.rs
acatangiu: 👍

src/vmm/src/builder.rs
acatangiu: What about `VmmUserData` or `VmmRegistration`?

The name should hint that:
It is data that Vmm doesn't understand, but owns, and is therefore implicitly tied to the Vmm lifetime.

andreeaflorescu: In this case, `Opaque` could also work.

acatangiu: Yes, `Opaque` is used in such cases. Should we go with simple `trait Opaque` or more descriptive `trait VmmOpaqueData`?

src/arch/src/aarch64/gic.rs
acatangiu: This is done in the latest PR version.

src/vmm/src/builder.rs
acatangiu: I've renamed to `trait VmmEventsObserver` and updated comments and description accordingly.

acatangiu:
@alexandruag @andreeaflorescu @sandreim please review

acatangiu:
The last commit reverts the `net over epoll-context` EDGE Triggered events functionality.

Doing edge triggered events in this PR would involve doing edge triggered event handling in cascaded epoll which is hard throwaway work.

We will reintroduce that functionality once we get rid of `epoll-context`.

acatangiu:
> The last commit reverts the net over epoll-context EDGE Triggered events functionality.
Doing edge triggered events in this PR would involve doing edge triggered event handling in cascaded epoll which is hard throwaway work.
We will reintroduce that functionality once we get rid of epoll-context.

We chose to merge the `net over polly PR` first so the commit reverting the edge triggered tap events has been removed from this PR.

Rebased over that, please re-review.

acatangiu:
@andreeaflorescu @alexandruag @sandreim PR has been rebased over net-with-polly, please review.

src/vmm/src/builder.rs
andreeaflorescu: I find the code to read nicer as it was before (i.e. without the `get` function). Is this function really needed?

src/vmm/src/rpc_interface.rs
```diff
@@ -156,22 +158,58 @@ impl<'a> PrebootApiController<'a> {
...
+        seccomp_level: u32,
+        epoll_context: &mut EpollContext,
+        event_manager: &mut EventManager,
+        firecracker_version: String,
+        recv_req: F,
+        respond: G,
+    ) -> (VmResources, Arc<Mutex<Vmm>>)
+    where
+        F: Fn() -> VmmAction,
+        G: Fn(std::result::Result<VmmData, VmmActionError>),
```
andreeaflorescu: nit: use more representative names than F & G.

PR #1558, opened 2020-01-29
tests: Run 'screen -ls' in a busy-waiting loop

gc-plp:
Signed-off-by: Gabriel Ionescu <gbi@amazon.com>

## Reason for This PR

Occasionally tests would fail due to the fact that the output of 'screen -ls' did not contain the screen session name.

Due to this issue, re.search() would return 'None', which would throw a generic exception.

## Description of Changes

Run 'screen -ls' in a busy waiting loop to wait for the given regular expression object. 
If the object is not found within the default time frame, raise an exception.

## License Acceptance

## PR Checklist

- [x] No API changes are included in this PR
- [x] The changes in this PR have no user impact
- [x] No new `unsafe` code has been added

### DIFF
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -18,10 +18,12 @@ import time
 from subprocess import run, PIPE
 
 from retry import retry
+from retry.api import retry_call
 
 import host_tools.memory as mem_tools
 import host_tools.network as net_tools
 
+import framework.utils as utils
 from framework.defs import MICROVM_KERNEL_RELPATH, MICROVM_FSFILES_RELPATH
 from framework.http import Session
 from framework.jailer import JailerContext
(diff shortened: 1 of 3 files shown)


tests/framework/utils.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
raduiliescu: Maybe have been nicer to just have a function that would run a command, then verify with a regexp and use retry_call - https://pypi.org/project/retry/ for handling the retry part?

dianpopa: I like @raduiliescu s suggestion. We already use `retry_call` in a bunch of places: https://github.com/firecracker-microvm/firecracker/blob/master/tests/framework/jailer.py#L201, https://github.com/firecracker-microvm/firecracker/blob/master/tests/framework/microvm.py#L357. We could stick to it across the CI framework.

gc-plp: I updated the PR to use the reply module.

tests/framework/microvm.py
```diff
@@ -333,10 +335,22 @@ def spawn(self):
...
-            out = run('screen -ls', shell=True, stdout=PIPE)\
-                .stdout.decode('utf-8')
-            screen_pid = re.search(r'([0-9]+)\.{}'.format(self._session_name),
-                                   out).group(1)
+            # Build a regex object to match (number).session_name
+            regex_object = re.compile(
+                r'([0-9]+)\.{}'.format(self._session_name))
+
+            # Check the output of 'screen -ls' until we get the PID or
+            # a timeout
```
iulianbarbu: What is this timeout referring to? It is the timeout specified in `pytest.ini`?

gc-plp: It refers to an exception thrown by the retry() call. I'll reword the comment.

tests/framework/microvm.py
```diff
@@ -18,9 +18,11 @@
 from subprocess import run, PIPE
 
 from retry import retry
+from retry.api import retry_call
 
 import host_tools.memory as mem_tools
 import host_tools.network as net_tools
+import framework.utils as utils
```
aghecenco: nit: move this lower in the block of imports from `framework`

tests/framework/utils.py
```diff
@@ -0,0 +1,32 @@
...
+# SPDX-License-Identifier: Apache-2.0
+"""Generic utility functions that are used in the framework."""
+import subprocess
```
aghecenco: nit: order importa alphabetically

gc-plp: Will fix, thanks.

I think that these checks should be added in the linter. It should be doable to do it through one of the current linters.

dianpopa:
gitlint complains:
```
1: T1 Title exceeds max length (51>50): "tests/licenses: add year 2020 as valid license year"
```

PR #1559, opened 2020-01-30
vmm: add KVM dirty bitmap fetching functionality

aghecenco:
## Reason for This PR

Adds a function in `vmm` to fetch the KVM dirty bitmap.

## Description of Changes

* the 1st commit cleans up some build warnings on `aarch64`, to make new warnings easier to spot.
* the 2nd commit reverts https://github.com/firecracker-microvm/firecracker/commit/ebc15f2eacaa4c26782a05f8e0cebe4c47512553 and *enables* KVM dirty page tracking by default, so that the dirty bitmap is always available.
* the 3rd commit introduces the actual function, which just calls the corresponding fn in [`kvm-ioctls`](https://github.com/rust-vmm/kvm-ioctls/blob/ee9a81a1241b320b4067b7d9bcf911da438e4db6/src/ioctls/vm.rs#L792) and aggregates the results per memory region.

## License Acceptance

## PR Checklist

- [x] The reason for this PR is  clearly provided.
- [x] The changes in this PR have user impact and have been added to the `CHANGELOG.md` file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
 - Decreased release binary size by ~15%.
 - Changed default API socket path to `/run/firecracker.socket`. This path
   also applies when running with the jailer.
-- Disabled KVM dirty page tracking by default.
+- Enabled KVM dirty page tracking by default.
 - Removed redundant RescanBlockDevice action from the /actions API.
   The functionality is available through the PATCH /drives API.
   See `docs/api_requests/patch-block.md`.
--- a/src/vmm/src/error.rs
+++ b/src/vmm/src/error.rs
@@ -333,6 +333,28 @@ impl Display for StartMicrovmError {
     }
 }
 
+/// Types of errors associated with retrieving the guest's dirty page bitmap.
+#[cfg_attr(test, derive(Debug))]
+pub enum DirtyBitmapError {
+    /// The guest memory is invalid.
+    GuestMemory(GuestMemoryError),
+    /// The guest isn't running.
+    InstanceNotInitialized,
+    /// The `KVM_GET_DIRTY_LOG` ioctl failed.
+    Ioctl(kvm_ioctls::Error),
+}
+
+impl Display for DirtyBitmapError {
+    fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
+        use DirtyBitmapError::*;
+        match *self {
+            GuestMemory(ref e) => write!(f, "Invalid guest memory: {:?}.", e),
+            InstanceNotInitialized => write!(f, "The guest is not running."),
+            Ioctl(ref e) => write!(f, "KVM get dirty bitmap ioctl failed. {}", e),
+        }
+    }
+}
+
 /// Types of errors associated with vmm actions.
 #[derive(Clone, Debug, PartialEq)]
 pub enum ErrorKind {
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -24,7 +24,7 @@
 - Decreased release binary size by ~15%.
 - Changed default API socket path to `/run/firecracker.socket`. This path
   also applies when running with the jailer.
-- Disabled KVM dirty page tracking by default.
+- Enabled KVM dirty page tracking by default.
```
dianpopa: I think the title of the PR would make a better changelog entry than this one. That `by default` is rather confusing.

aghecenco: I didn't mention the new functionality in the changelog because without an associated api it's not user facing; what if I remove the `by default`? I think the changelog should still mention that a flag previously disabled is now enabled.

src/vmm/src/error.rs
```diff
@@ -333,6 +333,28 @@ impl Display for StartMicrovmError {
     }
 }
 
+/// Types of errors associated with retrieving the guest's dirty page bitmap.
+#[derive(Debug)]
```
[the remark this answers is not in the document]
aghecenco: Only in unit tests, will update with a `cfg_attr` so that we don't have it in the normal build

CHANGELOG.md
dianpopa: What about the classical `Support for KVM dirty page tracking`?

src/vmm/src/vstate.rs
```diff
@@ -225,7 +228,7 @@ impl Vm {
                     guest_phys_addr: region.start_addr().raw_value() as u64,
                     memory_size: region.len() as u64,
                     userspace_addr: host_addr as u64,
-                    flags: 0,
+                    flags: KVM_MEM_LOG_DIRTY_PAGES,
```
serban300: Could this change cause any performance degradation ? Do we have any data on this ?

aghecenco: We don't have data (we don't even have a baseline to compare to), but it should introduce a performance penalty with all the new work the kernel has to do to account for dirty pages. I'm working on an additional enable/disable dirty logging mechanism so that we don't take the penalty if we don't care about the dirty bitmap.

acatangiu:
Please consider developing this functionality on the `refactoring/epoll_handler` branch.

I'm in the process of rebasing that branch on top of master so that we can merge. Any intermediate work should go to the refactoring branch so that it doesn't create conflicts.

aghecenco:
> Please consider developing this functionality on the `refactoring/epoll_handler` branch.
> 
> I'm in the process of rebasing that branch on top of master so that we can merge. Any intermediate work should go to the refactoring branch so that it doesn't create conflicts.

Closing this in favor of a new PR on the refactoring branch.

src/vmm/src/lib.rs
```diff
@@ -1763,6 +1795,90 @@ mod tests {
...
+                .unwrap(); // set PC
+            vcpu_fd
+                .set_one_reg(core_reg_base + 2 * 0, mmio_addr)
+                .unwrap(); // set X0
+        }
+
+        // Prepare a VM with 1 MB memory and 1 vCPU.
+        // Dirty a page and run the VM.
+        // Return after the first VM exit.
+        fn setup_with_code_and_run(&mut self) {
```
serban300: I can't find the new PR so I'm writing here just to make sure I don't forget.

This logic seems quite complicated. Could we test this by loading the `test_elf.bin` from the `kernel` crate and expect for a certain dirty map ? Or create a special elf that dirties only one page ?

aghecenco: If we load a full blown kernel, there's no way of knowing for sure how many pages it will dirty; and creating a custom kernel seems more complicated to me than loading a handful of asm (I also find it more transparent to view the asm itself). Do you think we should add more comments to explain the instructions?

aghecenco: Sounds good. I'll restructure the code for a more natural flow.
What kind of write do you mean? An MMIO is one option - it triggers a VM exit and validates the MMIO flow, but doesn't change anything in the dirty log - or an actual write in guest memory? The latter will dirty one more page, so we'd need to change the test passing condition (one page dirtied on `aarch64` and two on `x86_64`).

serban300: I meant an actual write in the guest memory. It would be nice to have the same logic for `ARM` and `x86_64`. So a write in the guest memory + a MMIO write. I thought this would dirty only one page.

Also do we actually need to generate a HALT or MMIO write ? Can't we get the dirty log while `KVM_RUN` is executing ?

Issue #1456, opened 2019-12-06
Occasional KVM_EXIT_SHUTDOWN and bad syscall (14) during VM shutdown


sipsma:
Firecracker-containerd recently updated a test case that previously ran 5 VMs to now run 100 and started seeing the following errors occur in ~0-3 VMs per test case ([issue](https://github.com/firecracker-microvm/firecracker-containerd/issues/358)).

As part of terminating VMs, we tell systemd in the guest to reboot the system. Most of the time this works fine, we see `reboot: machine restart` in the systemd logs and the VMM exits with 0 as expected right after.

Occasionally though, the guest reaches that `reboot: machine restart` line and then we see a few logs lines from Firecracker like:
```
2019-12-06T20:29:30.007995741 [anonymous-instance:INFO] Received KVM_EXIT_SHUTDOWN signal
2019-12-06T20:29:30.008087608 [anonymous-instance:ERROR] Shutting down VM after intercepting a bad syscall (14).
```
Followed by the VMM exiting with code 148. The [KVM_EXIT_SHUTDOWN case](https://github.com/firecracker-microvm/firecracker/blob/b359722c23ec69a27b74fbfe2de434a7caa1a4b5/src/vmm/src/vstate.rs#L586) and bad syscall (which appears to be rt_sigprocmask) ~seem to only ever occur together (never just one or the other)~ and only right after the reboot call should be occurring. 
  * EDIT: I've now very rarely seen `KVM_EXIT_SHUTDOWN` occur without an accompanying `rt_sigprocmask` call that results in the seccomp violation. When `KVM_EXIT_SHUTDOWN` occurs by itself, the VMM exits with 0 with no apparent issues.

Additionally, I don't see this for every VM that exits like that, but am occasionally getting a line like this in dmesg after a run of the test case where failures occur:
```
[10548769.845897] traps: fc_vcpu0[52462] trap invalid opcode ip:527d36 sp:7f7ee67f0590 error:0 in firecracker[400000+1ab000]
```
No idea if it's related or, if it is, why it wouldn't be written every time a VM exits with KVM_EXIT_SHUTDOWN+the bad syscall, but figured it was worth mentioning.

This same behavior occurs in both 0.19 (w/ rust1.35+musl) and a 0.20 binary I built from [here](https://github.com/serban300/firecracker/tree/release) (w/ rust1.38+musl).

[...]

PR #1586, opened 2020-02-05
vmm: keep vCPU threads running to avoid calling sigprocmask

kzys:
## Reason for This PR

#1456 - musl calls sigprocmask from pthread_exit(), but we don't want to whitelist the syscall.

This change is currently blocking firecracker-containerd to use Firecracker 0.20.0. While the issue did exist in 0.19.0, we are seeing this seccomp violation more than before somehow. See https://buildkite.com/firecracker-microvm/firecracker-containerd/builds/1508 for the actual test failure.

## Description of Changes

Basically follows what @acatangiu has proposed on #1456 - keep vCPU threads running to avoid calling pthread_exit().

## License Acceptance

## PR Checklist

- [x] Either this PR is linked to an issue, or, the reason for this PR is
      clearly provided.
      code is included in this PR.

### DIFF
--- a/src/vmm/src/vstate.rs
+++ b/src/vmm/src/vstate.rs
@@ -11,6 +11,7 @@ use std::io;
 use std::result;
 use std::sync::atomic::{fence, Ordering};
 use std::sync::mpsc::{channel, Receiver, Sender, TryRecvError};
+use std::sync::{Arc, Barrier};
 use std::thread;
 
 use super::TimestampUs;
@@ -570,6 +571,10 @@ impl Vcpu {
     pub fn start_threaded(mut self, seccomp_filter: BpfProgram) -> Result<VcpuHandle> {
         let event_sender = self.event_sender.take().unwrap();
         let response_receiver = self.response_receiver.take().unwrap();
+
+        let exit_barrier = Arc::new(Barrier::new(2));
+        let barrier = exit_barrier.clone();
+
         let vcpu_thread = thread::Builder::new()
             .name(format!("fc_vcpu {}", self.cpu_index()))
             .spawn(move || {
@@ -577,6 +582,11 @@ impl Vcpu {
                     .expect("Cannot cleanly initialize vcpu TLS.");
 
                 self.run(seccomp_filter);
+
+                // Don't exit from the thread until we have another cue on the VcpuHandle below.
+                // Thread cleanup on a libc may call syscalls which we don't want to whitelist.
+                // For example, musl calls sigprocmask. See #1456 for the detail.
+                barrier.wait();
             })
             .map_err(Error::VcpuSpawn)?;
 
@@ -584,6 +594,7 @@ impl Vcpu {
             event_sender,
             response_receiver,
             vcpu_thread,
+            exit_barrier,
         })
     }
 
@@ -825,6 +836,7 @@ pub struct VcpuHandle {
     event_sender: Sender<VcpuEvent>,
     response_receiver: Receiver<VcpuResponse>,
     vcpu_thread: thread::JoinHandle<()>,
+    exit_barrier: Arc<Barrier>,
 }
 
 impl VcpuHandle {
@@ -846,6 +858,7 @@ impl VcpuHandle {
 
     #[allow(dead_code)]
     pub fn join_vcpu_thread(self) -> thread::Result<()> {
+        self.exit_barrier.wait();
         self.vcpu_thread.join()
     }
 }


src/vmm/src/lib.rs
```diff
@@ -1108,6 +1122,13 @@ impl Vmm {
             self.exit_evt = Some(exit_poll_evt_fd);
         }
 
+        #[cfg(target_arch = "aarch64")]
+        {
+            self.epoll_context
+                .add_epollin_event(self.exit_evt.as_ref().unwrap(), EpollDispatch::Exit)
+                .map_err(|_| StartMicrovmError::RegisterEvent)?;
+        }
+
```
sandreim: Please see my first comment. With that change in place I think this whole function can be simplified to look like this: 
```rust
fn register_events(&mut self) -> std::result::Result<(), StartMicrovmError> {
            self.epoll_context
                .add_epollin_event(self.exit_evt.as_ref().unwrap(), EpollDispatch::Exit)
                .map_err(|_| StartMicrovmError::RegisterEvent)?;
}
```

src/vmm/src/vstate.rs
```diff
@@ -1245,13 +1272,17 @@ mod tests {
         // Stop it by sending exit.
         assert!(vcpu_handle.send_event(VcpuEvent::Exit).is_ok());
 
-        // Validate vCPU thread ends execution.
-        vcpu_handle
-            .join_vcpu_thread()
```
ioanachirca: I think now we can delete the unused `join_vcpu_thread` method.

src/vmm/src/lib.rs
```diff
@@ -438,14 +438,22 @@ impl Vmm {
...
         let vm = Vm::new(kvm.fd()).map_err(Error::Vm)?;
 
+        #[cfg(target_arch = "x86_64")]
+        let exit_evt = None;
+
+        #[cfg(target_arch = "aarch64")]
+        let exit_evt = EventFd::new(libc::EFD_NONBLOCK)
+            .map(Some)
+            .map_err(Error::EventFd)?;
+
```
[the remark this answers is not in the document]
kzys: You are right. It is calling `PortIODeviceManager::new()` right after that.

src/vmm/src/lib.rs
kzys: Thanks. Let me try.

src/vmm/src/vstate.rs
kzys: Thanks. Will do.

src/vmm/src/lib.rs
kzys: Addressed in dc7eeff

src/vmm/src/lib.rs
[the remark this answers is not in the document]
kzys: Addressed in dc7eeff.

src/vmm/src/vstate.rs
kzys: Addressed in 8ee01ae78c8935b6c29b4a9d67fe815024a68bc2

kzys:
The x86 build is failing, but it seems like not related to the changes.

> [Firecracker devtool] Error accessing Docker. Please make sure the Docker daemon is running and that you are part of the docker group. For more information, see https://docs.docker.com/install/linux/linux-postinstall/

kzys:
@andreeaflorescu That is true. `for _i in 0..10 {` is pretty arbitrary. How about changing that to `loop {`? pytest can kill the test if it takes forever.

src/vmm/src/vstate.rs
```diff
@@ -1245,13 +1267,17 @@ mod tests {
...
+        // Instead of waiting the completion of the thread, we need to poll the fd.
+        for _i in 0..10 {
+            match vcpu_exit_evt.read() {
+                Ok(x) => assert_eq!(x, 1),
+                Err(e) if e.kind() == io::ErrorKind::WouldBlock => {
+                    thread::sleep(Duration::from_millis(100))
+                }
+                Err(e) => panic!("failed to read the exit event: {}", e),
+            }
+        }
```
kzys: Using `queue_event_expect_response(&vcpu_handle, VcpuEvent::Exit, VcpuResponse::Paused)` is another way. It *looks* nicer than the for loop.

However `queue_event_expect_response()` is calling `recv_timeout(Duration::from_millis(100))`. So it doesn't address the concern actually.

andreeaflorescu:
@kzys My problem is with how we block the vcpu from exiting by doing a busy loop as follows:
```exited -> pause -> exited -> pause -> ... ``` until the exit from the vmm loop is actually called.

Conceptually it also looks a bit weird that we have a state machine where we run in a loop until `end_state` is true. With the current code `end_state` will never be set to true.

Instead of calling `exited` -> `pause` in a loop, we should find a better solution for avoid sigprocmask. I noticed that in the issue @acatangiu also mentioned using a barrier as an alternative. Is that something that we could do instead?

**Comment by kzys** [comment]
@andreeaflorescu I think we could. In production, the transition to `exit` calls `Barrier#wait()` to block indefinitely. In tests, the test code calls `Barrier#wait()` to unblock the thread, then read exit_evt.

**Comment by kzys** [comment]
It looks much better! Thanks for the suggestion. [The aarch64 build][https://firecracker-pr.s3.amazonaws.com/pr-logs/aarch64/01586/kzys-firecracker-579b1a9.log] is failing, but it seems unrelated to the changes I'm making.

I didn't squash the last commit (579b1a939ce1c3d43feeb5c94666a8dd5b90b954) into the first one (8ee01ae78c8935b6c29b4a9d67fe815024a68bc2) for making code review easier, but would do after review.

**Comment by serban300** [comment]
@kzys I would suggest separating this into 2 PRs. 

I would create:
- a PR containing only the `exit_evt` changes since it seems that there is no disagreement there and we could merge it easily.
- a different PR for fixing the `KVM_EXIT_SHUTDOWN`

This way the PR containing the fix for the `KVM_EXIT_SHUTDOWN` would be smaller, easier to review, and we could discuss it separately in case we want to spend more time analyzing it.

**Comment by kzys** [comment]
@serban300 Hmm, how did you reproduce the issue on your end? Just starting multiple Firecrackers to simulate heavy load?

Splitting the PR makes sense. Here is the one without KVM_EXIT_SHUTDOWN fix: #1598.

**Comment by serban300** [comment]
> @serban300 Hmm, how did you reproduce the issue on your end? Just starting multiple Firecrackers to simulate heavy load?
> 
> Splitting the PR makes sense. Here is the one without KVM_EXIT_SHUTDOWN fix: #1598.

@kzys Thanks for splitting the PR.

For reproducing the issue I just started a VM with 32 vCPUs and then issued a reboot command. I tried this a couple of times (max 5 times) and the issue reproduced once. I made sure that I was on your branch.

**Comment by serban300** [comment]
> > @serban300 Hmm, how did you reproduce the issue on your end? Just starting multiple Firecrackers to simulate heavy load?
> > Splitting the PR makes sense. Here is the one without KVM_EXIT_SHUTDOWN fix: #1598.
> 
> @kzys Thanks for splitting the PR.
> 
> For reproducing the issue I just started a VM with 32 vCPUs and then issued a reboot command. I tried this a couple of times (max 5 times) and the issue reproduced once. I made sure that I was on your branch.

Sorry @kzys I did the tests on an old binary. I can't reproduce the issue with your fix. But I would like to analyze the solution a bit.

**Comment by kzys** [comment]
@serban300 Good to hear :) Let me know if anything I can help. Currently this issue is blocking us to use Firecracker post-0.20.0 on firecracker-containerd.

**Comment by serban300** [comment]
@kzys I investigated the problem. I think we should change the way we handle the `KVM_SHUTDOWN` signal. Right now we treat it as an [`VcpuUnhandledKvmExit` error](https://github.com/firecracker-microvm/firecracker/blob/c9e900c181554b65b929d89ad576ecfee6ab1db6/src/vmm/src/vstate.rs#L641), in which case we end up [sending a signal to the VMM thread to exit with status 1](https://github.com/firecracker-microvm/firecracker/blob/c9e900c181554b65b929d89ad576ecfee6ab1db6/src/vmm/src/vstate.rs#L790). This is usually an error code, and doesn't seem fit for this use case.

Personally I would probably send a success code instead, pause the vCPU thread and wait for the VMM thread to kill the process. The barrier is also an option, but I wouldn't send an error status code anyway.

@firecracker-microvm/compute-capsule what do you think ?

**Comment by kzys** [comment]
@serban300 I don't have strong opinion on that. Regarding the exit code, doesn't it send `FC_EXIT_CODE_OK`?

https://github.com/firecracker-microvm/firecracker/blob/7fb2e4658a6763760161909f6526aed2f02a6c9b/src/vmm/src/lib.rs#L1270

**Comment by acatangiu** [comment]
I guess the cleanest solution here is to drop the `exited` state and just keep `running` and `paused`. This also keeps our logic simple and clear.

On errors or valid exit commands the `Vcpu`s will write to the `exit_evfd` to notify the `Vmm` then move to the `paused` state. The responsibility and control is thus passed to the `Vmm` thread.

**Comment by kzys** [comment]
@acatangiu 

Calling `StateMachine::finish(Self::paused)` instead of `StateMachine::finish(Self::exited)` is possible. We have some error cases that transition the state machine to `exited`. Do you want to change all of them to use `paused`?

What do you think about `Error::VcpuUnhandledKvmExit`? I agree it is not "unhandled", but technically all of the error cases are not "unhandled". So I don't have strong opinion on that and want to see on @firecracker-microvm/compute-capsule team's agreement regarding the approach first.

**Comment by serban300** [comment]
@kzys you're right, it exits with `FC_EXIT_CODE_OK`. I thought that it exits with the value that has been written in `exit_evt`. This still looks like a problem to me, because it also exits with `FC_EXIT_CODE_OK` in case a vCPU receives a `KVM_EXIT_INTERNAL_ERROR` for example.

**Comment by acatangiu** [comment]
> I guess the cleanest solution here is to drop the exited state and just keep running and paused. This also keeps our logic simple and clear.

I jumped the gun a bit here. It seems the `exited` state does have value. I propose the following mechanism:

`running` state --- encounters exit reason X (where X could be an exit command, an error, etc) ---> `return StateMachine::next(Self::exited)` ---> `exited` state.

`exited` state:
1. `vcpu.exit_evfd.write(1)`
2. use the existing communication channels to send to Vmm the exit reason: `VcpuResponse::Exited(X)`
3. block on waiting for another command from Vmm: `vcpu.event_receiver.recv()`
4. another command should never come, but even if it does, this state should return next state as `StateMachine::finish(Self::exited)` which will just stop the state machine.

This flow should solve all our requirements.

**Comment by serban300** [comment]
> > I guess the cleanest solution here is to drop the exited state and just keep running and paused. This also keeps our logic simple and clear.
> 
> I jumped the gun a bit here. It seems the `exited` state does have value. I propose the following mechanism:
> 
> `running` state --- encounters exit reason X (where X could be an exit command, an error, etc) ---> `return StateMachine::next(Self::exited)` ---> `exited` state.
> 
> `exited` state:
> 
>     1. `vcpu.exit_evfd.write(1)`
> 
>     2. use the existing communication channels to send to Vmm the exit reason: `VcpuResponse::Exited(X)`
> 
>     3. block on waiting for another command from Vmm: `vcpu.event_receiver.recv()`
> 
>     4. another command should never come, but even if it does, this state should return next state as `StateMachine::finish(Self::exited)` which will just stop the state machine.
> 
> 
> This flow should solve all our requirements.

This could work.

**Comment by serban300** [comment]
> > > I guess the cleanest solution here is to drop the exited state and just keep running and paused. This also keeps our logic simple and clear.
> > 
> > 
> > I jumped the gun a bit here. It seems the `exited` state does have value. I propose the following mechanism:
> > `running` state --- encounters exit reason X (where X could be an exit command, an error, etc) ---> `return StateMachine::next(Self::exited)` ---> `exited` state.
> > `exited` state:
> > ```
> > 1. `vcpu.exit_evfd.write(1)`
> > 
> > 2. use the existing communication channels to send to Vmm the exit reason: `VcpuResponse::Exited(X)`
> > 
> > 3. block on waiting for another command from Vmm: `vcpu.event_receiver.recv()`
> > 
> > 4. another command should never come, but even if it does, this state should return next state as `StateMachine::finish(Self::exited)` which will just stop the state machine.
> > ```
> > 
> > 
> > This flow should solve all our requirements.
> 
> This could work.

Although it might be easier to have separate eventfds for exit: one for `i8042` and one for all the vCPUs. When we receive data on `i8042_evt`, we exit with success. When we receive data on `vcpus_exit_evt` we read the exit code from it. 

Otherwise we have to query all the vCPUs for `VcpuResponse::Exited(X)` since we don't know which one issued the exit command.

**Comment by acatangiu** [comment]
I wouldn't use separate events for `i8042` and vcpu exits for a few reasons:

- on an `exit event` the `Vmm` needs to query all `Vcpu`s for the reason, but that query needs to be a `recv_timeout()` (a dozen ms or so) since not all `Vcpu`s might have exited. If none report any error exit reason, then process can end with `retcode 0`. This works nicely for the `i8042` path as well as the error path and kvm-exit-shutdown path.
- further down the line I'd like us to investigate why on newer kernels we get `KVM_EXIT_SHUTDOWN` from kvm while on older ones we don't. If we can make that work on all versions we can simply drop the `i8042` mechanism.

**Comment by serban300** [comment]
> I wouldn't use separate events for `i8042` and vcpu exits for a few reasons:
> 
>     * on an `exit event` the `Vmm` needs to query all `Vcpu`s for the reason, but that query needs to be a `recv_timeout()` (a dozen ms or so) since not all `Vcpu`s might have exited. If none report any error exit reason, then process can end with `retcode 0`. This works nicely for the `i8042` path as well as the error path and kvm-exit-shutdown path.
> 
>     * further down the line I'd like us to investigate why on newer kernels we get `KVM_EXIT_SHUTDOWN` from kvm while on older ones we don't. If we can make that work on all versions we can simply drop the `i8042` mechanism.

Ok, looks good to me. I think it's ok to use the same eventf for `i8042` and for the vCPUs exit signal.

I have only one comment. I wouldn't block on `vcpu.event_receiver.recv()` in the exited state. I wouldn't want to risk receiving another command. Who knows. I would do an infinite wait with something like:

```
        let barrier = Barrier::new(2);
        barrier.wait();
```

acatangiu:
Infinite wait should work fine.

> I wouldn't want to risk receiving another command. Who knows.

Even if another command is received, `exited` state should `return StateMachine::finish(Self::exited)` which will end the thread loop.

Whether we block forever or wait for a message which should never arrive is fine with me. The second option allows a bit more flexibility in case we ever want to actually join the thread (we control when it finishes and can be joined), but practically we should never need that so again, either option works for me.

serban300:
@kzys @acatangiu @andreeaflorescu I created a [draft](https://github.com/firecracker-microvm/firecracker/pull/1603) with a possible implementation of the solutions that we discussed. Please take a look !

kzys:
@serban300 Thanks! I've tested the draft PR on firecracker-containerd and seems it's working!

Let me close this pull request.

acatangiu:
Closed in favor of #1603

PR #1588, opened 2020-02-06
Fixes for the arg parser

lauralt:
## Reason for This PR

Some renames and small changes in order to have in `arg_parser` file only
the functionality of a command line argument parser.

## Description of Changes

App struct was renamed ArgParser and app specific fields
were removed in order to keep in `arg_parser.rs` only the
functionality that is related to a command line argument
parser.
ArgInfo struct was renamed Argument because it contains
the `user_value` field which is not actually an info about
the argument.

## License Acceptance

## PR Checklist

- [x] The reason for this PR is
      clearly provided.
- [x] Code-level documentation for touched
      code is included in this PR.

### DIFF
--- a/src/jailer/src/main.rs
+++ b/src/jailer/src/main.rs
@@ -19,7 +19,7 @@ use std::process;
 use std::result;
 
 use env::Env;
-use utils::arg_parser::{App, ArgInfo, Error as ParsingError};
+use utils::arg_parser::{ArgParser, Argument, Error as ParsingError};
 use utils::validators;
 
 const JAILER_VERSION: &str = env!("CARGO_PKG_VERSION");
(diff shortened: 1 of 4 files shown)


src/jailer/src/main.rs
```diff
@@ -322,17 +319,20 @@ fn main() {
             process::exit(1);
         }
         _ => {
-            if let Some(help) = app.arguments().value_as_bool("help") {
+            if let Some(help) = arg_parser.arguments().value_as_bool("help") {
                 if help {
-                    println!("{}", app.format_help());
+                    println!(
```
andreeaflorescu: Let's also add the banner here:
```println!("Firecracker v{}", FIRECRACKER_VERSION)```

File: src/utils/src/arg_parser.rs [hunk 2]
```diff
@@ -48,59 +48,36 @@ impl fmt::Display for Error {
...
 
     /// Parse the command line arguments.
-    pub fn parse_cmdline_args(&mut self) -> Result<()> {
+    pub fn parse_from_cmdline(&mut self) -> Result<()> {
         self.arguments.parse_from_cmdline()
     }
 
     /// Concatenate the `help` information of every possible argument
     /// in a message that represents the correct command line usage
     /// for the application.
```
andreeaflorescu: I would've expected the banner to be optional and not related to the ArgumentParser at all. For example, users could print the `header` before calling `formatted_help` if they want to.

In any case, if you want to keep it as is, please update the documentation.

lauralt: You're right, it makes more sense to be separated.

PR #1589, opened 2020-02-06
Raise coverage for vmm builder

iulianbarbu:
## Reason for This PR

#1578 

## Description of Changes

Raised the coverage for `builder.rs`.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -177,17 +177,21 @@ impl SerialStdin {
         SerialStdin(io::stdin())
     }
 }
+
 impl io::Read for SerialStdin {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
         self.0.read(buf)
     }
 }
+
 impl AsRawFd for SerialStdin {
     fn as_raw_fd(&self) -> RawFd {
         self.0.as_raw_fd()
     }
 }
+
 impl devices::legacy::ReadableFd for SerialStdin {}
+
 impl VmmEventsObserver for SerialStdin {
     fn on_vmm_boot(&mut self) -> std::result::Result<(), utils::errno::Error> {
         // Set raw mode for stdin.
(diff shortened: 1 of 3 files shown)


andreeaflorescu:
Update on this. Worked with @iulianbarbu and identified the code paths that are not tested as follows:
- MMIO
- VirtioDevice implementation for Net (some functions)
- BlockDeviceConfig & NetDeviceConfig
These will be tested in a subsequent PR.

We also looked at the builder functions that can be tested in a meaningful way and we agreed on the following:
- create_vcpus_*
- attach_*_devices

For the rest we can only have coverage tests and some of them actually come with the disadvantage that it makes it impossible to run the tests in parallel in the future.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,677 @@ pub mod tests {
...
 
+    #[cfg(target_arch = "x86_64")]
+    fn default_portio_device_manager() -> PortIODeviceManager {
         let read_tempfile = TempFile::new("/tmp/serial_read_").unwrap();
         let read_file = File::open(read_tempfile.as_path()).unwrap();
         let read_handle = SerialInput(read_file);
         let mut event_manager = EventManager::new().expect("Unable to create EventManager");
-        setup_serial_device(
+
+        let serial = setup_serial_device(
```
andreeaflorescu: It is a bit weird for the Event Manager to be created here and then not be accessible any more. It should be created outside of this function and passed as a parameter.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,677 @@ pub mod tests {
...
+        assert_eq!(
+            format!("{}", err),
+            format!(
+                "Invalid kernel command line: {}",
+                kernel::cmdline::Error::HasSpace.to_string()
+            )
+        );
+    }
+
+    // TODO: implement the test after vsock refactoring PR is merged
```
andreeaflorescu: `build_microvm` should not be tested in unit tests because it is the equivalent of `start_microvm`. The problem with testing this in a unit test is that we do not have a way to properly clean up the vmm resources which means that the KVM structures are going to be allocated for the duration of the unit tests, which might trigger ENOMEM errors when running other kvm related tests.

src/vmm/src/builder.rs
iulianbarbu: I agree. I've changed that. Take a look please.

src/vmm/src/builder.rs
iulianbarbu: Yep. It is a left over.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,659 @@ pub mod tests {
...
+    ) -> crate::Result<Vec<Vcpu>> {
         let mut event_manager = EventManager::new().expect("Unable to create EventManager");
-        setup_serial_device(
-            &mut event_manager,
-            Box::new(read_handle),
-            Box::new(io::stdout()),
+        let guest_memory = create_guest_memory(128).unwrap();
+        let mut vm = setup_kvm_vm(&guest_memory).unwrap();
+        setup_interrupt_controller(&mut vm).unwrap();
+        let vcpu_config_t2_temp = VcpuConfig {
```
lauralt: Can we rename this to `vcpu_config` or something that doesn't contain `t2`?

iulianbarbu: Yes. It's a leftover.

acatangiu:
> Update on this. Worked with @iulianbarbu and identified the code paths that are not tested as follows:
> 
>     * MMIO
> 
>     * VirtioDevice implementation for Net (some functions)
> 
>     * BlockDeviceConfig & NetDeviceConfig
>       These will be tested in a subsequent PR.
> 
> 
> We also looked at the builder functions that can be tested in a meaningful way and we agreed on the following:
> 
>     * create_vcpus_*
> 
>     * attach_*_devices
> 
> 
> For the rest we can only have coverage tests and some of them actually come with the disadvantage that it makes it impossible to run the tests in parallel in the future.

I somewhat disagree here. The idea of unit-testing is to test each individual unit. Start from the smallest unit and test other units above with the assumptions that the units below are also tested and function correctly.
Therefore I believe testing the individual functions and that they return `Ok()` when that's expected is fine and is not false coverage - **as long as we also independently test the underlying units**.

src/vmm/src/builder.rs
acatangiu: Everything should be cleaned up nicely, I don't think we have a problem with that anymore. But I agree this function is a complex recipe that is hard to unit-test properly and calling it in tests provides false coverage.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,657 @@ pub mod tests {
...
+            assert!(vmm
+                .mmio_device_manager
+                .get_device(DeviceType::Virtio(TYPE_BLOCK), "secondary")
+                .is_some());
+            // Check if these two block devices are inserted in kernel_cmdline.
+            assert!(vmm
+                .kernel_cmdline
+                .as_str()
+                .contains("virtio_mmio.device=4K@0xd0000000:5 virtio_mmio.device=4K@0xd0001000:6"));
+        }
```
acatangiu: Duplicate block of code.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,657 @@ pub mod tests {
...
+
+    #[test]
+    fn test_stdin_wrapper() {
+        let wrapper = SerialStdin(io::stdin());
+        assert_eq!(wrapper.as_raw_fd(), io::stdin().as_raw_fd())
+    }
+
+    #[test]
+    #[cfg(target_arch = "x86_64")]
+    fn test_create_vcpus_x86_with_t2_template() {
```
acatangiu: I don't think we should test the different templates in this file. The tests here should just verify the correctness of the `create_vcpus_x86_64()` function not also verify the correctness of the functions it calls.

To that purpose, I believe it should be enough to just verify a valid `Vec<Vcpu>` of the right length is returned.

iulianbarbu:
I think I've addressed your feedback! @lauralt , @acatangiu , @andreeaflorescu, please take a look again.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,568 @@ pub mod tests {
...
+        let mut vm = setup_kvm_vm(&guest_memory).unwrap();
+        setup_interrupt_controller(&mut vm).unwrap();
+        let vcpu_config = VcpuConfig {
+            vcpu_count,
+            ht_enabled: false,
+            cpu_template: None,
+        };
+        let boot_config = default_boot_config();
+        let entry_addr = load_kernel(&boot_config, &guest_memory).unwrap();
+        let pio_dev_manager = default_portio_device_manager(&mut event_manager);
```
acatangiu: You don't need to instantiate a `pio_dev_manager` here. You just need a `devices::Bus::new()` when creating vcpus.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,568 @@ pub mod tests {
...
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    fn default_portio_device_manager(event_manager: &mut EventManager) -> PortIODeviceManager {
         let read_tempfile = TempFile::new("/tmp/serial_read_").unwrap();
         let read_file = File::open(read_tempfile.as_path()).unwrap();
         let read_handle = SerialInput(read_file);
+
+        let serial =
+            setup_serial_device(event_manager, Box::new(read_handle), Box::new(io::stdout()))
```
acatangiu: I would also keep the original test for setting up the serial device.

With this code, it is still tested, but if we were to introduce a bug there, instead of explicitly failing a `test_serial_setup` test we'd only fail tests like `test_attach_net_devices` and `test_attach_block_devices`.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,568 @@ pub mod tests {
...
+
+        let exit_evt = EventFd::new(libc::EFD_NONBLOCK)
+            .map_err(Error::EventFd)
+            .map_err(StartMicrovmError::Internal)
+            .unwrap();
+
+        let vm = setup_kvm_vm(&guest_memory).unwrap();
+        let mmio_device_manager = default_mmio_device_manager();
+        #[cfg(target_arch = "x86_64")]
+        let pio_device_manager = default_portio_device_manager(event_manager);
```
acatangiu: We don't need a functioning `serial` for tests just like we don't need a functioning `i8042`.

This can be replaced by:
```rust
PortIODeviceManager::new(
    Arc::new(Mutex::new(Serial::new_sink(EventFd::new(libc::EFD_NONBLOCK).unwrap()))),
    EventFd::new(libc::EFD_NONBLOCK).unwrap()
).unwrap()
```

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,568 @@ pub mod tests {
...
+            mmio_device_manager,
+            #[cfg(target_arch = "x86_64")]
+            pio_device_manager,
+            seccomp_level: SECCOMP_LEVEL_ADVANCED,
+        }
+    }
+
+    #[test]
+    fn test_stdin_wrapper() {
+        let wrapper = SerialStdin(io::stdin());
```
acatangiu: This is a miss on the original test I added, you can improve it here:

```suggestion
        let wrapper = SerialStdin::get();
```

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,568 @@ pub mod tests {
...
+        let vcpu_count = 2;
+
+        let vcpu_config = VcpuConfig {
+            vcpu_count,
+            ht_enabled: false,
+            cpu_template: None,
+        };
+
+        let boot_config = default_boot_config();
+        let entry_addr = load_kernel(&boot_config, &guest_memory).unwrap();
```
acatangiu: Same here, you don't need to actually load a kernel for this test, just pass some random `entry_addr`. The vcpu will never be booted in this test anyway.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,568 @@ pub mod tests {
...
+        let guest_memory = create_guest_memory(128).unwrap();
+        let mut vm = setup_kvm_vm(&guest_memory).unwrap();
+        setup_interrupt_controller(&mut vm).unwrap();
+        let vcpu_config = VcpuConfig {
+            vcpu_count,
+            ht_enabled: false,
+            cpu_template: None,
+        };
+        let boot_config = default_boot_config();
+        let entry_addr = load_kernel(&boot_config, &guest_memory).unwrap();
```
acatangiu: You don't need to actually load a kernel for this test, just pass some random `entry_addr`. The vcpu will never be booted in this test anyway.

src/devices/src/virtio/mod.rs
```diff
@@ -60,6 +60,8 @@ pub const VIRTIO_MMIO_INT_CONFIG: u32 = 0x02;
 /// queue events.
 pub const NOTIFY_REG_OFFSET: u32 = 0x50;
 
+pub const QUEUE_SIZE: u16 = 256;
```
andreeaflorescu: This change doesn't seem related to the builder tests.

The queue size is also a device specific configuration, so I am not super sure we want to have it globally used even though it's the same value everywhere.

iulianbarbu: I thought because it is the same, we can make it common to the block and net device. However, I understand your concern, and I will get rid of these changes.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,558 @@ pub mod tests {
...
+            let res = attach_block_devices(&mut vmm, &block_dev_configs, &mut event_manager);
+            assert!(res.is_ok());
+            assert!(vmm.kernel_cmdline.as_str().contains("root=/dev/vda ro"));
+            assert!(vmm
+                .mmio_device_manager
+                .get_device(DeviceType::Virtio(TYPE_BLOCK), "root")
+                .is_some());
+        }
+
+        // Use Case 2: Root Block Device is specified through PARTUUID.
```
lauralt: Nit: Can you please add a short use case similar to this one, but with `is_read_only` field `true` to cover one more line (line 635) :D? I tried to change the other use cases but it resulted in other lines uncovered :(.

iulianbarbu: Addressed this line. @lauralt , take a look again, please.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,543 @@ pub mod tests {
...
+            exit_evt,
+            vm,
+            mmio_device_manager,
+            #[cfg(target_arch = "x86_64")]
+            pio_device_manager,
+            seccomp_level: SECCOMP_LEVEL_ADVANCED,
+        }
+    }
+
+    fn helper_test_attach_block_device(event_manager: &mut EventManager, custom_block_cfgs: Vec<CustomBlockConfig>) -> (Vmm, Vec<TempFile>) {
```
lauralt: It looks like we never use this helper function for the `block_files` member of the returned tuple. Could we return only the vmm and maybe rename this function to something like `vmm_with_block_devices` or something like this :-??.
Also, why do we need to create `block_files` vec in this function?

iulianbarbu: I will rename the function with your suggestion. The `Vec` is created because `TempFile`s are created inside the for loop scope. Once we get out of that scope, they are destroyed, making `attach_block_devices` return error. Keeping them inside a `Vec`, will make them available for the `attach_block_devices` call, and returning them from `helper..` will pass them to the outer scope of the function call, besides the vmm object. We can dispose them inside `helper_`, and not return them at all, but that will make the vm block devices point to deleted files. This is not a big deal, in the context of these unit tests, since we are not testing block functionality.

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,572 @@ pub mod tests {
...
+        assert_eq!(
+            format!("{}", err),
+            format!(
+                "Invalid kernel command line: {}",
+                kernel::cmdline::Error::HasSpace.to_string()
+            )
+        );
     }
+
+    // TODO: implement the test after vsock refactoring PR is merged
```
andreeaflorescu: Should we add them before merging this since vsock is merged?

src/vmm/src/builder.rs
```diff
@@ -763,39 +767,605 @@ pub mod tests {
...
+
+        impl TempSockFile {
+            fn new(tmp_file: TempFile) -> Self {
+                TempSockFile {
+                    path: String::from(
+                        fs::canonicalize(tmp_file.as_path())
+                            .unwrap()
+                            .to_str()
+                            .unwrap(),
+                    ),
```
lauralt: Hmm, is the `canonicalize()` needed or can we simplify this with: `String::from(tmp_file.as_path().to_str().unwrap())`?

src/vmm/src/builder.rs
iulianbarbu: Done, please take a look.

src/vmm/src/builder.rs
iulianbarbu: I wanted to take the absolute path of the created `TempFile`. The method you propose will return the file name for a specific `TempFile`. We can use the file name, though, but it will cause problems when removing the file, at the drop time, because it will try to remove the file from the current path of test execution.

lauralt: I may be wrong, but `as_path()` doesn't return the absolute path to a file? :-?

iulianbarbu: Actually, this is not true. `TempFile::as_path` returns a `Path` which can be relative or absolute, depending on how the `TempFile` was created, with an absolute path or a relative one. I will update as you suggested.

PR #1603, opened 2020-02-13
Change vCPU threads exit behavior in order to avoid sigprocmask

### LINKED ISSUE #1456: Occasional KVM_EXIT_SHUTDOWN and bad syscall (14) during VM shutdown

serban300:
## Reason for This PR

fixes #1456 

## Description of Changes

Don't allow vCPU threads to exit. If they finished execution, they will signal the VMM to exit and then wait forever.

## License Acceptance

## PR Checklist

- [x] Either this PR is linked to an issue, or, the reason for this PR is
      clearly provided.
      code is included in this PR.

### DIFF
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -1271,7 +1271,21 @@ impl Vmm {
             match self.epoll_context.dispatch_table[event.data as usize] {
                 Some(EpollDispatch::Exit) => {
                     self.exit_evt.read().map_err(Error::EventFd)?;
-                    self.stop(i32::from(FC_EXIT_CODE_OK));
+
+                    // Query each vcpu for the exit_code.
+                    // If the exit_code can't be found on any vcpu, it means that the exit signal
+                    // has been issued by the i8042 controller in which case we exit with
+                    // FC_EXIT_CODE_OK.
+                    let exit_code = self
+                        .vcpus_handles
+                        .iter()
+                        .find_map(|handle| match handle.response_receiver().try_recv() {
+                            Ok(VcpuResponse::Exited(exit_code)) => Some(exit_code),
+                            _ => None,
+                        })
+                        .unwrap_or(FC_EXIT_CODE_OK);
+
+                    self.stop(i32::from(exit_code));
                 }
                 Some(EpollDispatch::Stdin) => {
                     let mut out = [0u8; 64];
--- a/src/vmm/src/vstate.rs
+++ b/src/vmm/src/vstate.rs
@@ -14,7 +14,7 @@ use std::sync::mpsc::{channel, Receiver, Sender, TryRecvError};
 use std::thread;
 
 use super::TimestampUs;
-use super::FC_EXIT_CODE_GENERIC_ERROR;
+use super::{FC_EXIT_CODE_GENERIC_ERROR, FC_EXIT_CODE_OK};
 #[cfg(target_arch = "aarch64")]
 use arch::aarch64::gic::GICDevice;
 #[cfg(target_arch = "x86_64")]
@@ -926,11 +926,11 @@ impl Vcpu {
                 }
                 VcpuExit::Hlt => {
                     info!("Received KVM_EXIT_HLT signal");
-                    Err(Error::VcpuUnhandledKvmExit)
+                    Ok(VcpuEmulation::Stopped)
                 }
                 VcpuExit::Shutdown => {
                     info!("Received KVM_EXIT_SHUTDOWN signal");
-                    Err(Error::VcpuUnhandledKvmExit)
+                    Ok(VcpuEmulation::Stopped)
                 }
                 // Documentation specifies that below kvm exits are considered
                 // errors.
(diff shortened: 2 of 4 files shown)


src/vmm/src/lib.rs
```diff
@@ -1270,7 +1270,26 @@ impl Vmm {
...
+                    // If the exit_code can't be found on any vcpu, it means that the exit signal
+                    // has been issued by the i8042 controller in which case we exit with
+                    // FC_EXIT_CODE_OK.
+                    let exit_code = self
+                        .vcpus_handles
+                        .iter()
+                        .find_map(|handle| {
+                            match handle
+                                .response_receiver()
+                                .recv_timeout(Duration::from_millis(1))
```
acatangiu: I think we could spare a few more milliseconds here :smile:

src/vmm/src/vstate.rs
```diff
@@ -774,18 +775,33 @@ impl Vcpu {
...
-                StateMachine::next(Self::exited)
+                self.exit(FC_EXIT_CODE_GENERIC_ERROR)
             }
         }
     }
 
-    // This is the main loop of the `Exited` state.
-    fn exited(&mut self) -> StateMachine<Self> {
+    // Transition to the exited state
+    fn exit(&mut self, exit_code: u8) -> StateMachine<Self> {
```
acatangiu: nit: why is this a separate function and not just the `exited` state?

src/vmm/src/vstate.rs
```diff
@@ -718,7 +719,7 @@ impl Vcpu {
                 // Emulation was interrupted, check external events.
                 Ok(VcpuEmulation::Interrupted) => break,
                 // Emulation errors lead to vCPU exit.
-                Err(_) => return StateMachine::next(Self::exited),
+                Err(_) => return self.exit(FC_EXIT_CODE_GENERIC_ERROR),
```
acatangiu: In this case, you should change `run_emulation()` to not mark `KVM_EXIT_SHUTDOWN` as an error.
In case of `KVM_EXIT_SHUTDOWN`, we want `FC_EXIT_CODE_OK`.

src/vmm/src/lib.rs
kzys: +1 We actually hit this issue under heavy load due to testing (launching multiple VMs). 10d1f4aa59da50f8d39db9692844740dbacde5e6

acatangiu: This is an error case and only impacts the return code. We can't be extra generous either and optimize for the error case because then we'd be hurting the common good case by delaying the actual process end by `num_vcpus * timeout`.

serban300: We shouldn't need more than 1 since the vCPU first writes the response and then it writes to `exit_evt`. But I can increase this number. How many would you suggest ?

src/vmm/src/vstate.rs
serban300: Yes. I did this in the last commit.

src/vmm/src/vstate.rs
serban300: Because it needs to receive the `exit_code` as a parameter. From what I understand a `StateFn` can't receive parameters.

src/vmm/src/lib.rs
acatangiu: If `KVM_EXIT_SHUTDOWN` is not returned (as we've seen on certain cpu/kernel-version combinations) we're only relying on the `i8042` exit event which triggers the `Vmm` exit code without having any `Vcpu` write any exit reason. It's a valid case explained by your comment.

acatangiu: @serban300 is right, no need for bigger timeout.

1. in case of vcpu exit/error: `Vcpu` populates the channel **then** hits the event-fd, so by the time vmm picks up the event the channel data is guaranteed to be there.
2. in case of `i8042`: `Vcpu` doesn't provide data on the channel and Vmm will simply find it empty.

In all cases we don't need to wait.

nit: replace `.recv_timeout()` with `.try_wait()`.

serban300: Fixed.

kzys:
firecracker-containerd's integration tests are passing with this change!

https://github.com/firecracker-microvm/firecracker-containerd/pull/383

acatangiu:
Actually a `CHANGELOG::#Fixed` entry would also be in order.

serban300:
> Actually a `CHANGELOG::#Fixed` entry would also be in order.

Makes sense. Done.

src/vmm/src/vstate.rs
serban300: Personally I don't like it too much because exit_code is not generic state for the vCPU. It's just a code that has to be use once, right before exit. It would be really nice if we could send parameters to the state functions. Since it's not possible for the moment, I personally prefer to have the intermediate function `exit()` that receives the exit_code parameter.

PR #1606, opened 2020-02-13
tests: Reduce execution time of Python style check

gc-plp:
## Reason for This PR

The Python style check test was doing all the operations by serializing commands, which took a lot of time.

This PR runs all commands using the async module, reducing execution time from 40s to 20s on a typical desktop.

## Description of Changes

Run all commands asynchronously and take into account all executed commands including their stdout, stderr and return code.

## License Acceptance

## PR Checklist

- [x] No docs need to be updated as part of this PR
- [x] No API changes are included in this PR
- [x] The changes in this PR have no user impact
- [x] No new `unsafe` code has been added


andreeaflorescu:
Should we just remove some of the linters?

tests/framework/utils.py
```diff
@@ -30,3 +34,94 @@ def search_output_from_cmd(cmd: str,
 
     raise RuntimeError("Could not find '%s' in output for '%s'" %
                        (find_regex.pattern, cmd))
+
+
+def get_files_from(find_path: str,
```
aghecenco: nit: either rename the function or add a flag to suggest that it traverses `find_path` recursively

tests/framework/utils.py
```diff
@@ -30,3 +34,94 @@ def search_output_from_cmd(cmd: str,
...
+def run_cmd_list_async(cmd_list):
+    """
+    Run a list of commands asynchronously and wait for them to finish.
+
+    :param cmd_list: list of commands to execute
+    :return: None
+    """
+    loop = asyncio.get_event_loop()
+
+    cmds = []
```
aghecenco: nit
```suggestion
    cmds = [run_cmd_async(cmd) for cmd in cmd_list]
```

gc-plp:
We could, but we need to make sure that we don't lose the current things that we check for. 
As far as I've seen, we actually need to check for more things, such as alphabetically sorted imports, which can be done by adding flake8.

Either way, it's something that needs to be looked into, but it should be done in a different PR.

The point of this PR is to reduce test duration and add more clarity to commands executed in separate shells.

tests/integration_tests/build/test_style.py
```diff
@@ -32,57 +33,78 @@ def test_rust_style():
...
+    """
+    base_cmd = \
+        r'python3 -m pylint --jobs=0 --persistent=no --score=no ' \
+        r'--output-format=colorized --attr-rgx="[a-z_][a-z0-9_]{1,30}$" ' \
+        r'--argument-rgx="[a-z_][a-z0-9_]{1,30}$" ' \
+        r'--variable-rgx="[a-z_][a-z0-9_]{1,30}$" --disable=' \
+        r'bad-continuation,fixme,too-many-instance-attributes,' \
+        r'too-many-locals,too-many-arguments'
+
+    cmd_list = []
```
aghecenco: nit: can be replaced with list comprehension, same as above

tests/integration_tests/build/test_style.py
```diff
@@ -32,57 +33,78 @@ def test_rust_style():
...
+def gen_pydocstyle_cmds(file_list):
+    """
+    Return a list of pydocstyle commands that should be executed for each file.
+
+    :param file_list: list of files to check
+    :return: list of commands
+    """
+    base_cmd = "python3 -m pydocstyle --explain --source"
+
+    cmd_list = []
```
aghecenco: nit: can be replaced with list comprehension, same as above

tests/integration_tests/build/test_style.py
```diff
@@ -32,57 +33,78 @@ def test_rust_style():
...
+    python_files = utils.get_files_from(
+        find_path="..",
+        pattern="*.py",
+        exclude_names=["build"])
+
+    # Assert if somehow no python files were found
+    assert len(python_files) != 0
+
+    # Get all the linter commands that we should run
+    linter_cmds = []
```
aghecenco: nit: the list can be constructed in one go without successive `.extend`s

tests/pytest.ini
```diff
@@ -11,3 +11,9 @@ timeout = 300
 ; Set the cache dir location to our build dir, so we don't litter the source
 ; tree.
 cache_dir = ../build/pytest_cache
+
+; Set logger format and level
+log_level = DEBUG
+log_format = %(asctime)s %(name)s: %(levelname)s %(message)s
+; Uncomment the line below to get the live logger output
+;log_cli = true
```
aghecenco: nit: add newline at eof

tests/framework/utils.py
```diff
@@ -30,3 +34,94 @@ def search_output_from_cmd(cmd: str,
...
+                   exclude_names: list = None):
+    """
+    Return a list of files from a given path, recursively.
+
+    :param find_path: path where to look for files
+    :param pattern: what pattern to apply to file names
+    :param exclude_names: list of folder names to exclude
+    :return: list of found files
+    """
+    found = []
```
dianpopa: In order to avoid using `os` related functions mixed with `glob`, we could use an include/exclude approach based only on glob, something like:
```
    all = set((glob.glob(f"{find_path}/**/{pattern}",
                        recursive=True)))
    found = all
    for exclude in exclude_names:

        exclude_set = set(glob.glob(f"{find_path}/{exclude}/**/{pattern}",
                        recursive=True))
        found = found - exclude_set
    return found
```

tests/integration_tests/build/test_style.py
```diff
@@ -32,57 +33,78 @@ def test_rust_style():
...
+    """
+    Return a list of pydocstyle commands that should be executed for each file.
+
+    :param file_list: list of files to check
+    :return: list of commands
+    """
+    base_cmd = "python3 -m pydocstyle --explain --source"
+
+    cmd_list = []
+    for check_file in file_list:
```
dianpopa: I like more the approach with one function for all linters since we would avoid calling this `for` in each such function. Something like:
```
fn gen_linters_cmd()
cmd_for_linter1
cmd_for_linter2
for check_file in file_list:
   cmd_list.append(f"{cmd_for_linter1} {check_file}")
   cmd_list.append(f"{cmd_for_linter2} {check_file}")
```

tests/framework/utils.py
gc-plp: Is there a reason why we would like to avoid using `os` functions?

tests/integration_tests/build/test_style.py
gc-plp: The problem is that not all linters check the project through similar commands. 
Some need the file name, while others only need a path, which would not make the code very clear overall.

tests/framework/utils.py
dianpopa: No, I was saying that we can avoid **mixing** glob and os functions, but no strong reason against them. THis is more of a nit, anyways.

tests/integration_tests/build/test_style.py
dianpopa: Indeed, I think you are referring to pycodestyle, right? Actually, seems like providing the name of the file (just like we do for pydocstyle and pylint) works for pycodestyle too. Right now we do:
```
python3 -m pycodestyle --show-pep8 --show-source '
            '--exclude=../build ../
```
But I tried
``python3 -m pycodestyle --show-pep8 --show-source ../tests/integration_tests/functional/test_api.py 
```
and it does work, so we could do that for pycodestyle too. 
This is just a suggestion. Changes look good overall.

gc-plp: Fair enough, didn't check if this works. Will update it and put everything into one function.

Issue #1632, opened 2020-02-26
Args: `--version/-V` was missing


Pennyzct:
Hi guys
I've found that in new release Firecracker `v0.21.0`, the command line arg `--version/-V` is missing.
```
$ ./firecracker-v0.21.0-x86_64 --version
2020-02-26T03:15:13.135125190 [anonymous-instance:ERROR:src/firecracker/src/main.rs:132] Arguments parsing error: Found argument 'version' which wasn't expected, or isn't valid in this context.
```
Checking out the existing flags through `--help`, see `--version/-V` is also not there. But `--help` will 
display the version info in the first line. 
```
Firecracker v0.21.0

--api-sock <api-sock>: Path to unix domain socket used by the API.
--id <id>: MicroVM unique identifier.
--help: Show the help message.
--no-api: Optional parameter which allows starting and using a microVM without an active API socket.
--start-time-us <start-time-us>
--seccomp-level <seccomp-level>: Level of seccomp filtering that will be passed to executed path as argument.

                        - Level 0: No filtering.

                        - Level 1: Seccomp filtering by syscall number.

                        - Level 2: Seccomp filtering by syscall number and argument values.

```
So `--version/-V` has already been deprecated in the new release??

Pennyzct:
I've found that in PR https://github.com/firecracker-microvm/firecracker/pull/1360, we've replaced clap with our own args parser, and maybe forgot to add the dealing with `--version/-V`.

dianpopa:
Solved by patch release #1654.

PR #1360, opened 2019-10-29
Replace clap with our own args parser

lauralt:
Signed-off-by: Laura Loghin <lauralg@amazon.com>

## Reason for This PR

Get rid of `clap` crate dependencies and reduce this way the binary size.

## Description of Changes

Implements command line arguments parsing and validating and replaces `clap` crate functionality.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria. Where there are two options, keep one.]`  
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [x] All commits in this PR are signed (`git commit -s`).
- [x] The reason for this PR is clearly provided. 
- [x] The description of changes is clear and encompassing.
- [x] No docs need to be updated as part of this PR.
- [x] Code-level documentation for touched code is included in this PR.
- [x] No API changes are included in this PR.
- [x] The changes in this PR have no user impact.

PR #1633, opened 2020-02-26
Args: Add flag `--version`

Pennyzct:
## Reason for This PR
#1632 

## Description of Changes
In new release Firecracker `v0.21.0`, the command line arg `--version` is missing.
```
$ ./firecracker-v0.21.0-x86_64 --version
2020-02-26T03:15:13.135125190 [anonymous-instance:ERROR:src/firecracker/src/main.rs:132] Arguments parsing error: Found argument 'version' which wasn't expected, or isn't valid in this context.
```
And the `--version` is quite an universal command line arg to print application version number.

## License Acceptance

## PR Checklist

      clearly provided.
- [ ] Either no docs need to be updated as part of this PR, or, the required
- [ ] Either no code has been touched, or, code-level documentation for touched
      code is included in this PR.
- [ ] Either no API changes are included in this PR, or, the API changes are
- [ ] Either the changes in this PR have no user impact, or, the changes in
- [ ] Either no new `unsafe` code has been added, or, the newly added `unsafe`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,10 @@
 - Added `app_name` field in InstanceInfo struct for storing the
   application name.
 
-## Changed
+### Fixed
+- Added `--version` flag to both Firecracker and Jailer.
+
+### Changed
 - Updated CVE-2019-3016 mitigation information in
   [Production Host Setup](docs/prod-host-setup.md)
 - In case of using an invalid JSON as a 'config-file' for Firecracker,
--- a/src/firecracker/src/main.rs
+++ b/src/firecracker/src/main.rs
@@ -144,6 +144,14 @@ fn main() {
                     process::exit(i32::from(vmm::FC_EXIT_CODE_OK));
                 }
             }
+
+            if let Some(version) = arg_parser.arguments().value_as_bool("version") {
+                if version {
+                    println!("Firecracker v{}\n", FIRECRACKER_VERSION);
+                    process::exit(i32::from(vmm::FC_EXIT_CODE_OK));
+                }
+            }
+
             arg_parser.arguments()
         }
     };
--- a/src/utils/src/arg_parser.rs
+++ b/src/utils/src/arg_parser.rs
@@ -11,6 +11,7 @@ pub type Result<T> = result::Result<T, Error>;
 const ARG_PREFIX: &str = "--";
 const ARG_SEPARATOR: &str = "--";
 const HELP_ARG: &str = "--help";
+const VERSION_ARG: &str = "--version";
 
 /// Errors associated with parsing and validating arguments.
 #[derive(Debug, PartialEq)]
@@ -267,6 +268,16 @@ impl<'a> Arguments<'a> {
             return Ok(());
         }
 
+        // If `--version` is provided as a parameter, we artificially skip the parsing of other
+        // command line arguments by adding just the version argument to the parsed list and
+        // returning.
+        if args.contains(&VERSION_ARG.to_string()) {
+            let mut version_arg = Argument::new("version").help("Print the binary version number.");
+            version_arg.user_value = Some(Value::Bool(true));
+            self.insert_arg(version_arg);
+            return Ok(());
+        }
+
         // Otherwise, we continue the parsing of the other arguments.
         self.populate_args(args)
     }
@@ -455,9 +466,10 @@ mod tests {
     #[test]
     fn test_parse() {
         let arg_parser = build_arg_parser();
-        let mut arguments = arg_parser.arguments().clone();
 
         // Test different scenarios for the command line arguments provided by user.
+        let mut arguments = arg_parser.arguments().clone();
+
         let args = vec!["binary-name", "--exec-file", "foo", "--help"]
             .into_iter()
             .map(String::from)
@@ -466,7 +478,16 @@ mod tests {
         assert!(arguments.parse(&args).is_ok());
         assert!(arguments.args.contains_key("help"));
 
-        let arg_parser = build_arg_parser();
+        arguments = arg_parser.arguments().clone();
+
+        let args = vec!["binary-name", "--exec-file", "foo", "--version"]
+            .into_iter()
+            .map(String::from)
+            .collect::<Vec<String>>();
+
+        assert!(arguments.parse(&args).is_ok());
+        assert!(arguments.args.contains_key("version"));
+
         arguments = arg_parser.arguments().clone();
 
         let args = vec![
@@ -486,7 +507,6 @@ mod tests {
             Err(Error::MissingValue("api-sock".to_string()))
         );
 
-        let arg_parser = build_arg_parser();
         arguments = arg_parser.arguments().clone();
 
         let args = vec![
@@ -507,7 +527,6 @@ mod tests {
             Err(Error::DuplicateArgument("api-sock".to_string()))
         );
 
-        let arg_parser = build_arg_parser();
         arguments = arg_parser.arguments().clone();
 
         let args = vec!["binary-name", "--api-sock", "foo"]
@@ -520,7 +539,6 @@ mod tests {
             Err(Error::MissingArgument("exec-file".to_string()))
         );
 
-        let arg_parser = build_arg_parser();
         arguments = arg_parser.arguments().clone();
 
         let args = vec![
@@ -540,7 +558,6 @@ mod tests {
             Err(Error::UnexpectedArgument("invalid-arg".to_string()))
         );
 
-        let arg_parser = build_arg_parser();
         arguments = arg_parser.arguments().clone();
 
         let args = vec![
@@ -562,7 +579,6 @@ mod tests {
             Err(Error::MissingArgument("config-file".to_string()))
         );
 
-        let arg_parser = build_arg_parser();
         arguments = arg_parser.arguments().clone();
 
         let args = vec![
@@ -582,7 +598,6 @@ mod tests {
             Err(Error::MissingValue("id".to_string()))
         );
 
-        let arg_parser = build_arg_parser();
         arguments = arg_parser.arguments().clone();
 
         let args = vec![
@@ -603,7 +618,6 @@ mod tests {
             Err(Error::UnexpectedArgument("foobar".to_string()))
         );
 
-        let arg_parser = build_arg_parser();
         arguments = arg_parser.arguments().clone();
 
         let args = vec![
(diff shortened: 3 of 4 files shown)


KarthikNedunchezhiyan:
@Pennyzct how will this combine with jailer (for jailer users). I think there will be some inconsistency if we handled version param only in firecracker without jailer 🙂 WDYT?

Pennyzct:
Hi @KarthikNedunchezhiyan 
thanks for notice. 😊
jailer is also using the new args parser(PR #1360) and `--version` is also missing for jailer binary. 
```
$ ./jailer-v0.21.0-x86_64 --version
Arguments parsing error: Found argument 'version' which wasn't expected, or isn't valid in this context.

```
And for jailer user wants to print firecracker version number, this flag should be specified after `--` (end of command options) in order to be forwarded to firecracker process. ;)

KarthikNedunchezhiyan:
> for jailer user wants to print firecracker version number, this flag should be specified after -- (end of command options) in order to be forwarded to firecracker process. ;)

@Pennyzct Yes we can do like this but we need to provide all arguments with valid values to jailer. also, the jailer will create the root folder and cgroups just to know the version of firecracker 🙂.

lauralt:
Hello @Pennyzct, thanks for the PR, we didn't, indeed, add support for `--version` flag in the new arg parser implementation. I took a look at your commit and I think `--version` argument should be treated like `--help`, i.e. check directly for this argument in `arg_parser.rs` in `parse()` function and not insert it when creating the ArgParser object in main. As for the jailer, we should also add support for this flag because jailer can be seen as another application and user may want to see the version for it too (`jailer 0.21.0` for example). So basically, just add a check for this argument in `main()` function from `jailer`. What do you think?
Please, ping me if you need any help with the PR or if something is not clear.

Pennyzct:
Hi~ @lauralt thanks. ;)
You're right~~~It should be treated like `--help`, then it will also fix what @KarthikNedunchezhiyan worried~~~
I'll update asap for you guys to review~~~

KarthikNedunchezhiyan:
@lauralt personally I will not suggest keeping jailer version (`jailer 0.21.0`) as separate from the firecracker version (`firecracker 0.21.0`). It will lead to introducing a compatibility table between firecracker and jailer (just my opinion) 🙂. WDYT?

Pennyzct:
Hi~ @lauralt 
updated~~~ plz take a review. ;). and do you think it should be back-ported to `v0.21.0`?

lauralt:
> @lauralt personally I will not suggest keeping jailer version (`jailer 0.21.0`) as separate from the firecracker version (`firecracker 0.21.0`). It will lead to introducing a compatibility table between firecracker and jailer (just my opinion) . WDYT?

In my opinion, it makes sense to introduce this flag also for the jailer as it is another binary and the version for it is taken from a different `Cargo.toml` file.

iulianbarbu:
> @lauralt personally I will not suggest keeping jailer version (`jailer 0.21.0`) as separate from the firecracker version (`firecracker 0.21.0`). It will lead to introducing a compatibility table between firecracker and jailer (just my opinion) . WDYT?

@KarthikNedunchezhiyan , `jailer` and `firecracker` are considered crates in our project repository. Every crate has a specific package version, which is updated before a release of the crate. In our specific case, we have multiple crates, with dependencies between them, and a `workspace`, which manages these crates, as to what binaries to extract from this multiple crate project. A new release version of firecracker comes with two artifacts, namely the `firecracker` and `jailer` binaries, which correspond to `firecracker` and `jailer` crates. With this in mind, this means that the binaries are related to the crates versions, which can diverge, presumably, but not really in our case. Because they can diverge, it makes sense to keep `firecracker` version and `jailer` version separated.

KarthikNedunchezhiyan:
@lauralt @iulianbarbu ok 👍

lauralt:
> Hi~ @lauralt
> updated~~~ plz take a review. ;). and do you think it should be back-ported to `v0.21.0`?

I will, thanks. No, I don't think it should be backported as it is not actually a bug.

src/utils/src/arg_parser.rs
```diff
@@ -277,6 +278,13 @@ impl<'a> Arguments<'a> {
             return Ok(());
         }
```
lauralt: Can you please remove this newline and also update the comment from lines 271-273 to apply for both `--help` and `--version` flags, something like: ``If `--help` or `--version` is provided as a parameter, we artificially skip the parsing of other command line arguments by adding just that argument to the parsed list and returning.``

lauralt:
@Pennyzct Changes look good. I would rephrase a little bit the commit message, something like: 
The `--version` flag is quite an universal command line argument for
printing the application version number.
Therefore added flag `--version` to both firecracker and jailer.

Pennyzct:
Hi~ @lauralt thanks!
code and commit updated. plz take a review. ;)

> No, I don't think it should be backported as it is not actually a bug.
FYI, I've raised this request because of that in some app(e.g. kata-containers) which are using firecracker may use `--version` to acquire version info for pre-run compatibility check.

lauralt:
@Pennyzct The code coverage integration test is failing right now. Can you please update the target value from `test_coverage.py` with the current value (80.00) because, unfortunately, the newly added lines can not be tested properly :( ?

src/utils/src/arg_parser.rs
```diff
@@ -277,6 +278,16 @@ impl<'a> Arguments<'a> {
...
 
+        // If `--version` is provided as a parameter, we artificially skip the parsing of other
+        // command line arguments by adding just the version argument to the parsed list and
+        // returning.
+        if args.contains(&VERSION_ARG.to_string()) {
+            let mut version_arg = Argument::new("version").help("Print the binary version number.");
+            version_arg.user_value = Some(Value::Bool(true));
+            self.insert_arg(version_arg);
+            return Ok(());
+        }
```
andreeaflorescu: We should be able to add a test case for this code path. Isn't there one for parsing help as well?

lauralt: Yep, sorry, it is indeed.
@Pennyzct can you, please, update `test_parse()` with a test also for `--version` argument?

Pennyzct:
Hi~ @lauralt @andreeaflorescu 

> @Pennyzct can you, please, update test_parse() with a test also for --version argument?

I've already updated `test_parse()` with a new test for `--version` argument. And a little bit more, since we use clone to acquire the arguments, we probably don't need to `build_arg_parser()` for each small test union.

> @Pennyzct The code coverage integration test is failing right now. Can you please update the target value from test_coverage.py with the current value (80.00) because, unfortunately, the newly added lines can not be tested properly :( ?

thanks for the debugging. ;). Since we add test for the` --version` arg, the coverage test may not fail?? If it still fails, I'll then edit the value target accordingly.

src/firecracker/src/main.rs
```diff
@@ -144,6 +144,14 @@ fn main() {
                     process::exit(i32::from(vmm::FC_EXIT_CODE_OK));
                 }
             }
+
+            if let Some(version) = arg_parser.arguments().value_as_bool("version") {
```
andreeaflorescu: Since this is fixing a bug (version was previously available in Firecracker), we also need a changelog entry under `# Fixed` section.

Pennyzct: okay. update asap. Since if this is fixing a bug, could we back-port it to v0.21.0? 😊

andreeaflorescu: I think it makes sense. @firecracker-microvm/compute-capsule what do you think?

lauralt:
> Hi~ @lauralt @andreeaflorescu
> 
> > @Pennyzct can you, please, update test_parse() with a test also for --version argument?
> 
> I've already updated `test_parse()` with a new test for `--version` argument. And a little bit more, since we use `clone()` to acquire the arguments, we probably don't need to `build_arg_parser()` every time for each small test union, just the one in the beginning is enough.
> 
Nice catch! 
> > @Pennyzct The code coverage integration test is failing right now. Can you please update the target value from test_coverage.py with the current value (80.00) because, unfortunately, the newly added lines can not be tested properly :( ?
> 
> thanks for the debugging. ;). Since we add test for the` --version` arg, the coverage test may not fail?? If it still fails, I'll then edit the value target accordingly.

src/utils/src/arg_parser.rs
```diff
@@ -476,7 +488,16 @@ mod tests {
         assert!(arguments.parse(&args).is_ok());
         assert!(arguments.args.contains_key("help"));
 
-        let arg_parser = build_arg_parser();
+        let mut arguments = arg_parser.arguments().clone();
```
lauralt: Little nit: no need here for the `let mut` part.

src/firecracker/src/main.rs
lauralt: :+1:

src/utils/src/arg_parser.rs
Pennyzct: done. thanks. ;)

Pennyzct:
Hi~ @lauralt @andreeaflorescu 
If you're now all fine with the back-porting to `v0.21.0`, I added an new `FIXED` entry in `CHANGELOG` under v0.21.0. plz take a review~~~ ;)

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 - The HTTP header parsing is now case insensitive.
 - The `put_api_requests` and `patch_api_requests` metrics for net devices were
   un-swapped.
+- Added `--version` flag to both Firecracker and Jailer.
```
andreeaflorescu: We cannot release a patch version from this commit because the master branch is not stable just now and includes other changes as well. Can you move this under the `# Unreleased` section? For creating a patch release we need a new branch that starts from the commit of the v0.21.0 release, and only has this commit on top of it. I can create a branch and then we can open a new PR against that release branch.

Pennyzct: okay. thanks for the explanation~ updated. ;)

Pennyzct:
sorry. will re-base asap. ;)

CHANGELOG.md
Pennyzct: okay. thanks for the explanation. ;). updated.

andreeaflorescu:
I created a release branch for v0.21.1: https://github.com/firecracker-microvm/firecracker/tree/release_v0.21.1

You should be able to create a PR against that branch and then we can do a patch release.

lauralt:
@Pennyzct Can you please resolve the conflicts and do another `push -f`?

PR #1636, opened 2020-03-01
Codebase cleanup

KarthikNedunchezhiyan:
## Description of Changes

This PR includes:

- removed some superfluous code from arg parser.
- removed UUID module from jailer which is leftover of #441.

## License Acceptance

## PR Checklist

      clearly provided.
      code is included in this PR.


src/firecracker/src/main.rs
```diff
@@ -138,11 +138,9 @@ fn main() {
         }
         _ => {
             if let Some(help) = arg_parser.arguments().value_as_bool("help") {
-                if help {
```
andreeaflorescu: The default value of `help` is false, so we do need to check that help was configured (meaning that it is some), and that its value is true.

KarthikNedunchezhiyan: yes Andreea, the default value of arg parser which takes no value is `false` but `--help` is handled specially in this current model. So based on the current model only two values are possible either `None` or `Some(true)` for `--help`.

If we are planning to alter the logic of arg parser in the future then yes as you say it is better to leave that `if condition` to support future implementation. I am ok with both the flow 🙂

src/utils/src/arg_parser.rs
```diff
@@ -330,12 +328,8 @@ impl<'a> Arguments<'a> {
...
             self.validate_arg(arg)?;
 
-            // If the `arg` argument is indeed an expected one, set the value provided by user
-            // if it's a valid one.
-            let argument = self
-                .args
-                .get_mut(&arg[ARG_PREFIX.len()..])
-                .ok_or_else(|| Error::UnexpectedArgument(arg[ARG_PREFIX.len()..].to_string()))?;
+            // Safe to unwrap as arg is validated beforehand
+            let argument = self.args.get_mut(&arg[ARG_PREFIX.len()..]).unwrap();
```
lauralt: I know it's safe to unwrap here but, in case we will change the logic in `validate_arg()`, the `ok_or_else()` approach is safer, so I would leave here as it was before.

KarthikNedunchezhiyan: @lauralt done 👍

src/firecracker/src/main.rs
KarthikNedunchezhiyan: @andreeaflorescu done 👍

andreeaflorescu:
@KarthikNedunchezhiyan can you please update the commit message of your second commit to specify that you removed uid? This way it is easier to find it in the history in case we need to reintroduce that at some point.

KarthikNedunchezhiyan:
@andreeaflorescu changed commit as like you suggest 🙂.

src/utils/src/arg_parser.rs
```diff
@@ -208,9 +208,7 @@ impl<'a> Arguments<'a> {
     fn value_of(&self, arg_name: &'static str) -> Option<&Value> {
         if let Some(argument) = self.args.get(arg_name) {
             match &argument.user_value {
-                Some(val) => {
-                    return Some(val);
-                }
+                Some(val) => return Some(val),
                 None => return argument.default_value.as_ref(),
```
aghecenco: I'm biased towards one-liners and chained operators, so dropping in a Rust style nit:
```suggestion
       self.args.get(arg_name).and_then(|argument| {
            argument
                .user_value
                .as_ref()
                .or_else(|| argument.default_value.as_ref())
        })
```

KarthikNedunchezhiyan: @aghecenco changes done 👍 , also changed some other functions to match this style 🙂.

PR #1638, opened 2020-03-02
VirtioDevice trait cleanup

### LINKED ISSUE #1620: Unify virtio device activation path
Currently the block and net devices use `VirtioDevice::set_activated()` while vsock uses `VirtioDevice::activate()`. We should have a unified activation mechanism.

I am personally inclined to use `VirtioDevice::activate()` and `VirtioDevice::reset()` as these naturally map to the virtio commands as specified in the standard.

acatangiu:
## Reason for This PR

Fixes #1620 

## Description of Changes

Some small changes to the VirtioDevice trait interface:
- The VirtioDevice trait can work with references instead of owned objects for Queues and QueueEvents so no cloning is necessary. Also use slice references instead of owned Vecs.
- For a short while we were also using `activate`/`reset` and `set_device_activated`. This moves all devices and tests to using `activate`/`reset` to activate, respectivelly reset a VirtioDevice.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -679,12 +679,8 @@ impl VirtioDevice for Net {
         self.device_activated
     }
 
-    fn set_device_activated(&mut self, device_activated: bool) {
-        self.device_activated = device_activated;
-    }
-
-    fn activate(&mut self, _mem: GuestMemoryMmap) -> ActivateResult {
-        // TODO: to be removed
+    fn activate(&mut self) -> ActivateResult {
+        self.device_activated = true;
         Ok(())
     }
 }
@@ -816,7 +812,7 @@ mod tests {
             self.queues.clear();
             self.queues.push(rxq);
             self.queues.push(txq);
-            self.set_device_activated(true);
+            self.activate().unwrap();
         }
     }
 
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -621,24 +621,19 @@ impl VirtioDevice for Net {
         TYPE_NET
     }
 
-    fn get_queues(&mut self) -> &mut Vec<Queue> {
+    fn queues(&mut self) -> &mut [Queue] {
         &mut self.queues
     }
 
-    fn get_queue_events(&self) -> result::Result<Vec<EventFd>, std::io::Error> {
-        let mut queue_evts_copy: Vec<EventFd> = Vec::new();
-        for event_fd in &self.queue_evts {
-            queue_evts_copy.push(event_fd.try_clone()?);
-        }
-
-        Ok(queue_evts_copy)
+    fn queue_events(&self) -> &[EventFd] {
+        &self.queue_evts
     }
 
-    fn get_interrupt(&self) -> std::io::Result<EventFd> {
-        Ok(self.interrupt_evt.try_clone()?)
+    fn interrupt_evt(&self) -> &EventFd {
+        &self.interrupt_evt
     }
 
-    fn get_interrupt_status(&self) -> Arc<AtomicUsize> {
+    fn interrupt_status(&self) -> Arc<AtomicUsize> {
         self.interrupt_status.clone()
     }
 
@@ -1640,25 +1635,23 @@ mod tests {
         net.assign_queues(rxq.create_queue(), txq.create_queue());
 
         // Test queues count (TX and RX).
-        let queues = net.get_queues();
+        let queues = net.queues();
         assert_eq!(queues.len(), QUEUE_SIZES.len());
         assert_eq!(queues[RX_INDEX].size, rxq.size());
         assert_eq!(queues[TX_INDEX].size, txq.size());
 
         // Test corresponding queues events.
-        let queues_evts = net.get_queue_events().unwrap();
-        assert_eq!(queues_evts.len(), QUEUE_SIZES.len());
+        assert_eq!(net.queue_events().len(), QUEUE_SIZES.len());
 
         // Test interrupts.
-        let interrupt_status = net.get_interrupt_status();
+        let interrupt_status = net.interrupt_status();
         interrupt_status.fetch_or(VIRTIO_MMIO_INT_VRING as usize, Ordering::SeqCst);
         assert_eq!(
             interrupt_status.load(Ordering::SeqCst),
             VIRTIO_MMIO_INT_VRING as usize
         );
 
-        let interrupt_evt = net.get_interrupt().unwrap();
-        interrupt_evt.write(1).unwrap();
-        assert_eq!(interrupt_evt.read().unwrap() as usize, 1);
+        net.interrupt_evt().write(1).unwrap();
+        assert_eq!(net.interrupt_evt().read().unwrap() as usize, 1);
     }
 }
--- a/src/vmm/src/device_manager/mmio.rs
+++ b/src/vmm/src/device_manager/mmio.rs
@@ -110,12 +110,12 @@ impl MMIODeviceManager {
             return Err(Error::IrqsExhausted);
         }
 
-        let queue_evts = mmio_device
+        for (i, queue_evt) in mmio_device
             .locked_device()
-            .get_queue_events()
-            .map_err(Error::EventFd)?;
-
-        for (i, queue_evt) in queue_evts.iter().enumerate() {
+            .queue_events()
+            .iter()
+            .enumerate()
+        {
             let io_addr = IoEventAddress::Mmio(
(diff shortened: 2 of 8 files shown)


acatangiu:
> What do you think if we handle the events registration after activation for block and net devices like vsock does? With the vsock pattern of events registration, we do not register the queue/backend/rate limiters events until the activation, but only after. We wouldn't need to have the corresponding checks inside handle_read of net and block devices, and the actual event handling flow will seem more natural, IMO.

We could do that (definitely in a separate PR), but it'd require a two-step event registration process (like vsock does):
1. register an `activate` event
2. on receipt of an `activate` event, we register the other events.

It'd be formally cleaner, but I don't have any problem with the current flow either.

iulianbarbu:
> > What do you think if we handle the events registration after activation for block and net devices like vsock does? With the vsock pattern of events registration, we do not register the queue/backend/rate limiters events until the activation, but only after. We wouldn't need to have the corresponding checks inside handle_read of net and block devices, and the actual event handling flow will seem more natural, IMO.
> 
> We could do that (definitely in a separate PR), but it'd require a two-step event registration process (like vsock does):
> 
>     1. register an `activate` event
> 
>     2. on receipt of an `activate` event, we register the other events.
> 
> 
> It'd be formally cleaner, but I don't have any problem with the current flow either.

We can do it in another PR. Basically, what we would gain choosing the `vsock` way of events registration is proper handling of device events after device activation. We will still need the `set_device_activated` and `is_activated` interface, for logic related to device `reset`. I've opened an issue to track this alignment for all the devices:  #1639 .

iulianbarbu:
> > > What do you think if we handle the events registration after activation for block and net devices like vsock does? With the vsock pattern of events registration, we do not register the queue/backend/rate limiters events until the activation, but only after. We wouldn't need to have the corresponding checks inside handle_read of net and block devices, and the actual event handling flow will seem more natural, IMO.
> > 
> > 
> > We could do that (definitely in a separate PR), but it'd require a two-step event registration process (like vsock does):
> > ```
> > 1. register an `activate` event
> > 
> > 2. on receipt of an `activate` event, we register the other events.
> > ```
> > 
> > 
> > It'd be formally cleaner, but I don't have any problem with the current flow either.
> 
> We can do it in another PR. Basically, what we would gain choosing the `vsock` way of events registration is proper handling of device events after device activation. We will still need the `set_device_activated` and `is_activated` interface, for logic related to device `reset`. I've opened an issue to track this alignment for all the devices: #1639 .

Synced offline. We do not need `set_device_activated`. Also, the PR LGTM.

src/vmm/src/device_manager/mmio.rs
```diff
@@ -110,12 +110,12 @@ impl MMIODeviceManager {
             return Err(Error::IrqsExhausted);
         }
 
-        let queue_evts = mmio_device
+        for (i, queue_evt) in mmio_device
```
andreeaflorescu: With this change we're keeping the lock on device while registering all events. Is this behavior change on purpose? If yes, can you please specify this in the commit message?

src/devices/src/virtio/net/device.rs
```diff
@@ -679,12 +679,8 @@ impl VirtioDevice for Net {
         self.device_activated
     }
 
-    fn set_device_activated(&mut self, device_activated: bool) {
-        self.device_activated = device_activated;
-    }
-
-    fn activate(&mut self, _mem: GuestMemoryMmap) -> ActivateResult {
-        // TODO: to be removed
+    fn activate(&mut self) -> ActivateResult {
```
andreeaflorescu: How does this play out with resetting the device? I am not very sure, but I think we should be able to say that the device is no longer active in case of error, which implies this function should receive a parameter for setting the activate on true or false. I can't find the `reset` code in the code base. Was it removed?

src/vmm/src/device_manager/mmio.rs
acatangiu: No, this is not on purpose, but I didn't expect rust to hold the lock, I should do some experiments and confirm the behavior here.

In case it does hold the lock, I can:
- revert to the explicit syntax used before where the lock is released,
- just leave it like this since during its registration, the device has no other user so there is no contention on the lock whatsoever; nothing really to report in the commit message afaict

Reviewer's choice ^

src/devices/src/virtio/net/device.rs
acatangiu: So none of our devices support `reset` so they just inherit the default trait implementation:
```rust
    /// Optionally deactivates this device and returns ownership of the guest memory map, interrupt
    /// event, and queue events.
    fn reset(&mut self) -> Option<(EventFd, Vec<EventFd>)> {
        None
    }
```

This is used in `src/devices/src/virtio/mmio.rs:194` when the driver requests a reset:
```rust
            _ if status == 0 => {
                if self.locked_device().is_activated() {
                    let mut device_status = self.device_status;
                    let reset_result = self.locked_device().reset();
                    match reset_result {
                        Some((_interrupt_evt, mut _queue_evts)) => {}
                        None => {
                            device_status |= FAILED;
                        }
                    }
                    self.device_status = device_status;
                }

                // If the backend device driver doesn't support reset,
                // just leave the device marked as FAILED.
                if self.device_status & FAILED == 0 {
                    self.reset();
                }
            }
```

So we just mark our devices as `FAILED`.

andreeaflorescu: Ok, we can add a parameter to `activate` when we implement `reset`.

src/vmm/src/device_manager/mmio.rs
andreeaflorescu: I prefer the previous syntax as there is no need to hold the lock for this operation. Please let me know if your experiments confirm the behavior.

iulianbarbu: I think @andreeaflorescu is right. I've managed to reproduce what she suggests. The conclusion is that the syntax introduced in this PR will keep the lock for the scope of the for loop. How I've reproduced:

```
use std::sync::{Arc, Mutex};
use std::{thread, time};
 
struct HoldsQueueEvts {
     queue_evts: Vec<u32>
}

fn main() {
        let queue_evts = vec![3,2,1];
        let device = Arc::new(Mutex::new(HoldsQueueEvts { queue_evts: queue_evts }));
        let mut handles = Vec::new();
        
        for i in 0..2 {
                let device_clone = device.clone();
                handles.push(thread::spawn(move || {
                        // This thread will block the other from processing.
                        // START
                        //for (idx, evt) in device_clone.lock().unwrap().queue_evts.clone().iter().enumerate() {
                                //println!("{}: {} {}", i, idx, evt);   
                                //if i == 0 {
                                //      let ten_millis = time::Duration::from_millis(2000);
                                //      thread::sleep(ten_millis);
                                //}   
                        //}
                        // END


                        // This is how it wouldn't block the other thread.
                        // START
                        let queue_evts = device_clone.lock().unwrap().queue_evts.clone();
                        for (idx, val) in  queue_evts.iter().enumerate() {
                                println!("{}: {} {}", i, idx, val);
                                if i == 0 {
                                        let ten_millis = time::Duration::from_millis(2000);
                                        thread::sleep(ten_millis);
                                }
                        }
                        // END
                }));
        }

        for handle in handles {
                handle.join().unwrap();
        }
}
```

**OUTPUT 1**:
0: 0 3
0: 1 2
0: 2 1
1: 0 3
1: 1 2
1: 2 1

**OUTPUT 2**:
0: 0 3
1: 0 3
1: 1 2
1: 2 1
0: 1 2
0: 2 1

acatangiu: @andreeaflorescu @iulianbarbu 

Holding the lock is actually necessary now because we're taking a reference to the inner object data. In the old code we used to explicitly clone it under the lock then release the lock and provide the independent clone.

The new interface does not force cloning resources anymore. The reference can be safely used while the lock is active. If one would need to release the lock they could do it like before and explicitly clone the data under the reference then release the lock.

In this particular line, I think it's more expensive to clone the data than holding the lock for the duration of the loop, because holding the lock here is free since there is no user user of the device at this time that can suffer from contention.

iulianbarbu: I agree that at this point we do not have any other user of the device. Getting the lock and unlocking is cheap, since it does not imply any syscall, because there is no contention. Also, now we do not clone the queue events anymore, through the interface, but we borrow them, which would imply an explicit cloning step if we want to stick to the old syntax. Furthermore, I think it is not safe to approach this through cloning, since it means that we would have two owners for some specific event fds, which can be closed by one owner, but used still opened by the other, which would not provide shaky consistency for their usage. 

My thought is that we can stick to this syntax at the moment, because theoretically, there will be one single user which registers one specific device, when we build the microVM.

PR #1656, opened 2020-03-06
vsock: Use TempFile when generating file names

gc-plp:
## Reason for This PR

Occasionally, the vsock tests fail because the created file descriptors already exist. This PR tries to avoid that situation by trying multiple file names.
This can also happen when concurrent firecracker test runs are executed from the same path.

## Description of Changes

When a file is created, first, check if it exists. If it does, try to append a number to the file name.

## License Acceptance

## PR Checklist

- [x] no docs need to be updated as part of this PR
- [x] code-level documentation for touched code is included in this PR.
- [x] no API changes are included in this PR
- [x] the changes in this PR have no user impact
- [x] Either no new `unsafe` code has been added

### DIFF
--- a/src/devices/src/virtio/vsock/unix/muxer.rs
+++ b/src/devices/src/virtio/vsock/unix/muxer.rs
@@ -754,6 +754,7 @@ mod tests {
     use std::ops::Drop;
     use std::os::unix::net::{UnixListener, UnixStream};
     use std::path::{Path, PathBuf};
+    use utils::tempfile::TempFile;
 
     use super::super::super::csm::defs as csm_defs;
     use super::super::super::tests::TestContext as VsockTestContext;
@@ -776,6 +777,17 @@ mod tests {
         }
     }
 
+    // Create a TempFile with a given prefix and return it as a nice String
+    fn get_file(fprefix: &str) -> String {
+        let listener_path = TempFile::new_with_prefix(fprefix.to_owned()).unwrap();
+        listener_path
+            .as_path()
+            .as_os_str()
+            .to_str()
+            .unwrap()
+            .to_owned()
+    }
+
     impl MuxerTestContext {
         fn new(name: &str) -> Self {
             let vsock_test_ctx = VsockTestContext::new();
@@ -786,9 +798,8 @@ mod tests {
                     .unwrap(),
             )
             .unwrap();
-            let uds_path = format!("test_vsock_{}.sock", name);
-            let muxer = VsockMuxer::new(PEER_CID, uds_path).unwrap();
 
+            let muxer = VsockMuxer::new(PEER_CID, get_file(name)).unwrap();
             Self {
                 _vsock_test_ctx: vsock_test_ctx,
                 pkt,


src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -786,7 +810,8 @@ mod tests {
                     .unwrap(),
             )
             .unwrap();
-            let uds_path = format!("test_vsock_{}.sock", name);
+
+            let uds_path = get_available_file(format!("test_vsock_{}.sock", name));
```
andreeaflorescu: Can we use `Tempfile` instead to generate a unique path?

dianpopa:
I did experiment these failures myself. The `MuxerTestContext` the one binding to the `uds_path` has `Drop` implemented which deletes this path, so in theory we should never have to deal with these failures. The `LocalListener` (the one that listens on `uds_path_port`) also has a Drop implemented  that deletes the path on which it listened for connections. It would be interesting to see why these test fail even if in theory the paths used across the tests are removed. One other thing I noticed: for me there is one test that fails every time (`test_muxer_killq`).

src/devices/src/virtio/vsock/unix/muxer.rs
gc-plp: Yep, used it. 

See latest update, please.

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -776,6 +777,17 @@ mod tests {
...
+    // Create a TempFile with a given prefix and return it as a nice String
+    fn get_file(fprefix: &str) -> String {
+        let listener_path = TempFile::new_with_prefix(fprefix.to_owned()).unwrap();
+        listener_path
+            .as_path()
+            .as_os_str()
+            .to_str()
+            .unwrap()
+            .to_owned()
+    }
```
acatangiu, line 789: This will indeed use a unique path, but will not cleanup the socket UDS when the test is complete.

The TempFile here will cleanup itself when it goes out of scope (end of this fn), the path is passed to the muxer which creates an UDS there, but nobody cleans it up.

You can check `src/vmm/src/vmm_config/vsock.rs` tests to see an example to make sure it's also cleaned up.

Ideally we should create a `TempPathObject` in some tests utility crate so we can reuse the model everywhere instead of duplicating it around (it's currently duplicated in a few places around our codebase).

```rust
    /// When this object goes out of scope, the internal path is `unlinked` from the file-system.
    pub(crate) struct TempPathObject {
        path: String,
    }
    impl TempPathObject {
        pub fn new(fprefix: &str) -> Self {
            let obj_path = TempFile::new_with_prefix(fprefix.to_owned()).unwrap();
            TempPathObject {
                path: obj_path
                    .as_path()
                    .as_os_str()
                    .to_str()
                    .unwrap()
                    .to_owned(),
            }
        }
        pub fn path(&self) -> &String {
            &self.path
        }
    }
    impl Drop for TempPathObject {
        fn drop(&mut self) {
            let _ = std::fs::remove_file(&self.path);
        }
    }
```

gc-plp: > This will indeed use a unique path, but will not cleanup the socket UDS when the test is complete.

Line `774` implements the `Drop` trait for `MuxerTestContext`.

In the ideal scenario where this test works and finishes successfully, the `host_sock_path` is cleaned up and another test run can be executed.
If the test panics for some reason, there will be files left in the test folder.

The reason for dropping the `listener_path` variable there is that the `VsockMuxer` tries to open the file again and since the `TempPath` implementation opens the file by default, we just need a random unopened path.

andreeaflorescu: >  If the test panics for some reason, there will be files left in the test folder.

Drop is called on panic as well, so we shouldn't have this problem.

acatangiu: You're right @gc-plp, the UDS is in fact cleaned up by the MuxerTestContext drop impl.
I'm resolving this comment, feel free to merge the PR.

Issue #1624, opened 2020-02-21
Add tests for vmm::lib.rs


acatangiu:
The Vmm now exposes a crisp and simple programmatic interface for microVm runtime operations.
We should add unit tests for those functions where it makes sense (some might be better tested with integration tests).

iulianbarbu:
The units from `vmm::lib.rs` are tested through integration tests. I've identified the following units, and I think we have thorough integration tests which test the functionality:

- `pub fn get_bus_device(
        &self,
        device_type: DeviceType,
        device_id: &str,
    ) -> Option<&Mutex<dyn BusDevice>>` which is a wrapper over `MmioDeviceManager::get_bus_device`, which is tested as a unit inside tests for `device_manager/mmio.rs`.

- `pub fn start_vcpus(
        &mut self,
        mut vcpus: Vec<Vcpu>,
        vmm_seccomp_filter: BpfProgram,
        vcpu_seccomp_filter: BpfProgramRef,
    ) -> Result<()>` which is used every time we build and start a microVM, basically in every integration test which boots a kernel.

- `pub fn resume_vcpus(&mut self) -> Result<()>` used inside `Vmm::start_vcpus`.

- `pub fn configure_system(&self, vcpus: &[Vcpu], initrd: &Option<InitrdConfig>) -> Result<()>` used every time we build and start a microVM, for x86_64 and aarch64 respectively.

- `pub fn send_ctrl_alt_del(&mut self) -> Result<()>` tested by `functional/test_api::test_send_ctrl_alt_del`.

- `pub fn stop(&mut self, exit_code: i32)` tested by `functional/test_api::test_send_ctrl_alt_del`, when rebooting.

- `fn log_boot_time(t0_ts: &TimestampUs)` tested by `performance/test_bootime`.

PR #1661, opened 2020-03-09
test_shut_down.py: assert reboot flushes metrics

iulianbarbu:
## Reason for This PR

#1624

## Description of Changes

Updated `test_shut_down.py` to assert the flushing of metrics when a reboot is issued from within the booted microVM.

`mod tests` from vmm/lib.rs is not neccessary for this file, since the units testing fit better for integration tests, which are already covered by our current integration tests.

Modified the API for creation and flags manipulation of a `logging::Fifo`. These changes offer flexibility in choosing if the opened pipe is blocking or non blocking and support for changing file descriptor flags for the opened fifo.

## License Acceptance

## PR Checklist

- [x] Bo docs need to be updated as part of this PR.

### DIFF
--- a/tests/host_tools/logging.py
+++ b/tests/host_tools/logging.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Utilities for testing the logging system (metrics, common logs)."""
 
+import fcntl
 import os
 import sys
 
@@ -15,22 +16,41 @@ class Fifo:
     path = None
     fifo = None
 
-    def __init__(self, path):
+    def __init__(self, path, blocking=False):
         """Create a new named pipe."""
         if os.path.exists(path):
             raise FileExistsError("Named pipe {} already exists.".format(path))
+
         os.mkfifo(path)
-        # Open the FIFO and set O_NONBLOCK flag.
-        fd = os.open(path, os.O_RDONLY | os.O_NONBLOCK)
-        self.fifo = os.fdopen(fd, "r")
+        if not blocking:
+            fd = os.open(path, os.O_NONBLOCK)
+            self.fifo = os.fdopen(fd, "r")
+        else:
+            self.fifo = open(path, "r")
+
         self.path = path
 
     def sequential_reader(self, max_lines):
-        """Return up to `max_lines` lines from fifo `fifo_index`.
+        """Return up to `max_lines` lines from a non blocking fifo.
 
         :return: A list containing the read lines.
         """
         return self.fifo.readlines()[:max_lines]
+
+    @property
+    def flags(self):
+        """Return flags of the opened fifo.
+
+        :return An integer with flags of the opened file.
+        """
+        fd = self.fifo.fileno()
+        return fcntl.fcntl(fd, fcntl.F_GETFL)
+
+    @flags.setter
+    def flags(self, flags):
+        """Set new flags for the opened fifo."""
+        fd = self.fifo.fileno()
+        fcntl.fcntl(fd, fcntl.F_SETFL, flags)
 
     def threaded_reader(self, check_func, *args):
         """Start a thread to read fifo.
--- a/tests/host_tools/logging.py
+++ b/tests/host_tools/logging.py
@@ -2,12 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 """Utilities for testing the logging system (metrics, common logs)."""
 
-import fcntl
 import os
 import sys
 
 from queue import Queue
-from subprocess import run
 from threading import Thread
 
 
@@ -22,8 +20,7 @@ class Fifo:
         if os.path.exists(path):
             raise FileExistsError("Named pipe {} already exists.".format(path))
         os.mkfifo(path)
-        
-	# Open the FIFO and set O_NONBLOCK flag.
+        # Open the FIFO and set O_NONBLOCK flag.
         fd = os.open(path, os.O_RDONLY | os.O_NONBLOCK)
         self.fifo = os.fdopen(fd, "r")
         self.path = path
@@ -60,10 +57,9 @@ class Fifo:
         Failures and exceptions are propagated to the main thread
         through the `exceptions_queue`.
         """
-        fifo = self._open_nonblocking()
         max_iter = 20
         while max_iter > 0:
-            data = fifo.readline()
+            data = self.fifo.readline()
             if not data:
                 break
             try:
--- a/tests/host_tools/logging.py
+++ b/tests/host_tools/logging.py
@@ -15,13 +15,17 @@ class Fifo:
     """Facility for creating and working with named pipes (FIFOs)."""
 
     path = None
+    fifo = None
 
     def __init__(self, path):
         """Create a new named pipe."""
         if os.path.exists(path):
             raise FileExistsError("Named pipe {} already exists.".format(path))
-        cmd = 'mkfifo ' + path
-        run(cmd, shell=True, check=True)
+        os.mkfifo(path)
+        
+	# Open the FIFO and set O_NONBLOCK flag.
+        fd = os.open(path, os.O_RDONLY | os.O_NONBLOCK)
+        self.fifo = os.fdopen(fd, "r")
         self.path = path
 
     def sequential_reader(self, max_lines):
@@ -29,8 +33,7 @@ class Fifo:
 
         :return: A list containing the read lines.
         """
-        fifo = self._open_nonblocking()
-        return fifo.readlines()[:max_lines]
+        return self.fifo.readlines()[:max_lines]
 
     def threaded_reader(self, check_func, *args):
         """Start a thread to read fifo.
@@ -48,14 +51,6 @@ class Fifo:
         )
         metric_reader_thread.start()
         return exceptions_queue
-
-    def _open_nonblocking(self):
-        """Open a FIFO as read-only and non-blocking."""
-        fifo = open(self.path, "r")
-        fd = fifo.fileno()
-        flag = fcntl.fcntl(fd, fcntl.F_GETFL)
-        fcntl.fcntl(fd, fcntl.F_SETFL, flag | os.O_NONBLOCK)
-        return fifo
 
     def _do_thread_reader(self, exceptions_queue, check_func, *args):
         """Read from a FIFO opened as read-only.
--- a/tests/integration_tests/functional/test_metrics.py
+++ b/tests/integration_tests/functional/test_metrics.py
@@ -3,7 +3,6 @@
 """Tests the metrics system."""
 
 import os
-
 import host_tools.logging as log_tools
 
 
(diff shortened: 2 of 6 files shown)


dianpopa:
I do think there are some potential scenarios that vmm/src/lib.rs should get unit tested/ci tested against:

==Unit tests==
* `get_bus_device( device_type: DeviceType,
        device_id: &str,)` -> isn 't it worth it to test that not registering a (device_type, id) tuple would make this function return `None`? Or also what about making sure that registering a net device with the same id as a block device works correctly?
* `guest_memory` -> why not test that it is actually returns what we configured?
*  `pub fn kvm_vm(&self)` is not used at all

==CI==
* Are  metrics flushed upon  `pub fn stop(&mut self, exit_code: i32)` ?
* Does `configure_system` fail if there is not enough memory to setup the mptable?

iulianbarbu:
> I do think there are some potential scenarios that vmm/src/lib.rs should get unit tested/ci tested against:
> 
> ==Unit tests==
> 
>     * `get_bus_device( device_type: DeviceType, device_id: &str,)` -> isn 't it worth it to test that not registering a (device_type, id) tuple would make this function return `None`? Or also what about making sure that registering a net device with the same id as a block device works correctly?

Testing this unit means testing the `MmioDeviceManager::get_bus_device` (because it is a wrapper which does not have any related logic), which is covered by `device_manager/mmio.rs::test_device_info` unit test. I genuinely do no think it would bring any value to have a unit test for this function.

> 
>     * `guest_memory` -> why not test that it is actually returns what we configured?
> 
>     * `pub fn kvm_vm(&self)` is not used at all

Unit testing for getters, if they do not have any complex logic, it would require enough effort just for nothing, in my opinion. I've excluded them from the report (#1624) just because they lack any related logic that would bring value for a unit test.
 
> 
> ==CI==
> 
>     * Are  metrics flushed upon  `pub fn stop(&mut self, exit_code: i32)` ?

I think this is a scenario that can be tested inside an integration testing. I will add an integration test for this. 

>     * Does `configure_system` fail if there is not enough memory to setup the mptable?

`vmm/lib.rs::configure_system` is a wrapper over `arch_*::configure_system`, without any related logic to it, besides effective calls to `arch_*::configure_system`. Testing it for `NotEnoughMemory` for mptable means testing `configure_system` from `arch_*`. This is already covered in unit tests.

dianpopa:
> > I do think there are some potential scenarios that vmm/src/lib.rs should get unit tested/ci tested against:
> > ==Unit tests==
> > ```
> > * `get_bus_device( device_type: DeviceType, device_id: &str,)` -> isn 't it worth it to test that not registering a (device_type, id) tuple would make this function return `None`? Or also what about making sure that registering a net device with the same id as a block device works correctly?
> > ```
> 
> Testing this unit means testing the `MmioDeviceManager::get_bus_device` (because it is a wrapper which does not have any related logic), which is covered by `device_manager/mmio.rs::test_device_info` unit test. I genuinely do no think it would bring any value to have a unit test for this function.
> 
> > ```
> > * `guest_memory` -> why not test that it is actually returns what we configured?
> > 
> > * `pub fn kvm_vm(&self)` is not used at all
> > ```
> 
> Unit testing for getters, if they do not have any complex logic, it would require enough effort just for nothing, in my opinion. I've excluded them from the report (#1624) just because they lack any related logic that would bring value for a unit test.
> 
> > ==CI==
> > ```
> > * Are  metrics flushed upon  `pub fn stop(&mut self, exit_code: i32)` ?
> > ```
> 
> I think this is a scenario that can be tested inside an integration test. I will add an integration test for this.
> 
> > ```
> > * Does `configure_system` fail if there is not enough memory to setup the mptable?
> > ```
> 
> `vmm/lib.rs::configure_system` is a wrapper over `arch_*::configure_system`, without any related logic to it, besides effective calls to `arch_*::configure_system`. Testing it for `NotEnoughMemory` for mptable means testing `configure_system` from `arch::x86_64`. This is already covered in unit tests.

andreeaflorescu:
> I am not on the same page here regarding not making sense to test simple bits of code. Unit tests are not made only for complex units. Testing a getter ensures that the getter stays a getter. Unit tests should also be seen as an extra layer of protection against malicious or unintentional mistakes in PRs.

I agree with you that you generally want to test even the simple bits of code. In this particular usecase, testing the `Vmm` requires a lot of setup and comes with tests that are hard to maintain. The value added by testing setters and getters tends to fade away when taking into consideration the long time required for maintaining the tests and their setup. I think it makes more sense to invest this time in writing integration tests. WDYT?

dianpopa:
> > I am not on the same page here regarding not making sense to test simple bits of code. Unit tests are not made only for complex units. Testing a getter ensures that the getter stays a getter. Unit tests should also be seen as an extra layer of protection against malicious or unintentional mistakes in PRs.
> 
> I agree with you that you generally want to test even the simple bits of code. In this particular usecase, testing the `Vmm` requires a lot of setup and comes with tests that are hard to maintain. The value added by testing setters and getters tends to fade away when taking into consideration the long time required for maintaining the tests and their setup. I think it makes more sense to invest this time in writing integration tests. WDYT?

I think we should invest time in writing integration tests regardless of whether we write unit tests for the vmm or not.
As far as my concern, no exception should be made as far as unit tests. It is a dangerous direction to start being subjective about what you unit test and what you do not. Unit test should exercise exactly a single method from a public API regardless of what it is built upon or what calls it. That is why declaring the `impl vmm` as unit tested is wrong IMO.

However, this PR can go through.

tests/integration_tests/functional/test_metrics.py
```diff
@@ -33,3 +35,55 @@ def test_flush_metrics(test_microvm_with_api):
...
+    metrics_file_path = os.path.join(microvm.path, 'metrics_file')
+    metrics_file = log_tools.File(metrics_file_path)
+
+    response = microvm.metrics.put(
+        metrics_fifo=microvm.create_jailed_resource(metrics_file.path)
+    )
+    assert microvm.api_session.is_status_no_content(response.status_code)
+
+    microvm.start()
+    time.sleep(2)
```
acatangiu: I don't think this is robust code, sleeping for two seconds is arbitrarily too long or too short depending on the machine this test runs and/or how loaded the machine is.

It would be better to use a solution that deterministic-ally finds when a microVM has booted, then continues running the test.

aghecenco:
Re: the unit testing of small bits of code: I agree with @dianpopa that we should test everything, even if it's already covered. I've seen way too many getters that were sneaking in a bit of logic on the side before returning a value. I believe that code that isn't thoroughly tested has a better chance of not aging well.

tests/integration_tests/functional/test_metrics.py
```diff
@@ -3,7 +3,9 @@
 """Tests the metrics system."""
 
 import os
-
+import time
```
aghecenco: nit: please group imports alphabetically.

tests/integration_tests/functional/test_metrics.py
```diff
@@ -33,3 +35,55 @@ def test_flush_metrics(test_microvm_with_api):
...
     lines = metrics_fifo.sequential_reader(how_many_flushes)
     assert len(lines) == how_many_flushes
+
+
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
+    reason="not yet implemented on aarch64"
+)
+def test_stop_and_flush_metrics(test_microvm_with_atkbd):
+    """Check if stop/reboot flush the metrics."""
```
aghecenco:
```suggestion
    """Check if stop/reboot flushes the metrics."""
```

tests/host_tools/logging.py
```diff
@@ -11,32 +11,32 @@
 from threading import Thread
 
 
-class Fifo:
-    """Facility for creating and working with named pipes (FIFOs)."""
+class File:
```
aghecenco: Why this change? Why not keep using fifos? I think we should stick to fifos because in the long run we'll add high-density tests. Fifos at least have limited capacity, so, if nobody's reading from the other end, they won't accept new incoming data. Files on the other hand will fill up the disk if anything's wrong. If there's a synchronization or I/O issue with fifos, please fix that instead.

acatangiu: Disclaimer: I'm fine with both files or fifos here.

There are pros and cons to each, but just to make sure we care about the right ones, will this code ever be used for high-density testing (honest question)?

iulianbarbu:
@dianpopa , @acatangiu , @aghecenco , I think I've managed your concerns. Please take a look again.

tests/host_tools/logging.py
```diff
@@ -2,35 +2,43 @@
...
-        fifo = self._open_nonblocking()
-        return fifo.readlines()[:max_lines]
+        return self.fifo.readlines()[:max_lines]
+
+    def sequential_blocking_reader(self, max_lines):
+        """Return up to `max_lines` lines from fifo which blocks.
+
+        :return: A list containing the read lines.
+        """
+        blocking_fifo = open(self.path, "r")
```
aghecenco: Why open a new one instead of `fcntl`'ing the existing fd?

aghecenco: Actually, why not specify in the constructor if you want the fifo to be blocking or not, and use a `self.blocking` attribute to control the behavior?

tests/host_tools/logging.py
aghecenco: I'd see it that way, yup. I could see the use of throughput metrics in high density, as well as the boot time log (which we looked for in the `test_multiple_microvms` scenarios). So unless we rewrite the whole test harness (or write a new one for perf&stress), I think this framework might be used in density tests too.

tests/host_tools/logging.py
iulianbarbu: Yes, specifying the blocking behavior in the constructor makes sense. I will add also a method for changing the blocking behavior. E.g. opening a pipe in blocking mode, use it, then deciding that we want to use it later on with non blocking behavior, or vice versa. WDYT?

tests/host_tools/logging.py
```diff
@@ -15,23 +16,45 @@ class Fifo:
...
+        """
+        fd = self.fifo.fileno()
+        return fcntl.fcntl(fd, fcntl.F_GETFL)
+
+    def or_flags(self, flags):
+        """Or between `flags` arg and opened filed existing flags."""
+        new_flags = self.flags() | flags
+        self.set_flags(new_flags)
+
+    def set_flags(self, flags):
```
aghecenco: The getter/setter could be replaced with a `@property` like here: https://github.com/firecracker-microvm/firecracker/blob/eef40ed38718b8ef065a7bf2d53c64b30ffafb62/tests/framework/microvm.py#L169

iulianbarbu: @aghecenco , please take a look! I've added @property decorator, as suggested.

dianpopa: The nice thing about this type of decorators is that you can simply do:
```
# getter example
fl  = fifo.flags 
# setter
fifo.flags = OS_NONBLOCK
```
So the perk here is that you can use same name for both getter and setter when decorating them. However, I see you used "set_flags" for the setter.

iulianbarbu: Oh, right. I will push pushed an updated with your usage. Thanks.

PR #1669, opened 2020-03-11
Fix API race during microVM configuration

acatangiu:
## Reason for This PR

This fixes a race during config/pre-boot phase where the VMM thread reads this event_fd before API thread gets the chance to increment it.

## Description of Changes

The API event_fd is used by the API thread to signal the VMM thread that it has pending messages.

During config/pre-boot phase, this event_fd is used in a simple blocking loop that waits for an API event, then processes the request.
To keep this scenario simple, make the event_fd a blocking event_fd.

During runtime/post-boot phase the event_fd is registered with polly alongside multiple other events (from multiple sources), and is only read when EPOLLIN is reported on it. Having it as a blocking event_fd is safe.

## License Acceptance

## PR Checklist

      clearly provided.
- [ ] Either no docs need to be updated as part of this PR, or, the required
- [ ] Either no code has been touched, or, code-level documentation for touched
      code is included in this PR.
- [ ] Either no API changes are included in this PR, or, the API changes are
- [ ] Either the changes in this PR have no user impact, or, the changes in
- [ ] Either no new `unsafe` code has been added, or, the newly added `unsafe`

### DIFF
--- a/src/firecracker/src/api_server_adapter.rs
+++ b/src/firecracker/src/api_server_adapter.rs
@@ -59,7 +59,6 @@ impl Subscriber for ApiServerAdapter {
         let event_set = event.event_set();
 
         if source == self.api_event_fd.as_raw_fd() && event_set == EventSet::IN {
-            let _ = self.api_event_fd.read();
             match self.from_api.try_recv() {
                 Ok(api_request) => {
                     let response = self.controller.handle_request(*api_request);
@@ -76,6 +75,7 @@ impl Subscriber for ApiServerAdapter {
                     panic!("The channel's sending half was disconnected. Cannot receive data.");
                 }
             };
+            let _ = self.api_event_fd.read();
         } else {
             error!("Spurious EventManager event for handler: ApiServerAdapter");
         }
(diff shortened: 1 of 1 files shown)


src/firecracker/src/api_server_adapter.rs
```diff
@@ -76,6 +75,7 @@ impl Subscriber for ApiServerAdapter {
                     panic!("The channel's sending half was disconnected. Cannot receive data.");
                 }
             };
+            let _ = self.api_event_fd.read();
```
iulianbarbu: Why it's necessary to consume the API event after we read from the channel?

sandreim: We need to consume the source of the event as an ack that we processed the event. If we do not consume it, epoll_wait will exit every time and report the event leading to an epoll driven busy wait loop.

iulianbarbu: Yes, this is clear for me. But the changes from this PR moved the `self.api_event_fd.read()` call after the `match self.from_api.try_recv()`. I do not understand why we need this. I believe that it does not matter when we consume it, before or after the channel read.

acatangiu: Moving the `event.read()` _after_ the `channel.try_read()` is just paranoia for the following unlikely scenario:

> We introduce some faulty logic that also consumes the api_event from some other place (I know, near zero probability, therefore characterizing this move as paranoia). What could happen is that epoll wakes this thread up, but the event is incorrectly consumed by some theoretical faulty code on a different thread, and because the event_fd is now blocking, this function blocks indefinitely.

By moving the event_fd read after the `channel.try_read()` we still get the same behavior but at least we get the channel "spurious event" warning in the logs and that would help pin-point the problem fairly fast.

Again, this move is not necessary, it's just defensive programming and it doesn't cost anything.

hfingler:
Just started having this issue and this fixed it. Thanks!

PR #1678, opened 2020-03-18
Gracefully handle EINTR events in our async loops

### LINKED ISSUE #197: Firecracker unexpectedly crashes even on 'safe' signals
For example:

Catch the EINTR error when running the epoll loop.
Error:
thread '<unnamed>' panicked at 'VMM thread fail: Poll(Error { repr: Os { code: 4, message: "Interrupted system call" } })', /checkout/src/libcore/result.rs:906:4

I ran into this issue when my laptop went to sleep. I guess the firecracker process got a EINTR signal that we don't handle properly in the epoll loop in run_control.

acatangiu:
## Reason for This PR

Fixes #197 

## Description of Changes

Gracefully handle EINTR errors in our `EventManager` and `HttpServer` requests loops, they are internally consumed and treated as a `epoll` timeout with no events rather than io (fatal) errors.

The inner `vsock epoll` loop doesn't care about EINTR because it always runs non-blocking (with `timeout=0`) and also doesn't `break` on errors.

### Fix Validation

**Before**
```
kata@kata-devo:~/code/firecracker-demo$ sudo gdb -p `pidof firecracker`

Attaching to process 9345
[New LWP 9346]
[New LWP 9466]
0x000000000079a6a7 in epoll_pwait ()
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /home/kata/code/firecracker-demo/firecracker.
Use `info auto-load python-scripts [REGEXP]' to list them.
(gdb) c
Continuing.

Thread 1 "firecracker" received signal SIGSYS, Bad system call.
0x00000000007a069b in getcwd ()
(gdb) 
Continuing.
[LWP 9466 exited]
[LWP 9346 exited]
[Inferior 1 (process 9345) exited with code 0224]
```

**After**
```
kata@kata-devo:~/code/firecracker-demo$ sudo gdb -p `pidof firecracker`

Attaching to process 8020
[New LWP 8021]
[New LWP 8468]
0x000000000079a807 in epoll_pwait ()
warning: Missing auto-load script at offset 0 in section .debug_gdb_scripts
of file /home/kata/code/firecracker-demo/firecracker.
Use `info auto-load python-scripts [REGEXP]' to list them.
(gdb) c
Continuing.
^C
Thread 1 "firecracker" received signal SIGINT, Interrupt.
0x000000000079a807 in epoll_pwait ()
(gdb) c
Continuing.
```

## License Acceptance

## PR Checklist

### DIFF
--- a/src/micro_http/src/server.rs
+++ b/src/micro_http/src/server.rs
@@ -300,10 +300,11 @@ impl HttpServer {
         // current thread until at least one event is received.
         // The received notifications will then populate the `events` array with
         // `event_count` elements, where 1 <= event_count <= MAX_CONNECTIONS.
-        let event_count = self
-            .epoll
-            .wait(MAX_CONNECTIONS, -1, &mut events[..])
-            .map_err(ServerError::IOError)?;
+        let event_count = match self.epoll.wait(MAX_CONNECTIONS, -1, &mut events[..]) {
+            Ok(event_count) => event_count,
+            Err(e) if e.raw_os_error() == Some(libc::EINTR) => 0,
+            Err(e) => return Err(ServerError::IOError(e)),
+        };
         // We use `take()` on the iterator over `events` as, even though only
         // `events_count` events have been inserted into `events`, the size of
         // the array is still `MAX_CONNECTIONS`, so we discard empty elements
--- a/src/polly/src/event_manager.rs
+++ b/src/polly/src/event_manager.rs
@@ -178,14 +178,15 @@ impl EventManager {
     /// Wait for events for a maximum timeout of `miliseconds`. Dispatch the events to the
     /// registered signal handlers.
     pub fn run_with_timeout(&mut self, milliseconds: i32) -> Result<usize> {
-        let event_count = self
-            .epoll
-            .wait(
-                EventManager::EVENT_BUFFER_SIZE,
-                milliseconds,
-                &mut self.ready_events[..],
-            )
-            .map_err(Error::Poll)?;
+        let event_count = match self.epoll.wait(
+            EventManager::EVENT_BUFFER_SIZE,
+            milliseconds,
+            &mut self.ready_events[..],
+        ) {
+            Ok(event_count) => event_count,
+            Err(e) if e.raw_os_error() == Some(libc::EINTR) => 0,
+            Err(e) => return Err(Error::Poll(e)),
+        };
         self.dispatch_events(event_count);
 
         Ok(event_count)
--- a/tests/integration_tests/functional/test_signals.py
+++ b/tests/integration_tests/functional/test_signals.py
@@ -3,12 +3,13 @@
 """Tests scenarios for Firecracker signal handling."""
 
 import os
-from signal import SIGBUS, SIGSEGV
+from signal import SIGBUS, SIGRTMIN, SIGSEGV
 from time import sleep
 
 import pytest
 
 import host_tools.logging as log_tools
+import host_tools.network as net_tools
 
 
 @pytest.mark.parametrize(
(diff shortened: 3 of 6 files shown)


acatangiu:
I will also add an integration test for it in this PR.

acatangiu:
@andreeaflorescu I've added an integration test for this. The added test fails before the fix and works with the fix as expected.

tests/integration_tests/functional/test_signals.py
```diff
@@ -52,3 +53,36 @@ def test_sigbus_sigsegv(test_microvm_with_api, signum):
...
+    # Just validate a simple command: `nproc`
+    cmd = "nproc"
+    _, stdout, stderr = ssh_connection.execute_command(cmd)
+    assert stderr.read().decode("utf-8") == ""
+    assert int(stdout.read().decode("utf-8")) == 2
+
+    # We have a handler installed for this signal.
+    os.kill(firecracker_pid, SIGRTMIN+1)
+
+    # Validate the microVM is still up and running.
```
lauralt: Suggestion: If we only want to check that the vm is still running, we could send a request and check the response, for example a GET on the machine-config. This way you won't need the network config anymore.

acatangiu: I chose the ssh route as it covers many components functioning right. I think the test is much more valuable when it validates that the guest is unaffected (at least up and responding), not just that `firecracker` is still running.

src/polly/src/event_manager.rs
```diff
@@ -178,14 +178,15 @@ impl EventManager {
...
-                &mut self.ready_events[..],
-            )
-            .map_err(Error::Poll)?;
+        let event_count = match self.epoll.wait(
+            EventManager::EVENT_BUFFER_SIZE,
+            milliseconds,
+            &mut self.ready_events[..],
+        ) {
+            Ok(event_count) => event_count,
+            Err(e) if e.raw_os_error() == Some(libc::EINTR) => 0,
```
andreeaflorescu: We should just return here. There is no need to call `dispatch_events`.

src/micro_http/src/server.rs
```diff
@@ -300,10 +300,11 @@ impl HttpServer {
         // current thread until at least one event is received.
         // The received notifications will then populate the `events` array with
         // `event_count` elements, where 1 <= event_count <= MAX_CONNECTIONS.
-        let event_count = self
-            .epoll
-            .wait(MAX_CONNECTIONS, -1, &mut events[..])
-            .map_err(ServerError::IOError)?;
+        let event_count = match self.epoll.wait(MAX_CONNECTIONS, -1, &mut events[..]) {
+            Ok(event_count) => event_count,
+            Err(e) if e.raw_os_error() == Some(libc::EINTR) => 0,
```
andreeaflorescu: We can directly return `parsed_requests` here.

acatangiu: Since this is a very unlikely code path, I prefer to keep it simple at a very small expense in efficiency.

In terms of separation of concerns, the logic in this code block is currently contained to epoll-related logic. There is _a lot_ of code happening after this code-block. To short-circuit the function here would mean to introduce an assumption in this code block that all following code is irrelevant in an EINTR case (which is true right now, but adds validation burden to anyone making changes here in the future).

I will reluctantly short-circuit the function if you have a strong opinion we should do it. Let me know.

src/polly/src/event_manager.rs
acatangiu: The same _separation of concerns paradigm_ vs _optimizing an unlikely case_ could be applied here as well, but since this is a very simple and clear function I think it's perfectly fine to short-circuit it.

sandreim: I wanted to suggest avoiding the dispatch_events extra call for the unlikely usecase, but I think @acatangiu is in point here.

andreeaflorescu: For this code path I guess is fine to not return 0, but in the `micro_http` case, you have to read quite a lot of code to see that if you receive an interrupt an empty vector is returned.

From my point of view it actually makes the code harder to read, not easier. The separation of concerns main advantage is code readability and maintainability. It typically applies to components, not to how code is structured in a single function. Both of my comments were nits, so I think we can merge the code how it is now.

PR #1679, opened 2020-03-18
device.rs: increment metric instead of logging error

### LINKED ISSUE #1677: Lots of print out of "RX queue is empty"
I'm seeing lots of print out in VM with the following text:

anonymous-instance:ERROR:src/devices/src/virtio/net/device.rs:546] The RX queue is empty, there is no available buffer.

The exact case might be hard to reproduce. I'm running GPU virtualization using https://github.com/utcs-scea/ava with TensorFlow inference workload. The GPU virtualization software is forwarding apis which creates lots of network traffic. The guest os is Linux.

iulianbarbu:
We've introduced an error log for the case when there is no available buffer for RX data. This can happen for heavy network traffic usage, and it is not an actual error, rather a limitation for which we must increment a metric.

Signed-off-by: iulianbarbu <iul@amazon.com>

## Reason for This PR

Fixes #1677 

## Description of Changes

Removed an unnecessary error log.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -224,6 +224,10 @@ impl Block {
             used_any = true;
         }
 
+        if !used_any {
+            METRICS.block.no_avail_buffer.inc();
+        }
+
         used_any
     }
 
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -222,6 +222,7 @@ impl Net {
         let rx_queue = &mut self.queues[RX_INDEX];
         let mut next_desc = rx_queue.pop(&self.mem);
         if next_desc.is_none() {
+            METRICS.net.no_rx_avail_buffer.inc();
             return false;
         }
 
@@ -248,9 +249,7 @@ impl Net {
                         }
                         Err(e) => {
                             error!("Failed to write slice: {:?}", e);
-
                             METRICS.net.rx_fails.inc();
-
                             if let GuestMemoryError::PartialBuffer { completed, .. } = e {
                                 write_count += completed;
                             }
@@ -379,6 +378,7 @@ impl Net {
                 }
             }
         }
+
         if self.rx_deferred_irqs {
             self.rx_deferred_irqs = false;
             self.signal_used_queue()
@@ -471,7 +471,6 @@ impl Net {
                     Err(e) => {
                         error!("Failed to read slice: {:?}", e);
                         METRICS.net.tx_fails.inc();
-
                         if let GuestMemoryError::PartialBuffer { completed, .. } = e {
                             read_count += completed;
                         }
@@ -498,6 +497,8 @@ impl Net {
 
         if raise_irq {
             self.signal_used_queue()?;
+        } else {
+            METRICS.net.no_tx_avail_buffer.inc();
         }
 
         // An incoming frame for the MMDS may trigger the transmission of a new message.
@@ -543,8 +544,7 @@ impl Net {
     pub fn process_tap_rx_event(&mut self) {
         METRICS.net.rx_tap_event_count.inc();
         if self.queues[RX_INDEX].is_empty(&self.mem) {
-            error!("The RX queue is empty, there is no available buffer.");
-            METRICS.net.event_fails.inc();
+            METRICS.net.no_rx_avail_buffer.inc();
             return;
         }
 
@@ -593,8 +593,8 @@ impl Net {
                 self.resume_rx().unwrap_or_else(report_net_event_fail);
             }
             Err(e) => {
-                METRICS.net.event_fails.inc();
                 error!("Failed to get rx rate-limiter event: {:?}", e);
+                METRICS.net.event_fails.inc();
             }
         }
     }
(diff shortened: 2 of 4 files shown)


src/devices/src/virtio/net/device.rs
```diff
@@ -543,7 +543,6 @@ impl Net {
     pub fn process_tap_rx_event(&mut self) {
         METRICS.net.rx_tap_event_count.inc();
         if self.queues[RX_INDEX].is_empty(&self.mem) {
-            error!("The RX queue is empty, there is no available buffer.");
             METRICS.net.event_fails.inc();
```
acatangiu: I suggest we use a specific metric for this case. This metric is also used for error-paths.

src/devices/src/virtio/net/device.rs
```diff
@@ -543,7 +543,6 @@ impl Net {
     pub fn process_tap_rx_event(&mut self) {
         METRICS.net.rx_tap_event_count.inc();
         if self.queues[RX_INDEX].is_empty(&self.mem) {
-            error!("The RX queue is empty, there is no available buffer.");
```
acatangiu: Same problem in line 575 and I expect others. Please have the fix address all instances of this behavior.

src/devices/src/virtio/net/device.rs
```diff
@@ -652,7 +639,6 @@ impl VirtioDevice for Net {
     fn read_config(&self, offset: u64, mut data: &mut [u8]) {
         let config_len = self.config_space.len() as u64;
         if offset >= config_len {
-            error!("Failed to read config space");
```
andreeaflorescu: Reading and writing the config space should not be on the critical path. We should keep this log and the next one.

src/devices/src/virtio/block/device.rs
```diff
@@ -144,7 +144,6 @@ impl Block {
     pub(crate) fn process_queue_event(&mut self) {
         METRICS.block.queue_event_count.inc();
         if let Err(e) = self.queue_evts[0].read() {
-            error!("Failed to get queue event: {:?}", e);
```
acatangiu: This is an actual error that shouldn't happen. Keep it.

src/devices/src/virtio/block/device.rs
```diff
@@ -204,7 +203,6 @@ impl Block {
                             VIRTIO_BLK_S_OK
                         }
                         Err(e) => {
-                            error!("Failed to execute request: {:?}", e);
```
acatangiu: Also valid error that shouldn't happen in a valid scenario. Keep the log.

src/devices/src/virtio/block/device.rs
```diff
@@ -215,7 +213,6 @@ impl Block {
                     self.mem.write_obj(status, request.status_addr).unwrap();
                 }
                 Err(e) => {
-                    error!("Failed to parse available descriptor chain: {:?}", e);
```
acatangiu: Keep it

src/devices/src/virtio/block/device.rs
```diff
@@ -232,7 +229,6 @@ impl Block {
             .fetch_or(VIRTIO_MMIO_INT_VRING as usize, Ordering::SeqCst);
 
         self.interrupt_evt.write(1).map_err(|e| {
-            error!("Failed to signal used queue: {:?}", e);
```
acatangiu: Keep it

src/devices/src/virtio/block/device.rs
```diff
@@ -290,7 +286,6 @@ impl VirtioDevice for Block {
     fn read_config(&self, offset: u64, mut data: &mut [u8]) {
         let config_len = self.config_space.len() as u64;
         if offset >= config_len {
-            error!("Failed to read config space");
```
acatangiu: keep it

iulianbarbu:
@acatangiu , @andreeaflorescu , I've addressed your feedback! Per my current analysis, I've added metrics for no available buffers case, for the block & net devices (separate metrics for rx and tx queues).

src/devices/src/virtio/net/device.rs
```diff
@@ -543,8 +542,7 @@ impl Net {
     pub fn process_tap_rx_event(&mut self) {
         METRICS.net.rx_tap_event_count.inc();
         if self.queues[RX_INDEX].is_empty(&self.mem) {
-            error!("The RX queue is empty, there is no available buffer.");
-            METRICS.net.event_fails.inc();
+            METRICS.net.no_rx_avail_buffer.inc();
```
acatangiu: You need to increment this metric in _Line 225_ too.

```rust
    fn rx_single_frame(&mut self) -> bool {
        let rx_queue = &mut self.queues[RX_INDEX];
        let mut next_desc = rx_queue.pop(&self.mem);
        if next_desc.is_none() {
            // No buffers available <---------------------- increment metric
            return false;
        }
        ...
    }
```

iulianbarbu: This is true. I'll increment the no_rx_avail_buffer metric here, too.

iulianbarbu:
@acatangiu , @andreeaflorescu , please take a look again. I had to update the tests/integration_tests/build/test_coverage.py.

PR #1680, opened 2020-03-18
Added MMDS custom IPv4 support

### LINKED ISSUE #1228: Support for configurable MMDS Endpoint
For the moment, the MMDS Endpoint is hardcoded to `169.254.169.254`, the way IMDS does it.  However, we want to let users configure that (through the MMDS associated API request) to be something else if they need to.

iulianbarbu:
## Reason for This PR

Fixes #1228 

## Description of Changes

This PR is a proposal for supporting custom IPv4 address configuration for our MicroVM Metadata Service. It comprises a new API for configuring the MMDS IPv4 address and some changes/refactoring in our current MMDS logic, for cleansing. 

## License Acceptance

## PR Checklist

### DIFF
--- a/docs/mmds.md
+++ b/docs/mmds.md
@@ -55,6 +55,36 @@ maximum size bound for MMDS contents; the users are free to provide arbitrarily
 large inputs. However, the HTTP server is likely to encounter/become a
 bottleneck first, which means any API resource may have this potential issue.
 
+#### MMDS Configuration
+
+Users can set up a custom IPv4 address for MMDS. This can be achieved through a
+`PUT` request to the API server, before booting up the guest, with the path 
+`/mmds/config`. An example of MMDS configuration is the following:
+
+```json
+{
+"ipv4_address": "169.254.169.200"
+}
+```
+
+The `ipv4_address` value must be a valid IPv4 address. The configured IPv4 address
+can be used from within the microVM to issue `GET` requests  to the MMDS. An
+example of MMDS request from within the guest might look like the following:
+
+```bash
+curl -s "http://169.254.169.200/latest/meta-data/ami-id"
+```
+
+Depending on which configured network interface a user wants to communicate
+with the MMDS, an additional IP route is needed. E.g.:
+
+```bash
+ip route add 169.254.169.200 dev eth0
+```
+
+If MMDS configuration is not provided before booting up the guest, the MMDS
+IPv4 address defaults to `169.254.169.254`.
+
 ### Example use case: credential rotation
 
 For this example, the guest expects to find some sort of credentials (say, a
--- a/src/api_server/src/parsed_request.rs
+++ b/src/api_server/src/parsed_request.rs
@@ -56,7 +56,7 @@ impl ParsedRequest {
             (Method::Put, "logger", Some(body)) => parse_put_logger(body),
             (Method::Put, "machine-config", Some(body)) => parse_put_machine_config(body),
             (Method::Put, "metrics", Some(body)) => parse_put_metrics(body),
-            (Method::Put, "mmds", Some(body)) => parse_put_mmds(body),
+            (Method::Put, "mmds", Some(body)) => parse_put_mmds(body, path_tokens.get(1)),
             (Method::Put, "network-interfaces", Some(body)) => {
                 parse_put_net(body, path_tokens.get(1))
             }
@@ -169,7 +169,7 @@ pub enum Error {
 
 impl std::fmt::Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        match *self {
+        match self {
             Error::Generic(_, ref desc) => write!(f, "{}", desc),
             Error::EmptyID => write!(f, "The ID cannot be empty."),
             Error::InvalidID => write!(
@@ -349,7 +349,6 @@ mod tests {
              {}",
             ApiServer::basic_json_body("fault_message", "message")
         );
-
         assert_eq!(&buf[..], expected_response.as_bytes());
 
         // Empty ID error.
@@ -661,6 +660,17 @@ mod tests {
                 b"PUT /mmds HTTP/1.1\r\n\
                 Content-Type: application/json\r\n\
                 Content-Length: 2\r\n\r\n{}",
+            )
+            .unwrap();
+        assert!(connection.try_read().is_ok());
+        let req = connection.pop_parsed_request().unwrap();
+        assert!(ParsedRequest::try_from_request(&req).is_ok());
+
+        sender
+            .write_all(
+                b"PUT /mmds/config HTTP/1.1\r\n\
+                Content-Type: application/json\r\n\
+                Content-Length: 26\r\n\r\n{\"ipv4_address\":\"1.1.1.1\"}",
             )
             .unwrap();
         assert!(connection.try_read().is_ok());
--- a/src/api_server/src/request/mmds.rs
+++ b/src/api_server/src/request/mmds.rs
@@ -1,16 +1,33 @@
 // Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use super::super::VmmAction;
+use micro_http::StatusCode;
 use request::{Body, Error, ParsedRequest};
+use vmm::vmm_config::mmds::MmdsConfig;
 
 pub fn parse_get_mmds() -> Result<ParsedRequest, Error> {
     Ok(ParsedRequest::GetMMDS)
 }
 
-pub fn parse_put_mmds(body: &Body) -> Result<ParsedRequest, Error> {
-    Ok(ParsedRequest::PutMMDS(
-        serde_json::from_slice(body.raw()).map_err(Error::SerdeJson)?,
-    ))
+pub fn parse_put_mmds(
+    body: &Body,
+    path_second_token: Option<&&str>,
+) -> Result<ParsedRequest, Error> {
+    match path_second_token {
+        Some(config_path) => match *config_path {
+            "config" => Ok(ParsedRequest::Sync(VmmAction::SetMmdsConfiguration(
+                serde_json::from_slice::<MmdsConfig>(body.raw()).map_err(Error::SerdeJson)?,
+            ))),
+            _ => Err(Error::Generic(
+                StatusCode::BadRequest,
+                format!("Unrecognized PUT request path `{}`.", *config_path),
+            )),
+        },
+        None => Ok(ParsedRequest::PutMMDS(
+            serde_json::from_slice(body.raw()).map_err(Error::SerdeJson)?,
+        )),
+    }
 }
 
 pub fn parse_patch_mmds(body: &Body) -> Result<ParsedRequest, Error> {
@@ -33,10 +50,31 @@ mod tests {
         let body = r#"{
                 "foo": "bar"
               }"#;
-        assert!(parse_put_mmds(&Body::new(body)).is_ok());
+        assert!(parse_put_mmds(&Body::new(body), None).is_ok());
+        let invalid_body = "invalid_body";
+        assert!(parse_put_mmds(&Body::new(invalid_body), None).is_err());
 
-        let body = "invalid_body";
-        assert!(parse_put_mmds(&Body::new(body)).is_err());
+        let body = r#"{
+                "ipv4_address": "1.1.1.1"
+              }"#;
+        let path = "config";
+        assert!(parse_put_mmds(&Body::new(body), Some(&path)).is_ok());
+
+        let body = r#"{
+                "ipv4_address": ""
+              }"#;
+        assert!(parse_put_mmds(&Body::new(body), Some(&path)).is_err());
+
+        // Equivalent to reset the mmds configuration.
+        let empty_body = r#"{}"#;
+        assert!(parse_put_mmds(&Body::new(empty_body), Some(&path)).is_ok());
+
+        let invalid_config_body = r#"{
+                "invalid_config": "invalid_value"
+              }"#;
+        assert!(parse_put_mmds(&Body::new(invalid_config_body), Some(&path)).is_err());
+        assert!(parse_put_mmds(&Body::new(body), Some(&"invalid_path")).is_err());
+        assert!(parse_put_mmds(&Body::new(invalid_body), Some(&path)).is_err());
     }
 
     #[test]
@@ -44,9 +82,7 @@ mod tests {
         let body = r#"{
                 "foo": "bar"
               }"#;
-        assert!(parse_put_mmds(&Body::new(body)).is_ok());
-
-        let body = "invalid_body";
-        assert!(parse_put_mmds(&Body::new(body)).is_err());
+        assert!(parse_patch_mmds(&Body::new(body)).is_ok());
+        assert!(parse_patch_mmds(&Body::new("invalid_body")).is_err());
     }
 }
--- a/src/api_server/swagger/firecracker.yaml
+++ b/src/api_server/swagger/firecracker.yaml
@@ -317,6 +317,30 @@ paths:
           schema:
             $ref: "#/definitions/Error"
 
+  /mmds/config:
+    put:
+      summary: Set MMDS configuration. Pre-boot only.
+      description:
+        Creates MMDS configuration to be used by the MMDS network stack.
+      parameters:
+        - name: body
+          in: body
+          description: The MMDS configuration as JSON.
+          required: true
+          schema:
+            type: "#/definitions/MmdsConfig"
+      responses:
+        204:
+          description: MMDS configuration was created/updated.
+        400:
+          description: MMDS configuration cannot be updated due to bad input.
+          schema:
+            $ref: "#/definitions/Error"
+        default:
+          description: Internal server error
+          schema:
+            $ref: "#/definitions/Error"
+
   /network-interfaces/{iface_id}:
     put:
       summary: Creates a network interface. Pre-boot only.
@@ -564,6 +588,14 @@ definitions:
       metrics_path:
         type: string
         description: Path to the named pipe or file where the JSON-formatted metrics are flushed.
+
+  MmdsConfig:
+    type: object
+    description:
+      Defines the MMDS configuration.
+    properties:
+      ipv4_address:
+        type: string
 
   NetworkInterface:
     type: object
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -142,11 +142,10 @@ impl Net {
         let queues = QUEUE_SIZES.iter().map(|&s| Queue::new(s)).collect();
 
         let mmds_ns = if allow_mmds_requests {
-            Some(MmdsNetworkStack::new_with_defaults())
+            Some(MmdsNetworkStack::new_with_defaults(None))
         } else {
             None
         };
-
         Ok(Net {
             id,
             tap,
@@ -183,6 +182,11 @@ impl Net {
     /// Provides the MAC of this net device.
     pub fn guest_mac(&self) -> Option<&MacAddr> {
         self.guest_mac.as_ref()
+    }
+
+    /// Provides a mutable reference to the `MmdsNetworkStack`.
+    pub fn mmds_ns_mut(&mut self) -> Option<&mut MmdsNetworkStack> {
+        self.mmds_ns.as_mut()
     }
 
     fn signal_used_queue(&self) -> result::Result<(), DeviceError> {
--- a/src/dumbo/src/ns.rs
+++ b/src/dumbo/src/ns.rs
@@ -28,6 +28,7 @@ const DEFAULT_MAX_PENDING_RESETS: usize = 100;
 
 #[cfg_attr(test, derive(Debug, PartialEq))]
 enum WriteArpFrameError {
+    NoPendingArpReply,
     Arp(ArpFrameError),
     Ethernet(EthernetFrameError),
 }
@@ -49,17 +50,16 @@ impl From<handler::WriteNextError> for WritePacketError {
 }
 
 pub struct MmdsNetworkStack {
+    // Network interface MAC address used by frames/packets heading to MMDS server.
+    remote_mac_addr: MacAddr,
     // The Ethernet MAC address of the MMDS server.
     mac_addr: MacAddr,
-    // Whenever we detour a frame, we update the value of the remote MAC address. We need this
-    // because we don't send ARP request ourselves.
-    remote_mac_addr: MacAddr,
-    // The IPv4 address of the MMDS server.
+    // MMDS server IPv4 address.
     ipv4_addr: Ipv4Addr,
-    // We only remember the most recently received ARP request, and store the remote IPv4 address
-    // here (we keep the remote MAC address in self.remote_mac_addr), to be used when the next
-    // opportunity to send a frame presents itself.
-    pending_arp_reply: Option<Ipv4Addr>,
+    // ARP reply destination IPv4 address (requester of address resolution reply).
+    // It is the Ipv4Addr of the network interface for which the MmdsNetworkStack
+    // routes the packets.
+    pending_arp_reply_dest: Option<Ipv4Addr>,
     // This handles MMDS<->guest interaction at the TCP level.
     tcp_handler: TcpIPv4Handler,
 }
@@ -73,10 +73,10 @@ impl MmdsNetworkStack {
         max_pending_resets: NonZeroUsize,
     ) -> Self {
         MmdsNetworkStack {
-            mac_addr,
             remote_mac_addr: mac_addr,
+            mac_addr,
             ipv4_addr,
-            pending_arp_reply: None,
+            pending_arp_reply_dest: None,
             tcp_handler: TcpIPv4Handler::new(
                 ipv4_addr,
                 tcp_port,
@@ -86,10 +86,10 @@ impl MmdsNetworkStack {
         }
     }
 
-    pub fn new_with_defaults() -> Self {
+    pub fn new_with_defaults(mmds_ipv4_addr: Option<Ipv4Addr>) -> Self {
         // The unwrap is safe if parse_str() is implemented properly.
         let mac_addr = MacAddr::parse_str(DEFAULT_MAC_ADDR).unwrap();
-        let ipv4_addr = Ipv4Addr::from(DEFAULT_IPV4_ADDR);
+        let ipv4_addr = mmds_ipv4_addr.unwrap_or_else(|| Ipv4Addr::from(DEFAULT_IPV4_ADDR));
 
         // The unwrap()s are safe because the given literals are greater than 0.
         Self::new(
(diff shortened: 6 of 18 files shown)


src/dumbo/src/ns.rs
```diff
@@ -229,49 +248,59 @@ impl MmdsNetworkStack {
...
-        Ok(Some(
-            // The unwrap() is safe because arp_len > 0.
-            NonZeroUsize::new(eth_unsized.with_payload_len_unchecked(arp_len).len()).unwrap(),
-        ))
     }
 
     fn write_packet(&mut self, buf: &mut [u8]) -> Result<Option<NonZeroUsize>, WritePacketError> {
         let mut eth_unsized = self
             .prepare_eth_unsized(buf, ETHERTYPE_IPV4)
             .map_err(WritePacketError::Ethernet)?;
```
acatangiu: This can also go inside the `if` body below, no point in preparing the buffer if it's not going to be used.

iulianbarbu:
@acatangiu , I've done some significant changes to this PR. Please, take a look again.

src/mmds/src/data_store.rs
```diff
@@ -7,7 +7,7 @@ use serde_json::Value;
 
 /// The Mmds is the Microvm Metadata Service represented as an untyped json.
 #[derive(Clone)]
-pub struct Mmds {
+pub struct MmdsDataStore {
```
andreeaflorescu: Hmmm, naming this DataStore is a bit weird because you have a `DataStore` structure in the `data_store` module. The fact that `Mmds` is a data-store is a given because of the namespace (in this case the `data_store` module) from which you use it.

iulianbarbu: I understand your logic. I've changed it back to how it was, and I do not have a strong suggestion right now. At the moment, we can let it be, and get back to it once we feel a better motivation to take action. Changes related to this aspect are not in the scope of this PR. A little bit of background for why I chose to rename it: 

In this specific `mmds::data_store` case, `Mmds` struct is the only struct inside the module. Maybe it was designed with this in mind. However, at a first look, `Mmds` name has a strong resonance and I can't associate it with a data store, because of what it means: `MicroVM metadata service`. To work around this naming issue, I think we can stop this namespace hierarchy at `mmds` level, defining a struct called `DataStore` or `MmdsDataStore` inside it, to have suggestive naming but also clear separation through namespaces. If we would need to define structs specific to the data store (e.g structs that compose a data store), then I think it would make sense to add another level in the namespace hierarchy, called `data_store`. It is only my thought process, and it is not backed by any data which proves that it is better to have it this way. The only data/anecdote is the clarity which will bring to me. When I'll need to use, and see it, I'll ask myself: `Ok, Mmds struct from mmds::data_store namespace, but what it has to do with the data store, and how can I use it?` Pretty common question, but imagine if we would have it named `mmds::MmdsDataStore`, or simply `mmds::DataStore`. I'll say then: `Oh, DataStore struct. This must be the MMDS data store. I do not know how to use it, yet, but I know what it is designed for.`

iulianbarbu:
@alexandruag , @acatangiu , @andreeaflorescu , this PR is ready for functional/code review. Can you please take a look?

docs/mmds.md
```diff
@@ -55,6 +55,39 @@ maximum size bound for MMDS contents; the users are free to provide arbitrarily
 large inputs. However, the HTTP server is likely to encounter/become a
 bottleneck first, which means any API resource may have this potential issue.
 
+#### MMDS Configuration
+
+Users can setup an optional pool of IPv4 addresses for which the MMDS will
+respond from within the guest. This can be achieve through a `PUT` request
```
lauralt, line 61:
```suggestion
respond from within the guest. This can be achieved through a `PUT` request
```

src/api_server/src/parsed_request.rs
```diff
@@ -165,11 +166,13 @@ pub enum Error {
     InvalidPathMethod(String, Method),
     // An error occurred when deserializing the json body of a request.
     SerdeJson(serde_json::Error),
+    // An error occured while processing mmds configuration request.
+    MmdsConfig(MmdsConfigError),
```
lauralt, line 170: `MmdsConfig()` seems rather a very specific error for this enum, which should contain only generic errors for parsing requests.

src/api_server/src/request/mmds.rs
```diff
@@ -1,13 +1,53 @@
...
 pub fn parse_get_mmds() -> Result<ParsedRequest, Error> {
     Ok(ParsedRequest::GetMMDS)
 }
 
-pub fn parse_put_mmds(body: &Body) -> Result<ParsedRequest, Error> {
+pub fn parse_put_mmds(body: &Body, first_token: Option<&&str>) -> Result<ParsedRequest, Error> {
+    let mut config_request = false;
+    if let Some(config_path) = first_token {
+        if *config_path == "config" {
+            config_request = true;
```
lauralt, line 17: Hmm, why do you need the extra `config_request` bool and not include the whole logic from the next `if` block here?

src/api_server/src/request/mmds.rs
```diff
@@ -1,13 +1,53 @@
...
+    }
+
+    if config_request {
+        let mmds_config =
+            serde_json::from_slice::<MmdsConfig>(body.raw()).map_err(Error::SerdeJson)?;
+
+        let mmds_ipv4_addr_pool = mmds_config.ipv4_addr_pool();
+        if mmds_ipv4_addr_pool.is_err() {
+            // Safe, checked above.
+            return Err(Error::MmdsConfig(mmds_ipv4_addr_pool.err().unwrap()));
```
lauralt, line 33: I don't think we should wrap the error returned by `ip4_addr_pool()` in yet another MMDS specific error. The error returned by the aforementioned function should give us the necessary MMDS information and we could just wrap it in the `BadRequest` error type like you already do in the `is_empty()` scenario.
Also it looks like `ipv4_addr_pool()` doesn't have a single and clear purpose. In the context of parsing the HTTP request you are using it just to see if the `Vec` of addresses is not empty and in the vmm builder context you are actually using those addresses to configure a net device.
Also, I'm not so sure, when parsing the request body, is it okay to throw a `BadRequest` error if this `Vec` is empty? When you are attaching a net device it looks like the `empty` array scenario is accepted.
Anyway, I believe that the `ipv4_addr_pool()` logic should be a little bit rethought.
Actually, how about changing the type of `ipv4_address_pool` from `Vec<String>` to `Vec<Ipv4Addr>` (`Ipv4Addr` is a "standard" type like `PathBuf`, so it should make sense). I imagine that `serde` will take care of the parsing logic and throwing the right errors if those strings are invalid in terms of ipv4 addresses. This way, you won't need the `ipv4_addr_pool()` anymore, you can just create a getter for the `ipv4_address_pool` field, that you can call when attaching a net device. Also the whole error handling will be greatly simplified this way (no need, for example, for the `IPv4ParseError` enum variant). The code would look something like this:

request/mmds.rs:

```rust
pub fn parse_put_mmds(body: &Body, request_type: Option<&&str>) -> Result<ParsedRequest, Error> {
    match request_type {
        Some(&config_path) => match config_path {
            "config" => {
                return Ok(ParsedRequest::Sync(VmmAction::SetMmdsConfiguration(
                    serde_json::from_slice::<MmdsConfig>(body.raw()).map_err(|e| {
                        Error::SerdeJson(e)
                    })?,
                )));
            }
            _ => {
                return Err(Error::Generic(
                    StatusCode::BadRequest,
                    format!("Unrecognized PUT request path `{}`.", config_path),
                ))
            }
        },
        None => {
            return Ok(ParsedRequest::PutMMDS(
                serde_json::from_slice(body.raw()).map_err(Error::SerdeJson)?,
            ))
        }
    }
}

```
vmm_config/mmds.rs:

```rust
pub struct MmdsConfig {
    /// Pool of MMDS endpoints.
    ipv4_address_pool: Vec<Ipv4Addr>,
}

impl MmdsConfig {
    pub fn ipv4_address_poll(&self) -> Vec<Ipv4Addr> {
        self.ipv4_address_pool.clone()
    }
}
```

 What do you think? :D

src/api_server/swagger/firecracker.yaml
```diff
@@ -318,6 +318,29 @@ paths:
...
+    put:
+      summary: Set MMDS configuration.
+      description:
+        Creates a MMDS configuration to be used by the MMDS network stack.
+      parameters:
+        - name: body
+          in: body
+          description: The MMDS configuration as JSON.
+          schema:
+            type: #/definitions/MmdsConfig
```
lauralt, line 331: You have to also create the `MmdsConfig` definition. You can find all the definitions at the end of this file.

src/devices/src/virtio/net/device.rs
```diff
@@ -710,6 +712,7 @@ mod tests {
     use logger::{Metric, METRICS};
     use polly::event_manager::{EventManager, Subscriber};
     use rate_limiter::{RateLimiter, TokenBucket, TokenType};
+    use std::str::FromStr;
```
lauralt, line 715: This `use` should be placed with the other `use`s from std.

src/dumbo/src/ns.rs
```diff
@@ -104,32 +109,40 @@ impl MmdsNetworkStack {
...
-            && !test_speculative_dst_addr(src, self.ipv4_addr)
-        {
+        let mut arp_reply_ipv4_addr = None;
+        for ipv4_addr in self.ipv4_addr_pool.iter() {
+            // Check if the frame contains an ARP frame or an IPv4 packet sent to one
+            // of the MMDS server IPv4 addresses.
+            if test_speculative_tpa(src, *ipv4_addr) || test_speculative_dst_addr(src, *ipv4_addr) {
+                arp_reply_ipv4_addr = Some(*ipv4_addr);
+            }
+        }
```
lauralt, line 119: Question: Do we want here to keep only the last ipv4 address that meets one of these conditions? If yes, could we iterate the `Vec` in reverse order and add a `break` if we found a "good" `ipv4_addr`?

src/dumbo/src/ns.rs
```diff
@@ -221,10 +236,16 @@ impl MmdsNetworkStack {
...
         buf: &mut [u8],
-        dst_ipv4: Ipv4Addr,
     ) -> Result<Option<NonZeroUsize>, WriteArpFrameError> {
+        let arp_reply_info_opt = self.pending_arp_reply_dest.first();
+        if arp_reply_info_opt.is_none() {
+            return Err(WriteArpFrameError::NoPendingArpReply);
+        }
+
+        // Safe to unwrap because it is checked before.
+        let arp_reply_info = arp_reply_info_opt.unwrap();
```
lauralt, line 248: This can be replaced with:
```rust
let arp_reply_info = self
            .pending_arp_reply_dest
            .first()
            .ok_or_else(|| WriteArpFrameError::NoPendingArpReply)?;
```

src/dumbo/src/ns.rs
```diff
@@ -363,15 +389,18 @@ mod tests {
     #[test]
     #[allow(clippy::cognitive_complexity)]
     fn test_ns() {
-        let mut ns = MmdsNetworkStack::new_with_defaults();
+        let mut ns = MmdsNetworkStack::new_with_defaults(vec![Ipv4Addr::from(DEFAULT_IPV4_ADDR)]);
```
lauralt, line 392: If you want to check the default scenario, you should pass as argument an empty `Vec` here, not one that already contains the default value. Also, for testing `new_with_defaults()` function properly, we should also call this function with some non-empty `Vec<IpvdAddr>` as parameter.

src/dumbo/src/ns.rs
```diff
@@ -104,32 +109,40 @@ impl MmdsNetworkStack {
...
+        }
+
+        if arp_reply_ipv4_addr.is_none() {
             return false;
         }
 
         if let Ok(eth) = EthernetFrame::from_bytes(src) {
             match eth.ethertype() {
-                ETHERTYPE_ARP => return self.detour_arp(eth),
+                ETHERTYPE_ARP => return self.detour_arp(eth, arp_reply_ipv4_addr.unwrap()),
```
lauralt, line 127: I know it's pretty obvious that this `unwrap()` is safe but can we also add a comment about this?

iulianbarbu:
@lauralt , I've addressed your comments! PTAL.

src/vmm/src/vmm_config/mmds.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::net::Ipv4Addr;
+
+/// This struct represents the configuration realted to the MMDS service.
```
lauralt, line 6:
```suggestion
/// Keeps the IPv4 address that can be used for the MMDS configuration. 
```

src/dumbo/src/ns.rs
```diff
@@ -360,10 +369,21 @@ mod tests {
...
+        assert_eq!(ns.ipv4_addr, Ipv4Addr::LOCALHOST);
+    }
+
     #[test]
     #[allow(clippy::cognitive_complexity)]
     fn test_ns() {
-        let mut ns = MmdsNetworkStack::new_with_defaults();
+        let mut ns = MmdsNetworkStack::new_with_defaults(Some(Ipv4Addr::from(DEFAULT_IPV4_ADDR)));
         assert_eq!(ns.mac_addr, MacAddr::parse_str(DEFAULT_MAC_ADDR).unwrap());
         assert_eq!(ns.ipv4_addr, Ipv4Addr::from(DEFAULT_IPV4_ADDR));
```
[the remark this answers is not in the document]
iulianbarbu: > Are these asserts useful anymore? It looks like you are already checking these in the previous test.

They are not needed. Removed them.

>Also, I think new_with_defaults() should be called here with None parameter as calling it with the default value seems like a very unlikely scenario, what do you think?

I've used this way of initializing the `MmdsNetworkStack` because I wanted to make obvious what IPv4 address is used, to be more readable. It might be simpler instead if the code is less polluted. Simplified by calling `new_with_defaults(None)`, as you suggested. Removed also the asserts which unit test the `new_with_default` function, actually.

src/vmm/src/builder.rs
```diff
@@ -753,13 +761,19 @@ fn attach_net_devices(
             .transpose()
             .map_err(CreateRateLimiter)?;
 
+        let mmds_ipv4_addr = match mmds_config {
+            Some(cfg) => cfg.ipv4_addr(),
+            None => None,
+        };
```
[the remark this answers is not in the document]
iulianbarbu: This crashes if `mmds_config` is `None`.

src/vmm/src/vmm_config/mmds.rs
```diff
@@ -0,0 +1,20 @@
...
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct MmdsConfig {
+    /// MMDS IPv4 configured address.
+    ipv4_address: Option<Ipv4Addr>,
+}
+
+impl MmdsConfig {
+    /// Parse String to Ipv4Addr.
+    /// Error out if there parse error encountered.
```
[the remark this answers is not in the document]
iulianbarbu: Updated.

src/vmm/src/vmm_config/mmds.rs
iulianbarbu: Updated with: `/// Keeps the MMDS configuration`.

src/api_server/src/request/mmds.rs
```diff
@@ -33,20 +47,44 @@ mod tests {
...
+        assert!(parse_put_mmds(&Body::new(body), None).is_ok());
+
+        let body = "invalid_body";
+        assert!(parse_put_mmds(&Body::new(body), None).is_err());
+
+        let body = r#"{
+                "ipv4_address": "1.1.1.1"
+              }"#;
+        let path = "config";
+        assert!(parse_put_mmds(&Body::new(body), Some(&path)).is_ok());
```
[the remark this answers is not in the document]
lauralt: Yep, definitely. But before, even if you were using the same value for the `path` variable, you were reinitializing it before every check :(. I should've probably suggested to remove the second initialization.

src/vmm/src/builder.rs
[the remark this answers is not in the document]
lauralt: True, sorry, I should've double-checked.

src/devices/src/virtio/net/device.rs
```diff
@@ -98,6 +99,7 @@ impl Net {
         guest_mac: Option<&MacAddr>,
         rx_rate_limiter: RateLimiter,
         tx_rate_limiter: RateLimiter,
+        mmds_ipv4_addr: Option<Ipv4Addr>,
         allow_mmds_requests: bool,
```
acatangiu, line 103: `mmds_ipv4_addr` doesn't make sense when `allow_mmds_requests == false`.

The two args can be combined in a single enum, smth like:
```rust
enum MmdsConfig {
    Disabled,
    EnabledDefaultIp,
    EnabledCustomIp(Ipv4Addr),
}
```

acatangiu: Another option would be to just pass Option<MmdsNetworkStack> and create that during the building phase instead of passing configuration and having the network device build it.

src/dumbo/src/ns.rs
```diff
@@ -139,35 +139,38 @@ impl MmdsNetworkStack {
...
-                                    METRICS.mmds.connections_created.inc();
-                                    METRICS.mmds.connections_destroyed.inc();
-                                }
-                                RecvEvent::EndpointDone => {
-                                    METRICS.mmds.connections_destroyed.inc();
-                                }
-                                _ => (),
+            if ip.protocol() == PROTOCOL_TCP {
+                // Note-1: `remote_mac_address` is actually the network device mac address, on which
+                // this TCP segment came from.
```
acatangiu, line 144: Confusing phrasing, suggest
```suggestion
                // Note-1: `remote_mac_address` is actually the network device mac address,
                // where this TCP segment came from.
```

src/dumbo/src/ns.rs
```diff
@@ -139,35 +139,38 @@ impl MmdsNetworkStack {
...
+                // Note-2: For every routed packet we will have a single source MAC address, because
+                // the MmdsNetworkStack routes packets for only one network device.
```
acatangiu, line 146:
```suggestion
                // each MmdsNetworkStack routes packets for only one network device.
```

src/api_server/swagger/firecracker.yaml
```diff
@@ -664,6 +688,14 @@ definitions:
         description: The amount of milliseconds it takes for the bucket to refill.
         minimum: 0
 
+  MmdsConfig:
```
lauralt, line 691: Hmm, what I was trying to say here by alphabetical order was to put this definition between `Metrics` and `NetworkInterface`. Sorry for not being clearer.

src/vmm/src/resources.rs
```diff
@@ -111,6 +116,10 @@ impl VmResources {
         if let Some(vsock_config) = vmm_config.vsock_device {
             resources.set_vsock_device(vsock_config);
         }
+
+        if let Some(mmds_config) = vmm_config.mmds_config {
+            resources.set_mmds_config(mmds_config);
+        }
```
lauralt, line 122: Can you also test this in `test_from_json()` by creating a new json with, for example, an invalid `ipv4_address` value and checking the returned error?

src/vmm/src/vmm_config/mmds.rs
```diff
@@ -0,0 +1,20 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
lauralt, line 1:
```suggestion
// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

src/vmm/src/vmm_config/mmds.rs
```diff
@@ -0,0 +1,20 @@
...
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct MmdsConfig {
+    /// MMDS IPv4 configured address.
+    ipv4_address: Option<Ipv4Addr>,
+}
+
+impl MmdsConfig {
+    /// MMDS IPv4 address getter. Since IPv4 address configuration
+    /// is optional, returns an `Option<Ipv4Addr>`.
```
lauralt, line 16:
```suggestion
    /// Returns the MMDS IPv4 address if one was configured.
    /// Otherwise returns None.
```

iulianbarbu:
@lauralt , @acatangiu , PTAL! I've addressed your comments.

src/devices/src/virtio/net/device.rs
iulianbarbu: As a matter of fact, this issue describes on the same lines but shorter what I have in mind: https://github.com/firecracker-microvm/firecracker/issues/1761.

src/devices/src/virtio/net/device.rs
```diff
@@ -185,6 +184,11 @@ impl Net {
         self.guest_mac.as_ref()
     }
 
+    /// Provides a mutable reference to the `MmdsNetworkStack`.
+    pub fn mmds_ns(&mut self) -> Option<&mut MmdsNetworkStack> {
```
acatangiu, line 188: nit:
```suggestion
    pub fn mmds_ns_mut(&mut self) -> Option<&mut MmdsNetworkStack> {
```

src/vmm/src/builder.rs
```diff
@@ -723,16 +729,26 @@ fn attach_block_devices(
...
+        match mmds_config {
+            None => (),
+            Some(cfg) => {
+                cfg.ipv4_addr().map_or((), |ipv4_addr| {
+                    if let Some(mmds_ns) = net_device.lock().unwrap().mmds_ns() {
+                        mmds_ns.set_ipv4_addr(ipv4_addr);
+                    };
+                });
+            }
+        };
```
acatangiu, line 751: I think this would be better placed in the `resources` part instead of here since:

- it's not really part of _attaching_ a `Net` device to a `Vmm`.
- locality is better if you set the data as soon as you get it instead of deferring to a later step

What do you think?

src/vmm/src/vmm_config/mmds.rs
```diff
@@ -0,0 +1,27 @@
...
+
+impl MmdsConfig {
+    /// Returns the MMDS IPv4 address if one was configured.
+    /// Otherwise returns None.
+    pub fn ipv4_addr(&self) -> Option<Ipv4Addr> {
+        self.ipv4_address
+    }
+
+    #[cfg(test)]
+    pub fn new_with_ipv4_addr(ipv4_addr: Ipv4Addr) -> Self {
```
lauralt, line 22: Will this function be used sometime in the future? At this moment, I can't find any usage of it.

src/vmm/src/resources.rs
```diff
@@ -221,18 +230,41 @@ impl VmResources {
...
+    /// Settter for mmds config.
+    pub fn set_mmds_config(&mut self, config: MmdsConfig) {
+        // Update existing built network device `MmdsNetworkStack` IPv4 address.
+        for net_device in self.net_builder.iter_mut() {
+            config.ipv4_addr().map_or((), |ipv4_addr| {
+                if let Some(mmds_ns) = net_device.lock().unwrap().mmds_ns_mut() {
+                    mmds_ns.set_ipv4_addr(ipv4_addr);
+                };
+            });
+        }
```
acatangiu, line 265: The API currently allows setting MMDS ipv4 to `None` aka use default, need to support setting it back to `None` as well.

```suggestion
        for net_device in self.net_builder.iter_mut() {
            if let Some(mmds_ns) = net_device.lock().unwrap().mmds_ns_mut() {
                mmds_ns.set_ipv4_addr(ipv4_addr);
            }
        }
```

src/vmm/src/vmm_config/net.rs
```diff
@@ -110,34 +110,33 @@ impl fmt::Display for NetworkInterfaceError {
...
+    /// Returns a mutable iterator over the network devices.
     pub fn iter_mut(&mut self) -> ::std::slice::IterMut<Arc<Mutex<Net>>> {
-        self.if_list.iter_mut()
+        self.net_devices.iter_mut()
     }
 
-    /// Inserts `netif_config` in the network interfaces list.
-    /// If an entry with the same id already exists, it will overwrite it.
-    pub fn insert(&mut self, netif_config: NetworkInterfaceConfig) -> Result<()> {
+    /// Builds a network device based on a network interface config.
```
acatangiu, line 138:
```suggestion
    /// Builds a network device based on a network interface config. Keeps a device reference
    /// in the builder's internal list.
```

src/vmm/src/builder.rs
```diff
@@ -216,7 +216,7 @@ impl VmmEventsObserver for SerialStdin {
 /// An `Arc` reference of the built `Vmm` is also plugged in the `EventManager`, while another
 /// is returned.
 pub fn build_microvm(
-    vm_resources: &super::resources::VmResources,
+    vm_resources: &mut super::resources::VmResources,
```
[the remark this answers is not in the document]
iulianbarbu: Nope.

src/api_server/swagger/firecracker.yaml
```diff
@@ -317,6 +317,30 @@ paths:
           schema:
             $ref: "#/definitions/Error"
 
+  /mmds/config:
+    put:
+      summary: Set MMDS configuration.
```
acatangiu, line 322: Be explicit about restrictions on the API call.
```suggestion
      summary: Set MMDS configuration. Pre-boot only.
```

src/vmm/src/resources.rs
```diff
@@ -574,19 +609,59 @@ mod tests {
                     "network-interfaces": [
                         {{
                             "iface_id": "netif",
-                            "host_dev_name": "hostname8"
+                            "host_dev_name": "hostname8",
+			    "allow_mmds_requests": true
```
acatangiu, line 613: nit:
```suggestion
                            "host_dev_name": "hostname8",
                            "allow_mmds_requests": true
```

src/vmm/src/resources.rs
```diff
@@ -574,19 +609,59 @@ mod tests {
...
-                            "ht_enabled": false
-                     }}
+                    "machine-config": {{
+                        "vcpu_count": 2,
+                        "mem_size_mib": 1024,
+                        "ht_enabled": false
+                     }},
+     	             "mmds-config": {{
+	                 "ipv4_address": "169.254.170.2"
+	             }}
```
acatangiu, line 623: alignment nit:
```suggestion
                    "machine-config": {{
                        "vcpu_count": 2,
                        "mem_size_mib": 1024,
                        "ht_enabled": false
                     }},
                     "mmds-config": {{
                         "ipv4_address": "169.254.170.2"
                     }}
```

src/vmm/src/resources.rs
```diff
@@ -574,19 +609,59 @@ mod tests {
...
+			    "allow_mmds_requests": true
+                        }}
+                    ],
+                    "machine-config": {{
+                        "vcpu_count": 2,
+                        "mem_size_mib": 1024,
+                        "ht_enabled": false
+                     }},
+     	             "mmds-config": {{
+	             }}
```
acatangiu, line 660: nit: same here, these lines have some TABs as whitespaces, and this messes up alignment

iulianbarbu:
@acatan, @lauralt , I've addressed your comments! Please, take a look again.

PR #1687, opened 2020-03-19
Align device events registration after device activation

### LINKED ISSUE #1639: Align device events registration after device activation
Currently, `vosck` device implements a mechanism of registering device events after the device activation. `net` and `block` devices are registering the device events at creation time, which entail handling of device events when the device is not ready to do so.

We need the natural way of handling events, just after the device has been activated. Align `net` and `block` devices with the `vsock` implementation.

acatangiu:
## Reason for This PR

Fixes #1639 

### Device access to guest memory on/after activation

Passing memory during device activation instead of during construction enables the following story: #1702 #1708 #1709 and finally #1713

The changes in this PR are in line with the model currently defined in https://github.com/rust-vmm/vm-virtio.
They are also compatible with the https://github.com/rust-vmm/vm-virtio/pull/10 proposal where memory is also passed to the device during activation:

```rust
    /// Associated guest memory
    type M: GuestMemory;

    ...

    /// Activates this device for real usage.
    /// The ownership of the VirtioDeviceConfig object moves into the VirtioDevice object if
    /// activate succeeds, otherwise it should return the ownership back.
    fn activate(&mut self, config: VirtioDeviceConfig<M>) -> ActivateResult<VirtioDeviceConfig<M>>;
```

### Events registration during activate

Postpone external events registration to device activation time.
This makes the block and net devices unaware of external events prior to their activation.

During creation register a dedicated activation event which will notify the device when it's time to register the other external events sources.

This activation event is unregistered after successful device activation.

Coverage slightly decreased because untestable EventFd error cases.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -21,7 +21,7 @@ use virtio_gen::virtio_blk::*;
 use vm_memory::{Bytes, GuestMemoryMmap};
 
 use super::{
-    super::{ActivateResult, Queue, VirtioDevice, TYPE_BLOCK, VIRTIO_MMIO_INT_VRING},
+    super::{ActivateResult, DeviceState, Queue, VirtioDevice, TYPE_BLOCK, VIRTIO_MMIO_INT_VRING},
     request::*,
     Error, CONFIG_SPACE_SIZE, QUEUE_SIZES, SECTOR_SHIFT, SECTOR_SIZE,
 };
@@ -88,15 +88,15 @@ pub struct Block {
     avail_features: u64,
     acked_features: u64,
     config_space: Vec<u8>,
+    pub(crate) activate_evt: EventFd,
 
     // Transport related fields.
     queues: Vec<Queue>,
     interrupt_status: Arc<AtomicUsize>,
     interrupt_evt: EventFd,
     pub(crate) queue_evts: [EventFd; 1],
-    mem: GuestMemoryMmap,
 
-    device_activated: bool,
+    pub(crate) device_state: DeviceState,
 
     // Implementation specific fields.
     pub(crate) rate_limiter: RateLimiter,
@@ -107,7 +107,6 @@ impl Block {
     ///
     /// The given file must be seekable and sizable.
     pub fn new(
-        mem: GuestMemoryMmap,
         mut disk_image: File,
         is_disk_read_only: bool,
         rate_limiter: RateLimiter,
@@ -132,12 +131,12 @@ impl Block {
             acked_features: 0u64,
             config_space: build_config_space(disk_size),
             rate_limiter,
-            mem,
             interrupt_status: Arc::new(AtomicUsize::new(0)),
             interrupt_evt: EventFd::new(libc::EFD_NONBLOCK)?,
             queue_evts,
             queues,
-            device_activated: false,
+            device_state: DeviceState::Inactive,
+            activate_evt: EventFd::new(libc::EFD_NONBLOCK)?,
         })
     }
 
@@ -161,11 +160,16 @@ impl Block {
     }
 
     pub(crate) fn process_queue(&mut self, queue_index: usize) -> bool {
+        let mem = match self.device_state {
+            DeviceState::Activated(ref mem) => mem,
+            // This should never happen, it's been already validated in the event handler.
+            DeviceState::Inactive => unreachable!(),
+        };
         let queue = &mut self.queues[queue_index];
         let mut used_any = false;
-        while let Some(head) = queue.pop(&self.mem) {
+        while let Some(head) = queue.pop(mem) {
             let len;
-            match Request::parse(&head, &self.mem) {
+            match Request::parse(&head, mem) {
                 Ok(request) => {
                     // If limiter.consume() fails it means there is no more TokenType::Ops
                     // budget and rate limiting is in effect.
@@ -196,7 +200,7 @@ impl Block {
                     let status = match request.execute(
                         &mut self.disk_image,
                         self.disk_nsectors,
-                        &self.mem,
+                        mem,
                         &self.disk_image_id,
                     ) {
                         Ok(l) => {
@@ -212,7 +216,7 @@ impl Block {
                     };
                     // We use unwrap because the request parsing process already checked that the
                     // status_addr was valid.
-                    self.mem.write_obj(status, request.status_addr).unwrap();
+                    mem.write_obj(status, request.status_addr).unwrap();
                 }
                 Err(e) => {
                     error!("Failed to parse available descriptor chain: {:?}", e);
@@ -220,7 +224,7 @@ impl Block {
                     len = 0;
                 }
             }
-            queue.add_used(&self.mem, head.index, len);
+            queue.add_used(mem, head.index, len);
             used_any = true;
         }
 
@@ -318,17 +322,24 @@ impl VirtioDevice for Block {
     }
 
     fn is_activated(&self) -> bool {
-        self.device_activated
+        match self.device_state {
+            DeviceState::Inactive => false,
+            DeviceState::Activated(_) => true,
+        }
     }
 
-    fn activate(&mut self, _: GuestMemoryMmap) -> ActivateResult {
-        self.device_activated = true;
+    fn activate(&mut self, mem: GuestMemoryMmap) -> ActivateResult {
+        if self.activate_evt.write(1).is_err() {
+            error!("Block: Cannot write to activate_evt");
+            return Err(super::super::ActivateError::BadActivate);
+        }
+        self.device_state = DeviceState::Activated(mem);
         Ok(())
     }
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use std::fs::metadata;
     use std::os::unix::io::AsRawFd;
     use std::thread;
@@ -352,7 +363,7 @@ mod tests {
     }
 
     impl Block {
-        fn set_queue(&mut self, idx: usize, q: Queue) {
+        pub(crate) fn set_queue(&mut self, idx: usize, q: Queue) {
             self.queues[idx] = q;
         }
 
@@ -366,7 +377,7 @@ mod tests {
     }
 
     /// Create a default Block instance to be used in tests.
-    fn default_block() -> Block {
+    pub fn default_block() -> Block {
         // Create backing file.
         let f = TempFile::new().unwrap();
         let block_file = f.into_file();
@@ -375,12 +386,14 @@ mod tests {
         // Rate limiting is enabled but with a high operation rate (10 million ops/s).
         let rate_limiter = RateLimiter::new(0, None, 0, 100_000, None, 10).unwrap();
 
-        let mem = GuestMemoryMmap::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap();
-
-        Block::new(mem, block_file, true, rate_limiter).unwrap()
+        Block::new(block_file, true, rate_limiter).unwrap()
     }
 
-    fn initialize_virtqueue(vq: &VirtQueue) {
+    pub fn default_mem() -> GuestMemoryMmap {
+        GuestMemoryMmap::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap()
+    }
+
+    pub fn initialize_virtqueue(vq: &VirtQueue) {
         let request_type_desc: usize = 0;
         let data_desc: usize = 1;
         let status_desc: usize = 2;
@@ -496,7 +509,7 @@ mod tests {
     #[test]
     fn test_invalid_request() {
         let mut block = default_block();
-        let mem = block.mem.clone();
+        let mem = default_mem();
         let vq = VirtQueue::new(GuestAddress(0), &mem, 16);
         block.set_queue(0, vq.create_queue());
         block.activate(mem.clone()).unwrap();
@@ -521,7 +534,7 @@ mod tests {
     #[test]
     fn test_request_execute_failures() {
         let mut block = default_block();
-        let mem = block.mem.clone();
+        let mem = default_mem();
         let vq = VirtQueue::new(GuestAddress(0), &mem, 16);
         block.set_queue(0, vq.create_queue());
         block.activate(mem.clone()).unwrap();
@@ -579,7 +592,7 @@ mod tests {
     #[test]
     fn test_unsupported_request_type() {
         let mut block = default_block();
-        let mem = block.mem.clone();
+        let mem = default_mem();
         let vq = VirtQueue::new(GuestAddress(0), &mem, 16);
         block.set_queue(0, vq.create_queue());
         block.activate(mem.clone()).unwrap();
@@ -609,7 +622,7 @@ mod tests {
     #[test]
     fn test_read_write() {
         let mut block = default_block();
-        let mem = block.mem.clone();
+        let mem = default_mem();
         let vq = VirtQueue::new(GuestAddress(0), &mem, 16);
         block.set_queue(0, vq.create_queue());
         block.activate(mem.clone()).unwrap();
@@ -668,7 +681,7 @@ mod tests {
     #[test]
     fn test_flush() {
         let mut block = default_block();
-        let mem = block.mem.clone();
+        let mem = default_mem();
         let vq = VirtQueue::new(GuestAddress(0), &mem, 16);
         block.set_queue(0, vq.create_queue());
         block.activate(mem.clone()).unwrap();
@@ -711,7 +724,7 @@ mod tests {
     #[test]
     fn test_get_device_id() {
         let mut block = default_block();
-        let mem = block.mem.clone();
+        let mem = default_mem();
         let vq = VirtQueue::new(GuestAddress(0), &mem, 16);
         block.set_queue(0, vq.create_queue());
         block.activate(mem.clone()).unwrap();
@@ -777,7 +790,7 @@ mod tests {
     #[test]
     fn test_bandwidth_rate_limiter() {
         let mut block = default_block();
-        let mem = block.mem.clone();
+        let mem = default_mem();
         let vq = VirtQueue::new(GuestAddress(0), &mem, 16);
         block.set_queue(0, vq.create_queue());
         block.activate(mem.clone()).unwrap();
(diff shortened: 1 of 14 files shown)


acatangiu:
Still need to add unit tests for the new code.

acatangiu:
Tests added, PR is complete.

src/devices/src/virtio/vsock/device.rs
```diff
@@ -131,10 +128,15 @@ where
     /// otherwise.
     pub fn process_rx(&mut self) -> bool {
         debug!("vsock: process_rx()");
+        let mem = match self.device_state {
```
andreeaflorescu: I think it was more efficient how it was written before with device_state being a simple bool. Processing the event is on the critical path and it is more efficient to just check a flag than to do this match to get the memory.

Any reason for having the memory in the enum?

src/devices/src/virtio/net/event_handler.rs
```diff
@@ -8,15 +8,91 @@ use polly::event_manager::{EventManager, Subscriber};
...
+            .unwrap();
+
+        event_manager
+            .register(
+                self.queue_evts[RX_INDEX].as_raw_fd(),
+                EpollEvent::new(EventSet::IN, self.queue_evts[RX_INDEX].as_raw_fd() as u64),
+                self_subscriber.clone(),
+            )
+            .unwrap_or_else(|e| {
+                error!(
```
andreeaflorescu: Do we want to continue the execution if we cannot register the events needed for this device? It might lead to silent failures.

src/devices/src/virtio/block/event_handler.rs
```diff
@@ -27,25 +58,111 @@ impl Subscriber for Block {
...
+                    _ if queue_evt == source => self.process_queue_event(),
+                    _ if rate_limiter_evt == source => self.process_rate_limiter_event(),
+                    _ if activate_fd == source => self.process_activate_event(evmgr),
+                    _ => warn!("Block: Spurious event received: {:?}", source),
+                }
+            }
+            DeviceState::Inactive => warn!(
+                "Block: The device is not yet activated. Spurious event received: {:?}",
+                source
+            ),
```
andreeaflorescu: This error case can be tested and should probably be tested. We can check that we do not trigger `process_*_event` functions when the device is not activated. This applies to both block and net devices. It looks like the activate for vsock device is not tested at all, so maybe we should open an issue for that one.

src/devices/src/virtio/block/event_handler.rs
```diff
@@ -27,25 +58,111 @@ impl Subscriber for Block {
...
+
+    use super::*;
+    use crate::virtio::block::device::tests::*;
+    use crate::virtio::device::VirtioDevice;
+    use crate::virtio::queue::tests::*;
+    use virtio_gen::virtio_blk::*;
+    use vm_memory::{Bytes, GuestAddress};
+
+    #[test]
+    fn test_interest_list() {
```
andreeaflorescu: This test does not bring much value because `interest_list` just returns a static vector created in the function body. To test this basically means to duplicate the code in the test. This is typically a bad engineering practice referred to as WET. You can read more about it here: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

src/devices/src/virtio/device.rs
```diff
@@ -10,6 +10,14 @@ use std::sync::{atomic::AtomicUsize, Arc};
 use super::{ActivateResult, Queue};
 use crate::virtio::AsAny;
 use utils::eventfd::EventFd;
+use vm_memory::GuestMemoryMmap;
+
+/// Enum that indicates if a VirtioDevice is inactive or has been activated
+/// and memory attached to it.
+pub enum DeviceState {
+    Inactive,
+    Activated(GuestMemoryMmap),
```
andreeaflorescu: Having the `GuestMemoryMmap` here doesn't seem very flexible. If we want to have another state that also just applies to activated devices, we would need to convert this to a tuple. The code becomes harder to read then. As we are thinking about implementing reset on devices, the time where extending Activated is not that far because on reset we need to release some other resources.

src/devices/src/virtio/block/event_handler.rs
acatangiu: Sounds good. Will add a test for this error case.

src/devices/src/virtio/device.rs
acatangiu: Each state variant can have its own data (including overlapping data) so we wouldn't use tuples, just another state with or without `mem`.

Regarding `reset` support, the changes in this PR should help in that scenario.

Adding `mem` to `enum DeviceState` was just a solution that felt natural to me, but which I'm not hell-bent on doing.

The problem which this PR is trying to solve is allowing devices to be created without `guest memory`. Memory will be added during device activation. We need *some way* to allow this plugging of memory after creation.

One solution is `Option<Memory>`, but
```rust
struct SomeDevice {
    ...
    is_active: bool,
    mem: Option<Memory>,
    ...
}
```
is C-style code where you have an invisible error-prone connection between the two fields.

In Rust you can enforce that connection at compile time using an enum like the one in the PR.

```rust
enum DevState {
    Inactive,
    Active(Memory),
};
struct SomeDevice {
    ...
    state: DevState,
    ...
}
```

I'm open to alternatives.

src/devices/src/virtio/net/event_handler.rs
acatangiu: I believe that once a guest is booted (customer workload has possibly started) we should always avoid crashing.

A malfunctioning device has the blast-radius limited to operations on that device. The guest workload might or might not be affected and data might be recoverable if the malfunction is detected.

src/devices/src/virtio/vsock/device.rs
acatangiu: The previous version of the code had a single check of the `is_active: bool` per event handle while the code in this PR has one or sometimes two (like the case in this line) checks of the `device_state: enum DeviceState` per event handle.

`match enum { ... }` and `if boolean { ... }` are equivalent in terms of efficiency (it's still compare and jump instructions).

The extra `if` per event handle is inconsequential to this code's performance where the amount of work done for an event is substantial.

Even so, I am open to alternatives. The problem trying to be solved here is allowing device creation without memory and having memory given to the device some time later (during activation for example). Better ideas to support that are welcome! 😃

acatangiu: > Any reason for having the memory in the enum?

See this https://github.com/firecracker-microvm/firecracker/pull/1687#discussion_r397885208 for a more detailed explanation.

src/devices/src/virtio/block/event_handler.rs
andreeaflorescu: I agree we should be consistent. If people from the team agree with this, we shouldn't introduce this test in this PR just for the sake of consistency though :D

src/devices/src/virtio/device.rs
andreeaflorescu: I was discussing about this with @dhrgit at some point. Hopefully the following lines are a good summary of our discussion.

Devices should always be initialized after the memory is initialized because this is also how it works in a real world scenario. When you power on a machine, the memory is initialized, then the devices use the memory to initialize themselves. This would be mapped in code with the device constructor always receiving a reference to the memory. This way you also break the dependency between `mem` and `is_active`. The device cannot exist without the memory, so the `mem` field will always be `Memory` and not `Option<Memory>`. The activation can happen later in the logic of the device and does not have any impact on the memory.

acatangiu: There were also discussions with both @dhrgit and @alexandruag where we were comparing a device with a physical device which is first built (like in a factory in China 😆 ) and at some later point is, like you say initialized when it's plugged in a machine. The device can surely exist without `memory` or anything else that is outside of it.

Where I am trying to get with this change and some following ones pending on this, is a place where a device can be _built_ during it's configuration phase as opposed to being built during boot. Booting a microVM then becomes a matter of plugging all things together and hitting the switch.

This model allows user-errors (like invalid resources or permissions) to be reported as part of the configuration step rather than later at attempted boot.
It also removes the time window between configuration and boot where changes to the host system (creating Taps, changing permissions, etc) could cause microVM boot failures (race condition between validation and commitment of resources).

Such a model should also decreases code complexity and should provide better efficiency to the process of configuring and starting a microVM (no more moving configurations around in memory and validating resources both at config time and boot-time).

andreeaflorescu: I am not necessarily against that model. But you can have both of them by introducing a restriction at the API level that the memory needs to be configured before you configure the devices.

Indeed, not all devices need memory, but network, block and vsock do need it and these are the ones we are trying to model now.

acatangiu: That is a valid alternative!

Though, I personally think that such a limitation at the API level increases UX complexity as the user doesn't naturally understand such an ordering requirement (and it gets even more confusing when supporting default values for memory like we do), while the current proposal has no disadvantages AFAICT.

If you like the API-level restriction better we can set up a meeting with the maintainers, get busy with the whiteboard and commit to an option. Let me know.

andreeaflorescu: Yap. I would prefer a meeting because this approach is somewhat going in a very different direction compared to the rust-vmm approach.

andreeaflorescu: This is the interface that is currently merged in rust-vmm: https://github.com/rust-vmm/vm-virtio/blob/master/src/device.rs

It will probably suffer some changes though once we add more functionality to the crate.

acatangiu: > this approach is somewhat going in a very different direction compared to the rust-vmm approach

How so? The rust-vmm device interface _already_ has memory passed to the device during activation (exactly what this PR proposes).

andreeaflorescu: That is a temporary state because the code was very similar to what we had in Firecracker before. You can take a look at what adjustment are in the pipeline: https://github.com/rust-vmm/vm-virtio/pull/10 and https://github.com/rust-vmm/vm-virtio/pull/13.

When we discussed about unifying the activation path I though we will apply the same mechanism from vsock without introducing other changes, so I am a bit unprepared for a design discussion and I would need more time to think about this. I personally prefer the vsock design for now. This can be improved in the future, but it probably needs more investment. I find it more clear for `device_activated` to be a bool and the memory to be directly accessible from a field (it doesn't need to be field of the VirtioDevice, we can have it in a struct if we find something that makes sense) and not from an enum variant. I also strongly believe that virtio devices should not be initialized without a reference to the memory.

It looks like this PR is in preparation for a bigger change that is yet to come, so the full picture is needed to understand if this change is indeed need in its current form. Of course, we wouldn't need to block on the next PR availability, but a few paragraphs on the overall design would be helpful.

Of course all my rambling can come off as a personal preference, so I think it is better to leave it at the submitter choice because otherwise is going to be a too long discussion on something that can be updated later if we realize that's not the right direction.

acatangiu: > You can take a look at what adjustment are in the pipeline: rust-vmm/vm-virtio#10 and rust-vmm/vm-virtio#13.

Both of these are compatible with the model proposed in this PR. In https://github.com/rust-vmm/vm-virtio/pull/10 you can even see that the `VirtioDevice` trait defines activation as:

```rust
    /// Associated guest memory
    type M: GuestMemory;

    ...

    /// Activates this device for real usage.
    /// The ownership of the VirtioDeviceConfig object moves into the VirtioDevice object if
    /// activate succeeds, otherwise it should return the ownership back.
    fn activate(&mut self, config: VirtioDeviceConfig<M>) -> ActivateResult<VirtioDeviceConfig<M>>;
```

So the model is to pass virtio device configuration (including memory - the `M` in `VirtioDeviceConfig`) to the device during its activation.

This is a sane model. Most of the configuration is not available during device creation, some of it can be dynamic (like in the case of PCI). 

-----------------

> I also strongly believe that virtio devices should not be initialized without a reference to the memory.

Can you provide some supporting data/theories for that?

-----------------

> It looks like this PR is in preparation for a bigger change that is yet to come, so the full picture is needed to understand if this change is indeed need in its current form.

That is right, it is required, although not necessarily in its current form, for the following story: https://github.com/firecracker-microvm/firecracker/issues/1702 https://github.com/firecracker-microvm/firecracker/issues/1708 https://github.com/firecracker-microvm/firecracker/issues/1709 and finally https://github.com/firecracker-microvm/firecracker/issues/1713

The end result is reaching a state where our components are loose enough so we can easily swap them for [rust-vmm](https://github.com/rust-vmm) ones.

andreeaflorescu: I looked again at the code, my bad. I thought that `DeviceState` leaks in the interface and that would've been a problem. As long as `DeviceState` is not leaking in the interface, we can easily replace this abstraction with something else in the future (for example when consuming virito devices from rust-vmm).

Nevertheless, I don't think that `DeviceState` is the right abstraction as it complicates the code (with the match in process and deciding whether a device is activated or not). I still don't understand why not use the vsock design, but again it is too long of a conversation for a few lines of code.

Regarding the virtio devices requiring memory to be created, we are using static devices, and no PCI. So for us these resources can be created after the memory is initialized, which would simplify things because you don't need an Option<Memory> or the memory to be obfuscated behind the DeviceState which again would make access to the memory very simple instead of a match.

alexandruag:
Hi everyone! Taking a step back, it seems a lot complexity stems from the multi-step Firecracker configuration process. Is there any reason for devices, memory, mmds, etc. to be configured via separate API calls at this point? If the API only allowed the "one call" configuration we talked about a while ago (or, to push things even further, Firecracker is always started using a config file and only keeps the API for runtime PATCHes, etc), quite a lot of logic gets eliminated and we get configuration information upfront. This is a breaking change, but it seems worthwhile to seriously consider sooner rather than later.

Also wanted to mention the `VirtioDevice` interface from rust-vmm is still mostly based on the old crosvm/Firecracker code, so unfortunately it's unclear at this point how it will end up looking like. FWIW, not immediately going back to tying device memory configuration to activation might be better in terms of keeping options open.

acatangiu:
I agree with @alexandruag that the current multi-step configuration process introduces a complexity cost and imposes limitations on the Firecracker design that far outweigh any benefits it brings in a production scenario.

Unfortunately, like Alex mentioned, removing the multi-step configuration is a breaking change in terms of API and usage patterns and needs research and buy-in from customers/users of Firecracker before we can do it.

In the meantime, we should move forward with the changes in this PR as their blast-radius is very small and reverting the way memory is passed to the device can be easily done in the future. By that time we might even have an agreed-upon interface in rust-vmm that we can also adhere to.

src/devices/src/virtio/block/event_handler.rs
acatangiu: Added a test for the vsock device event handler.

Enhanced handler tests for all devices to validate correct events handling for both pre- and post-activation.

src/devices/src/virtio/block/event_handler.rs
acatangiu: I've removed the WET tests.

src/devices/src/virtio/vsock/event_handler.rs
```diff
@@ -191,24 +192,32 @@ where
...
 
-        let mut raise_irq = false;
-
-        match source {
-            _ if source == rxq => raise_irq = self.handle_rxq_event(event),
-            _ if source == txq => raise_irq = self.handle_txq_event(event),
-            _ if source == evq => raise_irq = self.handle_evq_event(event),
-            _ if source == backend => {
-                raise_irq = self.notify_backend(event);
+        match self.device_state {
```
andreeaflorescu: Shouldn't this be using `is_activated` function instead?

src/devices/src/virtio/block/device.rs
```diff
@@ -161,11 +160,16 @@ impl Block {
     }
 
     pub(crate) fn process_queue(&mut self, queue_index: usize) -> bool {
+        let mem = match self.device_state {
+            DeviceState::Activated(ref mem) => mem,
+            // This should never happen, it's been already validated in the event handler.
```
andreeaflorescu: If this should never happen, should we use a panic instead to avoid programming errors?

src/devices/src/virtio/vsock/event_handler.rs
```diff
@@ -547,4 +569,70 @@ mod tests {
...
+            guest_txvq,
+            ..
+        } = test_ctx.create_event_handler_context();
+
+        let vsock = Arc::new(Mutex::new(device));
+        event_manager.add_subscriber(vsock.clone()).unwrap();
+
+        // Push a queue event
+        // - the driver has something to send (there's data in the TX queue); and
+        // - the backend has no pending RX data.
```
iulianbarbu: This comment says that the backend has no pending RX data, but in the code block below, you do:
`device.backend.set_pending_rx(true)`.

src/devices/src/virtio/block/event_handler.rs
```diff
@@ -27,25 +58,120 @@ impl Subscriber for Block {
...
+                }
+            }
+            DeviceState::Inactive => warn!(
+                "Block: The device is not yet activated. Spurious event received: {:?}",
+                source
+            ),
+        };
     }
 
     // Returns the rate_limiter and queue event fds.
```
ioanachirca:
```suggestion
    // Returns the activate event fd.
```

ioanachirca: Could also remove the comment, as it's not essential and requires maintenance with each change of the function's meaning.

src/devices/src/virtio/block/device.rs
acatangiu: I'm usually very apprehensive of adding crash conditions because of the risk of crashing in production.

In this area however, we have a lot of unit-tests so any future programming errors will be better caught with a panic as you say. Also because of the high-degree of coverage, I believe it is safe to say that we shouldn't crash in prod while tests are passing.

Per your suggestion, I've modified all instances of this check to have `unreachable!()` on the _unreachable_ path.

src/devices/src/virtio/vsock/event_handler.rs
acatangiu: You're right, I had updated the test but forgot to update the comment.
Fixed.

src/devices/src/virtio/vsock/event_handler.rs
acatangiu: Sure. I've made the changes so all devices now use `is_activated()` in their event handler.

src/devices/src/virtio/block/event_handler.rs
acatangiu: Done.

PR #1692, opened 2020-03-20
tests: set logger for every spawned microvm

lauralt:
## Reason for This PR

We want to have a better monitoring of the console output using our logging system.

## Description of Changes

Configured the logger in every integration test that spawns a microvm in order to always check that the Firecracker process has indeed started. The logger is configured in test_logging.py
still via API requests and in the other tests via command line parameter. We create a new thread that redirects the console output to the logging file for a better monitoring of the errors that may appear.

## License Acceptance

## PR Checklist


tests/framework/microvm.py
```diff
@@ -371,12 +399,24 @@ def spawn(self):
...
             self._wait_create()
+        if create_logger:
+            self._wait_for_message("Running Firecracker", 0.5)
 
     @retry(delay=0.2, tries=5)
     def _wait_create(self):
         """Wait until the API socket and chroot folder are available."""
         os.stat(self._jailer.api_socket_path())
 
+    def _wait_for_message(self, message, timeout):
```
dianpopa: We could make use of the retry decorator here too. The retry decorator will re-run the function 3 times with a delay of 0.5 seconds upon catching any exception (in our case the exception will be AssertionError):
```
@retry(tries=3, delay=0.5)
def check_log_message:
'''Retry on AssertionError, raise error after 3 attempts, sleep 0.5 seconds between attempts.'''
assert message in self._log_data
```
This is just an example, you can tweak the parameters. WDYT?

tests/framework/microvm.py
```diff
@@ -497,6 +537,32 @@ def start(self):
...
+        """
+        def monitor_fd(microvm, path):
+            try:
+                fd = open(path, "r")
+                while True:
+                    if microvm.logging_thread.stopped():
+                        return
+                    data = fd.readline()
+                    if data:
+                        microvm.append_to_log_data(data)
```
dianpopa: With the @setter decorator this will become: microvm.log_data = data

tests/integration_tests/functional/test_signals.py
```diff
@@ -26,33 +26,21 @@ def test_sigbus_sigsegv(test_microvm_with_api, signum):
...
 
     msg = 'Shutting down VM after intercepting signal {}'.format(signum)
-    log_fifo.flags = log_fifo.flags & ~os.O_NONBLOCK
-    lines = log_fifo.sequential_reader(5)
 
-    msg_found = False
-    for line in lines:
-        if msg in line:
-            msg_found = True
+    start = time()
```
dianpopa: We could make the `_wait_for_message` function public and call it from here, right?

tests/integration_tests/performance/test_boottime.py
```diff
@@ -136,20 +126,9 @@ def _configure_vm(microvm, network_info=None, initrd=False):
...
-    )
-    assert microvm.api_session.is_status_no_content(response.status_code)
-
     microvm.start()
-    return log_fifo, _tap if network_info else None
+    if initrd:
+        time.sleep(0.8)
+    else:
+        time.sleep(0.4)
+    return microvm.log_data, _tap if network_info else None
```
dianpopa: The tests can access themselves `microvm.log_data` at any time, right?

tests/framework/microvm.py
lauralt: I was thinking that maybe we will use that function with a different timeout at some point, but until that, this simplifies things a lot :ok_hand:. I've updated with your suggestion.

tests/integration_tests/functional/test_signals.py
lauralt: Done.

tests/framework/microvm.py
```diff
@@ -221,6 +234,10 @@ def memory_events_queue(self, queue):
         """Set the memory usage events queue."""
         self._memory_events_queue = queue
 
+    def append_to_log_data(self, data):
```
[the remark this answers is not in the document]
gc-plp: I find that a setter for a functionality that accumulates data is not very intuitive.
In principle a setter should only be used to set and eventually process a value, in order for that value to be held until another set is called. When a subsequent set is called the old value should be overwritten.

tests/framework/microvm.py
```diff
@@ -497,6 +537,32 @@ def start(self):
...
+            try:
+                fd = open(path, "r")
+                while True:
+                    if microvm.logging_thread.stopped():
+                        return
+                    data = fd.readline()
+                    if data:
+                        microvm.append_to_log_data(data)
+            except IOError as error:
+                print(str(error))
```
gc-plp: Print statements are not necessarily shown by pytest.
You could use a logger instance declared at the top of the file and output the error there:

```
import logging
log = logging.getLogger("microvm")

...

log.error(f"[{microvm.id}] IOError while monitoring fd: {error}")
```

tests/integration_tests/performance/test_boottime.py
lauralt: True, I was under the impression that I have to return the `microvm` here for changes to be propagated. That's why I've moved the particular `time.sleep()`s here and returned instead `log_data`. But you are right,`log_data` can accessed also in the caller functions :thumbsup:. I've also moved back those `sleep()`s.

tests/framework/microvm.py
lauralt: Done.

PR #1693, opened 2020-03-20
Add functionality in VMM to fetch the KVM dirty bitmap

aghecenco:
## Reason for This PR

Added a new `pub fn` in the VMM runtime controller that retrieves the KVM dirty bitmap.

## Description of Changes

1. dirty bitmap fetching functionality in the `vmm`
1. support for enabling/disabling dirty page tracking at runtime in the `vmm`
1. rustacean `vmm` integration tests with a noisy guest

The core component is already present in `kvm-ioctls`, and the rest is glue.

## License Acceptance

## PR Checklist


src/vmm/src/controller.rs
```diff
@@ -225,4 +227,243 @@ impl VmmController {
...
+        }
+
+        fn set_dirty_page_tracking(&self, track_dirty_pages: bool) {
+            let vmm = self.vmm.lock().unwrap();
+            config_guest_memory(vmm.guest_memory(), track_dirty_pages, vmm.vm.fd());
+        }
+    }
+
+    #[test]
+    fn test_dirty_bitmap() {
```
serban300: Maybe we should also test the case where the VM ran and dirtied 1 page, but KVM_DIRTY_BITMAP wasn't enabled. So the result should be an empty bitmap.

src/vmm/src/vstate.rs
```diff
@@ -1310,14 +1310,17 @@ mod tests {
...
+    use std::sync::{Arc, Barrier};
     #[cfg(target_arch = "x86_64")]
     use std::time::Duration;
 
     use super::super::devices;
     use super::*;
 
     use utils::signal::validate_signal_num;
+    #[cfg(target_arch = "x86_64")]
     use vmm_config::boot_source::DEFAULT_KERNEL_CMDLINE;
```
dianpopa: nit: or you can move this in the x86_64 specific function that uses it (i.e load_good_kernel)

dianpopa:
> I just added 2 comments for the moment. They are nit-ish. I am not sure about the unit tests. It's a lot of code that has to be added just for testing one small thing. I have to spend some more time thinking about it.

Indeed, there is a pretty big quantity of code for one small thing in the same place. One idea that pops in mind would be to move the code that does the general setup of the vmm (i.e impl Vmm, create_guest_workload, create_device_managers) in a `tests/` folder inside the vmm crate. That way we (hopefully) make this setup code reusable from multiple unit tests. WDYT?

src/vmm/tests/mod.rs -> 
```
pub fn setup -> Vmm
```

src/vmm/src/controller.rs -> 
```
Vmm = setup()
let mut vmm_controller =
            VmmController::new(VmResources::default(), Arc::new(Mutex::new(vmm)));
``

**Comment by iulianbarbu** [comment]
> > I just added 2 comments for the moment. They are nit-ish. I am not sure about the unit tests. It's a lot of code that has to be added just for testing one small thing. I have to spend some more time thinking about it.
> 
> Indeed, there is a pretty big quantity of code for one small thing in the same place. One idea that pops in mind would be to move the code that does the general setup of the vmm (i.e impl Vmm, create_guest_workload, create_device_managers) in a `tests/` folder inside the vmm crate. That way we (hopefully) make this setup code reusable from multiple unit tests. WDYT?
> 
> src/vmm/tests/mod.rs ->
> 
> ```
> pub fn setup -> Vmm
> ```
> 
> src/vmm/src/controller.rs ->
> 
> ```
> Vmm = setup()
> let mut vmm_controller =
>             VmmController::new(VmResources::default(), Arc::new(Mutex::new(vmm)));
> ``
> ```

We should keep in mind the unit tests from `builder.rs`, which tests building a vmm, but it does not run it. I think it is worth looking at these tests because it might have common pieces. It feels like two pieces of work to me: creating unit tests for kvm dirty bitmap and common* infrastructure for vmm crate unit tests (`builder.rs` and `controller.rs` specific so far).

Or, we can make this building block functionality for `controller.rs` unit tests in this PR, since this is the only thing relevant to it, and track through an issue the part where we have to investigate if it is the case to use this unit test functionality in `builder.rs` as well. WDYT?

File: src/vmm/src/controller.rs [hunk 3]
```diff
@@ -225,4 +227,243 @@ impl VmmController {
...
+            |slot: usize, region: &GuestRegionMmap| -> std::result::Result<(), VmmActionError> {
+                let bitmap_region = vmm_handle
+                    .vm
+                    .fd()
+                    .get_dirty_log(slot as u32, region.len() as usize)
+                    .map_err(|e| VmmActionError::InternalVmm(super::Error::DirtyBitmap(e)))?;
+                bitmap.insert(slot, bitmap_region);
+                Ok(())
+            },
+        )?;
```
acatangiu, line 247: I would add this functionality as a `struct Vmm` method. The `VmmController` is just a wrapper over `Vmm` to translate between vmm interface and upper layers.

Getting dirty pages sounds like functionality worthy of being part of the `Vmm` programmatic interface.

src/vmm/src/controller.rs
```diff
@@ -225,4 +227,243 @@ impl VmmController {
 
         Ok(())
     }
+
+    /// Retrieves the KVM dirty bitmap for each of the guest's memory regions.
+    pub fn get_dirty_bitmap(
+        &self,
+    ) -> std::result::Result<HashMap<usize, Vec<u64>>, VmmActionError> {
+        let mut bitmap: HashMap<usize, Vec<u64>> = HashMap::new();
```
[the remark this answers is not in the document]
aghecenco: There's no reason for slot indices to be contiguous and some users might want to leverage that in the future, so I wouldn't simplify it to a plain `Vec` (a `Vec<Option<...>>` would work but it would make the code really ugly). The `Vec<(slot, Vec<u64>)>` solution gets rid of the hashing and keeps the map-like correspondence, but it increases per-slot access time for the consumer, and I'd rather avoid that at the expense of hashing.

src/vmm/src/controller.rs
aghecenco: Added. The result is an error (`ENOENT`) because KVM can't find any regions with dirty tracking enabled.

src/vmm/src/controller.rs
aghecenco: I moved the implementation and the unit tests out of the `controller` to the `vmm`. I left the `controller` action and entry point, but I'm considering whether we'll need them. Without an HTTP API, I think not.

aghecenco:
We discussed offline and agreed that this functionality (and others like it) are better tested through Rust integ tests. This PR is therefore blocked on creating adequate tests. #1764

aghecenco:
Will rebase on top of #1791

aghecenco:
Blocked by https://github.com/firecracker-microvm/firecracker/issues/1834

src/vmm/tests/integration_tests.rs
```diff
@@ -133,3 +135,88 @@ fn test_vmm_seccomp() {
...
+            let resources: VmResources = MockVmResources::new()
+                .with_boot_source(boot_source_cfg)
+                .into();
+            let mut event_manager = EventManager::new().unwrap();
+            let empty_seccomp_filter = get_seccomp_filter(SeccompLevel::None).unwrap();
+
+            let vmm = build_microvm(&resources, &mut event_manager, &empty_seccomp_filter).unwrap();
+            assert_eq!(
+                format!("{:?}", vmm.lock().unwrap().get_dirty_bitmap().err()),
+                "Some(DirtyBitmap(Error(2)))"
```
ioanachirca, line 157: nit: maybe add a comment explaining that if the requested bitmap doesn't exist, the kernel returns ENOENT (2). On a cursory glance that 2 looks a bit mysterious 😄

src/vmm/src/vstate.rs
```diff
@@ -563,6 +540,34 @@ impl Vm {
...
+    pub(crate) fn set_kvm_memory_regions(
+        &self,
+        guest_mem: &GuestMemoryMmap,
+        track_dirty_pages: bool,
+    ) -> Result<()> {
+        let flags = if track_dirty_pages {
+            KVM_MEM_LOG_DIRTY_PAGES
+        } else {
+            0
+        };
```
serban300, line 553: I would do something like:

```
let mut flags = 0;
if track_dirty_pages {
    flags &= KVM_MEM_LOG_DIRTY_PAGES;
}
```

src/vmm/tests/integration_tests.rs
```diff
@@ -133,3 +135,88 @@ fn test_vmm_seccomp() {
...
+}
+
+#[test]
+#[cfg(target_arch = "x86_64")]
+fn test_dirty_bitmap_succes() {
+    // This test is `x86_64`-only until we come up with an `aarch64` kernel that dirties a lot
+    // of pages.
+    let pid = unsafe { libc::fork() };
+    match pid {
+        0 => {
```
serban300, line 181: I think this scenario becomes increasingly popular lately. Would it be possible to have a helper struct for handling this ?

src/vmm/src/vstate.rs
```diff
@@ -563,6 +540,34 @@ impl Vm {
...
+        guest_mem
+            .with_regions(|index, region| {
+                let memory_region = kvm_userspace_memory_region {
+                    slot: index as u32,
+                    // It's safe to unwrap because the guest address is valid.
```
lauralt, line 558: Is this comment accurate?

src/vmm/src/vstate.rs
```diff
@@ -563,6 +540,34 @@ impl Vm {
...
+                    guest_phys_addr: region.start_addr().raw_value() as u64,
+                    memory_size: region.len() as u64,
+                    userspace_addr: guest_mem.get_host_address(region.start_addr()).unwrap() as u64,
```
iulianbarbu, line 561: Can we add a comment to why is safe to unwrap?

src/vmm/tests/integration_tests.rs
aghecenco: Yup, good idea, I logged https://github.com/firecracker-microvm/firecracker/issues/1887 to track it separately, this PR is already big and this is out of scope-ish.

src/vmm/src/vstate.rs
aghecenco: It belongs 2 lines below, where https://github.com/firecracker-microvm/firecracker/pull/1693#discussion_r423624542 noticed it missing.

aghecenco:
@iulianbarbu @ioanachirca feedback addressed, approves lost, please take another look

PR #1697, opened 2020-03-21
[issue #1685] Support default value for bool arguments

### LINKED ISSUE #1685: arg_parser: bool values should have the default set to false
Right now the arg_parser only supports custom default values for `String` values. Specifying a custom default values for bool arguments (also called flags) doesn't make a lot of sense because when the flag is not passed as an argument it means that its value is false, and when the flag is passed as an argument its value is true. Even though custom default values don't make a lot of sense for `Value::bool` we should nevertheless have a default value for these types of arguments, and that default value should always be false.

I couldn't really find a clean way for implementing this, so ideas are welcomed.

One option that I was thinking about is to specify in the constructor of `Argument` the default value. Right now the constructor of `Argument` assumes that the type of the argument is flag (because it specifies that takes_value = false):
```rust
        Argument {
            name,
            required: false,
            requires: None,
            takes_value: false,
            default_value: None,
            help: None,
            user_value: None,
        }
```

Since this is the case, we could also set the default_value to Some(false).
When the `takes_value` function is called with `true`, we can reset the `default_value` to None in case that value was Some(false). This is a bit a not very nice workaround and is also a bit weird because the order in which you call `takes_value` and `default_value` matters.

Another option is to implement `default_value` for both `Value::String` and `Value::bool`, but for `Value::bool` only allow the default to be false.

From these two options, I think I prefer the second one because it is a less weird workaround :))

As I said both options are pretty meh. Other suggestions are welcomed.

AnishShah:
Signed-off-by: Anish Shah <shah.anish07@gmail.com>

## Reason for This PR

Fixes #1685 

## Description of Changes

Changes in arg_parse.rs to support default value for boolean arguments 

## License Acceptance

## PR Checklist

### DIFF
--- a/src/utils/src/arg_parser.rs
+++ b/src/utils/src/arg_parser.rs
@@ -222,11 +222,11 @@ impl<'a> Arguments<'a> {
             .and_then(|arg_value| arg_value.as_string())
     }
 
-    /// Return the value of an argument if the argument exists and has the type
-    /// bool. Otherwise return None.
+    /// Return true if the argument exists and has the type bool. 
+    /// Otherwise return false.
     pub fn value_as_bool(&self, arg_name: &'static str) -> Option<bool> {
         self.value_of(arg_name)
-            .and_then(|arg_value| arg_value.as_bool())
+            .map_or(Some(false), |arg_value| arg_value.as_bool())
     }
 
     /// Get the extra arguments (all arguments after `--`).


lauralt:
Hi, @AnishShah and thanks for your interest!
A little bit of context for the issue linked to your PR: If you take a look, for example, in `main.rs`, you can see how we check if a flag is present in the arguments passed by user: 
`let api_enabled = !arguments.value_as_bool("no-api").unwrap_or(false);`
We want to safely replace that `unwrap_or()` with an `unwrap()`. If we do this now, it will panic :( (more precisely `value_as_bool()` will panic). That's because, when we build the `ArgParser` with the possible arguments, all of the arguments (flags and arguments with actual values) have a `None` value. If the user passes some flag, the value set for that one will be `Bool(true)`. But there is no possible scenario when that value can be `Bool(false)`. What we meant by `support default value for bool arguments` was that we should find a way to implicitly "say" that the default value for flags is `false`. In my opinion, this logic should be handled in `arg_parser.rs`, there is no need for the user to set a default value for a flag because it would be indeed `false`.
One way to fix this problem would be to update `value_as_bool()` to never return `None` or to find a way to set also the default value for flags, not only for the ones that take `String` values.
What do you think?

andreeaflorescu:
Hey @AnishShah. The issue description was a bit fuzzy, sorry for that. I will shortly update it so that it makes more sense and it gives some hints about how this can be implemented.

As @lauralt said, it doesn't make a lot of sense to allow to specify a default (other than false) for the arguments that don't take a value because if the default value is true, there is no way for users to change that default. The presence of a flag means that its value is true, the absence of the flag should mean that its value is false. How this can actually be implemented is a bit more complicated than I initially anticipated. We need to think about a better way to have defaults as false. I will also sync with @lauralt and we will think about a solution. Please let us know if you have any other ideas about how to map this in the arg parser.

andreeaflorescu:
Issue updated with some options: https://github.com/firecracker-microvm/firecracker/issues/1685

AnishShah:
Thanks for the quick reply on this PR - @lauralt, @andreeaflorescu and @acatangiu. I have made some changes to `value_as_bool` fn. Let me know what do you think.

acatangiu:
@AnishShah the change you propose in this PR feels like throwing a rug over a stain instead of cleaning it up :stuck_out_tongue: 

I think we should change the underlying representation to better suit the usage scenario.

@lauralt @andreeaflorescu what do you think of this option: https://github.com/firecracker-microvm/firecracker/issues/1685#issuecomment-602632241

From a bird's-eye view it looks like it would naturally fit the scenario.

lauralt:
> @AnishShah the change you propose in this PR feels like throwing a rug over a stain instead of cleaning it up 
> 
> I think we should change the underlying representation to better suit the usage scenario.
> 
> @lauralt @andreeaflorescu what do you think of this option: [#1685 (comment)](https://github.com/firecracker-microvm/firecracker/issues/1685#issuecomment-602632241)
> 
> From a bird's-eye view it looks like it would naturally fit the scenario.

The idea seems good, I didn't have the chance to try anything in this direction. But I think the changes involved in your proposal would be pretty significant. I'm also okay with merging the commit in its current form and then come later with a subsequent PR.
@AnishShah What do you think, do you want to give it a shot with what @acatangiu proposed or do you want to let us try that later?

AnishShah:
Thanks @lauralt. I can wait till we finalize a design.

AnishShah:
Closing this in favor of a new PR.

Issue #1675, opened 2020-03-16
Make the Logger more modular


serban300:
We should at least decouple the record formatting logic into a different module (e.g. `LogFormatter`). Check if there's anything else that we can do to make the logger more modular.

JonathanWoollett-Light:
A recent refactor in https://github.com/firecracker-microvm/firecracker/pull/4047 significantly simplified the logger, which I believe should address this issue.

Please re-open the issue if you still have outstanding concerns.

Issue #1686, opened 2020-03-19
Remove the `configure` workaround in the logger


andreeaflorescu:
Once #1361 is merged, we will have the possibility for adding logging from the start of the Firecracker process. With this new feature, we can remove the previous workaround by which the logging would be printed at stdout before the logging is initialized. The workaround was added so that we can have logging from startup which can now be achieved by simply configuring the logger from the start.

One problem that we might face is that when installing the panic hook we don't know if the logger was configured or not. We can workaround this issue, by checking if the logger is configured. If it is not configured, we can just use `println` instead of `error` to print the error message from the panic hook.

iulianbarbu:
I remember that we had some discussions in which it was unclear if checking if the logger is initialized will also initialize it. Just wanted to put this here, because I am unsure if it still stands.

serban300:
I strongly agree that we have to rethink our strategy here, but IMO we should find a solution that involves setting the logger right at the beginning of `main()`. Using `println()` until the logger is initialized still feels like a workaround to me and I think it can turn into a burden in the long run. We'll have to pay attention to this special case whenever we add a piece of code in `main()` . We'll have to be very aware whether the code is added before configuring the logger or after and we'll have to use `println` or `info, warn, error, etc` accordingly. This seems very error prone to me.

As I see it, we need 2 phases anyway:

1. Right at the beginning of `main()` we should set the actual logger (one way or another we have to end up calling `set_logger()`). This logger should be used for the entire lifetime of the application.
2. We should be able to reconfigure some of the parameters of the logger during execution. For example the destination, the format, etc. Which is what we'll also do when we receive the API call.

I don't think that `configure()` is the main issue here. The main issue is the number of `if` branches and special cases. For example:

```
    fn write_log(&self, msg: String, msg_level: Level) {
        if STATE.load(Ordering::Relaxed) == INITIALIZED {
            if let Some(guard) = buf_guard(&self.log_buf).as_mut() {
                // No need to explicitly call flush because the underlying LineWriter flushes
                // automatically whenever a newline is detected (and we always end with a
                // newline the current write).
                if guard.write(&(format!("{}\n", msg)).as_bytes()).is_err() {
                    // No reason to log the error to stderr here, just increment the metric.
                    METRICS.logger.missed_log_count.inc();
                }
            } else {
                METRICS.logger.missed_log_count.inc();
                panic!("Failed to write to the provided log destination due to poisoned lock");
            }
[...]

PR #1701, opened 2020-03-23
Logger simplifications

serban300:
## Reason for This PR

Introduce some logger simplifications that will make it easier to work on #1675 

## Description of Changes

Introduce some logger simplifications that will make it easier to work on #1675 

## License Acceptance

## PR Checklist

### DIFF
--- a/src/logger/src/logger.rs
+++ b/src/logger/src/logger.rs
@@ -109,7 +109,7 @@ pub type Result<T> = result::Result<T, LoggerError>;
 // Values used by the Logger.
 const IN_PREFIX_SEPARATOR: &str = ":";
 const MSG_SEPARATOR: &str = " ";
-const DEFAULT_LEVEL: LevelFilter = LevelFilter::Warn;
+const DEFAULT_MAX_LEVEL: LevelFilter = LevelFilter::Warn;
 
 // Synchronization primitives used to run a one-time global initialization.
 const UNINITIALIZED: usize = 0;
@@ -262,7 +262,7 @@ impl Logger {
     ///
     /// fn main() {
     ///     let l = LOGGER.deref();
-    ///     l.set_level(log::LevelFilter::Warn);
+    ///     l.set_max_level(log::LevelFilter::Warn);
     ///     assert!(l.configure(Some("MY-INSTANCE".to_string())).is_ok());
     ///     info!("An informational log message");
     ///     warn!("A test warning message");
@@ -273,7 +273,7 @@ impl Logger {
     /// 2018-11-07T05:34:25.180751152 [MY-INSTANCE:INFO:logger/src/lib.rs:389] A test warning
     /// message
     /// ```
-    pub fn set_level(&self, level: LevelFilter) -> &Self {
+    pub fn set_max_level(&self, level: LevelFilter) -> &Self {
         set_max_level(level);
         self
     }
@@ -334,10 +334,10 @@ impl Logger {
     }
 
     /// if the max level hasn't been configured yet, set it to default
-    fn try_init_level(&self) {
+    fn try_init_max_level(&self) {
         // if the max level hasn't been configured yet, set it to default
         if max_level() == LevelFilter::Off {
-            self.set_level(DEFAULT_LEVEL);
+            self.set_max_level(DEFAULT_MAX_LEVEL);
         }
     }
 
@@ -373,7 +373,7 @@ impl Logger {
             self.set_instance_id(some_instance_id);
         }
 
-        self.try_init_level();
+        self.try_init_max_level();
 
         STATE.store(UNINITIALIZED, Ordering::SeqCst);
 
@@ -413,7 +413,7 @@ impl Logger {
             *g = Some(log_dest);
         }
 
-        self.try_init_level();
+        self.try_init_max_level();
 
         STATE.store(INITIALIZED, Ordering::SeqCst);
         self.write_log(header, Level::Info);
@@ -550,7 +550,7 @@ mod tests {
 
         l.set_include_origin(true, true)
             .set_include_level(true)
-            .set_level(log::LevelFilter::Info);
+            .set_max_level(log::LevelFilter::Info);
         assert_eq!(l.show_line_numbers(), true);
         assert_eq!(l.show_file_path(), true);
         assert_eq!(l.show_level(), true);
--- a/src/logger/src/logger.rs
+++ b/src/logger/src/logger.rs
@@ -97,7 +97,7 @@ use std::result;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Mutex, RwLock};
 
-use log::{set_logger, set_max_level, Level, Log, Metadata, Record};
+use log::{max_level, set_logger, set_max_level, Level, LevelFilter, Log, Metadata, Record};
 use metrics::{Metric, METRICS};
 use utils::time::LocalTime;
 
@@ -109,7 +109,7 @@ pub type Result<T> = result::Result<T, LoggerError>;
 // Values used by the Logger.
 const IN_PREFIX_SEPARATOR: &str = ":";
 const MSG_SEPARATOR: &str = " ";
-const DEFAULT_LEVEL: Level = Level::Warn;
+const DEFAULT_LEVEL: LevelFilter = LevelFilter::Warn;
 
 // Synchronization primitives used to run a one-time global initialization.
 const UNINITIALIZED: usize = 0;
@@ -137,23 +137,17 @@ pub struct Logger {
     show_level: AtomicBool,
     show_file_path: AtomicBool,
     show_line_numbers: AtomicBool,
-    level: AtomicUsize,
     instance_id: RwLock<String>,
 }
 
 impl Logger {
-    // Creates a new instance of the current logger.
-    //
-    // The default log level is `WARN` and the default destination is stdout/stderr based on level.
+    /// Creates a new instance of the current logger.
     fn new() -> Logger {
         Logger {
             log_buf: Mutex::new(None),
             show_level: AtomicBool::new(true),
             show_line_numbers: AtomicBool::new(true),
             show_file_path: AtomicBool::new(true),
-            level:
-            // DEFAULT_LEVEL is warn so the destination output is stderr.
-            AtomicUsize::new(DEFAULT_LEVEL as usize),
             instance_id: RwLock::new(String::new()),
         }
     }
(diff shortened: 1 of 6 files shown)


src/logger/src/logger.rs
```diff
@@ -475,24 +468,20 @@ impl fmt::Display for LoggerError {
 
 /// Implements the "Log" trait from the externally used "log" crate.
 impl Log for Logger {
-    // Test whether the level of the log line should be outputted or not based on the currently
-    // configured level. If the configured level is "warning" but the line is logged through "info!"
-    // marco then it will not get logged.
-    fn enabled(&self, metadata: &Metadata) -> bool {
-        metadata.level() as usize <= self.level.load(Ordering::SeqCst)
+    // This is currently not used.
+    fn enabled(&self, _metadata: &Metadata) -> bool {
```
dianpopa: nit: Why not delete it?

serban300: It's part of the `Log` trait and all the implementers have to define this method. Otherwise the compiler throws a `not all trait items implemented` error.

serban300:
I fixed the unit tests failures. @dianpopa @acatangiu please take another look ! Thanks !

src/logger/src/logger.rs
```diff
@@ -268,7 +263,7 @@ impl Logger {
     ///
     /// fn main() {
     ///     let l = LOGGER.deref();
-    ///     l.set_level(log::Level::Info);
+    ///     l.set_level(log::LevelFilter::Info);
```
andreeaflorescu: I think the documentation also needs to be updated. Is the default level still warning?

src/logger/src/logger.rs
```diff
@@ -420,7 +420,7 @@ impl Logger {
                 // No need to explicitly call flush because the underlying LineWriter flushes
                 // automatically whenever a newline is detected (and we always end with a
                 // newline the current write).
-                if guard.write(&(format!("{}\n", msg)).as_bytes()).is_err() {
+                if guard.write_all(&(format!("{}\n", msg)).as_bytes()).is_err() {
```
andreeaflorescu: Why switch to `write_all`? My understanding is that depending on the writer it can be more inefficient than `write` + `flush`.

src/vmm/src/vmm_config/logger.rs
```diff
@@ -120,13 +120,14 @@ pub fn init_logger(
         .set_include_origin(logger_cfg.show_log_origin, logger_cfg.show_log_origin)
         .set_include_level(logger_cfg.show_level);
 
+    let writer = FifoLineWriter::new(
```
andreeaflorescu: With this change, can we still use regular files with the logger? Our logger documentation says that you can use regular files as well and we just renamed `logger_fifo` to `logger_path` for that reason. Do we get any benefit from using `FifoLineWriter` instead of `Writer`?

src/vmm/src/vmm_config/mod.rs
```diff
@@ -194,14 +175,14 @@ mod tests {
...
 
     #[test]
-    fn test_log_writer() {
+    fn test_fifo_line_writer() {
         let log_file_temp =
             TempFile::new().expect("Failed to create temporary output logging file.");
         let good_file = log_file_temp.as_path().to_path_buf();
-        let res = Writer::new(good_file);
-        assert!(res.is_ok());
+        let maybe_fifo = open_fifo(&good_file);
```
andreeaflorescu: nit: `maybe_*` is usually used in the async context. `fifo` should suffice here.

src/logger/src/logger.rs
serban300: From what I understand, the intention here is to write an entire log record. This guarantee can only be satisfied by `write_all`. `write` can write only a part of the log and return with `Ok(num_written_bytes)`. This can lead to discarding part of the log record. `flush` shouldn't matter here. `write_all` doesn't necessarily lead to a flush. `flush` doesn't necessarily have to be called after `write`.

This may affect performance or not. Depends on the implementation of the writer. But I think that we should offer the guarantee that the entire log message is written.

src/logger/src/logger.rs
serban300: Sorry, can you expand on that please ? I'm not sure I understand how.

src/vmm/src/vmm_config/logger.rs
serban300: Yes, we can still use regular files. There is no change in behavior here. No benefit or penalty. Just renaming. And removin the `impl Write for Writer` which wasn't needed, since it was just forwarding the calls to the underlying `LineWriter`.

Maybe `open_fifo` is not necessarily the best name. Maybe `open_fd_nonblock` would be better. And `NonblockFdLineWriter`.

src/vmm/src/vmm_config/mod.rs
serban300: I used `maybe_` because this returns an option. It can contain something or it can be `None`. I think I saw this in the rust examples somewhere, but I don't remember where exactly.

src/logger/src/logger.rs
andreeaflorescu: The default is Warning only in the vmm crate. The logger crate does not have a default any more and this is the documentation from the logger crate. Am I missing something?

src/logger/src/logger.rs
andreeaflorescu: I think we are calling flush after each write anyway. But I guess it makes sense to call write_all instead. Can you also specify that you are using write_all instead of write in the commit message as well?

src/vmm/src/vmm_config/logger.rs
andreeaflorescu: Ah, okay, I see. So I thought that `open_fifo` and `FifoLineWriter` are functions from std that operated only with fifos. My bad. I just saw that you are defining them in `mod.rs`

src/vmm/src/vmm_config/mod.rs
```diff
@@ -97,38 +97,19 @@ impl TryInto<RateLimiter> for RateLimiterConfig {
...
-impl io::Write for Writer {
-    fn write(&mut self, msg: &[u8]) -> Result<(usize)> {
-        self.line_writer.write_all(msg).map(|()| msg.len())
-    }
-
-    fn flush(&mut self) -> Result<()> {
-        self.line_writer.flush()
-    }
-}
+type FifoLineWriter = io::LineWriter<File>;
```
andreeaflorescu: How about `FcLineWriter` because it is a specialization of a `LineWriter` for the Firecracker usecase. Using `Fifo` in the name implies that it can only use fifo, but as you mentioned in your previous comment it works with regular files as well.

src/vmm/src/vmm_config/mod.rs
```diff
@@ -97,38 +97,19 @@ impl TryInto<RateLimiter> for RateLimiterConfig {
...
 type Result<T> = std::result::Result<T, std::io::Error>;
 
-/// Structure `Writer` used for writing to a FIFO.
-pub struct Writer {
-    line_writer: io::LineWriter<File>,
+/// Create and open a FIFO for writing to it.
+/// In order to not block the instance if nobody is consuming the message that is flushed to the
+/// two pipes, we are opening it with `O_NONBLOCK` flag. In this case, writing to a pipe will
+/// start failing when reaching 64K of unconsumed content.
+fn open_fifo(fifo_path: &PathBuf) -> Result<File> {
```
andreeaflorescu: Can we make this a function of `FifoLineWriter` so that we do not have orphan functions?

src/logger/src/logger.rs
andreeaflorescu: My comment was more of an FYI and it is related to the other issue actually: https://github.com/firecracker-microvm/firecracker/issues/1686.

Let's sync on this offline as it is a lot to write in a comment.

src/vmm/src/signal_handler.rs
```diff
@@ -144,6 +145,7 @@ mod tests {
 
     #[test]
     fn test_signal_handler() {
+        LOGGER.set_max_level(LevelFilter::Off);
```
andreeaflorescu: It looks a bit arbitrary that you need to use the logger and disable it for the signal_handler test to work. Can you add a comment to why is this actually needed? I assume it is because the seccomp filters.

src/logger/src/logger.rs
dianpopa: @serban300 I think @andreeaflorescu  is referring to these 2 comments:  https://github.com/firecracker-microvm/firecracker/blob/4687bc8a9a8e23cdd6c258c656a23e665be0c49a/src/logger/src/logger.rs#L147 and https://github.com/firecracker-microvm/firecracker/blob/4687bc8a9a8e23cdd6c258c656a23e665be0c49a/src/logger/src/logger.rs#L254. We need to carve out the part that mentions a default level as from now on the default level will be that of the underlying log crate.

serban300: You're right. Done.

src/logger/src/logger.rs
serban300: Sure. Done.

src/vmm/src/signal_handler.rs
serban300: Done.

src/vmm/src/vmm_config/mod.rs
serban300: The compiler doesn't allow defining functions on `FifoLineWriter` since it's basically a `LineWriter<File>` ...

src/vmm/src/vmm_config/mod.rs
serban300: Done. Also I renamed `open_fifo` to `open_fd_nonblock`.

src/logger/src/logger.rs
serban300: Actually let's keep the default level in order to make sure we keep exactly the same behavior as before.

src/vmm/src/vmm_config/mod.rs
andreeaflorescu: Yap, it would need to be a struct as it was before. I don't see why the change in this commit is necessary. I preferred it as it was before. Better yet, `LineWriter` can use the [new type pattern](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html).

src/logger/src/logger.rs
```diff
@@ -334,10 +334,10 @@ impl Logger {
     }
 
     /// if the max level hasn't been configured yet, set it to default
-    fn try_init_level(&self) {
+    fn try_init_max_level(&self) {
```
andreeaflorescu: nit: `try_` functions typically return a Result. You can look at examples here: https://doc.rust-lang.org/nightly/core/index.html?search=try_

PR #1717, opened 2020-03-27
arg_parse.rs: Add two static method - value_argument and flag_argument to Argument struct

### LINKED ISSUE #1685: arg_parser: bool values should have the default set to false

AnishShah:
Signed-off-by: Anish Shah <shah.anish07@gmail.com>

## Reason for This PR

Fixes #1685

## Description of Changes

Introduced two static method to Argument struct - 
1. `value_argument` - returns an argument that can take user value 
2.  `flag_argument` -  returns a flag.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/utils/src/arg_parser.rs
+++ b/src/utils/src/arg_parser.rs
@@ -91,27 +91,47 @@ impl<'a> ArgParser<'a> {
     }
 }
 
+/// Type of argument.
+#[derive(Clone, Debug, PartialEq)]
+enum ArgumentType {
+    ValueArgument,
+    FlagArgument,
+}
+
 /// Stores the characteristics of the `name` command line argument.
 #[derive(Clone, Debug, PartialEq)]
 pub struct Argument<'a> {
     name: &'a str,
     required: bool,
     requires: Option<&'a str>,
-    takes_value: bool,
+    arg_type: ArgumentType,
     default_value: Option<Value>,
     help: Option<&'a str>,
     user_value: Option<Value>,
 }
 
 impl<'a> Argument<'a> {
-    /// Create a new `Argument` that keeps the necessary information for an argument.
-    pub fn new(name: &'a str) -> Argument<'a> {
+    /// Create a new `ValueArgument` that keeps the necessary information for a value argument.
+    pub fn value_argument(name: &'a str) -> Argument<'a> {
         Argument {
             name,
             required: false,
             requires: None,
-            takes_value: false,
+            arg_type: ArgumentType::ValueArgument,
             default_value: None,
+            help: None,
+            user_value: None,
+        }
+    }
+
+    /// Create a new `FlagArgument` that keeps the necessary information for a flag argument.
+    pub fn flag_argument(name: &'a str) -> Argument<'a> {
+        Argument {
+            name,
+            required: false,
+            requires: None,
+            arg_type: ArgumentType::FlagArgument,
+            default_value: Some(Value::Bool(false)),
             help: None,
             user_value: None,
         }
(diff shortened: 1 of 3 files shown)


src/utils/src/arg_parser.rs
```diff
@@ -91,32 +91,52 @@ impl<'a> ArgParser<'a> {
...
 pub struct Argument<'a> {
     name: &'a str,
     required: bool,
     requires: Option<&'a str>,
-    takes_value: bool,
+    arg_type: ArgumentType,
     default_value: Option<Value>,
     help: Option<&'a str>,
     user_value: Option<Value>,
 }
```
acatangiu: Changing `takes_value: bool` with `arg_type: ArgumentType` does provide a little extra clarity, but just as before you still have a invisible/assumed connection between the two fields in the struct `arg_type: ArgumentType` and `user_value: Option<Value>` whose values implicitly depend on one another.

If I understand correctly, with this version the compiler allows `user_value == Some(Value)` while `arg_type == ArgumentType::FlagArgument`, while the code logic considers this case to be invalid.

One of the powerful tools in rust to remove these error-prone implicit dependencies and replace them with explicit compile-time validated dependencies is using a single enum that defines only valid states and their possible data:

```rust
enum ArgumentType {
    ValueArgument(Value),
    FlagArgument,
}
```
If this field can be None you can wrap this enum in an `Option`, or even better code is to add another variant to it `NotPresent` or smth.

andreeaflorescu: While looking a bit more at the code, there are quite a few methods that don't make sense for both `ValueArgument` and `FlagArgument`.

You can call the following methods on the `ValueArgument`:
- required
- requires
- default_value
- help

But on the `FlagArgument` only the following methods make sense:
- requires
- help

An `Argument` trait might still make sense even if we duplicate a few lines of code in both `ValueArgument` and `FlagArgument`.

Looks like we are very deep in the rabbit hole :))

lauralt:
I tried some things and played a little to see what this trait approach really involves and to me it doesn't seem like a good fit anymore.
First, the duplicate code is indeed a lot and that's because the "common" functionality is mostly consisted of the setter functions and, even if their implementations are exactly the same for both `ValueArgument` and `FlagArgument`, they must be implemented by the structs (a default trait method implementation can't change the fields of the trait implementors).
Secondly, the functions that parse and validate the arguments, in most of the validations, have to know the type of the Argument (flag or value argument), which, from what I've tried, also implies some duplicate code. Like Andreea said, for flags, a lot of fields won't make sense anymore and it actually looks like a value argument and a flag argument, even though they are both in fact arguments, don't have that much in common: you can't say that a flag is required or that a flag has a value (or, more precisely, a user value; this would be indeed true; `default_value` field also doesn't make sense; a `value` or `is_present` bool field would be enough: 0 when initializing that argument and 1 when it is passed by user).
Also, at this moment, in our arg parser implementation we don't allow the user to pass any argument two or more times, but if I think more about this, it should be no problem if the user passes the same flag 2, 3... times (for value arguments it makes sense this type of validation because the user could give, this way, 2 values for the same argument). With the trait approach, it will be pretty ugly to check these particular things that I've just mentioned as they depend a lot on the implementor type.
Maybe the cleanest way to solve this would be to have two separate collections for flags and arguments with values, but I imagine there will still be some duplicate code.

AnishShah:
Thanks @lauralt for trying out. I had all these same thoughts when I tried the trait approach. I think to do this correctly we have to rethink designing arg_parse.rs. I don't feel this is a good first issue for me. I will try to take up some other issue. Sorry for abandoning this. :(

andreeaflorescu:
@AnishShah no problem at all. Let us know if you need help in selecting the next issue. I was a bit rushing when I added the label "Good First Issue" on this one and I should've double check. Sorry.

PR #1729, opened 2020-03-31
cargo/config: disallow build warnings

dianpopa:
## Reason for This PR

Treat build errors as warnings.

## Description of Changes

Changed cargo/config.
**Obs** Tests are expected to fail until #1693 gets merged.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/host_tools/cargo_build.py
+++ b/tests/host_tools/cargo_build.py
@@ -44,8 +44,8 @@ def cargo_test(path, extra_args='', extra_env=''):
     """Trigger unit tests depending on flags provided."""
     path = os.path.join(path, CARGO_UNITTEST_REL_PATH)
     cmd = 'CARGO_TARGET_DIR={} {} RUST_TEST_THREADS=1 RUST_BACKTRACE=1 ' \
-          'cargo test {} ' \
-          '--all --no-fail-fast'.format(path, extra_env, extra_args)
+          'RUSTFLAGS="{}" cargo test {} --all --no-fail-fast'.format(
+            path, extra_env, get_rustflags(), extra_args)
     utils.run_cmd(cmd)
 
 
@@ -57,7 +57,7 @@ def get_firecracker_binaries():
     """
     target = DEFAULT_BUILD_TARGET
     cd_cmd = "cd {}".format(FC_WORKSPACE_DIR)
-    env_cmd = "TARGET_CC=musl-gcc"
+    env_cmd = 'TARGET_CC=musl-gcc RUSTFLAGS="{}"'.format(get_rustflags())
     cargo_cmd = "cargo build --release --target {}".format(target)
     cmd = "{} && {} {}".format(cd_cmd, env_cmd, cargo_cmd)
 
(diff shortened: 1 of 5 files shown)


.cargo/config
```diff
@@ -8,5 +8,11 @@ target-dir = "build/cargo_target"
 # No need for the `crt_static` flag as rustc appends it by default.
 rustflags = [
 	"-C", "link-arg=-lgcc",
+	"-D", "warnings",
 	"-C", "link-arg=-lfdt",
 ]
+
+[target.'cfg(not(any(target_arch="arm", target_arch="aarch64")))']
```
andreeaflorescu, line 15: This can easily become annoying when doing development. If you have code in flux and you want to test it, my understanding is that the build will fail locally. Can we instead have this as a separate test in the CI to check that when trying to push something upstream there are no build warnings?

acatangiu: Can we do this in cargo for release builds only?

dianpopa: I moved the check inside the integration tests. Take a look!

tests/host_tools/cargo_build.py
```diff
@@ -70,3 +70,11 @@ def get_firecracker_binaries():
     jailer_bin_path = "{}/{}".format(out_dir, JAILER_BINARY_NAME)
 
     return fc_bin_path, jailer_bin_path
+
+
+def get_rustflags():
+    """Get the relevant rustflags for building/unit testing."""
+    rustflags = "-D warnings"
+    if platform.machine() == "aarch64":
+        rustflags += " -C link-arg=-lgcc -C link-arg=-lfdt "
```
iulianbarbu, line 79: If I understand correctly, we would link against `-lgcc` and `-lfdt` for the firecracker binary used inside the integration tests and when cargo testing on arm. Why is this needed now, when we disallow warnings?

dianpopa: We need this flag for aarch64 binary to link (these flags are always used when doing `cargo build` on aarch64. Take a look at the cargo/.config for arm64. Unfortunately there is no way (or I do not know any) of getting the flags set in cargo/.config for some platform from our python tests.

iulianbarbu: Ok, so when compiling with `cargo build`, using `TARGET_CC=musl-gcc`, this has to link on aarch64 against `-lgcc` and `-lfdt` always (because musl-gcc lacks some symbols used in firecracker), and these flags can be found in `$HOME/.cargo/config`. I've looked on my x86 machine, and found this config file inside the .cargo dir inside the project, as well as this part responsible for aarch64 specific linking flags.

It makes sense, once we compile using rustc, and not `cargo build`, to specify these flags in the command line.

iulianbarbu: Oh, actually, looked once more, and found out that we actually use `cargo build` to build the firecracker binaries. Using a .cargo/config with this directive, inside the project where this build happens, should be enough:

[target.'cfg(any(target_arch="arm", target_arch="aarch64"))']
rustflags = [
        "-C", "link-arg=-lgcc",
        "-C", "link-arg=-lfdt",
]

However, this is more like a nit now. I do not have a strong preference, to use this or what you've already pushed in this PR.

dianpopa:
@iulianbarbu @serban300 please re-review. Tests have passed!

Issue #1713, opened 2020-03-26
Decouple Vmm from its Firecracker-specific RPC interface


acatangiu:
This issue is part of a story that tracks a state where the `vmm` crate is no longer Firecracker-specific.

### Benefits:

- swapping out `Vmm` implementations in Firecracker - the `vmm` crate used in Firecracker could be one from [rust-vmm](https://github.com/rust-vmm).
- upstreaming/outsourcing the current `Vmm` implementation in Firecracker to other projects.
- reducing the barrier-to-entry in Firecracker code-base by having much more contained components.
- clear language-enforced dependencies in the form of crates helps keeping the code clean by forcing contributors to work at the right level with the right resources.

### Scope:

`vmm_config`, `VmResources`, `Pre-boot and Runtime Controllers` and `rpc_interface` to be moved to an external `rpc_interface` crate that depend on `Vmm`. Dependency graph will be:

```
+---------------+     Dependencies
|  Firecracker  |           |
+---------------+           |
|  API Server   |           |
+---------------+           |
| RPC Interface |           |
+---------------+           |
|      Vmm      |           |
+---------------+           V
```

With `Vmm` decoupled from all Firecracker-specific configuration primitives and logic.

acatangiu:
Depends on:

- [ ] #1776 
- [ ] #1777 
- [ ] #1778

luminitavoicu:
We don't have an immediate plan for this. Will reopen in the future if necessary.

Issue #1408, opened 2019-11-18
Virtio Improvements


andreeaflorescu:
Redesign virtio devices in Firecracker. This needs to be scoped.
The plan is to consume virtio devices building blocks from rust-vmm. For this goal, we need to make sure that:
- MMIO transport can be easily replaced with PCI
- Rate Limiter is not tightly coupled to the emulation of virtio
- Metrics are opt-in
- Logging is done via the default Rust logger implementation
- others?

TBD Task breakdown.
TBD Actual design.

aghecenco:
re:others:
* device identification is split into device type (currently exposed by the `VirtioDevice` trait) + user-facing device ID. See https://github.com/firecracker-microvm/firecracker/pull/1737#discussion_r405509819

andreeaflorescu:
@aghecenco how about we created a separate issue for the device identification? This issue was created before the decoupling work was started and I assume that now we have a few more things that need to be done. We can link the separate issues in this one.

andreeaflorescu:
Task Breakdown:
- [ ]  https://github.com/firecracker-microvm/firecracker/issues/1761
- [ ] https://github.com/firecracker-microvm/firecracker/issues/1760
- [ ] https://github.com/firecracker-microvm/firecracker/issues/1762
- [ ] #1759 
- [ ] # 1758

aghecenco:
Done & added #1763 to the list above

serban300:
Just wanted to mention a couple of very high-level points that might be worth taking into consideration here. I didn't create tasks since I don't have a detailed design.

### Modularity

A virtual net device is mostly an intermediary. It forwards Ethernet frames from the host or MMDS to the guest and vice-versa. If we consider its purpose and also take into account special cases where a frame can’t be forwarded because of lack of space or other constraints, a basic net device could contain only frame forwarding, frame buffering and constraints checking (e.g. rate limiting) logic.

In the Firecracker net device this basic logic is also coupled with some extra logic. In order to keep the net device as simple as possible it would be nice to decouple this extra logic in different components. At least 3 come to mind:

* a `FrontEnd` or `GuestEnd` that would contain the logic for reading/writing to virtio queues. It could just implement Read and Write:

```
pub struct GuestEnd {
    queues: Vec<Queue>,
    queue_evts: Vec<EventFd>,
    interrupt_evt: EventFd,
}

impl GuestEnd {
[...]

PR #1736, opened 2020-04-01
Create Block devices during configuration instead of boot-time

### LINKED ISSUE #1702: Create Block devices during configuration instead of boot-time
Following our decoupling work, we can now create block devices during microVM configuration instead of pushing creation to boot-time.
This will improve decoupling even further and should decrease code complexity overall.

acatangiu:
## Reason for This PR

Fixes #1702 

Prerequisite for #1713 

## Description of Changes

Following the more decoupled design we can now create `block` devices on their configuration path, rather than saving a config and delaying creation to boot-time.

Such a model allows user-errors (like invalid resources or permissions) to be reported as part of the configuration step rather than later at attempted boot.

It would also remove the time window between configuration and boot where changes to the host system (deleting backing files, changing permissions, etc) would cause microVM boot failures (race condition between validation and commitment of resources).

Such a model should also decrease code complexity and should provide better efficiency to the process of configuring and starting a microVM (no more moving configurations around in memory and validating resources both at config time and boot-time).

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -95,10 +95,12 @@ pub struct Block {
     interrupt_status: Arc<AtomicUsize>,
     interrupt_evt: EventFd,
     pub(crate) queue_evts: [EventFd; 1],
-
     pub(crate) device_state: DeviceState,
 
     // Implementation specific fields.
+    id: String,
+    partuuid: Option<String>,
+    root_device: bool,
     pub(crate) rate_limiter: RateLimiter,
 }
 
@@ -107,8 +109,11 @@ impl Block {
     ///
     /// The given file must be seekable and sizable.
     pub fn new(
+        id: String,
         mut disk_image: File,
+        partuuid: Option<String>,
         is_disk_read_only: bool,
+        is_disk_root: bool,
         rate_limiter: RateLimiter,
     ) -> io::Result<Block> {
         let disk_size = disk_image.seek(SeekFrom::End(0))? as u64;
@@ -124,6 +129,9 @@ impl Block {
         let queues = QUEUE_SIZES.iter().map(|&s| Queue::new(s)).collect();
 
         Ok(Block {
+            id,
+            root_device: is_disk_root,
+            partuuid,
             disk_image_id: build_disk_image_id(&disk_image),
             disk_image,
             disk_nsectors: disk_size / SECTOR_SIZE,
@@ -259,6 +267,26 @@ impl Block {
         METRICS.block.update_count.inc();
         Ok(())
     }
+
+    /// Provides the ID of this block device.
+    pub fn id(&self) -> &String {
+        &self.id
+    }
+
+    /// Provides the PARTUUID of this block device.
+    pub fn partuuid(&self) -> Option<&String> {
+        self.partuuid.as_ref()
+    }
+
+    /// Specifies if this block device is read only.
+    pub fn is_read_only(&self) -> bool {
+        self.avail_features & (1u64 << VIRTIO_BLK_F_RO) != 0
+    }
+
+    /// Specifies if this block device is read only.
+    pub fn is_root_device(&self) -> bool {
+        self.root_device
+    }
 }
 
 impl VirtioDevice for Block {
@@ -386,7 +414,8 @@ pub(crate) mod tests {
         // Rate limiting is enabled but with a high operation rate (10 million ops/s).
         let rate_limiter = RateLimiter::new(0, None, 0, 100_000, None, 10).unwrap();
 
-        Block::new(block_file, true, rate_limiter).unwrap()
+        let id = "test".to_string();
+        Block::new(id, block_file, None, true, false, rate_limiter).unwrap()
     }
 
     pub fn default_mem() -> GuestMemoryMmap {
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -3,9 +3,7 @@
 
 //! Enables pre-boot setup, instantiation and booting of a Firecracker VMM.
 
-use std::convert::TryInto;
 use std::fmt::{Display, Formatter};
-use std::fs::OpenOptions;
 use std::io::{self, Read, Seek, SeekFrom};
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::sync::{Arc, Mutex};
@@ -25,18 +23,16 @@ use utils::terminal::Terminal;
 use utils::time::TimestampUs;
 use vm_memory::{Bytes, GuestAddress, GuestMemoryMmap};
 use vmm_config::boot_source::BootConfig;
-use vmm_config::drive::BlockDeviceConfigs;
+use vmm_config::drive::BlockBuilder;
 use vmm_config::net::NetBuilder;
-use vmm_config::RateLimiterConfig;
 use vstate::{KvmContext, Vcpu, VcpuConfig, Vm};
 use {device_manager, VmmEventsObserver};
 
 /// Errors associated with starting the instance.
 #[derive(Debug)]
 pub enum StartMicrovmError {
-    /// Unable to seek the block device backing file due to invalid permissions or
-    /// the file was deleted/corrupted.
-    CreateBlockDevice(io::Error),
+    /// Unable to attach block device to Vmm.
+    AttachBlockDevice(io::Error),
     /// Internal errors are due to resource exhaustion.
     CreateNetDevice(devices::virtio::net::Error),
     /// Failed to create a `RateLimiter` object.
@@ -87,12 +83,9 @@ impl Display for StartMicrovmError {
     fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
         use self::StartMicrovmError::*;
         match *self {
-            CreateBlockDevice(ref err) => write!(
-                f,
-                "Unable to seek the block device backing file due to invalid permissions or \
-                 the file was deleted/corrupted. Error number: {}",
-                err
-            ),
+            AttachBlockDevice(ref err) => {
+                write!(f, "Unable to attach block device to Vmm. Error: {}", err)
+            }
             CreateRateLimiter(ref err) => write!(f, "Cannot create RateLimiter: {}", err),
             CreateNetDevice(ref err) => {
                 let mut err_msg = format!("{:?}", err);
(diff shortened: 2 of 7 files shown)


raduiliescu:
Will this help in being able to attach block devices to a running Firecracker?

acatangiu:
It is a step in that direction yes, but we're far from it since the biggest missing piece there is PCI + Hotplug

src/vmm/src/vmm_config/drive.rs
```diff
@@ -70,109 +81,114 @@ pub struct BlockDeviceConfig {
...
+        let position = self.get_index_of_drive_id(&config.drive_id);
+        let block_dev = Arc::new(Mutex::new(Self::create_block(config)?));
+        // If the id of the drive already exists in the list, the operation is update/overwrite.
+        match position {
+            // New block device.
+            None => {
+                // Check whether the Device Config belongs to a root device,
+                // we need to satisfy the condition by which a VMM can only have one root device.
+                if is_root_device {
+                    if self.has_root_block {
```
sandreim, line 123: You can rule out duplicate root device before looking at it's position.

src/vmm/src/vmm_config/drive.rs
```diff
@@ -70,109 +81,114 @@ pub struct BlockDeviceConfig {
...
+            // Update existing block device.
+            Some(mut index) => {
+                // Check if the root block device is being updated.
+                if index == 0 && self.has_root_block {
+                    // Set root flag according to the updated block config.
+                    self.has_root_block = is_root_device;
+                } else if is_root_device {
+                    // Check if a second root block device is being added.
+                    if self.has_root_block {
+                        return Err(DriveError::RootBlockDeviceAlreadyAdded);
```
sandreim, line 145: Checking for duplicate block device at top will greatly simplify this code.

src/vmm/src/vmm_config/drive.rs
```diff
@@ -70,109 +81,114 @@ pub struct BlockDeviceConfig {
...
 }
 
-/// Wrapper for the collection that holds all the Block Devices Configs
+/// Wrapper for the collection that holds all the Block Devices
 #[derive(Default)]
-pub struct BlockDeviceConfigs {
-    /// A list of `BlockDeviceConfig` objects.
-    pub config_list: VecDeque<BlockDeviceConfig>,
-    has_root_block: bool,
+pub struct BlockDevices {
```
sandreim, line 86: Is it possible to have a generic struct definition/interface for Vsock/Block/Net stores/builders ?

src/vmm/src/builder.rs
```diff
@@ -655,75 +650,47 @@ fn attach_mmio_device(
...
+            .map_err(RegisterEvent)?;
 
+        let id = block.lock().unwrap().id().clone();
+        // The device mutex mustn't be locked here otherwise it will deadlock.
         attach_mmio_device(
             vmm,
-            drive_config.drive_id.clone(),
-            MmioTransport::new(vmm.guest_memory().clone(), block_device.clone())
-                .map_err(CreateBlockDevice)?,
+            id,
```
sandreim, line 691: I would avoid adding the id parameter here. Are there any options of fetching it inside the function ?

src/vmm/src/vmm_config/drive.rs
```diff
@@ -70,109 +81,114 @@ pub struct BlockDeviceConfig {
...
-        if block_device_config.is_root_device {
-            if self.has_root_block {
-                return Err(DriveError::RootBlockDeviceAlreadyAdded);
-            } else {
-                self.has_root_block = true;
-                // Root Device should be the first in the list whether or not PARTUUID is specified
-                // in order to avoid bugs in case of switching from partuuid boot scenarios to
-                // /dev/vda boot type.
-                self.config_list.push_front(block_device_config);
+    pub fn insert(&mut self, config: BlockDeviceConfig) -> Result<()> {
```
sandreim, line 112: Would it make sense to use this path when restoring devices from state?. One option of doing so would be to make config an Enum with 2 variants : BlockDeviceConfig, BlockState. I am not sure if this would work out fine given the fact that we also need other stuff like guest memory.

src/vmm/src/vmm_config/drive.rs
acatangiu: That's right. This is over-engineered to do everything in a single pass, but it's not really necessary. Number of block devices is never high enough to introduce any performance hit by traversing this list multiple times.

I will refactor it to try and make it simpler.

src/vmm/src/vmm_config/drive.rs
acatangiu: Right, will do.

src/vmm/src/vmm_config/drive.rs
```diff
@@ -70,109 +81,114 @@ pub struct BlockDeviceConfig {
...
+/// Wrapper for the collection that holds all the Block Devices
 #[derive(Default)]
-pub struct BlockDeviceConfigs {
-    /// A list of `BlockDeviceConfig` objects.
-    pub config_list: VecDeque<BlockDeviceConfig>,
-    has_root_block: bool,
+pub struct BlockDevices {
+    /// The list of block devices.
+    pub list: VecDeque<Arc<Mutex<Block>>>,
+    /// Index of the root block device, if any.
```
ioanachirca, line 89:
```suggestion
    /// If there is a root device, it is on the first position.
```

src/vmm/src/vmm_config/drive.rs
```diff
@@ -70,109 +81,114 @@ pub struct BlockDeviceConfig {
...
+        let rate_limiter = block_device_config
+            .rate_limiter
+            .map(super::RateLimiterConfig::try_into)
+            .transpose()
+            .map_err(DriveError::CreateRateLimiter)?;
+
+        // Create and return the Block device
+        devices::virtio::Block::new(
+            block_device_config.drive_id,
+            block_file,
```
ioanachirca, line 186: WDYT about passing the block file path to the device instead of the file? This way we wouldn't need an outside config with the path when saving/restoring the device.
This can be done in the block serialization PR.

acatangiu:
This PR is a prerequisite for #1713

src/vmm/src/vmm_config/drive.rs
acatangiu: Sure, I don't have any strong reason against it.

src/vmm/src/vmm_config/drive.rs
acatangiu: done

src/vmm/src/builder.rs
```diff
@@ -1294,182 +1262,78 @@ pub mod tests {
...
-        assert_eq!(
-            format!("{}", err),
-            format!(
-                "Unable to seek the block device backing file due to invalid permissions or \
-                 the file was deleted/corrupted. Error number: {}",
-                io::Error::from_raw_os_error(0)
-            )
-        );
+        let err = AttachBlockDevice(io::Error::from_raw_os_error(0));
+        let _ = format!("{}{:?}", err, err);
```
andreeaflorescu, line 1266: What is the point of calling format here without checking anything?

src/devices/src/virtio/block/device.rs
```diff
@@ -95,10 +95,12 @@ pub struct Block {
     interrupt_status: Arc<AtomicUsize>,
     interrupt_evt: EventFd,
     pub(crate) queue_evts: [EventFd; 1],
-
     pub(crate) device_state: DeviceState,
 
     // Implementation specific fields.
+    id: String,
+    partuuid: Option<String>,
+    root_device: bool,
```
andreeaflorescu, line 103: Virtio block doesn't care about partuuid, if the block device is root or not and about the id. While I understand simplification of code as an argument for having these here, I don't think we should go this far.

How about we create a wrapper over `VirtioBlock` in which we extend the Block device struct with the fields that are Firecracker specific?

src/vmm/src/builder.rs
acatangiu: It tests that `Display` and `Debug` are implemented.

andreeaflorescu: Ok. It looks a bit like it is not needed to call that, but that's just a nit.

src/devices/src/virtio/block/device.rs
acatangiu: Would be against doing that incrementally? All of our devices have Firecracker-specific logic in them right now.
Decoupling them from that logic needs careful thinking and design so it works uniformly for all devices.

I would prefer not to burden the current step/PR with that responsibility. I don't think having some plain-old-data in the device (like partuuid and id and root bool) is dangerous in any way (like we already have `mmds` in net for example).

Later on, when we get to decoupling the devices from firecracker-specific impl/logic we can think of the proper solution.

What do you think?

andreeaflorescu: All virtio devices unfortunately have other state in them (mmds in net, rate limiter in net & block). But I would be worry to add even more non-virtio specific logic to it because we would need to remove it later on. This solution of merging `BlockDeviceConfig` with `VirtioDevice` is only a temporary one. For the final solution we would need to create that separate VirtioDevice wrapper for Firecracker anyway, so I was thinking we can do this now.

With this change, we are introducing API logic to the Virtio Device, which was previously in the VMM. So if we go down this road, after fixing https://github.com/firecracker-microvm/firecracker/issues/1713, we will need to create another issue for "decouple virtio devices from Firecracker specific RPC interface". The fact that we have an id in Virtio Block is RPC specific.

What are your thoughts on this?

acatangiu: I disagree that any of that data is rpc or api specific. The id in the device is used in mmio hashtable to be able to find a device, not needed for rpc or api in any way. The `partuuid` is needed for kernel cmdline which is the same, so is the `root_flag`.

The current PR(s) move things in the right direction on multiple fronts with this disadvantage correctly called out by you that devices now have to hold a `String` or a `bool` or some plain data. It is worth mentioning that they don't take dependency on any struct or crate because of this. It's just primitives needed by upper layers.

I agree with you that _at some point_ we will decouple the strict _generic_ block/net/vsock device from the firecracker-impl-specific parts but as of now we don't know these and can't do them. Maybe this ID will be used in rust-vmm mmio-indexing as well - you have to use _something_. Even the device in rust-vmm will need to modify the cmdline if it's not PCI so it might need partuuid or root-block info. The truth is we don't know yet.

As long as we don't know what the most future-proof solution is (we need to define that in rust-vmm), I say we go with the cleanest solution for Firecracker. What do you think?

andreeaflorescu: I don't think this is related at all to rust-vmm. IMHO, Virtio Devices should do only Virtio Device specific things. We can take the shortest path for now, but I think we are moving the problem of separation of concerns to lower abstraction levels by adding this fields here.

I dismissed my review.

acatangiu: That’s right, we should solve all of these in a unified way alongside the rest tracked in https://github.com/firecracker-microvm/firecracker/issues/1408

I’ll try to do that in the near future when I get some free cycles.

acatangiu:
@ioanachirca @aghecenco I've rebased, please take a look.

PR #1737, opened 2020-04-01
Create Vsock device during configuration instead of boot-time

### LINKED ISSUE #1709: Create Vsock device during configuration instead of boot-time
Following the more decoupled design we can now create the Vsock device on its configuration path, rather than saving a config and delaying creation to boot-time.
    
Such a model allows user-errors (like invalid resources or permissions) to be reported as part of the configuration step rather than later at attempted boot.

It would also remove the time window between configuration and boot where changes to the host system (creating Sockets, changing permissions, etc) would cause microVM boot failures (race condition between validation and commitment of resources).
    
Such a model should also decrease code complexity and should provide better efficiency to the process of configuring and starting a microVM (no more moving configurations around in memory and validating resources both at config time and boot-time).

acatangiu:
## Reason for This PR

Fixes #1709 

Prerequisite for #1713 

## Description of Changes

Following the more decoupled design we can now create the `Vsock` device on its configuration path, rather than saving a config and delaying creation to boot-time.

Such a model allows user-errors (like invalid resources or permissions) to be reported as part of the configuration step rather than later at attempted boot.

It would also remove the time window between configuration and boot where changes to the host system (creating Sockets, changing permissions, etc) would cause microVM boot failures (race condition between validation and commitment of resources).

Such a model should also decrease code complexity and should provide better efficiency to the process of configuring and starting a microVM (no more moving configurations around in memory and validating resources both at config time and boot-time).

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/vsock/device.rs
+++ b/src/devices/src/virtio/vsock/device.rs
@@ -107,6 +107,10 @@ where
         Self::with_queues(cid, backend, queues)
     }
 
+    pub fn id(&self) -> &str {
+        defs::VSOCK_DEV_ID
+    }
+
     pub fn cid(&self) -> u64 {
         self.cid
     }
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -17,7 +17,7 @@ use arch::InitrdConfig;
 use device_manager::legacy::PortIODeviceManager;
 use device_manager::mmio::MMIODeviceManager;
 use devices::legacy::Serial;
-use devices::virtio::MmioTransport;
+use devices::virtio::{MmioTransport, Vsock, VsockUnixBackend};
 use polly::event_manager::{Error as EventManagerError, EventManager};
 use seccomp::BpfProgramRef;
 use utils::eventfd::EventFd;
@@ -27,7 +27,6 @@ use vm_memory::{Bytes, GuestAddress, GuestMemoryMmap};
 use vmm_config::boot_source::BootConfig;
 use vmm_config::drive::BlockDeviceConfigs;
 use vmm_config::net::NetBuilder;
-use vmm_config::vsock::VsockDeviceConfig;
 use vmm_config::RateLimiterConfig;
 use vstate::{KvmContext, Vcpu, VcpuConfig, Vm};
 use {device_manager, VmmEventsObserver};
@@ -42,10 +41,6 @@ pub enum StartMicrovmError {
     CreateNetDevice(devices::virtio::net::Error),
     /// Failed to create a `RateLimiter` object.
     CreateRateLimiter(io::Error),
-    /// Failed to create the backend for the vsock device.
-    CreateVsockBackend(devices::virtio::vsock::VsockUnixBackendError),
-    /// Failed to create the vsock device.
-    CreateVsockDevice(devices::virtio::vsock::VsockError),
     /// Memory regions are overlapping or mmap fails.
     GuestMemoryMmap(vm_memory::Error),
     /// Cannot load initrd due to an invalid memory configuration.
@@ -99,10 +94,6 @@ impl Display for StartMicrovmError {
                 err
             ),
             CreateRateLimiter(ref err) => write!(f, "Cannot create RateLimiter: {}", err),
-            CreateVsockBackend(ref err) => {
-                write!(f, "Cannot create backend for vsock device: {:?}", err)
-            }
-            CreateVsockDevice(ref err) => write!(f, "Cannot create vsock device: {:?}", err),
             CreateNetDevice(ref err) => {
                 let mut err_msg = format!("{:?}", err);
                 err_msg = err_msg.replace("\"", "");
@@ -356,8 +347,8 @@ pub fn build_microvm(
 
     attach_block_devices(&mut vmm, &vm_resources.block, event_manager)?;
     attach_net_devices(&mut vmm, &vm_resources.network_interface, event_manager)?;
-    if let Some(vsock) = vm_resources.vsock.as_ref() {
-        attach_vsock_device(&mut vmm, vsock, event_manager)?;
+    if let Some(vsock) = vm_resources.vsock.get() {
+        attach_unixsock_vsock_device(&mut vmm, vsock, event_manager)?;
     }
 
     // Write the kernel command line to guest memory. This is x86_64 specific, since on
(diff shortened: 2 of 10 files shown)


src/vmm/src/vmm_config/vsock.rs
```diff
@@ -16,20 +42,125 @@ pub struct VsockDeviceConfig {
...
+    pub fn new() -> Self {
+        Self { inner: None }
+    }
+
+    /// Inserts a Unix backend Vsock in the store.
+    /// If an entry already exists, it will overwrite it.
+    pub fn insert(&mut self, cfg: VsockDeviceConfig) -> Result<()> {
+        // Make sure to drop the old one and remove the socket before creating a new one.
+        if let Some(existing) = self.inner.take() {
+            let _ = std::fs::remove_file(existing.uds_path);
```
sandreim, line 67: We are not checking/returning the remove_file error. 
```suggestion
            std::fs::remove_file(existing.uds_path)?;
```

src/devices/src/virtio/vsock/device.rs
```diff
@@ -75,6 +76,7 @@ where
     B: VsockBackend,
 {
     pub(crate) fn with_queues(
+        id: String,
```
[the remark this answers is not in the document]
acatangiu: That's right, I didn't want to introduce API breaking changes with this PR so it stays focused.

A `String` `id` is used in our device map which is hashed based on `(DeviceType, id)` so we can't just not use one at all for `Vsock`.

I will make a change to this PR to simply ignore the one from API (it's not useful since we only ever have a single vsock), and hardcode a vsock id string in the vsock implementation. A `vsock.id()` will just provide a static hardcoded id.

What do you think?

src/vmm/src/vmm_config/vsock.rs
```diff
@@ -16,20 +42,125 @@ pub struct VsockDeviceConfig {
...
 
-impl Display for VsockError {
-    fn fmt(&self, f: &mut Formatter) -> Result {
-        use self::VsockError::*;
-        match *self {
-            UpdateNotAllowedPostBoot => {
-                write!(f, "The update operation is not allowed after boot.",)
+/// A builder of Vsock with Unix backend from 'VsockDeviceConfig'.
+#[derive(Default)]
+pub struct VsockStore {
```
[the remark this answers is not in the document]
acatangiu: You're right, naming is all over the place. I will rename it to `VsockBuilder`

src/vmm/src/vmm_config/vsock.rs
acatangiu: This is attempting to remove any leftover uds. I thought we can ignore failures here and report errors in Vsock creation where the right error will be reported (permission errors for example).

Getting the right error here feels a bit like duplicating logic.

I will add a comment, or if you insist I will duplicate the error handling logic.

aghecenco: I agree that the error should be propagated and the backend creation should fail fast. If there's a leftover that can't be removed, it could indicate that something went wrong before, and give the user a chance to diagnose, or maybe someone wants to keep using the previous uds, or some other strange case that might lead to trouble.

src/vmm/src/vmm_config/vsock.rs
```diff
@@ -16,20 +42,125 @@ pub struct VsockDeviceConfig {
     pub uds_path: String,
 }
 
-/// Errors associated with `VsockDeviceConfig`.
-#[derive(Debug)]
-pub enum VsockError {
-    /// The update is not allowed after booting the microvm.
-    UpdateNotAllowedPostBoot,
+struct VsockAndUnixPath {
+    vsock: Arc<Mutex<Vsock<VsockUnixBackend>>>,
```
aghecenco, line 46: I think `Arc<Mutex<Vsock<VsockUnixBackend>>>` is worthy of its own type alias.

src/vmm/src/builder.rs
```diff
@@ -781,31 +772,23 @@ fn attach_net_devices(
...
+        .add_subscriber(unix_vsock.clone())
+        .map_err(RegisterEvent)?;
 
+    let id = String::from(unix_vsock.lock().unwrap().id());
+    // The device mutex mustn't be locked here otherwise it will deadlock.
     attach_mmio_device(
         vmm,
-        vsock.vsock_id.clone(),
-        MmioTransport::new(vmm.guest_memory().clone(), vsock_device)
+        id,
```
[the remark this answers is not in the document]
acatangiu: This cannot be removed as `Block` still uses it. Only the last merging PR can remove it (this or #1736 whichever is last to go in).

Furthermore, after looking a bit at the code implications I don't know if it's worth removing it.

Mmio device identification happens on a generic `VirtioDevice` based on tuple `(dev.type_id(), dev_id)` where `dev_id` is string.
`type_id()` is part of the `VirtioDevice` trait inteface, but `id()` isn't and doesn't belong there either. So we can't remove it from the param while still using generic `VirtioDevice`s in the inner hashmap. To remove it we'd need to define a wrapper trait that also adds `id()` to the list of virtio methods and use that as a type in the hashmap.

Don't feel it's worth it anymore.

src/vmm/src/vmm_config/vsock.rs
acatangiu: done

src/vmm/src/vmm_config/vsock.rs
acatangiu: done

src/vmm/src/builder.rs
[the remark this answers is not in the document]
aghecenco: Yeah I thought the Block PR would go in first so this PR would have needed to enclose all the modifications to remove `id` from all devices. I understand your point. The ID for the application that consumes a device is represented by that tuple. On its own (if you took the physical device out of the computer) it would still have an identifier (you can observe its type); that would be the `device_type` in the `VirtioDevice`, although a device doesn't necessarily need to be `virtio` to have an observable type :thinking: Anyway, putting the device back into the computer, its identity is extended, it's in a certain slot etc - we skip that; with the OS running, the device will have an identifier allocated. That's where the 2nd component comes in. So conceptually, the way I see it, the "id" is a layered thing composed of multiple identifiers that come from different facets of what a device represents. Put into Rust, it would be a tuple of ids coming from multiple traits. Not necessarily a wrapper, but maybe a `VirtioDevice + OtherDeviceTraitThatHasId`. I.e. thing that makes sense for a standalone device, plus thing that makes sense only when the device is plugged into Firecracker.

...That said, I can see how it's way out of scope for now. I still think it's the right thing to do, and I'd track it as a part of https://github.com/firecracker-microvm/firecracker/issues/1408. WDYT?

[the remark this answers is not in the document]
acatangiu: Yup. Everything you said makes perfect sense.

PR #1739, opened 2020-04-01
Implement Versionize for Arc, GuestAddress and std::sync::atomic::*

sandreim:
## Reason for This PR

Implements #1738 

## Description of Changes

Provided in commit messages.

- [ ] This functionality can be added in [`rust-vmm`](https://github.com/rust-vmm).

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria.]`
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [x] All commits in this PR are signed (`git commit -s`).
- [x] The reason for this PR is clearly provided (issue no. or explanation).
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] Any newly added `unsafe` code is properly documented.
- [x] Any API changes are reflected in `firecracker/swagger.yaml`.
- [x] Any user-facing changes are mentioned in `CHANGELOG.md`.

Issue #1715, opened 2020-03-26
Define a state serialization/deserialization interface for devices


sandreim:
We need to define a trait for saving/restoring the device state to/from a snapshot.

PR #1743, opened 2020-04-02
Define trait Persist: interface for saving/restoring a component from state.

sandreim:
## Reason for This PR

Implements #1715

## Description of Changes

Provided in commit message.
Contains all commits from #1739 that should be reviewed in that PR.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/snapshot/src/persist.rs
+++ b/src/snapshot/src/persist.rs
@@ -1,0 +1,25 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//! Defines an abstract interface for saving/restoring a component from state.
+#![deny(missing_docs)]
+
+/// An abstract interface for saving/restoring a component from state.
+pub trait Persist
+where
+    Self: Sized,
+{
+    /// The type of the object representing the state of the component.
+    type State;
+    /// The type of the object holding the constructor arguments.
+    type ConstructorArgs;
+    /// The type of the error that can occur while constructing the object.
+    type Error;
+
+    /// Returns the current state of the component.
+    fn save(&self) -> Self::State;
+    /// Constructs a component from a specified state.
+    fn restore(
+        constructor_args: Self::ConstructorArgs,
+        state: &Self::State,
+    ) -> Result<Self, Self::Error>;
+}
--- a/src/versionize/Cargo.toml
+++ b/src/versionize/Cargo.toml
@@ -10,6 +10,7 @@ bincode = "1.2.1"
 versionize_derive = { path = "../versionize_derive" }
 crc64 = "1.0.0"
 vmm-sys-util = ">=0.4.0"
+vm-memory = { version = ">=0.2.0" }
 
 [build-dependencies]
 proc-macro2 = ">=1.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -713,6 +713,7 @@ dependencies = [
  "serde_derive 1.0.105 (registry+https://github.com/rust-lang/crates.io-index)",
  "syn 1.0.17 (registry+https://github.com/rust-lang/crates.io-index)",
  "versionize_derive 0.1.0",
+ "vm-memory 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "vmm-sys-util 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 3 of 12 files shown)


src/snapshot/src/persist.rs
```diff
@@ -0,0 +1,25 @@
...
+    type ConstructorArgs;
+    /// The type of the error that can occur while constructing the object.
+    type Error;
+
+    /// Returns the current state of the component.
+    fn save(&self) -> Self::State;
+    /// Constructs a component from a specified state.
+    fn restore(
+        constructor_args: Self::ConstructorArgs,
+        state: &Self::State,
```
ioanachirca, line 23: Could we pass ownership of the State, or do we need to use the State after restoring the component?

sandreim: Usually that state will be part of a bigger structure, so we will pass a reference in the general case as we expect these to be larger.

src/snapshot/src/persist.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//! Defines an abstract interface for saving/restoring a component from state.
+#![deny(missing_docs)]
```
aghecenco, line 4: This only needs to be in the crate's `lib.rs`.

src/versionize/Cargo.toml
```diff
@@ -10,6 +10,7 @@ bincode = "1.2.1"
 versionize_derive = { path = "../versionize_derive" }
 crc64 = "1.0.0"
 vmm-sys-util = ">=0.4.0"
+vm-memory = { version = ">=0.2.0" }
```
aghecenco, line 13: I don't agree that the `versionize` crate should depend on abstractions of the guest memory. I think it shouldn't care about the guest memory at all. It defines serialization for structs. It doesn't even need to be virtualization related. When versioned serialization is available in `rust-vmm`, we'll have to figure out if `vm-memory` needs to pull it in as a dependency and learn to serialize its `GuestAddress`es. Personally I think it shouldn't. Anyway that's a discussion for later.

In the meantime, since we can't `impl Versionize for GuestAddress` in the VMM (not the `versionize` crate) and we don't support any mechanism like `versionize_this_unversionizable_member_with()`, I think we should shadow the structs that have un-versionizable members with "state" structs where said members are broken down into either nested state, or primitives.

For example, the [`virtio-net` device](https://github.com/firecracker-microvm/firecracker/blob/0b7ba621d4c8a3f8948e7ebce3421f0262208eba/src/devices/src/virtio/net/device.rs#L60) contains a [`Tap`](https://github.com/firecracker-microvm/firecracker/blob/0b7ba621d4c8a3f8948e7ebce3421f0262208eba/src/utils/src/net/tap.rs#L49), which in turns contains a `File`, which we won't save. We'll only save the tap's name in a separate struct.

sandreim: I agree that the dependency shouldn't be there, but I believe we would end up shadowing a lot of structures and that would imply copying the fields by hand from one struct to another. Also this would also make it impossible to nest such a struct into a struct that fully supports Versionize on all fields.

I think for the short term it makes sense to at least try to refactor the state structures so we do not end up implementing Versionize for other other crate structs and otherwise fall back  to the currently implemented solution. We can incrementally fix this once we decide on a good solution for versioning `remote` crate structs.

For the net device we will need to modify the device constructor to receive the Path also that we will serialize and keep the File outside the state structure.

alexandruag:
Just wanted to raise two questions that came out after our initial discussion around the `Persist` trait, in case they help with thinking about the future direction of the newly added functionality. 

First, do we plan to leverage the fact that a type implements `Persist` anywhere in particular? For example, are there plans to pass `Persist` trait objects around, or have functions generic in terms of parameters/return values that implement `Persist`? (The stateful objects will implement state-related functionality anyway, the question here is whether it's meaningful to expose it via a trait vs just having simple methods on every such object).

Second, if we do plan to leverage the `Persist` trait, do we expect building blocks coming from rust-vmm (or external components with state in general) to implement it as well?

aghecenco:
My 2 cents: I don't think we'll see `dyn Persist`s around, but I don't like dropping the trait altogether. If anything, it's a useful reminder for future users of what can and should be persisted and what not, just in case anyone ever tries to save/restore live objects or, generally, stuff that isn't state.

Issue #1544, opened 2020-01-24
[GuestMemory] use the GuestMemory traits instead of the GuestMemoryMmap struct where possible


serban300:
Right now we're always using the `GuestMemoryMmap` concrete implementation. Try to use the `GuestMemory` interface instead where possible.

This is not needed for the migration to `rust-vmm/vm-memory`.

AlexandruCihodaru:
Our implementation is still dependent on our custom crate and we can not easily do this.

PR #1744, opened 2020-04-02
Implement Vsock Device serialization

sandreim:
## Reason for This PR

Implements #1426.

## Description of Changes

Provided in commit message (last commit)
Please review only the last commit (the others can be reviewed separately in #1739, #1743)

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/vsock/device.rs
+++ b/src/devices/src/virtio/vsock/device.rs
@@ -5,6 +5,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 
+use snapshot::Persist;
 use std::result;
 /// This is the `VirtioDevice` implementation for our vsock device. It handles the virtio-level
 /// device logic: feature negociation, device configuration, and device activation.
@@ -22,14 +23,15 @@ use std::result;
 /// - a backend FD.
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
-
 use utils::byte_order;
 use utils::eventfd::EventFd;
+use versionize::{VersionMap, Versionize, VersionizeResult};
+use versionize_derive::Versionize;
 use vm_memory::GuestMemoryMmap;
 
 use super::super::super::Error as DeviceError;
 use super::super::{
-    ActivateError, ActivateResult, DeviceState, Queue as VirtQueue, VirtioDevice, VsockError,
+    ActivateError, ActivateResult, DeviceStatus, Queue as VirtQueue, VirtioDevice, VsockError,
     VIRTIO_MMIO_INT_VRING,
 };
 use super::packet::VsockPacket;
@@ -47,14 +49,9 @@ pub(crate) const EVQ_INDEX: usize = 2;
 pub(crate) const AVAIL_FEATURES: u64 =
     1 << uapi::VIRTIO_F_VERSION_1 as u64 | 1 << uapi::VIRTIO_F_IN_ORDER as u64;
 
-pub struct Vsock<B> {
-    cid: u64,
-    pub(crate) queues: Vec<VirtQueue>,
+pub struct Vsock<B: 'static> {
     pub(crate) queue_events: Vec<EventFd>,
     pub(crate) backend: B,
-    avail_features: u64,
-    acked_features: u64,
-    pub(crate) interrupt_status: Arc<AtomicUsize>,
     pub(crate) interrupt_evt: EventFd,
     // This EventFd is the only one initially registered for a vsock device, and is used to convert
     // a VirtioDevice::activate call into an EventHandler read event which allows the other events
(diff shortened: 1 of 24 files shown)


src/devices/src/virtio/vsock/device.rs
```diff
@@ -47,22 +49,63 @@ pub(crate) const EVQ_INDEX: usize = 2;
...
+    activated: bool,
+}
+
+impl<B: VsockBackend> Persist for Vsock<B> {
+    type State = VsockState;
+    type ConstructorArgs = VsockConstructorArgs<B>;
+    type Error = VsockError;
+
+    fn save(&self) -> Self::State {
+        let mut state = self.state.clone();
```
andreeaflorescu, line 89: Hm, I am not really certain this is the best way forward for a couple of reasons.

1. State is used during the normal operation of the device. Since this is the case, state should always be accurate, and shouldn't need to be updated before saving it. My expectation would be that `state` is a simple snapshot of the current device and it shouldn't alter in any way the device. This might sound a bit purist, because changing the value of `state.activated` shouldn't really affect the device.

2. With this approach we now have 2 fields that give the same information about the device being activated. We have `DeviceStatus` which also has the memory and the field `activated`. I think is safe to say that generally you don't want 2 fields with the same purpose. It gets a bit worse because even though both `DeviceStatus` and `activated` give the same information, the latter only has an accurate value after calling `save`.

3. Do we really want to call `clone` here?

I think we could have two good-ish options here:

1. Replace `DeviceStatus` with `activated` so we don't have to do the clone and alter the state of the device when doing save. This would be the easy way out I guess.
2. Create the state when doing `save`
3. Other options?

As we also discussed offline, I prefer the second variant because it is implying a clean separation of concerns and versionize tags are only needed for the state which is completely separated from the normal device operation. We need some PoC to check how much code is needed and if we get any slow downs, so we can indefinitely postpone that :))

src/devices/src/virtio/vsock/device.rs
```diff
@@ -47,22 +49,63 @@ pub(crate) const EVQ_INDEX: usize = 2;
...
+pub struct VsockConstructorArgs<B> {
+    mem: GuestMemoryMmap,
+    backend: B,
+}
+
+/// The Vsock serializable state.
+#[derive(Clone, Versionize)]
+pub struct VsockState {
+    cid: u64,
+    queues: Vec<VirtQueue>,
```
andreeaflorescu, line 76: I think it is better for VirtQueue to have an associated state. This way, you don't need to implement Versonize for GuestAddress. In the state, you can simply save the underlying `u64`.

src/versionize/src/lib.rs
```diff
@@ -64,8 +65,8 @@ pub trait Versionize {
     where
         Self: Sized;
 
-    /// Returns the `Self` type id.
-    fn type_id() -> std::any::TypeId
+    /// Returns an unique indetifier for the type.
```
andreeaflorescu, line 68: I think the previous name was better :D

src/devices/src/virtio/vsock/device.rs
sandreim: > Hm, I am not really certain this is the best way forward for a couple of reasons.
> 
>     1. State is used during the normal operation of the device. Since this is the case, state should always be accurate, and shouldn't need to be updated before saving it. My expectation would be that `state` is a simple snapshot of the current device and it shouldn't alter in any way the device. This might sound a bit purist, because changing the value of `state.activated` shouldn't really affect the device.
> 
>     2. With this approach we now have 2 fields that give the same information about the device being activated. We have `DeviceStatus` which also has the memory and the field `activated`. I think is safe to say that generally you don't want 2 fields with the same purpose. It gets a bit worse because even though both `DeviceStatus` and `activated` give the same information, the latter only has an accurate value after calling `save`.
> 
>     3. Do we really want to call `clone` here?
Yes, we need to either clone the state in this implementation and you need to construct a new state as suggested in 1/2.
> 
> 
> I think we could have two good-ish options here:
> 
>     1. Replace `DeviceStatus` with `activated` so we don't have to do the clone and alter the state of the device when doing save. This would be the easy way out I guess.
> 
>     2. Create the state when doing `save`
> 
>     3. Other options?
> 
> 
> As we also discussed offline, I prefer the second variant because it is implying a clean separation of concerns and versionize tags are only needed for the state which is completely separated from the normal device operation. We need some PoC to check how much code is needed and if we get any slow downs, so we can indefinitely postpone that :))

The PoC will be implemented soon. I wouldn't worry about the performance but I am worried that we could end up implementing Persist for a lot of structs and defeating the purpose of using code generation for persisting states. I think we should draw a line as to where we derive Versionize and where we define a separate State structure, derive Versionize for it and implement Persist for that object.

src/devices/src/virtio/vsock/device.rs
sandreim: GuestAddress is basically an u64. The benefit of having Versionize implemented for GuestAddress is that you do not need to create a separate state struct for VirtuQueue or any other POD state that has GuestAddress members.  Please see my above reply about defining a clear way of separating POD states from live state. There is a lot of benefit in readability and consistency when relying on code generation where this doesn't hinder the separation of concerns.

src/devices/src/virtio/vsock/device.rs
andreeaflorescu: PoC available here: #1747

src/devices/src/virtio/vsock/device.rs
```diff
@@ -47,22 +49,63 @@ pub(crate) const EVQ_INDEX: usize = 2;
...
     // continuous triggers from happening before the device gets activated.
     pub(crate) activate_evt: EventFd,
-    pub(crate) device_state: DeviceState,
+    pub(crate) device_status: DeviceStatus,
+    pub(crate) state: VsockState,
+}
+
+/// A helper structure that holds the usual constructor parameters that are not serialized in VsockState
+pub struct VsockConstructorArgs<B> {
+    mem: GuestMemoryMmap,
```
andreeaflorescu, line 68: While writing the PoC, I realized we have a bit of a logic problem here. We can serialize the device at any given point, even when it is not activated. When the device is not activated, the memory is not available because the memory is only configured with `DeviceState::Activated`. It is not a big deal because in practice (at least with static devices that cannot be unplugged) because the memory will be configured anyway, and so we can create a clone of it and pass it here.

BTW, we should make sure to use `GuestMemory` instead of GuestMemoryMmap at least for new code we are introducing. For background, please check: https://github.com/firecracker-microvm/firecracker/issues/1544

PR #1747, opened 2020-04-03
PoC for serialization of devices using a separate device state struct

andreeaflorescu:
Related to #1744 
Not to be merged. This was created so it is easier to leave comments and try to reach a conclusion about how to implement serialization of devices. Only the last commit is relevant.

The code is not well optimized and some improvements can be made, but hopefully it is enough to grasp what this different approach implies.

@sandreim @alexandruag @acatangiu @ioanachirca please take a look.

### DIFF
--- a/src/devices/src/virtio/queue.rs
+++ b/src/devices/src/virtio/queue.rs
@@ -6,9 +6,12 @@
 // found in the THIRD-PARTY file.
 
 use std::cmp::min;
+use std::convert::{From, Into};
 use std::num::Wrapping;
 use std::sync::atomic::{fence, Ordering};
 
+use versionize::{VersionMap, Versionize, VersionizeResult};
+use versionize_derive::Versionize;
 use vm_memory::{Address, ByteValued, Bytes, GuestAddress, GuestMemory, GuestMemoryMmap};
 
 pub(super) const VIRTQ_DESC_F_NEXT: u16 = 0x1;
@@ -140,7 +143,7 @@ impl<'a> DescriptorChain<'a> {
     }
 }
 
-#[derive(Clone)]
+#[derive(Copy, Clone, Versionize)]
 /// A virtio queue's parameters.
 pub struct Queue {
     /// The maximal size in elements offered by the device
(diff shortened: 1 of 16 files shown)


src/devices/src/virtio/vsock/device.rs
```diff
@@ -289,10 +353,83 @@ where
...
 mod tests {
+    use super::super::tests::{TestBackend, TestContext};
+    use super::*;
     use crate::virtio::vsock::defs::uapi;
+    use versionize::VersionMap;
 
-    use super::super::tests::TestContext;
-    use super::*;
+    #[test]
+    fn test_persistence() {
```
andreeaflorescu, line 362: This is just copy paste from Andrei's PR.

src/devices/src/virtio/vsock/device.rs
```diff
@@ -65,6 +69,66 @@ pub struct Vsock<B> {
...
+    mem: Option<GuestMemoryMmap>,
+    backend: B,
+}
+
+impl<B: VsockBackend> Persist for Vsock<B> {
+    type State = VsockState;
+    type ConstructorArgs = VsockConstructorArgs<B>;
+    type Error = VsockError;
+
+    fn save(self) -> Self::State {
```
acatangiu, line 92: This cannot consume `self`, saving state is not necessarily a final operation (the microVM should _be able to_ continue running).

It's also very hard to get exclusive ownership of a device (it's wrapped in `Arc` and shared between multiple components of Firecracker).

src/devices/src/virtio/queue.rs
```diff
@@ -165,6 +168,60 @@ pub struct Queue {
...
+            ready: state.ready,
+            desc_table: GuestAddress::new(state.desc_table),
+            avail_ring: GuestAddress::new(state.avail_ring),
+            used_ring: GuestAddress::new(state.used_ring),
+            next_avail: state.next_avail,
+            next_used: state.next_used,
+        }
+    }
+}
+
```
acatangiu, line 224: This is a non-negligible amount of hand-written code to avoid deriving `Versionize` on `GuestAddress`.

If this is the case in many places around the code-base I am inclined to just `Versionize` `GuestAddress` (at least for now - we don't know yet what the final solution in rust-vmm will be).

andreeaflorescu: I was not very explicit to why I think there should be a `QueueState`. I will try to incorporate this in the PR somehow.

Just a few thoughts:

Not implementing serialize for `GuestAddress` is just one of the reason for keeping a `QueueState` and it might not be a strong one. What I find to be a strong reason is that I would like the serialization to be completely separated from the device logic. For that matter, I think all the serialization related things (that can be moved), should be in a separate module called `snapshot.rs` or  something to that extend. The reason is that we are a bit behind in rust-mvm on serialization, and we didn't yet agree on what will be the format. We also didn't decided if we are going to have cross-version support. If we strongly tie the implementation of devices (which we know we want to consume from rust-vmm) with the serialization, than there is a higher risk and we might need to make this separation anyway. Since we didn't yet serialized the devices, I think it is a good opportunity to find a way by which future changes (i.e. when consuming devices from rust-vmm) are unlikely to to happen or do not affect a large portion of the code.

When looking at this code in particular, I don't have a huge concern because there are only 50 lines of code and they are going to be used by all the devices.

If you're also okay with this, I would say to discuss this at length on Monday offline.

src/devices/src/virtio/vsock/device.rs
andreeaflorescu: That's right. It just made things super simple later on because I was able to implement `Into` and `From`. For save we shouldn't consume self.

src/devices/src/virtio/queue.rs
alexandruag: If we look at `GuestAddress` as a primitive, then `Queue` becomes its own state, and we can simply use a type alias to define `QueueState` as `Queue`. We might just do that here for now, since `Queue` is one of the basic building blocks we plan to get entirely from rust-vmm anyway when it's ready (as opposed to a particular device, which might end up as custom glue on top of rust-vmm building blocks).

At the same time, we can also keep the code added here since it's quite self contained, but at the very least we have to decide how we think `QueueState` will end up looking (i.e. `GuestAddress` vs `u64`).

src/devices/src/virtio/vsock/device.rs
```diff
@@ -65,6 +69,66 @@ pub struct Vsock<B> {
     pub(crate) device_state: DeviceState,
 }
 
+#[derive(Clone, Versionize)]
+pub struct VsockState {
```
alexandruag, line 73: I think this is a good potential example why keeping state as constituent part of a device post-restore might often not be a good fit. In this particular case, as soon as the `QueueState` type is no longer identical to `Queue` itself, the vsock device ends up holding both a `Vec` of `Queue`s and one of `QueueState`s if it contains a `VsockState` member during normal operation.

andreeaflorescu:
Changes will be incorporated in PR #1755. Closing this PoC.

PR #1749, opened 2020-04-03
Implement serialization for RateLimiter

ioanachirca:
Signed-off-by: Ioana Chirca <chioana@amazon.com>

## Reason for This PR

Being able to serialize the RateLimiter is a requirement for serializing the Block (#1424) and Net (#1425) devices.

## Description of Changes

Implemented the Persist trait for the RateLimiter and the TokenBucket structs and derived the Versionize trait for them.

## License Acceptance

## PR Checklist

- [x] Np new added `unsafe` code.
- [x] No API changes.
- [x] No user-facing changes.

### DIFF
--- a/src/rate_limiter/src/lib.rs
+++ b/src/rate_limiter/src/lib.rs
@@ -49,11 +49,16 @@ extern crate timerfd;
 extern crate utils;
 #[macro_use]
 extern crate logger;
+extern crate snapshot;
+extern crate versionize;
+extern crate versionize_derive;
 
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::time::{Duration, Instant};
 use std::{fmt, io};
 use timerfd::{ClockId, SetTimeFlags, TimerFd, TimerState};
+
+pub mod persist;
 
 #[derive(Debug)]
 /// Describes the errors that may occur while handling rate limiter events.
@@ -472,7 +477,7 @@ impl Default for RateLimiter {
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use super::*;
     use std::thread;
     use std::time::Duration;
@@ -494,6 +499,14 @@ mod tests {
 
         fn get_processed_refill_time(&self) -> u64 {
             self.processed_refill_time
+        }
+
+        // After a restore, we cannot be certain that the last_update field has the same value.
+        pub fn partial_eq(&self, other: &TokenBucket) -> bool {
+            (other.capacity() == self.capacity())
+                && (other.one_time_burst() == self.one_time_burst())
+                && (other.refill_time_ms() == self.refill_time_ms())
+                && (other.budget() == self.budget())
         }
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -499,8 +499,11 @@ version = "0.1.0"
 dependencies = [
  "libc 0.2.68 (registry+https://github.com/rust-lang/crates.io-index)",
  "logger 0.1.0",
+ "snapshot 0.1.0",
  "timerfd 1.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "utils 0.1.0",
+ "versionize 0.1.0",
+ "versionize_derive 0.1.0",
 ]
 
 [[package]]
--- a/src/rate_limiter/Cargo.toml
+++ b/src/rate_limiter/Cargo.toml
@@ -9,3 +9,6 @@ timerfd = ">=1.0"
 
 logger = { path = "../logger" }
 utils = { path = "../utils" }
+snapshot = { path = "../snapshot" }
+versionize = { path = "../versionize" }
+versionize_derive = { path = "../versionize_derive" }
--- a/src/rate_limiter/src/persist.rs
+++ b/src/rate_limiter/src/persist.rs
@@ -1,0 +1,205 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Defines the structures needed for saving/restoring a RateLimiter.
+
+use super::*;
+use snapshot::Persist;
+use versionize::{VersionMap, Versionize, VersionizeResult};
+use versionize_derive::Versionize;
+
+/// State for saving a TokenBucket.
+#[derive(Versionize)]
+pub struct TokenBucketState {
+    size: u64,
+    one_time_burst: Option<u64>,
+    refill_time: u64,
+    budget: u64,
+    elapsed_ns: u64,
+}
+
+impl Persist for TokenBucket {
+    type State = TokenBucketState;
+    type ConstructorArgs = ();
+    type Error = ();
+
+    fn save(&self) -> Self::State {
+        TokenBucketState {
+            size: self.size,
+            one_time_burst: self.one_time_burst,
+            refill_time: self.refill_time,
+            budget: self.budget,
+            elapsed_ns: self.last_update.elapsed().as_nanos() as u64,
+        }
+    }
+
+    fn restore(_: Self::ConstructorArgs, state: &Self::State) -> Result<Self, Self::Error> {
+        let now = Instant::now();
+        let last_update = now
+            .checked_sub(Duration::from_nanos(state.elapsed_ns))
+            .unwrap_or(now);
+
+        let mut token_bucket =
+            TokenBucket::new(state.size, state.one_time_burst, state.refill_time);
+
+        token_bucket.budget = state.budget;
+        token_bucket.last_update = last_update;
+
+        Ok(token_bucket)
+    }
+}
+
+/// State for saving a RateLimiter.
+#[derive(Versionize)]
+pub struct RateLimiterState {
+    ops: Option<TokenBucketState>,
+    bandwidth: Option<TokenBucketState>,
+}
+
+impl Persist for RateLimiter {
+    type State = RateLimiterState;
+    type ConstructorArgs = ();
+    type Error = io::Error;
+
+    fn save(&self) -> Self::State {
+        RateLimiterState {
+            ops: self.ops.as_ref().map(|ops| ops.save()),
+            bandwidth: self.bandwidth.as_ref().map(|bw| bw.save()),
+        }
+    }
+
+    fn restore(_: Self::ConstructorArgs, state: &Self::State) -> Result<Self, Self::Error> {
+        let rate_limiter = RateLimiter {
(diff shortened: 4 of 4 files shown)


src/rate_limiter/src/lib.rs
```diff
@@ -224,6 +231,60 @@ impl TokenBucket {
...
+            one_time_burst: state.one_time_burst,
+            refill_time: state.refill_time,
+            budget: state.budget,
+            last_update,
+            processed_capacity: state.processed_capacity,
+            processed_refill_time: state.processed_refill_time,
+        })
+    }
+}
+
```
acatangiu, line 287: I believe we should avoid the `TokenBucketState` manual code and derive `Versionize` directly on `TokenBucket`. It's composed of primitives and it's very well contained, it does not rely on any external elements.

ioanachirca: Do we consider `Instant` to be a primitive? there is no straight forward way to serialize it, we have to provide some methods that depend on the meaning of the `Instant`, in this case as a last_update field.

andreeaflorescu: How about we first decide on the guidelines for ser/de so that we can have a uniform implementation for all devices/other things that have state? We can also talk about exceptions to the rule, which I believe is something you are talking about here. We have a sync in a few hours.

acatangiu: Indeed. Manual translation is required here with or without using an `Instant` object.

Whichever way we try to represent that `Instant`, the actual value still **has to be post-processed** to take into consideration the time elapsed between serialization and deserialization.

Disregard my original comment.

src/rate_limiter/src/lib.rs
```diff
@@ -497,6 +621,15 @@ mod tests {
         }
     }
 
+    // After a restore, we cannot be certain that the last_update field has the same value.
+    fn equal_token_buckets(restored_tb: &TokenBucket, tb: &TokenBucket) -> bool {
```
aghecenco, line 625: nit: I'd move this inside `impl TokenBucket` with `&self` standing for `tb` and the other parameter being the restored one (or vice versa, doesn't matter as long as the convention is made obvious by naming or comments)

src/rate_limiter/src/lib.rs
```diff
@@ -471,6 +532,69 @@ impl Default for RateLimiter {
...
+        let ops: Option<TokenBucketState> = if let Some(ops) = &self.ops {
+            Some(ops.save())
+        } else {
+            None
+        };
+        let bandwidth: Option<TokenBucketState> = if let Some(bandwidth) = &self.bandwidth {
+            Some(bandwidth.save())
+        } else {
+            None
+        };
```
acatangiu, line 558: nit:
```suggestion
        let ops: self.ops.as_ref().map(|ops| ops.save());
        let bandwidth: self.bandwidth.as_ref().map(|bw| bw.save());
```

src/rate_limiter/src/lib.rs
```diff
@@ -224,6 +231,60 @@ impl TokenBucket {
...
+    fn restore(_: Self::ConstructorArgs, state: &Self::State) -> Result<Self, Self::Error> {
+        // When restoring last_update, we create a new Instant::now(), and then subtract the
+        // nanos elapsed since the last update took place, which were captured during save().
+        let elapsed = state.last_update;
+        let now = Instant::now();
+        let last_update = if let Some(instant) = now.checked_sub(Duration::from_nanos(elapsed)) {
+            instant
+        } else {
+            now
+        };
```
acatangiu, line 274: nit:
```suggestion
        let last_update = now.checked_sub(Duration::from_nanos(elapsed)).unwrap_or(now); 
```

src/rate_limiter/src/lib.rs
acatangiu: Yes, it would look better like Alexandra said.
FYI: to do that, you can add methods to `impl Whatever {}` inside `mod tests {}` too (as long as Whatever is in the current crate).

Since this is only tests you can even impl PartialEq and then just use `==`.
```rust
mod tests {
    impl PartialEq for TokenBucket {
        fn eq(&self, other: &Self) -> bool {...}
    }
}
```

src/rate_limiter/src/lib.rs
ioanachirca: Done.

src/rate_limiter/src/lib.rs
ioanachirca: Done.

src/rate_limiter/src/lib.rs
ioanachirca: I moved the method to `impl TokenBucket`. TokenBucket is already deriving `PartialEq` and this led to conflicts when trying to derive it in the `tests` module.

aghecenco: Yeah `TokenBucket` derives `PartialEq` because `RateLimiter` implements `PartialEq` and relies on comparing exact matches of `TokenBuckets`... This would be a *more* partial Eq but Rust doesn't have anything more flexible in terms of equivalence (the `Eq` trait exists but relies on `PartialEq` for comparison and only exists in a separate form to guarantee equivalence). Maybe you could get away with a `PartialOrd` in the tests module but it's hacky and doesn't really represent an ordering relationship. So I think it's better to leave the fn in `impl TokenBucket`

ioanachirca:
@aghecenco, @acatangiu I've addressed your comments, PTAL!

src/rate_limiter/src/lib.rs
```diff
@@ -224,6 +231,58 @@ impl TokenBucket {
     }
 }
 
+/// State for saving a TokenBucket.
+#[derive(Versionize)]
+pub struct TokenBucketState {
```
acatangiu, line 236: nit: I would add a separate file/module for serialization in general.

acatangiu:
> At a first glance, it looks like last_update, processed_capacity and processed_refill_time are not state, but Rate Limiter internal fields used for processing the budget. They should probably not be saved.
Temporarily blocking this PR to see if we can separate the state from the internal rate limiter logic.

Good catch!
`processed_capacity` and `processed_refill_time` can indeed be re-calculated at deserialization.
`last_update` is part of state that needs saving.

andreeaflorescu:
Right. `last_update` needs to be renamed a bit though because it will tell us the time elapsed.

src/rate_limiter/src/lib.rs
```diff
@@ -471,6 +519,52 @@ impl Default for RateLimiter {
...
+impl Persist for RateLimiter {
+    type State = RateLimiterState;
+    type ConstructorArgs = ();
+    type Error = io::Error;
+
+    fn save(&self) -> Self::State {
+        RateLimiterState {
+            ops: self.ops.as_ref().map(|ops| ops.save()),
+            bandwidth: self.bandwidth.as_ref().map(|bw| bw.save()),
+            timer_active: self.timer_active,
```
andreeaflorescu, line 539: I don't think we need to save `timer_active`. We actually need to know how much of the timer is left so that the restore is also correct. If the timer was armed, instead of starting off at the remaining duration, we start the timer at  `TIMER_REFILL_STATE`.

I don't have enough knowledge about the rate limiter so I don't really know what is the appropriate state here.

One proposal would be to define a `reset_time_ms` u32 value instead with the following meaning:
- value = 0 -> timer is disarmed -> we just initialize the timer with the default state & set timer_active = false
- 0 < value <= REFILL_TIMER_INTERVAL_MS -> we initialize the timer and set the state to `TimerState::Oneshot(Duration::from_millis(value))` and set timer_active = true
- value > REFILL_TIMER_INTERVAL_MS -> the state is wrong and we should exit with an error.

src/rate_limiter/src/lib.rs
```diff
@@ -495,6 +589,15 @@ mod tests {
...
             self.processed_refill_time
         }
+
+        // After a restore, we cannot be certain that the last_update field has the same value.
+        fn equal_to(&self, other: &TokenBucket) -> bool {
+            (other.capacity() == self.capacity())
+                && (other.one_time_burst() == self.one_time_burst())
+                && (other.refill_time_ms() == self.refill_time_ms())
+                && (other.budget() == self.budget())
+                && (other.last_update >= self.last_update)
```
andreeaflorescu, line 599: This is a bit weird to be present in an `equal_to` function. How about we implement `PartialEq` instead and ignore that field?

src/rate_limiter/src/lib.rs
acatangiu: The rate-limiter budget is passively updated by its users (when they consume, they also update the budget based on elapsed time).
When there is insufficient budget to serve a request, the rate-limiter _blocks_, and users of the ratelimiter might _block_ as well (our devices in Firecracker do so). A one-shot timer is activated to _actively_ update the ratelimiter budget in case all its users are blocked and there is no one to _passively_ update it.

This timer has hardcoded period of 100ms. I think we can avoid the added complexity of serializing and deserializing remaining duration and just let it run 100ms after deserialization.

acatangiu: I would actually go a step further and not even serialize whether the rate-limiter has pending timer.

I'm now thinking to just serialize budget-related state. On deserialization start off _unblocked_ (regardless of blocked/unblocked at serialization).

Post-deser all devices (all async objects really) need to be kicked anyway, so they will attempt to consume rate-limiter budget. If there is enough, it will go through, otherwise live ratelimiter code logic will do the blocking and timer setup.

This reduces ser/deser complexity, while being correct and fair behavior.

What do you think?

andreeaflorescu: Not serializing the `timer_active` at all and just not have the timer active sounds good to me.

ioanachirca: Removed `timer_active`, now the rate limiter starts with the timer turned off.

src/rate_limiter/src/lib.rs
ioanachirca: `PartialEq` is already derived for `TokenBucket`.  See [comment](https://github.com/firecracker-microvm/firecracker/pull/1749#discussion_r406274757)

src/rate_limiter/src/lib.rs
ioanachirca: will do.

src/rate_limiter/src/lib.rs
andreeaflorescu: `equal_to` is misleading though because you are not checking equality. You could still ignore the `last_update` field and call it `partial_eq` or something.

ioanachirca: Done.

PR #1755, opened 2020-04-07
Implement (de)serialization for block devices

### LINKED ISSUE #1424: Implement (de)serialization for Block Devices

ioanachirca:
Fixes #1424

## Description of Changes

- Introduce a VirtioDeviceState that captures the generic Virtio fields that all devices have
- Implement the Persist trait for the block device in order to be able to save/restore it
- Derive Versionize for the BlockState

## License Acceptance

## PR Checklist

- [x] No newly added `unsafe` code.
- [x] No API changes.
- [x] No user-facing changes.

### DIFF
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -13,4 +13,7 @@ utils = { path = "../utils" }
 net_gen = { path = "../net_gen" }
 polly = { path = "../polly" }
 rate_limiter = { path = "../rate_limiter" }
+snapshot = { path = "../snapshot" }
+versionize = { path = "../versionize" }
+versionize_derive = { path = "../versionize_derive" } 
 virtio_gen = { path = "../virtio_gen" }
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -15,5 +15,5 @@ polly = { path = "../polly" }
 rate_limiter = { path = "../rate_limiter" }
 snapshot = { path = "../snapshot" }
 versionize = { path = "../versionize" }
-versionize_derive = { path = "../versionize_derive" } 
+versionize_derive = { path = "../versionize_derive" }
 virtio_gen = { path = "../virtio_gen" }
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -7,9 +7,10 @@
 
 use std::cmp;
 use std::convert::From;
-use std::fs::File;
+use std::fs::{File, OpenOptions};
 use std::io::{self, Seek, SeekFrom, Write};
 use std::os::linux::fs::MetadataExt;
+use std::path::PathBuf;
 use std::result;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
@@ -81,26 +82,27 @@ fn build_disk_image_id(disk_image: &File) -> Vec<u8> {
 pub struct Block {
     // Host file and properties.
     disk_image: File,
+    pub(crate) disk_image_path: String,
     disk_nsectors: u64,
     disk_image_id: Vec<u8>,
 
     // Virtio fields.
-    avail_features: u64,
-    acked_features: u64,
+    pub(crate) avail_features: u64,
+    pub(crate) acked_features: u64,
     config_space: Vec<u8>,
     pub(crate) activate_evt: EventFd,
 
     // Transport related fields.
-    queues: Vec<Queue>,
-    interrupt_status: Arc<AtomicUsize>,
+    pub(crate) queues: Vec<Queue>,
+    pub(crate) interrupt_status: Arc<AtomicUsize>,
     interrupt_evt: EventFd,
     pub(crate) queue_evts: [EventFd; 1],
     pub(crate) device_state: DeviceState,
 
     // Implementation specific fields.
-    id: String,
-    partuuid: Option<String>,
-    root_device: bool,
+    pub(crate) id: String,
+    pub(crate) partuuid: Option<String>,
+    pub(crate) root_device: bool,
     pub(crate) rate_limiter: RateLimiter,
 }
 
@@ -110,12 +112,17 @@ impl Block {
     /// The given file must be seekable and sizable.
     pub fn new(
         id: String,
-        mut disk_image: File,
         partuuid: Option<String>,
+        disk_image_path: String,
         is_disk_read_only: bool,
         is_disk_root: bool,
         rate_limiter: RateLimiter,
     ) -> io::Result<Block> {
+        let mut disk_image = OpenOptions::new()
+            .read(true)
+            .write(!is_disk_read_only)
+            .open(PathBuf::from(&disk_image_path))?;
+
         let disk_size = disk_image.seek(SeekFrom::End(0))? as u64;
 
         let mut avail_features = (1u64 << VIRTIO_F_VERSION_1) | (1u64 << VIRTIO_BLK_F_FLUSH);
@@ -134,6 +141,7 @@ impl Block {
             partuuid,
             disk_image_id: build_disk_image_id(&disk_image),
             disk_image,
+            disk_image_path: disk_image_path.clone(),
             disk_nsectors: disk_size / SECTOR_SIZE,
             avail_features,
             acked_features: 0u64,
@@ -412,14 +420,22 @@ pub(crate) mod tests {
     pub fn default_block() -> Block {
         // Create backing file.
         let f = TempFile::new().unwrap();
-        let block_file = f.into_file();
-        block_file.set_len(0x1000).unwrap();
+        f.as_file().set_len(0x1000).unwrap();
 
         // Rate limiting is enabled but with a high operation rate (10 million ops/s).
         let rate_limiter = RateLimiter::new(0, None, 0, 100_000, None, 10).unwrap();
 
         let id = "test".to_string();
-        Block::new(id, block_file, None, true, false, rate_limiter).unwrap()
+        // The default block device is read-write and non-root.
+        Block::new(
+            id,
+            None,
+            f.as_path().to_str().unwrap().to_string(),
+            false,
+            false,
+            rate_limiter,
+        )
+        .unwrap()
     }
 
     pub fn default_mem() -> GuestMemoryMmap {
@@ -480,8 +496,7 @@ pub(crate) mod tests {
 
         assert_eq!(block.device_type(), TYPE_BLOCK);
 
-        let features: u64 =
-            (1u64 << VIRTIO_BLK_F_RO) | (1u64 << VIRTIO_F_VERSION_1) | (1u64 << VIRTIO_BLK_F_FLUSH);
+        let features: u64 = (1u64 << VIRTIO_F_VERSION_1) | (1u64 << VIRTIO_BLK_F_FLUSH);
 
         assert_eq!(block.avail_features_by_page(0), features as u32);
         assert_eq!(block.avail_features_by_page(1), (features >> 32) as u32);
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -294,7 +294,11 @@ impl VirtioDevice for Block {
         TYPE_BLOCK
     }
 
-    fn queues(&mut self) -> &mut [Queue] {
+    fn queues(&self) -> &[Queue] {
+        &self.queues
+    }
+
+    fn queues_mut(&mut self) -> &mut [Queue] {
         &mut self.queues
     }
 
--- a/src/devices/src/virtio/block/persist.rs
+++ b/src/devices/src/virtio/block/persist.rs
@@ -1,0 +1,146 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Defines the structures needed for saving/restoring block devices.
+
+use std::io;
+use std::sync::atomic::AtomicUsize;
+use std::sync::Arc;
+
+use rate_limiter::{persist::RateLimiterState, RateLimiter};
+use snapshot::Persist;
+use versionize::{VersionMap, Versionize, VersionizeResult};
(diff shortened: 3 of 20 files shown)


src/devices/src/virtio/block/device.rs
```diff
@@ -270,6 +270,10 @@ impl VirtioDevice for Block {
         &mut self.queues
     }
 
+    fn queues_clone(&self) -> Vec<Queue> {
+        self.queues.clone()
+    }
```
acatangiu, line 289: This is kinda unnecessary. Users of the `VirtioDevice` can always do
```rust
let cloned_queues = virtio_dev.queues().to_vec();
```
instead of exposing a new trait function to explicitly clone the queues.

src/devices/src/virtio/device.rs
```diff
@@ -5,20 +5,46 @@
...
 }
 
+/// State for saving the generic fields of a VirtioDevice.
+#[derive(Debug, Versionize)]
+pub struct VirtioDeviceState {
+    pub avail_features: u64,
+    pub acked_features: u64,
+    pub device_type: u32,
+    pub queues: Vec<Queue>,
+    pub interrupt_status: AtomicUsize,
```
acatangiu, line 32: This can just be `usize`, the state doesn't need to worry about synchronization.

src/devices/src/virtio/device.rs
```diff
@@ -105,4 +134,15 @@ pub trait VirtioDevice: AsAny + Send {
     fn reset(&mut self) -> Option<(EventFd, Vec<EventFd>)> {
         None
     }
+
+    fn state(&self) -> VirtioDeviceState {
```
acatangiu, line 138: Not sure if we should include this in `trait VirtioDevice` or add it to a child trait, smth like:
```rust
trait VirtioDeviceSerialization : VirtioDeviceState { // needs a better name
    fn state(&self) -> VirtioDeviceState { ... }
}
```

I am personally inclined towards the latter so that we don't pollute the `VirtioDevice` trait for users that don't need serialization.

src/devices/src/virtio/block/device.rs
```diff
@@ -342,14 +356,65 @@ impl VirtioDevice for Block {
...
+impl Persist for Block {
+    type State = BlockState;
+    type ConstructorArgs = BlockConstructorArgs;
+    type Error = io::Error;
+
+    fn save(&self) -> Self::State {
+        BlockState {
+            disk_path: self.disk_image_path.clone(),
+            virtio_state: self.state(),
+            rate_limiter_state: self.rate_limiter.save(),
```
acatangiu, line 379: nit: we should be consistent in naming across components: `dev.state()` vs `rate_limiter.save()`

src/devices/src/virtio/queue.rs
```diff
@@ -140,7 +143,7 @@ impl<'a> DescriptorChain<'a> {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, PartialEq, Versionize)]
```
andreeaflorescu, line 146: This is pending the decision on how we want to create the state for devices in general, but just wanted to lay out some disadvantages to this approach on queues.

The problem at hand here is that we are adding `Versionize` to the most common virtio device building block. This is a component that we want to consume from rust-vmm and it would be far easier to do so if the `Versonize` can be hidden behind a feature. For example, if we would have `Versonize` on `QueueState` instead of on `Queue`, the virtio crate could have a feature called `versioned_serialization`.

In the virtio `Cargo.toml` file:
```rust
....

# This way the dependencies for snapshot are only pulled when
# the feature `versioned_serialization` is used.

snapshot = { version = "0.1.0", optional = true }
versionize = { version = "0.1.0", optional = true }
versionize_derive = { version = "0.1.0", optional = true }

[features]
versioned_serialization = ["snapshot", "versionize", "versionize_derive"]
```

The `Queue` implementation stays the same irrespective of the feature `versioned_serialization` being specified or not. And the `QueueState` would be gated by the `versioned_serialization` feature as follows:

```rust
#[cfg(feature = "versioned_serialization")]
#[derive(Clone, Debug, Versionize)]
pub struct QueueState {
    max_size: u16,
    pub size: u16,
    pub ready: bool,
    pub desc_table: u64,
    pub avail_ring: u64,
    pub used_ring: u64,
    next_avail: Wrapping<u16>,
    next_used: Wrapping<u16>,
}
```
We don't need to have features in Firecracker of course, but I believe this approach would make it easier to consume the virtio building blocks from rust-vmm and it offers a cleaner separation of concerns, the concerns here being the virtio device functionality and the versioned serialization. We should also consider that `Queue` and `QueueState` are almost the same thing only now. If we add split queues or other features, this might not be the case any more.

acatangiu: I am inclined to not duplicate all this state since `Queue` is a stable object, and is completely comprised of plain-old-data.

@andreeaflorescu I think we can achieve what you want in rust-vmm having a `versioned_serialization` feature with the current code as well by deriving `Versionize` only when the feature is present:

```rust
#[derive(Debug, Clone)]
#[cfg_attr(feature = "versioned_serialization", derive(Versionize))]
pub struct Queue { ... }
```

without duplicating and maintaining a shadow state structure.

src/devices/src/virtio/device.rs
ioanachirca: Done.

src/devices/Cargo.toml
```diff
@@ -13,4 +13,10 @@ utils = { path = "../utils" }
 net_gen = { path = "../net_gen" }
 polly = { path = "../polly" }
 rate_limiter = { path = "../rate_limiter" }
+snapshot = { path = "../snapshot" }
+versionize = { path = "../versionize" }
+versionize_derive = { path = "../versionize_derive" }
 virtio_gen = { path = "../virtio_gen" }
+
+[dev-dependencies]
+tempfile = ">=3.0.2"
```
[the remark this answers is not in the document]
ioanachirca: Done.

src/devices/src/virtio/device.rs
```diff
@@ -5,20 +5,46 @@
...
+pub struct VirtioDeviceState {
+    pub avail_features: u64,
+    pub acked_features: u64,
+    pub device_type: u32,
+    pub queues: Vec<Queue>,
+    pub interrupt_status: AtomicUsize,
+    pub activated: bool,
+}
+
+impl PartialEq for VirtioDeviceState {
```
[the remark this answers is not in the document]
ioanachirca: Done.

src/devices/src/virtio/block/device.rs
ioanachirca: I'll remove it if it seems like VirtioDevice trait is not the place for it, but I included the method because it simplifies saving the state (having access to all these methods means we can directly implement `save` here and not have duplicate code in the devices).

src/devices/src/virtio/device.rs
ioanachirca: How about `VirtioDeviceSave`?

src/devices/src/virtio/block/device.rs
ioanachirca: Done. But this led to some interesting casts (e.g. `<Block as VirtioDeviceSave>::save(&self)`) to be able to differentiate between the `save` from `VirtioDeviceSave` and the one from `Persist`.

src/devices/src/virtio/block/device.rs
```diff
@@ -126,6 +139,7 @@ impl Block {
         Ok(Block {
             disk_image_id: build_disk_image_id(&disk_image),
```
acatangiu, line 140: We're not saving this, but rebuilding it from File. Which is correct, but it got me thinking how do we protect ourselves against getting a _wrong_ File on restore...

Would it make sense to both save the `disk_image_id` and also compute it from File on restore and sanity check them against each other?

Same for `disk_size` which is used to to compute other internals.

src/devices/src/virtio/device.rs
```diff
@@ -106,3 +112,28 @@ pub trait VirtioDevice: AsAny + Send {
...
+    pub acked_features: u64,
+    pub device_type: u32,
+    pub queues: Vec<Queue>,
+    pub interrupt_status: usize,
+    pub activated: bool,
+}
+
+/// Trait for saving the state of a VirtioDevice.
+pub trait VirtioDeviceSave : VirtioDevice {
+    fn save(&self) -> VirtioDeviceState {
```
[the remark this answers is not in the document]
acatangiu: ( I know I was the one who originally suggested the separate trait 🙈 )

src/devices/src/virtio/block/device.rs
acatangiu: The casts go away if we do it like suggested in https://github.com/firecracker-microvm/firecracker/pull/1755#discussion_r406124926

src/devices/src/virtio/queue.rs
andreeaflorescu: Yap, that's true. I like it more completely separated because you only need to use features at the `lib.rs` level so it should be easier to maintain. This is a soft requirement though.

We can leave it as it is in this PR, and only change it in the future in this case:
> We should also consider that Queue and QueueState are almost the same thing only now. If we add split queues or other features, this might not be the case any more.

The queue state can be the same irrespective on the type of queue (split/packed). But we might add some helper fields depending on the queue type. When it comes to that, we can easily split the structures in two (Queue and QueueState).

src/devices/src/virtio/block/device.rs
ioanachirca: Initially I tried that, but the problem is that `queues` takes a mutable reference to `self` and this method doesn't (and shouldn't if we want to call it from the `Persist` `save` context).

acatangiu: Yes, I see now. Seems like we do need to modify `trait VirtioDevice` in one way or another.

It is better to rename the existing `queues()` to `queues_mut()` since it provides a mut reference, and add a immutable getter for queues that doesn't clone, just provides `&[queue]`.

The reason for that, extracted from [this well-written answer](https://stackoverflow.com/a/35391084):

> Secondly, you should not clone in a getter. This is hugely wasteful if all the user wants to do is, for example, read from the value. It's better to return an immutable borrow, from which the user can clone if they need to.

src/devices/src/virtio/queue.rs
acatangiu: It's not clear which is the better long-term option to me either. Let's discuss it live in our sync-up meeting.

src/devices/src/virtio/block/device.rs
ioanachirca: Done.

src/devices/src/virtio/queue.rs
ioanachirca: We decided that it's best to have a separate `QueueState`. This allows possible future changes to `Queue` that will not affect the state and also has the benefit of not having to versionize the `GuestAddress`.

src/devices/src/virtio/device.rs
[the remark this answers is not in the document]
ioanachirca: Sure, I does look cleaner this way.

src/devices/src/virtio/block/device.rs
ioanachirca: That's a good idea, but I'm not sure how we can decide if this is an error or not.
For example, if the backing file gets bigger it should not be a problem.

Should we log a warning maybe?

src/devices/src/virtio/persist.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub avail_features: u64,
+    pub acked_features: u64,
+    pub device_type: u32,
+    pub queues: Vec<QueueState>,
+    pub interrupt_status: usize,
+    pub activated: bool,
+}
+
+impl VirtioDeviceState {
+    pub fn from_device(device: &dyn VirtioDevice) -> Self {
```
sandreim, line 88: Can we also implement Persist here? Personally I do not like the current solution as it looks like half the interface. Can we work around the usage of a Trait object so we can use Associated types ?

src/versionize/src/primitives.rs
```diff
@@ -61,6 +63,7 @@ impl_versionize!(f32);
 impl_versionize!(f64);
 impl_versionize!(char);
 impl_versionize!(String);
+impl_versionize!(PathBuf);
```
sandreim, line 66: See my prev comments.

src/devices/src/virtio/persist.rs
acatangiu: I think that would imply making `VirtioDeviceState<T: VirtioDevice>`. I'm not sure it's nicer.

Since we don't have a clear alternative here, I suggest we move forward with this one. It can be changed down the line when someone brings a better alternative (it's a two-way door, does not impose constraints on format).

What do you think?

src/devices/src/virtio/block/device.rs
```diff
@@ -110,12 +112,17 @@ impl Block {
...
         partuuid: Option<String>,
+        disk_image_path: &PathBuf,
         is_disk_read_only: bool,
         is_disk_root: bool,
         rate_limiter: RateLimiter,
     ) -> io::Result<Block> {
+        let mut disk_image = OpenOptions::new()
+            .read(true)
+            .write(!is_disk_read_only)
+            .open(disk_image_path)?;
```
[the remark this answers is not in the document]
ioanachirca: Done.

src/devices/src/virtio/block/persist.rs
```diff
@@ -0,0 +1,147 @@
...
+
+use crate::virtio::persist::VirtioDeviceState;
+use crate::virtio::{DeviceState, Queue};
+
+#[derive(Versionize)]
+pub struct BlockState {
+    id: String,
+    partuuid: Option<String>,
+    root_device: bool,
+    disk_path: PathBuf,
```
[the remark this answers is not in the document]
ioanachirca: Done.

src/devices/src/virtio/persist.rs
sandreim: It makes sense to go ahead with the current impl and rework this later.

PR #1772, opened 2020-04-10
MMDS JSON response support

### LINKED ISSUE #1775: Add MMDS JSON response support
MMDS is the MicroVM Metadata Service, implemented by Firecracker to share information from host to guest, or to share information between guest applications. MMDS data store is the storage for such information, and it consists in a stored JSON (https://www.json.org/json-en.html). At the moment, an application can query the data store and retrieve information only in a non-JSON format. The current MMDS format output has 0 depth, and to get a subtree of the entire JSON from the data store, an application would have to issue multiple requests to form it, which can cause performance issues and bad user experience.

iulianbarbu:
## Reason for This PR

Closes #1775. 

## Description of Changes

This PR adds full JSON support for MMDS, on both storage and retrieving metadata.
JSON response format is activated if an `Accept: application/json` header is present
in the MMDS request. Added relevant documentation in `mmds.md`. Further detailed documentation
is provided in #1800.

IMDS format response is kept intact.

## License Acceptance

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@
 - Added a [guide](docs/devctr-image.md) for updating the dev container image.
 - Added a new API call, `PUT /mmds/config`, for configuring the
   `MMDS` with a custom valid link-local IPv4 address.
-
+- Added experimental JSON response format support for MMDS guest applications
+  requests.
+   
 ### Fixed
 - Added `--version` flag to both Firecracker and Jailer.
 
@@ -32,6 +34,14 @@
   updates to existing configurations.
 - `PATCH /network-interfaces/{id}` only allowed post-boot. Use `PUT` for
   pre-boot updates to existing configurations.
+- Changed returned status code from `500 Internal Server Error` to `501 Not Implemented`,
+  for queries on the MMDS endpoint in IMDS format, when the requested resource value type
+  is unsupported.
+- Allowed the MMDS data store to be initialized with all supported JSON types.
+  Retrieval of these values within the guest, besides String, Array, and Dictionary,
+  is only possible in JSON mode.
+- `PATCH` request on `/mmds` before the data store is initialized returns 
+  `403 BadRequest`.
 
 ## [0.21.0]
 
@@ -211,7 +221,7 @@
 - When running with `jailer` the location of the API socket has changed to
   `<jail-root-path>/api.socket` (API socket was moved _inside_ the jail).
 - `PUT` and `PATCH` requests on `/mmds` with data containing any value type
-  other than `String`, `Array`, `Object` will return status code 400.
+  other than `String`, `Array`, `Object` will returns status code 400.
 - Improved multiple error messages.
 - Removed all kernel modules from the recommended kernel config.
 
--- a/docs/mmds.md
+++ b/docs/mmds.md
@@ -86,6 +86,37 @@ ip route add 169.254.169.200 dev eth0
 If MMDS configuration is not provided before booting up the guest, the MMDS
 IPv4 address defaults to `169.254.169.254`.
 
+# MMDS requests 
+
+Guests workloads that request metadata from MMDS  can choose to receive the response in IMDS or JSON format.
+IMDS documentation can be found [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html).
+
+Request example to retrieve IMDS format response:
+```bash
+curl -s "http://169.254.169.200/latest/meta-data/network/interfaces/macs/02:29:96:8f:6a:2d"
+```
+
+IMDS output:
+```text
+device-number
+local-hostname
+subnet-id
+```
+
+Request example to retrieve JSON format response:
+```bash
+curl -s -H "Accept: application/json" "http://169.254.169.200/latest/meta-data/network/interfaces/macs/02:29:96:8f:6a:2d"
+```
+
+JSON output:
+```text
+{
+    "device-number": "13345342",
+    "local-hostname": "localhost",
+    "subnet-id": "subnet-be9b61d"
+}
+```
+
 ### Example use case: credential rotation
 
 For this example, the guest expects to find some sort of credentials (say, a
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -215,14 +215,13 @@ impl ApiServer {
             .lock()
             .expect("Failed to acquire lock on MMDS info")
             .patch_data(value);
+
         match mmds_response {
             Ok(_) => Response::new(Version::Http11, StatusCode::NoContent),
             Err(e) => match e {
-                data_store::Error::NotFound => ApiServer::json_response(
-                    StatusCode::NotFound,
-                    ApiServer::json_fault_message(e.to_string()),
-                ),
-                data_store::Error::UnsupportedValueType => ApiServer::json_response(
+                data_store::Error::NotFound => unreachable!(),
+                data_store::Error::UnsupportedValueType => unreachable!(),
+                data_store::Error::NotInitialized => ApiServer::json_response(
                     StatusCode::BadRequest,
                     ApiServer::json_fault_message(e.to_string()),
                 ),
@@ -423,9 +422,6 @@ mod tests {
 
         let response = api_server.put_mmds(serde_json::Value::String("string".to_string()));
         assert_eq!(response.status(), StatusCode::NoContent);
-
-        let response = api_server.put_mmds(serde_json::Value::Bool(true));
-        assert_eq!(response.status(), StatusCode::BadRequest);
     }
 
     #[test]
@@ -440,7 +436,7 @@ mod tests {
         let to_vmm_fd = EventFd::new(libc::EFD_NONBLOCK).unwrap();
         let (api_request_sender, _from_api) = channel();
         let (_to_api, vmm_response_receiver) = channel();
-        let mmds_info = MMDS.clone();
+        let mmds_info = Arc::new(Mutex::new(Mmds::default()));
 
         let api_server = ApiServer::new(
             mmds_info,
@@ -451,14 +447,17 @@ mod tests {
         )
         .unwrap();
 
+        // MMDS data store is not yet initialized.
+        let response = api_server.patch_mmds(serde_json::Value::Bool(true));
+        assert_eq!(response.status(), StatusCode::BadRequest);
+
         let response = api_server.put_mmds(serde_json::Value::String("string".to_string()));
         assert_eq!(response.status(), StatusCode::NoContent);
 
-        let response = api_server.patch_mmds(serde_json::Value::String("string".to_string()));
+        let response = api_server.patch_mmds(serde_json::Value::String(
+            "{ \"key\" : \"value\" }".to_string(),
+        ));
         assert_eq!(response.status(), StatusCode::NoContent);
-
-        let response = api_server.patch_mmds(serde_json::Value::Bool(true));
-        assert_eq!(response.status(), StatusCode::BadRequest);
     }
 
     #[test]
--- a/src/micro_http/src/common/headers.rs
+++ b/src/micro_http/src/common/headers.rs
@@ -18,6 +18,8 @@ pub enum Header {
     TransferEncoding,
     /// Header `Server`.
     Server,
+    /// Header `Accept`
+    Accept,
 }
 
 impl Header {
@@ -29,6 +31,7 @@ impl Header {
             Self::Expect => b"Expect",
             Self::TransferEncoding => b"Transfer-Encoding",
             Self::Server => b"Server",
+            Self::Accept => b"Accept",
         }
     }
 
@@ -47,6 +50,7 @@ impl Header {
                 "expect" => Ok(Self::Expect),
                 "transfer-encoding" => Ok(Self::TransferEncoding),
                 "server" => Ok(Self::Server),
+                "accept" => Ok(Self::Accept),
                 _ => Err(RequestError::InvalidHeader),
             }
         } else {
@@ -80,6 +84,9 @@ pub struct Headers {
     /// server must support it. It is useful only when receiving the body of the request and should
     /// be known immediately after parsing the headers.
     chunked: bool,
+    /// `Accept` header might be used by HTTP clients to enforce server responses with content
+    /// formatted in a specific way.
+    accept: MediaType,
 }
 
 impl Default for Headers {
@@ -89,6 +96,9 @@ impl Default for Headers {
             content_length: Default::default(),
             expect: Default::default(),
             chunked: Default::default(),
+            // The default `Accept` media type is plain text. This is inclusive enough
+            // for structured and unstructured text.
+            accept: MediaType::PlainText,
         }
     }
 }
@@ -137,6 +147,13 @@ impl Headers {
                                 Err(_) => Err(RequestError::UnsupportedHeader),
                             }
                         }
+                        Header::Accept => match MediaType::try_from(entry[1].trim().as_bytes()) {
+                            Ok(accept_type) => {
+                                self.accept = accept_type;
+                                Ok(())
+                            }
+                            Err(_) => Err(RequestError::UnsupportedHeader),
+                        },
                         Header::TransferEncoding => match entry[1].trim() {
                             "chunked" => {
                                 self.chunked = true;
@@ -179,6 +196,11 @@ impl Headers {
         self.expect
     }
 
+    /// Returns the `Accept` header `MediaType`.
+    pub fn accept(&self) -> MediaType {
+        self.accept
+    }
+
     /// Parses a byte slice into a Headers structure for a HTTP request.
     ///
     /// The byte slice is expected to have the following format: </br>
@@ -218,6 +240,11 @@ impl Headers {
             return Ok(headers);
         }
         Err(RequestError::InvalidRequest)
+    }
+
+    /// Accept header setter.
+    pub fn set_accept(&mut self, media_type: MediaType) {
+        self.accept = media_type;
     }
 }
 
@@ -296,6 +323,7 @@ mod tests {
                 content_length,
                 expect,
                 chunked,
+                accept: MediaType::PlainText,
             }
         }
     }
@@ -343,14 +371,27 @@ mod tests {
     #[test]
     fn test_try_from_headers() {
         // Valid headers.
-        assert_eq!(
-            Headers::try_from(
-                b"Last-Modified: Tue, 15 Nov 1994 12:45:26 GMT\r\nContent-Length: 55\r\n\r\n"
-            )
-            .unwrap()
-            .content_length,
-            55
-        );
+        let headers =  Headers::try_from(
+            b"Last-Modified: Tue, 15 Nov 1994 12:45:26 GMT\r\nAccept: application/json\r\nContent-Length: 55\r\n\r\n"
+        )
+            .unwrap();
+        assert_eq!(headers.content_length, 55);
+        assert_eq!(headers.accept, MediaType::ApplicationJson);
+
+        // Valid headers.
+        let headers =  Headers::try_from(
+            b"Last-Modified: Tue, 15 Nov 1994 12:45:26 GMT\r\nAccept: text/plain\r\nContent-Length: 49\r\n\r\n"
+        )
+            .unwrap();
+        assert_eq!(headers.content_length, 49);
+        assert_eq!(headers.accept, MediaType::PlainText);
+
+        // Valid headers.
+        let headers = Headers::try_from(
+            b"Last-Modified: Tue, 15 Nov 1994 12:45:26 GMT\r\nContent-Length: 29\r\n\r\n",
+        )
+        .unwrap();
+        assert_eq!(headers.content_length, 29);
 
         let bytes: [u8; 10] = [130, 140, 150, 130, 140, 150, 130, 140, 150, 160];
         // Invalid headers.
@@ -416,6 +457,19 @@ mod tests {
         assert!(header
             .parse_header_line(b"Content-Type: application/json")
             .is_ok());
+
+        // Test valid accept media type.
+        assert!(header
+            .parse_header_line(b"Accept: application/json")
+            .is_ok());
+        assert!(header.accept == MediaType::ApplicationJson);
+        assert!(header.parse_header_line(b"Accept: text/plain").is_ok());
+        assert!(header.accept == MediaType::PlainText);
+
+        // Test invalid accept media type.
+        assert!(header
+            .parse_header_line(b"Accept: application/json-patch")
+            .is_err());
     }
 
     #[test]
@@ -442,5 +496,8 @@ mod tests {
 
         let header = Header::try_from(b"content-length").unwrap();
         assert_eq!(header.raw(), b"Content-Length");
+
+        let header = Header::try_from(b"Accept").unwrap();
+        assert_eq!(header.raw(), b"Accept");
     }
 }
--- a/src/mmds/src/data_store.rs
+++ b/src/mmds/src/data_store.rs
@@ -1,9 +1,8 @@
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::fmt;
-
 use serde_json::Value;
+use std::fmt;
 
 /// The Mmds is the Microvm Metadata Service represented as an untyped json.
 #[derive(Clone)]
@@ -12,9 +11,16 @@ pub struct Mmds {
     is_initialized: bool,
 }
 
+/// MMDS possible outputs.
+pub enum OutputFormat {
+    Json,
+    Imds,
+}
+
 #[derive(Debug, PartialEq)]
 pub enum Error {
     NotFound,
+    NotInitialized,
     UnsupportedValueType,
 }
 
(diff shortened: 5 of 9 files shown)


src/mmds/src/data_store.rs
```diff
@@ -89,15 +105,7 @@ impl Mmds {
         self.data_store.to_string()
     }
 
-    /// This function replicates the behavior of the Instance Metadata Service
```
andreeaflorescu, line 92: Can we move this comment to the below function instead of removing it?

src/mmds/src/data_store.rs
```diff
@@ -34,11 +40,21 @@ impl Default for Mmds {
...
             data_store: Value::default(),
             is_initialized: false,
+            response_format: ResponseFormat::Imds,
         }
     }
 }
 
 impl Mmds {
+    /// Response format setter.
+    pub fn set_json_format(&mut self, is_json: bool) {
```
andreeaflorescu, line 50: To be consistent with the naming, we should either name this `set_response_format` and send the `ResponseFormat` as a parameter, or name it `set_json_response_format` in which case we don't need to send the bool as a parameter.

src/mmds/src/data_store.rs
```diff
@@ -34,11 +40,21 @@ impl Default for Mmds {
...
+        self.response_format = if is_json {
+            ResponseFormat::Json
+        } else {
+            ResponseFormat::Imds
+        }
+    }
+
```
sandreim, line 57:
```suggestion
    /// Response format setter.
    pub fn set_response_format(&mut self, format: ResponseFormat) {
        self.response_format = format;
    }
```

src/mmds/src/data_store.rs
```diff
@@ -107,40 +115,40 @@ impl Mmds {
...
+                ResponseFormat::Imds => Mmds::get_imds_format_response(val),
+                ResponseFormat::Json => Ok(val
+                    .as_str()
+                    .map_or(val.to_string(), |val_str| val_str.to_string())),
+            },
+            None => Err(Error::NotFound),
+        }
+    }
+
+    fn get_imds_format_response(val: &Value) -> Result<String, Error> {
```
sandreim, line 128:
```suggestion
    fn format_imds_value(val: &Value) -> Result<String, Error> {
```

src/mmds/src/data_store.rs
```diff
@@ -107,40 +115,40 @@ impl Mmds {
...
-            Some(val) => {
+            Some(val) => match &self.response_format {
+                ResponseFormat::Imds => Mmds::get_imds_format_response(val),
+                ResponseFormat::Json => Ok(val
+                    .as_str()
+                    .map_or(val.to_string(), |val_str| val_str.to_string())),
+            },
+            None => Err(Error::NotFound),
+        }
+    }
```
sandreim, line 126: This would make it more readable:
```suggestion
            Some(val) => match &self.response_format {
                ResponseFormat::Imds => Mmds::format_imds_value(val),
                ResponseFormat::Json => Mmds::format_json_value(val),
            },
            None => Err(Error::NotFound),
        }
    }
    
    fn format_json_value(val: &Value) -> Result<String, Error> {
         Ok(val
                    .as_str()
                    .map_or(val.to_string(), |val_str| val_str.to_string())),
    }

```

docs/mmds.md
```diff
@@ -63,13 +63,15 @@ Users can set up a custom IPv4 address for MMDS. This can be achieved through a
 
 ```json
 {
-"ipv4_address": "169.254.169.200"
+"ipv4_address": "169.254.169.200",
+"json_response": true
```
sandreim, line 67: I would prefer this way of configuring MMDS:

```suggestion
"response_format": "JSON",
```
or

```suggestion
"response_format": "IMDS",
```

src/api_server/swagger/firecracker.yaml
```diff
@@ -596,6 +596,8 @@ definitions:
     properties:
       ipv4_address:
         type: string
+      json_response:
+        type: bool
```
sandreim, line 600:
```suggestion
      response_format:
        type: string
```

tests/integration_tests/functional/test_mmds.py
```diff
@@ -122,6 +122,70 @@ def test_custom_ipv4_support_config(test_microvm_with_ssh, network_config):
...
+
+    cmd = 'ip route add 169.254.169.250 dev eth0'
+    _, stdout, stderr = ssh_connection.execute_command(cmd)
+    _assert_out(stdout, stderr, '')
+
+    pre = 'curl -s http://169.254.169.250/'
+
+    cmd = pre + 'latest'
+    _, stdout, stderr = ssh_connection.execute_command(cmd)
+    _assert_out(stdout, stderr, '{"meta-data":{"ami-id":"ami-12345678",\
```
andreeaflorescu, line 179: Should we make sure that the output is a valid JSON as well?

tests/integration_tests/functional/test_mmds.py
```diff
@@ -122,6 +122,70 @@ def test_custom_ipv4_support_config(test_microvm_with_ssh, network_config):
...
+    assert test_microvm.api_session.is_status_ok(response.status_code)
+    assert response.json() == {}
+
+    data_store = {
+        'latest': {
+            'meta-data': {
+                'ami-id': 'ami-12345678',
+                'reservation-id': 'r-fea54097',
+                'local-hostname': 'ip-10-251-50-12.ec2.internal',
+                'public-hostname': 'ec2-203-0-113-25.compute-1.amazonaws.com',
```
andreeaflorescu, line 140: Is this a valid json? Are `,` accepted even though there is no other field afterwards?

src/mmds/src/data_store.rs
iulianbarbu: This comment is not relevant anymore. The method returns a plain JSON now.

src/mmds/src/lib.rs
```diff
@@ -76,14 +76,11 @@ pub fn parse_request(request_bytes: &[u8]) -> Response {
...
             match response {
-                Ok(response) => {
-                    let response_body = response.join("\n");
-                    build_response(
-                        request.http_version(),
-                        StatusCode::OK,
-                        Body::new(response_body),
-                    )
-                }
+                Ok(response_body) => build_response(
```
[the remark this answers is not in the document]
iulianbarbu: This is true. Added an `ApplicationJson` content type.

tests/integration_tests/functional/test_mmds.py
iulianbarbu: This is a python dict, so it doesn't really matter. Removed it since it is not needed.

src/mmds/src/lib.rs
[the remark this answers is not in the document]
iulianbarbu: Actually, the response is created by default with `MediaType::ApplicationJson`. We've always formed the MMDS response with a JSON content-type, although, we weren't actually sending back a valid JSON.

tests/integration_tests/functional/test_mmds.py
iulianbarbu: I think we are already covered through these checks, which confront the response content with an expected valid JSON. Do you want me to add further JSON validation checks on the expected JSON, to make sure we do not invalidate the expected JSON when modifying these tests?

src/api_server/src/parsed_request.rs
```diff
@@ -670,7 +670,7 @@ mod tests {
             .write_all(
                 b"PUT /mmds/config HTTP/1.1\r\n\
                 Content-Type: application/json\r\n\
-                Content-Length: 26\r\n\r\n{\"ipv4_address\":\"1.1.1.1\"}",
+                Content-Length: 32\r\n\r\n{\"ipv4_address\":\"169.254.170.2\"}",
```
andreeaflorescu, line 673: This change seems unrelated to this commit.

src/mmds/src/data_store.rs
```diff
@@ -190,7 +150,11 @@ mod tests {
                     "UK": "+44 1234567"
                 },
                 "mobile": "+44 2345678"
-            }
+            },
+            "phones2": [
```
andreeaflorescu, line 154: This is a bit funky. Why not replace phones or add something else that makes sense?

src/mmds/src/data_store.rs
```diff
@@ -107,39 +99,7 @@ impl Mmds {
...
-                            Some(str_val) => {
-                                ret.push(str_val.to_string());
-                                Ok(ret)
-                            }
-                            None => Err(Error::UnsupportedValueType),
-                        }
-                    }
-                }
-            }
+            Some(json) => Ok(json.to_string()),
```
andreeaflorescu, line 102: This can result in invalid JSON values. For example, in the tests where you do `mmds.get_value("/phones2".to_string()).unwrap()`, the return is something that is not a json.

This is breaking the contract of the function which specifies that it returns a json. There are a few options to fix this, but I am not sure which one is best.

1. Check that the path actually corresponds to a json (i.e. you cannot do get on a key that has as value a string/float/vec). This seems very restrictive.
2. When the value of a key is not JSON, create a valid JSON from it by doing something like:
``'{' + key + ':' + value '}'` (this is pseudo code :)) )

These are the only two I can think of, but there are probably others as well.

src/mmds/src/lib.rs
```diff
@@ -221,7 +218,7 @@ mod tests {
         // Test Ok path.
         let request = b"GET http://169.254.169.254/ HTTP/1.0\r\n\r\n";
         let mut expected_response = Response::new(Version::Http10, StatusCode::OK);
-        let body = "age\nname/\nphones/".to_string();
+        let body = r#"{"age":"43","name":{"first":"John","second":"Doe"},"phones":{"home":{"RO":"+40 1234567","UK":"+44 1234567"},"mobile":"+44 2345678"}}"#.to_string();
```
andreeaflorescu, line 221: Please use appropriate formatting here. Using `r#` is powerful because it allows you to use spaces and tabs to format the jsons.

src/mmds/src/lib.rs
```diff
@@ -231,7 +228,7 @@ mod tests {
 
         let request = b"GET /age HTTP/1.1\r\n\r\n";
         let mut expected_response = Response::new(Version::Http11, StatusCode::OK);
-        let body = "43".to_string();
+        let body = r#""43""#.to_string();
```
andreeaflorescu, line 231: Why is this needed?

tests/integration_tests/functional/test_mmds.py
```diff
@@ -88,39 +80,122 @@ def test_custom_ipv4_support_config(test_microvm_with_ssh, network_config):
...
     cmd = pre + 'latest/meta-data'
     _, stdout, stderr = ssh_connection.execute_command(cmd)
-    _assert_out_multiple(
+    _assert_out(
         stdout,
         stderr,
-        ['ami-id', 'reservation-id', 'local-hostname', 'public-hostname',
-            'network/']
-    )
+        '{"ami-id":"ami-12345678","local-hostname":\
```
andreeaflorescu, line 102: Can you format this JSON in a way that it is easier to read?

src/mmds/src/data_store.rs
raduiliescu: I would go with second option. Also I would add a unit test for this corner case that would validate with a tool that the json is in proper format.

andreeaflorescu: Disregard my comment. I read through the RFC again and apparently strings are valid jsons :))

https://tools.ietf.org/html/rfc7159#section-3

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@
 - Removed redundant RescanBlockDevice action from the /actions API.
   The functionality is available through the PATCH /drives API.
   See `docs/api_requests/patch-block.md`.
+- Within guest, MMDS response content is changed to JSON.
```
sandreim, line 71: The change needs to be properly documented plus provide one example on how MMDS responses look right now.

src/api_server/src/lib.rs
```diff
@@ -425,7 +418,7 @@ mod tests {
         assert_eq!(response.status(), StatusCode::NoContent);
 
         let response = api_server.put_mmds(serde_json::Value::Bool(true));
-        assert_eq!(response.status(), StatusCode::BadRequest);
+        assert_eq!(response.status(), StatusCode::NoContent);
```
andreeaflorescu, line 421: Any reason why we want to keep this check? It was previously there to check the error case. If we want to check that all values are valid values, then we should add a test for all. Otherwise, I would remove this one because it is confusing when reading the test.

The same applies to the following assert.

src/api_server/src/lib.rs
```diff
@@ -217,16 +216,10 @@ impl ApiServer {
...
-                data_store::Error::NotFound => ApiServer::json_response(
-                    StatusCode::NotFound,
-                    ApiServer::json_fault_message(e.to_string()),
-                ),
-                data_store::Error::UnsupportedValueType => ApiServer::json_response(
-                    StatusCode::BadRequest,
-                    ApiServer::json_fault_message(e.to_string()),
-                ),
-            },
+            Err(e) => ApiServer::json_response(
```
andreeaflorescu, line 219: This error case doesn't seem to be tested. Can you also add an assert/test for it?

src/mmds/src/data_store.rs
```diff
@@ -89,15 +64,8 @@ impl Mmds {
...
-    /// This function replicates the behavior of the Instance Metadata Service
-    /// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
-    /// 1. For a (key, value) pair where the value is a dictionary, it will return all the keys
-    /// in the dictionary.
-    /// 2. For a (key, value) pair where the value is a simple type (bool, string, number),
-    /// it will return the value.
-    ///
-    /// When the path is not found, a NotFound error is returned.
-    pub fn get_value(&self, path: String) -> Result<Vec<String>, Error> {
+    // Returns a JSON MMDS data store information mapped to a path.
```
andreeaflorescu, line 67: We can be a bit more descriptive here taking the example of the initial documentation for this function.

"Returns the subtree located at `path`. When the path corresponds to a leaf, it returns the value.

Returns `Error::NotFound` when the path is invalid."

src/vmm/src/resources.rs
```diff
@@ -633,51 +633,12 @@ mod tests {
...
-                    }},
-                    "mmds-config": {{
-                        "ipv4_address": "169.254.170.2"
                     }}
             }}"#,
             kernel_file.as_path().to_str().unwrap(),
             rootfs_file.as_path().to_str().unwrap(),
         );
         assert!(VmResources::from_json(json.as_str(), "some_version").is_ok());
-
```
andreeaflorescu, line 645: Why is this test removed?

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@
   updates to existing configurations.
 - `PATCH /network-interfaces/{id}` only allowed post-boot. Use `PUT` for
   pre-boot updates to existing configurations.
+- Within guest, MMDS response content has JSON format.
```
andreeaflorescu, line 35: The changelog entry should be added in the commit that adds the code. The are a few reasons for this. If we want to ever revert this change, we have to go through multiple commits. It keeps the history tidy and easier to read as unit changes are contained in a single commit.

src/mmds/src/data_store.rs
iulianbarbu: Done.

src/api_server/src/lib.rs
iulianbarbu: I've commented on a similar case to another review comment that you left.
TLDR;
To be able to test this scenario we would need to expose as `pub` a private field of a struct from another crate, or provide a getter, to be able to provide a side effect which will change the behavior of this method. Used this workaround at the moment.

Please let me know what you think.

src/micro_http/src/common/headers.rs
```diff
@@ -179,6 +196,12 @@ impl Headers {
         self.expect
     }
 
+    /// Returns the `Accept` header `MediaType`.
+    #[allow(unused)]
```
andreeaflorescu, line 200: This allow should not be needed as the function is pub.

src/micro_http/src/common/headers.rs
```diff
@@ -291,11 +314,12 @@ mod tests {
     use super::*;
 
     impl Headers {
-        pub fn new(content_length: i32, expect: bool, chunked: bool) -> Self {
+        pub fn new(content_length: i32, expect: bool, chunked: bool, accepted: MediaType) -> Self {
```
andreeaflorescu, line 317: /s/accepted/accept

Most of the tests don't use `accept`. How about we leave this constructor as it was before and when we want to test the accept header as well, we just set it?

src/micro_http/src/common/headers.rs
```diff
@@ -343,14 +367,12 @@ mod tests {
...
         // Valid headers.
-        assert_eq!(
-            Headers::try_from(
-                b"Last-Modified: Tue, 15 Nov 1994 12:45:26 GMT\r\nContent-Length: 55\r\n\r\n"
-            )
-            .unwrap()
-            .content_length,
-            55
-        );
+        let headers =  Headers::try_from(
```
andreeaflorescu, line 370: Can we also have a test for text/plain and for no accept header to check the default?

src/micro_http/src/common/headers.rs
```diff
@@ -416,6 +438,17 @@ mod tests {
         assert!(header
             .parse_header_line(b"Content-Type: application/json")
             .is_ok());
+
+        // Test valid accept media type.
+        assert!(header
```
andreeaflorescu, line 443: We should check the value of the header instead of just `is_ok`.

src/mmds/src/data_store.rs
```diff
@@ -89,15 +72,77 @@ impl Mmds {
...
+                    Some(str_val) => Ok(str_val.to_string()),
+                    None => Err(Error::UnsupportedValueType),
+                }
+            }
+        }
+    }
+
+    /// Returns the subtree located at path. When the path corresponds to a leaf, it returns the value.
+    /// Returns Error::NotFound when the path is invalid.
+    pub fn get_value(&self, path: String, accept: MediaType) -> Result<String, Error> {
```
andreeaflorescu, line 145: At this scope, we shouldn't "leak" information about the upper levels. As far as this layer is concerned, it should be able to return values from the data store in a specified format. In this context, `accept` doesn't make sense. We can rename it to `format`.

When looking at separation of concerns, it should also not use `MediaType` because that is the interface used by HTTP. The MMDS data store is not necessarily tied to HTTP and should define its own interface. The formats that the Data Store supports are JSON and IMDS. At a higher level we should make the translation between ApplicationJson -> Json and PlainText -> Imds.

In the data-store module:
```rust
enum OutputFormat {
    Json
    Imds
}

....
pub fn get_value(&self, path: String, format: OutputFormat) -> Result <String, Error> .....
```

In the `mmds/src/lib.rs` (this should eventually be moved some place else where it makes more sense):
```rust
impl Into<OutputFormat> for MediaType {
   fn into (&self) ... {
       match self {
                MediaType::ApplicationJson => OutputFormat::Json,
                MediaType::PlainText => OutputFormat::Imds,
       }
}
```

src/mmds/src/lib.rs
```diff
@@ -74,39 +74,30 @@ pub fn parse_request(request_bytes: &[u8]) -> Response {
...
+                        build_response(
+                            request.http_version(),
+                            StatusCode::NotFound,
+                            Body::new(error_msg),
+                        )
                     }
-                }
+                    MmdsError::UnsupportedValueType => build_response(
+                        request.http_version(),
+                        StatusCode::NotImplemented,
```
andreeaflorescu, line 96: Why change this from `InternalServerError` to `NotImplemented`?

src/mmds/src/data_store.rs
```diff
@@ -15,16 +15,19 @@ pub struct Mmds {
...
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
             Error::NotFound => write!(f, "The MMDS resource does not exist."),
-            Error::UnsupportedValueType => {
-                write!(f, "Cannot add non-strings values to the MMDS data-store.")
-            }
+            Error::NotInitialized => write!(f, "The MMDS data store is not initialized."),
+            Error::UnsupportedValueType => write!(
+                f,
+                "Retrieval of the requested resource value type is not supported."
```
andreeaflorescu, line 29: nit: "Cannot retrieve value. The value has an unsupported type."

src/micro_http/src/common/headers.rs
iulianbarbu: Done.

src/micro_http/src/common/headers.rs
iulianbarbu: Done.

src/micro_http/src/common/headers.rs
iulianbarbu: I thought that this unit test should check success and error cases for `Headers::try_from`. Parsing the accept headers is unit tested for `parse_header_line`, used by this method. Added test cases for text/plain and for no accept header to be extra sure.

src/micro_http/src/common/headers.rs
iulianbarbu: Done.

src/mmds/src/data_store.rs
iulianbarbu: Done.

src/mmds/src/data_store.rs
iulianbarbu: Done.

src/mmds/src/lib.rs
iulianbarbu: I thought that it does not make sense. HTTP RFC says: 500 Internal Server Error server error response code indicates that the server encountered an unexpected condition that prevented it from fulfilling the request. `UnsupportedValueType` error is an expected error, which happens every time a guest application tries to retrieve a data store value which can not be flattened (e.g. JSON arrays). This shouldn't happen. The exposed metadata has to be valid for retrieval from guest side, but we can not guarantee what metadata Firecracker customers will expose to guests. Please let me know what you think.

andreeaflorescu: Sounds good. It looks like this is a user-facing change. Should we also add it to the changelog?

src/micro_http/src/connection.rs
```diff
@@ -406,6 +406,20 @@ mod tests {
     use super::*;
     use common::{Method, Version};
     use std::os::unix::net::UnixStream;
+    use MediaType;
+
+    impl Headers {
```
andreeaflorescu, line 411: I don't think this change is needed. We are trying to test the connection here, so the accept header is not necessarily needed. We should be keeping this file as it was before.

iulianbarbu:
@andreeaflorescu , I've added some missing details to CHANGELOG.md. Please take a look again,

CHANGELOG.md
```diff
@@ -15,7 +15,10 @@
 - Added a [guide](docs/devctr-image.md) for updating the dev container image.
 - Added a new API call, `PUT /mmds/config`, for configuring the
   `MMDS` with a custom valid link-local IPv4 address.
-
+- Added JSON response format support for MMDS guest applications requests.
+- Added MMDS `403 BadRequest` failure response when trying to issue a `PATCH`
```
andreeaflorescu, line 19: This should be in another section: either `# Changed` or `# Fixed` because you are changing a behavior, not adding a new one.

nit: I would rephrase it a bit to something like: `PATCH` request on `/mmds` before the data store is initialized return `403 BadRequest`.

CHANGELOG.md
```diff
@@ -32,6 +35,13 @@
   updates to existing configurations.
 - `PATCH /network-interfaces/{id}` only allowed post-boot. Use `PUT` for
   pre-boot updates to existing configurations.
+- MMDS failure response for retrieval of unsupported JSON value types
+  (other than strings and objects) is changed from 500 InternalServerError
+  to 501 NotImplemented, if the request was made for an IMDS response format.
+- MMDS failure response to `PUT` and `PATCH` requests for MMDS data store,
```
andreeaflorescu, line 41: This can also be shortened to: "Allowed the MMDS data store to be initialized with all supported JSON types. Retrieval of these values (besides String, Array, Dictionary) is only possible in JSON mode."

CHANGELOG.md
```diff
@@ -15,7 +15,10 @@
 - Added a [guide](docs/devctr-image.md) for updating the dev container image.
 - Added a new API call, `PUT /mmds/config`, for configuring the
   `MMDS` with a custom valid link-local IPv4 address.
-
+- Added JSON response format support for MMDS guest applications requests.
```
andreeaflorescu, line 18: We should specify that this is a feature we are currently experimenting with and is subject to change.

CHANGELOG.md
```diff
@@ -32,6 +35,13 @@
   updates to existing configurations.
 - `PATCH /network-interfaces/{id}` only allowed post-boot. Use `PUT` for
   pre-boot updates to existing configurations.
+- MMDS failure response for retrieval of unsupported JSON value types
```
andreeaflorescu, line 38: Changed returned status code from 500 to `501 Not Implemented` for queries on the MMDS endpoint in IMDS format where the value is unsupported.

CHANGELOG.md
sandreim: I guess this would do it:
```suggestion
- Added experimental JSON response format support for MMDS guest applications requests.
```

docs/mmds.md
```diff
@@ -86,6 +86,37 @@ ip route add 169.254.169.200 dev eth0
 If MMDS configuration is not provided before booting up the guest, the MMDS
 IPv4 address defaults to `169.254.169.254`.
 
+# MMDS requests 
+
+From within a guest, an user can retrieve MMDS metadata in IMDS and JSON formats.
```
sandreim, line 91:
```suggestion
Guests workloads that request metadata from MMDS  can choose to receive the response in IMDS or JSON format.
```

PR #1773, opened 2020-04-10
Implement (de)serialization for net devices

### LINKED ISSUE #1425: Implement (de)serialization for Net Devices

ioanachirca:
## Reason for This PR

Fixes #1425 

## Description of Changes

- The Net device constructor receives the Tap interface name and opens it. This is similar to how the Block device handles the disk file.
- Implement the Persist trait for the Net device in order to be able to save/restore it.
- Derive Versionize for the NetState.

## License Acceptance

## PR Checklist

- [x] No newly added `unsafe` code.
- [x] No API changes.
- [x] No user-facing changes.

### DIFF
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -97,12 +97,14 @@ impl Net {
     /// Create a new virtio network device with the given TAP interface.
     pub fn new_with_tap(
         id: String,
-        tap: Tap,
+        tap_dev_name: String,
         guest_mac: Option<&MacAddr>,
         rx_rate_limiter: RateLimiter,
         tx_rate_limiter: RateLimiter,
         allow_mmds_requests: bool,
     ) -> Result<Self> {
+        let tap = Tap::open_named(&tap_dev_name).map_err(Error::TapOpen)?;
+
         // Set offload flags to match the virtio features below.
         tap.set_offload(
             net_gen::TUN_F_CSUM | net_gen::TUN_F_UFO | net_gen::TUN_F_TSO4 | net_gen::TUN_F_TSO6,
@@ -757,7 +759,6 @@ pub(crate) mod tests {
     use polly::event_manager::{EventManager, Subscriber};
     use rate_limiter::{RateLimiter, TokenBucket, TokenType};
     use utils::epoll::{EpollEvent, EventSet};
-    use utils::net::Tap;
     use virtio_gen::virtio_net::{
         virtio_net_hdr_v1, VIRTIO_F_VERSION_1, VIRTIO_NET_F_CSUM, VIRTIO_NET_F_GUEST_CSUM,
         VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_UFO, VIRTIO_NET_F_HOST_TSO4,
@@ -790,20 +791,20 @@ pub(crate) mod tests {
     impl Net {
         pub fn default_net(test_mutators: TestMutators) -> Net {
             let next_tap = NEXT_INDEX.fetch_add(1, Ordering::SeqCst);
-            let tap = Tap::open_named(&format!("net-device{}", next_tap)).unwrap();
-            tap.enable().unwrap();
+            let tap_dev_name = format!("net-device{}", next_tap);
 
             let guest_mac = Net::default_guest_mac();
 
             let mut net = Net::new_with_tap(
                 format!("net-device{}", next_tap),
-                tap,
+                tap_dev_name.clone(),
                 Some(&guest_mac),
                 RateLimiter::default(),
                 RateLimiter::default(),
                 true,
             )
             .unwrap();
+            net.tap.enable().unwrap();
             net.test_mutators = test_mutators;
 
             net
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -31,7 +31,7 @@ use virtio_gen::virtio_net::{
     VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_UFO, VIRTIO_NET_F_HOST_TSO4, VIRTIO_NET_F_HOST_UFO,
     VIRTIO_NET_F_MAC,
 };
-use vm_memory::{Bytes, GuestAddress, GuestMemoryError, GuestMemoryMmap};
+use vm_memory::{ByteValued, Bytes, GuestAddress, GuestMemoryError, GuestMemoryMmap};
 
 fn vnet_hdr_len() -> usize {
     mem::size_of::<virtio_net_hdr_v1>()
@@ -56,12 +56,29 @@ fn init_vnet_hdr(buf: &mut [u8]) {
     }
 }
 
+#[derive(Clone, Copy)]
+pub struct ConfigSpace {
+    pub guest_mac: [u8; MAC_ADDR_LEN],
+}
+
+impl Default for ConfigSpace {
+    fn default() -> ConfigSpace {
+        ConfigSpace {
+            guest_mac: [0; MAC_ADDR_LEN],
+        }
+    }
+}
+
+unsafe impl ByteValued for ConfigSpace {}
+
 pub struct Net {
-    id: String,
+    pub(crate) id: String,
 
     pub(crate) tap: Tap,
-    avail_features: u64,
-    acked_features: u64,
+    pub(crate) tap_if_name: String,
+
+    pub(crate) avail_features: u64,
+    pub(crate) acked_features: u64,
 
     pub(crate) queues: Vec<Queue>,
     pub(crate) queue_evts: Vec<EventFd>,
@@ -78,16 +95,16 @@ pub struct Net {
     tx_iovec: Vec<(GuestAddress, usize)>,
     tx_frame_buf: [u8; MAX_BUFFER_SIZE],
 
-    interrupt_status: Arc<AtomicUsize>,
+    pub(crate) interrupt_status: Arc<AtomicUsize>,
     interrupt_evt: EventFd,
 
-    config_space: Vec<u8>,
+    pub(crate) config_space: ConfigSpace,
     guest_mac: Option<MacAddr>,
 
     pub(crate) device_state: DeviceState,
     pub(crate) activate_evt: EventFd,
 
-    mmds_ns: Option<MmdsNetworkStack>,
+    pub(crate) mmds_ns: Option<MmdsNetworkStack>,
 
     #[cfg(test)]
     test_mutators: tests::TestMutators,
@@ -97,13 +114,13 @@ impl Net {
     /// Create a new virtio network device with the given TAP interface.
     pub fn new_with_tap(
         id: String,
-        tap_dev_name: String,
+        tap_if_name: String,
         guest_mac: Option<&MacAddr>,
         rx_rate_limiter: RateLimiter,
         tx_rate_limiter: RateLimiter,
         allow_mmds_requests: bool,
     ) -> Result<Self> {
-        let tap = Tap::open_named(&tap_dev_name).map_err(Error::TapOpen)?;
+        let tap = Tap::open_named(&tap_if_name).map_err(Error::TapOpen)?;
 
         // Set offload flags to match the virtio features below.
         tap.set_offload(
@@ -123,15 +140,12 @@ impl Net {
             | 1 << VIRTIO_NET_F_HOST_UFO
             | 1 << VIRTIO_F_VERSION_1;
 
-        let mut config_space;
+        let mut config_space = ConfigSpace::default();
         if let Some(mac) = guest_mac {
-            config_space = vec![0; MAC_ADDR_LEN];
-            config_space[..].copy_from_slice(mac.get_bytes());
+            config_space.guest_mac.copy_from_slice(mac.get_bytes());
             // When this feature isn't available, the driver generates a random MAC address.
             // Otherwise, it should attempt to read the device MAC address from the config space.
             avail_features |= 1 << VIRTIO_NET_F_MAC;
-        } else {
-            config_space = Vec::new();
         }
 
         let guest_mac = guest_mac.copied();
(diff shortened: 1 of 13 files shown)


src/devices/src/virtio/net/device.rs
```diff
@@ -150,6 +158,7 @@ impl Net {
         Ok(Net {
             id,
             tap,
+            tap_if_name: tap_if_name.clone(),
```
sandreim, line 161: No need to clone here.

src/versionize/src/primitives.rs
```diff
@@ -61,6 +63,37 @@ impl_versionize!(f32);
 impl_versionize!(f64);
 impl_versionize!(char);
 impl_versionize!(String);
+impl_versionize!(AtomicUsize);
```
sandreim, line 66: As per discussion, let's not impl Versionize for Atomics.

src/devices/src/virtio/net/device.rs
```diff
@@ -722,6 +735,76 @@ impl VirtioDevice for Net {
...
 }
 
+#[derive(Versionize)]
+pub struct NetState {
+    id: String,
+    tap_if_name: String,
+    rx_rate_limiter_state: RateLimiterState,
+    tx_rate_limiter_state: RateLimiterState,
+    allow_mmds_requests: bool,
+    guest_mac: Option<Vec<u8>>,
```
sandreim, line 745: If we want to be future looking about how we structure and name states, I would move guest_mac to a new struct NetConfigSpace and include it as config_space in the NetState. As we add features to any device in the future we will need to extend the config space with other fields that can be read/written by the guest. 

@andreeaflorescu might have some input her also.

src/rate_limiter/src/lib.rs
```diff
@@ -471,6 +519,52 @@ impl Default for RateLimiter {
...
+        RateLimiterState {
+            ops: self.ops.as_ref().map(|ops| ops.save()),
+            bandwidth: self.bandwidth.as_ref().map(|bw| bw.save()),
+            timer_active: self.timer_active,
+        }
+    }
+
+    fn restore(_: Self::ConstructorArgs, state: &Self::State) -> Result<Self, Self::Error> {
+        let mut rate_limiter = RateLimiter {
+            // Safe to unwrap because TokenBucket::restore doesn't return errors.
```
sandreim, line 545: More specifically, errors should be returned when serializing the state using semantic serialization. From this perspective it kind of makes sense that restore would return Self instead of Result.

src/devices/src/virtio/net/device.rs
ioanachirca: Done.

src/versionize/src/primitives.rs
ioanachirca: Done.

src/devices/src/virtio/net/device.rs
ioanachirca: Would it make sense to also keep `allow_mmds_requests` in the NetConfigSpace?

sandreim: That structure is actually treated as a buffer (`impl ByteValued for NetConfigSpace`) and read/written by the guest virtio net driver.

Only guest mac needs to be there for now. Other virtio features (for example multiqueue) will require us to add more fields in this structure.

src/devices/src/virtio/net/persist.rs
```diff
@@ -0,0 +1,157 @@
...
+    rx_rate_limiter_state: RateLimiterState,
+    tx_rate_limiter_state: RateLimiterState,
+    allow_mmds_requests: bool,
+    config_space: NetConfigSpace,
+    virtio_state: VirtioDeviceState,
+}
+
+pub struct NetConstructorArgs {
+    mem: GuestMemoryMmap,
+}
```
acatangiu, line 38: nit: all devices will also need to take a reference to `EventManager` to register their handlers on `restore()`.

But we can add that incrementally when we get there.

src/devices/src/virtio/net/persist.rs
```diff
@@ -0,0 +1,157 @@
...
+
+use super::*;
+
+use crate::virtio::persist::VirtioDeviceState;
+use crate::virtio::{DeviceState, Queue};
+
+#[derive(Versionize)]
+pub struct NetConfigSpace {
+    guest_mac: Vec<u8>,
+}
```
acatangiu, line 23: @ioanachirca @sandreim what do you think about moving `config_space` to generic `VirtioDeviceState`.
The `config_space: Vec<u8>` is part of the generic virtio device spec.

Independent devices implementations would then interpret that generic `Vec<u8>` as they need it.

ioanachirca: While this is more generic, vsock doesn't have a `config_space` and the block's `config_space` contains the disk size (and if the disk changes size between save and restore, overwriting the config with the old size would maybe lead to errors). I think these are 2 cases out of 3 in which we wouldn't use the config on restore.

acatangiu: That's right, but each device can do any impl-specific processing to the restored generic `config_space` pre-save/post-restore.

I'm just worried that we're ignoring saving parts of the spec based on assumptions stemming from our current device implementations.

ioanachirca: Makes sense. I added a commit that introduces `config_space` in `VirtioDeviceState` and only for the net device overwrites the config space on restore.

acatangiu: I just spoke to @sandreim and we have to keep it as a defined struct to be able to versionize it. So let's keep it like you proposed it in this PR.
Marking as resolved.

src/devices/src/virtio/net/persist.rs
```diff
@@ -0,0 +1,157 @@
...
+use vm_memory::GuestMemoryMmap;
+
+use super::*;
+
+use crate::virtio::persist::VirtioDeviceState;
+use crate::virtio::{DeviceState, Queue};
+
+#[derive(Versionize)]
+pub struct NetConfigSpace {
+    guest_mac: Vec<u8>,
```
sandreim, line 22: This should be an array of size 6.

src/devices/src/virtio/net/mod.rs
```diff
@@ -15,9 +15,11 @@ pub const TX_INDEX: usize = 1;
 
 pub mod device;
 pub mod event_handler;
+pub mod persist;
 
 pub use self::device::Net;
 pub use self::event_handler::*;
+pub use self::persist::*;
```
acatangiu, line 22: nit: I wouldn't `pub use persist::*` and keep it consistent with the other devices. Users could explicitly `use devices::net::persist::{X, Y}` since persist is a public module.

src/devices/src/virtio/net/persist.rs
```diff
@@ -0,0 +1,157 @@
...
+}
+
+#[derive(Versionize)]
+pub struct NetState {
+    id: String,
+    tap_if_name: String,
+    rx_rate_limiter_state: RateLimiterState,
+    tx_rate_limiter_state: RateLimiterState,
+    allow_mmds_requests: bool,
```
acatangiu, line 31: We also need to save and restore `mmds_ns: Option<MmdsNetworkStack>` since `MmdsNetworkStack` has state of its own.

`allow_mmds_requests` could be dropped from `NetState` and be inferred from `mmds_ns: Option<MmdsNetworkStack>`.

src/devices/src/virtio/net/persist.rs
ioanachirca: Done.

src/devices/src/virtio/net/mod.rs
ioanachirca: Done.

src/devices/src/virtio/net/persist.rs
ioanachirca: Done, now we save the state of the MmdsNetworkStack.

src/devices/src/virtio/net/persist.rs
```diff
@@ -0,0 +1,163 @@
...
+            virtio_state: VirtioDeviceState::from_device(self),
+        }
+    }
+
+    fn restore(
+        constructor_args: Self::ConstructorArgs,
+        state: &Self::State,
+    ) -> std::result::Result<Self, Self::Error> {
+        let rx_rate_limiter = RateLimiter::restore((), &state.rx_rate_limiter_state).unwrap();
+        let tx_rate_limiter = RateLimiter::restore((), &state.tx_rate_limiter_state).unwrap();
```
sandreim, line 66: How safe is the unwrap here?

src/dumbo/src/persist.rs
```diff
@@ -0,0 +1,102 @@
...
+use super::ns::MmdsNetworkStack;
+use super::*;
+
+const IPV4_ADDR_LEN: usize = 4;
+
+/// State of a MmdsNetworkStack.
+#[derive(Versionize)]
+pub struct MmdsNetworkStackState {
+    mac_addr: [u8; MAC_ADDR_LEN],
+    ipv4_addr: [u8; IPV4_ADDR_LEN],
```
sandreim, line 21: I would change this field to u32 as Ipv4Addr implements From/Into<u32>.

src/dumbo/src/persist.rs
```diff
@@ -0,0 +1,102 @@
...
+            max_pending_resets: self.tcp_handler.max_pending_resets,
+        }
+    }
+
+    fn restore(
+        _: Self::ConstructorArgs,
+        state: &Self::State,
+    ) -> std::result::Result<Self, Self::Error> {
+        let mac_addr = MacAddr::from_bytes_unchecked(&state.mac_addr);
+        let ipv4_addr = Ipv4Addr::new(
```
sandreim, line 50: Please see my above comment that would simplify the code here also.

src/devices/src/virtio/net/device.rs
```diff
@@ -78,16 +80,16 @@ pub struct Net {
     tx_iovec: Vec<(GuestAddress, usize)>,
     tx_frame_buf: [u8; MAX_BUFFER_SIZE],
 
-    interrupt_status: Arc<AtomicUsize>,
+    pub(crate) interrupt_status: Arc<AtomicUsize>,
     interrupt_evt: EventFd,
 
-    config_space: Vec<u8>,
+    pub(crate) config_space: Vec<u8>,
```
sandreim, line 86: nit: we can replace the Vec with structure for which we implement ByteValued.

src/dumbo/src/persist.rs
ioanachirca: Done.

src/devices/src/virtio/net/persist.rs
ioanachirca: Not very, that's true. I added a new `Error` enum.

src/devices/src/virtio/net/device.rs
ioanachirca: I added another commit in which I did this. Is this what you had in mind?

src/devices/src/virtio/net/device.rs
```diff
@@ -851,10 +867,8 @@ pub(crate) mod tests {
 
         fn set_mac(&mut self, mac: MacAddr) {
             self.guest_mac = Some(mac);
-            let mut config_space;
-            config_space = vec![0; MAC_ADDR_LEN];
-            config_space[..].copy_from_slice(mac.get_bytes());
-            self.config_space = config_space;
+            self.config_space = ConfigSpace::default();
```
acatangiu, line 870: this line should be removed:
- right now it's just superfluous since `guest_mac` is overwritten by the next instruction anyway.
- in a future where `ConfigSpace` contains more than just a `guest_mac`, this line would actually be a **bug** since it would change other fields of config space on a mac setter function.

src/devices/src/virtio/net/device.rs
```diff
@@ -56,6 +56,21 @@ fn init_vnet_hdr(buf: &mut [u8]) {
     }
 }
 
+#[derive(Clone, Copy)]
+pub struct ConfigSpace {
```
acatangiu, line 60: Does it make sense to directly `Versionize` this instead of having a separate identical `NetConfigSpace` struct in net/persist.rs?

If we do keep them separate, it would be nice to rename `NetConfigSpace` to `NetConfigSpaceState`.

src/devices/src/virtio/net/persist.rs
```diff
@@ -0,0 +1,174 @@
...
+            rx_rate_limiter,
+            tx_rate_limiter,
+            state.mmds_ns.is_some(),
+        )
+        .map_err(Error::CreateNet)?;
+
+        net.mmds_ns = state
+            .mmds_ns
+            .as_ref()
+            .map(|mmds_state| MmdsNetworkStack::restore((), &mmds_state).unwrap());
```
acatangiu, line 89: `MmdsNetworkStack::restore()` cannot fail _now_, but if we change that code and introduce any error conditions, this `unwrap()` will most likely be overlooked and might create problems.

We should avoid unwrap()s or add code comments in the places where the unwrap-is-safe assumption originates (in this case `MmdsNetworkStack::restore`)

src/devices/src/virtio/net/persist.rs
```diff
@@ -0,0 +1,174 @@
...
+        net.queues = state
+            .virtio_state
+            .queues
+            .iter()
+            .map(|queue_state| Queue::restore((), &queue_state).unwrap())
```
acatangiu, line 94: same `unwrap()` future-proofing comment

src/devices/src/virtio/net/device.rs
sandreim: I think we could follow the separate structures guidelines so we get consistency for all devices.

src/devices/src/virtio/net/device.rs
ioanachirca: That's right! Removed it.

src/devices/src/virtio/net/device.rs
ioanachirca: Renamed it to `NetConfigSpaceState`.

src/devices/src/virtio/net/persist.rs
ioanachirca: Done.

PR #1806, opened 2020-04-20
Support (de)ser for MmioTransport

### LINKED ISSUE #1714: Support (de)ser for MmioTransport
When (de)serializing a `VirtioDevice` we also need to save/restore Mmio-Virtio data not just device data.

Define a state struct for `MmioTransport` that completely describes said data.
Implement any missing functionality to be able to save/restore a `MmioTransport`

acatangiu:
## Reason for This PR

Fixes #1714 

## Description of Changes

Defines `MmioTransportState` and implements (de)serialization for `MmioTransport` primitives.
    
Creates (de)serialization framework and and hooks in the `Block` and `Vsock` devices (de)serialization to the `MmioTransport` devices that hold block or vsock devices. This allows (de)serialization of the inner device state along with `MmioTransport` primitives.

- [x] Block device hooked in.
- [x] Vsock device hooked in.
- [ ] Net device hooked in - **not done** - depends on #1773 

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/persist.rs
+++ b/src/devices/src/virtio/persist.rs
@@ -243,6 +243,15 @@ mod tests {
     }
 
     #[test]
+    fn test_net_over_mmiotransport_persistance() {
+        use crate::virtio::net::device::{tests::TestMutators, Net};
+        let mem = default_mem();
+        let net = Arc::new(Mutex::new(Net::default_net(TestMutators::default())));
+        let mmio_transport = MmioTransport::new(mem.clone(), net.clone());
+        generic_mmiotransport_persistance_test(mmio_transport, mem, net);
+    }
+
+    #[test]
     fn test_vsock_over_mmiotransport_persistance() {
         use crate::virtio::vsock::{Vsock, VsockUnixBackend};
         let mem = default_mem();
--- a/src/devices/src/virtio/persist.rs
+++ b/src/devices/src/virtio/persist.rs
@@ -5,14 +5,16 @@
 
 use super::device::*;
 use super::queue::*;
+use crate::virtio::MmioTransport;
 use crate::vm_memory::Address;
 use snapshot::Persist;
 use versionize::{VersionMap, Versionize, VersionizeResult};
 use versionize_derive::Versionize;
-use vm_memory::GuestAddress;
+use vm_memory::{GuestAddress, GuestMemoryMmap};
 
 use std::num::Wrapping;
 use std::sync::atomic::Ordering;
+use std::sync::{Arc, Mutex};
 
 #[derive(Clone, Debug, PartialEq, Versionize)]
 pub struct QueueState {
@@ -76,6 +78,7 @@ impl Persist for Queue {
 /// State of a VirtioDevice.
 #[derive(Debug, PartialEq, Versionize)]
 pub struct VirtioDeviceState {
+    pub device_type: u32,
     pub avail_features: u64,
     pub acked_features: u64,
     pub queues: Vec<QueueState>,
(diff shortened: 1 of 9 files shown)


src/devices/src/virtio/persist.rs
```diff
@@ -5,14 +5,26 @@
...
+use vm_memory::{GuestAddress, GuestMemoryMmap};
 
 use std::num::Wrapping;
 use std::sync::atomic::Ordering;
+use std::sync::{Arc, Mutex};
+
+#[derive(Debug)]
+pub enum Error {
+    BlockDeviceRestore(std::io::Error),
+    VsockBackendRestore(VsockUnixBackendError),
```
sandreim, line 25:
```suggestion
    VsockUnixBackendRestore(VsockUnixBackendError),
```

src/devices/src/virtio/persist.rs
```diff
@@ -76,6 +88,7 @@ impl Persist for Queue {
 /// State of a VirtioDevice.
 #[derive(Debug, PartialEq, Versionize)]
 pub struct VirtioDeviceState {
+    pub device_type: u32,
```
sandreim, line 91: Why do we need `device_type` in the state ?

src/devices/src/virtio/persist.rs
```diff
@@ -95,13 +109,122 @@ impl VirtioDeviceState {
...
+#[derive(Versionize)]
+pub struct MmioTransportState {
+    // The register where feature bits are stored.
+    features_select: u32,
+    // The register where features page is selected.
+    acked_features_select: u32,
+    queue_select: u32,
+    device_status: u32,
+    config_generation: u32,
+    device_state: DeviceTypeState,
```
sandreim, line 132: I would not bundle the actual device state inside the MmioTransportState. We should decouple these 2 things and the MmioTransport state should be serialized separately and the constructor struct should contain Arc<Mutex<dyn VirtioDevice>> that is constructed at the top level.

Actually we are restoring a device from lego parts that can be bundled together like this:
```
struct GenericDeviceState {
     transport_state: MmioTransportState,
     device_state: VsockState,
}
```

src/devices/src/virtio/persist.rs
```diff
@@ -95,13 +109,122 @@ impl VirtioDeviceState {
...
+impl Persist for MmioTransport {
+    type State = MmioTransportState;
+    type ConstructorArgs = MmioTransportConstructorArgs;
+    type Error = Error;
+
+    fn save(&self) -> Self::State {
+        let virtio_dev = self.device();
+        let mut unlocked_dev = virtio_dev.lock().unwrap();
+        let device_state = match unlocked_dev.device_type() {
+            TYPE_BLOCK => {
```
sandreim, line 148: The code here should be agnostic to the device type. We must not add any FC specific dependencies here.  Please see my prev comments.

src/devices/src/virtio/vsock/persist.rs
```diff
@@ -15,10 +15,16 @@ use vm_memory::GuestMemoryMmap;
...
 #[derive(Versionize)]
 pub struct VsockState {
-    cid: u64,
+    pub(crate) backend: VsockBackendState,
+    pub(crate) generic: VsockGenericState,
+}
+
+/// The Vsock serializable state.
+#[derive(Versionize)]
+pub struct VsockGenericState {
```
sandreim, line 26: How does `VsockFrontendState` sound? :)

src/devices/src/virtio/persist.rs
acatangiu: `device_type` is not used in our (de)serialization but is part of the generic virtio specification.

We should save all state irrespective of the _current_ actual implementation which may change.

src/devices/src/virtio/persist.rs
acatangiu: > Actually we are restoring a device from lego parts that can be bundled together like this:

You mean:
```rust
struct GenericDeviceState {
     transport_state: MmioTransportState,
     device_state: SomeEnumWithSpecificDeviceVariants,
}
```
right?

src/devices/src/virtio/persist.rs
acatangiu: Sounds good.. we can push the specific device type logic one (or more) levels up and keep the `MmioTransport` device-impl agnostic.

src/devices/src/virtio/vsock/persist.rs
acatangiu: sounds good

acatangiu: done

src/devices/src/virtio/persist.rs
acatangiu: done

src/devices/src/virtio/persist.rs
sandreim: I agree with that, I am questioning this because I think this cannot change in the future. The specific device ids are not going to changed.

acatangiu: I'm thinking of allowing using it as a validator/identifier if we ever need it. This ties the current `VirtioDeviceState` to a specific virtio device type.

I don't have a clear scenario where this would be imperatively necessary, but I think it's better to be safe than sorry. It's just one `u32` per device.

src/devices/src/virtio/persist.rs
```diff
@@ -136,4 +188,87 @@ mod tests {
...
+        f.as_file().set_len(0x1000).unwrap();
+        let block = default_block_with_path(f.as_path().to_str().unwrap().to_string());
+        let block = Arc::new(Mutex::new(block));
+
+        let mmio_transport = MmioTransport::new(mem.clone(), block.clone());
+        generic_mmiotransport_persistance_test(mmio_transport, mem, block);
+    }
+
+    #[test]
+    fn test_net_over_mmiotransport_persistance() {
```
sandreim, line 246: nit: Do we really need 3 tests for each device over mmio_transport that check the mmio transport ser/de and only assert the interrupt_status of the device. Would it make more sense to have only 1 test for this ?

src/devices/src/virtio/persist.rs
```diff
@@ -95,13 +99,61 @@ impl VirtioDeviceState {
...
 mod tests {
     use super::*;
+    use crate::virtio::block::device::tests::default_mem;
     use crate::virtio::mmio::tests::DummyDevice;
 
+    use utils::tempfile::TempFile;
+
     #[test]
-    fn test_persistance() {
+    fn test_queue_persistance() {
```
ioanachirca, line 156: nit: ```suggestion
    fn test_queue_persistence() {
```

Issue #1802, opened 2020-04-20
jailer doesn't ensure correct permissions for /dev inside the jail


aghecenco:
The jailer creates the `/dev`, `/dev/net` and `/run` folders inside the jail but doesn't ensure that they're accessible for the unprivileged user which firecracker will run under. This can lead to errors when attempting to open devices (`dev/kvm` and `/dev/net/tun`), even if the permissions on the devices themselves are correctly set.
https://github.com/firecracker-microvm/firecracker/blob/8943f4159bc4f48cf4b48c6cd4b5e4395dc08854/src/jailer/src/env.rs#L180

PR #1807, opened 2020-04-20
jailer: ensure correct permissions on folders inside jail

aghecenco:
## Reason for This PR

#1802  

## Description of Changes

The jailer enforces `0o755` (`rwx|r_x|r_x`)  permissions on the jail folder hierarchy, to make sure that the de-privileged firecracker can always access device nodes.

rust-vmm doesn't have a jailing crate yet.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::ffi::CStr;
-use std::fs::{self, canonicalize, File};
+use std::fs::{self, canonicalize, File, Permissions};
+use std::os::unix::fs::PermissionsExt;
 use std::os::unix::io::IntoRawFd;
 use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
@@ -30,6 +31,7 @@ const DEV_NULL_WITH_NUL: &[u8] = b"/dev/null\0";
 // Since libc::chown is not recursive, we cannot specify only /dev/net as we want
 // to walk through the entire folder hierarchy.
 const FOLDER_HIERARCHY: [&[u8]; 4] = [b"/\0", b"/dev\0", b"/dev/net\0", b"/run\0"];
+const FOLDER_PERMISSIONS: u32 = 0o700;
 
 // Helper function, since we'll use libc::dup2 a bunch of times for daemonization.
 fn dup2(old_fd: libc::c_int, new_fd: libc::c_int) -> Result<()> {
@@ -168,22 +170,27 @@ impl Env {
 
         SyscallReturnCode(unsafe { libc::chown(dev_path.as_ptr(), self.uid(), self.gid()) })
             .into_empty_result()
-            .map_err(|e| Error::ChangeFileOwner(dev_path.to_str().unwrap(), e))
+            // Safe to unwrap as we provided valid file names.
+            .map_err(|e| Error::ChangeFileOwner(PathBuf::from(dev_path.to_str().unwrap()), e))
     }
 
-    fn setup_jailed_folders(&mut self) -> Result<()> {
-        for folder in FOLDER_HIERARCHY.iter() {
-            let folder_cstr = CStr::from_bytes_with_nul(folder).map_err(Error::FromBytesWithNul)?;
+    pub fn setup_jailed_folder(&self, folder: &[u8]) -> Result<()> {
+        let folder_cstr = CStr::from_bytes_with_nul(folder).map_err(Error::FromBytesWithNul)?;
 
-            // This unwrap is safe as we provided strings that have valid utf8 chars.
-            let path = folder_cstr.to_str().unwrap();
-            fs::create_dir_all(path).map_err(|e| Error::CreateDir(PathBuf::from(path), e))?;
+        // Safe to unwrap as the byte sequence is UTF-8 validated above.
+        let path = folder_cstr.to_str().unwrap();
+        let path_buf = PathBuf::from(path);
+        fs::create_dir_all(path).map_err(|e| Error::CreateDir(path_buf.clone(), e))?;
+        fs::set_permissions(path, Permissions::from_mode(FOLDER_PERMISSIONS))
+            .map_err(|e| Error::Chmod(path_buf.clone(), e))?;
 
-            SyscallReturnCode(unsafe { libc::chown(folder_cstr.as_ptr(), self.uid(), self.gid()) })
-                .into_empty_result()
-                .map_err(|e| Error::ChangeFileOwner(folder_cstr.to_str().unwrap(), e))?;
-        }
-        Ok(())
+        #[cfg(target_arch = "x86_64")]
+        let folder_bytes_ptr = folder.as_ptr() as *const i8;
+        #[cfg(target_arch = "aarch64")]
+        let folder_bytes_ptr = folder.as_ptr();
+        SyscallReturnCode(unsafe { libc::chown(folder_bytes_ptr, self.uid(), self.gid()) })
+            .into_empty_result()
+            .map_err(|e| Error::ChangeFileOwner(path_buf, e))
     }
 
     pub fn run(mut self) -> Result<()> {
@@ -256,7 +263,10 @@ impl Env {
 
         // This will not only create necessary directories, but will also change ownership
         // for all of them.
-        self.setup_jailed_folders()?;
+        FOLDER_HIERARCHY
+            .iter()
+            .map(|f| self.setup_jailed_folder(*f))
+            .collect::<Result<()>>()?;
 
         // Here we are creating the /dev/kvm and /dev/net/tun devices inside the jailer.
         // Following commands can be translated into bash like this:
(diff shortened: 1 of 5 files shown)


src/jailer/src/env.rs
```diff
@@ -168,20 +171,34 @@ impl Env {
...
+                .map_err(Error::FromBytesWithNul)?;
+
+            // Safe to unwrap as the byte sequence is UTF-8 validated above.
             let path = folder_cstr.to_str().unwrap();
-            fs::create_dir_all(path).map_err(|e| Error::CreateDir(PathBuf::from(path), e))?;
-
-            SyscallReturnCode(unsafe { libc::chown(folder_cstr.as_ptr(), self.uid(), self.gid()) })
+            let path_buf = PathBuf::from(path);
+            fs::create_dir_all(path).map_err(|e| Error::CreateDir(path_buf.clone(), e))?;
+            fs::set_permissions(path, Permissions::from_mode(FOLDER_PERMISSIONS))
```
acatangiu, line 192: Doesn't this mean that each jail root is now accessible by any user on the system (including `uid`s of other firecracker instances? I would avoid this. Why isn't the `chown` enough?

src/jailer/src/env.rs
```diff
@@ -256,7 +273,7 @@ impl Env {
 
         // This will not only create necessary directories, but will also change ownership
         // for all of them.
-        self.setup_jailed_folders()?;
+        self.setup_jailed_folders(FOLDER_HIERARCHY.to_vec())?;
```
acatangiu, line 276: nit: this is slightly slower

src/jailer/src/env.rs
```diff
@@ -168,20 +171,34 @@ impl Env {
...
-            let folder_cstr = CStr::from_bytes_with_nul(folder).map_err(Error::FromBytesWithNul)?;
-
-            // This unwrap is safe as we provided strings that have valid utf8 chars.
+    fn setup_jailed_folders<P>(&self, folders: Vec<P>) -> Result<()>
+    where
+        P: AsRef<Path> + AsRef<OsStr>,
+    {
+        for folder in folders.iter() {
+            let mut folder_bytes = Path::new(folder).as_os_str().as_bytes().to_vec();
+            folder_bytes.push(b'\0');
```
acatangiu, line 184: nit: 
Using `&str` and dynamically pushing the 'NULL-terminating character' is more extensible, but it is also slower code on the hotpath..

Seeing how we're always setting up static things (like chroot hierarchy in this function), I would optimize the jailer code for speed rather than flexibility.

src/jailer/src/env.rs
acatangiu: After `chown`-ing the file/dir, I expect `0o700` (`rwx|___|___`) to be just right so only `uid` (and root) can access the resources in its jail.

src/jailer/src/env.rs
serban300: +1

Also this complicates the logic a bit.

src/jailer/src/env.rs
```diff
@@ -481,4 +498,44 @@ mod tests {
...
+        // );
+
+        // Success case.
+        let foo_dir = TempDir::new().unwrap();
+        let foo_path = foo_dir.as_path();
+        foo_dir.remove().unwrap();
+        assert!(env.setup_jailed_folders(vec![foo_path]).is_ok());
+
+        let metadata = fs::metadata(foo_path).unwrap();
+        // The mode bits will also have the directory bit set.
```
iulianbarbu, line 530: Is this comment referring by `directory bits` to the `setuid`, `setgid` and `sticky bit` attributes?
As far as I understand `0x4000` means that `setgid` and `sticky bit` is on. Why would these two attributes be required?

iulianbarbu: Sorry, just `sergid` actually.

src/jailer/src/env.rs
aghecenco: You're right that only the owner bits should be set. `chown` isn't enough in cases where a preconfigured `umask` removes access rights on the directory. I'll change to `700`.

src/jailer/src/env.rs
aghecenco: I'll try to figure out a way to specify the dir names with the nul terminator and without copying them around. My purpose with this change was to make `setup_jailed_folders` reusable (for the far future of an independent `jailer` crate) and unit test-able (for the immediate future of this PR).

What's the hot path you're referring to? This is done once per microvm, so I'm guessing you're thinking of the process startup time? That hasn't been exceeded (we have an integ test) but even so I'm not thrilled with the extra string operations so I'll try to get rid of them anyway.

src/jailer/src/env.rs
```diff
@@ -168,20 +171,34 @@ impl Env {
...
+            folder_bytes.push(b'\0');
+            let folder_cstr = CStr::from_bytes_with_nul(folder_bytes.as_slice())
+                .map_err(Error::FromBytesWithNul)?;
+
+            // Safe to unwrap as the byte sequence is UTF-8 validated above.
             let path = folder_cstr.to_str().unwrap();
-            fs::create_dir_all(path).map_err(|e| Error::CreateDir(PathBuf::from(path), e))?;
-
-            SyscallReturnCode(unsafe { libc::chown(folder_cstr.as_ptr(), self.uid(), self.gid()) })
+            let path_buf = PathBuf::from(path);
```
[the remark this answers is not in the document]
aghecenco: Most of the [errors](https://github.com/firecracker-microvm/firecracker/blob/608e135299b5ea46a3c7a3f9a240b229b8b03357/src/jailer/src/main.rs#L27) contain a `PathBuf`, the cloning is for convenience. A `Path` slice would have to be passed on with lifetime and all and it didn't seem quite worth the effort.

src/jailer/src/env.rs
aghecenco: But unit testable :grinning: To reiterate, I'm thinking some more on how to rework this section to get the best of both worlds

src/jailer/src/env.rs
aghecenco: Nope, it's `S_IFDIR`, the bit asserted by `S_ISDIR` (see the manpage for `stat`, or [here it is in `musl`](https://git.musl-libc.org/cgit/musl/tree/include/sys/stat.h#n31)). I'll reword the comment to make this more clear.

src/jailer/src/env.rs
acatangiu: > I'm guessing you're thinking of the process startup time? That hasn't been exceeded (we have an integ test)

You are right, that's why I marked this as a `nit`.

I am thinking in the context of future-proof work. The process start-up time we have right now is not great to begin with, and I think there's work to do to improve it. In that context, I would avoid changes that are (even theoretically) slower.

aghecenco: @acatangiu @serban300 I reworked that part to get rid of extra string copying and processing, PTAL.

src/jailer/src/env.rs
aghecenco: It's actually better to use `libc::S_IFDIR` directly instead of hardcoding some hex value, so I did that instead, PTAL.

iulianbarbu: Makes sense. Seems ok.

src/jailer/src/env.rs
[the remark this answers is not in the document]
iulianbarbu: I was referring to the `path` variable, from the same scope, which is a `str` slice. It can be used instead of the newly created `path_buf` in every place where `path` and `path_buf` are used, without a lifetime modifier. I've built it  without a lifetime modifier.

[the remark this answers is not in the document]
aghecenco: Yup I changed it back to `&str` but the error types still take a `PathBuf` so unless we change *all* of those (which I'd rather not do, at least not in this PR), some transformation is still necessary in thr `map_err` to make a `PathBuf` from a `&str`... unless I'm completely missing something, in which case, please help :grinning: The way I see it now, wither a `PathBuf` is precreated and `clone`d, or recreated at every `map_err` with `path.into()`. I think both options cost pretty much the same amount of computation.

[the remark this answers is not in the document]
iulianbarbu: Hmm, first time I suggested the change, I've managed to make it compile with `&str` everywhere. It seems it doesn't work at the moment. You can ignore my comment.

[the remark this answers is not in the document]
acatangiu: Generally, we shouldn't jump through hoops to optimize error cases. Those can be slow 😄

src/jailer/src/main.rs
```diff
@@ -136,6 +141,9 @@ impl fmt::Display for Error {
             GetOldFdFlags(ref err) => write!(f, "Failed to get flags from fd: {}", err),
             Gid(ref gid) => write!(f, "Invalid gid: {}", gid),
             InvalidInstanceId(ref err) => write!(f, "Invalid instance ID: {}", err),
+            Metadata(ref path, ref err) => {
```
dianpopa, line 144: I must be doing smth wrong, I do not see this used anywhere...

src/jailer/src/env.rs
```diff
@@ -481,4 +490,53 @@ mod tests {
...
+
+        // Error case: non UTF-8 paths.
+        let bad_string: &[u8] = &[0, 102, 111, 111, 0]; // A leading nul followed by 'f', 'o', 'o'
+        assert_eq!(
+            format!("{}", env.setup_jailed_folder(bad_string).err().unwrap()),
+            "Failed to decode string from byte array: data provided contains an interior nul byte at byte pos 0"
+        );
+
+        // Error case: inaccessible path - can't be triggered with unit tests running as root.
+        // assert_eq!(
```
dianpopa, line 509: Should we leave this here commented?

src/jailer/src/main.rs
aghecenco: Oops, leftover. Removed.

src/jailer/src/env.rs
```diff
@@ -27,9 +28,8 @@ const DEV_NULL_WITH_NUL: &[u8] = b"/dev/null\0";
 // Relevant folders inside the jail that we create or/and for which we change ownership.
 // We need /dev in order to be able to create /dev/kvm and /dev/net/tun device.
 // We need /run for the default location of the api socket.
-// Since libc::chown is not recursive, we cannot specify only /dev/net as we want
```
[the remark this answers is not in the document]
aghecenco: It does. Added it back.

aghecenco:
@iulianbarbu can you please take another look?

Issue #1770, opened 2020-04-10
Remove `dumbo` dependency on `mmds`


iulianbarbu:
`dumbo` crate is dependent on `mmds` crate. The reason is that the specific mmds network stack implementation (dumbo::ns) works very closely with `mmds`. To resolve this issue, we can move the `dumbo::ns` to `mmds` crate and pass a generic response retrieval callback to dumbo's `tcp::handler`, which currently depends on `mmds::parse_request` method.

PR #1813, opened 2020-04-22
Remove dumbo's dependency on mmds

moricho:
Signed-off-by: moricho <ikeda.morito@gmail.com>

## Reason for This PR

Fix: https://github.com/firecracker-microvm/firecracker/issues/1770

## Description of Changes

This PR 
・moves `dumbo::ns` to `mmds` crate.
・implements response retrieval callback passed to dumbo's tcp::handler
to remove the dumbo's dependency on `mmds`.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/dumbo/src/tcp/endpoint.rs
+++ b/src/dumbo/src/tcp/endpoint.rs
@@ -14,7 +14,7 @@
 use std::num::{NonZeroU16, NonZeroU64, Wrapping};
 
 use logger::{Metric, METRICS};
-use mmds::parse_request;
+use micro_http::{Body, Request, RequestError, Response, StatusCode, Version};
 use pdu::bytes::NetworkBytes;
 use pdu::tcp::TcpSegment;
 use pdu::Incomplete;
@@ -123,7 +123,11 @@ impl Endpoint {
         )
     }
 
-    pub fn receive_segment<T: NetworkBytes>(&mut self, s: &TcpSegment<T>) {
+    pub fn receive_segment<T: NetworkBytes>(
+        &mut self,
+        s: &TcpSegment<T>,
+        callback: fn(Request) -> Response,
+    ) {
         if self.stop_receiving {
             return;
         }
@@ -181,7 +185,7 @@ impl Endpoint {
             // available in self.receive_buf.
 
             // The following is some ugly but workable code that attempts to find the end of an
-            // HTTP 1.x request in receive_buf. We need to do this for now because parse_request()
+            // HTTP 1.x request in receive_buf. We need to do this for now because parse_request_bytes()
             // expects the entire request contents as parameter.
             if self.receive_buf_left > 2 {
                 let b = self.receive_buf.as_mut();
@@ -198,7 +202,8 @@ impl Endpoint {
                         };
 
                         // We found a potential request, let's parse it.
-                        let response = parse_request(&b[..end]);
+                        let response = parse_request_bytes(&b[..end], callback);
+
                         // The unwrap is safe because a Vec will allocate more space until all the
                         // writes succeed.
                         response.write_all(&mut self.response_buf).unwrap();
@@ -299,6 +304,51 @@ impl Endpoint {
     }
 }
 
+fn build_response(http_version: Version, status_code: StatusCode, body: Body) -> Response {
+    let mut response = Response::new(http_version, status_code);
+    response.set_body(body);
+    response
+}
+
+/// Parses the request bytes and builds a `micro_http::Response` by the given callback function.
+fn parse_request_bytes(byte_stream: &[u8], callback: fn(Request) -> Response) -> Response {
+    let request = Request::try_from(byte_stream);
+    match request {
+        Ok(request) => callback(request),
+        Err(e) => match e {
+            RequestError::InvalidHttpVersion(err_msg) => build_response(
+                Version::default(),
+                StatusCode::NotImplemented,
+                Body::new(err_msg.to_string()),
+            ),
+            RequestError::InvalidUri(err_msg) => build_response(
+                Version::default(),
+                StatusCode::BadRequest,
+                Body::new(err_msg.to_string()),
+            ),
+            RequestError::InvalidHttpMethod(err_msg) => build_response(
+                Version::default(),
+                StatusCode::NotImplemented,
+                Body::new(err_msg.to_string()),
+            ),
+            RequestError::InvalidRequest => build_response(
+                Version::default(),
+                StatusCode::BadRequest,
+                Body::new("Invalid request.".to_string()),
+            ),
+            RequestError::InvalidHeader => build_response(
+                Version::default(),
+                StatusCode::BadRequest,
+                Body::new("Invalid headers.".to_string()),
+            ),
+            // `micro-http` supports a predefined list of HTTP headers.
+            // It shouldn't reach this point, because it ignores the
+            // HTTP unsupported headers.
+            RequestError::UnsupportedHeader => unreachable!(),
+        },
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -308,6 +358,7 @@ mod tests {
 
     use pdu::tcp::Flags as TcpFlags;
     use tcp::connection::tests::ConnectionTester;
+    use tcp::tests::mock_callback;
 
     impl Endpoint {
         pub fn set_eviction_threshold(&mut self, value: u64) {
@@ -372,7 +423,7 @@ mod tests {
         ctrl.set_flags_after_ns(TcpFlags::ACK);
         ctrl.set_ack_number(endpoint_isn.wrapping_add(1));
         assert!(!e.connection.is_established());
-        e.receive_segment(&ctrl);
+        e.receive_segment(&ctrl, mock_callback);
         assert!(e.connection.is_established());
 
         // Also, there should be nothing to send now anymore, nor any timeout pending.
@@ -385,7 +436,7 @@ mod tests {
             data.set_flags_after_ns(TcpFlags::ACK);
             data.set_sequence_number(remote_isn.wrapping_add(1));
             data.set_ack_number(endpoint_isn.wrapping_add(1));
-            e.receive_segment(&data);
+            e.receive_segment(&data, mock_callback);
         }
 
         assert_eq!(e.receive_buf_left, incomplete_request.len());
@@ -414,7 +465,7 @@ mod tests {
             data.set_flags_after_ns(TcpFlags::ACK);
             data.set_sequence_number(remote_first_not_sent);
             data.set_ack_number(endpoint_isn + 1);
-            e.receive_segment(&data);
+            e.receive_segment(&data, mock_callback);
         }
 
         remote_first_not_sent =
@@ -432,8 +483,8 @@ mod tests {
             assert_eq!(s.inner().ack_number(), remote_first_not_sent);
 
             let response = from_utf8(s.inner().payload()).unwrap();
-            // The response should contain "404" because the MMDS is empty.
-            assert!(response.contains("404"));
+            // The response should contain "200" because the HTTP request is correct.
+            assert!(response.contains("200"));
 
             endpoint_first_not_sent = s
                 .inner()
@@ -466,7 +517,7 @@ mod tests {
                 data.set_flags_after_ns(TcpFlags::ACK);
                 data.set_sequence_number(remote_first_not_sent);
                 data.set_ack_number(endpoint_first_not_sent);
-                e.receive_segment(&data);
+                e.receive_segment(&data, mock_callback);
             }
 
             remote_first_not_sent = remote_first_not_sent.wrapping_add(request.len() as u32);
@@ -480,7 +531,7 @@ mod tests {
                 assert_eq!(s.inner().ack_number(), remote_first_not_sent);
 
                 let response = from_utf8(s.inner().payload()).unwrap();
-                assert!(response.contains("404"));
+                assert!(response.contains("200"));
 
                 endpoint_first_not_sent =
                     endpoint_first_not_sent.wrapping_add(s.inner().payload_len() as u32);
@@ -512,7 +563,7 @@ mod tests {
             data.set_flags_after_ns(TcpFlags::ACK);
             data.set_sequence_number(remote_first_not_sent);
             data.set_ack_number(endpoint_first_not_sent);
-            e.receive_segment(&data);
+            e.receive_segment(&data, mock_callback);
         }
 
         {
@@ -521,5 +572,58 @@ mod tests {
                 .unwrap();
             assert_eq!(s.inner().flags_after_ns(), TcpFlags::RST);
         }
+    }
+
+    #[test]
+    fn test_parse_request_bytes_error() {
+        // Test unsupported HTTP version.
+        let request_bytes = b"GET http://169.254.169.255/ HTTP/2.0\r\n\r\n";
+        let mut expected_response = Response::new(Version::Http11, StatusCode::NotImplemented);
+        expected_response.set_body(Body::new("Unsupported HTTP version.".to_string()));
+        let actual_response = parse_request_bytes(request_bytes, mock_callback);
+        assert_eq!(actual_response, expected_response);
+
+        // Test invalid URI (empty URI).
+        let request_bytes = b"GET   HTTP/1.0\r\n\r\n";
+        let mut expected_response = Response::new(Version::Http11, StatusCode::BadRequest);
+        expected_response.set_body(Body::new("Empty URI not allowed.".to_string()));
+        let actual_response = parse_request_bytes(request_bytes, mock_callback);
+        assert_eq!(actual_response, expected_response);
+
+        // Test invalid HTTP methods.
+        let invalid_methods = ["POST", "HEAD", "DELETE", "CONNECT", "OPTIONS", "TRACE"];
+        for method in invalid_methods.iter() {
+            let request_bytes = format!("{} http://169.254.169.255/ HTTP/1.0\r\n\r\n", method);
+            let mut expected_response = Response::new(Version::Http11, StatusCode::NotImplemented);
+            expected_response.set_body(Body::new("Unsupported HTTP method.".to_string()));
+            let actual_response = parse_request_bytes(request_bytes.as_bytes(), mock_callback);
+            assert_eq!(actual_response, expected_response);
+        }
+
+        // Test valid methods.
+        let valid_methods = ["PUT", "PATCH", "GET"];
+        for method in valid_methods.iter() {
+            let request_bytes = format!("{} http://169.254.169.255/ HTTP/1.0\r\n\r\n", method);
+            let expected_response = Response::new(Version::Http11, StatusCode::OK);
+            let actual_response = parse_request_bytes(request_bytes.as_bytes(), mock_callback);
+            assert_eq!(actual_response, expected_response);
+        }
+
+        // Test invalid HTTP format.
+        let request_bytes = b"GET / HTTP/1.1\r\n";
+        let mut expected_response = Response::new(Version::Http11, StatusCode::BadRequest);
+        expected_response.set_body(Body::new("Invalid request.".to_string()));
+        let actual_response = parse_request_bytes(request_bytes, mock_callback);
+        assert_eq!(actual_response, expected_response);
+
+        // Test invalid HTTP headers.
+        let request_bytes = b"PATCH http://localhost/home HTTP/1.1\r\n\
+                                 Expect: 100-continue\r\n\
+                                 Transfer-Encoding: identity; q=0\r\n\
+                                 Content-Length: 26\r\n\r\nthis is not\n\r\na json \nbody";
+        let mut expected_response = Response::new(Version::Http11, StatusCode::BadRequest);
+        expected_response.set_body(Body::new("Invalid headers.".to_string()));
+        let actual_response = parse_request_bytes(request_bytes, mock_callback);
+        assert_eq!(actual_response, expected_response);
     }
 }
--- a/src/dumbo/src/tcp/handler.rs
+++ b/src/dumbo/src/tcp/handler.rs
@@ -9,6 +9,7 @@ use std::collections::{HashMap, HashSet};
 use std::net::Ipv4Addr;
 use std::num::NonZeroUsize;
 
+use micro_http::{Request, Response};
 use pdu::bytes::NetworkBytes;
 use pdu::ipv4::{Error as IPv4PacketError, IPv4Packet, PROTOCOL_TCP};
 use pdu::tcp::{Error as TcpSegmentError, Flags as TcpFlags, TcpSegment};
@@ -123,11 +124,11 @@ pub struct TcpIPv4Handler {
     // Handler IPv4 address used for every connection.
     local_ipv4_addr: Ipv4Addr,
     // Handler TCP port used for every connection.
-    pub(crate) local_port: u16,
+    local_port: u16,
     // This map holds the currently active endpoints, identified by their connection tuple.
     connections: HashMap<ConnectionTuple, Endpoint>,
     // Maximum number of concurrent connections we are willing to handle.
-    pub(crate) max_connections: usize,
+    max_connections: usize,
     // Holds connections which are able to send segments immediately.
     active_connections: HashSet<ConnectionTuple>,
     // Remembers the closest timestamp into the future when one of the connections has to deal
@@ -136,7 +137,7 @@ pub struct TcpIPv4Handler {
     // RST segments awaiting to be sent.
     rst_queue: Vec<(ConnectionTuple, RstConfig)>,
     // Maximum size of the RST queue.
-    pub(crate) max_pending_resets: usize,
+    max_pending_resets: usize,
 }
 
 // Only used locally, in the receive_packet method, to differentiate between different outcomes
@@ -180,12 +181,33 @@ impl TcpIPv4Handler {
         self.local_ipv4_addr = ipv4_addr;
     }
 
+    /// Returns the local IPv4 address of this TCP handler.
+    pub fn local_ipv4_addr(&self) -> Ipv4Addr {
+        self.local_ipv4_addr
+    }
+
+    /// Returns the local port of this TCP handler.
+    pub fn local_port(&self) -> u16 {
+        self.local_port
+    }
+
+    /// Returns the max connections of this TCP handler.
+    pub fn max_connections(&self) -> usize {
+        self.max_connections
+    }
+
+    /// Returns the max pending resetes of this TCP handler.
+    pub fn max_pending_resets(&self) -> usize {
+        self.max_pending_resets
+    }
+
     /// Contains logic for handling incoming segments.
     ///
     /// Any changes to the state if the handler are communicated through an `Ok(RecvEvent)`.
     pub fn receive_packet<T: NetworkBytes>(
         &mut self,
         packet: &IPv4Packet<T>,
+        callback: fn(Request) -> Response,
     ) -> Result<RecvEvent, RecvError> {
         // TODO: We skip verifying the checksum, just in case the device model relies on offloading
         // checksum computation from the guest to some other entity. Clear this up at some point!
@@ -200,7 +222,7 @@ impl TcpIPv4Handler {
         let tuple = ConnectionTuple::new(packet.source_address(), segment.source_port());
 
         let outcome = if let Some(endpoint) = self.connections.get_mut(&tuple) {
-            endpoint.receive_segment(&segment);
+            endpoint.receive_segment(&segment, callback);
             if endpoint.is_done() {
                 RecvSegmentOutcome::EndpointDone
             } else {
@@ -484,14 +506,9 @@ impl TcpIPv4Handler {
 #[cfg(test)]
 mod tests {
     use pdu::bytes::NetworkBytesMut;
+    use tcp::tests::mock_callback;
 
     use super::*;
-
-    impl TcpIPv4Handler {
-        pub fn local_ipv4_addr(&self) -> Ipv4Addr {
-            self.local_ipv4_addr
-        }
-    }
 
     fn inner_tcp_mut<'a, 'b, T: NetworkBytesMut>(
         p: &'a mut IPv4Packet<'b, T>,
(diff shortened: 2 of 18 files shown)


src/dumbo/src/tcp/handler.rs
```diff
@@ -148,6 +149,48 @@ enum RecvSegmentOutcome {
     UnexpectedSegment(bool),
 }
 
+/// build_response build micro_http::Response with given http version, status code, and body.
+///
```
iulianbarbu, line 153: Remove this empty comment line.

src/dumbo/src/tcp/handler.rs
```diff
@@ -148,6 +149,48 @@ enum RecvSegmentOutcome {
     UnexpectedSegment(bool),
 }
 
+/// build_response build micro_http::Response with given http version, status code, and body.
```
iulianbarbu, line 152:
```suggestion
/// `build_response` builds the `micro_http::Response` with a given HTTP version, status code, and body.
```

src/dumbo/src/tcp/handler.rs
```diff
@@ -148,6 +149,48 @@ enum RecvSegmentOutcome {
...
+///
+pub fn build_response(http_version: Version, status_code: StatusCode, body: Body) -> Response {
+    let mut response = Response::new(http_version, status_code);
+    response.set_body(body);
+    response
+}
+
+/// parse_request parses the request and build response by given callback function.
+///
```
iulianbarbu, line 161: Remove this empty comment line.

src/dumbo/src/tcp/handler.rs
```diff
@@ -148,6 +149,48 @@ enum RecvSegmentOutcome {
...
 
+/// build_response build micro_http::Response with given http version, status code, and body.
+///
+pub fn build_response(http_version: Version, status_code: StatusCode, body: Body) -> Response {
+    let mut response = Response::new(http_version, status_code);
+    response.set_body(body);
+    response
+}
+
+/// parse_request parses the request and build response by given callback function.
```
iulianbarbu, line 160:
```suggestion
/// `parse_request` parses the request bytes and builds a `micro_http::Response` by the given callback function.
```

src/dumbo/src/tcp/handler.rs
```diff
@@ -148,6 +149,48 @@ enum RecvSegmentOutcome {
...
+                Version::default(),
+                StatusCode::BadRequest,
+                Body::new("Invalid request.".to_string()),
+            ),
+            RequestError::InvalidHeader => build_response(
+                Version::default(),
+                StatusCode::BadRequest,
+                Body::new("Invalid headers.".to_string()),
+            ),
+            RequestError::UnsupportedHeader => unreachable!(),
```
iulianbarbu, line 189: This will panic if the `Request` will contain an unsupported header. This can happen when trying to parse a `Request` and the `Content-type` is set to a content type different than `text/plain` or `application/json`. Can we return here a `BadRequest` error response as well?

src/dumbo/src/tcp/handler.rs
```diff
@@ -488,10 +532,87 @@ impl TcpIPv4Handler {
...
+        let response_body = String::from("test");
+        build_response(
+            request.http_version(),
+            StatusCode::OK,
+            Body::new(response_body),
+        )
+    }
+
+    #[test]
+    fn test_parse_request() {
```
iulianbarbu, line 559: Please add a test for `UnsupportedHeader` case as well .

moricho:
@iulianbarbu Thanks for your review! I fixed. In addition, I moved `persist` module from `dumbo` to `mmds` which was implemented in https://github.com/firecracker-microvm/firecracker/pull/1773 

ref: https://github.com/firecracker-microvm/firecracker/pull/1813/commits/7de102fbf3fe63c8afa7c29ff956939fe291e686

src/dumbo/src/tcp/handler.rs
```diff
@@ -148,6 +149,46 @@ enum RecvSegmentOutcome {
     UnexpectedSegment(bool),
 }
 
+/// `build_response` builds the `micro_http::Response` with a given HTTP version, status code, and body.
```
iulianbarbu, line 152: Why is this needed?

src/mmds/src/lib.rs
```diff
@@ -5,15 +5,24 @@
...
 
 pub mod data_store;
+pub mod ns;
+pub mod persist;
 
 use serde_json::{Map, Value};
 use std::sync::{Arc, Mutex};
 
 use data_store::{Error as MmdsError, Mmds};
-use micro_http::{Body, Request, RequestError, Response, StatusCode, Version};
```
iulianbarbu, line 16: My thought is that `build_response` function need to be implemented at the level of `mmds` crate. As you've provided right now, we would need to consume `build_response` from `dumbo` to create `micro-http::Response`s, which is not ergonomic when `dumbo::build_response` would need to be extended, to add other headers lets say, to the provided response. For example, if one needs to add an extra `micro-http::Header` to the response, it will need to call first `dumbo::build_response`, then take a mutable reference to it, and then add the header. Another option will be to adjust this `dumbo::build_response` function for multiple response building cases but I do not think it is in the scope of the `dumbo` crate. `micro-http::Response` already provides this API for building responses for multiple cases, so reinventing a build response API in `dumbo` does not worth it.  I suggest to make this response builder function implementation specific. What do you think?

iulianbarbu:
Hey, @moricho ! Let us know if you need any help with this.

moricho:
@iulianbarbu I’m sorry for the late reply! I'll start to work on this issue again soon :) Just a moment, please.

iulianbarbu:
> @iulianbarbu I’m sorry for the late reply! I'll start to work on this issue again soon :) Just a moment, please. Thank you for your help.

Cool. No problem. Take your time. Let us know if we can help. :)

src/mmds/src/lib.rs
moricho: @iulianbarbu 
I agree with your thought. Indeed I implemented `build_response` function in `dumbo` crate only for using it in `dumbo::parse_request` function. If we implement `build_response` in `mmds` crate, how should we pass `mmds::build_response` to `dumbo::parse_request`?
I think the first way to do that is passing `mmds::build_response` through one of the argument like `callback` as I've already implemented. Do you have any other ideas?

moricho:
@iulianbarbu Thank you for waiting! I resolved conflicts for now and replied to your comment(https://github.com/firecracker-microvm/firecracker/pull/1813#discussion_r417984014)

src/mmds/src/lib.rs
iulianbarbu: Passing the response through the callback is perfectly fine, as far as I can tell. I do not have any issue with it.

src/dumbo/src/tcp/handler.rs
```diff
@@ -180,12 +181,33 @@ impl TcpIPv4Handler {
         self.local_ipv4_addr = ipv4_addr;
     }
 
+    /// Getter for the local IPv4 address of this TCP handler.
+    pub fn get_local_ipv4_addr(&self) -> Ipv4Addr {
+        self.local_ipv4_addr
+    }
+
+    /// Getter for the local port of this TCP handler.
```
lauralt, line 189: There is no need to include `getter` in the comment. This comment applies for the other getters as well.

```suggestion
    /// Returns the local port of this TCP handler.
```

src/dumbo/src/tcp/handler.rs
```diff
@@ -493,6 +516,16 @@ mod tests {
         }
     }
 
+    fn mock_callback(request: Result<Request, RequestError>) -> Response {
```
lauralt, line 519: If we want to create a mock callback, wouldn't it be enough to return some `Response` type (for example `Response::new(Version::Http11, StatusCode::OK)`) and not even check the `request` variant via that match?

Also, this function looks like a duplicate. Can we get rid of one of them?

src/mmds/src/lib.rs
```diff
@@ -31,6 +39,13 @@ impl Into<OutputFormat> for MediaType {
     }
 }
 
+/// `build_response` builds the `micro_http::Response` with a given HTTP version, status code, and body.
+pub fn build_response(http_version: Version, status_code: StatusCode, body: Body) -> Response {
```
lauralt, line 43: Does this need to be `pub`?

src/dumbo/src/tcp/endpoint.rs
```diff
@@ -299,16 +303,36 @@ impl Endpoint {
     }
 }
 
+/// `parse_request` parses the request bytes and builds a `micro_http::Response` by the given callback function.
+pub fn parse_request(
+    request_bytes: &[u8],
+    callback: fn(Result<Request, RequestError>) -> Response,
+) -> Response {
```
iulianbarbu, line 310: This `parse_request` needs to map a stream of bytes to a `micro_http::Request`. If any error is raised during this mapping, this error should be sent to upper levels. 

```suggestion
) -> Result<Response, RequestError> {
```

src/dumbo/src/tcp/endpoint.rs
```diff
@@ -299,16 +303,36 @@ impl Endpoint {
...
+    let request = Request::try_from(request_bytes);
```
iulianbarbu, line 311:
```suggestion
    let request = Request::try_from(request_bytes)?;
```

src/dumbo/src/tcp/endpoint.rs
```diff
@@ -299,16 +303,36 @@ impl Endpoint {
...
+    callback(request)
```
iulianbarbu, line 312:
```suggestion
    Ok(callback(request))
```

src/dumbo/src/tcp/endpoint.rs
```diff
@@ -299,16 +303,36 @@ impl Endpoint {
...
     use super::*;
 
     use std::fmt;
     use std::str::from_utf8;
 
+    use micro_http::{Request, RequestError, Response, StatusCode, Version};
     use pdu::tcp::Flags as TcpFlags;
     use tcp::connection::tests::ConnectionTester;
 
+    fn mock_callback(request: Result<Request, RequestError>) -> Response {
```
iulianbarbu, line 326: Even the mock callback should process an already obtained `micro_http::Request` and return a correspondent response.
```suggestion
    fn mock_callback(request: Request) -> Response {
```

src/dumbo/src/tcp/endpoint.rs
```diff
@@ -299,16 +303,36 @@ impl Endpoint {
...
+        match request {
+            Ok(request) => {
+                let _uri = request.uri().get_abs_path();
+                Response::new(Version::Http11, StatusCode::OK)
+            }
+            Err(_) => Response::new(Version::Http11, StatusCode::BadRequest),
+        }
```
iulianbarbu, line 333:
```suggestion
            // If we really need some request based processing in the mock callback, we should do it here.
           // Otherwise, return a default response.
           Response::new(Version::Http11, StatusCode::OK)
        }
```

dianpopa:
If I understand correctly issue #1770 aims at decoupling the specific mmds logic that is found inside the dumbo crate. Another module that is totally specific to Firecracker's mmds is also the `endpoint` module. See this [doc comment](https://github.com/firecracker-microvm/firecracker/pull/1813/files#diff-3b36cb6cd2f1ea7852b501d50ad0ad32R4). So putting a callback in the receive_packet function does not solve our desire to make dumbo Firecracker independent. 
What I see as a clean fix for this is to have an endpoint interface defined in dumbo that asks the implementing crate for the needed functionality. Another thing that is Firecracker specific is `struct ConnectionTuple` which acts as a way of uniquely identifying an `Endpoint`:
So what I suggest is this:
* the handler module offers a `Endpoint` interface:
```
pub trait Endpoint {
    fn get_id(&self) -> EndpointId;
    fn receive_segment<T: NetworkBytes>(
        &mut self,
        s: &TcpSegment<T>) -> Response;
    ...
}
```
`EndpointId `will play the role of `ConnectionTuple` and it could be
`pub type EndpointId = u64;` the hash of the unique string that would result by appending the remote address and port (the old fields of ConnectionTuple).
* endpoint module from dumbo moves to mmds as an implementation for the endpoint interface from handler.rs
* TcpIPv4Handler becomes this:

```
pub struct TcpIPv4Handler {
    // Handler IPv4 address used for every connection.
    local_ipv4_addr: Ipv4Addr,
    // Handler TCP port used for every connection.
    local_port: u16,
    // This map holds the currently active endpoints, identified by their connection tuple.
    endpoints: HashMap<EndpointId, dyn Endpoint>,
    // Maximum number of concurrent connections we are willing to handle.
    max_connections: usize,
    // Holds connections which are able to send segments immediately.
    active_connections: HashSet<EndpointId>,
    // Remembers the closest timestamp into the future when one of the connections has to deal
    // with an RTO trigger.
    next_timeout: Option<(u64, EndpointId)>,
    // RST segments awaiting to be sent.
    rst_queue: Vec<(EndpointId, RstConfig)>,
    // Maximum size of the RST queue.
    max_pending_resets: usize,
}
```
WDYT? is this worth trying out?

iulianbarbu:
> If I understand correctly issue #1770 aims at decoupling the specific mmds logic that is found inside the dumbo crate. Another module that is totally specific to Firecracker's mmds is also the `endpoint` module. See this [doc comment](https://github.com/firecracker-microvm/firecracker/pull/1813/files#diff-3b36cb6cd2f1ea7852b501d50ad0ad32R4). So putting a callback in the receive_packet function does not solve our desire to make dumbo Firecracker independent.
> What I see as a clean fix for this is to have an endpoint interface defined in dumbo that asks the implementing crate for the needed functionality. Another thing that is Firecracker specific is `struct ConnectionTuple` which acts as a way of uniquely identifying an `Endpoint`:
> So what I suggest is this:
> 
>     * the handler module offers a `Endpoint` interface:
> 
> 
> ```
> pub trait Endpoint {
>     fn get_id(&self) -> EndpointId;
>     fn receive_segment<T: NetworkBytes>(
>         &mut self,
>         s: &TcpSegment<T>) -> Response;
>     ...
> }
> ```
> 
> `EndpointId `will play the role of `ConnectionTuple` and it could be
> `pub type EndpointId = u64;` the hash of the unique string that would result by appending the remote address and port (the old fields of ConnectionTuple).
> 
>     * endpoint module from dumbo moves to mmds as an implementation for the endpoint interface from handler.rs
> 
>     * TcpIPv4Handler becomes this:
> 
> 
> ```
> pub struct TcpIPv4Handler {
>     // Handler IPv4 address used for every connection.
>     local_ipv4_addr: Ipv4Addr,
>     // Handler TCP port used for every connection.
>     local_port: u16,
>     // This map holds the currently active endpoints, identified by their connection tuple.
>     endpoints: HashMap<EndpointId, dyn Endpoint>,
>     // Maximum number of concurrent connections we are willing to handle.
>     max_connections: usize,
>     // Holds connections which are able to send segments immediately.
>     active_connections: HashSet<EndpointId>,
>     // Remembers the closest timestamp into the future when one of the connections has to deal
>     // with an RTO trigger.
>     next_timeout: Option<(u64, EndpointId)>,
>     // RST segments awaiting to be sent.
>     rst_queue: Vec<(EndpointId, RstConfig)>,
>     // Maximum size of the RST queue.
>     max_pending_resets: usize,
> }
> ```
> 
> WDYT? is this worth trying out?

Not sure if this is worth trying out for this specific issue. As far as #1770 goes, the main topic, which I've thought about, was to remove the unnecessary dumbo dependency on MMDS.

One the other hand, while looking into your investigation, it seems like an interesting direction for further thought and maybe prototype. The dumbo's endpoint dependency on MMDS logic can be seen rather as a theoretical dependency, because of the fact that dumbo is designed to provide the necessary network stack for Ethernet/IP/TCP/HTTP communication. This shouldn't be necessary the case, because dumbo's Ethernet/IP/TCP dumbo network stack can be used also with an arbitrary application protocol (not necessary HTTP, required for MMDS). Hence, you propose an interface to be implemented by applications, for their network stack specific use case, where they can assert on their preference for the application protocol. This can be also generalized for the network protocol or for the link-layer protocol, but I can not tell if there is any data for such a need, rather than generalizing/decoupling dumbo's implementation and making it attractive for similar Firecracker use cases.

Please, let me know your thoughts.

src/mmds/src/lib.rs
```diff
@@ -59,14 +74,7 @@ pub fn json_patch(target: &mut Value, patch: &Value) {
...
 
-fn build_response(http_version: Version, status_code: StatusCode, body: Body) -> Response {
-    let mut response = Response::new(http_version, status_code);
-    response.set_body(body);
-    response
-}
-
-pub fn parse_request(request_bytes: &[u8]) -> Response {
-    let request = Request::try_from(request_bytes);
+pub fn callback(request: Result<Request, RequestError>) -> Response {
```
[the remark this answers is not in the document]
lauralt: +1 on renaming `callback()`. From what I see, this function is used in `parse_request()`, maybe it would make sense to rename that function to `parse_request_bytes()` and this one to `convert_to_response()`.

src/dumbo/src/tcp/endpoint.rs
lauralt: Hmm, it seems that this way we won't build the responses that were built before in the `Err(e)` match branch from `parse_request()` (or `callback()` now).
Are you thinking we should move that part somewhere else? It seems okay to build all the possible responses in the same place; maybe `callback()` should be somehow split in smaller functions, but I don't think we should treat those "error" cases in an upper level (as you mentioned I guess in a comment above).
Sorry in advance if I'm missing something, it's kind of the first time I'm looking at this code :sweat:.

src/mmds/src/lib.rs
```diff
@@ -156,19 +164,28 @@ pub fn parse_request(request_bytes: &[u8]) -> Response {
...
+        let actual_response = parse_request(request.as_bytes(), callback);
         assert!(expected_response.status() == actual_response.status());
         assert!(expected_response.body().unwrap() == actual_response.body().unwrap());
     }
 
+    fn check_callback_response(actual: Response, expected: Response) {
+        assert!(expected.status() == actual.status());
+        assert!(expected.body().unwrap() == actual.body().unwrap());
+        assert!(expected.http_version() == actual.http_version());
+    }
```
lauralt, line 185: This looks like a `PartialEq` implementation. Would it make sense to `impl PartialEq` here instead?

iulianbarbu:
Hey, @moricho ! How it's going? Are you planning to address the feedback anytime soon? We would be glad to take a stab on it, to address what is left, if you are out of bandwidth. Let us know if that's OK.

moricho:
@iulianbarbu Thanks for reminding me. I'll come back and address the feedback soon!

src/dumbo/src/tcp/endpoint.rs
lauralt: Sorry for the late reply, your proposal sounds good! From what it seems, there shouldn't be a lot of effort involved, at least from the implementation point of view; in tests, on the other hand, I think there will be several changes involved :-?. @moricho what do you think about this proposal? Does it make sense to you?

moricho: @iulianbarbu @lauralt It makes sense to me. Thanks for the detailed explanation! Based on your proposal, I'm now refactoring that part. 
In that regard, could I ask you to give me some advice?  Do you think it's reasonable that we implement a function to build error response at `endpoint` level to treat `parse_request` error, aside from existing one to build mmds response (`build_response` func) at `mmds` crate ?

iulianbarbu: @moricho , build response at the endpoint level makes sense for me, for responding to HTTP request parse errors.

I see it like this:

Dumbo assures the HTTP request is valid and if so it is passing it to MMDS. If the HTTP request is not valid, dumbo will emit a response corresponding to the `RequestError` encountered while parsing.

Before your PR, `RequestError` handling was done inside MMDS, where the request buffer was also parsed. Then, `mmds::build_response` was returning MMDS related errors (e.g `MmdsError::NotFound`) and micro-http request parsing error as well (e.g. `RequestError::InvalidUri`). Now, because we want to move the HTTP requests parsing in dumbo, we have to move `RequestError` handling (building and sending an HTTP response in case of parse error) in dumbo. Here, we can use a `build_response` function similar to `build_response` function from MMDS.

Let me know if this makes sense.

moricho: @iulianbarbu Thanks for your advice! It makes sense for me. Based on this discussion, I'll do refactoring continuously.

src/dumbo/src/tcp/endpoint.rs
```diff
@@ -522,4 +586,72 @@ mod tests {
...
+        // Test invalid HTTP format.
+        let request_bytes = b"GET / HTTP/1.1\r\n";
+        let mut expected_response = Response::new(Version::Http11, StatusCode::BadRequest);
+        expected_response.set_body(Body::new("Invalid request.".to_string()));
+        let actual_response = parse_request_bytes(request_bytes, mock_callback);
+        assert!(actual_response == expected_response);
+    }
+
+    #[test]
+    fn test_unsupported_http_method() {
```
iulianbarbu, line 615: This test is not relevant for Dumbo, but it is relevant for MMDS. Can you move it back to MMDS, please? Have in mind also that this test, in MMDS, was checking against `PUT` and `PATCH`, which were not allowed methods, in that context.

moricho: Thank you for comment. I have moved `test_unsupported_http_method ` from `dumbo` to `mmds` and also added `PUT` and `PATCH` cases.

src/mmds/src/lib.rs
moricho: @lauralt Thank you for review! That makes sense. I added `#[derive(PartialEq)]` for the necessary part and remove `check_callback_response` function.

src/mmds/src/lib.rs
```diff
@@ -78,108 +87,80 @@ fn sanitize_uri(mut uri: String) -> String {
...
-        let request = format!("{} http://169.254.169.254/ HTTP/1.1\r\n\r\n", method);
+        let request_string = format!("{} http://169.254.169.254/ HTTP/1.1\r\n\r\n", method);
+        let request_bytes = request_string.as_bytes();
         let mut expected_response = Response::new(Version::Http11, status);
         expected_response.set_body(Body::new(err_msg));
-        let actual_response = parse_request(request.as_bytes());
+        if status == StatusCode::MethodNotAllowed {
+            expected_response.allow_method(Method::Get);
+        }
+        expected_response.allow_method(Method::Get);
```
iulianbarbu, line 162: These lines conflicts themselves. You add an `Allow` header even if the expected response should not contain it (e.g. check method failure for "GET", StatusCode::Ok, "dont care"). Furthermore, the added Allow header is not asserted against the actual response.

```suggestion
        if status == StatusCode::MethodNotAllowed {
            expected_response.allow_method(Method::Get);
        }
```

src/mmds/src/lib.rs
```diff
@@ -78,108 +87,80 @@ fn sanitize_uri(mut uri: String) -> String {
...
+        let actual_response = parse_request_bytes(request_bytes, convert_to_response);
```
iulianbarbu, line 163:
```suggestion
        let actual_response = parse_request_bytes(request_bytes, convert_to_response);
        assert!(expected_response.allow() == actual_response.allow())
```

iulianbarbu:
For a better history, I would also squash all the commits inside the first one (`dumbo: remove dumbo's dependency on mmds`).

src/mmds/src/lib.rs
```diff
@@ -78,110 +87,80 @@ fn sanitize_uri(mut uri: String) -> String {
...
-                    Body::new("Not allowed HTTP method."),
-                );
-                response.allow_method(Method::Get);
-                return response;
-            }
-
-            // The data store expects a strict json path, so we need to
-            // sanitize the URI.
-            let json_pointer = sanitize_uri(uri.to_string());
+pub fn convert_to_response(request: Request) -> Response {
```
lauralt, line 90: This doesn't need to be `pub`.

src/mmds/src/lib.rs
```diff
@@ -78,110 +87,80 @@ fn sanitize_uri(mut uri: String) -> String {
...
     }
 }
 
 #[cfg(test)]
 mod tests {
     extern crate serde_json;
+
     use super::*;
+    use dumbo::tcp::parse_request_bytes;
+    use micro_http::{Request, Response, StatusCode, Version};
```
lauralt, line 152:
```suggestion
    use micro_http::Version;
```

src/mmds/src/lib.rs
```diff
@@ -78,110 +87,80 @@ fn sanitize_uri(mut uri: String) -> String {
...
         let mut expected_response = Response::new(Version::Http11, status);
         expected_response.set_body(Body::new(err_msg));
-        let actual_response = parse_request(request.as_bytes());
-        assert!(expected_response.status() == actual_response.status());
-        assert!(expected_response.body().unwrap() == actual_response.body().unwrap());
+        if status == StatusCode::MethodNotAllowed {
+            expected_response.allow_method(Method::Get);
+        }
+        let actual_response = parse_request_bytes(request_bytes, convert_to_response);
+        assert!(expected_response.allow() == actual_response.allow());
```
lauralt, line 163:
```suggestion
        assert_eq!(expected_response.allow(), actual_response.allow());
```

src/mmds/src/lib.rs
```diff
@@ -224,54 +203,44 @@ mod tests {
...
-        let request = b"HTTP/1.1";
-        let dummy_response = Response::new(Version::Http11, StatusCode::BadRequest);
-        assert!(parse_request(request).status() == dummy_response.status());
+        // Test resource not found.
+        let request_bytes = b"GET http://169.254.169.254/invalid HTTP/1.0\r\n\r\n";
+        let request = Request::try_from(request_bytes).unwrap();
+        let mut expected_response = Response::new(Version::Http10, StatusCode::NotFound);
+        expected_response.set_body(Body::new("Resource not found: /invalid.".to_string()));
+        let actual_response = convert_to_response(request);
+        assert!(actual_response == expected_response);
```
lauralt, line 212: nit:
```suggestion
        assert_eq!(actual_response, expected_response);
```
Applies to the entire test.

src/mmds/src/lib.rs
```diff
@@ -224,54 +203,44 @@ mod tests {
...
 
-        // Test unsupported HTTP version.
-        let request = b"GET http://169.254.169.255/ HTTP/2.0\r\n\r\n";
+        // Test NotImplemented
```
lauralt, line 214: nit:
```suggestion
        // Test NotImplemented.
```

src/dumbo/src/tcp/endpoint.rs
```diff
@@ -299,15 +304,62 @@ impl Endpoint {
     }
 }
 
+fn build_response(http_version: Version, status_code: StatusCode, body: Body) -> Response {
```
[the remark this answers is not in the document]
iulianbarbu: I can not find where `parse_request_bytes()` is used outside `dumbo` crate. Can you please tell where is the usage?

There are two drawbacks that I see for keeping one single `build_response`:
1) `build_response` from `dumbo` can change/be deleted and MMDS is impacted.
2) `build_response` is a wrapper over `micro_http::Request` building. Using it inside MMDS requires to make it a public method. This would blur the dumbos API by exposing a non-`dumbo` functionality.

Although they are duplicate, I would argue that keeping them duplicate enforces better crates separation, which is preferred. Please take a look at #1819 , which is introduced by a similar scenario. There, MMDS had a dependency on micro-http inner implementation, which assured that every HTTP request is a GET. However, this implementation had changed and the previous assumption did not hold anymore, resulting in that issue.

src/dumbo/src/tcp/mod.rs
```diff
@@ -92,6 +93,12 @@ pub fn seq_at_or_after(a: Wrapping<u32>, b: Wrapping<u32>) -> bool {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use micro_http::{Request, Response, StatusCode, Version};
+
+    pub fn mock_callback(request: Request) -> Response {
```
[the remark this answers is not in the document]
iulianbarbu: My 2 cents: The comment says the same thing as the code :)

pub - expect usage outside the module.
mock_.. - compile the minimum functionality to produce the needed side effect, even if it is dummy.

src/dumbo/src/tcp/endpoint.rs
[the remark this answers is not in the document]
lauralt: > I can not find where `parse_request_bytes()` is used outside `dumbo` crate. Can you please tell where is the usage?

https://github.com/moricho/firecracker/commit/68b6c0ab22cfa69b20ba39e1156d2f1c668b1a2b#diff-7d882cb8cb3286a3411207ff13d0c72bR162

and the import is here: https://github.com/moricho/firecracker/commit/68b6c0ab22cfa69b20ba39e1156d2f1c668b1a2b#diff-f7e4add39fc8dbdc081b77a61f5c586bR10 :(

> There are two drawbacks that I see for keeping one single `build_response`:
> 
> 1. `build_response` from `dumbo` can change/be deleted and MMDS is impacted.
> 2. `build_response` is a wrapper over `micro_http::Request` building. Using it inside MMDS requires to make it a public method. This would blur the dumbos API by exposing a non-`dumbo` functionality.
> 
> Although they are duplicate, I would argue that keeping them duplicate enforces better crates separation, which is preferred. Please take a look at #1819 , which is introduced by a similar scenario. There, MMDS had a dependency on micro-http inner implementation, which assured that every HTTP request is a GET. However, this implementation had changed and the previous assumption did not hold anymore, resulting in that issue.

Hmm, the situations are somehow not the same; assuring that every HTTP request is a GET sounds like something MMDS-specific, therefore the check/enforcement should be done at MMDS level, here `build_response` doesn't contain any dumbo/MMDS logic, it's more of a helper function, so that's why I suggested to leave just one. The only thing that I would expect to be different between a response that is built at MMDS level and one that is built at dumbo level is related to the allowed method(s). I agree it's weird to expose a non-dumbo function via the dumbo API, I don't like that either, we should definitely avoid it, another suggestion that I had in mind when I left the comment was to move it to `micro_http` since it's a pretty basic/generic function that can build a response based on some input.
We can also keep them duplicate to be super safe, but I would've liked to avoid that.

Also, it would be great if we could avoid using `parse_request_bytes` in MMDS tests. Btw, just curious, `MethodNotAllowed` seems already tested in `test_convert_to_response()`, why do we test it again in `test_unsupported_http_method`? Also, it looks like we are testing in MMDS the `micro_http` unsupported methods. It makes sense to test them again here as a sanity check since we must make sure that only `GET` is supported, but the problem that I see is that we don't actually check in MMDS for the `NotImplemented` methods, we kind of rely on dumbo to do the proper checks and build the right response. I would drop `test_unsupported_http_method()` until we make a clear separation here between dumbo and MMDS.

src/dumbo/src/tcp/mod.rs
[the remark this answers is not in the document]
lauralt: I'm fine either way :+1:. This function looks a bit random here and wanted to somehow explain what's up with it.

moricho:
@lauralt Thank you for review. I fixed some parts you mentioned.

@iulianbarbu @lauralt 
I'll reply to those comment and subsequent discussion later :)
・https://github.com/firecracker-microvm/firecracker/pull/1813#discussion_r448355890
・https://github.com/firecracker-microvm/firecracker/pull/1813#discussion_r448406433

src/dumbo/src/tcp/endpoint.rs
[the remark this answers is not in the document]
lauralt: And regarding `build_response()`, the logic from it seems to be the same in `dumbo`, `mmds` and `api_server` too (https://github.com/firecracker-microvm/firecracker/blob/master/src/api_server/src/parsed_request.rs#L94). Since building a response based on some specific inputs is rather application specific, it would make the most sense to have this in a helper crate from Firecracker, i.e. the `utils` one. But this way, `utils` will have an extra dependency on `micro-http` and we don't think it's worth it. It's fine to leave `build_response()` duplicated in `mmds` and `dumbo`.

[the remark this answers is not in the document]
moricho: I fixed!

Issue #1819, opened 2020-04-24
HTTP method check for guest MMDS requests


iulianbarbu:
When an application from within the guest want to retrieve metadata from MMDS, the documentation is specific in how this request has to be formed. However, we do not enforce validation on the HTTP method of the request, and we would end up responding to requests which use atypical HTTP methods for retrieving metadata, like `PUT` and `PATCH`.

iulianbarbu:
Fixed by #1820 .

PR #1800, opened 2020-04-17
docs: Added MMDS user guide

iulianbarbu:
## Reason for This PR

Closes #1801.

MMDS documentation is falling behind, once the new changes targeted on MMDS are on their way to be merged (#1790, #1786, #1772). Dedicated documentation was added in this PR, to be reviewed and addressed separated. I chose to add a separate PR with docs because adding it in every PR related to MMDS can not easily scale, and might be conflictual between separate PRs.

## Description of Changes

These changes segregates the information which describes MMDS design from the information which describes its usage.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

- [x] All commits in this PR are signed (`git commit -s`).
- [x] The reason for this PR is clearly provided (issue no. or explanation).
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] Any newly added `unsafe` code is properly documented.
- [x] Any API changes are reflected in `firecracker/swagger.yaml`.
- [x] Any user-facing changes are mentioned in `CHANGELOG.md`.

PR #1820, opened 2020-04-24
MMDS HTTP method check for guest requests

iulianbarbu:
## Reason for This PR

#1819

## Description of Changes

Add a validation check to enforce MMDS HTTP requests to use only `GET`.
Documentation about MMDS guest requests failures can be found in #1800 .

## License Acceptance

## PR Checklist

### DIFF
--- a/src/micro_http/src/response.rs
+++ b/src/micro_http/src/response.rs
@@ -6,6 +6,7 @@ use std::io::{Error as WriteError, Write};
 use ascii::{COLON, CR, LF, SP};
 use common::{Body, Version};
 use headers::{Header, MediaType};
+use Method;
 
 /// Wrapper over a response status code.
 ///
@@ -24,6 +25,8 @@ pub enum StatusCode {
     BadRequest,
     /// 404, Not Found
     NotFound,
+    /// 405, Method Not Allowed
+    MethodNotAllowed,
     /// 500, Internal Server Error
     InternalServerError,
     /// 501, Not Implemented
@@ -39,6 +42,7 @@ impl StatusCode {
             Self::NoContent => b"204",
             Self::BadRequest => b"400",
             Self::NotFound => b"404",
+            Self::MethodNotAllowed => b"405",
             Self::InternalServerError => b"500",
             Self::NotImplemented => b"501",
         }
@@ -75,6 +79,7 @@ pub struct ResponseHeaders {
     content_length: i32,
     content_type: MediaType,
     server: String,
+    allow: Vec<Method>,
 }
 
 impl Default for ResponseHeaders {
@@ -83,11 +88,31 @@ impl Default for ResponseHeaders {
             content_length: Default::default(),
             content_type: Default::default(),
             server: String::from("Firecracker API"),
+            allow: Vec::new(),
         }
     }
 }
 
 impl ResponseHeaders {
+    // The logic pertaining to `Allow` header writing.
+    fn write_allow_header<T: Write>(&self, buf: &mut T) -> Result<(), WriteError> {
+        if self.allow.is_empty() {
+            return Ok(());
+        }
+
+        buf.write_all(b"Allow: ")?;
+
+        let delimitator = b", ";
+        for (idx, method) in self.allow.iter().enumerate() {
+            buf.write_all(method.raw())?;
+            if idx < self.allow.len() - 1 {
+                buf.write_all(delimitator)?;
+            }
+        }
+
+        buf.write_all(&[CR, LF])
+    }
+
     /// Writes the headers to `buf` using the HTTP specification.
     pub fn write_all<T: Write>(&self, buf: &mut T) -> Result<(), WriteError> {
         buf.write_all(Header::Server.raw())?;
@@ -97,6 +122,8 @@ impl ResponseHeaders {
         buf.write_all(&[CR, LF])?;
         buf.write_all(b"Connection: keep-alive")?;
         buf.write_all(&[CR, LF])?;
+
+        self.write_allow_header(buf)?;
 
         if self.content_length != 0 {
             buf.write_all(Header::ContentType.raw())?;
@@ -171,6 +198,16 @@ impl Response {
         self.headers.set_server(server);
     }
 
+    /// Sets the HTTP allowed methods.
+    pub fn set_allow(&mut self, methods: Vec<Method>) {
+        self.headers.allow = methods;
+    }
+
+    /// Allows a specific HTTP method.
+    pub fn allow_method(&mut self, method: Method) {
+        self.headers.allow.push(method);
+    }
+
     fn write_body<T: Write>(&self, mut buf: T) -> Result<(), WriteError> {
         if let Some(ref body) = self.body {
             buf.write_all(body.raw())?;
@@ -215,6 +252,11 @@ impl Response {
     pub fn http_version(&self) -> Version {
         self.status_line.http_version
     }
+
+    /// Returns the allowed HTTP methods.
+    pub fn allow(&self) -> Vec<Method> {
+        self.headers.allow.clone()
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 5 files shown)


src/mmds/src/lib.rs
```diff
@@ -254,6 +259,33 @@ mod tests {
...
+        assert!(expected_response.body().unwrap() == actual_response.body().unwrap());
+
+        // Test `PATCH`. Supported by `micro-http`, but not supported by MMDS.
+        let request = b"PATCH http://169.254.169.254/ HTTP/1.0\r\n\r\n";
+        let mut expected_response = Response::new(Version::Http10, StatusCode::NotImplemented);
+        expected_response.set_body(Body::new("Unsupported HTTP method.".to_string()));
+        let actual_response = parse_request(request);
+        assert!(expected_response.status() == actual_response.status());
+        assert!(expected_response.body().unwrap() == actual_response.body().unwrap());
+    }
```
raduiliescu, line 287: Please also add a test with DELETE and HEAD, and if possible send an invalid HTTP method like "let request = b"NOMETHOD http:////169.254.169.254/ HTTP/1.0\r\n\r\n"

src/mmds/src/lib.rs
```diff
@@ -69,6 +69,15 @@ pub fn parse_request(request_bytes: &[u8]) -> Response {
                 );
             }
 
+            let method = request.method();
+            if method != Method::Get {
+                return build_response(
+                    request.http_version(),
+                    StatusCode::NotImplemented,
```
andreeaflorescu, line 76: We should be using the status code 405 Method Not Allowed. In micro-http it is okay to return the status code `NotImplemented` for POST or other methods that are actually not implemented. In MMDS we actually do not support other methods besides GET, so the status should reflect this.

src/mmds/src/lib.rs
```diff
@@ -254,6 +259,33 @@ mod tests {
...
 
+    #[test]
+    fn test_unsupported_http_method() {
+        // Test invalid HTTP POST method. This is not supported by `micro-http`.
+        let request = b"POST http://169.254.169.254/ HTTP/1.0\r\n\r\n";
+        let mut expected_response = Response::new(Version::Http10, StatusCode::NotImplemented);
+        expected_response.set_body(Body::new("Unsupported HTTP method.".to_string()));
+        let actual_response = parse_request(request);
+        assert!(expected_response.status() == actual_response.status());
+        assert!(expected_response.body().unwrap() == actual_response.body().unwrap());
```
andreeaflorescu, line 270: It looks like this code is duplicated several times where the only difference is the method. Can you make a vector of methods instead and iterate over them in a for?
```rust
let unsupported_methods = vec!["PATCH", "PUT", ....];
let mut expected_response = Response::new(Version::Http10, StatusCode::NotImplemented);
expected_response.set_body(Body::new("Unsupported HTTP method.".to_string()));
for method in unsupported_methods {
        let request = b"{} http://169.254.169.254/ HTTP/1.0\r\n\r\n".format(method);
        let actual_response = parse_request(request);
        assert!(expected_response.status() == actual_response.status());
        assert!(expected_response.body().unwrap() == actual_response.body().unwrap());
}```

File: src/mmds/src/lib.rs (hunk 2 above)
Comment (sandreim) [reply src/mmds/src/lib.rs:76]: If we return 405 then we must also include an `Allow` header field in the response with a list of the currently supported methods.

Comment (raduiliescu) [reply src/mmds/src/lib.rs:76]: Why complicate? I would return bad request, because this is what it is, a bad request.

Comment (iulianbarbu) [reply src/mmds/src/lib.rs:76]: > The server MUST generate an Allow header field in a 405 response containing a list of the target resource's currently supported methods.

It's somehow more to this. `micro-http` supports only `GET`, `PATCH` and `PUT`. This would mean that MMDS would understand these three, but do not allow `PATCH` and `PUT`. So `Method not allowed` would be a good fit here. However, there are other HTTP methods which are not supported by `micro-http`, hence by MMDS as well, so to be 100% accurate, we would have to respond from within MMDS with `Not implemented` status for `DELETE`, `HEAD`, `POST`, `UPDATE`.

I think it is worth to make the distinction that this part here, where we create a response based on a request, is the actual HTTP server logic, which is reachable from within the guest, and we can choose our own response logic/content, depending on how the server perceives the request. It doesn't have to give insight into the support or the limitations of the `micro-http` crate, as far as I can tell, but I might be wrong.

I see several options here:
* choose `NotImplemented` status, which is a valid response status even if we understand some HTTP methods which are not allowed.
* choose `Method not allowed` status, which fits the best for this use case, and add support in `micro-http` as well.
* choose `Bad Request` status, which can bring the needed information to user, to understand that the method is not allowed. This can be seen as an client error, in that the request is deceptive, given the contract for MMDS guest applications requests.
* choose a combination of `NotImplement` and `Method not allowed`, with the considerations given at the beginning of the comment.

If am fine with all these options, but if it was my choice, I would simplify, and make it work with what we currently have, meaning `NotImplemented` or `Bad Request`. Please, let me know what you think.

Comment (sandreim) [reply src/mmds/src/lib.rs:76]: 400 has different semantics: the api server cannot process the request due to a an error on the client side, for example malformed request syntax.

Comment (iulianbarbu) [reply src/mmds/src/lib.rs:76]: I agree, it's not complicated. 

Other than this, where do we make the separation between HTTP methods which are not allowed, and functionality which is not yet implemented? Are these methods really not allowed? Why shouldn't be these methods allowed?

I am thinking that in the future there might come a need to support Firecracker MMDS user data. This would mean to support few HTTP methods like `PUT mmds_ipv4_addr/user-data/resource` and `PATCH mmds_ipv4_addr/user-data/resource`. However, should we simply assume in terms of allowing or not allowing these methods, or in terms of supporting functionality which would require these methods?

Please, let me know what you think.

Comment (iulianbarbu) [reply src/mmds/src/lib.rs:76]: Synced offline with @sandreim . The scenario I've described can introduce many security holes, and it is not how IMDS works when providing user data at instance runtime. User-data has to be configured/modified when the instance is stopped, or before launching it. This means that the response `Method not allowed` status is necessary for other HTTP supported methods than `GET`.

File: src/mmds/src/lib.rs (hunk 1 above)
Comment (iulianbarbu) [reply src/mmds/src/lib.rs:324]: I've added tests for all HTTP methods and a presumably invalid HTTP method.

**Comment by iulianbarbu** [comment]
@sandreim, @andreeaflorescu , @raduiliescu , please take a look again! I've added Method not allow micro-http support and used it in MMDS.

File: src/micro_http/src/response.rs [hunk 4]
```diff
@@ -98,6 +105,20 @@ impl ResponseHeaders {
...
+            for (idx, method) in self.allow.iter().enumerate() {
+                buf.write_all(method.raw())?;
+                if idx < self.allow.len() - 1 {
+                    buf.write_all(delimitator)?;
+                }
+            }
+
+            buf.write_all(&[CR, LF])?;
+        }
+
```
sandreim, line 121: nit: Keep this fn small and readable - move this code inside a `write_allow()` or something else.

src/mmds/src/lib.rs
```diff
@@ -78,6 +81,16 @@ pub fn parse_request(request_bytes: &[u8]) -> Response {
                 );
             }
 
+            let method = request.method();
+            if method != Method::Get {
```
sandreim, line 85:
```suggestion
            if request.method() != Method::Get {
```

src/mmds/src/lib.rs
```diff
@@ -78,6 +81,16 @@ pub fn parse_request(request_bytes: &[u8]) -> Response {
...
+                return build_response(
+                    request.http_version(),
+                    StatusCode::MethodNotAllowed,
+                    // Safe to unwrap, the method `bytes` are UTF-8 valid.
```
sandreim, line 89: Where is the unwrap ? 👓

iulianbarbu:
@raduiliescu , sorry, had some rust clippy issues. Had to force push again. @sandreim , please take a look again.

iulianbarbu:
Fixes #1820.

Issue #1798, opened 2020-04-16
Implement Snapshotting specific APIs


lauralt:

lauralt:
Fixed by #1821.

PR #1537, opened 2020-01-21
[RFC] Snapshotting design proposal

acatangiu:
## Firecracker snapshots

This PR describes the overall snapshotting proposal.

Fixes #1395

PR #1821, opened 2020-04-24
Snapshot API calls

lauralt:
## Reason for This PR

Fixes: #1796, #1798.

## Description of Changes

This is a proposal for the Snapshot API calls through which I tried to keep the API as RESTful as it can be with the newly added actions. I will update the integration tests with the snapshotting resources once we're okay with certain paths / calls.

## License Acceptance

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@
 - Added `track_dirty_pages` field to `machine-config`. If enabled, Firecracker
   can create incremental guest memory snapshots by saving the dirty guest pages
   in a sparse file.
+- Added a new API call, `PATCH /vm`, for changing the microVM state (to
+  `Paused` or `Resumed`).
+- Added a new API call, `PUT /snapshot/create`, for creating a full or diff
+  snapshot.
+- Added a new API call, `PUT /snapshot/load`, for loading a snapshot.
 
 ### Fixed
 - Added `--version` flag to both Firecracker and Jailer.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,10 @@
   `MMDS` with a custom valid link-local IPv4 address.
 - Added experimental JSON response format support for MMDS guest applications
   requests.
-   
+- Added `track_dirty_pages` field to `machine-config`. If enabled, Firecracker
+  can create incremental guest memory snapshots by saving the dirty guest pages
+  in a sparse file.
+
 ### Fixed
 - Added `--version` flag to both Firecracker and Jailer.
 - Return `405 Method Not Allowed` MMDS response for non HTTP `GET` MMDS
--- a/src/api_server/src/request/snapshot.rs
+++ b/src/api_server/src/request/snapshot.rs
@@ -1,0 +1,160 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use super::super::VmmAction;
+use request::{Body, Error, ParsedRequest, StatusCode};
+use vmm::vmm_config::snapshot::{CreateSnapshotParams, LoadSnapshotParams, Vm, VmState};
+use Method;
+
+pub fn parse_put_snapshot(
+    body: &Body,
+    request_type_from_path: Option<&&str>,
+) -> Result<ParsedRequest, Error> {
+    match request_type_from_path {
+        Some(&request_type) => match request_type {
+            "create" => Ok(ParsedRequest::Sync(VmmAction::CreateSnapshot(
+                serde_json::from_slice::<CreateSnapshotParams>(body.raw())
+                    .map_err(Error::SerdeJson)?,
+            ))),
+            "load" => Ok(ParsedRequest::Sync(VmmAction::LoadSnapshot(
+                serde_json::from_slice::<LoadSnapshotParams>(body.raw())
+                    .map_err(Error::SerdeJson)?,
+            ))),
+            _ => Err(Error::InvalidPathMethod(
+                format!("/snapshot/{}", request_type),
+                Method::Put,
+            )),
+        },
+        None => Err(Error::Generic(
+            StatusCode::BadRequest,
+            "Missing snapshot operation type.".to_string(),
+        )),
+    }
+}
+
+pub fn parse_patch_vm_state(body: &Body) -> Result<ParsedRequest, Error> {
+    let vm = serde_json::from_slice::<Vm>(body.raw()).map_err(Error::SerdeJson)?;
+
+    match vm.state {
+        VmState::Paused => Ok(ParsedRequest::Sync(VmmAction::Pause)),
+        VmState::Resumed => Ok(ParsedRequest::Sync(VmmAction::Resume)),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::path::PathBuf;
+
+    use super::*;
+    use vmm::vmm_config::snapshot::SnapshotType;
+
+    #[test]
+    fn test_parse_put_snapshot() {
+        let mut body = r#"{
+                "snapshot_type": "Diff",
+                "snapshot_path": "foo",
+                "mem_file_path": "bar",
+                "version": 2
+              }"#;
+
+        let mut expected_cfg = CreateSnapshotParams {
+            snapshot_type: SnapshotType::Diff,
+            snapshot_path: PathBuf::from("foo"),
+            mem_file_path: PathBuf::from("bar"),
+            version: Some(2),
+        };
+
+        match parse_put_snapshot(&Body::new(body), Some(&"create")) {
+            Ok(ParsedRequest::Sync(VmmAction::CreateSnapshot(cfg))) => {
+                assert_eq!(cfg, expected_cfg)
+            }
+            _ => panic!("Test failed."),
+        }
+
+        body = r#"{
+                "snapshot_path": "foo",
+                "mem_file_path": "bar"
+              }"#;
+
+        expected_cfg = CreateSnapshotParams {
+            snapshot_type: SnapshotType::Full,
+            snapshot_path: PathBuf::from("foo"),
+            mem_file_path: PathBuf::from("bar"),
+            version: None,
+        };
+
+        match parse_put_snapshot(&Body::new(body), Some(&"create")) {
+            Ok(ParsedRequest::Sync(VmmAction::CreateSnapshot(cfg))) => {
+                assert_eq!(cfg, expected_cfg)
+            }
+            _ => panic!("Test failed."),
+        }
+
+        let invalid_body = r#"{
+                "invalid_field": "foo",
+                "mem_file_path": "bar"
+              }"#;
+
+        assert!(parse_put_snapshot(&Body::new(invalid_body), Some(&"create")).is_err());
+
+        body = r#"{
+                "snapshot_path": "foo",
+                "mem_file_path": "bar"
+              }"#;
+
+        let mut expected_cfg = LoadSnapshotParams {
+            snapshot_path: PathBuf::from("foo"),
+            mem_file_path: PathBuf::from("bar"),
+            enable_diff_snapshots: false,
+        };
+        match parse_put_snapshot(&Body::new(body), Some(&"load")) {
+            Ok(ParsedRequest::Sync(VmmAction::LoadSnapshot(cfg))) => assert_eq!(cfg, expected_cfg),
+            _ => panic!("Test failed."),
+        }
+
+        body = r#"{
+                "snapshot_path": "foo",
+                "mem_file_path": "bar",
+                "enable_diff_snapshots": true
+              }"#;
+
+        expected_cfg = LoadSnapshotParams {
+            snapshot_path: PathBuf::from("foo"),
+            mem_file_path: PathBuf::from("bar"),
+            enable_diff_snapshots: true,
+        };
+
+        match parse_put_snapshot(&Body::new(body), Some(&"load")) {
+            Ok(ParsedRequest::Sync(VmmAction::LoadSnapshot(cfg))) => assert_eq!(cfg, expected_cfg),
+            _ => panic!("Test failed."),
+        }
+
+        assert!(parse_put_snapshot(&Body::new(body), Some(&"invalid")).is_err());
+        assert!(parse_put_snapshot(&Body::new(body), None).is_err());
+    }
+
+    #[test]
+    fn test_parse_patch_vm_state() {
+        let mut body = r#"{
+                "state": "Paused"
+              }"#;
+
+        assert!(parse_patch_vm_state(&Body::new(body))
+            .unwrap()
+            .eq(&ParsedRequest::Sync(VmmAction::Pause)));
+
+        body = r#"{
+                "state": "Resumed"
+              }"#;
+
+        assert!(parse_patch_vm_state(&Body::new(body))
+            .unwrap()
+            .eq(&ParsedRequest::Sync(VmmAction::Resume)));
+
+        let invalid_body = r#"{
+                "invalid": "Paused"
+              }"#;
+
+        assert!(parse_patch_vm_state(&Body::new(invalid_body)).is_err());
+    }
+}
--- a/src/api_server/swagger/firecracker.yaml
+++ b/src/api_server/swagger/firecracker.yaml
@@ -399,6 +399,84 @@ paths:
           schema:
             $ref: "#/definitions/Error"
 
+  /snapshot/create:
+    put:
+      summary: Creates a full or diff snapshot. Post-boot only.
+      description:
+        Creates a snapshot of the microVM state. The microVM should be
+        in the `Paused` state.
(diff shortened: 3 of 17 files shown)


src/api_server/swagger/firecracker.yaml
```diff
@@ -699,6 +811,19 @@ definitions:
...
+    description:
+      Defines the microVM state. It is especially useful in the snapshotting context.
+    required:
+      - state
+    properties:
+      state:
+        type: string
+        enum:
+          - Paused
+          - Resumed
```
acatangiu, line 825: This is the same design as `/actions` where we have an enum stating the action to do. `pause` and `resume` can go there to keep things consistent.

The reason https://github.com/firecracker-microvm/firecracker/pull/1537 avoids using `actions` is that the design there is not future-proof/flexible. It does not naturally support any _actions_ that have a body (extra data like identifiers or payload or such).
You can check the infamous `BlockDeviceRescan` action in past Firecracker versions code, which made the code a mess by introducing a `payload` next to the enum which had to be validated and interpreted differently depending on the enum variant (even for other actions which hadn't been aware of payloads).

acatangiu: Or if we feel it's more RESTful, all `actions` can go under `/vm`

src/api_server/swagger/firecracker.yaml
```diff
@@ -699,6 +811,19 @@ definitions:
         description: The amount of milliseconds it takes for the bucket to refill.
         minimum: 0
 
+  Vm:
```
acatangiu, line 814: Please add a commit on top of https://github.com/firecracker-microvm/firecracker/pull/1537 to update that proposal with the details in this PR.

CHANGELOG.md
```diff
@@ -15,6 +15,14 @@
 - Added a [guide](docs/devctr-image.md) for updating the dev container image.
 - Added a new API call, `PUT /mmds/config`, for configuring the
   `MMDS` with a custom valid link-local IPv4 address.
+- Added `track_dirty_pages` field to `machine-config`, which specifies
+  whether dirty pages should be tracked and which allows incremental
+  snapshots to be created.
```
sandreim, line 20:
```suggestion
- Added `track_dirty_pages` field to `machine-config`. If enabled Firecracker can create
incremental guest memory snapshots by saving the dirty guest pages in a sparse file.
```

CHANGELOG.md
```diff
@@ -15,6 +15,14 @@
...
+- Added a new API call, `PATCH /vm`, for changing the microVM state (to
+  `Paused` or `Resumed`).
+- Added a new API call, `PUT /snapshot/create`, for creating a full or diff
```
sandreim, line 23: The [design proposal](https://github.com/firecracker-microvm/firecracker/blob/9d427b33d989c3225d874210f6c2849465941dc0/docs/snapshotting/design.md) introduces separate APIs for creating incremental and full snapshots. Are you proposing a single endpoint ?

lauralt: Yes. Because both `/create` and `/create-diff` would have the same request body, I was thinking to make that distinction via another field in the body and reduce this way the swagger file size (and probably avoid also some duplicate code).

sandreim: One of the reasons they are separated was to provide a clear distinction eliminating the need to have a parameter for that. This is also future looking, for example if in the future you would add another parameter that only makes sense for `create` it would create some extra complexity by requiring extra conditionals as well as json ser/de issues for `create-diff` requests.

lauralt: Are you suggesting that, when `create` and `create-diff` will diverge (from the request bodies point of view) we will add a new `SnapshotCreateConfig`, `SnapshotCreateConfigDiff` for example, or that we should make this distinction at this moment, i.e. have 2 different structs with the same fields? From what I understand from the design proposal RFC, these requests have the same body, so I supposed they will also use the same struct for ser/de.

sandreim: The main concern here is extensibility and complexity. From my point of view things are much more simpler whit as few parameters to configure as possible.

lauralt: What I was trying to say is that when those requests will have different bodies, we will also have to add some new `SnapshotCreateConfig` struct besides the existing one (or more precisely, we will rename the existing one to something like `SnapshotCreateConfigDiff` and create a new one with all the fields required for the updated `create` action). This also means some extra code. I am okay with having 2 different APIs for creating incremental and full snapshots, I just wanted to make sure that at this moment, in this PR, the struct that will be used for ser/de for both requests should be the same and not have a duplicate (`SnapshotCreateConfig` and `SnapshotCreateConfigDiff` - both with the same fields).

sandreim: > I just wanted to make sure that at this moment, in this PR, the struct that will be used for ser/de for both requests should be the same and not have a duplicate (`SnapshotCreateConfig` and `SnapshotCreateConfigDiff` - both with the same fields).

Definetly only one: SnapshotCreateConfig.

andreeaflorescu: I think a diff snapshot is a special case of a snapshot. Since the `load` API only takes full snapshots, does it make sense to have the default of `snapshot/create` create a full snapshot as well? This way the two operations are by default mirrored in a sense.

I have a few comments on naming as well.
There is no need to name them `..Config`. Besides Mmds (which is named wrong, ooopsi), all the other objects are named without that sufix. I see no reason to keep the suffix for snapshot.
Nevertheless, naming will be a bit awkward for the resource because `Snapshot` could be seen as a resource, while `snapshot/create` and `snapshot/load` cannot. To make it less awkward from the REST API point of view, can we use the same body for both requests instead? I would go as far as having the Snapshot resource (on `/snapshot`) and then an `action` field enum with the values load and create. We can make sure in the code that for load we ignore everything else that is not the memory path and the state path.

```swagger
Snapshot:
  type: object
  required:
    - snapshot_path
    - mem_file_path
    - action
  properties:
    version:
      type: integer
      minimum: TBD
      maximum: TBD
      description: Optional field for snapshot format version. If omitted
                   Firecracker will default to creating a snapshot with the
                   current version.
    snapshot_path:
      type: string
      description: Path to file which will be populated with the vmm state.
    mem_file_path:
      type: string
      description: Path to file which will be populated with the guest memory.
                   The file will contain a copy of the guest memory or copies
                   of just the dirtied pages, depending on which URL this
                   object is used.
    type: string
         enum: [full, diff]
         description: Type of snapshot to create. By default a full snapshot is created. This field
          only makes sense for the create operation and is ignored otherwise.
    action: string
        enum: [load, create]
        description: Action to take on the snapshot resource.
    support_inc_snapshots: bool
        description: Support incremental snapshots. This only makes sense for load operations and is ignored otherwise.
```

To create a snapshot:

1. Full snapshot:  
PUT /snapshot {action: "create", snapshot_file: "/blah/1231", "mem_file_path": "/blah/1"}

2. Diff snapshot:
PUT /snapshot {action: "create", type: "diff", snapshot_file: "/blah/1231", "mem_file_path": "/blah/1"}

3. Snapshot at a certain version:
PUT /snapshot {action: "create", version: 12, snapshot_file: "/blah/1231", "mem_file_path": "/blah/1"}

To load a snapshot:
PUT /snapshot {action: "load", snapshot_file: "/blah/1231", "mem_file_path": "/blah/1"}

I don't know if having actions is a common problem when designing restful APIs, but there are definitely other people that struggled with the same problem as we have now :)). Having the `action` field is not uncommon.

Whatever path forward we take here (either action in path or action in body), we should find some better names for `SnapshotCreate` and `SnapshotLoad` as these are rather operations than nouns. These names are going to be used by everyone using this functionality (and generating clients from our swagger), so we should try to make them less confusing.

andreeaflorescu: Idea no 2 which should be less intrusive is to rename `SnapshotCreate` and `SnapshotLoad` to `CreateSnapshotParams` and `LoadSnapshotParams`.

CHANGELOG.md
lauralt: Done.

CHANGELOG.md
lauralt: Updated with `CreateSnapshotParams` and `LoadSnapshotParams`.

src/api_server/swagger/firecracker.yaml
```diff
@@ -665,6 +743,43 @@ definitions:
...
+        - snapshot_path
+        - mem_file_path
+      properties:
+        snapshot_path:
+          type: string
+          description: Path to the file that contains the microVM state to be loaded.
+        mem_file_path:
+          type: string
+          description: Path to the file that contains the guest memory to be loaded.
+        support_inc_snapshots:
```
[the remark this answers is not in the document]
lauralt: Done.

src/api_server/swagger/firecracker.yaml
```diff
@@ -577,6 +652,9 @@ definitions:
         description: Flag for enabling/disabling Hyperthreading
       cpu_template:
         $ref: "#/definitions/CpuTemplate"
+      track_dirty_pages:
+        type: boolean
+        description: Enable dirty page tracking. If this is enabled, then incremental VM snapshots can be created.
```
sandreim, line 657: I think the description can be improved by differentiating between full/diff snapshot - sparse file.

src/api_server/swagger/firecracker.yaml
```diff
@@ -399,6 +399,81 @@ paths:
           schema:
             $ref: "#/definitions/Error"
 
+  /snapshot/create:
+    put:
+      summary: Creates a full or diff snapshot. Post-boot only.
```
sandreim, line 404: microVM needs to be in the Paused state.

src/api_server/swagger/firecracker.yaml
```diff
@@ -399,6 +399,81 @@ paths:
...
+        default:
+          description: Internal server error
+          schema:
+            $ref: "#/definitions/Error"
+
+  /vm:
+    patch:
+      summary: Updates the microVM state.
+      description:
+        Sets the desired state for the microVM.
```
sandreim, line 456: Please make it more explicit: Pause/Resume

src/api_server/swagger/firecracker.yaml
```diff
@@ -668,6 +746,46 @@ definitions:
...
+        description: Path to the file that will contain the microVM state.
+      mem_file_path:
+        type: string
+        description: Path to the file that will contain the guest memory.
+      snapshot_type: string
+        enum:
+          - Full
+          - Diff
+        description:
+          Type of snapshot to create. By default, a full snapshot is created.
```
sandreim, line 766: Is this parameter optional ?

src/api_server/swagger/firecracker.yaml
```diff
@@ -668,6 +746,46 @@ definitions:
...
+        - snapshot_path
+        - mem_file_path
+      properties:
+        snapshot_path:
+          type: string
+          description: Path to the file that contains the microVM state to be loaded.
+        mem_file_path:
+          type: string
+          description: Path to the file that contains the guest memory to be loaded.
+        support_inc_snapshots:
```
sandreim, line 783: Just to be consistent with the above enum I suggest:
```suggestion
        enable_diff_snapshots:
```

src/api_server/swagger/firecracker.yaml
```diff
@@ -668,6 +746,46 @@ definitions:
...
+          type: bool
+          description:
+            Marks if the resulting microVM should track dirty pages, in order to be
+            snapshotted again later.
```
sandreim, line 787:
```suggestion
            Enable support for incremental(diff) snapshots by tracking dirty guest pages.
```

src/api_server/swagger/firecracker.yaml
```diff
@@ -699,6 +817,19 @@ definitions:
         description: The amount of milliseconds it takes for the bucket to refill.
         minimum: 0
 
+  Vm:
+    type: object
+    description:
+      Defines the microVM state. It is especially useful in the snapshotting context.
```
sandreim, line 823: I guess it is worth saying here that this refers to the running state of the microVM, not to be confused with the microvm state file.

src/vmm/src/snapshot.rs
```diff
@@ -0,0 +1,70 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sandreim, line 1: Please move this file to src/vmm/src/vmm_config.rs.

src/api_server/swagger/firecracker.yaml
lauralt: Done, I added this in description to keep the summary short.

src/api_server/swagger/firecracker.yaml
lauralt: Done.

src/api_server/swagger/firecracker.yaml
lauralt: Done.

src/api_server/swagger/firecracker.yaml
lauralt: Yes, I updated the description.

src/api_server/swagger/firecracker.yaml
lauralt: Done.

src/api_server/swagger/firecracker.yaml
lauralt: Agree, done.

src/vmm/src/snapshot.rs
lauralt: I didn't want to add it there initially because I was thinking it will contain some functionality that won't fit in `vmm_config`. But in its current form, it seems better to move it there. Done.

src/api_server/src/request/snapshot.rs
```diff
@@ -0,0 +1,160 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0<Paste>
```
ioanachirca, line 2:
```suggestion
// SPDX-License-Identifier: Apache-2.0
```

lauralt: Done.

src/api_server/swagger/firecracker.yaml
```diff
@@ -668,6 +751,46 @@ definitions:
...
+      mem_file_path:
+        type: string
+        description: Path to the file that will contain the guest memory.
+      snapshot_type: string
+        enum:
+          - Full
+          - Diff
+        description:
+          Type of snapshot to create. It is optional and by default, a full
+          snapshot is created.
```
acatangiu, line 772: fyi: one of the topics of discussions with Lambda customers when we reviewed the API proposal was that all these knobs become hard to manage. That is why `/snapshot/create` and `/snapshot/create-diff` were ultimately different api endpoints (to simplify this config body).

In its current format you also need to document the behavior of `snapshot_type == Diff` in rapport with `enable_diff_snapshots` from a previous snapshot load or `enable_diff_snapshots` from vm-config in case this is a _fresh_ microvm.

src/api_server/swagger/firecracker.yaml
```diff
@@ -399,6 +399,82 @@ paths:
...
+        default:
+          description: Internal server error
+          schema:
+            $ref: "#/definitions/Error"
+
+  /snapshot/load:
+    put:
+      summary: Loads a snapshot. Pre-boot only.
+      description:
+        Loads the microVM state from a snapshot.
```
acatangiu, line 432: Only accepted on a fresh/unconfigured `firecracker` process.

src/api_server/swagger/firecracker.yaml
sandreim: This also brings up some extra checks required when validating the request.Does it make sense to make the simplifying assumption that fresh microvms cannot create incremental snapshots ?

src/api_server/swagger/firecracker.yaml
```diff
@@ -577,6 +653,13 @@ definitions:
         description: Flag for enabling/disabling Hyperthreading
       cpu_template:
         $ref: "#/definitions/CpuTemplate"
+      track_dirty_pages:
+        type: boolean
+        description:
+          Enable dirty page tracking. If this is enabled, then incremental guest memory
+          snapshots can be created. These are called diff snapshots and they are saving
+          in a sparse file the memory dirtied since the last snapshot, which can be a
+          full or diff one. A full snapshot contains a full copy of the guest memory.
```
acatangiu, line 662: nit:
```suggestion
          snapshots can be created. These are called diff snapshots containing only
          the memory dirtied since a previous snapshot. Full snapshots each contain a
          full copy of the guest memory.
```

src/api_server/swagger/firecracker.yaml
acatangiu: > Does it make sense to take the simplifying assumption that fresh microvms cannot create incremental snapshots ?

No...
- a `full` snapshot on a fresh microvm means a full memory file of size `==` to guest memory
- a `diff` snapshot on a fresh microvm means a sparse memory file containing only written pages.

PR #1832, opened 2020-04-28
Fix unit test coverage for the cpuid crate

ahmedaabouzied:
## Reason for This PR
Solve issue #1705 

## Description of Changes
Set the mod block to `mod tests` instead of `mod test` in the `cpuid` crate. This will make the test code in `cpuid` crate get excluded by kcov. If it's not excluded kcov counts the test code as code and then the test coverage result gets messed up.

- [x] This functionality can be added in [`rust-vmm`](https://github.com/rust-vmm).

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

- [x] All commits in this PR are signed (`git commit -s`).
- [x] The reason for this PR is clearly provided (issue no. or explanation).
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] Any newly added `unsafe` code is properly documented.
- [x] Any API changes are reflected in `firecracker/swagger.yaml`.
- [x] Any user-facing changes are mentioned in `CHANGELOG.md`.

Issue #1704, opened 2020-03-25
Add a mechanism to validate that unit tests are implemented in the`tests` mod block.


sandreim:
Kcov reports coverage only for tests defined in the `mod tests` block, so we need to enforce this to avoid bogus coverage results.
https://github.com/firecracker-microvm/firecracker/blob/e6436c599756b653931a57d5f5c418f7d22cac40/tests/integration_tests/build/test_coverage.py#L61

andreeaflorescu:
Just a tiny clarification: kcov reports the coverage for all the tests, but the problem is that the report that calculates the coverage is wrong without having the test section excluded because it is counting the test line as code lines as well.

tnotw:
Would this work?  An additional test in the test_coverage module that diffs the list of files containing the "#[cfg(test)]" attribute with the list of files containing the string "mod tests {".   
```
+def test_ensure_mod_tests():
+    """Fail if mod_tests hasn't been defined as the test module name."""
+    # List all source files containing rust test attribute.
+    # List all files containing string "mod tests {".
+    # Diff the two lists.
+    # Diff returns zero if file lists match, non-zero if different.
+    # Redirect file diff list to stderr.
+    cmd = (
+        '/bin/bash '
+        '-c '
+        '"diff '
+        '<(grep --extended-regexp '
+        '--files-with-matches '
+        '--recursive '
+        '\'^#\\[cfg\\(test\\)\\]$\' ../src) '
+        '<(grep '
+        '--files-with-matches '
+        '--recursive '
+        '\'mod tests {\' ../src)" 1>&2'
+    )
```
Giving: 
```[ unsafe | 1 tests | 1 worker(s) ]
  FAILED    integration_tests/build/test_coverage.py::test_ensure_mod_tests

========================================================================================== FAILURES ===========================================================================================
____________________________________________________________________________________ test_ensure_mod_tests ____________________________________________________________________________________
integration_tests/build/test_coverage.py:67: in test_ensure_mod_tests
    stdout=PIPE
/usr/lib/python3.6/subprocess.py:438: in run
    output=stdout, stderr=stderr)
E   subprocess.CalledProcessError: Command '/bin/bash -c "diff <(grep --extended-regexp --files-with-matches --recursive '^#\[cfg\(test\)\]$' ../src) <(grep --files-with-matches --recursive 'mod tests {' ../src)" 1>&2' returned non-zero exit status 1.
[...]

PR #1833, opened 2020-04-28
tests: add test to ensure rust test scope

tnotw:
Proposed mechanism to address #1704.  Add a new test to ensure that every rust source file that contains the rust test attribute,  also contains the string "mod tests {".  

Test should break the CI run until after #1832 is merged up,  where the failure error should list the non-conforming files. 

## Reason for This PR

## Description of Changes

## License Acceptance

## PR Checklist


dianpopa:
What will happen with this test for source files that have `cfg[test]` in it but do not have any unit test (i.e no test module) declared inside that source file?

tnotw:
I think it would still fail, and flag the file.  Wouldn't that be ok because any tests not in a module called "mod tests" would break the kcov coverage report anyway?  So you might want that to be flagged?  

But yes, this test is very specific to how the current codebase looks to me,  rather than accommodating any possible coding variations.  Appreciate that that might not be the requirement.

dianpopa:
> I think it would still fail, and flag the file. Wouldn't that be ok because any tests not in a module called "mod tests" would break the kcov coverage report anyway? So you might want that to be flagged?
> 
> But yes, this test is very specific to how the current codebase looks to me, rather than accommodating any possible coding variations. Appreciate that that might not be the requirement.

Ok, so this PR is tying the presence of a `cfg[test]` anywhere in the source file with the presence of a test module inside the same file. As far as my concern, this seems like a logical dependency. One would not bother to put `cfg[test]` in a source file unless one were trying to write unit tests for that code, right?
The PR looks good to me otherwise! Waiting for #1832 to get merged.

tnotw:
Yeah,  that's what I was going for.  I think the logic holds, but I'm not certain how future proof it is.   I guess it's pretty disposable if the shape of the tests change in any way.

lauralt:
> Yeah, that's what I was going for. I think the logic holds, but I'm not certain how future proof it is. I guess it's pretty disposable if the shape of the tests change in any way.

We started making efforts in writing rust integration tests, especially for the crates that have long source files or can not be properly tested by unit tests. I think this will be a future direction for our testing system. These tests do not require `cfg[test]` or `mod tests`, so from this point of view, the proposed changes should be ok.

lauralt:
Still, I don't think we should rely on the fact that every file that contains a `#[cfg(test)]` will have a `mod tests` defined. I remembered that at some moment we had a `#[cfg(test)]` for some `macro_use` in a file with no unit tests; see the commit here: https://github.com/firecracker-microvm/firecracker/commit/324e3cc690e8575a3cc0ccdf3f9d5e34a8c73903. In that particular situation, later on it was proven to be an alternative for it, but I don't think we should enforce it by this test.
Maybe a better option would be to search for that diff in files that have also a `#[test]` in them, but it's really ugly :disappointed:.

tnotw:
Ok,  so if this proposed test had been run against the commit you mention,  it would have flagged up the file containing `#[cfg(test)]` without the corresponding `mod tests {`. But might that be correct from a kcov perspective?  i.e. code flagged `#[cfg(test)]` in the codebase that isn't being excluded from the kcov report?  Appreciate it's not desirable from a code perspective. 

If it did flag, adjusting the `--exclude-region` option in [kcov](https://manpages.debian.org/unstable/kcov/kcov.1.en.html) to include regions marked `#[cfg(test)]` might be possible?   Something like: 
```
--exclude-region 'mod tests {', #[cfg(test)]:#[cfg(not(test))]
```
Then adjusting the proposed test here to accommodate both scenarios,  i.e. all files containing `#[cfg(test)]` would have to either contain a `mod test {`  string or a closing `#[cfg(not(test))]` 

Or maybe having the exclusion tied to `mod tests {` is actually imprecise and really the problem.  Should it really be getting tied to `#[cfg(test)]` exclusively?  Then we don't care about module name typos like `mod test {`.   Not sure really, I guess it is like this for a reason.

But yeah, back to the drawing board, I think.

lauralt:
> Ok, so if this proposed test had been run against the commit you mention, it would have flagged up the file containing `#[cfg(test)]` without the corresponding `mod tests {`. But might that be correct from a kcov perspective? i.e. code flagged `#[cfg(test)]` in the codebase that isn't being excluded from the kcov report? Appreciate it's not desirable from a code perspective.
> 
For the aforementioned situation, no because kcov excludes in its report the imports and `use`s region (This is what I was talking about: https://github.com/lauralt/firecracker/blob/f13717a4e3d928a64d313263e7bca2e9fc77a218/src/logger/src/lib.rs#L8). But your point is valid when it comes to functions that are marked with `#[cfg(test)]` and are not defined in `mod tests`. We are trying to avoid these situations, but sometimes people prefer to define them there because it probably seems easier/nicer to import them from there than from a tests module. I took a look in our codebase and it seems like we have a few places with test functions that are not defined in a test module and can be easily moved, I'll open an issue for that.

> If it did flag, adjusting the `--exclude-region` option in [kcov](https://manpages.debian.org/unstable/kcov/kcov.1.en.html) to include regions marked `#[cfg(test)]` might be possible? Something like:
> 
> ```
> --exclude-region 'mod tests {', #[cfg(test)]:#[cfg(not(test))]
> ```
> 
> Then adjusting the proposed test here to accommodate both scenarios, i.e. all files containing `#[cfg(test)]` would have to either contain a `mod test {` string or a closing `#[cfg(not(test))]`
> 
It would be really nice to do that but unfortunately we don't usually mark the end of a `#[cfg(test)]` section with a `#[cfg(not(test))]`. As suggested here: https://stackoverflow.com/questions/42440192/how-do-i-exclude-test-functions-from-code-coverage-when-using-kcov, we could enforce people to only add this kind of functions at the end of the non test module. But we can also have `#[cfg(test)]` for some statements or fields, and again, `#[cfg(not(test))]` wouldn't be required. So it looks like we can't add this kind of restriction.

> Or maybe having the exclusion tied to `mod tests {` is actually imprecise and really the problem. Should it really be getting tied to `#[cfg(test)]` exclusively? Then we don't care about module name typos like `mod test {`. Not sure really, I guess it is like this for a reason.
> 
We could name test modules whatever we want, but then we would lose consistency and it would be pretty weird I guess to have tests in all sorts of modules (also would be hard to tell kcov which sections to exclude).

> But yeah, back to the drawing board, I think.

I gave some more thought to this and, at this moment, the cleanest solution seems to be the following:
- search for `mod tests {` only in the files that contain a `#[test]` in them because in the end this is a clear sign that we have tests in that file (as opposed to `#[cfg(test)]`).
- make sure to do this search only in the `src` dirs of crates, not in the `tests` ones because the last ones contain integration tests (which are also marked with `#[test]` attribute) and we shouldn't care about them for the scope of this PR (also kcov excludes these dirs in its report).

What do you think?

tnotw:
Hi @lauralt,  sounds good to me, thank you very much for the clarification and explanation.  

So I've amended as follows:  

```
@ -49,11 +51,13 @@ def test_ensure_mod_tests():
         '<(grep --extended-regexp '
         '--files-with-matches '
         '--recursive '
-        '\'^#\\[cfg\\(test\\)\\]$\' ../src) '
+        '--exclude-dir=src/*_gen/* ' 
+        '\'#\\[test\\]\' ../src/*/src) '
         '<(grep '
         '--files-with-matches '
         '--recursive '
-        '\'mod tests {\' ../src)"'
+        '--exclude-dir=src/*_gen/* '
+        '\'mod tests {\' ../src/*/src)"'
```
With the additional step of excluding the generated files,  which I assume we don't want flagged.

The only file that is being flagged after these changes is this one:  `src/devices/src/virtio/vsock/mod.rs` which appears to have a "mod tests {" but no tests defined.  I'm guessing we shouldn't really be flagging this, and can probably remove it by anchoring the 'mod tests {' pattern to the start of line.  But it's beyond my rust skills to say definitively.

lauralt:
> Hi @lauralt, sounds good to me, thank you very much for the clarification and explanation.
> 
> So I've amended as follows:
> 
> ```
> @ -49,11 +51,13 @@ def test_ensure_mod_tests():
>          '<(grep --extended-regexp '
>          '--files-with-matches '
>          '--recursive '
> -        '\'^#\\[cfg\\(test\\)\\]$\' ../src) '
> +        '--exclude-dir=src/*_gen/* ' 
> +        '\'#\\[test\\]\' ../src/*/src) '
>          '<(grep '
>          '--files-with-matches '
>          '--recursive '
> -        '\'mod tests {\' ../src)"'
> +        '--exclude-dir=src/*_gen/* '
> +        '\'mod tests {\' ../src/*/src)"'
> ```
> 
> With the additional step of excluding the generated files, which I assume we don't want flagged.
> 
> The only file that is being flagged after these changes is this one: `src/devices/src/virtio/vsock/mod.rs` which appears to have a "mod tests {" but no tests defined. I'm guessing we shouldn't really be flagging this, and can probably remove it by anchoring the 'mod tests {' pattern to the start of line. But it's beyond my rust skills to say definitively.

Hmm, I kind of hoped that we won't / don't have `mod tests {` without unit tests :( :(. But we should allow that kind of "special" test modules, i.e. helpers for the modules that actually contain tests. If you're asking me, that module could be renamed to `test_helper` or something like that without any big concern (except the fact that kcov won't exclude it, but it will be an isolated case anyway).
That's why I proposed to search for `mod tests {` only in the files where `#[test]` is present; this way you won't even care about the file that you mentioned. I think this PR should stick to just one purpose: check that unit tests are in a `mod tests {`, don't check also that a `mod tests {` has at least one unit test. From the `kcov` perspective I agree it's not the best thing to have a `mod <name_different_that_tests> {` marked with `#[cfg(test)]`, that doesn't contain any test, not excluded from the report, but I assume the number of this kind of modules will be very very small).
The solution that you propose is indeed elegant, but it kind of does 2 things at a time. I would like the test in the end to tell me something like this if it fails: `you have unit test(s) in a module that is not called `tests`. Please check `blah blah` file.`.

tnotw:
Apologies, a few, ahem, reading and comprehension challenges at my end.  :-)

I've had another shot, hopefully it's closer to what you had in mind.

tests/integration_tests/build/test_coverage.py
```diff
@@ -32,6 +32,44 @@
...
+        '\'mod tests {\' '
+        '\\$(grep '
+        '--files-with-matches '
+        '--recursive '
+        '--exclude-dir=src/*_gen/* '
+        '\'\\#\\[test\\]\' ../src/*/src)" '
+    )
+
+    # the outer grep returns 0 if it finds files without the match, so we
+    # ignore the return codes.
```
lauralt, line 59:
```suggestion
    # The outer grep returns 0 even if it finds files without the match, so we
    # ignore the return codes.
```

tests/integration_tests/build/test_coverage.py
```diff
@@ -32,6 +32,44 @@
...
+    result = utils.run_cmd(cmd, no_shell=False, ignore_return_code=True)
+
+    ensure_mod_tests_msg = (
```
lauralt, line 62: I would rename this variable to `error_msg`.

tests/integration_tests/build/test_coverage.py
```diff
@@ -32,6 +32,44 @@
...
+        'Tests found in files without a "tests" module:\n {}'
+        'To ensure code coverage is reported correctly, please check that '
+        'your tests are in a module named "tests".'.format(result.stdout)
+    )
+
+    # stdout should be empty, where empty strings are false
+    assert not result.stdout, \
+        ensure_mod_tests_msg
```
lauralt, line 70:
```suggestion
    assert not result.stdout, error_msg
```

tests/integration_tests/build/test_coverage.py
```diff
@@ -32,6 +32,44 @@
 """Regex for extracting number of total executable lines found by kcov."""
 
 
+@pytest.mark.timeout(120)
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
+    reason="kcov hangs on aarch64"
+)
+def test_ensure_mod_tests():
+    """Fail if files containing tests don't also have a 'tests' module."""
```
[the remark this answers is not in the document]
tnotw: Yup, amended.

tests/integration_tests/build/test_coverage.py
```diff
@@ -32,6 +32,44 @@
...
+    # List all source files containing rust #[test] attribute,
```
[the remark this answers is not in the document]
tnotw: Yup, amended.

tests/integration_tests/build/test_coverage.py
```diff
@@ -32,6 +32,44 @@
...
+    # (excluding generated files).
+    # Take the list and check each file contains 'mod tests {', fail if not.
```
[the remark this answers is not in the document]
tnotw: Yup, amended.

tests/integration_tests/build/test_coverage.py
tnotw: Ta.  Sorry :-)

tests/integration_tests/build/test_coverage.py
tnotw: Yup,  amended.

tests/integration_tests/build/test_coverage.py
tnotw: Done.

tests/integration_tests/build/test_coverage.py
```diff
@@ -32,6 +32,44 @@
 """Regex for extracting number of total executable lines found by kcov."""
 
 
+@pytest.mark.timeout(120)
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
+    reason="kcov hangs on aarch64"
```
lauralt, line 38: We should change the reason here because this test doesn't actually use in any way `kcov`.
```suggestion
    reason="no need to test it on multiple platforms"
```

tests/integration_tests/build/test_coverage.py
```diff
@@ -32,6 +32,44 @@
...
+    # ignore the return code.
+    result = utils.run_cmd(cmd, no_shell=False, ignore_return_code=True)
+
+    error_msg = (
+        'Tests found in files without a "tests" module:\n {}'
+        'To ensure code coverage is reported correctly, please check that '
+        'your tests are in a module named "tests".'.format(result.stdout)
+    )
+
+    # stdout should be empty, where empty strings are false
```
lauralt, line 69: Please remove this comment or, if you want to keep it, rephrase it, something like this: `if stdout is not empty, the test should fail`. But I vote for removing it as it is pretty clear from the previous lines what you are asserting here (in my previous comment for this section I was removing it but it was probably hard to notice, sorry :D).

tests/integration_tests/build/test_coverage.py
tnotw: Done.

tests/integration_tests/build/test_coverage.py
tnotw: Done and apologies,  I did miss that.

tnotw:
Thanks very much @lauralt for your help, time and patience,  much appreciated.

tests/integration_tests/build/test_coverage.py
```diff
@@ -32,6 +32,43 @@
 """Regex for extracting number of total executable lines found by kcov."""
 
 
+@pytest.mark.timeout(120)
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
```
iulianbarbu, line 37: nit: I understand that we can assume that this test could run once per CI run. However, if one developer uses an arm machine for development and this test fails, he would find out only after the CI finished running the tests. The issue with this is that it will add a delay in getting the PR in a "green" state, assuming that the development on one architecture must be compliant with the tests on any other architecture and the PR author can not test on all architectures, before submitting the PR. I can not tell if this is a very pressing issue, because I do not know how many developers are pushing code to firecracker from an arm machine exclusively and we should take this issue into consideration only if it is a real issue. I just wanted it to call it out, because it seems to that this can be easily fixed if we would have some kind of CI part which is common for all tested platforms and must run once. I do not know if we should think about this too much right now or take it into consideration, but I am curious how you guys feels about it.

tnotw: If a developer were developing on arm, then because kcov never runs on arm, the coverage report would always be unknown,  and they'd have to push and trigger a CI run to get the coverage percentage anyway (maybe? Or run it on x86_64 to get the figure).  At this point they'd find out if they'd misnamed anything.  

It feels like this test and the kcov run should be linked via architecture, but I don't mind changing it if anyone feels strongly about it.

lauralt: Hmm, I suggested to change the reason there to also not link 100% with `kcov` because also for the uniformity / consistency reason I prefer to have all unit tests in `mod tests`.
Anyway, @iulianbarbu has raised a valid issue here, at this moment we are skipping quite a few tests on arm (mostly style tests). Maybe we should add an issue for that? In the meantime this PR can be unblocked I guess.

iulianbarbu: > If a developer were developing on arm, then because kcov never runs on arm, the coverage report would always be unknown, and they'd have to push and trigger a CI run to get the coverage percentage anyway (maybe? Or run it on x86_64 to get the figure). At this point they'd find out if they'd misnamed anything.
> 
> It feels like this test and the kcov run should be linked via architecture, but I don't mind changing it if anyone feels strongly about it.

Yes, at the moment, because of this test and kcov on arm, it seems that makes sense to have some tests run via architecture. `kcov` on arm has some instability issues. I do not know if this is the reason, or there are others, but because of this lack of 100% support on arm, we are coerced to enforce coverege test to run on `x86_64` arch only.

The reason I was proposing the platform independent execution CI part, which executes once all the tests that are platform independent, is because this test and others (style test), does some static analysis on the source code. If there would be platform dependent results, than it will make sense to make the test platform aware. In our current CI infrastructure, to run this test as if it is a platform independent test, once, we would need the `platform.machine()` check to skip it if the platform is not `x86_64`. However, this will not be consistent if the CI is run only on an arm machine, as well as if a developer runs the integration tests on an arm machine only.

To change this means to change the style checks as well. My thought is that we can stick to this solution at the moment and track this suggestion of platform independent tests a issue. I will open an issue for this, and link it to this PR.

tnotw: Yes, sorry, you're absolutely right. :-)

Issue #1834, opened 2020-04-28
False positives on VMM Rust integration tests


aghecenco:
The integration tests that exercise full VMM functionalities introduced by #1791 issue a `vmm.stop(0)` to gracefully stop the vcpu loops. This exits the integration test process with a success code and interferes with `cargo kcov` and `cargo test` normal functioning. This is why new tests that do this do not account for test coverage.

The `vmm.stop()` can be removed because the process exits anyway **but**
1. the vcpu threads will not exit gracefully
1. this plus the CPU pinning `kcov` does to serialize instructions for coverage computation breaks the VMM seccomp integration test.

We should:
1. remove the `vmm.stop()` in the Rust VMM integ tests; let the test harness (`cargo`) control how the test process exits
1. figure out how to gracefully stop the vcpu threads before letting `cargo` force-end the test process, bringing them down or panicking them;. the cleanest way is to have the test guest kernel exit and let them go through the `exited` state
1. figure out why the `seccomp` integ test doesn't work

acatangiu:
I think we can get everything working nicely with different `vmm.stop()` implementations for tests and release.
`Vcpu` cleanup can be done like we did in https://github.com/firecracker-microvm/firecracker/pull/1753

aghecenco:
The integration tests don't compile the `#[cfg(test)]`-marked code. The VMM crate it uses is exactly the same as for release.

acatangiu:
> The integration tests don't compile the #[cfg(test)]-marked code.

That's right..

I wonder if coverage reporting would work fine if we `fork` during tests. We could run the actual test in the child and the parent would only report the status.

andreeaflorescu:
We could also whitelist sysprocmask and let the vcpu threads exit cleanly.

aghecenco:
This isn't a seccomp issue in the unit tests. The problem with the vmm exiting is that it takes down the integration test process with it. Also, there's no way to force the vcpus to exit but leave the process running using just the public facing programmatic api.

aghecenco:
Another fun thing I discovered is that the seccomp test I added in the new integration tests is broken. The signal handler never gets called in the Rust integ tests when running under kcov.
* `cargo test` -> integ tests are run, signal handler is called, seccomp test works
* `cargo kcov` -> integ tests are run, signal handler **not** called, seccomp test broken

acatangiu:
> Also, there's no way to force the vcpus to exit but leave the process running using just the public facing programmatic api.

[...]

PR #1839, opened 2020-04-29
vmm: fix rust integration tests

aghecenco:
## Reason for This PR

#1834 

## Description of Changes

* the tests that run a vmm now `fork`. The vmm itself runs in the child process and terminates either with `vmm.stop()` (exit code 0) or `libc::exit(errcode)` in case of panic. The parent waits on the vmm process and asserts on its exit status.
* signal handling doesn't work under `kcov` => changed the default action in the test seccomp filter to `Kill` instead of `Trap`. `Trap`ping doesn't work either under `kcov` (!!) and, now that the vmm runs in a forked child, we can assert on its `SIGSYS`-marked exit code.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/tests/integration_tests.rs
+++ b/src/vmm/tests/integration_tests.rs
@@ -13,6 +13,7 @@ extern crate vmm_sys_util;
 mod mock_devices;
 mod mock_resources;
 mod mock_seccomp;
+mod test_utils;
 
 use std::io;
 use std::thread;
@@ -20,7 +21,6 @@ use std::time::Duration;
 
 use polly::event_manager::EventManager;
 use seccomp::{BpfProgram, SeccompLevel};
-use utils::signal::register_signal_handler;
 use vmm::builder::{build_microvm, setup_serial_device};
 use vmm::default_syscalls::get_seccomp_filter;
 use vmm::resources::VmResources;
@@ -29,7 +29,8 @@ use vmm_sys_util::tempfile::TempFile;
 
 use mock_devices::MockSerialInput;
 use mock_resources::{MockBootSourceConfig, MockVmResources};
-use mock_seccomp::{mock_sigsys_handler, MockSeccomp, SIGSYS_RECEIVED};
+use mock_seccomp::MockSeccomp;
+use test_utils::{restore_stdin, set_panic_hook};
 
 #[test]
 fn test_setup_serial_device() {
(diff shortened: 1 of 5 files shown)


src/vmm/tests/integration_tests.rs
```diff
@@ -47,48 +48,83 @@ fn test_setup_serial_device() {
...
-        let mut event_manager = EventManager::new().unwrap();
-        let empty_seccomp_filter = get_seccomp_filter(SeccompLevel::None).unwrap();
+            let vmm = build_microvm(&resources, &mut event_manager, &empty_seccomp_filter).unwrap();
 
-        let vmm = build_microvm(&resources, &mut event_manager, &empty_seccomp_filter).unwrap();
-        // This exits the process, so we won't get the output from cargo.
-        vmm.lock().unwrap().stop(0);
+            // This test kernel exits after saying "Hello world". Give the vCPUs some time to exit.
+            thread::sleep(Duration::from_millis(200));
+            vmm.lock().unwrap().stop(0);
```
acatangiu, line 82: If the test kernel actually exits with `HALT`/`SHUTDOWN`, then we could validate that shutdown path as well:

```suggestion
            // The vmm should exit once its workload completes and signals the exit event.
            let ev_count = event_manager.run_timeout(Duration::from_millis(500));
            // If we got here, something went wrong.
            vmm.lock().unwrap().stop(-1);
```

src/vmm/tests/integration_tests.rs
```diff
@@ -47,48 +48,83 @@ fn test_setup_serial_device() {
...
-    assert!(unsafe { SIGSYS_RECEIVED });
-    // This exits the process, so we won't get the output from cargo.
-    vmm.lock().unwrap().stop(0);
+            // The customer "forgot" to whitelist the KVM_RUN ioctl.
+            let filter: BpfProgram = MockSeccomp::new().without_kvm_run().into();
+            let _ = build_microvm(&resources, &mut event_manager, &filter).unwrap();
+            // Give the vCPUs a chance to attempt KVM_RUN.
+            thread::sleep(Duration::from_millis(200));
+            // Should never get here.
+            loop {}
```
acatangiu, line 117:
```suggestion
            vmm.lock().unwrap().stop(some_invalid_exit_code);
```

otherwise parent `waitpid` might spin forever (or at least rust test timeout period).

src/vmm/tests/mock_seccomp/mod.rs
```diff
@@ -207,7 +204,7 @@ impl MockSeccomp {
             ]
             .into_iter()
             .collect(),
-            default_action: SeccompAction::Trap,
+            default_action: SeccompAction::Kill,
```
iulianbarbu, line 207: Seccomp [documentation](http://man7.org/linux/man-pages/man2/seccomp.2.html) pertaining to `SECCOMP_RET_KILL_PROCESS` says that:
> The process terminates as though killed by a SIGSYS signal.
    Even if a signal handler has been registered for SIGSYS, the
    handler will be ignored in this case and the process always
    terminates.

Based on this doc, I can not follow the logic. The `mock_sigsys_handler` would have to be executed, and then the `SIGSYS_RECEIVED` boolean changed to `true`, for this test to work with success. However, the docs says that when receiving `SECCOMP_RET_KILL_PROCESS`, the  registered SIGSYS handler won't be executed.

Am I missing something?

aghecenco: The doc says that the signal handler will be ignored even if present. I deleted it. The test now passes on asserting that the child's exit code is "death by `SIGSYS`".

The reason behind this change is that, with `SECCOMP_RET_TRAP` (as was before), the signal handler is **never** executed **when running under `kcov`**. The same happens in unit tests, where if `kcov` is detected (by relying on the fact that it forces CPU affinity to 1), the signal handler part is not verified.

iulianbarbu: Oh, right. I've looked again on the diff, and missed some part of the changes. It makes sense now, thanks.

src/vmm/tests/integration_tests.rs
aghecenco: The workload only completes on `x86_64`, the `aarch64` test kernel doesn't halt (there's no equivalent of `HLT` either) so I'll change this to let the kernel run its course on `x86_64` but force the vmm to stop on `aarch64`.

src/vmm/tests/integration_tests.rs
aghecenco: Done

aghecenco:
@acatangiu @iulianbarbu feedback addressed :smiley: Can you please take another look?

src/vmm/tests/integration_tests.rs
```diff
@@ -47,48 +48,88 @@ fn test_setup_serial_device() {
...
-        let empty_seccomp_filter = get_seccomp_filter(SeccompLevel::None).unwrap();
-
-        let vmm = build_microvm(&resources, &mut event_manager, &empty_seccomp_filter).unwrap();
-        // This exits the process, so we won't get the output from cargo.
-        vmm.lock().unwrap().stop(0);
+            let vmm = build_microvm(&resources, &mut event_manager, &empty_seccomp_filter).unwrap();
+
+            // On x86_64, the vmm should exit once its workload completes and signals the exit event.
+            // On aarch64, the test kernel doesn't exit, so the vmm is force-stopped.
+            let _ = event_manager.run_with_timeout(500).unwrap();
```
acatangiu, line 82: nit: I hope 500ms is enough for slow machines too

src/vmm/tests/integration_tests.rs
```diff
@@ -47,48 +48,88 @@ fn test_setup_serial_device() {
...
+
+            #[cfg(target_arch = "x86_64")]
+            vmm.lock().unwrap().stop(-1); // If we got here, something went wrong.
+            #[cfg(target_arch = "aarch64")]
+            vmm.lock().unwrap().stop(0);
```
acatangiu, line 87: > The workload only completes on x86_64, the aarch64 test kernel doesn't halt (there's no equivalent of HLT either)

nit: your PR explanation would be useful here in the code as well

src/vmm/tests/integration_tests.rs
```diff
@@ -47,48 +48,88 @@ fn test_setup_serial_device() {
...
-    // Give the signal handler some time to complete.
-    thread::sleep(Duration::from_millis(30));
-    assert!(unsafe { SIGSYS_RECEIVED });
-    // This exits the process, so we won't get the output from cargo.
-    vmm.lock().unwrap().stop(0);
+            // The customer "forgot" to whitelist the KVM_RUN ioctl.
+            let filter: BpfProgram = MockSeccomp::new().without_kvm_run().into();
+            let vmm = build_microvm(&resources, &mut event_manager, &filter).unwrap();
+            // Give the vCPUs a chance to attempt KVM_RUN.
+            thread::sleep(Duration::from_millis(200));
```
acatangiu, line 120: same nit: I hope `200 ms` is enough for running tests on slow machines

Issue #1809, opened 2020-04-21
Intermittent failure of test_rx_rate_limiting test


raduiliescu:
test_rx_rate_limiting is failing intermittently.
Example of output:

```
=================================== FAILURES ===================================
____________________ test_rx_rate_limiting[ubuntu_with_ssh] ____________________
integration_tests/functional/test_rate_limiter.py:147: in test_rx_rate_limiting
    _check_rx_rate_limiting(test_microvm, guest_ips)
integration_tests/functional/test_rate_limiter.py:283: in _check_rx_rate_limiting
    assert (
E   assert 11.539077758789062 < 10
E    +  where 11.539077758789062 = _get_difference(92758000.0, 104857600.0)
------------------------------ Captured log setup ------------------------------
09:51:19 asyncio: DEBUG Using selector: EpollSelector
09:51:19 commands: DEBUG 
[8375] Command:
cd /firecracker && TARGET_CC=musl-gcc cargo build --release --target x86_64-unknown-linux-musl
[8375] stderr:
    Finished release [optimized] target(s) in 0.03s
------------------------------ Captured log call -------------------------------
09:51:19 asyncio: DEBUG Using selector: EpollSelector
09:51:19 commands: DEBUG 
[8378] Command:
ip netns add 7f369f93-a01c-4183-b4ec-00dd48078101
09:51:19 asyncio: DEBUG Using selector: EpollSelector
09:51:19 commands: DEBUG 
[8380] Command:
ln -f /tmp/fctest-efdxs5su/7f369f93-a01c-4183-b4ec-00dd48078101/log_fifo /srv/jailer/firecracker/7f369f93-a01c-4183-b4ec-00dd48078101/root/log_fifo
09:51:19 asyncio: DEBUG Using selector: EpollSelector
09:51:19 commands: DEBUG 
[8382] Command:
chown 1234:1234 /srv/jailer/firecracker/7f369f93-a01c-4183-b4ec-00dd48078101/root/log_fifo
09:51:19 asyncio: DEBUG Using selector: EpollSelector
09:51:19 commands: DEBUG 
[8385] Command:
[...]

PR #1842, opened 2020-05-03
Remove dependency of `utils` crate on `dumbo` crate

### LINKED ISSUE #1769: Remove `utils` crate dependency on `dumbo` crate
`dumbo` is used in `utils` to unit test the `Tap`, as in sending IPv4 packets and ARP frames on the tap, and verifying at the other end if the packet/frames retrieval is successful and if the payload is the same. It will be just enough to send some plain data on the tap, and receive it back with success, and keep the same behavior for the `Tap` unit tests, and for the `Tap` implementation as well.

ahmedaabouzied:
## Reason for This PR

Fixes  #1769. 
Remove the dependency of crate `utils` on crate `dumbo`. Dubmo was used to test sending IPv4 packets on the Tap interface. This refactor is for better testing for `utils` crate. Since now if the tests fail they only refer to problems in `utils` crate and not in `dumbo`.

## Description of Changes

*  Change `test_write` to write raw IPV4 UDP packet with a constant payload and assert that the same payload was received on the other end.
* Change `test_write` to parse the ARP request and respond without needing `dumbo`. And validating the payload read from the TAP device.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/utils/src/net/tap.rs
+++ b/src/utils/src/net/tap.rs
@@ -226,7 +226,6 @@ impl AsRawFd for Tap {
 
 #[cfg(test)]
 mod tests {
-    extern crate dumbo;
 
     use std::io::Read;
     use std::mem;
@@ -236,17 +235,15 @@ mod tests {
     use std::sync::atomic::{AtomicUsize, Ordering};
     use std::time::Duration;
 
-    use self::dumbo::{
-        EthIPv4ArpFrame, EthernetFrame, IPv4Packet, MacAddr, UdpDatagram, ETHERNET_PAYLOAD_OFFSET,
-        ETHERTYPE_ARP, ETHERTYPE_IPV4, ETH_IPV4_FRAME_LEN, PROTOCOL_UDP, UDP_HEADER_SIZE,
-    };
-
     use super::*;
 
     const DATA_STRING: &str = "test for tap";
     const SUBNET_MASK: &str = "255.255.255.0";
     const TAP_IP_PREFIX: &str = "192.168.241.";
-    const FAKE_MAC: &str = "12:34:56:78:9a:bc";
+    const IP_HEADER_LENGTH: usize = 20;
+    const UDP_HEADER_LENGTH: usize = 8;
+    const ARP_PACKET_LENGTH: usize = 52;
+    const UDP_PAYLOAD_OFFSET: usize = 52;
 
     // We skip the first 10 bytes because the IFF_VNET_HDR flag is set when the interface
     // is created, and the legacy header is 10 bytes long without a certain flag which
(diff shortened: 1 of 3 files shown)


src/utils/src/net/tap.rs
```diff
@@ -543,40 +543,11 @@ mod tests {
...
-        );
-
-        let udp = UdpDatagram::write_incomplete_datagram(ipv4.payload_mut(), payload).unwrap();
-        udp.finalize(
-            src_addr.port(),
-            dst_addr.port(),
-            Some((*src_addr.ip(), *dst_addr.ip())),
-        );
+    fn build_packet(payload: &[u8]) -> Vec<u8> {
+        let mut base = Vec::new();
```
andreeaflorescu, line 547: Can you add some comments to what these bytes represent?

For example, you can use `#[rufmt::skip]` on the initialization of base and then split the content per logical component. 
```rust
base = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 18, 52, 86, 120, 154, 189, // source port
            ...............         // destination port
            ...................... // payload length
            .......
]
```

BTW, shouldn't the payload length be written based on the `payload` parameter size?

ahmedaabouzied: I separated this ugly array into multiple variables with each representing a particular segment of the Ethernet, IP, UDP headers. I also added some comments explaining each variable. 
https://github.com/firecracker-microvm/firecracker/blob/72d58ff43065e8188b51e65600f3fb9e4bc56917/src/utils/src/net/tap.rs#L447-L494

src/utils/src/net/tap.rs
```diff
@@ -608,27 +579,15 @@ mod tests {
         let src_ip: Ipv4Addr = format!("{}{}", TAP_IP_PREFIX, next_ip + 1).parse().unwrap();
         let src_port = 44444;
         let src_addr = SocketAddrV4::new(src_ip, src_port);
-        let src_mac = MacAddr::parse_str(FAKE_MAC).unwrap();
+        // let src_mac = MacAddr::parse_str(FAKE_MAC).unwrap();
```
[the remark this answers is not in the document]
ahmedaabouzied: done

src/utils/src/net/tap.rs
```diff
@@ -543,40 +543,11 @@ mod tests {
 
     // Given a buffer of appropriate size, this fills in the relevant fields based on the
     // provided information. Payload refers to the UDP payload.
```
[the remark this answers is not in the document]
ahmedaabouzied: Done. Also renamed the function to `build_udp_packet` to better represent what it does.
https://github.com/firecracker-microvm/firecracker/blob/72d58ff43065e8188b51e65600f3fb9e4bc56917/src/utils/src/net/tap.rs#L446-L447

ahmedaabouzied:
@iulianbarbu, Now I removed the dependency on `dempo` in `test_read` which makes the `utils` crate now independent of `dempo`.

Right now we do the packet parsing required for `test_read` and `test_write` in the functions `build_udp_packet` and `build_arp_response` which still construct the packets with Ethernet, IPV4, UDP/ARP headers.

Your idea of using [packet sockets](http://man7.org/linux/man-pages/man7/packet.7.html) seems interesting and elegant. I don't think we can just send raw packets to the TAP device because as mentioned in the [TUN/TAP docs](https://www.kernel.org/doc/Documentation/networking/tuntap.txt) tap devices work with Ethernet frames.
> Depending on the type of device chosen the userspace program has to read/write
  IP packets (with tun) or ethernet frames (with tap). Which one is being used depends on the flags given with the ioctl().

Sending a packet without a valid Ethernet header would return an error `invalid argument` as I have encountered. Check out this [Stackoverflow comment](https://stackoverflow.com/questions/31179304/a-write-to-tun-tap-interface-returns-invalid-argument#comment50710409_31179304) on facing a similar problem.

We still can use a packet socket to avoid using IP, UDP packets and only send valid Ethernet frames to the TAP device. Which seems easier to do.

Since [std::net](https://doc.rust-lang.org/std/net/index.html) doesn't have an implementation of packet sockets. We can either use an external library to work with the packet sockets (which defeats the purpose of this issue I guess) or we use `libc` using some unsafe rust. We create the packet socket, then bind it to an appropriate IP address to intercept packets from the TAP device. Something like the following : 
```rust
unsafe {
    // create the packet socket
    let sock_fd =  libc::socket(libc::AF_PACKET, libc::SOCK_RAW, libc::ETH_P_ALL) };
         if sock_fd < 0 {
            println!("error creating socket");
         }
   // bind the socket to IP address to only get packets from the tap device
   let sockaddr = libc::sockaddr_ll {
            sll_family:   libc::AF_PACKET as u16,
            sll_protocol: libc::ETH_P_ALL as u16,
            sll_ifindex:  0,
            sll_hatype:   0,
            sll_pkttype:  0,
            sll_halen:    6,
            sll_addr:  <tap IP address as [u8]>
        };
   let res = libc::bind(self.lower,
                                 &sockaddr as *const libc::sockaddr_ll as *const libc::sockaddr,
                                 mem::size_of::<libc::sockaddr_ll>() as u32);
            if res == -1 { 
                   println!("error binding tap socket")
            }
}
```
I think this might work well for `test_write` but still in `test_read` we need to respond to the ARP packet to resolve the address of the tap device like here : 
https://github.com/firecracker-microvm/firecracker/blob/7cc940edc1665c6b60dbf4ce9358a1eadd1969cc/src/utils/src/net/tap.rs#L499-L508

So even when using packet sockets to avoid the packet parsing part, we still need to parse packets to send that ARP response according to the ARP request sent to the tap device. 
So what can we do about that @iulianbarbu?

**PS: Sorry for the late response. I only got to work on this over the weekend.**

ahmedaabouzied:
@iulianbarbu, @andreeaflorescu
We would need to squash these two commits into one with a decent commit message before a merge.

ahmedaabouzied:
I see that the integration test `test_rx_rate_limiting` is failing. 
```bash
=================================== FAILURES ===================================
____________________ test_rx_rate_limiting[ubuntu_with_ssh] ____________________
integration_tests/functional/test_rate_limiter.py:152: in test_rx_rate_limiting
    _check_rx_rate_limit_patch(test_microvm, guest_ips)
integration_tests/functional/test_rate_limiter.py:373: in _check_rx_rate_limit_patch
    _check_rx_bandwidth(test_microvm, guest_ips[0], bandwidth_kb)
integration_tests/functional/test_rate_limiter.py:426: in _check_rx_bandwidth
    assert diff_pc < MAX_BYTES_DIFF_PERCENTAGE
E   assert 11.8828125 < 10
```
I'll check it out. Any ideas why that may be?

iulianbarbu:
Let me respond to each of your issues:

---- 
> So even when using packet sockets to avoid the packet parsing part, we still need to parse packets to send that ARP response according to the ARP request sent to the tap device.
So what can we do about that @iulianbarbu?

In `test_read` case, what do you think if we do not use IPv4 and any IP based lower-level/higher-level protocol at all? We can get rid of ARP replies this way. For a `test_read` test we would need to send some data using a `AF_PACKET, SOCK_DGRAM, libc::ETH_P_ALL` packet socket. This way we would send to the tap file and read from it some custom link-layer packets (ethernet frames with custom payload - plain data - not any of the well known IP based protocols). We will still need parsing of ethernet frames (to reach the payload and assert on it), but we wouldn't need any logic related to building any replies. I saw that you used `SOCK_RAW` in the provided examples, but I think `[SOCK_DGRAM](http://man7.org/linux/man-pages/man7/packet.7.html)` would be a better fit, because it will delegate to the kernel the task of prepending or removing the ethernet headers:

>  SOCK_DGRAM operates on a slightly higher level. The physical header is removed before passing  the  packet  to  the user (and prepended before sending it).

In `test_write` case we still need to build an ethernet frame either way, but building it will be simpler if we use `AF_PACKET, SOCK_DGRAM, libc::ETH_P_ALL` packet socket as well,  because we wouldn't need to add information about any IP based protocol.

However, these suggestions are only a call out to get your feedback and understand if it would make sense, in terms of reducing complexity and thorough testing the tap. I want to let you know that it seems to me that you've already resolved the issue and that looks good to me, so any continuation to change this PR to the packet socket alternative is your choice. My thoughts is that we can stick with this solution if there is no different opinion.

------
> We would need to squash these two commits into one with a decent commit message before a merge.

Yes, I agree, we should wrap this up into one single commit, with a decent commit message.

------
> I see that the integration test `test_rx_rate_limiting` is failing.

We have an instability issue with this test and we are currently trying to understand why it is randomly failing and how can we make it stable. I think that you must try to trigger again the CI tests. Anyway, you will need to force push to the branch that one commit with both tests changes, so you will trigger the CI then.

andreeaflorescu:
> I see that the integration test `test_rx_rate_limiting` is failing.
> 
> ```shell
> =================================== FAILURES ===================================
> ____________________ test_rx_rate_limiting[ubuntu_with_ssh] ____________________
> integration_tests/functional/test_rate_limiter.py:152: in test_rx_rate_limiting
>     _check_rx_rate_limit_patch(test_microvm, guest_ips)
> integration_tests/functional/test_rate_limiter.py:373: in _check_rx_rate_limit_patch
>     _check_rx_bandwidth(test_microvm, guest_ips[0], bandwidth_kb)
> integration_tests/functional/test_rate_limiter.py:426: in _check_rx_bandwidth
>     assert diff_pc < MAX_BYTES_DIFF_PERCENTAGE
> E   assert 11.8828125 < 10
> ```
> 
> I'll check it out. Any ideas why that may be?

@ahmedaabouzied  If the test is not failing regularly with your PR, it is most likely unrelated to these changes. We already logged an issue with this test intermittently failing: https://github.com/firecracker-microvm/firecracker/issues/1809.

ahmedaabouzied:
@iulianbarbu, First of all, thanks a lot for this discussion. I was introduced to a lot of new concepts because of your suggestions.
I will investigate what you've suggested about using `SOCK_DGRAM` because that seems interesting to learn about but as you've said if it now works then let's go with the current solution.

iulianbarbu:
> @iulianbarbu, First of all, thanks a lot for this discussion. I was introduced to a lot of new concepts because of your suggestions.

Cool. Glad to hear. 👍 

> I will investigate what you've suggested about using `SOCK_DGRAM` because that seems interesting to learn about but as you've said if it now works then let's go with the current solution.

Sure. Let us know when you squash those two commits into one and we will see then if there is any other feedback to send, or we can merge it.

ahmedaabouzied:
Hey, I struggled with squashing the commits after you merged master into the branch so I rebased the branch on master instead. I hope that's okay.
Let me know if the commit message could be improved.

src/utils/src/net/tap.rs
```diff
@@ -464,6 +436,101 @@ mod tests {
...
+        // The IP total length value.
+        let total_length: [u8; 2] = [
+            0,
+            IP_HEADER_LENGTH as u8 + UDP_HEADER_LENGTH as u8 + payload_len,
+        ];
+
+        // Fixed header values:
+        // Identification, Flags, Fragment Offset, Time To Live, Protocol number (17 for UDP),
+        // Header Checksum.
+        // let ipv4_header_values = [174, 247, 64, 0, 64, 17, 40, 118];
```
iulianbarbu, line 471: Is this a leftover?

src/utils/src/net/tap.rs
```diff
@@ -464,6 +436,101 @@ mod tests {
...
+        let ipv4_version = [69, 0];
+
+        // The IP total length value.
+        let total_length: [u8; 2] = [
+            0,
+            IP_HEADER_LENGTH as u8 + UDP_HEADER_LENGTH as u8 + payload_len,
+        ];
+
+        // Fixed header values:
+        // Identification, Flags, Fragment Offset, Time To Live, Protocol number (17 for UDP),
```
iulianbarbu, line 469: Are all of these headers embedded in the `ipv4_header_values` variable? If so, can you put this comment right above the place where you initialize the variable?

src/utils/src/net/tap.rs
ahmedaabouzied: I'll check it out

iulianbarbu:
> Hey, I struggled with squashing the commits after you merged master into the branch so I rebased the branch on master instead. I hope that's okay.
> Let me know if the commit message could be improved.

Related to this, it's totally fine how you've rebased it. It's preferable to update the branch rebasing on top of master, rather than merging master into the branch, anyway. Feel free to rebase again on top of master. :)

src/utils/src/net/tap.rs
ahmedaabouzied: done

src/utils/src/net/tap.rs
ahmedaabouzied: It's now "Static header values" instead of the confusing "Fixed"

PR #1864, opened 2020-05-05
Fix MAC spoofing metric

### LINKED ISSUE #1856: Incorrect MAC spoofing metric
Currently, the Net device MAC spoofing metric goes into effect only for the MAC address specified through the Net constructor.
If we don't specify one by writing it in the config space, the driver generates an address. We should also  increment the spoofing metric for this case.

ioanachirca:
Signed-off-by: Ioana Chirca <chioana@amazon.com>

## Reason for This PR

Fixes #1856 

## Description of Changes

Update the Mac address whenever there is a write to the config space.
This way, we update the spoofing metric for both cases: guest MAC provided by the user or
by the guest driver.

## License Acceptance

## PR Checklist

- [x] No new added `unsafe` code.
- [x] No API changes.
- [x] No user-facing changes.

### DIFF
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -99,7 +99,7 @@ pub struct Net {
     interrupt_evt: EventFd,
 
     pub(crate) config_space: ConfigSpace,
-    guest_mac: Option<MacAddr>,
+    pub(crate) guest_mac: Option<MacAddr>,
 
     pub(crate) device_state: DeviceState,
     pub(crate) activate_evt: EventFd,
@@ -148,8 +148,6 @@ impl Net {
             avail_features |= 1 << VIRTIO_NET_F_MAC;
         }
 
-        let guest_mac = guest_mac.copied();
-
         let mut queue_evts = Vec::new();
         for _ in QUEUE_SIZES.iter() {
             queue_evts.push(EventFd::new(libc::EFD_NONBLOCK).map_err(Error::EventFd)?);
@@ -183,8 +181,8 @@ impl Net {
             device_state: DeviceState::Inactive,
             activate_evt: EventFd::new(libc::EFD_NONBLOCK).map_err(Error::EventFd)?,
             config_space,
-            guest_mac,
             mmds_ns,
+            guest_mac: guest_mac.copied(),
 
             #[cfg(test)]
             test_mutators: tests::TestMutators::default(),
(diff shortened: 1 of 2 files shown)


src/devices/src/virtio/net/device.rs
```diff
@@ -197,8 +193,8 @@ impl Net {
     }
 
     /// Provides the MAC of this net device.
-    pub fn guest_mac(&self) -> Option<&MacAddr> {
-        self.guest_mac.as_ref()
+    pub fn guest_mac(&self) -> MacAddr {
+        MacAddr::from_bytes_unchecked(&self.config_space.guest_mac[..MAC_ADDR_LEN])
```
sandreim, line 197: This fn seems to be called on the hot path. Why not cache the MacAddr when it's changed by the guest and always return a reference.

iulianbarbu: Because we verify on the hotpath if frames source mac address is different than the interface assigned mac address (to update `METRICS.net.tx_spoofed_mac_count`), parsing the mac address from the config every time we process TX, would add few cycles to this processing step, and this can be avoided using a cache. However, every time the guest writes to the config space, we would need to update this cache accordingly. +1 to this suggestion.

ioanachirca: Done.

src/devices/src/virtio/net/device.rs
```diff
@@ -141,14 +141,19 @@ impl Net {
...
+                .copy_from_slice(guest_mac.get_bytes());
             // When this feature isn't available, the driver generates a random MAC address.
             // Otherwise, it should attempt to read the device MAC address from the config space.
             avail_features |= 1 << VIRTIO_NET_F_MAC;
-        }
 
-        let guest_mac = guest_mac.copied();
+            mac = *guest_mac;
+        } else {
+            mac = MacAddr::from_bytes_unchecked(&config_space.guest_mac[..MAC_ADDR_LEN]);
```
iulianbarbu, line 155: My understanding is that if the config space is not given, it will be set at device activation time. This means that when we configure/build a network device without the `guest_mac` information, this information remains unknown until the network device gets activated. I think we still need `Option<MacAddr>` for guest_mac field of the Net device.

To have the `guest_mac` cache in sync with the content from the config space, we would need to update this cache when this content is change. This is done in `write_config` implementation of `VirtioDevice` trait for `Net`.

ioanachirca: I was hoping that we can get rid of the Option, but you're right, we have to know if we have a valid MAC address. 
For the save/restore case, should we save the config space *and* the Option<MacAddr> or can we assume that we save the device after the microvm booted and we have a valid address?

iulianbarbu: For the save/restore case, as we've discussed offline, it is sufficient to save only the config space and populate at restore time the `guest_mac` as well, besides the config space.

It is important to note that this parsing of the config space, to update the guest_mac cache, is done in the device activation, as well in the device restore, which are on the hotpath as well. The tradeoff is that we will have no degradation in the network device performance, but we will have some degradation on the microVM booting/restoring from snapshot. I do not know what to expect from this degradation, and if the performance tests are green, I guess it is not significant at all.

PR #1874, opened 2020-05-08
Create Snapshot

### LINKED ISSUE #1846: Vmm support for creating a microVM snapshot
This tracks complete snapshot creating capability offered through the vmm programmatic API.

Once this is done, we tie it to Firecracker HTTP API for full stack snapshot creating functionality.

ioanachirca:
## Reason for This PR

Fixes #1846

## Description of Changes

- add a new `VcpuEvent` to save state and possible `VcpuResponses`
- add `create_snapshot` vmm method

## License Acceptance

## PR Checklist

### DIFF
--- a/src/snapshot/Cargo.toml
+++ b/src/snapshot/Cargo.toml
@@ -8,7 +8,7 @@ autobenches = false
 serde = ">=1.0.27"
 serde_derive = ">=1.0.27"
 bincode = "1.2.1"
-versionize = { git = "https://github.com/firecracker-microvm/versionize", tag = "v0.1.1" }
+versionize = { git = "https://github.com/firecracker-microvm/versionize", tag = "v0.1.0" }
 versionize_derive = { git = "https://github.com/firecracker-microvm/versionize_derive", tag = "v0.1.0" }
 
 [dev-dependencies]
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -12,12 +12,14 @@
 
 extern crate kvm_bindings;
 extern crate kvm_ioctls;
+extern crate lazy_static;
 extern crate libc;
 extern crate polly;
 extern crate serde;
 #[macro_use]
 extern crate serde_derive;
 extern crate serde_json;
+extern crate sysconf;
 
 extern crate arch;
 #[cfg(target_arch = "x86_64")]
@@ -40,6 +42,7 @@ pub mod builder;
 /// Syscalls allowed through the seccomp filter.
 pub mod default_syscalls;
 pub(crate) mod device_manager;
+pub(crate) mod memory_dump;
 /// Save/restore utilities.
 pub mod persist;
 /// Resource store for configured microVM resources.
@@ -48,6 +51,8 @@ pub mod resources;
 pub mod rpc_interface;
 /// Signal handling utilities.
 pub mod signal_handler;
+/// microVM state versions.
+pub mod version_map;
 /// Wrappers over structures used to configure the VMM.
 pub mod vmm_config;
 mod vstate;
@@ -56,6 +61,8 @@ use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
 use std::io;
 use std::os::unix::io::AsRawFd;
+#[cfg(target_arch = "x86_64")]
+use std::sync::mpsc::RecvTimeoutError;
 use std::sync::Mutex;
 use std::time::Duration;
 
@@ -65,12 +72,19 @@ use device_manager::legacy::PortIODeviceManager;
 use device_manager::mmio::MMIODeviceManager;
 use devices::BusDevice;
 use logger::{LoggerError, MetricsError, METRICS};
+
+#[cfg(target_arch = "x86_64")]
+use persist::{MicrovmState, SaveMicrovmStateError, VmInfo};
 use polly::event_manager::{self, EventManager, Subscriber};
 use seccomp::{BpfProgram, BpfProgramRef, SeccompFilter};
+#[cfg(target_arch = "x86_64")]
+use snapshot::Persist;
 use utils::epoll::{EpollEvent, EventSet};
 use utils::eventfd::EventFd;
 use utils::time::TimestampUs;
 use vm_memory::{GuestMemory, GuestMemoryMmap, GuestMemoryRegion, GuestRegionMmap};
+#[cfg(target_arch = "x86_64")]
+use vstate::VcpuState;
 use vstate::{Vcpu, VcpuEvent, VcpuHandle, VcpuResponse, Vm};
 
 /// Success exit code.
@@ -365,6 +379,63 @@ impl Vmm {
     /// Returns a reference to the inner KVM Vm object.
     pub fn kvm_vm(&self) -> &Vm {
         &self.vm
+    }
+
+    /// Saves the state of a paused Microvm.
+    #[cfg(target_arch = "x86_64")]
+    pub fn save_state(&mut self) -> std::result::Result<MicrovmState, SaveMicrovmStateError> {
+        let vcpu_states = self.save_vcpu_states()?;
+
+        let vm_state = self
+            .vm
+            .save_state()
+            .map_err(SaveMicrovmStateError::InvalidVmState)?;
+
+        let device_states = self.mmio_device_manager.save();
+
+        let mem_size_mib =
+            self.guest_memory()
+                .map_and_fold(0, |(_, region)| region.len(), |a, b| a + b)
+                >> 20;
+
+        Ok(MicrovmState {
+            vm_info: VmInfo { mem_size_mib },
+            vm_state,
+            vcpu_states,
+            device_states,
+        })
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    fn save_vcpu_states(&mut self) -> std::result::Result<Vec<VcpuState>, SaveMicrovmStateError> {
+        for handle in self.vcpus_handles.iter() {
+            handle
+                .send_event(VcpuEvent::SaveState)
+                .map_err(SaveMicrovmStateError::SignalVcpu)?;
+        }
+
+        let vcpu_responses = self
+            .vcpus_handles
+            .iter()
+            // `Iterator::collect` can transform a `Vec<Result>` into a `Result<Vec>`.
+            .map(|handle| {
+                handle
+                    .response_receiver()
+                    .recv_timeout(Duration::from_millis(400))
+            })
+            .collect::<std::result::Result<Vec<VcpuResponse>, RecvTimeoutError>>()
+            .map_err(|_| SaveMicrovmStateError::InvalidVcpuState)?;
+
+        let vcpu_states = vcpu_responses
+            .into_iter()
+            .map(|response| match response {
+                VcpuResponse::SaveState(state) => Ok(*state),
+                VcpuResponse::SaveStateFailed(_) => Err(SaveMicrovmStateError::InvalidVcpuState),
+                _ => Err(SaveMicrovmStateError::InvalidVcpuState),
+            })
+            .collect::<std::result::Result<Vec<VcpuState>, SaveMicrovmStateError>>()?;
+
+        Ok(vcpu_states)
     }
 
     /// Retrieves the KVM dirty bitmap for each of the guest's memory regions.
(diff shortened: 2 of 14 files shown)


src/vmm/src/lib.rs
```diff
@@ -461,6 +467,143 @@ impl Vmm {
...
+        self.snapshot_memory_to_file(&params.mem_file_path, mem_size_mib, params.snapshot_type)?;
+
+        let mut snapshot_file = OpenOptions::new()
+            .create(true)
+            .write(true)
+            .open(params.snapshot_path)
+            .map_err(CreateSnapshotError::SnapshotBackingFile)?;
+
+        // TODO: fallback to current version.
+        let version_map = VersionMap::new();
```
sandreim, line 507: I would put the Firecracker version map into a separate file.

src/vmm/src/lib.rs
```diff
@@ -461,6 +467,143 @@ impl Vmm {
...
+        let version = match params.version {
+            Some(version) => version,
+            _ => 1,
```
sandreim, line 510: Why not do this instead of hardcoding it ?
```suggestion
            _ => version_map.latest_version(),
```

src/vmm/src/lib.rs
```diff
@@ -461,6 +467,143 @@ impl Vmm {
...
+        };
+        let mut snapshot = Snapshot::new(version_map.clone(), version);
```
sandreim, line 512: You do not need to clone de version map.

src/vmm/src/lib.rs
```diff
@@ -461,6 +467,143 @@ impl Vmm {
...
+        snapshot
+            .save(&mut snapshot_file, &microvm_state)
+            .map_err(CreateSnapshotError::SerializeMicrovmState)?;
+
+        Ok(())
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    fn save_vcpu_states(&mut self) -> std::result::Result<Vec<VcpuState>, CreateSnapshotError> {
```
sandreim, line 521: Can we move these functions outside of the lib.rs of Vmm ?

src/vmm/src/lib.rs
```diff
@@ -461,6 +467,143 @@ impl Vmm {
...
+                    Err(CreateSnapshotError::SaveVcpuState(Some(err)))
+                }
+                _ => Err(CreateSnapshotError::SaveVcpuState(None)),
+            })
+            .collect::<std::result::Result<Vec<VcpuState>, CreateSnapshotError>>()?;
+
+        Ok(vcpu_states)
+    }
+
+    fn snapshot_memory_to_file(
```
sandreim, line 554: I think we can simplify by defining the `trait DumpMemory`:
```
trait DumpMemory<T> {
     // No need for a mem size parameter, it can be computed.
     fn dump<T: std::io::Write>(writer: &mut T) -> std::io::Result;
     fn dump_diff<T: std::io::Write>(writer: &mut T) -> std::io::Result; // might need a different Result type since this calls into KVM.
}
```
Next implement this for GuestMemory. Then you can just call self.guest_memory().dump() or dump_diff().

ioanachirca: done, I implemented this interface in `memory_dump.rs`

src/vmm/src/lib.rs
ioanachirca: Now in `lib.rs` there are only `save_state` and `save_vcpu_states` (the vcpu handles are part of the vmm so I think it's ok for this method to remain here).

src/vmm/src/memory_dump.rs
```diff
@@ -0,0 +1,87 @@
...
+        let mut writer_offset = 0;
+
+        self.with_regions_mut(|slot, region| {
+            let bitmap = dirty_bitmap.get(&slot).unwrap();
+
+            for (i, v) in bitmap.iter().enumerate() {
+                for j in 0..64 {
+                    let is_dirty_page = ((v >> j) & 1u64) != 0u64;
+                    if is_dirty_page {
+                        let page_offset = ((i * 64) + j) * 4096;
```
sandreim, line 66: It would be nice to define a const for PAGE_SIZE.

src/vmm/src/memory_dump.rs
```diff
@@ -0,0 +1,87 @@
...
+
+                        // Seek forward in case we skipped some unmodified pages.
+                        writer
```
sandreim, line 69: This is good for nor but I think it is worth not calling seek between adjacent pages and reducing the number of syscalls required to perform the operation.

src/vmm/src/persist.rs
```diff
@@ -32,6 +43,139 @@ pub struct MicrovmState {
...
+    snapshot_path: &PathBuf,
+    version: Option<u16>,
+) -> std::result::Result<(), CreateSnapshotError> {
+    let mut snapshot_file = OpenOptions::new()
+        .create(true)
+        .write(true)
+        .open(snapshot_path)
+        .map_err(CreateSnapshotError::SnapshotBackingFile)?;
+
+    // TODO: VersionMap should be managed at a higher level.
```
sandreim, line 126: We can manage it at the higher level, in that case define a parameter for it. 
The version map is not transient so It looks comfy to just use lazy static. @firecracker-microvm/compute-capsule WDYT ?

sandreim:
This PR needs to also adjust binary size integration test:
`E   AssertionError: Current firecracker binary size of 2930992 bytes is above the target of 2751144 bytes with 179848 bytes.`

src/vmm/src/persist.rs
```diff
@@ -32,6 +43,139 @@ pub struct MicrovmState {
     pub device_states: DeviceStates,
 }
 
+/// Errors related to saving Microvm state.
+#[derive(Debug)]
+pub enum SaveMicrovmStateError {
+    /// Failed to save vCPU state.
+    SaveVcpuState(Option<vstate::Error>),
```
acatangiu, line 50: nit: this is a bit weird, suggest
```suggestion
    InvalidVcpuState,
    SaveVcpuState(vstate::Error),
```

src/vmm/src/persist.rs
acatangiu: That sounds fine

src/vmm/src/memory_dump.rs
aghecenco: Or get it from [`sysconf`](http://man7.org/linux/man-pages/man2/getpagesize.2.html).

src/vmm/src/memory_dump.rs
aghecenco: Yup, a (limited) buffer for adjacent dirty memory pages is a good next step

src/vmm/src/persist.rs
aghecenco: :+1:

src/vmm/tests/integration_tests.rs
```diff
@@ -178,3 +184,60 @@ fn test_pause_resume_microvm() {
...
+                MockBootSourceConfig::new().with_default_boot_args().into();
+            let resources: VmResources = MockVmResources::new()
+                .with_boot_source(boot_source_cfg)
+                .into();
+            let mut event_manager = EventManager::new().unwrap();
+            let empty_seccomp_filter = get_seccomp_filter(SeccompLevel::None).unwrap();
+
+            let vmm = build_microvm(&resources, &mut event_manager, &empty_seccomp_filter).unwrap();
+
+            // Give the vCPUs a chance to attempt KVM_RUN.
```
aghecenco, line 208: This looks imported from the seccomp test, where the vcpus fail at KVM_RUN, please update the comment so that it doesn't hint at bad things happening

src/vmm/tests/integration_tests.rs
```diff
@@ -178,3 +184,60 @@ fn test_pause_resume_microvm() {
         }
     }
 }
+
+#[test]
+#[cfg(target_arch = "x86_64")]
+fn test_create_snapshot() {
+    let snapshot_path = PathBuf::from("snapshot_file");
```
aghecenco, line 191: I'd use `tempfile`s for these, just in case an `assert` fails

src/vmm/tests/integration_tests.rs
```diff
@@ -178,3 +184,60 @@ fn test_pause_resume_microvm() {
...
+            let pid_done = unsafe { libc::waitpid(vmm_pid, &mut vmm_status, 0) };
+            assert_eq!(pid_done, vmm_pid);
+            restore_stdin();
+            // If any panics occurred, its exit status will be != 0.
+            assert!(unsafe { libc::WIFEXITED(vmm_status) });
+            assert_eq!(unsafe { libc::WEXITSTATUS(vmm_status) }, 0);
+        }
+    }
+
+    std::fs::remove_file(snapshot_path).expect("failed to delete snapshot file");
```
aghecenco, line 241: `tempfile` would also get rid of these

src/vmm/tests/integration_tests.rs
```diff
@@ -178,3 +184,60 @@ fn test_pause_resume_microvm() {
...
+        }
+        vmm_pid => {
+            // Parent process: wait for the vmm to exit.
+            let mut vmm_status: i32 = -1;
+            let pid_done = unsafe { libc::waitpid(vmm_pid, &mut vmm_status, 0) };
+            assert_eq!(pid_done, vmm_pid);
+            restore_stdin();
+            // If any panics occurred, its exit status will be != 0.
+            assert!(unsafe { libc::WIFEXITED(vmm_status) });
+            assert_eq!(unsafe { libc::WEXITSTATUS(vmm_status) }, 0);
```
aghecenco, line 237: I think the parent should at least assert that the snapshot files aren't empty after the child has exited

aghecenco: One step further would be for it to actually deserialize the state file and verify its contents

src/vmm/src/persist.rs
ioanachirca: Done.

src/vmm/tests/integration_tests.rs
ioanachirca: Done.

src/vmm/tests/integration_tests.rs
ioanachirca: Done.

src/vmm/tests/integration_tests.rs
ioanachirca: Done.

src/vmm/src/lib.rs
ioanachirca: I put it in `vmm/version_map.rs`. The logic can be improved, I used the simplest way I could find to match Firecracker versions to snapshot data versions.

src/vmm/src/memory_dump.rs
ioanachirca: Added batching, now adjacent dirty pages are written in one call and we seek only at the start of a new batch.

src/snapshot/Cargo.toml
```diff
@@ -8,7 +8,7 @@ autobenches = false
 serde = ">=1.0.27"
 serde_derive = ">=1.0.27"
 bincode = "1.2.1"
-versionize = { git = "https://github.com/firecracker-microvm/versionize", tag = "v0.1.1" }
+versionize = { git = "https://github.com/firecracker-microvm/versionize", tag = "v0.1.0" }
```
acatangiu, line 11: Is this intentional?

sandreim: You can leave it like that, we will later propagate that upgrade to all deps that need it.

src/vmm/src/version_map.rs
```diff
@@ -0,0 +1,26 @@
...
+    /// Static instance used for handling microVM state versions.
+    pub static ref VERSION_MAP: VersionMap = {
+        VersionMap::new()
+    };
+
+    /// Static instance used for mapping Firecracker release version to
+    /// snapshot data format version.
+    pub static ref FC_VERSION_TO_SNAP_VERSION: HashMap<u16, u16> = {
+        let mut hm = HashMap::new();
+        hm.insert(21, 1);
```
sandreim, line 22: Why 21 ? this is already released. With the current plan 0.23 will be the first to support snapshots.

src/snapshot/Cargo.toml
ioanachirca: I had a compile problem when the `snapshot` crate was using `versionize 0.1.1` and the other crates `versionize 0.1.0` and I saw that there is another PR which bumps the version, so I took this shortcut 😅

Issue #1808, opened 2020-04-21
Add integration tests for network performance


goandrei:
Add a integration test which measures the overhead introduced by Firecracker.

dianpopa:
We have added performance tests for network latency and network tcp throughput: https://github.com/firecracker-microvm/firecracker/tree/master/tests/integration_tests/performance.

PR #1409, opened 2019-11-19
Add block/net device benchmark tests.

sandreim:
Signed-off-by: Andrei Sandu <sandreim@amazon.com>

## Reason for This PR

Benchmark test for net and block devices: #1097 .

## Description of Changes

- devtool: add -r option to map a ramdisk as /srv
- add env pytest marker to skip/run benchmark tests
- iperf3 based benchmark for net device
- fio based benchmark for block device
- add initial benchmarking docs

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria. Where there are two options, keep one.]`  
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [x] All commits in this PR are signed (`git commit -s`).
- [x] This PR is linked to an issue and the reason for this PR is
      clearly provided. 
- [x] The description of changes is clear and encompassing.
- [x] The required doc changes are included in this PR. Docs in scope are all `*.md` files
      located either in the repository root, or in the `docs/` directory.
- [x] Code-level documentation for touched code is included in this PR.
- [x] No API changes are included in this PR.
- [x] The changes in this PR have no user impact.
- [x] No new `unsafe` code has been added.

PR #1877, opened 2020-05-11
Network throughput integration test

goandrei:
Signed-off-by: Andrei Gorneanu <gorneanu@amazon.com>

## Reason for This PR

Addresses #1808 

## Description of Changes

This is work in progress.

The test is based on #1409. Changes introduced & details:
- fixed windows size(-w) : 10.
- variable bandwidth(-b) for both UDP(100,5000,10000) and TCP(>5000).
- variable packet size(-l) for TCP : 128, 256, 512, 1024, 1500.
- 2 directions : h2g(-R) and g2h; must add bidirectional in a future PR.
- skip the first seconds of TCP(-O)
- iperf threads pinning(-A) and VM threads pinned as well.
- multiple streams per test(-P); right now there are only 2 tests(1&2 streams). Must add tests for 4, 8, 32.
- added support for multiple runs and statistics. Introduced _class Statistics_ which holds the results for a single metric and computes its min,max,stdev, mean and median. May be used in the future for p99, p50 etc.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -18,6 +18,8 @@ import re
 import select
 import time
 
+from subprocess import run
+
 from retry import retry
 from retry.api import retry_call
 
@@ -105,6 +107,7 @@ class Microvm:
         self.machine_cfg = None
         self.vm = None
         self.vsock = None
+        self.cset_name = None
 
         # Initialize the logging subsystem.
         self.logging_thread = None
@@ -146,6 +149,9 @@ class Microvm:
         if self._memory_events_queue and not self._memory_events_queue.empty():
             raise mem_tools.MemoryUsageExceededException(
                 self._memory_events_queue.get())
+
+        if self.cset_name is not None:
+            self.remove_vcpus_shield()
 
     @property
     def api_session(self):
@@ -296,6 +302,33 @@ class Microvm:
         os.makedirs(self._path, exist_ok=True)
         os.makedirs(self._kernel_path, exist_ok=True)
         os.makedirs(self._fsfiles_path, exist_ok=True)
+
+    def remove_vcpus_shield(self):
+        """Remove the cset created for the microvm."""
+
+        #Don't check since this may fail if no shield exists.
+        cmd = 'cset set --destroy --set={}'.format(self.cset_name)
+        run(cmd, shell=True)
+
+    def shield_vm_vcpus(self, cset_name, cpus_list, numa_node):
+        """Create a cset and insert the fc_vcpus in there."""
+
+        self.cset_name = cset_name
+        self.remove_vcpus_shield()
+
+        # Create a new cset.
+        cpus_list = ','.join(map(str, cpus_list))
+        #TODO: use --cpu_exclusive
+        cmd = 'cset set --cpu={} --mem={} --set={}'.format(cpus_list, numa_node, cset_name)
+        utils.run_cmd(cmd)
+
+        # Add Firecracker's threads in the cset.
+        cmd = 'cset proc --toset={} --move --pid $(pidof firecracker) --threads -k'.format(cset_name)
+        utils.run_cmd(cmd)
+
+        # Move kernel threads from this set.
+        cmd = 'cset proc --kthread --fromset={} --toset=root --force'.format(cset_name)
+        utils.run_cmd(cmd)
 
     def spawn(self, create_logger=True):
         """Start a microVM as a daemon or in a screen session."""
--- a/tests/integration_tests/performance/test_net_tput.py
+++ b/tests/integration_tests/performance/test_net_tput.py
@@ -1,0 +1,282 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Tests the network throughput overhead added by Firecracker."""
+
+import json
+import time
+import statistics
+import socket
+from subprocess import run, PIPE
+import pytest
+
+from retry import retry
+
+import host_tools.network as net_tools
+from framework.statistics import Statistics
+from framework.utils import get_cpus_in_numa
+
+IPERF = 'iperf3'
+OUTPUT_JSON = 'test_net_tput_{}_iperfs.json'
+CSET_NAME = 'test_tput'
+
+class IperfClient:
+    # Flow direction and corresponding modes for iperf.
+    MODES = {
+        'h2g': '-R',
+        'g2h': ''
+    }
+    # Iteration duration of a single run in seconds.
+    ITERATION_DURATION = 1
+    # Number of test runs.
+    RUNS = 1
+    PINNING_CPU = 1
+
+    def __init__(self, protocol, host_ip, iperfs, microvm):
+        self.host_ip = host_ip
+        self.iperfs = iperfs
+        self.microvm = microvm
+        self.statistics = self.create_statistics()
+
+    def build_params_map(self):
+        """Build parameters map used for iperf."""
+        specific_params = self.get_specific_params()
+        params = {
+            'host_ip': self.host_ip,
+            'iperfs': self.iperfs,
+            'protocol': specific_params['protocol'],
+        }
+
+        for _ in range(self.RUNS):
+            for bw in specific_params['bandwidth']:
+                for mode in self.MODES.items():
+                    params['bw'] = bw
+                    params['mode'] = mode[1]
+
+                    yield params
+
+    def build_iperf_cmd(self, params):
+        """Build the iperf command applying given params."""
+
+        cmd = '{} -c {} -t{} -f MBytes -J -b {}M {} {} -A {} -P {}'.format(
+                IPERF,
+                params['host_ip'],
+                self.ITERATION_DURATION,
+                params['bw'],
+                params['mode'],
+                params['protocol'],
+                self.PINNING_CPU,
+                params['iperfs'],
+            )
+
+        return cmd
+
+    def spawn_iperf_guest(self, iperf_cmd):
+        """Start iperf in client mode and return resulting JSON."""
+        _, stdout, stderr = self.ssh_connection.execute_command(iperf_cmd)
+        assert stderr.read() == ''
+
+        out = stdout.read()
+        return json.loads(out)
+
+    def run_iperf_client(self, results):
+        """Run an iperf client."""
+
+        self.ssh_connection = net_tools.SSHConnection(self.microvm.ssh_config)
+
+        for params in self.build_params_map():
+            cmd = self.build_iperf_cmd(params)
+            iperf_out = self.spawn_iperf_guest(cmd)
+            result = self.parse_result(iperf_out, params)
+
+            results.append(result)
+        
+        return statistics
+
+
+class IperfUdpClient(IperfClient):
+    def __init__(self, host_ip, iperfs, microvm):
+        super().__init__('udp', host_ip, iperfs, microvm)
+
+    def create_statistics(self):
+        """Create a map which holds metric statistics registered for UDP."""
+        return {
+            'tput': Statistics(),
+            'lost': Statistics(),
+            'pps': Statistics(),
+            'jitter': Statistics(),
+        }
+
+    def get_specific_params(self):
+        """Get specific parameters for udp."""
+        return {
+            'protocol' : '-u',
+            'bandwidth' : [100, 5000, 10000],
+        }
+
+    def parse_result(self, result, params):
+        """Parse iperf UDP test output and return result JSON."""
+        test = 'udp-{}-bw{}'.format(
+                    'h2g' if params['mode'] == '-R' else 'g2h', 
+                    params['bw'], 
+                )
+
+        totals = result['end']['sum']
+        total_bytes = int(totals['bytes'])
+        total_packets = int(totals['packets'])
+        lost_packets = int(totals['lost_packets'])
+        duration = float(totals['seconds'])
+        jitter = round(float(totals['jitter_ms']), 2)
+        tput = round((total_bytes*8) / (1024*1024*duration), 2)
+        pps = round(total_packets / duration, 2)
+
+        if test not in self.statistics:
+            self.statistics[test] = self.create_statistics()
+
+        self.statistics[test]['tput'].add(tput)
+        self.statistics[test]['lost'].add(lost_packets)
+        self.statistics[test]['pps'].add(pps)
+        self.statistics[test]['jitter'].add(jitter)
+
+        return {
+            'test': test,
+            'tput': tput,
+            'lost': lost_packets,
+            'pps': pps,
+            'jitter': jitter,
+        }
+
+
+class IperfTcpClient(IperfClient):
+    WINDOW_SIZE_MB = 10
+    # Skip the first OMIT_TIME seconds when testing TCP
+    # to avoid slow start.
+    OMIT_TIME = 0
+    PACKET_SIZE_KB = [128, 256, 1024, 1500]
+
+    def __init__(self, host_ip, iperfs, microvm):
+        super().__init__('tcp', host_ip, iperfs, microvm)
+
+    def create_statistics(self):
+        """Create a map which holds metric statistics registered for TCP."""
+        return {
+            'tput': Statistics(),
+            'retransmits': Statistics(),
+        }
+
+    def get_specific_params(self):
+        """Create specific parameters for tcp."""
+        return {
+            'protocol' : '',
+            'bandwidth' : [5000, 10000, 20000, 0],
+        }
+
+    def build_params_map(self):
+        """Build a parameters list specific to tcp."""
+        for params in super().build_params_map():
+            for window_size in range(2):
+                for packet_size in self.PACKET_SIZE_KB:
+                    params['window_size'] = window_size
+                    params['packet_size'] = packet_size
+
+                    yield params
+
+    def build_iperf_cmd(self, params):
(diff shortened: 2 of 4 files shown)


tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,290 @@
...
+RUNS = 1
+
+IPERF = 'iperf3'
+ITERATION_DURATION = 1
+WINDOW_SIZE = 10
+
+# BW in Mbits.
+BANDWIDTHS = {
+    'udp': [100, 5000, 10000],
+    'tcp': [5000, 10000, 20000, 0]
```
raduiliescu, line 22: The only case I see where it makes sense for tcp bandwith parameter is if we take the cpu usage of firecracker from host or if we check the retransmissions. For example we want to have 1Gbps with 10% CPU usage and 0 retransmissions.
If you don't plan to get this tcp bandwith parameter doesn't make sense, since we want to get the peak throughput

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,290 @@
...
+from subprocess import run
+import pytest
+
+import host_tools.network as net_tools
+
+RUNS = 1
+
+IPERF = 'iperf3'
+ITERATION_DURATION = 1
+WINDOW_SIZE = 10
```
raduiliescu, line 17: While a fixed window size gets more predictability you should also run tests without windows size. 
Please also put a comment or rename it to WINDOW_SIZE_MB that this is in MB, also on iperf duration what it means.

tests/framework/microvm.py
```diff
@@ -280,6 +280,24 @@ def setup(self):
         os.makedirs(self._kernel_path, exist_ok=True)
```
raduiliescu, line 280: A general comment. This PR is ok, but you should add in a next one info about the CPU usage. You need to look from host on the CPU taken by Firecracker emulation.

tests/framework/microvm.py
```diff
@@ -280,6 +280,24 @@ def setup(self):
...
+        assert _p.stderr == ''
+
+        vcpus = _p.stdout.split('\n')
+        vcpus = vcpus[0:-1]
+
+        for index, vcpu in enumerate(vcpus):
+            columns = vcpu.split()
+            pin_cpu = cpu_count - index - 1
+            _p = utils.run_cmd('taskset -cp {} {}'.format(pin_cpu, columns[1]))
+
```
serban300, line 300: I have a couple of comments and questions here:

1. I think this method should be able to receive some parameters. Either the numa node on which it should pin the vcpus or the exact physical cpus. The problem is that jailer also does numa node pinning. And it pins the firecracker process to node 0 by default. If you always pin the vcpus to the last physical cpus (node > 0), this will create issues.
2. Can we do cpu shielding with `cset` instead of pinning ? It should lead to more predictible results.
3. Just to make sure we don't forget. We'll also have to setup NTP or PTP sync in the guest for more stable results. But we can do it in a future PR.

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,290 @@
...
+# Packet size used for TCP.
+PACKET_SIZE = [128, 256, 1024, 1500]
+
+# From server to client and reversed.
+MODES = ['h2g', 'g2h']
+# Corresponding MODES for iperf options.
+MODES_OPTIONS = {
+    'h2g': '-R',
+    'g2h': ''
+}
```
serban300, line 34: can we merge these variables ?

I'm not sure about the exact syntax, but 'm thinking about something like :

```
Modes: {
h2g: {
options: '-R'
}
...
```

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,290 @@
...
+
+PROTOCOL = ['udp', 'tcp']
+PROTOCOL_OPTIONS = {
+    'tcp': '',
+    'udp': '-u'
+}
```
serban300, line 40: Can we also merge this 2 variables ?

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,290 @@
...
+OMIT_TIME = 0
+
+PINNING_CPU = 1
+
+OFFSET_PORT = 23412
+
+OUTPUT_JSON = 'test_net_tput_{}_iperfs.json'
+
+
+class Statistics:
```
serban300, line 53: I would move this class in a separate file inside the `framework` folder.

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,290 @@
...
+
+    # Don't check the result of this command because it can fail if no iperf
+    # is running.
+    run(iperf_cmd, shell=True, check=False)
+
+    iperf_cmd = '{} {} -sD'.format(netns_cmd_prefix, IPERF)
+    run(iperf_cmd, shell=True, check=True)
+
+    # Wait for the iperf daemon to start.
+    time.sleep(2)
```
serban300, line 107: Can we do something more deterministic here ? Does the `run` method return any result that would help us identify the moment when the iper process started ?

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,290 @@
...
+                    )
+
+                    iperf_out = spawn_guest_iperf(microvm, cmd)
+                    result = parse_iperf_tcp_result(iperf_out, mode,
+                                                    bw, packet_size,
+                                                    tcp_statistics)
+                    results.append(result)
+                    print(result)
+
+    return tcp_statistics
```
serban300, line 236: These 2 methods look very similar. Can we have merge them somehow in order to have less duplicate code here ?

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,290 @@
...
+
+    results = []
+    _ = run_udp(microvm, host_ip, results, iperfs)
+    _ = run_tcp(microvm, host_ip, results, iperfs)
+
+    # TODO : compute statistics
+
+    output_name = OUTPUT_JSON.format(iperfs)
+    with open(output_name, 'w') as outfile:
+        json.dump(results, outfile, indent=4)
```
serban300, line 290: these 2 methods also look very similar. Can we have a method `test_network_tput(num_iperf_processes)` instead ? In order to have less duplicate code.

tests/integration_tests/performance/test_net_tput.py
goandrei: I am going to introduce CPU usage as a next step since even iperf3 offers some details about CPU usage, so it would be an easy change.

tests/integration_tests/performance/test_net_tput.py
goandrei: Good catch, I will update the name and run TCP without window size too.

tests/framework/microvm.py
goandrei: 1. Since we already have jailer which is capable of pinning the numa node I think it would be enought to have only pinning on the guest vcpus using the parameters and checking with the numa node used by jailer to make sure we avoid the actual error. What do you think?
2. Sure, this would much better.
3. I will add a TODO in the code too. Anyway, this would be very helpful in every integration test, I will try to do it as fast as I can.

tests/integration_tests/performance/test_net_tput.py
goandrei: Sure, we can do this

tests/integration_tests/performance/test_net_tput.py
goandrei: I let this here for now since I am not literally using its capabilities. Probably in the future it will be extended so yes, in the next PR it will have its own file.

tests/integration_tests/performance/test_net_tput.py
goandrei: My guess would be that the server is spawned as another process and then the parent returns its exit code and goes away, so I think we can't find out with `run`'s help, but I will check it out closely.

tests/integration_tests/performance/test_net_tput.py
goandrei: There are some differences in terms of parameters used, but maybe it could be done by creating some kind of common ground command with the common parameters and then add specific ones in each protocol's method.

tests/integration_tests/performance/test_net_tput.py
goandrei: for sure, will refactor this

tests/integration_tests/performance/test_net_tput.py
raduiliescu: You can look busy wait on netstat -lp to see if there a listen socket opened for the server port. If there is a socket opened than you are good to go. This is the correct method to solve this.

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+RUNS = 1
+
+IPERF = 'iperf3'
+
+# Iteration duration of a single run in seconds.
+ITERATION_DURATION = 1
+
+WINDOW_SIZE_MB = 10
```
raduiliescu, line 24: Not clear if you run tests without the specifying the windows size.

tests/integration_tests/performance/test_net_tput.py
raduiliescu: I don't think CPU usage from iperf would help. Here you need to take CPU usage of the emulation thread from Firecracker. You need to take this from the host.

tests/integration_tests/performance/test_net_tput.py
goandrei: I do, in the `run_iperf` method there is a for loop for window_size too which enables/disables the -w flag.

tests/integration_tests/performance/test_net_tput.py
goandrei: I tried this too, but the command has no entries when running it inside the docker image, so I couldn't see when the server started.

tests/integration_tests/performance/test_net_tput.py
goandrei: Actually I think it helps. The thing is that Firecracker uses 3 threads: firecracker, fc_api and fc_vcpu 0(supposing we spawn a microvm with one vcpu). The emulation usage is summed up with the vcpu usage in the firecracker thread, so I think the emulation cost would be computed in this way :  

`emulation_cpu_usage = firecracker_cpu_usage - iperf_cpu_usage`

That's why I was saying the iperf3 info about the cpu usage would be useful(of course if I'm not missing smth).

tests/integration_tests/performance/test_net_tput.py
raduiliescu: then try doing a connect to server in a loop and then close it

tests/integration_tests/performance/test_net_tput.py
raduiliescu: In fact is this way - firecracker_cpu_usage - iperf_cpu_usage = guest_kernel_cpu + other small stuff + emulation. If there is no technical way for getting the emulation thread usage is fine to assume that the small stuff is small and the guest kernel cpu usage is constant related to the iperf test. But here you can take using htop the cpu usage for emulation thread.

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,278 @@
...
+                        if protocol == 'tcp':
+                            result = parse_iperf_tcp_result(iperf_out, mode[0],
+                                                            bw, window_size,
+                                                            packet_size, run_statistics)
+                        else:
+                            result = parse_iperf_udp_result(iperf_out, mode[0],
+                                                            bw, window_size, 
+                                                            run_statistics)
+                        results.append(result)
+                        print(result)
```
serban300, line 200: I wouldn't add protocol-related logic here. I see `run_iperf` as a generic method that just runs the iperf client and takes care of the logic that's common to tcp and udp. 

We can have classes, for example IperfClient and IperfUdpClient/IperfTcpClient that extend IperfClient. Or we can just have methods. But I would separate the logic that's specific to Udp/Tcp from the common logic somehow. Since we also have different statistics for UDP and TCP, maybe it would make more sense to create different different classes for TCP and UDP that takes care both of the specific iperf client logic and also of the specific statistics logic.

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,278 @@
...
+        start, end = cpu_list.split('-')
+        start = int(start)
+        end = int(end)
+        if end - start + 1 < vcpus:
+            vcpus -= end - start + 1
+            cpus.extend(range(start, end + 1))        
+        else:
+            cpus.extend(range(start, start + vcpus))
+            vcpus = 0
+            break
```
serban300, line 228: this logic seems quite complicated. Maybe it would be easier to use the info in  `cat /sys/devices/system/node/node*/cpumap` or `cat /sys/devices/system/node/node*/cpulist` .

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,330 @@
...
+
+class IperfClient:
+    def __init__(self, protocol, host_ip, iperfs, microvm):
+        assert protocol == 'udp' or protocol == 'tcp'
+        self.protocol = protocol
+        self.host_ip = host_ip
+        self.iperfs = iperfs
+        self.microvm = microvm
+
+    def run_iperf_client(self):
```
serban300, line 67: In `IperfClient` the `run_iperf_client` method returns a map of parameters. In `IperfTcpClient` and `IperfUdpClient` it actually runs the iperf command. Also this methods take different parameters. This is a bit confusing. I would have `run_iperf_client` run the iperf command, as the name suggests, and have a different method for building the params map.

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+RUNS = 1
+
+IPERF = 'iperf3'
+
+# Iteration duration of a single run in seconds.
+ITERATION_DURATION = 1
+
+WINDOW_SIZE_MB = 10
```
serban300, line 25: This seems to be Tcp specific, so I would define it inside `IperfTcpClient`:

```
class IperfTcpClient(IperfClient):
    WINDOW_SIZE_MB = 10

    def __init__(self, host_ip, iperfs, microvm):
        super().__init__('tcp', host_ip, iperfs, microvm)
    ...
```

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,330 @@
...
+            'pps': pps,
+            'jitter': jitter,
+        }
+
+    def run_iperf_client(self, results):
+        """Build udp parameters upon the common ones."""
+        udp_statistics = {}
+
+        for _ in range(RUNS):
+            for params in super().run_iperf_client():
```
serban300, line 163: these `for`s are duplicate for `IperfTcpClient` and `IperfUdpClient`.  Can we have this logic inside `IperfClient` which is common for both ?

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,330 @@
...
+                cmd = self.build_iperf_cmd(params)
+                iperf_out = super().spawn_iperf_guest(cmd)
+                result = self.parse_result(iperf_out, params, udp_statistics)
```
serban300, line 166: These 3 lines are also almost completely duplicated for `IperfTcpClient` and `IperfUdpClient`. Can we have this logic inside `IperfClient` which is common for both ?

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,322 @@
...
+                    params['bw'] = bw
+                    params['mode'] = mode[1]
+
+                    yield params
+
+    def build_iperf_cmd(self, params):
+        """Build the iperf command applying given params."""
+        self.ssh_connection = net_tools.SSHConnection(self.microvm.ssh_config)
+
+        cmd = '{} -c {} -t{} -f MBytes -J -b {}M {} {} -A {} -P {}'.format(
```
serban300, line 61: There are a lot of hardcoded strings here, and the logic seems quite hard to read. I was wondering if it would make sense to have a builder class here and do something like:

```
cmd = iperf_cmd_builder.client(ip).iteration_duration(self.ITERATION_DURATION).bandwidth(params['bw'])....build()
```
Each function call would append to a string. For example `iperf_cmd_builder.client(ip)` would do something like `self.cmd += '-c ' + ip`. And build() would return the final command.

Maybe this would also help us remove the `build_params_map()`, since we could use the `iper_cmd_builder` as a temporary storage for the incomplete command, instead of the `params_map`

tests/integration_tests/performance/test_net_tput.py
```diff
@@ -0,0 +1,322 @@
...
+            for bw in specific_params['bandwidth']:
+                for mode in self.MODES.items():
+                    params['bw'] = bw
+                    params['mode'] = mode[1]
+
+                    yield params
+
+    def build_iperf_cmd(self, params):
+        """Build the iperf command applying given params."""
+        self.ssh_connection = net_tools.SSHConnection(self.microvm.ssh_config)
```
serban300, line 59: I think this shouldn't be here. This method should only build the command, not execute it.

iulianbarbu:
Close in favor of #2227.

Issue #1659, opened 2020-03-08
Add `at least one root block device` check in firecracker


KarthikNedunchezhiyan:
In the firecracker, we need to check whether the user has supplied at least one root block device and throw an appropriate error message if not provided.

In current flow, if the user has not supplied any root block device he will end up in kernel panic like below.

> [    0.145223] VFS: Cannot open root device "(null)" or unknown-block(0,0): error -6
> [    0.146475] Please append a correct "root=" boot option; here are the available partitions:
> [    0.147855] fe00           30720 vda 
> [    0.147857]  driver: virtio_blk
> [    0.149050] Kernel panic - not syncing: VFS: Unable to mount root fs on unknown-block(0,0)

But recent versions of firecracker support initrd. Users may handle rootfs mounting logic there 😕. Need to find a way to handle this elegantly.

acatangiu:
@KarthikNedunchezhiyan are you interested in contributing with a fix to this issue? We can discuss a strategy together if that's the case.

acatangiu:
Actually, thinking about this more, I think we don't even need such a check. Having a root device or not is an optional use-case-specific choice.

For example users of Firecracker which boot unikernels might not want any block device.

I believe the current Firecracker behavior of not enforcing such requirements is actually the correct one.

acatangiu:
Closing this. Anyone, feel free to open if needed.

PR #1883, opened 2020-05-12
Device-API functionality matrix & docs: fix code format

### LINKED ISSUE #1866: Device-API functionality matrix
Thanks for the effort everyone has put into making Firecracker available.

Context: 
We are building some [firecracker-kernel infrastructure](https://github.com/bbros-dev/firecracker-kernels/tree/develop).  We (eventually) aim to support any LTS kernel and allow for arbitrary (kernel) variants - currently `default` and `ignite` (`k8s`?), soon to land are `tiny` and `otu` (one trusted user), etc..  We have chosen to use [Yocto]() to manage the 'machine'/'board'/'hardware' and kernel configuration, and will work on a `firecracker` BSP layer. This allows us to layer the "hardware" and kernel configuration use cases in a way that should be reusable. 
Ideally we'd like to use some artifact name suffix to indicate Firecracker device support ( `*-bcknv`, `*-cnv`, etc.) for the  devices (`b=virtio-block` device, etc.).

With that context in mind:
It would help us and, we think, others too, if the project could add to the documentation a Device-API functionality matrix showing which devices are required for which API calls. 
Specifically we find ambiguity around which devices are required - even if not used - for what firecracker functionality.

Examples:

1. Guest `reboot` is a graceful way to shutdown.  Does this mean we can build the kernel using a BSP configuration layer without any keyboard - and then "only" loose the `SendCtlAltDel` API?
2. We can build a static nginx (busyboc, etc. etc.) into the initrd, and not use the rootfs. Does this mean we can build a BSP kernel without a block device and still expect firecracker to run? What API calls do we loose?

What we had in mind:

**O** (Optional): Device (column) not required for Firecracker functionality, other than the API action (row). If the device (column) is omitted from a kernel build the API action (row) is not available (returns `404`?).
**R** (Required): Device (column) required for Firecracker functionality, including the API action (row).

|API| virtio-net | virtio-block| virtio-vsock| serial console| keyboard |
|----|--------------|----------------|---------------|-------------------|--------------|
| `SendCtlAltDel`|    |     |               |                    | O             |

etc

Appreciate any feedback.

bbros-dev:
## Reason for This PR

Fixes #1866 

The Device-API functionality matrices show which devices are required for which API calls.
Specifically, this aims to remove ambiguity around which devices are required - even if not used - for what firecracker functionality.

## Description of Changes

The device-API relations are broken down by:

- Endpoint
- Input Schema
- Output Schema
- Instance Action

## License Acceptance

## PR Checklist

- ~~[x] Any newly added `unsafe` code is properly documented.~~
- ~~[x] Any API changes are reflected in `firecracker/swagger.yaml`.~~
- ~~[x] Any user-facing changes are mentioned in `CHANGELOG.md`.~~


bbros-dev:
On reflection

- The terms matrix/ces should really be table/s?

docs/api_requests/patch-block.md
```diff
@@ -1,6 +1,6 @@
 # Updating a Block Device
 
-Attached block devices require a PATCH /drives API call when the backing
+Attached block devices require a `PATCH /drives/{id}` API call when the backing
```
dianpopa, line 3: I would add a "docs: fix code format" in the commit title so that it points to a more specific destination in the project.

bbros-dev: Yes. I really should break this out to a separate patch.

docs/device-api.md
```diff
@@ -0,0 +1,106 @@
...
+| `metrics`                 |    O     |    O           |   O          |   O        |   O          |
+| `mmds`                    |    O     |    O           |   O          | **R**      |   O          |
+| `mmds/config`             |    O     |    O           |   O          | **R**      |   O          |
+| `network-interfaces/{id}` |    O     |    O           |   O          | **R**      |   O          |
+| `snapshot/create`         |    O     |    O           |   O          |   O        |   O          |
+| `snapshot/load`           |    O     |    O           |   O          |   O        |   O          |
+| `vm`                      |    O     |    O           |   O          |   O        |   O          |
+| `vsock`                   |    O     |    O           |   O          |   O        | **R**        |
+
+## Input Schema
```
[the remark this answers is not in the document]
bbros-dev: > What is the purpose of showing the properties?

Completeness.

> I see they all appear as required, no discriminant between them.

Yes, that is for more authoritative sources  ;) to change.

docs/api_requests/patch-block.md
dianpopa: As far as my concern, you can do it in this PR.

dianpopa: I may have not been clear enough. I was referring to the title in the commit not the title of the PR. You can change it by doing `git commit --amend` and then `git push -f`. You also need to do a rebase. Let me know if you need help.

bbros-dev:
OK it appears that block devices are optional by design:

https://github.com/firecracker-microvm/firecracker/issues/1659

This suggests the matrix value for block devices can be changed from `R` to `O`.
Not yet clear what the HTTP API returns.

Also don't see any `test_no_drives` scenario:
https://github.com/firecracker-microvm/firecracker/blob/master/tests/integration_tests/functional/test_drives.py

raduweiss:
> OK it appears that block devices are optional by design:
> 
> #1659
> 
> This suggests the matrix value for block devices can be changed from `R` to `O`.
> Not yet clear what the HTTP API returns.
> 
> Also don't see any `test_no_drives` scenario:
> https://github.com/firecracker-microvm/firecracker/blob/master/tests/integration_tests/functional/test_drives.py

Good catch, I'll open an issue for that.

docs/device-api.md
```diff
@@ -0,0 +1,106 @@
...
+
+**O** (Optional): The device (column) **is not required** for a Firecracker
+microVM to run. If the device (column) is omitted from a kernel build the API
+element (row) is **TBA**.
+
+**R** (Required): The device (column) **is required** for a Firecracker microVM
+to run. If the device (column) is omitted from a kernel build the API
+element (row) response is **TBA**.
+
+## API Endpoints
```
iulianbarbu, line 14: Can we mention that the API endpoints are described in the [firecracker swagger file](https://github.com/firecracker-microvm/firecracker/blob/master/src/api_server/swagger/firecracker.yaml)?

docs/device-api.md
```diff
@@ -0,0 +1,106 @@
+# Device-API Functionality
+
+The Device-API following functionality matrix indicates which devices are
+required for an API call to be usable.
+
+**O** (Optional): The device (column) **is not required** for a Firecracker
+microVM to run. If the device (column) is omitted from a kernel build the API
+element (row) is **TBA**.
```
iulianbarbu, line 8: The request for an API endpoint which lacks the needed Firecracker component for that API to complete returns a response with an error.

iulianbarbu (commented):
Hi, @bbros-dev ! I see that this PR was not addressed for a while. Are you still interested in merging this documentation?

docs/device-api.md
```diff
@@ -0,0 +1,106 @@
...
+| `snapshot/create`         |    O     |    O           |   O          |   O        |   O          |
+| `snapshot/load`           |    O     |    O           |   O          |   O        |   O          |
+| `vm`                      |    O     |    O           |   O          |   O        |   O          |
+| `vsock`                   |    O     |    O           |   O          |   O        | **R**        |
+
+## Input Schema
+
+| Schema                    | Property              | keyboard | serial console | virtio-block | virtio-net | virtio-vsock |
+|---------------------------|-----------------------|:--------:|:--------------:|:------------:|:----------:|:------------:|
+| `BootSource`              | boot_args             |   O      |   O            |   O          |   O        |   O          |
```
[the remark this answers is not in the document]
bbros-dev: Utterly confused by that description.

docs/device-api.md
```diff
@@ -0,0 +1,106 @@
...
+|                           | ops                   |   O      |   O            | **R**        | **R**      |   O          |
+| `TokenBucket`             | one_time_burst        |   O      |   O            | **R**        | **R**      |   O          |
+|                           | refill_time           |   O      |   O            | **R**        | **R**      |   O          |
+|                           | size                  |   O      |   O            | **R**        | **R**      |   O          |
+| `Vm`                      | state                 |   O      |   O            |   O          |   O        |   O          |
+| `Vsock`                   | guest_cid             |   O      |   O            |   O          |   O        | **R**        |
+|                           | uds_path              |   O      |   O            |   O          |   O        | **R**        |
+|                           | vsock_id              |   O      |   O            |   O          |   O        | **R**        |
+
+## Output Schema
```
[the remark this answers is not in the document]
bbros-dev: Ditto.

docs/device-api.md
bbros-dev: Sure, but it seems extraneous: Or at least I generally don't care how an API is implemented, and I don't want to think I have to understand swaggers peculiarities to know how to use this API.

docs/device-api.md
bbros-dev: OK. However, we have not used the word component on this page so that statement will need to be reworded, or the definition of 'component' added somewhere herein.

docs/device-api.md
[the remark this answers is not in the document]
iulianbarbu: > IMO that aid is best given by being explicit and not having them decode some summary/abbreviation conventions.

I see. My concern here is that we are required to keep APIs, input and outputs schemas duplicated in both swagger file and Device-API matrix. Users will eventually want to checkout the swagger file (and understand the semantics of this file) if they are interested in the source of truth for APIs and schema definitions. When users want to understand what devices are required for specific APIs endpoints or schemas, they need to look it up in the Device-API matrix. Having an abbreviation convention means that some API endpoint, input or output schema will fall under the wildcard, which means that they will not explicitly appear in the document, which might cause confusion for end users. Adding a backing definition in the document for the wildcard will address this issue.

In the end, I would like to say that I do not have a strong preference for my suggestion. It was just a callout for getting more comments and see if it is something which you or others would prefer instead of the long tables form. Feel free to choose the option you think it is better for end users, I am OK with both.

docs/device-api.md
```diff
@@ -0,0 +1,109 @@
...
+|                         | id                |   O      |   O            |   O          |   O        |   O          |
+|                         | state             |   O      |   O            |   O          |   O        |   O          |
+|                         | vmm_version       |   O      |   O            |   O          |   O        |   O          |
+| `MachineConfiguration`  | cpu_template      |   O      |   O            |   O          |   O        |   O          |
+|                         | ht_enabled        |   O      |   O            |   O          |   O        |   O          |
+|                         | mem_size_mib      |   O      |   O            |   O          |   O        |   O          |
+|                         | track_dirty_pages |   O      |   O            |   O          |   O        |   O          |
+|                         | vcpu_count        |   O      |   O            |   O          |   O        |   O          |
+
+## Instance Actions
```
iulianbarbu, line 103: For consistency sake, I would add again the swagger file reference here.

docs/device-api.md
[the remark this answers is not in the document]
bbros-dev: > My concern here is that we are required to keep APIs, input and outputs schemas duplicated in both swagger file and Device-API matrix.

Fully agree and share that concern.  However we are documenting the connection between two distinct systems: Swagger API and Firecracker uVM, so it seems unavoidable that some hand maintained document is required.
IIUC, Firecracker's focus means I'd expect its device model and even API to be very slowly changing, if at all once v1 is released. so changes here should be infrequent.

In case it turns out there are frequent of changes:
We might (eventually) be able to script this table from some Firecracker JSON master definition file (is there such a thing?) and the Swagger API definition?  Updating this document would be part of the release process.
But that is not possible now, and may require adding linking data to their Swagger or JSON master file.

I'm having trouble understanding the conventions you suggest and how they would work. 
Maybe they can be made as a PR once this lands?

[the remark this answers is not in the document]
iulianbarbu: We do not have frequent changes in API area, but if we do, we need to update the release process to sync the Device-API matrix. We can further improve this documentation in following PRs, if needed. Will resolve this comment for now.

docs/device-api.md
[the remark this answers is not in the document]
iulianbarbu: Please do not resolve the comment until we reach a resolution. For example, I do not see any update for this comment, and no response as well.

iulianbarbu (commented):
Hi, @bbros-dev ! From my point of view, this PR is almost done. See my comments below. Let me know how do you plan to address them.

docs/device-api.md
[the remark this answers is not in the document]
bbros-dev: I've made the change just haven't pushed the change.

docs/device-api.md
bbros-dev: Done. Incoming.

docs/device-api.md
bbros-dev: Not sure what to place here in the stead of **TBA**.

iulianbarbu: Please consider the following suggestion. WDYT?
```suggestion
element request (row) returns with an HTTP response containing a corresponding error message.
```

iulianbarbu: Just another thing which I realized, since the table refers to the device requirement for APIs to succeed, is that `/mmds` and `/mmds/config` does not have any hard requirement on the existence of a configured network device. A network device is relevant only for MMDS operation, but not for its configuration. With this in mind, I think it makes sense to change the `virtio-net` column value corresponding to both `/mmds` and `/mmds/config` from `**R**` to `**O**`.

bbros-dev: I've not used MMDS in anger.  My instinct is that configuring something and not using it sounds like an unusual corner case?
Is there a mainstream use case where MMDS would be configured for a uVM that does not have a network device, but that MMDS configuration is then used on a uVM with a network device?

bbros-dev: > request (row) returns with an HTTP response containing a corresponding error message.

Is there a link to the error messages we can use, or should that be documented here?

iulianbarbu: > Is there a mainstream use case where MMDS would be configured for a uVM that does not have a network device, but that MMDS configuration is then used on a uVM with a network device?
I am not aware of any mainstream case as the one you've described.

>  My instinct is that configuring something and not using it sounds like an unusual corner case.
I agree. The misalignment which might cause customer confusion comes from the fact that the MMDS configuration requirements does not correspond with MMDS operation requirements. I can not say this is a required design choice or simply an arbitrary happening due to unclear requirements. My understanding is that it is not a blocker of any sort for Firecracker usage and configuration since we haven't had any issues with it so far. I also believe that with the advent of the Device-API matrix it will be easier to think for the future in terms of API refinement in this area. Feel free to open any issue on this topic.

iulianbarbu: I do not think we need to handle this specific occurrence of TBA as in what errors would an end user get if the optional requirements are not satisfied. Here we should say that the API endpoint call will succeed if the requirement is not satisfied. I would add an expectancy in terms of returned HTTP responses only for the `Required` section.

> Is that a 404 response?

We do not have yet an unified interface for expected responses if users do not respect mandatory requirements of an API endpoint. At the moment, we can work backwards from two API endpoints: `drives/{id}` and `network-interfaces/{id}`. Only these two have a hard mandatory requirement on the existence of a configured virtio-block/virtio-net device with the same `id` used by the API call.

* `drives/{id}` results in 400 - BadRequest - HTTP response if the block device identified by `id` was not previously configured.
* `network-interfaces/{id}` API call will act as a no-op if the network device identified by `id` was not previously configured.

> Is there a link to the error messages we can use, or should that be documented here?

We do not have a documentation for the error messages returned in response to `drives/{id}` or `network-interfaces/{id}`. I would suggest you to mention the returned errors response code for `drives/{id}` and mention also that `network-interfaces/{id}` is considered as no-op in case of error. This is not unified across devices. Feel free to open an issue for this as well.

bbros-dev: Done.  What about `vsock`?

iulianbarbu: @bbros-dev , please see my comment updates above.

> What about vsock?

`vsock` API endpoint call does not require the pre-existance of a configured `vsock` device, so we do not need **R** for virtio-vsock section of the `vsock` API endpoint. If we have it **O** then we do not need to set an expectancy for not satisfying the API endpoint mandatory requirements, since there are not any.

> Also `SendCtlAltDel` when no keyboard is present?

The keyboard device is always present for a Firecracker uVM. On a separate note, this API endpoint is only available on `x86_64`. Firecracker at the moment has arch dependencies for this API endpoint and for `CreateSnapshot`/`LoadSnapshot` as well. I think this is another dependency we want to capture in the document, but with this information the document scope will change from `Device-API matrix` to `Firecracker API dependency matrix`. Please let me know what you think.

bbros-dev: >  I think this is another dependency we want to capture in the document, but with this information the document scope will change from  Device-API matrix to Firecracker API dependency matrix. Please let me know what you think.

Agree. Perhaps an separate issue to track updating this with ARM64 specifics once that architecture is stabilized?

iulianbarbu: Yes, we can definitely work with another issue for tracking Firecracker API arch dependency dimension. Feel free to open it.

bbros-dev: Okay, so we are done?

iulianbarbu (commented):
The PR looks good to me. I think it would be best to squash all the commits into a single one.
I'll approve afterwards.

bbros-dev:
> The PR looks good to me. I think it would be best to squash all the commits into a single one and to rebase as well.
> I'll approve afterwards.

@iulianbarbu , Thanks for your help getting this PR into shape. Much appreciated.

iulianbarbu:
Hi, @bbros-dev ! I believe you closed the PR by mistake. I'll reopen it.

iulianbarbu (approved):

docs/device-api.md
```diff
@@ -0,0 +1,124 @@
...
+|                           | iface_id              |    O     |       O        |      O       |   **R**    |      O       |
+|                           | rx_rate_limiter       |    O     |       O        |      O       |   **R**    |      O       |
+|                           | tx_rate_limiter       |    O     |       O        |      O       |   **R**    |      O       |
+| `PartialDrive`            | drive_id              |    O     |       O        |    **R**     |     O      |      O       |
+|                           | path_on_host          |    O     |       O        |    **R**     |     O      |      O       |
+| `PartialNetworkInterface` | iface_id              |    O     |       O        |      O       |   **R**    |      O       |
+|                           | rx_rate_limiter       |    O     |       O        |      O       |   **R**    |      O       |
+|                           | tx_rate_limiter       |    O     |       O        |      O       |   **R**    |      O       |
+| `RateLimiter`             | bandwidth             |    O     |       O        |    **R**     |   **R**    |      O       |
```
dianpopa, line 87: IMO this should not be placed in the matrix for 2 reasons:
* RateLimiter` is not an API request that can be issued on its own. It can be used in either Net or Drive specific API requests.
* Someone reading this matrix can think that if they want to use the rate limiter configuration they would need to enable both virtio net and virtio block drivers which is not true.
My proposal is to remove this from here.

docs/device-api.md
```diff
@@ -0,0 +1,124 @@
...
+|                           | ops                   |    O     |       O        |    **R**     |   **R**    |      O       |
+| `TokenBucket`             | one_time_burst        |    O     |       O        |    **R**     |   **R**    |      O       |
```
dianpopa, line 89: The same case as with the "RateLimiter".

dianpopa (commented):
LGTM! Just a little observation regarding "RateLimiter" and "TokenBucket" configs.

bbros-dev:
Ok, who else is doing a review and please can they:
- Do it now
- Indicate when it will be done.

Once all the reviewers indicate they have had had their say, I'll address any changes.

This drip-drip process is not pleasant.

dianpopa:
> Ok, who else is doing a review (@aghecenco @dianpopa @raduweiss ?) and please can they either:
> 
> * Do it now, and say when they are completed.
> * Indicate when it will be done, and likewise explicitly say when they are done.
> 
> Once all the reviewers indicate they have had had their say, and say they have no more comments, I'll address any changes.
> 
> This drip-drip process is not pleasant.

Hello! In order for a PR to be merged you need at least 2 approving reviews (me and @iulianbarbu will be those 2). The PR was already approved by Iulian. After my comments will be addressed I will also approve the PR and it will then be merged.

docs/device-api.md
bbros-dev: > RateLimiter` is not an API request that can be issued on its own. It can be used in either Net or Drive specific API requests

Perhaps I misunderstand your objection? However the whole point of this table is to show which devices are required for and API request to 'work'.  Removing these from the table would suggest they can be expected to work regardless of the devices present.

Given the RateLimiter API should 'work with either device, a better approach would be to repeat the entry.  One row showing net device is required and block device optional.
The other entry showing the block device is required and the net device optional.

Read together, they both suggest that API will work if both devices are present.

Are their any other such API calls?

dianpopa: As you said, "the whole point of this table is to show which devices are required for an API request to work" but RateLimiter does not have an API endpoint, it can't be issued as a standalone API request. Am I mistaken?

bbros-dev: > ... it can't be issued as a standalone API request.

But it is a API request?

docs/device-api.md
bbros-dev: Again if this is not an API request it should be remove.  If it is an API request, whether standalone or not, it should stay.

If it is an API request do you agree breaking it into two rows as for RateLimiter communicates the fact either or both devices are required?

dianpopa (dismissed):

dianpopa:
You need to rebase your branch (hopefully you can do that without push -f so that the approvals are not lost) and then you can merge your PR. Thanks!

docs/device-api.md
dianpopa: I can now understand better your view on the scope of matrix, I guess I understood its scope differently.  We can leave this as it is.

bbros-dev:
@dianpopa and @georgepisaltu is that a more accurate summary of the `RateLimiter` input schema and device relations?

bbros-dev:
@dianpopa and @georgepisaltu is that a more accurate summary of the TokenBucket input schema and device relations?

dianpopa (dismissed):

iulianbarbu (dismissed):

iulianbarbu:
Hi, @bbros-dev ! Please squash the commits into a single one. We should add this new document from one go. The last three commits change the first commit, which introduce bloated history. Me and @dianpopa will approve again shortly after.

dianpopa (approved):

iulianbarbu (approved):

iulianbarbu:
We need to rebase again. 😞 @bbros-dev , I think this can be easily done directly from this PR view, by issuing an "Update branch" after you deactivate branch protection on your branch against other users merges (not sure though if you have it by default or if it was explicitly set). You can either do this or rebase again.

PR #1891, opened 2020-05-13
virtio-queue: use map_err for error handling

xiekeyang:
Use `err_map` to handle error cases when getting desc variable, to
instead `match` statement.

I feel we need not the line:
```rust
            Err(_) => {
                ...
                return None;
            }
```
for check has been done, and if memory broken happens, just throw out error message.

Signed-off-by: keyangxie <keyang.xie@gmail.com>

## Reason for This PR

## Description of Changes

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/queue.rs
+++ b/src/devices/src/virtio/queue.rs
@@ -77,14 +77,13 @@ impl<'a> DescriptorChain<'a> {
         mem.checked_offset(desc_head, 16)?;
 
         // These reads can't fail unless Guest memory is hopelessly broken.
-        let desc = match mem.read_obj::<Descriptor>(desc_head) {
-            Ok(ret) => ret,
-            Err(_) => {
+        let desc = mem
+            .read_obj::<Descriptor>(desc_head)
+            .map_err(|_| {
                 // TODO log address
                 error!("Failed to read from memory");
-                return None;
-            }
-        };
+            })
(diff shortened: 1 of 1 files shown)


xiekeyang:
@aghecenco test coverage failed. I would update the target value later if you think this pr is OK. thanks.

src/devices/src/virtio/queue.rs
```diff
@@ -77,14 +77,13 @@ impl<'a> DescriptorChain<'a> {
...
+        let desc = mem
+            .read_obj::<Descriptor>(desc_head)
+            .map_err(|_| {
                 // TODO log address
                 error!("Failed to read from memory");
-                return None;
-            }
-        };
+            })
+            .unwrap();
```
andreeaflorescu, line 86: We cannot use unwrap here because the contract of this function is that it returns Some(desc) if a descriptor exists, or None otherwise.

xiekeyang:
@andreeaflorescu , thanks for your reply.

I had thought `None` was unreachable case because `reads can't fail unless Guest memory is hopelessly broken`, so `unwrap` is reliable and next programme is safe.

But if we have to keep the contract then there might be no more graceful statement to instead `match`.

I'd close this pr, as I'm still a little confused how to setup the test case for `return None`.

PR #1892, opened 2020-05-13
Add some infrastructure for snapshot integration test.

### LINKED ISSUE #1849: Snapshot create/restore integ tests infrastructure

sandreim:
## Reason for This PR
Fixes #1849.

## Description of Changes
Adds the flexibility required to generate a test matrix for save/restore tests using multiple snapshot images and firecracker binaries.

This PR introduces a new way for spinning up microvms in integration tests. Firstly, it does not rely on the hidden logic of fixtures to create and configure the microvms. Secondly, it defines a decoupled view of test resources, machine configurations, microvm configuration and implements  simple APIs that manage these. And thirdly, it allows a single test function to be run multiple times in a different setup (machine, disks, kernels, snapshots).

Layers and components:
- Resources layer components: **ArtifactCollection**, **Artifact**, **ArtifactSet**, **MicrovmBuilder**
- Test layer components: **TestContext**, **TestMatrix**

The resources exposed by these APIs are stored separately(`ci-artifacts` folder in S3 bucket) from the other resources used in integration tests.

The snapshot_pause_resume.py test was updated and now runs pause/resume on these 3 distinct microvm configurations: [2 vcpu 256MB](https://s3.amazonaws.com/spec.ccfc.min/ci-artifacts/microvms/2vcpu_256mb.json), [2 vcpu 512MB](https://s3.amazonaws.com/spec.ccfc.min/ci-artifacts/microvms/2vcpu_512mb.json), [2 vcpu 1024MB](https://s3.amazonaws.com/spec.ccfc.min/ci-artifacts/microvms/2vcpu_1024mb.json) in 8 seconds.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -117,7 +117,7 @@ if os.geteuid() != 0:
     raise PermissionError("Test session needs to be run as root.")
 
 
-def test_images_s3_bucket():
+def _test_images_s3_bucket():
     """Auxiliary function for getting this session's bucket name."""
     return os.environ.get(
         ENV_TEST_IMAGES_S3_BUCKET,
@@ -125,7 +125,7 @@ def test_images_s3_bucket():
     )
 
 
-MICROVM_S3_FETCHER = MicrovmImageS3Fetcher(test_images_s3_bucket())
+MICROVM_S3_FETCHER = MicrovmImageS3Fetcher(_test_images_s3_bucket())
 
 
 def init_microvm(root_path, bin_cloner_path):
--- a/tests/framework/artifacts.py
+++ b/tests/framework/artifacts.py
@@ -80,9 +80,6 @@ class ArtifactCollection:
     ARTIFACTS_MICROVMS = '/microvms'
     ARTIFACTS_SNAPSHOTS = '/snapshots/' + platform.machine()
 
-    s3 = None
-    artifacts_bucket = None
-
     def __init__(
         self,
         artifacts_bucket
@@ -92,44 +89,58 @@ class ArtifactCollection:
         self.s3 = boto3.resource('s3', config=config)
         self.artifacts_bucket = artifacts_bucket
 
-    def microvms(self):
+    def _fetch_artifacts(self,
+                         artifact_root,
+                         artifact_dir,
+                         artifact_ext,
+                         artifact_type,
+                         artifact_class,
+                         keyword=None):
+        bucket = self.s3.Bucket(self.artifacts_bucket)
+        artifacts = []
+        prefix = artifact_root + artifact_dir
+        files = bucket.objects.filter(Prefix=prefix)
+        for file in files:
+            if (
+                # Filter by extensions.
+                file.key.endswith(artifact_ext)
+                # Filter by userprovided keyword if any.
+                and (keyword is None or keyword in file.key)
+            ):
+                artifacts.append(artifact_class(bucket,
+                                                file.key,
+                                                type=artifact_type))
+        return artifacts
+
+    def microvms(self, keyword=None):
         """Return all microvms artifacts for the current arch."""
-        bucket = self.s3.Bucket(self.artifacts_bucket)
-        microvm_artifacts = []
-        microvm_prefix = self.ARTIFACTS_ROOT + self.ARTIFACTS_MICROVMS
-        microvms = bucket.objects.filter(Prefix=microvm_prefix)
-        for microvm in microvms:
-            if microvm.key.endswith(MICROVM_CONFIG_EXTENSION):
-                microvm_artifacts.append(Artifact(bucket,
-                                                  microvm.key,
-                                                  type="microvm"))
+        return self._fetch_artifacts(
+            self.ARTIFACTS_ROOT,
+            self.ARTIFACTS_MICROVMS,
+            MICROVM_CONFIG_EXTENSION,
+            "microvm",
+            Artifact,
+            keyword=keyword
+        )
 
-        return microvm_artifacts
-
-    def kernels(self):
+    def kernels(self, keyword=None):
         """Return all microvms kernels for the current arch."""
-        bucket = self.s3.Bucket(self.artifacts_bucket)
-        kernel_artifacts = []
-        kernel_prefix = self.ARTIFACTS_ROOT + self.ARTIFACTS_KERNELS
-        kernels = bucket.objects.filter(Prefix=kernel_prefix)
-        for kernel in kernels:
-            if kernel.key.endswith(MICROVM_KERNEL_EXTENSION):
-                kernel_artifacts.append(Artifact(bucket,
-                                                 kernel.key,
-                                                 type="kernel"))
+        return self._fetch_artifacts(
+            self.ARTIFACTS_ROOT,
+            self.ARTIFACTS_KERNELS,
+            MICROVM_KERNEL_EXTENSION,
+            "kernel",
+            Artifact,
+            keyword=keyword
+        )
 
-        return kernel_artifacts
-
-    def disks(self):
+    def disks(self, keyword=None):
         """Return all disk artifacts for the current arch."""
-        bucket = self.s3.Bucket(self.artifacts_bucket)
-        disk_artifacts = []
-        disk_prefix = self.ARTIFACTS_ROOT + self.ARTIFACTS_DISKS
-        disks = bucket.objects.filter(Prefix=disk_prefix)
-        for disk in disks:
-            if disk.key.endswith(MICROVM_DISK_EXTENSION):
-                disk_artifacts.append(DiskArtifact(bucket,
-                                                   disk.key,
-                                                   type="disk"))
-
-        return disk_artifacts
+        return self._fetch_artifacts(
+            self.ARTIFACTS_ROOT,
+            self.ARTIFACTS_DISKS,
+            MICROVM_DISK_EXTENSION,
+            "disk",
+            DiskArtifact,
+            keyword=keyword
+        )
--- a/tests/framework/artifacts.py
+++ b/tests/framework/artifacts.py
@@ -6,42 +6,51 @@ import os
 import platform
 import re
 
+from enum import Enum
 from shutil import copyfile
 from typing import List
-from os import path
 from pathlib import Path
+from os import path
 
 import boto3
 import botocore.client
 
 
-MICROVM_CONFIG_EXTENSION = ".json"
-MICROVM_KERNEL_EXTENSION = ".bin"
-MICROVM_DISK_EXTENSION = ".ext4"
+class ArtifactType(Enum):
+    """Supported artifact types."""
+
+    MICROVM = "microvm"
+    KERNEL = "kernel"
+    DISK = "disk"
+    SSH_KEY = "ssh_key"
+    MISC = "misc"
 
 
 class Artifact:
     """A generic artifact manipulation class."""
 
-    bucket = None
-    key = None
-    local_folder = None
-
-    def __init__(self, bucket, key, type="misc"):
+    def __init__(self, bucket, key, type=ArtifactType.MISC):
         """Initialize bucket, key and type."""
         self.bucket = bucket
         self.key = key
-        self.type = type
+        self.local_folder = None
+        self._type = type
+
+    @property
+    def type(self):
+        """Return the artifact type."""
+        return self._type
 
     def name(self):
         """Return the artifact name."""
-        return self.key.rsplit("/", 1)[1]
+        return Path(self.key).name
 
     def local_dir(self):
         """Return the directory containing the downloaded artifact."""
+        assert self.local_folder is not None
         return "{}/{}/{}".format(
             self.local_folder,
-            self.type,
+            self.type.value,
             platform.machine(),
         )
 
@@ -66,12 +75,17 @@ class DiskArtifact(Artifact):
 
     def ssh_key(self):
         """Return a ssh key artifact."""
-        key_file_path = self.key.rsplit(".", 1)[0] + '.id_rsa'
-        return Artifact(self.bucket, key_file_path, type="ssh_key")
+        # Replace extension.
+        key_file_path = str(Path(self.key).with_suffix('.id_rsa'))
+        return Artifact(self.bucket, key_file_path, type=ArtifactType.SSH_KEY)
 
 
 class ArtifactCollection:
     """Provides easy access to different artifact types."""
+
+    MICROVM_CONFIG_EXTENSION = ".json"
+    MICROVM_KERNEL_EXTENSION = ".bin"
+    MICROVM_DISK_EXTENSION = ".ext4"
 
     # S3 bucket structure.
     ARTIFACTS_ROOT = 'ci-artifacts'
@@ -113,34 +127,67 @@ class ArtifactCollection:
         return artifacts
 
     def microvms(self, keyword=None):
-        """Return all microvms artifacts for the current arch."""
+        """Return microvms artifacts for the current arch."""
         return self._fetch_artifacts(
-            self.ARTIFACTS_ROOT,
-            self.ARTIFACTS_MICROVMS,
-            MICROVM_CONFIG_EXTENSION,
-            "microvm",
+            ArtifactCollection.ARTIFACTS_ROOT,
+            ArtifactCollection.ARTIFACTS_MICROVMS,
+            ArtifactCollection.MICROVM_CONFIG_EXTENSION,
+            ArtifactType.MICROVM,
             Artifact,
             keyword=keyword
         )
 
     def kernels(self, keyword=None):
-        """Return all microvms kernels for the current arch."""
+        """Return kernel artifacts for the current arch."""
         return self._fetch_artifacts(
-            self.ARTIFACTS_ROOT,
-            self.ARTIFACTS_KERNELS,
-            MICROVM_KERNEL_EXTENSION,
-            "kernel",
+            ArtifactCollection.ARTIFACTS_ROOT,
+            ArtifactCollection.ARTIFACTS_KERNELS,
+            ArtifactCollection.MICROVM_KERNEL_EXTENSION,
+            ArtifactType.KERNEL,
             Artifact,
             keyword=keyword
         )
 
     def disks(self, keyword=None):
-        """Return all disk artifacts for the current arch."""
+        """Return disk artifacts for the current arch."""
         return self._fetch_artifacts(
-            self.ARTIFACTS_ROOT,
-            self.ARTIFACTS_DISKS,
-            MICROVM_DISK_EXTENSION,
-            "disk",
+            ArtifactCollection.ARTIFACTS_ROOT,
+            ArtifactCollection.ARTIFACTS_DISKS,
+            ArtifactCollection.MICROVM_DISK_EXTENSION,
+            ArtifactType.DISK,
             DiskArtifact,
             keyword=keyword
         )
+
+
+class ArtifactSet:
+    """Manages a set of artifacts with the same type."""
+
+    def __init__(self, artifacts):
+        """Initialize type and artifact array."""
+        assert len(artifacts) > 0
+        self._type = artifacts[0].type
+        self._artifacts = []
+        self.add(artifacts)
+
+    def add(self, artifacts):
+        """Add artifacts to set."""
+        assert len(artifacts) > 0
+        for artifact in artifacts:
+            assert artifact.type == self._type
+            self._artifacts.append(artifact)
+
+    def artifact(self, index):
+        """Return the current artifact in the list."""
+        assert len(self) > 0
+        assert index < len(self)
+        return self._artifacts[index]
+
+    @property
+    def artifacts(self):
+        """Return the artifacts array."""
+        return self._artifacts
+
+    def __len__(self):
+        """Return the artifacts array len."""
+        return len(self._artifacts)
--- a/tests/framework/artifacts.py
+++ b/tests/framework/artifacts.py
@@ -4,13 +4,9 @@
 
 import os
 import platform
-import re
 
 from enum import Enum
-from shutil import copyfile
-from typing import List
 from pathlib import Path
-from os import path
 
 import boto3
 import botocore.client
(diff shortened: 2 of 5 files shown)


tests/integration_tests/functional/test_pause_resume.py
```diff
@@ -1,65 +1,108 @@
...
+    artifacts = ArtifactCollection(test_images_s3_bucket())
+
+    microvm_artifacts = artifacts.microvms()
+    kernel_artifacts = artifacts.kernels()
+    disk_artifacts = artifacts.disks()
+
+    # Create a test matrix. Push logger and network as context variables.
+    test_matrix = TestMatrix(context={
+            'network_config': network_config,
+            'logger': logger
```
gc-plp, line 98: I think that you could give a name to the logger here. What do you think?
You can _name_ logger instances by passing a string to the `getLogger` call, in order to have all log entries marked with the given name.

sandreim: done.

tests/framework/artifacts.py
```diff
@@ -0,0 +1,146 @@
...
+MICROVM_KERNEL_EXTENSION = ".bin"
+MICROVM_DISK_EXTENSION = ".ext4"
+
+
+class Artifact:
+    """A generic artifact manipulation class."""
+
+    bucket = None
+    key = None
+    local_folder = None
```
gc-plp, line 28: If `local_folder` is left initialized to `None`, an immediate call to `local_dir()` after instantiation would yield `None` as path.

It would be good to move the variables inside the `__init__` function and perhaps check for uninitialized values for `local_folder`. If `local_dir()` and `local_path()` are not designed to be public, the functions should also be marked as private.

tests/framework/artifacts.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def __init__(self, bucket, key, type="misc"):
```
gc-plp, line 30: Could the artifact `type` be part of an enum, so that we explicitly allow only a predefined set of types?

tests/framework/artifacts.py
```diff
@@ -0,0 +1,146 @@
...
+        """Initialize bucket, key and type."""
+        self.bucket = bucket
+        self.key = key
+        self.type = type
+
+    def name(self):
+        """Return the artifact name."""
+        return self.key.rsplit("/", 1)[1]
```
gc-plp, line 38: Instead of `rsplit`, you could also use `Path(self.key).name`

tests/framework/artifacts.py
```diff
@@ -0,0 +1,146 @@
...
+            self.name()
+        )
+
+
+class DiskArtifact(Artifact):
+    """Specializes the generic artifact."""
+
+    def ssh_key(self):
+        """Return a ssh key artifact."""
+        key_file_path = self.key.rsplit(".", 1)[0] + '.id_rsa'
```
gc-plp, line 69: Same proposal here, as above. You could use `Path` to make the purpose of statement clearer.

tests/framework/artifacts.py
sandreim: I think this is a good idea.

tests/framework/artifacts.py
```diff
@@ -0,0 +1,146 @@
...
+            ):
+                artifacts.append(artifact_class(bucket,
+                                                file.key,
+                                                type=artifact_type))
+        return artifacts
+
+    def microvms(self, keyword=None):
+        """Return all microvms artifacts for the current arch."""
+        return self._fetch_artifacts(
+            self.ARTIFACTS_ROOT,
```
gc-plp, line 118: Instead of `self` you can also reference the variables by `ArtifactCollection.ARTIFACTS_ROOT`. Using it this way is clearer that you're referencing a class variable that is read only.

tests/framework/artifacts.py
```diff
@@ -0,0 +1,146 @@
...
+from shutil import copyfile
+from typing import List
+from os import path
+from pathlib import Path
+
+import boto3
+import botocore.client
+
+
+MICROVM_CONFIG_EXTENSION = ".json"
```
gc-plp, line 18: Does any other class except `ArtifactCollection` use these definitions? Would it make more sense to either move them under `ArtifactCollection` or move the variables defined in `ArtifactCollection` inside an enum?

tests/framework/artifacts.py
sandreim: self is shorter 😃

tests/framework/artifacts.py
sandreim: Done

tests/framework/artifacts.py
sandreim: done

tests/framework/artifacts.py
sandreim: done

tests/integration_tests/functional/test_pause_resume.py
```diff
@@ -1,65 +1,111 @@
...
     exit_code, _, _ = ssh_connection.execute_command("ls")
     assert exit_code == 0
+
+    microvm.kill()
+    del tap
+
+
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
+    reason="Not supported yet."
```
ioanachirca, line 80: Pause/resume is available on ARM also.

tests/framework/artifacts.py
```diff
@@ -0,0 +1,191 @@
...
+
+class ArtifactSet:
+    """Manages a set of artifacts with the same type."""
+
+    def __init__(self, artifacts):
+        """Initialize type and artifact array."""
+        assert len(artifacts) > 0
+        self._type = artifacts[0].type
+        self._artifacts = []
+        self.add(artifacts)
```
lauralt, line 169: IMO it seems a little bit weird to have an `add` function called in an initialization function. If `add()` is only used here can we give up on this function and insert here, those artifacts (this way the second `assert` is not required)?
If we need in the future a function that adds new artifacts to the set we can make a separate one that will be used for this scope, i.e. adding (appending) not initializing :-??

tests/framework/artifacts.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def add(self, artifacts):
+        """Add artifacts to set."""
+        assert len(artifacts) > 0
+        for artifact in artifacts:
+            assert artifact.type == self._type
+            self._artifacts.append(artifact)
+
+    def artifact(self, index):
```
lauralt, line 178: Is this used?
Also, just a personal opinion: the code gets hard to follow when there are lot of `assert`s (in non test files). Is it possible to replace them with `if`s and return `None` or some kind of error if those conditions are not met / reduce the number of `assert`s / group them on the same line?

tests/integration_tests/functional/test_pause_resume.py
sandreim: They work indeed, but have no real usecase by themselves.

tests/framework/artifacts.py
```diff
@@ -0,0 +1,191 @@
...
+            ArtifactCollection.ARTIFACTS_ROOT,
+            ArtifactCollection.ARTIFACTS_DISKS,
+            ArtifactCollection.MICROVM_DISK_EXTENSION,
+            ArtifactType.DISK,
+            DiskArtifact,
+            keyword=keyword
+        )
+
+
+class ArtifactSet:
```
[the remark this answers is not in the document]
sandreim: In this context `Set` does not relate any language but to a concept: a collection of artifacts that have the same type similar to the noun `set`. The ArtifactSet class offers the guarantee that all artifacts have the same type.

tests/framework/artifacts.py
sandreim: `add()` contains logic that guarantees all artifacts in set are the same. It can be called to add more artifacts to the set after construction. Why would I duplicate it ?

tests/framework/artifacts.py
sandreim: I think you are right, this fn was used at some point but forgot to delete it as I was refactoring the code. 

Asserts are important to enforce guarantees so that undefined behavior does not propagate 
to actual test files.

tests/framework/builder.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # Download ssh key into microvm root.
+        ssh_key.download(self.root_path)
+        vm.ssh_config['ssh_key_path'] = ssh_key.local_path()
+        os.chmod(vm.ssh_config['ssh_key_path'], 0o400)
+
+        # Start firecracker.
+        vm.spawn()
+
+        # Apply the microvm artifact configuration and start microvm.
```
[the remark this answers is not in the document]
sandreim: I'll move the comment before the `basic_config()` call.

tests/framework/artifacts.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class Artifact:
+    """A generic artifact manipulation class."""
+
+    def __init__(self, bucket, key, artifact_type=ArtifactType.MISC):
+        """Initialize bucket, key and type."""
+        self.bucket = bucket
+        self.key = key
+        self.local_folder = None
```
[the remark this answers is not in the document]
sandreim: Sure, makes sense.

tests/framework/artifacts.py
gc-plp: `assert`s are a good way of detecting whether an invalid state is detected from which the execution path will not recover.  

Since we don't always want to recover from a faulty test setup, I think that asserts are useful.

Perhaps explaining why they're there would be nice - pytest supports assert messages as shown [here](https://docs.pytest.org/en/latest/assert.html):

```python
assert a % 2 == 0, "value was odd, should be even"
```

tests/framework/matrix.py
```diff
@@ -0,0 +1,160 @@
...
+        self._set_index = 0
+        if not os.path.exists(cache_dir):
+            os.mkdir(cache_dir)
+
+    @property
+    def sets(self):
+        """Return the artifact sets."""
+        return self._sets
+
+    def add_set(self, artifact_set: ArtifactSet):
```
lauralt, line 106: This also looks unused. Can we add it when it will be used? It is easier to review this way :D.

tests/framework/artifacts.py
```diff
@@ -0,0 +1,191 @@
...
+        self,
+        bucket
+    ):
+        """Initialize S3 client."""
+        config = botocore.client.Config(signature_version=botocore.UNSIGNED)
+        # pylint: disable=E1101
+        # fixes "E1101: Instance of '' has no 'Bucket' member (no-member)""
+        self.bucket = boto3.resource('s3', config=config).Bucket(bucket)
+
+    def _fetch_artifacts(self,
```
lauralt, line 105: Do we plan on calling this with `artifact_root` different than `ArtifactCollection.ARTIFACTS_ROOT`?

tests/framework/artifacts.py
sandreim: Renamed to insert().

tests/framework/builder.py
```diff
@@ -0,0 +1,59 @@
...
+        fc_binary, jailer_binary = build_tools.get_firecracker_binaries()
+
+        vm = Microvm(
+            resource_path=self.root_path,
+            fc_binary_path=fc_binary,
+            jailer_binary_path=jailer_binary,
+            microvm_id=microvm_id,
+            bin_cloner_path=self.bin_cloner_path
+        )
+        vm.setup()
```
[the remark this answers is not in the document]
sandreim: Unfortunately it doesn't work 😢 if I try to import it from conftest.py . This function will need to be changed in the future when we add firecracker binary artifacts, so indeed we are duplicating some code, with the promise that it will diverge at some point soon.

tests/framework/builder.py
```diff
@@ -0,0 +1,59 @@
...
+
+class MicrovmBuilder:
+    """Build fresh microvms or restore from snapshot."""
+
+    def __init__(self, root_path, bin_cloner_path):
+        """Initialize microvm root and cloning binary."""
+        self.root_path = root_path
+        self.bin_cloner_path = bin_cloner_path
+
+    def build(self, kernel: Artifact, disks: [DiskArtifact], config: Artifact):
```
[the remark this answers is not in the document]
sandreim: This is required to build a microvm with multiple block devices, but there is not such test defined yet. There is more work pending here, like for example adding NetDeviceArtifact support.

tests/framework/builder.py
```diff
@@ -0,0 +1,59 @@
...
+        with open(config.local_path()) as microvm_config_file:
+            microvm_config = json.load(microvm_config_file)
+
+        vm.basic_config(vcpu_count=int(microvm_config['vcpu_count']),
+                        mem_size_mib=int(microvm_config['mem_size_mib']),
+                        ht_enabled=bool(microvm_config['ht_enabled']))
+
+        return vm
+
+    def build_from_snapshot(self, snapshot: Artifact):
```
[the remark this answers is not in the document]
sandreim: I'll post a PR with with snapshots this week or next week with such a test. I'll add a comment here explaining how this is going to work.

tests/integration_tests/functional/test_pause_resume.py
```diff
@@ -1,65 +1,111 @@
...
-    test_microvm = test_microvm_with_ssh
-    test_microvm.spawn()
-
-    # Set up the microVM with 2 vCPUs, 256 MiB of RAM and a root file
-    # system with the rw permission.
-    test_microvm.basic_config()
-
-    # Create tap before configuring interface.
-    _tap1, _, _ = test_microvm.ssh_network_config(network_config, '1')
+"""Basic tests scenarios for snapshot save/restore."""
```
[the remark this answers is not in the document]
sandreim: done

tests/framework/matrix.py
```diff
@@ -0,0 +1,160 @@
...
+"""Generate multiple microvm configurations and run tests.
+
+Implements a component that calls a specific function in the context
+of the cartesian product of all artifact sets.
+"""
+
+import os
+from framework.artifacts import ArtifactSet
+
+
```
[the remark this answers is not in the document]
sandreim: done

tests/framework/artifacts.py
```diff
@@ -0,0 +1,191 @@
...
+class ArtifactCollection:
+    """Provides easy access to different artifact types."""
+
+    MICROVM_CONFIG_EXTENSION = ".json"
+    MICROVM_KERNEL_EXTENSION = ".bin"
+    MICROVM_DISK_EXTENSION = ".ext4"
+
+    # S3 bucket structure.
+    ARTIFACTS_ROOT = 'ci-artifacts'
+    ARTIFACTS_DISKS = '/disks/' + platform.machine()
```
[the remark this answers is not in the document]
sandreim: done.

tests/integration_tests/functional/test_pause_resume.py
```diff
@@ -1,65 +1,111 @@
...
+
+    logger.info("Testing microvm: \"{}\" with kernel {} and disk {} "
+                .format(context.microvm.name(),
+                        context.kernel.name(),
+                        context.disk.name()))
+
+    microvm = vm_builder.build(context.kernel,
+                               [context.disk],
+                               context.microvm)
+    tap = microvm.ssh_network_config(context.custom['network_config'], '1')
```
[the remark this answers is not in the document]
sandreim: It is used, please see below del tap. If I do not explicitly delete it I get these:
```
Cannot open network namespace "82b35ec3-1ffb-4a45-b214-b8aafa52af45": No such file or directory

Returned error code: 1
stderr:
Cannot open network namespace "82b35ec3-1ffb-4a45-b214-b8aafa52af45": No such file or directory
```

tests/framework/matrix.py
sandreim: Removed it, the constructor is enough.

tests/framework/artifacts.py
sandreim: Nope, I've removed that parameter.

sandreim:
@lauralt @ioanachirca PTAL.

tests/framework/builder.py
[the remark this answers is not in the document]
lauralt: Okay, I thought this function is designed to do just a basic configuration and adding non root block devices, net devices, etc. will be done separately.

tests/framework/builder.py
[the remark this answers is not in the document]
lauralt: Hmm, I just tried that now and it worked for me :-?.
```python
from conftest import init_microvm
...
vm = init_microvm(self.root_path, self.bin_cloner_path)
```
Isn't this working or am I doing something wrong?

tests/integration_tests/functional/test_pause_resume.py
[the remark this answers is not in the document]
lauralt: Yep, `del _tap` is needed indeed, but it looks like if we only use that variable for deleting it later, python considers `tap` as unused and marks it with that gray annoying highlight, so that's why I've suggested to rename it to `_tap`.

tests/integration_tests/functional/test_pause_resume.py
```diff
@@ -1,65 +1,112 @@
...
+    platform.machine() != "x86_64",
+    reason="Not supported yet."
+)
+def test_pause_resume(test_session_root_path,
+                      network_config,
+                      bin_cloner_path):
+    """Test scenario: boot/pause/resume for all available configurations."""
+    logger = logging.getLogger("pause_resume")
+    # Currently, artifacts share the bucket with all other resources.
+    artifact_collection = ArtifactCollection(test_images_s3_bucket())
```
lauralt, line 89: Would be nice to add a comment here why we have that `integration_tests/functional/test_pause_resume.py::test_images_s3_bucket` in the `tools/devtool test` output (in case you didn't fix that meantime).

tests/framework/matrix.py
```diff
@@ -0,0 +1,152 @@
...
+    """Computes the cartesian product of artifacts."""
+
+    __test__ = False
+
+    def __init__(self,
+                 artifact_sets,
+                 context=TestContext(),
+                 cache_dir=ARTIFACTS_LOCAL_ROOT):
+        """Initialize the cache directory."""
+        self._cache_dir = cache_dir
```
lauralt, line 88: nit: Would it make sense to move this at the end of the function? It makes things a little bit clearer I think to initialize that field after being sure that `cache_dir` path exists.

tests/framework/matrix.py
```diff
@@ -0,0 +1,152 @@
...
+
+            # Pop the previous artifact from solution.
+            cartesian_product.pop()
+            # Walk down 1 level.
+            self._set_index -= 1
+
+    def _run_test_fn(self, artifacts, test_fn):
+        """Patch context and call test_fn."""
+        for artifact in artifacts:
+            self._context.set_any(artifact.type.value, artifact)
```
lauralt, line 136: nit: it would be nice to have a function in `Artifact` class that returns `artifact.type.value`.

sandreim: The type is currently an accessor for the type (which is an enum). Value is an accesor for the Enum object.

tests/framework/matrix.py
sandreim: The cache dir needs to be initialized regardless if it exists.

lauralt:
```python
        self._context = context
        # Stores the artifact sets array.
        self._sets = artifact_sets
        # ArtifactSet stack pointer.
        self._set_index = 0
        if not os.path.exists(cache_dir):
            os.mkdir(cache_dir)
        self._cache_dir = cache_dir
```

That's what I suggested, it seems somehow more natural this flow, but it's a nit anyway.

tests/framework/matrix.py
lauralt: Yep, I preferred to kind of hide at this level the `type` type (enum), and just call a function from `Artifact` class: `artifact.type_value()`
```
def type_value(self):
        """Return the artifact type value."""
        return self._type.value
```
(but that's pretty subjective I think :-?)

tests/framework/builder.py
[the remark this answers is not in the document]
sandreim: Added comment.

tests/integration_tests/functional/test_pause_resume.py
sandreim: I've reverted back to `from conftest import init_microvm`. Seems to work now, maybe there was something wrong locally.

The PR does not fail on unused checks, neither is my IDE. It is used as it is reference by the delete.

tests/framework/matrix.py
sandreim: done.

tests/framework/matrix.py
sandreim: We can just use strings instead of enums if we make that change. Artifact types are pretty much opaque to the tests as they are fetched and manipulated through proper APIs.

tests/framework/builder.py
```diff
@@ -0,0 +1,54 @@
...
+class MicrovmBuilder:
+    """Build fresh microvms or restore from snapshot."""
+
+    def __init__(self, root_path, bin_cloner_path):
+        """Initialize microvm root and cloning binary."""
+        self.root_path = root_path
+        self.bin_cloner_path = bin_cloner_path
+
+    def build(self, kernel: Artifact, disks: [DiskArtifact], config: Artifact):
+        """Build and start a fresh microvm."""
```
ioanachirca, line 23: This comment might have slipped through - the method only builds the microvm.

tests/framework/builder.py
```diff
@@ -0,0 +1,54 @@
...
+        vm.basic_config(vcpu_count=int(microvm_config['vcpu_count']),
+                        mem_size_mib=int(microvm_config['mem_size_mib']),
+                        ht_enabled=bool(microvm_config['ht_enabled']))
+
+        return vm
+
+    # TBD: Snapshot support is not fully merged. Once that is read
+    # this function will provide a way to spin up clones.
+    def build_from_snapshot(self, snapshot: Artifact):
+        """Build and start a microvm from a snapshot artifact."""
```
ioanachirca, line 54: The method above only builds the microvm, will it be the same for this one?

tests/framework/builder.py
sandreim: It also starts it in basic_config().

tests/framework/builder.py
sandreim: Not sure yet, it's either paused or running.

tests/conftest.py
```diff
@@ -117,15 +117,15 @@ def test_with_any_microvm(test_microvm_any):
     raise PermissionError("Test session needs to be run as root.")
 
 
-def test_images_s3_bucket():
+def _test_images_s3_bucket():
```
lauralt, line 120: :tada:

Issue #1822, opened 2020-04-24
Enforce data length validity when rewriting virtio mmio device config space


aghecenco:
When an MMIO exit is triggered, depending on the address (offset), several things can happen (see [virtio specification](https://docs.oasis-open.org/virtio/virtio/v1.0/cs04/virtio-v1.0-cs04.html#x1-1110002) for details):
a) a "normal" device operation
b) a config space rewrite
c) invalid MMIO write, nothing happens ([Firecracker logs a warning](https://github.com/firecracker-microvm/firecracker/blob/31c5c2af1d1e2028e59e8b3bce12c930562ad46e/src/devices/src/virtio/mmio.rs#L316))

In situation b) according to the [virtio spec](https://docs.oasis-open.org/virtio/virtio/v1.0/cs04/virtio-v1.0-cs04.html#x1-1110002) the validity of the data written here depends on the device and driver. Firecracker does a rudimentary data length check to make sure that it fits in the config space but, if the data is smaller, it panics ([here in case of `virtio-block` device](https://github.com/firecracker-microvm/firecracker/blob/31c5c2af1d1e2028e59e8b3bce12c930562ad46e/src/devices/src/virtio/block/device.rs#L361)).

To avoid the panic, we can enforce the size checks and, on failure, increment the `block.cfg_fails` metric (and `net.cfg_fails`; the vsock device has no config space to get rewritten).

PR #1903, opened 2020-05-15
devices: support partial config space write

iulianbarbu:
## Reason for This PR

Fixes: #1822.

## Description of Changes

Support partial writes to network & block devices config space.

## License Acceptance

## PR Checklist

- ~~[ ] Any newly added `unsafe` code is properly documented~~.
- ~~[ ] Any API changes are reflected in `firecracker/swagger.yaml`~~.
- ~~[ ] Any user-facing changes are mentioned in `CHANGELOG.md`~~.

### DIFF
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -728,9 +728,8 @@ impl VirtioDevice for Net {
             METRICS.net.cfg_fails.inc();
             return;
         }
-        let (_, right) = config_space_bytes.split_at_mut(offset as usize);
-        right.copy_from_slice(&data[..]);
 
+        config_space_bytes[offset as usize..(offset + data_len) as usize].copy_from_slice(data);
         self.guest_mac = Some(MacAddr::from_bytes_unchecked(
             &self.config_space.guest_mac[..MAC_ADDR_LEN],
         ));
@@ -979,6 +978,14 @@ pub(crate) mod tests {
         // Check that the guest MAC was updated.
         let expected_guest_mac = MacAddr::from_bytes_unchecked(&new_config);
         assert_eq!(expected_guest_mac, net.guest_mac.unwrap());
+
+        // Partial write (this is how the kernel sets a new mac address) - byte by byte.
+        let new_config = [0x11, 0x22, 0x33, 0x44, 0x55, 0x66];
+        for i in 0..new_config.len() {
+            net.write_config(i as u64, &new_config[i..=i]);
+        }
+        net.read_config(0, &mut new_config_read);
+        assert_eq!(new_config, new_config_read);
 
         // Invalid write.
         net.write_config(5, &new_config);
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -357,8 +357,8 @@ impl VirtioDevice for Block {
             METRICS.block.cfg_fails.inc();
             return;
         }
-        let (_, right) = self.config_space.split_at_mut(offset as usize);
-        right.copy_from_slice(&data[..]);
+
+        self.config_space[offset as usize..(offset + data_len) as usize].copy_from_slice(data);
     }
 
     fn is_activated(&self) -> bool {
@@ -543,8 +543,16 @@ pub(crate) mod tests {
         block.read_config(0, &mut actual_config_space);
         assert_eq!(actual_config_space, expected_config_space);
 
+        // If priviledged user writes to `/dev/mem`, in block config space - byte by byte.
+        let expected_config_space = [0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x00, 0x11];
+        for i in 0..expected_config_space.len() {
+            block.write_config(i as u64, &expected_config_space[i..=i]);
+        }
+        block.read_config(0, &mut actual_config_space);
+        assert_eq!(actual_config_space, expected_config_space);
+
         // Invalid write.
-        let new_config_space: [u8; CONFIG_SPACE_SIZE] = [0xd, 0xe, 0xa, 0xd, 0xb, 0xe, 0xe, 0xf];
+        let new_config_space = [0xd, 0xe, 0xa, 0xd, 0xb, 0xe, 0xe, 0xf];
         block.write_config(5, &new_config_space);
         // Make sure nothing got written.
         block.read_config(0, &mut actual_config_space);
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -192,11 +192,12 @@ def test_session_tmp_path(test_session_root_path):
     shutil.rmtree(tmp_path)
 
 
-def _gcc_compile(src_file, output_file):
+def _gcc_compile(src_file, output_file, extra_flags="-static -O3"):
     """Build a source file with gcc."""
-    compile_cmd = 'gcc {} -o {} -static -O3'.format(
+    compile_cmd = 'gcc {} -o {} {}'.format(
         src_file,
-        output_file
+        output_file,
+        extra_flags
     )
     utils.run_cmd(compile_cmd)
 
@@ -232,6 +233,22 @@ def bin_vsock_path(test_session_root_path):
         vsock_helper_bin_path
     )
     yield vsock_helper_bin_path
+
+
+@pytest.fixture(scope='session')
+def change_net_config_space_bin(test_session_root_path):
+    """Build a binary that changes the MMIO config space."""
+    # pylint: disable=redefined-outer-name
+    change_net_config_space_bin = os.path.join(
+        test_session_root_path,
+        'change_net_config_space'
+    )
+    _gcc_compile(
+        'host_tools/change_net_config_space.c',
+        change_net_config_space_bin,
+        extra_flags=""
+    )
+    yield change_net_config_space_bin
 
 
 @pytest.fixture(scope='session')
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -10,6 +10,7 @@ destroy microvms.
 - Use the Firecracker Open API spec to populate Microvm API resource URLs.
 """
 
+import json
 import logging
 import os
 from queue import Queue
@@ -237,6 +238,22 @@ class Microvm:
     def memory_events_queue(self, queue):
         """Set the memory usage events queue."""
         self._memory_events_queue = queue
+
+    def flush_metrics(self, metrics_fifo):
+        """Flush the microvm metrics.
+
+        Requires specifying the configured metrics file.
+        """
+        # Empty the metrics pipe.
+        _ = metrics_fifo.sequential_reader(100)
+
+        response = self.actions.put(action_type='FlushMetrics')
+        assert self.api_session.is_status_no_content(response.status_code)
+
+        lines = metrics_fifo.sequential_reader(100)
+        assert len(lines) == 1
+
+        return json.loads(lines[0])
 
     def append_to_log_data(self, data):
         """Append a message to the log data."""
--- a/tests/host_tools/change_net_config_space.c
+++ b/tests/host_tools/change_net_config_space.c
@@ -1,0 +1,88 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// This is used by the `test_net_config_space.py` integration test, which writes
+// into the microVM configured network device config space a new MAC address.
+
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <sys/mman.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+int show_usage() {
+    printf("Usage: ./change_net_config_space.bin [dev_addr_start] [mac_addr]\n");
+    printf("Example:\n");
+    printf("> ./change_net_config_space.bin 0xd00001000 0x060504030201\n");
+    return 0;
+}
+
+int main(int argc, char *argv[]) {
+    int fd, i, offset;
+    uint8_t *map_base;
+    volatile uint8_t *virt_addr;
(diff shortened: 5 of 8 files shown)


tests/host_tools/change_config_space.c
```diff
@@ -0,0 +1,103 @@
...
+     * for the net device starts.
+     * Net device config space offset. From 0xD0000000
+     * to 0xD0001000 is the MMIO space for the root device
+     * (block device). */
+    config_offset = 0x00001100;
+#endif
+
+#ifdef __aarch64__
+    mmio_start_address = 0x40001000;
+    /* The offset in MMIO space where the config space
```
andreeaflorescu, line 39: The config offset is the same on both platform (0x100) as specified by the virtio specification. What is changed on different platforms is the location of the devices on the bus. This should be reflected in the code as well.

I think @sandreim had a nice suggestion of finding the start address of a specific device using information from `/proc`. This is more of a nit, but we might get unexpected failures if we switch the order in which we are adding devices on the bus in Firecracker.

sandreim:
```bash
cat /proc/cmdline
console=ttyS0 reboot=k panic=1 pci=off root=/dev/vda rw virtio_mmio.device=4K@0xd0000000:5 virtio_mmio.device=4K@0xd0001000:6 virtio_mmio.device=4K@0xd0002000:7
```

```bash
cat /sys/devices/virtio-mmio-cmdline/virtio-mmio.2/virtio2/uevent
DRIVER=virtio_net
MODALIAS=virtio:d00000001v00000000
```

lauralt: If I remember correctly on arm it was harder to take this infos from the guest.

sandreim: That is correct @lauralt . That information lives in the FDT. @iulianbarbu you can try [fdt python module](https://pypi.org/project/fdt/) or something similar.

lauralt:
Question: The description says it fixes #1822; for block device there will be a separate PR or what's the plan for it?

iulianbarbu:
> Question: The description says it fixes #1822; for block device there will be a separate PR or what's the plan for it?

Honestly, missed to look into block config space updates. I'll update the PR description with info about it, but I am not sure if I can tell much. 

It seems that the driver sets the config space only when its probing the device. Moreover, it gives some hints about the [direction](https://elixir.bootlin.com/linux/latest/source/drivers/block/virtio_blk.c#L365) of such config space update/initialization. Currently, we support updates for the backing file of a firecracker block device, but this is driven only by the host, which is trusted. Letting the guest resize the disk capacity might be dangerous if the disk is truncated. 

To be able to answer your question we would need to decide if we want to support privileged partial writes, or even partial writes at all, to the block device config space, after booting the device. Currently, this can be done writing to the device memory, inside its config space, the entire 8 bytes of the config space. Writing less will issue a panic. @firecracker-microvm/compute-capsule , WDYT about supporting writes/partial writes to the blk device config space?

iulianbarbu:
> To be able to answer your question we would need to decide if we want to support privileged partial writes, or even partial writes at all, to the block device config space, after booting the device. Currently, this can be done writing to the device memory, inside its config space, the entire 8 bytes of the config space. Writing less will issue a panic. @firecracker-microvm/compute-capsule , WDYT about supporting writes/partial writes to the blk device config space?

I've dived deep into the virtio mmio bus kernel implementation and found out the following:

- [static const struct virtio_config_ops virtio_mmio_config_ops](https://elixir.bootlin.com/linux/latest/source/drivers/virtio/virtio_mmio.c#L503) is responsible for operations on the config space of a virtio mmio device.
- [vm_set](https://elixir.bootlin.com/linux/latest/source/drivers/virtio/virtio_mmio.c#L505) field of this struct, with its implementation [here](https://elixir.bootlin.com/linux/latest/source/drivers/virtio/virtio_mmio.c#L189), is used for writing to the virtio device config space, via the virtio-mmio bus. It can write a buffer of variable length byte by byte (for virtio legacy devices, keeping the system's endianess), or once per function call, when the buffer can be safely casted without losses to __le8, __le16, __le32 and __le64.
- [virtio_blk driver](https://elixir.bootlin.com/linux/latest/source/drivers/block/virtio_blk.c#L902) do not support changing the config space, but supports [handling notifications about config space changes](https://elixir.bootlin.com/linux/latest/source/drivers/block/virtio_blk.c#L405), external to the guest. Hence, when the notification is sent, the driver is queuing some work, in order to [sync the kernel disk struct](https://elixir.bootlin.com/linux/latest/source/drivers/block/virtio_blk.c#L357) with the new config changes.
- there is no support for `virtio_blk` config space updates through any ioctl interface. The only dev ops, which might be touched by any ioctl interface is [here](https://elixir.bootlin.com/linux/latest/source/drivers/block/virtio_blk.c#L319).

After these findings I can confirm that the guest does not have any hooks which can propagate any changes to the config space. It make sens to drop support for config space writes originating from guest, after the device activation. We can not distinguish between config space writes originating from host (when updating the block device backing file path) or guest, because the host uses the same interface, namely the interface which handles virtio mmio bus accesses. Because of this common code path, we will unintentionally support guest config space writes, which will be eventually available for custom `virtio_blk` drivers with support for config space writes.

Let me know if all of these makes sense or if you have a different understanding.

tests/host_tools/change_config_space.c
iulianbarbu: @sandreim, @andreeaflorescu , PTAL! I've addressed your feedback.

iulianbarbu:
> Question: The description says it fixes #1822; for block device there will be a separate PR or what's the plan for it?

@lauralt , PTAL. I've added support for block config space partial writes.

src/devices/src/virtio/net/device.rs
```diff
@@ -31,7 +31,9 @@ use virtio_gen::virtio_net::{
     VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_UFO, VIRTIO_NET_F_HOST_TSO4, VIRTIO_NET_F_HOST_UFO,
     VIRTIO_NET_F_MAC,
 };
-use vm_memory::{ByteValued, Bytes, GuestAddress, GuestMemoryError, GuestMemoryMmap};
+use vm_memory::{
+    ByteValued, Bytes, GuestAddress, GuestMemoryError, GuestMemoryMmap, VolatileMemory,
```
acatangiu, line 35: I can't figure out where `VolatileMemory` is used..

tests/integration_tests/functional/test_config_space.py
```diff
@@ -0,0 +1,351 @@
...
+def test_net_change_mac_address(test_microvm_with_ssh, network_config,
+                                change_config_space_bin):
+    """Test changing the MAC address of the network device."""
+    global PAYLOAD_DATA_SIZE
+
+    test_microvm = test_microvm_with_ssh
+    test_microvm.spawn()
+    test_microvm.basic_config(boot_args="ipv6.disable=1")
+
+    # Data exchange interface ('eth0' in guest).
```
raduiliescu, line 30: does it worth adding a tests with 2 block devices and 2 network interfaces. The block devices added first?

iulianbarbu: At the moment, on both x86_64 and aarch64, the test microvm contains one blk device and two net devices. The side effect of adding one more blk device would be the shift by one page in memory mappings for the net devices.  If we target the logic which determines the start address of a network device memory mapping, it is idempotent to this change because the start address is not dependent on any order of the memory mappings (it is determined based on the interface name - which is determined in relation with the configured guest IPv4 address of the network interface of interest). I think that adding one more blk device to the test microvm does not bring a side effect which might change the behavior of the test.

src/devices/src/virtio/net/device.rs
iulianbarbu: It's a leftover. Removed it.

tests/host_tools/change_config_space.c
```diff
@@ -0,0 +1,106 @@
...
+    *(virt_addr + 3) = (uint8_t) (mac >> 16);
+    sprintf(mac_addr + offset, ":%02x", *(virt_addr + 3));
+    offset += 3;
+
+    *(virt_addr + 4) = (uint8_t) (mac >> 8);
+    sprintf(mac_addr + offset, ":%02x", *(virt_addr + 4));
+    offset += 3;
+
+    *(virt_addr + 5) = (uint8_t) mac;
+    sprintf(mac_addr + offset, ":%02x", *(virt_addr + 5));
```
aghecenco, line 96: nit: all of this can also be loop-ified

tests/integration_tests/functional/test_config_space.py
```diff
@@ -0,0 +1,351 @@
...
+        iterations
+    )
+
+    # `tx_spoofed_mac_count` metric was incremented due to the MAC address
+    # change.
+    fc_metrics = _flush_metrics(test_microvm, metrics_fifo)
+    assert fc_metrics['net']['tx_spoofed_mac_count'] > 0
+
+    if platform.machine() == "x86_64":
+        net_addr_base = _get_net_mem_addr_base_on_x86_64(ssh_connection_ctl,
```
aghecenco, line 97: I'd use a single `_get_net_mem_addr_base` fn and hide the `if platform.machine()` check inside it for better readability, this function is already long

tests/integration_tests/functional/test_config_space.py
```diff
@@ -0,0 +1,351 @@
...
+                                                         guest_if1_name)
+    elif platform.machine() == "aarch64":
+        net_addr_base = _get_net_mem_addr_base_on_aarch64(ssh_connection_ctl,
+                                                          guest_if1_name)
+    else:
+        assert False
+
+    assert net_addr_base is not None
+
+    # Write into '/dev/mem' the same mac address, byte by byte.
```
aghecenco, line 107: meta-nit: is there a virtio-friendly way to trigger this without `/dev/mem`? Mmap-ing `/dev/mem` and writing some bytes at some address is not exactly the first thing that comes to mind when thinking about changing a MAC address.

tests/integration_tests/functional/test_config_space.py
```diff
@@ -0,0 +1,351 @@
...
+        test_microvm.jailer,
+        ssh_connection_ctl,
+        host_ip1,
+        host_port,
+        iterations
+    )
+    fc_metrics = _flush_metrics(test_microvm, metrics_fifo)
+    assert fc_metrics['net']['tx_spoofed_mac_count'] == 0
+
+    # Try again, just to be extra sure.
```
aghecenco, line 136: :+1:

tests/integration_tests/functional/test_config_space.py
```diff
@@ -0,0 +1,351 @@
...
+    )
+
+    # Wait server to initialize.
+    exit_code, _, stderr = ssh_connection.execute_command(cmd)
+    # If this assert fails, a connection refused happened.
+    assert exit_code == 0
+    assert stderr.read() == ''
+
+
+def _flush_metrics(test_microvm, metrics_fifo):
```
aghecenco, line 224: nit: this could be moved to `host_tools` and used both here and in `test_flush_metrics`

tests/integration_tests/functional/test_config_space.py
```diff
@@ -0,0 +1,351 @@
...
+    for idx in virtio_devs_idx:
+        _, guest_if_name, _ = ssh_connection.execute_command(
+            cmd.format(sys_virtio_mmio_cmdline, idx, idx)
+        )
+        if guest_if_name.read().strip() == if_name:
+            return devs_addr[int(idx)]
+    return None
+
+
+def _get_net_if_name(ssh_connection, guest_ip):
```
aghecenco, line 347: This looks like it belongs in `host_tools/network.py`

tests/integration_tests/functional/test_config_space.py
iulianbarbu: Done.

tests/integration_tests/functional/test_config_space.py
iulianbarbu: Done.

tests/host_tools/change_config_space.c
iulianbarbu: Changing this to a `for` makes it fail on aarch64. It hangs and it is stopped due to integration test timeout > 300s.

iulianbarbu: Writing to `/dev/mem` usually involves denoting the mapped memory as volatile, because it can be modified out of the context of the current thread. A summary for what happens is described bellow, based on [this](c0x.coding-guidelines.com/6.7.3.html):

"A volatile declaration may be used to describe an object corresponding to a memory-mapped input/output port or an object accessed by an asynchronously interrupting function. Actions on objects so declared shall not be ‘‘optimized out’’ by an implementation or reordered except as permitted by the rules for evaluating expressions."

Currently, this code hangs on `aarch64` because our integration tooling compiles it with `-O3`. Haven't looked into the generated asm on `aarch64` or `x86_64`, to understand how is the loop unrolled in both cases, but I took a shorter path and came out with the following plan:
- keep the implementation intent visible, marking the '/dev/mem' mapped memory as volatile.
- simplify the code, using the for loop.
- compile without any optimizations.

It seems that following this plan works well on `aarch64` too, it is intent clear, the code is reduced, as you suggested and we follow C spec recommendations. 

Moreover, following this plan on `aarch64` with `O3` compiler optimization, hangs as well, even if the compiler should omit optimizations around the volatile memory mapping (have to look into the generated asm to confirm or force skipping unrolling the loop).

Let me know what you think.

aghecenco: Really cool deep dive, and really appreciate the thorough explanation. In this situation I'm OK with leaving the C code unchanged and the loop explicitly unrolled, maybe with a comment mentioning the compiler's behavior.

iulianbarbu: Cool, OK. I'll keep the loop unrolled in the code and I'll add a comment related to loop unrolling behavior, when using `-O3`.

Issue #1555, opened 2020-01-28
Occasional "vCPUs resume failed"


kzys:
Hello folks,

I'm upgrading Firecracker from 0.19.0 to 0.20.0 on firecracker-containerd (https://github.com/firecracker-microvm/firecracker-containerd/pull/383). One of the tests we have is launching micro 100 VMs, and it occasionally got "vCPUs resume failed" error.

Apparently the test was hitting the receive timeout Firecracker internally has (https://github.com/firecracker-microvm/firecracker/blob/53cf1bacadc23a21df361a894ac2887cafcb7139/src/vmm/src/lib.rs#L941) and changing the timeout from 100ms to 1000ms mitigated the issue.

But I'm not so sure what would be the right way to fix the issue;

- Changing the timeout from 100ms to 1000ms or something longer? It worked for us, but there is no guarantees that 1000ms is enough for everyone.
- No timeout? We could let clients handle timeout. At least it is possible for firecracker-containerd.
- Don't start vcpus as Paused mode? I don't know this is technically possible.

Thanks,

acatangiu:
Hi @kzys, if the vcpu threads don't start in 100ms, the host must be highly overloaded.

Of course, changing the timeout to something larger is fine, there is no point to optimize for the error-case with a small timeout. Feel free to open a PR for that.

More interesting to me is why is this happening for a number of microVMs as low as 100?
- What type of host machine are you using (specs)?
- Are you starting all 100 at the exact same time or did you allow a bit of spacing between them to avoid a [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem)?

kzys:
Thanks. I'm going to open a PR that changes the timeout.

We are running our tests on m5d.metal. The test is trying to start all 100 at the exact same time, to make sure we don't have any race condition around our logic.

Issue #1850, opened 2020-05-04
Snapshot create/restore integ tests


acatangiu:

sandreim:
Two new tests needed to close this issue:
- Restore snapshot artifacts
- Create snapshots with binary firecracker artifact and restore with current Firecracker build

sandreim:
Test suite completed by PR #2067.

PR #1907, opened 2020-05-18
Load Microvm snapshot support

### LINKED ISSUE #1847: Vmm support for restoring a microVM from snapshot
This tracks complete snapshot loading capability. Builds `Vmm` from a `snapshot`.

Once this is done, we tie it to Firecracker HTTP API for full stack snapshot loading functionality.

### LINKED ISSUE #1870: Extend snapshotted memory state
In the context of snapshotting guest memory, we will create a file containing either the full guest memory (full snapshot) or a sparse file with dirtied pages (diff snapshot).

In order to restore the memory from file, we need to save some state information in the bigger MicrovmState. Depending on how we will recreate the guest memory regions, the memory state will look different.

Options:

1. The simple way
- state = guest memory size
- upon restore, use the same "hardcoded" way of choosing the guest memory regions as on fresh boot ( `arch::arch_memory_regions`)

2. The more flexible way
- state = Vec<(Address, Size, FileOffset)>
- if we know the expected regions (start address and size) and the corresponding file offset that contains the memory for that region, it is easier to map them (directly use `vm_memory::GuestMemory::from_ranges_with_files` to create the guest memory)

We will start with the first option, but should keep in mind to later switch to the second.

acatangiu:
## Reason for This PR

Fixes #1847 
Fixes #1870 

## Description of Changes

Splits microvm building logic in boot-specific code and generic code. The latter is to be reused in the snapshot path.

Adds code to load microvm state from snapshot file and to build microvm from state definition.

Implements missing plumbing to have end-to-end functionality.

Integration tests will be added as part of #1850 

## License Acceptance

## PR Checklist


src/vmm/src/memory_dump.rs
```diff
@@ -3,50 +3,104 @@
...
+            guest_memory_state.regions.push(GuestMemoryRegionState {
+                base_address: region.start_addr().0,
+                size: region.len() as usize,
+                offset,
+            });
+
+            offset += region.len();
+            Ok(())
+        })
+        .map_err(Error::WriteMemory)?;
```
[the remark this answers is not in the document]
ioanachirca: It doesn't seem like this method can encounter an error

src/vmm/tests/integration_tests.rs
```diff
@@ -367,13 +370,60 @@ fn create_snapshot(is_diff: bool) {
...
+        0 => {
+            set_panic_hook();
+            let mut event_manager = EventManager::new().unwrap();
+            let empty_seccomp_filter = get_seccomp_filter(SeccompLevel::None).unwrap();
+
+            snapshot_file.as_file().seek(SeekFrom::Start(0)).unwrap();
+            let microvm_state: MicrovmState =
+                Snapshot::load(&mut snapshot_file.as_file(), VERSION_MAP.clone()).unwrap();
+
+            // TODO: get this from mem file.
```
ioanachirca, line 389: GuestMemoryMmap::restore should work.

src/vmm/src/lib.rs
```diff
@@ -421,23 +422,63 @@ impl Vmm {
...
+        }
+
+        let vcpu_responses = self
+            .vcpus_handles
+            .iter()
+            // `Iterator::collect` can transform a `Vec<Result>` into a `Result<Vec>`.
+            .map(|handle| {
+                handle
+                    .response_receiver()
+                    .recv_timeout(Duration::from_millis(1000))
```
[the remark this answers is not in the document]
acatangiu: I've increased all timeouts to 1000ms to give a better chance to highly oversubscribed scenarios/machines.

We've seen scenarios in the past where people were running stress tests and where channel timeouts would happen simply because CPUs were so contended that threads of a particular Firecracker process were scheduled to run very sparsely.

We do not enforce in Firecracker a time-limit for these operations, they will run only as fast as the host CPU runs them. The timeout's purpose here is to catch error cases where things go wrong and the thread on the other end of the channel is not responding anymore.

I'm not sure even 1000ms is enough, but it is what was used to fix https://github.com/firecracker-microvm/firecracker/issues/1555 so we can continue with that for now.

src/vmm/tests/integration_tests.rs
acatangiu: done

src/vmm/src/memory_dump.rs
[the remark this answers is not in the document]
acatangiu: right, removed.

src/vmm/src/builder.rs
```diff
@@ -324,21 +259,64 @@ pub fn build_microvm(
...
+        vm_resources.net_builder.iter(),
         event_manager,
     )?;
+    if let Some(unix_vsock) = vm_resources.vsock.get() {
+        attach_unixsock_vsock_device(&mut vmm, &mut boot_cmdline, unix_vsock, event_manager)?;
+    }
+
+    #[cfg(target_arch = "aarch64")]
+    attach_legacy_devices_aarch64(event_manager, &mut vmm, &mut boot_cmdline)
+        .map_err(StartMicrovmError::Internal)?;
```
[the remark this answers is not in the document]
acatangiu: done

src/vmm/tests/integration_tests.rs
```diff
@@ -362,13 +370,61 @@ fn create_snapshot(is_diff: bool) {
...
 
 #[cfg(target_arch = "x86_64")]
 #[test]
-fn test_create_snapshot() {
+fn test_create_and_load_snapshot() {
     // Create diff snapshot.
-    create_snapshot(true);
+    let (snapshot_file, memory_file) = verify_create_snapshot(true);
+    // Create a new microVm from snapshot.
+    verify_load_snapshot(snapshot_file, memory_file);
```
lauralt, line 424: Are we testing here loading from a diff snapshot?

ioanachirca: The problem appears after we resume the microVM, it should encounter no error before. The test seems to be about verifying that `create_snapshot` produces valid snapshot and memory files and that `load_snapshot` can build the microVM from these files. The actual content of the memory file and what happens after resume can be checked in another test, I think. To test functionality after loading from a diff snapshot we'd also need a tool for merging the sparse files.

src/vmm/src/lib.rs
```diff
@@ -42,7 +42,7 @@ pub mod builder;
 /// Syscalls allowed through the seccomp filter.
 pub mod default_syscalls;
 pub(crate) mod device_manager;
-pub(crate) mod memory_dump;
+pub mod memory_dump;
```
ioanachirca, line 45: We could find a better name for this module, since now it can do more than dumping memory. Maybe `memory_snapshot` ?

src/vmm/src/memory_dump.rs
```diff
@@ -92,19 +149,123 @@ impl DumpMemory for GuestMemoryMmap {
...
             writer_offset += region.len();
             Ok(())
         })
-        .map_err(Error::WriteMemory)?;
+        .map_err(Error::WriteMemory)
+    }
+
+    /// Creates a GuestMemoryMmap give a `file` containing the data
+    /// and a `state` containing mapping information.
+    fn restore(file: &File, state: &GuestMemoryState) -> std::result::Result<Self, Error> {
```
ioanachirca, line 164: Should this include the fix to use a `MAP_PRIVATE` mapping?

src/vmm/src/memory_dump.rs
```diff
@@ -3,52 +3,113 @@
...
                 .map(|_| ())
         })
-        .map_err(Error::WriteMemory)?;
-
-        Ok(())
+        .map_err(Error::WriteMemory)
     }
 
+    /// Dumps all pages of GuestMemoryMmap present in `dirty_bitmap` to a writer.
     fn dump_dirty<T: std::io::Write + std::io::Seek>(
```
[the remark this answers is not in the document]
ioanachirca: I think it can be done:
- create a GuestMemoryMmap with a couple of pages and write something on each of them
- create a DirtyBitmap that marks only some of these pages
- check that the resulting file after `dump_dirty` has content (non-zero pages) only for the marked pages

src/vmm/tests/integration_tests.rs
lauralt: Yup, it's kinda confusing though to have a test that is basically checking for a happy case when that case is expected to not be a happy one :-?, the microVM can be built, but it can't be used. I don't know if it makes much sense to test loading a microVM from snapshot without resuming it too.
Anyway, a comment here why is this working (since we're telling in docs that loading should be done only from a full snapshot) would be useful.

tests/framework/microvm.py
```diff
@@ -535,6 +556,48 @@ def start(self):
...
+        """Disable network namespaces functionality, to use with cloning."""
+        # Cloning and network namespaces don't work well together yet.
+        self.jailer.netns = None
+        self.ssh_config['netns_file_path'] = None
+
+    def pause_to_snapshot(self, mem_file_path=None, snapshot_path=None):
+        """Pauses the microVM, and creates snapshot.
+
+        This function validates that the microVM pauses successfully and
+        returns the name of the snapshot file.
```
ioanachirca, line 569:
```suggestion
        returns the name of the snapshot files.
```

src/vmm/src/vstate.rs
```diff
@@ -980,7 +980,7 @@ impl Vcpu {
 
     #[allow(unused)]
```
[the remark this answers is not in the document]
acatangiu: Right! will remove

src/vmm/src/memory_dump.rs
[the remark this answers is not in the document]
acatangiu: sounds good

tests/integration_tests/functional/test_snapshot.py
```diff
@@ -0,0 +1,97 @@
...
+        snapshot_path=test_microvm.tmp_path()
+    )
+    assert test_microvm.api_session.is_status_bad_request(resp.status_code)
+    assert "requested operation is not supported before starting the microVM"\
+           in resp.text
+
+    # Start microVM.
+    test_microvm.start()
+
+    # The APIs for snapshot related procedures does not get timed (yet?).
```
ioanachirca, line 58: nit:
```suggestion
    # The APIs for snapshot related procedures do not get timed (yet?).
```

tests/framework/resources.py
```diff
@@ -428,6 +428,75 @@ def create_json(
...
+    def put(self, **args):
+        """Create a snapshot of the microvm."""
+        datax = self.create_json(**args)
+        return self._api_session.put(
+            "{}".format(self._snapshot_cfg_url),
+            json=datax
+        )
+
+    @staticmethod
+    def create_json(mem_file_path, snapshot_path, diff=False):
```
ioanachirca, line 452: What about version? Even though we're not testing that now, it would be good to add a `version=None` param.

tests/integration_tests/functional/test_snapshot.py
```diff
@@ -0,0 +1,97 @@
...
+def test_snapshot(test_microvm_with_ssh, network_config):
+    """Test snapshotting functionality."""
+    test_microvm = test_microvm_with_ssh
+
+    # Cloning and network namespaces don't work well together yet.
+    test_microvm.disable_netns()
+
+    test_microvm.spawn()
+
+    # Set up the microVM with 1 vCPUs, 256 MiB of file-backed RAM, no network
```
ioanachirca, line 27: nit:
```suggestion
    # Set up the microVM with 1 vCPU, 256 MiB of file-backed RAM, no network
```

src/vmm/tests/integration_tests.rs
acatangiu: Most of our rust integration tests only verify functions run with no errors but don't verify the guest workload. I think the main reason for that is that we have our python integration tests which have a complex framework for doing that.

To verify how guest behaves in rust integration tests would need framework to communicate with the guest; to verify devices still function in the guest would need framework to set up devices on the host and framework to test black-box device functionality; etc.

So the current test only verifies that *building* a microvm works from a *code-logic* point of view.

src/vmm/src/memory_dump.rs
acatangiu: yes

tests/integration_tests/functional/test_snapshot.py
```diff
@@ -0,0 +1,97 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Tests scenarios for pausing/resuming a microVM."""
```
lauralt, line 3: This should be updated.

src/vmm/src/persist.rs
```diff
@@ -284,32 +384,35 @@ mod tests {
         let err = InvalidVersion;
         let _ = format!("{}{:?}", err, err);
 
+        let err = InvalidVmState(vstate::Error::NotEnoughMemorySlots);
```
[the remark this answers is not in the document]
ioanachirca: Yup, it's similar to other tests that check `Display` implementations, only for coverage.

src/vmm/src/memory_dump.rs
ioanachirca: Updated.

src/vmm/src/lib.rs
ioanachirca: Done.

src/vmm/src/memory_dump.rs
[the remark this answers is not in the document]
ioanachirca: Done.

src/vmm/src/persist.rs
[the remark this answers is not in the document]
acatangiu: It also tests that `Display` and `Debug` **are** implemented. It doesn't verify message string, but it does verify that there **is** a `Display` and `Debug` implementation.

src/api_server/swagger/firecracker.yaml
```diff
@@ -412,7 +412,7 @@ paths:
           description: The configuration used for creating a snaphot.
           required: true
           schema:
-            $ref: "#/definitions/CreateSnapshotParams"
+            $ref: "#/definitions/SnapshotCreateParams"
       responses:
         501:
```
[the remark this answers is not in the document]
acatangiu: good catch! will do

src/vmm/src/builder.rs
```diff
@@ -324,21 +269,128 @@ pub fn build_microvm(
...
+        guest_memory,
+        track_dirty_pages,
+        vcpu_config.vcpu_count,
+    )?;
+
     attach_block_devices(
         &mut vmm,
         &mut boot_cmdline,
-        &vm_resources.block,
+        vm_resources.block.list.iter(),
```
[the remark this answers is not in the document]
acatangiu: > Having an iterator as function parameter doesn't look like something very intuitive or idiomatic

Why not?

--------------

This change is ~required~ _really-nice-to-have_ to break the tight coupling between this generic microVM building code from the Firecracker-specific `BlockBuilder`/`NetBuilder`/`VsockBuilder`.

Using a generic iterator here allows calling this function from any specific code as long as that code/class/object can provide an iterator over the actual devices.

src/vmm/src/lib.rs
```diff
@@ -384,22 +385,22 @@ impl Vmm {
...
-            .save_state()
-            .map_err(MicrovmStateError::SaveVmState)?;
+        let vm_state = self.vm.save_state().map_err(SaveVmState)?;
 
         let device_states = self.mmio_device_manager.save();
 
         let mem_size_mib =
             self.guest_memory()
                 .map_and_fold(0, |(_, region)| region.len(), |a, b| a + b)
                 >> 20;
```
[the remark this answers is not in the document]
acatangiu: yup, will do

src/vmm/src/vstate.rs
```diff
@@ -1306,36 +1319,40 @@ pub struct VcpuState {
 
 // Allow currently unused Pause and Exit events. These will be used by the vmm later on.
```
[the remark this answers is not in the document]
acatangiu: yup, will do

src/vmm/tests/integration_tests.rs
```diff
@@ -288,7 +295,7 @@ fn test_dirty_bitmap_success() {
 }
 
 #[cfg(target_arch = "x86_64")]
-fn create_snapshot(is_diff: bool) {
+fn verify_create_snapshot(is_diff: bool) -> (TempFile, TempFile) {
```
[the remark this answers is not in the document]
acatangiu: I'll take a stab at it

tests/framework/resources.py
acatangiu: yup, will add it

tests/framework/resources.py
```diff
@@ -428,6 +428,75 @@ def create_json(
...
+        return datax
+
+
+class SnapshotLoad():
+    """Facility for sending snapshot instructions on the microvm."""
+
+    SNAPSHOT_LOAD_URL = 'snapshot/load'
+
+    __snapshot_cfg_url = None
+    __api_session = None
```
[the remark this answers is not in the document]
acatangiu: Good catch, I initially added these two classes with class methods, then switched `SnapshotCreate` to object methods but didn't also switch `SnapshotLoad`.

It works either way, but I will change this one too to keep the approach consistent with the others.

src/vmm/src/vstate.rs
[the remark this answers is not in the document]
acatangiu: done

src/vmm/src/vstate.rs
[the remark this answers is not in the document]
acatangiu: done

src/vmm/src/vstate.rs
```diff
@@ -1306,36 +1319,40 @@ pub struct VcpuState {
...
 }
 
 /// List of responses that the Vcpu reports.
 pub enum VcpuResponse {
+    /// Requested action encountered an error.
+    #[cfg(target_arch = "x86_64")]
+    Error(Error),
+    /// Vcpu is stopped.
+    Exited(u8),
+    /// Vcpu state not allowed while running.
```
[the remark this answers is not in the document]
acatangiu: done

src/vmm/src/lib.rs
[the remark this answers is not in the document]
acatangiu: done

src/vmm/tests/integration_tests.rs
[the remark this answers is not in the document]
acatangiu: I don't see an opportunity to de-duplicate any code and still keep tests clear and easy to follow.

> For example the part where we deserialize the microvm state and check if the result is the expected one can be a separate function

I personally feel that just moving those lines in two other _extra_ functions that nobody else will use, actually hurts readability with no added benefits..

src/api_server/swagger/firecracker.yaml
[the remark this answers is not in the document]
acatangiu: done

tests/framework/microvm.py
acatangiu: done

tests/framework/resources.py
acatangiu: done

tests/framework/resources.py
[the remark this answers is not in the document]
acatangiu: done

tests/integration_tests/functional/test_snapshot.py
acatangiu: done

tests/integration_tests/functional/test_snapshot.py
acatangiu: done

tests/integration_tests/functional/test_snapshot.py
acatangiu: done

src/vmm/src/builder.rs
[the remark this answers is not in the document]
lauralt: > > Having an iterator as function parameter doesn't look like something very intuitive or idiomatic
> 
> Why not?
> 
Well, it's not the first thing you're thinking about when you want to write a function that does something with the elements of a specific collection :D. Anyway, my observation is indeed pretty subjective, but I find a function which takes a reference to a collection of `<T>` instead of an iterator over `<T>` elements a little bit more readable and easier to understand, also there's no need to add explicit lifetimes which in general affect readability too.

> This change is ~required~ _really-nice-to-have_ to break the tight coupling between this generic microVM building code from the Firecracker-specific `BlockBuilder`/`NetBuilder`/`VsockBuilder`.
> 
> Using a generic iterator here allows calling this function from any specific code as long as that code/class/object can provide an iterator over the actual devices.

I agree it decouples things a little, the fact that we're using for example a `VecDeque` in `BlockBuilder` is pretty specific, but at this moment those functions are used only in the `BlockBuilder`, `NetBuilder` contexts, so I don't know if this change is very useful now. Maybe it would be a good idea to refactor a bit `insert_block_devices()` and `insert_net_device()` from tests to not use `BlockBuilder`, `NetBuilder` anymore and this way your change will be even more justified.

src/vmm/src/persist.rs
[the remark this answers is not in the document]
lauralt: Ok, sounds good, it could be a good idea to add a test also for `LoadSnapshotError`. I noticed that the coverage decreased and I think (not 100% sure though) there isn't a lot of new functionality that can't be rust tested (either with unit tests or with integration tests).

src/vmm/tests/integration_tests.rs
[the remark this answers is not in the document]
lauralt: I suggested to move that part to a separate function, because in the context of verifying if the snapshot creation is successful, we shouldn't care about the specific checks (same memory size, same vcpu count) that we're doing. Those checks seem to correspond to the deserialization of a "default" microVMstate (no devices, one vcpu ...).

```rust
fn check_default_microvm_deser(memory_file: &TempFile, snapshot_file: &TempFile) {
    let restored_microvm_state: MicrovmState =
        Snapshot::load(&mut snapshot_file.as_file(), VERSION_MAP.clone()).unwrap();
   .......
    assert_eq!(restored_microvm_state.vcpu_states.len(), 1);
}
```
 I think that `verify_create_snapshot()` is more readable when you hide that logic somehow.

Moreover, I just noticed now that this part:
```rust
let boot_source_cfg: BootSourceConfig = MockBootSourceConfig::new()
                .with_default_boot_args()
                .with_kernel(NOISY_KERNEL_IMAGE)
                .into();
            let resources: VmResources = MockVmResources::new()
                .with_boot_source(boot_source_cfg)
                .into();
            let mut event_manager = EventManager::new().unwrap();
            let empty_seccomp_filter = get_seccomp_filter(SeccompLevel::None).unwrap();

            // The vmm will start with dirty page tracking = OFF.
            let vmm = build_microvm_for_boot(&resources, &mut event_manager, &empty_seccomp_filter)
                .unwrap();
```

 is common to a lot of tests/functions in this file. It might not be in the scope of this PR and sorry for bothering you with this :(, but the size of this file is beginning to increase substantially and it's really hard to follow the tests anymore and the scope of each one of them. Since this PR introduces two new big functions, this could be a good time to de-duplicate some code.

That function could have the following signature:
`fn default_vmm(kernel_image: Option<&str>) ->Arc<Mutex<Vmm>> { };`

If it's not the right time to do this change, feel free to dismiss my comment.

src/vmm/tests/integration_tests.rs
lauralt: Got it, still, a more detailed comment on line 423 would be great.

src/vmm/src/rpc_interface.rs
```diff
@@ -279,7 +282,18 @@ impl<'a> PrebootApiController<'a> {
                 .build_net_device(netif_body)
                 .map(|_| VmmData::Empty)
                 .map_err(VmmActionError::NetworkConfig),
-            LoadSnapshot(_snapshot_load_cfg) => Ok(VmmData::NotFound),
```
lauralt, line 282: `VmmData::NotFound` is not used anymore, but I don't know if we should keep it around for future APIs that will temporarily lack an implementation, what do you think?

src/vmm/tests/integration_tests.rs
acatangiu: done

src/vmm/src/builder.rs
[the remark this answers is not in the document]
acatangiu: In tests we're using `BlockBuilder`/`NetBuilder`/`VsockBuilder` to get free device building - so we don't duplicate code to build the device.

I, too, don't like how the dependency graph is a mess when looking at tests.
We should address this as part of https://github.com/firecracker-microvm/firecracker/issues/1713.

src/vmm/tests/integration_tests.rs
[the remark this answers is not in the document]
acatangiu: done

src/vmm/src/persist.rs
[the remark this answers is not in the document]
acatangiu: done

src/vmm/src/rpc_interface.rs
```diff
@@ -359,10 +369,11 @@ impl RuntimeApiController {
             | ConfigureMetrics(_)
             | InsertBlockDevice(_)
             | InsertNetworkDevice(_)
-            | LoadSnapshot(_)
             | SetVsockDevice(_)
             | SetMmdsConfiguration(_)
             | SetVmConfiguration(_) => Err(VmmActionError::OperationNotSupportedPostBoot),
+            #[cfg(target_arch = "x86_64")]
+            LoadSnapshot(_) => Err(VmmActionError::OperationNotSupportedPreBoot),
```
lauralt, line 376:
```suggestion
            LoadSnapshot(_) => Err(VmmActionError::OperationNotSupportedPostBoot),
```

tests/framework/microvm.py
```diff
@@ -300,6 +312,7 @@ def setup(self):
     def spawn(self, create_logger=True):
         """Start a microVM as a daemon or in a screen session."""
         # pylint: disable=subprocess-run-check
+        # pylint: disable=too-many-statements
```
lauralt, line 315: I think one way to avoid this would be to move, for example, the following part:
```python
self.actions = Actions(self._api_socket, self._api_session)
...      
self.vsock = Vsock(self._api_socket, self._api_session)
```
to a separate function (`initialize_resources()` or something like that).

`spawn()` is pretty huge at this moment.

tests/framework/resources.py
```diff
@@ -428,6 +428,74 @@ def create_json(
...
+        datax = self.create_json(**args)
+        return self._api_session.put(
+            "{}".format(self._snapshot_cfg_url),
+            json=datax
+        )
+
+    @staticmethod
+    def create_json(mem_file_path, snapshot_path):
+        """Compose the json associated to this type of API request."""
+        datax = {
```
lauralt, line 492: This should contain `enable_diff_snapshots` too.

tests/framework/resources.py
```diff
@@ -428,6 +428,74 @@ def create_json(
         return datax
 
 
+class SnapshotCreate():
+    """Facility for sending snapshot instructions on the microvm."""
```
lauralt, line 432: Can we differentiate this comment and the one from line 470 somehow?

src/vmm/src/lib.rs
```diff
@@ -421,23 +411,63 @@ impl Vmm {
             .map(|handle| {
                 handle
                     .response_receiver()
-                    .recv_timeout(Duration::from_millis(400))
+                    .recv_timeout(Duration::from_millis(1000))
```
sandreim, line 414: Is this enough when the host is under CPU pressure ?

src/firecracker/src/main.rs
```diff
@@ -271,7 +271,7 @@ fn build_microvm_from_json(
             );
             process::exit(i32::from(vmm::FC_EXIT_CODE_BAD_CONFIGURATION));
         });
-    let vmm = vmm::builder::build_microvm(&vm_resources, event_manager, &seccomp_filter)
+    let vmm = vmm::builder::build_microvm_for_boot(&vm_resources, event_manager, &seccomp_filter)
```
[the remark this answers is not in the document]
acatangiu: I like it better like it is now. The function builds a microvm and does very specific configurations to memory and vcpus for booting Linux. I think it's useful to have that information reflected in the function name.

src/vmm/src/builder.rs
```diff
@@ -324,21 +278,123 @@ pub fn build_microvm(
...
+    vmm.mmio_device_manager =
+        MMIODeviceManager::restore(mmio_ctor_args, &microvm_state.device_states).unwrap();
+
+    // Move vcpus to their own threads and start their state machine in the 'Paused' state.
+    vmm.start_vcpus(vcpus, seccomp_filter).map_err(Internal)?;
+
+    // Restore vcpus kvm state.
+    vmm.restore_vcpu_states(microvm_state.vcpu_states).unwrap();
+
+    // Load seccomp filters for the VMM thread.
```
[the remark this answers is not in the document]
acatangiu: I guess there could be usecases for that. But let's treat it as an enhancement. It is not required for our current milestone.

src/vmm/src/builder.rs
```diff
@@ -324,21 +278,123 @@ pub fn build_microvm(
...
+    // Execution panics if filters cannot be loaded, use --seccomp-level=0 if skipping filters
+    // altogether is the desired behaviour.
+    SeccompFilter::apply(seccomp_filter.to_vec())
+        .map_err(Error::SeccompFilters)
+        .map_err(Internal)?;
+
+    let vmm = Arc::new(Mutex::new(vmm));
```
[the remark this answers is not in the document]
acatangiu: done

src/vmm/src/lib.rs
```diff
@@ -383,35 +375,33 @@ impl Vmm {
 
     /// Saves the state of a paused Microvm.
     #[cfg(target_arch = "x86_64")]
-    pub fn save_state(&mut self) -> std::result::Result<MicrovmState, SaveMicrovmStateError> {
+    pub fn save_state(&mut self) -> std::result::Result<MicrovmState, MicrovmStateError> {
```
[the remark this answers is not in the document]
acatangiu: We're not using it outside the crate, yet.

But it is public because it is part of the public `Vmm` programmatic interface.

src/vmm/src/lib.rs
acatangiu: Raised to `1s` to address exactly that. I am sure that in some extreme scenarios or on very slow hardware, `1s` might not be enough.

We can go even higher right now or tune this later if necessary.

`1000ms` is the value used to fix #1555

src/vmm/src/lib.rs
```diff
@@ -421,23 +411,63 @@ impl Vmm {
...
             })
-            .collect::<std::result::Result<Vec<VcpuState>, SaveMicrovmStateError>>()?;
+            .collect::<std::result::Result<Vec<VcpuState>, MicrovmStateError>>()?;
 
         Ok(vcpu_states)
     }
 
+    #[cfg(target_arch = "x86_64")]
+    /// Restores vcpus kvm states.
+    pub fn restore_vcpu_states(
```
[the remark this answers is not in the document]
acatangiu: We're not using it outside the crate, yet.

But it is public because it is part of the public `Vmm` programmatic interface.

src/vmm/src/persist.rs
```diff
@@ -131,25 +185,26 @@ fn snapshot_state_to_file(
     version: Option<String>,
     version_map: VersionMap,
 ) -> std::result::Result<(), CreateSnapshotError> {
+    use self::CreateSnapshotError::*;
     let mut snapshot_file = OpenOptions::new()
         .create(true)
         .write(true)
         .open(snapshot_path)
-        .map_err(CreateSnapshotError::SnapshotBackingFile)?;
+        .map_err(SnapshotBackingFile)?;
```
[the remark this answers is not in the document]
acatangiu: I think it's fine like this. `map_err` tells without uncertainty that we're dealing with an error.

src/vmm/src/persist.rs
```diff
@@ -159,38 +214,75 @@ fn snapshot_memory_to_file(
     mem_file_path: &PathBuf,
     snapshot_type: SnapshotType,
 ) -> std::result::Result<(), CreateSnapshotError> {
+    use self::CreateSnapshotError::*;
     let mut file = OpenOptions::new()
         .write(true)
         .create(true)
         .truncate(true)
         .open(mem_file_path.clone())
```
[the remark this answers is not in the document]
acatangiu: fixed

src/vmm/src/persist.rs
```diff
@@ -159,38 +214,75 @@ fn snapshot_memory_to_file(
...
+
+fn guest_memory_from_file(
+    mem_file_path: &PathBuf,
+    mem_state: &GuestMemoryState,
+) -> std::result::Result<GuestMemoryMmap, LoadSnapshotError> {
+    use self::LoadSnapshotError::{DeserializeMemory, MemoryBackingFile};
+    let mem_file = OpenOptions::new()
+        .write(true)
+        .read(true)
+        .open(mem_file_path.clone())
```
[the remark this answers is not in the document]
acatangiu: fixed

src/vmm/src/rpc_interface.rs
acatangiu: good catch! fixed

tests/framework/microvm.py
acatangiu: done

tests/framework/resources.py
acatangiu: done

src/vmm/src/builder.rs
```diff
@@ -324,21 +278,123 @@ pub fn build_microvm(
...
+        vm: vmm.vm.fd(),
+    };
+    vmm.mmio_device_manager =
+        MMIODeviceManager::restore(mmio_ctor_args, &microvm_state.device_states).unwrap();
+
+    // Move vcpus to their own threads and start their state machine in the 'Paused' state.
+    vmm.start_vcpus(vcpus, seccomp_filter).map_err(Internal)?;
+
+    // Restore vcpus kvm state.
+    vmm.restore_vcpu_states(microvm_state.vcpu_states).unwrap();
```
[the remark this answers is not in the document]
lauralt: > Do we need to explicitly call resume from the API to resume the VCPUs ? Can't we just do that here so a new API call is not necessary. IMO the only logical action after load is to resume.
> 

Hmm, it's a good point and would simplify things a bit. But since `CreateSnapshot` involves pausing and resuming to be done explicitly, we would lose I guess some kind of a symmetry this way :-??.
Maybe we should start thinking more about the config file option also for load, as you already proposed in another thread. We could add a new cmd line parameter, `--snapshot-file` or `--snapshot-config-file` which would include the load config and also the optional logger and metrics configs. 

> I guess this unwrap is not in the final version of the PR. Error needs to travel up the stack and presented to the user.

tests/framework/resources.py
acatangiu: done

src/vmm/src/builder.rs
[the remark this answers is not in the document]
acatangiu: > Do we need to explicitly call resume from the API to resume the VCPUs ? Can't we just do that here so a new API call is not necessary. IMO the only logical action after load is to resume.

Yes, that is the proposed and approved API flow.

We can change that outside this PR, as it implies getting buy-in from users and updating documentation as well.

acatangiu:
@sandreim @lauralt @ioanachirca I've addressed all your comments, please review.

acatangiu:
@sandreim @lauralt @ioanachirca this PR is up-to-date. Please review.

tests/framework/microvm.py
```diff
@@ -552,6 +576,48 @@ def start(self):
         response = self.actions.put(action_type='InstanceStart')
         assert self._api_session.is_status_no_content(response.status_code)
 
+    def disable_netns(self):
+        """Disable network namespaces functionality, to use with cloning."""
+        # Cloning and network namespaces don't work well together yet.
```
ioanachirca, line 581: Could you add a little more detail about this (especially the `yet`part)?

acatangiu: There is no support yet 😛 

This is a very basic test. @sandreim is working on adding a capable testing framework that can handle cloning.

ioanachirca: Got it, from from the comment it was not clear whether we were lacking in functionality or testing framework.

acatangiu:
I removed that _dumb_ integ test from this PR. A highly improved version of it will be added by @sandreim as part of #1850

tests/integration_tests/build/test_binary_size.py
```diff
@@ -11,10 +11,10 @@
 MACHINE = platform.machine()
 """ Platform definition used to select the correct size target"""
 
-FC_BINARY_SIZE_TARGET = 3013464 if MACHINE == "x86_64" else 3251464
+FC_BINARY_SIZE_TARGET = 3200000 if MACHINE == "x86_64" else 3251464
 """Firecracker target binary size in bytes"""
 
-FC_BINARY_SIZE_LIMIT = 3200000 if MACHINE == "x86_64" else 3400000
+FC_BINARY_SIZE_LIMIT = 3500000 if MACHINE == "x86_64" else 3400000
```
lauralt, line 17: The `3500000` limit is pretty big comparing with the current binary size (`3194680` from what I've seen when running this test and forcing it it to fail).
Also, more of a nit, but can we update the `FC_BINARY_SIZE_TARGET` with a more accurate value? I think it's nice to modify with precise values when the test requires to update these values (it keeps somehow the history clearer).

tests/integration_tests/build/test_coverage.py
```diff
@@ -23,7 +23,7 @@
 # this contains the frequency while on AMD it does not.
 # Checkout the cpuid crate. In the future other
 # differences may appear.
-COVERAGE_DICT = {"Intel": 84.63, "AMD": 84.60}
+COVERAGE_DICT = {"Intel": 84.43, "AMD": 84.39}
```
lauralt, line 26: This is a rather big decrease :(. Are we sure all the code that can be rust tested is indeed tested? I know we will have some awesome python integration tests, but it's nice to add also unit tests (where it is possible).

Issue #1751, opened 2020-04-06
vsock: error reading from backing stream


stefano-garzarella:
While developing the vhost-user-vsock application for cloud-hypervisor, I had an issue that happens also with the vsock device emulation: https://github.com/cloud-hypervisor/cloud-hypervisor/issues/1001

As @sboeuf suggested, I tried with Firecracker and I had a very similar issue:

```
2020-04-06T11:52:33.632842533 [anonymous-instance:ERROR:src/devices/src/virtio/vsock/csm/connection.rs:242] vsock: error reading from backing stream: lp=5201, pp=1031, err=Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }
```
After the error, the throughput goes to 0 and the peers are able to communicate only restarting the connection, but the error happens every time I tried to run `iperf-vsock` (guest connecting to the host)

## How to reproduce

I used [iperf-vsock](https://github.com/stefano-garzarella/iperf-vsock) to stress the vsock connection.

Kernel: hello-vmlinux.bin (Linux (none) 4.14.55-84.37.amzn2.x86_64 #1 SMP Wed Jul 25 18:47:15 UTC 2018 x86_64 Linux)
Firecracker: firecracker-v0.21.1-x86_64

### VM config
```json
{
  "boot-source": {
    "kernel_image_path": "./hello-vmlinux.bin",
    "boot_args": "console=ttyS0 reboot=k panic=1 pci=off"
  },
  "drives": [
    {
      "drive_id": "rootfs",
      "path_on_host": "rootfs.ext4",
      "is_root_device": true,
      "is_read_only": false
    }
  ],
  "machine-config": {
    "vcpu_count": 2,
    "mem_size_mib": 1024,
    "ht_enabled": false
  },
  "vsock": {
      "vsock_id": "1",
      "guest_cid": 4,
      "uds_path": "/tmp/vm4.vsock"
  }
}
```

### guest
```shell
~# mkdir /tmp     # iperf3 uses mkstemp(3)
~# iperf3 --vsock -c 2
Connecting to host 2, port 5201
[  5] local 4 port 1033 connected to 2 port 5201
[ ID] Interval           Transfer     Bitrate
[  5]   0.00-1.00   sec  91.0 MBytes   763 Mbits/sec                  
[...]

PR #1910, opened 2020-05-18
Update vsock logic to address stalling problems

gc-plp:
## Reason for This PR

Addresses #1751 

## Description of Changes

This PR changes some of the internal mechanisms in the vsock device, that were causing the problems described in #1751.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/vsock/csm/connection.rs
+++ b/src/devices/src/virtio/vsock/csm/connection.rs
@@ -178,81 +178,94 @@ where
             return Ok(());
         }
 
+        if self.pending_rx.remove(PendingRx::Rw) {
+            // We're due to produce a data packet, by reading the data from the host-side
+            // Unix socket.
+
+            match self.state {
+                // A data packet is only valid for established connections, and connections for
+                // which our peer has initiated a graceful shutdown, but can still receive data.
+                ConnState::Established | ConnState::PeerClosed(false, _) => (),
+                _ => {
+                    // Any other connection state is invalid at this point, and we need to kill it
+                    // with fire.
+                    pkt.set_op(uapi::VSOCK_OP_RST);
+                    return Ok(());
+                }
+            }
+
+            // Oh wait, before we start bringing in the big data, can our peer handle receiving so
+            // much bytey goodness?
+            if self.need_credit_update_from_peer() {
+                self.last_fwd_cnt_to_peer = self.fwd_cnt;
+                pkt.set_op(uapi::VSOCK_OP_CREDIT_REQUEST);
+                return Ok(());
+            }
+
+            let buf = pkt.buf_mut().ok_or(VsockError::PktBufMissing)?;
+
+            // The maximum amount of data we can read in is limited by both the RX buffer size and
+            // the peer available buffer space.
+            let max_len = std::cmp::min(buf.len(), self.peer_avail_credit());
+
+            // Read data from the stream straight to the RX buffer, for maximum throughput.
+            match self.stream.read(&mut buf[..max_len]) {
+                Ok(read_cnt) => {
+                    if read_cnt == 0 {
+                        // A 0-length read means the host stream was closed down. In that case,
+                        // we'll ask our peer to shut down the connection. We can neither send nor
+                        // receive any more data.
+                        self.state = ConnState::LocalClosed;
+                        self.expiry = Some(
+                            Instant::now() + Duration::from_millis(defs::CONN_SHUTDOWN_TIMEOUT_MS),
+                        );
+                        pkt.set_op(uapi::VSOCK_OP_SHUTDOWN)
+                            .set_flag(uapi::VSOCK_FLAGS_SHUTDOWN_RCV)
+                            .set_flag(uapi::VSOCK_FLAGS_SHUTDOWN_SEND);
+                    } else {
+                        // On a successful data read, we fill in the packet with the RW op, and
+                        // length of the read data.
+                        pkt.set_op(uapi::VSOCK_OP_RW).set_len(read_cnt as u32);
+                    }
+                    self.rx_cnt += Wrapping(pkt.len());
+                    self.last_fwd_cnt_to_peer = self.fwd_cnt;
+                    return Ok(());
+                }
+                Err(err) if err.kind() == ErrorKind::WouldBlock => {
+                    // This shouldn't actually happen (receiving EWOULDBLOCK after EPOLLIN), but
+                    // apparently it does, so we need to handle it greacefully.
+                    warn!(
+                        "vsock: unexpected EWOULDBLOCK while reading from backing stream: \
+                         lp={}, pp={}, err={:?}",
+                        self.local_port, self.peer_port, err
+                    );
+                }
+                Err(err) => {
+                    // We are not expecting any other errors when reading from the underlying
+                    // stream. If any show up, we'll immediately kill this connection.
+                    error!(
+                        "vsock: error reading from backing stream: lp={}, pp={}, err={:?}",
+                        self.local_port, self.peer_port, err
+                    );
+                    pkt.set_op(uapi::VSOCK_OP_RST);
+                    self.last_fwd_cnt_to_peer = self.fwd_cnt;
+                    return Ok(());
+                }
+            };
+        }
+
         // A credit update is basically a no-op, so we should only waste a perfectly fine RX
-        // buffer on it if we really have nothing else to say.
+        // buffer on it if we really have nothing else to say, hence we check for this RX
+        // indication last.
         if self.pending_rx.remove(PendingRx::CreditUpdate) && !self.has_pending_rx() {
             pkt.set_op(uapi::VSOCK_OP_CREDIT_UPDATE);
             self.last_fwd_cnt_to_peer = self.fwd_cnt;
             return Ok(());
         }
 
-        // Alright, if we got to here, we need to cough up a data packet. We've already checked
-        // for all other pending RX indications.
-        if !self.pending_rx.remove(PendingRx::Rw) {
-            return Err(VsockError::NoData);
-        }
-
-        match self.state {
-            // A data packet is only valid for established connections, and connections for
-            // which our peer has initiated a graceful shutdown, but can still receive data.
-            ConnState::Established | ConnState::PeerClosed(false, _) => (),
-            _ => {
-                // Any other connection state is invalid at this point, and we need to kill it
-                // with fire.
-                pkt.set_op(uapi::VSOCK_OP_RST);
-                return Ok(());
-            }
-        }
-
-        // Oh wait, before we start bringing in the big data, can our peer handle receiving so
-        // much bytey goodness?
-        if self.need_credit_update_from_peer() {
-            self.last_fwd_cnt_to_peer = self.fwd_cnt;
-            pkt.set_op(uapi::VSOCK_OP_CREDIT_REQUEST);
-            return Ok(());
-        }
-
-        let buf = pkt.buf_mut().ok_or(VsockError::PktBufMissing)?;
-
-        // The maximum amount of data we can read in is limited by both the RX buffer size and
-        // the peer available buffer space.
-        let max_len = std::cmp::min(buf.len(), self.peer_avail_credit());
-
-        // Read data from the stream straight to the RX buffer, for maximum throughput.
-        match self.stream.read(&mut buf[..max_len]) {
-            Ok(read_cnt) => {
-                if read_cnt == 0 {
-                    // A 0-length read means the host stream was closed down. In that case,
-                    // we'll ask our peer to shut down the connection. We can neither send nor
-                    // receive any more data.
-                    self.state = ConnState::LocalClosed;
-                    self.expiry = Some(
-                        Instant::now() + Duration::from_millis(defs::CONN_SHUTDOWN_TIMEOUT_MS),
-                    );
-                    pkt.set_op(uapi::VSOCK_OP_SHUTDOWN)
-                        .set_flag(uapi::VSOCK_FLAGS_SHUTDOWN_RCV)
-                        .set_flag(uapi::VSOCK_FLAGS_SHUTDOWN_SEND);
-                } else {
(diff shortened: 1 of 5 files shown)


sandreim:
Is the issue fixed by this PR ?

gc-plp:
> Is the issue fixed by this PR ?

Hi Andrei. I added more details in the comment [here](https://github.com/firecracker-microvm/firecracker/issues/1751#issuecomment-630281988)

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -143,16 +143,22 @@ impl VsockChannel for VsockMuxer {
...
                         .set_fwd_cnt(0);
+                    self.rxq.pop().unwrap();
                     return Ok(());
                 }
 
                 // We'll defer building the packet to this connection, since it has something
                 // to say.
                 MuxerRx::ConnRx(key) => {
                     let mut conn_res = Err(VsockError::NoData);
+                    let mut do_pop = true;
```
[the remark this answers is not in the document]
gc-plp: > it is necessary to keep the connection inside the muxer rxq after handling a pending RX instruction for that specific connection

The purpose of this change is to keep the connection inside the `rxq`, in the situation where there are multiple flags pending.

Previously to this, the muxer would permanently pop a connection from the `rxq` even if it had multiple flags pending, which made the unhandled flag to be dropped and lost forever. If that flag would have been a credit update, the flow control would be broken.

I agree that instructions for one connection would be prioritized, but since the instructions are flag based and multiple similar instructions cannot be queued at the same time, I don't think that there is a significant impact on neighboring connections. 
We could implement a _fair_ scheduler, but we also need to keep in mind that this code path is executed very often and we need an efficient way to look-up flags and pop elements.

If you have a proposal on this, let's discuss it.

src/devices/src/virtio/vsock/csm/connection.rs
```diff
@@ -439,7 +452,13 @@ where
                         "vsock: error flushing TX buf for (lp={}, pp={}): {:?}",
                         self.local_port, self.peer_port, err
                     );
-                    self.kill();
+                    match err {
+                        Error::TxBufFlush(inner) if inner.kind() == ErrorKind::WouldBlock => {
+                            // This should never happen (EWOULDBLOCK after EPOLLOUT), but
```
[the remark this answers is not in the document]
gc-plp: Yes, you can get `WouldBlock` errors on both read and write calls.

andreeaflorescu:
Since this looks like a corner case that wasn't handle, should we add a regression test here as well?

gc-plp:
Yes, we want to have regression tests, but I was thinking that we could add them in a separate PR.

andreeaflorescu:
Regression tests are typically merged in the same PR and the reviewers should check that the regression test fails before the fix and passes after the fix.

gc-plp:
@andreeaflorescu, unit tests added, please review.

src/devices/src/virtio/vsock/csm/connection.rs
```diff
@@ -655,6 +655,17 @@ where
     }
 }
 
+#[cfg(test)]
+impl<S> VsockConnection<S>
```
iulianbarbu, line 659: Does it makes sense to add this impl inside the `mod tests` and then make the module `pub`?

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -1339,4 +1339,33 @@ mod tests {
...
         assert_eq!(conn.fwd_cnt().0, 0);
     }
+
+    #[test]
+    fn test_regression_rxq_pop() {
+        let mut ctx = MuxerTestContext::new("regression_rxq_pop");
+        let peer_port = 1025;
+        let (mut stream, local_port) = ctx.local_connect(peer_port);
+
+        // Send some data
```
iulianbarbu, line 1349: Nit: full stop at the end.

src/devices/src/virtio/vsock/csm/connection.rs
gc-plp: We could do that, but I think that this is the least invasive way of accessing private members declared in `connection.rs`

src/devices/src/virtio/vsock/unix/muxer.rs
gc-plp: Done

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -1320,4 +1320,23 @@ mod tests {
 
         assert!(!ctx.muxer.has_pending_rx());
     }
+
+    #[test]
+    fn test_regression_handshake() {
+        let mut ctx = MuxerTestContext::new("regression_handshake");
+        let peer_port = 1025;
+
+        // Create a local connection
```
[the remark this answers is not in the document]
gc-plp: Done

src/devices/src/virtio/vsock/csm/connection.rs
iulianbarbu: I do not have a strong preference. My idea was to group all the tests relevant information in the `mod tests`, but I can not say that keeping this implementation outside the `mod tests` would counterfeit my idea of grouping. Eventually, we can group in the future such implementations for different structs in the `mod tests`, if they increase in number.

src/devices/src/virtio/vsock/csm/connection.rs
```diff
@@ -619,11 +650,27 @@ where
...
             .set_fwd_cnt(self.fwd_cnt.0)
     }
 }
 
+#[cfg(test)]
+impl<S> VsockConnection<S>
+where
+    S: Read + Write + AsRawFd,
+{
+    /// Get the fwd_cnt value from the connection
```
iulianbarbu, line 663: Nit: full stop as well.

src/devices/src/virtio/vsock/csm/connection.rs
lauralt: > We could do that, but I think that this is the least invasive way of accessing private members declared in `connection.rs`

You can extend the struct implementation in `mod tests` without doing any other change and still be able to access private members; also the test module doesn't need to be `pub`. It's nice to keep the testing functionality in tests module because those functions won't be included in the `kcov` report, as opposed to how it is now implemented.

src/devices/src/virtio/vsock/csm/connection.rs
gc-plp: Done

src/devices/src/virtio/vsock/csm/connection.rs
gc-plp: Yep, you're right, that works too.

iulianbarbu: As far as I understood, this impl is used outside the mod tests, so the private members are not available there. That's why the methods must be declared as pub(crate) (combined with the fact that the struct is public).

Otherwise, if the impl would be included inside the mod tests, the module would need to be public in order to make use from the public interface exposed by the implementation outside the module/file scope.

src/devices/src/virtio/vsock/csm/connection.rs
```diff
@@ -749,6 +749,17 @@ mod tests {
         }
     }
 
+    #[cfg(test)]
```
andreeaflorescu, line 752: You don't need to specify the `cfg(test)` macro because this implementation is defined in `mod tests` which is all only compiled for tests.

src/devices/src/virtio/vsock/csm/connection.rs
```diff
@@ -749,6 +749,17 @@ mod tests {
...
+    impl<S> VsockConnection<S>
+    where
+        S: Read + Write + AsRawFd,
+    {
+        /// Get the fwd_cnt value from the connection.
+        pub(crate) fn fwd_cnt(&mut self) -> Wrapping<u32> {
```
andreeaflorescu, line 758: You shouldn't need `mut` here.

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -1320,4 +1320,23 @@ mod tests {
 
         assert!(!ctx.muxer.has_pending_rx());
     }
+
+    #[test]
+    fn test_regression_handshake() {
```
andreeaflorescu, line 1325: What I noticed at another PR is that they also added a comment with the GitHub issue link that was related to the regression test. It would be nice to have it here as well.

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -1339,4 +1339,33 @@ mod tests {
...
+        let conn = ctx.muxer.conn_map.get_mut(&key).unwrap();
+
+        // Forcefully insert another flag.
+        conn.insert_credit_update();
+
+        // Call recv twice in order to check that the connection is still
+        // in the rxq.
+        assert!(ctx.muxer.has_pending_rx());
+        ctx.recv();
+        assert!(ctx.muxer.has_pending_rx());
```
andreeaflorescu, line 1368: Shouldn't the assert be after the call to `recv`?

src/devices/src/virtio/vsock/csm/connection.rs
```diff
@@ -1048,21 +1095,29 @@ mod tests {
 
         // Force a stale state, where the peer hasn't been updated on our credit situation.
         ctx.conn.last_fwd_cnt_to_peer = Wrapping(0);
-        ctx.conn.fwd_cnt = Wrapping(csm_defs::CONN_CREDIT_UPDATE_THRESHOLD as u32);
 
-        // Fake a data send from the peer, to bring us over the credit update threshold.
+        // Initialize the fwd counter at 6 bytes below the threshold
```
lauralt, line 1099: This comment is a bit misleading, as I would expect something like this: `initial_fwd_cnd = csm_defs::CONN_CREDIT_UPDATE_THRESHOLD as u32 - 6;` Can we update the comment or rename `CONN_CREDIT_UPDATE_THRESHOLD`?
Also, nit: full stop for all comments in this test.

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -1339,4 +1339,33 @@ mod tests {
         // Check that fwd_cnt is 0 - "OK ..." was not accounted for.
         assert_eq!(conn.fwd_cnt().0, 0);
     }
+
+    #[test]
+    fn test_regression_rxq_pop() {
```
andreeaflorescu, line 1344: Same comment as for the other test. It would be useful to also have a link to the GitHub issue that this regression test is addressing.

src/devices/src/virtio/vsock/csm/connection.rs
```diff
@@ -718,6 +749,22 @@ mod tests {
         }
     }
 
+    #[cfg(test)]
```
lauralt, line 752: this can be removed.

src/devices/src/virtio/vsock/csm/connection.rs
```diff
@@ -1048,21 +1095,29 @@ mod tests {
 
         // Force a stale state, where the peer hasn't been updated on our credit situation.
         ctx.conn.last_fwd_cnt_to_peer = Wrapping(0);
-        ctx.conn.fwd_cnt = Wrapping(csm_defs::CONN_CREDIT_UPDATE_THRESHOLD as u32);
 
-        // Fake a data send from the peer, to bring us over the credit update threshold.
+        // Initialize the fwd counter at 6 bytes below the threshold
+        let initial_fwd_cnd =
```
lauralt, line 1100:
```suggestion
        let initial_fwd_cnt =
```

src/devices/src/virtio/vsock/csm/connection.rs
gc-plp: Added a more comprehensive comment.

src/devices/src/virtio/vsock/csm/connection.rs
gc-plp: Removed

src/devices/src/virtio/vsock/csm/connection.rs
gc-plp: Typo fixed

src/devices/src/virtio/vsock/csm/connection.rs
gc-plp: Removed

src/devices/src/virtio/vsock/csm/connection.rs
gc-plp: `mut` removed

src/devices/src/virtio/vsock/unix/muxer.rs
gc-plp: Link added

src/devices/src/virtio/vsock/unix/muxer.rs
gc-plp: Link added

src/devices/src/virtio/vsock/unix/muxer.rs
gc-plp: Actually, one assert was missing, because first of all we check that there is something pending, then we `recv`. After the first `recv` we should still have something pending, so we `recv` again.

Finally, we check that there is nothing pending.

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -1305,4 +1320,63 @@ mod tests {
 
         assert!(!ctx.muxer.has_pending_rx());
     }
+
+    #[test]
+    fn test_regression_handshake() {
+        // Address one of the issues found while fixing the following issue:
+        // https://github.com/firecracker-microvm/firecracker/pull/1910
```
lauralt, line 1327: Hmm, actually this links to a PR, not issue; same for the next test.

gc-plp: Pasted the wrong link :)

Fixed now.

PR #1912, opened 2020-05-19
Reregister devices with event-manager on restore

acatangiu:
## Reason for This PR

Part of #1847 

## Description of Changes

Small reworking of devices event registering logic. Devices now own reporting their `interest-list` depending on their `activate` state. The mechanism to register events now becomes context independent and can be used and is used when restoring devices from snapshot.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/event_handler.rs
+++ b/src/devices/src/virtio/block/event_handler.rs
@@ -10,40 +10,34 @@ use crate::virtio::VirtioDevice;
 
 impl Block {
     fn process_activate_event(&self, event_manager: &mut EventManager) {
+        debug!("block: activate event");
+        if let Err(e) = self.activate_evt.read() {
+            error!("Failed to consume block activate event: {:?}", e);
+        }
+        let activate_fd = self.activate_evt.as_raw_fd();
         // The subscriber must exist as we previously registered activate_evt via
         // `interest_list()`.
-        let self_subscriber = event_manager
-            .subscriber(self.activate_evt.as_raw_fd())
-            .unwrap();
+        let self_subscriber = match event_manager.subscriber(activate_fd) {
+            Ok(subscriber) => subscriber,
+            Err(e) => {
+                error!("Failed to process block activate evt: {:?}", e);
+                return;
+            }
+        };
 
-        event_manager
-            .register(
-                self.queue_evts[0].as_raw_fd(),
-                EpollEvent::new(EventSet::IN, self.queue_evts[0].as_raw_fd() as u64),
-                self_subscriber.clone(),
-            )
-            .unwrap_or_else(|e| {
-                error!("Failed to register block queue with event manager: {:?}", e);
-            });
+        // Interest list changes when the device is activated.
+        let interest_list = self.interest_list();
+        for event in interest_list {
+            event_manager
+                .register(event.data() as i32, event, self_subscriber.clone())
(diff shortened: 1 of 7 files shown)


src/devices/src/virtio/block/event_handler.rs
```diff
@@ -85,10 +79,17 @@ impl Subscriber for Block {
     }
 
     fn interest_list(&self) -> Vec<EpollEvent> {
-        vec![EpollEvent::new(
-            EventSet::IN,
-            self.activate_evt.as_raw_fd() as u64,
-        )]
+        if self.is_activated() {
```
sandreim, line 82: I would add a comment here to tell the story of when this happens. FreshVM, Restored clone.

src/devices/src/virtio/net/event_handler.rs
```diff
@@ -12,82 +12,34 @@ use crate::virtio::{VirtioDevice, RX_INDEX, TX_INDEX};
 
 impl Net {
     fn process_activate_event(&self, event_manager: &mut EventManager) {
+        debug!("net: activate event");
+        if let Err(e) = self.activate_evt.read() {
+            error!("Failed to consume net activate event: {:?}", e);
```
sandreim, line 17: Do we need to bail out and never activate this device ?

acatangiu: The device is already activated at this point, I think this falls under the category of: log error and carry on. These are highly unlikely errors which used to be `unwrap()`s before and we switched them to logging to make sure we don't crash the whole guest on a device error.

The perfect solution would probably be to reset the device on such errors, but we don't have that support yet.

src/devices/src/virtio/block/event_handler.rs
acatangiu: done

src/devices/src/virtio/vsock/event_handler.rs
```diff
@@ -113,70 +113,30 @@ where
...
-            });
-
-        event_manager
-            .unregister(self.activate_evt.as_raw_fd())
-            .unwrap_or_else(|e| {
-                error!("Failed to unregister vsock activate evt: {:?}", e);
-            })
+        event_manager.unregister(activate_fd).unwrap_or_else(|e| {
+            error!("Failed to unregister vsock activate evt: {:?}", e);
+        });
```
ioanachirca, line 139: `handle_activate_event` should be `process_activate_event` to match the corresponding methods in the other devices. Or even better, we could define an interface for this since the implementation is exactly the same for all devices.

acatangiu: An interface would definitely be nice for code de-duplication, but the snag I hit when trying to do it is that it doesn't really belong in either `VirtioDevice` or `Subscriber`; it would have to be a `trait SubscriberVirtioDevice: Subscriber + VirtioDevice` and it would live in a separate file under devices so that we don't make a mess of things so the end result adds enough extra complexity that I just feel it's not worth it.

I'm not against it, but I don't see a nice, clear and simple way to do it.

ioanachirca: I think this is fine and more elegant than having the same method in 3 places (or more if we're going to add devices). We can log an issue for this improvement in order to think a bit more about it.

PR #1916, opened 2020-05-20
Draft network latency test

goandrei:
Signed-off-by: Andrei Gorneanu <gorneanu@amazon.com>

## Reason for This PR

Related to #1808 

## Description of Changes

This tests the latency of a Firecracker microvm by pinging the host. 

- added a goldfile which defines how the assertion should be done. It contains a mean value and a possible deviation from it.
- the ping command sends 500 requests with an interval of 0.2 seconds between them.

The tests won't work without a specialized rootfs which has iputils-ping installed for both x86 and aarch64.

## License Acceptance

## PR Checklist


tests/goldfile.json
```diff
@@ -0,0 +1,15 @@
+{
```
raduiliescu, line 1: I would put the goldfile in the same directory as the other code - tests/integration-tests/performance

tests/integration_tests/performance/test_net_latency.py
```diff
@@ -0,0 +1,73 @@
...
+    return stdout.read()
+
+
+def get_statistics(output):
+    """Get statistics out of a ping result."""
+    output = output.split('\n')[-2]
+    return re.findall(FLOAT_REG, output)
+
+
+def check_results(result):
```
raduiliescu, line 45: would be also nice to test the p90 or something like that. average can be misleading. Also should check for lost packets. What happens if only one of the pings worked and the time for that is better than gold?

tests/goldfile.json
goandrei: I will do this

tests/integration_tests/performance/test_net_latency.py
goandrei: That's a good point, I will do this check and add some perencetiles too

tests/integration_tests/performance/test_net_latency.py
```diff
@@ -0,0 +1,109 @@
...
+
+    _tap, host_ip, _ = microvm.ssh_network_config(network_config, '1')
+    microvm.start()
+
+    ssh_connection = net_tools.SSHConnection(microvm.ssh_config)
+
+    # Ping g2h
+    g2h = do_ping_on_guest(host_ip, ssh_connection)
+    g2h_statistics = get_statistics(g2h)
+    print(g2h_statistics)
```
gc-plp, line 107: Is this print statement useful? 
Looking at the code, it seems to just print a dict(). If we really want to have logged messages stating the violation of performance values, we could add a reason to the `assert` statements, such as `assert value >= limit, "value less than limit"`.

tests/integration_tests/performance/test_net_latency.py
```diff
@@ -0,0 +1,109 @@
...
+
+import host_tools.network as net_tools
+
+# Number of requests issued.
+REQUESTS = 1000
+
+# Interval time between requests.
+INTERVAL = 0.2
+
+PING = 'ping -c {} -i {} {}'
```
gc-plp, line 18: nit: you could compact the global variable list.

tests/integration_tests/performance/test_net_latency.py
goandrei: It's purpose was to display the percentiles too, even if the test failed or not. But, if the test passes they won't be showed anyway, so it's indeed useless. I think it would be better if I would dump the dict() in a json.

tests/integration_tests/performance/goldfile.json
```diff
@@ -0,0 +1,15 @@
+{
+    "network_latency": {
+        "description" : "Average network latency for a Firecracker VM. The results are round-trip.",
+        "x86_64" : {
+            "target" : 0.140,
```
iulianbarbu, line 5: What units? Can we mention the units in the description?

goandrei: Good idea. I will update this

iulianbarbu: Just to double check: we have here 0.140 milliseconds or 140 milliseconds?

goandrei: These are 0.140 milliseconds

iulianbarbu: Cool. Thanks.

acatangiu:
@gc-plp @iulianbarbu @raduiliescu This PR looks good and the functionality useful, do we want to merge it?

sandreim:
> @gc-plp @iulianbarbu @raduiliescu This PR looks good and the functionality useful, do we want to merge it?

IMO we should merge it after performance framework/ci changes are done.

iulianbarbu:
> > @gc-plp @iulianbarbu @raduiliescu This PR looks good and the functionality useful, do we want to merge it?
> 
> IMO we should merge it after performance framework/ci changes are done.

Indeed, that's how I see it. I think we can let it open until we put to rest the two RFCs pending around testing framework ( #2041 and #2051 ), which will provide reproducible runs for this test.

iulianbarbu:
Close in favor of #2051 .

Issue #1865, opened 2020-05-05
Create a simple mmio device to support logging the guest boot-time


acatangiu:
Right now Firecracker has some hardcoded logic for a `PIO` address on `x86_64` and a `MMIO` address on `aarch64` to trigger logging of elapsed time since `InstanceStart` command.
See https://github.com/firecracker-microvm/firecracker/blob/76dfe3af573f6fa641efc192896c6b42bc15b60d/src/vmm/src/vstate.rs#L1032 and https://github.com/firecracker-microvm/firecracker/blob/76dfe3af573f6fa641efc192896c6b42bc15b60d/src/vmm/src/vstate.rs#L1048

This is both inefficient (check against this magic address and value happens on every `MmioWrite` and `IoWrite`) and hacky (this mmio address space is not managed in any way).

We should unify this mechanism to work over `Mmio` for both `aarch64` and `x86_64` and provide it through a "device" that implements `trait BusDevice` so it can be properly managed by a device manager.

lauralt:
#1921

PR #1921, opened 2020-05-22
devices: btimer: Add boot timer device.

tnotw:
Proposed solution to #1865 

Still work in progress and untested, but at a high-level is this going in the right direction?

Where I've: 

1. Added a new pseudo `BootTimer` device 
2. Added it to the first slot on the MMIO bus
3. Removed the boot signal handling and timestamping from Vcpu (AFAICT,  the timestamp was only used for the boot log time)
4. Removed the logging statement to the BootTimer device.  

Still got to: 

- Write an `/sbin/init` wrapper for x86_64 in order to generate the signal at the correct address.
- Complete the "logic" of the device.
- Test.
- Add missing tests. 

Hope that makes sense.

## Reason for This PR

## Description of Changes

Add a device to improve kernel boot time logging. 

## License Acceptance

## PR Checklist

### DIFF
--- a/resources/tests/init.c
+++ b/resources/tests/init.c
@@ -1,0 +1,44 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Init wrapper for boot timing. Alpine-specific because it points at /sbin/openrc-init.
+
+#include <sys/types.h>
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+// Base address values are defined in arch/src/lib.rs as arch::MMIO_MEM_START.
+// Values are computed in arch/src/<arch>/mod.rs from the architecture layouts.
+// Position on the bus is defined by MMIO_LEN increments, where MMIO_LEN is
+// defined as 0x1000 in vmm/src/device_manager/mmio.rs.
+#ifdef __x86_64__
+#define MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE 0xd0000000
+#endif
+#ifdef __aarch64__
+#define MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE 0x40000000
+#endif
+
+#define MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE 123
+
+int main () {
+   int fd = open("/dev/mem", (O_RDWR | O_SYNC | O_CLOEXEC));
+   int mapped_size = getpagesize();
+
+   char *map_base = mmap(NULL,
+        mapped_size,
+        PROT_WRITE,
+        MAP_SHARED,
+        fd,
+        MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE);
+
+   *map_base = MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE;
+   msync(map_base, mapped_size, MS_ASYNC);
+
+   const char *init = "/sbin/openrc-init";
+
+   char *const argv[] = { "/sbin/init", NULL };
+   char *const envp[] = { };
+
+   execve(init, argv, envp);
+}
--- a/src/devices/src/pseudo/boot_timer.rs
+++ b/src/devices/src/pseudo/boot_timer.rs
@@ -1,0 +1,42 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use utils::time::TimestampUs;
+
+use crate::bus::BusDevice;
+
+const MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE: u8 = 123;
+
+/// Pseudo device to record the kernel boot time.
+pub struct BootTimer {
+    start_ts: TimestampUs,
+}
+
+impl BusDevice for BootTimer {
+    fn write(&mut self, offset: u64, data: &[u8]) {
+        // Only handle byte length instructions at a zero offset.
+        if data.len() != 1 || offset != 0 {
+            return;
+        }
+
+        if data[0] == MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE {
+            let now_tm_us = TimestampUs::default();
+
+            let boot_time_us = now_tm_us.time_us - self.start_ts.time_us;
+            let boot_time_cpu_us = now_tm_us.cputime_us - self.start_ts.cputime_us;
+            info!(
+                "Guest-boot-time = {:>6} us {} ms, {:>6} CPU us {} CPU ms",
+                boot_time_us,
+                boot_time_us / 1000,
+                boot_time_cpu_us,
+                boot_time_cpu_us / 1000
+            );
+        }
+    }
+}
+
+impl BootTimer {
+    pub fn new(start_ts: TimestampUs) -> BootTimer {
+        BootTimer { start_ts }
+    }
+}
--- a/src/devices/src/pseudo/mod.rs
+++ b/src/devices/src/pseudo/mod.rs
@@ -1,0 +1,6 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+mod boot_timer;
+
+pub use self::boot_timer::BootTimer;
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -204,8 +204,6 @@ fn create_vmm_and_vcpus(
     vcpu_count: u8,
 ) -> std::result::Result<(Vmm, Vec<Vcpu>), StartMicrovmError> {
     use self::StartMicrovmError::*;
-    // Timestamp for measuring microVM boot duration.
-    let request_ts = TimestampUs::default();
 
     // Set up Kvm Vm and register memory regions.
     let mut vm = setup_kvm_vm(&guest_memory, track_dirty_pages)?;
@@ -218,11 +216,8 @@ fn create_vmm_and_vcpus(
     // Instantiate the MMIO device manager.
     // 'mmio_base' address has to be an address which is protected by the kernel
     // and is architectural specific.
-    #[allow(unused_mut)]
-    let mut mmio_device_manager = MMIODeviceManager::new(
-        &mut (arch::MMIO_MEM_START as u64),
-        (arch::IRQ_BASE, arch::IRQ_MAX),
-    );
+    let mmio_device_manager =
+        MMIODeviceManager::new(arch::MMIO_MEM_START, (arch::IRQ_BASE, arch::IRQ_MAX));
 
     let vcpus;
     // For x86_64 we need to create the interrupt controller before calling `KVM_CREATE_VCPUS`
@@ -230,7 +225,7 @@ fn create_vmm_and_vcpus(
     #[cfg(target_arch = "x86_64")]
     let pio_device_manager = {
         setup_interrupt_controller(&mut vm)?;
-        vcpus = create_vcpus(&vm, vcpu_count, request_ts, &exit_evt).map_err(Internal)?;
+        vcpus = create_vcpus(&vm, vcpu_count, &exit_evt).map_err(Internal)?;
 
         // Serial device setup.
         let serial_device = setup_serial_device(
@@ -254,7 +249,7 @@ fn create_vmm_and_vcpus(
     // Search for `kvm_arch_vcpu_create` in arch/arm/kvm/arm.c.
     #[cfg(target_arch = "aarch64")]
     {
-        vcpus = create_vcpus(&vm, vcpu_count, request_ts, &exit_evt).map_err(Internal)?;
+        vcpus = create_vcpus(&vm, vcpu_count, &exit_evt).map_err(Internal)?;
         setup_interrupt_controller(&mut vm, vcpu_count)?;
     }
 
@@ -301,12 +296,17 @@ pub fn build_microvm_for_boot(
     #[allow(unused_mut)]
     let mut boot_cmdline = boot_config.cmdline.clone();
 
+    // Timestamp for measuring microVM boot duration.
+    let request_ts = TimestampUs::default();
+
     let (mut vmm, mut vcpus) = create_vmm_and_vcpus(
         event_manager,
         guest_memory,
         track_dirty_pages,
         vcpu_config.vcpu_count,
     )?;
+
+    attach_boot_timer_device(&mut vmm, request_ts)?;
 
     attach_block_devices(
         &mut vmm,
@@ -601,29 +601,17 @@ fn attach_legacy_devices_aarch64(
         .map_err(Error::RegisterMMIODevice)
 }
 
-fn create_vcpus(
-    vm: &Vm,
-    vcpu_count: u8,
-    request_ts: TimestampUs,
-    exit_evt: &EventFd,
-) -> super::Result<Vec<Vcpu>> {
+fn create_vcpus(vm: &Vm, vcpu_count: u8, exit_evt: &EventFd) -> super::Result<Vec<Vcpu>> {
     let mut vcpus = Vec::with_capacity(vcpu_count as usize);
     for cpu_idx in 0..vcpu_count {
         let exit_evt = exit_evt.try_clone().map_err(Error::EventFd)?;
 
         #[cfg(target_arch = "x86_64")]
-        let vcpu = Vcpu::new_x86_64(
-            cpu_idx,
-            vm.fd(),
-            vm.supported_msrs().clone(),
-            exit_evt,
-            request_ts.clone(),
-        )
-        .map_err(Error::Vcpu)?;
+        let vcpu = Vcpu::new_x86_64(cpu_idx, vm.fd(), vm.supported_msrs().clone(), exit_evt)
+            .map_err(Error::Vcpu)?;
 
         #[cfg(target_arch = "aarch64")]
-        let vcpu = Vcpu::new_aarch64(cpu_idx, vm.fd(), exit_evt, request_ts.clone())
-            .map_err(Error::Vcpu)?;
+        let vcpu = Vcpu::new_aarch64(cpu_idx, vm.fd(), exit_evt).map_err(Error::Vcpu)?;
 
         vcpus.push(vcpu);
     }
@@ -713,6 +701,21 @@ fn attach_virtio_device<T: 'static + VirtioDevice + Subscriber>(
         .register_new_virtio_mmio_device(vmm.vm.fd(), id, device, cmdline)
         .map_err(RegisterMmioDevice)
         .map(|_| ())
+}
+
+pub(crate) fn attach_boot_timer_device(
+    vmm: &mut Vmm,
+    request_ts: TimestampUs,
+) -> std::result::Result<(), StartMicrovmError> {
+    use self::StartMicrovmError::*;
+
+    let boot_timer = devices::pseudo::BootTimer::new(request_ts);
+
+    vmm.mmio_device_manager
+        .register_new_mmio_boot_timer(boot_timer)
+        .map_err(RegisterMmioDevice)?;
+
+    Ok(())
 }
 
 fn attach_block_devices<'a>(
@@ -808,10 +811,7 @@ pub mod tests {
     }
 
     fn default_mmio_device_manager() -> MMIODeviceManager {
-        MMIODeviceManager::new(
-            &mut (arch::MMIO_MEM_START as u64),
-            (arch::IRQ_BASE, arch::IRQ_MAX),
-        )
+        MMIODeviceManager::new(arch::MMIO_MEM_START, (arch::IRQ_BASE, arch::IRQ_MAX))
     }
 
     #[cfg(target_arch = "x86_64")]
@@ -1004,7 +1004,7 @@ pub mod tests {
         #[cfg(target_arch = "x86_64")]
         setup_interrupt_controller(&mut vm).unwrap();
 
-        let vcpu_vec = create_vcpus(&vm, vcpu_count, TimestampUs::default(), &evfd).unwrap();
+        let vcpu_vec = create_vcpus(&vm, vcpu_count, &evfd).unwrap();
         assert_eq!(vcpu_vec.len(), vcpu_count as usize);
     }
 
@@ -1161,6 +1161,19 @@ pub mod tests {
                 .get_device(DeviceType::Virtio(TYPE_BLOCK), drive_id.as_str())
                 .is_some());
         }
+    }
+
+    #[test]
+    fn test_attach_boot_timer_device() {
+        let mut vmm = default_vmm();
+        let request_ts = TimestampUs::default();
+
+        let res = attach_boot_timer_device(&mut vmm, request_ts);
+        assert!(res.is_ok());
+        assert!(vmm
+            .mmio_device_manager
+            .get_device(DeviceType::BootTimer, &DeviceType::BootTimer.to_string())
+            .is_some());
     }
 
     #[test]
--- a/src/vmm/src/device_manager/mmio.rs
+++ b/src/vmm/src/device_manager/mmio.rs
@@ -13,6 +13,7 @@ use std::{fmt, io};
 use arch::aarch64::DeviceInfoForFDT;
 use arch::DeviceType;
 use devices;
+use devices::pseudo::BootTimer;
 use devices::{virtio::MmioTransport, BusDevice};
 use kernel::cmdline as kernel_cmdline;
 use kvm_ioctls::{IoEventAddress, VmFd};
@@ -93,15 +94,9 @@ pub struct MMIODeviceManager {
 
 impl MMIODeviceManager {
     /// Create a new DeviceManager handling mmio devices (virtio net, block).
-    pub fn new(mmio_base: &mut u64, irq_interval: (u32, u32)) -> MMIODeviceManager {
-        if cfg!(target_arch = "aarch64") {
-            // This hack is done to reserve the 'log boot time magic address'.
-            // TODO (issue #1865): we should create a simple mmio device for logging the boot time
-            // and use it on both x86_64 and aarch64, instead of this hacky thing we do right now.
-            *mmio_base += MMIO_LEN;
-        }
+    pub fn new(mmio_base: u64, irq_interval: (u32, u32)) -> MMIODeviceManager {
         MMIODeviceManager {
-            mmio_base: *mmio_base,
+            mmio_base,
             irq: irq_interval.0,
             last_irq: irq_interval.1,
             bus: devices::Bus::new(),
@@ -110,17 +105,21 @@ impl MMIODeviceManager {
     }
 
     /// Allocates resources for a new device to be added.
-    fn allocate_new_slot(&mut self) -> Result<MMIODeviceInfo> {
-        if self.irq > self.last_irq {
+    fn allocate_new_slot(&mut self, irq_count: u32) -> Result<MMIODeviceInfo> {
+        if self.irq + irq_count > self.last_irq + 1 {
             return Err(Error::IrqsExhausted);
+        }
+        let mut irqs = Vec::with_capacity(irq_count as usize);
+        for _ in 0..irq_count {
+            irqs.push(self.irq);
+            self.irq += 1;
         }
         let slot = MMIODeviceInfo {
             addr: self.mmio_base,
             len: MMIO_LEN,
-            irqs: vec![self.irq],
+            irqs,
         };
         self.mmio_base += MMIO_LEN;
-        self.irq += 1;
         Ok(slot)
     }
 
@@ -190,7 +189,7 @@ impl MMIODeviceManager {
         mmio_device: MmioTransport,
         _cmdline: &mut kernel_cmdline::Cmdline,
     ) -> Result<MMIODeviceInfo> {
-        let mmio_slot = self.allocate_new_slot()?;
+        let mmio_slot = self.allocate_new_slot(1)?;
         self.register_virtio_mmio_device(vm, device_id, mmio_device, &mmio_slot)?;
         #[cfg(target_arch = "x86_64")]
         Self::add_virtio_device_to_cmdline(_cmdline, &mmio_slot)?;
@@ -204,7 +203,7 @@ impl MMIODeviceManager {
         vm: &VmFd,
         serial: Arc<Mutex<devices::legacy::Serial>>,
     ) -> Result<()> {
-        let slot = self.allocate_new_slot()?;
+        let slot = self.allocate_new_slot(1)?;
         vm.register_irqfd(
             &serial.lock().expect("Poisoned lock").interrupt_evt(),
             slot.irqs[0],
@@ -231,13 +230,22 @@ impl MMIODeviceManager {
     /// Create and register a new MMIO RTC device.
     pub fn register_new_mmio_rtc(&mut self, vm: &VmFd) -> Result<()> {
         // Create and attach a new RTC device.
-        let slot = self.allocate_new_slot()?;
+        let slot = self.allocate_new_slot(1)?;
         let rtc_evt = EventFd::new(libc::EFD_NONBLOCK).map_err(Error::EventFd)?;
         let device = devices::legacy::RTC::new(rtc_evt.try_clone().map_err(Error::EventFd)?);
         vm.register_irqfd(&rtc_evt, slot.irqs[0])
             .map_err(Error::RegisterIrqFd)?;
 
         let identifier = (DeviceType::RTC, DeviceType::RTC.to_string());
+        self.register_mmio_device(identifier, slot, Arc::new(Mutex::new(device)))
+    }
+
+    /// Create and register a boot timer device.
+    pub fn register_new_mmio_boot_timer(&mut self, device: BootTimer) -> Result<()> {
+        // Create and attach a new boot timer device.
+        let slot = self.allocate_new_slot(0)?;
+
+        let identifier = (DeviceType::BootTimer, DeviceType::BootTimer.to_string());
         self.register_mmio_device(identifier, slot, Arc::new(Mutex::new(device)))
     }
 
@@ -393,7 +401,7 @@ mod tests {
             GuestMemoryMmap::from_ranges(&[(start_addr1, 0x1000), (start_addr2, 0x1000)]).unwrap();
         let mut vm = builder::setup_kvm_vm(&guest_mem, false).unwrap();
         let mut device_manager =
-            MMIODeviceManager::new(&mut 0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
+            MMIODeviceManager::new(0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
 
         let mut cmdline = kernel_cmdline::Cmdline::new(4096);
         let dummy = Arc::new(Mutex::new(DummyDevice::new()));
@@ -415,7 +423,7 @@ mod tests {
             GuestMemoryMmap::from_ranges(&[(start_addr1, 0x1000), (start_addr2, 0x1000)]).unwrap();
         let mut vm = builder::setup_kvm_vm(&guest_mem, false).unwrap();
         let mut device_manager =
-            MMIODeviceManager::new(&mut 0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
+            MMIODeviceManager::new(0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
 
         let mut cmdline = kernel_cmdline::Cmdline::new(4096);
         #[cfg(target_arch = "x86_64")]
@@ -460,8 +468,7 @@ mod tests {
 
     #[test]
     fn test_error_messages() {
-        let device_manager =
-            MMIODeviceManager::new(&mut 0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
+        let device_manager = MMIODeviceManager::new(0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
         let mut cmdline = kernel_cmdline::Cmdline::new(4096);
         let e = Error::Cmdline(
             cmdline
@@ -522,7 +529,7 @@ mod tests {
         assert!(builder::setup_interrupt_controller(&mut vm, 1).is_ok());
 
         let mut device_manager =
-            MMIODeviceManager::new(&mut 0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
+            MMIODeviceManager::new(0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
         let mut cmdline = kernel_cmdline::Cmdline::new(4096);
         let dummy = Arc::new(Mutex::new(DummyDevice::new()));
 
@@ -547,5 +554,35 @@ mod tests {
         assert!(device_manager
             .get_device(DeviceType::Virtio(type_id), &id)
             .is_none());
+    }
+
+    #[test]
+    fn test_slot_irq_allocation() {
+        let mut device_manager =
+            MMIODeviceManager::new(0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
+        let _addr = device_manager.allocate_new_slot(0);
+        assert_eq!(device_manager.irq, arch::IRQ_BASE);
+        let _addr = device_manager.allocate_new_slot(1);
+        assert_eq!(device_manager.irq, arch::IRQ_BASE + 1);
+        assert_eq!(
+            format!(
+                "{}",
+                device_manager
+                    .allocate_new_slot(arch::IRQ_MAX - arch::IRQ_BASE + 1)
+                    .unwrap_err()
+            ),
+            "no more IRQs are available".to_string()
+        );
+
+        let _addr = device_manager.allocate_new_slot(arch::IRQ_MAX - arch::IRQ_BASE - 1);
+        assert_eq!(device_manager.irq, arch::IRQ_MAX);
+        assert_eq!(
+            format!("{}", device_manager.allocate_new_slot(2).unwrap_err()),
+            "no more IRQs are available".to_string()
+        );
+
+        let _addr = device_manager.allocate_new_slot(1);
+        assert_eq!(device_manager.irq, arch::IRQ_MAX + 1);
+        assert!(device_manager.allocate_new_slot(0).is_ok());
     }
 }
--- a/src/vmm/src/device_manager/persist.rs
+++ b/src/vmm/src/device_manager/persist.rs
@@ -13,6 +13,7 @@ use std::sync::{Arc, Mutex};
 
 use super::mmio::*;
 
+use devices::pseudo::BootTimer;
 use devices::virtio::block::persist::{BlockConstructorArgs, BlockState};
 use devices::virtio::block::Block;
 use devices::virtio::net::persist::{Error as NetError, NetConstructorArgs, NetState};
@@ -24,6 +25,7 @@ use devices::virtio::{MmioTransport, TYPE_BLOCK, TYPE_NET, TYPE_VSOCK};
 use kvm_ioctls::VmFd;
 use polly::event_manager::{Error as EventMgrError, EventManager};
 use snapshot::Persist;
+use utils::time::TimestampUs;
 use versionize::{VersionMap, Versionize, VersionizeResult};
 use versionize_derive::Versionize;
 use vm_memory::GuestMemoryMmap;
@@ -86,7 +88,7 @@ pub struct DeviceStates {
     pub block_devices: Vec<ConnectedBlockState>,
     /// Net device states.
     pub net_devices: Vec<ConnectedNetState>,
-    /// Vsock device tests.
+    /// Vsock device state.
     pub vsock_device: Option<ConnectedVsockState>,
 }
 
@@ -114,6 +116,11 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                 .unwrap()
                 .lock()
                 .expect("Poisoned lock");
+
+            if let Some(boot_timer) = bus_device.as_any().downcast_ref::<BootTimer>() {
+                // No need to save BootTimer state.
+                continue;
+            }
 
             let mmio_transport = bus_device
                 .as_any()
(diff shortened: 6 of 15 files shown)


iulianbarbu:
Hey, @tnotw ! Thanks for your PR. To me looks promising.
While we're looking at it, please let us know if you need any help with it.

tnotw:
Hi @iulianbarbu,  thanks very much for taking a look!

I think I have a snapshot test failure of some sort to go and investigate.  I assume that adding a device has had some impact I've not thought about when taking snapshots.  To be honest, I've not really been following the snapshot work. 

Also,  I think using a slot on the bus consumes an IRQ,  which will be unused (?).  I've got a vague suspicion this might be a bit wasteful on `x86_64`, where IRQ numbers seem limited, and was wondering if I should code round this somehow?

src/vmm/src/builder.rs
```diff
@@ -273,13 +284,16 @@ pub fn build_microvm(
         (arch::IRQ_BASE, arch::IRQ_MAX),
     );
 
+    // attach boot timer as the first mmio device on either architecture
+    attach_boot_timer_device(&mut mmio_device_manager, request_ts)?;
```
acatangiu, line 288: Please move this next to the other device attach functions (further down in this function).

src/vmm/src/builder.rs
```diff
@@ -628,6 +635,19 @@ pub fn configure_system_for_boot(
...
+    request_ts: TimestampUs,
+) -> std::result::Result<(), StartMicrovmError> {
+    use self::StartMicrovmError::*;
+
+    mmio_device_manager
+        .register_new_mmio_boot_timer(request_ts.clone())
+        .map_err(RegisterBootTimerDevice)?;
+
+    Ok(())
+}
```
acatangiu, line 649: The boot-timer device shouldn't be created by `mmio_device_manager.register_new_mmio_boot_timer()`. That function should take the boot-timer-dev as a param and only register it.

The boot-timer device creation can be done in this function in `builder.rs`

src/vmm/src/device_manager/mmio.rs
```diff
@@ -234,6 +229,17 @@ impl MMIODeviceManager {
         self.register_mmio_device(identifier, slot, Arc::new(Mutex::new(device)))
     }
 
+    /// Create and register a boot timer device.
+    pub fn register_new_mmio_boot_timer(&mut self, request_ts: TimestampUs) -> Result<()> {
+        // Create and attach a new boot timer device.
+        let slot = self.allocate_new_slot()?;
```
acatangiu, line 235: You can modify `allocate_new_slot()` to take `irq_count` as a param and allocate the right number of IRQs - in this case `0`.

src/vmm/src/builder.rs
tnotw: Done.

src/vmm/src/builder.rs
tnotw: Done.

src/vmm/src/device_manager/mmio.rs
tnotw: Done,  although I have some reservations about the implementation.

tnotw:
Thanks @acatangiu for your review and feedback.   

I've attempted to address your points above and have added an `init` implementation.  I did have a `build.rs` script to compile it,  but when I updated the container recently `musl-tools` (and hence `musl-gcc`) had been removed, which scuppered that idea.  Not sure how sensible it was anyway. 

Still working on the `persist` part....

acatangiu:
> Thanks @acatangiu for your review and feedback.
> 
> I've attempted to address your points above and have added an `init` implementation. I did have a `build.rs` script to compile it, but when I updated the container recently `musl-tools` (and hence `musl-gcc`) had been removed, which scuppered that idea. Not sure how sensible it was anyway.
> 
> Still working on the `persist` part....

Why do you need `musl-tools`? That was used to compile C code using `musl`.

If you use rust, the container should have everything you need for both `rust musl` binaries and `rust glibc` binaries.
You can also write the `init` binary in C using glibc. There should be everything you need for it in the container.

tnotw:
> 
> Why do you need `musl-tools`? That was used to compile C code using `musl`.
> 
> If you use rust, the container should have everything you need for both `rust musl` binaries and `rust glibc` binaries.
> You can also write the `init` binary in C using glibc. There should be everything you need for it in the container.

Right, sorry, I though maybe there was some reason for it to be in C and that `muscl-gcc' was required compiled to it.  Maybe wrong on both counts.  Want me to redo it in rust?

acatangiu:
> Right, sorry, I though maybe there was some reason for it to be in C and that `muscl-gcc' was required compiled to it. Maybe wrong on both counts. Want me to redo it in rust?

Either works. Like I said, a small C program compiled against glibc (using default `gcc`) should work nicely.

tnotw:
Ok,  well it was contained in that last push at `./src/device/src/pseudo/init.c`,  if you did want to take a peek.  I appreciate that none of this stuff is a priority. 

The reason I believed 'musl-gcc' was required was because a gcc built object wouldn't run on the musl-based Alpine image.  It couldn't find the correct library somehow.  The `musl-gcc' built object could.  Probably some lack of understanding on my part.  I will investigate further.

src/devices/src/pseudo/init.c
```diff
@@ -0,0 +1,40 @@
...
+#endif
+#ifdef __aarch64__
+#define MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE 0x40001000
+#endif
+
+#define MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE 123
+
+int main () {
+
+   int fd = open("/dev/mem", (O_RDWR | O_SYNC));
```
acatangiu, line 20: The mapping goes away on `exec()` but this `fd` stays on. Add the `O_CLOEXEC` flag here:
```suggestion
   int fd = open("/dev/mem", (O_RDWR | O_SYNC | O_CLOEXEC));
```

src/devices/src/pseudo/init.c
```diff
@@ -0,0 +1,40 @@
...
+   char *map_base = mmap(NULL,
+        mapped_size,
+        PROT_WRITE,
+        MAP_SHARED,
+        fd,
+        MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE);
+
+   char *virt_addr = (char*)map_base;
+
+   *(volatile char*)virt_addr = MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE;
```
acatangiu, line 32: Marking this as `volatile` isn't necessary, and neither does it guarantee that the write will go through without IO buffering.

I found this thread to be useful: https://stackoverflow.com/questions/45753903/memory-mapped-files-and-pointers-to-volatile-objects

In this case you could do an `msync()` after the assignment to flush the data to `/dev/mem`:

```suggestion
   *map_base = MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE;
   msync(map_base, mapped_size, MS_ASYNC);
```

You can use either `MS_ASYNC` or `MS_SYNC` here. I went for `MS_ASYNC` since we don't **need** to wait for the flush before continuing our init process.

src/devices/src/pseudo/init.c
```diff
@@ -0,0 +1,40 @@
...
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#ifdef __x86_64__
+#define MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE 0xd0000000
+#endif
+#ifdef __aarch64__
+#define MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE 0x40001000
+#endif
```
acatangiu, line 14: Add some comments here pointing to the firecracker rust files where these values are defined/computed.

src/devices/src/pseudo/persist.rs
```diff
@@ -0,0 +1,82 @@
...
+pub struct BootTimerState {
+    start_ts: String,
+}
+
+pub struct BootTimerConstructorArgs {
+    pub mem: GuestMemoryMmap,
+}
+
+impl Persist<'_> for BootTimer {
+    type State = BootTimerState;
```
acatangiu, line 28: Is it possible to define `State` as `()` and avoid defining a dummy `BootTimerState`?
```suggestion
    type State = ();
```
If that doesn't work, just use an empty `struct BootTimerState {}`

Same for `BootTimerConstructorArgs`.

src/devices/src/pseudo/persist.rs
```diff
@@ -0,0 +1,82 @@
...
+        //let guest_mem = GuestMemoryMmap::from_ranges(&[(GuestAddress(0), 0x10000)]).unwrap();
+
+        // Restore the btimer device.
+        /*  let _restored_btimer = BootTimer::restore(
+            BootTimerConstructorArgs { mem: guest_mem },
+            &BootTimerState::deserialize(&mut mem.as_slice(), &version_map, 1).unwrap(),
+        )
+        .unwrap(); */
+
+        // How do we test this?
```
acatangiu, line 80:
```suggestion
        // I think it's enough to do:
        BootTimer::restore((), &()).unwrap();
```

src/devices/src/pseudo/persist.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Versionize)]
+pub struct BootTimerState {
+    start_ts: String,
+}
+
+pub struct BootTimerConstructorArgs {
+    pub mem: GuestMemoryMmap,
+}
+
+impl Persist<'_> for BootTimer {
```
acatangiu, line 27: I think there's no point in actually implementing `Persist` for `BootTimer`. @sandreim what do you think? how can we most elegantly handle a device with no state?

I need to check device-manager `Persist` impl, but I'm guessing that's where the trick will be. For this type of device don't serialize any device specific data, and on restore path just build `new()` one.

src/vmm/src/builder.rs
```diff
@@ -840,6 +861,11 @@ pub mod tests {
         block_files
     }
 
+    pub(crate) fn insert_boot_timer_device(vmm: &mut Vmm, request_ts: TimestampUs) {
+        let res = attach_boot_timer_device(vmm, request_ts);
+        assert!(res.is_ok());
+    }
```
acatangiu, line 867: instead of using this function, why not just do `attach_boot_timer_device(vmm, request_ts).unwrap()`?

src/vmm/src/device_manager/mmio.rs
```diff
@@ -110,17 +108,32 @@ impl MMIODeviceManager {
...
-        if self.irq > self.last_irq {
-            return Err(Error::IrqsExhausted);
+    fn allocate_new_slot(&mut self, irq_count: usize) -> Result<MMIODeviceInfo> {
+        let slot: MMIODeviceInfo;
+        match irq_count {
+            0 => {
+                slot = MMIODeviceInfo {
+                    addr: self.mmio_base,
+                    len: MMIO_LEN,
+                    irqs: vec![0],
```
acatangiu, line 118: Don't trick into thinking it has `irq 0`. Just give it an empty vec:
```suggestion
                    irqs: vec![],
```

src/vmm/src/device_manager/mmio.rs
```diff
@@ -110,17 +108,32 @@ impl MMIODeviceManager {
...
+                slot = MMIODeviceInfo {
+                    addr: self.mmio_base,
+                    len: MMIO_LEN,
+                    irqs: vec![self.irq],
+                };
+                self.irq += 1;
+            }
+            _ => {
+                return Err(Error::UnsupportedIrqCount);
+            }
```
acatangiu, line 134: I'm fine with allowing any number of IRQs really:

```suggestion
        if self.irq + irq_count > self.last_irq + 1 {
            return Err(Error::IrqsExhausted);
        }
        let irqs = Vec::with_capacity(irq_count);
        for _ in 0..irq_count {
            irqs.push(self.irq);
            self.irq += 1;
        }
        slot = MMIODeviceInfo {
            addr: self.mmio_base,
            len: MMIO_LEN,
            irqs,
        };
```

src/devices/src/pseudo/init.c
```diff
@@ -0,0 +1,40 @@
...
+        PROT_WRITE,
+        MAP_SHARED,
+        fd,
+        MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE);
+
+   char *virt_addr = (char*)map_base;
+
+   *(volatile char*)virt_addr = MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE;
+
+   const char *init = "/sbin/openrc-init";
```
acatangiu, line 34: This is `Alpine` specific. Add a comment at the top of the file making it clear that the resulting binary would be `Alpine` specific.

src/vmm/src/builder.rs
tnotw: I think it was because it follows the pattern of the other devices,  and makes it visible to the device_manger `persist.rs` tests.  I did try to remove it, but it then requires the `attach_boot_timer_device` function to be made public.

src/devices/src/pseudo/init.c
tnotw: Done.

src/devices/src/pseudo/init.c
tnotw: Done.

src/devices/src/pseudo/persist.rs
tnotw: I've removed `Persist` for `BootTimer` and had a stab at recreating a new device on restore.  It required a change to the restored device manager from using `dummy_mmio_base` to the actual `MMIO_MEM_START` address,  so that the timer get recreated with the correct address.  It would also assume that the `BootTimer` is always the first device on the bus,  which it is currently for `x86_64` but isn't for `aarch64`.   Not sure if that approach works,  but obviously happy to go with whatever you advise.....

src/devices/src/pseudo/init.c
tnotw: Thanks!!

src/devices/src/pseudo/init.c
tnotw: Thanks!!  Sorry, it was a bit cut and pasty.  The link helped :-)

src/vmm/src/builder.rs
acatangiu: > but it then requires the attach_boot_timer_device function to be made public.

It should work with `pub(crate)`.

src/devices/src/pseudo/init.c
```diff
@@ -0,0 +1,46 @@
...
+#define MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE 0xd0000000
+#endif
+#ifdef __aarch64__
+#define MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE 0x40001000
+#endif
+
+#define MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE 123
+
+int main () {
+
```
acatangiu, line 24: nit:
```suggestion
```

src/devices/src/pseudo/init.c
```diff
@@ -0,0 +1,46 @@
...
+   int mapped_size = getpagesize();
+
+   char *map_base = mmap(NULL,
+        mapped_size,
+        PROT_WRITE,
+        MAP_SHARED,
+        fd,
+        MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE);
+
+   char *virt_addr = (char*)map_base;
```
acatangiu, line 35: Please remove unused `virt_addr`:
```suggestion
```

src/vmm/src/builder.rs
```diff
@@ -141,6 +141,7 @@ impl Display for StartMicrovmError {
 
                 write!(f, "Cannot open the block device backing file. {}", err_msg)
             }
+
```
acatangiu, line 144: nit: remove newline
```suggestion
```

src/vmm/src/device_manager/mmio.rs
```diff
@@ -110,17 +105,22 @@ impl MMIODeviceManager {
     }
 
     /// Allocates resources for a new device to be added.
-    fn allocate_new_slot(&mut self) -> Result<MMIODeviceInfo> {
-        if self.irq > self.last_irq {
+    fn allocate_new_slot(&mut self, irq_count: u32) -> Result<MMIODeviceInfo> {
+        let slot: MMIODeviceInfo;
```
acatangiu, line 109: nit: you can remove this line and simply add a `let` where you assign `slot`.

src/vmm/src/device_manager/persist.rs
```diff
@@ -115,57 +117,59 @@ impl<'a> Persist<'a> for MMIODeviceManager {
...
-                    states.vsock_device = Some(ConnectedVsockState {
-                        device_id: device_id.clone(),
-                        device_state: vsock_state,
-                        transport_state,
-                        mmio_slot: device_info.clone(),
-                    });
-                }
-                _ => unreachable!(),
-            };
+            if let Some(boot_timer) = bus_device.as_any().downcast_ref::<BootTimer>() {
```
acatangiu, line 120: Why change the code following this `if let` code-block?

I suggest you keep this code block that skips `BootTimer` devices, and leave the rest untouched.

src/devices/src/pseudo/init.c
```diff
@@ -0,0 +1,45 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Init wrapper for boot timing.  Alpine-specific because it points at /sbin/openrc-init
```
andreeaflorescu, line 4: Why is this in `src/devices/` if it is Alpine specific?

src/vmm/src/builder.rs
```diff
@@ -301,13 +300,18 @@ pub fn build_microvm_for_boot(
...
+    let request_ts = TimestampUs::default();
+
     let (mut vmm, mut vcpus) = create_vmm_and_vcpus(
         event_manager,
         guest_memory,
         track_dirty_pages,
         vcpu_config.vcpu_count,
     )?;
 
+    attach_boot_timer_device(&mut vmm, request_ts)?;
```
andreeaflorescu, line 313: Is this method of getting the boot time generic enough to work with any other Linux guests? If not, do we want to include this device for every build of Firecracker or should we instead gate it by an optional feature?

src/devices/src/pseudo/init.c
acatangiu: What Andreea said. We need to find a better location for this. Maybe `/tools`? @andreeaflorescu what do you think?

TLDR this file represents source of binary that writes to magic port on userspace init to trigger boot time calculation.

src/vmm/src/builder.rs
acatangiu: This is generic. The device works with all Linux guests.

The binary used in the guest is Alpine specific because our "minimal-boot" rootfs is Alpine.

src/devices/src/pseudo/init.c
lauralt: Also, where is / when was this file used?

src/vmm/src/device_manager/mmio.rs
```diff
@@ -110,17 +105,22 @@ impl MMIODeviceManager {
     }
 
     /// Allocates resources for a new device to be added.
-    fn allocate_new_slot(&mut self) -> Result<MMIODeviceInfo> {
-        if self.irq > self.last_irq {
+    fn allocate_new_slot(&mut self, irq_count: u32) -> Result<MMIODeviceInfo> {
```
andreeaflorescu, line 108: This function does not keep the same abstraction as before this PR.

Previously, it was hiding the implementation details of a slot. A slot was defined as both a irq number & an MMIO base. Not sure this is the best approach for fixing the problem at hand (i.e. having a device that doesn't need interrupts).

To me it looks like a re-design is in order for the interface to still be clean.

src/devices/src/pseudo/init.c
tnotw: > Why is this in `src/devices/` if it is Alpine specific?

Think I originally had it building via a `build.rs` script in the root of the devices module.  That bit got dropped but I left it where it is because there is obviously the connection between the signal and the boot_timer device handling it.  It made sense to me,  but happy to put it wherever is considered sensible. 

> Also, where is / when was this file used?

It lives in the guests that are used to measure boot time, and is a drop in wrapper replacement for `/sbin/init` on Alpine guests (a small tweak is required for systemd guests).  It's hopefully pretty close to what is currently on these guests.   So on a clean Alpine image `/sbin/int` is a symlink to `/sbin/openrc-init`,  whereas on the guests used to measure boot time, there is a wrapper program that sends the boot time signal before executing `/sbin/openrc-init`.

To see the current version in action,  boot into the `boottime-rootfs.ext4` rootfs,  tail the FC logs, and execute `/sbin/init` a few times.   You should hopefully see the boot time logging message being triggered on each execution.  To test this PR,  build Firecracker,  compile and copy over the "new" `/sbin/init` wrapper to the rootfs and repeat the `/sbin/init` calls.  Boot time logging should hopefully still be working.

src/vmm/src/device_manager/mmio.rs
tnotw: Possibly beyond my contributor pay grade to comment, as I was working under guidance from @acatangiu here.  

Just to check though, are you saying that you'd preferred `allocate_new_slot` to remain parameterless, and for there to be some other mechanism for determining the number of IRQ's to allocate?

andreeaflorescu: Something like that, yap.

It looks a bit like allocate_new_slot is taking care of two things:
- allocate an MMIO slot which is needed for all devices
- allocate an IRQ number

Looking at the code it is a bit hard to fix. The problem seems to be that "slot" is used to define both the irq and the mmio slot. Typically you only call a "slot" the mmio slot.
The expectation would be that the device manager separates the allocation of a slot and of a irq number. Then each device requests irqs & slots as they need them. When a device requests an irq, than irq also needs to be registered against KVM.

You can ignore my comment and I can later add a technical dept issue so we can fix this.

On the other hand, it doesn't look like we ever register more than one irq for every other device. Should we use a bool instead?

acatangiu: I see the `mmio space` (composed of `base addr` + `length`) and `used irqs` as resources managed by the `MMIODeviceManager`. For our existing devices these are _hardcoded_ as:
 - `base addr` of current device starts where the previous device `mmio space` ends
 - `mmio space len` is `MMIO_LEN`
 - `used irq(s)` hardcoded to a single IRQ number, next available in the range.

`allocate_new_slot` is the interface through which these resources are reserved for configuration and use by the device being currently set up by the caller.

The way I see it is that this interface can be extended to provide control to the caller on _how many_ of these resources are required for the new device - in this case, the boot timer device doesn't need any IRQs.

Providing more control on what types and how many resources the device needs doesn't leak implementation details for the MMIODeviceManager IMHO. From an interface standpoint, this would ideally allow control over all types and values of resources (base, len, irq count), but I fear that brings a lot of complexity without a practical usecase.

@andreeaflorescu what do you think? Does this make sense?

src/devices/src/pseudo/init.c
andreeaflorescu: I don't think I quite grasp what is the purpose of this file. From what I understand, this file needs to be baked in the rootfs so that testing the bootime works. It seems to me that that this file is used for showcasing what needs to happen on `init`. Potentially you can build your own rootfs and include this file and then check the boot time. Are my assumptions correct?

Looking at this from a different angle:
- everything that is in `src/` is essentially shipped with the Firecracker release binary
- this file doesn't seem to be compiled
- is this file needed for release? if not, shouldn't it be somewhere in resources/tests?

src/vmm/src/builder.rs
andreeaflorescu: Ok, thanks.

src/vmm/src/device_manager/mmio.rs
andreeaflorescu: Yap, let's handle this in a separate PR.

src/devices/src/pseudo/init.c
tnotw: > Are my assumptions correct?

Yeah,  pretty much.  It's not just showcasing though, because it is needed for the bootime test.   Before merging this PR someone on the project side will need to compile this file and update the bootime rootfs images with the new init wrapper,  to correspond with the changes in this PR.   The main change being the address that the signal goes to i.e. currently on x86_64:  
```
const MAGIC_IOPORT_SIGNAL_GUEST_BOOT_COMPLETE: u64 = 0x03f0;
```
Which will no longer work when this change goes in,  because Firecracker will be listening for a signal at: 
```
0xd0000000
```
Which is the first slot on the MMIO bus,  i.e. the location of the BootTimer device.

`resources/tests`  sounds good to me,  I'll move it.

lauralt: @tnotw Thanks for the explanation, my question was more of a: why is the CI passing? :)) Was the rootfs image used by the boottime integration tests already updated in S3 with your init wrapper? :-? I think I'm missing something here.

tnotw: Ah, well,  I'm not sure it is to be honest.  I cannot see any `x86_64` logs being produced in the CI runs,  so I suspect something is rather broken.  And I have a suspicion that that boot time test for `aarch64` is disabled. 

To be honest,  I was still working on it and I'm not really sure it was 100% ready for review.

lauralt: Yep, it looks like the CI is not actually running if you open the PR as draft PR, but it's green and in the meantime it was marked as ready for review which made things even more confusing :(.
Also, boot time test for `aarch64` is not disabled.

src/devices/src/pseudo/boot_timer.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn test_boot_timer() {
+        let mut boot_timer = BootTimer::new(TimestampUs::default());
+
+        let good_data = [123];
+        let bad_data = [123, 123];
+        let good_offset: u64 = 0;
+        let bad_offset: u64 = 1;
+
+        boot_timer.write(good_offset, &bad_data);
+        boot_timer.write(bad_offset, &good_data);
```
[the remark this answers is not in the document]
tnotw: Yeah,  it was mostly just sanity-checking and line coverage.   As above though,  less just seemed to be more here.

src/devices/src/pseudo/init.c
tnotw: Ok,  so I'll try and trigger another run tonight and see what happens.

My other theory about why `'aarch64` works is that i've miscalculated where the bus timer lives on the MMIO bus,  and that both old and new locations are the same.  So it works with the old `/sbin/init` which points to `0x40000000` not  `0x40001000`, which I have in the new wrapper.   I've not done erm....... quite as much testing on `aarch64` as I would have liked :-)

src/devices/src/pseudo/boot_timer.rs
```diff
@@ -0,0 +1,65 @@
...
+
+pub struct BootTimer {
+    start_ts: TimestampUs,
+}
+
+impl BusDevice for BootTimer {
+    fn write(&mut self, offset: u64, data: &[u8]) {
+        // Only handle byte length instructions
+        if data.len() != 1 {
+            return;
```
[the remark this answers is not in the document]
lauralt: The suggestion for adding a warning here was not for making testing easier, but for having some proof (log) if an invalid write happened (which if I understand correctly can happen if, in the init wrapper, you are writing more than one byte). You can leave it as is if you don't think is worth it.

src/devices/src/pseudo/boot_timer.rs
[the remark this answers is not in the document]
lauralt: Adding tests just for line coverage doesn't bring value, I prefer having a smaller coverage than tests that just call functions without checking anything. Also, testing it by looking for the warning is, like I said, not the best idea, I think we can leave that part untested as well.

src/devices/src/pseudo/boot_timer.rs
```diff
@@ -0,0 +1,65 @@
+use utils::time::TimestampUs;
```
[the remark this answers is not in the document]
tnotw: Done.

src/devices/src/pseudo/boot_timer.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use crate::bus::BusDevice;
+
+const MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE: u8 = 123;
+
+pub struct BootTimer {
```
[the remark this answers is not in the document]
tnotw: Done.

src/devices/src/pseudo/boot_timer.rs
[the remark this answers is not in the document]
tnotw: Ok,  I know what you mean - it feels like something should go here - but unless anyone else objects,  I think I'll just leave it as it is....

src/devices/src/pseudo/boot_timer.rs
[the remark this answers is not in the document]
tnotw: Yeah,  maybe a better way of putting it would be:  "I put some tests in so no one would ask where the tests are" :-)

src/vmm/src/device_manager/persist.rs
tnotw: Done.

src/devices/src/pseudo/boot_timer.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub fn new(start_ts: TimestampUs) -> BootTimer {
+        BootTimer { start_ts }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use utils::time::TimestampUs;
```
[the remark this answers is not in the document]
tnotw: Done

src/devices/src/pseudo/init.c
tnotw: Seems to be better now,  in the sense that the right tests are failing,  i.e. the x86_64 boottime tests.

src/vmm/src/device_manager/mmio.rs
```diff
@@ -94,12 +95,6 @@ pub struct MMIODeviceManager {
 impl MMIODeviceManager {
     /// Create a new DeviceManager handling mmio devices (virtio net, block).
     pub fn new(mmio_base: &mut u64, irq_interval: (u32, u32)) -> MMIODeviceManager {
-        if cfg!(target_arch = "aarch64") {
```
andreeaflorescu, line 97: Since the `mmio_base` is not changed anymore, the input parameter shouldn't be mut.

This change needs to be propagated in `builder.rs` and we can then removed the `allow(unused_mut)`: https://github.com/firecracker-microvm/firecracker/blob/e630dfc62322cf7eaf17aa3b65d6d5a82f535bc0/src/vmm/src/builder.rs#L221

src/devices/src/pseudo/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod boot_timer;
```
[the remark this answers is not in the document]
tnotw: Apologies,  added now.

src/vmm/src/device_manager/persist.rs
```diff
@@ -275,8 +287,10 @@ impl<'a> Persist<'a> for MMIODeviceManager {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use builder::attach_boot_timer_device;
     use builder::tests::*;
     use utils::tempfile::TempFile;
+    use utils::time::TimestampUs;
```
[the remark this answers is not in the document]
tnotw: Removed.

src/vmm/src/device_manager/persist.rs
```diff
@@ -174,14 +181,19 @@ impl<'a> Persist<'a> for MMIODeviceManager {
...
+        let mut mmio_base = arch::MMIO_MEM_START as u64;
+        // Only used during initial registration.
         let dummy_irq_range = (0, 0);
-        let mut dev_manager = MMIODeviceManager::new(&mut dummy_mmio_base, dummy_irq_range);
+        let mut dev_manager = MMIODeviceManager::new(&mut mmio_base, dummy_irq_range);
         let mem = &constructor_args.mem;
         let vm = constructor_args.vm;
         let event_manager = constructor_args.event_manager;
 
+        let boot_timer = devices::pseudo::BootTimer::new(TimestampUs::default());
```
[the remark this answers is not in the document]
tnotw: Done.

src/vmm/src/device_manager/mmio.rs
```diff
@@ -548,4 +556,34 @@ mod tests {
...
+            MMIODeviceManager::new(&mut 0xd000_0000, (arch::IRQ_BASE, arch::IRQ_MAX));
+        let _addr = device_manager.allocate_new_slot(0);
+        assert_eq!(device_manager.irq, arch::IRQ_BASE);
+        let _addr = device_manager.allocate_new_slot(1);
+        assert_eq!(device_manager.irq, arch::IRQ_BASE + 1);
+        assert_eq!(
+            format!(
+                "{}",
+                device_manager
+                    .allocate_new_slot(arch::IRQ_MAX + 1)
```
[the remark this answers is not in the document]
tnotw: > Here I assume you want to test the smallest number of irqs that can no longer be allocated. In this case, this should be replaced with:
> `.allocate_new_slot(arch::IRQ_MAX - arch::IRQ_BASE + 1)`

Ta,  Done. 

> 
> Moreover, I would test some more corner case scenarios, e.g.: all the irqs are taken, but you try to allocate a new slot with 0 irqs (this case should be successful), you have only one irq available but you try to reserve two etc.

Given the agreement to open a follow up PR to discuss resource allocation, I would not add additional tests here.  Since the current allocation implementation might be fated, I would be concerned that further tests are just churn.  Discussion as part of a specific PR, and then additional tests if required, sounds better to me.

src/vmm/src/device_manager/mmio.rs
```diff
@@ -548,4 +556,34 @@ mod tests {
...
+        let start_addr1 = GuestAddress(0x0);
+        let start_addr2 = GuestAddress(0x1000);
+        let guest_mem =
+            GuestMemoryMmap::from_ranges(&[(start_addr1, 0x1000), (start_addr2, 0x1000)]).unwrap();
+        let mut vm = builder::setup_kvm_vm(&guest_mem, false).unwrap();
+
+        #[cfg(target_arch = "x86_64")]
+        assert!(builder::setup_interrupt_controller(&mut vm).is_ok());
+        #[cfg(target_arch = "aarch64")]
+        assert!(builder::setup_interrupt_controller(&mut vm, 1).is_ok());
```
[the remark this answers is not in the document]
tnotw: Done.

src/vmm/src/device_manager/mmio.rs
tnotw: Done,  but this also caught an unused `mut` on the `mmio_device_manager`.  I removed it,  but maybe a sanity check would be sensible.  Also removed what I think in an unused cast here.

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
lauralt: I might be too rigid here, but I don't really agree with you on that:
- we don't know when that PR will be opened, merged etc.; it looks like a re-design is involved there and I suppose it will take time;
- you are changing the logic here for a pretty important function; since I expect we want to keep the new irqs allocation logic (i.e. let the device request as many irqs as it wants) in the other design as well, these tests won't be a waste of time at all, they will probably need some `allocate_new_slot()` -> `allocate_irqs()` change in them; moreover you will make the re-design implementer more aware of what needs to be tested.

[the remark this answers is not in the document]
tnotw: Hi @lauralt,  apologies for the delay getting back to you.  I've attempted to add the two tests you've highlighted, if you had a moment to have a look...

src/vmm/src/device_manager/mmio.rs
andreeaflorescu: Thanks!

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
lauralt: It's great, thanks! Overall, the PR looks good, I guess we should update now the boottime image with the init wrapper, and then we will let you know when it's done.

tnotw:
Shall I squash the  commits?  Just to tidy things up.....

Also, I still not 100% sure I understand why the CI/CD pipeline sometimes fails catastrophically,  as opposed to failing at the right tests.....

lauralt:
> Shall I squash the commits? Just to tidy things up.....
> 
> Also, I still not 100% sure I understand why the CI/CD pipeline sometimes fails catastrophically, as opposed to failing at the right tests.....

Squashing the commits would be great, thanks. Regarding the CI pipeline, we will take a look at it soon, sorry for that!

gc-plp:
Hi @tnotw!

I've updated the root filesystem used for measuring boot time, but you will need to update the boot time tests by specifying a custom init path:

[Here](https://github.com/firecracker-microvm/firecracker/blob/master/tests/integration_tests/performance/test_boottime.py#L76), you should add `'boot_args': 'init=/sbin/boottime_init console=ttyS0 reboot=k panic=1 pci=off'`.

Once you do that we should be able to run the CI cycle again and check if anything else needs to be done.

Thanks!

tnotw:
Hi @gc-plp, 

Thanks very much for picking this up!  I've updated the `boot_args` as requested.  Unfortunately,  I'm not seeing the update to the `boottime-rootfs.ext4` file; both old and new files look the same to me: 

```
sha256sum *
635d86bf8b7ffc6859cb75b0ba4764b4e08b664e589c49ff60c82fb568ca021a  boottime-rootfs.ext4
635d86bf8b7ffc6859cb75b0ba4764b4e08b664e589c49ff60c82fb568ca021a  boottime-rootfs_old.ext4
```
Am I looking in the right place?  Or was there a wee s3 push/sync issue? 

I guess that the `aarch64` boottime image and both `initrd.img` files will also need updating if we're taking this approach of specifying a new init file location.

gc-plp:
@tnotw the rootfs files should be different:

```
sha256sum build/img/x86_64/minimal/fsfiles/boottime-rootfs*
639e376bcb6464b2e794cda531ad173a4b0428ffcf4d4afad9357980a7ae36d5  build/img/x86_64/minimal/fsfiles/boottime-rootfs.ext4
635d86bf8b7ffc6859cb75b0ba4764b4e08b664e589c49ff60c82fb568ca021a  build/img/x86_64/minimal/fsfiles/boottime-rootfs_old.ext4

```

`boottime-rootfs_old.ext4` is the original rootfs, while `boottime-rootfs.ext4` is the updated one.

Could you delete the `build/` folder and run the build again? 
This folder is used for file caching. A file is downloaded only if the test system realizes that it's missing, so if it's updated upstream, it won't be downloaded again.

Indeed, the `initrd` will probably also have to be updated. I'll do that once the non-initrd tests pass and retrigger the build from our end.

tnotw:
> Could you delete the `build/` folder and run the build again?

Thanks @gc-plp,  apologies,  I hadn't realised the images were cached.

tnotw:
Hi @gc-plp, 

Booting seems to go a whole lot quicker without `console=ttyS0`.  Are these parameters definitely pathed in somewhere at the moment?
```
console=ttyS0 reboot=k panic=1 pci=off
```
We've not added them in unintentionally?

gc-plp:
@tnotw good catch on the `console=ttyS0` parameter. That was the problem :)

I'll update the initrd and `aarch64` images.

resources/tests/init.c
```diff
@@ -0,0 +1,44 @@
...
+#define MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE 0xd0000000
+#endif
+#ifdef __aarch64__
+#define MAGIC_MMIO_SIGNAL_GUEST_BOOT_COMPLETE 0x40000000
+#endif
+
+#define MAGIC_VALUE_SIGNAL_GUEST_BOOT_COMPLETE 123
+
+int main () {
+   int fd = open("/dev/mem", (O_RDWR | O_SYNC | O_CLOEXEC));
```
gc-plp, line 25: If we are going to integrate this into the `initrd.img` file, we need to make sure that `/dev/mem` exists when it's executed.

If it doesn't exist, we need to create it: `mknod("/dev/mem", S_IFCHR, makedev(1, 1));`

I updated the `initrd.img` with a modified version of this executable that calls `mknod` first, but the boot time seems to be quite long - about 1.2s. I haven't finished investigating what's the cause for the long boot time, but if nothing comes up, I can continue on Tuesday.

tnotw: Ok,  so yeah,  this looks a whole load more complicated than I thought :-(

This [post](https://unix.stackexchange.com/questions/52713/how-to-populate-dev-directory-when-building-my-own-initrd) sort of suggests that an `/init` wrapper containing `mount -t devtmpfs none /dev` before calling `/sbin/init` (or `/sbin/boottime_init` in our case) might be an option.   Don't know that would be quicker than making a static device.

This might would work if we just had a single boot time mechanism,  but I'm struggling to see how it would work for trying to maintain two.  AFACIT,  the `init` parameter gets ignored when loading an `initrd` image,  and `/init` is called no matter what.  Unless I'm doing something wrong.

acatangiu: Why not stick with the simple solution we had before?

Directly bake and use the `init` wrapper in the image. The wrapper writes to the magic addr then runs the real `init`.

Simplifying assumptions:
1. Since we need the wrapper tool baked in the image, just place it as the default `init`, avoid using the `init=` boot param. Even when boot time is not measured, it only cost a few extra cycles at userspace initialization.
2. Since we control the image, use one that has `/dev/mem` set up right at boot. Don't add the extra complexity of handling mounting `/dev/mem`. We do need to continue gracefully if `/dev/mem` is not present (just run _real_ `init`).

@gc-plp @tnotw what do you think?

tnotw: Simple sounds good to me :-)

gc-plp: @acatangiu, in the current initrd the first assumption is already in place, so even if we don't specify a custom `init=` the magic number will still be written.

We can also use a custom kernel to simplify things. I tested the idea by adding:
```diff
diff --git a/init/main.c b/init/main.c
index 0ee9c6866ada..31a4207dd20c 100644
--- a/init/main.c
+++ b/init/main.c
@@ -1001,6 +1001,11 @@ static int __ref kernel_init(void *unused)
 
        rcu_end_inkernel_boot();
 
+       void *base;
+       base = ioremap(0xd0000000, 0x4);
+       iowrite8(123, base);
+       iounmap(base);
+
        if (ramdisk_execute_command) {
                ret = run_init_process(ramdisk_execute_command);
                if (!ret)
```

in the 4.14 kernel just before calling `/init` and it works, with the following results:
- `Boot time with no network is: 86636 us`
- `Boot time with network configured is: 86456 us`
- `Boot time with initrd is: 146258 us`

The advantage of this method is that we don't need to patch multiple images to measure boot time (e.g. `number of architectures`x`number of images) and it's distro agnostic.
What do you think?

acatangiu: Yes, that works too. But it just moves the burden of managing rootfs images to managing kernel images.
It can also be considered _cheating_ since that's still _kernel boot code_ 😝 .

**Pro**: we generally have more rootfs images than kernel images so fewer objects to maintain
**Con**: it is much easier to update rootfs images (simple file operations) than to patch and build kernel images

I personally lean toward doing it in userspace.

gc-plp: @tnotw After a bit of struggling with packing the initrd image, it seems to work now. Apparently I was gzipping the initrd image, which introduced extra boot time.

Could you revert the `boot_args` parameter addition? Sorry for the back and forth.

tnotw: Sure, no problem.  Every day is a school day, as they say :smile:

gc-plp:
The PR seems to work with the updated images, but if we keep the updated images in the S3 bucket, other PRs will fail.

I switched the images back to the old ones.

Let me know when this is ready to be merged so that I can switch back to the new images and retrigger the tests. Once the tests pass, we can merge this immediately.

acatangiu:
> There are a few paths that we can follow:
    1. This PR should use different fixture names, so that the tests pull images from a different folder. The disadvantage of this method is that the size of the S3 bucket will increase because we need to keep old images there.
    2. The init should trigger both the new boot time method and the old one. The disadvantage is that we remove code that handles the old boot time trigger method.
    3. We go back to using a separate init and check if we can achieve an under 160ms boot time with a custom init.

I like option 1. It keeps things clean, each version with its own independent mechanism. S3 bucket increase is irrelevant IMHO.

raduiliescu:
> 
> 
> > There are a few paths that we can follow:
> > 1. This PR should use different fixture names, so that the tests pull images from a different folder. The disadvantage of this method is that the size of the S3 bucket will increase because we need to keep old images there.
> > 2. The init should trigger both the new boot time method and the old one. The disadvantage is that we remove code that handles the old boot time trigger method.
> > 3. We go back to using a separate init and check if we can achieve an under 160ms boot time with a custom init.
> 
> I like option 1. It keeps things clean, each version with its own independent mechanism. S3 bucket increase is irrelevant IMHO.

I'd also go with option 1. On long term would be nice to version our images and have ways to replicate them.

gc-plp:
By popular demand, option (1) has been implemented for x86_64 :+1: 

@tnotw, we are once again asking for your support in changing the boottime test. You basically need to update the fixtures that are passed as test parameters:

```diff
diff --git a/tests/integration_tests/performance/test_boottime.py b/tests/integration_tests/performance/test_boottime.py
index 9a90e25b..30abf388 100644
--- a/tests/integration_tests/performance/test_boottime.py
+++ b/tests/integration_tests/performance/test_boottime.py
@@ -22,34 +22,34 @@ INITRD_BOOT_TIME_US = 160000 if platform.machine() == "x86_64" else 500000
 TIMESTAMP_LOG_REGEX = r'Guest-boot-time\s+\=\s+(\d+)\s+us'


-def test_boottime_no_network(test_microvm_with_boottime):
+def test_boottime_no_network(test_microvm_with_boottime_timer):
     """Check guest boottime of microvm without network."""
-    _ = _configure_vm(test_microvm_with_boottime)
+    _ = _configure_vm(test_microvm_with_boottime_timer)
     time.sleep(0.4)
-    boottime_us = _test_microvm_boottime(test_microvm_with_boottime.log_data)
+    boottime_us = _test_microvm_boottime(test_microvm_with_boottime_timer.log_data)
     print("Boot time with no network is: " + str(boottime_us) + " us")


 def test_boottime_with_network(
-        test_microvm_with_boottime,
+        test_microvm_with_boottime_timer,
         network_config
 ):
     """Check guest boottime of microvm with network."""
-    _tap = _configure_vm(test_microvm_with_boottime, {
+    _tap = _configure_vm(test_microvm_with_boottime_timer, {
         "config": network_config, "iface_id": "1"
     })
     time.sleep(0.4)
-    boottime_us = _test_microvm_boottime(test_microvm_with_boottime.log_data)
+    boottime_us = _test_microvm_boottime(test_microvm_with_boottime_timer.log_data)
     print("Boot time with network configured is: " + str(boottime_us) + " us")


 def test_initrd_boottime(
-        test_microvm_with_initrd):
+        test_microvm_with_initrd_timer):
     """Check guest boottime of microvm with initrd."""
-    _tap = _configure_vm(test_microvm_with_initrd, initrd=True)
+    _tap = _configure_vm(test_microvm_with_initrd_timer, initrd=True)
     time.sleep(0.8)
     boottime_us = _test_microvm_boottime(
-        test_microvm_with_initrd.log_data, max_time_us=INITRD_BOOT_TIME_US)
+        test_microvm_with_initrd_timer.log_data, max_time_us=INITRD_BOOT_TIME_US)
     print("Boot time with initrd is: " + str(boottime_us) + " us")
```

tnotw:
@gc-plp,  think that's done.  I had to tweak the pylint config to take the expanded argument name, but I think that's ok.  The x86_64 CI run failed for some unknown reason....

gc-plp:
@tnotw everything is passed now. 
Thank you for this awesome contribution!

tnotw:
Phew!!  Thanks very much @iulianbarbu, @acatangiu, @andreeaflorescu, @lauralt, @gc-plp,  for all your help, reviews and feedback.  Very much appreciated.

Issue #1875, opened 2020-05-11
serial console - host hangup results in noisy firecracker log


iulianbarbu:
When opening a Firecracker microVM with serial console and the some point closing the received console, the serial console receives "IN | ERR | HANGUP" events, in level trigger manner, resulting in noisy firecracker log.

Issue #1627, opened 2020-02-24
Handle epoll errors for when handling device events 


iulianbarbu:
At the moment, the current device event handling implementation, for net, block and vsock devices, does not handle errors which might occur on the epoll registered event `fd`s (`EPOLLHUP`, `EPOLLRDHUP`, `EPOLERR`). We must `unregister` these `fd`s for which we receieved error events.

acatangiu:
I don't think we should unregister in case of any of those events. But all devices do need to gracefully handle them.

serban300:
In the meanwhile we switched to using the EventHandler from rust-vmm. We should check if it handles these events or if we should move this issue there.

PR #1931, opened 2020-05-27
serial: handle HANG_UP

iulianbarbu:
## Reason for This PR

Firecracker serial device is emulated around the standard input of a Firecracker process. When the Firecracker standard input is set up as an end of a pipe and the other end of the pipe is closed, then this event will level-trigger an `EPOLLHUP` event on the Firecracker standard input, which is monitored for driving the serial device emulation.

`EPOLLHUP` is currently an unexpected event and when it arrives, it will issue a warning. This warning is logged continuously, raising the Firecracker emulation thread to 100% utilization, starving the emulation and the guest, if quotas are imposed on the Firecracker process as a whole.

Fixes: #1875 .

## Description of Changes

Added handling of `EPOLLHUP` event received on the serial device input. When it is received, consume any bytes which might be inflight, then unregister the serial input from monitoring.

## License Acceptance

## PR Checklist

- ~~[ ] Any API changes are reflected in `firecracker/swagger.yaml`.~~
- ~~[ ] Any user-facing changes are mentioned in `CHANGELOG.md`.~~

### DIFF
--- a/src/devices/src/legacy/serial.rs
+++ b/src/devices/src/legacy/serial.rs
@@ -157,25 +157,21 @@ impl Serial {
         }
     }
 
-    fn thr_empty(&mut self) -> io::Result<()> {
+    fn thr_empty_interrupt(&mut self) -> io::Result<()> {
         if self.is_thr_intr_enabled() {
             self.add_intr_bit(IIR_THR_BIT);
-            self.trigger_interrupt()?
+            self.interrupt_evt.write(1)?
         }
         Ok(())
     }
 
-    fn recv_data(&mut self) -> io::Result<()> {
+    fn recv_data_interrupt(&mut self) -> io::Result<()> {
         if self.is_recv_intr_enabled() {
             self.add_intr_bit(IIR_RECV_BIT);
-            self.trigger_interrupt()?
+            self.interrupt_evt.write(1)?
         }
         self.line_status |= LSR_DATA_BIT;
         Ok(())
-    }
-
-    fn trigger_interrupt(&mut self) -> io::Result<()> {
-        self.interrupt_evt.write(1)
     }
 
     fn iir_reset(&mut self) {
@@ -195,7 +191,7 @@ impl Serial {
                 if self.is_loop() {
                     if self.in_buffer.len() < LOOP_SIZE {
                         self.in_buffer.push_back(value);
-                        self.recv_data()?;
+                        self.recv_data_interrupt()?;
                     }
                 } else {
                     if let Some(out) = self.out.as_mut() {
@@ -204,7 +200,7 @@ impl Serial {
                         out.flush()?;
                         METRICS.uart.flush_count.inc();
                     }
-                    self.thr_empty()?;
+                    self.thr_empty_interrupt()?;
                 }
             }
             IER => self.interrupt_enable = value & IER_FIFO_BITS,
@@ -244,10 +240,26 @@ impl Serial {
         }
     }
 
+    fn recv_bytes(&mut self) -> io::Result<usize> {
+        if let Some(input) = self.input.as_mut() {
+            let mut out = [0u8; 32];
+            return input.read(&mut out).and_then(|count| {
+                if count > 0 {
+                    self.raw_input(&out[..count])?;
+                    Ok(count)
+                } else {
+                    Ok(0)
+                }
+            });
+        }
+
+        Ok(0)
+    }
+
     fn raw_input(&mut self, data: &[u8]) -> io::Result<()> {
         if !self.is_loop() {
             self.in_buffer.extend(data);
-            self.recv_data()?;
+            self.recv_data_interrupt()?;
         }
         Ok(())
     }
@@ -276,35 +288,47 @@ impl BusDevice for Serial {
 }
 
 impl Subscriber for Serial {
-    /// Handle a read event (EPOLLIN) on the serial input fd.
-    fn process(&mut self, event: &EpollEvent, _: &mut EventManager) {
+    /// Handle events on the serial input fd.
+    fn process(&mut self, event: &EpollEvent, ev_mgr: &mut EventManager) {
         let source = event.fd();
-        let event_set = event.event_set();
 
-        // TODO: also check for errors. Pending high level discussions on how we want
-        // to handle errors in devices.
-        let supported_events = EventSet::IN;
-        if !supported_events.contains(event_set) {
-            warn!(
-                "Received unknown event: {:?} from source: {:?}",
-                event_set, source
-            );
+        // We expect to be interested only in serial input.
+        let interest_list = self.interest_list();
+        if interest_list.len() != 1 {
+            warn!("Unexpected events/sources interest list.");
             return;
         }
 
-        if let Some(input) = self.input.as_mut() {
-            if input.as_raw_fd() == source {
-                let mut out = [0u8; 32];
-                match input.read(&mut out[..]) {
-                    Ok(count) => {
-                        self.raw_input(&out[..count])
-                            .unwrap_or_else(|e| warn!("Serial error on input: {}", e));
-                    }
-                    Err(e) => {
-                        warn!("error while reading stdin: {:?}", e);
+        // Safe to unwrap. Checked before if interest list has one element.
+        let supported_event = interest_list.first().unwrap();
+
+        // Check if the event source is the serial input.
+        if supported_event.fd() != source {
+            warn!("Unexpected event source: {}", source);
+            return;
+        }
+
+        // We expect to receive: `EventSet::IN`, `EventSet::HANG_UP` or
+        // `EventSet::ERROR`. To process all these events we just have to
+        // read from the serial input.
+        match self.recv_bytes() {
+            Ok(count) => {
+                // Check if the serial input have to be unregistered.
+                let event_set = event.event_set();
+                let unregister_condition =
+                    event_set.contains(EventSet::ERROR) | event_set.contains(EventSet::HANG_UP);
+                if count == 0 && unregister_condition {
+                    // Unregister the serial input source.
+                    match ev_mgr.unregister(supported_event.fd()) {
+                        Ok(_) => warn!("Detached the serial input due to peer error/close."),
+                        Err(e) => error!(
+                            "Peer is unreachable. Could not detach the serial input: {:?}",
+                            e
+                        ),
                     }
                 }
             }
+            Err(e) => error!("error while reading stdin: {:?}", e),
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/devices/src/legacy/serial.rs
```diff
@@ -276,34 +288,72 @@ impl BusDevice for Serial {
...
-    fn process(&mut self, event: &EpollEvent, _: &mut EventManager) {
+    /// Handle events on the serial input fd.
+    fn process(&mut self, event: &EpollEvent, ev_mgr: &mut EventManager) {
         let source = event.fd();
         let event_set = event.event_set();
 
-        // TODO: also check for errors. Pending high level discussions on how we want
-        // to handle errors in devices.
-        let supported_events = EventSet::IN;
-        if !supported_events.contains(event_set) {
```
acatangiu, line 287: The problem seems to be limited to this line.

The `epoll_ctl` man page says:
```
EPOLLHUP
              Hang up happened on the associated file descriptor.  epoll_wait(2) will always wait for this event; it is not necessary to set it in
              events.
```

Meaning we should always be prepared to handle `EventSet::HANG_UP`.

I believe just removing this error shortcircuit here by adding `EventSet::HANG_UP` to `supported_events` will fix the issue.

**FYI: This problem exists for all our devices, so it should be fixed everywhere**, also documentation for `EventManager` should be updated to clearly specify that `EventSet::HANG_UP` can be triggered even if it wasn't declared in `interest_list()`.

src/devices/src/legacy/serial.rs
```diff
@@ -244,10 +240,26 @@ impl Serial {
...
+        if let Some(input) = self.input.as_mut() {
+            let mut out = [0u8; 32];
+            return match input.read(&mut out) {
+                Ok(count) if count > 0 => {
+                    self.raw_input(&out[..count])?;
+                    Ok(count)
+                }
+                Err(e) => Err(e),
+                _ => Ok(0),
+            };
```
aghecenco, line 253: nit that gets rid of the strange `Err(e) => Err(e)`
```suggestion
            input.read(&mut out).and_then(|count| {
                if count > 0 {
                    self.raw_input(&out[..count])?;
                    Ok(count)
                } else {
                    Ok(0)
                }
            })?;
```

src/devices/src/legacy/serial.rs
```diff
@@ -276,34 +288,72 @@ impl BusDevice for Serial {
...
+        if supported_event.fd() != source {
+            return;
+        }
+
+        // Handle `EventSet::IN` to dispose inflight bytes.
+        if event_set.contains(EventSet::IN) {
+            match self.recv_bytes() {
+                Ok(_) => {}
+                Err(e) => warn!("error while reading stdin: {:?}", e),
+            }
```
aghecenco, line 323:
```suggestion
            if let Err(e) = self.recv_bytes() {
                warn!("error while reading stdin: {:?}", e);
            }
```

src/devices/src/legacy/serial.rs
```diff
@@ -312,7 +362,10 @@ impl Subscriber for Serial {
     /// If serial input is present, register the serial input FD as readable.
```
aghecenco, line 362: Mention `HANG_UP` too.

src/devices/src/legacy/serial.rs
andreeaflorescu: Do we already have an issue for the other devices as well?

Similarly, for edge triggered events we need to handle EPOLLRDHUP.

iulianbarbu: > I believe just removing this error shortcircuit here by adding EventSet::HANG_UP to supported_events will fix the issue.

I've tested this shortcircuit removal and to me it seems that it isn't enough. `EventSet::HANG_UP` it is level-triggered and keeps coming again and again, even if we handle all the inflight bytes.

This keeps the emulation thread at 100% (due to the continuously event manager dispatching of the event to the serial). The solution from this PR for this is to unregister the stdin fd and basically dropping later on any serial input altogether.

As far as my testing goes, the way one process can use *pipes* is by using *anonymous pipes* or *named pipes*.

If one process uses anonymous pipes (through `int pipe(int pipefd[2])` or similar `pipe2` call) and write end of the pipe is closed, then the read end will continuously receive `HANG_UP`, because it can not communicate anymore with the closed end. This event can not be fixed and the communication is broken/should terminate.

However, if a process uses named pipes (through `int mkfifo(const char *pathname, mode_t mode)` it behaves different. 

Let's open a read end and a write end on top of this named pipe:
```
let rc = unsafe { libc::mkfifo(path.as_ptr(), libc::O_NONBLOCK as u32) };
assert!(rc == 0);

let mut fds: [c_int; 2] = [0; 2];

// Open as an non-blocking read end pipe, to be able to work on a single thread.
fds[0] = unsafe { libc::open(path.as_ptr(), libc::O_RDONLY | libc::O_NONBLOCK) };
assert!(fds[0] >= 0);

fds[1] = unsafe { libc::open(path.as_ptr(), libc::O_WRONLY) };
assert!(fds[1] >= 0);
```

Now, let's close the write end. This will start flooding the read end with `HANG_UP`s. However, we can stop these events by re-opening a file on top of the named pipe with `O_WRONLY` flag, which basically fixes the write end of the pipe. We can send now bytes on this pipe as if nothing happened. However, a simple bash example somehow contradicts this practice:
```
# Terminal 1 - read end.
mkfifo fifo
# Blocking fifo.
cat < fifo
# Terminal 2 - write end.
cat > fifo &
export write_pid=$!
sudo gdb -p  $write_pid
> print close(1) # Close process standard input (pipe write end).
```

This makes the process reading from the pipe to end. This is the default behavior for `cat` at least, since `HANG_UP` is seen as an end-of-file.

My thoughts about this is that it is mandatory to not keep the emulation thread at 100% utilization because of flooding it with `HANG_UP`s. At the same time, it would be great if we could support scenarios of named pipes, like the one described above, even if they seem a hack/workaround for something I can't think of right now. At the moment, I can not see a solution without compromises and if we have to choose, I think the 100% issue is prioritary.

Do you guys see any scenario where unregistering the stdin fd from the event manager (and stopping the HANG_UP) might impair any use case which tries to re establish the pipe "connection" after closing the write end? Should we really take it into consideration for a real world scenario?

iulianbarbu: > Do we already have an issue for the other devices as well?

We have an issue for this: #1627.

> Similarly, for edge triggered events we need to handle EPOLLRDHUP.

It is already mentioned in the issue above.

andreeaflorescu: > It is already mentioned in the issue above.

Cool, thanks!

acatangiu: > My thoughts about this is that it is mandatory to not keep the emulation thread at 100% utilization because of flooding it with HANG_UPs. At the same time, it would be great if we could support scenarios of named pipes.

We shouldn't worry about the named pipe scenario. We can work with the limitation that once the STDIN pipe is closed, it is irrecoverable.

With this info brought to light by @iulianbarbu's investigation, I am fine with unregistering the FDs on `HANG_UP`.

acatangiu: But in this case, I am thinking whether it'd be better to do that in `EventManager` instead of repeating same code for all `Subscriber`s.
@andreeaflorescu what do you think?

iulianbarbu: > But in this case, I am thinking whether it'd be better to do that in `EventManager` instead of repeating same code for all `Subscriber`s.

I think there might be cases that depend on `HANG_UP` handling outside the event manager loop, inside the application thread. Wouldn't this drop the support for such cases?

I do not have a real example, at the moment, but because pipes can be used as IPC, it might replace some behavior obtained through unix signaling. In case of `SIGHUP`, some daemons might [reload their configuration](https://stackoverflow.com/questions/19052354/sighup-for-reloading-configuration).

This implies custom handling that needs to be done and if the event manager assumes a default behavior for `HANG_UP`, this custom handling wouldn't be possible, right?

andreeaflorescu: I wouldn't add this in the EventManager. I find it pretty hard to assume a correct, desired, default behavior for `HANG_UP`. We could make sure to not duplicate the code everywhere by providing some Firecracker specific handlers, that can be used in all devices implementation.

For example, we can define an `EventHandler<T> where T: EventSubscriber` and have methods on it for `handle_error()`, `handle_rhup()` and so on. This way we can reduce the code duplication. I didn't put much thought into this, I am sure we can find some other better alternatives.

src/devices/src/legacy/serial.rs
```diff
@@ -276,34 +288,72 @@ impl BusDevice for Serial {
...
+                            break;
+                        }
                     }
                     Err(e) => {
                         warn!("error while reading stdin: {:?}", e);
+                        break;
                     }
-                }
+                };
+            }
```
acatangiu, line 344: Do we need to consume pending bytes? I don't see an obvious reason to do so if we unregister the fd anyway in the next step.
(less code is better)

src/devices/src/legacy/serial.rs
```diff
@@ -276,34 +288,72 @@ impl BusDevice for Serial {
...
+
+            // Unregister serial input, due to HANG_UP.
+            match ev_mgr.unregister(source) {
+                Ok(()) => warn!(
+                    "Received {:?} from source: {:?}. Detached the serial console.",
+                    event_set, source
+                ),
+                Err(e) => warn!(
```
acatangiu, line 352:
```suggestion
                Err(e) => error!(
```

src/devices/src/legacy/serial.rs
```diff
@@ -244,10 +240,26 @@ impl Serial {
         }
     }
 
+    fn recv_bytes(&mut self) -> io::Result<usize> {
+        if let Some(input) = self.input.as_mut() {
+            let mut out = [0u8; 32];
+            return match input.read(&mut out) {
```
acatangiu, line 246: This could be run in a loop to minimize epoll events when there is a lot of data pending on the pipe

acatangiu: > However, I can not tell if we have any data about the need to optimize it and I am not sure if that would help significantly for any serial console use case, at the moment.

It's just a simple optimizing suggestion from looking at the code.

> Furthermore, if you think the serial input needs optimizations on the processing path, I think we have to address this in a separate issue/PR.

Sure, if you feel this is complicated, we can do it in a separate PR.

src/devices/tests/integration_tests.rs
```diff
@@ -0,0 +1,93 @@
...
+    )));
+
+    // Make reading fd non blocking to read just what is inflight.
+    let flags = unsafe { libc::fcntl(fds[0], libc::F_GETFL, 0) };
+    let mut rc = unsafe { libc::fcntl(fds[0], libc::F_SETFL, flags | libc::O_NONBLOCK) };
+    assert!(rc == 0);
+
+    // Write some dummy data on the writing end of the pipe to handle it later on.
+    // 33 bytes are read in two rounds of serial input processing because
+    // it  is handled in batches of 32 bytes at maximum.
```
acatangiu, line 38: nit
```suggestion
    // it is handled in batches of 32 bytes at maximum.
```

iulianbarbu:
@aghecenco , please take a look when you can. I've addressed your feedback.

iulianbarbu:
@aghecenco , @acatangiu , I've forced push again to fix the arm CI failure (got a failure due to an unstable part of the integration tests infrastructure #1941).

PR #1936, opened 2020-06-01
tests: add some assertions to metrics case

xiekeyang:
Improve metrics test case, to read metrics back and assert 2 items
values inside result.

Values in these 2 items should be always equal to expected under normal case.

Signed-off-by: xiekeyang <keyang.xie@gmail.com>

## Reason for This PR

## Description of Changes

## License Acceptance

## PR Checklist

~~- [ ] Any required documentation changes (code and docs) are included in this PR.~~
~~- [ ] Any newly added `unsafe` code is properly documented.~~

### DIFF
--- a/tests/integration_tests/functional/test_metrics.py
+++ b/tests/integration_tests/functional/test_metrics.py
@@ -3,6 +3,7 @@
 """Tests the metrics system."""
 
 import os
+import json
 import host_tools.logging as log_tools
 
 
@@ -22,5 +23,29 @@ def test_flush_metrics(test_microvm_with_api):
     assert microvm.api_session.is_status_no_content(response.status_code)
 
     microvm.start()
+
+    res = metrics_fifo.sequential_reader(1)
+    metrics = json.loads(res[0])
+
+    exp_keys = [
+        'utc_timestamp_ms',
+        'api_server',
+        'block',
+        'get_api_requests',
+        'i8042',
+        'logger',
+        'mmds',
+        'net',
+        'patch_api_requests',
+        'put_api_requests',
+        'rtc',
+        'seccomp',
+        'vcpu',
+        'vmm',
+        'uart',
+        'signals',
+    ]
+
+    assert set(metrics.keys()) == set(exp_keys)
 
     microvm.flush_metrics(metrics_fifo)


tests/integration_tests/functional/test_metrics.py
```diff
@@ -23,4 +26,41 @@ def test_flush_metrics(test_microvm_with_api):
 
     microvm.start()
 
+    cmd = 'read line < {} && echo $line'.format(metrics_fifo_path)
+    res = utils.run_cmd(cmd)
```
serban300, line 30: Can we use `logging.py::Fifo::sequential_reader()` here ? Or `metrics_fifo_path.readlines()` ? Or `metrics_fifo_path.read()` ?

tests/integration_tests/functional/test_metrics.py
```diff
@@ -23,4 +26,41 @@ def test_flush_metrics(test_microvm_with_api):
...
+      "flush_count": 0,
+      "queue_event_count": 0,
+      "rate_limiter_event_count": 0,
+      "update_count": 0,
+      "update_fails": 0,
+      "read_bytes": 0,
+      "write_bytes": 0,
+      "read_count": 0,
+      "write_count": 0
+    }
```
serban300, line 64: Can we check just a couple of individual metrics instead of groups ? For example:

`metrics['put_api_requests']['actions_count'] == 1`
`metrics['put_api_requests']['metrics_count'] == 1`

dianpopa: I think the grouping approach is more useful since it will force code that adds new metrics to also update this test. My proposal would be to create a separate json in our CI folder that holds the expected json output (just like the one located in tests/framework/config_vm.json), load it in the test and check it against the one obtain  by the test. WDYT?

tests/integration_tests/functional/test_metrics.py
xiekeyang: Done.

tests/integration_tests/functional/test_metrics.py
xiekeyang: I agree with @dianpopa 's comment, which indicates my idea for this pr. I actually want to include more groups to deeply check if the metrics values is as excepted under integration test cases. But so far I only put 2 important and expectable group to this pr. Maybe we can cover almost all groups  in future.

>> My proposal would be to create a separate json in our CI folder that holds the expected json output (just like the one located in tests/framework/config_vm.json), load it in the test and check it against the one obtain by the test. WDYT?

I had considered this, but 1. some groups value may be un-exceptable, so the json will seem to be incomplete;  2. we maybe add more process to metrics test, and have to compare with different groups or values, so currently pasting the expected values to code maybe looks  more clear. WDYT?

serban300: My concern is that this test would end up being very complex if we want to check all the metrics values here. Also it would be responsible to check some things that are not really related to the metrics system. For example I think that checking that the block metrics values are correct is the responsibility of the block related tests, rather then the responsibility of this test. An option would be to disregard the values here, and just check that all the metrics are printed. And maybe we could also check that not all are equal to 0. What do you think ?

xiekeyang: @serban300 , this sounds great! We can just check the metrics output keys, with disregarding their values.

xiekeyang: @serban300 @dianpopa Updated, PTAL.

xiekeyang: I change to check only metrics root keys are as expected. The nested keys (like inside `block`, `put_api_requests `...) should be put on the relative module test.

serban300: Looks good to me. Thanks for doing the changes !

PR #1862, opened 2020-05-04
Improve command-line arguments

edigaryev:
## Reason for This PR

#1688

## Description of Changes

#1688, with the following additions:

* documented `--start-time-us` and `--start-time-cpu-us`, fixed `start-time-cpu-us` typo
* one unit test for jailed (#363) to counter imprecise kcov results preventing coverage test to pass
  * the most relevant issue for this behavior I've found is https://github.com/rust-lang/rust/issues/55352

...and with the exception for:

>Segregate mandatory and optional arguments properly.

See https://github.com/firecracker-microvm/firecracker/issues/1688#issuecomment-623550854.

- [ ] This functionality can be added in [`rust-vmm`](https://github.com/rust-vmm).

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [ ] All commits in this PR are signed (`git commit -s`).
- [ ] The reason for this PR is clearly provided (issue no. or explanation).
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this PR.
- [ ] Any newly added `unsafe` code is properly documented.
- [ ] Any API changes are reflected in `firecracker/swagger.yaml`.
- [ ] Any user-facing changes are mentioned in `CHANGELOG.md`.

Issue #363, opened 2018-06-20
Unit test the jailer 


dianpopa:
Kcov reports 0% coverage for all the files that are jailer related.

dianpopa:
Coverage reported for jailer crate right now: 42.6%. Not closing this issue yet since there is room for improvement, see above.
[cov.zip](https://github.com/firecracker-microvm/firecracker/files/4319390/cov.zip)

pkj415:
Interested in working on this to ramp up on Firecracker! (Just some context - I am a graduate student at UT trying to know more about Firecracker and will be working with Xiang Gu).

Does commenting here implicitly mean this issue gets assigned to me?

dianpopa:
Hi @pkj415 , 

Are you still interested in working on this issue? I will assign it to you after confirmation.

pkj415:
Hi @dianpopa, yes you can assign this to me. Sorry for the delay in reply, I missed this notification somehow.

dianpopa:
All the unit testable jailer code is covered.

PR #1940, opened 2020-06-02
jailer: increase code coverage

lauralt:
## Reason for This PR

A step forward for #363.

## Description of Changes

With the recently merged #1862, realized the coverage for jailer can be quite easily increased.
Split `run()` in two in order to be able to unit-test some part of it.
Added a few unit tests and made a small cleanup in jailer env tests.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -1,7 +1,7 @@
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::ffi::CStr;
+use std::ffi::{CStr, OsString};
 use std::fs::{self, canonicalize, File, Permissions};
 use std::os::unix::fs::PermissionsExt;
 use std::os::unix::io::IntoRawFd;
@@ -210,14 +210,11 @@ impl Env {
             .map_err(|e| Error::ChangeFileOwner(path_buf, e))
     }
 
-    pub fn run(mut self) -> Result<()> {
+    fn copy_exec_to_chroot(&mut self) -> Result<OsString> {
         let exec_file_name = self
             .exec_file_path
             .file_name()
             .ok_or_else(|| Error::FileName(self.exec_file_path.clone()))?;
-
-        let chroot_exec_file = PathBuf::from("/").join(exec_file_name);
-
         // We do a quick push here to get the global path of the executable inside the chroot,
         // without having to create a new PathBuf. We'll then do a pop to revert to the actual
         // chroot_dir right after the copy.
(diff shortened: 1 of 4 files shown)


src/jailer/src/env.rs
```diff
@@ -210,14 +210,11 @@ impl Env {
             .map_err(|e| Error::ChangeFileOwner(path_buf, e))
     }
 
-    pub fn run(mut self) -> Result<()> {
+    fn prepare_for_chroot(&mut self) -> Result<()> {
```
ioanachirca, line 213: The name is a bit misleading, this method doesn't do all the necessary preparing for chroot (cgroups, open /dev/null). Could the 2 things that this method deals with be split in different methods and tested separately?

src/jailer/src/env.rs
```diff
@@ -609,6 +609,80 @@ mod tests {
...
+            // Trying to create again the same device node is not allowed.
+            assert_eq!(
+                format!("{}", env.mknod_and_own_dev(dev, major, minor).unwrap_err()),
+                format!(
+                    "Failed to create {}\u{0} via mknod inside the jail: File exists (os error 17)",
+                    dev_str
+                )
+            );
+            // Remove the device node.
+            fs::remove_file(dev_str).expect("Could not remove file.");
```
dianpopa, line 622: Could we use`TempFile::with_prefix` instead of `/dev/net/tun-test` etc.? This way we make sure that the device files are deleted even if this test function quits before reaching `remove_file`.

src/jailer/src/env.rs
```diff
@@ -253,6 +250,17 @@ impl Env {
...
                 .map_err(Error::CloseNetNsFd)?;
         }
+        Ok(())
+    }
+
+    pub fn run(mut self) -> Result<()> {
+        self.prepare_for_chroot()?;
+
+        // Safe to unwrap because we check in the call above that `file_name()` is successful.
+        // The borrow checker doesn't let us to nicely use an `&OsStr` as function parameter
```
dianpopa, line 260: `OsStr` is a borrowed reference to an `OsString`from inside `self`. That is why the `self` is seen as mutably borrowed by the checker. What you can do is make `prepare_for_chroot `return an `OsString`. Something around:
```
fn prepare_for_chroot(&mut self) -> Result<(OsString)> {
...
Ok(exec_file_name.to_os_string())
}
```

```
pub fn run {
let exec_file_name = self.prepare_for_chroot()?;
let chroot_exec_file = PathBuf::from("/").join(&exec_file_name);

let cgroup = Cgroup::new(&self.id, self.numa_node, &exec_file_name)?;
}
```
WDYT?

src/jailer/src/env.rs
```diff
@@ -609,6 +609,80 @@ mod tests {
...
+    fn test_prepare_for_chroot() {
+        // Create a standard environment.
+        let arg_parser = build_arg_parser();
+        let mut args = arg_parser.arguments().clone();
+        // Not using here `ArgVals::new()` because `fs::copy()` from `prepare_for_chroot()`
+        // sees `/proc/cpuinfo` as an empty file and doesn't copy anything to destination.
+        let some_arg_vals = ArgVals {
+            node: "1",
+            id: "bd65600d-8669-4903-8a14-af88203add38",
+            exec_file: "file1",
```
dianpopa, line 636: Can`t we use `TempFile` for zzns and file1  and TempDir for `chroot_base`?

src/jailer/src/env.rs
lauralt: Nice! Indeed I tried something similar, but with `OsStr` instead of `OsString`:woman_facepalming:.
I'll update, thanks!

src/jailer/src/env.rs
lauralt: The files are created inside `mknod_and_own_dev()`. This function takes a `&'static [u8]` as parameter, I don't think `Tempfile` could help here.

I added an extra check at the beginning to be sure the file doesn't exist before attempting to create it, should be safer now :-??.

src/jailer/src/env.rs
lauralt: :+1: done.

Issue #1848, opened 2020-05-04
Snapshot create/restore performance metrics


acatangiu:

lauralt:
Fixed by #1945.

PR #1945, opened 2020-06-05
snapshot: add metrics for load and create

lauralt:
## Reason for This PR

#1848

## Description of Changes

Added metrics that measure the duration of creating a full or diff snapshot and of loading from a snapshot.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -105,8 +105,7 @@ impl ApiServer {
         let mut server = HttpServer::new(path).expect("Error creating the HTTP server");
 
         if let Some(start_time) = start_time_us {
-            let delta_us =
-                (utils::time::get_time(utils::time::ClockType::Monotonic) / 1000) - start_time;
+            let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic) - start_time;
             METRICS
                 .api_server
                 .process_startup_time_us
@@ -115,7 +114,7 @@ impl ApiServer {
 
         if let Some(cpu_start_time) = start_time_cpu_us {
             let delta_us =
-                utils::time::get_time(utils::time::ClockType::ProcessCpu) / 1000 - cpu_start_time;
+                utils::time::get_time_us(utils::time::ClockType::ProcessCpu) - cpu_start_time;
             METRICS
                 .api_server
                 .process_startup_time_cpu_us
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -151,6 +151,9 @@ impl ApiServer {
                                 error!("API Server encountered an error on response: {}", e);
                                 Ok(())
                             })?;
+                        let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic)
+                            - request_processing_start_us;
+                        debug!("Total previous API call duration: {} us.", delta_us);
                     }
                 }
                 Err(e) => {
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -21,7 +21,7 @@ use std::path::PathBuf;
 use std::sync::{mpsc, Arc, Mutex, RwLock};
 use std::{fmt, io};
 
-use logger::{Metric, METRICS};
+use logger::{update_metric_with_elapsed_time, Metric, METRICS};
 pub use micro_http::{
     Body, HttpServer, Method, Request, RequestError, Response, ServerError, ServerRequest,
     ServerResponse, StatusCode, Version,
@@ -33,6 +33,8 @@ use seccomp::{BpfProgram, SeccompFilter};
 use utils::eventfd::EventFd;
 use vmm::rpc_interface::{VmmAction, VmmActionError, VmmData};
 use vmm::vmm_config::instance_info::InstanceInfo;
+#[cfg(target_arch = "x86_64")]
+use vmm::vmm_config::snapshot::SnapshotType;
 
 /// Shorthand type for a request containing a boxed VmmAction.
 pub type ApiRequest = Box<VmmAction>;
@@ -136,10 +138,14 @@ impl ApiServer {
             match server.requests() {
                 Ok(request_vec) => {
                     for server_request in request_vec {
+                        let request_processing_start_us =
+                            utils::time::get_time_us(utils::time::ClockType::Monotonic);
                         server
                             .respond(
                                 // Use `self.handle_request()` as the processing callback.
-                                server_request.process(|request| self.handle_request(request)),
+                                server_request.process(|request| {
+                                    self.handle_request(request, request_processing_start_us)
+                                }),
                             )
                             .or_else(|e| {
                                 error!("API Server encountered an error on response: {}", e);
@@ -157,9 +163,11 @@ impl ApiServer {
         }
     }
 
-    fn handle_request(&self, request: &Request) -> Response {
+    fn handle_request(&self, request: &Request, request_processing_start_us: u64) -> Response {
         match ParsedRequest::try_from_request(request) {
-            Ok(ParsedRequest::Sync(vmm_action)) => self.serve_vmm_action_request(vmm_action),
+            Ok(ParsedRequest::Sync(vmm_action)) => {
+                self.serve_vmm_action_request(vmm_action, request_processing_start_us)
+            }
             Ok(ParsedRequest::GetInstanceInfo) => self.get_instance_info(),
             Ok(ParsedRequest::GetMMDS) => self.get_mmds(),
             Ok(ParsedRequest::PatchMMDS(value)) => self.patch_mmds(value),
@@ -171,13 +179,47 @@ impl ApiServer {
         }
     }
 
-    fn serve_vmm_action_request(&self, vmm_action: Box<VmmAction>) -> Response {
+    fn serve_vmm_action_request(
+        &self,
+        vmm_action: Box<VmmAction>,
+        request_processing_start_us: u64,
+    ) -> Response {
+        let metric_with_action = match *vmm_action {
+            #[cfg(target_arch = "x86_64")]
+            VmmAction::CreateSnapshot(ref params) => match params.snapshot_type {
+                SnapshotType::Full => Some((
+                    &METRICS.latencies_us.full_create_snapshot,
+                    "create full snapshot",
+                )),
+                SnapshotType::Diff => Some((
+                    &METRICS.latencies_us.diff_create_snapshot,
+                    "create diff snapshot",
+                )),
+            },
+            #[cfg(target_arch = "x86_64")]
+            VmmAction::LoadSnapshot(_) => {
+                Some((&METRICS.latencies_us.load_snapshot, "load snapshot"))
+            }
+            VmmAction::Pause => Some((&METRICS.latencies_us.pause_vm, "pause vm")),
+            VmmAction::Resume => Some((&METRICS.latencies_us.resume_vm, "resume vm")),
+            _ => None,
+        };
+
         self.api_request_sender
             .send(vmm_action)
             .expect("Failed to send VMM message");
         self.to_vmm_fd.write(1).expect("Cannot update send VMM fd");
         let vmm_outcome = *(self.vmm_response_receiver.recv().expect("VMM disconnected"));
-        ParsedRequest::convert_to_response(vmm_outcome)
+        let response = ParsedRequest::convert_to_response(&vmm_outcome);
+
+        if vmm_outcome.is_ok() {
+            if let Some((metric, action)) = metric_with_action {
+                let elapsed_time_us =
+                    update_metric_with_elapsed_time(metric, request_processing_start_us);
+                info!("'{}' API request took {} us.", action, elapsed_time_us);
+            }
+        }
+        response
     }
 
     fn get_instance_info(&self) -> Response {
@@ -271,9 +313,12 @@ mod tests {
     use micro_http::HttpConnection;
     use mmds::MMDS;
     use utils::tempfile::TempFile;
+    use utils::time::ClockType;
     use vmm::builder::StartMicrovmError;
     use vmm::rpc_interface::VmmActionError;
     use vmm::vmm_config::instance_info::InstanceInfo;
+    #[cfg(target_arch = "x86_64")]
+    use vmm::vmm_config::snapshot::CreateSnapshotParams;
 
     #[test]
     fn test_error_messages() {
@@ -331,8 +376,50 @@ mod tests {
                 StartMicrovmError::MicroVMAlreadyRunning,
             ))))
             .unwrap();
-        let response = api_server.serve_vmm_action_request(Box::new(VmmAction::StartMicroVm));
+        let response = api_server.serve_vmm_action_request(Box::new(VmmAction::StartMicroVm), 0);
         assert_eq!(response.status(), StatusCode::BadRequest);
+
+        let start_time_us = utils::time::get_time_us(ClockType::Monotonic);
+        assert_eq!(METRICS.latencies_us.pause_vm.count(), 0);
+        to_api.send(Box::new(Ok(VmmData::Empty))).unwrap();
+        let response =
+            api_server.serve_vmm_action_request(Box::new(VmmAction::Pause), start_time_us);
+        assert_eq!(response.status(), StatusCode::NoContent);
+        assert_ne!(METRICS.latencies_us.pause_vm.count(), 0);
+
+        #[cfg(target_arch = "x86_64")]
+        {
+            assert_eq!(METRICS.latencies_us.diff_create_snapshot.count(), 0);
+            to_api
+                .send(Box::new(Err(VmmActionError::OperationNotSupportedPreBoot)))
+                .unwrap();
+            let response = api_server.serve_vmm_action_request(
+                Box::new(VmmAction::CreateSnapshot(CreateSnapshotParams {
+                    snapshot_type: SnapshotType::Diff,
+                    snapshot_path: PathBuf::new(),
+                    mem_file_path: PathBuf::new(),
+                    version: None,
+                })),
+                start_time_us,
+            );
+            assert_eq!(response.status(), StatusCode::BadRequest);
+            // The metric should not be updated if the request wasn't successful.
+            assert_eq!(METRICS.latencies_us.diff_create_snapshot.count(), 0);
+
+            to_api.send(Box::new(Ok(VmmData::Empty))).unwrap();
+            let response = api_server.serve_vmm_action_request(
+                Box::new(VmmAction::CreateSnapshot(CreateSnapshotParams {
+                    snapshot_type: SnapshotType::Diff,
+                    snapshot_path: PathBuf::new(),
+                    mem_file_path: PathBuf::new(),
+                    version: None,
+                })),
+                start_time_us,
+            );
+            assert_eq!(response.status(), StatusCode::NoContent);
+            assert_ne!(METRICS.latencies_us.diff_create_snapshot.count(), 0);
+            assert_eq!(METRICS.latencies_us.full_create_snapshot.count(), 0);
+        }
     }
 
     #[test]
@@ -495,21 +582,21 @@ mod tests {
             .unwrap();
         assert!(connection.try_read().is_ok());
         let req = connection.pop_parsed_request().unwrap();
-        let response = api_server.handle_request(&req);
+        let response = api_server.handle_request(&req, 0);
         assert_eq!(response.status(), StatusCode::BadRequest);
 
         // Test a Get Info request.
         sender.write_all(b"GET / HTTP/1.1\r\n\r\n").unwrap();
         assert!(connection.try_read().is_ok());
         let req = connection.pop_parsed_request().unwrap();
-        let response = api_server.handle_request(&req);
+        let response = api_server.handle_request(&req, 0);
         assert_eq!(response.status(), StatusCode::OK);
 
         // Test a Get Mmds request.
         sender.write_all(b"GET /mmds HTTP/1.1\r\n\r\n").unwrap();
         assert!(connection.try_read().is_ok());
         let req = connection.pop_parsed_request().unwrap();
-        let response = api_server.handle_request(&req);
+        let response = api_server.handle_request(&req, 0);
         assert_eq!(response.status(), StatusCode::OK);
 
         // Test a Put Mmds request.
@@ -522,7 +609,7 @@ mod tests {
             .unwrap();
         assert!(connection.try_read().is_ok());
         let req = connection.pop_parsed_request().unwrap();
-        let response = api_server.handle_request(&req);
+        let response = api_server.handle_request(&req, 0);
         assert_eq!(response.status(), StatusCode::NoContent);
 
         // Test a Patch Mmds request.
@@ -535,7 +622,7 @@ mod tests {
             .unwrap();
         assert!(connection.try_read().is_ok());
         let req = connection.pop_parsed_request().unwrap();
-        let response = api_server.handle_request(&req);
+        let response = api_server.handle_request(&req, 0);
         assert_eq!(response.status(), StatusCode::NoContent);
 
         // Test erroneous request.
@@ -548,7 +635,7 @@ mod tests {
             .unwrap();
         assert!(connection.try_read().is_ok());
         let req = connection.pop_parsed_request().unwrap();
-        let response = api_server.handle_request(&req);
+        let response = api_server.handle_request(&req, 0);
         assert_eq!(response.status(), StatusCode::BadRequest);
     }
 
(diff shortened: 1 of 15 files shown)


src/logger/src/metrics.rs
```diff
@@ -194,6 +194,9 @@ pub trait Metric {
     }
     /// Returns current value of the counter.
     fn count(&self) -> usize;
+    #[cfg(target_arch = "x86_64")]
+    /// Stores `value` to the current counter.
+    fn store(&self, value: usize);
```
aghecenco, line 199: This is missing a unit test, and I think `test_metric` could be extended with one

aghecenco: I think this is a good place for `store` too https://github.com/firecracker-microvm/firecracker/blob/4cb9b39f03f58498195995b209e9119c075df9a0/src/api_server/src/lib.rs#L112
We've made do with `add` so far because the metric is only emitted once but `store` is conceptually better.

src/logger/src/metrics.rs
```diff
@@ -221,6 +224,13 @@ impl Metric for SharedMetric {
     fn count(&self) -> usize {
         self.0.load(Ordering::Relaxed)
     }
+
+    #[cfg(target_arch = "x86_64")]
+    // It is necessary to also reset the old value in order to serialize later the correct one.
```
aghecenco, line 229: The comment is a bit unclear - first things first, by `serialize` you refer to metric flush, not `persist`? If so, could you make the comment a bit more verbose (or point to `impl Serialize`) so that it's clear that after flush, the metric value will be the `delta` of `self.0 - self.1`?

src/logger/src/metrics.rs
```diff
@@ -489,6 +499,33 @@ pub struct SerialDeviceMetrics {
...
+// These store the duration of creating/loading a snapshot.
+// If there are more than one `/snapshot/create` request in a minute
+// (until the metrics are flushed), only the duration of the last
+// snapshot creation is stored in the metric. If the user is interested
+// in all the durations, a `FlushMetrics` request should be sent after
+// each `create` request.
+#[derive(Default, Serialize)]
+pub struct SnapshotMetrics {
+    /// Measures the snapshot full create time in microseconds.
+    pub last_full_create_us: SharedMetric,
```
aghecenco, line 522: nit: I think `last_` is too verbose and doesn't add value, metrics usually express the last instance of some event

src/vmm/src/rpc_interface.rs
```diff
@@ -427,10 +442,32 @@ impl RuntimeApiController {
     }
 
     #[cfg(target_arch = "x86_64")]
-    fn create_snapshot(&mut self, params: CreateSnapshotParams) -> ActionResult {
+    fn create_snapshot(&mut self, create_params: CreateSnapshotParams) -> ActionResult {
         let mut locked_vmm = self.vmm.lock().unwrap();
-        persist::create_snapshot(&mut locked_vmm, params, VERSION_MAP.clone())
-            .map_err(VmmActionError::CreateSnapshot)
+        let create_start_us = utils::time::get_time(utils::time::ClockType::Monotonic) / 1000;
```
aghecenco, line 447: nit: `utils::time::get_time(utils::time::ClockType::Monotonic) / 1000` shows up a lot, maybe it deserves a helper fn in `utils::time` - `get_time_us(ClockType)`?

src/logger/src/metrics.rs
lauralt: Done.

src/logger/src/metrics.rs
lauralt: I updated the comment, hope it's clearer now.

src/logger/src/metrics.rs
lauralt: I was thinking that this way it will be clearer that we don't store all the `create` values, but since this is the default behavior, it makes sense to not add `last_`.

src/vmm/src/rpc_interface.rs
lauralt: Done, also renamed `get_time` to `get_time_ns` to make things clearer.

src/vmm/src/rpc_interface.rs
```diff
@@ -325,6 +319,26 @@ impl<'a> PrebootApiController<'a> {
             CreateSnapshot(_) | SendCtrlAltDel => Err(VmmActionError::OperationNotSupportedPreBoot),
         }
     }
+
+    #[cfg(target_arch = "x86_64")]
+    fn load_snapshot(&mut self, load_params: LoadSnapshotParams) -> ActionResult {
+        let load_start_us = utils::time::get_time_us(utils::time::ClockType::Monotonic);
```
sandreim, line 325: load_start_us should be computed on the API thread. In case of high CPU load we expect some latency here and we should see it in the metrics.

src/vmm/src/rpc_interface.rs
```diff
@@ -427,10 +441,26 @@ impl RuntimeApiController {
     }
 
     #[cfg(target_arch = "x86_64")]
-    fn create_snapshot(&mut self, params: CreateSnapshotParams) -> ActionResult {
+    fn create_snapshot(&mut self, create_params: CreateSnapshotParams) -> ActionResult {
         let mut locked_vmm = self.vmm.lock().unwrap();
-        persist::create_snapshot(&mut locked_vmm, params, VERSION_MAP.clone())
-            .map_err(VmmActionError::CreateSnapshot)
+        let create_start_us = utils::time::get_time_us(utils::time::ClockType::Monotonic);
```
sandreim, line 446: Same comment here regarding create_start_us.

src/logger/src/metrics.rs
```diff
@@ -489,6 +498,33 @@ pub struct SerialDeviceMetrics {
...
+// in all the durations, a `FlushMetrics` request should be sent after
+// each `create` request.
+#[derive(Default, Serialize)]
+pub struct SnapshotMetrics {
+    /// Measures the snapshot full create time in microseconds.
+    pub full_create_us: SharedMetric,
+    /// Measures the snapshot diff create time in microseconds.
+    pub diff_create_us: SharedMetric,
+    /// Measures the snapshot load time in microseconds.
+    pub load_us: SharedMetric,
```
sandreim, line 525: The way I understand this metric is that it measures how much it takes for us from receiving the API call to when the VMM is built.

We also need a metric which measures the resume time. We could then do load_us + resume_use = restore_time in the dashboard.

The problem with this is that it does not show how much time is lost between load and resume. Another option would be to compute the resume_us metric to span from load api call to resume completion.

src/vmm/src/rpc_interface.rs
acatangiu: > load_start_us should be computed on the API thread

I disagree:
- That would break modularity and spread the logic over multiple components. It would also add vmm-specific logic to the otherwise completely decoupled api server.
- The metric itself will lose granularity in regards to the snapshot operation. This metric should ideally measure exclusively snapshot code.
- Regressions or inadequate performance in the API server code or the API<->VMM communication should have their own metrics.

src/logger/src/metrics.rs
lauralt: With the current implementation, resuming the microVM from snapshot is completely independent of the load operation. If we measure them separately and then do `load_us + resume_us`, the result won't tell us what we want to know, i.e. how much does it take to resume from snapshot. If we want to define the `resume_us` metric as spanning from load api call to resume completion, we kind of assume that the user will resume the vm immediately after loading and, even if it's not the most expected scenario, resuming can happen also after a few minutes (and basically anytime) and the metric won't be that useful.

Anyway, this metric (`resume_us`) would make sense only if we include indeed the API flow too(as you've suggested in another comment here), but since these operations are separate (and moreover, one of them is pre boot and the after post boot), I'm not a big fan of combining them in a metric :(. I also expect the code to get pretty ugly and counter-intuitive (among others, we will have to check in the resume function if the microVM was started from a snapshot, if we don't do that , `resume_us` will be modified also when resuming after a `create` and this way the metric will have a different meaning). I expect this metric would make more sense when resuming from a snapshot in one shot and not with 2 separate APIs.

src/vmm/src/rpc_interface.rs
sandreim: I don't fully disagree with @acatangiu's point, but from an UX point of view I want to be able to get see how much time does it take since receiving the API call up until the snapshot is loaded. 

I do not have a hard preference on how this should be implemented, if it is possible to make them modular so one can  add api_latency_us + load_time_us + resume_time_us and get the e2e timing.

src/logger/src/metrics.rs
sandreim: Resuming in one shot is possible if the REST client reuses the connection and does not create one for each request: Connection: keep-alive header.

sandreim:
I want to make sure we implement this in such a way that these metrics provide maximum value to the end user and not just for development. Let's sync up offline on this one.

tests/integration_tests/functional/test_metrics.py
```diff
@@ -45,6 +47,8 @@ def test_flush_metrics(test_microvm_with_api):
         'uart',
         'signals',
     ]
+    if platform.machine() != "aarch64":
```
dianpopa, line 50: I would not label the snapshotting metrics as being x86_64 specific in logger/metrics. The build is fine that way too.

src/vmm/src/rpc_interface.rs
```diff
@@ -325,6 +319,26 @@ impl<'a> PrebootApiController<'a> {
             CreateSnapshot(_) | SendCtrlAltDel => Err(VmmActionError::OperationNotSupportedPreBoot),
         }
     }
+
+    #[cfg(target_arch = "x86_64")]
+    fn load_snapshot(&mut self, load_params: LoadSnapshotParams) -> ActionResult {
```
dianpopa, line 324: This could be 
```suggestion
    fn load_snapshot(&mut self, load_params: &LoadSnapshotParams) -> ActionResult {
```

src/vmm/src/rpc_interface.rs
```diff
@@ -427,10 +441,26 @@ impl RuntimeApiController {
     }
 
     #[cfg(target_arch = "x86_64")]
-    fn create_snapshot(&mut self, params: CreateSnapshotParams) -> ActionResult {
+    fn create_snapshot(&mut self, create_params: CreateSnapshotParams) -> ActionResult {
```
dianpopa, line 444:
```suggestion
    fn create_snapshot(&mut self, create_params: &CreateSnapshotParams) -> ActionResult {
```

src/vmm/src/rpc_interface.rs
```diff
@@ -427,10 +441,26 @@ impl RuntimeApiController {
...
+
+        persist::create_snapshot(&mut locked_vmm, create_params.clone(), VERSION_MAP.clone())
+            .map_err(VmmActionError::CreateSnapshot)?;
+
+        let delta_us =
+            utils::time::get_time_us(utils::time::ClockType::Monotonic) - create_start_us;
+        match create_params.snapshot_type {
+            SnapshotType::Full => {
+                METRICS.snapshot.full_create_us.store(delta_us as usize);
+                info!("Creating the full snapshot took {} us.", delta_us);
```
dianpopa, line 456: nit: us -> microseconds

src/logger/src/metrics.rs
lauralt: I added it in the same commit to kind of justify the need for it (the snapshot context), but since it can be used in other places and it's an independent thing I'll move it though to another commit.

src/vmm/src/rpc_interface.rs
```diff
@@ -325,6 +319,26 @@ impl<'a> PrebootApiController<'a> {
...
+        let loaded_vmm = persist::load_snapshot(
+            &mut self.event_manager,
+            &self.seccomp_filter,
+            load_params,
+            VERSION_MAP.clone(),
+        );
+
+        let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic) - load_start_us;
+        METRICS.snapshot.load_us.store(delta_us as usize);
+        info!("Loading the snapshot took {} us.", delta_us);
```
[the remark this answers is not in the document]
lauralt: `us` is a widely accepted simplification for `μs`. `microseconds` is too verbose in this message IMO and I wouldn't change it with `μs` since this may involve Unicode issues for some editors I guess :-?.

src/vmm/src/vmm_config/snapshot.rs
```diff
@@ -7,7 +7,7 @@ use std::path::PathBuf;
 
 /// The snapshot type options that are available when
 /// creating a new snapshot.
-#[derive(Debug, Deserialize, PartialEq, Serialize)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
```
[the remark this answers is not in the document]
lauralt: I agree, at that moment the `&` change seemed to involve too many, unrelated to this PR, changes (in persist code, in tests) so I kind of took my mind off it for a while and unfortunately didn't insist after that on it :(, sorry for that. I'll add a separate commit with this change.

tests/integration_tests/functional/test_metrics.py
lauralt: I thought about this for a while and decided in the end to leave them only on the platform on which snapshotting is available now (and remove the `x86_64` label when we will have support on `aarch64` too). I think it's confusing to print metrics (even with `0` values) for a feature that it's not actually supported.

src/vmm/src/vmm_config/snapshot.rs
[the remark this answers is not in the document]
acatangiu: Seeing how `enum SnapshotType` is just an identifier (small in size), I'm fine with using clones as well. References will not be faster.

src/vmm/src/rpc_interface.rs
```diff
@@ -387,20 +402,42 @@ impl RuntimeApiController {
...
             .lock()
             .expect("Poisoned lock")
             .pause_vcpus()
-            .map_err(VmmActionError::InternalVmm)
+            .map_err(VmmActionError::InternalVmm)?;
+
+        update_metric_with_elapsed_time(
+            &METRICS.latencies_us.vmm_pause_vm,
+            pause_start_us,
+            "pausing the microVM",
```
sandreim, line 416: I think it would be much better like this:
```suggestion
            "pause_vcpus()",
```

src/vmm/src/rpc_interface.rs
```diff
@@ -387,20 +402,42 @@ impl RuntimeApiController {
...
             .lock()
             .expect("Poisoned lock")
             .resume_vcpus()
-            .map_err(VmmActionError::InternalVmm)
+            .map_err(VmmActionError::InternalVmm)?;
+
+        update_metric_with_elapsed_time(
+            &METRICS.latencies_us.vmm_resume_vm,
+            resume_start_us,
+            "resuming the microVM",
```
sandreim, line 436: resume_vcpus()

src/vmm/src/rpc_interface.rs
```diff
@@ -427,10 +464,32 @@ impl RuntimeApiController {
...
+
+        persist::create_snapshot(&mut locked_vmm, create_params, VERSION_MAP.clone())
+            .map_err(VmmActionError::CreateSnapshot)?;
+
+        match create_params.snapshot_type {
+            SnapshotType::Full => {
+                update_metric_with_elapsed_time(
+                    &METRICS.latencies_us.vmm_full_create_snapshot,
+                    create_start_us,
+                    "creating the full snapshot",
```
sandreim, line 479: Hmmm, now "create_snapshot()" as action string would be bad here as we also need to account for inc/full. So how about "create_snapshot(diff/full)" ?

src/api_server/src/lib.rs
```diff
@@ -137,15 +138,25 @@ impl ApiServer {
...
+                                }),
                             )
                             .or_else(|e| {
                                 error!("API Server encountered an error on response: {}", e);
                                 Ok(())
                             })?;
+                        let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic)
+                            - request_processing_start_us;
+                        debug!(
+                            "Handling the request from the previous API call, including waiting for the VMM response, took {} us.",
```
acatangiu, line 157: Please use a short message like `Total API call time: %u`.
All the long static strings throughout the codebase add up and bloat the binary size.

src/logger/src/lib.rs
```diff
@@ -32,3 +32,17 @@ fn extract_guard<G>(lock_result: LockResult<G>) -> G {
...
 }
+
+pub fn update_metric_with_elapsed_time(
+    metric: &SharedMetric,
+    start_time_us: u64,
+    action_to_log: &str,
+    thread_name: &str,
+) {
+    let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic) - start_time_us;
+    info!(
```
acatangiu, line 43: keep this the log level consistent, you currently have a mix of `info!` and `debug!`

src/logger/src/lib.rs
```diff
@@ -32,3 +32,17 @@ fn extract_guard<G>(lock_result: LockResult<G>) -> G {
         Err(poisoned) => poisoned.into_inner(),
     }
 }
+
+pub fn update_metric_with_elapsed_time(
+    metric: &SharedMetric,
+    start_time_us: u64,
+    action_to_log: &str,
+    thread_name: &str,
```
acatangiu, line 40: Instead of passing in thread name as `&str` just use [ThreadId](https://doc.rust-lang.org/std/thread/struct.ThreadId.html) in the debug message.

src/api_server/src/lib.rs
```diff
@@ -172,13 +185,45 @@ impl ApiServer {
...
+        &self,
+        vmm_action: Box<VmmAction>,
+        action_start_us: u64,
+    ) -> Response {
+        let metric_with_action = match *vmm_action {
+            #[cfg(target_arch = "x86_64")]
+            VmmAction::CreateSnapshot(ref params) => match params.snapshot_type {
+                SnapshotType::Full => Some((
+                    &METRICS.latencies_us.full_create_snapshot,
+                    "creating the full snapshot",
```
acatangiu, line 198: nit:
Since these are debug messages (developer-oriented), they can be shorter and not necessarily prettified:
```suggestion
                    "create full snapshot",
```

`load snapshot`
`pause vm`
`resume vm`
etc

src/api_server/src/lib.rs
```diff
@@ -172,13 +185,45 @@ impl ApiServer {
...
             .send(vmm_action)
             .expect("Failed to send VMM message");
         self.to_vmm_fd.write(1).expect("Cannot update send VMM fd");
         let vmm_outcome = *(self.vmm_response_receiver.recv().expect("VMM disconnected"));
-        ParsedRequest::convert_to_response(vmm_outcome)
+        let response = ParsedRequest::convert_to_response(&vmm_outcome);
+
+        if vmm_outcome.is_ok() {
+            if let Some((metric, action)) = metric_with_action {
+                update_metric_with_elapsed_time(metric, action_start_us, action, "API");
```
acatangiu, line 223: I'm not sure about the added value of the `update_metric_with_elapsed_time()` function.

```suggestion
                let elapsed = utils::elapsed_us(action_start_us);
                metric.store(elapsed);
                debug!("API thread took {} us for {:?}", elapsed, vmm_action);
```

You get a couple duplicated lines but it's easier to read and maintain (less bytes used too).

Avoids the `action` static string description and only formats `vmm_action` when log-level is debug.

src/logger/src/lib.rs
lauralt: `debug!` is only used for printing the total API time for each request (not related to the snapshot ones). I used `info!` in the other places in case the user is interested in those durations (at the `Info` level) and doesn't want to configure the metrics (also for `create` requests you will have all those durations without calling `FlushMetrics` more times). Do you think `debug!` would be more appropriate for these too?
In case we decide for `debug!` everywhere, I can update all the log messages to not be prettified, for `info!` on the other hand I think those sentences should sound somehow right/clear.

src/api_server/src/lib.rs
lauralt: I agree it's much clearer without the separate function that does all that stuff, I actually wanted to separate the logging part and add a function in a `SharedMetric` implementation: `update_with_elapsed_time(&self, start_time)` in order to avoid the ugly metric parameter.
This duplicate code happens six times, so I thought it deserves a separate function though (and not pollute the ones from the vmm crate with all these logs/metrics).
I think I'll leave `update_metric_with_elapsed_time(metric, action_start_us)` outside of any `SharedMetric` implementation and return the elapsed time (and the logging part will be done in `vmm`); sounds like a good compromise solution to me :-?.

src/logger/src/lib.rs
acatangiu: Production scenarios most likely use metrics and not logs to measure these things - logs don't scale well. Even so, I'm not sure what's the right level for these log messages... Both `debug!` and `info!` make sense, so I leave it to your personal choice.

> In case we decide for debug! everywhere, I can update all the log messages to not be prettified, for info! on the other hand I think those sentences should sound somehow right/clear.

I partially disagree. _Clear_ messages should always be the case, _pretty_ messages are just nice to have - but not at the expense of other things. I'm happy with clear even pretty messages as long as they can be kept short. I believe a small binary and small logfile footprint is much more appreciated than highly descriptive log messages.

src/api_server/src/lib.rs
acatangiu: I thought about `update_metric_with_elapsed_time(metric, action_start_us)` to return elapsed time as well, but in the end I think the function is still weird.

```rust
let elapsed = utils::elapsed_us(action_start_us);
metric.store(elapsed);
```
does the same thing in a nice way keeping separation of concerns, vs.
```rust
let elapsed = update_metric_with_elapsed_time(metric, action_start_us);
```

The second option is also _fine_ if you insist on it, not gonna block the PR for it.

src/api_server/src/lib.rs
```diff
@@ -172,13 +185,45 @@ impl ApiServer {
         }
     }
 
-    fn serve_vmm_action_request(&self, vmm_action: Box<VmmAction>) -> Response {
+    fn serve_vmm_action_request(
+        &self,
+        vmm_action: Box<VmmAction>,
+        action_start_us: u64,
```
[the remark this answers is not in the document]
lauralt: Yep, I hoped the shorter name would be accurate too for this function context, but it is somehow confusing. I updated it.

src/api_server/src/lib.rs
lauralt: Done.

src/api_server/src/lib.rs
lauralt: `let elapsed = SOME_METRIC.update_with_elapsed_time(action_start_us);` is definitely weird and wanted to avoid. Option 2 seems ok to me.

src/logger/src/lib.rs
lauralt: 'create full snapshot' VMM action took 2754442 us.
'create full snapshot' API request took 2755308 us.
Shorter than this doesn't seem clear to me. Unfortunately, sometimes is actually hard to be (very) clear without lot of words :(.

src/api_server/src/lib.rs
lauralt: Done.

src/vmm/src/rpc_interface.rs
```diff
@@ -325,6 +317,26 @@ impl<'a> PrebootApiController<'a> {
...
+
+        let loaded_vmm = persist::load_snapshot(
+            &mut self.event_manager,
+            &self.seccomp_filter,
+            load_params,
+            VERSION_MAP.clone(),
+        );
+
+        let elapsed_time_us =
+            update_metric_with_elapsed_time(&METRICS.latencies_us.vmm_load_snapshot, load_start_us);
```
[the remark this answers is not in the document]
lauralt: I can do persistence too :joy::

Those lines and `elapsed_us_since()` definition compensate each other :D.
Anyway, it was a matter of coverage too, I expect the coverage will decrease even more with this change and it already decreased pretty much in this PR.
Btw, `cargo fmt` does the following thing for some of them:
```rust
                let elapsed_us = utils::time::elapsed_us_since(create_start_us);
                METRICS
                    .latencies_us
                    .vmm_full_create_snapshot
                    .store(elapsed_us);
                info!(
                    "'create full snapshot' VMM action took {} us.",
                    elapsed_time_us
                );
```
I understand your point, but I don't think there is a solution which is clearly better than the other as the logging and metrics stuff are longer anyway than what we actually want to do in the calling functions (create a snapshot etc.). This is why I tried at the beginning to put everything that is logger/metrics related in just one function and separate stuff this way.

lauralt:
@acatangiu I had to rebase the coverage file, please re-approve/take another look, thanks!

lauralt:
@sandreim @acatangiu I rebased again :(, please re-approve, thanks!

sandreim:
> @sandreim @acatangiu I rebased again :(, please re-approve, thanks!

Looks like you still need to rebase again :)

PR #1954, opened 2020-06-11
Remove CPUID caching.

### LINKED ISSUE #1942: Snapshot: Local timer interrupt is never raised after resuming from 2nd incremental snapshot.
Steps to reproduce:

- Start fresh microvm
- Pause/Create snapshot
- Kill Firecracker.
- Start new Firecracker.
- Load/Resume previous snapshot
- Guest is receiving local timer interrupts (run `cat /proc/interrupts` several times to verify this)
- Pause/Create snapshot
- Kill Firecracker.
- Start new Firecracker.
- Load/Resume previous snapshot
- No local timer interrupts are raised (run `cat /proc/interrupts` several times to verify this)

**Root cause**: The Firecracker implementation does not use KVM_GET_CPUID2 to retrieve the cpuid entries and caches the values used with KVM_SET_CPUID2. The snapshot restore path does not cache these values and an invalid default value will be saved in the next snapshot created from this resumed state. When this new snapshot is restored it will no set the TSC_DEADLINE_TIMER flag and the KVM  in-kernel LAPIC emulation will not use the TSC deadline feature and will fallback to LVT_TIMER_ONESHOT. The guest is unaware of this underlying change in CPU features and does not attempt to reconfigure the LAPIC.

Solutions: 

1. Cache the cpuid values from the snapshot on restore time.
2. No caching, always call KVM_GET_CPUID2.

sandreim:
## Reason for This PR

Fixes #1942

## Description of Changes

Use KVM_GET_CPUID2 in Vcpu::save_state().
    
Currently the CPUID is cached only at microvm boot time. The restore path
fails to cache the CPUID from the snapshot and will save a bogus value when
a new snapshot is created from the current state. When such a snapshot is loaded
it will cause the in-kernel KVM LAPIC emulation to setup incorrectly and
other undefined behaviour.

This commit removes CPUID caching from Vcpu and calls KVM_GET_CPUID2 on
the Vcpu::save_state() path.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/vstate.rs
+++ b/src/vmm/src/vstate.rs
@@ -142,6 +142,9 @@ pub enum Error {
     /// Cannot run the VCPUs.
     VcpuRun(kvm_ioctls::Error),
     #[cfg(target_arch = "x86_64")]
+    /// Failed to get KVM vcpu cpuid.
+    VcpuGetCpuid(kvm_ioctls::Error),
+    #[cfg(target_arch = "x86_64")]
     /// Failed to set KVM vcpu cpuid.
     VcpuSetCpuid(kvm_ioctls::Error),
     #[cfg(target_arch = "x86_64")]
@@ -278,6 +281,8 @@ impl Display for Error {
             VcpuGetXcrs(e) => write!(f, "Failed to get KVM vcpu xcrs: {}", e),
             #[cfg(target_arch = "x86_64")]
             VcpuGetXsave(e) => write!(f, "Failed to get KVM vcpu xsave: {}", e),
+            #[cfg(target_arch = "x86_64")]
+            VcpuGetCpuid(e) => write!(f, "Failed to get KVM vcpu cpuid: {}", e),
             #[cfg(target_arch = "x86_64")]
             VcpuSetCpuid(e) => write!(f, "Failed to set KVM vcpu cpuid: {}", e),
             #[cfg(target_arch = "x86_64")]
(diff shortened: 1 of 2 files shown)


src/vmm/src/vstate.rs
```diff
@@ -717,7 +720,6 @@ impl Vcpu {
     pub fn new_x86_64(
         id: u8,
         vm_fd: &VmFd,
-        cpuid: CpuId,
```
lauralt, line 720: `cpuid` should be removed from the list of arguments in the function comment. Also, the comment on line 731 should be updated/removed.

sandreim: done.

sandreim:
> Looks good.
> nit: would be nice to update commit message for [163c36b](https://github.com/firecracker-microvm/firecracker/commit/163c36ba65409dc78676bf8ce9b8c7ef16dc5b25) to follow the 50/72 rule.

I've updated the commit message.

src/vmm/src/vstate.rs
```diff
@@ -960,8 +960,12 @@ impl Vcpu {
             .fd
             .get_vcpu_events()
             .map_err(Error::VcpuGetVcpuEvents)?;
+
         Ok(VcpuState {
-            cpuid: self.cpuid.clone(),
+            cpuid: self
+                .fd
+                .get_cpuid2(kvm_bindings::KVM_MAX_CPUID_ENTRIES)
```
acatangiu, line 967: Are there any ordering requirements between `GET_CPUID` and the other `GET_*` ? If there are and we don't follow them we could introduce subtle and sporadic failures.

src/vmm/src/vstate.rs
```diff
@@ -806,16 +805,19 @@ impl Vcpu {
     /// * `machine_config` - The machine configuration of this microvm needed for the CPUID configuration.
```
lauralt, line 805: Probably some old leftover :(.

src/vmm/src/vstate.rs
sandreim: AFAIK there are no explicit requirements here.

Issue #1652, opened 2020-03-06
strip firecracker release binary


serban300:
Looks like simply calling `strip` on the firecracker release binary reduces the size from 2.5 MB to 1.4 MB. Investigate if this affects in any way the functionality.

sandreim:
IMO this will affect debugability by less detailed or useless backtraces, but it should not affect any functionality. I am not sure if it is possible with rust, but if it's done at compile time it can boost performance.

raduweiss:
Yeah, according to the `strip` command's description, the ability to interpret back/stack traces is affected. But, what do we gain? Does the binary size reduction also result in memory usage reduction? If not, does the binary size reduction help in any other way?

petreeftime:
Probably `strip --strip-debug --discard-locals` is a better option than what `strip` does by default for production builds. You can still get some useful information from stack traces that way.

serban300:
@raduweiss in this case the binary size reduction doesn't result in a memory usage reduction. It's only a disk size optimization. Does this bring any value ? If not, we can close the issue.

acatangiu:
Closing this issue as we yet have no indicators for relevant positive gains which would result from stripping the firecracker release binary.

We can reopen if later needed.

iulianbarbu:
Reopening this.

iulianbarbu:
The PR was merged.

PR #1957, opened 2020-06-12
devtool: strip release binaries out of debug info

iulianbarbu:
## Reason for This PR

Cargo adds by default DWARF sections to release binaries. It is recommended to strip release binaries out of debug information because of security concerns, but also because this information is not relevant in a production environment.

#1652

## Description of Changes

The provided solution adds a devtool command (`devtool strip [--target-libc (musl|gnu)]`) which strips release binaries out of debug information (DWARF sections). This solution is based on `GNU strip` tool, which strips away only debug information (using --strip-debug).

## License Acceptance

## PR Checklist


.cargo/config
```diff
@@ -1,12 +1,3 @@
...
-
-[target.'cfg(any(target_arch="arm", target_arch="aarch64"))']
-# On aarch64 musl depends on some libgcc functions (i.e `__addtf3` and other `*tf3` functions) for logic that uses
-# long double. Such functions are not builtin in the rust compiler, so we need to get them from libgcc.
-# No need for the `crt_static` flag as rustc appends it by default.
-rustflags = [
-	"-C", "link-arg=-lgcc",
-	"-C", "link-arg=-lfdt",
-]
+target-dir = "build/cargo_target"
```
aghecenco, line 3: nit: add newline

tools/devtool
```diff
@@ -475,14 +475,22 @@ cmd_build() {
         cargo_args+=(--target "$target")
     fi
 
-    [ $profile = "release" ] && cargo_args+=("--release")
+    machine=`uname -m`
+    # On aarch64 musl depends on some libgcc functions (i.e `__addtf3` and other `*tf3` functions) for logic that uses
+    # long double. Such functions are not builtin in the rust compiler, so we need to get them from libgcc.
+    # No need for the `crt_static` flag as rustc appends it by default.
+    [ $machine = "aarch64" ] && RUSTFLAGS="-C link-arg=-lgcc -C link-arg=-lfdt"
```
aghecenco, line 482: nit: you can [collapse](https://doc.rust-lang.org/rustc/codegen-options/index.html#link-args) multiple linker args into `link-args=-lgcc -lfdt`

tools/devtool
```diff
@@ -475,14 +475,22 @@ cmd_build() {
...
+    # Strip away debug info, since cargo adds them by default event when building artifacts for release.
+    # See: https://github.com/rust-lang/cargo/issues/4122.
+    [ $profile = "release" ] && cargo_args+=("--release") && RUSTFLAGS="$RUSTFLAGS -C link-arg=-s"
```
aghecenco, line 485: I would leave it to the user to decide whether or not they want symbols. We might need the unstripped binary for debugging, and with this change the user can't force a build with symbols. How about a new devtool parameter (`devtool build --release --keepsyms` or smth) that lets us to both?

iulianbarbu: Good idea. To not break things, we might want to leave the default behavior to be as it is and provide a flag `--strip_debug_info`, to enforce debug info stripping.

sandreim:
I am not sure we need to do this as part of the build system. I would leave the build system just as it is and move the stripping a step further into the release pipeline where the release binary is stripped and a copy of the original is kept in cold storage for debugging purposes.

tools/devtool
dianpopa: I would not bother modifying the `devtool `for this since users always have the choice of using "cargo build --release" if they want the unaltered version of what cargo obtains.

tools/devtool
```diff
@@ -475,14 +475,22 @@ cmd_build() {
         cargo_args+=(--target "$target")
     fi
 
-    [ $profile = "release" ] && cargo_args+=("--release")
+    machine=`uname -m`
+    # On aarch64 musl depends on some libgcc functions (i.e `__addtf3` and other `*tf3` functions) for logic that uses
+    # long double. Such functions are not builtin in the rust compiler, so we need to get them from libgcc.
+    # No need for the `crt_static` flag as rustc appends it by default.
+    [ $machine = "aarch64" ] && RUSTFLAGS="-C link-arg=-lgcc -C link-arg=-lfdt"
+    # Strip away debug info, since cargo adds them by default event when building artifacts for release.
```
dianpopa, line 483: nit: typo for "event"

tools/devtool
petreeftime: This removes all symbols, not only debug symbols. This means stack traces would be difficult to read. Keeping a small subset of symbols via `strip --discard-locals --strip-debug` would allow you to remove most symbols without breaking stack traces. It also means you don't have to change the linking, which doesn't work the same on all toolchains.

iulianbarbu: > I would not bother modifying the devtool for this since users always have the choice of using "cargo build --release" if they want the unaltered version of what cargo obtains.
> I would not bother modifying the devtool for this since users always have the choice of using "cargo build --release" if they want the unaltered version of what cargo obtains.

@dianpopa , do you suggest to implement stripping support as the default way of building the release binaries? Now I would prefer adding _optional_ support for stripping, when building with `tools/devtool build --release`, rather than adding by default stripping support while being open to break customers. Using `cargo build --release` to get unstripped binaries depends on setting up the rust dev environment 

I see two ways for implementing this, but I prefer the first one, since it involves the minimum changes to devtool which will not break customers or force them to install any dependency on their machines, to be able to build Firecracker like before:
* adding a flag to devtool, which strips away debug info would not break customers and will provide optional support for stripping debug info.
* stripping default info by default, when building the release binaries with devtool, will drop the support for building release binaries the "cargo" way (with debug info). If user need to build release binaries with `cargo build --release` they would need to set up rust dev environment (rust, cargo, rustup, toolchains etc).

> This removes all symbols, not only debug symbols. This means stack traces would be difficult to read. [...] 

@petreeftime , stripping only what is necessary is a valid point. I wasn't aware of selective stripping until now. I'll try to incorporate your suggestion. 

> It also means you don't have to change the linking, which doesn't work the same on all toolchains.

This argument conflicts with the suggestion to use `strip`. `strip` is part of the `GNU binutils`, which currently provides `GNU ld` and `GNU as`, used by `tools/devtool build --release`. So even if we change the toolchain, we can not say that we move away from GNU binutils (this is a core part of Linux toolchains in general). To me it seems that `GNU ld` and `GNU as` (GNU binutils in general) are core parts of Linux toolchains and changing stripping alternative from linker flags to strip tool it shouldn't make any difference while `GNU binutils` is still used as a core part of the toolchain we build against.

andreeaflorescu:
> I am not sure we need to do this as part of the build system. I would leave the build system just as it is and move the stripping a step further into the release pipeline where the binary is stripped and a copy of the original is kept in cold storage for debugging purposes.
> @firecracker-microvm/compute-capsule WDYT ?

+1 on this.

I was imagining this being implemented such that we have a devtool command that does building release binaries + stripping them at once (or as separate commands), without impacting in any way `cargo build` or `./tool/devtool build`.

With that in mind, I wouldn't like us to modify `./cargo/config`. Can this be done at the `./tools/devtool ` level only? Maybe with overriding `RUSTFLAGS` and exposing a new devtool command such as `build-release-stripped` (or some other better name). This command can then be used by an automated release process to create & upload the release artifacts.

iulianbarbu:
> I am not sure we need to do this as part of the build system. I would leave the build system just as it is and move the stripping a step further into the release pipeline where the binary is stripped 

Why to separate the stripping of debug symbols (embedded into the release binaries at build time) in a separate step, inside the release pipeline? My understanding is that these debug symbols can be added or not when we build the executable files, so to me it seems that this is the step were we should provide optional build flags which can be used for building the binaries with our without debug info. Let me know what you think.

> and a copy of the original is kept in cold storage for debugging purposes.

We wouldn't need to store binaries with debug symbols. We can get these binaries if we checkout on the release tag and build it.

iulianbarbu:
> +1 on this.
> 
> I was imagining this being implemented such that we have a devtool command that does building release binaries + stripping them at once (or as separate commands), without impacting in any way `cargo build` or `./tool/devtool build`. With that in mind, I wouldn't like us to modify `./cargo/config`. 

Right, this makes sense. However, having something inside the devtool for building and stripping at once was not the suggestion of @sandreim , as far as I can tell. I will keep the .cargo/config, to not impact `cargo build`.
 
> Can this be done at the `./tools/devtool ` level only? Maybe with overriding `RUSTFLAGS` and exposing a new devtool command such as `build-release-stripped` (or some other better name). This command can then be used by an automated release process to create & upload the release artifacts.

We can expose support for binaries stripping with a `devtool/build flag`, not necessary another build & strip command. What do you think?

andreeaflorescu:
> Right, this makes sense. However, having something inside the devtool for building and stripping at once was not the suggestion of @sandreim , as far as I can tell. I will keep the .cargo/config, to not impact `cargo build`.
> 
I probably misunderstood then, sorry.

> > Can this be done at the `./tools/devtool ` level only? Maybe with overriding `RUSTFLAGS` and exposing a new devtool command such as `build-release-stripped` (or some other better name). This command can then be used by an automated release process to create & upload the release artifacts.
> 
> We can expose support for binaries stripping with a `devtool/build` flag, not necessary another build & strip command. What do you think?

I think we should keep `./devtool build` as is. The current logic of `./devtool build` is to pass all the flags directly to `cargo build`. I've personally abused this functionality (and I find it very nice) to do things such as building firecracker with gnu. I personally prefer to have another `build-stripped` command that doesn't pass the flags to cargo build directly.

sandreim:
> I think we should keep `./devtool build` as is. The current logic of `./devtool build` is to pass all the flags directly to `cargo build`. I've personally abused this functionality (and I find it very nice) to do things such as building firecracker with gnu. I personally prefer to have another `build-stripped` command that doesn't pass the flags to cargo build directly.

+1 on this. Let's just have `devtool build` unchanged. Looks like strip is available in the dev container, why not make it available via `devtool strip`

iulianbarbu:
> > I think we should keep `./devtool build` as is. The current logic of `./devtool build` is to pass all the flags directly to `cargo build`. I've personally abused this functionality (and I find it very nice) to do things such as building firecracker with gnu. I personally prefer to have another `build-stripped` command that doesn't pass the flags to cargo build directly.
> 
> +1 on this. Let's just have `devtool build` unchanged. Looks like strip is available in the dev container, why not make it available via `devtool strip`

Correct. Looked within the container with `tools/devtool shell -p` and strip seems to be present in the container. I looked on the Dockerfile and couldn't see where it is pulled from. However, I guess it is shipped with `FROM ubuntu:18.04`. I'll incorporate your suggestions. Thanks.

iulianbarbu:
> I think you should also shrink the binary size limits in the associated integ test, the new binaries should be smaller.

@aghecenco , adding stripping support for devtool does not interfere with building firecracker binaries inside integration tests. Do you suggest to strip the binaries in the integration tests as well and then update the binaries sizes in the corresponding test?

andreeaflorescu:
> > I think you should also shrink the binary size limits in the associated integ test, the new binaries should be smaller.
> 
> @aghecenco , adding stripping support for devtool does not interfere with building firecracker binaries inside integration tests. Do you suggest to strip the binaries in the integration tests as well and then update the binaries sizes in the corresponding test?

The integration tests should be run with the binary that ends up being used by our customers.

iulianbarbu:
> > > I think you should also shrink the binary size limits in the associated integ test, the new binaries should be smaller.
> > 
> > 
> > @aghecenco , adding stripping support for devtool does not interfere with building firecracker binaries inside integration tests. Do you suggest to strip the binaries in the integration tests as well and then update the binaries sizes in the corresponding test?
> 
> The integration tests should be run with the binary that ends up being used by our customers.

I've stripped away debug_info, after integration tests build step and updated the `test_binary_size.py` test. PTAL! There were similar suggestions to have this in our integration tests as well, from @dpopa and @aghecen as well. PTAL!

tools/devtool
```diff
@@ -500,6 +509,54 @@ cmd_build() {
...
+        "$CTR_CARGO_TARGET_DIR/$target/$profile/jailer"
+    ret=$?
+
+    [ $ret -eq 0 ] && {
+        cargo_bin_dir="$CARGO_TARGET_DIR/$target/$profile"
+        say "Stripping was successful."
+        say "Stripped binaries placed under $cargo_bin_dir."
+    }
+
+    return $?
```
iulianbarbu, line 557: It needs one more change, before merging. I'll update soon.

```suggestion
    return $ret
```

iulianbarbu: Done.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
   snapshot.
 - Added a new API call, `PUT /snapshot/load`, for loading a snapshot.
 - Added metrics for the vsock device.
+- Added devtool strip command to strip release binaries out of DWARF sections.
```
dianpopa, line 28: hm, "strip the binary out of?" We are not removing DWARF sections, right? Should we?

Anyhow, I'd change this to something more obvious "Added `devtool strip` command which removes debug symbols from the release binaries."

tools/devtool
```diff
@@ -500,6 +509,54 @@ cmd_build() {
...
+                target="$(uname -m)-unknown-linux-${libc}"
+                ;;
+            *)
+                die "Unknown argument: $1. Please use --help for help."
+            ;;
+        esac
+        shift
+    done
+
+    # Check prerequisites
```
dianpopa, line 535: Another prerequisite would be to check that release binaries exist at the expected location, right?

tests/host_tools/cargo_build.py
```diff
@@ -69,6 +69,9 @@ def get_firecracker_binaries():
     fc_bin_path = "{}/{}".format(out_dir, FC_BINARY_NAME)
     jailer_bin_path = "{}/{}".format(out_dir, JAILER_BINARY_NAME)
 
+    cmd = "strip --strip-debug {} {}".format(fc_bin_path, jailer_bin_path)
+    utils.run_cmd(cmd)
```
gc-plp, line 73:
```suggestion
    utils.run_cmd(
        "strip --strip-debug {} {}".format(fc_bin_path, jailer_bin_path))
```

tools/devtool
iulianbarbu: I've added a check for release binaries existance.

iulianbarbu:
@dpopa, @gc-plp , I've addressed your feedback! PTAL.

dianpopa:
> @dpopa, @gc-plp , I've addressed your feedback! PTAL.

nit: @dianpopa :D
I will take a look!

Issue #1963, opened 2020-06-18
MMDS  requests containing a valid URL with double slash fails.


sandreim:
This works ``` curl http://169.254.170.2/path/to/resource ```
This doesnt: ``` curl http://169.254.170.2/path//to/resource ```

PR #1964, opened 2020-06-18
Sanitize Mmds request URI

sandreim:
## Reason for This PR

Fixes: #1963 

## Description of Changes

Provided in commit messages.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/mmds/src/lib.rs
+++ b/src/mmds/src/lib.rs
@@ -65,6 +65,19 @@ fn build_response(http_version: Version, status_code: StatusCode, body: Body) ->
     response
 }
 
+// Make the URI a correct JSON pointer value.
+fn sanitize_uri(mut uri: String) -> String {
+    let mut len = u32::MAX as usize;
+    // Loop while the deduping decreases the sanitized len.
+    // Each iteration will attempt to dedup "//".
+    while uri.len() < len {
+        len = uri.len();
+        uri = uri.replace("//", "/");
+    }
+
+    uri
+}
+
 pub fn parse_request(request_bytes: &[u8]) -> Response {
     let request = Request::try_from(request_bytes);
     match request {
@@ -88,12 +101,16 @@ pub fn parse_request(request_bytes: &[u8]) -> Response {
                 return response;
             }
 
+            // The data store expects a strict json path, so we need to
+            // sanitize the URI.
+            let json_pointer = sanitize_uri(uri.to_string());
+
             // The lock can be held by one thread only, so it is safe to unwrap.
             // If another thread poisoned the lock, we abort the execution.
             let response = MMDS
                 .lock()
                 .expect("Poisoned lock")
-                .get_value(uri.to_string(), request.headers.accept().into());
+                .get_value(json_pointer, request.headers.accept().into());
 
             match response {
                 Ok(response_body) => build_response(
--- a/src/mmds/src/lib.rs
+++ b/src/mmds/src/lib.rs
@@ -185,6 +185,25 @@ mod tests {
     }
 
     #[test]
+    fn test_sanitize_uri() {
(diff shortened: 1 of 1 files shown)


src/mmds/src/lib.rs
```diff
@@ -65,6 +65,20 @@ fn build_response(http_version: Version, status_code: StatusCode, body: Body) ->
...
 
+// Make the URI a correct JSON pointer value.
+fn sanitize_uri(uri: &str) -> String {
+    let mut sanitized_uri = uri.to_owned();
+    let mut len = u32::MAX as usize;
+    // Loop while the deduping decreases the sanitized len.
+    // Each iteration will attempt to dedup "//".
+    while sanitized_uri.len() < len {
+        len = sanitized_uri.len();
+        sanitized_uri = sanitized_uri.replace("//", "/");
```
iulianbarbu, line 76: Does it make sense to use a pattern and replace all the occurrences of "//+" with "/"? This wouldn't involve a while loop to reduce the "//" resulted from previous reduction in steps.

src/mmds/src/lib.rs
```diff
@@ -65,6 +65,20 @@ fn build_response(http_version: Version, status_code: StatusCode, body: Body) ->
     response
 }
 
+// Make the URI a correct JSON pointer value.
+fn sanitize_uri(uri: &str) -> String {
+    let mut sanitized_uri = uri.to_owned();
+    let mut len = u32::MAX as usize;
+    // Loop while the deduping decreases the sanitized len.
```
lauralt, line 72:
```suggestion
    // Loop until the sanitized len is no longer modified in an iteration. 
```

src/mmds/src/lib.rs
lauralt: I was searching that too but it seems rust doesn't offer support for regex in the std library.

sandreim: That sounds like we need to pull in some extra dependency to use regular expressions. This code is not on any hotpath so I do not want to make such trade-offs,

src/mmds/src/lib.rs
```diff
@@ -65,6 +65,20 @@ fn build_response(http_version: Version, status_code: StatusCode, body: Body) ->
     response
 }
 
+// Make the URI a correct JSON pointer value.
+fn sanitize_uri(uri: &str) -> String {
```
lauralt, line 69: Would it make sense to make `uri` a String here and avoid the following line?

src/mmds/src/lib.rs
sandreim: I think my initial comment has same meaning, why do you prefer it to be longer ?

src/mmds/src/lib.rs
sandreim: The `uri` variable is a &str as we retrieve it here. I would still have to call to_string() somewhere else.
```
let uri = request.uri().get_abs_path();
```

src/mmds/src/lib.rs
lauralt: For me the comment wasn't clear about when does the looping end and it seemed that this was its purpose. From the comment I'm thinking of an infinite loop. Anyway, it was more of a nit :D.

src/mmds/src/lib.rs
lauralt: I know, the idea here was to call to_string() when calling `sanitize_uri()`, not inside of it; it was a matter of having one less code line.

sandreim: Done.

src/mmds/src/lib.rs
iulianbarbu: In this case, I am OK with the while loop.

PR #1985, opened 2020-06-30
micro-http: unsigned Content-Length + several checked arithmetic operations

### LINKED ISSUE #1977: micro-http: validate that the content-length is valid
We are not validating that the content-length is valid when parsing requests via the `connection` module.

One problem is that Content-Length valued is defined to be an `i32`. Valid lengths though are > 0. This is problematic when the headers are parsed as part of the connection module because there we are using `Content-Length` to read from the stream.
As no other bound checks are done, this results in a Firecracker crash. 

## Steps to reproduce

Input:
```bash
curl --unix-socket /tmp/firecracker.sock -i \
    -X GET 'http://localhost/actions' \
    -H 'Accept: application/json' \
    -H 'Content-Length: -1' \
```
Output with debug builds:
```
2020-06-25T12:43:58.057850423 [anonymous-instance:ERROR:src/firecracker/src/main.rs:64] Firecracker panicked at 'attempt to add with overflow', src/micro_http/src/connection.rs:287:37
Aborted (core dumped)
```

### TODO

- [ ] Content-Length type should be changed to something that does not accept negative values (i.e. we can work with u32 instead); The same needs to be applied to 
- [ ] Validate the Content-Length (and all other fields are valid) before doing operations that might result in index out of bounds or overflows. For example, if content-length == 0 -> don't try to read the body.
- [ ] Add negative tests for Content-Length values.

aghecenco:
## Reason for This PR

Fixes #1977

## Description of Changes

* Changed `Content-Length` from `i32` to `u32`.
* Added a bunch of fixes for unchecked unsigned arithmetic in `connection.rs`. 
* Added more unit tests too that exercise failure cases for over/underflows in math operations.
* Replaced some `unwrap`s with error propagation, introducing new error variants.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/micro_http/src/common/headers.rs
+++ b/src/micro_http/src/common/headers.rs
@@ -74,7 +74,7 @@ impl Header {
 pub struct Headers {
     /// The `Content-Length` header field tells us how many bytes we need to receive
     /// from the source after the headers.
-    content_length: i32,
+    content_length: u32,
     /// The `Expect` header field is set when the headers contain the entry "Expect: 100-continue".
     /// This means that, per HTTP/1.1 specifications, we must send a response with the status code
     /// 100 after we have received the headers in order to receive the body of the request. This
@@ -134,7 +134,7 @@ impl Headers {
                 }
                 if let Ok(head) = Header::try_from(entry[0].as_bytes()) {
                     match head {
-                        Header::ContentLength => match entry[1].trim().parse::<i32>() {
+                        Header::ContentLength => match entry[1].trim().parse::<u32>() {
                             Ok(content_length) => {
                                 self.content_length = content_length;
                                 Ok(())
@@ -180,7 +180,7 @@ impl Headers {
     }
 
     /// Returns the content length of the body.
-    pub fn content_length(&self) -> i32 {
+    pub fn content_length(&self) -> u32 {
         self.content_length
     }
 
@@ -318,7 +318,7 @@ mod tests {
     use super::*;
 
     impl Headers {
-        pub fn new(content_length: i32, expect: bool, chunked: bool) -> Self {
+        pub fn new(content_length: u32, expect: bool, chunked: bool) -> Self {
             Self {
                 content_length,
                 expect,
@@ -471,6 +471,12 @@ mod tests {
         assert!(header
             .parse_header_line(b"Accept: application/json-patch")
             .is_err());
+
+        // Invalid content length.
+        assert_eq!(
+            header.parse_header_line(b"Content-Length: -1"),
+            Err(RequestError::InvalidHeader)
+        );
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


src/micro_http/src/connection.rs
```diff
@@ -213,16 +242,27 @@ impl<T: Read + Write> HttpConnection<T> {
...
+                        RequestError::HeadersWithoutPendingRequest,
+                    ))?;
                 // The `line_end_index` relative to the whole buffer.
-                let line_end_index = relative_line_end_index + *line_start_index;
+                let line_end_index = relative_line_end_index
+                    .checked_add(*line_start_index)
+                    .ok_or(ConnectionError::ParseError(RequestError::Overflow))?;
 
                 // Get the line slice and parse it.
+                // The slice access is safe becase `line_end_index` is a sum of `line_end_index`
```
ioanachirca, line 264: maybe
```suggestion
                // The slice access is safe because `line_end_index` is a sum of `line_start_index`
```
?

src/micro_http/src/connection.rs
```diff
@@ -265,12 +308,21 @@ impl<T: Read + Write> HttpConnection<T> {
...
+            .ok_or(ConnectionError::ParseError(RequestError::Underflow))?
+            as u32;
+        if self.body_bytes_to_be_read > start_to_end {
             // Append everything that we read to our current incomplete body and update
             // `body_bytes_to_be_read`.
+            // The slice access is safe, otherwise `checked_sub` would have failed.
             self.body_vec
                 .extend_from_slice(&self.buffer[*line_start_index..end_cursor]);
-            self.body_bytes_to_be_read -= end_cursor as i32 - *line_start_index as i32;
+            // Safe to substract directly as the `if` condition prevents underflow.
```
ioanachirca, line 324: nit: 
```suggestion
            // Safe to subtract directly as the `if` condition prevents underflow.
```

src/mmds/src/lib.rs
```diff
@@ -137,6 +137,16 @@ pub fn parse_request(request_bytes: &[u8]) -> Response {
             }
         }
         Err(e) => match e {
+            RequestError::BodyWithoutPendingRequest => build_response(
```
lauralt, line 140: The changes in this file will need a small rebase and after that, we should test in `test_parse_request_bytes_error()` also the newly added error types.

src/micro_http/src/common/headers.rs
```diff
@@ -134,7 +134,7 @@ impl Headers {
                 }
                 if let Ok(head) = Header::try_from(entry[0].as_bytes()) {
                     match head {
-                        Header::ContentLength => match entry[1].trim().parse::<i32>() {
+                        Header::ContentLength => match entry[1].trim().parse::<u32>() {
```
lauralt, line 137: nit: as a double-check, I would add the following:

```rust
        // Invalid content length.
        assert_eq!(
            header.parse_header_line(b"Content-Length: -1"),
            Err(RequestError::InvalidHeader)
        );
```
in `test_parse_header_line()`, even though this is tested in `connection.rs` too.

src/mmds/src/lib.rs
aghecenco: After rebasing and looking more carefully I realized that the overflow checks I'd added were superfluous, so I documented the safety of the operations instead

PR #1988, opened 2020-06-30
virtio net: don't rely on vnet header always being present

aghecenco:
## Reason for This PR

This PR removes a potential panic condition caused by the virtio net device expecting to find a vnet header in every frame. In case a malformed shorter frame shows up on the TX path (large enough buffers make sure that it can't happen on RX), error messages are now logged and a dedicated metric is emitted. With this occasion, several unused device error variants have been removed.

## Description of Changes

* When processing a TX frame, Firecracker no longer `panic`s if it's missing the vnet header. Instead, it propagates an error and increments a metric.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -17,13 +17,13 @@ use dumbo::{EthernetFrame, MacAddr, MAC_ADDR_LEN};
 use libc::EAGAIN;
 use logger::{Metric, METRICS};
 use rate_limiter::{BucketUpdate, RateLimiter, TokenType};
-#[cfg(not(test))]
-use std::io::Read;
 use std::io::Write;
+#[cfg(not(test))]
+use std::io::{self, Read};
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
-use std::{cmp, io, mem, result};
+use std::{cmp, mem, result};
 use utils::eventfd::EventFd;
 use utils::net::Tap;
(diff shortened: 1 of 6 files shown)


src/devices/src/virtio/net/device.rs
```diff
@@ -523,14 +536,20 @@ impl Net {
...
                 &mut self.tx_rate_limiter,
                 &self.tx_frame_buf[..read_count],
                 &mut self.tap,
                 self.guest_mac,
-            ) && !self.rx_deferred_frame
-            {
+            )
+            .unwrap_or_else(|_| {
+                // Log an error, but don't stop processing the malformed frame.
+                error!("VNET header missing in the TX frame.");
```
dianpopa, line 548: Here we are relying on the fact that `write_to_mmds_or_tap` will only ever return error when vnet header is missing. This might be prone to bugs IMO. I d move this logic (increase metric, log an error) in the `write_mmds_or_tap` function after checking the result of `frame_bytes_from_buf `.

dianpopa:
PS: coverage on AMD needs updating

src/devices/src/virtio/net/device.rs
aghecenco: I moved the error message and metric as you suggested. I could have reverted `write_to_mmds_or_tap` to its previous form where it returned `bool` instead of `Result<bool>` but left it with `Result` as it makes easier to control the execution flow (early exit on error).

aghecenco:
@ioanachirca rebased, please re-approve. Thanks!

Issue #1973, opened 2020-06-25
Feature: devtool/build - Podman support


bbros-dev:
We don't have Docker. We likely won't ever.  Docker only support can be worked around, but it would be nicer to have `devtool/build` not care.

In our experience so far the tools are interchangeable so there should be no additional maintenance burden.

Would adding Podman support to `devtool/build` be acceptable?

tnotw:
Are there any cgroups version considerations to doing this?  For example, the Fedora (post-31) defaults are podman/cgroups v2.  For me, podman worked ok, but the tests for the jailer failed (I believe it still requires cgroups v1), so I had to revert to docker/cgroups v1.

bbros-dev:
TL;DR: No not for these changes. I think ;)

> For me, podman worked ok

Interesting? Was that as parameterized and configured in `devtool`, or using your own incantation?

As rootless it should have broken on the `--tty` with the device trees shared.

My understanding is the cgroup v1 vs v2 is about what engine you can run, or vice versa, the engine you run (currently) sets the cgroup ver you use.

When we turn to determining the actual/minimal capabilities each contain is granted then the v1 vs v2 issue will loom. This change doesn't address that.

Frankyl - I hope that shortly we can dump these containers and have Firecracker uVM build Firecracker.
In fact I'd suggest the time investigating the "least permissions required" for these containers could be better spent on porting the build to a couple of uVM's.

Or is there some benefit from the container setup that I am missing?

tnotw:
It was my own incantation.  It was probably just sufficient to run, but nothing more. 

This is the error that led me to believe rootless podman (which appears to require cgroups v2) might not fly: 
```
E   Exception: thread 'main' panicked at 'Jailer error: cpu configurations not found in /proc/mounts', src/jailer/src/main.rs:353:27
```
Which I interpreted as the the jailer confirming the cgroups v1 structures in /proc/mounts,  which don't appear to be there under Fedora/cgroups v2.  Perhaps this is just a Fedora problem,  or some other cgroups configuration issue though.

bbros-dev:
Thanks for the additional info. +1 for uVM builds.

acatangiu:
> I hope that shortly we can dump these containers and have Firecracker uVM build Firecracker.
Or is there some benefit from the container setup that I am missing?

One problem to using uVM builds is the need for virtualization support in dev machines.
Any dev environment running on a VM without nested virtualization support would be unusable.

[...]

Issue #1980, opened 2020-06-27
Security: Test container `run`s use root and --privileged.


bbros-dev:
We currently execute test container `run`s with root access and `--privileged`.
`--privileged` is the same as running with `--cap-add=all`.

Ideally we should just activate the capabilities the root process needs to run the tests, via `--cap-drop=all --cap-add=...`.
Alternatively if 'nearly' all capabilities are needed we can use `--cap-add=all --cap-drop=...`

Does root really need all its capabilities to run the tests?

bbros-dev:
An immediate workaround is to add a confirmation step so that Firecracker contributors/developers are aware of what they are running on their machine.

Agreed?

bbros-dev:
The issue of running the fcuvm/dev container as root with elevated privileges is thorny.  We likely can never run (at least) some parts of the test suite without elevated privileges, doing all the testing and analysis to work out which test items require which capabilities is a substantial effort and (in our opinion) likely not the best use of dev cycles right now. 
Example: a higher priority is writing, documenting and working out the equivalent jailer details.

There are two workarounds readily available to anyone who comes along this path and wants to run the test suite, in order or ease:

Workaround 1:
Only run the test suite on external CI/CD providers.  These generally run as root, and your systems aren't touched by any libraries the test suite pulls in and runs.

Workaround 2: 
Only run the test suite as rootless, either Docker (hard to configure) or Podman (no setup required). 
The Podman path is provided by PR #1973 and the instructions therein.

Closing as fixed for us, and for anyone else there are two workarounds at hand.

PR #1981, opened 2020-06-28
Permission: Run tests as root with all cap. on

bbros-dev:
## Reason for This PR

Closes issue #1980.

## Description of Changes

While Docker is often used with root access to all capabilities,
this can be a surprise to users coming from podman.

We don't immediately know the minimal set of capabilities the
root user needs to run the Firecracker test suite.

Until then the best we can do is:
1. Tell the user what is about to happen.
2. Get user agreement.
3. Log what they have agreed to.

The logging is generally useful. However it is primarily intended
to inform users who may have inadvertently done something like:

   $(yes | ./tools/devtool test)

This is a workaround until a we identify the minimal set of root
capabilities needed to run the Firecracker test suite.

Signed-off-by: Begley Brothers Inc <begleybrothers@gmail.com>

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

- [x] All commits in this PR are signed (`git commit -s`).
- [x] The reason for this PR is clearly provided (issue no. or explanation).
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] Any newly added `unsafe` code is properly documented.
- ~~[ ] Any API changes are reflected in `firecracker/swagger.yaml`.~~
- [x] Any user-facing changes are mentioned in `CHANGELOG.md`.

PR #1991, opened 2020-07-01
OCI engine agnostic devtool (Issue #1973)

bbros-dev:
## Reason for This PR

Closes issue #1973 
Workaround issue #1980

## Description of Changes

OCI engine agnostic devtool

Closes issue #1973.
Workaround issue #1980.

These changes allow `tools/devtool build` and `tools/devtool test`
to Just-Work(TM) when a user has Podman and not Docker installed.

They provide a workaround for issue 1980 in so far as a user can
build and test under a typical user account - albeit with that
account has elevated privileges.
AFAIK this is not the same as building as the root with elevated
privileges.

Moving to rootless builds exposes several issues that are
odscured by building as root with `--cap-add=all`.
The first is arises from handing over the hosts whole device tree
to the container via `--volume dev:dev`.
Unprivleged, this generates a permissions error in the presense
of the `--tty` option.
Those issues have been resolved in this commit.

- Respects a DEBUG=1|y|yes|true flag that turns on xtrace for
  the script and adds RUST_LOG=rls=trace,cargo=trace for the
  Rust build.
- OPT_UNATTENDED is still respected, and users can also
  use devtool in pipes, such as `yes|tools/devtool test`
- Add `oci_engine_args_init` to accomodate the fact that Docker
  handles user namespacing at the daemon level and Podman
  uses an option.
- `ensure_docker` -> `ensure_oci_engine`. This now returns the OCI
  engine path. Where both Podman and Docker are installed,
  Podman is preferred. Simply because 1) the Podman default
  is to run rootless, 2) for users without either, Podman can be
  installed without elevated rights and need to configure a
  daemon for user name space functionality.
- `ensure_devctr` is refactored to consume the OCI engine
  returned by `ensure_oci_engine`.
- `ensure_build_dir` permissions change fallsback to use
   `cmd_fix_perms` in cases where root owns the folder.
- `cmd_fix_perms` allows us to do away with running a
  container to get elevated permissions to make the
  change - the dirty hack is gone.
- `cmd_fix_perms` now respects OPT_UNATTENDED,
  and `check_superuser` before invoking sudo for the
  recursive `chmod`. After sudo is used it is revoked
  in `uncheck_superuser` - here check has its "checkbox"
  meaning and not "inspect" meaning.
- `get_user_confirmation` now allows piped confirmation
  and respects OPT_UNATTENDED.
- `run_devctr` selects the OCI engine using the logic
  in `ensure_oci_engine`. Podman namespacing option
  is added in `oci_engine_args_init`. The `--tty` option is
  removed without loss of functionality, and this allows
  builds be done rootless.  Test runs still require root.
  The `--volume dev:dev` appears unneccessary to
  complete a build. The host exposing its device tree
  to the guest seems excessive.
- `cmd_build`: `cargo_args` fixes `--target ...` insertion - this broke
   my builds but I can't find any other reports. Nonetheless, it seems
  the right way to add space delimited items to a shell array. A
  `--ulimit 4096:4096` options is added. Withoput this my builds
  broke and this options was used elsewhere so it appears to have
  been an oversight here.
- `check_binary` is a new function. This is needed because it is so
  tricky to pass a bash function to sudo. Instead we set a flag and
  use that to invoke sudo on a file (the OCI engine).
- `check_superuser` sets the `run_with_sudo` flag, and gets the
  elevated root access if not already in hand.
- `uncheck_sudo` revokes the elevated sudo access.
- `run_as_root` does that.
- `main` now checks for specific known commands and so is limited
  in what it will run.

Together the 3 additional functions introduced allow us to remove
the workaround of changiung permissions by running an OCI engine
(Docker).
Changes, other than the above, are cosmetic.

Signed-off-by: Begley Brothers Inc <begleybrothers@gmail.com>

~~- [] This functionality can be added in [`rust-vmm`](https://github.com/rust-vmm).~~

## License Acceptance

## PR Checklist


gc-plp:
> Where both Podman and Docker are installed,
> Podman is preferred. Simply because 1) the Podman default
> is to run rootless, 2) for users without either, Podman can be
> installed without elevated rights and need to configure a
> daemon for user name space functionality

Adding Podman as the default option will be confusing for people that have both runtimes installed, because it changes the way of working with `devtool`, for example when being asked for the root password.

The PR should conserve the current command parameters so that they're passed on as before and add Podman as an alternative to the default engine, which should be `docker`.

bbros-dev:
@gc-plp, thanks for taking time to look over this. Appreciate the feedback.

> Podman as the default option will be confusing for people that have both runtimes installed

If they have Podman installed they likely already know it runs rootless by default.  
The fact that Docker *silently* runs a process as root with elevated privileges is *precisely* the reason it should be sent to the back of the queue ;)
That a Podman workflow prompts the user before going ahead with elevated access is the intent of the changes - a feature not a bug. Unless of course they did `sudo devtool build`, but most Podman users seem to know one point for having Podman is that things run in user space, and prompt you when they don't.

You are correct that users should not be surprised. However, doing something like taking the whole of the hosts device tree and handing it over to a 3rd party container, is ~~bad~~surprising enough. Doing it with root access is still more surprising. Doing it with `--cap-add=all` is yet more surprising. Doing all this behind their back is not the same as 'not surprising them'.  

Surprise or disappointment really depends on what your expectations were.  
See PR #1981 for a workaround aimed at informing users of what is going on - hopefully that leads to informed expectations.

> The PR should conserve the current command parameters so that they're passed on as before.

Agreed this is an alternate implementation.
However it doesn't address the issues identified with the Docker setup. 

Are the `build` or `test` broken when run with Docker?

gc-plp:
> The fact that Docker silently runs a process as root

This does not necessarily need to happen. In order to launch `devtool`, you either [add your user to the `docker` group](https://github.com/firecracker-microvm/firecracker/blob/master/docs/getting-started.md#appendix-b-setting-up-docker) or run the command as `sudo`. Both options are done voluntarily.

> handing it over to a 3rd party container

Are you referring to the fact that the container is built by someone else?

This container is actually built, maintained and updated by the Firecracker team, with the `Dockerfile` available [here](https://github.com/firecracker-microvm/firecracker/blob/master/tools/devctr).

> Are the `build` or `test` broken when run with Docker?

I just tested it and `devtool build` works fine, but `devtool test` needs to be launched with `sudo` rights, because the jailer creates cgroups when running tests.

`devtool shell` does not seem to launch a shell.

Adding Podman is a good idea, but we should be cautious when introducing changes in how people work with the updated components.

I would also talk about being able to run Podman as an alternative to Docker in the quickstart guide.

bbros-dev:
> Are you referring to the fact that the container is built by someone else?

Correct.

> `devtool test` needs to be launched with `sudo` rights

Correct. As indicated, this PR does not address the minimal set of capabilities needed to run the test suite.

> `devtool shell` does not seem to launch a shell.

My apologies. Update forthcoming.

bbros-dev:
I believe I made the changes requested.

>... we should be cautious when introducing changes in how people work with the updated components.

Agreed.  Likewise we shouldn't assume users come to Firecracker with the existing team's setup and habits.  I've added comments to the docs addressing users with Docker configured as rootless. 

> I would also talk about being able to run Podman as an alternative to Docker in the quickstart guide.

I've added some more context.

bbros-dev:
>... people onboarding to Firecracker that may stumble upon devtool issues

I don't see any issues other than for users who run their containers as non-root encountering scripts that assumed root was the default.    
Note **this includes Docker users** who have configured their system to run containers as non-root, and who may have organizational policies to prohibit running containers as root.  
Note in these organizations arranging for a security review for a 3rd party container to be run as root can take months. In my experience this container would likely not pass - there is no documentation on why each of the privileges is required.  The review would have to reverse engineer each test case, work out the related privilege to see what their security policy was and then investigate the libraries the test case code relied on.

> it would be good if the default engine is Docker, with Podman as a fallback for advanced users.

Sorry, we don't agree. We can't commit code that promotes a runtime whose default setup we consider a security risk over a runtime whose default is to run as the users privileges. 
To setup a sane and secure Docker you have to be a relatively advanced user:
- https://docs.docker.com/engine/security/rootless/

We can't spend more time on this.  Feel free to amend or reject as you see fit - we don't take these things personally ;)

bbros-dev:
Closing as 'fixed for us`, and we believe the following is workable for anyone else.

Anyone else who want this changed setup can apply the patch then:

```bash
git update-index --assume-unchanged tools/devtool
```

PR #1238, opened 2019-08-31
Remove pnet dependency

shakram02:
## Reason for This PR

Fixes #612 

## Description of Changes

1. parse UDP datagrams
2. replace pnet::packet with dumbo's sutff
3. move MacAddr from `net_util` to dumbo
4. use UdpSocket in `test_write`, `test_read` in `tap.rs` ARP reply construction was needed to make this work
5. Use host-IPs instead of hardcoded ones since the packets will be routed
## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria. Where there are two options, keep one.]`  
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [ ] All commits in this PR are signed (`git commit -s`).
- [x] Either this PR is linked to an issue, or, the reason for this PR is
      clearly provided. 
- [x] The description of changes is clear and encompassing.
- [x] Either no docs need to be updated as part of this PR, or, the required
      doc changes are included in this PR. Docs in scope are all `*.md` files
      located either in the repository root, or in the `docs/` directory.
- [x] Either no code has been touched, or, code-level documentation for touched
      code is included in this PR.
- [x] Either no API changes are included in this PR, or, the API changes are
      reflected in `firecracker/swagger.yaml`.
- [x] Either the changes in this PR have no user impact, or, the changes in
      this PR have user impact and have been added to the `CHANGELOG.md` file.

PR #2016, opened 2020-07-07
Test tap by simulating traffic

serban300:
## Reason for This PR

Test tap by simulating frame receival/transmission. We will reuse this code later in order to add unit tests for the net device. Those tests will cover the entire net flow better.

## Description of Changes

Test tap by simulating frame receival/transmission. We will reuse this code later in order to add unit tests for the net device. Those tests will cover the entire net flow better.

## License Acceptance

## PR Checklist


src/utils/Cargo.toml
```diff
@@ -10,3 +10,7 @@ libc = ">=0.2.39"
 vmm-sys-util = ">=0.2.1"
 
 net_gen = { path = "../net_gen" }
+
+[dev-dependencies]
```
andreeaflorescu, line 14: This was the previous way of testing the Tap. We removed the dependencies on pnet on purpose. See PR: https://github.com/firecracker-microvm/firecracker/pull/1238

serban300: Ok, thanks ! I will try to find a solution in order to avoid the pnet dependency.

serban300: The problem here is the fact that we need the frame injection logic both in the unit tests for the `Tap` and in the unit tests for the `virtio net` device. But the logic defnied inside `mod tests` can't be exported across crates. So we would have to duplicate this logic. If we use pnet, there are only a couple of lines of code, so duplicating them wouldn't be a big issue. If we don't use pnet there would be more than 100 lines of code that we would need to duplicate and this would be more problematic.

For the moment the `Tap` struct is only used inside the net emulation code. So I propose moving the tap inside the devices crate. This way the entire crate will be able to access the logic defined inside the `Tap` unit tests. @firecracker-microvm/compute-capsule  does anyone have any concern related to this approach ?

andreeaflorescu: I would also add that the current `Tap` implementation is very virtio-net specific at the moment as we inject the vnet information in it. It would need some changes to be truly library worth it.

serban300:
I removed the pnet dependency. The PR is ready for review.

src/devices/src/virtio/net/tap.rs
```diff
@@ -35,6 +36,40 @@ ioctl_iow_nr!(TUNSETIFF, TUNTAP, 202, ::std::os::raw::c_int);
...
+
+    fn flags(mut self, flags: i16) -> Self {
+        // Since we don't call as_mut on the same union field more than once, this block is safe.
+        let ifru_flags = unsafe { self.0.ifr_ifru.ifru_flags.as_mut() };
+        *ifru_flags = flags;
+
+        self
+    }
+
+    fn execute<F: AsRawFd>(mut self, socket: &F, ioctl: u64) -> Result<ifreq> {
```
iulianbarbu, line 62: Would be better if we have this on Tap? It is not relevant to `ifreq` building.

serban300: I was a bit unsure here. It's true `execute` is not part of the `ifreq` building process. But it's not really related to the tap either. But since it's more related to the `ifreq` than to the `tap`, I defined it here. Maybe it's a naming issue ? How about renaming it `build_and_execute` ? 

Another option would be to have a new class `IfReq` that holds a `ifreq` inside. And define `execute` there and `build` on `IfReqBuilder`. I think it would be more fair, but it would add some more code and right now we don't really need it.

WDYT ?

iulianbarbu: It is a bit uncertain for me why we would need to execute anything, while building an `ifreq`. Then, why executing  means doing an ioctl on the socket. I was seeing it more like a method used for configuring a network device [1]. I would rather rename the method to something like `configure_netif`, which will end building an `ifreq`, or move it on the Tap as `configure`. WDYT?

[1] netdevice documentation: "Linux supports some standard ioctls to configure network devices. They can be used on any socket's file descriptor regardless of the family or type. Most of them pass an ifreq structure [...]".

iulianbarbu: Talked with @serban300 offline. `execute` will have as an side effect issuing an `ifreq`, which stands for interface request. Resolving this.

PR #2019, opened 2020-07-10
Gracefully handle cmd line logger errors

lauralt:
## Reason for This PR

If the logger initialization from command line
arguments failed, the process aborted with a coredump.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/firecracker/src/main.rs
+++ b/src/firecracker/src/main.rs
@@ -191,7 +191,8 @@ fn main() {
         // It's safe to unwrap here because the field's been provided with a default value.
         let level = arguments.value_as_string("level").unwrap();
         let logger_level = LoggerLevel::from_string(level).unwrap_or_else(|err| {
-            panic!("Invalid value for logger level: {}. Possible values: [Error, Warning, Info, Debug]", err);
+            error!("Invalid value for logger level: {}. Possible values: [Error, Warning, Info, Debug]", err);
+            process::exit(i32::from(vmm::FC_EXIT_CODE_GENERIC_ERROR));
         });
         let show_level = arguments.value_as_bool("show-level").unwrap_or(false);
         let show_log_origin = arguments.value_as_bool("show-log-origin").unwrap_or(false);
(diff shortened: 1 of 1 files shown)


src/firecracker/src/main.rs
```diff
@@ -191,7 +191,8 @@ fn main() {
         // It's safe to unwrap here because the field's been provided with a default value.
         let level = arguments.value_as_string("level").unwrap();
         let logger_level = LoggerLevel::from_string(level).unwrap_or_else(|err| {
-            panic!("Invalid value for logger level: {}. Possible values: [Error, Warning, Info, Debug]", err);
+            error!("Invalid value for logger level: {}. Possible values: [Error, Warning, Info, Debug]", err);
+            process::exit(i32::from(vmm::FC_EXIT_CODE_GENERIC_ERROR));
```
ioanachirca, line 195: Would `FC_EXIT_CODE_ARG_PARSING` be more explicit?

serban300, line 195: nit: I see we already use this in 2 places. If we plan to make extended use of it, we can define a macro, just like `panic!`. Something like `fc_exit!(exit_code, message, args)`.

lauralt: Hmm, I'm not so sure, when I added this code, I was thinking that it makes sense only for errors returned from the `ArgParser` processing. This `from_string` validation isn't necessary related to the actual argument parsing.

lauralt: I think it would affect readability pretty much (I tried something similar here: https://github.com/firecracker-microvm/firecracker/pull/1945 and the result wasn't so nice) and I also expect the number of code lines won't decrease. I would leave it as is if you don't have a strong preference against it.

serban300: Ok. I don't have a strong preference.

ioanachirca: Alright. We can use the generic code since the error is printed.

PR #2041, opened 2020-07-23
[RFC] tools: add support for environment shielding

iulianbarbu:
## Reason for This PR

When running performance tests, one might want to run them in a shielded and isolated environment, where there is as little as possible system tasks interference. It can be done with external tools, but this would imply adding specific dependencies on the host. To avoid this, but to also offer the bare minimum for setting up a shielded and isolated environment, `devtool` was augmented with the needed support.

## Description of Changes

Added `devtool test` flag: -s|--shield {NUMA_NODE_ID};{CPUS_COUNT}

The flag computes the available CPUs and memory specified through the pattern. It will bind the tests docker container
to those. It also isolates the docker container from outside tasks interference, by trying to move all exisiting running tasks
to a separate cpuset, which will use other resources compared to the shield. However, if there is only a single NUMA node online on the platform, the system cpuset can not be created and the tests will run as if there is no shield present.

## License Acceptance

## PR Checklist

~~[] Any newly added `unsafe` code is properly documented.~~
~~[] Any API changes are reflected in `firecracker/swagger.yaml`.~~
~~[] All added/changed functionality is tested.~~

### DIFF
--- a/tools/devtool
+++ b/tools/devtool
@@ -57,6 +57,12 @@
 #
 #
 # TODO:
+#   - Transform `devtool` with dependency injection in mind. `devtool` commands
+#     can be seen as individual entries/dependencies for `devtool`, which is a
+#     thin layer over them. Keep in mind that some commands, and thus `devtool`,
+#     may require different host dependencies. The ideal objective would be to have
+#     zero host dependencies and offload dependencies to the development docker container,
+#     where is possible.
 #   - Cache test binaries, preserving them across `./devtool test` invocations.
 #     At the moment, Firecracker is rebuilt everytime a test is run.
 #   - List tests by parsing the `pytest --collect-only` output.
@@ -118,36 +124,60 @@ CTR_MICROVM_IMAGES_DIR="$CTR_FC_BUILD_DIR/img"
 # These options are not command-specific, so we store them as global vars
 OPT_UNATTENDED=false
 
+# Shield relevant mountpoints.
+SYSFS_MOUNTPOINT=$(cat /proc/mounts | grep "^sysfs \/.\+ sysfs .*$" | cut -d' ' -f2 | head -n 1)
+CPUSET_MOUNTPOINT=$(cat /proc/mounts | grep -P "^[^ ]+ (/.+) (?:cpuset |cgroup (?:[^ ]*,)?cpuset[, ])"\
+                                               | cut -d' ' -f2 | head -n 1)
+
+# Shield relevant exit codes.
+ESKIP=12
+ESINGLENUMANODE=13
 
 # Send a decorated message to stdout, followed by a new line
 #
 say() {
-    [ -t 1 ] && [ -n "$TERM" ] \
-        && echo "$(tput setaf 2)[$MY_NAME]$(tput sgr0) $*" \
-        || echo "[$MY_NAME] $*"
+    run_devctr \
+    --user "$(id -u):$(id -g)" \
+    --workdir "$CTR_FC_ROOT_DIR" \
+    -- \
+    python3 -c "
+import tools.log_utils as lu
+lu.say_info(\"$1\")"
 }
 
 # Send a decorated message to stdout, without a trailing new line
 #
 say_noln() {
-    [ -t 1 ] && [ -n "$TERM" ] \
-        && echo -n "$(tput setaf 2)[$MY_NAME]$(tput sgr0) $*" \
-        || echo "[$MY_NAME] $*"
+    run_devctr \
+    --user "$(id -u):$(id -g)" \
+    --workdir "$CTR_FC_ROOT_DIR" \
+    -- \
+    python3 -c "
+import tools.log_utils as lu
+lu.say_info(\"$1\", False)"
 }
 
 # Send a text message to stderr
 #
 say_err() {
-    [ -t 2 ] && [ -n "$TERM" ] \
-        && echo -e "$(tput setaf 1)[$MY_NAME] $*$(tput sgr0)" 1>&2 \
-        || echo -e "[$MY_NAME] $*" 1>&2
+    run_devctr \
+    --user "$(id -u):$(id -g)" \
+    --workdir "$CTR_FC_ROOT_DIR" \
+    -- \
+    python3 -c "
+import tools.log_utils as lu
+lu.say_err(\"$1\")"
 }
 
 # Send a warning-highlighted text to stdout
 say_warn() {
-    [ -t 1 ] && [ -n "$TERM" ] \
-        && echo "$(tput setaf 3)[$MY_NAME] $*$(tput sgr0)" \
-        || echo "[$MY_NAME] $*"
+    run_devctr \
+    --user "$(id -u):$(id -g)" \
+    --workdir "$CTR_FC_ROOT_DIR" \
+    -- \
+    python3 -c "
+import tools.log_utils as lu
+lu.say_warn(\"$1\")"
 }
 
 # Exit with an error message and (optional) code
@@ -159,7 +189,8 @@ die() {
         code="$2"
         shift 2
     }
-    say_err "$@"
+
+    [ ! -z "$@" ] && say_err "$@"
     exit $code
 }
 
@@ -168,6 +199,69 @@ die() {
 ok_or_die() {
     code=$?
     [[ $code -eq 0 ]] || die -c $code "$@"
+}
+
+validate_shield() {
+    local shield=$1
+
+    run_devctr \
+    --user "$(id -u):$(id -g)" \
+    --workdir "$CTR_FC_ROOT_DIR" \
+    --mount type=bind,source="$SYSFS_MOUNTPOINT",target="/host_sys" \
+    --mount type=bind,source="$CPUSET_MOUNTPOINT",target="/host_cpuset" \
+    -- \
+    python3 -c "
+import tools.shield_utils as su
+su.validate(\"$shield\")"
+    ok_or_die
+}
+
+shield_environment() {
+    local numa_node_id=$1
+    local cpus_count=$2
+    if [ ! -z $numa_node_id ] && [ ! -z $cpus_count ]; then
+        # Get available cpuset cpus list.
+        local cpuset_cpus=$(run_devctr \
+        --user "$(id -u):$(id -g)" \
+        --workdir "$CTR_FC_ROOT_DIR" \
+        --mount type=bind,source="$SYSFS_MOUNTPOINT",target="/host_sys" \
+        --mount type=bind,source="$CPUSET_MOUNTPOINT",target="/host_cpuset" \
+        -- \
+        python3 -c "import tools.shield_utils as su
+print(su.ensure_environment($numa_node_id, $cpus_count, return_to_stdout=True))")
+        ok_or_die "Failed to ensure the shielded environment ($?)."
+
+        # Create system cpuset.
+        run_devctr \
+        --privileged \
+        --workdir "$CTR_FC_ROOT_DIR" \
+        --mount type=bind,source="$SYSFS_MOUNTPOINT",target="/host_sys" \
+        --mount type=bind,source="$CPUSET_MOUNTPOINT",target="/host_cpuset" \
+        -- \
+        python3 -c "import tools.shield_utils as su
+su.create_system_cpuset($numa_node_id)" || ENVIRONMENT_SHIELDED=$? > /dev/null
+
+        # Verify exit codes.
+        [[ -z $ENVIRONMENT_SHIELDED ]] && ENVIRONMENT_SHIELDED=0
+        [[ $ENVIRONMENT_SHIELDED -eq $ESINGLENUMANODE ]] || [[ $ENVIRONMENT_SHIELDED -eq $ESKIP ]] || [[ $ENVIRONMENT_SHIELDED -eq 0 ]]
+        ok_or_die "Failed to create the system cpuset ($ENVIRONMENT_SHIELDED)."
+
+        retval=$cpuset_cpus
+    fi
+}
+
+cleanup_shield() {
+    # Teardown the created 'system' cpuset.
+    [[ -z $ENVIRONMENT_SHIELDED ]] || \
+    [[ $ENVIRONMENT_SHIELDED -eq $ESINGLENUMANODE ]] || \
+    run_devctr \
+        --user "$(id -u):$(id -g)" \
+        --workdir "$CTR_FC_ROOT_DIR" \
+        --mount type=bind,source="$SYSFS_MOUNTPOINT",target="/host_sys" \
+        --mount type=bind,source="$CPUSET_MOUNTPOINT",target="/host_cpuset" \
+        -- \
+        python3 -c "import tools.shield_utils as su
+su.teardown_system_cpuset()" || die
 }
 
 # Check if Docker is available and exit if it's not.
@@ -194,7 +288,6 @@ ensure_docker() {
 # available on this system.
 #
 ensure_devctr() {
-
     # We depend on having Docker present.
     ensure_docker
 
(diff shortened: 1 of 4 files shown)


tools/devtool
```diff
@@ -81,6 +87,10 @@ MY_NAME="Firecracker $(basename "$0")"
 # Full path to the Firecracker tools dir on the host.
 FC_TOOLS_DIR=$(cd "$(dirname "$0")" && pwd)
 
+# Relative path to the Firecracker tools dir on the host.
+REL_FC_TOOLS_DIR=$(dirname "$0")
+MODULE_PY_DIR=`echo $REL_FC_TOOLS_DIR | sed 's/\//\./g'`
```
gc-plp, line 92: This method of calling Python code seems a bit hacky and adds obscurity to the code.

Instead of using it like this, we could still use plain bash, without needing Python3 on the host, and add multiple entry points in the container. For example, to wrap around say, we can add a file in the container with the following content:

```
# Assuming that this file is located in tools/ctr_utils/say
import tools.shield_utils as su

import sys

if __name__ == "__main__":
    # Pass the argv list to the target function
    su.say(*sys.argv)
```

And the bash entry point would just look like this:
```
say() {
    run_devctr -- tools/ctr_utils/say $1
}
```

If we need to handle return codes from the launched application, we should be able to pick those up.

tools/shield_utils.py
```diff
@@ -0,0 +1,368 @@
...
+        for line in fp:
+            res = mount_point_regex.search(line)
+            if res:
+                path = res.group(1)
+                break
+
+    return path
+
+
+# Parse list formats for cpuset and mems.
```
gc-plp, line 71: Instead of adding function descriptions as comments, you should enclose the comments in pydoc comments:

```
This is a python block
```

tools/shield_utils.py
```diff
@@ -0,0 +1,368 @@
...
+import random
+import string
+# Add the file directory to PATH to be able to import other modules from within
+# the same directory.
+sys.path.append(os.path.dirname(__file__))
+import log_utils as log
+
+
+# Exit codes.
+EINVPATTERN = 1
```
gc-plp, line 38: You can these constants in a class, that's decorated with `enum`: https://docs.python.org/3/library/enum.html

tools/shield_utils.py
```diff
@@ -0,0 +1,368 @@
...
+def _parse_list_format(content):
+    content = content.strip()
+    arr = set()
+    ranges=content.split(",")
+    for group in ranges:
+        m = re.search(RANGE_PATTERN, group)
+        # group is a singular value.
+        if m is None:
+            arr.add(int(group))
+        else: # group is a tuple.
```
gc-plp, line 83: Are we sure that it can't be something else than a tuple?

tools/shield_utils.py
```diff
@@ -0,0 +1,368 @@
...
+        if raw:
+            return fp.readline()
+        return _parse_list_format(fp.readline())
+
+
+def _is_memory_node_available(numa_node_id):
+    cpuset_path = _locate_mount_point(CPUSET_MOUNT_PATTERN)
+    (_, dirnames, _) = next(os.walk(cpuset_path))
+    for dirname in dirnames:
+        if dirname != "docker":
```
gc-plp, line 122: Removes an indentation block:
```suggestion
        if dirname == "docker":
            continue
```

tools/shield_utils.py
```diff
@@ -0,0 +1,368 @@
...
+            with open(cpuset_path + "/" + dirname + "/cpuset.mems") as fp:
```
gc-plp, line 123: An indentation block can also be removed here by opening the file directly: `fp = open(cpuset_path + "/" + dirname + "/cpuset.mems")`. Since the function is quite small, the file will be closed upon exiting the scope.

tools/shield_utils.py
```diff
@@ -0,0 +1,368 @@
...
+        else:
+            log.say_warn(str(err))
+
+    dir_exists = os.path.isdir(cpuset_system_path)
+    if dir_exists:
+        log.die("'system' cpuset could not be removed.", ETEARDOWNSYS)
+
+    log.say("'system' cpuset was succesfully removed.")
+
+if __name__ == "__main__":
```
gc-plp, line 364: Not really necessary if we don't implement an implicit behavior.

tools/devtool
```diff
@@ -575,11 +603,21 @@ cmd_strip() {
...
 
     # Parse any command line args.
     while [ $# -gt 0 ]; do
         case "$1" in
             "-h"|"--help")      { cmd_help; exit 1; } ;;
+            "-s"|"--shield")
+                shift
+                validate_shield $1
+                ok_or_die
+                [[ "$1" =~ ^(0|[1-9][0-9]*)\;([1-9][0-9]*)$ ]] && \
```
gc-plp, line 617: I think that this block needs a bit of more commenting :)

tools/shield_utils.py
```diff
@@ -0,0 +1,368 @@
...
+
+import os
+import re
+import io
+import sys
+import random
+import string
+# Add the file directory to PATH to be able to import other modules from within
+# the same directory.
+sys.path.append(os.path.dirname(__file__))
```
gc-plp, line 33: Is this really needed?

tools/shield_utils.py
```diff
@@ -0,0 +1,368 @@
...
+                path = res.group(1)
+                break
+
+    return path
+
+
+# Parse list formats for cpuset and mems.
+# See: https://man7.org/linux/man-pages/man7/cpuset.7.html.
+#
+def _parse_list_format(content):
```
gc-plp, line 74: The return value of this function seems to be used in a few different ways.

Could the return value be encapsulated in an object that would make it easier to understand how it's used?

tools/shield_utils.py
```diff
@@ -0,0 +1,368 @@
...
+    # use by 'docker' cpuset. The available cpus are computed based on
+    # cpusets different than `root` and `docker` usage of slices within
+    # the specified `numa_node_id`.
+    cpulist = _get_cpulist(numa_node_id)
+    non_avail_cpus = _get_non_available_cpus()
+    avail_cpus = []
+    # For every NUMA node core, check if it is available. Add it to the
+    # avail_cpus list if it is not present inside the non_avail_cpus list.
+    for cpu_group in cpulist:
+        if type(cpu_group) is tuple:
```
gc-plp, line 160:
```suggestion
        if type(cpu_group) is not tuple:
            continue
```

tools/devtool
iulianbarbu: This might work. An issue I see with this approach is that we may need to access host resources from within the docker container (filesystem and devices), under the R/W permissions. For the shield case, we would need to expose sysfs cgroups, parse some files and modify them. I will experiment with a similar setup and see if we can get rid of python3 host dependency.

tools/log_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def _say(msg, msg_color, tag, tag_color, file, newline=True):
+    if not file.isatty():
+        return
+    if msg_color is None:
+        msg_color = ""
+    if tag_color is None:
+        tag_color = ""
+    print(tag_color + tag + Colors.ENDC.value + " " + msg_color + msg
```
gc-plp, line 37:
```suggestion
    print(f"{tag_color}{tag}{Colors.ENDC.value} {msg_color}{msg}{msg_color}{Colors.ENDC.value}", file=file, end="\n" if newline else "")
```

iulianbarbu: Done.

tools/shield_utils.py
```diff
@@ -0,0 +1,427 @@
...
+        if dirname == "docker":
+            continue
+
+        with open(root + "/" + dirname + "/cpuset.cpus") as fp:
+            non_avail_cpus |= set(_parse_list_format(fp.readline()))
+
+    return list(non_avail_cpus)
+
+
+def _get_available_cpus(numa_node_id):
```
serban300, line 209: This logic is quite complicated, but it's also not bulletproof. A cpu may be available when the test starts, and then it can be added to another cpuset, which would affect the results of the test. It would be much easier and reliable if we used the `cpuset.cpu_exclusive` flag in order to mark the cpus that we use as exclusive to our cpuset. WDYT ?

tools/shield_utils.py
```diff
@@ -0,0 +1,427 @@
...
+        log.die("Can not provide {} CPUs for NUMA node {}.".
+                format(cpus_count, numa_node_id), ExitCode.ENOAVAILCPU)
+
+    if return_to_stdout:
+        return ','.join(map(str, avail_cpus))
+
+    return avail_cpus
+
+
+def create_system_cpuset(excluded_numa_node_id):
```
serban300, line 322: This is very similar to what `cset shield` does. Could we try to use it instead ?

tools/devtool
```diff
@@ -170,6 +193,75 @@ ok_or_die() {
...
+        cpuset_mountpoint=$(cat /proc/mounts | grep -P "^[^ ]+ (/.+) (?:cpuset |cgroup (?:[^ ]*,)?cpuset[, ])" | cut -d' ' -f2 | head -n 1)
+
+        # Get available cpuset cpus list.
+        CPUSET_CPUS=$(run_devctr \
+        --user "$(id -u):$(id -g)" \
+        --workdir "$CTR_FC_ROOT_DIR" \
+        --mount type=bind,source="$sysfs_mountpoint",target="/host_sys" \
+        --mount type=bind,source="$cpuset_mountpoint",target="/host_cpuset" \
+        -- \
+        python3 -c "import tools.shield_utils as su
```
serban300, line 228: My feeling is that writing the shielding logic in python introduces some complicated interactions between bash and python. Not just here, but in multiple places. I understand that python is easier to use in some situations, but I think we  could simplify that logic a lot sunch that python wouldn't make a big difference, or maybe just use `cset shield`. In this case maybe we could consider rewriting the shielding logic in bash.

tools/shield_utils.py
```diff
@@ -0,0 +1,427 @@
...
+    avail_cpus = []
+
+    # For every NUMA node core, check if it is available. Add it to the
+    # avail_cpus list if it is not present inside the non_avail_cpus list.
+    for cpu in cpulist:
+        cpu_is_avail = True
+        for non_avail_cpu in non_avail_cpus:
+            if cpu == non_avail_cpu:
+                cpu_is_avail = False
+                break
```
serban300, line 231: `non_avail_cpu` is a set. Maybe there's a predefined function that checks if a set contains a certain element.

tools/shield_utils.py
```diff
@@ -0,0 +1,427 @@
...
+                        containers.
+    """
+    cpulist = _get_cpulist(numa_node_id)
+    non_avail_cpus = _get_non_available_cpus()
+    avail_cpus = []
+
+    # For every NUMA node core, check if it is available. Add it to the
+    # avail_cpus list if it is not present inside the non_avail_cpus list.
+    for cpu in cpulist:
+        cpu_is_avail = True
```
serban300, line 227: Maybe there's something like `iter.filter()` in rust.

tools/devtool
iulianbarbu: We can not simply use `cset shield` because of few considerations that I thought are somehow blocking the progress. Running `cset shield` on host means bringing undesired host dependencies on python and `cset`.  Running `cset shield` within docker container is posible only if we patch the `cset shield` to actually look within custom directories, where the shared host sysfs and cgroup cpuset are, other than the mountpoints defined in `/proc/mounts` for the docker container counterpart of sysfs and cgroup cpuset. The implementation details can be later changed if we decide that we must simplify by other means and adjust to `cset shield`. Please let me know what are your thoughts on this.

tools/devtool
```diff
@@ -575,11 +674,24 @@ cmd_strip() {
...
             "-h"|"--help")      { cmd_help; exit 1; } ;;
+            "-s"|"--shield")
+                shift
+                validate_shield $1
+
+                # Pattern match on the shield: {TESTS_NUMA_NODE_ID};{CPUS_COUNT}.
+                [[ "$1" =~ ^(0|[1-9][0-9]*)\;([1-9][0-9]*)$ ]] && \
+                TESTS_NUMA_NODE_ID=${BASH_REMATCH[1]} && \
+                CPUS_COUNT=${BASH_REMATCH[2]}
+                ;;
```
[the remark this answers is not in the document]
iulianbarbu: I am not sure having custom flags or parsing flags is a better approach in this case. It will provide clearer arguments but it will be unusual, in my opinion. The plan here is to provide a digestible pattern but also to provide ways to extend it for scenarios where we must set shields cross NUMA nodes. Working with the provided pattern, we can easily extend it for shields cross numa nodes like so: `{TESTS_NUMA_NODE_ID1,TESTS_NUMA_NODE_ID2,...};{CPUS_COUNT1,CPUS_COUNT2,....}`. This is somehow similar to the jailer `--cgroups` param pattern added in #2012. At the moment, I am not sure where the cross numa node scenario might be useful, but it might be in the near future. The support can be added later on a need basis.

tools/shield_utils.py
iulianbarbu: Good idea. Will try this.

tools/shield_utils.py
iulianbarbu: Yes, indeed. This PR was inspired by how the tool is designed. Please see my reasoning above about why it is difficult to use `cset shield` in this case.

tools/devtool
iulianbarbu: Also, using `cset shield` implies having an intrusive view over the topology of the platform, where shielded tests will run. This means that besides knowing how many NUMA nodes are available on the platform and how many CPUs every node has, we must find out which NUMA node are online, what is the NUMA node cpumask, and then choose a subset of the cpulist for a specific online NUMA node, depending on the requirements of the test. Again, for the NUMA node memory module similar steps must be followed. Most of these steps are automated by this PR. A descriptive error is returned if the steps are not satisfied.

tools/devtool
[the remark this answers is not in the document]
serban300: Cross numa-node cpusets could be created by specifying multiple numa nodes: 

```
--node_0 1 --node_1 1 
```

Also, now that I think about it, we should also specify where the memory should come from. We might have scenarios where we want to allocate CPUs from node 0 and memory from node 1. Maybe not in the performance tests, but we might want to run some manual tests with such configurations sometimes. I definitely had to run such tests before.

tools/devtool
serban300: If we move the shielding logic outside docker, the extra dependency on `cset` would be hurting. If we decide to run the shielding logic in docker, `cset` might help. It would still bring some functionality that we could use. Not `cset shield` specifically, but `cset`.

As we discussed offline, first we should decide whether to have the shielding logic inside or outside docker.

tools/devtool
```diff
@@ -170,6 +193,75 @@ ok_or_die() {
...
+    ok_or_die
+}
+
+shield_environment() {
+    numa_node_id=$1
+    cpus_count=$2
+    if [ ! -z $numa_node_id ] && [ ! -z $cpus_count ]; then
+        # Locate sysfs.
+        sysfs_mountpoint=$(cat /proc/mounts | grep "^sysfs \/.\+ sysfs .*$" | cut -d' ' -f2 | head -n 1)
+        cpuset_mountpoint=$(cat /proc/mounts | grep -P "^[^ ]+ (/.+) (?:cpuset |cgroup (?:[^ ]*,)?cpuset[, ])" | cut -d' ' -f2 | head -n 1)
```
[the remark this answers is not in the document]
serban300: Thanks ! You're right. I missed this.

tools/log_utils.py
```diff
@@ -0,0 +1,68 @@
...
+    if msg_color is None:
+        msg_color = ""
+    if tag_color is None:
+        tag_color = ""
+    print(f"{tag_color}{tag}{Colors.ENDC.value} "
+          f"{msg_color}{msg}{msg_color}{Colors.ENDC.value}",
+          file=file, end="\n" if newline else "")
+
+
+def say(msg, newline=True):
```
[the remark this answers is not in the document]
iulianbarbu: Done.

tools/shield_utils.py
```diff
@@ -0,0 +1,427 @@
...
+        with open(root + "/" + dirname + "/cpuset.mems") as fp:
+            log.say(root + "/" + dirname + "/cpuset.mems")
+            mems = _parse_list_format(fp.readline())
+            if not len(list(filter(lambda m: m == numa_node_id, mems))) == 0:
+                return False
+
+    return True
+
+
+def _get_non_available_cpus():
```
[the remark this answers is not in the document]
iulianbarbu: Done.

tools/shield_utils.py
iulianbarbu: Done.

tools/shield_utils.py
iulianbarbu: Done.

tools/shield_utils.py
```diff
@@ -0,0 +1,427 @@
...
+        sys.exit(ExitCode.ESKIP)
+
+    online_numa_nodes_path = SYSFS_MOUNTPOINT + "/devices/system/node/online"
+    with open(online_numa_nodes_path) as fp:
+        online_numa_nodes = _parse_list_format(fp.readline())
+        avail_nodes = list(filter(lambda node: node != excluded_numa_node_id,
+                                  online_numa_nodes))
+        if len(avail_nodes) == 0:
+            log.say_warn("There is a single NUMA node on the system where"
+                         "tests can run. Skipping creation for the system"
```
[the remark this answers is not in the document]
iulianbarbu: Done.

tools/devtool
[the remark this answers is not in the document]
iulianbarbu: I've restructured the code to have these as global variable. Associated comments were given to explain that these global variables are relevant only for shielding functionality. PTAL. :)

iulianbarbu:
Decided to rewrite the functionality in Rust.

iulianbarbu:
Closing this in favor of #2126 .

Issue #2046, opened 2020-07-28
AArch64: graceful power-off from external


Pennyzct:
Hi guys

Got requirement from kata-containers, we want a graceful way to power off the VM from external on aarch64, since it's brutal to use `kill -9` at the first place and may leave stale resource behind.

Since x86_64 has provided API action `SendCtrlAltDel` to gracefully shutdown the microvm from external. I'd like to have the same functionality on AArch64.

dianpopa:
Hi @Pennyzct , it is really nice to see this! We were looking forward to this kind of functionality, see #1339 which was being tracked as part of our efforts to match all x86_64 functionality on aarch64 too. I will be helping you in having your PR reviewed and passing the CI testing.

Thanks,
Diana

14sea:
Hi — I have a working implementation of external graceful shutdown on aarch64 and would like to gauge whether the maintainers would accept it before opening a PR.

**Approach** (matches the direction suggested in #1339): expose a minimal **PL061 GPIO controller** as an MMIO device on aarch64 and describe a `gpio-keys` power button (`KEY_POWER`) in the FDT. The existing `SendCtrlAltDel` action is reused — on x86_64 it injects CTRL+ALT+DEL via i8042 (unchanged); on aarch64 it drives a short press/release pulse on the virtual power button. A guest with the standard `gpio-keys` driver and a power-key consumer (e.g. `systemd-logind`, `HandlePowerKey=poweroff` by default) then shuts down cleanly, and Firecracker exits on the resulting `KVM_SYSTEM_EVENT_SHUTDOWN`.

**Validated on real aarch64 + KVM hardware** (RK3568, Ubuntu guest):
- guest probes the PL061 and binds `gpio-keys`; `/dev/input/event0` receives the `KEY_POWER` press/release.
- the SPI is declared edge-triggered to match Firecracker's plain irqfd model, so the GIC interrupt count increments by exactly two per action (press + release) with no interrupt storm and ~0% host CPU.
- with `systemd-logind` active the microVM powers off automatically and Firecracker exits `exit_code=0`.
- full snapshot → restore cycle: the PL061 register state is saved/restored, so the power button still works on the restored microVM.

**Questions before I open a PR:**
1. Firecracker intentionally keeps the device model minimal. Are you open to adding the PL061 device for this, and if so should it be **opt-in / feature gated** rather than always compiled in?
[...]

PR #2047, opened 2020-07-28
AArch64: graceful power-off from external

Pennyzct:
## Reason for This PR

#2046 

## Description of Changes

Add a new API action `PressGPIOPowerOff` to inject GPIO Pin 3 keypress to the microvm.
This can be used to trigger a graceful shutdown of the microVM, if the guest has support for gpio pl061 and gpio-keys.
Firecracker needs to add emulated gpio pl061 controller and gpio-keys node in microvm.
Related commit follows ARM PrimeCell General Purpose Input/Output(PL061) specification.
See https://static.docs.arm.com/ddi0190/b/DDI0190.pdf

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/Cargo.toml
+++ b/src/devices/Cargo.toml
@@ -18,3 +18,6 @@ versionize = { version = "0.1.1" }
 versionize_derive = { git = "https://github.com/firecracker-microvm/versionize_derive", tag = "v0.1.0" }
 virtio_gen = { path = "../virtio_gen" }
 mmds = { path = "../mmds" }
+kvm-ioctls = { git = "https://github.com/firecracker-microvm/kvm-ioctls", tag = "v0.5.0-2" }
+timer = ">=0.1.3"
+chrono = ">=0.2"
--- a/src/vmm/src/device_manager/mmio.rs
+++ b/src/vmm/src/device_manager/mmio.rs
@@ -72,6 +72,16 @@ const MMIO_LEN: u64 = 0x1000;
 /// to its configuration space.
 pub const MMIO_CFG_SPACE_OFF: u64 = 0x100;
 
+/// The irq field is interpreted like this:
+/// bits:  | 31 ... 24 | 23  ... 16 | 15    ...    0 |
+/// field: | irq_type  | vcpu_index |     irq_id     |
+
+/// The irq_type field has the following values:
+/// - irq_type[0]: out-of-kernel GIC: irq_id 0 is IRQ, irq_id 1 is FIQ
+/// - irq_type[1]: in-kernel GIC: SPI, irq_id between 32 and 1019(incl.), the vcpu_index field is ignored
+/// - irq_type[2]: in-kernel GIC: PPI, irq_id between 16 and 31 (incl.)
+const SPI_IRQ_BASE: u32 = 0x01000020;
+
 /// Stores the address range and irq allocated to this device.
 #[derive(Clone, Debug, PartialEq, Versionize)]
 pub struct MMIODeviceInfo {
@@ -86,6 +96,8 @@ pub struct MMIODeviceInfo {
 /// Manages the complexities of registering a MMIO device.
 pub struct MMIODeviceManager {
     pub(crate) bus: devices::Bus,
+    #[cfg(target_arch = "aarch64")]
+    pub gpio_pl061: Option<Arc<Mutex<devices::legacy::GPIO>>>,
     mmio_base: u64,
     irq: u32,
     last_irq: u32,
@@ -101,6 +113,8 @@ impl MMIODeviceManager {
             last_irq: irq_interval.1,
             bus: devices::Bus::new(),
             id_to_dev_info: HashMap::new(),
+            #[cfg(target_arch = "aarch64")]
+            gpio_pl061: None,
         }
     }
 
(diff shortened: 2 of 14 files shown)


Pennyzct:
@weichen81

src/devices/Cargo.toml
```diff
@@ -18,3 +18,6 @@ versionize = { version = "0.1.1" }
 versionize_derive = { git = "https://github.com/firecracker-microvm/versionize_derive", tag = "v0.1.0" }
 virtio_gen = { path = "../virtio_gen" }
 mmds = { path = "../mmds" }
+kvm-ioctls = { git = "https://github.com/firecracker-microvm/kvm-ioctls", tag = "v0.5.0-2" }
+timer = ">=0.1.3"
+chrono = ">=0.2"
```
acatangiu, line 23: I think we need a different mechanism than relying on `timer` (which also brings in `chrono`) for multiple reasons:
 - `timer` crate has an incompatible `MPL-2.0` license
 - `timer` is implemented using a multi-thread timer model: spawns a new thread to do the work after a certain time. This is not in line with our security model/posture where we want to keep strict control over process threads (our `seccomp` filters do not allow spawning new threads).
 - `chrono` is a bloated dependency with a long-ish trail of child dependencies. We'd like to avoid it since the use-case for it is simple enough to cover with `std::time`.

Possible alternative: use `TimerFd` - for example that's what we use in our Rate Limiter implementation: https://github.com/firecracker-microvm/firecracker/blob/master/src/rate_limiter/src/lib.rs

src/vmm/src/device_manager/mmio.rs
```diff
@@ -86,6 +96,8 @@ pub struct MMIODeviceInfo {
 /// Manages the complexities of registering a MMIO device.
 pub struct MMIODeviceManager {
     pub(crate) bus: devices::Bus,
+    #[cfg(target_arch = "aarch64")]
+    pub gpio_pl061: Option<Arc<Mutex<devices::legacy::GPIO>>>,
```
acatangiu, line 100: Would it make sense to create a `gpio` bus (which only currently has this device)?

src/devices/Cargo.toml
Pennyzct: thanks. I'll try to replace them with `TimerFd`. ;)

Pennyzct:
Hi @acatangiu @dianpopa thanks for the review~
The reason why I chose gpio-keys to power off VM lies [here](http://lists.infradead.org/pipermail/linux-arm-kernel/2015-May/341037.html). 
ACPI may be the consistent way across archs . ;). Although, right now, AArch64 needs extra UEFI support to boot from ACPI, and that may lead to extra boot-up time and memory footprint.

src/devices/Cargo.toml
acatangiu: Our device model uses a single emulation thread that asynchronously handles device events. `TimerFd` would be such a `gpio_pl061` device event.

The device needs to use our `EventManager` [Subscriber interface](https://github.com/firecracker-microvm/firecracker/blob/db84b948461bddcc89c0b69aa5639368a7d7c3ef/src/polly/src/event_manager.rs#L50) to register async events.
A simple example device with a single event can be seen in our `Serial device` implementation: https://github.com/firecracker-microvm/firecracker/blob/db84b948461bddcc89c0b69aa5639368a7d7c3ef/src/devices/src/legacy/serial.rs#L290

Pennyzct:
Hi @acatangiu @dianpopa 
For issuing level-triggered interrupt, I'm now using `KVM_IRQ_LINE` ioctl, and using `timer` to emulate outbound irq line being raised for 100ms. _**Although, it seems to me that `KVM_IRQ_LINE` hasn't been included in syscall whitelist.**_
Or we could still use `KVM_IRQFD` ioctl to do level-triggered interrupt, but then we also need to implement a new func `register_irqfd_resample` in https://github.com/firecracker-microvm/kvm-ioctls/blob/master/src/ioctls/vm.rs to allow resample.

> With KVM_IRQFD_FLAG_RESAMPLE, KVM_IRQFD supports a de-assert and notify
> mechanism allowing emulation of level-triggered, irqfd-based interrupts.
> When KVM_IRQFD_FLAG_RESAMPLE is set the user must pass an additional eventfd in the kvm_irqfd.resamplefd field.  
> When operating in resample mode, posting of an interrupt through kvm_irq.fd asserts
> the specified gsi in the irqchip.  When the irqchip is resampled, such as from an EOI, 
> the gsi is de-asserted and the user is notified via kvm_irqfd.resamplefd.  
> 
So Wdyt? ☺

dianpopa:
Hi ,

Sorry for the late late reply. We want to take some time and think about the design decisions implied by adding support for shutdown on aarch64. We have lots of other priorities so it might take a while until we come back to your PR. Thanks for your contribution and your patience @Pennyzct !

dianpopa:
Hi @Pennyzct ,

Last year was a very busy one for us and I wanted to let you know that we are still trying to reach a conclusion regarding your PR. We are still interested in implementing `shutdown` functionality on arm64 too but we still need some time to decide on how this functionality is going to integrate with Firecracker. The reason for that is that we are currently reconsidering our decision to use FDT for describing the hardware components of an aarch64 microVM and until we have not reached a conclusion we would like to postpone merging this PR. We will get back to you when we have decided on a path forward! Thanks!

dianpopa:
Hi @Pennyzct,

We reached a conclusion regarding this functionality. Later this year we are planning on adding ACPI PM support in Firecracker, check [this](https://github.com/firecracker-microvm/firecracker/issues/2493) out. Having ACPI power management will allow for a more straightforward way of doing reboot on all platforms currently supported without the need to use a dedicated device. As a consequence, I am closing this PR for the moment. Thanks for your implication!

PR #2048, opened 2020-07-28
rate_limiter: enforce limit for large requests

### LINKED ISSUE #259: rate limiting: handle or disallow scenarios where max virtio req size > token bucket size
Rate limiting currently works by allowing or temporarily blocking virtio virtqueue requests based on the request size and the ratelimiter token bucket remaining budget.
We can easily reproduce scenarios where a virtio request can exceed the token bucket size, in such scenarios traffic is forever blocked since there will never be enough budget to serve that request.

Solutions:
1. Augment the design and implementation to be able to do partial processing of virtio requests. This means we need to do extra internal accounting of queue processing.
2. Enforce limits and validations in rate limiter configurations to disallow such scenarios.

georgepisaltu:
Handled the corner case where a request to the rate limiter wanted
to consume more tokens than the bucket size by emptying the bucket
and arming the timer with a custom duration such that the rate
limiting is consistent.

Signed-off-by: George Pisaltu <gpl@amazon.com>

## Reason for This PR

Fixes #259 

## Description of Changes

## License Acceptance

## PR Checklist

### DIFF
--- a/src/rate_limiter/src/lib.rs
+++ b/src/rate_limiter/src/lib.rs
@@ -86,6 +86,17 @@ fn gcd(x: u64, y: u64) -> u64 {
     x
 }
 
+/// Enum describing the outcomes of a `reduce()` call on a `TokenBucket`.
+#[derive(Clone, Debug, PartialEq)]
+pub enum BucketReduction {
+    /// There are not enough tokens to complete the operation.
+    Failure,
+    /// A part of the available tokens have been consumed.
+    Success,
+    /// A number of tokens `inner` times larger than the bucket size have been consumed.
+    OverConsumption(f64),
+}
+
 /// TokenBucket provides a lower level interface to rate limiting with a
 /// configurable capacity, refill-rate and initial burst.
 #[derive(Clone, Debug, PartialEq)]
@@ -148,7 +159,7 @@ impl TokenBucket {
     /// Attempts to consume `tokens` from the bucket and returns whether the action succeeded.
     // TODO (Issue #259): handle cases where a single request is larger than the full capacity
     // for such cases we need to support partial fulfilment of requests
-    pub fn reduce(&mut self, mut tokens: u64) -> bool {
+    pub fn reduce(&mut self, mut tokens: u64) -> BucketReduction {
         // First things first: consume the one-time-burst budget.
         if self.one_time_burst > 0 {
             // We still have burst budget for *all* tokens requests.
@@ -156,7 +167,7 @@ impl TokenBucket {
                 self.one_time_burst -= tokens;
                 self.last_update = Instant::now();
                 // No need to continue to the refill process, we still have burst budget to consume from.
-                return true;
+                return BucketReduction::Success;
             } else {
                 // We still have burst budget for *some* of the tokens requests.
                 // The tokens left unfulfilled will be consumed from current `self.budget`.
@@ -180,24 +191,24 @@ impl TokenBucket {
         }
 
         if tokens > self.budget {
-            // TODO (Issue #259) remove this block when issue is resolved
+            // This operation requests a bandwidth higher than the bucket size
             if tokens > self.size {
                 error!(
-                    "Trying to consume more tokens {} than the total capacity {}",
+                    "Consumed {} tokens from bucket of size {}",
                     tokens, self.size
                 );
-                // best effort rate-limiting, this is a dirty workaround for Issue #259
-                if self.budget == self.size {
-                    self.budget = 0;
-                    return true;
-                }
+                // Empty the bucket and report an overconsumption of
+                // (remaining tokens / size) times larger than the bucket size
+                tokens -= self.budget;
+                self.budget = 0;
+                return BucketReduction::OverConsumption(tokens as f64 / self.size as f64);
             }
             // If not enough tokens consume() fails, return false.
-            return false;
+            return BucketReduction::Failure;
         }
 
         self.budget -= tokens;
-        true
+        BucketReduction::Success
     }
 
     /// "Manually" adds tokens to bucket.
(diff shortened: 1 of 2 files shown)


src/rate_limiter/src/lib.rs
```diff
@@ -180,24 +190,25 @@ impl TokenBucket {
...
                     tokens, self.size
                 );
-                // best effort rate-limiting, this is a dirty workaround for Issue #259
+                // If the bucket is full, empty it and report an overconsumption of
+                // (tokens / size) times larger than the bucket size
                 if self.budget == self.size {
                     self.budget = 0;
-                    return true;
+                    return BucketReduction::OverConsumption(tokens as f64 / self.size as f64);
                 }
```
acatangiu, line 204: Instead of waiting for bucket to be full before doing the `OverConsumption` you should do it in place. By waiting for bucket to be full you're also losing precision by not taking the 'waiting' period in the rate-limiting accounting.

```rust
// If the bucket is full, empty it and report an overconsumption of
// (tokens / size) times larger than the bucket size
tokens -= self.budget;
self.budget = 0;
return BucketReduction::OverConsumption(tokens as f64 / self.size as f64);
```

src/rate_limiter/src/lib.rs
```diff
@@ -347,6 +358,23 @@ impl RateLimiter {
         })
     }
 
+    // Arm the timer of the rate limiter with `duration` milliseconds.
+    // If `duration` is `None`, the minimum refill duration will be used.
+    fn activate_timer(&mut self, duration: Option<u64>) {
```
acatangiu, line 363: You could drop the extra complexity of the `duration: Option<u64>` parameter. It involves the caller knowing that `None` maps to a default and `Some` maps to the custom value.

Just take `TimerState` as a param.

```suggestion
    fn activate_timer(&mut self, new_state: TimerState) {
```

src/rate_limiter/src/lib.rs
```diff
@@ -347,6 +358,23 @@ impl RateLimiter {
...
+        if !self.timer_active {
```
acatangiu, line 364: I think we can drop this check. Timer should never be active when we get here, and even if it was, we should update it to its new deadline.
```suggestion
```

src/rate_limiter/src/lib.rs
```diff
@@ -357,23 +385,38 @@ impl RateLimiter {
...
-            Some(bucket) => bucket.reduce(tokens),
+        match token_bucket {
+            Some(bucket) => {
+                let refill_time = bucket.refill_time_ms();
+                match bucket.reduce(tokens) {
+                    // When we report budget is over, there will be no further calls here,
+                    // register a timer to replenish the bucket and resume processing;
+                    // make sure there is only one running timer for this limiter.
+                    BucketReduction::Failure => {
+                        self.activate_timer(None);
```
acatangiu, line 396: Ties into the comment above: clearer to do:
```suggestion
                        self.activate_timer(TIMER_REFILL_STATE);
```

acatangiu:
Approach is solid, the fix should work great! Some implementation details changes requested.

src/rate_limiter/src/lib.rs
georgepisaltu: Revised the logic in `reduce()` according to this.

src/rate_limiter/src/lib.rs
georgepisaltu: Done

src/rate_limiter/src/lib.rs
georgepisaltu: It should indeed overwrite the older timer config anyway.

src/rate_limiter/src/lib.rs
georgepisaltu: Done

src/rate_limiter/src/lib.rs
```diff
@@ -357,23 +374,40 @@ impl RateLimiter {
...
+        match token_bucket {
+            Some(bucket) => {
+                let refill_time = bucket.refill_time_ms();
+                match bucket.reduce(tokens) {
+                    // When we report budget is over, there will be no further calls here,
+                    // register a timer to replenish the bucket and resume processing;
+                    // make sure there is only one running timer for this limiter.
+                    BucketReduction::Failure => {
+                        self.activate_timer(TIMER_REFILL_STATE);
+                        false
```
acatangiu, line 386: Actually, this new code changes the relation between timer state (active/inactive) and the return value of the main `consume()` method (success/failure).

With the old code there it was guaranteed that timer would only be active/activated when a `consume()` operation fails. This version of the code activates the timer on `OverConsumption` **while** allowing (returning `true`) on that `consume()` call.

As such, the `activate_timer()` should set default state in case of `BucketReduction::Failure` only if `timer.active == false`.

You can reintroduce the timer state check in `activate_timer()` or wrap it here locally where the reason/need for it is localized:
```suggestion
                        if !self.timer_active {
                            self.activate_timer(TIMER_REFILL_STATE);
                        }
                        false
```

src/rate_limiter/src/lib.rs
```diff
@@ -357,23 +374,40 @@ impl RateLimiter {
...
+                        // The operation "borrowed" a number of tokens `ratio` times
+                        // greater than the size of the bucket, and since it takes
+                        // `refill_time` milliseconds to fill an empty bucket, in
+                        // order to enforce the bandwidth limit we need to prevent
+                        // further calls to the rate limiter for
+                        // `ratio * refill_time` milliseconds.
+                        self.activate_timer(TimerState::Oneshot(Duration::from_millis(
+                            (ratio * refill_time as f64) as u64,
+                        )));
+                        true
```
acatangiu, line 402: There another conceptual problem with this fix.. :

If **all** your packets/consume-requests are larger than bucket size, they **all** go through with _no rate-limiting_ whatsoever. This happens because you allow them through but never block on the limiter. The current logic relies on the idea that a future packet with size < bucket.size will block the limiter.

acatangiu: Nevermind, we also expose the `is_blocked()` api which reports the rate-limiter as blocked if there is a timer active, so above statement is false.

src/rate_limiter/src/lib.rs
```diff
@@ -549,26 +602,28 @@ pub(crate) mod tests {
         let refill_ms = 1000;
         let mut tb = TokenBucket::new(capacity, 0, refill_ms as u64).unwrap();
 
-        assert!(tb.reduce(123));
+        assert!(tb.reduce(123) == BucketReduction::Success);
```
acatangiu, line 605: consistency nit:
```suggestion
        assert_eq!(tb.reduce(123), BucketReduction::Success);
```

Same for all below.

src/rate_limiter/src/lib.rs
georgepisaltu: Added this as well as a check at the beginning of `consume()` to not let the rate limiter consume any tokens while the timer is active.

src/rate_limiter/src/lib.rs
```diff
@@ -508,6 +550,25 @@ pub(crate) mod tests {
         }
     }
 
+    impl PartialEq for BucketReduction {
```
sandreim, line 553: Why not let generate this impl ? 
`#[derive(PartialEq)]`

src/rate_limiter/src/lib.rs
```diff
@@ -735,10 +798,63 @@ pub(crate) mod tests {
         assert!(l.consume(100, TokenType::Ops));
         // try and succeed on another 100 bytes this time
         assert!(l.consume(100, TokenType::Bytes));
+    }
+
+    #[test]
+    fn test_rate_limiter_overconsumption() {
```
sandreim, line 804: Does this test fail without your fix ?

src/rate_limiter/src/lib.rs
```diff
@@ -347,33 +358,64 @@ impl RateLimiter {
...
+
         // Identify the required token bucket.
         let token_bucket = match token_type {
             TokenType::Bytes => self.bandwidth.as_mut(),
             TokenType::Ops => self.ops.as_mut(),
         };
         // Try to consume from the token bucket.
-        let success = match token_bucket {
-            Some(bucket) => bucket.reduce(tokens),
+        match token_bucket {
```
sandreim, line 383: `if let` would make it more readable. Nested matches usually are hard to read.

src/rate_limiter/src/lib.rs
```diff
@@ -180,24 +191,24 @@ impl TokenBucket {
...
-                // best effort rate-limiting, this is a dirty workaround for Issue #259
-                if self.budget == self.size {
-                    self.budget = 0;
-                    return true;
-                }
+                // Empty the bucket and report an overconsumption of
+                // (remaining tokens / size) times larger than the bucket size
+                tokens -= self.budget;
+                self.budget = 0;
+                return BucketReduction::OverConsumption(tokens as f64 / self.size as f64);
```
sandreim, line 204: Can self.size be 0 ?

georgepisaltu: No, as you cannot construct a `TokenBucket` with size 0.

src/rate_limiter/src/lib.rs
georgepisaltu: I didn't think it can be derived, fixed now.

src/rate_limiter/src/lib.rs
georgepisaltu: Changed it to an if-let statement.

src/rate_limiter/src/lib.rs
georgepisaltu: It does.

PR #2051, opened 2020-07-28
tests: added statistics component

iulianbarbu:
## Reason for This PR

This PR adds a generic statistical information gathering component. It can be used for performance testing, where statistical information can be inferred from multiple iterations.

An usage example of this component was added for `test_network_latency.py` performance test, provided in #1916 .

## Description of Changes

The component has three entities:
* Statistics core: base class which exposes the loop for managing multiple measurements.
* Statistics producer: base class which is used by the core to run measurements on host or guest and produce raw data output.
* Statistics consumer: base class which consumes multiple results, in multiple iterations, parses them and computes the statistics.

**Note**: This PR is focused only on the results computation of statistical tests. The statistics result is a JSON which will be augmented on a need basis, depending on how we want to individualize/customize the details about a test run (environment, platform, tests inputs etc.). These results can be considered tests artifacts and are supposed to be indexed later on, queried/visualized and monitored for alarming.

## License Acceptance

## PR Checklist

- ~~[ ] Any newly added `unsafe` code is properly documented.~~
- ~~[ ] Any API changes are reflected in `firecracker/swagger.yaml`.~~
- ~~[ ] Any user-facing changes are mentioned in `CHANGELOG.md`.~~

### DIFF
--- a/tests/framework/scheduler.py
+++ b/tests/framework/scheduler.py
@@ -29,6 +29,8 @@ from random import random
 from select import select
 from time import sleep
 import pytest
+# Needed to force deselect nonci tests.
+from _pytest.mark import Expression, MarkMatcher
 from _pytest.main import ExitCode
 
 from . import mpsing  # pylint: disable=relative-beyond-top-level
@@ -88,6 +90,21 @@ class PytestScheduler(mpsing.MultiprocessSingleton):
         This will execute in the worker / child context.
         """
         self._add_report(report)
+
+    @staticmethod
+    def filter_batch(config, batch, marker_name):
+        """Deselect marked tests which are not explicitly selected."""
+        deselected = []
+        expr = Expression.compile(config.option.markexpr)
+        for item in batch['items'][:]:
+            for key in item.keywords:
+                if key is marker_name and \
+                        not expr.evaluate(MarkMatcher.from_item(item)):
+                    deselected.append(item)
+                    batch['items'].remove(item)
+                    break
+
+        config.hook.pytest_deselected(items=deselected)
 
     def pytest_runtestloop(self, session):
         """Pytest hook. The main test scheduling and running loop.
@@ -155,17 +172,27 @@ class PytestScheduler(mpsing.MultiprocessSingleton):
         for item in session.items:
             # A test can match any of the patterns defined by the batch,
             # in order to get assigned to it.
-            next(
+            for batch in schedule:
                 # Found a matching batch. No need to look any further.
-                batch['items'].append(item) for batch in schedule
                 if re.search(
                     "|".join(["({})".format(x) for x in batch['patterns']]),
                     "/".join(item.listnames()),
-                ) is not None
-            )
+                ) is not None:
+                    batch['items'].append(item)
+                    break
 
         # Filter out empty batches.
         schedule = [batch for batch in schedule if batch['items']]
+
+        # Evaluate marker expression only for the marked batch items.
+        # If pytest runs with a marker expression which does not include
+        # `nonci` marked tests (e.g `-m "not nonci" or non-existent marker
+        # expression), the tests marked with `nonci` marker will be skipped.
+        for batch in schedule:
+            PytestScheduler.filter_batch(session.config,
+                                         batch,
+                                         marker_name="nonci")
+            break
 
         for batch in schedule:
             self._raw_stdout(
--- a/tests/framework/statistics/consumer.py
+++ b/tests/framework/statistics/consumer.py
@@ -1,0 +1,145 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Module for multiple statistics consumers."""
+
+from abc import ABC, abstractmethod
+from typing import Dict, Any
+from collections import defaultdict
+from .types import MeasurementDef, StatisticDef
+
+
+class Consumer(ABC):
+    """Base class for statistics aggregation class."""
+
+    UNIT_KEY = "unit"
+    DATA_KEY = "data"
+
+    # pylint: disable=W0102
+    def __init__(self,
+                 consume_stats,
+                 custom=dict()):
+        """Initialize a consumer."""
+        self._results = defaultdict()  # Aggregated results.
+        self._measurements_defs = defaultdict(MeasurementDef)
+        self._statistics_defs = defaultdict()
+        self._consume_stats = consume_stats
+        self._custom = custom
+        self._statistics = dict()
+        self._iteration = 0
+
+    @abstractmethod
+    def ingest(self, iteration: int, raw_data: Any):
+        """Abstract method for ingesting the raw result."""
+
+    def consume_stat(self, st_name: str, ms_name: str, value: MeasurementDef):
+        """Aggregate statistics."""
+        results = self._results.get(ms_name)
+        if not results:
+            self._results[ms_name] = dict()
+        st_data = self._results[ms_name].get(st_name)
+        if not st_data:
+            self._results[ms_name][st_name] = list()
+        self._results[ms_name][st_name].append(value)
+
+    def consume_measurement(self,
+                            ms_name: str,
+                            value: MeasurementDef):
+        """Aggregate measurement."""
+        results = self._results.get(ms_name)
+        if results is None:
+            self._results[ms_name] = dict()
+            self._results[ms_name][self.DATA_KEY] = list()
+        self._results[ms_name][self.DATA_KEY].append(value)
+
+    def consume_custom(self, name, value: Any):
+        """Aggregate custom information."""
+        if not self._custom[self._iteration]:
+            self._custom[self._iteration] = {}
+        self._custom[self._iteration][name] = value
+
+    def set_stat_def(self, value: StatisticDef):
+        """Set statistics definition."""
+        if not self._statistics_defs.get(value.measurement_name):
+            self._statistics_defs[value.measurement_name] = dict()
+
+        self._statistics_defs[value.measurement_name][value.name] = value
+
+    def set_measurement_def(self, value: MeasurementDef):
+        """Set measurement definition."""
+        self._measurements_defs[value.name] = value
+
+    def process(self) -> Dict[str, dict]:
+        """Generate statistics as a dictionary."""
+        # Verify that the statistics/measurements correspondence
+        # is backed by corresponding measurements definitions.
+        for ms_name in self._statistics_defs:
+            assert ms_name in self._measurements_defs
+
+        if self._consume_stats:
+            # Verify if the gathered results for a measurement are
+            # backed by measurements definitions.
+            for ms_name in self._results:
+                assert ms_name in self._measurements_defs
+                # Verify if the gathered statistics are backed by
+                # statistics definitions.
+                for st_name in self._results[ms_name]:
+                    assert st_name in self._statistics_defs[ms_name]
+        else:
+            # Verify if the gathered measurements are backed by
+            # measurements definitions.
+            for ms_name in self._results:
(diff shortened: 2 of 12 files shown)


tests/framework/statistics/core.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+from collections import namedtuple
+import statistics
+
+
+DEFAULT_ITERATIONS_COUNT = 10
+
+
+Config = namedtuple("Config", "name fc_version platform iterations")
```
gc-plp, line 17: If these objects are not part of the API, we can hide these behind classes.

gc-plp (commented):
Looks good overall!

iulianbarbu:
@gc-plp , please take a look!

gc-plp (dismissed):

gc-plp (approved):

tests/integration_tests/performance/test_network_latency.py
```diff
@@ -0,0 +1,202 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Tests the network latency of a Firecracker guest."""
+
+import logging
+import platform
+import re
+import pytest
+import host_tools.network as net_tools
+from conftest import _test_images_s3_bucket
```
gc-plp, line 10: We should not import symbols starting with `_` as they're designed for internal use within the file where they are declared.

tests/integration_tests/performance/test_network_latency.py
```diff
@@ -0,0 +1,202 @@
...
+    64 bytes from 8.8.8.8: icmp_seq=1 ttl=118 time=17.7 ms
+    64 bytes from 8.8.8.8: icmp_seq=2 ttl=118 time=17.7 ms
+    64 bytes from 8.8.8.8: icmp_seq=3 ttl=118 time=17.4 ms
+    64 bytes from 8.8.8.8: icmp_seq=4 ttl=118 time=17.8 ms
+
+    --- 8.8.8.8 ping statistics ---
+    4 packets transmitted, 4 received, 0% packet loss, time 3005ms
+    rtt min/avg/max/mdev = 17.478/17.705/17.808/0.210 ms
+    """
+    eager_map(cons.consume_measurement_def, measurements())
```
gc-plp, line 81: nit: might as well be:

```
[cons.consume_measurement_def(i) for i in measurements()]
[cons.consume_stat_def(i) for i in stats()]
```

tests/integration_tests/performance/test_network_latency.py
iulianbarbu: Done.

tests/integration_tests/performance/test_network_latency.py
iulianbarbu: This results in python styling issues, since the result of the list comprehension is not assigned/used by anything, so we would need to discard by assigning them to `don't care ( _ )`. I prefer an `eager_map` approach instead.

tests/integration_tests/performance/test_network_latency.py
```diff
@@ -0,0 +1,199 @@
...
+        time = re.findall(FLOAT_REG, time)[0]
+        times.append(time)
+
+    times.sort()
+    cons.consume_result(name=consumer.StatisticDef.P50_KEY,
+                        value=times[int(requests * 0.5)])
+    cons.consume_result(name=consumer.StatisticDef.P90_KEY,
+                        value=times[int(requests * 0.9)])
+    cons.consume_result(name=consumer.StatisticDef.P99_KEY,
+                        value=times[int(requests * 0.99)])
```
serban300, line 120: Shouldn't the consumer or aggregator compute this automatically from the measurements that we supply ?

tests/framework/statistics/core.py
```diff
@@ -0,0 +1,96 @@
...
+    # pylint: disable=W0102
+    def __init__(self, name, iterations, custom={}):
+        """Core constructor."""
+        self._pipes = defaultdict(Pipe)
+        self._statistics = Statistics(name=name,
+                                      iterations=iterations,
+                                      results={},
+                                      custom=custom)
+
+    def add_pipe(self, producer: Producer, consumer: Consumer, tag=None):
```
serban300, line 39: Are there use-cases for more pipes ? Do we need this ?

serban300 (commented):
Overall I find the producer-consumer model very complex. Will we need this in the future ? Could we get away with just a statstistics aggregator ?

tests/framework/statistics/consumer.py
```diff
@@ -0,0 +1,306 @@
...
+        """Return the 90th percentile."""
+        return cls._percentile(results, 90)
+
+    @classmethod
+    def p99(cls, results):
+        """Return the 99th percentile."""
+        return cls._percentile(results, 99)
+
+
+class Criteria:
```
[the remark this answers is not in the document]
iulianbarbu: `Operator` naming for this class is not intent visible. We want it to be used for enforcing hard requirements over the tests behavior, meaning a `criteria`/condition for this test to pass would need to have some measurement grater than a baseline. The goal here would be to have a lazy assertion mechanism.

tests/integration_tests/performance/test_network_latency.py
iulianbarbu: Generally, there are two modes of driving the producer/consumer pipes. Some of them can produce *measurements data points*, which can be transformed in statistics by applying  on them statistical methods in the consumer statistics generation step while others can directly *produce the statistics data*, which should be directly forwarded to the consumer statistics generation step. However, both measurements and statistics definitions are needed in either way for statistics generation, since statistics units are set by measurements definitions.

In this specific case we initialize the `LambdaConsumer` with `consume_stats=True`. This means we directly consume statistics for our measurements, rather than measurements data.

On a separate note, I must avoid asserting on the measurements consumption expectancy, from `Consumer::process` method: i.e if the consumer is initialized with `consume_stats=True` than we must not check if measurements definitions have corresponding data points consumed. I'll let you know after I fix it.

tests/framework/statistics/consumer.py
```diff
@@ -0,0 +1,306 @@
...
+                self._statistics[stat.name][self.RESULT_KEY] = \
+                    stat.apply(self._results[m_name])
+
+            # Check pass criteria.
+            if callable(stat.check) and stat.check.target:
+                res = self._statistics[stat.name][self.RESULT_KEY]
+                assert stat.check(res), "%r criteria for %r failed." \
+                                        " Target: %r vs actual: %r." \
+                                        % (stat.check.name, stat.name,
+                                           str(stat.check.target), str(res))
```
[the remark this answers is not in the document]
iulianbarbu: I see this as an oversimplification because to have a general ready to use framework we must answer to the following questions:
1. Who will push the data to the aggregator? 
2. What format has the data pushed to the aggregator and who process it?
3. What interface has the entity which pushes data to the aggregator?

The `Producer/Consumer` pattern encapsulates the answers for these questions.
1. A raw data producer will push data to the "aggregator" (Consumer).
2. The "aggregator" (Consumer) is responsible for parsing and extracting the needed information from the produced raw data.
3. A `Producer` will produce raw data if an implementation for `produce` abstract method is provided. More specifically, in the case of a `LambdaProducer`, the contract is that it receives a Pyhton function and optionally kwargs for that function which will produce a raw output (stream of bytes, strings, raw bytes etc). The rest is covered by the inner implementation. 

Let me know your thoughts.

iulianbarbu:
> Overall I find the producer-consumer model very complex. Will we need this in the future ? Could we get away with just a statstistics aggregator ?

Shortly put, I think we need this for the near future because there are lots of tests we need to implement around it, which are in need of such functionality (i.e performance tests). I can not easily see how a statistics aggregator can provide all the needs detailed above. Let me know if you have something specific in mind.

tests/framework/statistics/core.py
serban300: Oh, I see. This might explain the overall complexity of the statistics component. I have to give it more thought.

serban300 (approved):
I did another review iteration on the PR. Looks good to me now. I resolved all the comments.

iulianbarbu:
@serban300 , @gc-plp , I've fixed the coverage. For some reason it differs between my x86_64 machine and the CI x86_64 machine. PTAL.

iulianbarbu:
Another issue is that the network latency performance test is supported at the moment only on x86_64. We need baselines for aarch64 and amd64 as well. This can be done in future PRs. @serban300 , @gc-plp , WDYT?

tests/framework/scheduler.py
```diff
@@ -89,6 +91,26 @@ def pytest_runtest_logreport(self, report):
         """
         self._add_report(report)
 
+    @staticmethod
+    def deselect_if_marked(config, batch, marker_name):
+        """Deselect marked tests which are not explicitly selected."""
+        deselected = []
+        remaining = []
```
serban300, line 98: I'm not sure we need this array. We can just remove from `batch[items]` if ?I understand correctly.

tests/framework/scheduler.py
```diff
@@ -89,6 +91,26 @@ def pytest_runtest_logreport(self, report):
         """
         self._add_report(report)
 
+    @staticmethod
+    def deselect_if_marked(config, batch, marker_name):
```
serban300, line 95: I would rename this to something like `filter_batch()`

tests/framework/scheduler.py
iulianbarbu: You are right. I was under the impression that we can remove while iterating. Changed it as you suggested.

tests/framework/scheduler.py
iulianbarbu: Done.

tests/framework/scheduler.py
serban300: Oh I was thinking at something like 

```
for i=0; i < items.len; i++ {
    if (...) {
        items.remove(i);
        i--;
    }
}
```

I'm not sure that `batch['items'].remove(item)` works correctly. Did you check it ? I found a thread on a forum stating something like:

```
>>> somelist = range(10)
>>> for x in somelist:
...     somelist.remove(x)
>>> somelist
[1, 3, 5, 7, 9]

>>> somelist = range(10)
>>> for x in somelist[:]:
...     somelist.remove(x)
>>> somelist
[]
```

So the result depends on the type of list that you're iterating on.

gc-plp: > I was under the impression that we can't remove while iterating

We can remove while iterating, but it's not safe to do so.

When iterating over an array, the generator will keep track of the yielded element index. If you remove one element, the generator sequence may be mixed up.

For example, let's assume that we have:
```
arr = [1, 2, 3, 4]

for i in arr:
    print(i)
```

Running this snippet will print:
```
1
2
3
4
```

What if while looping, we remove one of the elements?

```
arr = [1, 2, 3, 4]

for i in arr:
    print(i)

    if i == 2:
        arr.remove(i)
```

Running this snippet will print:

```
1
2
4
```
Why wasn't `3` printed?
What happened here was that when element `2` was removed the list length decreased and the indexes of the elements following `2` were decremented:
```
 1   2   3   4 
[0] [1] [2] [3]
```
Became:
```
 1   3   4
[0] [1] [2]
```

Since at removal time the index was `1`, the next yielded element was `arr[2]`.

In order to remove elements while iterating one should either create a copy of that list or create a separate list of elements to be removed and remove them after the loop.

iulianbarbu: Yes, you are right. I've checked it before for a list with a single element, which works. But having an example like yours it does not work anymore. I've added the sugar you mentioned above `[:]`. PTAL!

serban300 (dismissed):

serban300 (approved):

gc-plp (approved):

PR #2055, opened 2020-07-31
Improved error handling for rate limiter

luminitavoicu:

## Reason for This PR

Invalid inputs for rate limiter might cause Firecracker to panic while loading a snapshot.

## Description of Changes

Propagated the error upwards.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/rate_limiter/src/persist.rs
+++ b/src/rate_limiter/src/persist.rs
@@ -71,15 +71,16 @@ impl Persist<'_> for RateLimiter {
 
     fn restore(_: Self::ConstructorArgs, state: &Self::State) -> Result<Self, Self::Error> {
         let rate_limiter = RateLimiter {
-            // Safe to unwrap because TokenBucket::restore doesn't return errors.
-            ops: state
-                .ops
-                .as_ref()
-                .map(|ops| TokenBucket::restore((), ops).unwrap()),
-            bandwidth: state
-                .bandwidth
-                .as_ref()
-                .map(|bw| TokenBucket::restore((), bw).unwrap()),
+            ops: if let Some(ops) = state.ops.as_ref() {
+                Some(TokenBucket::restore((), ops)?)
+            } else {
+                None
+            },
+            bandwidth: if let Some(bw) = state.bandwidth.as_ref() {
+                Some(TokenBucket::restore((), bw)?)
+            } else {
+                None
+            },
             timer_fd: TimerFd::new_custom(ClockId::Monotonic, true, true)?,
             timer_active: false,
         };


src/rate_limiter/src/persist.rs
```diff
@@ -72,10 +75,12 @@ impl Persist<'_> for RateLimiter {
...
         let rate_limiter = RateLimiter {
             // Safe to unwrap because TokenBucket::restore doesn't return errors.
-            ops: state
-                .ops
-                .as_ref()
-                .map(|ops| TokenBucket::restore((), ops).unwrap()),
+            ops: state.ops.as_ref().map(|ops| {
+                TokenBucket::restore((), ops).unwrap_or_else(|e| {
+                    error!("Failed to restore rate limiter: {}", e);
+                    std::process::exit(i32::from(EXIT_CODE_GENERIC_ERROR));
```
serban300, line 81: I think the error should be propagated upwards and handled somewhere else, but I'm not very familiar with the snapshoting code.

ioanachirca, line 81: Why exit the process here instead of propagating the error? AFAIK we're not doing this in the other restore contexts.

ioanachirca:
Could you change the commit message to something like "rate limiter: improve error handling"? You can add more context in the following lines (i.e. that this unwrap might cause an error when loading a snapshot), but the title should be a bit more specific 😄

src/rate_limiter/src/persist.rs
acatangiu: Indeed, as a general rule, if the _current_ function returns a `Result` you should propagate any local errors through that Result.
Exceptions to this rule are error cases where running further instructions is either not safe or doesn't make sense from a program logic point of view (for example: mutex lock is poisoned - continuing the program generally results in Undefined Behavior).

luminitavoicu: Good point. I'll change that.

src/rate_limiter/src/persist.rs
```diff
@@ -71,15 +71,18 @@ impl Persist<'_> for RateLimiter {
...
-                .map(|bw| TokenBucket::restore((), bw).unwrap()),
+            ops: if let Some(ops) = state.ops.as_ref() {
+                let tb = TokenBucket::restore((), ops)?;
+                Some(tb)
+            } else {
+                None
+            },
+            bandwidth: if let Some(bw) = state.bandwidth.as_ref() {
+                let tb = TokenBucket::restore((), bw)?;
+                Some(tb)
```
ioanachirca, line 82: you could directly use `Some(TokenBucket::restore((), bw)?)`

luminitavoicu: Sure, I missed that. Solved.

src/rate_limiter/src/persist.rs
```diff
@@ -71,15 +71,16 @@ impl Persist<'_> for RateLimiter {
...
-                .map(|ops| TokenBucket::restore((), ops).unwrap()),
-            bandwidth: state
-                .bandwidth
-                .as_ref()
-                .map(|bw| TokenBucket::restore((), bw).unwrap()),
+            ops: if let Some(ops) = state.ops.as_ref() {
+                Some(TokenBucket::restore((), ops)?)
+            } else {
+                None
+            },
```
serban300, line 78: Nit: This can also be done with:
```
state
                .ops
                .as_ref()
                .map(|ops| TokenBucket::restore((), ops))
                .transpose()?
```

Issue #745, opened 2018-12-05
Improve 'firecracker --version`


rn:
`firecracker` built from HEAD of master (commit `497b4fe358c5`) shows:
```
% ./firecracker --version
firecracker 0.11.0
```
This is clearly a few commits ahead of the 0.11.0 release.

It would be better if it reports something like `0.11.0+ <git hash>` to easier identify the version (note the `+`), or just the commit hash for none releases . If `firecracker` was build from a dirty repository (ie local, un-committed changes) indicating this with `dirty` would also help.

aghecenco:
Would the output of `git describe` do the trick?

Release commit 3ad0b7b6284853f3b38d1ec7c2d81e1fccfcddfd :
```bash
git describe 3ad0b7b6284853f3b38d1ec7c2d81e1fccfcddfd
v0.12.0
```

Following commit d9c0086bda99e5ec183eb80feccee2f158647fea :
```bash
git describe d9c0086bda99e5ec183eb80feccee2f158647fea
v0.12.0-1-gd9c0086
```

rn:
yeah, that looks good. I would maybe also add a `-dirty` if build from a repository which does not have uncommitted changes. The reliable way to get this is with something like this:
```
git update-index -q --refresh && git diff-index --quiet HEAD -- . || echo "-dirty"
```
[the `update-index` is necessary in case the timestamp (but no the content) of a local file changed]

61131:
Within another code base, I use the following for generating a short description for version reporting:

`GIT_VERSION:= $(shell git --no-pager describe --tags --always --dirty)`

For a current checkout, it reports the version as:

`v0.12.0-26-g608aa56`

aghecenco:
@rn I'm confused, are repos **without** uncommitted changes dirty? I was thinking the opposite.

Jibbow:
There is a library which does exactly that: https://docs.rs/git-version/0.3.2/git_version/index.html
However, quickly pulling it into Firecracker gives me a `No such file or directory` error inside the macro. Should I investigate this error further or is this the wrong way to go anyways because this adds a new dependency?

aghecenco:
See also https://github.com/cloud-hypervisor/cloud-hypervisor/pull/746/commits/24195b01026890696a5902543076ba42617c6ae3

vbmade2000:
May be https://docs.rs/built/0.4.2/built/ cab help ?

alindima:
Issue was fixed by PR #2061

PR #2061, opened 2020-08-04
Improve firecracker --version output

alindima:
## Reason for This PR

Fix for #745 

Improving version output to include number of commits over latest tag, commit hash and dirty flag.

## Description of Changes

Added a build script that, on every `devtool build` sets the FIRECRACKER_VERSION environment variable to the output of "git describe --dirty" command.

This is a special environment variable which is embedded in the executable code.

## License Acceptance

## PR Checklist


raduiliescu:
> 
> 
> I like the idea and the approach, but in its current form this PR would make our release binaries also report commit hash instead of plain version numbers:
> 
> `0.21.0-442-gd006277` instead of `v0.21.0` as it used to be.
> 
> Do we want to change our release version numbering policy like so, or should add functionality to this PR to treat releases differently?

I would use --verbose parameter to --version and only there add info on branches and githash on top of current info.

alindima:
Thank you for your feedback.
I agree with you, maybe it would be best if the release binary would show only the simplified version number from `Cargo.toml`.

I also agree with @raduiliescu that it would be useful to have a verbose parameter for the version, just in case somebody needs it.

Are there any concerns regarding this approach?

acatangiu:
> I would use --verbose parameter to --version and only there add info on branches and githash on top of current info.

That should work, but it makes it harder to see whether a particular binary is a release or a custom build. `--version` would be the same for both, while `--verbose` would differ only by commit hash so it requires the user to go to the repo and compare hashes to find out if that binary is a release binary or a custom build.

I'd like it if we could use `git tags` to make that differentiation and do something like: use `--version` which outputs:
 - for a release binary (local `HEAD` is at a particular `vX.Y.Z` tag): output is `vX.Y.Z`
 - for other builds (local `HEAD` does not match any release tag): output is `vX.Y.Z-branch-commit[-dirty]`

WDYT?

alindima:
> > I would use --verbose parameter to --version and only there add info on branches and githash on top of current info.
> 
> That should work, but it makes it harder to see whether a particular binary is a release or a custom build. `--version` would be the same for both, while `--verbose` would differ only by commit hash so it requires the user to go to the repo and compare hashes to find out if that binary is a release binary or a custom build.
> 
> I'd like it if we could use `git tags` to make that differentiation and do something like: use `--version` which outputs:
> 
> * for a release binary (local `HEAD` is at a particular `vX.Y.Z` tag): output is `vX.Y.Z`
> * for other builds (local `HEAD` does not match any release tag): output is `vX.Y.Z-branch-commit[-dirty]`
> 
> WDYT?

Well, actually, the behaviour that is currently implemented in this PR is the one you mentioned.

The `git describe --dirty` command outputs just vX.Y.Z if `HEAD` matches a tag.
The command only appends the commit and dirty flags if it does not match a tag.

alindima:
I have found that there might be something fundamentally flawed about the approach.

The `build.rs` script that I have introduced is per-package. That means that it runs when cargo detects that a change has occurred in that package. The package in this case is `firecracker`.

However, there are changes that should trigger the re-run even if they didn't happen inside the firecracker crate. For example, an update in the `jailer` crate should also potentially be reflected in the version number of the firecracker crate.

One way of fixing this is by using in the `build.rs` script an instruction like `cargo:rerun-if-changed=PATH`. Still, this is not enough because it only does a shallow check (according to the [docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath)). Therefore, if a file from a nested folder is changed, this wouldn't be reflected in the version number. This could be mitigated by making the script loop through all the files and folders recursively, adding each of them to a separate `cargo:rerun-if-changed=PATH` instruction, while also checking the .gitignore file. (This could also be a build-time performance impact and seems like a great overkill).

This could be fixed if there were a single top-level crate. We could then define the build.rs script to watch for changes in the entire package. Currently, since the top-level `Cargo.toml` only defines two separate workspaces, this isn't possible.

This problem is rooted in the fact that the firecracker crate version and the firecracker repository version are not the same thing, since the repo also includes other crates, like the jailer for example.

Some help and feedback on this would be greatly appreciated.

alindima:
The team and I synced offline on this issue.

The behaviour of "git describe --dirty" is actually the behaviour we expect; that is, when a binary is build from a release tag, the output returned by the command is just the tag (`vX.Y.Z`) and only when a binary is build from a repo that has some commits over a release tag or some dirty (modified and uncommitted) files does this command output the long version (`vX.Y.Z-nr-hash-[dirty]`). This is actually what we want.

In regards to the implementation, the concerns outlined in the previous comment still remain.

I have explored on the latest commit an option that does a `touch` command on the `build.rs` file as the last step of the build script. This artificially updates the time when the build script was modified and automatically triggers a re-run on every build.

This approach seems to work and has minimal impact.

WDYT?

sandreim:
> The team and I synced offline on this issue.
> 
> The behaviour of "git describe --dirty" is actually the behaviour we expect; that is, when a binary is build from a release tag, the output returned by the command is just the tag (`vX.Y.Z`) and only when a binary is build from a repo that has some commits over a release tag or some dirty (modified and uncommitted) files does this command output the long version (`vX.Y.Z-nr-hash-[dirty]`). This is actually what we want.
> 
> In regards to the implementation, the concerns outlined in the previous comment still remain.
> 
> I have explored on the latest commit an option that does a `touch` command on the `build.rs` file as the last step of the build script. This artificially updates the time when the build script was modified and automatically triggers a re-run on every build.
> 
> This approach seems to work and has minimal impact.
> 
> WDYT?

This is clearly an improvement. I have no other concerns to add.

src/firecracker/build.rs
```diff
@@ -19,6 +19,8 @@ fn main() {
             env!("CARGO_PKG_VERSION")
         ),
     };
+
+    Command::new("touch").args(&["build.rs"]).output().unwrap();
```
acatangiu, line 23: This basically `touch`es itself as part of running, right? Does cargo consistently pick up the timestamp as _newer than last build_ even though _last build_ is basically still in progress?

If this is reliable, then fine. Otherwise we can put this `touch` in `tools/devtool` _before_ running the `cargo build [--release]` command.

alindima: As we discussed offline, this should be reliable since cargo should cache the timestamp when the script started executing, hence this touch would be newer than last build.
And it works as expected

alindima:
Putting this on hold while we discuss a testing strategy

alindima:
After syncing with the team, we decided for now, to remove the integration test for the `--version` flag

src/firecracker/build.rs
```diff
@@ -0,0 +1,39 @@
...
+
+// this build script is called on every firecracker package build,
+// embedding the FIRECRACKER_VERSION directly in the resulting binary
+fn main() {
+    let git_version_command = Command::new("git").args(&["describe", "--dirty"]).output();
+
+    match git_version_command {
+        Ok(result) => println!(
+            "cargo:rustc-env=FIRECRACKER_VERSION={}",
+            remove_leading_v(String::from_utf8(result.stdout).unwrap())
```
serban300, line 14: Can we do `result.stdout[1..]` instead of `remove_leading_v` ?

serban300: Or `result.trim_start_matches`

alindima: the `remove_leading_v` function takes into account the UTF-8 encoding. By simply doing `result.stdout[1..]`, we assume that the command returns ASCII characters.

However, the `trim_start_matches` looks perfect. I wasn't aware of this method

tests/host_tools/cargo_build.py
```diff
@@ -45,7 +45,7 @@ def cargo_test(path, extra_args=''):
     path = os.path.join(path, CARGO_UNITTEST_REL_PATH)
     cmd = 'CARGO_TARGET_DIR={} RUST_TEST_THREADS=1 RUST_BACKTRACE=1 ' \
           'RUSTFLAGS="{}" cargo test {} --all --no-fail-fast'.format(
-            path, get_rustflags(), extra_args)
+              path, get_rustflags(), extra_args)
```
dianpopa, line 48: Was this needed?

src/firecracker/build.rs
dianpopa: If I understand correctly this is a workaround for triggering the `build.rs` every time `cargo build` is called, right?

tests/host_tools/cargo_build.py
```diff
@@ -55,24 +55,29 @@ def get_firecracker_binaries():
     Returns the location of the firecracker related binaries eventually after
     building them in case they do not exist at the specified root_path.
     """
+    if not hasattr(get_firecracker_binaries, "was_built"):
```
dianpopa, line 58: If any source file changes and we do `./tools/devtool build`, will the build get trigger again?

tests/host_tools/cargo_build.py
alindima: in my opinion, not. I see that the rust formatter in VSCode added this automatically

tests/host_tools/cargo_build.py
alindima: yes. this attribute lives as long as the script execution. when we do a new `tools/devtool build`, the build will get triggered again.

This was needed because until now, this function did a `cargo build` on every call. However, since we now update the `build.rs` on every build, this was taking too long and some tests were failing with timeout.

src/firecracker/build.rs
alindima: done

tests/host_tools/cargo_build.py
acatangiu: The thing I don't like about this is: with this PR _every time_ I run `pytest` or `devtool test` when doing development, it is going to do a top-level `release` build. That time can really add up during a development session where you run a test repeatedly.

I'd like to have a way to either:
- switch this logic off (based on env variable toggled from `devtool`: **on** for `build`, **off** for everything else `test/shell/etc`)
- use something mechanism independent of `cargo`/`build.rs` so that we get finer control over when it is enabled or not (for example we want it in `devtool build` but not in `devtool test`)
- some other idea to avoid rebuilding each time we instantiate `pytest`

tests/host_tools/cargo_build.py
```diff
@@ -55,24 +55,29 @@ def get_firecracker_binaries():
...
-        "strip --strip-debug {} {}"
-        .format(fc_bin_path, jailer_bin_path)
-    )
+    if not get_firecracker_binaries.was_built:
+        utils.run_cmd(cmd)
+
+        utils.run_cmd(
+            "strip --strip-debug {} {}"
+            .format(fc_bin_path, jailer_bin_path)
+        )
```
acatangiu, line 79: Why do we `strip` binaries used in tests anyways? This step adds unnecessary time to tests.

@firecracker-microvm/compute-capsule I say we remove it.

src/firecracker/build.rs
```diff
@@ -0,0 +1,31 @@
...
+    } else {
+        // if this is not a git repo, return the standard cargo package version
+
+        println!(
+            "cargo:rustc-env=FIRECRACKER_VERSION={}",
+            env!("CARGO_PKG_VERSION")
+        );
+    }
+
+    Command::new("touch").args(&["build.rs"]).output().unwrap();
```
acatangiu, line 30: This line is better suited in `devtool` on the `build` path.

We don't want this mechanism to always re-trigger even in tests or local development (we don't want to wait for repeated unnecessary `release` linking each time we run a `pytest` test).

`devtool build [--release]` can `touch` this and force the binary being rebuilt and having the right `version`.
All other `devtool` paths (`test/shell`) should not _force_ rebuilding the binary.

WDYT?

alindima: I agree with you, but what if someody does a build using `cargo build` instead of using devtool?
I figured that implementing it this way, this would be devtool-agnostic.
WDYT?

tests/host_tools/cargo_build.py
alindima: With this PR, cargo is only going to recompile and rerun the build script, since nothing else is changed in the source files.

I have run the `time tools/devtool build` command to see how much it affects build time and it is only `2.785s`.

Is this really a concern?

src/firecracker/build.rs
acatangiu: Local (non-devtool/non-container) builds should never reach clients, all our automated scripts and runbooks for manual steps work with `devtool`.

I would like to support `cargo build` to _always_ update the right version, but not at the expense on development speed.

tests/host_tools/cargo_build.py
acatangiu: Can you do a quick profile of the `get_firecracker_binaries()` function when running `pytest <some-test>` with and without the rebuild? Ideally not new vs old code as old code had unnecessary `strip`.

I would like to compare two runs where the code hasn't changed (a rebuild should not trigger) (without `strip`):
1. with the `touch build.rs` which will force re-trigger build
2. without `touch build.rs` which will be basically a no-op

If the difference is 1-2 seconds, then I am fine with it, but I was under the impression that the binary is _always_ rebuilt (including linking) which on `--release` (such as tests) is considerable.

acatangiu: So I've done some testing:

This is how this PR behaves in its current form (where `touch src/firecracker/build.rs` always happens):

```
acatan@uf6ed9c851f4556:~/code/firecracker$ git log -2 --oneline --decorate
321433b (HEAD -> improve_version, alin/improve_version) Merge branch 'master' into improve_version
a2a6822 Improve firecracker --version output. Update build script to build the binaries only once.

acatan@uf6ed9c851f4556:~/code/firecracker$ cargo build --release
Updating crates.io index
  Downloaded versionize v0.1.2
   Compiling libc v0.2.68
...
   Compiling api_server v0.1.0 (/home/ANT.AMAZON.COM/acatan/code/firecracker/src/api_server)
    Finished release [optimized] target(s) in 1m 36s

acatan@uf6ed9c851f4556:~/code/firecracker$ cargo build --release
   Compiling firecracker v0.21.0 (/home/ANT.AMAZON.COM/acatan/code/firecracker/src/firecracker)
    Finished release [optimized] target(s) in 20.70s

acatan@uf6ed9c851f4556:~/code/firecracker$ cargo build --release
   Compiling firecracker v0.21.0 (/home/ANT.AMAZON.COM/acatan/code/firecracker/src/firecracker)
    Finished release [optimized] target(s) in 20.74s
```

**Without changing any files** in the repository, each `build --release` **takes 20.74s** on my local machine (which is faster than a EC2 m5d.metal).

Removing the forced `touch src/firecracker/build.rs`:

```
acatan@uf6ed9c851f4556:~/code/firecracker$ git diff
diff --git a/src/firecracker/build.rs b/src/firecracker/build.rs
index 83cbeec..7bb283c 100644
--- a/src/firecracker/build.rs
+++ b/src/firecracker/build.rs
@@ -27,5 +27,4 @@ fn main() {
         );
     }
 
-    Command::new("touch").args(&["build.rs"]).output().unwrap();
 }

acatan@uf6ed9c851f4556:~/code/firecracker$ cargo build --release
   Compiling firecracker v0.21.0 (/home/ANT.AMAZON.COM/acatan/code/firecracker/src/firecracker)
    Finished release [optimized] target(s) in 20.68s

acatan@uf6ed9c851f4556:~/code/firecracker$ cargo build --release
    Finished release [optimized] target(s) in 0.05s

acatan@uf6ed9c851f4556:~/code/firecracker$ cargo build --release
    Finished release [optimized] target(s) in 0.04s
```

This is correct behavior, nothing changed, iterative `cargo build --release` take **0.04s**.

tests/host_tools/cargo_build.py
dianpopa: If we want our release binaries to be stripped, we also need to run the CI with them in this form.

tests/host_tools/cargo_build.py
alindima: Indeed. I was only taking into account the `devtool build` command, without the `--release` flag.
I will have to update the implementation to account for this. 20 seconds is not acceptable.

I will adapt the implementation to do a `touch build.rs` only from `devtool/build --release`

tests/host_tools/cargo_build.py
acatangiu: In theory, yes.
But we could also make the simplifying assumption that the `strip` binary in the container that we control is correct and does not corrupt the target binary.

This `strip` step/command adds 2 seconds to each `pytest` run (this cost is felt most during test development when we do `pytest` over and over with a single test), so it's not that bad. I would like to not pay those 2 seconds, but at the same time it's not such a big price to pay if we don't trust `strip` and want to verify.

src/firecracker/build.rs
alindima: done. added this `touch` command in devtool on the build path

src/firecracker/build.rs
```diff
@@ -0,0 +1,31 @@
...
+
+// this build script is called on every change in
+// the firecracker crate source files, on every `devtool build --release`
+// and on every change in this file,
+// embedding the FIRECRACKER_VERSION directly in the resulting binary
+fn main() {
+    let git_version_command = Command::new("git")
+        .args(&["describe", "--dirty"])
+        .output()
+        .unwrap();
```
serban300, line 14: We should also handle this possible panic.

andreeaflorescu:
Does this change have any impact on reproducible builds?

tools/devtool
```diff
@@ -499,6 +499,9 @@ cmd_build() {
 
     [ $profile = "release" ] && cargo_args+=("--release")
 
+    # Artificially trigger the build script to always get the latest `git describe` output
```
andreeaflorescu, line 502: If we move the build.rs to the workspace level we might not need this hack anymore.

alindima: done

src/firecracker/build.rs
```diff
@@ -0,0 +1,31 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
alindima: done in this new revision

CHANGELOG.md
```diff
@@ -82,6 +82,7 @@
 - Segregated MMDS documentation in MMDS design documentation and MMDS user
   guide documentation.
 - The logger `level` field is now case-insensitive.
+- Updated firecracker --version flag to include `git describe --dirty` output, where available
```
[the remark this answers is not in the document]
alindima: fixed

src/firecracker/build.rs
alindima: handled

alindima:
> Does this change have any impact on reproducible builds?

As long as the folder containing the sources has the git directory and information, reproducible builds should not be affected

src/firecracker/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "firecracker"
 version = "0.21.0"
 authors = ["Amazon Firecracker team <firecracker-devel@amazon.com>"]
 edition = "2018"
+build = "../../build.rs"
```
acatangiu, line 6: Shouldn't we do this for `jailer` too?

CHANGELOG.md
```diff
@@ -82,6 +82,9 @@
 - Segregated MMDS documentation in MMDS design documentation and MMDS user
   guide documentation.
 - The logger `level` field is now case-insensitive.
+- Updated firecracker --version flag to include `git describe --dirty` output,
```
acatangiu, line 85: If we do it for `jailer` as well, update entry here too.

build.rs
```diff
@@ -0,0 +1,35 @@
...
+                        .unwrap()
+                        .trim_start_matches('v')
+                );
+            } else {
+                // if this is not a git repo, return the standard cargo package version
+
+                println!(
+                    "cargo:rustc-env=FIRECRACKER_VERSION={}",
+                    env!("CARGO_PKG_VERSION")
+                );
```
serban300, line 27: This code block is duplicated. Also `"cargo:rustc-env=FIRECRACKER_VERSION={}"` is duplicated a couple of times.

How about something like:

```
    let firecracker_version = Command::new("git")
        .args(&["describe", "--dirty"])
        .output()
        .ok()
        .and_then(|output| {
            if output.status.success() {
                return Some(output.stdout);
            }
            None
        })
        .and_then(|version_bytes| {
            String::from_utf8(version_bytes).ok()
        })
        .and_then(|version_string| {
            Some(version_string.trim_start_matches('v').to_string())
        })
        .unwrap_or(env!("CARGO_PKG_VERSION").to_string());

    println!("cargo:rustc-env=FIRECRACKER_VERSION={}", firecracker_version);
```

If I understand correctly, this is run at build time so performance is not extremely important, but we can also play with `String` and `&str` a bit to optimize it.

build.rs
```diff
@@ -0,0 +1,35 @@
...
+fn main() {
+    let git_version_command = Command::new("git").args(&["describe", "--dirty"]).output();
+
+    match git_version_command {
+        Ok(outcome) => {
+            if outcome.status.success() {
+                println!(
+                    "cargo:rustc-env=FIRECRACKER_VERSION={}",
+                    String::from_utf8(outcome.stdout)
+                        .unwrap()
```
serban300, line 18: I just realized this is another possible panic. In theory it shouldn't happen, but it would be better to handle it.

src/firecracker/Cargo.toml
alindima: This comment is outdated. In the current state of the PR, it is not possible, since both of these versions would be identical and would point to the same firecracker version number

build.rs
alindima: done

alindima:
I have yet to investigate for a git command in order to also set the dirty flag when there are any new, unstaged and uncommited files added.

The `git describe --dirty` command does not report this as dirty.

WDYT? Is this what we want?

alindima:
> I have yet to investigate for a git command in order to also set the dirty flag when there are any new, unstaged and uncommited files added.
> 
> The `git describe --dirty` command does not report this as dirty.
> 
> WDYT? Is this what we want?

I have added in the latest commit a bash command that has the desired behaviour. What do you think?

build.rs
```diff
@@ -0,0 +1,36 @@
...
+        // the dirty flag reported by git describe --dirty does not suffice,
+        // since it is not affected by new, unstaged files;
+        // we check the output of git status to account for this case.
+        .args(&[
+            "-c",
+            "err=0; trap \"err=1\" ERR; git describe >/dev/null;\
+            echo -n \"$(git describe)\";\
+            [[ $(git status --porcelain | wc -l) -eq 0 ]] && : || echo '-dirty';\
+            test $err = 0",
+        ])
```
acatangiu, line 19: This looks complicated enough that I actually prefer just the simple `git describe --dirty` even if it misses untracked files.
From my point of view, this mechanism is to help us avoid _accidentally_ confusing a dirty binary with a release one. _Accidentally_ having untracked files that cleanly compile is highly improbable and I'm not worried about covering it.

The above is just a preference, I am also fine with the current "complicated" version.

acatangiu (approved):

build.rs
dianpopa: I am also in favor of the less complicated output (i.e "git describe only").

build.rs
```diff
@@ -0,0 +1,36 @@
...
+            if output.status.success() {
+                return Some(output.stdout);
+            }
+            None
+        })
+        .and_then(|version_bytes| String::from_utf8(version_bytes).ok())
+        .map(|version_string| version_string.trim_start_matches('v').to_string())
+        .unwrap_or_else(|| env!("CARGO_PKG_VERSION").to_string());
+
+    println!(
```
dianpopa, line 32: We should add another line with
```
println!(
        "cargo:rustc-env=JAILER_VERSION={}",
        firecracker_version
    );
```
and then update [this](https://github.com/alindima/firecracker/blob/improve_version/src/jailer/src/main.rs#L19) accordingly.

CHANGELOG.md
dianpopa: I d rephrase this entry to something more concise and clear:

"`firecracker --version` gets updated on each `devtool build` to the output of `git describe --dirty`.
Also, if we keep the long version of the command (i.e porcelain stuff) this is not available anymore.

build.rs
alindima: It seems that this is the general opinion, so I will roll back to `git describe --version`

build.rs
alindima: This opens another discussion. Doesn't it make sense then to have a single environment variable, to avoid duplicating this? Something like "REPO_VERSION" or just "FIRECRACKER_VERSION" for both jailer and firecracker (since the repo is named firecracker and the jailer is only a sub-crate). WDYT?

acatangiu: You're right, since we're basing on a `git` command, the result will always be repo-level, so using a single env variable makes sense - `REPO_VERSION` and `FIRECRACKER_VERSION` are both fine.

Don't forget to consume the env var in the `jailer` binary as well.

alindima: done in latest commit

CHANGELOG.md
alindima: edited

dianpopa (approved):

acatangiu (approved):

PR #2067, opened 2020-08-06
Add cross-version save/load integration tests.

### LINKED ISSUE #2077: Intermittent fail in CI - test_concurrency.py::test_run_concurrency
```bash
____________ test_run_concurrency[ubuntu_with_ssh, 20 instance(s)] _____________
integration_tests/functional/test_concurrency.py:24: in test_run_concurrency
    _ = net_tools.SSHConnection(microvm.ssh_config)
host_tools/network.py:39: in __init__
    self._init_connection()
<decorator-gen-2>:2: in _init_connection
    ???
/usr/local/lib/python3.6/dist-packages/retry/api.py:74: in retry_decorator
    logger)
/usr/local/lib/python3.6/dist-packages/retry/api.py:33: in __retry_internal
    return f()
host_tools/network.py:74: in _init_connection
    raise ConnectionError
E   ConnectionError
```

sandreim:
## Reason for This PR

This completes the test suite mentioned in issue #1850. 
Also fixes #2077 and #2075 - collateral from snapshot networking config changes.

## Description of Changes

Add two snapshot version integration tests:
- Use current version to load snapshots created with older versions.
- Create snapshots with current versions and load in older releases.

Framework changes:
- added binary path parameters in init_microvm()
- added FirecrackerArtifact class
- added SnapshotArtifact class
- added support to specify the fc version when creating a snapshot
- use static network configuration for snapshots

More details in commit messages.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -128,20 +128,26 @@ def _test_images_s3_bucket():
 MICROVM_S3_FETCHER = MicrovmImageS3Fetcher(_test_images_s3_bucket())
 
 
-def init_microvm(root_path, bin_cloner_path):
+def init_microvm(root_path, bin_cloner_path,
+                 fc_binary=None, jailer_binary=None):
     """Auxiliary function for instantiating a microvm and setting it up."""
     # pylint: disable=redefined-outer-name
     # The fixture pattern causes a pylint false positive for that rule.
     microvm_id = str(uuid.uuid4())
-    fc_binary, jailer_binary = build_tools.get_firecracker_binaries()
+
+    if fc_binary is None:
+        fc_binary, jailer_binary = build_tools.get_firecracker_binaries()
+
+    # Make sure we always have both binaries.
+    assert fc_binary
+    assert jailer_binary
 
     vm = Microvm(
-        resource_path=root_path,
-        fc_binary_path=fc_binary,
-        jailer_binary_path=jailer_binary,
-        microvm_id=microvm_id,
-        bin_cloner_path=bin_cloner_path
-    )
+         resource_path=root_path,
+         fc_binary_path=fc_binary,
+         jailer_binary_path=jailer_binary,
+         microvm_id=microvm_id,
+         bin_cloner_path=bin_cloner_path)
     vm.setup()
     return vm
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -135,7 +135,7 @@ def init_microvm(root_path, bin_cloner_path,
     # The fixture pattern causes a pylint false positive for that rule.
     microvm_id = str(uuid.uuid4())
 
-    if fc_binary is None:
+    if fc_binary is None or jailer_binary is None:
         fc_binary, jailer_binary = build_tools.get_firecracker_binaries()
 
     # Make sure we always have both binaries.
--- a/tests/framework/artifacts.py
+++ b/tests/framework/artifacts.py
@@ -166,7 +166,7 @@ class SnapshotArtifact:
         # Get the name of the snapshot folder.
         snapshot_name = self.name
         self._local_folder = os.path.join(ARTIFACTS_LOCAL_ROOT,
-                                          self.type.strip('/'),
+                                          self.type.value,
                                           snapshot_name)
 
     @property
@@ -349,7 +349,7 @@ class ArtifactCollection:
             # Select only files with specified keyword.
             if (key[-1] == "/" and key != prefix and
                (keyword is None or keyword in snapshot_dir.key)):
-                artifact_type = ArtifactCollection.ARTIFACTS_SNAPSHOTS
+                artifact_type = ArtifactType.SNAPSHOT
                 artifacts.append(SnapshotArtifact(self.bucket,
                                                   key,
                                                   artifact_type=artifact_type))
--- a/tests/framework/artifacts.py
+++ b/tests/framework/artifacts.py
@@ -160,7 +160,6 @@ class SnapshotArtifact:
         for disk in snaphot_disks:
             artifact = Artifact(self._bucket, disk.key,
                                 artifact_type=ArtifactType.DISK)
-            # artifact.download()
             self._disks.append(artifact)
 
         # Get the name of the snapshot folder.
--- a/tests/framework/artifacts.py
+++ b/tests/framework/artifacts.py
@@ -14,6 +14,9 @@ import botocore.client
 from host_tools.snapshot_helper import merge_memory_bitmaps
 
 
+ARTIFACTS_LOCAL_ROOT = "/tmp/ci-artifacts"
+
+
 class ArtifactType(Enum):
     """Supported artifact types."""
 
@@ -24,6 +27,9 @@ class ArtifactType(Enum):
     MEM = "mem"
     VMSTATE = "vmstate"
     MISC = "misc"
+    SNAPSHOT = "snapshot"
+    FC = "firecracker"
+    JAILER = "jailer"
 
 
 class Artifact:
@@ -61,6 +67,10 @@ class Artifact:
         """Return the artifact name."""
         return Path(self.key).name
 
+    def base_name(self):
+        """Return the base name (without extension)."""
+        return Path(self.key).stem
+
     def local_dir(self):
         """Return the directory containing the downloaded artifact."""
         assert self._local_folder is not None
@@ -70,7 +80,7 @@ class Artifact:
             platform.machine(),
         )
 
-    def download(self, target_folder, force=False):
+    def download(self, target_folder=ARTIFACTS_LOCAL_ROOT, force=False):
         """Save the artifact in the folder specified target_path."""
         assert self.bucket is not None
         self._local_folder = target_folder
@@ -88,7 +98,7 @@ class Artifact:
             self.name()
         )
 
-    def copy(self):
+    def copy(self, file_name=None):
         """Create a writeable copy of the artifact."""
         assert os.path.exists(self.local_path()), """File {} not found.
         call download() first.""".format(self.local_path())
@@ -99,10 +109,16 @@ class Artifact:
             self.type.value,
             platform.machine()
         )
-        # The temp file suffix is the artifact type.
-        suffix = "-{}".format(self.type.value)
-        # The key for the new artifact is the full path to the file.
-        new_key = tempfile.mktemp(dir=new_dir, suffix=suffix)
+
+        if file_name is None:
+            # The temp file suffix is the artifact type.
+            suffix = "-{}".format(self.type.value)
+            # The key for the new artifact is the full path to the file.
+            new_key = tempfile.mktemp(dir=new_dir, suffix=suffix)
+        else:
+            # Caller specified new name.
+            new_key = os.path.join(new_dir, file_name)
+
         # Create directories if needed.
         Path(new_dir).mkdir(parents=True, exist_ok=True)
         # Copy to local artifact.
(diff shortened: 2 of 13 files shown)


tests/conftest.py
```diff
@@ -128,20 +128,26 @@ def _test_images_s3_bucket():
...
-def init_microvm(root_path, bin_cloner_path):
+def init_microvm(root_path, bin_cloner_path,
+                 fc_binary=None, jailer_binary=None):
     """Auxiliary function for instantiating a microvm and setting it up."""
     # pylint: disable=redefined-outer-name
     # The fixture pattern causes a pylint false positive for that rule.
     microvm_id = str(uuid.uuid4())
-    fc_binary, jailer_binary = build_tools.get_firecracker_binaries()
+
+    if fc_binary is None:
```
gc-plp, line 138: Shouldn't we also check for `jailer_binary` being `None`? 

If `jailer_binary` is `None` and `fc_binary` not `None` then it would fail at line `143`.

I think that we can either split the `get_firecracker_binaries` function into separate functions for each binary or do the following:

```
default_fc_binary, default_jailer_binary = build_tools.get_firecracker_binaries()

if not fc_binary:
    fc_binary = default_fc_binary

if not jailer_binary:
    jailer_binary = default_jailer_binary
```

tests/framework/artifacts.py
```diff
@@ -118,8 +134,117 @@ def copy(self):
...
+                                 artifact_type=ArtifactType.SSH_KEY)
+        self._disks = []
+
+        disk_prefix = "{}disk".format(key)
+        snaphot_disks = self._bucket.objects.filter(Prefix=disk_prefix)
+
+        for disk in snaphot_disks:
+            artifact = Artifact(self._bucket, disk.key,
+                                artifact_type=ArtifactType.DISK)
+            # artifact.download()
```
gc-plp, line 163: Should this have been deleted? :)

tests/framework/artifacts.py
```diff
@@ -118,8 +134,117 @@ def copy(self):
...
+            self._disks.append(artifact)
+
+        # Get the name of the snapshot folder.
+        snapshot_name = self.name
```
gc-plp, line 167: Does `self` have the `name` attribute?

gc-plp, line 167:
```suggestion
        self._local_folder = os.path.join(ARTIFACTS_LOCAL_ROOT,
                                          self.type.strip('/'),
                                          self.name)
```

tests/framework/artifacts.py
```diff
@@ -178,6 +322,40 @@ def _fetch_artifacts(self,
...
+        prefix += ArtifactCollection.ARTIFACTS_SNAPSHOTS
+        snaphot_dirs = self.bucket.objects.filter(Prefix=prefix)
+        for snapshot_dir in snaphot_dirs:
+            key = snapshot_dir.key
+            # Filter out the snapshot artifacts root folder.
+            # Select only files with specified keyword.
+            if (key[-1] == "/" and key != prefix and
+               (keyword is None or keyword in snapshot_dir.key)):
+                artifact_type = ArtifactCollection.ARTIFACTS_SNAPSHOTS
+                artifacts.append(SnapshotArtifact(self.bucket,
```
gc-plp, line 353: `artifact_type` is not the same type as the `artifact_type` parameter declared in the `SnapshotArtifact` constructor.

tests/framework/artifacts.py
```diff
@@ -118,8 +134,117 @@ def copy(self):
...
+        for disk in snaphot_disks:
+            artifact = Artifact(self._bucket, disk.key,
+                                artifact_type=ArtifactType.DISK)
+            # artifact.download()
+            self._disks.append(artifact)
+
+        # Get the name of the snapshot folder.
+        snapshot_name = self.name
+        self._local_folder = os.path.join(ARTIFACTS_LOCAL_ROOT,
+                                          self.type.strip('/'),
```
gc-plp, line 169: If `self.type` returns the default `ArtifactType.SNAPSHOT` value, the return value does not have `strip()` as an attribute.

tests/framework/artifacts.py
```diff
@@ -178,6 +322,40 @@ def _fetch_artifacts(self,
...
+        # Snapshot folder /ci-artifacts/snapshots/x86_64/fc_snapshot_v0.22:
+        # - vm.mem
+        # - vm.vmstate
+        # - disk0 <---- this is the root disk
+        # - disk1
+        # - diskN
+        # - ssh_key
+
+        artifacts = []
+        prefix = ArtifactCollection.ARTIFACTS_ROOT
```
gc-plp, line 343: Could this be initialized in one line?
```
prefix = ArtifactCollection.ARTIFACTS_ROOT + ArtifactCollection.ARTIFACTS_SNAPSHOTS
``` 
?

tests/framework/artifacts.py
```diff
@@ -178,6 +322,40 @@ def _fetch_artifacts(self,
...
+        prefix += ArtifactCollection.ARTIFACTS_SNAPSHOTS
+        snaphot_dirs = self.bucket.objects.filter(Prefix=prefix)
+        for snapshot_dir in snaphot_dirs:
+            key = snapshot_dir.key
+            # Filter out the snapshot artifacts root folder.
+            # Select only files with specified keyword.
+            if (key[-1] == "/" and key != prefix and
```
gc-plp, line 350: Comparing `key[-1] with "/"` is a bit unclear. Is the purpose of this comparison to check if it's a folder?

tests/framework/microvms.py
```diff
@@ -0,0 +1,126 @@
...
+# pylint: disable=R0903
+class VMBase:
+    """Base class for constructing microvms."""
+
+    @classmethod
+    def from_artifacts(cls, bin_cloner_path, config,
+                       kernel, disks, cpu_template, start=False):
+        """Spawns a new Firecracker and applies specified config."""
+        artifacts = ArtifactCollection(_test_images_s3_bucket())
+        config = artifacts.microvms(keyword=config)[0]
```
gc-plp, line 60: Could you add a comment explaining why `[0]` is used?

tests/framework/microvms.py
```diff
@@ -0,0 +1,126 @@
...
+        return VmInstance(config, kernel, attached_disks, ssh_key, basevm)
+
+
+class C3nano(VMBase):
+    """Create VMs with 2vCPUs and 256 MB RAM."""
+
+    @classmethod
+    def spawn(cls, bin_cloner_path, start=False):
+        """Spawns and optionally starts the vm."""
+        return cls.from_artifacts(bin_cloner_path,
```
gc-plp, line 100: We should call the constructor by specifying the class parameters explicitly, in order to make it easier to read.

tests/framework/microvms.py
```diff
@@ -0,0 +1,126 @@
...
+                                  start)
+
+
+class C3micro(VMBase):
+    """Create VMs with 2vCPUs and 512 MB RAM."""
+
+    @classmethod
+    def spawn(cls, bin_cloner_path, start=False):
+        """Spawns and optionally starts the vm."""
+        return cls.from_artifacts(bin_cloner_path,
```
gc-plp, line 114: We should call the constructor by specifying the class parameters explicitly, in order to make it easier to read.

tests/framework/builder.py
```diff
@@ -24,7 +24,6 @@ def __init__(self, bin_cloner_path,
         self.bin_cloner_path = bin_cloner_path
         self.init_root_path()
 
-        # Update permissions to owner RWX.
```
gc-plp: Can we leave the comment? :)

tests/integration_tests/functional/test_snapshot_version.py
```diff
@@ -0,0 +1,122 @@
...
+from framework.builder import MicrovmBuilder, SnapshotBuilder, SnapshotType
+from framework.microvms import C3micro
+import host_tools.network as net_tools  # pylint: disable=import-error
+
+
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
+    reason="Not supported yet."
+)
+def test_restore_from_past_versions(bin_cloner_path):
```
serban300, line 19: For the moment we don't have cross Intel families snapshot restore. So this test may fail depending on the processor model. For example on my Intel laptop it fails. Also creating snapshot artificats seems unnecessary for the moment. In order to prevent machine-dependent fails and unnecessary complexity, wouldn't it be better to just create a snapshot using the old firecracker versions (that are present as artifacts in S3) and restore it using the current version ? This would also make it easier to test cross-version snapshot restore on AMD.

tests/framework/artifacts.py
sandreim: yup.

tests/framework/artifacts.py
```diff
@@ -138,14 +278,18 @@ class ArtifactCollection:
...
     ARTIFACTS_ROOT = 'ci-artifacts'
-    ARTIFACTS_DISKS = '/disks/' + PLATFORM
-    ARTIFACTS_KERNELS = '/kernels/' + PLATFORM
-    ARTIFACTS_MICROVMS = '/microvms'
-    ARTIFACTS_SNAPSHOTS = '/snapshots/' + PLATFORM
+    ARTIFACTS_DISKS = '/disks/' + PLATFORM + "/"
+    ARTIFACTS_KERNELS = '/kernels/' + PLATFORM + "/"
+    ARTIFACTS_MICROVMS = '/microvms/'
+    ARTIFACTS_SNAPSHOTS = '/snapshots/' + PLATFORM + "/"
+    ARTIFACTS_BINARIES = '/binaries/' + PLATFORM + "/"
```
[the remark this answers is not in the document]
sandreim: I will explore the suggestion. Thanks.

tests/framework/artifacts.py
sandreim: That might be a very long line according to the linter.

tests/framework/artifacts.py
sandreim: Yup, wanted to filter out the folder.

tests/framework/microvms.py
sandreim: Thanks, I'll try this.

tests/conftest.py
sandreim: I spent some time on this here. The idea was to avoid building the fc binaries when caller provided the fc_binary. The assert should point to the obvious error that the jailer is missing. Also, the caller should always provide both binaries to avoid using mixed binaries.

tests/framework/artifacts.py
```diff
@@ -14,6 +14,9 @@
 from host_tools.snapshot_helper import merge_memory_bitmaps
 
 
+ARTIFACTS_LOCAL_ROOT = "/tmp/ci-artifacts"
```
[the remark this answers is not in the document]
sandreim: For this one, I guess it's shorter as a string.

tests/framework/artifacts.py
sandreim: Thanks, good finding.

tests/framework/microvms.py
sandreim: Added comment.

tests/integration_tests/functional/test_snapshot_version.py
sandreim: We have CPU templates so this should work, but you ar right the m5d.metal snapshots do not restore on local dev machines. I think it is a good idea to create snaps instead of getting them from S3. I will look into this.

sandreim: I implemented this as a simpler way to make this work for now.

tests/framework/utils.py
```diff
@@ -190,3 +191,9 @@ def run_cmd(cmd, ignore_return_code=False, no_shell=False):
         run_cmd_async(cmd=cmd,
                       ignore_return_code=ignore_return_code,
                       no_shell=no_shell))
+
+
+def is_amd():
```
serban300, line 196: Something similar is also used for the `cpu_features` tests. There's a local function defined in that test suite. In order to avoid duplicate code, I would suggest to move that logic in an utility class and use it both for `test_cpu_features` and for `test_snapshot_perf`. Something like: https://github.com/serban300/firecracker/commit/9dbf11b859c23fa28e54e9ff8caef41ad329153f . Of course your code for extracting the brand string is shorter and more efficient so we should leverage it instead of the old logic.

sandreim: Thanks. I'll update the existing tests.

serban300 (approved):

gc-plp (approved):

Issue #1571, opened 2020-02-03
[Devices] virtio memory balloon


acatangiu:
Adding support for a memory balloon device will allow Firecracker users to reclaim memory unused by their guests.

We should emulate a `virtio-balloon` device as described in the `VIRTIO` spec.

More details on memory ballooning concept [here](https://en.wikipedia.org/wiki/Memory_ballooning).

Related issues: #1570

aghecenco:
Previously prototyped in #1198

dianpopa:
Memory ballooning is in production as of [release v0.24.0](https://github.com/firecracker-microvm/firecracker/releases/tag/v0.24.0).

PR #2079, opened 2020-08-11
Memory ballooning support

georgepisaltu:
## Reason for This PR
#1571 

## Description of Changes

Added a virtio balloon device that supports statistics.

Note: Because Firecracker uses a private anonymous mapping for freshly booted microVMs and private file-backed mapping for snapshot resumed microVMs, a madvise call for releasing memory is not sufficient to work in both cases.
This PR currently uses a hacky approach: before madvise, mmap a fresh private anonymous region over the one that we want the kernel to release.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/balloon/device.rs
+++ b/src/devices/src/virtio/balloon/device.rs
@@ -11,6 +11,8 @@ use timerfd::{ClockId, SetTimeFlags, TimerFd, TimerState};
 
 use logger::{error, Metric, METRICS};
 use ::utils::eventfd::EventFd;
+use versionize::{VersionMap, Versionize, VersionizeResult};
+use versionize_derive::Versionize;
 use virtio_gen::virtio_blk::*;
 use vm_memory::{Address, ByteValued, Bytes, GuestAddress, GuestMemoryMmap};
 
@@ -24,11 +26,11 @@ use super::{
 
 use crate::{report_balloon_event_fail, Error as DeviceError};
 
-const SIZE_OF_U32: usize = 4;
-const SIZE_OF_STAT: usize = 10;
+const SIZE_OF_U32: usize = std::mem::size_of::<u32>();
+const SIZE_OF_STAT: usize = std::mem::size_of::<BalloonStat>();
 
 #[repr(C)]
-#[derive(Clone, Copy, Default)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Versionize)]
 pub(crate) struct ConfigSpace {
     num_pages: u32,
     actual_pages: u32,
@@ -50,7 +52,7 @@ struct BalloonStat {
 unsafe impl ByteValued for BalloonStat {}
 
 // BalloonStats holds statistics returned from the stats_queue.
-#[derive(Default, Debug, PartialEq)]
+#[derive(Clone, Default, Debug, PartialEq, Versionize)]
 pub struct BalloonStats {
     pub swap_in: Option<u64>,
     pub swap_out: Option<u64>,
@@ -99,6 +101,7 @@ pub struct Balloon {
     pub(crate) device_state: DeviceState,
 
     // Implementation specific fields.
+    pub(crate) restored: bool,
     pub(crate) stats_polling_interval_s: u16,
     pub(crate) stats_timer: TimerFd,
     // The index of the previous stats descriptor is saved because
@@ -113,7 +116,8 @@ impl Balloon {
         must_tell_host: bool,
         deflate_on_oom: bool,
         stats_polling_interval_s: u16,
-    ) -> io::Result<Balloon> {
+        restored: bool,
+    ) -> Result<Balloon, Error> {
         let mut avail_features = 1u64 << VIRTIO_F_VERSION_1;
 
         if must_tell_host {
@@ -150,7 +154,8 @@ impl Balloon {
             queue_evts,
             queues,
             device_state: DeviceState::Inactive,
-            activate_evt: EventFd::new(libc::EFD_NONBLOCK)?,
+            activate_evt: EventFd::new(libc::EFD_NONBLOCK).map_err(Error::EventFd)?,
+            restored,
             stats_polling_interval_s,
             stats_timer,
             stats_desc_index: None,
@@ -252,6 +257,7 @@ impl Balloon {
             match remove_range(
                 &mem,
                 (guest_addr, u64::from(range_len) << VIRTIO_BALLOON_PFN_SHIFT),
+                self.restored,
             ) {
                 Ok(_) => continue,
                 Err(e) => {
@@ -423,12 +429,13 @@ impl VirtioDevice for Balloon {
     }
 
     fn activate(&mut self, mem: GuestMemoryMmap) -> ActivateResult {
+        self.device_state = DeviceState::Activated(mem);
         if self.activate_evt.write(1).is_err() {
             error!("Balloon: Cannot write to activate_evt");
             METRICS.balloon.activate_fails.inc();
+            self.device_state = DeviceState::Inactive;
             return Err(super::super::ActivateError::BadActivate);
         }
-        self.device_state = DeviceState::Activated(mem);
 
         if self.stats_enabled() {
             let timer_state = TimerState::Periodic {
--- a/src/devices/src/virtio/balloon/device.rs
+++ b/src/devices/src/virtio/balloon/device.rs
@@ -2,19 +2,20 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::cmp;
-use std::io::{self, Write};
-use std::result;
+use std::io::Write;
+use std::result::Result;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
-use timerfd::{ClockId, SetTimeFlags, TimerFd, TimerState};
 
-use logger::{error, Metric, METRICS};
+use ::timerfd::{ClockId, SetTimeFlags, TimerFd, TimerState};
+
+use ::logger::{error, Metric, METRICS};
 use ::utils::eventfd::EventFd;
-use versionize::{VersionMap, Versionize, VersionizeResult};
-use versionize_derive::Versionize;
-use virtio_gen::virtio_blk::*;
-use vm_memory::{Address, ByteValued, Bytes, GuestAddress, GuestMemoryMmap};
+use ::versionize::{VersionMap, Versionize, VersionizeResult};
+use ::versionize_derive::Versionize;
+use ::virtio_gen::virtio_blk::*;
+use ::vm_memory::{Address, ByteValued, Bytes, GuestAddress, GuestMemoryMmap};
 
 use super::*;
 use super::{
@@ -24,10 +25,21 @@ use super::{
     utils::{compact_page_frame_numbers, remove_range},
 };
 
-use crate::{report_balloon_event_fail, Error as DeviceError};
+use crate::report_balloon_event_fail;
+use crate::virtio::balloon::Error as BalloonError;
 
 const SIZE_OF_U32: usize = std::mem::size_of::<u32>();
 const SIZE_OF_STAT: usize = std::mem::size_of::<BalloonStat>();
+
+macro_rules! mem_of_active_device {
+    ($state:expr) => {
+        match $state {
+            DeviceState::Activated(ref mem) => mem,
+            // This should never happen, it's been already validated in the event handler.
+            DeviceState::Inactive => unreachable!(),
+        }
+    };
+}
 
 #[repr(C)]
 #[derive(Clone, Copy, Debug, Default, PartialEq, Versionize)]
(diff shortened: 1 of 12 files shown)


src/devices/src/virtio/balloon/device.rs
```diff
@@ -0,0 +1,920 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+use timerfd::{ClockId, SetTimeFlags, TimerFd, TimerState};
+
+use ::utils::eventfd::EventFd;
+use logger::{error, Metric, METRICS};
+use versionize::{VersionMap, Versionize, VersionizeResult};
+use versionize_derive::Versionize;
+use virtio_gen::virtio_blk::*;
+use vm_memory::{Address, ByteValued, Bytes, GuestAddress, GuestMemoryMmap};
```
acatangiu, line 18: nit:
Since Rust 2018 external crates should be imported like `use ::logger::bla` for clarity.

src/devices/src/virtio/balloon/device.rs
```diff
@@ -0,0 +1,920 @@
...
+            self.process_inflate()
+                .unwrap_or_else(report_balloon_event_fail);
+        }
+    }
+
+    pub(crate) fn process_deflate_queue_event(&mut self) {
+        if let Err(e) = self.queue_evts[DEFLATE_INDEX].read() {
+            error!("Failed to get queue event: {:?}", e);
+            METRICS.balloon.event_fails.inc();
+        } else if self.process_deflate_queue() {
```
acatangiu, line 183: Don't we need `report_balloon_event_fail` in case `process_deflate_queue()` fails?

src/devices/src/virtio/balloon/device.rs
```diff
@@ -0,0 +1,920 @@
...
+
+            // Acknowledge the receipt of the descriptor.
+            // 0 is number of bytes the device has written to memory.
+            queue.add_used(&mem, head.index, 0);
+            needs_interrupt = true;
+        }
+
+        if needs_interrupt {
+            self.signal_used_queue()
+                .unwrap_or_else(report_balloon_event_fail);
```
acatangiu, line 253: Doesn't this lead to double counting of the event failure?

src/devices/src/virtio/balloon/device.rs
```diff
@@ -0,0 +1,920 @@
...
+
+    pub(crate) fn process_inflate_queue_event(&mut self) {
+        if let Err(e) = self.queue_evts[INFLATE_INDEX].read() {
+            error!("Failed to get queue event: {:?}", e);
+            METRICS.balloon.event_fails.inc();
+        } else {
+            self.process_inflate()
+                .unwrap_or_else(report_balloon_event_fail);
+        }
+    }
```
acatangiu, line 177: `process_inflate_queue_event()`, `process_deflate_queue_event()`, `process_stats_queue_event()` and `process_stats_timer_event()` - could be made to return `Result` and dedup failure code and hardish-to-read branching.

```suggestion
    pub(crate) fn process_inflate_queue_event(&mut self) -> Result<(), Error> {
        self.queue_evts[INFLATE_INDEX].read().map_err(Error::EventFd)?;
        self.process_inflate()
    }
```

And have the caller do `.unwrap_or_else(report_balloon_event_fail)`

src/devices/src/virtio/balloon/utils.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    // Don't forget to push the last range to the result.
+    result.push((previous, length));
+
+    result
+}
+
+pub(crate) fn remove_range(
```
acatangiu, line 51: @alexandruag @serban300 would it make sense to implement (and expose) this in `vm_memory` instead?

src/devices/src/virtio/balloon/device.rs
```diff
@@ -0,0 +1,920 @@
...
+                    error!("Error removing memory range: {:?}", e);
+                }
+            };
+        }
+
+        Ok(())
+    }
+
+    pub(crate) fn process_deflate_queue(&mut self) -> bool {
+        let mem = match self.device_state {
```
sandreim, line 278: Can this be moved to a function that returns the guest memory?

src/devices/src/virtio/balloon/device.rs
```diff
@@ -0,0 +1,920 @@
...
+    super::{
+        ActivateResult, DeviceState, Queue, VirtioDevice, TYPE_BALLOON, VIRTIO_MMIO_INT_VRING,
+    },
+    utils::{compact_page_frame_numbers, remove_range},
+};
+
+use crate::virtio::balloon::Error;
+use crate::{report_balloon_event_fail, Error as DeviceError};
+
+const SIZE_OF_U32: usize = 4;
```
sandreim, line 31: We can replace these constants with `std::mem::size_of::<u32>()` and `std::mem::size_of::<BalloonStat>()`. We can also add unit tests that assert the size of of BalloonsStat is 10.

src/devices/src/virtio/balloon/device.rs
```diff
@@ -0,0 +1,920 @@
...
+                let addr = head
+                    .addr
+                    .checked_add(index as u64)
+                    .ok_or(DeviceError::MalformedDescriptor)?;
+                let stat = mem
+                    .read_obj::<BalloonStat>(addr)
+                    .map_err(|_| DeviceError::MalformedDescriptor)?;
+                self.latest_stats
+                    .update_with_stat(&stat)
+                    .map_err(|_| DeviceError::MalformedPayload)?;
```
sandreim, line 325: No metric in case of stats update fail ?

src/devices/src/virtio/balloon/device.rs
```diff
@@ -0,0 +1,920 @@
...
+        }
+    }
+
+    fn activate(&mut self, mem: GuestMemoryMmap) -> ActivateResult {
+        if self.activate_evt.write(1).is_err() {
+            error!("Balloon: Cannot write to activate_evt");
+            METRICS.balloon.activate_fails.inc();
+            return Err(super::super::ActivateError::BadActivate);
+        }
+        self.device_state = DeviceState::Activated(mem);
```
sandreim, line 440: I think there is a possible race condition if the vCPU thread sleeps right before this line and the emulation thread attempts to do some work. This will lead to a panic via `DeviceState::Inactive => unreachable!(),`. I know this is highly unlikely as the activation part is driven by only one CPU in the guest, but as a best practice we should set the `device_state` before sending the activation notification.

It looks like this is so for all devices so it is worth digging deeper.

src/devices/src/virtio/balloon/utils.rs
```diff
@@ -0,0 +1,190 @@
...
+            return Err(RemoveRegionError::MalformedRange);
+        }
+        let phys_address = guest_memory
+            .get_host_address(guest_address)
+            .map_err(|_| RemoveRegionError::AddressTranslation)?;
+
+        // Mmap a new anonymous region over the present one in order to create a hole.
+        // This workaround is (only) needed after resuming from a snapshot because the guest memory
+        // is mmaped from file as private and there is no `madvise` flag that works for this case.
+        let ret = unsafe {
```
sandreim, line 68: Why not do this only when running a resumed VM ? the Baloon device has 2 constructors: for brand new devices and for restored devices; on the restore path you can set a flag so we avoid doing this for fresh microvms.

src/devices/src/virtio/balloon/device.rs
acatangiu: The device is only accessed through an `Arc<Mutex>` so there is no race condition here. A lot of logic internal to the device relies on exclusive access of `self`.

sandreim: > The device is only accessed through an `Arc<Mutex>` so there is no race condition here. A lot of logic internal to the device relies on exclusive access of `self`.

Yes, you are right, but still I would prefer swapping the 2 statements.

src/devices/src/virtio/balloon/device.rs
georgepisaltu: We do not as all operations in `process_deflate_queue()` are safe, the only things that can fail when dealing with the deflate queue are failure to read the `EventFd` and failure to signal the used queue.

The boolean return value corresponds to whether or not an interrupt is needed after processing the deflate queue, not whether or not the processing has succeeded.

src/devices/src/virtio/balloon/device.rs
georgepisaltu: No because if the `signal_used_queue()` call fails, the event will be logged as an error through `report_balloon_event_fail` and the execution of the function continues with the potential to return an `Ok` value, thus not reporting it again in `process_inflate_queue_event()`.

src/devices/src/virtio/balloon/device.rs
georgepisaltu: Unfortunately it would result in a double mutable borrow, as the reference to the guest memory needs to be mutable.

src/devices/src/virtio/balloon/device.rs
georgepisaltu: Added a metric in case of stats update failure.

src/devices/src/virtio/balloon/device.rs
georgepisaltu: Done.

src/devices/src/virtio/balloon/device.rs
georgepisaltu: Changed it to match the Rust 2018 style.

src/devices/src/virtio/balloon/utils.rs
georgepisaltu: Added a `restored` flag in order to have a way of finding out if the device is on a restored VM or not. Now we check for this and only perform the extra `mmap` when we are on a restored device.

src/devices/src/virtio/balloon/device.rs
georgepisaltu: I found returning the error to the caller to be difficult as it can't use the very simple `?` mechanism. A lot of the time the error is only logged and does not stop the flow of the function in which it appears.

Furthermore, the caller (`Subscriber::process()`) also handles the activation event in the same place and the errors for these types of events are different. Even if we find a way to make the errors reach the caller, we would need to either unify the error variants of the device with the activation specific ones or treat them in separate places.

acatangiu: > I found returning the error to the caller to be difficult as it can't use the very simple ? mechanism. A lot of the time the error is only logged and does not stop the flow of the function in which it appears.

Most of these `process_X_event` functions have that wild `if else if` format which can be simplified (for readability and ease of maintenance) using `?`. From what I see, the current code is logically structured to _not do other things_ after the error which is equivalent with _stopping the flow of the function_.

> Furthermore, the caller (`Subscriber::process()`) also handles the activation event in the same place and the errors for these types of events are different.

This is where you would consume or map the errors, they don't need to be all the same.
```rust
impl Subscriber for Balloon {
    fn process(self, ev: event) {
        match ev {
            X => self.process_x().unwrap_or_else(|| { /* process X err here */ }),
            Y => self.process_y().unwrap_or_else(|| { /* process Y err here */ }),
        }
    } 
}
```

To do this you would also have to clean up the `Device::Error` hierarchy a bit as it's currently a bit all over the place.

WDYT? Would it be worth it?

src/devices/src/virtio/balloon/device.rs
acatangiu: Could be a macro to dedup code 🤷‍♂️

src/devices/src/virtio/balloon/device.rs
georgepisaltu: I like the idea much more than how the processing happens now. I will try to refactor it a bit and see where we end up.

src/devices/src/virtio/balloon/device.rs
sandreim: Ok, so this was just purely cosmetical. We can postpone cosmetics for a later time.

src/devices/src/virtio/balloon/event_handler.rs
```diff
@@ -0,0 +1,183 @@
...
+        });
+    }
+}
+
+impl Subscriber for Balloon {
+    fn process(&mut self, event: &EpollEvent, evmgr: &mut EventManager) {
+        let source = event.fd();
+        let event_set = event.event_set();
+
+        // TODO: also check for errors. Pending high level discussions on how we want
```
sandreim, line 51: This comment is very vague.

src/devices/src/virtio/balloon/device.rs
```diff
@@ -0,0 +1,923 @@
...
+    utils::{compact_page_frame_numbers, remove_range},
+};
+
+use crate::report_balloon_event_fail;
+use crate::virtio::balloon::Error as BalloonError;
+
+const SIZE_OF_U32: usize = std::mem::size_of::<u32>();
+const SIZE_OF_STAT: usize = std::mem::size_of::<BalloonStat>();
+
+macro_rules! check_device_state {
```
acatangiu, line 34: nit: rename
```suggestion
macro_rules! mem_of_active_device {
```

sandreim (approved):

acatangiu (approved):

acatangiu (approved):

sandreim (approved):

PR #2084, opened 2020-08-17
Improved virtqueue error handling for used ring

luminitavoicu:

## Reason for This PR

During snapshot load, attempting add available descriptor into the virtio used ring caused errors when the address was off.

## Description of Changes

Added error and exited gracefully to handle the case described above.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/lib.rs
+++ b/src/devices/src/lib.rs
@@ -14,6 +14,7 @@ pub mod pseudo;
 pub mod virtio;
 
 pub use self::bus::{Bus, BusDevice, Error as BusError};
+use crate::virtio::QueueError;
 use logger::{error, Metric, METRICS};
 
 // Function used for reporting error in terms of logging
@@ -33,4 +34,6 @@ pub enum Error {
     IoError(io::Error),
     /// Device received malformed payload.
     MalformedPayload,
+    /// Error during queue processing.
+    QueueError(QueueError),
 }
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -272,7 +272,13 @@ impl Block {
                     len = 0;
                 }
             }
-            queue.add_used(mem, head.index, len);
+
+            queue.add_used(mem, head.index, len).unwrap_or_else(|e| {
+                error!(
+                    "Failed to add available descriptor head {}: {}",
+                    head.index, e
+                )
+            });
             used_any = true;
         }
 
--- a/src/devices/src/virtio/queue.rs
+++ b/src/devices/src/virtio/queue.rs
@@ -7,9 +7,12 @@
 
 use logger::error;
 use std::cmp::min;
+use std::fmt;
 use std::num::Wrapping;
 use std::sync::atomic::{fence, Ordering};
-use vm_memory::{Address, ByteValued, Bytes, GuestAddress, GuestMemory, GuestMemoryMmap};
+use vm_memory::{
+    Address, ByteValued, Bytes, GuestAddress, GuestMemory, GuestMemoryError, GuestMemoryMmap,
+};
 
 pub(super) const VIRTQ_DESC_F_NEXT: u16 = 0x1;
 pub(super) const VIRTQ_DESC_F_WRITE: u16 = 0x2;
@@ -21,6 +24,29 @@ pub(super) const VIRTQ_DESC_F_WRITE: u16 = 0x2;
 //
 // The Virtio Spec 1.0 defines the alignment of VirtIO descriptor is 16 bytes,
 // which fulfills the explicit constraint of GuestMemoryMmap::read_obj_from_addr().
+
+#[derive(Debug)]
+pub enum QueueError {
+    /// Descriptor index out of bounds.
+    DescIndexOutOfBounds(u16),
+    /// Attempted an invalid write into the used ring.
+    UsedRing(GuestMemoryError),
+}
+
+impl fmt::Display for QueueError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use self::QueueError::*;
+
+        match &*self {
+            DescIndexOutOfBounds(val) => write!(f, "Descriptor index out of bounds: {}", val),
+            UsedRing(e) => write!(
+                f,
+                "Failed to write value into the virtio queue used ring: {}",
+                e
+            ),
+        }
+    }
+}
 
 /// A virtio descriptor constraints with C representive.
 #[repr(C)]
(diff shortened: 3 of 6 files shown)


src/devices/src/virtio/queue.rs
```diff
@@ -309,31 +330,42 @@ impl Queue {
...
         let used_ring = self.used_ring;
         let next_used = u64::from(self.next_used.0 % self.actual_size());
         let used_elem = used_ring.unchecked_add(4 + next_used * 8);
 
-        // These writes can't fail as we are guaranteed to be within the descriptor ring.
-        mem.write_obj(u32::from(desc_index), used_elem).unwrap();
-        mem.write_obj(len as u32, used_elem.unchecked_add(4))
-            .unwrap();
+        mem.write_obj(u32::from(desc_index), used_elem)
+            .map_err(|_e| QueueError::InvalidUsedRingAddress(desc_index as u32, used_elem))?;
```
serban300, line 352: This error is not necessarily related to an InvalidAddress. It can be any of the [following](https://github.com/rust-vmm/vm-memory/blob/171e91e1bb057d3af9a65adf1e45952f24da514b/src/guest_memory.rs#L51).

I would define the error as `GuestMemoryError(GuestMemoryError)`.

src/devices/src/virtio/block/device.rs
```diff
@@ -272,7 +272,11 @@ impl Block {
                     len = 0;
                 }
             }
-            queue.add_used(mem, head.index, len);
+
+            queue.add_used(mem, head.index, len).unwrap_or_else(|e| {
+                error!("Failed to add available descriptor {}: {}", head.index, e);
+                std::process::exit(1);
```
serban300, line 278: I wouldn't exit on an `add_used()` failure. It can be a transitory error, a corrupted descriptor, etc. The process might continue to work correctly.

src/devices/src/virtio/queue.rs
```diff
@@ -789,15 +821,29 @@ pub(crate) mod tests {
...
-        assert_eq!(x.len, 0x1000);
+            //should be ok
+            q.add_used(m, 1, 0x1000).unwrap();
+            assert_eq!(vq.used.idx.get(), 1);
+            let x = vq.used.ring[0].get();
+            assert_eq!(x.id, 1);
+            assert_eq!(x.len, 0x1000);
+        }
+
+        //Invalid used ring address addresses
```
serban300, line 838: Nit: Maybe `Invalid queue addresses configuration` would be better.

luminitavoicu: Right, I will correct this.

src/devices/src/virtio/block/device.rs
luminitavoicu: I was considering the possibility for this to fail due to a malicious guest/attacker, in which case the process should not continue to work past this point. WDYT?

src/devices/src/virtio/queue.rs
luminitavoicu: That is right. I will update this.

src/devices/src/virtio/queue.rs
```diff
@@ -22,6 +23,26 @@ pub(super) const VIRTQ_DESC_F_WRITE: u16 = 0x2;
...
+}
+
+impl fmt::Display for QueueError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use self::QueueError::*;
+
+        match &*self {
+            GuestMemoryError(e) => write!(
+                f,
+                "Failed to write value into the virtio queue used ring: {}",
```
acatangiu, line 39: I think this should be a generic guest mem error description so it'd be applicable to other cases not just the virtio queue used ring.

`GuestMemory access error: {}` sounds good to me for example

src/devices/src/virtio/queue.rs
```diff
@@ -309,31 +330,40 @@ impl Queue {
...
+        desc_index: u16,
+        len: u32,
+    ) -> Result<(), QueueError> {
         if desc_index >= self.actual_size() {
             error!(
                 "attempted to add out of bounds descriptor to used ring: {}",
                 desc_index
             );
-            return;
+            return Ok(());
```
acatangiu, line 344: This should be an error. You can add a special variant for it like `QueueError::DescIndexOutOfBounds(u16)`

src/devices/src/virtio/block/device.rs
acatangiu: Like Serban said, we shouldn't `exit()` on these events. If the error cannot be cascaded up (like in these device functions) we should log the error and move on. The invalid access is blocked and logged.

src/devices/src/virtio/queue.rs
```diff
@@ -789,15 +819,29 @@ pub(crate) mod tests {
...
+
+        //Invalid queue addresses configuration
+        {
+            q.used_ring = GuestAddress(0xffff_ffff);
+
+            //should fail
+            assert!(match q.add_used(m, 1, 0x1000) {
+                Err(QueueError::GuestMemoryError(_)) => true,
+                _ => false,
+            });
```
lauralt, line 844: Since both `QueueError` and `GuestMemoryError` implement Display, I would test `add_used()` errors like this:
```rust
assert_eq!(format!("{}", q.add_used(m, 1, 0x1000).unwrap_err()), format!("Failed to write value into the virtio queue used ring: Guest memory error: invalid guest address {}", invalid_addr));
```
and I would somehow compute `invalid_addr` based on the other addresses / offsets.
This way, the testing is more precise and you also test the Display functionality that you implemented for `QueueError`.

Also, just curious: is it possible to test the other two `QueueError` scenarios too?

src/devices/src/virtio/queue.rs
```diff
@@ -309,31 +330,40 @@ impl Queue {
...
 
         // This fence ensures all descriptor writes are visible before the index update is.
         fence(Ordering::Release);
 
-        mem.write_obj(self.next_used.0 as u16, used_ring.unchecked_add(2))
-            .unwrap();
+        let next_used_addr = used_ring.unchecked_add(2);
+        mem.write_obj(self.next_used.0 as u16, next_used_addr)
+            .map_err(QueueError::GuestMemoryError)?;
+        Ok(())
```
lauralt, line 366:
```suggestion
        mem.write_obj(self.next_used.0 as u16, next_used_addr)
            .map_err(QueueError::GuestMemoryError)
```

src/devices/src/virtio/queue.rs
acatangiu: We decided a while ago to not duplicate implementation in tests. @andreeaflorescu had a nice article around it.

Tests that were matching exact error messages were deemed unuseful. Matching string literal duplicated both in impl and tests is redundant and not really useful.

We decided to test `Debug` and `Display` by just doing `format!("{:?}{}", err, err)` which verifies (at test compile time) that `Debug` and `Display` are implemented.

lauralt: Ok, sounds good too.

src/devices/src/virtio/queue.rs
luminitavoicu: Good point.

sandreim: I think that errors generated in specific code can benefit from being very explicit at the cost of increasing the binary size.

acatangiu: We can keep the specific error description but then we should also change the error name to `QueueError::UsedRing` or smth specific.

src/devices/src/virtio/queue.rs
luminitavoicu: Done.

src/devices/src/virtio/block/device.rs
luminitavoicu: Handled it.

src/devices/src/virtio/queue.rs
luminitavoicu: Done

src/devices/src/lib.rs
```diff
@@ -33,4 +34,6 @@ pub enum Error {
     IoError(io::Error),
     /// Device received malformed payload.
     MalformedPayload,
+    /// Error during queue processing.
+    QueueError(QueueError),
```
acatangiu, line 38: nit: we try to keep `enum` variants alphabetically ordered (to be easy to find/navigate)

acatangiu (approved):

src/devices/src/lib.rs
luminitavoicu: Solved

src/devices/src/virtio/queue.rs
```diff
@@ -22,6 +25,29 @@ pub(super) const VIRTQ_DESC_F_WRITE: u16 = 0x2;
 // The Virtio Spec 1.0 defines the alignment of VirtIO descriptor is 16 bytes,
 // which fulfills the explicit constraint of GuestMemoryMmap::read_obj_from_addr().
 
+#[derive(Debug)]
+pub enum QueueError {
+    /// Descriptor index out of bounds.
+    DescIndexOutOfBounds(u16),
+    /// Attempted to write to an invalid address into the used ring.
+    UsedRing(GuestMemoryError),
```
lauralt, line 33: nit:
To avoid any confusion, I would update this comment with:

`
\\\ Attempted to do an invalid write into the used ring.
`

since we can also receive, for example, the `PartialBuffer` error, which doesn't exactly refer to an invalid address.

lauralt (dismissed):

src/devices/src/virtio/queue.rs
luminitavoicu: Good catch. Updated the comment.

lauralt (approved):

acatangiu (approved):

PR #2086, opened 2020-08-19
aarch64: implement save restore for VcpuState

### LINKED ISSUE #1826: aarch64: implement save/restore for VcpuState
* `pub fn save_state`
* `pub fn restore_state`

dianpopa:
## Reason for This PR

Saving/restoring the VcpuState paves the way to achieving snapshot create/resume on aarch64. 
Fixes #1826.

## Description of Changes

Read commits independently.

## License Acceptance

## PR Checklist


src/arch/Cargo.toml
```diff
@@ -5,8 +5,8 @@ authors = ["The Chromium OS Authors"]
 edition = "2018"
 
 [dependencies]
-kvm-bindings = { git = "https://github.com/firecracker-microvm/kvm-bindings", tag = "v0.2.0-2", features = ["fam-wrappers"] }
-kvm-ioctls = { git = "https://github.com/firecracker-microvm/kvm-ioctls", tag = "v0.5.0-2" }
+kvm-bindings = { path = "../kvm-bindings", features = ["fam-wrappers"]  }
+kvm-ioctls = { path = "../kvm-ioctls" }
```
sandreim, line 9: Did we ingest the crates back in Firecracker ?

src/arch/src/aarch64/regs.rs
```diff
@@ -176,6 +207,237 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+pub fn save_core_registers(vcpu: &VcpuFd, state: &mut kvm_regs) -> Result<()> {
+    let mut off = offset__of!(user_pt_regs, regs);
+    // There are 31 user_pt_regs:
+    // https://elixir.free-electrons.com/linux/v4.14.174/source/arch/arm64/include/uapi/asm/ptrace.h#L72
+    // These actually are the general-purpose registers of the Armv8-a
+    // architecture (i.e x0-x30 if used as a 64bit register or w0-30 when used as a 32bit register).
+    for i in 0..31 {
+        state.regs.regs[i] = vcpu
+            .get_one_reg(arm64_core_reg_id!(KVM_REG_SIZE_U64, off))
+            .map_err(Error::GetCoreRegister)?;
```
sandreim, line 225: Can we make the error more specific so that it is obvious which reg failed ?

src/arch/src/aarch64/regs.rs
```diff
@@ -176,6 +207,237 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+    reg_list.as_slice().to_vec().sort();
+
+    reg_list.retain(|regid| is_system_register(*regid));
+
+    // Now, for the rest of the registers left in the previously fetched register list, we are
+    // simply calling KVM_GET_ONE_REG.
+    let indices = reg_list.as_slice();
+    for (_pos, index) in indices.iter().enumerate() {
+        if _pos > 230 {
+            break;
```
sandreim, line 396: Why ?

sandreim (commented):
I've taken a first pass at this. It looks like the PR contains a lot of stuff that should be removed(kvm-bindings and versionize crate code). Please clean it up a little and I'll take another look.

src/arch/src/aarch64/regs.rs
```diff
@@ -196,6 +207,236 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+///
+/// * `vcpu` - Structure for the VCPU that holds the VCPU's fd.
+/// * `state` - Structure for returning the state of the core registers.
+pub fn save_core_registers(vcpu: &VcpuFd, state: &mut kvm_regs) -> Result<()> {
+    let mut off = offset__of!(user_pt_regs, regs);
+    // There are 31 user_pt_regs:
+    // https://elixir.free-electrons.com/linux/v4.14.174/source/arch/arm64/include/uapi/asm/ptrace.h#L72
+    // These actually are the general-purpose registers of the Armv8-a
+    // architecture (i.e x0-x30 if used as a 64bit register or w0-30 when used as a 32bit register).
+    for i in 0..31 {
```
serban300, line 222: Can we use `state.regs.regs.iter_mut()` in order to avoid the hardcoded value `31` here ?

src/arch/src/aarch64/regs.rs
```diff
@@ -196,6 +207,236 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+    // Exception Link Register for EL1, when taking an exception to EL1, this register
+    // holds the address to which to return afterwards.
+    let off = offset__of!(kvm_regs, elr_el1);
+    state.elr_el1 = vcpu
+        .get_one_reg(arm64_core_reg_id!(KVM_REG_SIZE_U64, off))
+        .map_err(Error::GetCoreRegister)?;
+
+    // Saved Program Status Registers, there are 5 of them used in the kernel.
+    let mut off = offset__of!(kvm_regs, spsr);
+    for i in 0..KVM_NR_SPSR as usize {
```
serban300, line 263: I would use `state.spsr.iter_mut()` if possible.

src/arch/src/aarch64/regs.rs
```diff
@@ -196,6 +207,236 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+        state.spsr[i] = vcpu
+            .get_one_reg(arm64_core_reg_id!(KVM_REG_SIZE_U64, off))
+            .map_err(Error::GetCoreRegister)?;
+        off += std::mem::size_of::<u64>();
+    }
+
+    // Now moving on to floting point registers which are stored in the user_fpsimd_state in the kernel:
+    // https://elixir.free-electrons.com/linux/v4.9.62/source/arch/arm64/include/uapi/asm/kvm.h#L53
+    let mut off = offset__of!(kvm_regs, fp_regs) + offset__of!(user_fpsimd_state, vregs);
+    for i in 0..32 {
```
serban300, line 273: I would use `state.fp_regs.vregs.iter_mut()` if possible.

src/arch/src/aarch64/regs.rs
```diff
@@ -196,6 +207,236 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
     vcpu.get_one_reg(MPIDR_EL1).map_err(Error::GetSysRegister)
 }
 
+/// Get the state of the core registers.
+///
+/// # Arguments
+///
+/// * `vcpu` - Structure for the VCPU that holds the VCPU's fd.
+/// * `state` - Structure for returning the state of the core registers.
+pub fn save_core_registers(vcpu: &VcpuFd, state: &mut kvm_regs) -> Result<()> {
```
serban300, line 216: It looks like we save the entire `kvm_regs` structure. Is it possible to also use `get_reg_list()` here ? instead of iterating through the entire structure manually ?

src/arch/src/aarch64/regs.rs
```diff
@@ -196,6 +207,236 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+    let mut off = offset__of!(user_pt_regs, regs);
+    // There are 31 user_pt_regs:
+    // https://elixir.free-electrons.com/linux/v4.14.174/source/arch/arm64/include/uapi/asm/ptrace.h#L72
+    // These actually are the general-purpose registers of the Armv8-a
+    // architecture (i.e x0-x30 if used as a 64bit register or w0-30 when used as a 32bit register).
+    for i in 0..31 {
+        state.regs.regs[i] = vcpu
+            .get_one_reg(arm64_core_reg_id!(KVM_REG_SIZE_U64, off))
```
serban300, line 224: Aren't the id's continuous ? Instead of incrementing the offset, could we start with `id = arm64_core_reg_id!(KVM_REG_SIZE_U64, 0)` and increment the id after processing each register ?

dianpopa:
@sandreim @serban300 thanks for the review, nice suggestions! I did a `push -f` to use latest kvm-ioctls, will address your comments at the start of next week.

src/arch/src/aarch64/regs.rs
```diff
@@ -196,6 +207,236 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+    vcpu.get_reg_list(&mut reg_list)
+        .map_err(Error::GetRegList)?;
+
+    // At this point reg_list should contain: core registers and system registers.
+    // The register list contains the number of registers and their ids. We will be needing to
+    // call KVM_GET_ONE_REG on each id in order to save all of them. We carve out from the list
+    // the core registers which are represented in the kernel by kvm_regs structure and for which
+    // we can calculate the id based on the offset in the structure.
+
+    reg_list.retain(|regid| *regid != 0);
```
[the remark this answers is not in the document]
dianpopa: Yes, the intention is to remove the extra elements. Indeed the retain does not do anythin. Removed it.

src/arch/src/aarch64/regs.rs
```diff
@@ -196,6 +207,236 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+    reg_list.as_slice().to_vec().sort();
```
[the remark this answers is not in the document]
dianpopa: No, it is not needed.

src/arch/src/aarch64/regs.rs
dianpopa: This is a debug leftover also, removed it.

src/arch/src/aarch64/regs.rs
```diff
@@ -196,6 +207,236 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+    Ok(())
+}
+
+/// Save the state of the core registers.
+///
+/// # Arguments
+///
+/// * `vcpu` - Structure for the VCPU that holds the VCPU's fd.
+/// * `state` - Structure containing the state of the core registers.
+pub fn restore_core_registers(vcpu: &VcpuFd, state: &kvm_bindings::kvm_regs) -> Result<()> {
```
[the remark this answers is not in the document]
dianpopa: I did think about this but I did not like what I obtained. At every step you need to check for every register if  you are in a get or a set since the offset is calculated progressively and that makes the function uglier than this. I d also need to use a function which receives mutable objects just for the "get" usecase and ignore the mutable for the set. I find this alternative harder to read, wdyt?

dianpopa (commented):
@serban300 @sandreim I addressed your feedback. Please take another look!

src/arch/src/aarch64/regs.rs
serban300: Sorry, I'm not very familiar with ARM and this KVM methods. so `get_reg_list()` doesn't contain the core registers ?

serban300 (dismissed):

serban300 (approved):

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -105,36 +109,58 @@ impl KvmVcpu {
...
+
+        state.mpidr = arch::aarch64::regs::read_mpidr(&self.fd).map_err(Error::SaveState)?;
+
+        Ok(state)
     }
 
     /// Use provided state to populate KVM internal state.
-    pub fn restore_state(&self, _state: &VcpuState) -> Result<()> {
-        Err(Error::UnsupportedAction("Restoring the state"))
+    pub fn restore_state(&self, state: &VcpuState) -> Result<()> {
```
iulianbarbu, line 154: Shouldn't we restore MPIDR as well?

src/arch/src/aarch64/regs.rs
iulianbarbu: I am wondering too if we can get the core registers by using `get_reg_list()` and retain only what is not `system` registers. Would this filtering result in other registers besides core registers?

src/vmm/src/vstate/vcpu/aarch64.rs
dianpopa: MPIDR is saved here because we need it in order to save the GIC's registers. This functionality is needed when saving/restoreing vm state (next PR).

src/arch/src/aarch64/regs.rs
```diff
@@ -196,6 +211,226 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+///
+/// * `vcpu` - Structure for the VCPU that holds the VCPU's fd.
+/// * `state` - Structure for returning the state of the core registers.
+pub fn save_core_registers(vcpu: &VcpuFd, state: &mut kvm_regs) -> Result<()> {
+    let mut off = offset__of!(user_pt_regs, regs);
+
+    // There are 31 user_pt_regs:
+    // https://elixir.free-electrons.com/linux/v4.14.174/source/arch/arm64/include/uapi/asm/ptrace.h#L72
+    // These actually are the general-purpose registers of the Armv8-a
+    // architecture (i.e x0-x30 if used as a 64bit register or w0-30 when used as a 32bit register).
```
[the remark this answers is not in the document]
dianpopa: Not applicable anymore.

src/arch/src/aarch64/regs.rs
dianpopa: OOh now I understand what you are saying. Indeed we can do it this way, nothing that can stop us. Never realized it since I started development with saving/retoring core_regs and then discovered reg list for the system registers. I updated the code to your suggestion.

dianpopa (commented):
@iulianbarbu @serban300 addressed your comments. Please take another look. Thanks!

src/vmm/src/vstate/vcpu/aarch64.rs
iulianbarbu: Actually, we restore the MPDIR at 151 line. Sorry for the wasted cycles. :D I'll resolve this.

iulianbarbu (approved):
LGTM!

src/arch/src/aarch64/regs.rs
```diff
@@ -209,6 +229,74 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
...
+    // The register list contains the number of registers and their ids. We will be needing to
+    // call KVM_GET_ONE_REG on each id in order to save all of them. We carve out from the list
+    // the core registers which are represented in the kernel by kvm_regs structure and for which
+    // we can calculate the id based on the offset in the structure.
+    reg_list.retain(|regid| is_core_register(*regid) || is_system_register(*regid));
+
+    // Now, for the rest of the registers left in the previously fetched register list, we are
+    // simply calling KVM_GET_ONE_REG.
+    let indices = reg_list.as_slice();
+    for index in indices.iter() {
```
serban300, line 255: It might be more efficient to just do something like `reg_list.as_slice().iter().filter(|regid| is_core_register(*regid) || is_system_register(*regid))` ?  `retain` removes elements from the `FamStruct`. `filter` should only do the check.

dianpopa: I will try it. Did not know there exists 'filter'.

src/arch/src/aarch64/regs.rs
dianpopa: @iulianbarbu @serban300 I reverted this change as it was not working on the arm CI machines. 
The reason is that prior to this fix: https://patchwork.kernel.org/project/linux-arm-kernel/patch/20190620112301.138137-3-marc.zyngier@arm.com/, KVM_GET_REGLIST would also return invalid registers ids (not only core and system ids).
What tricked me into thinking that it worked was the fact that my dev machine has 5.3.0 kernel while on the CI machines the host kernel is version 4.14 and does not contain the fix mentioned above. Please re-review the reverted code!

serban300: That's unfortunate :( Isn't there any way to filter these invalid registers from the userspace ?

dianpopa: Yes, the code was super simple and nice before indeed. I spent quite some time trying to figure out what is happening, looked in kernel code for some way to filter those invalid registers and did not find any way. What happens is that KVM_GRET_REG_LIST returns a list of registers that apparently look like core registers but when I call kvm_get_one_reg the kernel returns EINVAL. Looking at the fix posted by Mark Zyngier regarding this issue, I see that he also uses `struct kvm_regs` for validating those, so it seems that this is the safest method to do it.

serban300 (approved):

iulianbarbu (approved):

PR #2092, opened 2020-08-21
Snapshot restore - devices sanity checks

### LINKED ISSUE #2099: [Bug] unhandled return result in device_manager/persist
## Describe the bug

There are [errors](https://github.com/firecracker-microvm/firecracker/blob/master/src/vmm/src/device_manager/persist.rs#L212) inside the aforementioned file that do not get propagated to the upper level. They are hiding behind this [allow unused](https://github.com/firecracker-microvm/firecracker/blob/master/src/vmm/src/device_manager/persist.rs#L9).

acatangiu:
## Description of Changes

Added sanity checks along the restore from snapshot path for device manager and virtio devices.

Also provided support for explicit user-facing error when attempting snapshot-create without pausing.

Fixes #2099 

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/persist.rs
+++ b/src/devices/src/virtio/persist.rs
@@ -16,6 +16,11 @@ use std::num::Wrapping;
 use std::sync::atomic::Ordering;
 use std::sync::{Arc, Mutex};
 
+#[derive(Debug)]
+pub enum Error {
+    InvalidInput,
+}
+
 #[derive(Clone, Debug, PartialEq, Versionize)]
 pub struct QueueState {
     /// The maximal size in elements offered by the device
@@ -38,6 +43,25 @@ pub struct QueueState {
 
     next_avail: Wrapping<u16>,
     next_used: Wrapping<u16>,
+}
+
+impl QueueState {
+    /// Does a sanity check against expected values.
+    pub fn sanity_check(&self, queue_max_size: u16) -> std::result::Result<(), Error> {
+        // Cannot use `q.is_valid()` because snapshot can happen at any time,
+        // including during device configuration/activation when fields are only
+        // partially configured.
+        // We can't even check if GuestAddresses are valid guest phys addresses because
+        // their configuration happens in two steps for the two u32 halves of the
+        // u64 address. This means a snapshot can capture them only partially configured.
+        //
+        // The best we can do is sanity check queue size and max size.
+        if self.max_size != queue_max_size || self.size > queue_max_size {
+            Err(Error::InvalidInput)
+        } else {
+            Ok(())
+        }
+    }
 }
 
 impl Persist<'_> for Queue {
@@ -97,6 +121,30 @@ impl VirtioDeviceState {
             activated: device.is_activated(),
         }
     }
+
+    /// Does a sanity check against expected values.
+    pub fn sanity_check(
+        &self,
+        device_type: u32,
+        num_queues: usize,
+        queue_max_size: u16,
+    ) -> std::result::Result<(), Error> {
+        // Check:
+        // - right device type,
+        // - acked features is a subset of available ones,
+        // - right number of queues,
+        if self.device_type != device_type
+            || (self.acked_features & !self.avail_features) != 0
+            || self.queues.len() != num_queues
+        {
+            return Err(Error::InvalidInput);
+        }
+        // Queues are the expected size.
+        for q in self.queues.iter() {
+            q.sanity_check(queue_max_size)?;
+        }
+        Ok(())
+    }
 }
 
 #[derive(Debug, PartialEq, Versionize)]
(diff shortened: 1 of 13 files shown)


src/devices/src/virtio/persist.rs
```diff
@@ -97,6 +121,30 @@ impl VirtioDeviceState {
...
+        device_type: u32,
+        num_queues: usize,
+        queue_max_size: u16,
+    ) -> std::result::Result<(), Error> {
+        // Check:
+        // - right device type,
+        // - acked features is a subset of available ones,
+        // - right number of queues,
+        if self.device_type != device_type
+            || (self.acked_features & !self.avail_features) != 0
```
georgepisaltu, line 137: nit: use xor for this evaluation.

`self.acked_features ^ self.avail_features != 0`

acatangiu: actually no, while `a & !b` implies `a ^ b` the reverse is not necessarily true.
`a ^ b` = `(a & !b) | (!a &b)` so using `xor` could evaluate true to because of `(!a &b)` which would provide unwanted behavior in our case.

Practical example:
```
acked = 0b0001
avail = 0b0101
# above is correct config example
acked & !avail = 0 # checking against 0 confirms config is good -> correct
acked ^ avail = 0b0100 # checking against 0 says the config isn't good -> incorrect
```

src/vmm/src/vstate/vcpu/mod.rs
```diff
@@ -313,7 +313,9 @@ impl Vcpu {
             // SaveState or RestoreState cannot be performed on a running Vcpu.
             Ok(VcpuEvent::SaveState) | Ok(VcpuEvent::RestoreState(_)) => {
                 self.response_sender
-                    .send(VcpuResponse::NotAllowed)
+                    .send(VcpuResponse::NotAllowed(String::from(
+                        "not allowed on running Vcpu",
```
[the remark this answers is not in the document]
acatangiu: done

serban300 (dismissed):

src/devices/src/virtio/persist.rs
georgepisaltu: I see, I completely forgot that the device does not necessarily need to accept all available features, therefore the XOR becomes incorrect.

src/devices/src/virtio/persist.rs
```diff
@@ -40,6 +45,25 @@ pub struct QueueState {
...
 
+impl QueueState {
+    /// Does a sanity check against expected values.
+    pub fn sanity_check(&self, queue_max_size: u16) -> std::result::Result<(), Error> {
+        // Cannot use `q.is_valid()` because snapshot can happen at any time,
+        // including during device configuration/activation when fields are only
+        // partially configured.
+        // We can't even check if GuestAddresses are valid guest phys addresses because
+        // their configuration happens in two steps for the two u32 halves of the
+        // u64 address. This means a snapshot can capture them only partially configured.
```
serban300, line 56: Only if the device status is active could we also check `q.is_valid()` though ?

acatangiu: I looked into doing that, but the current design doesn't easily allow it.
The queues and their state are decoupled from the (virtio)MmioTransport where the `active` information is stored.

The queues are restored as part of device restoration, then after we have a device we proceed to re-register/restore MmioTransport state.

I don't see an elegant solution to doing it..

src/vmm/src/device_manager/mmio.rs
```diff
@@ -81,43 +84,66 @@ pub struct MMIODeviceInfo {
...
     pub fn new(mmio_base: u64, irq_interval: (u32, u32)) -> MMIODeviceManager {
         MMIODeviceManager {
+            #[cfg(target_arch = "x86_64")]
             mmio_base,
-            irq: irq_interval.0,
+            next_avail_mmio: mmio_base,
+            #[cfg(target_arch = "x86_64")]
+            first_irq: irq_interval.0,
             last_irq: irq_interval.1,
+            next_avail_irq: irq_interval.0,
```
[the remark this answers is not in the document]
acatangiu: > Another thing that stands out here is that next_avail_irq doesn't seem to be restored when loading a snapshot. Which might cause problems if we decide to do device hotplug in the future.

Most of the current code assumes no further device registrations post-boot. The device manager will move to a totally different model (PCI) in order to add support for hoptplug, so I don't see any reason to save/restore any pre-boot state.

> Using 3 variables in order to keep track of the available IRQs seems a bit too much.

I like your suggested Vec(Deque) approach!

src/vmm/src/device_manager/mmio.rs
```diff
@@ -81,43 +84,66 @@ pub struct MMIODeviceInfo {
...
     }
 
+    #[cfg(target_arch = "x86_64")]
+    /// Does a slot sanity check against expected values.
+    pub fn slot_sanity_check(&self, slot: &MMIODeviceInfo) -> Result<()> {
+        if slot.addr < self.mmio_base || slot.len != MMIO_LEN {
+            return Err(Error::InvalidInput);
+        }
+        for irq in slot.irqs.iter() {
+            if self.first_irq > *irq || *irq > self.last_irq {
```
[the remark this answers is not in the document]
acatangiu: I don't see it as more readable 🙈 
It'd also turn two checks into an iteration over a range with a check at each step..

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
acatangiu: I ended up with the separate object with a cleaner interface than using `VecDeque`.
@serban300 @georgepisaltu  PTAL

src/vmm/src/device_manager/mmio.rs
```diff
@@ -78,46 +81,92 @@ pub struct MMIODeviceInfo {
...
+        let mut irqs = Vec::with_capacity(count as usize);
+        for _ in 0..count {
+            irqs.push(self.next_avail);
+            self.next_avail += 1;
+        }
+        Ok(irqs)
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    pub fn check(&self, irqs: &[u32]) -> Result<()> {
```
serban300, line 114: Nit: `check()` is very vague. I would rename this function to something like `are_irqs_valid()` or `are_irqs_available()`

serban300 (approved):

georgepisaltu (approved):

PR #2095, opened 2020-08-25
Expose the balloon device through the API

georgepisaltu:
## Reason for This PR
#1571 

## Description of Changes

Exposed the balloon device interface to the vmm and API server in order to make the device accessible to users.

## License Acceptance

## PR Checklist


src/vmm/src/rpc_interface.rs
```diff
@@ -86,6 +95,9 @@ pub enum VmmAction {
     /// driver is listening on the guest end, this can be used to shut down the microVM gracefully.
     #[cfg(target_arch = "x86_64")]
     SendCtrlAltDel,
+    /// Update a network interface, after microVM start. Currently, the only updatable properties
+    /// are the RX and TX rate limiters.
+    UpdateBalloon(BalloonUpdateConfig),
```
acatangiu, line 100: Wrong doc comment

src/vmm/src/rpc_interface.rs
```diff
@@ -174,6 +189,8 @@ pub enum VmmData {
     Empty,
     /// The microVM configuration represented by `VmConfig`.
     MachineConfiguration(VmConfig),
+    /// The latest balloon device statistics.
+    BalloonStats(BalloonStats),
```
acatangiu, line 193: nit: alphabetic

src/vmm/src/rpc_interface.rs
```diff
@@ -485,6 +514,84 @@ impl RuntimeApiController {
         Ok(())
     }
 
+    /// Updates configuration for the balloon device as described in `balloon_update`.
+    fn latest_balloon_stats(&self) -> result::Result<VmmData, VmmActionError> {
```
acatangiu, line 518: Please move this to a `Vmm` function. The RPC interface is supposed to only do RPC and only care about commands ordering requirements, it shouldn't have virtualization logic.

src/vmm/src/rpc_interface.rs
```diff
@@ -485,6 +514,84 @@ impl RuntimeApiController {
...
+            ))
+        } else {
+            Err(VmmActionError::BalloonConfig(
+                BalloonConfigError::DeviceNotFound,
+            ))
+        }
+    }
+
+    /// Updates configuration for the balloon device as described in `balloon_update`.
+    fn update_balloon_config(&mut self, balloon_update: BalloonUpdateConfig) -> ActionResult {
```
acatangiu, line 554: same with this: move to `Vmm`

(I know block and net updates are also here, but those should also move to the vmm impl/interface - outside the scope of this PR though)

Cargo.lock
```diff
@@ -240,6 +240,7 @@ dependencies = [
  "net_gen 0.1.0",
  "polly 0.0.1",
  "rate_limiter 0.1.0",
+ "serde 1.0.105 (registry+https://github.com/rust-lang/crates.io-index)",
```
acatangiu, line 243: nit: this should be in the `devices: balloon: expose balloon API` commit - that's where you take `serde` as a dependency in `devices`

tests/framework/microvm.py
```diff
@@ -345,6 +347,7 @@ def init_snapshot_api(self):
             self._api_session
         )
 
+    # pylint: disable=too-many-statements
```
acatangiu, line 350: I'm so sorry you were unlucky enough to hit the limit with just **one** line added... but you were nonetheless.

Please split the function instead of disabling the lint.

tests/host_tools/fillmem.c
```diff
@@ -0,0 +1,81 @@
...
+    size_t len = 100 * MB;
+    while(1) {
+        ptr = mmap(NULL, len * sizeof(char), PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+        if (ptr != NULL) {
+            for (i = 0; i < len; i++) {
+                ptr[i] = '0';
+            }
+        }
+    }
+}*/
```
acatangiu, line 45: remove this maybe?

tests/host_tools/fillmem.c
```diff
@@ -0,0 +1,81 @@
...
+        if (WIFSIGNALED(status)) {
+            char buf[200];
+            sprintf(buf, "OOM Killer stopped the program with signal %d, exit code %d\n", WTERMSIG(status), WEXITSTATUS(status));
+            write(fd, buf, strlen(buf) + 1);
+            close(fd);
+
+            return 0;
+        }
+
+        write(fd, "Memory filling was successful\n", 31);
```
acatangiu, line 76: nit: all uses of `fill` are a bit misleading, the program attempts to allocate and use `arg#1` MB of memory, not _fill all available memory_.

src/vmm/src/device_manager/persist.rs
```diff
@@ -84,6 +100,8 @@ pub struct ConnectedVsockState {
 #[derive(Versionize)]
 /// Holds the device states.
 pub struct DeviceStates {
+    /// Balloon device state.
+    pub balloon_device: Option<ConnectedBalloonState>,
```
acatangiu, line 104: This needs bumping the snapshot version.

Any change to any `Versionized` struct requires bumping snapshot version.

src/vmm/src/device_manager/persist.rs
```diff
@@ -132,6 +151,19 @@ impl<'a> Persist<'a> for MMIODeviceManager {
 
             let locked_device = mmio_transport.locked_device();
             match locked_device.device_type() {
+                TYPE_BALLOON => {
```
acatangiu, line 154: Please add an integ test with snaps+balloon

src/vmm/src/resources.rs
```diff
@@ -275,6 +288,14 @@ impl VmResources {
         self.vsock.insert(config)
     }
 
+    /// Sets a balloon device to be attached when the VM starts.
+    pub fn set_balloon_device(
```
sandreim, line 292: Do we actually need this fn? Why not just call ballon.insert directly ?

src/vmm/tests/integration_tests.rs
```diff
@@ -312,7 +312,7 @@ fn verify_create_snapshot(is_diff: bool) -> (TempFile, TempFile) {
                 snapshot_type,
                 snapshot_path: snapshot_file.as_path().to_path_buf(),
                 mem_file_path: memory_file.as_path().to_path_buf(),
-                version: Some(String::from("0.23.0")),
+                version: Some(String::from("0.24.0")),
```
sandreim, line 315: This only validates that we can create snapshots for current version. I would think if it is worth creating a snapshot for current aand all previous version.

tests/integration_tests/build/test_coverage.py
```diff
@@ -23,7 +23,7 @@
 # this contains the frequency while on AMD it does not.
 # Checkout the cpuid crate. In the future other
 # differences may appear.
-COVERAGE_DICT = {"Intel": 84.50, "AMD": 84.45}
+COVERAGE_DICT = {"Intel": 84.35, "AMD": 84.33}
```
sandreim, line 26: Coverage dropping. Does this PR contain new paths that cannot be easily covered ?

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,687 @@
...
+from framework.artifacts import ArtifactCollection, ArtifactSet
+from framework.matrix import TestMatrix, TestContext
+from framework.builder import MicrovmBuilder, SnapshotBuilder, SnapshotType
+
+import host_tools.network as net_tools  # pylint: disable=import-error
+
+
+def get_free_mem_ssh(ssh_connection):
+    """Get how much free memory a guest sees, over an ssh connection."""
+    def _available_mem(free_output):
```
sandreim, line 23: Can we not use a local function definition here?  it makes it harder to read.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,687 @@
...
+
+from conftest import _test_images_s3_bucket
+from framework.artifacts import ArtifactCollection, ArtifactSet
+from framework.matrix import TestMatrix, TestContext
+from framework.builder import MicrovmBuilder, SnapshotBuilder, SnapshotType
+
+import host_tools.network as net_tools  # pylint: disable=import-error
+
+
+def get_free_mem_ssh(ssh_connection):
```
sandreim, line 21: I would move this to the utils.py file. It might make sense in other tests as well.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,687 @@
...
+
+    # Get and open an ssh connection.
+    ssh_connection = net_tools.SSHConnection(test_microvm.ssh_config)
+
+    # Get the free memory before ballooning.
+    available_mem_deflated = get_free_mem_ssh(ssh_connection)
+
+    # Inflate 64 MB == 16384 page balloon.
+    response = test_microvm.balloon.patch(
+        num_pages=16384,
```
sandreim, line 246: Constant ?

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,687 @@
...
+        stats_polling_interval_s=0
+    )
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+    time.sleep(5)
```
sandreim, line 250: Do we really need to sleep 5 seconds and stall the CI pipeline ?

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,687 @@
...
+
+    # Start the microvm.
+    test_microvm.start()
+
+    # Get an ssh connection to the microvm.
+    ssh_connection = net_tools.SSHConnection(test_microvm.ssh_config)
+
+    # Inflate the balloon
+    response = test_microvm.balloon.patch(
+        num_pages=44000,
```
sandreim, line 289: Constant ?

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,687 @@
...
+    # Get an ssh connection to the microvm.
+    ssh_connection = net_tools.SSHConnection(test_microvm.ssh_config)
+
+    # Inflate the balloon.
+    response = test_microvm.balloon.patch(
+        num_pages=44000,
+        stats_polling_interval_s=0
+    )
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+    time.sleep(5)
```
sandreim, line 336: Same question, do we really need this ?

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,687 @@
...
+    # Check memory usage.
+    first_reading = get_rss_mem_by_pid(firecracker_pid)
+
+    # Have the guest drop its caches.
+    ssh_connection.execute_command('sync; echo 3 > /proc/sys/vm/drop_caches')
+    time.sleep(5)
+
+    # Now inflate the balloon.
+    response = test_microvm.balloon.patch(
+        num_pages=10000,
```
sandreim, line 453: Another constant ?

src/vmm/src/version_map.rs
```diff
@@ -5,13 +5,27 @@
...
 use lazy_static::lazy_static;
 use versionize::VersionMap;
+#[cfg(target_arch = "x86_64")]
+use versionize::Versionize;
 
 lazy_static! {
     // Note: until we have a better design, this needs to be updated when the version changes.
     /// Static instance used for handling microVM state versions.
     pub static ref VERSION_MAP: VersionMap = {
+        #[cfg(target_arch = "x86_64")]
```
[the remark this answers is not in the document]
georgepisaltu: We do because of the difference between architectures. Compiling without that `cfg` would lead to a warning at compilation on aarch64.

tests/host_tools/fillmem.c
```diff
@@ -0,0 +1,70 @@
...
+
+#define MB (1024 * 1024)
+
+
+int fill_mem(int mb_count) {
+    int i, j;
+    char *ptr = NULL;
+    for(j = 0; j < mb_count; j++) {
+        ptr = NULL;
+        while (ptr == NULL) {
```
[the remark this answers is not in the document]
georgepisaltu: There is a weird behavior on aarch64 when `mmap`ing the whole thing at once where you get a SIGSEGV before the OOM killer should send a SIGKILL in case of large memory chunks.

src/vmm/src/resources.rs
georgepisaltu: It currently mirrors the interface of all other devices. It is not *needed* but it looks symmetrical.

tests/integration_tests/build/test_coverage.py
georgepisaltu: It does, especially in the RPC side of things and with IO operations that fail because of external reasons. Will take another look to see where can I make some improvements.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: We do because the time it takes the guest driver to actually fulfill the device's requests is very volatile. I have tried multiple values for this and even implemented a system which periodically checks whether the assert is successful if it wasn't in the past, but this was actually the best option out of what I've tried.

tests/host_tools/fillmem.c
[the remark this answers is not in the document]
sandreim: Please add some code comments about this.

src/vmm/src/resources.rs
sandreim: I think it is more important to make it simpler than symmetrical.

tests/integration_tests/functional/test_balloon.py
sandreim: Sleep is not a good way to synchronize expectations with the guest. I am concerned about intermittent failures.

tests/host_tools/fillmem.c
```diff
@@ -0,0 +1,70 @@
...
+            ptr = mmap(NULL, MB * sizeof(char), PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+        }
+        for (i = 0; i < MB; i++) {
```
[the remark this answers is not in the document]
georgepisaltu: It's the same thing, just more explicit. This also doesn't get optimized away by the compiler.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: There is no way you can get notified by the firecracker process that the balloon driver has done the work. We can only poll the RSS and watch for changes. This chosen waiting time (5 seconds) is what I have (subjectively) determined to be the time I am willing to let the driver do something. If there is no result after the waiting time has passed, the test will consider it a failure.

The only other option that I see available is polling every second until we get the assert right, but I think it's messier.

src/vmm/src/device_manager/persist.rs
```diff
@@ -90,6 +106,29 @@ pub struct DeviceStates {
     pub net_devices: Vec<ConnectedNetState>,
     /// Vsock device state.
     pub vsock_device: Option<ConnectedVsockState>,
+    /// Balloon device state.
+    #[version(
+        start = 2,
+        default_fn = "balloon_default",
```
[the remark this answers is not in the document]
georgepisaltu: I removed it.

src/vmm/src/persist.rs
```diff
@@ -351,14 +362,21 @@ mod tests {
...
+        let mut version_map = VersionMap::new();
 
-        microvm_state
+        assert!(microvm_state
             .serialize(&mut buf.as_mut_slice(), &version_map, 1)
+            .is_err());
+
+        version_map
+            .new_version()
+            .set_type_version(DeviceStates::type_id(), 2);
```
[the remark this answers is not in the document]
georgepisaltu: There is an integration test that checks for this as it would have been hard to change the current unit tests to fit this need.

src/vmm/src/vmm_config/balloon.rs
```diff
@@ -0,0 +1,149 @@
...
+
+impl BalloonBuilder {
+    /// Creates an empty Balloon Store.
+    pub fn new() -> Self {
+        Self { inner: None }
+    }
+
+    /// Inserts a Balloon device in the store.
+    /// If an entry already exists, it will overwrite it.
+    pub fn insert(&mut self, cfg: BalloonDeviceConfig) -> Result<()> {
```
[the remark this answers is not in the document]
georgepisaltu: Done.

tests/host_tools/fillmem.c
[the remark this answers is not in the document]
georgepisaltu: I have replaced it with `memset` as it does not get optimized by the compiler. I encountered a problem elsewhere and made a wrong assumption. Please disregard my previous comment on this.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,687 @@
...
+def build_test_matrix(network_config, bin_cloner_path, logger):
+    """Build a test matrix using the kernel with the balloon driver."""
+    artifacts = ArtifactCollection(_test_images_s3_bucket())
+    # Testing matrix:
+    # - Guest kernel: Linux 4.9/4.14
+    # - Rootfs: Ubuntu 18.04
+    # - Microvm: 2vCPU with 512 MB RAM
+    # TODO: Multiple microvm sizes must be tested in the async pipeline.
+    microvm_artifacts = ArtifactSet(artifacts.microvms(keyword="2vcpu_256mb"))
+    kernel_artifacts = ArtifactSet(artifacts.kernels(
```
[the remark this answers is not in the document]
georgepisaltu: Done.

src/vmm/src/resources.rs
georgepisaltu: I made it simpler rather than symmetrical.

src/vmm/tests/integration_tests.rs
georgepisaltu: We have an integration test checking for this.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: Done.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: Removed.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: Removed.

PR #2101, opened 2020-08-27
aarch64: snapshot functionality

### LINKED ISSUE #2034: aarch64: add support for saving and restoring VmState
This will mostly mean saving and restoring the vgic's registers: distributor, redistributor and the state of the cpu interfaces.

### LINKED ISSUE #2035: aarch64: update seccomp list for allowing full snapshot create/load

### LINKED ISSUE #2036: aarch64: run successful CreateSnapshot/LoadSnapshot

dianpopa:
## Reason for This PR

Fixes #2036 
Fixes #2035 
Fixes #2034

## Description of Changes

Read commits.

## License Acceptance

## PR Checklist


src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -0,0 +1,209 @@
...
+        flags: 0,
+    };
+    if set == true {
+        fd.set_device_attr(&gic_dist_attr).map_err(|e| {
+            Error::DeviceAttribute(e, true, kvm_bindings::KVM_DEV_ARM_VGIC_GRP_DIST_REGS)
+        })?;
+    } else {
+        fd.get_device_attr(&mut gic_dist_attr).map_err(|e| {
+            Error::DeviceAttribute(e, false, kvm_bindings::KVM_DEV_ARM_VGIC_GRP_DIST_REGS)
+        })?;
```
serban300, line 90: Personally I would find it more readable if we had 2 separate methods for `set` and `get`. Or having an enum instead of the bool flag.

src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -0,0 +1,209 @@
...
+    let mut state = Vec::new();
+
+    for dreg in VGIC_DIST_REGS {
+        let mut base = dreg.base + REG_SIZE as u32 * dreg.bpi as u32;
+        let end = compute_reg_len(fd, &dreg, base)?;
+
+        while base < end {
+            let val: u32 = 0;
+            dist_attr_access(fd, base, &val, false)?;
+            state.push(val);
```
serban300, line 171: Nit: We can preinitialize the Vec with capacity for VGIC_DIST_REGS, and then just do `dist_attr_access(fd, base, &val[i], false)?;` . Push can lead to reallocations.

serban300 (commented):
As a general question I was curious if this logic applies both for GICv2 and v3. And can it diverge in the future between v2 and v3, v3.1, etc ? Because we have different classes for these devices, so AFAIU they can be quite different.

dianpopa (commented):
@serban300 addressed your comments, PTAL!

src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -0,0 +1,209 @@
...
+    Ok(end)
+}
+
+/// Set distributor registers of the GIC.
+pub fn set_dist_regs(fd: &DeviceFd, state: &Vec<u32>) -> Result<()> {
+    let mut idx = 0;
+
+    for dreg in VGIC_DIST_REGS {
+        let mut base = dreg.base + REG_SIZE as u32 * dreg.bpi as u32;
+        let end = compute_reg_len(fd, &dreg, base)?;
```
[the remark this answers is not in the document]
dianpopa: Done, take a look!

src/vmm/src/builder.rs
```diff
@@ -608,6 +609,14 @@ fn attach_legacy_devices_aarch64(
         .map_err(Error::RegisterMMIODevice)
 }
 
+#[cfg(target_arch = "aarch64")]
+fn init_vcpus(vm: &Vm, vcpus: &[Vcpu]) -> super::Result<()> {
+    for vcpu in vcpus {
+        vcpu.kvm_vcpu.init(vm.fd()).map_err(Error::VcpuInit)?;
```
aghecenco, line 624: Would it make sense to bake this `init()` in `aarch64::Vcpu::new`?

src/vmm/src/builder.rs
```diff
@@ -553,9 +549,9 @@ pub fn setup_interrupt_controller(
 /// Sets up the serial device.
 pub fn setup_serial_device(
     event_manager: &mut EventManager,
-    input: Box<dyn devices::legacy::ReadableFd + Send>,
     out: Box<dyn io::Write + Send>,
 ) -> super::Result<Arc<Mutex<Serial>>> {
+    let input = Box::new(SerialStdin::get());
```
aghecenco, line 562: I'm unsure about this because it removes an otherwise generic and configurable input source (`ReadableFd`) and replaces it with the very specific `SerialStdin`, which can't be substituted by anything else. I'm thinking of a user that consumes the programmatic API (which this fn is a part of, and which should stay flexible) and wants to plugin a fancy custom serial console that's not our `SerialStdin`.

aghecenco: How about making `SerialStdin` `pub(crate)`?

src/vmm/src/device_manager/mmio.rs
```diff
@@ -249,8 +249,13 @@ impl MMIODeviceManager {
...
         vm: &VmFd,
         serial: Arc<Mutex<devices::legacy::Serial>>,
+        dev_info_opt: Option<MMIODeviceInfo>,
     ) -> Result<()> {
-        let slot = self.allocate_new_slot(1)?;
+        let slot = if let Some(slot) = dev_info_opt {
+            slot
+        } else {
+            self.allocate_new_slot(1)?
+        };
```
aghecenco, line 251: nit: shorter:
```suggestion
        let slot = dev_info_opt.unwrap_or(self.allocate_new_slot(1)?);
```
(same below)

src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Relevant for registers that DO NOT share IRQs.
+    length: u16,
+}
+
+// All or at least the registers we are interested in are 32 bit, so
+// we use a constant for size(u32).
+const REG_SIZE: u8 = 4;
+
+// Creates a vgic distributor register.
+macro_rules! VGIC_DIST_REG {
```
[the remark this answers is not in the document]
serban300: I think it should work if you define the constructor as a const function: `const fn new() -> DistReg {}`.

src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -0,0 +1,220 @@
...
+/// in the documentation mentioned above.
+struct DistReg {
+    /// Offset from distributor address.
+    base: u32,
+    /// Bits per interrupt.
+    /// Relevant for registers that DO share IRQs.
+    bpi: u8,
+    /// Length of the register.
+    /// Relevant for registers that DO NOT share IRQs.
+    length: u16,
```
serban300, line 46: If I understand correctly, `bpi` and `length` are mutually exclusive ? If this is the case, maybe it would make sense to have 2 types of registers. Something like `SharedIrqReg` and `SimpleReg`. They can be united through a trait, enum, etc. This way `compute_reg_range` would be implemented differently for each one. I think it would make the 

```
        if self.length > 0 {
        ...
        }
        if self.bpi > 0 {
        ...
        }
```

logic more readable.

src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -0,0 +1,220 @@
...
+    VGIC_DIST_REG!(GICD_ICFGR, 2, 0),
+    VGIC_DIST_REG!(GICD_ICPENDR, 1, 0),
+    VGIC_DIST_REG!(GICD_ISPENDR, 1, 0),
+    VGIC_DIST_REG!(GICD_ICACTIVER, 1, 0),
+    VGIC_DIST_REG!(GICD_ISACTIVER, 1, 0),
+    VGIC_DIST_REG!(GICD_IPRIORITYR, 8, 0),
+];
+
+#[allow(clippy::trivially_copy_pass_by_ref)]
+fn kvm_set_dist_regs(fd: &DeviceFd, offset: u32, val: &u32) -> Result<()> {
```
serban300, line 102: Can we move this into the impl for `DistReg` ? Seems dist reg logic.

Also since it sets only one reg, should we remove the plural? `kvm_set_dist_reg` ?

The same comments for `kvm_get_dist_regs`.

src/arch/src/aarch64/gic/dist_regs.rs
dianpopa: We synced offline on this. For the moment const fn cannot be used from Traits so making 2 types of registers and then putting them in the same static list is not possible. We will leave this like this for the moment.

src/arch/src/aarch64/gic/dist_regs.rs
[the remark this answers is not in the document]
dianpopa: So I would need to do smth like this:
```
const fn new() {
 if (length > ) {
} else {}
}
```
Unfortunately, does not allow you to use `if` in a `const` function.

src/vmm/src/builder.rs
dianpopa: Indeed, I baked it into `Vcpu::new`, take a look.

src/vmm/src/builder.rs
dianpopa: Yup, I made it `pub (crate)`, take a look.

src/vmm/src/device_manager/mmio.rs
dianpopa: Done.

dianpopa (commented):
@serban300 @aghecenco addressed comments, PTAL!

aghecenco (dismissed):

src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -0,0 +1,210 @@
...
+            dreg.kvm_set_dist_reg(fd, base, &state[idx])?;
+            idx += 1;
+            base += REG_SIZE as u32;
+        }
+    }
+    Ok(())
+}
+
+/// Get distributor registers of the GIC.
+pub fn get_dist_regs(fd: &DeviceFd) -> Result<Vec<u32>> {
```
serban300, line 164: We discussed this offline. Just dropping it here.

Saving all the registers into a `Vec` might not leave us enough information. For example if we save on a firecracker version with n IRQs and try to restore on a possible later firecracker version with m IRQs, this will lead to assigning the values to wrong registers when restoring. It would be safer to have a structure that also remembers the exact register that the value belongs to. At least it would give us the option to do a translation in the future.

src/arch/src/aarch64/gic/icc_regs.rs
```diff
@@ -0,0 +1,225 @@
...
+                // calculate priority bits by reading the ctrl_el1 register.
+                icc_attr_access(fd, *icc_offset, i, &val, false)?;
+                // The priority bits are found in the ICC_CTLR_EL1 register (bits from  10:8).
+                // See page 194 from https://static.docs.arm.com/ihi0069/c/IHI0069C_gic_
+                // architecture_specification.pdf.
+                // Citation:
+                // "Priority bits. Read-only and writes are ignored. The number of priority bits
+                // implemented, minus one."
+                num_priority_bits =
+                    ((val & ICC_CTLR_EL1_PRIBITS_MASK) >> ICC_CTLR_EL1_PRIBITS_SHIFT) + 1;
```
serban300, line 115: Nit-ish: We have a specialized structure for reading bit ranges, for example 10:8. It is defined in the cpuid crate. It's called `BitRange`. We could move it to the `utils` crate an use it here instead of bit shifts.

dianpopa:
@acatangiu or @sandreim would you please help me with an additional opinion on this PR especially at commits that introduce "dev_mgr: adjust legacy registration on aarch64" and "vmm: save/restore for legacy devices on aarch64". Thanks!

src/vmm/src/device_manager/persist.rs
```diff
@@ -87,6 +97,9 @@ pub struct DeviceStates {
     pub net_devices: Vec<ConnectedNetState>,
     /// Vsock device state.
     pub vsock_device: Option<ConnectedVsockState>,
+    #[cfg(target_arch = "aarch64")]
+    // State of legacy devices in MMIO space.
+    pub legacy_devices: Option<Vec<ConnectedLegacyState>>,
```
acatangiu, line 102: nit: drop `Option`, can be empty `Vec` just like net or block.

src/vmm/src/device_manager/persist.rs
```diff
@@ -119,6 +134,23 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                 continue;
             }
 
+            #[cfg(target_arch = "aarch64")]
+            {
+                if *device_id == DeviceType::Serial.to_string()
+                    || *device_id == DeviceType::RTC.to_string()
```
acatangiu, line 140: nit: you can match on `device_type` directly instead of `device_id` and string transformations

src/vmm/src/device_manager/persist.rs
```diff
@@ -119,6 +134,23 @@ impl<'a> Persist<'a> for MMIODeviceManager {
...
+                {
+                    if states.legacy_devices.is_none() {
+                        states.legacy_devices = Some(Vec::new());
+                    }
+                    if let Some(legacy_devices) = states.legacy_devices.as_mut() {
```
acatangiu, line 145: nit: all of this becomes much simpler if you drop `Option<Vec>` and use only `Vec`

src/arch/src/aarch64/gic/dist_regs.rs
dianpopa: I gave it another thought and decided that doing this in this PR is pretty late in the game and I might break current working functionality or even degrade currently obtained performance. I prefer to do it in a subsequent PR after the CI is also setup to run aarch64 functionality. Opened #2234.

src/vmm/src/device_manager/persist.rs
```diff
@@ -78,6 +78,16 @@ pub struct ConnectedVsockState {
     pub mmio_slot: MMIODeviceInfo,
 }
 
+#[cfg(target_arch = "aarch64")]
+#[derive(Versionize)]
+/// Holds the state of a legacy device connected to the MMIO space.
+pub struct ConnectedLegacyState {
```
acatangiu, line 84: Would it make sense to also save `DeviceType`? Using the `device_id` to identify the device type works for now but seems fragile.

src/arch/src/aarch64/gic/icc_regs.rs
dianpopa: @serban300 I opened an issue for this #2236. I will do this as soon as I also integrate the snapshotting integration tests for aarch64 so that I make sure I do not break any functionality. Is this ok?

src/arch/src/aarch64/gic/dist_regs.rs
dianpopa: I also plan in actually add an integration test for this too so that I make sure that restoring from previous versions of Firecracker works. Is it ok with you?

src/vmm/src/device_manager/mmio.rs
```diff
@@ -249,8 +249,9 @@ impl MMIODeviceManager {
         &mut self,
         vm: &VmFd,
         serial: Arc<Mutex<devices::legacy::Serial>>,
+        dev_info_opt: Option<MMIODeviceInfo>,
```
acatangiu, line 245: nit: This works, but pls change function doc-comment to capture this change.

------------

For virtio devices the approach was to use two separate functions:
- `register_virtio_mmio_device()` does the actual registering
- `register_new_virtio_mmio_device()` allocates slot and calls the above. Also populates cmdline

You could do the same for Serial and RTC if you think it looks better.

dianpopa (commented):
@acatangiu I addressed your comments! @aghecenco @serban300 please prioritise taking another look, thanks!

acatangiu (commented):
LGTM 👍

src/arch/src/aarch64/gic/redist_regs.rs
```diff
@@ -0,0 +1,207 @@
...
+    Ok(())
+}
+
+fn access_redists_aux(
+    fd: &DeviceFd,
+    gicr_typer: &[u64],
+    state: &mut Vec<u32>,
+    reg_list: &'static [RdistReg],
+    idx: &mut usize,
+    set: bool,
```
serban300, line 105: I noticed a couple of issues here:
1. `idx` is only needed for set
2. state only needs to be mutable for get, but we always expect it to be `&mut`. Because of this, inside `set_redist_regs` we have to do `state.to_owned()` which AFAIU clones the data leading to an overhead.
3. I find `set: bool` a bit hard to read and easy to get wrong.

I would like to propose instead of `set: bool` something like 

```
enum Action<'a> {
    Set(&'a [u32], usize),
    Get(&'a mut Vec<u32>),
}
```
I'm not sure about the name. `Action` is not very suggestive. Anyway, then we can do:

```
fn access_redists_aux(
    fd: &DeviceFd,
    gicr_typer: &[u64],
    reg_list: &'static [RdistReg],
    action: &mut Action,
) -> Result<()> {
    for i in gicr_typer {
        for rdreg in reg_list {
            let mut base = rdreg.base;
            let end = base + rdreg.length as u32;

            while base < end {
                let mut val = 0;
                match action {
                    Action::Set(state, idx) => {
                        val = state[*idx];
                        redist_attr_access(fd, base, *i, &val, true)?;
                        *idx += 1;
                    }
                    Action::Get(state) => {
                        redist_attr_access(fd, base, *i, &val, false)?;
                        state.push(val);
                    }
                }
                base += REG_SIZE as u32;
            }
        }
    }
    Ok(())
}

/// Get redistributor registers.
pub fn get_redist_regs(fd: &DeviceFd, gicr_typer: &[u64]) -> Result<Vec<u32>> {
    let mut state = Vec::new();
    let mut action = Action::Get(&mut state);
    access_redists_aux(
        fd,
        &gicr_typer,
        VGIC_RDIST_REGS,
        &mut action,
    )?;

    access_redists_aux(
        fd,
        &gicr_typer,
        VGIC_SGI_REGS,
        &mut action,
    )?;
    Ok(state)
}

/// Set redistributor registers.
pub fn set_redist_regs(fd: &DeviceFd, gicr_typer: &[u64], state: &[u32]) -> Result<()> {
    let mut action = Action::Set(state, 0);

    access_redists_aux(
        fd,
        gicr_typer,
        VGIC_RDIST_REGS,
        &mut action,
    )?;
    access_redists_aux(
        fd,
        gicr_typer,
        VGIC_SGI_REGS,
        &mut action,
    )
}
```

src/arch/src/aarch64/gic/redist_regs.rs
```diff
@@ -0,0 +1,207 @@
...
+) -> Result<()> {
+    for i in gicr_typer {
+        for rdreg in reg_list {
+            let mut base = rdreg.base;
+            let end = base + rdreg.length as u32;
+
+            while base < end {
+                let mut val = 0;
+                if set {
+                    val = state[*idx];
```
serban300, line 115: I would skip this extra assignment. We could do something like `redist_attr_access(fd, base, *i, &state[*idx], true)?;`.

src/arch/src/aarch64/gic/redist_regs.rs
```diff
@@ -0,0 +1,207 @@
...
+    VGIC_RDIST_REG!(GICR_ICFGR0, 8),
+    VGIC_RDIST_REG!(GICR_ICPENDR0, 4),
+    VGIC_RDIST_REG!(GICR_ISPENDR0, 4),
+    VGIC_RDIST_REG!(GICR_ICACTIVER0, 4),
+    VGIC_RDIST_REG!(GICR_ISACTIVER0, 4),
+    VGIC_RDIST_REG!(GICR_IPRIORITYR0, 32),
+];
+
+#[allow(clippy::trivially_copy_pass_by_ref)]
+fn redist_attr_access(fd: &DeviceFd, offset: u32, typer: u64, val: &u32, set: bool) -> Result<()> {
```
serban300, line 82: I would call this `access_redist_attr`.

src/arch/src/aarch64/gic/redist_regs.rs
```diff
@@ -0,0 +1,207 @@
...
+        fd.set_device_attr(&gic_dist_attr)
+            .map_err(|e| Error::DeviceAttribute(e, true, KVM_DEV_ARM_VGIC_GRP_REDIST_REGS))?;
+    } else {
+        fd.get_device_attr(&mut gic_dist_attr)
+            .map_err(|e| Error::DeviceAttribute(e, false, KVM_DEV_ARM_VGIC_GRP_REDIST_REGS))?;
+    }
+    Ok(())
+}
+
+fn access_redists_aux(
```
serban300, line 99: Can we call this something like `access_redist_regs` or `access_redist_reg_list` ?

And can we also have such a function for the distributor regs ? In order to unify `set_dist_regs` and `get_dist_regs` ?

src/arch/src/aarch64/gic/redist_regs.rs
```diff
@@ -0,0 +1,207 @@
...
+const GICR_ICFGR0: u32 = GICR_SGI_OFFSET + 0x0C00;
+
+const KVM_DEV_ARM_VGIC_V3_MPIDR_SHIFT: u32 = 32;
+const KVM_DEV_ARM_VGIC_V3_MPIDR_MASK: u64 = 0xffff_ffff << KVM_DEV_ARM_VGIC_V3_MPIDR_SHIFT as u64;
+
+// All or at least the registers we are interested in are 32 bit, so
+// we use a constant for size(u32).
+const REG_SIZE: u8 = 4;
+
+/// This is how we represent the registers of a distributor.
```
serban300, line 34: nit: of a redistributor ?

src/arch/src/aarch64/gic/redist_regs.rs
```diff
@@ -0,0 +1,207 @@
...
+/// It is relrvant their offset from the base address of the
+/// distributor.
```
serban300, line 36: nit: redistributor ?

src/arch/src/aarch64/gic/redist_regs.rs
```diff
@@ -0,0 +1,207 @@
...
+/// Each register has a different number
+/// of bits_per_irq and is therefore variable length.
+/// First 32 interrupts (0-32) are private to each CPU (SGIs and PPIs).
+/// and so we save the first irq to identify between the type of the interrupt
+/// that the respective register deals with.
+struct RdistReg {
```
serban300, line 42: nit: I would call this `RedistReg`.

src/arch/src/aarch64/gic/redist_regs.rs
```diff
@@ -0,0 +1,207 @@
...
+
+#[allow(clippy::trivially_copy_pass_by_ref)]
+fn redist_attr_access(fd: &DeviceFd, offset: u32, typer: u64, val: &u32, set: bool) -> Result<()> {
+    let mut gic_dist_attr = kvm_device_attr {
+        group: KVM_DEV_ARM_VGIC_GRP_REDIST_REGS,
+        attr: (typer & KVM_DEV_ARM_VGIC_V3_MPIDR_MASK) | (offset as u64), // this needs the mpidr
+        addr: val as *const u32 as u64,
+        flags: 0,
+    };
+    if set {
```
serban300, line 89: Nit: We could also use an enum here instead of `set: bool`.

src/arch/src/aarch64/gic/icc_regs.rs
serban300: Yes, sounds good.

src/arch/src/aarch64/gic/dist_regs.rs
serban300: Yes, sounds good.

src/arch/src/aarch64/gic/dist_regs.rs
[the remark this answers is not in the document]
serban300: Hmmm maybe this was something related to the discussion about having 2 separate structures. With the current structure we should be able to do something like:

```
impl DistReg {
    const fn new(base: u32, bpi: u8, length: u16) -> DistReg {
        DistReg {
            base,
            bpi,
            length
        }
    }
}
```

The same comment for Redist.

src/arch/src/aarch64/gic/dist_regs.rs
serban300: I was thinking maybe we could do something to unify `set_dist_regs` and `get_dist_regs`. Maybe we'll revisit this as part of the improvement issues.

dianpopa (commented):
@serban300 addressed your comments! @acatangiu @aghecenco please re-review

src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -0,0 +1,206 @@
...
+            end = base + (self.bpi as u32 * (num_irq - super::super::layout::IRQ_BASE) / 8);
+            if self.bpi as u32 * (num_irq - super::super::layout::IRQ_BASE) % 8 > 0 {
+                end += REG_SIZE as u32;
+            }
+        }
+        Ok((base, end))
+    }
+
+    #[allow(clippy::trivially_copy_pass_by_ref)]
+    fn access_dist_regs(
```
serban300, line 105: Nit: It seems to access only one reg. Should we call it `access_dist_reg` ?

src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    for dreg in VGIC_DIST_REGS {
+        let (mut base, end) = dreg.compute_reg_range(fd)?;
+        while base < end {
+            let val: u32 = 0;
+            dreg.access_dist_regs(fd, base, &val, &Action::Get)?;
+            state.push(val);
+            base += REG_SIZE as u32;
+        }
+    }
```
serban300, line 156: Since `get_dist_regs` and `set_dist_regs` are very similar, can we use the same strategy that is used for redist regs and ICC and create a common method that is called both by `get_dist_regs` and `set_dist_regs` ?

src/arch/src/aarch64/gic/icc_regs.rs
```diff
@@ -0,0 +1,211 @@
...
+                || *icc_offset == SYS_ICC_AP1R2_EL1
+                || *icc_offset == SYS_ICC_AP1R3_EL1)
+                && num_priority_bits != 7
+            {
+                continue;
+            }
+            let val;
+            match action {
+                Action::Set(state, idx) => {
+                    val = state[*idx];
```
serban300, line 131: Nit: I would avoid the unnecessary allocation. We can just do `access_icc_attr(fd, *icc_offset, *i, &state[*idx], true)?;`.

dianpopa (commented):
@serban300 @acatangiu @aghecenco please re review! Thanks!

serban300 (approved):

aghecenco (approved):

PR #2112, opened 2020-09-02
snapshot: compute crc check before deserization

luminitavoicu:

## Reason for This PR

CRC checksum was computed after deserialization of the snapshot state file.

## Description of Changes

Changed API to compute and add CRC checksum during snapshot save/load and moved CRC check before snapshot deserialization.

## License Acceptance

## PR Checklist

### DIFF
--- a/docs/benchmarks/state-serialize.md
+++ b/docs/benchmarks/state-serialize.md
@@ -8,35 +8,35 @@ Snapshot size: 83886 bytes.
 - Architecture:        x86_64
 - CPU op-mode(s):      32-bit, 64-bit
 - Byte Order:          Little Endian
-- CPU(s):              12
+- CPU(s):              4
 - Thread(s) per core:  2
-- Core(s) per socket:  6
+- Core(s) per socket:  2
 - Socket(s):           1
 - NUMA node(s):        1
 - Vendor ID:           GenuineIntel
 - CPU family:          6
-- Model:               158
-- Model name:          Intel(R) Core(TM) i7-8700 - CPU @ 3.20GHz
-- Stepping:            10
-- CPU MHz:             800.077
-- CPU max MHz:         4600.0000
-- CPU min MHz:         800.0000
-- BogoMIPS:            6399.96
+- Model:               142
+- Model name:          Intel(R) Core(TM) i7-7600U CPU @ 2.80GHz
+- Stepping:            9
+- CPU MHz:             1100.008
+- CPU max MHz:         3900.0000
+- CPU min MHz:         400.0000
+- BogoMIPS:            5799.77
 - Virtualization:      VT-x
 - L1d cache:           32K
 - L1i cache:           32K
 - L2 cache:            256K
-- L3 cache:            12288K
-- NUMA node0 CPU(s):   0-11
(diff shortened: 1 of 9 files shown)


src/snapshot/src/lib.rs
```diff
@@ -128,25 +128,38 @@ impl Snapshot {
...
         O: Versionize,
     {
         let mut crc_reader = CRC64Reader::new(reader);
-        let format_vm = Self::format_version_map();
-        let object: O = Snapshot::load(&mut crc_reader, version_map)?;
+
+        // Extract snapshot data without stored checksum, which is 8 bytes in size
+        let raw_snapshot_len = snapshot_len - std::mem::size_of::<u64>();
+        let mut snapshot = vec![0u8; raw_snapshot_len];
+        let bytes_read = crc_reader.read(&mut snapshot).unwrap_or(0);
```
acatangiu, line 145: AFAIK `read()` does not guarantee reading all:

https://doc.rust-lang.org/std/io/trait.Read.html
> It is not an error if the returned value n is smaller than the buffer size, even when the reader is not at the end of the stream yet. This may happen for example because fewer bytes are actually available right now (e. g. being close to end-of-file) or because read() was interrupted by a signal.

Please use https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact

I'm also not sure if we need some sort of `BufferedReader` to avoid multiple underlying syscalls when using simple `Read`.

In a simple implementation I think the tradeoff between double buffering (using BufferedRead) and possible multiple syscalls would be more efficient to use double buffering...

@sandreim WDYT?

sandreim (commented):
PR looks good.

Please update the new perf numbers here: https://github.com/firecracker-microvm/firecracker/blob/master/docs/benchmarks/state-serialize.md.

And also here: https://s3.amazonaws.com/spec.ccfc.min/perf/snapshot-0.23/report/index.html

src/snapshot/src/lib.rs
luminitavoicu: Done

luminitavoicu:
@acatangiu @sandreim Comments addressed, please take another look

docs/benchmarks/state-serialize.md
```diff
@@ -8,35 +8,35 @@ Snapshot size: 83886 bytes.
...
 | Test                |      Mean     |
 |---------------------|---------------|
-| Serialize           |    354.18 us  |
-| Serialize + crc64   |    406.08 us  |
-| Deserialize         |    61.412 us  |
-| Deserialize + crc64 |    258.27 us  |
+| Serialize           |    371.38 us  |
+| Serialize + crc64   |    493.26 us  |
+| Deserialize         |    90.755 us  |
+| Deserialize + crc64 |    216.90 us  |
```
acatangiu, line 40: I'm reading this as new code is actually faster with CRC64 overhead of 240% instead of 422%

Any idea why?

src/snapshot/src/lib.rs
```diff
@@ -128,25 +128,37 @@ impl Snapshot {
...
         let mut crc_reader = CRC64Reader::new(reader);
-        let format_vm = Self::format_version_map();
-        let object: O = Snapshot::load(&mut crc_reader, version_map)?;
+
+        // Extract snapshot data without stored checksum, which is 8 bytes in size
+        let raw_snapshot_len = snapshot_len - std::mem::size_of::<u64>();
+        let mut snapshot = vec![0u8; raw_snapshot_len];
+        crc_reader
+            .read_exact(&mut snapshot)
+            .map_err(|ref err| Error::Io(err.raw_os_error().unwrap_or(0)))?;
```
acatangiu, line 147:
```suggestion
            .map_err(|ref err| Error::Io(err.raw_os_error().unwrap_or(libc::EINVAL)))?;
```

sandreim (dismissed):

acatangiu (dismissed):

docs/benchmarks/state-serialize.md
sandreim: Different CPU.

src/snapshot/src/lib.rs
luminitavoicu: Done

sandreim (approved):

acatangiu (approved):

PR #2127, opened 2020-09-09
Net: retry frame tx/rx in case of queue related error

serban300:
## Reason for This PR

Net: retry frame tx/rx in case of queue error

## Description of Changes

Net: retry frame tx/rx in case of queue error

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -211,14 +211,25 @@ impl Net {
         self.mmds_ns.as_mut()
     }
 
-    fn signal_used_queue(&self) -> result::Result<(), DeviceError> {
+    fn signal_used_queue(&mut self) -> result::Result<(), DeviceError> {
         self.interrupt_status
             .fetch_or(VIRTIO_MMIO_INT_VRING as usize, Ordering::SeqCst);
         self.interrupt_evt.write(1).map_err(|e| {
             error!("Failed to signal used queue: {:?}", e);
             METRICS.net.event_fails.inc();
             DeviceError::FailedSignalingUsedQueue(e)
-        })
+        })?;
+
+        self.rx_deferred_irqs = false;
+        Ok(())
+    }
+
+    fn signal_rx_used_queue(&mut self) -> result::Result<(), DeviceError> {
+        if self.rx_deferred_irqs {
+            return self.signal_used_queue();
+        }
+
+        Ok(())
     }
 
     // Attempts to copy a single frame into the guest if there is enough
@@ -442,12 +453,9 @@ impl Net {
             }
         }
 
-        if self.rx_deferred_irqs {
-            self.rx_deferred_irqs = false;
-            self.signal_used_queue()
-        } else {
-            Ok(())
-        }
+        // At this point we processed as many Rx frames as possible.
+        // We have to wake the guest if at least one descriptor chain has been used.
+        self.signal_rx_used_queue()
     }
 
     // Process the deferred frame first, then continue reading from tap.
@@ -456,13 +464,10 @@ impl Net {
             self.rx_deferred_frame = false;
             // process_rx() was interrupted possibly before consuming all
             // packets in the tap; try continuing now.
-            self.process_rx()
-        } else if self.rx_deferred_irqs {
-            self.rx_deferred_irqs = false;
-            self.signal_used_queue()
-        } else {
-            Ok(())
+            return self.process_rx();
         }
+
+        self.signal_rx_used_queue()
     }
 
     fn resume_rx(&mut self) -> result::Result<(), DeviceError> {
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -529,6 +529,7 @@ impl Net {
             self.tx_iovec.clear();
             while let Some(desc) = next_desc {
                 if desc.is_write_only() {
+                    self.tx_iovec.clear();
                     break;
                 }
                 self.tx_iovec.push((desc.addr, desc.len as usize));
@@ -569,12 +570,12 @@ impl Net {
                     }
                     Err(e) => {
                         error!("Failed to read slice: {:?}", e);
-                        if let GuestMemoryError::PartialBuffer { completed, .. } = e {
-                            read_count += completed;
-                            METRICS.net.tx_partial_reads.inc();
-                        } else {
-                            METRICS.net.tx_fails.inc();
+                        match e {
+                            GuestMemoryError::PartialBuffer { .. } => &METRICS.net.tx_partial_reads,
+                            _ => &METRICS.net.rx_fails,
                         }
+                        .inc();
+                        read_count = 0;
                         break;
                     }
                 }
@@ -1512,18 +1513,15 @@ pub mod tests {
 
         let desc_list = [(0, 100, 0), (1, 100, VIRTQ_DESC_F_WRITE), (2, 500, 0)];
         th.add_desc_chain(NetQueue::Tx, 0, &desc_list);
-        let frame = th.write_tx_frame(&desc_list, 700);
+        th.write_tx_frame(&desc_list, 700);
         th.event_manager.run_with_timeout(100).unwrap();
 
         // Check that the used queue advanced.
         assert_eq!(th.txq.used.idx.get(), 1);
         th.net().check_used_queue_signal(1);
         th.txq.check_used_elem(0, 0, 0);
-        // Check that the frame was partially sent to the tap.
-        let mut buf = vec![0; 1000];
-        assert!(tap_traffic_simulator.pop_rx_packet(&mut buf[vnet_hdr_len()..]));
-        assert_eq!(&buf[..100], &frame[..100]);
-        assert_eq!(&buf[100..1000], vec![0; 900].as_slice());
+        // Check that the frame was skipped.
+        assert!(!tap_traffic_simulator.pop_rx_packet(&mut []));
     }
 
     #[test]
@@ -1544,8 +1542,8 @@ pub mod tests {
         assert_eq!(th.txq.used.idx.get(), 1);
         th.net().check_used_queue_signal(1);
         th.txq.check_used_elem(0, 0, 0);
-        // Check that the frame wasn't sent to the tap.
-        assert!(!tap_traffic_simulator.pop_rx_packet(&mut [0; 1000]));
+        // Check that the frame was skipped.
+        assert!(!tap_traffic_simulator.pop_rx_packet(&mut []));
     }
 
     #[test]
@@ -1561,7 +1559,7 @@ pub mod tests {
         th.add_desc_chain(NetQueue::Tx, offset, &desc_list);
         let expected_len =
             (150 + th.mem.last_addr().raw_value() + 1 - th.txq.dtable[2].addr.get()) as usize;
-        let frame = th.write_tx_frame(&desc_list, expected_len);
+        th.write_tx_frame(&desc_list, expected_len);
         check_metric_after_block!(
             METRICS.net.tx_partial_reads,
             1,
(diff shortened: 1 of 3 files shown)


src/devices/src/virtio/net/device.rs
```diff
@@ -257,84 +277,94 @@ impl Net {
...
+        queue.add_used(mem, head_index, used_len).map_err(|e| {
+            error!("Failed to add available descriptor {}: {}", head_index, e);
+            FrontendError::AddUsed(e)
+        })?;
         self.rx_deferred_irqs = true;
 
-        if write_count >= self.rx_bytes_read {
-            METRICS.net.rx_bytes_count.add(write_count);
+        if let Ok(()) = result {
+            METRICS.net.rx_bytes_count.add(frame.len());
```
iulianbarbu, line 344: Are we always able to send the entire frame? I think we will miss deferring frames this way.

src/devices/src/virtio/net/device.rs
```diff
@@ -1291,7 +1314,7 @@ pub mod tests {
             ],
         );
         let frame = th.check_rx_deferred_frame(1000);
-        th.rxq.check_used_elem(0, 0, 100);
+        th.rxq.check_used_elem(0, 0, 0);
```
iulianbarbu, line 1317: So we can either send the entire frame or nothing? Is this intended?

src/devices/src/virtio/net/device.rs
```diff
@@ -257,84 +277,94 @@ impl Net {
...
-                    let source_slice = &self.rx_frame_buf[write_count..limit];
-                    let write_result = mem.write_slice(source_slice, desc.addr);
+        let queue = &mut self.queues[RX_INDEX];
+        let head_descriptor = queue.pop(mem).ok_or_else(|| {
+            METRICS.net.no_rx_avail_buffer.inc();
+            FrontendError::EmptyQueue
+        })?;
+        let head_index = head_descriptor.index;
+
+        let mut frame_slice = frame;
```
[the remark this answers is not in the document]
serban300: We need at least to remember the frame length. But I find it more natural to keep the frame reference. It shouldn't be that expensive.

src/devices/src/virtio/net/device.rs
```diff
@@ -257,84 +277,94 @@ impl Net {
...
+        let used_len = result.as_ref().map_or(0, |_| frame.len() as u32);
+        queue.add_used(mem, head_index, used_len).map_err(|e| {
+            error!("Failed to add available descriptor {}: {}", head_index, e);
+            FrontendError::AddUsed(e)
+        })?;
         self.rx_deferred_irqs = true;
 
-        if write_count >= self.rx_bytes_read {
-            METRICS.net.rx_bytes_count.add(write_count);
+        if let Ok(()) = result {
```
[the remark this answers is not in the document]
serban300: Done.

src/devices/src/virtio/net/device.rs
serban300: Yes, we should always be able to send the entire frame. Any Rx descriptor chain should be able to fit an entire frame. If not, it's a guest bug. We shouldn't miss deferring frames. I'm not completely sure what you mean. Could you expand on that please ?

src/devices/src/virtio/net/device.rs
```diff
@@ -442,12 +472,7 @@ impl Net {
             }
         }
 
-        if self.rx_deferred_irqs {
-            self.rx_deferred_irqs = false;
-            self.signal_used_queue()
-        } else {
-            Ok(())
-        }
+        self.signal_rx_used_queue()
```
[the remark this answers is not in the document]
serban300: I added a comment.

src/devices/src/virtio/net/device.rs
```diff
@@ -456,13 +481,10 @@ impl Net {
...
-            self.process_rx()
-        } else if self.rx_deferred_irqs {
-            self.rx_deferred_irqs = false;
-            self.signal_used_queue()
-        } else {
-            Ok(())
+            return self.process_rx();
         }
+
+        self.signal_rx_used_queue()
```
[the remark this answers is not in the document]
serban300: I think the context is self explanatory. We always signal rx used queue when there might be deferred irqs. And the name of the function is `handle_deferred_frame()`.

src/devices/src/virtio/net/device.rs
serban300: Yes, it's intended. We either send the entire frame or we skip the descriptor chain. There's no benefit in sending incomplete invalid frames.

src/devices/src/virtio/net/device.rs
iulianbarbu: My understanding is that the previous version of the code sends the rx received data even if the avail descriptors couldn't hold it entirely. If this is the case, the deferred frame is set to true, which means that the priority now is to send the rest of the rx frame instead of processing incoming data from the tap. You can observe this pattern throughout the code - e.g `process_tap_rx_event` or `handle_deffered_frame`.

> RX descriptor chain should be able to fit an entire frame, and if not, it's a guest bug. 

Do you have references about this certainty? I was not aware about it, and haven't found yet any reference myself.

src/devices/src/virtio/net/device.rs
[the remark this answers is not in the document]
serban300: `self.rx_bytes_count` can't be used everywhere because of the borrow checker. Sorry, I forgot to mention.

src/devices/src/virtio/net/device.rs
serban300: Now the behavior is the same. If the frame can't be sent, the descriptor chain is skipped, and the `deferred_frame` flag is set to true, cecause `write_frame_to_guest` returns `false`. This logic didn't change.

Quoting from the virtio standard:

```
5.1.6.3 Setting Up Receive Buffers

It is generally a good idea to keep the receive virtqueue as fully populated as possible: if it runs out, network performance will suffer.

If the VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_TSO6 or VIRTIO_NET_F_GUEST_UFO features are used, the maximum incoming packet will be to 65550 bytes long (the maximum size of a TCP or UDP packet, plus the 14 byte ethernet header), otherwise 1514 bytes. The 12-byte struct virtio_net_hdr is prepended to this, making for 65562 or 1526 bytes.
5.1.6.3.1 Driver Requirements: Setting Up Receive Buffers

    If VIRTIO_NET_F_MRG_RXBUF is not negotiated:
        If VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_TSO6 or VIRTIO_NET_F_GUEST_UFO are negotiated, the driver SHOULD populate the receive queue(s) with buffers of at least 65562 bytes.
        Otherwise, the driver SHOULD populate the receive queue(s) with buffers of at least 1526 bytes.
    If VIRTIO_NET_F_MRG_RXBUF is negotiated, each buffer MUST be at least the size of the struct virtio_net_hdr.

Note: Obviously each buffer can be split across multiple descriptor elements.

If VIRTIO_NET_F_MQ is negotiated, each of receiveq1…receiveqN that will be used SHOULD be populated with receive buffers. 
```

src/devices/src/virtio/net/device.rs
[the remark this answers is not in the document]
serban300: I don't have strong preferences here. Personally I wouldn't find a comment very helpful.

src/devices/src/virtio/net/device.rs
```diff
@@ -456,13 +464,10 @@ impl Net {
             self.rx_deferred_frame = false;
             // process_rx() was interrupted possibly before consuming all
             // packets in the tap; try continuing now.
-            self.process_rx()
-        } else if self.rx_deferred_irqs {
-            self.rx_deferred_irqs = false;
```
aghecenco, line 461: AFAICT, this could have been a bug that you fixed: if inside `signal_used_queue`, `interrupt_evt.write(1)` failed for whatever reason, this line would have marked no more deferred irqs, but the irq would have remain not delivered. With your changes,  `rx_deferred_irqs` is only set to `false` (deferred irqs are marked "handled") *after* successfully signaling the guest. This way, if there's an error in the interrupt event handling, the irq remains there for later retriggering. Did I get that right?

src/devices/src/virtio/net/device.rs
```diff
@@ -268,84 +277,94 @@ impl Net {
         success
     }
 
-    // Copies a single frame from `self.rx_frame_buf` into the guest. Returns true
-    // if a buffer was used, and false if the frame must be deferred until a buffer
-    // is made available by the driver.
-    fn rx_single_frame(&mut self) -> bool {
+    // Copies a single frame from `self.rx_frame_buf` into the guest.
+    pub fn do_write_frame_to_guest(&mut self) -> std::result::Result<(), FrontendError> {
```
aghecenco, line 281: No need for `pub`.

src/devices/src/virtio/net/device.rs
```diff
@@ -34,6 +34,15 @@ use virtio_gen::virtio_net::{
 };
 use vm_memory::{ByteValued, Bytes, GuestAddress, GuestMemoryError, GuestMemoryMmap};
 
+pub enum FrontendError {
```
aghecenco, line 37: No need for `pub`. This error type is completely swallowed by `write_frame_to_guest` which converts it (sort of) into a `bool`.

src/devices/src/virtio/net/device.rs
```diff
@@ -34,6 +34,15 @@ use virtio_gen::virtio_net::{
...
+    AddUsed(QueueError),
+    DescriptorChainTooSmall,
+    EmptyQueue,
+    GuestMemory(GuestMemoryError),
+    ReadOnlyDescriptor,
+    WriteOnlyDescriptor,
```
aghecenco, line 43: This isn't used anywhere.

src/devices/src/virtio/net/device.rs
```diff
@@ -34,6 +34,15 @@ use virtio_gen::virtio_net::{
 };
 use vm_memory::{ByteValued, Bytes, GuestAddress, GuestMemoryError, GuestMemoryMmap};
 
+pub enum FrontendError {
+    AddUsed(QueueError),
```
aghecenco, line 38: The inner `QueueError` is discarded. Any reason to propagate it at all?

src/devices/src/virtio/net/device.rs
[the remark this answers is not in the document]
aghecenco: I'd also prefer `frame` to go.
```rust
let mut frame_slice = &self.rx_frame_buf[..self.rx_bytes_read];
let frame_len = frame_slice.len();
```
gets rid of the extra ref and 2 calls to `frame.len()`.

src/devices/src/virtio/net/device.rs
```diff
@@ -268,84 +277,94 @@ impl Net {
...
-        } else {
-            false
         }
+        result
+    }
+
+    // Copies a single frame from `self.rx_frame_buf` into the guest. In case of an error retries
+    // the operation if possible. Returns true if the operation was successfull.
+    fn write_frame_to_guest(&mut self) -> bool {
+        let max_iterations = self.queues[RX_INDEX].actual_size();
```
aghecenco, line 353: Why this number of retries as opposed to something hardcoded?

src/devices/src/virtio/net/device.rs
serban300: You're right. I didn't realize that. Fixed.

src/devices/src/virtio/net/device.rs
serban300: To be honest, I didn't think about this initially. I just wanted to make it easier to follow the `rx_deferred_irqs` flag. But yes, looks like it addresses a corner case.

serban300:
@aghecenco  @iulianbarbu 

Thanks for the review. I addressed all your comments. PTAL when you have time !

aghecenco (approved):

src/devices/src/virtio/net/device.rs
```diff
@@ -257,84 +276,94 @@ impl Net {
...
+        let head_descriptor = queue.pop(mem).ok_or_else(|| {
+            METRICS.net.no_rx_avail_buffer.inc();
+            FrontendError::EmptyQueue
+        })?;
+        let head_index = head_descriptor.index;
+
+        let mut frame_slice = &self.rx_frame_buf[..self.rx_bytes_read];
+        let frame_len = frame_slice.len();
+        let mut maybe_next_descriptor = Some(head_descriptor);
+        while let Some(descriptor) = &maybe_next_descriptor {
```
iulianbarbu, line 298:
```
5.1.6.4 Processing of Incoming Packets

When a packet is copied into a buffer in the receiveq, the optimal path is to disable further used buffer notifications for the receiveq and process packets until no more are found, then re-enable them.

Processing incoming packets involves:

    num_buffers indicates how many descriptors this packet is spread over (including this one): this will always be 1 if VIRTIO_NET_F_MRG_RXBUF was not negotiated. This allows receipt of large packets without having to allocate large buffers: a packet that does not fit in a single buffer can flow over to the next buffer, and so on. In this case, there will be at least num_buffers used buffers in the virtqueue, and the device chains them together to form a single packet in a way similar to how it would store it in a single buffer spread over multiple descriptors. The other buffers will not begin with a struct virtio_net_hdr.
    If num_buffers is one, then the entire packet will be contained within this buffer, immediately following the struct virtio_net_hdr. 
```

We do not negotiate for `VIRTIO_NET_F_MRG_RXBUF` and this means that the packets will always fit a buffer, which is represented by a single descriptor. I think this means that we do not need to process descriptor chains after all and can directly use a single descriptor to hold incoming packets. @serban300 , have you verified if the packets are written to a descriptor chain ( >1 available descriptors) or to a single one?

src/devices/src/virtio/net/device.rs
```diff
@@ -257,84 +276,94 @@ impl Net {
...
+        if result.is_ok() && !frame_slice.is_empty() {
+            warn!("Receiving buffer is too small to hold frame of current size");
+            METRICS.net.rx_fails.inc();
+            result = Err(FrontendError::DescriptorChainTooSmall);
+        }
 
-        // Mark that we have at least one pending packet and we need to interrupt the guest.
+        // Mark the descriptor chain as used. If an error occurred, skip the descriptor chain.
+        let used_len = if result.is_err() { 0 } else { frame_len as u32 };
+        queue.add_used(mem, head_index, used_len).map_err(|e| {
```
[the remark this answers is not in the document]
serban300: If we provide a descriptor chain with 0 used bytes, the driver will simply skip it. This is not a waste of guest cycles. We actually have to notify the guest that the descriptor chain has been used, because otherwise the guest might not provide new descriptor chains for us. It will think that we already have enough available.

This is covered in the unit tests. But there we don't get to see the actual driver behavior. I can't think of a way to trigger it in the integration tests.

src/devices/src/virtio/net/device.rs
serban300: I tried this and I'm getting descriptor chains with multiple buffer of size 12 or 4080. I'm not sure why. What you're saying might be part of a new version of the virtio standard. We're using 0.95 as far as I know. Anyway, I would keep the logic for processing descriptor chains since it's already implemented and we might need it in the future.

iulianbarbu (dismissed):

iulianbarbu (approved):

aghecenco (approved):

PR #2135, opened 2020-09-14
Expose the balloon device through the API

georgepisaltu:
## Reason for This PR

#1571 

## Description of Changes

Exposed the balloon device interface to the vmm and API server in order to make the device accessible to users.

## License Acceptance

## PR Checklist


tests/integration_tests/functional/test_api.py
```diff
@@ -744,3 +744,60 @@ def test_api_vsock(test_microvm_with_api):
...
+    # Overwriting the existing device should give an error now.
+    response = test_microvm.balloon.put(
+        num_pages=3,
+        deflate_on_oom=False,
+        must_tell_host=True,
+        stats_polling_interval_s=3
+    )
+    assert test_microvm.api_session.is_status_bad_request(response.status_code)
+
+    time.sleep(1)
```
gc-plp, line 796: What's the purpose of calling `time.sleep(1)`?

tests/framework/utils.py
```diff
@@ -89,6 +89,22 @@ def get_files_from(find_path: str, pattern: str, exclude_names: list = None,
...
 
 
+def get_free_mem_ssh(ssh_connection):
+    """
+    Get how much free memory a guest sees, over an ssh connection.
+
+    :param ssh_connection: connection to the guest
+    :return: available mem column output of 'free'
+    """
+    _, stdout, stderr = ssh_connection.execute_command('free')
```
gc-plp, line 99: Wouldn't `cat /proc/meminfo | grep MemAvailable` be a more reliable method of extracting the data? 
`free` uses `meminfo` behind the scenes, so probably we could avoid the for loop:

def get_free_mem_ssh(ssh_connection):
    """
    Get how much free memory a guest sees, over an ssh connection.
    :param ssh_connection: connection to the guest
    :return: available mem column output of 'free'
    """
    _, stdout, stderr = ssh_connection.execute_command('cat /proc/meminfo | grep MemAvailable')
    assert stderr.read() == ''
    
    # Split "MemAvailable:   123456 kB" and validate it
    meminfo_data = stdout.split(" ")
    if len(meminfo_data) == 3:
            return free_kb[1]
        
    raise Exception('Available memory not found in `free` output')

tests/framework/utils.py
```diff
@@ -89,6 +89,22 @@ def get_files_from(find_path: str, pattern: str, exclude_names: list = None,
     return found
 
 
+def get_free_mem_ssh(ssh_connection):
+    """
+    Get how much free memory a guest sees, over an ssh connection.
```
gc-plp, line 94: I would add that we return the amount of free memory in kB

tests/integration_tests/functional/test_api.py
georgepisaltu: Removed.

tests/framework/utils.py
georgepisaltu: Done.

tests/framework/utils.py
georgepisaltu: I think the `free` output is also pretty standard, but I have no strong opinion on this. I could change this if it makes a difference.

aghecenco: I vote for `/proc/meminfo` too, particularly if in the future we might care about finer grained stats and since it's not a huge change (hopefully :smile:)

tests/host_tools/fillmem.c
```diff
@@ -0,0 +1,80 @@
...
+int fill_mem(int mb_count) {
+    int i, j;
+    char *ptr = NULL;
+    for(j = 0; j < mb_count; j++) {
+        ptr = NULL;
+        while (ptr == NULL) {
+            // We can't map the whole chunk of memory at once because
+            // on aarch64 when we try to allocate a large amount of
+            // memory to OOM the system, it often results in the
+            // process getting killed by a SIGSEGV instead of the
```
aghecenco, line 30: Interesting, any idea why?

tests/host_tools/fillmem.c
```diff
@@ -0,0 +1,80 @@
...
+            char buf[200];
+            sprintf(buf, "OOM Killer stopped the program with signal %d, exit code %d\n", WTERMSIG(status), WEXITSTATUS(status));
+            write(fd, buf, strlen(buf) + 1);
+            close(fd);
+
+            return 0;
+        }
+
+        write(fd, "Memory filling was successful\n", 31);
+        return 0;
```
aghecenco, line 76: nit: close the fd (or remove the `close` at line 70....but `close`ing fds is the better practice_

tests/host_tools/fillmem.c
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    return -2;
```
aghecenco, line 79:
```suggestion
        } else {
            write(fd, "Memory filling was successful\n", 31);
        }
        close(fd);
        return 0;
    }

    return -2;
```

aghecenco: Actually I'd remove the unreachable `return -2` too.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,713 @@
...
+        assert exit_code == 0 and (
+            "OOM Killer stopped the program with "
+            "signal 9, exit code 0"
+        ) in stdout.read()
+    else:
+        assert exit_code == 0 and (
+            "Memory filling was "
+            "successful"
+        ) in stdout.read()
+    time.sleep(0.1)
```
aghecenco, line 52: Is this needed?

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,713 @@
...
+def make_guest_dirty_memory(ssh_connection, should_oom=False, amount=8192):
+    """Tell the guest, over ssh, to dirty `amount` pages of memory."""
+    amount_in_mbytes = amount / MB_TO_PAGES
+
+    exit_code, _, _ = ssh_connection.execute_command(
+        "/sbin/fillmem {}".format(amount_in_mbytes)
+    )
+    time.sleep(0.1)
+
+    cmd = "cat /tmp/fillmem_output.txt"
```
aghecenco, line 40: Looks like this could be more deterministic if you only created and filled the file (in the C code) *after* you run the memory load, and you wait here for the file to show up on disk and be non-zero-sized.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,713 @@
...
+        stats_polling_interval_s=stats_interval
+    )
+    assert microvm.api_session.is_status_no_content(response.status_code)
+
+    return microvm
+
+
+def copy_fillmem_to_rootfs(rootfs_path):
+    """Build and copy the 'memfill' program to the rootfs."""
+    subprocess.check_call("gcc ./host_tools/fillmem.c -o fillmem", shell=True)
```
aghecenco, line 144: I think this belongs in the `bin_cloner_path` fixture

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,713 @@
...
+    subprocess.check_call("cp fillmem tmpfs/sbin/fillmem", shell=True)
+    subprocess.check_call("rm fillmem", shell=True)
+    subprocess.check_call("umount tmpfs", shell=True)
+    subprocess.check_call("rmdir tmpfs", shell=True)
+
+
+# pylint: disable=C0103
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
+    reason="The balloon device is less effective on arm64 architecture"
```
aghecenco, line 156: Why? What does "less effective" mean? I think this deserves at least a brief explanation, even if `aarch64` support is experimental; maybe especially because it's experimental, to offer info for new contributors / customers in the absence of SLAs

src/devices/src/virtio/balloon/persist.rs
```diff
@@ -20,12 +20,92 @@ use crate::virtio::balloon::device::{BalloonStats, ConfigSpace};
...
 use crate::virtio::{DeviceState, Queue};
 
-#[derive(Versionize)]
+#[derive(Clone, Versionize)]
+pub struct BalloonConfigSpaceState {
+    num_pages: u32,
+    actual_pages: u32,
+}
+
+impl BalloonConfigSpaceState {
```
sandreim, line 29: I would drop this impl since these fns are used only once.

src/vmm/src/device_manager/persist.rs
```diff
@@ -87,6 +103,21 @@ pub struct DeviceStates {
...
+    /// Balloon device state.
+    #[version(start = 2, ser_fn = "balloon_serialize")]
+    pub balloon_device: Option<ConnectedBalloonState>,
+}
+
+impl DeviceStates {
+    fn balloon_serialize(&mut self, target_version: u16) -> VersionizeResult<()> {
+        if target_version < 2 && self.balloon_device.is_some() {
+            return Err(VersionizeError::Semantic(
+                "Balloon is not compatible with the target version.".to_owned(),
```
sandreim, line 115: I think we should be more specific: "Target version does not implement the virtio-balloon device."

tests/host_tools/fillmem.c
sandreim: I appreciate the comments here, but it is a bit vague (often results?) and does not highlight the root cause.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,713 @@
...
+    microvm.start()
+
+    logger.info("Create {} #0.".format(snapshot_type))
+    # Create a snapshot builder from a microvm.
+    snapshot_builder = SnapshotBuilder(microvm)
+
+    # Create a snapshot with a balloon on version 0.23.0.
+    # This should fail as the balloon was introduced in 0.24.0.
+    try:
+        snapshot_builder.create(
```
sandreim, line 694: Does this test fail if this create does not fail as it should?

src/devices/src/virtio/balloon/persist.rs
```diff
@@ -20,12 +20,92 @@ use crate::virtio::balloon::device::{BalloonStats, ConfigSpace};
...
+        }
+    }
+}
+
+#[derive(Clone, Versionize)]
+pub struct BalloonStatsState {
+    target_pages: u32,
+    actual_pages: u32,
+    target_mb: u32,
+    actual_mb: u32,
```
sandreim, line 50: Are the target_* fields part of the stats or user input ?

sandreim: Also from a state snapshot point of view, actual_pages/mb seems to be duplicated in config space.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,713 @@
...
+        context=test_context,
+        artifact_sets=[
+            microvm_artifacts,
+            kernel_artifacts,
+            disk_artifacts
+        ]
+    )
+
+
+def create_vm(context, deflate=True, stats_interval=0):
```
sandreim, line 90: You can use VMMicro.spawn() to simplify this function.
Why are you enabling diff snapshots ?

sandreim: Also, I do not see where this fn is called from ?

tests/integration_tests/functional/test_balloon.py
```diff
@@ -0,0 +1,713 @@
...
+        enable_diff_snapshots=enable_diff_snapshots
+    )
+
+    network_config = net_tools.UniqueIPv4Generator.instance()
+    _, host_ip, guest_ip = microvm.ssh_network_config(
+        network_config,
+        '1',
+        tapname="tap0"
+    )
+    logger.debug("Host IP: {}, Guest IP: {}".format(host_ip, guest_ip))
```
sandreim, line 126: You need to remove this, build() already sets up networking and IPs.

tests/integration_tests/functional/test_balloon.py
sandreim: Not sure we want to rely on fixtures for this. I would prefer to do it explicitly.

tests/integration_tests/functional/test_balloon.py
sandreim: This is not a good reason to disable ARM tests. Why is the test not working for ARM?

src/vmm/src/lib.rs
```diff
@@ -46,13 +46,18 @@ use crate::device_manager::mmio::MMIODeviceManager;
 use crate::memory_snapshot::SnapshotMemory;
 #[cfg(target_arch = "x86_64")]
 use crate::persist::{MicrovmState, MicrovmStateError, VmInfo};
+use crate::rpc_interface::{ActionResult, VmmActionError, VmmData};
```
acatangiu, line 49: `rpc_interface` is one layer above the `vmm` (for now, only conceptually, later we should actually enforce it by taking it out of the `Vmm` crate).

This file `vmm/src/lib.rs` is strictly _vmm programmatic interface_, it should not have any knowledge (aka dependencies) to upper layers such as the `rpc_interface`.

Don't use `rpc_interface` here.

src/vmm/src/lib.rs
```diff
@@ -46,13 +46,18 @@ use crate::device_manager::mmio::MMIODeviceManager;
...
+use crate::vmm_config::balloon::{
```
acatangiu, line 50: Same with `vmm_config`: this is something RPC/API specific. The `vmm` implementation doesn't need to know about it (depend on it).

src/vmm/src/lib.rs
```diff
@@ -493,6 +498,116 @@ impl Vmm {
...
+            ));
+        }
+
+        Ok(())
+    }
+
+    /// Updates configuration for the balloon device as described in `balloon_stats_update`.
+    fn update_balloon_stats_config(
+        &mut self,
+        balloon_stats_update: BalloonUpdateStatsConfig,
```
acatangiu, line 578: This can directly take `stats_polling_interval_s`.

Either that, or define `BalloonUpdateStatsConfig` as part of the `devices::virtio::balloon` if it makes sense.

src/vmm/src/lib.rs
```diff
@@ -493,6 +498,116 @@ impl Vmm {
...
+            ))
+        } else {
+            Err(VmmActionError::BalloonConfig(
+                BalloonConfigError::DeviceNotFound,
+            ))
+        }
+    }
+
+    /// Updates configuration for the balloon device as described in `balloon_update`.
+    fn update_balloon_config(&mut self, balloon_update: BalloonUpdateConfig) -> ActionResult {
```
acatangiu, line 535: This can directly take `amount_mb`.

Either that, or define `BalloonUpdateConfig` as part of the `devices::virtio::balloon` if it makes sense.

src/devices/src/virtio/balloon/device.rs
```diff
@@ -154,7 +183,11 @@ impl Balloon {
             EventFd::new(libc::EFD_NONBLOCK).map_err(BalloonError::EventFd)?,
         ];
 
-        let queues = QUEUE_SIZES.iter().map(|&s| Queue::new(s)).collect();
+        let mut queues: Vec<Queue> = QUEUE_SIZES.iter().map(|&s| Queue::new(s)).collect();
+
+        if stats_polling_interval_s == 0 {
+            let _ = queues.pop();
```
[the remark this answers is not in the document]
acatangiu: This expects that the stats queue is always last. Is it safe to make that assumption - is the queue order defined in the standard or smth?

If not, I suggest moving this `queues.pop()` to a `remove_stats_queue()` function and add comments to both the queues building code and that function to make the ordering dependency easy to see.

acatangiu (changes requested):

tests/host_tools/fillmem.c
georgepisaltu: It works as expected on x86_64, but on arm64 there is an almost arbitrary interval of values for which allocations will succeed but will result in an OOM death of the process, but if we try to allocate memory of sizes above that interval it results in a straight SIGSEGV. From my experiments, the window for this successful death by OOM killer seems to be from the point of memory pressure to about 18 MB above that, and of course the point at which the system enters a memory pressure state varies from run to run. I don't know why this happens and I don't think it's relevant for this PR.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: I will just remove it altogether as the program should have written every byte and closed the fd by the time control is returned in the test from the ssh command.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: Less effective in this case means an approximately 60% drop in page throughput when it comes to inflate operations. I will, however, make the test run on arm64 and tweak the assertions to only check if there is an effect, rather than checking if there is a sufficiently large effect.

src/devices/src/virtio/balloon/device.rs
```diff
@@ -41,11 +41,25 @@ macro_rules! mem_of_active_device {
     };
 }
 
+macro_rules! mb_to_pages {
```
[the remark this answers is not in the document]
georgepisaltu: It doesn't feel like this fn would belong in the `Balloon` structure as a whole, as it is not device related. I chose to use a macro instead of always multiplying with the constant to avoid duplicating code.

src/devices/src/virtio/balloon/device.rs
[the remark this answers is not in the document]
georgepisaltu: Yes, the order is stated in the standard and the stats queue has the index 2, so it's the last queue.

src/vmm/src/lib.rs
georgepisaltu: Why `devices::virtio::balloon` and not `vmm::vmm_config::balloon`?

Also, I would prefer not to completely drop these structures as they are used in deserializing the API requests, but I agree the vmm functions should take the inner directly.

acatangiu: I don't have a strong preference where we define the structures as long as we don't break the logical dependency tree. `vmm_config` might be confusingly named, but if you look inside it you can see it defines API and RPC level configuration structures. The API component sits on top of RPC which sits on top of `Vmm`. We want the `Vmm` to be independent of API and RPC therefore we shouldn't take any dependencies like that.

We have an uncompleted decoupling item https://github.com/firecracker-microvm/firecracker/issues/1713 that would actually enforce the above and eliminate confusions.

src/devices/src/virtio/balloon/device.rs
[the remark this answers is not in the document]
acatangiu: In this case a comment here should suffice.

src/devices/src/virtio/balloon/device.rs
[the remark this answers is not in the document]
acatangiu: You can have a module/file level function that would not be part of `Balloon` (or any other struct for that matter).

The dedup is fine, the suggestion here is only to use a fn instead of a macro.
```suggestion
fn mb_to_pages(amount_mb: u32) -> std::result::Result<u32, BalloonError> {
```

src/devices/src/virtio/balloon/device.rs
```diff
@@ -41,11 +41,21 @@ macro_rules! mem_of_active_device {
...
 
+fn mb_to_pages(amount_mb: u32) -> Result<u32, BalloonError> {
+    amount_mb
+        .checked_mul(MB_TO_4K_PAGES)
+        .ok_or(BalloonError::TooManyPagesRequested)
+}
+
+fn pages_to_mb(amount_pages: u32) -> u32 {
+    amount_pages / MB_TO_4K_PAGES
+}
```
acatangiu, line 52: Do you treat cases where # of pages is not multiple of MBs? Should there be a comment here at least explaining why that shouldn't happen or smth?

200 pages here returns 0 MB for example.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: I removed it completely.

src/devices/src/virtio/balloon/mod.rs
```diff
@@ -47,16 +53,28 @@ const VIRTIO_BALLOON_S_HTLB_PGFAIL: u16 = 9;
 pub enum Error {
     /// Activation error.
     Activate(super::ActivateError),
+    /// No balloon device found.
+    DeviceNotFound,
+    /// Device was not activated because the driver is missing.
+    DriverNotFound,
```
acatangiu, line 59: nitish:
```suggestion
    /// Device not activated yet.
    DeviceNotActive,
```

I would avoid to imply knowing _why_ the guest hasn't activated the device (yet).

src/devices/src/virtio/balloon/persist.rs
georgepisaltu: Removed them as they are populated when a GET "/balloon/statistics" anyway.

src/vmm/src/device_manager/persist.rs
georgepisaltu: Changed.

src/devices/src/virtio/balloon/persist.rs
georgepisaltu: Done.

src/devices/src/virtio/balloon/device.rs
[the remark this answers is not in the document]
georgepisaltu: Added the comment and made the drop more specific.

src/vmm/src/lib.rs
```diff
@@ -46,13 +46,19 @@ use crate::device_manager::mmio::MMIODeviceManager;
 use crate::memory_snapshot::SnapshotMemory;
 #[cfg(target_arch = "x86_64")]
 use crate::persist::{MicrovmState, MicrovmStateError, VmInfo};
+/* use crate::rpc_interface::{ActionResult, VmmActionError, VmmData};
```
acatangiu, line 49: leftover commented code

src/devices/src/virtio/balloon/device.rs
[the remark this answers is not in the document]
georgepisaltu: Changed to `fn`s.

tests/host_tools/fillmem.c
georgepisaltu: Done.

src/vmm/src/lib.rs
```diff
@@ -493,6 +498,116 @@ impl Vmm {
...
+                virtio_device
+                    .lock()
+                    .expect("Poisoned lock")
+                    .as_mut_any()
+                    .downcast_mut::<Balloon>()
+                    .unwrap()
+                    .update_stats_polling_interval(balloon_stats_update.stats_polling_interval_s)
+                    .map_err(|_| {
+                        VmmActionError::BalloonConfig(BalloonConfigError::InvalidStatsUpdate)
+                    })?;
```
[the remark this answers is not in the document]
georgepisaltu: Done.

src/vmm/src/lib.rs
```diff
@@ -493,6 +498,116 @@ impl Vmm {
...
+                .map_err(|err| {
+                    VmmActionError::BalloonConfig(BalloonConfigError::UpdateFailure(err))
+                })?;
+        } else {
+            return Err(VmmActionError::BalloonConfig(
+                BalloonConfigError::DeviceNotFound,
+            ));
+        }
+
+        Ok(())
```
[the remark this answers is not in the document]
georgepisaltu: Done.

src/vmm/src/vmm_config/balloon.rs
```diff
@@ -0,0 +1,186 @@
...
+
+/// Errors associated with the operations allowed on the balloon.
+#[derive(Debug)]
+pub enum BalloonConfigError {
+    /// The user made a request on an inexistent balloon device.
+    DeviceNotFound,
+    /// The user tried to enable/disable the statistics after boot.
+    InvalidStatsUpdate,
+    /// Amount of pages requested cannot fit in `u32`.
+    MemorySizeOverflow,
```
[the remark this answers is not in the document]
georgepisaltu: Changed.

tests/framework/utils.py
georgepisaltu: Used `/proc/meminfo`.

tests/host_tools/fillmem.c
georgepisaltu: I found out that the SIGSEGV mentioned in this comment came from the fact that I was wrong to check the `mmap` result like this (`ptr == NULL`). What I should have done was `ptr == MAP_FAILED`.

The root cause of the issue is that when the program is used with the purpose of making the system enter an OOM state, it is trying to allocate and demand-page in more memory than is available. If it does this as one big allocation, it will fail since the memory manager will just refuse the allocation with `ENOMEM`. In this case it makes sense to try multiple small allocations since they are much more likely to succeed and this is by no means a performance critical code path.

tests/host_tools/fillmem.c
georgepisaltu: Done.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: Removed it.

tests/integration_tests/functional/test_balloon.py
georgepisaltu: You're right, it didn't. It does now.

tests/host_tools/fillmem.c
acatangiu: Ok, that makes sense! Please update the comment to reflect this as well. 👍

src/vmm/src/lib.rs
```diff
@@ -493,6 +495,98 @@ impl Vmm {
             .set_kvm_memory_regions(&self.guest_memory, enable)
             .map_err(Error::Vm)
     }
+
+    /// Updates configuration for the balloon device as described in `balloon_update`.
```
acatangiu, line 499: doc comment from different `fn`

src/vmm/src/device_manager/persist.rs
```diff
@@ -184,6 +229,31 @@ impl<'a> Persist<'a> for MMIODeviceManager {
...
+        if let Some(balloon_state) = &state.balloon_device {
+            let ctor_args = BalloonConstructorArgs { mem: mem.clone() };
+
+            let device = Arc::new(Mutex::new(
+                Balloon::restore(ctor_args, &balloon_state.device_state).map_err(Error::Balloon)?,
+            ));
+
+            let device_id = balloon_state.device_id.clone();
+            let transport_state = &balloon_state.transport_state;
+            let mmio_slot = &balloon_state.mmio_slot;
```
acatangiu, line 241:
```suggestion
            let mmio_slot = &balloon_state.mmio_slot;
            dev_manager
                .slot_sanity_check(mmio_slot)
                .map_err(Error::DeviceManager)?;
```

Somewhere along the line we should dedup all this code: For every device type code block, after building/restoring the actual device there is the same duplicated code for restoring `MmioTransport` and registering the device to dev-manager and event-manager.

acatangiu (commented):
Love the PR's commits structure 👍

acatangiu (approved):

sandreim (approved):

PR #2136, opened 2020-09-15
Feature/micro http accept encoding

### LINKED ISSUE #2096: Implement support for accept-encoding
## Why is this feature request important?  What are the use cases?  Please describe.
Without this is possible that requests for accept-encoding are wrongfully parsed. 

There should be added a handler for the accept-encoding request otherwise the request will be wrongfully parsed resulting in a undesired behavior.

N\A

## Checks

AlexandruCihodaru:
## Reason for This PR

Fixes #2096 .
Added support for accept encoding. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

## Description of Changes

## License Acceptance

## PR Checklist

### DIFF
--- a/src/micro_http/src/common/headers.rs
+++ b/src/micro_http/src/common/headers.rs
@@ -21,6 +21,8 @@ pub enum Header {
     Server,
     /// Header `Accept`
     Accept,
+    /// Header `Accept-Encoding`
+    AcceptEncoding,
 }
 
 impl Header {
@@ -33,6 +35,7 @@ impl Header {
             Self::TransferEncoding => b"Transfer-Encoding",
             Self::Server => b"Server",
             Self::Accept => b"Accept",
+            Self::AcceptEncoding => b"Accept-Encoding",
         }
     }
 
@@ -52,6 +55,7 @@ impl Header {
                 "transfer-encoding" => Ok(Self::TransferEncoding),
                 "server" => Ok(Self::Server),
                 "accept" => Ok(Self::Accept),
+                "accept-encoding" => Ok(Self::AcceptEncoding),
                 invalid_key => Err(RequestError::HeaderError(HttpHeaderError::UnsupportedName(
                     invalid_key.to_string(),
                 ))),
@@ -199,6 +203,7 @@ impl Headers {
                             )),
                         },
                         Header::Server => Ok(()),
+                        Header::AcceptEncoding => Encoding::try_from(entry[1].trim().as_bytes()),
                     }
                 } else {
                     Err(RequestError::HeaderError(
@@ -283,6 +288,63 @@ impl Headers {
     /// Accept header setter.
     pub fn set_accept(&mut self, media_type: MediaType) {
         self.accept = media_type;
+    }
+}
+
+/// Wrapper over supported AcceptEncoding.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Encoding {}
+
+impl Encoding {
+    /// Parses a byte slice and checks if identity encoding is invalidated. Encoding
+    /// must be ASCII, so also UTF-8 valid.
+    ///
+    /// # Errors
+    /// `InvalidRequest` is returned when the byte stream is empty.
+    ///
+    /// `InvalidValue` is returned when the identity encoding is invalidated.
+    ///
+    /// `InvalidUtf8String` is returned when the byte stream contains invalid characters.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use micro_http::Encoding;
+    ///
+    /// assert!(Encoding::try_from(b"deflate").is_ok());
+    /// assert!(Encoding::try_from(b"identity;q=0").is_err());
+    /// ```
+    pub fn try_from(bytes: &[u8]) -> Result<(), RequestError> {
+        if bytes.is_empty() {
+            return Err(RequestError::InvalidRequest);
+        }
+        match std::str::from_utf8(bytes) {
+            Ok(headers_str) => {
+                let entry = headers_str.split(',').collect::<Vec<&str>>();
+
+                for encoding in entry {
+                    match encoding.trim() {
+                        "identity;q=0" => {
+                            Err(RequestError::HeaderError(HttpHeaderError::InvalidValue(
+                                "Accept-Encoding".to_string(),
+                                encoding.to_string(),
+                            )))
+                        }
+                        "*;q=0" if !headers_str.contains("identity") => {
+                            Err(RequestError::HeaderError(HttpHeaderError::InvalidValue(
+                                "Accept-Encoding".to_string(),
+                                encoding.to_string(),
+                            )))
+                        }
+                        _ => Ok(()),
+                    }?;
+                }
+                Ok(())
+            }
+            Err(utf8_err) => Err(RequestError::HeaderError(
+                HttpHeaderError::InvalidUtf8String(utf8_err),
+            )),
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/micro_http/src/common/headers.rs
```diff
@@ -90,6 +94,8 @@ pub struct Headers {
     /// `Accept` header might be used by HTTP clients to enforce server responses with content
     /// formatted in a specific way.
     accept: MediaType,
+    accept_encoding_flag: bool,
+    accept_encoding: Encoding,
```
georgepisaltu, line 98: Instead of having a `field: T` and a `field_flag: bool`, you can make use of Rust's `Option<T>`, but in this particular case the `accept_encoding` field won't be necessary anyway.

src/micro_http/src/common/headers.rs
```diff
@@ -199,6 +207,24 @@ impl Headers {
                             )),
                         },
                         Header::Server => Ok(()),
+                        Header::AcceptEncoding => {
+                            match Encoding::try_from(entry[1].trim().as_bytes()) {
```
georgepisaltu, line 211: I think the match arms can be better expressed with `map` and `map_err`. This would also reduce the indentation level.

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +312,174 @@ impl Headers {
     }
 }
 
+/// Wrapper over supported AcceptEncoding.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum Encoding {
```
georgepisaltu, line 317: We only need the `Identity` variant of this enum as it is the only one we currently support. In this case we can consider removing the variants altogether and only keeping the methods.

src/micro_http/src/request.rs
```diff
@@ -511,5 +511,15 @@ mod tests {
         assert_eq!(request.headers.expect(), false);
         assert_eq!(request.headers.content_length(), 0);
         assert!(request.body.is_none());
+
+        let request = Request::try_from(b"GET http://localhost/ HTTP/1.0\r\n\
+                                                                                    Accept-Encoding: identity;q=0\r\n\r\n");
```
georgepisaltu, line 516: Consider using `cargo fmt --all` to automatically format your code. I think this indentation would also go away but if not, please bring it back a bit.

georgepisaltu (commented):
In the case of the `Accept-Encoding` header, I don't think we need to add any functionality other than checking the "identity" value is not invalidated by the user. This can be done by simply ensuring that in the value string of the header there is no explicit `identity;q=0` or `*;q=0`, the latter considered only when `identity` is not mentioned in the preceding values.

This would greatly reduce the size and complexity of the parsing function and suits our use case perfectly. It would also eliminate the need for an `accept_encoding` field in our header structure. If the header structure gets too complex, we should consider switching to a `HashMap` structure.

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +312,174 @@ impl Headers {
...
+                                        )),
+                                    ),
+
+                                    Ok(_q_value) => Ok(Self::Identity),
+                                },
+                                _ => Err(RequestError::HeaderError(HttpHeaderError::InvalidValue(
+                                    "Accept-Encoding".to_string(),
+                                    encoding.to_string(),
+                                ))),
+                            }
```
serban300, line 443: This seems very similar to what happens on the `2 =>` branch. Can we deduplicate this code ?

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +312,174 @@ impl Headers {
...
+                        }
+                        not_supported_encoding => Err(RequestError::HeaderError(
```
serban300, line 445: I would rename this to `unsupported_encoding`.

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +312,174 @@ impl Headers {
...
+                        // The client wants encoding to be gzip.
+                        "gzip" => Ok(Self::Gzip),
+                        // The client wants encoding to be compress.
+                        "compress" => Ok(Self::Compress),
+                        // The client wants encoding to be deflate.
+                        "deflate" => Ok(Self::Deflate),
+                        // The client wants encoding to be brotli.
+                        "br" => Ok(Self::Br),
+                        // The client wants no encoding.
+                        "identity" => {
```
serban300, line 377: This branch is quite complex. Maybe it would be worth it separating it in another method.

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +298,72 @@ impl Headers {
...
+                for encoding in entry {
+                    let compression = encoding.split(';').collect::<Vec<&str>>()[0].trim();
+                    if !compression_types.contains(&compression) {
+                        return Err(RequestError::HeaderError(
+                            HttpHeaderError::UnsupportedValue(
+                                "Accept-Encoding".to_string(),
+                                encoding.to_string(),
+                            ),
+                        ));
+                    }
```
georgepisaltu, line 341: Why is this check necessary? I think we can just ignore unsupported encodings (even unknown ones) as long as we can use `identity`, which is checked below.

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +298,72 @@ impl Headers {
...
+                    match encoding.trim() {
+                        "identity;q=0" => {
+                            Err(RequestError::HeaderError(HttpHeaderError::InvalidValue(
+                                "Accept-Encoding".to_string(),
+                                encoding.to_string(),
+                            )))
+                        }
+                        "*;q=0" if !q_value_identity => {
```
georgepisaltu, line 349: I think you can replace `q_value_identity` with the actual expression, as it is only used once.

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +298,72 @@ impl Headers {
...
+                            Err(RequestError::HeaderError(HttpHeaderError::InvalidValue(
+                                "Accept-Encoding".to_string(),
+                                encoding.to_string(),
+                            )))
+                        }
+                        _accepted_encoding => Ok(()),
```
georgepisaltu, line 355: I would drop the `_accepted_encoding` name as it is not only not used, but misleading; it implies that we accept multiple encodings, when we only do `identity`.
```suggestion
                        _ => Ok(()),
```

src/micro_http/src/common/headers.rs
```diff
@@ -199,6 +205,12 @@ impl Headers {
                             )),
                         },
                         Header::Server => Ok(()),
+                        Header::AcceptEncoding => Encoding::try_from(entry[1].trim().as_bytes())
+                            .map(|_| {
+                                self.accept_encoding = true;
+                                Ok(())
+                            })
+                            .map_err(|e| e)?,
```
georgepisaltu, line 213: If you drop the previous `accept_encoding` field, which is unused later anyway, this becomes much simpler.
```suggestion
                        Header::AcceptEncoding => Encoding::try_from(entry[1].trim().as_bytes()),
```

src/micro_http/src/common/headers.rs
```diff
@@ -90,6 +94,7 @@ pub struct Headers {
     /// `Accept` header might be used by HTTP clients to enforce server responses with content
     /// formatted in a specific way.
     accept: MediaType,
+    accept_encoding: bool,
```
georgepisaltu, line 97: I don't think we need this field as it is not used later.

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +298,72 @@ impl Headers {
...
+
+impl Encoding {
+    /// Parses a byte slice into a Encoding structure for a HTTP request. Encoding
+    /// must be ASCII, so also UTF-8 valid.
+    ///
+    /// # Errors
+    /// The function returns `InvalidRequest` when the byte stream is empty, `InvalidValue` when the
+    /// byte stream contains a logical error, `UnsupportedValue` when the byte stream contains a
+    /// value that is not supported yet and `InvalidUtf8String` when the byte stream contains
+    /// invalid characters.
```
georgepisaltu, line 313: It's better that you start the explanation for each error variant with the error variant name: `ErrorOne is returned when scenario one happens`. Each possible error and its explanation should be on their own line. Additionally, you should leave an empty line between each error variant so that when formatted by cargo-doc it is displayed like a list. I know this is not how other comments in the crate are written, but it is a good practice nonetheless.

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +298,72 @@ impl Headers {
     }
 }
 
+/// Wrapper over supported AcceptEncoding.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Encoding {}
+
+impl Encoding {
+    /// Parses a byte slice into a Encoding structure for a HTTP request. Encoding
```
georgepisaltu, line 306:
```suggestion
    /// Parses a byte slice into an Encoding structure for a HTTP request. Encoding
```

src/snapshot/Cargo.toml
```diff
@@ -10,6 +10,6 @@ libc = "0.2"
 versionize = "0.1.2"
 versionize_derive = ">=0.1.1"
 
-[[bench]]
-name = "main"
-harness = false
+#[[bench]]
+#name = "main"
+#harness = false
```
serban300, line 15: Is this intentional ?

src/micro_http/src/common/headers.rs
```diff
@@ -286,6 +298,72 @@ impl Headers {
...
+    /// ```
+    pub fn try_from(bytes: &[u8]) -> Result<(), RequestError> {
+        if bytes.is_empty() {
+            return Err(RequestError::InvalidRequest);
+        }
+        match std::str::from_utf8(bytes) {
+            Ok(headers_str) => {
+                let q_value_identity = headers_str.contains("identity");
+                let entry = headers_str.split(',').collect::<Vec<&str>>();
+                let compression_types = ["br", "compress", "deflate", "gzip", "identity", "*"];
```
serban300, line 331: can we make this a global const ? Also having a set instead or an array could be faster.

src/snapshot/Cargo.toml
AlexandruCihodaru: No, will change it. Thanks.

georgepisaltu (approved):

serban300 (approved):

PR #2121, opened 2020-09-07
tests: pin processes to cpu cores

iulianbarbu:
## Reason for This PR

Performance tests might need core pinning  for spawned processes.

## Description of Changes

Added `psutil` python3 package to a new version of dev container and used it to get/set cpu affinity for processes but retrieve a process threads as well. 

- [ ] This functionality can be added in [`rust-vmm`](https://github.com/rust-vmm).

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria.]`
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [x] All commits in this PR are signed (`git commit -s`).
- [x] The reason for this PR is clearly provided (issue no. or explanation).
- [x] The description of changes is clear and encompassing.
~~- [] Any required documentation changes (code and docs) are included in this PR.~~
- [x] Any newly added `unsafe` code is properly documented.~~
~~- [] Any API changes are reflected in `firecracker/swagger.yaml`.~~
~~- [] Any user-facing changes are mentioned in `CHANGELOG.md`.~~
~~- [] All added/changed functionality is tested.~~

PR #2147, opened 2020-09-25
Add mmap with MAP_SHARED to allowlist. Modify devtool to trigger it.

alindima:
Signed-off-by: alindima <alindima@amazon.com>

## Reason for This PR

With the previous allowlist, Firecracker raised a SIGSYS when running non-jailed.
This was because the `/etc/localtime` was getting memory mapped by the `localtime_r` musl implementation, used in the logger.
It was not caught by our CI because we don't mount the `/etc/localtime` in the jailer.

## Description of Changes

 - Allowed the `mmap` syscall with the `MAP_SHARED` flag.
 - Added a new docker container tag that installs the `tzdata` package - `fcuvm/dev:feature_seccomp_audit_v1` - in order to have access to the `/etc/localtime` file in the container.
 - By creating a new docker image, the python style checker complained about some uses of `super()` -> also fixed them
 - Modified the test framework to mount `/etc/localtime` in the jailer root so that this error will not reappear in the future.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/default_syscalls/filters.rs
+++ b/src/vmm/src/default_syscalls/filters.rs
@@ -55,7 +55,7 @@ pub fn default_filter() -> Result<SeccompFilter, Error> {
                     Cond::new(2, ArgLen::DWORD, Eq, super::FCNTL_FD_CLOEXEC)?,
                 ],],
             ),
-            // Used for drive patching & rescanning
+            // Used for drive patching & rescanning, for reading the local timezone
             allow_syscall(libc::SYS_fstat),
             // Used for snapshotting
             #[cfg(target_arch = "x86_64")]
@@ -96,6 +96,16 @@ pub fn default_filter() -> Result<SeccompFilter, Error> {
             allow_syscall(libc::SYS_mremap),
             // Used for freeing memory
             allow_syscall(libc::SYS_munmap),
+            // Used for reading the timezone in LocalTime::now()
+            allow_syscall_if(
+                libc::SYS_mmap,
+                or![and![Cond::new(
+                    3,
+                    ArgLen::DWORD,
+                    Eq,
+                    libc::MAP_SHARED as u64
+                )?],],
+            ),
             #[cfg(target_arch = "x86_64")]
             allow_syscall(libc::SYS_open),
             #[cfg(target_arch = "aarch64")]
--- a/tests/framework/jailer.py
+++ b/tests/framework/jailer.py
@@ -162,6 +162,25 @@ class JailerContext:
             utils.run_cmd(cmd)
         return jailed_p
 
+    def copy_into_root(self, file_path, create_jail=False):
+        """Copy a file in the jail root, owned by uid:gid.
+
+        Copy a file in the jail root, creating the folder path if
+        not existent, then change their owner to uid:gid.
+        """
+        global_path = os.path.join(
+            self.chroot_path(), file_path.strip(os.path.sep))
+        if create_jail:
+            os.makedirs(self.chroot_path(), exist_ok=True)
+
+        os.makedirs(os.path.dirname(global_path), exist_ok=True)
+
+        shutil.copy(file_path, global_path)
+
+        cmd = 'chown {}:{} {}'.format(
+            self.uid, self.gid, global_path)
+        utils.run_cmd(cmd)
+
     def netns_file_path(self):
         """Get the host netns file path for a jailer context.
 
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -82,6 +82,10 @@ class Microvm:
         )
         self.jailer_clone_pid = None
 
+        # Copy the /etc/localtime file in the jailer root
+        self.jailer.copy_into_root(
+            "/etc/localtime", create_jail=True)
+
         # Now deal with the things specific to the api session used to
         # communicate with this machine.
         self._api_session = None
--- a/tools/devtool
+++ b/tools/devtool
@@ -73,7 +73,7 @@
 # Development container image (name:tag)
 # This should be updated whenever we upgrade the development container.
 # (Yet another step on our way to reproducible builds.)
-DEVCTR_IMAGE="fcuvm/dev:v17"
+DEVCTR_IMAGE="fcuvm/dev:feature_seccomp_audit_v1"
 
 # Naming things is hard
 MY_NAME="Firecracker $(basename "$0")"
--- a/tests/framework/decorators.py
+++ b/tests/framework/decorators.py
@@ -14,7 +14,7 @@ def timed_request(method):
 
         def __init__(self, duration, method, resource, payload):
             """Compose the error message from the API call components."""
-            super(ApiTimeoutException, self).__init__(
+            super().__init__(
                 'API call exceeded maximum duration: {:.2f} ms.\n'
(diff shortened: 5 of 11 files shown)


tools/devtool
```diff
@@ -369,6 +369,7 @@ run_devctr() {
     docker run "${docker_args[@]}" \
         --rm \
         --volume /dev:/dev \
+        -v /etc/localtime:/etc/localtime \
```
gc-plp, line 372: I think that if we install `tzdata` in the contaiener we won't need to mount the `localtime` file.

tests/framework/microvm.py
```diff
@@ -82,6 +82,10 @@ def __init__(
         )
         self.jailer_clone_pid = None
 
+        # Copy the /etc/localtime file in the jailer root
+        self.jailer.copy_into_root(
```
gc-plp, line 86: If we install `tzdata` in the container we may not need to add an extra function, thus simplifying the PR.

alindima: Installing `tzdata` would just be another way of mounting `/etc/localtime`. We still need to copy this file into the jailer, so we can't get rid of this function

tools/devtool
alindima: Indeed. However, why do you think installing the package would be a better option?

gc-plp: The host may not necessarily have `/etc/localtime`.

alindima: makes sense

src/vmm/src/default_syscalls/filters.rs
```diff
@@ -96,6 +96,16 @@ pub fn default_filter() -> Result<SeccompFilter, Error> {
             allow_syscall(libc::SYS_mremap),
             // Used for freeing memory
             allow_syscall(libc::SYS_munmap),
+            // Used for reading the timezone in LocalTime::now()
+            allow_syscall_if(
```
aghecenco, line 100: Is `Localtime::now`'s behavior the same on all toolchains?

src/vmm/src/default_syscalls/filters.rs
```diff
@@ -96,6 +96,16 @@ pub fn default_filter() -> Result<SeccompFilter, Error> {
...
+                libc::SYS_mmap,
+                or![and![Cond::new(
+                    3,
+                    ArgLen::DWORD,
+                    Eq,
+                    libc::MAP_SHARED as u64
```
aghecenco, line 106: nit: specify in commit message that `mmap` is now allowed only with `MAP_SHARED`

tests/framework/jailer.py
```diff
@@ -162,6 +162,27 @@ def jailed_path(self, file_path, create=False, create_jail=False):
             utils.run_cmd(cmd)
         return jailed_p
 
+    def copy_into_root(self, file_path, create_jail=False):
+        """Copy a file in the jail root, owned by uid:gid.
+
+        Copies a file in the jail root, creating the folder path if
+        not existent and changes their owner to uid:gid
```
aghecenco, line 169: nit: use active voice, same as above.
```suggestion
        Copy a file in the jail root, creating the folder path if
        not existent, then change their owner to uid:gid.
```

tests/framework/jailer.py
```diff
@@ -162,6 +162,27 @@ def jailed_path(self, file_path, create=False, create_jail=False):
...
+        global_path = os.path.join(
+            self.chroot_path(), file_path.strip(os.path.sep))
+        if create_jail:
+            os.makedirs(self.chroot_path(), exist_ok=True)
+
+        os.makedirs(os.path.dirname(global_path), exist_ok=True)
+
+        cmd = 'cp {} {}'.format(
+            file_path, global_path)
+        utils.run_cmd(cmd)
```
aghecenco, line 180: No need to spawn a new process for a copy. You can use [`shutil.copy`](https://docs.python.org/3/library/shutil.html).

tests/framework/jailer.py
```diff
@@ -162,6 +162,27 @@ def jailed_path(self, file_path, create=False, create_jail=False):
...
+
+        cmd = 'chown -R {}:{} {}'.format(
```
aghecenco, line 182: Depending on when this is called, it could inadvertently change the owner for files in the jail that *don't* require an ownership change. I'd only `chown` the file you put in the jail.

tests/framework/microvm.py
```diff
@@ -82,6 +82,10 @@ def __init__(
...
+            "/etc/localtime", create_jail=True)
```
aghecenco, line 87: What if there's no `/etc/localtime` on the host?

src/vmm/src/default_syscalls/filters.rs
aghecenco: What about `aarch64`?

alindima: yes. the `mmap` is triggered for both `x86_64` and `aarch64`

tests/framework/microvm.py
alindima: this will be handled by installing `tzdata` in the dev container

tools/devtool
alindima: done

src/vmm/src/default_syscalls/filters.rs
alindima: done

tests/framework/jailer.py
alindima: done

tests/framework/jailer.py
alindima: done

tests/framework/microvm.py
alindima: handled

tests/framework/jailer.py
alindima: I agree. Changed this, please take a look

dianpopa:
@alindima  I am trying to understand what was the issue here
1. What functionality from the logger triggered the `localtime_r`?
2. If the /`etc/localtime` was not mounted in the chroot, how would the logger achieve the functionality from 1.?

Thanks,
Diana

alindima:
Hello @dianpopa,

1. The `Localtime::now()` triggered the `localtime_r`, [at this line](https://github.com/alindima/firecracker/blob/feature/seccomp_audit/src/logger/src/logger.rs#L410).

2. When the `/etc/localtime` was not present in the chroot-ed folder, the musl implementation would pick the UTC timezone as a default.

dianpopa (commented):
Functional changes look good, some style comments only:
1. The commit message does not follow 50/72
2. I would have split the commit into 2-3 commits to be easier to review and for a more readable history.

alindima:
@dianpopa thank you for the suggestions, please take another look

dianpopa (approved):

gc-plp:
[This](https://github.com/firecracker-microvm/firecracker/pull/2147/commits/7c21495b564fc0785b525b1873d3cf7e447225d3) commit seems to have been already been pushed as part of #2121. 

Could you could sync with `master` first?

alindima:
As @gc-plp pointed out, I need to do a rebase from master, prior to merging this PR. I will close this PR and will create a duplicate once the rebase is complete

alindima:
the new PR is here: https://github.com/firecracker-microvm/firecracker/pull/2160

Issue #1997, opened 2020-07-03
Incremental snapshots: Memory snapshot is inconsistent.


sandreim:
Steps to reproduce (1 snapshot):
- Start fresh microvm
- Pause microvm
- Create diff snapshot
- Kill microvm
- Start fresh Firecracker
- Load previous snapshot
- Resume guest.

Guest crashes on resume:
```
[    1.979938] BUG: stack guard page was hit at ffffc90000dfbff8 (stack is ffffc90000dfc000..ffffc90000dfffff)
[    1.980796] kernel stack overflow (double-fault): 0000 [#1] SMP PTI
[    1.981294] PANIC: double fault, error_code: 0x0
[    1.981665] CPU: 0 PID: 770 Comm: (agetty) Not tainted 4.14.55-84.37.amzn2.x86_64 #1
[    1.982324] task: ffff8801298c0e40 task.stack: ffffc90000dfc000
[    1.982351] RIP: 0010:+0x4/0x30
[    1.982352] RSP: 0018:fffffe0000007000 EFLAGS: 00010093
[    1.982353] RAX: 0000000000000000 RBX: 0000000000000001 RCX: ffffffff81600a27
[    1.982353] RDX: 0000000000000002 RSI: 0000000000000002 RDI: fffffe0000007028
[    1.982354] RBP: fffffe0000007000 R08: 0000000000000000 R09: 0000000000000000
[    1.982359] R10: 0000000000000000 R11: 0000000000000000 R12: 0000000000000000
[    1.982360] R13: 0000000000000000 R14: 0000000000000000 R15: 0000000000000000
[    1.982360] FS:  00007f7b7dbc4e00(0000) GS:ffff88012fc00000(0000) knlGS:0000000000000000
[    1.982361] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[    1.982361] CR2: fffffe0000006ff8 CR3: 0000000129a84006 CR4: 00000000003606b0
[    1.982369] DR0: 0000000000000000 DR1: 0000000000000000 DR2: 0000000000000000
[    1.982369] DR3: 0000000000000000 DR6: 00000000fffe0ff0 DR7: 0000000000000400
[    1.982370] Call Trace:
[...]

PR #2149, opened 2020-09-29
Start dirty page tracking implementation in vm-memory.

sandreim:
## Reason for This PR

Bootstrap implementation for dirty page tracking in vm-memory - #1997

Thanks @alexandruag for providing insight/PoC on the proxy crate approach.

## Description of Changes

Create a "proxy" crate that Firecracker will use as a dependency for all current vm-memory imports. This crate will export local GuestMemoryMmap and GuestRegionMap implementations that will contain dirty page tracking functionality. All other re-exports will be fetched from upstream vm-memory.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/arch/Cargo.toml
+++ b/src/arch/Cargo.toml
@@ -8,8 +8,7 @@ edition = "2018"
 kvm-bindings = { git = "https://github.com/firecracker-microvm/kvm-bindings", tag = "v0.2.0-2", features = ["fam-wrappers"] }
 kvm-ioctls = { git = "https://github.com/firecracker-microvm/kvm-ioctls", tag = "v0.5.0-2" }
 libc = ">=0.2.39"
-vm-memory = { version = ">=0.2.2", features = ["backend-mmap"] }
-
+vm-memory = { path = "../vm-memory" }
 arch_gen = { path = "../arch_gen" }
 utils = { path = "../utils" }
 
--- a/src/vm-memory/Cargo.toml
+++ b/src/vm-memory/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "vm-memory"
+version = "0.1.0"
+authors = ["Amazon Firecracker team <firecracker-devel@amazon.com>"]
+edition = "2018"
+
+[dependencies]
+vm-memory-upstream = { package = "vm-memory", version = ">=0.2.2", features = ["backend-mmap"] }
+
+[dev-dependencies]
+vmm-sys-util = ">= 0.4.0"
--- a/src/vm-memory/src/bitmap.rs
+++ b/src/vm-memory/src/bitmap.rs
@@ -1,0 +1,132 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+//! Atomic bitmap implementation.
+
+// Temporarly disable unused warnings.
+// TODO: remove these once the Bitmap integration is completed.
+#![allow(dead_code)]
+
+use std::sync::atomic::{AtomicU64, Ordering};
+
+/// `Bitmap` implements a simple bit map on the page level with test and set operations. It is
+/// page-size aware, so it converts addresses to page numbers before setting or clearing the bits.
+#[derive(Debug)]
+pub struct Bitmap {
+    map: Vec<AtomicU64>,
+    size: usize,
+    page_size: usize,
+}
+
+impl Bitmap {
+    /// Create a new bitmap of `byte_size`, with one bit per `page_size`.
+    /// In reality this is rounded up, and you get a new vector of the next multiple of 64 bigger
+    /// than `size` for free.
+    pub fn new(byte_size: usize, page_size: usize) -> Self {
+        // Bit size is the number of bits in the bitmap, always at least 1 (to store the state of
+        // the '0' address).
+        let bit_size = std::cmp::max(1, byte_size / page_size);
+        // Create the map of `AtomicU64`, allowing the bit set operations to be done on a non-mut
+        // `Bitmap`, avoiding the need for a Mutex or other serialization.
+        let map_size = ((bit_size - 1) >> 6) + 1;
+        let map: Vec<AtomicU64> = (0..map_size).map(|_| AtomicU64::new(0)).collect();
+
+        Bitmap {
+            map,
+            size: bit_size,
+            page_size,
+        }
+    }
+
+    /// Is bit `n` set? Bits outside the range of the bitmap are always unset.
+    #[inline]
+    pub fn is_bit_set(&self, n: usize) -> bool {
+        if n <= self.size {
+            (self.map[n >> 6].load(Ordering::SeqCst) & (1 << (n & 63))) != 0
+        } else {
+            // Out-of-range bits are always unset.
+            false
+        }
+    }
+
+    /// Is the bit corresponding to address `addr` set?
+    pub fn is_addr_set(&self, addr: usize) -> bool {
+        self.is_bit_set(addr / self.page_size)
+    }
+
+    /// Set a range of bits starting at `start_addr` and continuing for the next `len` bytes.
+    pub fn set_addr_range(&self, start_addr: usize, len: usize) {
+        let first_bit = start_addr / self.page_size;
+        let page_count = (len + self.page_size - 1) / self.page_size;
+        for n in first_bit..(first_bit + page_count) {
+            if n > self.size {
+                // Attempts to set bits beyond the end of the bitmap are simply ignored.
+                break;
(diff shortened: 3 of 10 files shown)


src/arch/Cargo.toml
```diff
@@ -8,8 +8,7 @@ edition = "2018"
 kvm-bindings = { git = "https://github.com/firecracker-microvm/kvm-bindings", tag = "v0.2.0-2", features = ["fam-wrappers"] }
 kvm-ioctls = { git = "https://github.com/firecracker-microvm/kvm-ioctls", tag = "v0.5.0-2" }
 libc = ">=0.2.39"
-vm-memory = { version = ">=0.2.2", features = ["backend-mmap"] }
-
+vm-memory-proxy = { path = "../vm-memory-proxy" }
```
serban300, line 11: Can we keep the old name (`vm-memory`) ? Chaning it has 2 disadvantages:
1. It's very verbose
2. We'll have to undo this change when we switch back to the `vm-memory` crate.

sandreim: I agree on both points. My idea here was to make the switch be explicit and much clear.

aghecenco:
I may have gotten it backwards - I was thinking that Firecracker will *continue* consuming `rust-vmm/vm-memory`, but instead of using `rust-vmm/vm-memory::GuestMemoryMmap` all over the place, Firecracker would add its own `struct GuestMemoryMmapWithTracking` (or something) and `impl rust-vmm/vm-memory::GuestMemory for firecracker::GuestMemoryMmapWithTracking` (disregard rustlang blasphemy). That way we don't have to kill the dependency. Did we move away from that? What's the advantage of removing the dependency, other than renaming `GuestMemoryMmap` everywhere?

sandreim:
> I may have gotten it backwards - I was thinking that Firecracker will _continue_ consuming `rust-vmm/vm-memory`, but instead of using `rust-vmm/vm-memory::GuestMemoryMmap` all over the place, Firecracker would add its own `struct GuestMemoryMmapWithTracking` (or something) and `impl rust-vmm/vm-memory::GuestMemory for firecracker::GuestMemoryMmapWithTracking` (disregard rustlang blasphemy). That way we don't have to kill the dependency. Did we move away from that? What's the advantage of removing the dependency, other than renaming `GuestMemoryMmap` everywhere?

The dependency to vm-memory from upstream is kept in the local proxy crate `vm-memory-proxy` and we override only what is needed with a local implementation. The rest of the FC code depends on the proxy crate which re-exports the upstream stuff.

src/arch/Cargo.toml
sandreim: I've updated the PR and now the proxy crate has the same name: `vm-memory`

acatangiu: Only 11 files changed now as opposed to the previous version with 50 files changed 👍

aghecenco:
> we override only what is needed with a local implementation

Why not create a new implementor for `GuestMemory`?

raduweiss:
> > we override only what is needed with a local implementation
> 
> Why not create a new implementor for `GuestMemory`?

FWIW I also understood the coupling in the way @aghecenco described it above.

sandreim:
> > > we override only what is needed with a local implementation
> > 
> > 
> > Why not create a new implementor for `GuestMemory`?
> 
> FWIW I also understood the coupling in the way @aghecenco described it above.

The goal here is to do as few changes as possible to the interfaces we currently use in Firecracker with the hope that tracking will be available in upstream. The Firecracker code uses both the GuestMemory  interface and also the GuestMemoryMmap directly. If we implement a new backend like GuestMemoryMmapTracked we would need to do more changes in Firecracker.

The approach in this PR is less invasive and completely transparent to Firecracker as we re-export all the stuff from upstream and we override only what's needed to support our usecase.

src/vm-memory/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "vm-memory"
+version = "0.1.0"
+authors = ["Amazon Firecracker team <firecracker-devel@amazon.com>"]
+edition = "2018"
+
+[dependencies]
+vm-memory = { version = ">=0.2.2", features = ["backend-mmap"] }
```
aghecenco, line 8: This
```suggestion
vm-memory-upstream = { package = "vm-memory", version = ">=0.2.2", features = ["backend-mmap"] }
```
and replacing occurrences of `vm_memory` with `vm_memory_upstream` inside the `firecracker/vm_memory` crate should do the trick for the doctests

sandreim: Thanks, this fixes the doctest issue.

src/vm-memory/src/bitmap.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Get the length of the bitmap in bits (i.e. in how many pages it can represent).
+    pub fn len(&self) -> usize {
+        self.size
+    }
+
+    /// Is the bitmap empty (i.e. has zero size)? This is always false, because we explicitly
+    /// round up the size when creating the bitmap.
+    pub fn is_empty(&self) -> bool {
+        false
+    }
```
luminitavoicu, line 79: I don't think this function is needed, since we know the bitmap is never empty. WDYT?

sandreim: Makes sense, I'll remove it.

luminitavoicu (approved):

src/vm-memory/src/mmap.rs
```diff
@@ -0,0 +1,1184 @@
...
+        .unwrap();
+        assert!(gm.find_region(start_addr).is_some());
+        let region = gm.find_region(start_addr).unwrap();
+        assert!(region.file_offset().is_some());
+    }
+
+    // Windows needs a dedicated test where it will retrieve the allocation
+    // granularity to determine a proper offset (other than 0) that can be
+    // used for the backing file. Refer to Microsoft docs here:
+    // https://docs.microsoft.com/en-us/windows/desktop/api/memoryapi/nf-memoryapi-mapviewoffile
```
luminitavoicu, line 1098: I don't think we should we keep this indication.

sandreim: You are right.

src/vm-memory/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the THIRD-PARTY file.
```
acatangiu, line 6: nit: is this still the case?

src/vm-memory/src/mmap.rs
```diff
@@ -0,0 +1,1191 @@
+// Copyright (C) 2019 Alibaba Cloud Computing. All rights reserved.
+//
+// Portions Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
acatangiu, line 3: nit: For this local file where we add dirty tracking support, these two might be better reversed.

src/vm-memory/src/bitmap.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+impl Bitmap {
+    /// Create a new bitmap of `byte_size`, with one bit per `page_size`.
+    /// In reality this is rounded up, and you get a new vector of the next multiple of 64 bigger
+    /// than `size` for free.
+    pub fn new(byte_size: usize, page_size: usize) -> Self {
+        // Bit size is the number of bits in the bitmap, always at least 1 (to store the state of
+        // the '0' address).
+        let bit_size = std::cmp::max(1, byte_size / page_size);
```
acatangiu, line 28: nit:
```suggestion
        let num_pages = std::cmp::max(1, byte_size / page_size);
```
for clarity.

src/vm-memory/src/mmap.rs
```diff
@@ -206,14 +200,27 @@ impl GuestMemoryRegion for GuestRegionMmap {
...
+        // This is safe because we mapped the area at addr ourselves, so this slice will not
+        // overflow. However, it is possible to alias.
+        Some(std::slice::from_raw_parts(
+            self.mapping.as_ptr(),
+            self.mapping.size(),
+        ))
     }
 
+    // TODO: This implementation is temporary.
+    // We need to return None here once we refactor vsock.
```
acatangiu, line 215: Let's add an issue to track this.

acatangiu (approved):

src/vm-memory/src/mmap.rs
sandreim: Created #2153

PR #2157, opened 2020-10-05
Added dirty page tracking functionality to local vm-memory

luminitavoicu:
## Reason for This PR

Bootstrap implementation for dirty page tracking in vm-memory - #1997

## Description of Changes

- Added a dedicated dirty page bitmap for GuestRegionMap.
- Updated local vm-memory's API with capabilities to enable dirty page tracking.
- Updated dirty page bitmap every time the memory is altered.
- Added unit tests for both GuestRegionMap and GuestMemoryMmap.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vm-memory/src/mmap.rs
+++ b/src/vm-memory/src/mmap.rs
@@ -22,7 +22,9 @@ use vm_memory_upstream::address::Address;
 use vm_memory_upstream::guest_memory::{
     self, FileOffset, GuestAddress, GuestMemory, GuestMemoryRegion, GuestUsize, MemoryRegionAddress,
 };
-use vm_memory_upstream::volatile_memory::{VolatileMemory, VolatileSlice};
+use vm_memory_upstream::volatile_memory::{
+    Error as VolatileMemoryError, VolatileMemory, VolatileSlice,
+};
 use vm_memory_upstream::{ByteValued, Bytes};
 
 use vmm_sys_util::errno;
@@ -107,15 +109,16 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
     type E = guest_memory::Error;
 
     fn write(&self, buf: &[u8], addr: MemoryRegionAddress) -> guest_memory::Result<usize> {
-        // TODO: here be write!
         let maddr = addr.raw_value() as usize;
-        self.local_volatile_slice()
+        let bytes = self
+            .local_volatile_slice()
             .write(buf, maddr)
-            .map_err(Into::into)
+            .map_err(Into::<guest_memory::Error>::into)?;
+        self.mark_dirty_pages(maddr, bytes);
+        Ok(bytes)
     }
 
     fn read(&self, buf: &mut [u8], addr: MemoryRegionAddress) -> guest_memory::Result<usize> {
-        // TODO: here be read!
         let maddr = addr.raw_value() as usize;
         self.local_volatile_slice()
             .read(buf, maddr)
@@ -123,15 +126,22 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
     }
 
     fn write_slice(&self, buf: &[u8], addr: MemoryRegionAddress) -> guest_memory::Result<()> {
-        // TODO: here be write!
         let maddr = addr.raw_value() as usize;
-        self.local_volatile_slice()
-            .write_slice(buf, maddr)
-            .map_err(Into::into)
+        match self.local_volatile_slice().write_slice(buf, maddr) {
+            Ok(()) => {
+                self.mark_dirty_pages(maddr, buf.len());
+                Ok(())
+            }
+            Err(e) => {
+                if let VolatileMemoryError::PartialBuffer { completed, .. } = e {
+                    self.mark_dirty_pages(maddr, completed);
+                }
+                Err(e.into())
+            }
+        }
     }
 
     fn read_slice(&self, buf: &mut [u8], addr: MemoryRegionAddress) -> guest_memory::Result<()> {
-        // TODO: here be read!
         let maddr = addr.raw_value() as usize;
         self.local_volatile_slice()
             .read_slice(buf, maddr)
@@ -164,11 +174,13 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
     where
         F: Read,
     {
-        // TODO: here be write!
         let maddr = addr.raw_value() as usize;
-        self.local_volatile_slice()
+        let bytes = self
+            .local_volatile_slice()
             .read_from::<F>(maddr, src, count)
-            .map_err(Into::into)
+            .map_err(Into::<guest_memory::Error>::into)?;
+        self.mark_dirty_pages(maddr, bytes);
+        Ok(bytes)
     }
 
     fn read_exact_from<F>(
@@ -180,11 +192,12 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
     where
         F: Read,
     {
-        // TODO: here be write!
         let maddr = addr.raw_value() as usize;
         self.local_volatile_slice()
             .read_exact_from::<F>(maddr, src, count)
-            .map_err(Into::into)
+            .map_err(Into::<guest_memory::Error>::into)?;
+        self.mark_dirty_pages(maddr, count);
+        Ok(())
     }
 
     fn write_to<F>(
@@ -196,7 +209,6 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
     where
         F: Write,
     {
-        // TODO: here be read!
         let maddr = addr.raw_value() as usize;
         self.local_volatile_slice()
             .write_to::<F>(maddr, dst, count)
@@ -212,7 +224,6 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
     where
         F: Write,
     {
-        // TODO: here be read!
         let maddr = addr.raw_value() as usize;
         self.local_volatile_slice()
             .write_all_to::<F>(maddr, dst, count)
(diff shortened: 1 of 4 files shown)


src/vm-memory/src/mmap.rs
```diff
@@ -269,14 +339,32 @@ impl GuestMemoryMmap {
...
+
+    /// Creates a container, allocates anonymous memory for guest memory regions and enables dirty
+    /// page tracking.
+    ///
+    /// Valid memory regions are specified as a slice of (Address, Size) tuples sorted by Address.
+    pub fn from_ranges_with_tracking(
+        ranges: &[(GuestAddress, usize)],
+    ) -> result::Result<Self, Error> {
+        Self::from_ranges_with_files(ranges.iter().map(|r| (r.0, r.1, None)), true)
     }
```
serban300, line 353: Is this needed ? Can't we just add `track_dirty_pages` as a parameter to `from_ranges` ?

src/vm-memory/src/mmap.rs
```diff
@@ -51,9 +77,39 @@ impl GuestRegionMmap {
...
+        guest_base: GuestAddress,
+        bitmap: Bitmap,
+    ) -> result::Result<Self, Error> {
+        if guest_base.0.checked_add(mapping.len() as u64).is_none() {
+            return Err(Error::InvalidGuestRegion);
+        }
+        Ok(GuestRegionMmap {
+            mapping,
+            guest_base,
+            dirty_bitmap: Some(bitmap),
```
serban300, line 97: Shouldn't we build the bitmap here ? Since we know all the details about the region. What's the advantage of accepting the bitmap as a parameter ?

src/vm-memory/src/mmap.rs
```diff
@@ -51,9 +77,39 @@ impl GuestRegionMmap {
         Ok(GuestRegionMmap {
             mapping,
             guest_base,
+            dirty_bitmap: None,
+        })
+    }
+
+    /// Create a new memory-mapped memory region for the guest's physical memory.
+    /// This region will own a dedicated bitmap to handle dirty page tracking.
+    pub fn new_with_bitmap(
```
serban300, line 86: Do we need a new function ? Wouldn't it be better to supply a flag to `new()` ?

src/vm-memory/src/mmap.rs
```diff
@@ -380,6 +478,21 @@ impl GuestMemoryMmap {
 
         Err(Error::InvalidGuestRegion)
     }
+
+    /// Mutably apply the specified function on each memory region, including the dirty bitmap.
+    pub fn with_regions_bitmaps_mut<F>(&self, mut cb: F) -> result::Result<(), GuestRegionMmapError>
```
serban300, line 483: Is this needed ? Can't we just call `with_regions_mut` and call `region.bitmap()` for each region ?

src/vm-memory/src/mmap.rs
luminitavoicu: From_ranges function has a lot of usages inside Firecracker, so adding an additional dedicated function to handle dirty page tracking helped minimize the impact on the current code.

src/vm-memory/src/mmap.rs
```diff
@@ -74,31 +130,35 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
...
-        self.local_volatile_slice()
-            .write(buf, maddr)
-            .map_err(Into::into)
+        match self.local_volatile_slice().write(buf, maddr) {
+            Ok(bytes) => {
+                self.mark_dirty_pages(maddr, bytes);
+                Ok(bytes)
+            }
+            Err(e) => Err(e.into()),
+        }
```
sandreim, line 140:
```suggestion
        let bytes = self.local_volatile_slice().write(buf, maddr)?;
        self.mark_dirty_pages(maddr, bytes);
        Ok(bytes)
```

src/vm-memory/src/mmap.rs
```diff
@@ -450,6 +563,62 @@ mod tests {
...
+
+        // check read_from
+        let mut file = File::open(Path::new("/dev/zero")).unwrap();
+        assert!(!mmap.dirty_bitmap().unwrap().is_addr_set(356));
+        assert!(mmap
+            .read_from(MemoryRegionAddress(356), &mut file, mem::size_of::<u32>())
+            .is_ok());
+        assert!(mmap.dirty_bitmap().unwrap().is_addr_set(356));
+
+        // check read_exact_from
```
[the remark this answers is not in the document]
luminitavoicu: You are right. I will add more test cases.

src/vm-memory/src/mmap.rs
```diff
@@ -12,6 +12,11 @@
 //!
 //! This implementation is mmap-ing the memory of the guest into the current process.
 
+// Temporarly disable unused warnings.
+// TODO: remove these once dirty tracking support is completed.
+#![allow(dead_code)]
```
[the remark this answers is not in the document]
luminitavoicu: I used this for the `with_regions_bitmaps_mut` function, but since we are getting rid of that, these lines should indeed be removed.

src/vm-memory/src/mmap.rs
luminitavoicu: You are right, we can do that.

src/vm-memory/src/mmap.rs
luminitavoicu: I will add a dedicated function to enable dirty page tracking instead and use the original constructor for both cases.

src/vm-memory/src/mmap.rs
```diff
@@ -51,9 +57,34 @@ impl GuestRegionMmap {
...
+    /// Provide the region with a dedicated bitmap to handle dirty page tracking.
+    pub fn enable_dirty_page_tracking(&mut self) {
+        let page_size = match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
+            -1 => panic!(
+                "Failed to enable dirty page tracking: {}",
+                errno::Error::last()
+            ),
+            ps => ps as usize,
+        };
+        self.dirty_bitmap = Some(Bitmap::new(self.len() as usize, page_size));
```
serban300, line 73: I would do this only if `dirty_bitmap` is None.

src/vm-memory/src/mmap.rs
serban300: Can we also add a method 'enable_dirty_tracking()' here ? Instead of having a different constructor ?

src/vm-memory/src/mmap.rs
```diff
@@ -347,7 +404,7 @@ impl GuestMemoryMmap {
     /// Insert a region into the `GuestMemoryMmap` object and return a new `GuestMemoryMmap`.
     ///
     /// # Arguments
-    /// * `region`: the memory region to insert into the guest memory object.
+    /// * `region` - the memory region to insert into the guest memory object.
     pub fn insert_region(
```
serban300, line 408: AFAIK we don't use this function. But we should handle the case where the new region doesn't have dirty_tracking enabled while the others have. Or the opposite.

src/vm-memory/src/mmap.rs
```diff
@@ -74,31 +105,32 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
...
 
     fn write(&self, buf: &[u8], addr: MemoryRegionAddress) -> guest_memory::Result<usize> {
-        // TODO: here be write!
         let maddr = addr.raw_value() as usize;
-        self.local_volatile_slice()
+        let bytes = self
+            .local_volatile_slice()
             .write(buf, maddr)
-            .map_err(Into::into)
+            .map_err(Into::<guest_memory::Error>::into)?;
```
serban300, line 112: Just mentioning this here. There are cases where we get an error, but actually we managed to do a partial write. If this happens I'm not sure whether we should consider the page dirtied or not.

src/vm-memory/src/mmap.rs
```diff
@@ -347,13 +415,20 @@ impl GuestMemoryMmap {
...
         &self,
-        region: Arc<GuestRegionMmap>,
+        mut region: GuestRegionMmap,
     ) -> result::Result<GuestMemoryMmap, Error> {
+        let dirty_page_tracking = self.is_dirty_tracking_enabled();
+        if !dirty_page_tracking && region.dirty_bitmap().is_some() {
+            region.dirty_bitmap = None;
+        } else if dirty_page_tracking {
+            region.enable_dirty_page_tracking();
+        }
```
serban300, line 428: I think  `&& region.dirty_bitmap().is_some()` is not needed.

src/vm-memory/src/mmap.rs
```diff
@@ -380,6 +455,11 @@ impl GuestMemoryMmap {
 
         Err(Error::InvalidGuestRegion)
     }
+
+    /// Return true if dirty page tracking is enabled for `GuestMemoryMmap`, and else otherwise.
+    pub fn is_dirty_tracking_enabled(&self) -> bool {
+        self.regions[0].dirty_bitmap().is_some()
```
serban300, line 461: In theory we might have a GuestMemory with no regions. Anyway I would use `self.regions.iter().all()` here.

src/vm-memory/src/mmap.rs
```diff
@@ -74,31 +109,39 @@ impl Bytes<MemoryRegionAddress> for GuestRegionMmap {
...
+        match self.local_volatile_slice().write_slice(buf, maddr) {
+            Ok(()) => {
+                self.mark_dirty_pages(maddr, buf.len());
+                Ok(())
+            }
+            Err(e) => {
+                if let VolatileMemoryError::PartialBuffer { completed, .. } = e {
+                    self.mark_dirty_pages(maddr, completed);
+                }
+                Err(e.into())
```
serban300, line 139: We should also do this for `write`, `read_from`, `read_exact_from`.

luminitavoicu: I think `write_slice` is the only one of them  that can return a `PartialBuffer` error.

src/vm-memory/src/mmap.rs
luminitavoicu: Yes, that's better.

serban300 (dismissed):

src/vm-memory/src/mmap.rs
```diff
@@ -51,9 +59,36 @@ impl GuestRegionMmap {
...
+    }
+
+    /// Get a copy of the dirty page bitmap representative for this memory region (if any).
+    pub fn dirty_bitmap(&self) -> Option<Bitmap> {
+        self.dirty_bitmap.clone()
+    }
+
+    /// Mark pages dirty starting from 'start_addr' and continuing for 'len' bytes.
+    pub fn mark_dirty_pages(&self, start_addr: usize, len: usize) {
+        if let Some(bitmap) = self.dirty_bitmap.as_ref() {
```
dianpopa, line 87:
```suggestion
        if let Some(bitmap) = self.dirty_bitmap() {
```

src/vm-memory/src/mmap.rs
```diff
@@ -347,13 +415,20 @@ impl GuestMemoryMmap {
...
     pub fn insert_region(
         &self,
-        region: Arc<GuestRegionMmap>,
+        mut region: GuestRegionMmap,
     ) -> result::Result<GuestMemoryMmap, Error> {
+        let dirty_page_tracking = self.is_dirty_tracking_enabled();
+        if dirty_page_tracking {
+            region.enable_dirty_page_tracking();
+        } else {
+            region.dirty_bitmap = None;
```
dianpopa, line 427: Is this necessary? The `GuestRegionMmap `defaults to `None` for the "dirty_bitmap" field, right?

tests/integration_tests/build/test_coverage.py
```diff
@@ -23,7 +23,7 @@
 # this contains the frequency while on AMD it does not.
 # Checkout the cpuid crate. In the future other
 # differences may appear.
-COVERAGE_DICT = {"Intel": 85.08, "AMD": 84.35}
+COVERAGE_DICT = {"Intel": 85.15, "AMD": 84.41}
```
dianpopa, line 26: 👍

src/vm-memory/src/mmap.rs
```diff
@@ -51,9 +59,36 @@ impl GuestRegionMmap {
...
+        };
+        if self.dirty_bitmap.is_none() {
+            self.dirty_bitmap = Some(Bitmap::new(self.len() as usize, page_size));
+        }
+    }
+
+    /// Get a copy of the dirty page bitmap representative for this memory region (if any).
+    pub fn dirty_bitmap(&self) -> Option<Bitmap> {
+        self.dirty_bitmap.clone()
+    }
```
dianpopa, line 83:
```suggestion
     pub fn dirty_bitmap(&self) -> Option<&Bitmap> {
        self.dirty_bitmap.as_ref()
    }
```

src/vm-memory/src/mmap.rs
luminitavoicu: Yes, indeed. But dirty page tracking can be enabled for the new `GuestRegionMmap` before the insertion. If dirty tracking is not activated for the `GuestMemoryMmap,` it makes no sense to insert there a region which owns a bitmap for dirty pages. WDYT?

src/vm-memory/src/mmap.rs
```diff
@@ -51,9 +59,36 @@ impl GuestRegionMmap {
         Ok(GuestRegionMmap {
             mapping,
             guest_base,
+            dirty_bitmap: None,
         })
     }
 
+    /// Provide the region with a dedicated bitmap to handle dirty page tracking.
+    pub fn enable_dirty_page_tracking(&mut self) {
+        let page_size = match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
```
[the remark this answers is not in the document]
dianpopa: had an offline discussion. Since sysconf can only return error when the parameter is not an accepted one (which will not ever be the case for us), I think it is ok to leave the expect

src/vm-memory/src/mmap.rs
dianpopa: I see, you re right, keep it this way.

dianpopa (approved):

Issue #2154, opened 2020-10-02
Replace Polly with rust-vmm/event-manager


sandreim:

humancalico:
@sandreim Can I work on this?

dianpopa:
Hi @humancalico ,

We were planning on working on this in the near future but if you want to start work on this you can go ahead and I can help you with guidance. I encourage you to join our [slack channel](https://tinyurl.com/firecracker-microvm) and ping me there for anything.

acatangiu:
Fixed by #2604

PR #2208, opened 2020-10-21
RFC: Add support for pausing and resuming emulation

### LINKED ISSUE #2193: Potential race condition during Pause/CreateSnapshot sequence
Looking at the implementation of Firecracker's snapshot functionality, the events happen in the following order:
- Pause Vcpus (VMM thread is still running and can process a Block request - for example a write to disk)
- At this point the Firecracker customers can copy the block device files to a new location; This is where the race condition can happen because the write request is still in transit and VMM emulation is still running
- CreateSnapshot

Even when saving the block device happens after the CreateSnapshot API, there is still the opportunity of a race because the Pause and CreateSnapshot requests can be sent back to back via API connection reuse.

Thanks @andreeaflorescu for finding this.

acatangiu:
## Reason for This PR

Fixes #2193 
Alternate fix: #2213 

## Draft/RFC status

Currently RFC to get early feedback. Still needs unit-tests and integ-tests to promote to full PR.

## Description of Changes

When pausing or resuming the VM, alongside vcpus device emulation is now also paused or resumed.
    
For reduced complexity and runtime overhead, this change only affects virtio devices. Legacy devices are controlled by vcpus (which get paused) and use no resources external to firecracker so their state remains consistent even if they're not explicitly paused.

### Implementation details:

    We want to be able to pause and resume event handling.
    Usually this boils down to unregistering from event handling on
    pause, and re-registering on resume.
    
    At the very least we need an 'update()' function in the Subscriber
    trait to allow on-demand updates to EventManager for said subscriber.
    
    I've chosen instead to add pause and resume specific functions with
    default implementations to minimize boiler-plate code across all
    Subscribers that don't need custom implementations.
    To keep an unified model for both fresh booting a VM and restoring
    one from snapshot, we follow the same pattern with devices as with
    vcpus: devices start off in the paused state and they are resumed
    alongside vcpus.

    In devices, override default implementation of Subscriber functions that
    control when events are registered and unregistered.
    
    The resulting behaviour is:
     * (pause):  on paused->paused nothing happens,
     * (pause):  on running->paused transition device events are
                 unregistered from EventManager, effectively
                 removing the device from the emulation thread,
     * (resume): on running->running nothing happens,
     * (resume): on paused->running transition device events are
                 re-registered with EventManager.
    
    The MMIODeviceManager was enhanced with a list of subscribers to be
    able to directly access the Subscriber trait functions on the devices
    without having to downcast BusDevices all the way down to their
    concrete types then upcast back to Arc<Mutex<dyn Subscriber>>.
    
    Device events registration is moved from device attachment to Vmm,
    to device resume path since `resume_emulation()` is called on both
    boot path and snapshot load path.

## License Acceptance

## PR Checklist

- [ ] ~Any user-facing changes are mentioned in `CHANGELOG.md`.~

### DIFF
--- a/src/polly/src/event_manager.rs
+++ b/src/polly/src/event_manager.rs
@@ -61,6 +61,24 @@ pub trait Subscriber {
 
     /// Returns a list of `EpollEvent` that this subscriber is interested in.
     fn interest_list(&self) -> Vec<EpollEvent>;
+
+    /// Callback for the `Subscriber` to handle stopping/pausing its event handling.
+    ///
+    /// The `Subscriber` is also responsible for unregistering its events from `EventManager`.
+    fn stop(&mut self, _: &mut EventManager) -> std::result::Result<(), String> {
+        Ok(())
+    }
+
+    /// Callback for the Subscriber to handle starting/resuming its event handling.
+    ///
+    /// The `Subscriber` is also responsible for registering its events from `EventManager`.
+    fn start(
+        &mut self,
+        _: Arc<Mutex<dyn Subscriber>>,
+        _: &mut EventManager,
+    ) -> std::result::Result<(), String> {
+        Ok(())
+    }
 }
 
 /// Manages I/O notifications using epoll mechanism.
--- a/src/devices/src/virtio/balloon/device.rs
+++ b/src/devices/src/virtio/balloon/device.rs
@@ -117,6 +117,7 @@ pub struct Balloon {
     pub(crate) device_state: DeviceState,
 
     // Implementation specific fields.
+    pub(crate) events_registered: bool,
     pub(crate) restored: bool,
     pub(crate) stats_polling_interval_s: u16,
     pub(crate) stats_timer: TimerFd,
@@ -172,6 +173,7 @@ impl Balloon {
             queues,
             device_state: DeviceState::Inactive,
             activate_evt: EventFd::new(libc::EFD_NONBLOCK).map_err(BalloonError::EventFd)?,
+            events_registered: false,
             restored,
             stats_polling_interval_s,
             stats_timer,
--- a/src/devices/src/virtio/balloon/event_handler.rs
+++ b/src/devices/src/virtio/balloon/event_handler.rs
@@ -2,15 +2,16 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::os::unix::io::AsRawFd;
+use std::result::Result;
+use std::sync::{Arc, Mutex};
 
 use logger::{debug, error, warn};
 use polly::event_manager::{EventManager, Subscriber};
 use utils::epoll::{EpollEvent, EventSet};
 
 use crate::report_balloon_event_fail;
-use crate::virtio::{
-    balloon::device::Balloon, VirtioDevice, DEFLATE_INDEX, INFLATE_INDEX, STATS_INDEX,
-};
+use crate::virtio::balloon::device::Balloon;
+use crate::virtio::{VirtioDevice, DEFLATE_INDEX, INFLATE_INDEX, STATS_INDEX};
 
 impl Balloon {
     fn process_activate_event(&self, event_manager: &mut EventManager) {
@@ -125,6 +126,36 @@ impl Subscriber for Balloon {
                 self.activate_evt.as_raw_fd() as u64,
             )]
         }
+    }
+
+    fn stop(&mut self, evmgr: &mut EventManager) -> Result<(), String> {
+        if self.events_registered {
+            for event in self.interest_list() {
+                evmgr
+                    .unregister(event.data() as i32)
+                    .map_err(|e| format!("Failed to unregister events: {:?}", e))?;
+            }
+            self.events_registered = false;
+        }
+        Ok(())
+    }
+
+    fn start(
+        &mut self,
+        self_subscriber: Arc<Mutex<dyn Subscriber>>,
+        evmgr: &mut EventManager,
+    ) -> Result<(), String> {
+        if !self.events_registered {
+            for event in self.interest_list() {
+                evmgr
+                    .register(event.data() as i32, event, self_subscriber.clone())
+                    .map_err(|e| format!("Failed to re-register events: {:?}", e))?;
+                // Kick the device to make up for lost, in-flight events.
+                self.process(&event, evmgr);
+            }
+            self.events_registered = true;
+        }
+        Ok(())
     }
 }
 
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -145,6 +145,7 @@ pub struct Block {
     pub(crate) partuuid: Option<String>,
     pub(crate) root_device: bool,
     pub(crate) rate_limiter: RateLimiter,
+    pub(crate) events_registered: bool,
 }
 
 impl Block {
@@ -186,6 +187,7 @@ impl Block {
             queues,
             device_state: DeviceState::Inactive,
             activate_evt: EventFd::new(libc::EFD_NONBLOCK)?,
+            events_registered: false,
         })
     }
 
--- a/src/devices/src/virtio/block/event_handler.rs
+++ b/src/devices/src/virtio/block/event_handler.rs
@@ -1,13 +1,15 @@
 // Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 use std::os::unix::io::AsRawFd;
+use std::result::Result;
+use std::sync::{Arc, Mutex};
 
 use logger::{debug, error, warn};
 use polly::event_manager::{EventManager, Subscriber};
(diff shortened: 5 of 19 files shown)


src/polly/src/event_manager.rs
```diff
@@ -61,6 +61,68 @@ pub trait Subscriber {
 
     /// Returns a list of `EpollEvent` that this subscriber is interested in.
     fn interest_list(&self) -> Vec<EpollEvent>;
+
+    /// Function used by the default implementation of `pause()` to decide if it should
+    /// unregister `Subscriber` events from `EventManager`.
+    ///
+    /// Override this function with custom logic to control the default implementation
+    /// of `pause()`.
+    fn unregister_events_on_pause(&self) -> bool {
```
andreeaflorescu, line 70: Adding the pause/resume specific functions in polly looks incompatible with the initial polly design which was intended as a reusable event management compoennt. Changing the interface of polly at this level would also make it really hard to switch to event_manager (#2154) in the future as it unlikely that such a functionality will be implemented at this level.

Would it be possible to abstract this at a different level? If boiler plate code is the concern that generated these changes to happen at the interface level, it might make sense to explore other technical solutions as well. WDYT?

andreeaflorescu (commented):
I think I might be missing something, but isn't it still possible to have a race with this implementation? It looks like we're just removing the event from the epoll loop, but this change will only be of effect at the next epoll `run` loop.

Beside this, when we pause the device, shouldn't we first check that there are no inflight requests? I would expect us to wait for any unfinished requests, and only then mark the device as `Pause` because otherwise we can still get in inconsistent states.

sandreim:
This looks a bit complicated and I agree with @andreeaflorescu about polly interface changes in the context of switching to rust-vmm/event-manager. I am wondering if we could simplify this. 

Another approach that might be considered "hacky"  is to have a secondary event manager that we would run when the device emulation is stopped. The only thing that needs to be done here is to move the API thread subscriber to this secondary EM on Pause and back on Resume.

src/polly/src/event_manager.rs
acatangiu: I agree with your concerns, but it all boils down with whether we want `polly` or the upstream `EventManager` to be able to offer control to their underlying `Subscribers` over registering and un-registering their (or other's) events in the current `EventManager` loop/`run()`.

### *The Scenario*

Let me try to set a clear scenario and we can definitely explore other technical solutions for handling it:
- there are multiple registered `Subscribers`,
- one of them is the main one, the driver if you will, or in our case the `Vmm`,
- at some point the driver/vmm, wants one of the other `subscribers`, in our case _a device_, to register or unregister their events **based on the device's internal state and/or data**

### Options (identified so far)

To do this I can identify two options:
1. Using the existing `polly/event_manager` interface:
   1. The `vmm` downcasts whatever handle it has for the device (`trait BusDevice` or some `trait Bla` that has the ability to navigate down to the concrete type; `trait Subscriber` for example doesn't currently have that ability) all the way down to the concrete type.
   2. Having the concrete type, it inspects the device's internal state/data, and based on that it decides whether to register/unregister that device to/from `event_manager`.
   3. To be able to call `event_manager.register()` we need to upcast the concrete type to `Arc<Mutex<dyn Subscriber>>` (needed by `register()`).
   4. To do this upcast we need to add *a lot of god ugly boiler-plate* conversions to get from `MutexGuard<ConcreteType>` to `Arc<Mutex<dyn Subscriber>>`.
2. Enhancing `trait Subscriber` interface to allow native/internal handling of the scenario:
   1. `Subscriber` offers an interface for the device to do it itself. We can add better, generic names, but for this example let's stick with `pause/resume`.
   2. Unlike `1. i`, when working through the trait, there is no need to do any downcasting, so `vmm` simply calls `device.pause()`. The device implementation natively runs on the `Self` concrete type.
   3. Unlike `1. iii`, we're already working with `Arc<Mutex<dyn Subscriber>>` to start with, so we can easily use `event_manager.register()` in `device.resume()` with no dynamic up or down casting necessary.
3. (Hacky) Existing `polly` interface: use _fake events_ in every device and manually call `process()`.
   - Works same as **option 2** but instead of adding `pause/resume` we define fake `EpollEvents` that decode as `pause` or `resume` commands.
   - `vmm` calls `device.process()` and pushes one of these fake events.
   - when defining/choosing fake event values, we need to take care not to have conflicts between real events (`FDs`) and our fake ones.

### Personal preference

Obviously *option 2* since that's the one that made it to the PR 😄 
You can also see *option 1* as work-in-progress in [this branch](https://github.com/acatangiu/firecracker/tree/pause-emulation), but it is incomplete as I just gave up on all the up and down casting somewhere along the way 😆 and looked for alternatives.

Aside from that, option 1 also has the downside of leaking or intentionally exposing internal logic as needed by step `1. ii`. The device is not independent, the `vmm` is the one to externally call `register/unregister` for the device; the device itself can't do that outside `trait Subscriber` functions without taking `&mut EventManager` outside of its `impl Subscriber`.

#### Option 2 interface improvements?

1. More generic naming, maybe not `pause/resume`?
2. Don't offer default impl for `pause/resume`, then you don't need other four control functions that control the default impl. Downside is more boilerplate code.
3. Probably many more 🙈 

### The big question

Like I said in the beginning, when evaluating options, we need to first answer this question:

Is *The Scenario* something that more users of `EventManager` are interested in? Or is it something `Firecracker` specific?

My money is on the former.

acatangiu (commented):
@andreeaflorescu 

> I think I might be missing something, but isn't it still possible to have a race with this implementation? It looks like we're just removing the event from the epoll loop, but this change will only be of effect at the next epoll run loop.

So let's say an `epoll.wait()` comes back with following events: `dev1-ev1, dev2-ev1, vmm-pause, dev3-ev1, dev2-ev2`:
1. It runs the associated `process()` functions sequentially so `dev1-ev1` is fully processed, then same for `dev2-ev1`.
2. On `vmm-pause` processing, `devs 1/2/3` unregister all their events.
3. When trying to process `dev3-ev1` and later `dev2-ev2`, `EventManager` doesn't find them in its hashmap anymore and just carries on: https://github.com/firecracker-microvm/firecracker/blob/5a9abb13a0072350bc560c077f063bc8a0a27289/src/polly/src/event_manager.rs#L201

Unless I'm missing smth, the effects are immediate, not at the next epoll run loop.

This obviously depends on the internal implementation of `EventManager` which is another reason why this functionality has better separation of concerns when part of the `Subscriber` interface instead of living outside.

> Beside this, when we pause the device, shouldn't we first check that there are no inflight requests? I would expect us to wait for any unfinished requests, and only then mark the device as Pause because otherwise we can still get in inconsistent states.

I might be missing something, but I don't think so. Any pending events are just dropped and on re-registration we just blindly kick all.

If we're worried about `Subscribers` that are sensitive to spurious events (afaik not the case in firecracker), you are right, we'd need to track pending events and only kick those on resume/re-registration.

@sandreim 

> Another approach that might be considered "hacky" is to have a dedicated event manager that we would run when the device emulation is stopped. This EM instance will have the API handler as its single subscriber. We only need to move the API handler subscriber to this secondary EM on Pause and back on Resume.

This sounds totally legit, I will do a quick PoC with this approach to validate it and open another RFC with it. Then we can more accurately compare between them.
In either case, I believe the right solution for upstream `EventManager` depends on *The big question* section in my other comment.

acatangiu:
I've hit a blocker on the two event managers approach:

The plan was to switch API to a _secondary evmgr_ when pausing emulation, and continue running the _secondary evmgr_ instead of the _main_ one for the duration of the pause. On resume, simply switch back to the _main evmgr_.

The issue is that on a given `evmgr.run()` loop, events from multiple sources are processed iteratively. After processing API event and do the switch, the _main evmgr_ does not run further loop iterations but **does finish the current loop**, possibly meaning device emulation events.

The end result is that the race is still there, device emulation can still happen _after_ `pause vm` completed.

We can fix this by:
1. Make the Pause API async 🤮 
2. Change `EventManager` to process events based on a priority queue, make API event lowest prio so it's always last.
3. Instead of using two `evmgrs` side by side and switch between them, use two cascaded `evmgrs` with API in the top one and emulation in the bottom one. On pause, un-register the bottom evmgr from the top one.

`1` is API breaking change, and a different model, customers would have to do some apis sync some async, adds a lot of complexity to firecracker, is just ugly from my point of view.
`2` looks like the best option if we go this way, but still needs `EventManager` changes
`3` only touches top-level application logic (nice), but adds one extra layer of indirection for each emulation event (one extra syscall - epoll_wait() - and another layer of hashmap search and dynamic dispatch for events), which sounds bad but probably won't affect performance in any practically measureable way.

I still like the approach in this RFC more than all of the above 😋

acatangiu:
@andreeaflorescu @sandreim @alexandruag please take another look.

The relevant commits are the middle 3:
```
  polly: add start and stop to Subscriber trait
  devices: add support for pausing and resuming emulation
  Add support for pausing and resuming emulation 
```

Please review these in order, you can ignore the other commits as they are only pseudo-related code beautification.

acatangiu:
Closing in favor of #2215

PR #2213, opened 2020-10-22
RFC: Cascaded EventManager for pausing and resuming emulation

### LINKED ISSUE #2193: Potential race condition during Pause/CreateSnapshot sequence

acatangiu:
## Reason for This PR

Fixes #2193 
Alternative to #2208 

## Description of Changes

Use two cascaded `EventManagers` with API in the top one and emulation in the bottom one. On VMM pause, un-register the bottom `EventManager` from the top one. On VMM resume, re-register the emulation `EventManager` under the main/top one.

### The good

I like this approach because it also separates `vmm` and emulation from other app-specific logic like API handling, metrics, etc.

The `vmm` owns its emulation `EventManager` which is unregistered or re-registered for running by the runtime-controller (the api controller) on vmm pause and respectively vmm resume.

There are a lot of line changes mainly because `builder`, `snapshot` and `tests` don't need to carry around the main `&mut EventManager`, the `Vmm` simply works with its internal evmgr and only exposes the FD for registering it under the main evmgr. The builder now builds a simple `Vmm` instead of an `Arc<Mutex<Vmm>>` already registered to an `EventManager`. The top layers use the `Vmm` whichever way they like.

Also we're not touching the `EventManager/Subscriber` interface - which is interpreted as good by some, irrelevant for others.

### The bad

Emulation goes through two layers of `EventManager` syscalls, lookups and dynamic dispatches. From a few rounds of simple tests, it looks like performance impact is visible:

| iperf3: guest as: |    Original    | Cascaded EvMgr |
|-------------------|:--------------:|----------------|
| client run 1      | 20.4 Gbits/sec | 20.1 Gbits/sec |
| client run 2      | 20.5 Gbits/sec | 19.4 Gbits/sec |
| server run 1      | 24.7 Gbits/sec | 21.5 Gbits/sec |
| server run 2      | 21.7 Gbits/sec | 20.2 Gbits/sec |

### The ugly

The current code in this PR is not enough.
If we're coming off a snapshot we may have lost pending events, so it's not enough to register the emulation event manager, we also need to kick all of its Subscribers.

This means, on resume, we either:
 - add a `.kick()` to `trait VirtioDevice` and go thru all `BusDevices` and `kick()` the `type == Virtio` ones,
 - add a `.kick()` to `trait Subscriber` and keep a list of subscribers like in the other RFC and just run thru them and kick them,
    or
 - (uglyish, but works) go through all `BusDevices`, dynamic cast them down to their concrete types, manually call `kick()/process()`

L.E. we might also need to do some device specific things on pause as well: block devices could be flushed for example (but I don't think that's mandatory), in the future we might identify corner cases with mandatory actions.

### DIFF
--- a/src/firecracker/src/api_server_adapter.rs
+++ b/src/firecracker/src/api_server_adapter.rs
@@ -168,16 +168,11 @@ pub fn run_with_api(
 
     // Configure, build and start the microVM.
     let (vm_resources, vmm) = match config_json {
-        Some(json) => super::build_microvm_from_json(
-            seccomp_filter,
-            &mut event_manager,
-            json,
-            &instance_info,
-            boot_timer_enabled,
-        ),
+        Some(json) => {
+            super::build_microvm_from_json(seccomp_filter, json, &instance_info, boot_timer_enabled)
+        }
         None => PrebootApiController::build_microvm_from_requests(
             seccomp_filter,
-            &mut event_manager,
             instance_info,
             || {
                 let req = from_api
@@ -198,6 +193,10 @@ pub fn run_with_api(
             boot_timer_enabled,
         ),
     };
+    let vmm = Arc::new(Mutex::new(vmm));
+    event_manager
+        .add_subscriber(vmm.clone())
+        .expect("Cannot register the vmm to the events manager.");
 
     // Start the metrics.
     firecracker_metrics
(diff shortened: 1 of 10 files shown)


acatangiu:
Closing in favor of #2215

Issue #1933, opened 2020-05-28
Does firecracker supports macvtap?


Vesnica:
I'm following the getting started guide, and trying to bind a macvtap to guest, which always give me this:
```
2020-05-28T14:03:41.904918179 [anonymous-instance:ERROR:src/api_server/src/parsed_request.rs:90] Received Error. Status code: 400 Bad Request. Message: Cannot open TAP device. Invalid name/permissions. CreateTap(Os { code: 22, kind: InvalidInput, message: Invalid argument })
```
I use following command to create macvtap and try to setup network for guest:
```
sudo ip link add link eno1 name macvtap0 type macvtap mode bridge
curl --unix-socket /tmp/firecracker.socket -i \
         -X PUT 'http://localhost/network-interfaces/eth0' \
         -H 'Accept: application/json'           \
         -H 'Content-Type: application/json'     \
         -d "{
                \"iface_id\": \"eth0\",
                \"host_dev_name\": \"macvtap0\"
            }"
```
firecracker version: `0.21.1`

sandreim:
Hi @Vesnica! Firecracker does not support macvtap yet. What is your exact usecase ? Macvtap simplifies the networking setup, but otherwise you can still use tap + bridge.

Vesnica:
@sandreim Using a bridge requires me to reconfig my network (move ip to bridge etc), using macvtap can avoid these troubles. 

I'm looking forward to this feature.

pothos:
To accepting macvtap devices you can either require them to be in the format `/dev/tapXY` and then opening here https://github.com/firecracker-microvm/firecracker/blob/49003d0f2f8573f6e1af6d22232c0917a0253189/src/devices/src/virtio/net/tap.rs#L113 and skipping that step https://github.com/firecracker-microvm/firecracker/blob/49003d0f2f8573f6e1af6d22232c0917a0253189/src/devices/src/virtio/net/tap.rs#L123
Actually it would be nicer if the `XY` number for `/dev/tapXY` is detected from its interface name via something like `let mut ifreq = ifreq_for(name);; ifreq_ioctl(socketfd, &mut ifreq, SIOCGIFINDEX).map(|ifindex| ifindex as usize);`

pothos:
[...]

PR #2217, opened 2020-10-24
Add prototype support for macvtap interfaces

upxe:
Firecracker uses /dev/net/tun to open/create tap interfaces. On the host these are typically bridged, using the standard Linux bridge implementation.

Linux macvtap interfaces are similar to bridges but simpler: the kernel doesn't snoop frames to maintain a FIB to map MAC addresses to interfaces and doesn't need to use flooding, because the full topology is known to the kernel.

However, macvtap interfaces are exposed to userland as /dev/tap%d device nodes, which Firecracker doesn't support. This commit adds support by checking whether the `host_dev_name` in the network configuration looks like a path name, and opening the device node if it does.

Tested thus:

```bash
    # Create a dummy interface to host the macvtap interfaces, plus two macvtap interfaces in bridge mode
    sudo ip link add dummy0 type dummy
    sudo ip link add link dummy0 name vtap0 type macvtap mode bridge
    sudo ip link add link dummy0 name vtap1 type macvtap mode bridge
    sudo ip link set dummy0 up
    sudo ip link set vtap0 up
    sudo ip link set vtap1 up
    sudo chmod 660 /dev/tap*             # For proof-of-concept purposes only
    sudo chown root:$(id -g) /dev/tap*   # For proof-of-concept purposes only
```
Configure two VMs thus; look in `/dev` for device node names; use `ip link` to get tap MAC address etc:
```
    {
        ...
            "network-interfaces" [
                {
                    ...
                    "host_dev_name: "/dev/tapXX",
                    "guest_mac": "<see host's ip link output>"
                }
            ]
        ...
    }
```
Start the VMs, give them IP addresses on the same subnet; they can then ping each other etc.

You must use the correct MAC address in the VM configuration, i.e. the MAC address that the host kernel has assigned to the tap interface.

## Reason for This PR

macvtap (see #1933) allows VMs on a host to be networked without using the Linux bridge code; it is simpler and not prone to ARP spoofing.

The implementation in the commit proves that macvtap-based networking works; I'm after feedback please as to whether macvtap support would be considered for merging. If so then I'd look at configuring properly rather than leeching off the existing /dev/net/tun configuration code.

Alternatively, a more flexible approach would be to allow Firecracker to adopt a file descriptor as a network interface and read/write frames to it. (In other words, in this case, open /dev/tapXX as e.g. FD 3, then exec() firecracker and have it treat FD 3 as the tap interface). This would let me do other useful things on the host, but would require a jailer change to keep the relevant FD[s] open.

## Description of Changes

Extend `Tap::open_named()` in `virtio/net/tap.rs` to support opening `/dev/tapXX` directly, rather than using `/dev/net/tun` to open/create tap interfaces. This allows Firecracker to use tap interfaces create via `macvtap` on the host.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

## PR Checklist

I haven't looked at these as this is an RFC; if macvtap (or passing in FDs) is acceptable then I'll submit a PR that meets them.

### DIFF
--- a/src/devices/src/virtio/net/tap.rs
+++ b/src/devices/src/virtio/net/tap.rs
@@ -5,13 +5,23 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 
-use net_gen::ifreq;
-use std::fs::File;
+use std::fs::{File, OpenOptions};
 use std::io::{Error as IoError, Read, Result as IoResult, Write};
-use std::os::raw::*;
-use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
-use utils::ioctl::{ioctl_with_mut_ref, ioctl_with_ref, ioctl_with_val};
-use utils::{ioctl_expr, ioctl_ioc_nr, ioctl_iow_nr};
+use std::os::{
+    raw::*,
+    unix::{
+        fs::{FileTypeExt, OpenOptionsExt},
+        io::{AsRawFd, FromRawFd, RawFd},
+    },
+};
+use std::path::Path;
+
+use net_gen::ifreq;
+use utils::{
+    ioctl::{ioctl_with_mut_ref, ioctl_with_ref, ioctl_with_val},
+    ioctl_expr, ioctl_ioc_nr, ioctl_iow_nr,
+    net::macvtap::MacVTap,
+};
 
 // As defined in the Linux UAPI:
 // https://elixir.bootlin.com/linux/v4.17/source/include/uapi/linux/if.h#L33
@@ -24,10 +34,16 @@ pub enum Error {
     CreateTap(IoError),
     /// Invalid interface name.
     InvalidIfname,
+    /// Tap interface device is not a character device.
+    InvalidTapDevType,
     /// ioctl failed.
     IoctlError(IoError),
+    /// Unable to open tap interface device.
+    OpenTapDev(IoError),
     /// Couldn't open /dev/net/tun.
     OpenTun(IoError),
+    /// Unable to stat tap interface device for macvtap interface.
+    StatTapDev(IoError),
 }
 
 pub type Result<T> = ::std::result::Result<T, Error>;
@@ -100,11 +116,66 @@ impl IfReqBuilder {
 }
 
 impl Tap {
-    /// Create a TUN/TAP device given the interface name.
+    /// * `if_name` - the name of the interface.
+    /// Create a TUN/TAP device given the tap or macvtap interface name.
     /// # Arguments
     ///
     /// * `if_name` - the name of the interface.
     pub fn open_named(if_name: &str) -> Result<Tap> {
+        // Options:
+        //  - /dev/net/<if_name> exists; open it.
+        //  - It's a macvtap device: determine by checking /sys; open the
+        //    corresponding /dev/tapX node.
+        //  - It's a tap device: open /dev/net/tun and allocate via SETIFF.
+        if let Ok(path) = MacVTap::get_device_node(if_name) {
+            Self::macvtap_open_named(if_name, &path)
+        } else {
+            Self::tap_open_named(if_name)
+        }
+    }
+
+    /// Create a TUN/TAP device given the macvtap interface name and device node.
+    /// # Arguments
+    ///
+    /// * `if_name` - the name of the interface.
+    /// * `dev_path` - location of the interface's device node.
+    fn macvtap_open_named(if_name: &str, dev_path: &Path) -> Result<Tap> {
+        // Open the device node
+        let mut opts = OpenOptions::new();
+        let tap_file = opts
+            .read(true)
+            .write(true)
+            .custom_flags(libc::O_NONBLOCK | libc::O_CLOEXEC)
+            .open(dev_path)
+            .map_err(Error::OpenTapDev)?;
+
+        // Must be a char device
+        let md = tap_file.metadata().map_err(Error::StatTapDev)?;
+        if !md.file_type().is_char_device() {
+            return Err(Error::InvalidTapDevType);
+        }
+
+        // The length check is probably unnecessary because we know that the
+        // network interface is valid at this point, but it doesn't hurt.
+        let name_bytes = if_name.as_bytes();
+        if name_bytes.len() >= IFACE_NAME_MAX_LEN {
+            return Err(Error::InvalidIfname);
+        }
+
+        let mut ret = Tap {
+            tap_file,
+            if_name: [0; IFACE_NAME_MAX_LEN],
+        };
+
+        ret.if_name[..name_bytes.len()].copy_from_slice(name_bytes);
+        Ok(ret)
+    }
+
+    /// Create a TUN/TAP device given the tap interface name.
+    /// # Arguments
+    ///
+    /// * `if_name` - the name of the interface.
+    fn tap_open_named(if_name: &str) -> Result<Tap> {
         let terminated_if_name = build_terminated_if_name(if_name)?;
 
         let fd = unsafe {
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -9,11 +9,12 @@ use std::os::unix::process::CommandExt;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
-use crate::cgroup;
 use crate::cgroup::Cgroup;
 use crate::chroot::chroot;
+use crate::{cgroup, to_cstring};
 use crate::{Error, Result};
 use utils::arg_parser::Error::MissingValue;
+use utils::net::macvtap::MacVTap;
 use utils::syscall::SyscallReturnCode;
 use utils::{arg_parser, validators};
 
@@ -64,6 +65,7 @@ pub struct Env {
     start_time_cpu_us: u64,
     extra_args: Vec<String>,
     cgroups: Vec<Cgroup>,
+    macvtaps: Vec<String>,
 }
 
 impl Env {
@@ -157,6 +159,15 @@ impl Env {
             }
         }
 
+        // macvtap arg format: --macvtap if_name => create device node /dev/net/if_name in the chroot.
+        let mut macvtaps = Vec::new();
+        // We do not create here the MacVTaps since we need to do join_netns before that.
+        if let Some(macvtap_args) = arguments.multiple_values("macvtap") {
+            for arg in macvtap_args {
+                macvtaps.push(arg.to_string());
+            }
+        }
+
         Ok(Env {
             id: id.to_owned(),
             chroot_dir,
@@ -169,6 +180,7 @@ impl Env {
             start_time_cpu_us,
             extra_args: arguments.extra_args(),
             cgroups,
+            macvtaps,
         })
     }
 
@@ -275,12 +287,57 @@ impl Env {
         // namespace. Safe because we are passing valid parameters.
         SyscallReturnCode(unsafe { libc::close(netns_fd) })
             .into_empty_result()
-            .map_err(Error::CloseNetNsFd)
+            .map_err(Error::CloseNetNsFd)?;
+
+        // Since namespaces are shared by default when creating a new process using fork or clone,
+        // unshare() is used to disassociate (unshare) the current process from the mount namespace:
+        // https://linux.die.net/man/2/unshare.
+        SyscallReturnCode(unsafe { libc::unshare(libc::CLONE_NEWNS) })
+            .into_empty_result()
+            .map_err(Error::MountSysfs)?;
+
+        // It is not sufficient to join the new network namespace,
+        // we also need to mount a version of /sys that describes
+        // the network namespace.
+
+        // Don't let any mounts propagate back to the parent.
+        // This means that the sysfs of the network namespace will only
+        // get mounted for the jailer process.
+        SyscallReturnCode(unsafe {
+            libc::mount(
+                to_cstring("")?.as_ptr(),
+                to_cstring("/")?.as_ptr(),
+                to_cstring("none")?.as_ptr(),
+                libc::MS_SLAVE | libc::MS_REC,
+                std::ptr::null(),
+            )
+        })
+        .into_empty_result()
+        .map_err(Error::MountSysfs)?;
+
+        // Unmount the current sysfs since it's describing the previous namespace.
+        let cstr_sys = to_cstring(Path::new("/sys"))?;
+        SyscallReturnCode(unsafe { libc::umount2(cstr_sys.as_ptr(), libc::MNT_DETACH) })
+            .into_empty_result()
+            .map_err(Error::UmountSysfs)?;
+
+        // Actually mount the sysfs corresponding to the current namespace.
+        SyscallReturnCode(unsafe {
+            libc::mount(
+                std::ptr::null(),
+                cstr_sys.as_ptr(),
+                to_cstring("sysfs")?.as_ptr(),
+                0,
+                std::ptr::null(),
+            )
+        })
+        .into_empty_result()
+        .map_err(Error::MountSysfs)
     }
 
     #[cfg(target_arch = "aarch64")]
     fn copy_cache_info(&self) -> Result<()> {
-        use crate::{readln_special, to_cstring, writeln_special};
+        use crate::{readln_special, writeln_special};
 
         const HOST_CACHE_INFO: &str = "/sys/devices/system/cpu/cpu0/cache";
         // Based on https://elixir.free-electrons.com/linux/v4.9.62/source/arch/arm64/kernel/cacheinfo.c#L29.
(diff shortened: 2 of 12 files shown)


serban300:
Hi @upxe ,

Thanks for the contribution ! This might be very useful. We would like to do some performance tests first, and we can discuss next steps after. In the meanwhile, just dropping a few questions:

- is this compatible with the snapshotting functionality ? For example right now when we restore from a snapshot we just recreate a tap with the same name as the old one. Do you see any complications here when using a macvtap instead ?
- currently we have some unit tests that simulate frame tx/rx on the tap. Is this possible using the macvtap ?

upxe:
> * is this compatible with the snapshotting functionality ? For example right now when we restore from a snapshot we just recreate a tap with the same name as the old one. Do you see any complications here when using a macvtap instead ?

I think that this depends on how you name your tap devices. The default (on Debian at any rate) isn't very useful: the `XX` in `/tap/XX` is the interface index, entirely unrelated to whatever name you gave the interface, and that appears in the `ip link` output. You might be able to convince `udev` (or whatever you use to populate `/dev`) to give you a better name, but just opening the same device node by name is probably a bad idea for these interfaces.

However, `/sys/devices/virtual/net/<interface name>/macvtap/*/dev` tells you what device node to open (or, indeed, to create in the jailer chroot). So: shouldn't be a problem to figure out what device to create or open, but you couldn't use the same code path as the existing tap device does.

This all means that we would probably need to distinguish between tap and macvtap interfaces at the configuration level, rather than relying on the interface name looking like `/dev/something`. That's probably a good idea anyway though.

> * currently we have some unit tests that simulate frame tx/rx on the tap. Is this possible using the macvtap ?

Haven't look into this yet - I raised the PR to see whether there was interest in principle in supporting this before diving in too deep. In theory, shouldn't be a problem, because once you've opened the device it's more-or-less a normal tap device. I shall have a look, but have had a few problems running the test suite - the whole unit test + integration test run (`tools/devtool test`) takes ages on my laptop, so I'm running `cargo test` within a shell in the Docker container. That seems to work, but a few tests fail in the current master.

gc-plp:
Hi @upxe. 
I just wanted to do a quick follow-up on this PR. Recently, I ran a quick test on this PR using iperf3 and noticed a slight decrease in throughput when reported to the master Firecracker version, but also saw that the overall CPU usage in the host is lower, which is something that we would like to look into more.

I will redo the tests and come back with the exact test scenario and numbers.

pclesr:
Being able to specify the absolute path to the taped is extremely useful, as it allows unique names . 

In an embedded environment, there is the desire to put each VM into a network namespace. Using MacVTap makes the configuration easier, as the Macvtap can be attached directly to the `veth` device in the network namespace without having to make another bridge. This simplifies the setup considerably. The environment is also sandboxed (like with the jailer) with other namespaces like mount, pid, etc to help isolate the environment.

acatangiu:
Also asked for in #1933

gc-plp:
Hi @upxe. I think that we can start working towards merging this, but the first step would be to point the PR towards a feature branch. I've created one called [feature/macvtap](https://github.com/firecracker-microvm/firecracker/tree/feature/macvtap) - please update the PR when you have the time.

upxe:
I've updated the PR; unless you're happy with the current approach (which was just a PoC, and I think is a bit too simple-minded), it would be useful to have a steer on how you want to proceed, e.g.

 - Whether you want to support both the current `/dev/net/tun` approach and `macvtap` interfaces, or see the macvtap approach replacing the existing one.

 - Whether checking for an absolute path is OK to use the `macvtap` implementation, or whether it look up the interface index given the interface name and see whether a suitable `/dev/tapN` exists. (Would probably play better with snapshotting.)

Apart from the above, I'm aware at the moment that I need to:

 - Add jailer support.
 - Ensure that snapshotting works. This will have to be aware of MAC addresses - the current code appears to persist the MAC address, in which case snapshot restoration will also have to set the host MAC address of the `macvtap` interface. I think that that's fine though.
 - Look into unit testing.
 - Consider doc changes.
 - Sign commits.

gc-plp:
@upxe, indeed, all the items that you mentioned will have to be added, but I haven't looked into what the best approach would be to steer it towards the final implementation, given all the implications that it may have with snapshotting and other areas such as isolation between microVMs. This is why, as a first step, we should start working towards merging it into a feature branch.

Related to your first two bullets, preferably we would like to have both `/dev/net/tun` and `macvtap` available. Having absolute paths inside the snapshots would probably be a bad idea, correct.

Also, I would also test it through the Python framework, not only using rust unit tests. The snapshotting feature would also have to be tested through Python tests when macvtap snapshots are restored.

georgepisaltu:
Hi @upxe!

Do you still plan on contributing to this PR?

upxe:
> Do you still plan on contributing to this PR?

I do; the day job intervened in December. Might get a chance to look at it this weekend.

btw, I had (thought that I'd) put it onto a feature branch on 28 Nov so was a little confused by @gc-plp's comment about "merging towards a feature branch" - if I haven't interpreted that correctly then please let me know.

Thanks.

gc-plp:
@upxe, the PR is pointed to the correct branch now, but before merging we would need to have the items that you mentioned [here](https://github.com/firecracker-microvm/firecracker/pull/2217#issuecomment-735219189), addressed.

casibbald:
@upxe Been going through this and this PR is a great addition to FC, thank you so much. It happens to be functionality that is a blocker on a concept we are looking at.
Day job permitting, how are you progressing with the additional requirements to get the PR ready for final review and merging?

upxe:
I'm looking at Jailer integration and there's a snag: macvtap interfaces are (AFAICT - shall investigate further) accessed via per-interface device nodes in /dev, rather than via /dev/net/tun. These individual device nodes aren't in the jailed firecracker's view of /dev, and the Jailer can't create then (like it does for /dev/kvm and /dev/net/tun) because it doesn't know which interfaces the VM will use.

Options that I can think of (unless I can somehow make the interfaces accessible via /dev/net/tun):

1. Put a read-only sysfs mount into the chroot. Simplest and most flexible, but exposes rather more of the host to the VM than is ideal.

2. Jailer finds all macvtap interfaces (via sysfs) and creates device nodes in the chroot for all such interfaces that the firecracker uid/gid would be able to open. Potentially exposes unnecessary interfaces to the VM, but can be avoided by assigning a unique UID/GID to each VM, which is probably a good idea anyway. The macvtap interfaces must exist before running the Jailer.

3. Make a jailer user declare via command-line args the names of any macvtap interfaces that the VM will use, so that the Jailer can create the relevant device nodes. Puts minimal device nodes into the chroot, at the expense of some managerial/orchestration complexity.

Do the maintainers have any views on those? I'm minded to prefer option 2.

casibbald:
Hi Upxe,

could you drop me a private email? I have a few questions about the various
implementation ideas.

Kind regards

Charles

gc-plp:
@upxe I think that the question to ask here would be, what happens if a sandbox escape occurs and an attacker is able to access the jailer environment? Which of the three options would ensure that the attacker doesn't disturb a neighboring VM?

From this angle, option 3 seems to be ensure a minimal exposure of host devices inside the jailer. As you mentioned, there is some additional complexity passed on to the orchestrator, but a piece of extra logic would still have to be added in order to connect a microVM through a macvtap interface (e.g. creating the actual interface before launching the jailer).

gc-plp:
Hi @upxe 
Have you had any chance to continue the work on this PR?

upxe:
I'm happy to go with a command line argument in the jailer to have it create some additional device nodes.

The last remaining issue for now (haven't considered save/restore yet) is that I'm currently using sysfs to map interface names to device nodes, and sysfs isn't available in the jailed root. But if we're creating device nodes in the jailer then we just name them appropriately and have firecracker look for them if /sys isn't mounted.

If the host has a macvtap interface called foo that the user wants to use in a VM then we end up with:

1. User runs the jailer with e.g. `--macvtap foo`.
2. Jailer looks in host sysfs for the foo interface's major/minor, and creates /dev/net/foo in the jailed root.
3. When firecracker wants to open a network interface "foo":
    1. If /dev/net/foo exists then open it and treat as a macvtap interface.
    2. Else if /sys exists (i.e. no jailer) then see whether it's a macvtap interface, act as per jailer (and the existing code on this branch).
    3. Else (no sys, no /dev/net/foo) use /dev/net/tun as per normal.

Incidentally, we can distinguish tap from mavctap via the ethtool GDRVINFO ioctl, but AFAICT we'd still need sysfs to figure out the major/minor numbers for the device node.

dianpopa:
> I'm happy to go with a command line argument in the jailer to have it create some additional device nodes.
> 
> The last remaining issue for now (haven't considered save/restore yet) is that I'm currently using sysfs to map interface names to device nodes, and sysfs isn't available in the jailed root. But if we're creating device nodes in the jailer then we just name them appropriately and have firecracker look for them if /sys isn't mounted.
> 
> If the host has a macvtap interface called foo that the user wants to use in a VM then we end up with:
> 
> 1. User runs the jailer with e.g. `--macvtap foo`.
> 2. Jailer looks in host sysfs for the foo interface's major/minor, and creates /dev/net/foo in the jailed root.
> 3. When firecracker wants to open a network interface "foo":
>    
>    1. If /dev/net/foo exists then open it and treat as a macvtap interface.
>    2. Else if /sys exists (i.e. no jailer) then see whether it's a macvtap interface, act as per jailer (and the existing code on this branch).
>    3. Else (no sys, no /dev/net/foo) use /dev/net/tun as per normal.
> 
> Incidentally, we can distinguish tap from mavctap via the ethtool GDRVINFO ioctl, but AFAICT we'd still need sysfs to figure out the major/minor numbers for the device node.

Hi @upxe 

I will be helping you with steps on how to proceed further here. Currently I am a little busy with some aarch64 work but I will be coming back with an answer as soon as I can.

Thanks for the involvement,
Diana

dianpopa (commented):
Hi @upxe,

I started reviewing your PR, I have one high level request. Would be much easier to go through the changes if the commits were rearranged such that:
* First commit (i.e add prototype support for macvtap interfaces), second one and the last could be squashed into one commit, When doing the squash, you can leave the messages from the first and second commit since I think both of them are important to keep around
* After the squash, I would move the commit that adds macvtap utils support to be the first one
* last commit would be the one adding jailer support
WDYT?

dianpopa:
@upxe in order to make it easier to exchange suggestions on the PR,  I encourage you to join our [slack workspace](https://join.slack.com/t/firecracker-microvm/shared_invite/enQtNDY2NTUwMzQ3MDE1LWIwMzA0OWFkMTZhMTlmMDZiMmFkYjMyODMxMGQ1ZjliMzJjNjJiNWRhNWNkOGEyNmUxNmRkMjZhYTc3MmVjZjM). Feel free to ping me if you decide to join. Thanks!

dianpopa:
Hi @upxe ,

We have raised the priority on merging this since we need to have this implementation in Firecracker as soon as possible. In order to speed things up, I would like to be able to contribute to this PR too. Would this be ok with you? 
Looking forward to your answer. Thanks!

upxe:
@dianpopa - no problem at all wrt contributing. Main things from my perspective:

 - I haven't looked at snapshot/restore at all - I haven't needed that functionality myself, so am not familiar with that part of the codebase. It might just work, but then again, it might not.
 - Testing: I can't easily test on ARM, and last time I looked at AMD64 there were some failures on my box that I don't think are related to these changes.
 - I was going to reorganise the commits to be a bit more coherent for merging, so your prod on that is timely. I'll look at that this weekend (I don't have time / permission to work on this in the week, so it's at most half a day at the weekend unfortunately).

BTW, since my work on this has been sporadic I've been rebasing my branch and force-pushing - I think that that gives a cleaner history than periodically merging master into a long-lived feature branch. However, it seems to confuse Github a bit, although your force-push looks to have sorted that out. If you want me to do something else in this regard then please let me know.

src/jailer/src/env.rs
```diff
@@ -275,12 +287,56 @@ impl Env {
         // namespace. Safe because we are passing valid parameters.
         SyscallReturnCode(unsafe { libc::close(netns_fd) })
             .into_empty_result()
-            .map_err(Error::CloseNetNsFd)
+            .map_err(Error::CloseNetNsFd)?;
+
+        // Since namespaces are shared by default when creating a new process using fork or clone,
+        // unshare() is used to disassociate the current process from the default shared namespace.
+        SyscallReturnCode(unsafe { libc::unshare(libc::CLONE_NEWNS) })
```
acatangiu, line 294: I would explicitly add in the comment that this unshares the _mount_ namespace.

src/jailer/src/env.rs
```diff
@@ -275,12 +287,56 @@ impl Env {
...
+                to_cstring("/")?.as_ptr(),
+                to_cstring("none")?.as_ptr(),
+                libc::MS_SLAVE | libc::MS_REC,
+                std::ptr::null(),
+            )
+        })
+        .into_empty_result()
+        .map_err(Error::MountSysfs)?;
+
+        // Unmount the default sysfs.
```
acatangiu, line 317: nit:
```suggestion
        // Unmount the current sysfs since it's describing the previous namespace.
```

tests/integration_tests/functional/test_net.py
```diff
@@ -50,3 +55,65 @@ def test_high_ingress_traffic(test_microvm_with_ssh, network_config):
...
     exit_code, _, _ = ssh_connection.execute_command('echo success\n')
     assert exit_code == 0
+
+
+@decorators.test_context('ssh_and_balloon', NO_OF_VMS)
+def test_macvtaps(test_multiple_microvms):
+    """Check the MacVTap functionality."""
+    microvms = test_multiple_microvms
+    # Creating a bridge for the macvtap interfaces.
+    bridge = net_tools.Bridge("bridge{}".format(randint(1, 10000)))
```
gc-plp, line 65: Do we need to `randint()` here? If we're trying to make sure that we don't create multiple bridges with the same name we could just check if the bridge exists and tear it down if it does.

src/jailer/src/env.rs
dianpopa: Done.

src/jailer/src/env.rs
dianpopa: Done.

src/devices/src/virtio/net/tap.rs
```diff
@@ -100,11 +116,66 @@ impl IfReqBuilder {
...
+            .open(dev_path)
+            .map_err(Error::OpenTapDev)?;
+
+        // Must be a char device
+        let md = tap_file.metadata().map_err(Error::StatTapDev)?;
+        if !md.file_type().is_char_device() {
+            return Err(Error::InvalidTapDevType);
+        }
+
+        // The length check is probably unnecessary because we know it's an extant
```
georgepisaltu, line 158: ultra-nit: Although an _extant_ word in English dictionaries, "extant" is not a term I'd use in this comment.

```diff
- // The length check is probably unnecessary because we know it's an extant
+ // The length check is probably unnecessary because we know it's an existent/valid
```

georgepisaltu (approved):

src/utils/src/net/macvtap.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    #[test]
+    fn test_is_normal_path_component() {
+        assert!(!is_normal_path_component(""));
+        assert!(!is_normal_path_component("."));
+        assert!(!is_normal_path_component(".."));
+        assert!(!is_normal_path_component("/"));
+        assert!(!is_normal_path_component("./x"));
+        assert!(!is_normal_path_component("../x"));
+        assert!(!is_normal_path_component("../x"));
```
luminitavoicu, line 154: super nit: these two checks seem to be duplicated. You could remove one of them.

src/utils/src/net/macvtap.rs
```diff
@@ -0,0 +1,165 @@
...
+        assert!(!is_normal_path_component("x/./x"));
+        assert!(!is_normal_path_component("x/"));
+        assert!(!is_normal_path_component("x/y"));
+        assert!(!is_normal_path_component("/.."));
+        assert!(!is_normal_path_component("../"));
+        assert!(!is_normal_path_component("x/y"));
```
luminitavoicu, line 160: Same with this one: duplicate of the one on line 157.

dianpopa: Done.

src/utils/src/net/macvtap.rs
dianpopa: Done.

dianpopa (commented):
@gc-plp @georgepisaltu @acatangiu @luminitavoicu I addressed your feedback, PTAL! Thanks

georgepisaltu (approved):

luminitavoicu (approved):

majek:
Rebase please  - the branch conflicts with main.

upxe:
done on upxe:macvtap; don't know how to update this pull request because it's showing as merged + closed

dzenc:
Despite the fact that this is marked merged + closed, it doesn't seem to be in there?   Did this "die on the vine" somewhere?  This seems like a lot of work, and a lot of value to be dropped.  (I would like to make use of it!)

What can I do to help?

dzenc:
Comment added to #1933.  Thanks for listening!  :)

PR #2226, opened 2020-10-27
Improvements around network latency performance test

iulianbarbu:
## Reason for This PR

The test is failing intermittently.

## Description of Changes

* Pinned uVM threads to physical cores.
* Adjusted latency baseline and delta to cover lows and highs which were observed outside the current interval.
* Fixed a mapping issue: AVG latency was MAX latency.

## License Acceptance

## PR Checklist

- ~~[] Any required documentation changes (code and docs) are included in this PR.~~
- ~~[] Any newly added `unsafe` code is properly documented.~~
- ~~[] Any API changes are reflected in `firecracker/swagger.yaml`.~~
- ~~[] Any user-facing changes are mentioned in `CHANGELOG.md`.~~
- ~~[] All added/changed functionality is tested.~~

### DIFF
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -136,6 +136,7 @@ class Microvm:
         # Cpu load monitoring has to be explicitly enabled using
         # the `enable_cpu_load_monitor` method.
         self._cpu_load_monitor = None
+        self._vcpus_count = None
 
         # External clone/exec tool, because Python can't into clone
         self.bin_cloner_path = bin_cloner_path
@@ -354,6 +355,43 @@ class Microvm:
             self._api_socket,
             self._api_session
         )
+
+    @property
+    def vcpus_count(self):
+        """Get the vcpus count."""
+        return self._vcpus_count
+
+    @vcpus_count.setter
+    def vcpus_count(self, vcpus_count: int):
+        """Set the vcpus count."""
+        self._vcpus_count = vcpus_count
+
+    def pin_vmm(self, cpu_id: int) -> bool:
+        """Pin the firecracker process VMM thread to a cpu list."""
+        if self.jailer_clone_pid:
+            for thread in utils.ProcessManager.get_threads(
+                    self.jailer_clone_pid)["firecracker"]:
+                utils.ProcessManager.set_cpu_affinity(thread, [cpu_id])
+                return True
+        return False
+
+    def pin_vcpu(self, vcpu_id: int, cpu_id: int):
+        """Pin the firecracker vcpu thread to a cpu list."""
+        if self.jailer_clone_pid:
+            for thread in utils.ProcessManager.get_threads(
+                    self.jailer_clone_pid)[f"fc_vcpu {vcpu_id}"]:
+                utils.ProcessManager.set_cpu_affinity(thread, [cpu_id])
+            return True
+        return False
+
+    def pin_api(self, cpu_id: int):
+        """Pin the firecracker process API server thread to a cpu list."""
+        if self.jailer_clone_pid:
+            for thread in utils.ProcessManager.get_threads(
+                    self.jailer_clone_pid)["fc_api"]:
+                utils.ProcessManager.set_cpu_affinity(thread, [cpu_id])
+            return True
+        return False
 
     def spawn(self, create_logger=True, log_file='log_fifo', log_level='Info'):
         """Start a microVM as a daemon or in a screen session."""
--- a/tests/integration_tests/performance/test_network_latency.py
+++ b/tests/integration_tests/performance/test_network_latency.py
@@ -13,14 +13,15 @@ from framework.matrix import TestMatrix, TestContext
 from framework.builder import MicrovmBuilder
 from framework.statistics import core, consumer, producer, types, criteria,\
     function
-from framework.utils import eager_map
+from framework.utils import eager_map, CpuMap
+from framework.builder import DEFAULT_HOST_IP
 
 
 PING = "ping -c {} -i {} {}"
 BASELINES = {
     "x86_64": {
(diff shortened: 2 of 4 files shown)


tests/integration_tests/performance/test_network_latency.py
```diff
@@ -180,10 +181,26 @@ def _g2h_send_ping(context):
...
-
     _tap, host_ip, _ = basevm.ssh_network_config(network_config, '1')
 
     basevm.start()
+
+    # Pin uVM threads to physical cores.
+    match = re.match("([0-9]+)vcpu_[0-9]+mb", context.microvm.name())
+    parallel = int(match.group(1))
+    basevm_threads = ProcessCpuAffinity(basevm.jailer_clone_pid).get_threads()
+    cpu_aff_curr = 0
```
serban300, line 192: Nit: I think `current_cpu_id` sounds better.

tests/integration_tests/performance/test_network_latency.py
```diff
@@ -180,10 +181,26 @@ def _g2h_send_ping(context):
                               ssh_key=ssh_key,
                               config=context.microvm,
                               network_config=network_config)
-
     _tap, host_ip, _ = basevm.ssh_network_config(network_config, '1')
 
     basevm.start()
+
+    # Pin uVM threads to physical cores.
+    match = re.match("([0-9]+)vcpu_[0-9]+mb", context.microvm.name())
```
serban300, line 189: Can't we find the number of vcpus in a different way ? Do we have access to the microvm macine config for example ?

iulianbarbu: We can issue a `nproc` on the guest. I'll try it and let you know how it goes.

tests/integration_tests/performance/test_network_latency.py
iulianbarbu: Sure.

tests/integration_tests/performance/test_network_latency.py
gc-plp: Can we implement an accessor within the microvm() class so that we pick up the number of cpus set at startup?

iulianbarbu: This sounds like a really nice approach. I'll try this.

tests/integration_tests/performance/test_network_latency.py
```diff
@@ -180,10 +181,26 @@ def _g2h_send_ping(context):
...
+    parallel = int(match.group(1))
+    basevm_threads = ProcessCpuAffinity(basevm.jailer_clone_pid).get_threads()
```
gc-plp, line 191: Could the `ProcessCpuAffinity` class be part of the microvm class?

Creating objects out at runtime that are spawned out of the current PIDs is something that should be done under the hood. For example:

```
basevm_threads = basevm.get_threads() # returns a map of ProcessCpuAffinity objects

basevm_threads["fc_api"].set_cpu_affinity(cpu_aff_curr)
cpu_aff_curr += 1

basevm_threads["firecracker"].set_cpu_affinity(cpu_aff_curr)
cpu_aff_curr += 1

```

Or even simpler (given that it's a list):

```
for pid in basevm.get_threads():
    pid.set_cpu_affinity(cpu_aff_curr)
    cpu_aff_curr += 1

```

iulianbarbu:
@gc-plp , @serban300 , please take a look again! I've addressed your feedback.

tests/framework/microvm.py
```diff
@@ -355,6 +356,43 @@ def init_snapshot_api(self):
...
+    def pin_vmm(self, cpu_ids_list: list) -> bool:
+        """Pin the firecracker process VMM thread to a cpu list."""
+        if self.jailer_clone_pid:
+            for thread in utils.ProcessManager.get_threads(
+                    self.jailer_clone_pid)["firecracker"]:
+                utils.ProcessManager.set_cpu_affinity(thread, cpu_ids_list)
+                return True
+        return False
+
+    def pin_vcpu(self, vcpu_id: int, cpu_ids_list: list):
```
serban300, line 378: why do we need a list of cpu_ids ? Would it make sense to pin a single vCPU to multiple CPUs ?

iulianbarbu: I can not tell if there is a specific usecase right now, but we might want to set the affinity for different vCPU threads to different cpu lists, pertaining to different NUMA nodes, not necessarily to a single core. I think we should see this API as support for process cpu affinity modifications rather than strict process/cpu pinning. We can restrict the support for now or rename the API to suggest better that it is intended to be used for setting cpu affinity for microvm threads, as in "set_vcpu_thread_cpu_affinity`. WDYT?

gc-plp:
Looks good, but it seems like we have an unrelated failure.

tests/framework/microvm.py
serban300: Personally I would restrict the support for now. I would only offer the ability to pin a vCPU to 1 cpu_id. The same for vmm, api, etc.

gc-plp: Could we just pass a mask here? (e.g. `0x3`)

It would work for both one physical core and multiple cores.

iulianbarbu: Done.

iulianbarbu:
> Looks good, but it seems like we have an unrelated failure.

I've opened an issue: https://github.com/firecracker-microvm/firecracker/issues/2257. I see that the CI fails again for the test and on amd we have unrelated failures for `test_5_full_snapshots`.

serban300 (dismissed):

gc-plp (approved):

serban300 (approved):

PR #2227, opened 2020-10-27
[performance] added network TCP throughput test

iulianbarbu:
## Reason for This PR

Lack of TCP network throughput performance test.

## Description of Changes

Added a test for TCP network throughput for 1/2 vCPUs uVMs. The test contains:
- guest to host workloads
- host to guest workloads
- bidirectional workloads

## License Acceptance

## PR Checklist

- ~~[] Any required documentation changes (code and docs) are included in this PR.~~
- ~~[] Any newly added `unsafe` code is properly documented.~~
- ~~[] Any API changes are reflected in `firecracker/swagger.yaml`.~~
- ~~[] Any user-facing changes are mentioned in `CHANGELOG.md`.~~
- ~~[] All added/changed functionality is tested.~~


tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -0,0 +1,396 @@
...
+        "g2h": [""],
+        "h2g": ["-R"],
+        "bd": ["", "-R"]
+    },
+    "protocols": [
+        {
+            "name": "tcp",
+            "omit": 1,  # seconds
+            # Used as KBytes - congestion window.
+            "window_size": [10, 64, 128, 256],
```
raduiliescu, line 35: I think you have too many options for window size. Also you should leave it to run with default params, including packet length.

tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -0,0 +1,396 @@
...
+from framework.matrix import TestMatrix, TestContext
+from framework.builder import MicrovmBuilder
+from framework.statistics import core, consumer, producer, criteria, \
+    types, function
+from framework.utils import CpuMap, ProcessCpuAffinity, CmdBuilder, run_cmd, \
+    eager_map
+import host_tools.network as net_tools
+
+CONFIG = {
+    "time": 2,  # seconds
```
raduiliescu, line 23: time is too small, TCP takes some time to get to the best window size. I recommend at least 15s omitting first 4-5s.

tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -0,0 +1,396 @@
...
+    "protocols": [
+        {
+            "name": "tcp",
+            "omit": 1,  # seconds
+            # Used as KBytes - congestion window.
+            "window_size": [10, 64, 128, 256],
+            "pkt_size": [128, 256, 512, 1024],
+        }
+    ],
+    "baseline_bw": {  # Mibps
```
raduiliescu, line 39: this is hard to read... I also suggest to put it in a different file so it can be generated and name them goldvalues, or something more suggestive.
Is pretty hard to understand and change them. I recommend you add a mode where you can generate the gold for example a command like ./test_throughput --generate_goldvalues --delta=0.1 ... which would generate the files.

tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -0,0 +1,396 @@
...
+DURATION_TOTAL = "duration_total"
+RETRANSMITS = "retransmits"
+RETRANSMITS_TOTAL = "retransmits_total"
+CPU_UTILIZATION_HOST = "cpu_utilization_host"
+CPU_UTILIZATION_HOST_TOTAL = "cpu_utilization_host_total"
+CPU_UTILIZATION_GUEST = "cpu_utilization_guest"
+CPU_UTILIZATION_GUEST_TOTAL = "cpu_utilization_guest_total"
+
+BASE_PORT = 5000
+
```
raduiliescu, line 157: While the tests are ok, they are missing a point - how does the emulation perform in a high throughput scenarios.
Do these tests fill up the emulation? How does the performance look when you generate a lot of contention?
I recommend to run tests with different vm sizes - 1vcpus, 2, 4, 8, 16 and also try multiple iperfs processes for each vm type 1, 2, 4, 8, 16, 128. 
Also after this PR please try with multiple rate-limiter configs, so we can know the penalty added by the rate-limiter.

iulianbarbu: Trying with different uVMs machine configs definitely makes sense. Adding these scenarios is straight forward, starting from this test, but I suggest we keep this as it is right now, following up with uVMs of 4, 8, 16, 128 vCPUs in the future. As we have seen for what we have right now, getting baselines for these uVMs take some time. Once this PR is merged, it is just a matter of adding a new microvm artifact and getting the baselines. WDYT?

tests/integration_tests/performance/test_network_tcp_throughput.py
iulianbarbu: Reduced the options for window size. PTAL!

tests/integration_tests/performance/test_network_tcp_throughput.py
iulianbarbu: Raised the workload runtime to 20 seconds.

tests/integration_tests/performance/test_network_tcp_throughput.py
iulianbarbu: From my experience of running this test, generating gold values is a process which should run the test for many iterations, extracting the minimum and the maximum for each test combination. The gold values contains a target and a delta. The target can be the mean: (MIN + MAX) / 2 and delta: (MAX - MIN) / 2 + a not significant buffer space.

This means that we can not simply overwrite the baselines, running the test once. We do not have an assurance that this process will provide the best resilience to intermittent failures.

I've improved the information presented in the baseline. Please let me know if you still find the baselines hard to read.

dianpopa:
@iulianbarbu is this a duplicate of https://github.com/firecracker-microvm/firecracker/pull/1877? Do we think we can close it?

iulianbarbu:
> Not clear do you also run with default params for window size and packet size? If not please make it clear here with adding something like 0 in this list, otherwise I don't see how you can handle it generically

I test close to default params, but not explicitly the defaults. The default params depend on the configured socket buffer size, on the OS distribution we run. I'll make it more obvious that we test the defaults.

raduiliescu:
> 
> 
> > Not clear do you also run with default params for window size and packet size? If not please make it clear here with adding something like 0 in this list, otherwise I don't see how you can handle it generically
> 
> I test close to default params, but not explicitly the defaults. The default params depend on the configured socket buffer size, on the OS distribution we run. I'll make it more obvious that we test the defaults.

That is not ok. You should run without a strict window size so it can autoadjust according to the TCP algorithms.
Also for the packet size default it will run with maximum one.

tests/integration_tests/performance/network_tcp_throughput_test_config.py
```diff
@@ -0,0 +1,181 @@
...
+        "g2h": [""],
+        "h2g": ["-R"],
+        "bd": ["", "-R"]
+    },
+    "protocols": [
+        {
+            "name": "tcp",
+            "omit": 5,
+            "window_size": [16, None],
+            "pkt_size": [None, 1024],
```
raduiliescu, line 33: I don't get what you did now. Is really important to test with small packets... My suggestion was to keep old values and add a None...

iulianbarbu: The previous packet length was 128K, which is the default packet length used by iperf3 for TCP packets. So there is no change in terms of packet size. The value for 256K related to window size, which I've replaced with `None`, is very similar to the results obtained by using the default window size (or more specifically, as TCP decides). Please let me know if I misunderstood something.

tests/host_tools/network.py
```diff
@@ -38,9 +38,9 @@ def __init__(self, ssh_config):
 
         self._init_connection()
 
-    def execute_command(self, cmd_string):
+    def execute_command(self, cmd_string, with_asyncio=True):
```
serban300, line 41: What's the difference between executing a command with asyncio and without asyncio ? Why do we need this ?

tests/framework/statistics/criteria.py
```diff
@@ -7,6 +7,29 @@
 from abc import ABC, abstractmethod
 
 
+class Result:
```
serban300, line 10: Another option would be to throw an Exception

tests/framework/utils.py
```diff
@@ -359,3 +365,44 @@ def eager_map(func, iterable):
...
+def get_cpu_load(pid: int, iterations=20, omit=0):
+    """Get process threads CPU load."""
+    assert iterations > 0, "Get cpu load iterations must be at least 1."
+
+    if omit > 0:
+        time.sleep(omit)
+
+    cpu_load = dict()
+    for i in range(iterations):
+        cp = subprocess.run(GET_CPU_LOAD.format(pid), shell=True,
```
serban300, line 400: what does `cp` stand for ?

tests/framework/utils.py
```diff
@@ -359,3 +365,44 @@ def eager_map(func, iterable):
...
+        # empty line.
+        if len(lines[-1]) == 0:
+            lines.pop()
+
+        if len(lines) == 0:
+            break
+
+        for line in lines:
+            info = line.strip().split(maxsplit=1, sep=" ")
+            if i == 0:
```
serban300, line 423: I would test if `cpu_load[info[1]]` is not initialized. Otherwise there might be corner cases. For example if we get here before Firecracker fully started, the vcpu threads won't exist, so we won't initialized the list. And after a couple of iterations the vcpu threads will have started.

tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -0,0 +1,357 @@
...
+                         guest_cmd_builder,
+                         current_avail_cpu,
+                         vcpus_count,
+                         runtime,
+                         omit,
+                         load_factor,
+                         modes):
+    """Produce iperf raw output from server-client connection."""
+    # Start the servers.
+    for determ in range(load_factor*vcpus_count):
```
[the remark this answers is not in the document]
serban300: I would rename it to something like `iperf_index` or `host_iperf_index`.

tests/integration_tests/performance/network_tcp_throughput_test_config.py
```diff
@@ -0,0 +1,359 @@
...
+        "g2h": [""],
+        "h2g": ["-R"],
+        "bd": ["", "-R"]
+    },
+    "protocols": [
+        {
+            "name": "tcp",
+            "omit": 5,
+            "window_size": [16, 256, None],
+            "pkt_size": [None, 1024],
```
raduiliescu, line 32: Please add also a small packet size 128 or 96. The 128K is not the packet size, but the payload size that is later segmented, since nobody is sending a 128K packet on the wire being limited first by mss at tcp level and finaly but MTU. The window size looks good now

tests/integration_tests/performance/test_network_tcp_throughput.py
raduiliescu: Please track this remaining requirement

tests/host_tools/network.py
iulianbarbu: The issue is not that the command ran on the same threads. In practice, `run_cmd` will start iperf3 server processes in the background (on the host), on the main thread, and we do not account for them. We account for the guest processes, started with `execute_command`. Those commands have to be tracked in order to get their `stdout`. This has to happen on different threads, while each thread monitors a specific iperf3 client stdout and returns it to the main thread. Then the issue here is that the inner implementation of `execute_command` depends on asyncio runtime, which provides an incompatible interface to be used within threads, because of [1].

[1] https://python.readthedocs.io/en/latest/library/asyncio-dev.html#concurrency-and-multithreading

tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -0,0 +1,367 @@
...
+                         current_avail_cpu,
+                         vcpus_count,
+                         runtime,
+                         omit,
+                         load_factor,
+                         modes):
+    """Produce iperf raw output from server-client connection."""
+    # Start the servers.
+    for determinant in range(load_factor*vcpus_count):
+        assigned_cpu = CpuMap(current_avail_cpu % vcpus_count)
```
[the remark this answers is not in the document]
iulianbarbu: For a `load_factor` of one this will probably assign the same CPU core to a FC thread and to iperf3 server processes as well. I try to avoid this by having this method aware of other processes/threads already pinned to CPU cores, which are not available anymore.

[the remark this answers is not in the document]
serban300: Right, sorry.

iulianbarbu (commented):
I just need to update baselines for the new payload length of 96K. I think I've addressed the rest of your comments. @raduiliescu , @serban300

tests/framework/utils.py
serban300: Hmmm I don't see any change to this `if`. It's still 

```
if i == 0:
    cpu_load[info[1]] = list()
```

tests/host_tools/network.py
serban300: We should understand what causes this behaviour. Why running guest commands with asyncio can lead to deadlocks. I don't think this is the desired behavior.

Your code works around the problem for this specific scenario. But we should make sure we don't encounter this issue in other scenarios as well.

raduiliescu:
Now it looks ok to me.

tests/framework/utils.py
iulianbarbu: Sorry, this comment somehow slipped. Please take a look again. The `get_cpu_percent` primitive is changed a bit.

iulianbarbu:
@serban300 , @raduiliescu , I've integrated your feedback, but found difficulty in updating the baselines - hence the multiple force-pushes done so far. At the moment, I am pleased with the pipeline results, based on the current baselines. We can have a final say after the weekend, by looking at the reliability of the pipeline.

PTAL and let me know if you find things I can improve/fix.

raduiliescu (approved):

serban300 (approved):

raduiliescu (approved):

PR #2223, opened 2020-10-27
Make virtio flush disabled by default

pclesr:
## Reason for This PR

The existing implementation incorrectly advertises VIRTIO_BLK_T_FLUSH, when it
does not actually flush data to media from the host page cache. Unfortunately,
making all devices respect the flush feature bit slows down booting because
on a RW root partition, there can be a number of flush requests during the boot
process.

## Description of Changes

This commit adds a drive option 'want_flush' that allows each
drive to specify whether they require consistency and are willing
to take the performance hit of syncing data out to media. The
default is the existing behaviour, namely no flush.

- [ ] This functionality can be added in [`rust-vmm`](https://github.com/rust-vmm).

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria.]`
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [x] All commits in this PR are signed (`git commit -s`).
- [x] The reason for this PR is clearly provided (issue no. or explanation).
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] Any newly added `unsafe` code is properly documented.
- [x] Any API changes are reflected in `firecracker/swagger.yaml`.
- [ ] Any user-facing changes are mentioned in `CHANGELOG.md`.
- [x] All added/changed functionality is tested.

PR #2229, opened 2020-10-27
metrics: fix bug triggered by memory reordering casualties

dianpopa:
With the passing of the time in Firecracker started
to shape out 2 types of metrics:
* the types of metrics needed for counting purposes -> they
are represented by 2 values, one of them is the active counter while
the other one is the value that was last flushed. With each flush, we would
 compute the delta between them
* the types of metrics needed for storing persistent values of a specific
measurement, for example the ones used for obtaining the process start up
time. This type of metric does not need to keep around a second value
since the value being flushed is always the one the system updates.

This approach also rids us of any memory reordering that may happen on weak
memory models.

## Reason for This PR

The first are of code that needs to check out first thing when starting the review is `src/logger/metrics.rs`.

## Description of Changes

## License Acceptance

## PR Checklist

### DIFF
--- a/src/logger/src/metrics.rs
+++ b/src/logger/src/metrics.rs
@@ -44,9 +44,14 @@
 //! * Since all metrics start at 0, we implement the `Default` trait via derive for all of them,
 //!   to avoid having to initialize everything by hand.
 //!
-//! Moreover, the value of a metric is currently NOT reset to 0 each time it's being written. The
-//! current approach is to store two values (current and previous) and compute the delta between
-//! them each time we do a flush (i.e by serialization). There are a number of advantages
+//! The system implements 2 types of metrics:
+//! * Shared Incremental Metrics (SharedIncMetrics) - dedicated for the metrics which need a counter
+//! (i.e the number of times an API request failed). These metrics are reset upon flush.
+//! * Shared Store Metrics (SharedStoreMetrics) - are targeted at keeping a persistent value, it is not
+//! intended to act as a counter (i.e for measure the process start up time for example).
+//!
+//! The current approach for the `SharedIncMetrics` type is to store two values (current and previous)
+//! and compute the delta between them each time we do a flush (i.e by serialization). There are a number of advantages
 //! to this approach, including:
 //! * We don't have to introduce an additional write (to reset the value) from the thread which
 //!   does to actual writing, so less synchronization effort is required.
@@ -181,11 +186,9 @@ impl fmt::Display for MetricsError {
     }
 }
 
-/// Used for defining new types of metrics that can be either incremented with an unit
-/// or an arbitrary amount of units.
-// This trait helps with writing less code. It has to be in scope (via an use directive) in order
-// for its methods to be available to call on structs that implement it.
-pub trait Metric {
+/// Used for defining new types of metrics that act as a counter (i.e they are continuously updated by
+/// incrementing their value).
+pub trait IncMetric {
     /// Adds `value` to the current counter.
     fn add(&self, value: usize);
     /// Increments by 1 unit the current counter.
@@ -194,6 +197,12 @@ pub trait Metric {
     }
     /// Returns current value of the counter.
     fn count(&self) -> usize;
+}
+
+/// Used for defining new types of metrics that do not need a counter and act as a persistent indicator.
+pub trait StoreMetric {
+    /// Returns current value of the counter.
+    fn fetch(&self) -> usize;
     /// Stores `value` to the current counter.
     fn store(&self, value: usize);
 }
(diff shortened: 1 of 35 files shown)


src/logger/src/metrics.rs
```diff
@@ -223,16 +237,19 @@ impl Metric for SharedMetric {
...
+impl StoreMetric for SharedStoreMetric {
+    fn fetch(&self) -> usize {
+        self.0.load(Ordering::Relaxed)
+    }
 
-    // It is necessary to also reset the old value in order to flush later the correct one,
-    // which is `value` (see the `Serialize` implementation for SharedMetric a few lines below).
     fn store(&self, value: usize) {
         self.0.store(value, Ordering::Relaxed);
-        self.1.store(0, Ordering::Relaxed)
```
acatangiu, line 231: Indeed this was racy between thread doing `store` and thread doing `serialize`, good catch!

acatangiu (commented):
LGTM 👍

src/logger/src/metrics.rs
```diff
@@ -185,7 +190,7 @@ impl fmt::Display for MetricsError {
 /// or an arbitrary amount of units.
 // This trait helps with writing less code. It has to be in scope (via an use directive) in order
```
alindima, line 191: I think we should also update this comment. Currently, it states the general purpose of traits, which is redundant. It should be similar to the comment associated to pub trait `StoreMetric`, stating the use case

src/logger/src/metrics.rs
alindima: hmm, was this racy? I think the atomic type prevented that, right?

src/logger/src/metrics.rs
```diff
@@ -774,16 +796,18 @@ mod tests {
...
+            metric.count(),
             M2_INITIAL_COUNT + NUM_THREADS_TO_SPAWN * NUM_INCREMENTS_PER_THREAD
         );
+    }
 
-        // Trying to store multiple values in the metric will result in keeping only the last
-        // value there.
-        m2.store(1);
-        m2.store(2);
-        assert_eq!(m2.0.load(Ordering::Relaxed), 2);
```
alindima, line 785: why did you remove this small test?

alindima (commented):
Looks good. Only a couple of nits/questions

src/logger/src/metrics.rs
alindima: makes total sense, thanks

src/logger/src/metrics.rs
alindima: given the new implementation, I now see that this test doesn't make sense any more

src/logger/src/metrics.rs
```diff
@@ -181,11 +186,9 @@ impl fmt::Display for MetricsError {
     }
 }
 
-/// Used for defining new types of metrics that can be either incremented with an unit
-/// or an arbitrary amount of units.
-// This trait helps with writing less code. It has to be in scope (via an use directive) in order
-// for its methods to be available to call on structs that implement it.
-pub trait Metric {
+/// Used for defining new types of metrics that as a counter (i.e they are continously updated by
```
alindima, line 189: nit:

```suggestion
/// Used for defining new types of metrics that act as a counter (i.e they are continously updated by
```

dianpopa: Done!

alindima (approved):

tests/integration_tests/build/test_binary_size.py
```diff
@@ -13,9 +13,9 @@
 
 SIZES_DICT = {
     "x86_64": {
-        "FC_BINARY_SIZE_TARGET": 1955168,
+        "FC_BINARY_SIZE_TARGET": 2059824,
         "JAILER_BINARY_SIZE_TARGET": 1439512,
-        "FC_BINARY_SIZE_LIMIT": 2052927,
+        "FC_BINARY_SIZE_LIMIT": 2100000,
```
raduiliescu, line 18: Do we understand this increase? Can you detail it? just looking on the code a 2% increase doesn't make sense to me

alindima: Given the delta of 5%, the increase is not based only on this PR. The size has likely increased over time, and this PR was unlucky enough to go overboard the 5%

alindima: I saw this exact same issue on this PR #2223 , for ARM

raduiliescu: Ok.

raduiliescu (approved):

PR #2231, opened 2020-10-27
Dirty page tracking

luminitavoicu:
## Reason for This PR
Integrated local vm-memory crate into Firecracker and updated current dirty page tracking functionality with Firecracker dirty bitmaps. - [#1997](https://github.com/firecracker-microvm/firecracker/issues/1997)

## Description of Changes

- Added a reset functionality to region bitmaps.
- Propagated dirty page tracking options through Firecracker.
- Updated dirty dumps to include the Firecracker bitmaps into the decision process.
- Re-enabled diff snapshot support.
- Added unit tests and updated current ones.
- Improved integration tests. 

## License Acceptance

## PR Checklist


src/vm-memory/src/bitmap.rs
```diff
@@ -78,6 +78,13 @@ impl Bitmap {
     pub fn is_empty(&self) -> bool {
         false
     }
+
+    /// Reset all bitmap bits to 0.
+    pub fn reset(&self) {
+        for n in 0..self.size {
+            self.map[n >> 6].fetch_and(0 << (n & 63), Ordering::SeqCst);
+        }
```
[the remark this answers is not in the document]
blitz: Using `Release` ordering would change the semantics and feels like premature optimization. If weakening the memory consistency is necessary, it should be documented why it is safe.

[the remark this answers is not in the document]
acatangiu: Since the operation also changes from `fetch_and` (read + update + write) to a simple `store` (write), is `SeqCst` still necessary?

The `Acquire` part is no longer necessary right? So that only leaves asking ourselves:
 - do we need all threads to see these operations sequentially consistent? Or is sequential ordering not relevant as long as update atomicity is upheld?

For the current Firecracker snapshot use-case, higher level program logic guarantees exclusive access to the emulation thread to this bitmap so I believe Ordering constraints are irrelevant.

For a possibly future live-snapshotting use-case where we'd keep vCPUs active, this becomes important, but still _strict sequential ordering_ doesn't look necessary, right?

For other use-cases I'm having a failure of imagination to come up with a scenario where sequential consistency is necessary when clearing this bitmap.

Disclaimer: I'm fine with keeping `Ordering::SeqCst` here since it's not in any hot-path whatsoever, but I wanted to take this opportunity to learn more about these ordering requirements and how to best use them.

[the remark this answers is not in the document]
sandreim: I think this is very relevant for usecases such as migrating a microvm. I would go for `store(0, Ordering::Release);`

sandreim (commented):
PR looks good, but I feel that we are missing an important integration test: pause vm/create full/create diff and then assert on sparse === diff file.

src/vm-memory/src/mmap.rs
```diff
@@ -460,6 +460,56 @@ impl GuestMemoryMmap {
...
+            count,
+            addr,
+            |offset,
+             len,
+             caddr,
+             region|
+             -> result::Result<usize, vm_memory_upstream::guest_memory::Error> {
+                // Check if something bad happened before doing unsafe things.
+                assert!(offset <= count);
+                let len = std::cmp::min(len, 4096);
```
serban300, line 483: Why do we do this ?

luminitavoicu: We don't. I will remove this.

src/vm-memory/src/mmap.rs
```diff
@@ -460,6 +460,56 @@ impl GuestMemoryMmap {
...
+                let mut buf = vec![0u8; len].into_boxed_slice();
+                let bytes_read = src
+                    .read(&mut buf[..])
+                    .map_err(vm_memory_upstream::guest_memory::Error::IOError)?;
```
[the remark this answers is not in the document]
luminitavoicu: You are right! I've made the update.

src/vm-memory/src/bitmap.rs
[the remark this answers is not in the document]
luminitavoicu: Done

luminitavoicu:
> PR looks good, but I feel that we are missing an important integration test: pause vm/create full/create diff and then assert on sparse === diff file.

Test added.

src/vm-memory/src/mmap.rs
serban300: More precisely, I meant why do we need this custom logic ? I see that the buffer len is equal to the page len . But why ? Can't we use the default GuestMemory implementation of this function ? I think the dirty tracking should work correctly anyway.

luminitavoicu: This custom logic is a part of the upstream implementation. The `read_from` function in upstream has two workflow branches: one for the regions which have `as_mut_slice` implemented and another one otherwise. 
For the second case, the content of the source (the object we want to read from) is read into a buffer and then the buffer is written in the region (which is what we want, because the region `write` function owns logic to mark the pages dirty).  
For the first case, the upstream uses a slice of the region and writes there directly from source by calling `read` on the object. 

Because of vsock, our local region has an implementation of `as_mut_slice`, so it matches this first case. Therefore, by using the upstream implementation, we would not be able to mark pages as dirty. To solve this issue, I brought locally the part we need from upstream.

serban300: Ok, I understand the issue now. Thanks for the explanation.

src/vm-memory/src/bitmap.rs
```diff
@@ -78,6 +78,13 @@ impl Bitmap {
     pub fn is_empty(&self) -> bool {
         false
     }
+
+    /// Reset all bitmap bits to 0.
+    pub fn reset(&self) {
+        for bit in self.map.iter() {
+            bit.store(0, Ordering::Release);
```
acatangiu, line 85: nit: rename `bit` to `element` or `it` or smth else since each loop we reset 64bits, not just one. The current name is misleading.

src/vm-memory/src/mmap.rs
```diff
@@ -460,6 +460,63 @@ impl GuestMemoryMmap {
...
+            |_offset,
+             len,
+             caddr,
+             region|
+             -> result::Result<usize, vm_memory_upstream::guest_memory::Error> {
+                let page_size = match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
+                    -1 => panic!("Failed to read: {}", errno::Error::last()),
+                    ps => ps as usize,
+                };
+                let len = std::cmp::min(len, page_size);
```
acatangiu, line 485: Calling `libc::sysconf()` on each of these calls can be avoided by caching the page_size in a `lazy_static` or some other global const.

But looking closer I think page-size has nothing to do with this part anyway.

The enforced `len` of max 4096 is there to split a large read/write to multiple underlying calls where each call does at most 4096 bytes using a temporary buffer.

I believe this is intended to avoid temp buffer allocations of unbounded (or at least very big sizes).

For example when reading a 1GB kernel file to guest memory, we don't want to allocate 1GB temporary vec. Instead we read 4k bytes at a time.

So the old code `let len = std::cmp::min(len, 4096)` was correct.
But pls use a global const to define 4096 and add a comment around its usage logic.

acatangiu (changes requested):

luminitavoicu:
Comments addressed. @acatangiu, please take another look.

acatangiu (dismissed):

serban300 (approved):

acatangiu:
@luminitavoicu you need to cherry-pick this commit from master branch:

```
    CI: move build independent tests to a folder
    
    Some of our integration tests can very well be run on a
    single platform.
    In order to do that all the build independent tests were
    moved to a separate folder and conftest was adjusted to ignore
    this folder on other platforms different than AMD.
```

Unfortunately we don't have per-branch CI params to define branch _capabilities_ or similar.

serban300 (approved):

acatangiu (approved):

PR #2240, opened 2020-10-28
[Balloon] Added GET call for balloon device state

georgepisaltu:
## Reason for This PR

We need an API call to poll the state of the balloon device.

## Description of Changes

Added an API call to poll the state of the balloon device.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/api_server/swagger/firecracker.yaml
+++ b/src/api_server/swagger/firecracker.yaml
@@ -60,6 +60,115 @@ paths:
           schema:
             $ref: "#/definitions/Error"
 
+  /balloon:
+    get:
+      summary: Returns the current balloon device configuration.
+      operationId: describeBalloonConfig
+      responses:
+        200:
+          description: The balloon device configuration
+          schema:
+            $ref: "#/definitions/Balloon"
+        400:
+          description: Balloon device not configured.
+          schema:
+            $ref: "#/definitions/Error"
+        default:
+          description: Internal Server Error
+          schema:
+            $ref: "#/definitions/Error"
+    put:
+      summary: Creates or updates a balloon device.
+      description:
+        Creates a new balloon device if one does not already exist, otherwise updates it, before machine startup.
+        This will fail after machine startup.
+        Will fail if update is not possible.
+      operationId: putBalloon
+      parameters:
+      - name: body
+        in: body
+        description: Balloon properties
+        required: true
+        schema:
+          $ref: "#/definitions/Balloon"
+      responses:
+        204:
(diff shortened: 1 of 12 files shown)


src/devices/src/virtio/balloon/device.rs
```diff
@@ -73,6 +73,15 @@ struct BalloonStat {
 // Safe because BalloonStat only contains plain data.
 unsafe impl ByteValued for BalloonStat {}
 
+// BalloonStats holds statistics returned from the stats_queue.
+#[derive(Clone, Default, Debug, PartialEq, Serialize)]
+pub struct BalloonState {
```
acatangiu, line 78: Suggest `BalloonConfig` or some other name; `BalloonState` misleads the reader to think of snapshots 😛

acatangiu (commented):
LGTM, just needs changing `BalloonState` to `BalloonConfig` or smth similar.

src/api_server/swagger/firecracker.yaml
```diff
@@ -60,6 +60,115 @@ paths:
...
+        default:
+          description: Internal Server Error
+          schema:
+            $ref: "#/definitions/Error"
+    patch:
+      summary: Updates a balloon device statistics polling interval.
+      description:
+        Updates an existing balloon device statistics interval, before or after machine startup.
+        Will fail if update is not possible.
+      operationId: patchStatsBalloon
```
sandreim, line 152: This sounds like we want to update the stats. I would rename this as patchBalloonStatsInterval to make it more specific.

sandreim (commented):
Changes look good. I have only one question around how the APIs would work when guest driver does not negotiate VIRTIO_BALLOON_F_STATS_VQ.

georgepisaltu:
> Changes look good. I have only one question around how the APIs would work when guest driver does not negotiate VIRTIO_BALLOON_F_STATS_VQ.

The `GET /balloon` will be always available and will return the configuration of the device, with the flags at install time and the current target size.

The `GET /balloon/statistics` will be available only after boot and will return only the statistics, as described in the stats structure. `VIRTIO_BALLOON_F_STATS_VQ` can only be negotiated at boot time, and we only do it when the `stats_polling_interval_s` is set to a non-zero value before boot. Therefore, this request will only be available post-boot if the polling interval was set to a non-zero value pre-boot.

sandreim (dismissed):

sandreim (approved):

acatangiu (approved):

Issue #2063, opened 2020-08-04
Add API to retrieve supported snapshot versions.


sandreim:

GeorgeLS:
Hello, I'm keen in systems programming and I would like to contribute to Firecracker.
Also, this is my first try in open source.
I read the documentation and the code about versioning control in firecracker but it's not clear to me what we would like to achieve here.
Is there a defined set of versions that Firecracker supports or we would like to persist all the snapshot versions that the user has created so far?
For the second one I would imagine the API something like this:

```
curl --unix-socket $socket -i \
    -X GET 'http://localhost/snapshot' \
    -H 'Accept: application/json' \
    -H 'Content-Type: application/json'
```

and the response being (in JSON):

```
{
    [
        {
            "snapshot_path": "/full/path/to/snapshot_file",
            "version": "0.23.0"
         },
        {
            "snapshot_path": "/full/path/to/snapshot_file",
            "version": "0.30.0"
        },
        ...
    ]
}
```

Please give me your thoughts/ideas on the matter so we can discuss it and potentially I could start working on it.

Thank you in advance,
George

luminitavoicu:
Hi @GeorgeLS! 

We greatly appreciate your initiative to contribute to Firecracker. 
Firecracker has an internal design to map the release version to the specific snapshot data format versions. What we were aiming for here, was to be able to fetch a list of the supported snapshot versions. We have not decided yet on the most suitable mechanism for this (API or command line parameter). 
This issue has already been picked up by a member of the Firecracker team and a RFC will be up soon, so you can check the progress there if you are interested in keeping up to date with this.

Thank you again,
Luminita

dianpopa:
Functionality was implemented by this [PR](https://github.com/firecracker-microvm/firecracker/pull/2247) by extending the "--version" command line parameter to also show supported snapshot versions.

PR #2247, opened 2020-10-30
Added command line parameter to retrieve supported snapshot versions

luminitavoicu:

## Reason for This PR

This is a draft PR meant to solve [#2063](https://github.com/firecracker-microvm/firecracker/issues/2063).

## Description of Changes

An added functionality to display an ordered list of the current supported snapshot data format versions to the `--version` command line parameter. 
The updated `./firecracker --version` command will print the information in the format below:

> Firecracker v0.x.0
> 
> Supported snapshot data format versions: v0.y.0, v0.z.0

## License Acceptance

## PR Checklist

### DIFF
--- a/src/firecracker/src/main.rs
+++ b/src/firecracker/src/main.rs
@@ -19,6 +19,7 @@ use utils::validators::validate_instance_id;
 use vmm::default_syscalls::get_seccomp_filter;
 use vmm::resources::VmResources;
 use vmm::signal_handler::register_signal_handlers;
+use vmm::version_map::FC_VERSION_TO_SNAP_VERSION;
 use vmm::vmm_config::instance_info::InstanceInfo;
 use vmm::vmm_config::logger::{init_logger, LoggerConfig, LoggerLevel};
 
@@ -140,7 +141,7 @@ fn main() {
         .arg(
             Argument::new("version")
                 .takes_value(false)
-                .help("Print the binary version number.")
+                .help("Print the binary version number and a list of supported snapshot data format versions.")
         );
 
     let arguments = match arg_parser.parse_from_cmdline() {
@@ -164,6 +165,7 @@ fn main() {
             if let Some(version) = arg_parser.arguments().value_as_bool("version") {
                 if version {
                     println!("Firecracker v{}\n", FIRECRACKER_VERSION);
+                    print_supported_snapshot_versions();
                     process::exit(i32::from(vmm::FC_EXIT_CODE_OK));
                 }
             }
@@ -258,6 +260,22 @@ fn main() {
             boot_timer_enabled,
         );
     }
+}
+
+// Print supported snapshot data format versions.
+fn print_supported_snapshot_versions() {
+    let mut snapshot_versions_str = "Supported snapshot data format versions:".to_string();
+    let mut snapshot_versions: Vec<String> = FC_VERSION_TO_SNAP_VERSION
+        .iter()
+        .map(|(key, _)| key.clone())
+        .collect();
+    snapshot_versions.sort();
+
+    snapshot_versions
+        .iter()
+        .for_each(|v| snapshot_versions_str.push_str(format!(" v{},", v).as_str()));
+    snapshot_versions_str.pop();
+    println!("{}\n", snapshot_versions_str);
 }
 
 // Configure and start a microVM as described by the command-line JSON.
--- a/src/firecracker/src/main.rs
+++ b/src/firecracker/src/main.rs
@@ -136,6 +136,11 @@ fn main() {
             Argument::new("boot-timer")
                 .takes_value(false)
                 .help("Whether or not to load boot timer device for logging elapsed time since InstanceStart command.")
+        )
+        .arg(
+            Argument::new("version")
+                .takes_value(false)
+                .help("Print the binary version number.")
         );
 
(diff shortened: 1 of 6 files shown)


src/utils/src/arg_parser.rs
```diff
@@ -11,6 +11,7 @@ pub type Result<T> = result::Result<T, Error>;
 const ARG_PREFIX: &str = "--";
 const ARG_SEPARATOR: &str = "--";
 const HELP_ARG: &str = "--help";
+const SNAPSHOT_VERSION_ARG: &str = "--snapshot-version";
```
sandreim, line 14: I would prefer to output the snapshot version information on the second line after the Firecracker version. No need for additional CLI parameter.

src/firecracker/src/main.rs
```diff
@@ -163,6 +164,23 @@ fn main() {
...
+                if snapshot_version {
+                    println!("Supported snapshot data format versions:");
+
+                    let mut snapshot_versions: Vec<String> = FC_VERSION_TO_SNAP_VERSION
+                        .iter()
+                        .map(|(key, _)| key.clone())
+                        .collect();
+                    snapshot_versions.sort();
+                    for version in snapshot_versions {
+                        println!("v{}", version);
```
sandreim, line 178: I would output these on a single line: 
`Supported snapshot versions: v0.23.0, v0.24.0`

sandreim (commented):
Added some comments.
Overall I like the approach with the command line parameter instead of an API call.
I am curious how @firecracker-microvm/compute-capsule and other folks think about this.

raduweiss:
My 2c: a command line parameters is required, since this attribute is about the capabilities of the program (which versions can it work with?), and we should not _force_ users to interact with the API server to get that info. We should work backwards from what users want with regard to _also_ adding an API here.

src/firecracker/src/main.rs
luminitavoicu: Done

src/utils/src/arg_parser.rs
luminitavoicu: Ack

acatangiu (commented):
Please add a sample output to the PR description to provide a visual representation of ur changes.
Otherwise PR LGTM.

src/firecracker/src/main.rs
```diff
@@ -159,6 +160,20 @@ fn main() {
             if let Some(version) = arg_parser.arguments().value_as_bool("version") {
                 if version {
                     println!("Firecracker v{}\n", FIRECRACKER_VERSION);
+
+                    print!("Supported snapshot data format versions:");
```
alindima, line 164: nit: I think it would be clearer to have a separate function that prints the snapshot versions

src/utils/src/arg_parser.rs
```diff
@@ -362,7 +362,10 @@ impl<'a> Arguments<'a> {
         // command line arguments by adding just the version argument to the parsed list and
         // returning.
         if args.contains(&VERSION_ARG.to_string()) {
-            let mut version_arg = Argument::new("version").help("Print the binary version number.");
+            let mut version_arg = Argument::new("version").help(
```
alindima, line 365: the `arg_parser` is more general-purpose and is used by multiple binaries in this workspace, that don't have any support for snapshots - jailer and more recently, seccompiler.

We should find another place for this

alindima (changes requested):

src/firecracker/src/main.rs
luminitavoicu: Done

src/firecracker/src/main.rs
```diff
@@ -255,6 +262,22 @@ fn main() {
...
+        .iter()
+        .map(|(key, _)| key.clone())
+        .collect();
+    snapshot_versions.sort();
+
+    let len = snapshot_versions.len();
+    for version in snapshot_versions.iter().take(len - 1) {
+        print!(" v{},", version);
+    }
+    println!(" v{}\n", snapshot_versions[len - 1]);
```
alindima, line 278: If I understand correctly, the only reason this `for` loop takes the first `n-1` elements is because we want to have a newline after the last printed version. 

Wouldn't this be simpler?

```suggestion
    snapshot_versions.sort().iter().for_each(|version| print!(" v{},", version));
    println!("\n");
```

alindima (commented):
The PR also needs a CHANGELOG.md update

src/firecracker/src/main.rs
luminitavoicu: I also wanted to avoid having a trailing comma at the end of the versions list.

alindima: I see. In this case we could build a string and pop off the last element right before outputting. Anyway, this is not a huge deal, but I think that it would make for more readable code. it's up to you

acatangiu (commented):
Changes LGTM, please add a Changelog entry as well.

luminitavoicu:
Comments addressed. @acatangiu and @alindima, please take another look.

src/firecracker/src/main.rs
luminitavoicu: Comment addressed.

alindima (approved):
LGTM

acatangiu (approved):

PR #2258, opened 2020-11-05
Fix reinflate test intermittent fail

### LINKED ISSUE #2257: [Bug] test_reinflate_balloon - intermittent failure on aarch64
## Describe the bug

Unstable test:

```
test_reinflate_balloon[debian_with_ssh_and_balloon]
   integration_tests/functional/test_balloon.py:330: in test_reinflate_balloon
      assert (third_reading - first_reading) <= 20000 E assert (154396 - 122916) <= 20000
```

## To Reproduce

Reproduced it once on: https://github.com/firecracker-microvm/firecracker/pull/2226, for aarch64.

## Expected behaviour

The test should always pass.

## Environment

CI for a PR.

## Additional context

N/A

## Checks
~~- [] Are you certain the bug being reported is a Firecracker issue?~~

georgepisaltu:
## Reason for This PR

The intermittent fail in `test_reinflate_balloon` has resurfaced and
this commit introduces the (hopefully) final fix for this test.

## Description of Changes

Fixes #2257 

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/integration_tests/functional/test_balloon.py
+++ b/tests/integration_tests/functional/test_balloon.py
@@ -298,6 +298,16 @@ def test_reinflate_balloon(test_microvm_with_ssh_and_balloon, network_config):
     firecracker_pid = test_microvm.jailer_clone_pid
     ssh_connection = net_tools.SSHConnection(test_microvm.ssh_config)
 
+    # First inflate the balloon to free up the uncertain amount of memory
+    # used by the kernel at boot and establish a baseline, then give back
+    # the memory.
+    response = test_microvm.balloon.patch(amount_mb=200)
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+    time.sleep(5)
+    response = test_microvm.balloon.patch(amount_mb=0)
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+    time.sleep(2)
+
     # Get the guest to dirty memory.
     make_guest_dirty_memory(ssh_connection)
     first_reading = get_rss_mem_by_pid(firecracker_pid)


tests/integration_tests/functional/test_balloon.py
```diff
@@ -298,6 +298,16 @@ def test_reinflate_balloon(test_microvm_with_ssh_and_balloon, network_config):
     firecracker_pid = test_microvm.jailer_clone_pid
     ssh_connection = net_tools.SSHConnection(test_microvm.ssh_config)
 
+    # First inflate the balloon to free up the uncertain amount of memory
+    # used by the kernel at boot and establish a baseline, then give back
+    # the memory.
+    response = test_microvm.balloon.patch(amount_mb=200)
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+    time.sleep(5)
```
gc-plp, line 306: Sleeping for an amount of time doesn't really guarantee that this test won't fail anymore. It also adds extra time to the test suite.

Could we instead check in a loop that the requested amount has been freed? If we exceed a large enough timeout we can fail.

The key is that instead of relying on a fixed amount of time to hopefully free the requested amount of memory, we sometimes may free it quicker than the 5 seconds and some other times need 6 seconds. If it doesn't happen within a reasonable time frame, then something is wrong and we would need to fail.

iulianbarbu, line 306: After the offline discussion that we had, I think these wait times need to be addressed in a generic way. As you said in the corresponding issue, we need a generic mechanism for Firecracker process RSS monitoring. Should we track this separately in a new issue?

georgepisaltu: The sleep time was one thing that previously caused intermittent failures. Indeed, if the sleep time was too low it wouldn't have time to free the memory, but 5 seconds for 200 MiB is sufficient in all cases. The real fix is establishing the baseline before running the test by constricting the VM to 56 MiB,  releasing the memory, and then start actually measuring the RSS.

I agree that the sleep usage is not ideal, but this is not the only place where it's used. To get rid of the sleeps, we need to perform the checks via a polling mechanism (similar to what you described) and that is a separate issue in my opinion.

georgepisaltu: Yes, I'll log a separate issue as soon as this is merged.

gc-plp: > that is a separate issue in my opinion

I don't feel strongly towards doing the fix in one change or two, but it sounds like we're introducing something now which we will remove later in order to fix the same underlying problem.

georgepisaltu: Again, all the balloon tests currently use `sleep` to wait for inflate/deflate completion. We should indeed introduce a mechanism more efficient than the `sleep` one in the future and it should replace the `sleep`s in the other tests as well, but for the purpose of fixing the intermittent failure, this fix doesn't introduce anything more than an extra inflate/deflate sequence.

acatangiu: At least open an issue to address all of these `sleep`s. We can pick it up later, it'd also be a good ramp-up issue.

georgepisaltu: Created the related issue separately.

#2261

gc-plp: Let's merge this for now as is. As per the discussion in #2257, the question that I rased above will be picked up by someone else.

gc-plp (approved):

iulianbarbu (approved):

PR #2264, opened 2020-11-09
Avoid hard-coded sleep() calls in balloon driver functional tests

pkj415:
Currently, tests in test_balloon.py use sleep() calls with hard coded values to
wait until the RSS of the microvm reaches a stable value for reading.

The behavior is changed and the test now waits until the percentage change in
RSS falls under a specified threshold. Right now 0.5% is used as the threshold.

## Reason for This PR

Tests in test_balloon.py use hard-coded sleep() calls. Fixing that to make tests more robust.
The hard-coded sleep() calls results in failure of tests on aarch64.

## Description of Changes

The hard-coded sleep calls are removed and the test now waits until the percentage change in
RSS falls under a specified threshold. Right now 0.5% is used as the threshold.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/integration_tests/functional/test_balloon.py
+++ b/tests/integration_tests/functional/test_balloon.py
@@ -9,12 +9,13 @@ import subprocess
 import time
 
 import pytest
+from retry import retry
 
 from conftest import _test_images_s3_bucket
 from framework.artifacts import ArtifactCollection, ArtifactSet
 from framework.builder import MicrovmBuilder, SnapshotBuilder, SnapshotType
 from framework.matrix import TestMatrix, TestContext
-from framework.utils import get_free_mem_ssh
+from framework.utils import get_free_mem_ssh, run_cmd
 
 import host_tools.network as net_tools  # pylint: disable=import-error
 
@@ -22,10 +23,27 @@ import host_tools.network as net_tools  # pylint: disable=import-error
 MB_TO_PAGES = 256
 
 
-def get_rss_mem_by_pid(pid):
-    """Get the RSS memory that a guest uses, given the pid of the guest."""
-    output = subprocess.check_output("pmap -X {}".format(pid), shell=True)
-    return int(output.decode('utf-8').split('\n')[-2].split()[1], 10)
+@retry(delay=0.5, tries=10)
+def get_stable_rss_mem_by_pid(pid, percentage_delta=0.5):
+    """
+    Get the RSS memory that a guest uses, given the pid of the guest.
+
+    Wait till the fluctuations in RSS drop below percentage_delta. If timeout
+    is reached before the fluctuations drop, raise an exception.
+    """
+
+    def get_rss_from_pmap():
+        _, output, _ = run_cmd("pmap -X {}".format(pid))
+        return int(output.split('\n')[-2].split()[1], 10)
+
+    first_rss = get_rss_from_pmap()
+    time.sleep(1)
+    second_rss = get_rss_from_pmap()
+
+    delta = (abs(first_rss - second_rss)/float(first_rss)) * 100
+    assert delta < percentage_delta
+
+    return second_rss
 
 
 def make_guest_dirty_memory(ssh_connection, should_oom=False, amount=8192):
@@ -127,15 +145,15 @@ def test_rss_memory_lower(test_microvm_with_ssh_and_balloon, network_config):
     # Using deflate_on_oom, get the RSS as low as possible
     response = test_microvm.balloon.patch(amount_mb=200)
     assert test_microvm.api_session.is_status_no_content(response.status_code)
-    time.sleep(5)
 
     # Get initial rss consumption.
-    init_rss = get_rss_mem_by_pid(firecracker_pid)
+    init_rss = get_stable_rss_mem_by_pid(firecracker_pid)
 
     # Get the balloon back to 0.
     response = test_microvm.balloon.patch(amount_mb=0)
     assert test_microvm.api_session.is_status_no_content(response.status_code)
-    time.sleep(2)
+    # This call will internally wait for rss to become stable.
+    _ = get_stable_rss_mem_by_pid(firecracker_pid)
 
     # Dirty memory, then inflate balloon and get ballooned rss consumption.
     make_guest_dirty_memory(ssh_connection)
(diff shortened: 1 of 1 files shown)


tests/integration_tests/functional/test_balloon.py
```diff
@@ -22,10 +22,22 @@
...
 
 
-def get_rss_mem_by_pid(pid):
+def get_stable_rss_mem_by_pid(pid):
     """Get the RSS memory that a guest uses, given the pid of the guest."""
-    output = subprocess.check_output("pmap -X {}".format(pid), shell=True)
-    return int(output.decode('utf-8').split('\n')[-2].split()[1], 10)
+
+    def get_rss_from_pmap():
+        output = subprocess.check_output("pmap -X {}".format(pid), shell=True)
```
gc-plp, line 29: For logging purposes, you can use `utils.run_cmd()`.

It's also helpful in avoiding `decode` calls.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -22,10 +22,22 @@
...
+        return int(output.decode('utf-8').split('\n')[-2].split()[1], 10)
+
+    rss = get_rss_from_pmap()
+    percentage_delta = 100
+    while percentage_delta > 0.5:
+        prev_rss = rss
+        rss = get_rss_from_pmap()
+        time.sleep(1)
+        percentage_delta = (abs(prev_rss - rss)/float(prev_rss)) * 100
```
gc-plp, line 38: What if `percentage_delta` never drops below `0.5`?

`get_stable_rss_mem_by_pid` should have a couple of extra default parameters: `delta` and `timeout`: `def get_stable_rss_mem_by_pid(pid, delta=0.5, timeout=30)`

tests/integration_tests/functional/test_balloon.py
```diff
@@ -22,10 +22,22 @@
...
+
+    def get_rss_from_pmap():
+        output = subprocess.check_output("pmap -X {}".format(pid), shell=True)
+        return int(output.decode('utf-8').split('\n')[-2].split()[1], 10)
+
+    rss = get_rss_from_pmap()
+    percentage_delta = 100
+    while percentage_delta > 0.5:
+        prev_rss = rss
+        rss = get_rss_from_pmap()
```
gc-plp, line 36: There is a chance that both `pmem` calls return the same value (or similar values) and the while loop exits early. I think that the `time.sleep(1)` command should be executed before the second `pmem` call.

tests/integration_tests/functional/test_balloon.py
pkj415: Updated it

tests/integration_tests/functional/test_balloon.py
pkj415: Oh yes, nice catch, thanks!

tests/integration_tests/functional/test_balloon.py
pkj415: Yes, added timeout and delta as parameters to the function. Also, I am raising an Exception in case of timeout. Let me know if that looks okay

pkj415:
Thanks a lot for the comments @gc-plp ! I have addressed them, will wait for further comments.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -14,18 +14,40 @@
...
+    delta = 100
+    while delta > percentage_delta and timeout_secs > 0:
+        prev_rss = rss
+        time.sleep(1)
+
+        rss = get_rss_from_pmap()
+        timeout_secs -= 1
+        delta = (abs(prev_rss - rss)/float(prev_rss)) * 100
+
+    if timeout_secs == 0:
```
gc-plp, line 47: The `if` condition should be in the `while` loop.

pkj415: It doesn't have to be right? It will exit the while loop if timeout_secs reaches 0. Let me know or maybe I am missing something?

gc-plp: Ah, sorry for the vague comment there :)

You can simplify the while loop condition and check `timeout_secs` only once inside the loop.

pkj415: Oh yes, that's a good idea, I didn't catch that from the comment, updated it.

gc-plp:
@pkj415 the PR looks good. Feel free to mark it as ready for review when ready.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -14,18 +14,41 @@
...
+
+        rss = get_rss_from_pmap()
+        timeout_secs -= 1
+        if timeout_secs == 0:
+            raise Exception(
+                "Timed out while reading RSS, it did not stabilize")
+
+        delta = (abs(prev_rss - rss)/float(prev_rss)) * 100
+
+    return rss
```
dianpopa, line 51: You can replace the while with a [retry decorator](https://pypi.org/project/retry/).
```suggestion
@retry(delay=0.5, tries=10)
def get_stable_rss_mem_by_pid(pid, percentage_delta=0.5):
    """
    Get the RSS memory that a guest uses, given the pid of the guest.

    Wait till the fluctuations in RSS drop below percentage_delta. If timeout
    is reached before the fluctuations drop, raise an exception.
    """

    def get_rss_from_pmap():
        _, output, _ = run_cmd("pmap -X {}".format(pid))
        return int(output.split('\n')[-2].split()[1], 10)

    first_rss = get_rss_from_pmap()
    time.sleep(1)
    second_rss = get_rss_from_pmap()
    delta = (abs(first_rss - second_rss)/float(first_rss)) * 100
    assert delta < percentage_delta
    return second_rss
```

and import the retry framework with "from retry import retry".

dianpopa (commented):
Please change the commit title to something that follows the 50/72 commit rule. I suggest something of the kind:
`CI:  refactor sleep calls in balloon tests`

tests/integration_tests/functional/test_balloon.py
pkj415: Nice suggestion, updated the PR

dianpopa (approved):

gc-plp (approved):

Issue #2116, opened 2020-09-04
[Bug] Firecracker exits before returning API response when loading an invalid snapshot.


sandreim:
## Describe the bug

Due to the current design, any failure on the vCPU resume path will cause Firecracker to exit before sending a response.

## To Reproduce

1. Create a snapshot that contains a corrupted vCPU state. 
2. Load the snapshot.

## Expected behaviour

Firecracker should exit after returning API response.

## Environment

- Firecracker version: v0.23.

## Checks
- [x] Have you searched the Firecracker Issues database for similar problems?
- [x] Have you read the existing relevant Firecracker documentation?
- [x] Are you certain the bug being reported is a Firecracker issue?

berciuliviu:
Fixed by #2282

PR #2282, opened 2020-11-18
Firecracker exits before returning API response FIX

berciuliviu:
## Reason for This PR

This PR addresses the following [bug](https://github.com/firecracker-microvm/firecracker/issues/2116). For now, the status is RFC since integration tests have to be added.

## Description of Changes

The important points of this change are the following:

- the VCPU errors that killed the process before, are now allowed to return to the API
thread.
- the API thread decides if an error is considered **fatal** or not. New errors can be added in the future which are deemed fatal
- in case of a **fatal** error, the Firecracker process is killed by the API thread,
since the VMM thread is already compromised
- this mechanism abides by the rules of the [Snapshotting specification](https://github.com/firecracker-microvm/firecracker/blob/master/docs/snapshotting/snapshot-support.md) which says that,
in the case of an error when _loading a snapshot_, _"a specific error is reported and then the current Firecracker process is ended"_

## License Acceptance

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Added devtool test `-m|--cpuset-mems` flag for memory confinement when tests
   run.
 - Added the virtio traditional memory ballooning device.
+- Added a mechanism to handle vCPU/VMM errors that result in process termination.
 
 ### Changed
 
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -25,6 +25,8 @@ use vmm::vmm_config::instance_info::InstanceInfo;
 #[cfg(target_arch = "x86_64")]
 use vmm::vmm_config::snapshot::SnapshotType;
 
+use vmm::FC_EXIT_CODE_BAD_CONFIGURATION;
+
 /// Shorthand type for a request containing a boxed VmmAction.
 pub type ApiRequest = Box<VmmAction>;
 /// Shorthand type for a response containing a boxed Result.
@@ -67,6 +69,9 @@ pub struct ApiServer {
     /// FD on which we notify the VMM that we have sent at least one
     /// `VmmRequest`.
     to_vmm_fd: EventFd,
+    /// If this flag is set, the process encountered a fatal error
+    /// and it is going to exit once it sends any pending API response.
+    vmm_fatal_error: bool,
 }
 
 impl ApiServer {
@@ -83,6 +88,7 @@ impl ApiServer {
             api_request_sender,
             vmm_response_receiver,
             to_vmm_fd,
+            vmm_fatal_error: false,
         })
     }
 
@@ -146,6 +152,18 @@ impl ApiServer {
                         let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic)
                             - request_processing_start_us;
                         debug!("Total previous API call duration: {} us.", delta_us);
+                        if self.vmm_fatal_error {
+                            // Flush the remaining outgoing responses
+                            // and proceed to exit
+                            server.flush_outgoing_writes();
+                            error!(
+                                "Fatal error with exit code: {}",
+                                FC_EXIT_CODE_BAD_CONFIGURATION
+                            );
+                            unsafe {
+                                libc::_exit(i32::from(FC_EXIT_CODE_BAD_CONFIGURATION));
+                            }
+                        }
                     }
                 }
                 Err(e) => {
@@ -158,7 +176,11 @@ impl ApiServer {
         }
     }
 
-    pub fn handle_request(&self, request: &Request, request_processing_start_us: u64) -> Response {
+    pub fn handle_request(
+        &mut self,
+        request: &Request,
+        request_processing_start_us: u64,
+    ) -> Response {
         match ParsedRequest::try_from_request(request) {
             Ok(ParsedRequest::Sync(vmm_action)) => {
                 self.serve_vmm_action_request(vmm_action, request_processing_start_us)
@@ -175,7 +197,7 @@ impl ApiServer {
     }
 
     fn serve_vmm_action_request(
-        &self,
+        &mut self,
         vmm_action: Box<VmmAction>,
         request_processing_start_us: u64,
     ) -> Response {
@@ -205,6 +227,8 @@ impl ApiServer {
             .expect("Failed to send VMM message");
         self.to_vmm_fd.write(1).expect("Cannot update send VMM fd");
         let vmm_outcome = *(self.vmm_response_receiver.recv().expect("VMM disconnected"));
+        #[cfg(target_arch = "x86_64")]
+        self.check_for_fatal_error(&vmm_outcome);
         let response = ParsedRequest::convert_to_response(&vmm_outcome);
 
         if vmm_outcome.is_ok() {
@@ -215,6 +239,14 @@ impl ApiServer {
             }
         }
         response
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    fn check_for_fatal_error(&mut self, response: &std::result::Result<VmmData, VmmActionError>) {
+        // Errors considered as fatal are added here
+        if let Err(VmmActionError::LoadSnapshot(_)) = response {
+            self.vmm_fatal_error = true;
+        }
     }
 
     fn get_instance_info(&self) -> Response {
@@ -355,7 +387,7 @@ mod tests {
         let (to_api, vmm_response_receiver) = channel();
         let mmds_info = MMDS.clone();
 
-        let api_server = ApiServer::new(
+        let mut api_server = ApiServer::new(
             mmds_info,
             vmm_shared_info,
             api_request_sender,
@@ -546,7 +578,7 @@ mod tests {
         let (to_api, vmm_response_receiver) = channel();
         let mmds_info = MMDS.clone();
 
-        let api_server = ApiServer::new(
+        let mut api_server = ApiServer::new(
             mmds_info,
             vmm_shared_info,
             api_request_sender,
--- a/src/micro_http/src/server.rs
+++ b/src/micro_http/src/server.rs
@@ -1,6 +1,7 @@
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use logger::error;
 use std::io::{Read, Write};
 use std::os::unix::io::AsRawFd;
 use std::os::unix::io::RawFd;
@@ -48,9 +49,9 @@ impl ServerRequest {
     /// The response is then wrapped in a `ServerResponse`.
     ///
     /// Returns a `ServerResponse` ready for yielding to the server
-    pub fn process<F>(&self, callable: F) -> ServerResponse
+    pub fn process<F>(&self, mut callable: F) -> ServerResponse
     where
-        F: Fn(&Request) -> Response,
+        F: FnMut(&Request) -> Response,
     {
         let http_response = callable(self.inner());
         ServerResponse::new(http_response, self.id)
@@ -377,6 +378,27 @@ impl HttpServer {
             .retain(|_, client_connection| !client_connection.is_done());
 
         Ok(parsed_requests)
+    }
+
+    /// This function is responsible with flushing any remaining outgoing
+    /// requests on the server.
+    ///
+    /// Note that this function can block the thread on write, since the
+    /// operation is blocking.
+    pub fn flush_outgoing_writes(&mut self) {
+        for (_, connection) in self.connections.iter_mut() {
+            while connection.state == ClientConnectionState::AwaitingOutgoing {
+                if let Err(e) = connection.write() {
+                    if let ServerError::ConnectionError(ConnectionError::InvalidWrite) = e {
+                        // Nothing is logged since an InvalidWrite means we have successfully
+                        // flushed the connection
+                    } else {
+                        error!("Connection write error: {}", e);
+                    }
+                    break;
+                }
+            }
+        }
     }
 
     /// The file descriptor of the `epoll` structure can enable the server to become
--- a/tests/README.md
+++ b/tests/README.md
@@ -20,6 +20,12 @@ To run tests from specific directories and/or files:
 tools/devtool test -- <test_dir_or_file_path>...
 ```
 
+To run a single specific test from a file:
+
+``` sh
+tools/devtool test -- <test_file_path>::<test_name>
+```
+
 The testing system is built around [pytest](https://docs.pytest.org/en/latest/).
 Any parameters passed to `tools/devtool test --` are passed to the `pytest`
 command. `devtool` is used to automate fetching of test dependencies (useful
@@ -247,6 +253,11 @@ local image cache.
 *Is there a way to speed up integration tests execution time?*
 `A5:`
 You can speed up tests execution time with any of these:
+
+`Q6:`
+*How can I get live logger output from the tests?*
+`A6:`
+Accessing **pytest.ini** will allow you to modify logger settings.
 
 1. Run the tests from inside the container and set the environment variable
    `KEEP_TEST_SESSION` to a non-empty value.
--- a/tests/integration_tests/functional/test_snapshot_basic.py
+++ b/tests/integration_tests/functional/test_snapshot_basic.py
@@ -5,8 +5,11 @@
 import filecmp
 import logging
 import os
+import subprocess
 import platform
 import tempfile
+from pathlib import Path
+import time
 import pytest
 from conftest import _test_images_s3_bucket
 from framework.artifacts import ArtifactCollection, ArtifactSet
@@ -350,6 +353,77 @@ def test_5_inc_snapshots(network_config,
                              ])
 
     test_matrix.run_test(_test_seq_snapshots)
+
+
+def _get_process_start_time(p_pid):
+    # Helper function returning:
+    # - start time of process with provided PID
+    # - return code of stat -c%X /proc/$$ command
+    proc = subprocess.Popen(["stat -c%X /proc/{}".format(p_pid)],
+                            stdout=subprocess.PIPE,
+                            shell=True,
+                            encoding='ascii')
+    out = proc.communicate()[0].strip('\n')
+    return out, proc.returncode
+
+
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
+    reason="Not supported yet."
+)
+def test_load_snapshot_failure_handling(bin_cloner_path):
+    """
+    Test scenario.
+
+    1. Create two empty files representing snapshot memory and
+    microvm state
+    2. Try to load a VM snapshot out of the empty files.
+    3. Verify that an error was shown and the FC process is terminated.
+    """
+    logger = logging.getLogger("snapshot_load_failure")
+    vm = MicrovmBuilder(bin_cloner_path).create_basevm()
+    vm.spawn(log_level='Info')
+
+    # Create two empty files for snapshot state and snapshot memory
+    chroot_path = vm.jailer.chroot_path()
+    snapshot_dir = os.path.join(chroot_path, "snapshot")
+    Path(snapshot_dir).mkdir(parents=True, exist_ok=True)
+
+    snapshot_mem = os.path.join(snapshot_dir, "snapshot_mem")
+    open(snapshot_mem, "w+").close()
+    snapshot_vmstate = os.path.join(snapshot_dir, "snapshot_vmstate")
+    open(snapshot_vmstate, "w+").close()
+
+    # Hardlink the snapshot files into the microvm jail.
+    jailed_mem = vm.create_jailed_resource(snapshot_mem)
+    jailed_vmstate = vm.create_jailed_resource(snapshot_vmstate)
+
+    # Get the Firecracker process starting time
+    fc_time, err = _get_process_start_time(vm.jailer_clone_pid)
+    assert err == 0
+    logger.info("FC process time before snapshoat load: %s.", fc_time)
+
+    # Load the snapshot
+    response = vm.snapshot_load.put(mem_file_path=jailed_mem,
+                                    snapshot_path=jailed_vmstate)
+
+    logger.info("Response status code %d, content: %s.",
+                response.status_code,
+                response.text)
+    assert vm.api_session.is_status_bad_request(response.status_code)
+
+    # Check if FC process is closed
+    pid_time, ret = _get_process_start_time(vm.jailer_clone_pid)
+    while True:
+        if ret != 0 or fc_time != pid_time:
+            break
+        time.sleep(0.1)
+        pid_time, ret = _get_process_start_time(vm.jailer_clone_pid)
+    # Check that either:
+    #  - pid no longer exists, or
+    #  - pid exists but is a different process (different process start time)
+    assert ret != 0 or fc_time != pid_time, "Firecracker process should \
+     have ended with error!"
 
 
 @pytest.mark.skipif(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -444,6 +444,7 @@ name = "micro_http"
 version = "0.1.0"
 dependencies = [
  "libc 0.2.80 (registry+https://github.com/rust-lang/crates.io-index)",
+ "logger 0.1.0",
  "utils 0.1.0",
 ]
 
--- a/src/api_server/Cargo.toml
+++ b/src/api_server/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2018"
 serde = ">=1.0.27"
 serde_derive = ">=1.0.27"
 serde_json = ">=1.0.9"
+libc = ">=0.2.39"
 
 logger = { path = "../logger" }
 micro_http = { path = "../micro_http" }
--- a/src/micro_http/Cargo.toml
+++ b/src/micro_http/Cargo.toml
@@ -8,3 +8,4 @@ edition = "2018"
 libc = ">=0.2.39"
 
 utils = { path = "../utils" }
+logger = { path = "../logger" }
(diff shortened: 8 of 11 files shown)


src/api_server/src/lib.rs
```diff
@@ -217,6 +231,19 @@ impl ApiServer {
         response
     }
 
+    fn check_for_fatal_error(&self, response: &std::result::Result<VmmData, VmmActionError>) {
```
acatangiu, line 234: If you make `&self` mutable, you can avoid the need for using `RefCell`.

```suggestion
    fn check_for_fatal_error(&mut self, response: &std::result::Result<VmmData, VmmActionError>) {
```

src/api_server/src/lib.rs
```diff
@@ -67,6 +70,9 @@ pub struct ApiServer {
     /// FD on which we notify the VMM that we have sent at least one
     /// `VmmRequest`.
     to_vmm_fd: EventFd,
+    /// If this flag is set with some value, the process encountered
+    /// a fatal error and it is going to exit.
+    encountered_fatal_error: RefCell<Option<u8>>,
```
acatangiu, line 75: You should save the error so you can log it before exiting. Also, ideally avoid `RefCell` unless necessary.

```suggestion
    fatal_error_encountered: Option<VmmActionError>,
```

src/api_server/src/lib.rs
```diff
@@ -129,6 +136,12 @@ impl ApiServer {
         loop {
             match server.requests() {
                 Ok(request_vec) => {
+                    if let Some(exit_code) = *self.encountered_fatal_error.borrow() {
```
acatangiu, line 139: This needs to happen inside the loop below; right after sending the HTTP response (after `server.respond()`).

Ideally it should also flush the http socket to make sure the response is sent before exiting the process.

src/api_server/src/lib.rs
```diff
@@ -129,6 +136,12 @@ impl ApiServer {
...
+                        unsafe {
+                            error!("Fatal error with exit code: {}", exit_code);
```
acatangiu, line 141: nit: take this `error!` out of the `unsafe` block.

src/api_server/src/lib.rs
```diff
@@ -217,6 +231,19 @@ impl ApiServer {
         response
     }
 
+    fn check_for_fatal_error(&self, response: &std::result::Result<VmmData, VmmActionError>) {
+        if let Err(e) = response {
+            let mut is_fatal = self.encountered_fatal_error.borrow_mut();
+            // Errors considered as fatal are added to this 'match'
+            match e {
```
acatangiu, line 238: save this `e` to log it before exiting.

src/api_server/src/lib.rs
```diff
@@ -67,6 +70,9 @@ pub struct ApiServer {
     /// FD on which we notify the VMM that we have sent at least one
     /// `VmmRequest`.
     to_vmm_fd: EventFd,
+    /// If this flag is set with some value, the process encountered
+    /// a fatal error and it is going to exit.
```
acatangiu, line 74: nit:

```suggestion
    /// a fatal error and it is going to exit once it sends any pending API response.
```

acatangiu (commented):
Love the approach, only have punctual code comments.

src/api_server/src/lib.rs
berciuliviu: What do you think if I create a separate structure containing both the error and an exit code meaningful to it? This way I could log both the error and pass the exit code to libc::exit.

src/api_server/src/lib.rs
berciuliviu: The problem here is the following: `check_for_fatal_error` is called by `serve_vmm_action_request`, which is called by `handle_request`. Both these functions don't have a _mutable self_ in their function header. If I am to add `&mut self` to both functions, it still wouldn't work since, on line **152**, `handle_request` is called from a closure, for which Rust says "cannot borrow `*self` as mutable, as `Fn` closures cannot mutate their captured variables". This is also the reason I tried using a RefCell. An Arc<Mutex<>> also worked, but I thought it would be too resourceful for a variable that is not used across threads and contains relatively small data.

src/api_server/src/lib.rs
berciuliviu: Will do.

src/api_server/src/lib.rs
berciuliviu: While I agree to this approach, the function `server.requests()` also does the job of responding to the client when there are `epoll::EventSet::OUT` events to be processed. Do you think it would be feasible to call again `server.requests()` before exiting the process, so that it could flush the current response queue?

src/api_server/src/lib.rs
berciuliviu: Will do. The reason for RefCell usage is explained on a comment below.

src/api_server/src/lib.rs
berciuliviu: After discussing with @acatangiu , we have decided to modify the `ServerRequest::process` function to allow for a mutable callable, instead of an immutable one. This change allows us the change the chain of functions above to accept a `&mut self`, allowing us to change the _RefCell_ to a _boolean_.

src/api_server/src/lib.rs
berciuliviu: After discussing with @acatangiu , who guided me to reach out to @georgepisaltu to dive deeper into the case, we have decided on the following: the verification for encountering fatal errors will be done after the server response at line 148. If we have a fatal error, a new HttpServer function will check if we have pending outgoing writes in a loop. If we do, we call `server.requests()` to address the current response queue, while also returning a 500 Internal Server Error to any new incoming requests. We kill the process afterwards.

src/api_server/src/lib.rs
```diff
@@ -217,6 +240,15 @@ impl ApiServer {
         response
     }
 
+    fn check_for_fatal_error(&mut self, response: &std::result::Result<VmmData, VmmActionError>) {
+        if let Err(e) = response {
+            // Errors considered as fatal are added here
+            if let VmmActionError::LoadSnapshot(_) = e {
```
acatangiu, line 246: nit: can be shortened as:
```suggestion
        // Errors considered as fatal are added here
        if let Err(VmmActionError::LoadSnapshot(_)) = response {
```

src/micro_http/src/server.rs
```diff
@@ -379,6 +379,19 @@ impl HttpServer {
...
     }
 
+    /// This function is responsible with flushing any remaining outgoing
+    /// requests on the server.
+    ///
+    /// Note that this function can block the thread on write, since the
+    /// operation is blocking.
+    pub fn flush_outgoing_writes(&mut self) {
+        for (_, connection) in self.connections.iter_mut() {
+            if connection.state == ClientConnectionState::AwaitingOutgoing {
```
acatangiu, line 389: should this be a loop?
```suggestion
            while connection.state == ClientConnectionState::AwaitingOutgoing {
```

acatangiu (commented):
Looks good!

src/micro_http/src/server.rs
sandreim: Can it be an infinite loop ? What if we cannot send data back to the client ?

src/api_server/src/lib.rs
```diff
@@ -67,6 +69,9 @@ pub struct ApiServer {
     /// FD on which we notify the VMM that we have sent at least one
     /// `VmmRequest`.
     to_vmm_fd: EventFd,
+    /// If this flag is set, the process encountered  a fatal error
+    /// and it is going to exit once it sends any pending API response.
+    encountered_fatal_error: bool,
```
sandreim, line 74: I would call it vmm_fatal_error.

src/micro_http/src/server.rs
```diff
@@ -379,6 +379,19 @@ impl HttpServer {
         Ok(parsed_requests)
     }
 
+    /// This function is responsible with flushing any remaining outgoing
+    /// requests on the server.
+    ///
+    /// Note that this function can block the thread on write, since the
+    /// operation is blocking.
+    pub fn flush_outgoing_writes(&mut self) {
```
sandreim, line 387: I think we should also call close() on the fds after writing all data.

sandreim:
The non-draft PR should also include the changelog update on the behavior change.

src/micro_http/src/server.rs
berciuliviu: @acatangiu yes, this could also be a loop. Thumbs up for the suggestion!
@sandreim yes, it can block on write if the operation waits on the client

sandreim: Can we have a timeout for the blocking case ? so that the behavior is predictible.

src/micro_http/src/server.rs
acatangiu: is it necessary? doesn't the same thing happen (process buffers are flushed) during/because-of `libc::exit()`?

src/micro_http/src/server.rs
acatangiu: hmm.. I can't think of any non-complex solution to handling the blocking case.

Is the extra complexity necessary though? Isn't it the same behavior as all other API responses?

It is a good catch @sandreim, but **if there is no easy way to do it**, I say take it as is.
I wouldn't invest more time and adding more complexity for these unlikely error paths which are irrelevant in production scenarios, since that comes at the expense of investing in things with much higher impact and value.

src/api_server/src/lib.rs
berciuliviu: Will do

src/api_server/src/lib.rs
berciuliviu: Will do

src/micro_http/src/server.rs
berciuliviu: I agree with @acatangiu

src/micro_http/src/server.rs
```diff
@@ -379,6 +379,21 @@ impl HttpServer {
...
+        for (_, connection) in self.connections.iter_mut() {
+            while connection.state == ClientConnectionState::AwaitingOutgoing {
+                if connection.write().is_err() {
```
berciuliviu, line 390: I have added this error handling for the case in which an error such as a ConnectionClosed would not allow the function to set the proper ClientConnectionState. In that case, we want to break out of the loop.

src/api_server/src/lib.rs
```diff
@@ -67,6 +69,9 @@ pub struct ApiServer {
     /// FD on which we notify the VMM that we have sent at least one
     /// `VmmRequest`.
     to_vmm_fd: EventFd,
+    /// If this flag is set, the process encountered  a fatal error
```
acatangiu, line 72: nit:
```suggestion
    /// If this flag is set, the process encountered a fatal error
```

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -351,6 +352,82 @@ def test_5_inc_snapshots(network_config,
...
+                                       ssh_key,
+                                       SnapshotType.FULL)
+
+    vm.kill()
+
+    # Corrupt the memory file of the snapshot
+    logger.info("Corrupt the memory file %s", snapshot.mem)
+
+    mem_fd = open(snapshot.mem, "wb")
+    mem_fd.write(bytearray([0xde, 0xad, 0xbe, 0xef, 0x00]))
```
acatangiu, line 399: How does this guarantee the `LoadSnapshot` API command will fail? Wouldn't it be more robust to just use an empty or completely zeroed out snapshot vm-state file?

Memory corruptions will affect guest and while we do have _some_ validations for it, it is not guaranteed that the snapshot load operation will fail because of it.

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -351,6 +352,82 @@ def test_5_inc_snapshots(network_config,
...
+
+    response = vm.snapshot_load.put(mem_file_path=jailed_mem,
+                                    snapshot_path=jailed_vmstate)
+
+    logger.info("Response status code %d, content: %s.", response.status_code, response.text)
+    assert vm.api_session.is_status_bad_request(response.status_code)
+
+    logger.info("The Firecracker PID is %d.", vm.jailer_clone_pid)
+    try:
+        os.kill(vm.jailer_clone_pid, 0)
```
acatangiu, line 425: This is brittle, the pid could be reused, chance is small for that to happen, but I would avoid it if possible.

`os.kill()` can also inadvertently kill other processes that might reuse the pid.

Instead of using `os.kill` you could check process age. See this for example https://unix.stackexchange.com/questions/238641/get-process-age-from-command-line

We can get `jailer_pid` process `start_time` using `/proc/<pid>/stat`, before calling the API. Then recheck after the API call, and it should be either:
 - no process with that PID 👍 
 or
 - process with that PID has newer `start_time` 👍 

Anything else: 👎

tests/integration_tests/functional/test_snapshot_basic.py
berciuliviu: You are correct. I will empty the file in this case.

tests/integration_tests/functional/test_snapshot_basic.py
berciuliviu: Understood, will do as such

berciuliviu: Done!

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -352,6 +353,106 @@ def test_5_inc_snapshots(network_config,
...
 
+DEFAULT_HOST_IP = "192.168.0.1"
+DEFAULT_GUEST_IP = "192.168.0.2"
+DEFAULT_TAP_NAME = "tap0"
+DEFAULT_DEV_NAME = "eth0"
+DEFAULT_NETMASK = 30
+
+
+def _get_process_start_time(p_pid):
+    # Helper function - returns the start_time and returncode for
```
acatangiu, line 364: This comment is misleading:

return code provided by this function is not `return code of provided PID`.

It is actually `return code of 'stat -c%X <pid>'`

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -352,6 +353,106 @@ def test_5_inc_snapshots(network_config,
...
+    mem_fd.truncate(0)
+    mem_fd.close()
+
+    # Create a new VM instance
+    vm = MicrovmBuilder(bin_cloner_path).create_basevm()
+    vm.spawn(log_level='Info')
+
+    # Hardlink all the snapshot files into the microvm jail.
+    jailed_mem = vm.create_jailed_resource(snapshot.mem)
+    jailed_vmstate = vm.create_jailed_resource(snapshot.vmstate)
```
acatangiu, line 425: Can you achieve the same without the first half of this test?

Instead of actually starting a base vm, creating a snapshot and corrupting that snapshot - just try to load empty `jailed_mem` and empty `jailed_vmstate`.

That would significantly cut both complexity and duration of this test.

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -352,6 +353,106 @@ def test_5_inc_snapshots(network_config,
...
+                response.text)
+    assert vm.api_session.is_status_bad_request(response.status_code)
+
+    fc_exit_time, err = _get_process_start_time(vm.jailer_clone_pid)
+    if err != 0:
+        assert True
+    else:
+        if fc_time == fc_exit_time:
+            assert False, "Firecracker process is not dead!"
+        assert True
```
acatangiu, line 453:
```suggestion
    assert err != 0 or fc_time != fc_exit_time, "Firecracker process should have ended with error!"
```

tests/integration_tests/functional/test_snapshot_basic.py
berciuliviu: Done

tests/integration_tests/functional/test_snapshot_basic.py
berciuliviu: Done

tests/integration_tests/functional/test_snapshot_basic.py
berciuliviu: Done

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Added devtool test `-m|--cpuset-mems` flag for memory confinement when tests
   run.
 - Added the virtio traditional memory ballooning device.
+- Added a mechanism for handling fatal errors, which will trigger a process kill when encountered.
```
sandreim, line 13: The process is not killed. I would make this more specific - Added a mechanism to handle vCPU/VMM errors that result in process termination.

src/micro_http/src/server.rs
sandreim: I agree this also works, but as a best practice we could handle cleanup ourselves.

src/micro_http/src/server.rs
sandreim: Please also log a warning/error here.

tests/README.md
```diff
@@ -20,6 +20,12 @@ To run tests from specific directories and/or files:
 tools/devtool test -- <test_dir_or_file_path>...
 ```
 
+To run a single specific test from a file:
```
sandreim, line 23: Please move these changes in a separate commit.

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -352,6 +355,90 @@ def test_5_inc_snapshots(network_config,
...
+                            encoding='ascii')
+    out = proc.communicate()[0].strip('\n')
+    return out, proc.returncode
+
+
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
+    reason="Not supported yet."
+)
+def test_load_snapshot_failure_handling(bin_cloner_path):
```
sandreim, line 381: We can simplify this:

```
from framework.microvms import VMNano
def test_load_snapshot_failure_handling(bin_cloner_path):
    vm_instance = VMNano.spawn(bin_cloner_path)
    vm = vm_intsance.vm
```

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -352,6 +355,90 @@ def test_5_inc_snapshots(network_config,
...
+
+    logger.info("Response status code %d, content: %s.",
+                response.status_code,
+                response.text)
+    assert vm.api_session.is_status_bad_request(response.status_code)
+
+    # Check if FC process is closed
+    seconds = 0.1
+    fc_exit_time, err = _get_process_start_time(vm.jailer_clone_pid)
+    while seconds < 300:
```
sandreim, line 432: You can just loop forever here, framework will timeout the test after 300 seconds.

src/micro_http/src/server.rs
georgepisaltu: Every FD backing a stream is set to non-blocking before creating the connection (`server.rs:489`). Writes can't be blocking as we would immediately receive `EAGAIN` or `EWOULDBLOCK`. As long as we check for these two error cases, the loop can't be infinite.

src/api_server/src/lib.rs
acatangiu: nit: leftover whitespace still there

src/api_server/src/lib.rs
```diff
@@ -217,6 +241,14 @@ impl ApiServer {
         response
     }
 
+    #[cfg(target_arch = "x86_64")]
```
acatangiu, line 244: nit: instead of making the function `x86_64` specific, maybe we can make just it's internal logic specific. The function could then be called on all platforms.

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -352,6 +355,90 @@ def test_5_inc_snapshots(network_config,
...
+                                    snapshot_path=jailed_vmstate)
+
+    logger.info("Response status code %d, content: %s.",
+                response.status_code,
+                response.text)
+    assert vm.api_session.is_status_bad_request(response.status_code)
+
+    # Check if FC process is closed
+    seconds = 0.1
+    fc_exit_time, err = _get_process_start_time(vm.jailer_clone_pid)
```
acatangiu, line 431: nit:
```suggestion
    pid_time, ret = _get_process_start_time(vm.jailer_clone_pid)
```

future `assert err != 0` makes me think we want `err` when in fact we want `ret == 0`
same for check `pid_time != fc_time` somewhat clearer than `fc_time != fc_exit_time`

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -352,6 +355,90 @@ def test_5_inc_snapshots(network_config,
...
+    while seconds < 300:
+        if err != 0 or fc_time != fc_exit_time:
+            break
+        seconds += 0.1
+        time.sleep(0.1)
+        fc_exit_time, err = _get_process_start_time(vm.jailer_clone_pid)
+    assert err != 0 or fc_time != fc_exit_time, "Firecracker process should \
```
acatangiu, line 438: nit:
leave a helpful comment for future generations
```suggestion
    # Check that either:
    #  - pid no longer exists, or
    #  - pid exists but is a different process (different process start time)
    assert err != 0 or pid_time != fc_time, "Firecracker process should \
```

acatangiu (commented):
Looks good, just a couple of nits from me and some good comments from Andrei.

src/micro_http/src/server.rs
berciuliviu: Done

tests/integration_tests/functional/test_snapshot_basic.py
berciuliviu: Done

src/api_server/src/lib.rs
berciuliviu: Due to Clippy errors, this will be done better when we have more fatal errors to handle, through a larger 'match'.

src/micro_http/src/server.rs
berciuliviu: `connection.write()`  checks for these cases through a more general StreamError approach, so we are covered in this regard. The respective does close the connection whenever it encounters an unexpected StreamError or a ConnectionClosed error.

acatangiu (dismissed):

src/micro_http/src/server.rs
```diff
@@ -379,6 +380,22 @@ impl HttpServer {
...
     }
 
+    /// This function is responsible with flushing any remaining outgoing
+    /// requests on the server.
+    ///
+    /// Note that this function can block the thread on write, since the
+    /// operation is blocking.
+    pub fn flush_outgoing_writes(&mut self) {
+        for (_, connection) in self.connections.iter_mut() {
+            while connection.state == ClientConnectionState::AwaitingOutgoing {
```
georgepisaltu, line 390: After the while loop, I would also set the connection state to `AwaitingIncoming`, so that if we use this method in the future in a context where we won't immediately exit the process, the state of the connection is still valid.

src/micro_http/src/server.rs
```diff
@@ -379,6 +380,22 @@ impl HttpServer {
...
+                if let Err(e) = connection.write() {
+                    error!("Connection write error: {}", e);
```
georgepisaltu, line 392: I suggest avoiding the logging with `error!` if the error we receive from `connection.write()` is `Err(ServerError::ConnectionError(ConnectionError::InvalidWrite))`, since that would indicate that we have nothing to write, so we have successfully flushed everything.

src/micro_http/src/server.rs
berciuliviu: From what I see, the connection state is also set to `AwaitingIncoming` on a successful write.

src/micro_http/src/server.rs
berciuliviu: I agree with @georgepisaltu . Any other error is caught inside the write function. If we still want to log an error, the actual logging could be moved inside the write function itself. What do you think?

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -352,6 +355,88 @@ def test_5_inc_snapshots(network_config,
...
+    3. Verify that an error was shown and the FC process is terminated.
+    """
+    logger = logging.getLogger("snapshot_load_failure")
+    vm = MicrovmBuilder(bin_cloner_path).create_basevm()
+    vm.spawn(log_level='Info')
+
+    vm.create_tap_and_ssh_config(host_ip=DEFAULT_HOST_IP,
+                                 guest_ip=DEFAULT_GUEST_IP,
+                                 netmask_len=DEFAULT_NETMASK,
+                                 tapname=DEFAULT_TAP_NAME)
```
acatangiu, line 396: nit: from what I can tell, you don't need a TAP or configuring microVM networking for this test
```suggestion
```

acatangiu (dismissed):

tests/integration_tests/functional/test_snapshot_basic.py
berciuliviu: You are correct

src/vmm/src/builder.rs
```diff
@@ -419,23 +419,32 @@ pub fn build_microvm_from_snapshot(
             .map_err(RestoreMicrovmState)?;
 
     // Move vcpus to their own threads and start their state machine in the 'Paused' state.
-    vmm.start_vcpus(vcpus, seccomp_filter)
-        .map_err(StartMicrovmError::Internal)?;
+    vmm.start_vcpus(vcpus, seccomp_filter).map_err(|e| {
+        let _ = vmm.exit_vcpus();
```
acatangiu, line 423: I just realized that we can avoid carrying this around every error-case exit by moving it to `Vmm::drop()`.

That would also be a more complete fix; we always want to stop vcpus on dropping the vmm that owns them, right?

src/micro_http/src/server.rs
berciuliviu: Done!

src/vmm/src/builder.rs
berciuliviu: Correct, I have implemented the approach you provided!

acatangiu (approved):
LGTM!

sandreim (approved):

PR #2284, opened 2020-11-19
Fixed function naming in mmio device registration

### LINKED ISSUE #2015: Unnatural flow of virtio initialization
The flow of registering a new virtio device now passes through the following function calls:
- `attach_virtio_device` (which calls into registration functions on both the device manager and the event manager)
- `register_new_virtio_mmio_device` (in the device manager)
- `register_virtio_mmio_device` (in device manager; called from `register_new_virtio_mmio_device`)
- `register_mmio_device` (in device manager; called from `register_virtio_mmio_device`)

We now have 3 functions with almost the same name in the device manager. The name is not helping in identifying what is the difference between these 3 functions, which makes the code harder to read.

HQ01:
Fixed unnatural naming of functions relating to mmio device
registration. 

Signed-off-by: HQ01 <qh384@nyu.edu>

## Reason for This PR

Fixes #2015 

## Description of Changes

Renamed "register_new_virtio_mmio-device" to
"register_mmio_virtio_add_cmdline", "register_virtio_mmio_device" to
"register_mmio_virtio", and "register_new_mmio_boot_timer" to
"register_mmio_boot_timer". New function names are more consistent with
other mmio device registration functions. "New" is now only present in
names of functions that create the corresponding device. Functions that
register a new virtio device now have names better reflecting their
jobs.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -730,7 +730,7 @@ fn attach_virtio_device<T: 'static + VirtioDevice + Subscriber>(
     // The device mutex mustn't be locked here otherwise it will deadlock.
     let device = MmioTransport::new(vmm.guest_memory().clone(), device);
     vmm.mmio_device_manager
-        .register_new_virtio_mmio_device(vmm.vm.fd(), id, device, cmdline)
+        .register_mmio_virtio_for_boot(vmm.vm.fd(), id, device, cmdline)
         .map_err(RegisterMmioDevice)
         .map(|_| ())
 }
@@ -744,7 +744,7 @@ pub(crate) fn attach_boot_timer_device(
     let boot_timer = devices::pseudo::BootTimer::new(request_ts);
 
     vmm.mmio_device_manager
-        .register_new_mmio_boot_timer(boot_timer)
+        .register_mmio_boot_timer(boot_timer)
         .map_err(RegisterMmioDevice)?;
 
     Ok(())
--- a/src/vmm/src/device_manager/mmio.rs
+++ b/src/vmm/src/device_manager/mmio.rs
@@ -172,6 +172,7 @@ impl MMIODeviceManager {
         self.irqs.check(&slot.irqs)
     }
 
+    /// Register a device at some MMIO address.
     fn register_mmio_device(
         &mut self,
         identifier: (DeviceType, String),
@@ -186,7 +187,7 @@ impl MMIODeviceManager {
     }
 
     /// Register a virtio-over-MMIO device to be used via MMIO transport at a specific slot.
-    pub fn register_virtio_mmio_device(
+    pub fn register_mmio_virtio(
         &mut self,
         vm: &VmFd,
         device_id: String,
@@ -236,7 +237,7 @@ impl MMIODeviceManager {
 
     /// Allocate slot and register an already created virtio-over-MMIO device. Also Adds the device
     /// to the boot cmdline.
-    pub fn register_new_virtio_mmio_device(
+    pub fn register_mmio_virtio_for_boot(
         &mut self,
         vm: &VmFd,
         device_id: String,
@@ -244,7 +245,7 @@ impl MMIODeviceManager {
         _cmdline: &mut kernel_cmdline::Cmdline,
     ) -> Result<MMIODeviceInfo> {
         let mmio_slot = self.allocate_new_slot(1)?;
-        self.register_virtio_mmio_device(vm, device_id, mmio_device, &mmio_slot)?;
+        self.register_mmio_virtio(vm, device_id, mmio_device, &mmio_slot)?;
         #[cfg(target_arch = "x86_64")]
         Self::add_virtio_device_to_cmdline(_cmdline, &mmio_slot)?;
         Ok(mmio_slot)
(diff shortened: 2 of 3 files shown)


dianpopa:
Hello,

The functions renamed by the current PR are functions that are used on all platforms supported by Fireracker (ARM included) for inserting some device to  the MMIO bus. While on x86 the location of the devices is made knows to the kernel through the cmdline, on ARM, this information is offered through the FDT. You can also see that these functions, have a [specific x86 cfg](https://github.com/firecracker-microvm/firecracker/blob/1be7cd71df2f5b6132de911dfbe9ee33165b34fa/src/vmm/src/device_manager/mmio.rs#L250) suggestive for these 2 different mechanism. As a consequence, renaming the functions to "_add_to_cmdline" would be misleading for the ARM scenario. Let me know your thoughts and whether you need further clarifications!

src/vmm/src/builder.rs
```diff
@@ -730,7 +730,7 @@ fn attach_virtio_device<T: 'static + VirtioDevice + Subscriber>(
     // The device mutex mustn't be locked here otherwise it will deadlock.
     let device = MmioTransport::new(vmm.guest_memory().clone(), device);
     vmm.mmio_device_manager
-        .register_new_virtio_mmio_device(vmm.vm.fd(), id, device, cmdline)
+        .register_mmio_virtio_add_to_cmdline(vmm.vm.fd(), id, device, cmdline)
```
acatangiu, line 733: The _new_ in the name is there to say that this is a previously never registered `new` device.

acatangiu (changes requested):
@HQ01 Thank you for your contribution.

[Naming is hard](https://martinfowler.com/bliki/TwoHardThings.html), and I agree we can find better names for these functions, but your current suggestions don't add value or extra clarity IMHO.

I suggest you take another look at the code, understand its purpose and the differences between the different registration functions and the differences of the contexts where each is used, and then you will be able to come up with actual better naming.

src/vmm/src/builder.rs
HQ01: @acatangiu thanks for the comment! Here I took the function "register_new_mmio_rtc" as reference, thought that "new" refers to the fact that a device is created in the function. Is the fact that this is a previously never registered new device is already implied by "register"? In another word, do we need to register a device that's previously registered?

acatangiu: Like I said, naming is hard, and current names are indeed confusing.

> Is the fact that this is a previously never registered new device is already implied by "register"? In another word, do we need to register a device that's previously registered?

On the snapshots path, we register a device that has previously been registered in the original microVM where the snapshot comes from. The difference being that there are two paths:
 1. the vm is fresh booted and a (new) device is registered on the first free slot
 2. the vm is a restore and a device is registered at a specific slot (same address where it was in the original/snapshot-source vm)

Alternative naming ideas that cross my mind right now:
 - `register_mmio_virtio_add_to_cmdline` for the boot and `register_mmio_virtio_at_slot` for snapshot
  or simply
 - `register_mmio_virtio_for_boot` and `register_mmio_virtio_from_snapshot`

HQ01:
> @HQ01 Thank you for your contribution.
> 
> [Naming is hard](https://martinfowler.com/bliki/TwoHardThings.html), and I agree we can find better names for these functions, but your current suggestions don't add value or extra clarity IMHO.
> 
> I suggest you take another look at the code, understand its purpose and the differences between the different registration functions and the differences of the contexts where each is used, and then you will be able to come up with actual better naming.

@acatangiu thanks for the suggestion! I'm new to Firecracker but would love to contribute/open to discussion. This is one of my first tries to contribute to Firecracker, so please correct me if I'm wrong/miss anything.

I do take a look at the three functions involved here and their context. `register_mmio_device` is a utility function that's used to register various devices at MMIO addresses and its name reflects its usage quite well, so I didn't change that. I think the major confusion in the original "unnatural flow of virtio initialization" comes from `register_new_virtio_mmio_device` and `register_virt_mmio_device`. `register_new_virtio_mmio_device` is essentially a wrapper of 1) actually registering the virtio device and 2) made the device's location known to the kernel. `register_virtio_mmio_device` is renamed to `register_mmio_virtio` to be consistent with `register_mmio_serial` and `register_mmio_boot_timer`.

So it boils down to how do we name the helper function `register_new_virtio_mmio_device` that includes virtio device registration and adding it to cmdline. I think @dianpopa made a great point pointing out that the second part has a specific x86 cfg.  Does the "new" keyword in the original function naming capture the second point well? Any thought/suggestion is welcomed!

src/vmm/src/builder.rs
HQ01: @acatangiu sounds good to me! Please correct me if I'm wrong, I think the restore path you are referring to is in src/device_manager/persist::restore, which call `register_virtio_mmio_device` directly. Should we just rename `register_new_virtio_mmio_device` to `register_mmio_virtio_for_boot`, without changing `register_virtio_mmio_device` to `register_mmio_virtio_from_snapshot`? This is because the second function is also called internally from the first one.

HQ01:
Hi @acatangiu function names are updated as discussed.

acatangiu (approved):

dianpopa (approved):

PR #2299, opened 2020-11-24
Remove ZSWAP from kernel configs

### LINKED ISSUE #1145: Compile zbud into the kernel used for running aarch64 tests
Missing module causes the following warnings when booting on aarch64:
```
[    2.353087] zswap: default zpool zbud not available
[    2.386900] zswap: pool creation failed
```

dianpopa:
## Reason for This PR

Fixes #1145.

## Description of Changes

* new working kernel configs for arm64 (kernel in S3 is now compiled with this)
* x86_64 config with balloon skipped features and zswap removed

## License Acceptance

## PR Checklist


resources/microvm-kernel-arm64.config
```diff
@@ -193,7 +193,6 @@ CONFIG_SIGNALFD=y
 CONFIG_TIMERFD=y
```
raduiliescu, line 193: There is a huge difference in the kernel configs and zswap is just a small part.  Why all these changes? How did you generate it and from where you started?

Why do we need a different config with initramfs, I guess if you don't specify initramfs at startup it shouldn't add any penalty even if the config has **support** for it.

dianpopa: Maybe I should have mentioned it, apart from removing ZSWAP, I also tried to bring the arm64 config as close to the x86_64 config. I mentioned this in the commit associated. One major difference between the 2 was that the arm64 config had a lot of loadable modules I.e (=m) while x86_64 did not have any and so I tried to make the 2 as similar as possible.

dianpopa: I will try to see if using initrd enabled kernel will work even if no initrd is offered.

dianpopa: Indeed, we do not need a different config with initramfs, @raduiliescu take another look.

georgepisaltu (approved):

raduiliescu (approved):

resources/microvm-kernel-x86_64.config
```diff
@@ -543,6 +541,8 @@ CONFIG_MEMORY_HOTPLUG_SPARSE=y
 CONFIG_MEMORY_HOTREMOVE=y
 CONFIG_SPLIT_PTLOCK_CPUS=4
 CONFIG_ARCH_ENABLE_SPLIT_PMD_PTLOCK=y
+CONFIG_MEMORY_BALLOON=y
+CONFIG_BALLOON_COMPACTION=y
```
sandreim, line 545: This is useful when guest is using hugepages as it helps improving memory defragmentation but at the cost of CPU overhead of an extra thread doing background work: https://elixir.bootlin.com/linux/latest/source/drivers/virtio/virtio_balloon.c#L753. 

Are we sure we want to have this by default ? I would think about a separate kernel configuration tuned for a specific usecase that requires guest huge page support.

sandreim: @georgepisaltu what is your thinking about this ?

dianpopa: I already talked to @georgepisaltu , these options are automatically selected by VIRTIO_BALLON. George removed them manually when updating the config with VIRTIO_BALLON.

Issue #2161, opened 2020-10-06
Freeze devtool pip package versions


gc-plp:
## Describe the bug

The devtool containers that we currently build are not reproducible.

If we build the container at one point in time and rebuild it after a period of time when the packages receive new updates, we may get a different container.

## Expected behaviour

The pip packages in devtool should contain fixed versions.

alindima:
I think the issue originated from this: when modifying the dockerfile and creating a new image tag, you may also update pip packages to their latest versions, without knowing it.
This can result in breaking changes, similar to one we recently had: a new version of pylint complained about some uses of `super()` that we had to modify in order to make the integration test pass. See [this commit](https://github.com/firecracker-microvm/firecracker/commit/cc63335bc82a7f1166409cf236b1403732a9ec25).

This kind of issue can extend to other python packages, potentially introducing breaking changes with their latest versions.

PR #2300, opened 2020-11-24
Use poetry to manage python package while building dev container

kumargu:
# Reason for This PR

Fixes https://github.com/firecracker-microvm/firecracker/issues/2161

## Description of Changes

Freeze versions of pip vended packages using [peotry](https://python-poetry.org/). 
These versions corresponds to latest version of packages on https://pypi.org/. We also
introduce a new script in devtool (build_devctr) which builds a new container image
defined in Dockerfile.

## License Acceptance

## PR Checklist


gc-plp:
Hi @kumargu. 

The PR looks good, but it seems a bit difficult to maintain, as one would have to remove the versions manually and check for changes.

I propose that we wrap this around a build script that uses a `requirements.txt` file and once the build finishes we generate a `requirements-freeze.txt` through [`pip freeze`](https://pip.pypa.io/en/stable/reference/pip_freeze/).

There should be two entry points: (1) `build.sh` that builds the container using the frozen assets and (2) `update.sh` that generates the updated requirements

kumargu:
Updated PR as suggested. Once this looks good, I will update the docs.

tools/requirements.txt
```diff
@@ -0,0 +1,16 @@
+setuptools==50.3.2
```
gc-plp, line 1: Shouldn't these requirements not be bound to a version?

gc-plp (commented):
The PR looks great, but the two shell scripts also need a short readme on how to use each one.

Could you create one in the tests folder or maybe in the docs folder?

tools/requirements.txt
kumargu: Yeah these need not be version bounded, removed them in new revision.

kumargu:
> The PR looks great, but the two shell scripts also need a short readme on how to use each one.
> 
> Could you create one in the tests folder or maybe in the docs folder?

added some docs in docs folder.

tools/devctr/Dockerfile.x86_64
```diff
@@ -94,5 +82,7 @@ RUN mkdir "$TMP_BUILD_DIR" \
 ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION_TAG}/tini-static-amd64 /sbin/tini
 RUN chmod +x /sbin/tini
 
+RUN python3 -m pip freeze > /tmp/requirements-freeze.txt
```
alindima, line 85: I am wondering why are we getting the freezed versions from the container. Shouldn't they be identical to the current `tools/requirements-freeze.txt`?

tools/update_requirements.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
```
alindima, line 1: Does this script really update the freezed versions?
As far as I can tell, the container installs packages according to `requirements-freeze.txt`. Then runs a `pip freeze`, which should give the exact same requirements file. Then this script just takes that same file and overwrites the `tools/requirements-freeze.txt` with the exact same file. Or am I missing something?

tools/requirements.txt
```diff
@@ -0,0 +1,14 @@
+boto3==1.16.24
```
alindima, line 1: shouldn't this file be versionless?
Also, why are is number of packages smaller than the number of packages in requirements-freeze.txt

alindima, line 1: On a second thought, I don't really understand why we need this unversioned file. It seems to only be used when there isn't a `requirements-freeze.txt` file present.
But this shouldn't be the case once we submit them to Github, right?

tools/devctr/Dockerfile.x86_64
alindima: I think we should have here an update command before freezing the packages, right?

alindima:
Thanks for the contribution!
Maybe also split the PR into 2-3 commits to make it easier to review. separate commit for doc updates and separate commit for requirements files.

docs/build_docker_image.md
```diff
@@ -0,0 +1,25 @@
+# Building docker image using freezed versions of Pip vended packages
```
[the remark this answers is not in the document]
kumargu: devctr-image.md

docs/update_requirememts.md
```diff
@@ -0,0 +1,12 @@
+# Updating python packages and their versions
```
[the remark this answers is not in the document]
kumargu: I have updated the devctr-image.md, as how the requirements-freeze.txt get updated. 
I still want to keep this doc file because this script could be executed outside of the devctr-image.md. Let's says you don't want to build a new docker images but just want a copy of requirements-freeze.txt from container (given you have an image id) you can use this script.

tools/build_docker_image.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+set -e
```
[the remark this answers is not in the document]
kumargu: just to make sure that if something breaks in the script the execution should get paused with error and next steps should not get triggered.

For example the first step is copying requirements.txt to requirements-freeze.txt. if that step breaks then docker build should not get triggered.

tools/build_docker_image.sh
```diff
@@ -0,0 +1,52 @@
...
+freeze_requirements_if_not_present() {
+    expected_requirements_freeze_file_path="$FIRECRACKER_TOOLS_DIR/requirements-freeze.txt"
+    if [[ ! -f "$expected_requirements_freeze_file_path" ]]; then
+       cp "$FIRECRACKER_TOOLS_DIR/requirements.txt" $expected_requirements_freeze_file_path
+    fi
+}
+
+build_docker_image() {
+    docker build --iidfile $TEMP_IMAGE_ID_LOCATION -t fcuvm/dev -f "tools/devctr/$1" . 
+    echo "Completed buidling docker image"
```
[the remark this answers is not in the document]
kumargu: ack

tools/requirements.txt
kumargu: > shouldn't this file be versionless?

```requirements.txt``` is used just once, for the fist time when we merge this code. I kept things simple to use the latest version for these packages. Otherwise I would have to run pip install for these packages on host and then create ```requirements-freeze.txt```

Also per my understanding requirements.txt are generally meant to be locked version, but that not a hard rule.  


>Also, why are is number of packages smaller than the number of packages in requirements-freeze.txt

This is because a single pip vended packages can bring multiple other dependant packages.
 Eg  ```python -m pip install aws-shell==0.2.2``` brings following dependencies.
```
aws-shell==0.2.2
awscli==1.18.190
boto3==1.16.30
botocore==1.19.30
colorama==0.4.3
configobj==5.0.6
docutils==0.15.2
jmespath==0.10.0
prompt-toolkit==1.0.18
pyasn1==0.4.8
Pygments==2.7.3
python-dateutil==2.8.1
PyYAML==5.3.1
rsa==4.5
s3transfer==0.3.3
six==1.15.0
urllib3==1.26.2
wcwidth==0.2.5
```

Comment (kumargu) [reply tools/requirements.txt:1]: You are right, requirements.txt won't be used in future. It's just the **base** list of packages we need for the container.  The actual list of pip vended freeze might be different after installations.

I can a separate PR to get rid of ```requirements.txt``` with some code cleanup. But as long as we refer in the code we should keep that file.

File: tools/update_requirements.sh (hunk 3 above)
Comment (kumargu) [reply tools/update_requirements.sh:1]: >Does this script really update the freezed versions?

The version would not change a lot of times, and that's the intention of the CR, to keep things freezed. But it could change. Following are the cases when this gets updated - 

* The first version of ```requirements-freeze``` has just the base packages (some 14) but after running  pip install commands run on container we get some 42 pip vended packages to lock. So we copy/overwrite them to requirements-freeze.txt. And this becomes our first git version if requirements-freeze.txt. 

* Let's say we added a new package dependency in requirements-freeze.txt, any new additions to package list should also get reflected in requirements-freeze.txt which is pulled from container.

* Same thing applies when we upgrade packages versions. 

Essentially it gives us more control to pick which versions we want to pick as part of new diff we see in the requirements-freeze.txt.

File: tools/devctr/Dockerfile.x86_64 (hunk 2 above)
Comment (kumargu) [reply tools/devctr/Dockerfile.x86_64:85]: > I am wondering why are we getting the freezed versions from the container. Shouldn't they be identical to the current `tools/requirements-freeze.txt`?

As discussed in another comment, the list of installed packages might be bigger than list of base packages to install.

Comment (kumargu) [reply tools/devctr/Dockerfile.x86_64:85]: > I think we should have here an update command before freezing the packages, right?

Hm..didn't actually follow. 
Line 68 ```RUN python3 -m pip install --requirement /tmp/requirements-freeze.txt ``` reruns package installations every-time docker image is build.

**Comment by kumargu** [comment]
> Thanks for the contribution!
> Maybe also split the PR into 2-3 commits to make it easier to review. separate commit for doc updates and separate commit for requirements files.

done.

File: tools/update_requirements.sh (hunk 3 above)
Comment (alindima) [reply tools/update_requirements.sh:1]: I see. I was under the impression that this script should also perform an upgrade of the pip packages.
Isn't there something we could do so that the upgrade process would not be manual. With the current state of the PR, we would have to manually edit the requirements file for every single upgraded package.

Something like `./update_requirements.sh --upgrade `?

File: tools/requirements.txt (hunk 4 above)
Comment (alindima) [reply tools/requirements.txt:1]: agree with you. Maybe it makes sense to keep `requirements.txt`. It is similar to a `Cargo.toml` file, while the freezed variant is similar to `Cargo.lock`

Comment (alindima) [reply tools/requirements.txt:1]: i see

File: docs/devctr-image.md [hunk 9]
```diff
@@ -47,9 +47,25 @@ registry. The Firecracker CI suite must also be updated to use the new image.
...
     ```
+    
+   To pull python dependencies required on the container the 
+   script uses ```requirements-freeze.txt``` which has a list of python
+   packages with their freezed versions. This ensures that same version of
+   packages are built on container in it's entire life-cycle, unless one
+   wants to intentionally update the freezed versions.  
+   #### How does requirements-freeze.txt gets updated?
+
+   After a successful docker build run, we run ```pip freeze``` on container
```
alindima, line 64: it's not obvious that this is done via the `update_docker_image.sh` script

docs/devctr-image.md
```diff
@@ -47,9 +47,25 @@ registry. The Firecracker CI suite must also be updated to use the new image.
...
+   to get all the python dependencies and their respective versions.
+   We then copy the dependencies on the host to ```requirements-freeze.txt```.
+
+   If you see a diff in requirements-freeze.txt you must validate for backward compatibility and raise a PR to update the diff.
```
alindima, line 68: Can this ever happen without knowingly updating the `requirements-freeze` file?
Also, the line is too long, if I remember correctly, we try to keep them under 100 chars

tools/devctr/Dockerfile.x86_64
alindima: I was under the impression that the update script should also run an upgrade command

tools/update_requirements.sh
alindima: As a side note, I would also rename the scripts to include `pip` in their name. `update_requirements` does not say much about what the requirements refer to. Maybe `update_pip_requirements`?

docs/update_requirememts.md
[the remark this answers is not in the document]
alindima: if we keep this file, maybe just link to it from the `devctr-image.md` file.

tools/update_requirements.sh
kumargu: > I see. I was under the impression that this script should also perform an upgrade of the pip packages.
> Isn't there something we could do so that the upgrade process would not be manual. With the current state of the PR, we would have to manually edit the requirements file for every single upgraded package.
> 
> Something like `./update_requirements.sh --upgrade `?

Right now upgrades are not supported, it would be manual, but I don't think it would be lot of pain. However we can make the support for upgrades as well later. 

One easy way to run upgrades would be to remove all the versions from requirements-freeze.txt and the new updated file will have upgraded versions.

kumargu: > As a side note, I would also rename the scripts to include `pip` in their name. `update_requirements` does not say much about what the requirements refer to. Maybe `update_pip_requirements`?

sounds good. done.

docs/devctr-image.md
kumargu: made some edits in the docs.

docs/devctr-image.md
kumargu: >Can this ever happen without knowingly updating the requirements-freeze file?
I think it can happen. a) if pip versions in Docker image gets updated, not sure tho b)new pip installs are defined in Docker images instead of defining in requirements-freeze

tools/devctr/Dockerfile.aarch64
```diff
@@ -53,23 +53,11 @@ RUN apt-get update \
     && python3 -m pip install \
```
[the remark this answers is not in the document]
kumargu: these need not be in the the requirements file, they just manage the installations, that's why they are special.
They need to be executed before we actually run pip install -r requirements.txt, otherwise the command will break. 
Also even you see they are not part of requirements-freeze which was copied from container. This is because they don't pull any pip vended packages.

docs/update_requirememts.md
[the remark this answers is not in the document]
kumargu: sounds good. done.

kumargu:
@alindima can you please take another look at the PR, replied back to your comments in the conversation.

tools/build_docker_image.sh
```diff
@@ -0,0 +1,52 @@
...
+            -f|--docker-file) docker_file_name="$2"; shift ;;
+            *) echo "Invalid parameter passed: $1"; exit 1 ;;
+        esac
+        shift
+    done
+    
+    freeze_requirements_if_not_present
+    build_docker_image $docker_file_name
+    image_id=$(get_image_id)
+    bash $FIRECRACKER_TOOLS_DIR/update_requirements.sh -i $image_id
```
gc-plp, line 49: This path doesn't exist. It should be renamed to `update_pip_requirements.txt`

kumargu: thanks a lot. fixed the bug in new iteration. 

Tested by re-running the script against Dockerfile.aarch64 and Dockerfile.x86_64.

gc-plp:
Hi @kumargu.

I took a look at the overall proposal and it's heading the right way.

I do have a few suggestions and things to note down:

1. There seem to be a few dependencies on paths, which are already defined in `devtool`. Perhaps we could reduce duplicating those variables by integrating the two scripts in `devtool`. What do you think?

2. The dockerfile definition can be automatically selected by checking out the output of `uname -m`.

3. `requirements.txt` and `requirements-freeze.txt` should be placed in the `devctr` folder.

4. The `requirements.txt` file should not contain any fixed versions. It's purpose is to act as an unbounded list of requirements.

5. The `build` step seems to copy the freeze file from the rootfs and then freeze it back, which is not actually doing anything. I think that the build step should simply use the freeze file to install packages. If we want to do the update in this step, during the build step we could run `pip --list outdated`. If outdated packages are found, then we can prompt the user and state that outdated packages have been found, print what packages are outdated. If the user wants to use the new versions, we run `pip update` and then generate the new freeze file.

6. The `update` seems to require an image ID, but we could simply call the build step using `requirements.txt`. If, as suggested in step 5, we include the update step in the build step, the update step would not be user interfacing anymore.

kumargu:
@gc-plp 
Addressed  and followed all comments from 1-4. 

For 5, I followed a slightly different approach for getting outdated packages, otherwise followed the suggestion of merging everything in build step. ```pip --list outdated``` needs the packages to be installed on the host which I wanted to avoid, The user might be using the host for other use-cases.

The script looks for packages in [PyPi](https://pypi.org/project/) store and fetch all versions for a package. 

* For adding new packages, the script prompts the user(as suggested) to pick from recent 5(can be increased) versions and update ```requirements-freeze.txt``` file with chosen versions. Addition of new packages is done by running a diff between ``requirements.txt```  and ``requirements-freeze.txt``` 

* For version upgrades we, the script prompts the user to pick from all version greater than current version defined in ``requirements-freeze.txt```  and then do an in-place update of the ```requirements-freeze.txt```

In the last step we build the image with any new packages + any new version.

tools/devtool
```diff
@@ -375,6 +408,117 @@ run_devctr() {
...
+}
+
+# Scans {arch-based}-requirements-freeze.txt file and looks for all package versions greater
+# than current freezed versions. Prompts the user to either upgrade the
+# {arch-based}-requirements-freeze.txt file with any of the newer versions or skip the upgrade.
+# 
+# Args:
+#   $1  path to the {arch-based}-requirements-freeze file.
+# 
+upgrade_version() {
```
gc-plp, line 496: pip provides a subcommand for checking for outdated packages: `pip list --outdated`

It would be much simpler to run this command and check if there are outdated packages. If there are, we just prompt the user and say something like: 
````
There are outdated packages: 
<output of pip list --outdated>. 

Would you like to upgrade these packages?
```

kumargu: Agree that the current logic is little complex. One reason I didn't choose ```pip list --outdated``` was because, on the host, there might be user installed packages other than defined in requirements-freeze.txt. If we run pip list --outdated and then freeze it back we bring unwanted changes and its will become hard to filter/identify what to pick and leave from git diff.

Had ```pip list --outdated``` taken a requirements files, that would have solved our problem.

tools/devtool
```diff
@@ -441,6 +563,8 @@ cmd_help() {
     echo "        Performs prerequisites checks needed to execute firecracker."
     echo "    fix_perms"
     echo "        Fixes permissions when devtool dies in the middle of a privileged session."
+    echo "    build_devctr"
+    echo "        Builds a development docker container."
```
gc-plp, line 567: nit: the description is a bit misleading, as this containers role is to build and test Firecracker - see the comments at the top of the file.

gc-plp:
PR looks good.Tested it and it works great :) 

I think that the second and third commits can be squashed together since both address the same functionality.

tools/devtool
```diff
@@ -375,6 +393,110 @@ run_devctr() {
...
+    say "Checking for newer package versions..."     
+    
+    image_id=$(docker images | grep fcuvm | awk '{print $3}' | head -n 1)
+    echo $image_id
+    outdated_packages=($(docker run -it $image_id python3 -m pip list --outdated --format=freeze | grep -v '^\-e' | cut -d = -f 1))
+
+    if [ "${#outdated_packages[@]}" -ne 0 ]; then
+        say "List of outdated packages"
+        printf '%s\n' "${outdated_packages[@]}"
+        say "Would you like to upgrade these packages?"
```
gc-plp, line 488: There is a primitive that prompts for user input named `get_user_confirmation`. 
It has built-in functionality for asking for `y/n` input.

tools/devtool
```diff
@@ -375,6 +393,110 @@ run_devctr() {
...
+    declare requirements_freeze_file_location="$1"
+    say "Checking for newer package versions..."     
+    
+    image_id=$(docker images | grep fcuvm | awk '{print $3}' | head -n 1)
+    echo $image_id
+    outdated_packages=($(docker run -it $image_id python3 -m pip list --outdated --format=freeze | grep -v '^\-e' | cut -d = -f 1))
+
+    if [ "${#outdated_packages[@]}" -ne 0 ]; then
+        say "List of outdated packages"
+        printf '%s\n' "${outdated_packages[@]}"
```
gc-plp, line 487: There is a primitive that prints text on the terminal called `say`. Could it be used here?

tools/devtool
kumargu: Ack. replaced with ```get_user_confirmation```

kumargu:
> PR looks good.Tested it and it works great :)
> 
> I think that the second and third commits can be squashed together since both address the same functionality.

Thank for the guidance.

Also squashed the commit.

tools/devtool
kumargu: Updated, let me know if the new one looks ok.

tools/devtool
```diff
@@ -375,6 +388,102 @@ run_devctr() {
         "$DEVCTR_IMAGE" "${ctr_args[@]}"
 }
 
+# Given a list of packages it adds them to {arch-based}-requirements_freeze.txt
+#
+# Args:
```
alindima, line 393: This args comment does not say anything about the other arguments, `$2..$n`

tools/devtool
```diff
@@ -375,6 +388,102 @@ run_devctr() {
...
+#   $1  path to the {arch-based}-requirements-freeze file.
+#
+perform_update() {
```
alindima, line 396: All of these helper functions should include pip/python in their names. Their logic is specific to pip, not to all packages installed in the docker container

tools/devtool
```diff
@@ -375,6 +388,102 @@ run_devctr() {
...
+# 
+perform_upgrade() {
+    declare image_id="$1"
+    declare requirements_freeze_file_location="$2"
+
+    dummy_container_name=$(uuidgen)
+    dummy_container_id=$(docker create -ti --name "$dummy_container_name" "$image_id" bash)    
+    docker start "$dummy_container_id"
+    
+    cmd="python3 -m pip list --outdated --format=freeze | grep -v '^\-e' | cut -d = -f 1 | xargs -n 1 python3 -m pip install -U; python3 -m pip freeze > /tmp/requirements-freeze.txt;"
```
alindima, line 451: please split this very long line into multiple lines. Same story in the `upgrade_version` function

tools/devtool
alindima: these functions also have very similar, generic names. It's hard to tell the difference from `update_packages` and `perform_update` without reading the code. You can try being a little more specific, where it makes sense

docs/devctr-image.md
```diff
@@ -46,9 +46,31 @@ registry. The Firecracker CI suite must also be updated to use the new image.
...
+
    ```bash
-    docker build -t fcuvm -f tools/devctr/Dockerfile.x86_64 .
-    ```
+   tools/devtool build_devctr
+   ```
+
+   The script uses two files -
+
+    a) ```requirements.txt``` - This contains list of python packages
```
alindima, line 57: This contains *the* list of python packages.

docs/devctr-image.md
```diff
@@ -46,9 +46,31 @@ registry. The Firecracker CI suite must also be updated to use the new image.
 
 1. Build a new container image with the updated Dockerfile.
 
+   The below script is used to build a docker image.
```
alindima, line 49: This line is a little redundant. Maybe replace it with `Run the following command:`

docs/devctr-image.md
```diff
@@ -46,9 +46,31 @@ registry. The Firecracker CI suite must also be updated to use the new image.
...
+   tools/devtool build_devctr
+   ```
+
+   The script uses two files -
+
+    a) ```requirements.txt``` - This contains list of python packages
+    with their unbounded versions.
+
+    b) {arch-based}-requirements-freeze.txt
+    (eg ```x86-requirements-freeze.txt```) - This contains exhaustive
```
alindima, line 61: I think we could phrase this a little clearer

What about:

`This file contains the exhaustive list of packages installed in the docker image, with their respective versions. Think of it like a Cargo.lock file for pip.`

docs/devctr-image.md
```diff
@@ -46,9 +46,31 @@ registry. The Firecracker CI suite must also be updated to use the new image.
...
+    requirements.txt. Unlike requirements.txt, package
+    version in this file are bounded.
+
+   #### How does requirements-freeze.txt gets updated?
+
+   In every run the script finds if new packages are added in
+   requirements.txt and adds newly added packages in
+   {arch-based}-requirements-freeze.txt.
+
+   The script also looks for all outdated packages on the container,
```
alindima, line 72: "The script also looks for any outdated packages in the container and prompts the user to upgrade them"

tools/devtool
```diff
@@ -375,6 +388,102 @@ run_devctr() {
...
+    say "Checking for newly added requirements..."
+    while IFS= read -r line; do
+        echo "${line%%=*}" >>$parsed_package_location
+    done <"$requirements_freeze_file_location"
+    unset IFS
+
+    diff=($(grep -v -f $parsed_package_location $FC_DEVCTR_DIR/requirements/requirements.txt))    
+    rm $parsed_package_location
+
+    if [ "${#diff[@]}" -ne 0 ]; then
```
alindima, line 424: Tested this locally and it does not handle the removed packages. We should also add support for this use case

kumargu: That's a good call-out. Removing a package is little tricky. 

For addition logic,  we check packages that are present in ```requirements.txt``` but not in ``requirements-freeze.txt```.

We might have to do reverse for deletion logic, check for packages that are in  ```requirements-freeze.txt``` but deleted from  ``requirements-freeze.txt``.  But that won't be accurate,  ``requirements-freeze.txt``` always has more packages than ```requirements.txt``` . 

We can just maintain a single file ```requirements-freeze.txt```(per arch) and add/remove packages directly to this file which solves the problem of addition and deletion. 

@gc-plp @alindima let me know what you guys think about this.

alindima: I think this is a very good idea. This would also simplify the code

gc-plp:
@kumargu did a bit of research and I don't think that pip fits this idea very well.

Seems that there's a different package manager that manages dependencies and creates lockfiles called `poetry`. Did a simple poc and added it [here](https://github.com/gc-plp/firecracker/commit/0fd38595f828a176980874cb87e3c467d24ebd51). I should have exported the lockfile from the container but that can be done separately. 

Let me know your thoughts on how it looks.

kumargu:
@gc-plp updated the PR to use poetry, the PR looks much simpler.

kumargu:
Also, the generated lock file on aarch64 and x86_64 were same.

tools/devtool
kumargu: resolving this comment as we changed the approach and moved to use Poetry.

docs/devctr-image.md
kumargu: changed

tools/devtool
kumargu: fixed across PR.

tools/devtool
kumargu: fixed

docs/devctr-image.md
```diff
@@ -46,9 +46,14 @@ registry. The Firecracker CI suite must also be updated to use the new image.
 
 1. Build a new container image with the updated Dockerfile.
 
+   Run the following command.
```
gc-plp, line 49: This line is not consistent with the one below. We should remove it and just end the sentence above with a `:`

gc-plp (commented):
Works great! Tested it with adding/removing/updating package versions and it reflects the changes in the lockfile.

I suggest updating the PR title and addressing the remaining comments.

docs/devctr-image.md
```diff
@@ -46,9 +46,14 @@ registry. The Firecracker CI suite must also be updated to use the new image.
...
+
    ```bash
-    docker build -t fcuvm -f tools/devctr/Dockerfile.x86_64 .
-    ```
+    tools/devtool build_devctr
+   ```
+
+   This internally, also looks for outdated packages and tries to
```
gc-plp, line 55: This phrase is not very clear on what happens during the update phase. It also does not specify the fact that we can disable updating.

kumargu: made some edits.

docs/devctr-image.md
kumargu: made some edits. let me know if it looks better. Happy to reiterate.

alindima (commented):
Almost there 🚀

tools/devtool
```diff
@@ -303,6 +304,35 @@ cmd_fix_perms() {
...
+            ;;
+        esac
+        shift
+    done
+
+    docker build -t fcuvm -f "$FC_DEVCTR_DIR/$docker_file_name" .
+
+    if [ "$should_update_python_packages" = true ]; then
+        update_python_package_version
+    fi
```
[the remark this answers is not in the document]
alindima: my bad, seems that this is already done by docker

docs/devctr-image.md
```diff
@@ -45,10 +45,21 @@ registry. The Firecracker CI suite must also be updated to use the new image.
 1. Commit the changes, if any.
 
 1. Build a new container image with the updated Dockerfile.
+   a: Additionally also checks for any outdated python packages
```
alindima, line 48: Please add another newline here to render it properly in markdown

tools/devtool
```diff
@@ -426,6 +456,35 @@ test_key() {
...
+# locks the new versions and copies back the ```poetry.lock```
+# file to the host. 
+#
+update_python_package_version() {
+    say "Updating python packages..."
+    
+    # defined in Dockerfile
+    poetry_dir_on_container="/tmp/poetry"
+    lock_file_location_on_host="$FC_DEVCTR_DIR/poetry.lock"
+    image_id=$(docker images | grep fcuvm | awk '{print $3}' | head -n 1)
```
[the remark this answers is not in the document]
kumargu: good point, updated to use ```DEVCTR_IMAGE```

tools/devtool
```diff
@@ -500,6 +559,9 @@ cmd_help() {
     echo "        Performs prerequisites checks needed to execute firecracker."
     echo "    fix_perms"
     echo "        Fixes permissions when devtool dies in the middle of a privileged session."
+    echo "    build_devctr [--no-python-package-update]"
+    echo "        Builds the development container from its Dockerfile."
+    echo "        -n, --no-python-package-update  Do-not update python packages."
```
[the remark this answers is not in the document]
kumargu: ack.

docs/devctr-image.md
```diff
@@ -45,10 +45,21 @@ registry. The Firecracker CI suite must also be updated to use the new image.
...
+   and tries to update the outdated packages. This makes sure
```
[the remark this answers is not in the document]
kumargu: ack

docs/devctr-image.md
```diff
@@ -45,10 +45,21 @@ registry. The Firecracker CI suite must also be updated to use the new image.
...
+   that python packages versions are up to date with latest versions.
 
    ```bash
-    docker build -t fcuvm -f tools/devctr/Dockerfile.x86_64 .
-    ```
+    tools/devtool build_devctr
+   ```
+
+   b: Builds a container image but skips performing updates of python
+   packages. This makes sure that all container images uses the same
```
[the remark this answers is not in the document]
kumargu: ack

docs/devctr-image.md
kumargu: ack

alindima (approved):
Thanks for the contribution!

kumargu:
> Thanks for the contribution!

Thanks for the thorough reviews!

gc-plp (approved):

Issue #2014, opened 2020-07-07
Separate allocate slot function in MMIO allocation and IRQ allocation


andreeaflorescu:
The allocate_slot takes care of two different actions:
- allocating a slot
- allocating an IRQ

For once, this makes the function to be improperly name (because it is not only allocation a slot, but also an irq). Having these two actions as part of the same function is problematic because while all MMIO devices do require a slot, not all MMIO devices require an IRQ. This later problem has been addressed in PR #1921 by having an input parameter which toggles whether an IRQ is needed or not. Yet another problem is that now because the function takes care of both the MMIO slot allocation and the IRQ allocation, the registering of the IRQ in KVM with kvm_register_irqfd takes place in a completely different place in the code. Allocating an IRQ and registering it to KVM should be handled in the same place because the reason for allocating that IRQ is to associate it with a device in KVM.

More on the topic in this conversation: https://github.com/firecracker-microvm/firecracker/pull/1921#discussion_r442159543

dianpopa:
Hi,

@andreeaflorescu my suggestion here would be to just rename the function `allocate_new_slot` to `register_mmio_resources`, similar to what this function is named in [vm-device](https://github.com/rust-vmm/vm-device/blob/master/src/device_manager.rs#L208). 
In regards to allocating the IRQ and registering it in the same place, this would imply significant refactoring that I think is not worth the effort since our goal should be to switch to rust-vmm's vm-device. What are your thoughts on this?

PR #2313, opened 2020-11-30
Split MMIO dev manager function

doracgp:
## Reason for This PR
#2014 
Further discussion on [#1921](https://github.com/firecracker-microvm/firecracker/pull/1921#discussion_r442159543)

## Description of Changes

- `MMIODeviceManager`'s `allocate_new_slot` is replaced by two new methods: `allocate_slot` and `allocate_irq`.
  - `allocate_slot` allocates a memory region.
  - `allocate_irq` allocates a single irq and registers it with KVM for the given VM and event.
- This localizes irq allocation and registration.
- This has the side effect of getting rid of `vm.register_irqfd` calls from various other functions.
- Previously, `allocate_new_slot` returned an `MMIODeviceInfo` struct. It is now the responsibility of methods such as `register_mmio_virtio` and `register_new_mmio_rtc` to do direct calls to `allocate_slot` and `allocate_irq` and construct their own instances of `MMIODeviceInfo`
  - You can look at 42c3ea7 to see a possible, more direct`allocate_new_slot` analog (`allocate_new_device_resources`, a configurable method that returns an `MMIODeviceInfo` directly)
  - However, the current solution was the cleanest way I could think of passing information to the `MMIODeviceManager`, since it allows a simple one-to-one between irq and VM/event, rather than having to pass in an optional IRQ registration struct into an `allocate_new_slot` analog.
  - I would appreciate feedback here and a sense of if this choice makes sense given the use cases.
- There are also a few other side effects:
  - Most notably, previously `register_mmio_virtio` used to atomically register IO events with IRQ events. This is no longer the case.
  - Also, `register_mmio_virtio` no longer catches where > 1 IRQs were allocated before registering them since the two are coupled now.
- `test_slot_allocation` was rewritten as `test_irq_allocation`, since it really just tested the bump allocation of irqs. More thorough unit testing can be added for the two new methods if they seem promising.

Otherwise, all these methods were private to the module, so I don't anticipate there being issues in other parts of the code.

The only change not in `mmio.rs` is in `persist.rs`, and that change was simply a naming change for the sanity check method.

## License Acceptance

## PR Checklist

- [X] Any API changes are reflected in `firecracker/swagger.yaml`.

## Contributor Comments
- I used `cargo fmt` for formatting.
- I am not super familiar with virtio and would love any pointers if this code has gaping holes!
- If this seems like a promising approach, I can go ahead and write some unit tests for the two new methods. 
- This is my first attempt at an open source contribution, so any feedback is welcome. :)

### DIFF
--- a/src/vmm/src/device_manager/mmio.rs
+++ b/src/vmm/src/device_manager/mmio.rs
@@ -21,7 +21,7 @@ use devices::BusDevice;
 use kernel::cmdline as kernel_cmdline;
 use kvm_ioctls::{IoEventAddress, VmFd};
 use logger::info;
-#[cfg(target_arch = "aarch64")]
+
 use utils::eventfd::EventFd;
 use versionize::{VersionMap, Versionize, VersionizeResult};
 use versionize_derive::Versionize;
@@ -156,38 +156,43 @@ impl MMIODeviceManager {
         }
     }
 
-    /// Allocates resources for a new device to be added.
-    fn allocate_new_slot(&mut self, irq_count: u32) -> Result<MMIODeviceInfo> {
-        let irqs = self.irqs.get(irq_count)?;
-        let slot = MMIODeviceInfo {
-            addr: self.next_avail_mmio,
-            len: MMIO_LEN,
-            irqs,
-        };
-        self.next_avail_mmio += MMIO_LEN;
-        Ok(slot)
+    // Allocates an irq and registers it with KVM for a given VM, event.
+    fn allocate_irq(&mut self, vm: &VmFd, evt: &EventFd) -> Result<u32> {
+        let irq = self.irqs.get(1)?[0];
+        vm.register_irqfd(evt, irq).map_err(Error::RegisterIrqFd)?;
+        Ok(irq)
+    }
+
+    // Allocates slot and returns its base.
+    fn allocate_slot(&mut self, len: u64) -> Result<u64> {
+        if len < 0x100 {
+            return Err(Error::InvalidInput);
+        }
+        let base = self.next_avail_mmio;
+        self.next_avail_mmio += len;
+        Ok(base)
     }
 
     #[cfg(target_arch = "x86_64")]
-    /// Does a slot sanity check against expected values.
-    pub fn slot_sanity_check(&self, slot: &MMIODeviceInfo) -> Result<()> {
-        if slot.addr < self.mmio_base || slot.len != MMIO_LEN {
+    /// Does a MMIO device resource sanity check against expected values.
+    pub fn mmio_device_info_sanity_check(&self, device_info: &MMIODeviceInfo) -> Result<()> {
+        if device_info.addr < self.mmio_base || device_info.len != MMIO_LEN {
             return Err(Error::InvalidInput);
         }
-        self.irqs.check(&slot.irqs)
+        self.irqs.check(&device_info.irqs)
     }
 
     /// Register a device at some MMIO address.
     fn register_mmio_device(
         &mut self,
         identifier: (DeviceType, String),
-        slot: MMIODeviceInfo,
+        device_info: MMIODeviceInfo,
         device: Arc<Mutex<dyn BusDevice>>,
     ) -> Result<()> {
         self.bus
-            .insert(device, slot.addr, slot.len)
+            .insert(device, device_info.addr, device_info.len)
             .map_err(Error::BusError)?;
-        self.id_to_dev_info.insert(identifier, slot);
+        self.id_to_dev_info.insert(identifier, device_info);
         Ok(())
     }
 
@@ -197,11 +202,11 @@ impl MMIODeviceManager {
         vm: &VmFd,
         device_id: String,
         mmio_device: MmioTransport,
-        slot: &MMIODeviceInfo,
+        device_info: &MMIODeviceInfo,
     ) -> Result<()> {
         // Our virtio devices are currently hardcoded to use a single IRQ.
         // Validate that requirement.
-        if slot.irqs.len() != 1 {
+        if device_info.irqs.len() != 1 {
             return Err(Error::InvalidInput);
         }
         let identifier;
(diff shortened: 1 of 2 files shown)


doracgp:
@andreeaflorescu , since you opened #2014, would appreciate if you could take a look. Thank you!

src/vmm/src/device_manager/mmio.rs
```diff
@@ -156,38 +156,44 @@ impl MMIODeviceManager {
...
-        let slot = MMIODeviceInfo {
-            addr: self.next_avail_mmio,
-            len: MMIO_LEN,
-            irqs,
-        };
-        self.next_avail_mmio += MMIO_LEN;
-        Ok(slot)
+    // Allocates an irq and registers it with KVM for a given VM, event.
+    fn allocate_irq(&mut self, vm: &VmFd, evt: &EventFd) -> Result<Vec<u32>> {
+        let irq = self.irqs.get(1)?;
```
acatangiu, line 161: This should either
- take a `count_irqs` argument and have support to allocate and register multiple IRQs
  or
- if it's only supposed to do one irq at a time and simply be called multiple times, it should return `Result<u32>` instead of a vec of irqs

src/vmm/src/device_manager/mmio.rs
```diff
@@ -156,38 +156,44 @@ impl MMIODeviceManager {
...
+        vm.register_irqfd(evt, irq[0])
+            .map_err(Error::RegisterIrqFd)?;
+        Ok(irq)
+    }
+
+    // Allocates slot and returns its base.
+    fn allocate_slot(&mut self, len: u64) -> Result<u64> {
+        if len < 0x100 {
+            return Err(Error::InvalidInput);
+        }
```
acatangiu, line 171: Where does the `0x100` limit come from?

src/vmm/src/device_manager/mmio.rs
```diff
@@ -249,11 +263,18 @@ impl MMIODeviceManager {
...
-        let mmio_slot = self.allocate_new_slot(1)?;
-        self.register_mmio_virtio(vm, device_id, mmio_device, &mmio_slot)?;
+        let mmio_device_info;
+        {
+            mmio_device_info = MMIODeviceInfo {
+                addr: self.allocate_slot(MMIO_LEN)?,
+                len: MMIO_LEN,
+                irqs: self.allocate_irq(vm, &mmio_device.locked_device().interrupt_evt())?,
+            };
+        }
```
acatangiu, line 273: Can't this be shortened?
```suggestion
        let mmio_device_info = MMIODeviceInfo {
            addr: self.allocate_slot(MMIO_LEN)?,
            len: MMIO_LEN,
            irqs: self.allocate_irq(vm, &mmio_device.locked_device().interrupt_evt())?,
        };
```

src/vmm/src/device_manager/mmio.rs
```diff
@@ -263,50 +284,63 @@ impl MMIODeviceManager {
...
-        .map_err(Error::RegisterIrqFd)?;
+        let device_info;
+        {
+            device_info = MMIODeviceInfo {
+                addr: self.allocate_slot(MMIO_LEN)?,
+                len: MMIO_LEN,
+                irqs: self
+                    .allocate_irq(vm, &serial.lock().expect("Poisoned lock").interrupt_evt())?,
+            };
+        }
```
acatangiu, line 295: Same here:
```suggestion
        let device_info = MMIODeviceInfo {
            addr: self.allocate_slot(MMIO_LEN)?,
            len: MMIO_LEN,
            irqs: self.allocate_irq(vm, &serial.lock().expect("Poisoned lock").interrupt_evt())?,
        };
```

acatangiu (commented):
Thanks for contributing this! The overall PR direction is good.
Some overall comments:
- The PR contains 8 commits containing different approaches, to get this merged you need to pick one and have a clean commit history. Looking at the sum of all changes it looks to me like you can simply squash all 8 commits in a single one.
- I really appreciate the amount of detail and context that you put in the PR and commit descriptions, keep that up! Small `nit` here: please wrap description according to the [50/72 rule](https://medium.com/@preslavrachev/what-s-with-the-50-72-rule-8a906f61f09c).
- Your changes seem to break snapshot functionality. Looking at the failed tests it seems the `restore` part gets broken. You can locally run specific tests like: `tools/devtool test -- integration_tests/functional/test_snapshot_basic.py`

src/vmm/src/device_manager/mmio.rs
doracgp: It came from [this line](https://github.com/firecracker-microvm/firecracker/blob/8101086bbb942ca7accfd4f6be863fabf72f3d9b/src/vmm/src/device_manager/mmio.rs#L77)

src/vmm/src/device_manager/persist.rs
```diff
@@ -239,7 +239,7 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                                   event_manager: &mut EventManager|
          -> Result<(), Self::Error> {
             dev_manager
-                .slot_sanity_check(slot)
+                .mmio_device_info_sanity_check(slot)
```
doracgp, line 242: The only change to anything `persist` related is here. However, I don't expect this to break tests since none of the logic is changed. Nor do I expect the rest of my code to break tests since the MMIODeviceInfo struct remains unchanged.

There are a lot of warnings in the code mentioning changes that would require a "snapshot version bump," should I look into that even though I don't think any of mine require it?

doracgp:
@acatangiu Thank you so much for the feedback!

I believe I have addressed all of the points you brought up in your review, except for the failing snapshot tests.

It seems something I did is breaking the snapshot restore, however, I'm not sure what would. Could you please take a look at [this comment](https://github.com/firecracker-microvm/firecracker/pull/2313#discussion_r541505879)? Would really appreciate any guidance there.

src/vmm/src/device_manager/mmio.rs
```diff
@@ -197,35 +202,38 @@ impl MMIODeviceManager {
...
-                let io_addr =
-                    IoEventAddress::Mmio(slot.addr + u64::from(devices::virtio::NOTIFY_REG_OFFSET));
+                let io_addr = IoEventAddress::Mmio(
+                    device_info.addr + u64::from(devices::virtio::NOTIFY_REG_OFFSET),
+                );
                 vm.register_ioevent(queue_evt, &io_addr, i as u32)
                     .map_err(Error::RegisterIoEvent)?;
             }
-            vm.register_irqfd(locked_device.interrupt_evt(), slot.irqs[0])
-                .map_err(Error::RegisterIrqFd)?;
```
acatangiu, line 218: This is why snapshot tests fail with these changes.

On the restore from snapshot path, `allocate_irq()` is never called so the IRQ FD is never registered.

src/vmm/src/device_manager/persist.rs
acatangiu: See this comment: https://github.com/firecracker-microvm/firecracker/pull/2313/files#r548548064

You can make your design work by adding a `restore_irq()` function called on the restore path just like `allocate_irq()` is called on the boot path.

acatangiu, line 242: nit: for consistency with the rest of your changes:
```suggestion
                .mmio_device_info_sanity_check(device_info)
```

acatangiu (approved):

acatangiu (changes requested):
Accidentally hit "Approve" in my original review

src/vmm/src/device_manager/mmio.rs
```diff
@@ -156,38 +156,43 @@ impl MMIODeviceManager {
...
-        let irqs = self.irqs.get(irq_count)?;
-        let slot = MMIODeviceInfo {
-            addr: self.next_avail_mmio,
-            len: MMIO_LEN,
-            irqs,
-        };
-        self.next_avail_mmio += MMIO_LEN;
-        Ok(slot)
+    // Allocates an irq and registers it with KVM for a given VM, event.
+    fn allocate_irq(&mut self, vm: &VmFd, evt: &EventFd) -> Result<u32> {
```
dianpopa, line 160: IMO this function should do just what the name says (i.e allocate the irq count that the device needs and let the device choose how many interrupts will use from the ones allocated and how it will use them) so what I propose here is to transform the current `allocate_irq` function you propose into:
```
fn allocate_irqs(&mut self, irq_count: u32) -> Result<Vec<u32>> {
self.irqs.get(irq_count)
}
``` 
and the `vm.register_irqfd` logic is done separately by each device (like it used to be), WDYT?

src/vmm/src/device_manager/mmio.rs
```diff
@@ -156,38 +156,43 @@ impl MMIODeviceManager {
...
+        let irq = self.irqs.get(1)?[0];
+        vm.register_irqfd(evt, irq).map_err(Error::RegisterIrqFd)?;
+        Ok(irq)
+    }
+
+    // Allocates slot and returns its base.
+    fn allocate_slot(&mut self, len: u64) -> Result<u64> {
+        if len < 0x100 {
```
dianpopa, line 168: nit: define a constant with this (i.e MMIO_MIN_LEN)

dianpopa (commented):
After resolving the conflicts you will need to do further adjustments to the aarch64 specific code which was changed to support persist functionality in the meantime.

acatangiu:
@doracgp are you still interested in contributing this PR?

dianpopa:
I will close this PR for the moment. @doracgp feel free to reopen in case you want to resume work here. Thanks!

Issue #2239, opened 2020-10-28
[Improvement] Improve readability for VGIC' distributor register handling


dianpopa:
Right now the code for the distributor registers contains only one list and one type of structure for all distributor registers. For more readability, it would be nice to divide this register into 2 types of register (simple ones and the shared IRQ type).
Any further ideas of readability improvement are welcome.

PR #2326, opened 2020-12-08
Improve readability for VGIC distributor register handling

serban300:
## Reason for This PR

#2239 

## Description of Changes

- GIC registers fixes
- simplify gic registers logic
- Improve readability for VGIC' distributor register handling

## License Acceptance

## PR Checklist

### DIFF
--- a/src/arch/src/aarch64/gic/dist_regs.rs
+++ b/src/arch/src/aarch64/gic/dist_regs.rs
@@ -8,20 +8,19 @@ use kvm_ioctls::DeviceFd;
 // Distributor registers as detailed at page 456 from
 // https://static.docs.arm.com/ihi0069/c/IHI0069C_gic_architecture_specification.pdf.
 // Address offsets are relative to the Distributor base address defined
-// by the system memory map. Unless otherwise stated in the register description,
-// all GIC registers are 32-bits wide.
-const GICD_CTLR: DistReg = DistReg::new(0x0, 0, 4);
-const GICD_STATUSR: DistReg = DistReg::new(0x0010, 0, 4);
-const GICD_IGROUPR: DistReg = DistReg::new(0x0080, 1, 0);
-const GICD_ISENABLER: DistReg = DistReg::new(0x0100, 1, 0);
-const GICD_ICENABLER: DistReg = DistReg::new(0x0180, 1, 0);
-const GICD_ISPENDR: DistReg = DistReg::new(0x0200, 1, 0);
-const GICD_ICPENDR: DistReg = DistReg::new(0x0280, 1, 0);
-const GICD_ISACTIVER: DistReg = DistReg::new(0x0300, 1, 0);
-const GICD_ICACTIVER: DistReg = DistReg::new(0x0380, 1, 0);
-const GICD_IPRIORITYR: DistReg = DistReg::new(0x0400, 8, 0);
-const GICD_ICFGR: DistReg = DistReg::new(0x0C00, 2, 0);
-const GICD_IROUTER: DistReg = DistReg::new(0x6000, 64, 0);
+// by the system memory map.
+const GICD_CTLR: DistReg = DistReg::simple(0x0, 4);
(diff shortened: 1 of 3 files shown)


src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -39,66 +39,73 @@ static VGIC_DIST_REGS: &[DistReg] = &[
...
-            // a register relates to multiple interrupts.
-            end = base + (self.bpi as u32 * (num_irq - super::super::layout::IRQ_BASE) / 8);
-            if self.bpi as u32 * (num_irq - super::super::layout::IRQ_BASE) % 8 > 0 {
-                end += REG_SIZE as u32;
+    }
+}
+
+impl DistReg {
+    fn mem_iter(&self) -> impl Iterator<Item = u32> {
+        let mut start = self.offset;
```
iulianbarbu, line 88: Can we assume `bits_per_irq` % 8 is always 0?  Also, can we assume `bits_per_irq` is > 0?

serban300: For a `DistRegSize::SharedIrq` register, `bits_per_irq` is always > 0. `bits_per_irq % 8` si not always 0.

iulianbarbu: Ok. Shouldn't any register be of `U32_MAX` length, however? I am just trying to understand if we can simplify the logic here, computing the shared irq dist regs based on register size, rather than `bits_per_irq` size, which might result in an offset which is not byte aligned.

serban300: The size of the registers will vary, but any register should be a multiple of `U32_SIZE`. The problem is that for any shared irq register, you want to avoid the regions below irq 32 and above irq IRQ_MAX, because they are RAZ/WI. That's why we need to compute these values.

src/arch/src/aarch64/gic/dist_regs.rs
```diff
@@ -39,66 +39,73 @@ static VGIC_DIST_REGS: &[DistReg] = &[
...
+                // read-as-zero/write-ignore (RAZ/WI) policy.
+                // The first part of a shared-irq register, the one corresponding to the
+                // SGI and PPI IRQs (0-32) is RAZ/WI, so we skip it.
+                start += IRQ_BASE * u32::from(bits_per_irq) / 8;
+
+                let size_in_bits = u32::from(bits_per_irq) * (IRQ_MAX - IRQ_BASE);
+                let mut size_in_bytes = size_in_bits / 8;
+                if size_in_bits % 8 > 0 {
+                    size_in_bytes += u32::from(U32_SIZE);
+                }
```
iulianbarbu, line 102: I am not sure how we should align these. If all regs are multiple of `U32_SIZE` and `size_in_bytes` is not a multiple of `U32_SIZE`, then the real end of the register should be `size_in_bytes` + (U32_SIZE - (size_in_bytes % U32_SIZE)). If we add u32::from(U32_SIZE) always we are going to enter in the RAZ/WI zone, right?

dianpopa (approved):

src/arch/src/aarch64/gic/dist_regs.rs
dianpopa: I do not think I understand the concern. If there is a raz/wi zone, then that zone will be at the start of the register space. So, we could not possible reach it by adding bits to the the register space...What we do in this code is we try to read the amount of register space that is of interest to us (i.e the amount of space that covers the interrupts that our system uses). 
The GICD registers that share interrupts function in a slightly weird way. 
Let us say that we have 15 interrupts in our system, and there is a GICD register that uses 4 bits to represent some info about 1 interrupt. The way we detect the amount of register space is like this:
15 * 4 = 60 (our interrupts cover 60 bits of space)
a u32 has 32 bits =>  60 / 32 = 1 mod 28 => we will read 2 u32s from the register space to get our info on the 15 interrupts. Does it make sense?

iulianbarbu (approved):

Issue #2008, opened 2020-07-06
Add create snapshot tests targeting previous versions.


sandreim:
The tests should create a snapshot targeting an older version and resume it using the corresponding Firecracker  binary artifact.

The tests must run on top of the test matrix defined in issue #2007.

Issue #2006, opened 2020-07-06
Add snapshot resume from older version integration tests  


sandreim:
Context: these tests must resume a microvm snapshot from CI artifacts.
The following areas should be covered:
- Snapshot is loaded successfully
- Previous workload is resumed and completes
- A new workload is started and complete successfully.

PR #2328, opened 2020-12-08
Add snapshot restore integration tests across all supported versions and devices.

sandreim:
## Reason for This PR

Implements #2008 #2006

## Description of Changes

Provided in commit messages.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/framework/artifacts.py
+++ b/tests/framework/artifacts.py
@@ -238,8 +238,9 @@ class SnapshotArtifact:
 
         return Snapshot(dst_mem_path,
                         dst_state_file,
-                        disk_paths,
-                        dst_ssh_key)
+                        disks=disk_paths,
+                        net_ifaces=None,
+                        ssh_key=dst_ssh_key)
 
 
 class DiskArtifact(Artifact):
@@ -433,7 +434,7 @@ class SnapshotType(Enum):
 class Snapshot:
     """Manages Firecracker snapshots."""
 
-    def __init__(self, mem, vmstate, disks, ssh_key):
+    def __init__(self, mem, vmstate, disks, net_ifaces, ssh_key):
         """Initialize mem, vmstate, disks, key."""
         assert mem is not None
         assert vmstate is not None
@@ -443,6 +444,7 @@ class Snapshot:
         self._vmstate = vmstate
         self._disks = disks
         self._ssh_key = ssh_key
+        self._net_ifaces = net_ifaces
 
     def rebase_snapshot(self, base):
         """Rebases current incremental snapshot onto a specified base layer."""
@@ -468,3 +470,58 @@ class Snapshot:
     def ssh_key(self):
         """Return the ssh key file path."""
         return self._ssh_key
+
+    @property
+    def net_ifaces(self):
+        """Return the list of net interface configs."""
+        return self._net_ifaces
+
+
+# Default configuration values for network interfaces.
+DEFAULT_HOST_IP = "192.168.0.1"
+DEFAULT_GUEST_IP = "192.168.0.2"
+DEFAULT_TAP_NAME = "tap0"
+DEFAULT_DEV_NAME = "eth0"
+DEFAULT_NETMASK = 30
+
+
+class NetIfaceConfig:
+    """Defines a network interface configuration."""
+
+    def __init__(self,
+                 host_ip=DEFAULT_HOST_IP,
+                 guest_ip=DEFAULT_GUEST_IP,
+                 tap_name=DEFAULT_TAP_NAME,
+                 dev_name=DEFAULT_DEV_NAME,
+                 netmask=DEFAULT_NETMASK):
+        """Initialize object."""
+        self._host_ip = host_ip
+        self._guest_ip = guest_ip
+        self._tap_name = tap_name
+        self._dev_name = dev_name
+        self._netmask = netmask
+
+    @property
+    def host_ip(self):
+        """Return the host IP."""
+        return self._host_ip
+
+    @property
+    def guest_ip(self):
+        """Return the guest IP."""
+        return self._guest_ip
+
+    @property
+    def tap_name(self):
+        """Return the tap device name."""
+        return self._tap_name
+
+    @property
+    def dev_name(self):
+        """Return the guest device name."""
+        return self._dev_name
+
+    @property
+    def netmask(self):
+        """Return the netmask."""
+        return self._netmask
--- a/tests/framework/builder.py
+++ b/tests/framework/builder.py
@@ -7,17 +7,12 @@ import os
 import tempfile
 from pathlib import Path
 from conftest import init_microvm
-from framework.artifacts import Artifact, DiskArtifact, Snapshot, SnapshotType
+from framework.artifacts import (
+    Artifact, DiskArtifact, Snapshot, SnapshotType, NetIfaceConfig
+)
 import framework.utils as utils
 import host_tools.logging as log_tools
 import host_tools.network as net_tools
-
-
-DEFAULT_HOST_IP = "192.168.0.1"
-DEFAULT_GUEST_IP = "192.168.0.2"
-DEFAULT_TAP_NAME = "tap0"
-DEFAULT_DEV_NAME = "eth0"
-DEFAULT_NETMASK = 30
 
 
 class MicrovmBuilder:
@@ -55,7 +50,7 @@ class MicrovmBuilder:
               disks: [DiskArtifact],
               ssh_key: Artifact,
               config: Artifact,
-              network_config: net_tools.UniqueIPv4Generator = None,
+              net_ifaces=None,
               enable_diff_snapshots=False,
               cpu_template=None):
         """Build a fresh microvm."""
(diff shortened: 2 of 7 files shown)


tests/framework/artifacts.py
```diff
@@ -468,3 +470,58 @@ def disks(self):
     def ssh_key(self):
         """Return the ssh key file path."""
         return self._ssh_key
+
+    @property
+    def net_ifaces(self):
+        """Return the ssh key file path."""
```
berciuliviu, line 476: Comment should be changed to something like "Return the net interfaces".

sandreim: Yup. Copy paste error :)

tests/framework/builder.py
```diff
@@ -8,18 +8,12 @@
 from pathlib import Path
 from conftest import init_microvm
 from framework.artifacts import Artifact, DiskArtifact, Snapshot, SnapshotType
+from framework.artifacts import NetIfaceConfig
```
berciuliviu, line 11: Nit:
If the line is not too long so the styling tests fail, "NetIfaceConfig" could be appended to the line above

tests/framework/microvms.py
```diff
@@ -126,4 +133,6 @@ def spawn(cls, bin_cloner_path, start=False,
                                      None,
                                      start,
                                      fc_binary,
-                                     jailer_binary)
+                                     jailer_binary,
+                                     net_ifaces=net_ifaces,
+                                     enable_diff_snapshots=diff_snapshots)
```
berciuliviu, line 138: Nit here and on lines 117-118:
To keep the style, you could pass directly **net_ifaces** instead of **net_ifaces=net_ifaces**, same with diff_snapshots :)

tests/framework/builder.py
gc-plp: Can be added as a multi-line import too.

tests/framework/artifacts.py
```diff
@@ -468,3 +470,58 @@ def disks(self):
...
+        return self._net_ifaces
+
+
+# Default configuration values for network interfaces.
+DEFAULT_HOST_IP = "192.168.0.1"
```
gc-plp, line 481: Do we want to have default IP addresses? Sounds like a recipe for duplicate IP addresses when writing tests.

tests/framework/microvms.py
gc-plp: When multiple parameters are passed, it's easier to understand what parameters are passed when mentioning each one by key word. 
I'd rather go the other way around and pass each parameter explicitly.

tests/integration_tests/functional/test_balloon.py
```diff
@@ -120,29 +120,8 @@ def copy_util_to_rootfs(rootfs_path, util):
     subprocess.check_call("rmdir tmpfs", shell=True)
 
 
-# pylint: disable=C0103
-def test_rss_memory_lower(test_microvm_with_ssh_and_balloon, network_config):
+def _test_rss_memory_lower(test_microvm):
     """Check inflating the balloon makes guest use less rss memory."""
```
berciuliviu, line 124: The comment here is not pretty clear to me, as it is the same with the other part on line 156. I think the usages changed a little with the split and this should be specified, since this test would inflate the balloon of a custom provided microVM, while the other "half" will be the complete test run for this usecase. From what I understand, `**_**test_rss_memory_lower` is the one to be called from other tests?

tests/framework/artifacts.py
sandreim: Two uVMs can use the same network configuration without any problem since they all live in different net namespaces. To make them communicate the multiple uVM tests would need to do some plumbing work.

tests/integration_tests/functional/test_balloon.py
sandreim: This function implements the actual testing part, while the test function just sets up the microvm and then calls _test_rss_memory_lower. The setup part is something that is present in all pytest tests, so it makes sense to comment what the test actually tests for. I think the same comment applies to both.

tests/framework/artifacts.py
gc-plp: Right, but on the host side there may be collisions.

tests/framework/artifacts.py
sandreim: Fixed.

tests/framework/builder.py
sandreim: Fixed by multiline.

tests/integration_tests/functional/test_snapshot_advanced.py
```diff
@@ -0,0 +1,259 @@
...
+)
+def test_restore_old_version_all_devices(bin_cloner_path):
+    """Test scenario: restore all previous version snapshots."""
+    # Microvm: 2vCPU 256MB RAM, baloon, 4 disks and 4 netdevices.
+    logger = logging.getLogger("snapshot_many_devices")
+
+    artifacts = ArtifactCollection(_test_images_s3_bucket())
+    # Fetch all firecracker binaries.
+    # With each binary create a snapshot and try to restore in current
+    # version.
```
luminitavoicu, line 98: This test resumes the snapshot using older binary artifacts. This comment should be updated.

tests/integration_tests/functional/test_snapshot_advanced.py
```diff
@@ -0,0 +1,259 @@
...
+        logger.debug("========== Firecracker restore snapshot log ==========")
+        logger.debug(microvm.log_data)
+
+
+@pytest.mark.skipif(
+    platform.machine() != "x86_64",
+    reason="Not supported yet."
+)
+def test_restore_old_version_all_devices(bin_cloner_path):
+    """Test scenario: restore all previous version snapshots."""
```
berciuliviu, line 91: Create current version snapshot and restore in older/previous versions?

tests/integration_tests/functional/test_snapshot_advanced.py
```diff
@@ -0,0 +1,259 @@
...
+        scratchdisk = drive_tools.FilesystemFile(tempfile.mktemp(), size=64)
+        vm.add_drive(scratch, scratchdisk.path)
+        disks.append(scratchdisk.path)
+
+        # Workaround FilesystemFile destructor removal of file.
+        scratchdisk.path = None
+
+    vm.start()
+
+    # Iterate and validate copnnectivity on all ifaces after boot.
```
luminitavoicu, line 221:
```suggestion
    # Iterate and validate connectivity on all ifaces after boot.
```

tests/integration_tests/functional/test_snapshot_advanced.py
```diff
@@ -0,0 +1,259 @@
...
+    for blk in scratch_drives:
+        # Create mount point and mount each device.
+        cmd = "mkdir -p /mnt/{blk} && mount /dev/{blk} /mnt/{blk}".format(
+            blk=blk
+        )
+        exit_code, _, _ = ssh_connection.execute_command(cmd)
+        assert exit_code == 0
+
+        # Create file using dd using O_DIRECT.
+        # Afer resume we will conpute md5sum on these files.
```
luminitavoicu, line 238:
```suggestion
        # After resume we will compute md5sum on these files.
```

tests/integration_tests/functional/test_snapshot_advanced.py
```diff
@@ -0,0 +1,259 @@
...
+        jailer = firecracker.jailer()
+        jailer.download()
+
+        logger.info("Creating snapshot with Firecracker: %s",
+                    firecracker.local_path())
+        logger.info("Using Jailer: %s", jailer.local_path())
+
+        target_version = firecracker.base_name()[1:]
+
+        # v0.23 does not support creating diff snapshots.
```
berciuliviu, line 61: How will this scale up when say, for example, we have v0.24, v0.25 and v0.26, with different new features? Will we have a check for every new feature as we have now? Will these tests need to be maintained across new versions?

sandreim: Yes, we need to test all devices that make sense in every version. We can split the test in multiple ways in the future if that will help maintainability.

tests/integration_tests/functional/test_snapshot_advanced.py
sandreim: Yes, it tries to restore all previous version snapshots in current version.

sandreim: I will update the description.

luminitavoicu: When creating a snapshot, the Firecracker API allows to provide a target_version (the microVM version for which we want to create the snapshot). This is optional and defaults to the current version if not specified. Firecracker v0.24 supports v0.23.0 as target version for snapshotting. 

This test creates snapshots using the current Firecracker version, but targeting the old supported versions. Then, the snapshots are resumed using the binary artifacts.

tests/integration_tests/functional/test_snapshot_advanced.py
sandreim: done

tests/integration_tests/functional/test_snapshot_advanced.py
sandreim: done

tests/integration_tests/functional/test_snapshot_advanced.py
sandreim: updated.

sandreim:
@luminitavoicu @berciuliviu @gc-plp updated, PTAL.

berciuliviu (approved):
Looks good to me.

luminitavoicu (approved):

Issue #1146, opened 2019-06-20
Setup cache and cpu topology on aarch64


andreeaflorescu:
We need to setup the cache and cpu topology as well as the machine model on aarch64 as well. I am not sure what mechanism is used on aarch64, but the errors seem to suggest that something should exist.

Numa node:
```
[    0.000000] NUMA: No NUMA configuration found
[    0.000000] NUMA: Faking a node at [mem 0x0000000000000000-0x0000000087ffffff]
[    0.000000] NUMA: NODE_DATA [mem 0x87edac80-0x87ef3fff]
```

Machine model:
```
[    0.000000] Boot CPU: AArch64 Processor [431f0a11]
[    0.000000] Machine model: linux,dummy-virt
```

Cache Topology:
```
[    1.785544] cacheinfo: Unable to detect cache hierarchy for CPU 0
```

petreeftime:
For aarch64 NUMA topology can be specified in the device tree: https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/numa.txt.

PR #2334, opened 2020-12-09
aarch64: setup cache topology

dianpopa:
## Reason for This PR

Setup cache topology for aarch64 guests.
Partially implements #1146.

## Description of Changes

* first part of the commits implement cache topolgy and changes are made in src/aarch64
* second part of the commits makes the sysfs required information available in the jailer
* third part restructures test_cpu_templates
* fourth part adds cache topology integration test for aarch64

## License Acceptance

## PR Checklist


AlexandruCihodaru (approved):
LGTM 🚀

dianpopa:
@AlexandruCihodaru and @sandreim tests have now passed, please re-review! Thanks!

AlexandruCihodaru (approved):

src/arch/src/aarch64/fdt.rs
```diff
@@ -300,27 +309,109 @@ fn generate_prop64(cells: &[u64]) -> Vec<u8> {
...
+        for cache in non_l1_caches.iter() {
+            // We append the next-level-cache node (the node that specifies the cache hierarchy)
+            // in the next iteration. For example,
+            // L2-cache {
+            //      cache-size = <0x8000> ----> first iteration
+            //      next-level-cache = <&l3-cache> ---> second iteration
+            // }
+            // The cpus per unit cannot be 0 since the sysfs will also include the current cpu
+            // in the list of shared cpus so it need to be at least 1. Firecracker trusts the host.
+            assert!(cache.cpus_per_unit > 0);
```
sandreim, line 368: Would it make sense to return an error here ?

src/jailer/src/env.rs
```diff
@@ -320,6 +380,8 @@ impl Env {
         } else {
             None
         };
+        #[cfg(target_arch = "aarch64")]
+        self.copy_cache_info()?;
```
sandreim, line 384: Are we doing this by default ? It would make sense to have a switch for it.

src/arch/src/aarch64/fdt.rs
dianpopa: Yup, it would be more elegant. I will return an error.

src/jailer/src/env.rs
dianpopa: You mean check whether or not the cache info is already present?

dianpopa: We synced offline on this. The problem here is that the cache information will be copied into the chroot irrespective of whether Firecracker needs them or not (for example, upon resuming a microVM from a snapshot we do not need this info but we still copy the cache information into the chroot). Unfortunately, there is no easy way to check from inside the jailer logic what is the purpose of the run. Making one available is not super obvious and we decided to log an issue for this and continue with the current functionality for the moment.

dianpopa:
@sandreim @AlexandruCihodaru please re-review! Thanks!

src/jailer/src/env.rs
acatangiu: If this would be conditional, the user would need to take extra care to only disable it when they are sure no other snapshots will be done. Not sure if the effort is worth it.

In the context of incremental snapshots it would look weird to have this extra knob that _disables further snapshots_ but buys you some microseconds.

sandreim (approved):
👍

AlexandruCihodaru (approved):

PR #2335, opened 2020-12-09
Improved snapshot creation performance tests

luminitavoicu:

## Reason for This PR

Ensure the latency for creating diff or full snapshots does not degrade over time.

## Description of Changes

Improved existing full snapshot creation test and added a new one for measuring latency for diff snapshots.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/integration_tests/performance/test_snapshot_perf.py
+++ b/tests/integration_tests/performance/test_snapshot_perf.py
@@ -7,28 +7,32 @@ import logging
 import os
 import platform
 import pytest
-from conftest import _test_images_s3_bucket
+from conftest import _test_images_s3_bucket, DEFAULT_TEST_IMAGES_S3_BUCKET
 from framework.artifacts import ArtifactCollection, ArtifactSet
 from framework.matrix import TestMatrix, TestContext
 from framework.microvms import VMMicro
 from framework.builder import MicrovmBuilder, SnapshotBuilder, SnapshotType
+from framework.utils import CpuMap
 import host_tools.network as net_tools  # pylint: disable=import-error
 import host_tools.logging as log_tools
-
 
 # How many latencies do we sample per test.
 SAMPLE_COUNT = 3
 USEC_IN_MSEC = 1000
 
 # Latencies in milliseconds.
+# The latency for snapshot creation has high variance due to scheduler noise.
+# The issue is tracked here:
+# https://github.com/firecracker-microvm/firecracker/issues/2346
+# TODO: Update baseline values after fix.
 CREATE_LATENCY_BASELINES = {
     '2vcpu_256mb.json': {
-        'FULL':  1000,
-        'DIFF':  0,
+        'FULL':  180,
+        'DIFF':  50,
     },
     '2vcpu_512mb.json': {
-        'FULL':  1000,
-        'DIFF':  0,
+        'FULL':  280,
+        'DIFF':  50,
     }
 }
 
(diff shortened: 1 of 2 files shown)


tests/integration_tests/performance/test_snapshot_perf.py
```diff
@@ -23,12 +28,12 @@
 # Latencies in milliseconds.
 CREATE_LATENCY_BASELINES = {
     '2vcpu_256mb.json': {
-        'FULL':  1000,
-        'DIFF':  0,
+        'FULL':  270,
```
georgepisaltu, line 31: How did we get to these values?

tests/integration_tests/performance/test_snapshot_perf.py
```diff
@@ -48,57 +53,67 @@ def _test_snapshot_create_latency(context):
...
-        # Create a fresh microvm from aftifacts.
-        vm = vm_builder.build(kernel=context.kernel,
-                              disks=[rw_disk],
-                              ssh_key=ssh_key,
-                              config=context.microvm,
-                              enable_diff_snapshots=enable_diff_snapshots)
+    # Test snapshot creation for every supported target version.
+    for target_version in FIRECRACKER_VERSIONS:
+        # v0.23 does not support diff snapshots
+        if "v0.23" in target_version and snapshot_type == SnapshotType.DIFF:
```
georgepisaltu, line 64: Do we expect any other exceptions of this kind in the future? If so, we should consider a structure for these exceptions for readability. If not then this is fine.

georgepisaltu (commented):
LGTM, just a couple of questions.

tests/integration_tests/performance/test_snapshot_perf.py
sandreim: Are we sure these values will not trigger intermittent failures ?

tests/integration_tests/performance/test_snapshot_perf.py
```diff
@@ -15,6 +15,11 @@
 import host_tools.network as net_tools  # pylint: disable=import-error
 import host_tools.logging as log_tools
 
+# Supported snapshot target versions for the current Firecracker version.
+FIRECRACKER_VERSIONS = {
```
sandreim, line 19: These are also available by parsing the binary ci artifacts from S3. One advantage of that approach is that it does not require to update any code and just upload the binary in S3. The downside is that it is not explicitly visible in the code and would require adding logs to help debug failures.

WDYT ?

tests/integration_tests/performance/test_snapshot_perf.py
sandreim: v0.23 supports loading diff snapshots, we should not skip this test.

tests/integration_tests/performance/test_snapshot_perf.py
sandreim: I would also create a framework method that returns this array.

luminitavoicu: Done

tests/integration_tests/performance/test_snapshot_perf.py
luminitavoicu: I updated the test to contain target and delta values. These results were computed based on the empirical rule, which states that 99.7% of data observed following a normal distribution falls within 3 standard deviations of the mean (µ ± 3σ). The delta value was calculated by multiplying the standard deviation by 3.

sandreim (approved):

georgepisaltu (approved):

PR #2338, opened 2020-12-10
[performance] block device test

iulianbarbu:
## Reason for This PR

Lack of a block device performance test.

## Description of Changes

Added a test for Firecracker block device performance regression detection:
* Used `fio` to compute: iops, bandwidth, latency.
* Used `top` to compute cpu utilization host/guest.
* `fio` workloads cover the following modes: randread, randrw, read, readwrite.
* Block sizes for disk requests vary between: 512, 1024, 4096, 65536 bytes.

**NOTE**: I am still working through validating the computed baselines.

## License Acceptance

## PR Checklist

~~- [ ] Any newly added `unsafe` code is properly documented.~~
~~- [ ] All added/changed functionality is tested.~~

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -87,29 +87,15 @@ import pytest
 import host_tools.cargo_build as build_tools
 import host_tools.network as net_tools
 import host_tools.proc as proc
-
-from framework.artifacts import ArtifactCollection
 import framework.utils as utils
+import framework.defs as defs
+from framework.artifacts import ArtifactCollection
 from framework.microvm import Microvm
 from framework.s3fetcher import MicrovmImageS3Fetcher
 from framework.scheduler import PytestScheduler
 
-
-SPEC_S3_BUCKET = 'spec.ccfc.min'
-"""The s3 bucket that holds global Firecracker specifications."""
-
-DEFAULT_TEST_IMAGES_S3_BUCKET = 'spec.ccfc.min'
-"""The default s3 bucket that holds Firecracker microvm test images."""
-
-ENV_TEST_IMAGES_S3_BUCKET = 'TEST_MICROVM_IMAGES_S3_BUCKET'
-"""Environment variable for configuring the test microvm s3 bucket.
-
-If variable exists in `os.environ`, its value will be used as the s3 bucket
-for microvm test images.
-"""
-
+# Tests root directory.
 SCRIPT_FOLDER = os.path.dirname(os.path.realpath(__file__))
-
 
 # This codebase uses Python features available in Python 3.6 or above
 if sys.version_info < (3, 6):
@@ -134,8 +120,8 @@ if "AMD" in proc.proc_type():
 def _test_images_s3_bucket():
     """Auxiliary function for getting this session's bucket name."""
     return os.environ.get(
-        ENV_TEST_IMAGES_S3_BUCKET,
-        DEFAULT_TEST_IMAGES_S3_BUCKET
+        defs.ENV_TEST_IMAGES_S3_BUCKET,
+        defs.DEFAULT_TEST_IMAGES_S3_BUCKET
     )
 
 
@@ -188,28 +174,38 @@ def pytest_addoption(parser):
     return PytestScheduler.instance().do_pytest_addoption(parser)
 
 
-@pytest.fixture(autouse=True, scope='session')
 def test_session_root_path():
-    """Ensure and yield the testrun root directory.
+    """Create and return the testrun session root directory.
 
-    Created at session initialization time, this directory will be
-    session-unique. This is important, since the scheduler will run
-    multiple pytest sessions concurrently.
+    Testrun session root directory confines any other test temporary file.
+    If it exists, consider this as a noop.
     """
-    root_path = tempfile.mkdtemp(prefix="fctest-")
+    os.makedirs(defs.DEFAULT_TEST_SESSION_ROOT_PATH, exist_ok=True)
+    return defs.DEFAULT_TEST_SESSION_ROOT_PATH
 
-    yield root_path
 
-    shutil.rmtree(root_path)
+@pytest.fixture(autouse=True, scope='session')
+def test_fc_session_root_path():
+    """Ensure and yield the fc session root directory.
+
+    Create a unique temporary session directory. This is important, since the
+    scheduler will run multiple pytest sessions concurrently.
+    """
+    fc_session_root_path = tempfile.mkdtemp(
+        prefix="fctest-",
+        dir=f"{test_session_root_path()}"
+    )
+    yield fc_session_root_path
+    shutil.rmtree(fc_session_root_path)
 
 
 @pytest.fixture
-def test_session_tmp_path(test_session_root_path):
+def test_session_tmp_path(test_fc_session_root_path):
     """Yield a random temporary directory. Destroyed on teardown."""
     # pylint: disable=redefined-outer-name
     # The fixture pattern causes a pylint false positive for that rule.
 
-    tmp_path = tempfile.mkdtemp(prefix=test_session_root_path)
+    tmp_path = tempfile.mkdtemp(prefix=test_fc_session_root_path)
     yield tmp_path
     shutil.rmtree(tmp_path)
 
@@ -225,7 +221,7 @@ def _gcc_compile(src_file, output_file, extra_flags="-static -O3"):
 
 
 @pytest.fixture(scope='session')
-def bin_cloner_path(test_session_root_path):
+def bin_cloner_path(test_fc_session_root_path):
     """Build a binary that `clone`s into the jailer.
 
     It's necessary because Python doesn't interface well with the `clone()`
@@ -233,7 +229,7 @@ def bin_cloner_path(test_session_root_path):
     """
     # pylint: disable=redefined-outer-name
     # The fixture pattern causes a pylint false positive for that rule.
-    cloner_bin_path = os.path.join(test_session_root_path, 'newpid_cloner')
+    cloner_bin_path = os.path.join(test_fc_session_root_path, 'newpid_cloner')
     _gcc_compile(
         'host_tools/newpid_cloner.c',
         cloner_bin_path
@@ -242,12 +238,12 @@ def bin_cloner_path(test_session_root_path):
 
 
 @pytest.fixture(scope='session')
-def bin_vsock_path(test_session_root_path):
+def bin_vsock_path(test_fc_session_root_path):
     """Build a simple vsock client/server application."""
     # pylint: disable=redefined-outer-name
     # The fixture pattern causes a pylint false positive for that rule.
     vsock_helper_bin_path = os.path.join(
-        test_session_root_path,
+        test_fc_session_root_path,
         'vsock_helper'
     )
     _gcc_compile(
@@ -258,11 +254,11 @@ def bin_vsock_path(test_session_root_path):
 
 
 @pytest.fixture(scope='session')
-def change_net_config_space_bin(test_session_root_path):
+def change_net_config_space_bin(test_fc_session_root_path):
     """Build a binary that changes the MMIO config space."""
     # pylint: disable=redefined-outer-name
     change_net_config_space_bin = os.path.join(
-        test_session_root_path,
+        test_fc_session_root_path,
         'change_net_config_space'
     )
     _gcc_compile(
@@ -274,7 +270,7 @@ def change_net_config_space_bin(test_session_root_path):
 
 
 @pytest.fixture(scope='session')
-def bin_seccomp_paths(test_session_root_path):
+def bin_seccomp_paths(test_fc_session_root_path):
     """Build jailers and jailed binaries to test seccomp.
 
     They currently consist of:
@@ -287,7 +283,7 @@ def bin_seccomp_paths(test_session_root_path):
     # pylint: disable=redefined-outer-name
     # The fixture pattern causes a pylint false positive for that rule.
     seccomp_build_path = os.path.join(
-        test_session_root_path,
+        test_fc_session_root_path,
         build_tools.CARGO_RELEASE_REL_PATH
     )
 
@@ -298,7 +294,7 @@ def bin_seccomp_paths(test_session_root_path):
                             src_dir='integration_tests/security/demo_seccomp')
 
     release_binaries_path = os.path.join(
-        test_session_root_path,
+        test_fc_session_root_path,
         build_tools.CARGO_RELEASE_REL_PATH,
         build_tools.RELEASE_BINARIES_REL_PATH
     )
@@ -337,17 +333,17 @@ def bin_seccomp_paths(test_session_root_path):
 
 
 @pytest.fixture()
-def microvm(test_session_root_path, bin_cloner_path):
+def microvm(test_fc_session_root_path, bin_cloner_path):
     """Instantiate a microvm."""
     # pylint: disable=redefined-outer-name
     # The fixture pattern causes a pylint false positive for that rule.
 
     # Make sure the necessary binaries are there before instantiating the
     # microvm.
-    vm = init_microvm(test_session_root_path, bin_cloner_path)
+    vm = init_microvm(test_fc_session_root_path, bin_cloner_path)
     yield vm
     vm.kill()
-    shutil.rmtree(os.path.join(test_session_root_path, vm.id))
+    shutil.rmtree(os.path.join(test_fc_session_root_path, vm.id))
 
 
 @pytest.fixture
@@ -382,7 +378,7 @@ def test_microvm_any(request, microvm):
 
 @pytest.fixture
 def test_multiple_microvms(
-        test_session_root_path,
+        test_fc_session_root_path,
         context,
         bin_cloner_path
 ):
@@ -400,7 +396,7 @@ def test_multiple_microvms(
 
     # When the context specifies multiple microvms, we use the first vm to
     # populate the other ones by hardlinking its resources.
-    first_vm = init_microvm(test_session_root_path, bin_cloner_path)
+    first_vm = init_microvm(test_fc_session_root_path, bin_cloner_path)
     MICROVM_S3_FETCHER.init_vm_resources(
         microvm_resources,
         first_vm
@@ -411,7 +407,7 @@ def test_multiple_microvms(
     # asserts that the `how_many` parameter is always positive
     # (i.e strictly greater than 0).
     for _ in range(how_many - 1):
-        vm = init_microvm(test_session_root_path, bin_cloner_path)
+        vm = init_microvm(test_fc_session_root_path, bin_cloner_path)
         MICROVM_S3_FETCHER.hardlink_vm_resources(
             microvm_resources,
             first_vm,
@@ -423,7 +419,7 @@ def test_multiple_microvms(
 
     for i in range(how_many):
         microvms[i].kill()
-        shutil.rmtree(os.path.join(test_session_root_path, microvms[i].id))
+        shutil.rmtree(os.path.join(test_fc_session_root_path, microvms[i].id))
 
 
 def pytest_generate_tests(metafunc):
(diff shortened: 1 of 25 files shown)


alindima:
we should also modify or check the host block device scheduler. Having a different scheduler would also influence the results

tests/framework/statistics/consumer.py
```diff
@@ -77,14 +77,16 @@ def _validate(self):
         """
         for ms_name in self._statistics_defs:
             if ms_name not in self._measurements_defs:
-                return False
```
raduiliescu, line 80: For the current PR this is ok, but I am not sure if this is the best way in terms of extensibility. This function will either return true or assert. There is no case for false. Maybe in the future you will want to reuse it, and this behaviors is quite strict. I would use a print or return a string, where empty string will be ok. This way the upper level function can decide if it wants to assert or not.

I am not blocking the PR on this. Is just food for thought.

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,541 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
raduiliescu, line 1: Copyright 2020

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,541 @@
...
+                CLAT_READ_STDDEV, CLAT_READ, st.function.Stddev),
+            st.consumer.StatisticDef(
+                LAT_READ_MEAN, LAT_READ, st.function.Avg,
+                criteria.EqualWith(
+                    baseline_lat_read["target"],
+                    baseline_lat_read["target"] *
+                    baseline_lat_read["delta_percentage"] / 100)),
+            st.consumer.StatisticDef(LAT_READ_STDDEV, LAT_READ,
+                                     st.function.Stddev)]
+    return [
```
raduiliescu, line 154: Can't this return list statements made generic, like a matrix where you say how do you want to calculate each param and then you just walk through the matrix and add the stuff? The code here is pretty hard to read.

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,541 @@
...
+                                 st.function.Stddev),
+        st.consumer.StatisticDef(LAT_READ_MEAN, LAT_READ, st.function.Avg),
+        st.consumer.StatisticDef(LAT_READ_STDDEV, LAT_READ,
+                                 st.function.Stddev)]
+
+
+def get_rw_stats(env_id, fio_id, cpu=None):
+    """Return read/write statistics."""
+    stats = get_read_stats(env_id, fio_id, cpu)
+    if cpu:
```
raduiliescu, line 174: Don't fully understand the differences between if cpu and if not cpu.  Can you please explain?

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,541 @@
...
+            if DEBUG:
+                result[CPU_UTILIZATION_VMM_SAMPLES_TAG] = data
+
+        vcpus_util = 0
+        for vcpu in range(basevm.vcpus_count):
+            # We expect a single fc_vcpu thread tagged with
+            # f`fc_vcpu {vcpu}`.
+            tag = f"fc_vcpu {vcpu}"
+            assert tag in cpu_load and len(cpu_load[tag]) == 1
+            for thread_id in cpu_load[tag]:
```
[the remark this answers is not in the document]
alindima: it's a dict, so `thread_id = list(cpu_load[tag])[0]`

tests/framework/statistics/consumer.py
alindima: Another way would be to have an extra configuration to this function -> `assert=True`

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,541 @@
...
+from framework.artifacts import ArtifactCollection, ArtifactSet
+from framework.builder import MicrovmBuilder
+from framework.matrix import TestContext, TestMatrix
+from framework.statistics import core, criteria
+from framework.utils import get_cpu_percent, CmdBuilder, eager_map, \
+    get_cpu_model_name
+import host_tools.drive as drive_tools
+import host_tools.network as net_tools  # pylint: disable=import-error
+import framework.statistics as st
+from integration_tests.performance.block_performance_test_config import \
```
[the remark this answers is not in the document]
alindima: Tried it. Unfortunately, using `import *`, pylint will give us a warning. This could be avoided by adding this comment: `# pylint: disable=W0401`.

This is not enough, since pytlint also throws errors, stating that every indentifier in the imported file is undefined. We could mitigate by this adding another supression comment.

However, I wouldn't do it, since this would make pylint ignore any other undefined variable error (some that should actually be errors, e.g.: `print(sadasnjkldsfndsksncksd)`

tests/integration_tests/performance/block_performance_test_config.py
```diff
@@ -0,0 +1,971 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
gc-plp: We should not store them at all in this format. 

It's difficult to read and difficult to update the values.

tests/integration_tests/performance/test_block_performance.py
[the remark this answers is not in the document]
gc-plp: Just use `from integration_tests.performance.block_performance_test_config as test_config`.

Then you can use `test_config.FIO.something`.

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,541 @@
...
+    """Define metrics based on the mode."""
+    ms = get_cpu_utilization_measurements()
+    if mode.endswith("read"):
+        ms.extend(get_read_measurements())
+        return ms
+    ms.extend(get_rw_measurements())
+    return ms
+
+
+def get_cpu_utilization_stats(env_id, fio_id, cpu=None):
```
gc-plp, line 74: Shouldn't we have a generic function for measuring CPU utilization and feeding the expected values?

I don't really understand why we need to define CPU usage comparisons criteria since these should be universal.

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,541 @@
...
+    """Return cpu utilization statistics."""
+    if cpu:
+        baseline_cpu_util_vmm = cpu["baseline_cpu_utilization"][env_id]["vmm"][
```
gc-plp, line 77: I don't think that using nested dicts is the way to go here.

We should rethink this.

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,541 @@
...
+def get_rw_stats(env_id, fio_id, cpu=None):
+    """Return read/write statistics."""
+    stats = get_read_stats(env_id, fio_id, cpu)
+    if cpu:
+        baseline_bw_write = cpu["baseline_bw_write"][env_id][fio_id]
+        baseline_iops_write = cpu["baseline_iops_write"][env_id][fio_id]
+        baseline_slat_write = cpu["baseline_slat_write"][env_id][fio_id]
+        baseline_clat_write = cpu["baseline_clat_write"][env_id][fio_id]
+        baseline_lat_write = cpu["baseline_lat_write"][env_id][fio_id]
+        stats.extend([
```
gc-plp, line 180: This block of code is hard to read and understand due to the mix of indentations and multiple elements being extended.

tests/integration_tests/performance/test_block_performance.py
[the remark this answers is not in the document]
gc-plp: Also, as a rule of thumb, keep imported symbols to a minimum. If multiple symbols need to be imported from one file, use the import name as a prefix as stated above.

Otherwise we add too many symbols in the current module and their origins become hard to track.

gc-plp (commented):
The overall test achieves its goal, but there are a few things to consider before merging this. I left a few comments above, but I think that a summary is needed:

1. The huge block_performance_test_config.py file needs to be changed into something else. Ideally this should act as a backing store for a class based approach that reads this data. Storing it as a Python file that also needs nice formatting to make the linters play cool is not creating an easy to maintain code base.
We should explore other options here. A CSV file perhaps? Breaking it down into multiple files?

2. Repeated use of nested dicts. Use carefully. Ideally use a class based approach.

3. A file that contains a test or tests should only define the high level steps that the test does in a very clear way. Other dependencies should be layered either in the framework that supports the test or secondary test files that do not define tests.

4. The statistics framework itself needs a lot of explicit elements that it has to be given. For example having to always append things to the `StatisticDef` member. Could the framework identify what elements it needs by reading the test baseline definition?
I'm thinking that the huge statistics file could instead define a bunch of key variables that tell the framework how the tata in that file is read and used. Right now we always need to trickle down information from the top level, but that information is very specific to test cases and could be also built from the ground up.

tests/integration_tests/performance/block_performance_test_config.py
```diff
@@ -0,0 +1,971 @@
...
+
+DEBUG = True
+TEST_ID = "block_device_performance"
+FIO = "fio"
+CONFIG = {
+    "block_device_size": 2048,  # MB
+    "time": 100,  # seconds
+    "omit": 10,  # seconds
+    "load_factor": 1,
+    "fio_blk_sizes": [4096],  # 1024, 512, 65536? bytes
```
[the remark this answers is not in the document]
iulianbarbu: Yes. I am trying to merge baselines in baby steps. Also, 4k block size is the most common in disks benchmarking. I think we can assume this is the goto relevant measurement to be used to assess virtio block device performance, for regression analysis.

tests/integration_tests/performance/block_performance_test_config.py
```diff
@@ -0,0 +1,971 @@
...
+    "hosts": {
+        "instances": {
+            "m5d.metal": {
+                "cpus": [
+                    {
+                        "model": "Intel(R) Xeon(R) Platinum 8259CL CPU @ "
+                                 "2.50GHz",
+                        'baseline_bw_read': {
+                            'vmlinux-4.14.bin/ubuntu-18.04.ext4': {
+                                'randrw-bs4096-2vcpu': {'target': 186782,
```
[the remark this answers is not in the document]
iulianbarbu: I am not aware of any reasons whatsoever. This test was adapted from a previous existing one, which was not interested in write only benchmarking. Having dedicated cases for other scenarios might be a follow up after merging this test, which is the foundation really. The scenarios can be really diverse, so it is a matter of what we find interesting or relevant for customers use cases.

tests/host_tools/network.py
```diff
@@ -59,6 +59,23 @@ def scp_file(self, local_path, remote_path):
         else:
             utils.run_cmd(cmd)
 
+    def scp_get_file(self, remote_path, local_path):
+        """Copy a files to the VM using scp."""
```
alindima, line 63: should it be `"""Copy a file from the VM using scp"""`? I see that there is already a function in this file that handles copying to the VM

tests/integration_tests/performance/test_vsock_throughput.py
```diff
@@ -13,83 +13,89 @@
...
     """Define the produced measurements for VSOCK workloads."""
     return [types.MeasurementDef(THROUGHPUT, THROUGHPUT_UNIT),
             types.MeasurementDef(DURATION, DURATION_UNIT),
-            types.MeasurementDef(CPU_UTILIZATION_HOST, CPU_UTILIZATION_UNIT),
-            types.MeasurementDef(CPU_UTILIZATION_GUEST, CPU_UTILIZATION_UNIT)]
+            types.MeasurementDef.cpu_utilization_vmm(),
+            types.MeasurementDef.cpu_utilization_vcpus_total()]
+
+
+def basic_vsock_stats():
```
alindima, line 39: maybe we should rename the function. the stats are not basic, they just don't have pass criteria

tests/integration_tests/performance/test_vsock_throughput.py
```diff
@@ -13,83 +13,89 @@
...
+            types.StatisticDef.avg(DURATION),
+            types.StatisticDef.get_first_observation(
+                st_name="value",
+                ms_name=CPU_UTILIZATION_VMM),
+            types.StatisticDef.get_first_observation(
+                st_name="value",
+                ms_name=CPU_UTILIZATION_VCPUS_TOTAL)]
+
+
+def vsock_stats(cpu_baselines: str, iperf3_id: str, env_id: str):
```
alindima, line 55: i'd say to remove this function and just paste the code in the caller, the same way as before.
either that or rename it to `stats_with_criteria` (and the `basic_vsock_stats` function to `stats_without_criteria`)

alindima (commented):
Great work with the framework refactoring 🚀

tests/framework/statistics/core.py
```diff
@@ -68,6 +69,9 @@ def run_exercise(self) -> Statistics:
             if len(custom) > 0:
                 self._statistics['custom'][tag] = custom
 
+        if not self._check:
```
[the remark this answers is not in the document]
raduiliescu: That would be really nice, but if that is not trivial to do, can we make it in a followup PR.

tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -13,81 +13,93 @@
...
 from framework.utils_cpuid import get_cpu_model_name
 import host_tools.network as net_tools
-from integration_tests.performance.network_tcp_throughput_test_config import \
-    THROUGHPUT, DURATION, RETRANSMITS, CPU_UTILIZATION_HOST, \
-    CPU_UTILIZATION_GUEST, CONFIG, CPU_UTILIZATION_VMM_TAG, \
-    CPU_UTILIZATION_VCPUS_TOTAL_TAG, THROUGHPUT_TOTAL, RETRANSMITS_TOTAL, \
-    DURATION_TOTAL, IPERF3, BASE_PORT, DEBUG, \
-    IPERF3_CPU_UTILIZATION_PERCENT_OUT_TAG, IPERF3_END_RESULTS_TAG, \
-    DEBUG_CPU_UTILIZATION_VMM_SAMPLES_TAG, DELTA_PERCENTAGE_TAG, TARGET_TAG
+from integration_tests.performance.configs.network_tcp_throughput_test_config \
```
gc-plp, line 21: Can we just use
```python
import integration_tests.performance.configs.network_tcp_throughput_test_config as test_cfg
```

And further down just use `test_cfg.THROUGHPUT` and so on.

WDYT?

tests/framework/statistics/core.py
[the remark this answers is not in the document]
iulianbarbu: Yes. I would focus on this in another PR.

tests/conftest.py
```diff
@@ -87,30 +87,15 @@ def test_with_any_microvm(test_microvm_any):
 import host_tools.cargo_build as build_tools
 import host_tools.network as net_tools
 import host_tools.proc as proc
-
-from framework.artifacts import ArtifactCollection
 import framework.utils as utils
+from framework.defs import SCRIPT_FOLDER, ENV_TEST_IMAGES_S3_BUCKET, \
```
gc-plp, line 91: Reduce clutter by importing `framework.defs` as an alias.

tests/framework/defs.py
```diff
@@ -20,3 +22,15 @@
...
+SPEC_S3_BUCKET = 'spec.ccfc.min'
+"""The s3 bucket that holds global Firecracker specifications."""
+DEFAULT_TEST_IMAGES_S3_BUCKET = 'spec.ccfc.min'
+"""The default s3 bucket that holds Firecracker microvm test images."""
+ENV_TEST_IMAGES_S3_BUCKET = 'TEST_MICROVM_IMAGES_S3_BUCKET'
+"""Environment variable for configuring the test microvm s3 bucket.
+If variable exists in `os.environ`, its value will be used as the s3 bucket
+for microvm test images.
+"""
+SCRIPT_FOLDER = os.path.dirname(os.path.realpath(__file__))
```
gc-plp, line 34: This variable definition is not robust as it references a different path than before, introducing breaking changes in the CI suite (e.g. gitlint test).

tests/integration_tests/performance/configs/block_performance_test_config_raw.json
```diff
@@ -0,0 +1,789 @@
...
+                  }
+                }
+              }
+            }
+          }
+        ]
+      }
+    }
+  }
+}
```
alindima, line 789: add a newline here

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,395 @@
...
+    import DEBUG, FIO, CONFIG, CPU_UTILIZATION_VMM_SAMPLES_TAG, FIO_OUTPUT, \
+    TEST_ID, CPU_UTILIZATION_VCPUS_TOTAL, IOPERSEC_UNIT, KIBPERSEC_UNIT, \
+    BLK_BASELINE_PROVIDER, IOPS, BW, CPU_UTILIZATION_VMM
+
+
+READ_OP = "read"
+WRITE_OP = "write"
+
+
+def get_cpu_utilization_measurements():
```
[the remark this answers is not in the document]
alindima: I see, the names are clearer now 👍🏻

iulianbarbu (commented):
@gc-plp  and @alindima , I've managed to produce stable baselines and to address your comments. Please take a look again! Thanks for the patience. :)

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -0,0 +1,472 @@
...
+SLAT = "slat_{}"
+READ_OP = "read"
+WRITE_OP = "write"
+CPU_UTILIZATION_VMM = DefaultMeasurement.CPU_UTILIZATION_VMM.name.lower()
+CPU_UTILIZATION_VMM_SAMPLES_TAG = f"{CPU_UTILIZATION_VMM}_samples"
+CPU_UTILIZATION_VCPUS_TOTAL = \
+    f"{DefaultMeasurement.CPU_UTILIZATION_VCPUS_TOTAL.name.lower()}"
+
+# Measurements units.
+FIO_OUTPUT = "fio_output"
```
alindima, line 48: this constant does not seem to be used

alindima (commented):
Looks good, just some minor comments

tests/conftest.py
```diff
@@ -87,30 +87,18 @@ def test_with_any_microvm(test_microvm_any):
...
-"""The default s3 bucket that holds Firecracker microvm test images."""
-
-ENV_TEST_IMAGES_S3_BUCKET = 'TEST_MICROVM_IMAGES_S3_BUCKET'
-"""Environment variable for configuring the test microvm s3 bucket.
-
-If variable exists in `os.environ`, its value will be used as the s3 bucket
-for microvm test images.
-"""
-
+# Environment variable for configuring the test microvm s3 bucket.
```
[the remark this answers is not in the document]
iulianbarbu: Yes. PTAL!

tests/conftest.py
```diff
@@ -188,28 +176,38 @@ def pytest_addoption(parser):
     return PytestScheduler.instance().do_pytest_addoption(parser)
 
 
-@pytest.fixture(autouse=True, scope='session')
 def test_session_root_path():
-    """Ensure and yield the testrun root directory.
+    """Ensure and return the testrun session root directory exists.
```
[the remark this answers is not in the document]
iulianbarbu: Done.

tests/integration_tests/performance/test_block_performance.py
iulianbarbu: Done.

alindima (approved):
LGTM, great job 🚀

gc-plp (approved):

Issue #2293, opened 2020-11-23
[Code improvement] deduplicate literal HTTP responses in tests


acatangiu:
There are many tests with literal hardcoded HTTP responses that bloat the code. Some of them even have data embedded in them, making those tests hard to maintain.

Example possible deduplication:

in https://github.com/firecracker-microvm/firecracker/blob/e8200f3c3eaba014220e447e8d426c8cf8607eec/src/api_server/src/parsed_request.rs#L511 :

```diff
-        let expected_response = format!(
-            "HTTP/1.1 200 \r\n\
-             Server: Firecracker API\r\n\
-             Connection: keep-alive\r\n\
-             Content-Type: application/json\r\n\
-             Content-Length: 122\r\n\r\n{}",
-            VmConfig::default().to_string()
-        );
+        let body = VmConfig::default().to_string();
+        let expected_response = http_response_ok(&body);
        );
```

where `http_response_ok()` could be reused in all tests.

Example `http_response_ok()` definition:

```rust
fn http_response_ok(body: &str) {
    format!(
        "HTTP/1.1 200 \r\n\
         Server: Firecracker API\r\n\
         Connection: keep-alive\r\n\
         Content-Type: application/json\r\n\
         Content-Length: {}\r\n\r\n{}",
         status_code,
         body.len(),
         body,
    )
}
```

EvanJP:
Hi!  I would like to be assigned to this if possible as a first issue to explore the code base (one of the many UT Austin students looking for issues to help out on)

EvanJP:
Would we want to be parsing the (firecracker.yaml)[https://github.com/firecracker-microvm/firecracker/blob/master/src/api_server/swagger/firecracker.yaml] file in case the API changes, or should we just keep it simple and hard-code the function definitions first? 

An example function definition might be `fn call_machine-config(request: Request, ht_enabled: bool, mem_size_mib: i32, track_dirty_pages: Option<bool>, vcpu_count: i32) -> String` where `Request` is simply an enum for `PATCH, POST,` etc.

If the usage of `Option` for the optional params seem good to y'all I'll go ahead with implementing these.

acatangiu:
I don't follow. I'm guessing you're talking about further improvements, this issue is very simple and only tracks deduplicating all the http static/hardcoded contents.

Example: define a function:
```rust
fn http_response_ok(body: &str) {
    format!(
        "HTTP/1.1 200 \r\n\
         Server: Firecracker API\r\n\
[...]

PR #2340, opened 2020-12-10
dedupe literal HTTP response and request strings

EvanJP:
## Reason for This PR

Deduplicate hard coded HTTP responses and requests in tests as per #2293.

## Description of Changes

- In `src/api_server/src/parsed_request.rs` testing:
    - Remove hard-coded literal HTTP responses and requests in tests.
    - `fn http_response(body: &str, status_code: i32) -> String` is introduced
        - Given the body and string, format to fit HTTP response string
    - `fn http_request(request_type: &str, end_point: &str, body: &str) -> String` is introduced
        - Given the request type, end point, and content fit to HTTP request string.
             - Every instance of this uses `.as_bytes()` in the testing, but I left it to return a String for consistency sake.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/api_server/src/parsed_request.rs
+++ b/src/api_server/src/parsed_request.rs
@@ -292,16 +292,53 @@ pub(crate) mod tests {
         }
     }
 
+    fn http_response(body: &str, status_code: i32) -> String {
+        let header = format!(
+            "HTTP/1.1 {} \r\n\
+             Server: Firecracker API\r\n\
+             Connection: keep-alive\r\n",
+            status_code
+        );
+        if status_code == 204 {
+            // No Content
+            return format!("{}{}", header, "\r\n");
+        } else {
+            let content = format!(
+                "Content-Type: application/json\r\n\
+                Content-Length: {}\r\n\r\n{}",
+                body.len(),
+                body,
+            );
+
+            format!("{}{}", header, content)
+        }
+    }
+
+    fn http_request(request_type: &str, endpoint: &str, body: Option<&str>) -> String {
+        let req_no_body = format!(
+            "{} {} HTTP/1.1\r\n\
+            Content-Type: application/json\r\n",
+            request_type, endpoint
+        );
+        if body.is_some() {
+            return format!(
+                "{}\
+                Content-Length: {}\r\n\r\n\
+                {}",
+                req_no_body,
+                body.unwrap().len(),
+                body.unwrap()
+            );
+        }
+        return format!("{}\r\n", req_no_body,);
+    }
+
     #[test]
     fn test_missing_slash() {
         let (mut sender, receiver) = UnixStream::pair().unwrap();
         let mut connection = HttpConnection::new(receiver);
         sender
-            .write_all(
-                b"GET none HTTP/1.1\r\n\
-                Content-Type: text/plain\r\n\
-                Content-Length: 4\r\n\r\nbody",
-            )
+            .write_all(http_request("GET", "none", Some("body")).as_bytes())
             .unwrap();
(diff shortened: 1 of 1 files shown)


src/api_server/src/parsed_request.rs
```diff
@@ -294,15 +294,45 @@ pub(crate) mod tests {
...
+    }
+
     #[test]
     fn test_missing_slash() {
         let (mut sender, receiver) = UnixStream::pair().unwrap();
         let mut connection = HttpConnection::new(receiver);
-        sender
-            .write_all(
-                b"GET none HTTP/1.1\r\n\
-                Content-Type: text/plain\r\n\
```
berciuliviu, line 304: Here, "text/plain" is used as the Content-Type, while the **http_request** function will default it to 'application/json' for every case. For this to be correct and retain the same behaviour, I believe we should also be able to set the content type.

src/api_server/src/parsed_request.rs
```diff
@@ -1014,11 +941,10 @@ pub(crate) mod tests {
     fn test_try_from_patch_mmds() {
         let (mut sender, receiver) = UnixStream::pair().unwrap();
         let mut connection = HttpConnection::new(receiver);
+        let body = "{}";
```
berciuliviu, line 913: NIT:

Since the body is basically "{}", I believe we could send it directly as a parameter.

http_request("PATCH", "/mmds", **"{}"**).as_bytes()

alindima:
Please also squash the `cargo fmt` commit into the previous one

src/api_server/src/parsed_request.rs
```diff
@@ -294,16 +294,47 @@ pub(crate) mod tests {
...
+             Content-Length: {}\r\n\r\n{}",
+            body.len(),
+            body,
+        );
+
+        if status_code == 204 {
+            // No Content
+            return format!("{}{}", header, "\r\n");
+        }
+        format!("{}{}", header, content)
```
acatangiu, line 316: No point in building `content` if not used, also the preferred style for conditional returning is `if/else` block:
```suggestion
        if status_code == 204 {
            // No Content
            format!("{}\r\n", header)
        } else {
            let content = format!(
                "Content-Type: application/json\r\n\
                 Content-Length: {}\r\n\r\n{}",
                body.len(),
                body,
            );
            format!("{}{}", header, content)
        }
```

src/api_server/src/parsed_request.rs
acatangiu: if `Content-Type: application/json` also works in all tests, I say we use that as the default and stop using `Content-Type: text/plain`

acatangiu (approved):

georgepisaltu:
@EvanJP can you please take a look at the review comments and address them?

dianpopa:
@acatangiu @berciuliviu I addressed all your concerns/suggestions. Please take another look!

src/api_server/src/parsed_request.rs
```diff
@@ -292,16 +292,53 @@ pub(crate) mod tests {
...
+    fn http_response(body: &str, status_code: i32) -> String {
+        let header = format!(
+            "HTTP/1.1 {} \r\n\
+             Server: Firecracker API\r\n\
+             Connection: keep-alive\r\n",
+            status_code
+        );
+        if status_code == 204 {
+            // No Content
+            return format!("{}{}", header, "\r\n");
```
acatangiu, line 304: nit:
```suggestion
            format!("{}{}", header, "\r\n")
```

acatangiu (approved):

berciuliviu (approved):

PR #2341, opened 2020-12-11
Mask XSAVE PKRU bit in C2/T3 templates

sandreim:
## Reason for This PR

Fixes an inconsistency in the templates - the OSPKE feature bit is masked, but the XSAVE PKRU bit is not. Snapshots created on a CPU with OSPKE from a T2/C3 uVM would fail to load on a CPU without OSPKE (Broadwell): https://elixir.bootlin.com/linux/latest/source/arch/x86/kvm/x86.c#L4299.

## Description of Changes

Provided in commit message.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/cpuid/src/cpu_leaf.rs
+++ b/src/cpuid/src/cpu_leaf.rs
@@ -240,6 +240,7 @@ pub mod leaf_0xd {
 
             pub const MPX_STATE_BITRANGE: BitRange = bit_range!(4, 3);
             pub const AVX512_STATE_BITRANGE: BitRange = bit_range!(7, 5);
+            pub const PKRU_BITINDEX: u32 = 9;
         }
     }
 
--- a/src/cpuid/src/template/intel/c3.rs
+++ b/src/cpuid/src/template/intel/c3.rs
@@ -123,6 +123,10 @@ fn update_xsave_features_entry(
         entry
             .eax
             .write_bits_in_range(&index0::eax::AVX512_STATE_BITRANGE, 0);
+
+        // OSPKE is masked in leaf_0x7 index 0 - RDPKRU/WRPKRU not exposed.
+        // Here we mask the XSAVE PKRU capabilities.
+        entry.eax.write_bit(index0::eax::PKRU_BITINDEX, false);
     }
 
     if entry.index == 1 {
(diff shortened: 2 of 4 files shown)


src/cpuid/src/cpu_leaf.rs
```diff
@@ -240,6 +240,7 @@ pub mod leaf_0xd {
 
             pub const MPX_STATE_BITRANGE: BitRange = bit_range!(4, 3);
             pub const AVX512_STATE_BITRANGE: BitRange = bit_range!(7, 5);
+            pub const PKRU_SHIFT: u32 = 9;
```
serban300, line 243: Nit: I would rename this to `PKRU_BITINDEX` in order to stay consistent with the other variables.

dianpopa:
It is not clear to me if we want to hide the feature also on hardware that has support for it?

andreeaflorescu (commented):
Good catch on the inconsistency! You should also update the integration test with the newly added features in the template so that we also get a regression test: https://github.com/firecracker-microvm/firecracker/blob/master/tests/integration_tests/functional/test_cpu_features.py#L281

Just wanted to call out that for other features we might need to update the restore path to only set the features that are available on the cpu, as masking unavailable features on both templates would make the templates useless.

LGTM, will approve once you also update the integration test.

src/cpuid/src/cpu_leaf.rs
sandreim: done

sandreim:
> Good catch on the inconsistency! You should also update the integration test with the newly added features in the template so that we also get a regression test: https://github.com/firecracker-microvm/firecracker/blob/master/tests/integration_tests/functional/test_cpu_features.py#L281
> 
There is no new feature added, just disabling the XCR0 PKRU bit. I've updated the existing template integration tests to look out for the PKRU bit.

> Just wanted to call out that for other features we might need to update the restore path to only set the features that are available on the cpu, as masking unavailable features on both templates would make the templates useless.
> 
 What CPUID we expose to the guest must remain the same during the microvm lifecycle which includes snapshots. Changing the exposed features on restore might break workloads that were running/using them when the snapshot was created.

> LGTM, will approve once you also update the integration test.

Please take another look.

andreeaflorescu (approved):

serban300 (approved):

PR #2350, opened 2020-12-14
Catch enosys

AlexandruCihodaru:
## Reason for This PR

ENOSYS was not handled and they caused `unexpected` errors.
## Description of Changes

Added handler for ENOSYS on aarch64.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/vstate/vcpu/mod.rs
+++ b/src/vmm/src/vstate/vcpu/mod.rs
@@ -509,6 +509,13 @@ impl Vcpu {
                         // Notify that this KVM_RUN was interrupted.
                         Ok(VcpuEmulation::Interrupted)
                     }
+                    libc::ENOSYS => {
+                        METRICS.vcpu.failures.inc();
+                        error!(
+                            "Received ENOSYS error because KVM failed to emulate an instruction."
+                        );
+                        Err(Error::FaultyKvmExit(format!("{}", e)))
+                    }
                     _ => {
                         METRICS.vcpu.failures.inc();
                         error!("Failure during vcpu run: {}", e);
--- a/tests/integration_tests/functional/test_error_code.py
+++ b/tests/integration_tests/functional/test_error_code.py
@@ -1,0 +1,43 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Tests scenarios for Firecracker kvm exit handling."""
+
+import os
+import platform
+import pytest
+from framework.utils import wait_process_termination
+
+
+@pytest.mark.skipif(
+    platform.machine() != "aarch64",
+    reason="The error code returned on aarch64 will not be returned on x86 "
+           "under the same conditions."
+)
+def test_enosys_error_code(test_microvm_with_initrd):
+    """Test that ENOSYS error is caught and firecracker exits gracefully."""
+    # On aarch64 we trigger this error by adding to initrd a C program that
+    # maps a file into memory and then tries to load the content from an
+    # offset in the file bigger than its length into a register asm volatile
+    # ("ldr %0, [%1], 4" : "=r" (ret), "+r" (buf));
+    vm = test_microvm_with_initrd
+    vm.jailer.daemonize = False
+    vm.spawn()
+    vm.memory_monitor = None
+
+    vm.initrd_file = os.path.join(vm.path, "fsfiles/", "initrd_enosys.img")
+    vm.basic_config(
+        add_root_device=False,
+        vcpu_count=1,
+        boot_args='console=ttyS0 reboot=k panic=1 pci=off',
+        use_initrd=True
+    )
+
+    vm.start()
+
+    # Check if FC process is closed
+    wait_process_termination(vm.jailer_clone_pid)
+
+    log_data = vm.log_data
+    assert "Received ENOSYS error because KVM failed to emulate " \
+           "an instruction." in log_data
+    assert "Vmm is stopping." in log_data
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -23,7 +23,7 @@ import host_tools.proc as proc
 # this contains the frequency while on AMD it does not.
 # Checkout the cpuid crate. In the future other
 # differences may appear.
-COVERAGE_DICT = {"Intel": 85.2, "AMD": 84.43, "ARM": 83.10}
+COVERAGE_DICT = {"Intel": 85.15, "AMD": 84.37, "ARM": 83.03}
 
 PROC_MODEL = proc.proc_type()
 


tests/integration_tests/functional/test_error_code.py
```diff
@@ -0,0 +1,42 @@
...
+
+import os
+import time
+import platform
+import pytest
+
+
+@pytest.mark.skipif(
+    platform.machine() == "x86_64",
+    reason="We only test with a kernel for aarch64"
```
alindima, line 13: `reason` should express why we skip the test for x86_64, and this is not clear.
why do we do this?

alindima: can't we also trigger this behaviour on x86?

tests/integration_tests/functional/test_error_code.py
```diff
@@ -0,0 +1,42 @@
...
+def test_enosys_error_code(test_microvm_with_ssh):
+    """Test that ENOSYS error_code doesn't kill the microVM."""
+    test_microvm = test_microvm_with_ssh
+    test_microvm.spawn()
+
+    # We don't need to monitor the memory for this test.
+    test_microvm.memory_monitor = None
+
+    test_microvm.kernel_file = \
+        test_microvm.kernel_file[:-11] + "enosys_vmlinux.bin"
```
alindima, line 24: as far as I can tell, this is essentially pop()-ing the last identifier from the `kernel_file` path and replacing it . the logic is brittle and we should replace it with some `os.path` method calls

alindima, line 24: we should also have a comment saying what this kernel does.
if we want to update this kernel further down the line, we should know how to re-add the specific faulty instruction

src/vmm/src/vstate/vcpu/mod.rs
```diff
@@ -509,6 +509,13 @@ impl Vcpu {
                         // Notify that this KVM_RUN was interrupted.
                         Ok(VcpuEmulation::Interrupted)
                     }
+                    libc::ENOSYS => {
+                        METRICS.vcpu.failures.inc();
+                        error!(
+                            "Received ENOSYS error because KVM failed to emulate an instruction"
```
alindima, line 515:
```suggestion
                            "Received ENOSYS error because KVM failed to emulate an instruction."
```

tests/integration_tests/functional/test_error_code.py
```diff
@@ -0,0 +1,42 @@
...
+    test_microvm.basic_config(vcpu_count=1)
+
+    test_microvm.start()
+
+    firecracker_pid = test_microvm.jailer_clone_pid
+
+    while True:
```
alindima, line 31: I've seen this logic in multiple tests. It would be nice to have a helper method for it. Not blocking the PR for this, though.

tests/integration_tests/functional/test_error_code.py
```diff
@@ -0,0 +1,35 @@
...
+        try:
+            os.kill(firecracker_pid, 0)
+            time.sleep(0.01)
+        except OSError:
+            break
+
+    if platform.machine() != "x86_64":
+        log_data = test_microvm.log_data
+        assert "Received ENOSYS error because KVM failed to emulate an instruction" in log_data
+        assert "Vmm is stopping." in log_data
```
berciuliviu: This integration test is implemented here and the next commit adds changes to it. Shouldn't it be better to implement & manage it in only one commit?

tests/integration_tests/functional/test_error_code.py
```diff
@@ -29,7 +36,7 @@ def test_enosys_error_code(test_microvm_with_ssh, network_config):
...
             break
 
-    if platform.machine() != "x86_64":
-        log_data = test_microvm.log_data
-        assert "Received ENOSYS error because KVM failed to emulate an instruction" in log_data
-        assert "Vmm is stopping." in log_data
+    log_data = test_microvm.log_data
+    assert "Received ENOSYS error because KVM failed to emulate " \
+           "an instruction" in log_data
+    assert "Vmm is stopping." in log_data
```
berciuliviu, line 42: Doesn't this check over here mean that the VMM is killed after logging the expected error? If so, the description at line 16 should specify that the purpose of this test is to have the error logged before exiting the VMM.

berciuliviu (commented):
The third "merge" commit could be removed if you rebase the branch and add your changes on top of it. Something like "git pull --rebase upstream master" could work in this case.

tests/integration_tests/functional/test_error_code.py
sandreim: Instead of adding another kernel binary, can we just set the first instruction that is going to be executed to something that cannot be emulated and results in ENOSYS ?

AlexandruCihodaru: The kernel used for this test is one randomly generated. Also how the error code is handled is not influenced in any way by the kernel version. In my opinion we should create a new issue and try to reproduce what @sandreim  suggested in this comment https://github.com/firecracker-microvm/firecracker/pull/2350#discussion_r543330663

tests/integration_tests/functional/test_error_code.py
AlexandruCihodaru: As far as I know we can, but the behavior would be the same. The portion of code that is handling this error code does not care about architecture.

alindima: This is a weak argument. We run many tests that should have the same behaviour across all platforms. Just because something should happen, doesn't mean it will.

alindima: At the very least, create an issue for this and link it in the `reason`

dianpopa:
We are currently blocked trying to find a way of finding a way of validating this PR on x86_64. What we need to realize is that we might spend quite some time trying to do this. Let me explain why:
As per the code changes done by Alex, you can see that this error (ENOSYS) is returned by KVM_RUN. KVM_RUN (i.e [kvm_arch_vcpu_ioctl_run](https://elixir.bootlin.com/linux/v4.4/source/virt/kvm/kvm_main.c#L2375) is not only dependent on the platform and but it also varies so much from platform to platform (https://elixir.bootlin.com/linux/v4.4/source/arch/x86/kvm/x86.c#L6783, https://elixir.bootlin.com/linux/v4.4/source/arch/arm/kvm/arm.c#L532). So this means that an `ENOSYS` returned by `KVM_RUN` on arm  in some conditions will not trigger ENOSYS in the same conditions. We might spend quite some time searching for the conditions which trigger ENOSYS on x86.
As a consequence, what I propose here is to just focus on having the test run on arm64 (the platform for which we know how to trigger it). @berciuliviu @alindima WDYT?

alindima:
> We are currently blocked trying to find a way of finding a way of validating this PR on x86_64. What we need to realize is that we might spend quite some time trying to do this. Let me explain why:
> As per the code changes done by Alex, you can see that this error (ENOSYS) is returned by KVM_RUN. KVM_RUN (i.e [kvm_arch_vcpu_ioctl_run](https://elixir.bootlin.com/linux/v4.4/source/virt/kvm/kvm_main.c#L2375) is not only dependent on the platform and but it also varies so much from platform to platform (https://elixir.bootlin.com/linux/v4.4/source/arch/x86/kvm/x86.c#L6783, https://elixir.bootlin.com/linux/v4.4/source/arch/arm/kvm/arm.c#L532). So this means that an `ENOSYS` returned by `KVM_RUN` on arm in some conditions will not trigger ENOSYS in the same conditions. We might spend quite some time searching for the conditions which trigger ENOSYS on x86.
> As a consequence, what I propose here is to just focus on having the test run on arm64 (the platform for which we know how to trigger it). @berciuliviu @alindima WDYT?

I'm fine with this, thanks for the explanation. We should document this in the aarch64 test

berciuliviu:
> We are currently blocked trying to find a way of finding a way of validating this PR on x86_64. What we need to realize is that we might spend quite some time trying to do this. Let me explain why:
> As per the code changes done by Alex, you can see that this error (ENOSYS) is returned by KVM_RUN. KVM_RUN (i.e [kvm_arch_vcpu_ioctl_run](https://elixir.bootlin.com/linux/v4.4/source/virt/kvm/kvm_main.c#L2375) is not only dependent on the platform and but it also varies so much from platform to platform (https://elixir.bootlin.com/linux/v4.4/source/arch/x86/kvm/x86.c#L6783, https://elixir.bootlin.com/linux/v4.4/source/arch/arm/kvm/arm.c#L532). So this means that an `ENOSYS` returned by `KVM_RUN` on arm in some conditions will not trigger ENOSYS in the same conditions. We might spend quite some time searching for the conditions which trigger ENOSYS on x86.
> As a consequence, what I propose here is to just focus on having the test run on arm64 (the platform for which we know how to trigger it). @berciuliviu @alindima WDYT?

I am also fine with it. As @alindima proposed, I consider it a good suggestion to document the explanation :)

acatangiu:
I think all this effort going into a test that is supposed to validate an error message is not worth it.
The `ENOSYS` kvm return code error message can be covered by a simple unit test that works on all platforms the same.

raduiliescu:
> 
> 
> I think all this effort going into a test that is supposed to validate an error message is not worth it.
> The `ENOSYS` kvm return code error message can be covered by a simple unit test that works on all platforms the same.

I would keep the current integration test for ARM, specify why is not working for x86 and add the unit test.

berciuliviu (dismissed):

tests/integration_tests/functional/test_error_code.py
dianpopa: You can use `wait_process_termination` from ``framework.utils`

alindima (dismissed):

berciuliviu (dismissed):

alindima (approved):

berciuliviu (approved):

PR #2314, opened 2020-12-01
Add option to logger to log thread names

Preston-Sundar:
Per issue firecracker-microvm#2117, added optional parameter to
allow logging of the current thread name in the prefix.
Added a rust unit test that spawns  a named thread and logs i
ts name in the prefix. Added python integration tests that
create logs with new show_thread_name parameter.

Signed-off-by: sundar.preston.789@gmail.com <sundar.preston.789@gmail.com>

Co-authored-by: Kevin Guo <kev.guo123@gmail.com>
Co-authored-by: Yichen Zhang <jakezhang2020@gmail.com>

## Reason for This PR

Addresses issue #2117 

## Description of Changes

* Modified `src/logger/src/logger.rs` to optionally allow appending the thread name to the log prefix. Enabled by setting `show_thread_name` to true. Also added unit test that creates a thread with custom name and checks the log's prefix for that name.
* Modified `src/vmm/src/vmm_config/logger.rs` so that `LoggerConfig`'s signature matches changes made to the logger.
* Added argument `show-thread-name` to the argument parser in `src/firecracker/src/main.rs`.
* Minor changes made to `tests/framework/resources.py`, `src/vmm/src/rpc_interface.rs`, and `src/api_server/src/request/logger.rs` to correctly call the logger with the new parameter.
* Added documentation changes in `docs/device-api.md`, `docs/logger.md`, and `src/api_server/swagger/firecracker.yaml` that correspond to the new parameter. 

- [ ] This functionality can be added in [`rust-vmm`](https://github.com/rust-vmm).

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria.]`
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [ ] All commits in this PR are signed (`git commit -s`).
- [ ] The reason for this PR is clearly provided (issue no. or explanation).
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this PR.
- [ ] Any newly added `unsafe` code is properly documented.
- [ ] Any API changes are reflected in `firecracker/swagger.yaml`.
- [ ] Any user-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.

Issue #2117, opened 2020-09-04
Improve logger to include thread name (vcpu#/api/vmm)


sandreim:
Multiple threads concurrently write the log file and the source of the logs is not obvious. If we add the thread id the logs would become more readable.

HQ01:
interested to help out!

sandreim:
Hey @HQ01! Thank you for your interest in contributing to Firecracker. 

I will try to summarize some next steps to get you started:
- Dive deep the Firecracker `logger` crate  which  provides an implementation for the API abstraction by the [log crate](https://crates.io/crates/log)
- Change the code so the log prefix also include thread ids. It should live in `Logger::create_prefix(&self, record: &Record)` 
- Would be nice if you find a solution to use the thread names instead of ids like: **vmm, api, vcpu#1, vcpu#2** ..
- Write unit and integration tests
- Publish PR

If you get stuck along the way please reach out to us via Slack or this ticket.

HQ01:
Hi @sandreim, thanks for the information! To provide you with more context, we are 4 students in UT Austin's virtualization class, and look forward to contributing to an open-source project as part of our coursework -- Firecracker really attracts us as an exciting project to work on :) We will start from relatively simple issues, but hopefully we can fix them all one by one. Will keep you posted as we start working on them!

2018kguo:
Hey @sandreim, just as a quick status update, we recently started coding this up and have added functionality to Logger::create_prefix such that it has an option to include the thread name in the prefix, We have added a few unit tests and have about to start writing integration tests. Things are still in a pretty rough state, but this is our current branch: [here](https://github.com/Preston-Sundar/firecracker/tree/Feature-Improve-logger-to-include-thread-name).

Preston-Sundar:
PR opened at https://github.com/firecracker-microvm/firecracker/pull/2314

acatangiu:
Fixed by #2352

PR #2352, opened 2020-12-15
Add thread name to logger prefix.

Preston-Sundar:
As per issue #2117, add capability to logger to log current
thread name in the prefix. Add unit test that spawns named
thread and checks the logged prefix for that name.

Signed-off-by: sundar.preston.789@gmail.com <sundar.preston.789@gmail.com>

Co-authored-by: Kevin Guo <kev.guo123@gmail.com>

## Reason for This PR

Addresses issue #2117 
Submitted earlier PR #2314, but accidentally messed up git rebase. :cry:.

## Description of Changes

* Modified `src/logger/src/logger.rs` to always append the thread name to the log prefix. Also added unit test that creates a thread with custom name and checks the log's prefix for that name.
* Updated `tests/integration_tests/functional/test_logging.py` to reflect newly added thread names in log prefix.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/logger/src/logger.rs
+++ b/src/logger/src/logger.rs
@@ -83,6 +83,7 @@ use std::io::{sink, stderr, stdout, Write};
 use std::result;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Mutex, RwLock};
+use std::thread;
 
 use crate::metrics::{IncMetric, METRICS};
 use lazy_static::lazy_static;
@@ -244,6 +245,11 @@ impl Logger {
         self
     }
 
+    /// Get the current thread's name.
+    fn get_thread_name(&self) -> String {
+        thread::current().name().unwrap_or("-").to_string()
+    }
+
     /// Creates the first portion (to the left of the separator)
     /// of the log statement based on the logger settings.
     fn create_prefix(&self, record: &Record) -> String {
@@ -253,6 +259,9 @@ impl Logger {
         if !instance_id.is_empty() {
             prefix.push(instance_id.to_string());
         }
+
+        // Attach current thread name to prefix.
+        prefix.push(self.get_thread_name());
 
         if self.show_level() {
             prefix.push(record.level().to_string());
@@ -527,6 +536,7 @@ mod tests {
     #[test]
     fn test_configure() {
         let logger = Logger::new();
+        let crnt_thread_name = logger.get_thread_name();
 
         // Assert that `configure()` can be called successfully any number of times.
         assert!(logger.configure(Some(TEST_INSTANCE_ID.to_string())).is_ok());
@@ -546,14 +556,17 @@ mod tests {
         logger.mock_log(Level::Info, "info");
         validate_log(
             &mut Box::new(&mut reader),
-            "[TEST-INSTANCE-ID:INFO:logger.rs:0] info\n",
+            &format!(
+                "[TEST-INSTANCE-ID:{}:INFO:logger.rs:0] info\n",
+                crnt_thread_name
+            ),
         );
     }
 
     #[test]
     fn test_init() {
         let logger = Logger::new();
-
+        let crnt_thread_name = logger.get_thread_name();
         // Assert that the first call to `init()` is successful.
         let (writer, mut reader) = log_channel();
         logger.set_instance_id(TEST_INSTANCE_ID.to_string());
@@ -568,7 +581,10 @@ mod tests {
         logger.mock_log(Level::Info, "info");
         validate_log(
             &mut Box::new(&mut reader),
-            "[TEST-INSTANCE-ID:INFO:logger.rs:0] info\n",
+            &format!(
+                "[TEST-INSTANCE-ID:{}:INFO:logger.rs:0] info\n",
+                crnt_thread_name
+            ),
         );
 
         // Assert that initialization works only once.
@@ -580,7 +596,10 @@ mod tests {
         logger.mock_log(Level::Info, "info");
         validate_log(
             &mut Box::new(&mut reader),
-            "[TEST-INSTANCE-ID:INFO:logger.rs:0] info\n",
+            &format!(
+                "[TEST-INSTANCE-ID:{}:INFO:logger.rs:0] info\n",
+                crnt_thread_name
+            ),
         );
         validate_log(&mut Box::new(&mut reader_2), "");
     }
@@ -589,7 +608,7 @@ mod tests {
     fn test_create_prefix() {
         let logger = Logger::mock_new();
         let mut reader = logger.mock_init();
-
+        let crnt_thread_name = logger.get_thread_name();
         // Test with empty instance id.
         logger.set_instance_id("".to_string());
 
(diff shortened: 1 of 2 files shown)


tests/integration_tests/functional/test_logging.py
```diff
@@ -55,6 +55,7 @@ def check_log_message_format(log_str, instance_id, level, show_level,
     strptime(timestamp, "%Y-%m-%dT%H:%M:%S")
```
berciuliviu, line 55: The description for this function should also be updated to match the new addition to the log format.

src/logger/src/logger.rs
```diff
@@ -645,10 +692,38 @@ mod tests {
...
+                "[TEST-INSTANCE-ID:{}:WARN:logger.rs:0] msg\n",
+                crnt_thread_name
+            ),
         );
     }
 
+    #[test]
+    fn test_thread_name_custom() {
+        // Name this thread "custom-thread"
+        let custom_thread_1 = thread::Builder::new()
```
berciuliviu, line 705: NIT:

Could name it simply to custom_thread.

src/logger/src/logger.rs
```diff
@@ -645,10 +692,38 @@ mod tests {
...
+            &format!(
+                "[TEST-INSTANCE-ID:{}:WARN:logger.rs:0] msg\n",
+                crnt_thread_name
+            ),
         );
     }
 
+    #[test]
+    fn test_thread_name_custom() {
+        // Name this thread "custom-thread"
```
berciuliviu, line 704: This comment can be removed, since the name of the test is pretty self-explanatory.

src/logger/src/logger.rs
```diff
@@ -517,6 +527,16 @@ mod tests {
...
     }
 
+    fn get_thread_name() -> String {
+        let handle = thread::current();
+        let thread_name;
+        match handle.name() {
+            Some(x) => thread_name = x.to_string(),
+            None => thread_name = "-".to_string(),
+        }
+        thread_name
```
berciuliviu, line 537: To make this shorter, thread_name can be removed and you could return **x.to_string()** or **"-".to_string()** directly from the match case. The same could be said about line 261. We could push the new strings directly without the need of a new variable.

Preston-Sundar:
@berciuliviu Addressed requested changes in latest commit.

src/logger/src/logger.rs
```diff
@@ -527,7 +543,7 @@ mod tests {
     #[test]
     fn test_configure() {
         let logger = Logger::new();
-
+        let crnt_thread_name = get_thread_name();
```
berciuliviu, line 546: NIT:
A new line should be added after this line.

src/logger/src/logger.rs
```diff
@@ -254,6 +255,13 @@ impl Logger {
             prefix.push(instance_id.to_string());
         }
 
+        // Attach current thread name to prefix.
+        let handle = thread::current();
+        match handle.name() {
+            Some(x) => prefix.push(x.to_string()),
```
berciuliviu, line 261: NIT: 
Some(x) -> Some(name), both here and below

src/logger/src/logger.rs
```diff
@@ -645,10 +688,37 @@ mod tests {
...
+            ),
         );
     }
 
+    #[test]
+    fn test_thread_name_custom() {
+        let custom_thread = thread::Builder::new()
+            .name("custom-thread".to_string())
+            .spawn(move || {
+                // test to check the thread name
```
berciuliviu, line 703: Comment can be removed.

src/logger/src/logger.rs
```diff
@@ -645,10 +688,37 @@ mod tests {
...
+                let logger = Logger::mock_new();
+                let mut reader = logger.mock_init();
+                // Test with a mock instance id.
```
berciuliviu, line 706: Is there a special reason we want to set the instance id here? If not, we could remove setting it and let the test verify only if we log the custom name or not, as the test name says :) .

tests/integration_tests/functional/test_logging.py
```diff
@@ -44,16 +44,17 @@ def check_log_message_format(log_str, instance_id, level, show_level,
     Parse the string representing the logs and look for the parts
     that should be there.
     The log line should look lie this:
-         YYYY-MM-DDTHH:MM:SS.NNNNNNNNN [ID:LEVEL:FILE:LINE] MESSAGE
+         YYYY-MM-DDTHH:MM:SS.NNNNNNNNN [ID:THREAD:LEVEL:FILE:LINE] MESSAGE
     where LEVEL and FILE:LINE are both optional.
-    e.g.:
-    `2018-09-09T12:52:00.123456789 [MYID:WARN:/path/to/file.rs:52] warning`
+    e.g. with THREAD as "t":
+    `2018-09-09T12:52:00.123456789 [MYID:t:WARN:/path/to/file.rs:52] warning`
```
berciuliviu, line 50: **t** could be replaced with "thread-name"/"threadname" and `with THREAD as "t"` removed, since we also don't specify a set value for the other parameters beforehand.

src/logger/src/logger.rs
```diff
@@ -254,6 +255,13 @@ impl Logger {
             prefix.push(instance_id.to_string());
         }
 
+        // Attach current thread name to prefix.
+        let handle = thread::current();
+        match handle.name() {
+            Some(x) => prefix.push(x.to_string()),
+            None => prefix.push("-".to_string()),
```
berciuliviu, line 262: I am thinking about the decision of setting the default thread name to "-". Citing the Rust documentation, ["By default, spawned threads are unnamed"](https://doc.rust-lang.org/std/thread/#naming-threads), it could mean that we could very well ignore it if no thread name is provided. @sandreim may I have your input on this?

berciuliviu (commented):
The patch looks good. Since `integration_tests/build/test_coverage.py` test has to be updated for aarch64, I have also added a few more suggestions to the code.

src/logger/src/logger.rs
```diff
@@ -517,6 +525,14 @@ mod tests {
         );
     }
 
+    fn get_thread_name() -> String {
```
acatangiu, line 528: move this function out of `mod tests` and use it in `create_prefix()` too.

tests/integration_tests/functional/test_logging.py
Preston-Sundar: Good point, however adding either of those strings pushes line 50 above the 80-character line limit. Is it okay to break the example logger string on line 50 into two lines?

berciuliviu: Good observation. In this case, keeping the format you decided in the first place is a way to go. I'd rather see the log on one line if we can keep at that. I would modify it a little as "with THREAD NAME as **TN**" and it should look good in my opinion.

src/logger/src/logger.rs
Preston-Sundar: When we set up the initial PR, we used an optional parameter to add the thread name after the instance id. We wanted to test the prefix formatting a little more and hence the remnants of that test. You're right! Now that we test the thread name on all the other unit tests, it's better to simplify this one.

tests/integration_tests/functional/test_logging.py
Preston-Sundar: Agreed! "TN" is more readable

Preston-Sundar:
@berciuliviu @acatangiu Latest commit addresses new change requests.

Preston-Sundar:
@berciuliviu @acatangiu Latest commit addresses requested changes.

src/logger/src/logger.rs
```diff
@@ -244,6 +245,15 @@ impl Logger {
...
     }
 
+    /// Get the current thread's name.
+    fn get_thread_name(&self) -> String {
+        let handle = thread::current();
+        match handle.name() {
+            Some(name) => name.to_string(),
+            None => "-".to_string(),
+        }
+    }
```
acatangiu, line 255: You can use `&str` instead of `String` to avoid the overhead of building `String` objects if not needed:

```suggestion
    fn get_thread_name(&self) -> &str {
        let handle = thread::current();
        match handle.name() {
            Some(name) => name,
            None => "-",
        }
    }
```
 
Which can be shortened to:

```suggestion
    fn get_thread_name(&self) -> &str {
        thread::current().name().unwrap_or("-")
    }
```

Preston-Sundar: Thank you for the suggestion, we added this to the latest commit! 

However, wouldn't the suggested code, specifically this: 
```rust
fn get_thread_name(&self) -> &str {
    thread::current().name().unwrap_or("-")
}
```
create a temporary reference? So when its called in `create_prefix`, a `"returns a value referencing data owned by the current function"` error is thrown. 

Instead we added the following code:
```rust
fn get_thread_name(&self) -> String {
    thread::current().name().unwrap_or("-").to_string()
}
```
 Unfortunately, It still creates the String object. Any suggestions around temporary reference problem?

Preston-Sundar:
@berciuliviu @acatangiu Latest commit addresses suggested changes

src/logger/src/logger.rs
berciuliviu: I do not think we can get around without using a String here. What Sundar said is correct. We also use String objects for any part of the prefix,  storing them in a **Vec<String>**, before joining them into a bigger String at the end of the function. I believe it is okay to leave it as it is.

src/logger/src/logger.rs
```diff
@@ -546,14 +556,17 @@ mod tests {
...
+                crnt_thread_name
+            ),
         );
     }
 
     #[test]
     fn test_init() {
         let logger = Logger::new();
-
+        let crnt_thread_name = logger.get_thread_name();
```
berciuliviu, line 569: NIT: New line after this line.

berciuliviu (approved):
Looks good to me.

acatangiu (approved):

PR #2364, opened 2020-12-17
Fix block used bytes reporting

iulianbarbu:
## Reason for This PR

Fix the reported used bytes for any virtio block request.

## Description of Changes

Report the correct/actual bytes written/read from the virtio block buffers, which are accounted as used, and reported back to driver.

## License Acceptance

## PR Checklist

~~- [] Any newly added `unsafe` code is properly documented.~~

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -19,7 +19,7 @@ use logger::{error, warn, IncMetric, METRICS};
 use rate_limiter::{BucketUpdate, RateLimiter, TokenType};
 use utils::eventfd::EventFd;
 use virtio_gen::virtio_blk::*;
-use vm_memory::{Bytes, GuestMemoryMmap};
+use vm_memory::{Bytes, GuestMemoryError, GuestMemoryMmap};
 
 use super::{
     super::{ActivateResult, DeviceState, Queue, VirtioDevice, TYPE_BLOCK, VIRTIO_MMIO_INT_VRING},
@@ -258,15 +258,46 @@ impl Block {
                             break;
                         }
                     }
+
                     let status = match request.execute(&mut self.disk, mem) {
                         Ok(l) => {
-                            len = l;
-                            VIRTIO_BLK_S_OK
+                            // Account for the status byte as well.
+                            // With a non-faulty driver, we shouldn't get to the point where we
+                            // overflow here (since data len must be a multiple of 512 bytes, so
+                            // it can't be u32::MAX). In the future, this should be fixed at the
+                            // request parsing level, so no data will actually be transferred in
+                            // scenarios like this one.
+                            if let Some(l) = l.checked_add(1) {
+                                len = l;
+                                VIRTIO_BLK_S_OK
+                            } else {
+                                len = l;
+                                VIRTIO_BLK_S_IOERR
+                            }
                         }
                         Err(e) => {
-                            error!("Failed to execute request: {:?}", e);
                             METRICS.block.invalid_reqs_count.inc();
-                            len = 1; // We need at least 1 byte for the status.
+                            match e {
+                                ExecuteError::Read(GuestMemoryError::PartialBuffer {
+                                    completed,
+                                    expected,
+                                }) => {
+                                    error!(
+                                        "Failed to execute virtio block read request: can only \
+                                        write {} of {} bytes.",
+                                        completed, expected
+                                    );
+                                    METRICS.block.read_bytes.add(completed);
+                                    // This can not overflow since `completed` < data len which is
+                                    // an u32.
+                                    len = completed as u32 + 1;
+                                }
+                                _ => {
+                                    error!("Failed to execute virtio block request: {:?}", e);
+                                    // Status byte only.
+                                    len = 1;
+                                }
+                            };
                             e.status()
                         }
                     };
@@ -690,7 +721,7 @@ pub(crate) mod tests {
 
             assert_eq!(vq.used.idx.get(), 1);
             assert_eq!(vq.used.ring[0].get().id, 0);
-            assert_eq!(vq.used.ring[0].get().len, 0);
+            assert_eq!(vq.used.ring[0].get().len, 1);
             assert_eq!(mem.read_obj::<u32>(status_addr).unwrap(), VIRTIO_BLK_S_OK);
         }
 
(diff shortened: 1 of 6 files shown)


docs/api_requests/patch-block.md
```diff
@@ -9,6 +9,21 @@ guest at the time of the API call, otherwise the call will silently fail -
 no error is returned from either the guest or the host, but the guest might end
 up in an inconsistent state.
 
+# Virtio block underlying host file resize operations
+
+The design of the PATCH /drives API guaranties sending updates for previous
+underlying host file changes to the guest and that no change is done by the API
+call to the underlying host file.
```
acatangiu, line 16:
```suggestion
The design of the PATCH /drives API does not mutate the host backing file.
It updates the guest emulated block device with the current host backing file properties.
```

docs/api_requests/patch-block.md
```diff
@@ -9,6 +9,21 @@ guest at the time of the API call, otherwise the call will silently fail -
...
+
+To this end, the sequence of resizing the virtio block underlying host file
+followed by a PATCH /drives API call is not an atomic operation. For a reliable
+and secure experience on host and guest we recommend that any change to the
+virtio block devices underlying host files to be followed by a corresponding
+Firecracker block device rescan API. Between these two operations, it is
+recommended to avoid/restrict any virtio block device usage on the guest.
+If needed, snapshot pause/resume API can be used to pause the guest activity
```
acatangiu, line 24: Pause/Resume are snapshot-independent commands.

```suggestion
If needed, pause/resume API can be used to pause the guest activity
```

tests/integration_tests/functional/test_drives.py
```diff
@@ -41,8 +43,19 @@ def test_rescan_file(test_microvm_with_ssh, network_config):
...
 
-    # Resize the filesystem from 256 MiB (default) to 512 MiB.
-    fs.resize(512)
+    # Resize the filesystem from 256 MiB (default) to 128 MiB.
+    truncated_size = 1
+    utils.run_cmd('truncate --size ' + str(truncated_size) + 'M ' + fs.path)
+
+    # Check if reading from the entire disk results in a file of the same size
+    # or errors out.
+    utils.run_cmd('truncate --size ' + str(truncated_size) + 'M ' + fs.path)
```
acatangiu, line 52: Leftover line:

```suggestion
```

acatangiu (commented):
I have a couple of wording comments, but otherwise PR looks good!

iulianbarbu (commented):
@acatangiu , I've addressed your review comments. Please take a look again!

sandreim:
Looks good to me, still need to fix the ARM coverage.

src/devices/src/virtio/block/request.rs
```diff
@@ -214,37 +214,40 @@ impl Request {
...
+                    METRICS.block.read_count.inc();
+                    self.data_len
+                })
+                .map_err(ExecuteError::Read),
+            RequestType::Out => mem
+                .write_all_to(self.data_addr, diskfile, self.data_len as usize)
+                .map(|_| {
+                    METRICS.block.write_bytes.add(self.data_len as usize);
+                    METRICS.block.write_count.inc();
+                    self.data_len
```
lauralt, line 230: Hmmm, why are we now returning `self.data_len` for `Write` request? The `len` field of an entry from the used ring should contain the number of bytes written to memory, not of the ones read from there. For`Write` request only the status byte is written to memory.

src/devices/src/virtio/block/request.rs
```diff
@@ -214,37 +214,40 @@ impl Request {
...
+                })
+                .map_err(ExecuteError::Flush),
             RequestType::GetDeviceID => {
                 let disk_id = disk.image_id();
                 if (self.data_len as usize) < disk_id.len() {
                     return Err(ExecuteError::BadRequest(Error::InvalidOffset));
                 }
                 mem.write_slice(disk_id, self.data_addr)
-                    .map_err(ExecuteError::Write)?;
+                    .map(|_| 0)
```
lauralt, line 246: Why are we mapping `Ok` to `0`? For `GetDeviceId` requests, the device is writing to memory the device id, so the number of bytes is VIRTIO_BLK_ID_BYTES.

src/devices/src/virtio/block/request.rs
iulianbarbu: Good catch! Updated.

src/devices/src/virtio/block/request.rs
iulianbarbu: Good catch again! :)) Updated.

sandreim: I think we also need to cover GetDeviceID in an integration test.

iulianbarbu: This is covered at the moment through an unit test. We would benefit for sure from an integration test as well, but I would not add it in this PR. WDYT?

sandreim (approved):

src/devices/src/virtio/block/device.rs
```diff
@@ -713,8 +732,68 @@ pub(crate) mod tests {
...
+
+            mem.write_obj::<u32>(VIRTIO_BLK_T_IN, request_type_addr)
+                .unwrap();
+            vq.dtable[1]
+                .flags
+                .set(VIRTQ_DESC_F_NEXT | VIRTQ_DESC_F_WRITE);
+
+            let size = block.disk.file.seek(SeekFrom::End(0)).unwrap();
+            block.disk.file.set_len(size / 2).unwrap();
+            mem.write_obj(5, GuestAddress(request_type_addr.0 + 8))
```
lauralt, line 783: Can you add a comment here/somewhere about `request_type_addr.0 + 8` being the address where the sector number is stored? (It's not at all obvious the difference between this test case scenario and the previous one)?

Also, a comment why the length from the used ring entry = 1 would be also very useful (you would expect that for a partial buffer error on read, `completed` to be > 0, not = 0).

acatangiu (commented):
I picked up driving this PR since @iulianbarbu is unavailable.

src/devices/src/virtio/block/device.rs
```diff
@@ -258,15 +258,35 @@ impl Block {
...
                             METRICS.block.invalid_reqs_count.inc();
-                            len = 1; // We need at least 1 byte for the status.
+                            match e {
+                                ExecuteError::Read(GuestMemoryError::PartialBuffer {
+                                    completed,
+                                    expected,
+                                }) => {
+                                    error!(
+                                        "Failed to execute virito block read request due to \
+                                       being able writing {} bytes to device buffers instead of {}.",
```
georgepisaltu, line 277: nit:
```suggestion
                                       being able to write {} bytes to device buffers instead of {}.",
```

georgepisaltu (approved):

src/devices/src/virtio/block/device.rs
lauralt: We can also test the partial buffer error with `completed` > 0.
For example:

```rust
            vq.used.idx.set(0);
            set_queue(&mut block, 0, vq.create_queue());

            mem.write_obj::<u32>(VIRTIO_BLK_T_IN, request_type_addr)
                .unwrap();
            vq.dtable[1]
                .flags
                .set(VIRTQ_DESC_F_NEXT | VIRTQ_DESC_F_WRITE);
            vq.dtable[1].len.set(1024);

            let size = block.disk.file.seek(SeekFrom::End(0)).unwrap();
            block.disk.file.set_len(size / 2).unwrap();
            mem.write_obj(3, GuestAddress(request_type_addr.0 + 8))
                .unwrap();

            invoke_handler_for_queue_event(&mut block);

            assert_eq!(vq.used.idx.get(), 1);
            assert_eq!(vq.used.ring[0].get().id, 0);

            // File has 4 sectors and we try to read from sector 3, which means we will read
            // 512 bytes (instead of 1024).
            assert_eq!(vq.used.ring[0].get().len, 513);
            assert_eq!(
                mem.read_obj::<u32>(status_addr).unwrap(),
                VIRTIO_BLK_S_IOERR
            );
```

src/devices/src/virtio/block/device.rs
```diff
@@ -258,15 +258,34 @@ impl Block {
                             break;
                         }
                     }
+
                     let status = match request.execute(&mut self.disk, mem) {
                         Ok(l) => {
-                            len = l;
+                            // Account for the status descriptor as well.
```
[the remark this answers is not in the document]
lauralt: > rephrased comment
> 
> didn't use `checked_add` since `len` or `l` cannot come close to `u32::MAX` bytes. That would mean single IO `read`/`write` of at least 4GB which is unlikely/not possible in normal usecases.
> 
> In case of naughty guest driver creating requests > 4GB what is the effect of overflow here? reported `len` smaller than actually written to virtio buffers - I understand that this is allowed by the virtio protocol, right?
> 
It is allowed indeed, but I still see this as an error case, since the driver should never send a request that will cause the `len` field from the used ring overflow an `u32` (it is no more naughty, but rather faulty). Also, since we are having a very simple implementation of the block device and we always assume that there is only one data descriptor in the request (which the spec doesn't enforce in any way, but it is rather a Linux driver particularity), this `len` from here can only be <= u32::MAX. And keeping in mind the [following](https://github.com/oasis-tcs/virtio-spec/blob/master/content.tex#L4651), it seems pretty bad to accept as valid u32::MAX.

> If we were to add overflow checks just to be extra safe, I'm not sure what we could do in case of an overflow. Would we `unwrap`/`panic`? Would that be better?

I'm not sure what would be the best option here (since we never validate for example that data len can't be u32::MAX I wouldn't go with an unwrap here :-?).

src/devices/src/virtio/block/device.rs
acatangiu: @lauralt sounds good. Will you push this test in this PR or do you want me to do it?

src/devices/src/virtio/block/device.rs
acatangiu: done

src/devices/src/virtio/block/device.rs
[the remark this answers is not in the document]
acatangiu: done

src/devices/src/virtio/block/device.rs
```diff
@@ -258,15 +258,40 @@ impl Block {
...
+                                    expected,
+                                }) => {
+                                    error!(
+                                        "Failed to execute virtio block read request: can only \
+                                        write {} of {} bytes.",
+                                        completed, expected
+                                    );
+                                    METRICS.block.read_bytes.add(completed);
+                                    // Account for the status byte as well.
+                                    len = (completed as u32).checked_add(1).unwrap_or(1);
```
lauralt, line 287: `Unwrap()` should be safe here since completed < data len from the request.

src/devices/src/virtio/block/device.rs
```diff
@@ -258,15 +258,40 @@ impl Block {
...
                     let status = match request.execute(&mut self.disk, mem) {
                         Ok(l) => {
-                            len = l;
-                            VIRTIO_BLK_S_OK
+                            // Account for the status byte as well.
+                            if let Some(l) = l.checked_add(1) {
+                                len = l;
+                                VIRTIO_BLK_S_OK
+                            } else {
+                                len = 1;
```
lauralt, line 269: I don't think this fixes the issue :(. If we're on the `Ok(l)` path, it means `l` bytes were transferred. The problem is we shouldn't actually get to this point (Request::parse() should fail before we get to execute the command and basically, no bytes won't be transferred), but this is somehow another issue. For now, I would change this to `len = l` or use `expect()`.

src/devices/src/virtio/block/device.rs
lauralt: Sure, you can add it.

src/devices/src/virtio/block/device.rs
lauralt: Hmm, or `unchecked_add()` or we can leave it as it was before: len = completed + 1 and add a comment about why it can not overflow.

acatangiu: This is a nit. Please mark it as such.

src/devices/src/virtio/block/device.rs
acatangiu: With current code, a faulty driver will get status error byte which is correct from a protocol point of view right?

We don’t want to add process fatal error/panic conditions on code paths controlled by the guest.

src/devices/src/virtio/block/device.rs
lauralt: I didn't see it as a nit, since I don't agree to store 1 in used length for this particular issue (overflow). Moreover, if this overflow is never possible, it seemed great to avoid at all the `unwrap_or(1)` and simply use `unwrap()` or the other suggestions. Also, sorry for insisting on this :(.

lauralt (dismissed):

lauralt (approved):

acatangiu (dismissed):

sandreim (dismissed):

acatangiu (approved):

acatangiu:
@lauralt @sandreim had to fix coverage conflict, PTAL

lauralt (approved):

acatangiu:
Closing in favor of #2379

PR #2402, opened 2021-01-11
Removed sysconf dependency

### LINKED ISSUE #2187: Remove sysconf crate dependency
## Describe the bug

Given that we strive to keep the number of dependencies to a minimum, it seems that sysconf crate is an easy candidate as far as minimizing Firecracker dependencies. The `sysconf` crate is used to obtain the size of a page on the host with `sysconf::page::pagesize();` This dependency could be easily replaced with `libc::sysconf()`.

After issue is solved, check if Cargo.lock still contains "sysconf" as a dependency.

luminitavoicu:

# Reason for This PR

In Firecracker the sysconf crate is only used for fetching the system's page size. In order to keep the dependencies list to a minimum, sysconf could be removed.
Fixes #2187 

## Description of Changes

Replaced sysconf's `pagesize()` with `libc::sysconf(libc::_SC_PAGESIZE)`.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/memory_snapshot.rs
+++ b/src/vmm/src/memory_snapshot.rs
@@ -15,6 +15,7 @@ use vm_memory::{
 };
 
 use crate::DirtyBitmap;
+use utils::errno;
 
 /// State of a guest memory region saved to file/buffer.
 #[derive(Debug, PartialEq, Versionize)]
@@ -69,6 +70,8 @@ pub enum Error {
     CreateMemory(vm_memory::Error),
     /// Cannot create region.
     CreateRegion(vm_memory::mmap::MmapRegionError),
+    /// Cannot fetch system's page size.
+    PageSize(errno::Error),
     /// Cannot dump memory.
     WriteMemory(GuestMemoryError),
 }
@@ -80,6 +83,7 @@ impl Display for Error {
             FileHandle(err) => write!(f, "Cannot access file: {:?}", err),
             CreateMemory(err) => write!(f, "Cannot create memory: {:?}", err),
             CreateRegion(err) => write!(f, "Cannot create memory region: {:?}", err),
+            PageSize(err) => write!(f, "Cannot fetch system's page size: {:?}", err),
             WriteMemory(err) => write!(f, "Cannot dump memory: {:?}", err),
         }
     }
@@ -117,8 +121,8 @@ impl SnapshotMemory for GuestMemoryMmap {
         writer: &mut T,
         dirty_bitmap: &DirtyBitmap,
     ) -> std::result::Result<(), Error> {
-        let page_size = sysconf::page::pagesize();
         let mut writer_offset = 0;
+        let page_size = get_page_size()?;
 
         self.with_regions_mut(|slot, region| {
             let kvm_bitmap = dirty_bitmap.get(&slot).unwrap();
@@ -200,6 +204,13 @@ impl SnapshotMemory for GuestMemoryMmap {
     }
 }
 
+fn get_page_size() -> Result<usize, Error> {
+    match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
+        -1 => Err(Error::PageSize(errno::Error::last())),
+        ps => Ok(ps as usize),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::collections::HashMap;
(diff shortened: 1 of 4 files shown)


src/vmm/src/memory_snapshot.rs
```diff
@@ -117,8 +118,11 @@ impl SnapshotMemory for GuestMemoryMmap {
         writer: &mut T,
         dirty_bitmap: &DirtyBitmap,
     ) -> std::result::Result<(), Error> {
-        let page_size = sysconf::page::pagesize();
         let mut writer_offset = 0;
+        let page_size = match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
+            -1 => panic!("Failed to dump dirty pages: {}", errno::Error::last()),
```
sandreim, line 123: I think we need to return an error here instead of a panic.

luminitavoicu: I chose to panic here instead of throwing an error because this is the behavior of `sysconf::page::pagesize()`:

```
fn priv_pagesize() -> usize {
    // sysconf(_SC_PAGESIZE) is required by POSIX 1003.1: http://www.unix.com/man-page/posix/3p/sysconf/
    ::raw::sysconf(::raw::SysconfVariable::ScPagesize).expect("sysconf(_SC_PAGESIZE) failed, but _SC_PAGESIZE is required by POSIX 1003.1") as usize
}
```

 and also because `libc::sysconf` can only return error when the parameter is not an accepted one (which would never be the case here). 

However, it would make for a more consistent and elegant approach to raise an error here. I will change the behavior.

berciuliviu: I've observed the same "panic" behaviour inside mmap.rs - line 79 - **enable_dirty_page_tracking**. I wonder if it is worth changing it there also to return an error? From my short investigation, I'd say we leave it like that, that since I believe will involve more complex changes, but it is worth taking it into consideration.

berciuliviu (commented):
Looks good to me. Just remove the dot following commit title for the buildkite tests to pass.

src/vmm/src/memory_snapshot.rs
```diff
@@ -117,8 +121,11 @@ impl SnapshotMemory for GuestMemoryMmap {
         writer: &mut T,
         dirty_bitmap: &DirtyBitmap,
     ) -> std::result::Result<(), Error> {
-        let page_size = sysconf::page::pagesize();
         let mut writer_offset = 0;
+        let page_size = match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
```
dianpopa, line 125: I suggest to write a function for this and use that in the unit tests, WDYT?

tests/integration_tests/build/test_coverage.py
```diff
@@ -23,7 +23,7 @@
 # this contains the frequency while on AMD it does not.
 # Checkout the cpuid crate. In the future other
 # differences may appear.
-COVERAGE_DICT = {"Intel": 85.18, "AMD": 84.47, "ARM": 83.25}
+COVERAGE_DICT = {"Intel": 85.18, "AMD": 84.47, "ARM": 83.37}
```
dianpopa, line 26: 💯

src/vmm/src/memory_snapshot.rs
luminitavoicu: We did take that into consideration at that time and decided it would be best to stick with the `panic` approach, because the vm-memory proxy uses errors imported from the upstream vm-memory crate.

src/vmm/src/memory_snapshot.rs
luminitavoicu: Good idea!

dianpopa (dismissed):

src/vmm/src/memory_snapshot.rs
```diff
@@ -200,18 +207,23 @@ impl SnapshotMemory for GuestMemoryMmap {
     }
 }
 
+fn get_page_size() -> i64 {
+    unsafe { libc::sysconf(libc::_SC_PAGESIZE) }
+}
```
acatangiu, line 212: I would include _error parsing_ in this function instead of moving that to the caller:
```suggestion
fn get_page_size() -> Result<usize> {
    match unsafe { libc::sysconf(libc::_SC_PAGESIZE) } {
        -1 => Err(Error::PageSize(errno::Error::last())),
        ps => Ok(ps),
    }
}
```

dianpopa (dismissed):

src/vmm/src/memory_snapshot.rs
luminitavoicu: Done

dianpopa (approved):

berciuliviu (approved):

PR #2423, opened 2021-01-22
Support and tools for easier baseline gathering

alindima:
# Reason for This PR

There is currently no standard way of gathering baselines when it comes to long-running performance tests.
In order to gather baselines:

- One has to modify the test code in order to output the raw results.
- After a number of test iterations are run, the resulting data has to be parsed with manual written scripts into the right baselines.

## Description of Changes

- Add support for pytest `--dump-results-to-file`, which will transform into a `results_file_dumper` fixture used for writing raw test data in the `test_results` folder
- Added a `parse_baselines` script that allows for custom logic for parsing the raw data into baselines.
- Added implementations for `parse_baselines` for vsock and network tcp tests.
- Refactored the `vsock_throughput` and `network_tcp_throughput` performance tests to use the `results_file_dumper` fixture and slight refactor of the baselines format.
- Fixed the `test_licenses` test which failed for Amazon 2021 copyright year.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -144,11 +144,11 @@ def init_microvm(root_path, bin_cloner_path,
     assert jailer_binary
 
     vm = Microvm(
-         resource_path=root_path,
-         fc_binary_path=fc_binary,
-         jailer_binary_path=jailer_binary,
-         microvm_id=microvm_id,
-         bin_cloner_path=bin_cloner_path)
+        resource_path=root_path,
+        fc_binary_path=fc_binary,
+        jailer_binary_path=jailer_binary,
+        microvm_id=microvm_id,
+        bin_cloner_path=bin_cloner_path)
     vm.setup()
     return vm
 
@@ -166,11 +166,12 @@ def pytest_configure(config):
 
 
 def pytest_addoption(parser):
-    """Pytest hook. Add concurrency command line option.
-
-    For some reason, pytest doesn't properly pick up this hook in our plugin
-    class, so we need to call it from here.
-    """
+    """Pytest hook. Add command line options."""
+    parser.addoption(
+        "--dump-results-to-file",
+        action="store_true",
+        help="Flag to dump test results to the test_results folder.",
+    )
     return PytestScheduler.instance().do_pytest_addoption(parser)
 
 
@@ -208,6 +209,15 @@ def test_session_tmp_path(test_fc_session_root_path):
     tmp_path = tempfile.mkdtemp(prefix=test_fc_session_root_path)
     yield tmp_path
     shutil.rmtree(tmp_path)
+
+
+@pytest.fixture
+def results_file_dumper(request):
+    """Yield the custom --dump-results-to-file test flag."""
+    if request.config.getoption("--dump-results-to-file"):
+        return utils.ResultsFileDumper(request.node.originalname)
+
+    return None
 
 
 def _gcc_compile(src_file, output_file, extra_flags="-static -O3"):
--- a/tests/framework/defs.py
+++ b/tests/framework/defs.py
@@ -28,3 +28,5 @@ DEFAULT_TEST_IMAGES_S3_BUCKET = 'spec.ccfc.min'
 ENV_TEST_IMAGES_S3_BUCKET = 'TEST_MICROVM_IMAGES_S3_BUCKET'
 """Default test session root directory path"""
 DEFAULT_TEST_SESSION_ROOT_PATH = "/srv"
+"""Absolute path to the test results folder"""
+TEST_RESULTS_DIR = Path(FC_WORKSPACE_DIR) / "test_results"
--- a/tests/framework/utils.py
+++ b/tests/framework/utils.py
@@ -14,6 +14,7 @@ import time
 from collections import namedtuple, defaultdict
 import psutil
 from retry import retry
+from framework import defs
 
 CommandReturn = namedtuple("CommandReturn", "returncode stdout stderr")
 CMDLOG = logging.getLogger("commands")
@@ -515,3 +516,26 @@ def compare_versions(first, second):
         return first[1] - second[1]
 
     return first[0] - second[0]
+
+
+class ResultsFileDumper:
+    """Class responsible with outputting test results to files."""
+
+    def __init__(self, test_name: str, append=True):
+        """Initialize the instance."""
+        if not append:
+            flags = "w"
+        else:
+            flags = "a"
+
+        self._root_path = defs.TEST_RESULTS_DIR
+
+        # Create the root directory, if it doesn't exist.
+        self._root_path.mkdir(exist_ok=True)
+
+        self._file = open(self._root_path / test_name, flags)
+
+    def writeln(self, data: str):
+        """Write the `data` string to the output file, appending a newline."""
+        self._file.write(data)
+        self._file.write("\n")
--- a/tests/integration_tests/performance/test_network_tcp_throughput.py
+++ b/tests/integration_tests/performance/test_network_tcp_throughput.py
@@ -33,19 +33,20 @@ def measurements():
 
 def criteria_stats(cpu_baseline: dict, iperf3_id: str, env_id: str):
     """Return statistics definitions based with pass criteria."""
-    baseline_bw = cpu_baseline["baseline_bw"][env_id][iperf3_id]
+    [kernel, rootfs] = env_id.split("/")
+
+    baseline_bw = cpu_baseline["throughput"][kernel][rootfs][iperf3_id]
     delta_throughput = \
         baseline_bw[test_cfg.DELTA_PERCENTAGE_TAG] * \
         baseline_bw[test_cfg.TARGET_TAG] / 100
-    baseline_cpu_util = cpu_baseline["baseline_cpu_utilization"][env_id]
-    baseline_cpu_host = baseline_cpu_util["vmm"][
+    baseline_cpu_host = cpu_baseline["cpu_utilization_vmm"][kernel][rootfs][
         iperf3_id]
     baseline_vmm_target = baseline_cpu_host[test_cfg.TARGET_TAG]
     baseline_vmm_delta = \
         baseline_cpu_host[test_cfg.DELTA_PERCENTAGE_TAG] * \
         baseline_vmm_target / 100
     baseline_cpu_vcpus_total = \
-        baseline_cpu_util["vcpus_total"][iperf3_id]
+        cpu_baseline["cpu_utilization_vcpus_total"][kernel][rootfs][iperf3_id]
     baseline_vcpus_total_target = baseline_cpu_vcpus_total[test_cfg.TARGET_TAG]
     baseline_vcpus_total_delta = \
         baseline_cpu_vcpus_total[test_cfg.DELTA_PERCENTAGE_TAG] * \
@@ -103,7 +104,7 @@ def produce_iperf_output(basevm,
     # Check if we have enough CPUs to pin the servers on the host.
     # The available CPUs are the total minus vcpus, vmm and API threads.
     assert load_factor * basevm.vcpus_count < CpuMap.len() - \
-           basevm.vcpus_count - 2
+        basevm.vcpus_count - 2
 
     # Start the servers.
     for server_idx in range(load_factor*basevm.vcpus_count):
@@ -336,7 +337,7 @@ def pipes(basevm, host_ip, current_avail_cpu, env_id):
 
 @pytest.mark.nonci
 @pytest.mark.timeout(3600)
-def test_network_throughput(bin_cloner_path):
+def test_network_tcp_throughput(bin_cloner_path, results_file_dumper):
     """Test network throughput driver for multiple artifacts."""
     logger = logging.getLogger("network_tcp_throughput")
     artifacts = ArtifactCollection(_test_images_s3_bucket())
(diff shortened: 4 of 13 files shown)


tests/framework/defs.py
```diff
@@ -28,3 +28,5 @@
 ENV_TEST_IMAGES_S3_BUCKET = 'TEST_MICROVM_IMAGES_S3_BUCKET'
 """Default test session root directory path"""
 DEFAULT_TEST_SESSION_ROOT_PATH = "/srv"
+"""Name of the performance test results folder"""
+PERF_TEST_RESULTS_DIR = "../test_artifacts"
```
gc-plp, line 32: Name and/or description should reflect the fact that this is a relative path.

Also, for defining relative paths, we should set in stone one variable that fixes that path (e.g. `FC_WORKSPACE_DIR`) and start from there.

tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -33,19 +35,20 @@ def measurements():
 
 def criteria_stats(cpu_baseline: dict, iperf3_id: str, env_id: str):
     """Return statistics definitions based with pass criteria."""
-    baseline_bw = cpu_baseline["baseline_bw"][env_id][iperf3_id]
+    [kernel, rootfs] = env_id.split("/")
```
gc-plp, line 38: What's the typical format of `env_id`? Could we describe it in a comment?

tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -353,6 +356,12 @@ def test_network_throughput(bin_cloner_path):
         'name': 'network_tcp_throughput'
     }
 
+    if gather_performance_data:
+        test_context.custom['results_path'] = os.path.join(
```
gc-plp, line 360: `PERF_TEST_RESULTS_DIR` could be defined as a `pathlib.Path` object.

If so, we can avoid calling `os.path.join` and just define:

```python
test_context.custom['results_path']  = PERF_TEST_RESULTS_DIR / 'test_network_tcp_throughput.out'

File: tests/integration_tests/performance/test_vsock_throughput.py [hunk 4]
```diff
@@ -303,7 +305,11 @@ def test_vsock_throughput(bin_cloner_path):
         'name': 'vsock_throughput'
     }
 
-    print(get_cpu_model_name())
+    if gather_performance_data:
+        test_context.custom['results_path'] = os.path.join(
```
gc-plp, line 309: Same comment as above

tests/conftest.py
```diff
@@ -210,6 +211,17 @@ def test_session_tmp_path(test_fc_session_root_path):
     shutil.rmtree(tmp_path)
 
 
+@pytest.fixture
+def gather_performance_data(request):
```
iulianbarbu, line 215: nit: I am imagining this fixture returning a resource. It should be `None` if the flag for dumping the test results was not used. If the flag is used, it can be an object (lets call it `Dumper`), which will be concerned with the root destination of tests dumps. Furthermore, the object should provide interface for setting specific test dump destination under the root destination. In the end, it should provide interface for writing to destination. I would keep the framework agnostic of this tests dump aspect (having it aware only for pass criteria check), delegating the dumping aspect to the test logic. Let me know if you find significant value in this suggestion.

tools/parse_baselines/main.py
```diff
@@ -0,0 +1,88 @@
...
+
+def get_data_files(args):
+    """Return a list of files that contain results for this test."""
+    assert os.path.isdir(args.data_folder)
+
+    file_list = []
+
+    # Get all files in the dir tree that have the right name.
+    for root, _, files in os.walk(args.data_folder):
+        for file in files:
```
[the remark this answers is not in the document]
alindima: Actually, the `os.walk` method does a top-down tree traversal, getting all the files with the given name recursively.
See: https://www.tutorialspoint.com/python/os_walk.htm

tests/integration_tests/performance/test_network_tcp_throughput.py
alindima: well, I think this line is self-explanatory. Having a comment that says exactly that `env_id` is in this case of format `kernel/rootfs` seems unnecessary.
This `env_id` variable is specific to this test, doesn't have smth to do with the framework.

tests/integration_tests/performance/test_network_tcp_throughput.py
alindima: after the latest changes, this no longer applies here. Changes the `TEST_RESULTS_DIR` to be a `pathlib.Path` object

alindima:
Addressed your comments. Please re-review @iulianbarbu @gc-plp

tools/parse_baselines/main.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # We need at least one file.
+    assert len(file_list) > 0
+
+    return file_list
+
+
+def concatenate_data_files(data_files):
+    """Create temp file to hold all concatenated results for this test."""
+    outfile = tempfile.NamedTemporaryFile(delete=False)
```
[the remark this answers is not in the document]
iulianbarbu: My thought was to try something like the following:
```
>>> def boo():
...     x = tempfile.NamedTemporaryFile()
...     return x
... 
>>> def foo():
...     f = boo()
...     print(os.path.exists(f.name))
...     return f.name
... 
>>> os.path.exists(foo())
True
False
```

In our case, at the end of the `main()` function, the file will be effectively closed (part because the program finishes executing and part due to becoming out of scope and hence garbage collected, but I am not sure which one is the real motif), being deleted from the filesystem as well. It will prove the usage of the `unlink` unnecessary. Feel free to integrate it if you find it useful.

tools/parse_baselines/main.py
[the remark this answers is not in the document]
iulianbarbu: 👍

tools/parse_baselines/providers/types.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
iulianbarbu, line 1: The commits says there is an implementation for vsock baselines as well, but it misses from this commit.

alindima: it's not actually missing, it's just that the `NetworkTcpDataParser` is used for both vsock and tcp tests.
I'll rename it to IperfDataParser

tools/parse_baselines/main.py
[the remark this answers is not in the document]
alindima: great idea, integrated this. I was expecting that the interpreter would free up the memory and therefore close the file, if it sees that we only get the name prop from it.
However, it works if we save the outfile in a local variable and use that, as in your example

tools/parse_baselines/providers/types.py
alindima: done, renamed to `Iperf3DataParser`

tests/framework/utils.py
```diff
@@ -515,3 +516,26 @@ def compare_versions(first, second):
...
+class ResultsFileDumper:
+    """Class responsible with outputting test results to files."""
+
+    def __init__(self, test_name: str):
+        """Initialize the instance."""
+        self._root_path = defs.TEST_RESULTS_DIR
+        self._path = self._root_path / test_name
+
+        # Create the root directory, if it doesn't exist.
+        os.makedirs(self._root_path, exist_ok=True)
```
gc-plp, line 530: nit: since `_root_path` is a Pathlib.Path instance you can run the built-in mkdir function https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

tests/framework/utils.py
```diff
@@ -515,3 +516,26 @@ def compare_versions(first, second):
...
+
+    def writeln(self, data: str, append=True):
```
gc-plp, line 532: Do we ever need to append to a file? 

Seems like this is only called with json files as parameters and serializing jsons creates an invalid file.

tests/framework/utils.py
```diff
@@ -515,3 +516,26 @@ def compare_versions(first, second):
...
+        """Write the `data` string to the output file, appending a newline."""
+        if not append:
+            flags = "w"
+        else:
+            flags = "a"
+
+        with open(self._path, flags) as file:
```
gc-plp, line 539: Why not open the file on class init and keep writing data as this function is called?

tools/parse_baselines/providers/iperf3.py
```diff
@@ -0,0 +1,106 @@
...
+        self._data = nesteddict()
+
+    # pylint: disable=R0201
+    def _calculate_baseline(self, data: List[float]) -> dict:
+        """Return the target and delta values, given a list of data points."""
+        avg = statistics.mean(data)
+        stddev = statistics.stdev(data)
+        return {
+            'target': math.ceil(round(avg, 2)),
+            # We add a 3% extra margin, to account for small variations that
```
gc-plp, line 42: Can we just output the raw numbers and leave the extra margin in the responsibility of the comparison tool?

Seems like we're losing some accuracy and data here by truncating the raw values.

tools/parse_baselines/providers/iperf3.py
```diff
@@ -0,0 +1,106 @@
...
+                data = self._data[cpu_model]
+                for kernel in data[baseline]:
+                    data = self._data[cpu_model][baseline]
+                    for rootfs in data[kernel]:
+                        data = self._data[cpu_model][baseline][kernel]
+                        for iperf_config in \
+                                data[rootfs]:
+                            data[rootfs][iperf_config] = \
+                                self._calculate_baseline(
+                                    data[rootfs][iperf_config])
```
gc-plp, line 104: `data` is being reused and it's kind of confusing.

tools/parse_baselines/providers/iperf3.py
iulianbarbu: This is a good point. Should we make it optional, as in provide a flag for adding extra margins? I think it makes sense to add these extra margins only for `delta_percentage`s.  An extra cmd line argument for parse_baseline tool like `--base-delta-percentage 3` should do the job. @gc-plp , @alindima , I am curious if you think this would be the right thing to do.

gc-plp: That may work too. It's just the fact that we're hardcoding things that's going to chase us later.

tools/parse_baselines/providers/iperf3.py
alindima: we can indeed avoid specifying data, but since this is a very deep nested dict, we'll get pylint angry at the line length.
I can't think of a cleaner solution here

tests/framework/utils.py
alindima: I don't quite understand what you mean by `serializing jsons creates an invalid file`. Writing a line of JSON to a file does not invalidate that file in any way. It's not binary serialized.

This method can be called with any string payload, depending on the test and the baselines parser. The fact that it uses JSON here is specific to this test.

Moreover, I added the default behaviour of appending to the file for the case where we want to run `devtool test ..` in a loop, gathering data from multiple iterations. Since we are appending to the file, one will not lose the data gathered from previous iterations.

tests/framework/utils.py
alindima: that's a good point, we can close the file on `__del__`

tools/parse_baselines/providers/iperf3.py
alindima: Actually, I can solve this problem by recursion, which should be cleaner and future-proof to adding another dimension to the dict

tests/framework/utils.py
gc-plp: It's closed when it goes out of scope

tests/framework/utils.py
alindima: done

tests/framework/utils.py
alindima: done

tools/parse_baselines/providers/iperf3.py
alindima: done

tools/parse_baselines/providers/iperf3.py
iulianbarbu: The course of action in my opinion is to simply remove the extra percentage from here. This might be needed or not, depending on how stable the test is. Being optional, we can omit it right now and add it once it is needed, through a specific support that lets user increase the delta percentages by an extra margin. @gc-plp , @alindima , let me know if you agree with this course of action. This is the only thing left on my plate before approving the PR.

gc-plp: The extra percentage is rather random and should be optional. The 3% value seems chosen at random and may work for some tests while it may fail for some others as @iulianbarbu said.

alindima: This X% extra value is totally optional. Every test parser will implement the `DataParser` abstract class. It is only added for the tcp vsock throughput tests

iulianbarbu (approved):

tools/parse_baselines/providers/iperf3.py
alindima: moved this to a constant

gc-plp (approved):

PR #2436, opened 2021-02-01
balloon: log error for duplicate pfns

georgepisaltu:
# Reason for This PR

The balloon device did not have any mechanism in place for dealing with duplicate page frame numbers coming from the guest driver. This is an edge case, as a normal driver would not be allowed to send duplicate pages to the balloon.

## Description of Changes

With this change, the balloon will log an error message and skip any duplicate page frame numbers sent by the guest driver.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/balloon/utils.rs
+++ b/src/devices/src/virtio/balloon/utils.rs
@@ -4,6 +4,7 @@
 use std::io;
 
 use super::{RemoveRegionError, MAX_PAGE_COMPACT_BUFFER};
+use logger::error;
 use vm_memory::{GuestAddress, GuestMemory, GuestMemoryMmap, GuestMemoryRegion};
 
 /// This takes a vector of page frame numbers, and compacts them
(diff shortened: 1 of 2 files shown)


src/devices/src/virtio/balloon/utils.rs
```diff
@@ -25,25 +26,33 @@ pub(crate) fn compact_page_frame_numbers(v: &mut [u32]) -> Vec<(u32, u32)> {
...
 
-    for page_frame_number in &v[1..] {
+    for pfn_index in 1..v.len() {
+        let page_frame_number = v[pfn_index];
+
+        // Skip duplicate pages.
+        if page_frame_number == v[pfn_index - 1] {
+            error!("Skipping duplicate PFN.");
+            continue;
+        }
```
acatangiu, line 39: I'm not sure I see the value of this check.. A misbehaving/malfunctioning driver can still add duplicated entries since this check only checks for adjacent duplicates.

What is the impact of having duplicates anyway? Is it anything else than emulation overhead? If not, then let's not worry about it, it would be pretty costly to guarantee catching it on device side.

src/devices/src/virtio/balloon/utils.rs
```diff
@@ -25,25 +26,33 @@ pub(crate) fn compact_page_frame_numbers(v: &mut [u32]) -> Vec<(u32, u32)> {
...
+
         // Check if the current page frame number is adjacent to the most recent page range.
-        if *page_frame_number == previous + length {
+        if page_frame_number == v[previous] + length {
```
acatangiu, line 42: This operation can overflow with the _right_ invalid input provided by a malfunctioning driver.

src/devices/src/virtio/balloon/utils.rs
sandreim: Since these are sorted above, all duplicates shall be adjacent.

src/devices/src/virtio/balloon/utils.rs
```diff
@@ -25,25 +26,33 @@ pub(crate) fn compact_page_frame_numbers(v: &mut [u32]) -> Vec<(u32, u32)> {
...
+    let mut previous = 0;
     let mut length = 1;
 
-    for page_frame_number in &v[1..] {
+    for pfn_index in 1..v.len() {
+        let page_frame_number = v[pfn_index];
+
+        // Skip duplicate pages.
+        if page_frame_number == v[pfn_index - 1] {
+            error!("Skipping duplicate PFN.");
```
sandreim, line 37: You can also print out the duplicate PFN.

src/devices/src/virtio/balloon/utils.rs
georgepisaltu: This operation is what led me to make this change.

When we reviewed this algorithm, I thought about whether or not this operation needs to be a `checked_add`, but I decided it's not needed as a normal driver would only send distinct PFNs.

This makes the overflow impossible as for whatever value `v[previous]` has in the u32 range, we know there are at least `length` consecutive numbers greater than it in the array (the greatest so far being `page_frame_number`), since `v[previous]` is before all of them in the sorted array and `length` was incremented for each consecutive one.

The overflow can happen when the numbers are not distinct and we have multiple max values. 

Suppose you have an array of PFNs `[u32::MAX, u32::MAX]`. When running the old code, `previous` would be `u32::MAX`, length would be `1`. This will overflow as `previous` is **_never_** supposed to be `u32::MAX` unless the array has a length of 1.

This actually solves the overflow issue. I also added a unit test covering this.

src/devices/src/virtio/balloon/utils.rs
georgepisaltu: Done.

sandreim (approved):

acatangiu (approved):

PR #2438, opened 2021-02-03
aarch64: relax cache topology restrictions

### LINKED ISSUE #2407: Commit 76528ea4 aarch64: setup cache topology regression
# Describe the bug

Commit 76528ea4 aarch64: setup cache topology causes a regression on systems that do not have FDT cache entries.

## To Reproduce

On an aarch64 system with a minimal FDT, run without this commit. 

Then, add the commit. Firecracker will fail with:
`2021-01-08T13:40:26.079775900 [vmnetns:main] Building VMM configured from cmdline json failed: ConfigureSystem(SetupFDT(ReadCacheInfo(InvalidCacheAttr("Line Size", "cannot parse integer from empty string"))))
`

## Expected behaviour

Firecracker should have sufficient defaults to handle this. Also, the code should check for the existence of the required entries in /sysfs and display an error warning instead of the above cryptic message.

## Environment

The FDT is basic, with no cache architecture lines in the FDT. As a result, `/sys/devices/system/cpu/cpu0/cache//index{0..2}` contains only
```
ls /sys/devices/system/cpu/cpu0/cache//index0
level power shared_cpu_list shared_cpu_map type uevent
```

`[ - Firecracker version.]`
Upstream master as of Jan 8. Last commits:
364cf32b (upstream/master, upstream_master) Increase Pytest Coverage
62d7a1b8 Implement Guard Paging Mechanism
bf4796da docs: updating Firecracker maintainers
1026a7c1 CI: modify func for retrieving files

`[ - Host and guest kernel versions.]`
Host: android 4.14
`[ - Rootfs used.]`
N/A
`[ - Architecture.]`
aarch64

## Additional context

I have to revert this commit to use firecracker.

Check for and supply reasonable defaults for cache params.

## Checks

- [ x] Have you searched the Firecracker Issues database for similar problems?
- [x ] Have you read the existing relevant Firecracker documentation?
- [x ] Are you certain the bug being reported is a Firecracker issue?

dianpopa:
# Reason for This PR

Fixes #2407 

## Description of Changes

* the commits contain the info needed

## License Acceptance

## PR Checklist

### DIFF
--- a/src/arch/src/aarch64/cache_info.rs
+++ b/src/arch/src/aarch64/cache_info.rs
@@ -5,6 +5,8 @@ use std::fmt::{Display, Formatter};
 use std::path::{Path, PathBuf};
 use std::{fs, io, result};
 
+use logger::warn;
+
 // Based on https://elixir.free-electrons.com/linux/v4.9.62/source/arch/arm64/kernel/cacheinfo.c#L29.
 const MAX_CACHE_LEVEL: u8 = 7;
 
@@ -12,26 +14,44 @@ const MAX_CACHE_LEVEL: u8 = 7;
 pub enum Error {
     FailedToReadCacheInfo(io::Error),
     InvalidCacheAttr(String, String),
-    NoMoreCacheLevels(u8),
+    MissingCacheConfig,
+    MissingOptionalAttr(String),
 }
 
 type Result<T> = result::Result<T, Error>;
 
-#[derive(Default)]
-pub struct CacheInfo {
+#[derive(Clone)]
+pub(crate) struct CacheInfo {
     // Cache Level: 1, 2, 3..
     pub level: u8,
-    pub size_: usize,
     // Type of cache: Unified, Data, Instruction.
     pub type_: CacheType,
-    pub number_of_sets: u16,
-    pub line_size: u16,
+    pub size_: Option<usize>,
+    pub number_of_sets: Option<u16>,
+    pub line_size: Option<u16>,
     // How many CPUS share this cache.
     pub cpus_per_unit: u16,
+    // The directory from which to read cache info.
+    // In tests this gets modified.
+    pub cache_dir: PathBuf,
 }
 
+impl Default for CacheInfo {
+    fn default() -> Self {
+        CacheInfo {
+            level: 0,
+            type_: CacheType::Unified,
+            size_: None,
+            number_of_sets: None,
+            line_size: None,
+            cpus_per_unit: 1,
+            cache_dir: PathBuf::from("/sys/devices/system/cpu/cpu0/cache"),
+        }
+    }
+}
+
+#[derive(Clone)]
 // Based on https://elixir.free-electrons.com/linux/v4.9.62/source/include/linux/cacheinfo.h#L11.
-#[derive(Debug)]
 pub enum CacheType {
     Instruction,
     Data,
@@ -45,11 +65,10 @@ impl Display for Error {
                 write!(f, "Failed to read cache information: {}", err)
             }
             Error::InvalidCacheAttr(ref attr, ref err) => {
-                write!(f, "Invalid \"{}\" cache attribute: {}", attr, err)
+                write!(f, "Invalid cache configuration found for {}: {}", attr, err)
             }
-            Error::NoMoreCacheLevels(ref index) => {
-                write!(f, "Reached maximum cache index: {}", index)
-            }
+            Error::MissingCacheConfig => write!(f, "Cannot proceed with reading cache info"),
+            Error::MissingOptionalAttr(ref msg) => write!(f, "{}", msg),
         }
     }
 }
@@ -61,7 +80,7 @@ impl CacheType {
             "Data" => Ok(Self::Data),
             "Unified" => Ok(Self::Unified),
             cache_type => Err(Error::InvalidCacheAttr(
-                "Type".to_string(),
+                "type".to_string(),
                 cache_type.to_string(),
             )),
         }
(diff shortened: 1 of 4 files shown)


src/arch/src/aarch64/fdt.rs
```diff
@@ -338,17 +338,27 @@ fn create_cpu_nodes(fdt: &mut Vec<u8>, vcpu_mpidr: &[u64]) -> Result<()> {
...
-                cache.type_.of_cache_sets(),
-                cache.number_of_sets as u32,
-            )?;
+            if cache.size_.is_some() {
+                append_property_u32(
+                    fdt,
+                    cache.type_.of_cache_size(),
+                    cache.size_.unwrap() as u32,
+                )?;
+            }
```
georgepisaltu, line 347: nit: As this is used multiple times below (and in similar fashion), maybe this can be a function or macro for deduplication and to avoid the use of `unwrap` with an `if let`.

georgepisaltu (dismissed):

dianpopa:
@georgepisaltu addressed your feedback and [pclesr's feedback](https://github.com/firecracker-microvm/firecracker/issues/2407#issuecomment-777470533). PTAL!

src/arch/src/aarch64/cache_info.rs
```diff
@@ -101,9 +118,76 @@ impl CacheType {
...
+        if let Ok(number_of_sets) = self.sysfs_cache(index, "number_of_sets") {
+            self.number_of_sets = Some(number_of_sets.parse::<u16>().map_err(|e| {
+                Error::InvalidCacheAttr("number_of_sets".to_string(), e.to_string())
+            })?);
+        } else {
+            *err_str += if *err_str != "" { ", " } else { "" };
+            *err_str += "number of sets";
+        }
+
+        Ok(())
```
alindima, line 186: I find it a little weird to use this `err_str` and return an `Ok` value. Is there a reason we're not using a regular `Error`?

src/arch/src/aarch64/cache_info.rs
```diff
@@ -101,9 +118,76 @@ impl CacheType {
...
+        } else {
+            *err_str += "shared cpu map";
+        }
+
+        if let Ok(coherency_line_size) = self.sysfs_cache(index, "coherency_line_size") {
+            self.line_size = Some(coherency_line_size.parse::<u16>().map_err(|e| {
+                Error::InvalidCacheAttr("coherency_line_size".to_string(), e.to_string())
+            })?);
+        } else {
+            *err_str += if *err_str != "" { ", " } else { "" };
```
alindima, line 166: we can simplify this logic by simply adding a comma after each append operation, and pop the last comma right before returning

src/arch/src/aarch64/cache_info.rs
```diff
@@ -318,24 +356,208 @@ mod tests {
...
+        );
+    }
+
+    #[test]
+    fn test_sysfs_read_caches() {
+        let mut l1_caches: Vec<CacheInfo> = Vec::new();
+        let mut non_l1_caches: Vec<CacheInfo> = Vec::new();
+        // We use sysfs for extracting the cache information.
+        assert!(sysfs_read_caches(&mut l1_caches, &mut non_l1_caches).is_ok());
+        assert_eq!(l1_caches.len(), 2);
```
alindima, line 560: Are these assumptions we're willing to make? These depend on the cache topology of the host, right?

If somebody is running these unit tests on another machine that does not have these, they will run into errors I'm assuming.

Can we somehow mock the sysfs for this test?

dianpopa: Indeed, these asserts depend on the host configuration and they might fail when run on other machines. I removed them.
Mocking a sysfs is not quite obvious given that `sysfs_read_caches` (the function that we are testing) constructs the `cache` object inside it and so I do not have access to the `cache_dir` field.

dianpopa (commented):
@alindima @georgepisaltu I addressed your latest feedback, please give it another look. Thanks

src/arch/src/aarch64/cache_info.rs
alindima: I understand what you're saying. However, I still find it confusing to have an `err_str` with an `Ok()` value.
Can't we distinguish the error types in the caller of this function? We can then match on the returned Error and decide whether it's a soft error (one that does not kill the process) or a hard error.

src/arch/src/aarch64/cache_info.rs
alindima: I see. However, can't we mock the `cache_dir` by implementing the `Default` trait in the test mod with a custom path?

dianpopa (commented):
Regarding https://github.com/firecracker-microvm/firecracker/pull/2438#discussion_r587413612, re-implementing `Default` in the tests has a tricky part associated. `TempDir` will [remove the associated path once it goes out of scope](https://github.com/rust-vmm/vmm-sys-util/blob/master/src/unix/tempdir.rs#L26). So, when `sysfs_read_caches` will try to make use of `cache_dir` will crash. Unfortunately, the `vmm-sys-util` implementation of `TempDir` does not have support for making the path persist after drop. 
Now, that I think about the big picture, this function is also called from the fdt code which will also obtain an FDT specific to the machine that our CI is run on (i.e m6g). So, my proposal here is to tackle this separately as part of this [issue](https://github.com/firecracker-microvm/firecracker/issues/2492) since it has a larger scope, WDYT?

src/arch/src/aarch64/cache_info.rs
alindima: looks good

alindima (approved):

alindima:
> Regarding [#2438 (comment)](https://github.com/firecracker-microvm/firecracker/pull/2438#discussion_r587413612), re-implementing `Default` in the tests has a tricky part associated. `TempDir` will [remove the associated path once it goes out of scope](https://github.com/rust-vmm/vmm-sys-util/blob/master/src/unix/tempdir.rs#L26). So, when `sysfs_read_caches` will try to make use of `cache_dir` will crash. Unfortunately, the `vmm-sys-util` implementation of `TempDir` does not have support for making the path persist after drop.
> Now, that I think about the big picture, this function is also called from the fdt code which will also obtain an FDT specific to the machine that our CI is run on (i.e m6g). So, my proposal here is to tackle this separately as part of this [issue](https://github.com/firecracker-microvm/firecracker/issues/2492) since it has a larger scope, WDYT?

I agree

alindima (approved):

georgepisaltu (approved):

Issue #2209, opened 2020-10-22
[RFC] File-based seccomp filters


alindima:
This issue is meant to be an RFC for a new Firecracker feature - **file-based seccomp filter customisation**.

Please leave below any comments/questions you may have. Any feedback is welcome!

Here is the issue tracking this feature: https://github.com/firecracker-microvm/firecracker/issues/1366

## What is the problem?

Firecracker is currently using the linux [**seccomp-bpf**](https://www.kernel.org/doc/Documentation/userspace-api/seccomp_filter.rst) feature to limit the process from making dangerous or unneeded syscalls to the underlying host OS. This is a powerful security mechanism that is put in place after the necessary FC setup and just before running client code in the microVM.

Currently, the seccomp filters are hardcoded into the source code. This means that every modification in the filter would need a recompilation and potentially, a new release. This adds up to a lot of unnecessary work.

This feature would also enable any Firecracker user to build seccomp support for their own platform and libc toolchain. This should be particularly useful for solving glibc compatibility issues, where the list of required syscalls is highly dependent on the version.

By implementing this feature, we also enable the following seccomp improvements:

* having **customer-tailored filters**. It would enable users to customise filters to their specific use-case. (For example, if you are not using snapshot-restore, you may be able to tighten the syscall allowlist).
* having **thread-specific seccomp filters** (Each Firecracker thread category has a different responsibility and, in turn, needs a different set of syscalls →  potential security and performance improvements, by having smaller allowlists).

## Proposed solution

We will expose a way for users to **optionally** provide custom seccomp filters for the Firecracker process, in the form of external files. These files will contain serialized binary BPF code that will get installed in the kernel.

### Changes in Firecracker’s API

Currently, the `--seccomp-level` parameter takes three options: `0` (no filtering), `1` (basic filtering - matches on syscall names), `2` (advanced filtering - matches on syscall names and arguments, the default).

The proposal is to **remove** the `--seccomp-level` parameter, and replace it with a combination of two other parameters:

```
--no-seccomp # seccomp is disabled, equivalent to level 0
--seccomp-filter "/path/to/bpf" # custom filters, at the given path
```

As a **default**, when no parameters are specified, Firecracker will use the default, advanced filters, just as before (equivalent to level 2).

This implies removing the option of basic filtering (level 1). The level 1 option is still achievable via custom filter files.

[...]

PR #2443, opened 2021-02-04
feature/file_based_seccomp: build mechanism for default filters

alindima:
# Reason for This PR

As per the RFC issue: #2209 , add a build mechanism for embedding the default seccomp filters in the Firecracker binary.

## Description of Changes

- Removes the default hardcoded filters from Firecracker, adding their JSON correspondents for x86_64_musl and aarch64_musl. Other platforms will receive empty filters by default.
- Add cargo build script that runs seccompiler and places the serialized BPF in the `build/` directory for Firecracker to consume at compile-time.
- Replace mechanism for getting the default filters in Firecracker with a call to `include_bytes!` which will embed the serialized filters in the binary.
- Updated changelog to state the existence of seccompiler and the new default that experimental gnu builds no longer use default seccomp filters (they use empty filters and users have the possibility of overriding them).

**Note**: the JSON files may seem huge. Keep in mind that the filter is currently duplicated for each thread category. An audit which will come in a subsequent PR will restrict the filters to the bare-minimum needed for that thread category.

## License Acceptance

## PR Checklist


resources/seccomp/x86_64-unknown-linux-musl.json
```diff
@@ -0,0 +1,1880 @@
...
+            },
+            {
+                "syscall": "accept4",
+                "comment": "Called by the api thread to receive data on socket",
+                "args": [
+                    {
+                        "arg_index": 3,
+                        "arg_type": "dword",
+                        "op": "eq",
+                        "val": 524288,
```
sandreim, line 78: Would it make sense to have these as hex ?

sandreim (commented):
Looks good overall. A few questions:
- are docs coming in a separate PR ?
- are integration tests for --seccomp-filter coming in a separate PR ?
- for usecases where snapshots and custom filters are being used, I am wondering if it would be any cons to bundle the compiled filters in the vm state since the custom filter loading would create additional latency. WDYT ?

resources/seccomp/x86_64-unknown-linux-musl.json
alindima: unfortunately, JSON does not support hex numbers

alindima:
> * are docs coming in a separate PR ?

Yes, documentation is coming in a future PR.

> * are integration tests for --seccomp-filter coming in a separate PR ?

There are already integration tests for --seccomp-filter. Take a look [here](https://github.com/firecracker-microvm/firecracker/blob/feature/file_based_seccomp/tests/integration_tests/security/test_custom_seccomp.py).
Let me know if you have some use-cases in mind that we do not currently cover.

> * for usecases where snapshots and custom filters are being used, I am wondering if it would be any cons to bundle the compiled filters in the vm state since the custom filter loading would create additional latency. WDYT ?

hmm this an interesting topic to think about. By embedding the custom filter in the vm state, there could be a conflict in the sense that the user could specify an additional filter and we would have to think about a conflict resolution situation.

I also think that it could increase the attack surface, since the filter directly affects the defense mechanisms of the process.
IMHO, it would be best to think about custom seccomp filters the same way we think about disk files on host. They can be part of the snapshot state, as a separate file, managed by the user independently.

I can take an action item to measure the snapshot load performance and boot time with a custom filter, WDYT @sandreim ?
If there are significant differences we can resume the discussion from there

resources/seccomp/x86_64-unknown-linux-musl.json
sandreim: Hex as string?

alindima: hmm this would add some extra code complexity (not much), but I don't see a benefit really. WDYT?

acatangiu:
I think the seccomp whitelist is more related to the host-side setup rather than the guest workload.
What I mean is, I look at it how one looks at the `jailer` or other host-prod configurations. It is more dependent on the service plans for that microVM than dependent on the guest workload.

Considering this, I would not include the whitelist in the snapshot.

resources/seccomp/x86_64-unknown-linux-musl.json
alindima: we synced offline with the team and decided that the benefit of adding hex values is not worth the extra complexity, for now

sandreim: Ack. Makes sense to keep it on radar as an eventual improvement if customers really need it.

sandreim:
> > * for usecases where snapshots and custom filters are being used, I am wondering if it would be any cons to bundle the compiled filters in the vm state since the custom filter loading would create additional latency. WDYT ?
> 
> hmm this an interesting topic to think about. By embedding the custom filter in the vm state, there could be a conflict in the sense that the user could specify an additional filter and we would have to think about a conflict resolution situation.
> 
> I also think that it could increase the attack surface, since the filter directly affects the defense mechanisms of the process.
> IMHO, it would be best to think about custom seccomp filters the same way we think about disk files on host. They can be part of the snapshot state, as a separate file, managed by the user independently.
> 
> Another con that I can think about is that the Firecracker process needs the filter for the API thread before it receives the `LoadSnapshot` command (before it even launches the API server even). Getting the filters as part of the vm state would not work in this case
> 
> I can take an action item to measure the snapshot load performance and boot time with a custom filter, WDYT @sandreim ?
> If there are significant differences we can resume the discussion from there

I agree, and +1 for @acatangiu comment - this is workload agnostic.

alindima:
> > > * for usecases where snapshots and custom filters are being used, I am wondering if it would be any cons to bundle the compiled filters in the vm state since the custom filter loading would create additional latency. WDYT ?
> > 
> > 
> > hmm this an interesting topic to think about. By embedding the custom filter in the vm state, there could be a conflict in the sense that the user could specify an additional filter and we would have to think about a conflict resolution situation.
> > I also think that it could increase the attack surface, since the filter directly affects the defense mechanisms of the process.
> > IMHO, it would be best to think about custom seccomp filters the same way we think about disk files on host. They can be part of the snapshot state, as a separate file, managed by the user independently.
> > Another con that I can think about is that the Firecracker process needs the filter for the API thread before it receives the `LoadSnapshot` command (before it even launches the API server even). Getting the filters as part of the vm state would not work in this case
> > I can take an action item to measure the snapshot load performance and boot time with a custom filter, WDYT @sandreim ?
> > If there are significant differences we can resume the discussion from there
> 
> I agree, and +1 for @acatangiu comment - this is workload agnostic.

Actually, it can also be dependent on the guest workload. In the case where a customer tailors the filter to remove some unneeded syscalls (removes the mmap needed for balloon device for example).  However, the other cons still outweigh the addition of the filter to the vm state. the filter can be considered another (optional) snapshot artifact

acatangiu:
> > > > * for usecases where snapshots and custom filters are being used, I am wondering if it would be any cons to bundle the compiled filters in the vm state since the custom filter loading would create additional latency. WDYT ?
> > > 
> > > 
> > > hmm this an interesting topic to think about. By embedding the custom filter in the vm state, there could be a conflict in the sense that the user could specify an additional filter and we would have to think about a conflict resolution situation.
> > > I also think that it could increase the attack surface, since the filter directly affects the defense mechanisms of the process.
> > > IMHO, it would be best to think about custom seccomp filters the same way we think about disk files on host. They can be part of the snapshot state, as a separate file, managed by the user independently.
> > > Another con that I can think about is that the Firecracker process needs the filter for the API thread before it receives the `LoadSnapshot` command (before it even launches the API server even). Getting the filters as part of the vm state would not work in this case
> > > I can take an action item to measure the snapshot load performance and boot time with a custom filter, WDYT @sandreim ?
> > > If there are significant differences we can resume the discussion from there
> > 
> > 
> > I agree, and +1 for @acatangiu comment - this is workload agnostic.
> 
> Actually, it can also be dependent on the guest workload. In the case where a customer tailors the filter to remove some unneeded syscalls (removes the mmap needed for balloon device for example). However, the other cons still outweigh the addition of the filter to the vm state. the filter can be considered another (optional) snapshot artifact

Yes, it will always be a combination of host requirements (how much would the microVM firecracker process be allowed to do) and guest workload requirements (what the firecracker process needs to be able to do for a given workload), but managing this balance falls to the service on the host side - when spawning a firecracker process, decide what its purpose and target workload would be (whether planning to load from snapshot or booting fresh is also part of the plans for that process and can influence the seccomp whitelist).

sandreim:
> > Actually, it can also be dependent on the guest workload. In the case where a customer tailors the filter to remove some unneeded syscalls (removes the mmap needed for balloon device for example). However, the other cons still outweigh the addition of the filter to the vm state. the filter can be considered another (optional) snapshot artifact
> 
> Yes, it will always be a combination of host requirements (how much would the microVM firecracker process be allowed to do) and guest workload requirements (what the firecracker process needs to be able to do for a given workload), but managing this balance falls to the service on the host side - when spawning a firecracker process, decide what its purpose and target workload would be (whether planning to load from snapshot or booting fresh is also part of the plans for that process and can influence the seccomp whitelist).

I think that seccomp filters should not be viewed as workload dependent - workloads cannot issue syscalls on the host - unless they've broken out of the VM. Custom filters allows to quickly fix production issues by allowing some extra syscalls  without rebuilding Firecracker. Also, they can take a stab at tightening them at the risk of crashing in prod - this still requires knowledge of the Firecracker implementation and a lot of testing. With great power comes great responsibility.

sandreim (approved):

georgepisaltu (approved):

PR #2446, opened 2021-02-08
[RFC] Scheduler update

gc-plp:
# Reason for This PR

This PR proposes mechanism of changing how we set test parallelism.

It changes the mechanism where we decide parallelism based on folder name to a mechanism where we decide parallelism upon what fixtures are used by the test.

For example, if a test uses a fixture that launches a VM, then parallelism for that test will be set to 1, which means that only that test can run at that time.

The PR also introduces a method of manually specifying parallelism for a test, which overrides the built-in behavior.

## Description of Changes

1. Removes dict in scheduler with per function based approach to inject test suites.
2. Add concurrency decorator to functions that seem to be able to be executed in parallel.
3. Move performance tests to different folder.

## License Acceptance

## PR Checklist


gc-plp:
Will split the PR in a few commits to make it clearer to review.

gc-plp:
Seems like we earn about 2 minutes of test time when measuring the AMD CI, where we break down tests by test folder thus making parallelism less likely to happen.

tests/integration_tests/build/test_unittests.py
```diff
@@ -17,6 +17,7 @@
     "target",
     TARGETS
 )
+@pytest.mark.concurrency(1)
```
alindima, line 20: why do we add this restriction? Do we need to also add it for `test_coverage.py`?

tests/framework/scheduler.py
```diff
@@ -6,12 +6,27 @@
...
+will be set to 1, which means that while that test is running, at most one test
+can run in parallel with it.
+
+Although these limitations are detected automatically, one can also define
+concurrency limits manually by adding the `@pytest.mark.concurrency(ARG)`
+decorator, where ARG can be a number or 'max'.
+If ARG is a number, that number will be set as the concurrency level for the
+test. If ARG is 'max' then the test will run using the maximum concurrency
+limit set.
+If ARG > maximum concurrency limit, then the concurrency level will not exceed
```
alindima, line 28: The concurrency limit should probably be a constant somewhere and this doc block should reference it.

tests/framework/scheduler.py
```diff
@@ -25,20 +40,134 @@
...
+    def get_runnable_items(self, session_items):
+        """Return a list of runnable items given the session items."""
+        items = list(session_items)
+
+        # A test can match any of the patterns defined by the batch,
+        # in order to get assigned to it.
+        to_remove = []
+        for item in items:
+            # Check if the file path matches any of the path lists
+            matches_batch = re.search(
```
alindima, line 79: This method could be split into two: `matches_path()` and `is_ignored_path()` which would make it easier to read

tests/framework/scheduler.py
alindima: Having looked over the code some more, we should just say that this limit is given to the `--concurrency` CLI param

tests/framework/scheduler.py
```diff
@@ -225,40 +291,69 @@ def pytest_sessionfinish(self, session, exitstatus):
...
+                    random() + len(workers) / 5.0 if len(workers) > 1 else 0
+                )
+
+                # Create the worker process and start it up.
+                worker = mp.Process(
+                    target=self._worker_main, args=(crt_item.item, _delay)
+                )
+                workers[worker] = crt_item
+                worker.start()
+            sleep(0.1)
```
alindima, line 346: why is this sleep needed?

tests/framework/scheduler.py
```diff
@@ -225,40 +291,69 @@ def pytest_sessionfinish(self, session, exitstatus):
...
+
+                # Avoid storming the host with too many workers
+                # started at once.
+                _delay = (
+                    random() + len(workers) / 5.0 if len(workers) > 1 else 0
+                )
+
+                # Create the worker process and start it up.
+                worker = mp.Process(
+                    target=self._worker_main, args=(crt_item.item, _delay)
```
alindima, line 342: it seems that we are using a worker for only one item, right? Can't we give a worker more items to handle sequentially?

tests/framework/scheduler.py
```diff
@@ -225,40 +291,69 @@ def pytest_sessionfinish(self, session, exitstatus):
...
+                )
+                workers[worker] = crt_item
+                worker.start()
+            sleep(0.1)
+
+        # Wait for all workers to die
```
alindima, line 348: isn't this equivalent to `watcher.join()`, if we move the `while(workers)` to the `wait_for_worker_to_die()` method?
I think it fits better there

tests/framework/scheduler.py
```diff
@@ -115,96 +241,36 @@ def pytest_runtestloop(self, session):
...
-                "\n[ ",
-                self._colorize('yellow', batch['name']),
-                " | ",
-                "{} tests".format(len(batch['items'])),
-                " | ",
-                "{} worker(s)".format(batch['concurrency']),
-                " ]\n"
-            )
-            self._run_batch(batch)
+        # First, go through the args and check if a file was specified
```
alindima, line 244: what if the user specified a folder? E.G. `tools/devtool test -- integration_tests/style`

tests/framework/scheduler.py
```diff
@@ -225,40 +291,69 @@ def pytest_sessionfinish(self, session, exitstatus):
         if "failed" in trep.stats:
             session.exitstatus = ExitCode.TESTS_FAILED
 
-    def _run_batch(self, batch):
-        """Run the tests in this batch, spread across multiple workers.
-
-        Called in the server process context.
+    def wait_for_worker_to_die(self, workers):
```
alindima, line 294: it looks like it is waiting for all workers to die

tests/framework/scheduler.py
```diff
@@ -73,7 +203,18 @@ def do_pytest_addoption(parser):
             action="store",
             type=int,
             default=default,
-            help="Concurrency level (max number of worker processes to spawn)."
+            help="Concurrency level (max number of worker \
+processes to spawn).",
+        )
+
+        parser.addoption(
```
alindima, line 210: Maybe we should add these API changes in the commit message, so that it is a bit clearer how to run tests now

alindima (commented):
While the changes look promising, I don't fully understand why we want to constrain the tests that use microVMs to not be parallelised. I might be missing something but I think we can do this for most tests, except for performance tests. the metals we use for testing are quite big and I think they should handle that.

Also, I don't really like the fixture-based mechanism of deciding concurrency. Fixtures can change and there could be two tests that use the same fixture but should have a different concurrency level (also, the new `TestMatrix` tests may not even use fixtures in the future).

My proposal would be to have a default concurrency for all tests and manually override it for tests where this is a problem (which _should_ not be many).

gc-plp:
> While the changes look promising, I don't fully understand why we want to constrain the tests that use microVMs to not be parallelised. I might be missing something but I think we can do this for most tests, except for performance tests. the metals we use for testing are quite big and I think they should handle that.

We already constrain all the tests to be executed one by one. We're not changing anything there.

The reason for doing this in the past was that in the test framework we measure various timing related aspects which may fail if multiple VMs are running at the same time. For example all API requests are timed and can fail easily.

> Also, I don't really like the fixture-based mechanism of deciding concurrency. Fixtures can change and there could be two tests that use the same fixture but should have a different concurrency level (also, the new `TestMatrix` tests may not even use fixtures in the future).

The reason for proposing the fixture based mechanism is that it should detect automatically what tests are launching VMs.

`TestMatrix` based tests should explicitly use fixtures that launch VMs as it gives better visibility on what are the test dependencies.

> My proposal would be to have a default concurrency for all tests and manually override it for tests where this is a problem (which _should_ not be many).

I think that setting test concurrency can be easily overlooked and can evolve in a can of worms if we implement it at this granularity level. We're currently in a similar situation where we set concurrency by folder/file name.

tests/integration_tests/build/test_unittests.py
gc-plp: Some unit tests pass criteria are based on time measurement.

tests/framework/scheduler.py
```diff
@@ -6,12 +6,27 @@
...
+parallel with others.
+There is an assumption made, that if a test starts a microvm, its performance
+may be affected by other tests running in parallel with it, so for this reason
+the test framework limits tests that launch microvms to be executed in parallel
+with others.
+
+In order to check if a test launches a VM, the framework looks up what fixtures
+the test uses. If one of those fixtures matches the list enumerated in
+`PytestScheduler.TestItem.CONC_FIXTURES_PREF`, then the tests concurrency limit
+will be set to 1, which means that while that test is running, at most one test
```
serban300, line 19: What does concurrency limit 2 or 3 mean for example ? How many tests can run in parallel ?

gc-plp:
Will resume work on this later as we will be refactoring some of the scheduler code as part of the tooling rework effort.

Issue #2207, opened 2020-10-20
Add mechanism to select per block device cache strategy.


sandreim:
We should investigate the opportunity and possibilities to allow customers to influence how host storage is being accessed in the Firecracker block device emulation. Here is a list of switches that might make sense to expose as configuration options:
- ignore flush commands (from guest virtio driver) for performance reasons
- use host page cache for read/write operations or prefer direct io with the block file
- synchronous mode (O_SYNC/O_DSYNC) - writes will be reported complete only when committed to the host storage device

berciuliviu:
Relates to [2223](https://github.com/firecracker-microvm/firecracker/pull/2223)

Issue #2172, opened 2020-10-12
Virtio block flush (VIRTIO_BLK_T_FLUSH) not syncing data to host


pclesr:
## Describe the bug

It appears that a sync on an ext4 file in the guest does not result in the data on the host also being sync'd out to disk.

## To Reproduce
Using the file tester `fio` with the following config file:
```
[globa]
rw=write
ioengine=sync
bs=32k
direct=0
size=128m
numjobs=1
fsync_on_close=1
loops=10
```

When running this on the host to establish a baseline and running the linux `perf` utility (with call graphs enabled), one sees on an ancient (android) 4.14 kernel many calls to sync
```
    5.50%     0.00%  fio              [kernel.kallsyms]     [k] sys_fsync
            |
            ---sys_fsync
               do_fsync
               vfs_fsync_range
               ext4_sync_file
               file_write_and_wait_range
```

perf is run with 
`perf record -a --call-graph dwarf,16384 -F 200 -o /tmp/perf.data /tmp/fio /tmp/fio.job
`
Then the results with `perf report `and `perf script`.

When running the exact same `fio` and config in the VM, the bandwidth reported by `fio` is 4x that of the host baseline, indicating that the data is never actually sync'd to the medium. The page cache on the host is functioning as a huge disk cache.  Running perf  on the host during the execution of the fio test in the VM results shows no calls to sync.

The test is run against a virtio disk (/dev/vdb) that is formatted as an ext4 disk on the host, and then specified in the list of devices in the VMs config.json and mounted in the guest. 

If perf is run in the VM (with a 5.7 kernel), then perf shows calls both to sync and the ext4 write call in the VM kernel:
```
   20.56%     0.03%  fio              [kernel.kallsyms]  [k] new_sync_write
            |
             --20.54%--new_sync_write
                       ext4_file_write_iter
```
so I do not believe it is a problem with fio. 

It does not seem to matter what the VM config is. Firecracker is master as of 31 August 2020.

## Expected behaviour

I would have expected that synchronous IO in the VM would be worse than on the host. 

## Environment
Firecracker v0.21.0-473-gf2f6d8f8
Host: Android 4.14
Guest: 5.7 kernel.org
rootfs: Alpine 
Arch: arm64
perf version 4.14.133.g2e3484a
fio-3.16-32-g8c302

Running on an Android system with a 2 core A57 with a 14gig eMMC. 
```
[    0.490202] mmc0: new HS400 MMC card at address 0001
[...]

Issue #2188, opened 2020-10-15
Log errors for block device close()


alindima:
The file backing Firecracker's block device, like any other rust `File` is closed when going out of scope or when calling `drop` on the `File`, without any error handling: https://github.com/rust-lang/rust/blob/master/library/std/src/sys/unix/fd.rs#L284.

This may be a design flaw of rust's `File` struct, since the linux man page states that:

> Not checking the return value of close() is a common but nevertheless serious programming error. It is quite possible that errors on a previous write(2) operation are first reported at the final close(). Not checking the return value when closing the file may lead to silent loss of data.

One thing we can do to ensure that the subsequent `close()` will run successfully is calling `fsync()` on the file descriptor in advance.
This can be done as a `Drop` implementation for the `DiskProperties` member struct of the block device.

As far as I can tell, there is not much we can do to handle the potential error and data loss, apart from logging it, which may still be useful.

jain98:
Hi, first timer here. I'd be interested in giving this a shot.

berciuliviu:
Hey @jain98! I wonder if you are still looking/working on this issue? If not, I could tackle it since I've got some time on my hands.
Thanks!

jain98:
@berciuliviu feel free to take up the issue.

berciuliviu:
Linked with [2223](https://github.com/firecracker-microvm/firecracker/pull/2223)

berciuliviu:
Closed as duplicate to issue [2207](https://github.com/firecracker-microvm/firecracker/issues/2207)

PR #2447, opened 2021-02-08
[Block] Add option to configure block device flush

georgepisaltu:
# Reason for This PR

#2172 

## Description of Changes

This commit introduces the notion of cache type to the virtio block device implementation. This allows users to choose from two cache types:

- `unsafe`: this option will advertise the flush virtio capability to the guest block driver, but will acknowledge and ignore incoming flush requests;
- `writethrough`: this option will advertise the flush virtio capability to the guest block driver and will perform a `fsync` syscall on incoming flush requests;

This functionality is exposed to users through an optional `ignore_flush` field in the block configuration API request. When absent, this field defaults to `true`, ignoring incoming flush requests from the guest driver and keeping the behavior of the block device the same as before this change. Setting it to `false` will make the device call `fsync` on the backing file when processing flush requests. 

## License Acceptance

## PR Checklist

### DIFF
--- a/docs/api_requests/block-caching.md
+++ b/docs/api_requests/block-caching.md
@@ -1,0 +1,70 @@
+# Block device caching strategies
+
+Firecracker offers the possiblity of choosing the block device caching
+strategy. Caching strategy affects the path data written from inside the
+microVM takes to the host persistent storage.
+
+## How it works
+
+When installing a block device through a PUT /drives API call, users can choose
+the caching strategy by inserting a `cache_type` field in the JSON body of the
+request. The available cache types are:
+
+- `Unsafe`
+- `Writeback`
+
+### Unsafe mode (default)
+
+When configuring the block caching strategy to `Unsafe`, the device will
+advertise the VirtIO `flush` feature to the guest driver. If negotiated when
+activating the device, the guest driver will be able to send flush requests
+to the device, but the device will just acknowledge the request without
+actually performing any flushing on the host side. The data which was not
+yet committed to disk will continue to reside in the host page cache.
+
+### Writeback mode
+
+When configuring the block caching strategy to `Writeback`, the device will
+advertise the VirtIO `flush` feature to the guest driver. If negotiated when
+activating the device, the guest driver will be able to send flush requests
+to the device. When the device executes a flush request, it will perform an
+`fsync` syscall on the backing block file, committing all data in the host
+page cache to disk.
+
+## Supported use cases
+
+The caching strategy should be used in order to make a trade-off:
+
+- `Unsafe`
+  - enhances performance as fewer syscalls and IO operations are performed when
+    running workloads
+  - sacrifices data integrity in situations where the host simply loses the
+    contents of the page cache without committing them to the backing storage
+    (such as a power outage)
+  - recommended for use cases with ephemeral storage, such as serverless
+    environments
+- `Writeback`
+  - ensures that once a flush request was acknowledged by the host, the data
+    is committed to the backing storage
+  - sacrifices performance, from boot time increases to greater
+    emulation-related latencies when running workloads
+  - recommended for use cases with low power environments, such as embedded
+    environments
+
+## How to configure it
+
+Example sequence that configures a block device with a caching strategy:
+
+```bash
+curl --unix-socket ${socket} -i \
+     -X PUT "http://localhost/drives/dummy" \
+     -H "accept: application/json" \
+     -H "Content-Type: application/json" \
+     -d "{
+             \"drive_id\": \"dummy\",
+             \"path_on_host\": \"${drive_path}\",
+             \"is_root_device\": false,
+             \"is_read_only\": false,
+             \"cache_type\": \"Writeback\"
+         }"
+```
--- a/src/api_server/src/request/drive.rs
+++ b/src/api_server/src/request/drive.rs
@@ -216,20 +216,33 @@ mod tests {
         assert!(parse_put_drive(&Body::new("invalid_payload"), None).is_err());
         assert!(parse_put_drive(&Body::new("invalid_payload"), Some(&"id")).is_err());
 
-        // PATCH with invalid fields.
+        // PUT with invalid fields.
         let body = r#"{
                 "drive_id": "bar",
                 "is_read_only": false
               }"#;
         assert!(parse_put_drive(&Body::new(body), Some(&"2")).is_err());
 
-        // PATCH with invalid types on fields. Adding a drive_id as number instead of string.
+        // PUT with missing all optional fields.
+        let body = r#"{
+            "drive_id": "1000",
+            "path_on_host": "dummy",
+            "is_root_device": true,
+            "is_read_only": true
+        }"#;
+        assert!(parse_put_drive(&Body::new(body), Some(&"1000")).is_ok());
+
+        // PUT with invalid types on fields. Adding a drive_id as number instead of string.
+        assert!(parse_put_drive(&Body::new(body), Some(&"foo")).is_err());
+
+        // PUT with the complete configuration.
         let body = r#"{
                 "drive_id": "1000",
                 "path_on_host": "dummy",
                 "is_root_device": true,
                 "partuuid": "string",
                 "is_read_only": true,
+                "cache_type": "Unsafe",
                 "rate_limiter": {
                     "bandwidth": {
                         "size": 0,
@@ -244,7 +257,5 @@ mod tests {
                 }
             }"#;
         assert!(parse_put_drive(&Body::new(body), Some(&"1000")).is_ok());
-
-        assert!(parse_put_drive(&Body::new(body), Some(&"foo")).is_err());
     }
 }
--- a/src/api_server/swagger/firecracker.yaml
+++ b/src/api_server/swagger/firecracker.yaml
@@ -753,6 +753,11 @@ definitions:
     properties:
       drive_id:
         type: string
+      cache_type:
+        type: string
+        description:
+          Represents the caching strategy for the block device.
+        default: "Unsafe"
       is_read_only:
         type: boolean
       is_root_device:
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -30,8 +30,29 @@ use super::{
 use crate::virtio::VIRTIO_MMIO_INT_CONFIG;
 use crate::Error as DeviceError;
 
+use serde::{Deserialize, Serialize};
+
+/// Configuration options for disk caching.
+#[derive(Clone, Copy, Debug, Deserialize, PartialEq, Serialize)]
+pub enum CacheType {
+    /// Flushing mechanic will be advertised to the guest driver, but
+    /// the operation will be a noop.
+    Unsafe,
+    /// Flushing mechanic will be advertised to the guest driver and
+    /// flush requests coming from the guest will be performed using
+    /// `fsync`.
+    Writeback,
+}
+
+impl Default for CacheType {
+    fn default() -> CacheType {
+        CacheType::Unsafe
+    }
+}
+
 /// Helper object for setting up all `Block` fields derived from its backing file.
 pub(crate) struct DiskProperties {
+    cache_type: CacheType,
     file_path: String,
     file: File,
     nsectors: u64,
@@ -39,7 +60,11 @@ pub(crate) struct DiskProperties {
 }
 
 impl DiskProperties {
-    pub fn new(disk_image_path: String, is_disk_read_only: bool) -> io::Result<Self> {
+    pub fn new(
+        disk_image_path: String,
+        is_disk_read_only: bool,
+        cache_type: CacheType,
+    ) -> io::Result<Self> {
         let mut disk_image = OpenOptions::new()
             .read(true)
             .write(!is_disk_read_only)
@@ -57,6 +82,7 @@ impl DiskProperties {
         }
 
         Ok(Self {
+            cache_type,
             nsectors: disk_size >> SECTOR_SHIFT,
             image_id: Self::build_disk_image_id(&disk_image),
             file_path: disk_image_path,
@@ -121,6 +147,31 @@ impl DiskProperties {
         }
         config
     }
+
+    pub fn cache_type(&self) -> CacheType {
+        self.cache_type
+    }
+}
+
+impl Drop for DiskProperties {
+    fn drop(&mut self) {
+        match self.cache_type {
+            CacheType::Writeback => {
+                // flush() first to force any cached data out.
+                if self.file.flush().is_err() {
+                    error!("Failed to flush block data on drop.");
+                }
+                // Sync data out to physical media on host.
+                if self.file.sync_all().is_err() {
+                    error!("Failed to sync block data on drop.")
+                }
+                METRICS.block.flush_count.inc();
+            }
+            CacheType::Unsafe => {
+                // This is a noop.
+            }
+        };
+    }
 }
 
 /// Virtio device for exposing block level read/write operations on a host file.
@@ -155,12 +206,13 @@ impl Block {
     pub fn new(
         id: String,
         partuuid: Option<String>,
+        cache_type: CacheType,
         disk_image_path: String,
         is_disk_read_only: bool,
         is_disk_root: bool,
         rate_limiter: RateLimiter,
     ) -> io::Result<Block> {
-        let disk_properties = DiskProperties::new(disk_image_path, is_disk_read_only)?;
+        let disk_properties = DiskProperties::new(disk_image_path, is_disk_read_only, cache_type)?;
 
         let mut avail_features = (1u64 << VIRTIO_F_VERSION_1) | (1u64 << VIRTIO_BLK_F_FLUSH);
 
@@ -343,7 +395,8 @@ impl Block {
 
     /// Update the backing file and the config space of the block device.
     pub fn update_disk_image(&mut self, disk_image_path: String) -> io::Result<()> {
-        let disk_properties = DiskProperties::new(disk_image_path, self.is_read_only())?;
+        let disk_properties =
+            DiskProperties::new(disk_image_path, self.is_read_only(), self.cache_type())?;
         self.disk = disk_properties;
         self.config_space = self.disk.virtio_block_config_space();
 
@@ -379,6 +432,10 @@ impl Block {
     /// Specifies if this block device is read only.
     pub fn is_root_device(&self) -> bool {
         self.root_device
+    }
+
+    pub fn cache_type(&self) -> CacheType {
+        self.disk.cache_type()
     }
 }
 
@@ -491,8 +548,12 @@ pub(crate) mod tests {
         let size = SECTOR_SIZE * num_sectors;
         f.as_file().set_len(size).unwrap();
 
-        let disk_properties =
-            DiskProperties::new(String::from(f.as_path().to_str().unwrap()), true).unwrap();
+        let disk_properties = DiskProperties::new(
+            String::from(f.as_path().to_str().unwrap()),
+            true,
+            CacheType::Unsafe,
+        )
+        .unwrap();
 
         assert_eq!(size, SECTOR_SIZE * num_sectors);
         assert_eq!(disk_properties.nsectors, num_sectors);
@@ -504,7 +565,9 @@ pub(crate) mod tests {
         // Testing `backing_file.virtio_block_disk_image_id()` implies
         // duplicating that logic in tests, so skipping it.
 
-        assert!(DiskProperties::new("invalid-disk-path".to_string(), true).is_err());
+        assert!(
+            DiskProperties::new("invalid-disk-path".to_string(), true, CacheType::Unsafe).is_err()
+        );
     }
 
     #[test]
--- a/src/devices/src/virtio/block/persist.rs
+++ b/src/devices/src/virtio/block/persist.rs
@@ -7,9 +7,10 @@ use std::io;
 use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
 
+use logger::warn;
 use rate_limiter::{persist::RateLimiterState, RateLimiter};
 use snapshot::Persist;
-use versionize::{VersionMap, Versionize, VersionizeResult};
+use versionize::{VersionMap, Versionize, VersionizeError, VersionizeResult};
 use versionize_derive::Versionize;
 use virtio_gen::virtio_blk::VIRTIO_BLK_F_RO;
 use vm_memory::GuestMemoryMmap;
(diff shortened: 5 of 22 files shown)


src/devices/src/virtio/block/device.rs
```diff
@@ -30,16 +30,39 @@ use super::{
...
+/// Configuration options for disk caching.
+#[derive(Clone, Copy)]
+pub enum CacheType {
+    /// Flushing mechanic will be advertised to the guest driver, but
+    /// the operation will be a noop.
+    Unsafe,
+    /// Flushing mechanic will be advertised to the guest driver and
+    /// flush requests coming from the guest will be performed using
+    /// `fsync`.
+    Writethrough,
```
acatangiu, line 42: `writethrough` is not the best name since that usually means **no host FS caching** which is not the case here.

Unfortunately I don't have a better name right now 😄 Suggestions are welcome!

src/devices/src/virtio/block/device.rs
```diff
@@ -146,6 +174,7 @@ pub struct Block {
     pub(crate) partuuid: Option<String>,
     pub(crate) root_device: bool,
     pub(crate) rate_limiter: RateLimiter,
+    pub(crate) ignore_flush: bool,
```
acatangiu, line 177: Why not use the same `enum CacheType` here as well? Ideally even use it from API all the way down without transformations.

src/devices/src/virtio/block/persist.rs
```diff
@@ -24,12 +24,34 @@ use crate::virtio::{DeviceState, TYPE_BLOCK};
...
     virtio_state: VirtioDeviceState,
     rate_limiter_state: RateLimiterState,
 }
 
+impl BlockState {
+    fn block_ignore_flush_ser(&mut self, target_version: u16) -> VersionizeResult<()> {
+        if target_version < 3 && !self.ignore_flush {
+            return Err(VersionizeError::Semantic(
+                "Target version does not implement writethrough block cache.".to_owned(),
+            ));
```
acatangiu, line 44: Should we make this backward-compatible, where the cache/flush just reverts to old behavior (also print a `warn!` or smth)?

src/api_server/swagger/firecracker.yaml
```diff
@@ -753,6 +753,11 @@ definitions:
     properties:
       drive_id:
         type: string
+      ignore_flush:
+        type: boolean
+        description:
+          Makes the device ignore flush requests coming from the guest driver.
+        default: true
```
acatangiu, line 760: Why not use the `CacheType` enum here instead of this very specific bool?

The enum is forward-looking as it allows us to add options to it or change implementation details for each `CacheType`.
This bool is clearer as it is self-explanatory, but also limited and not extensible.

src/devices/src/virtio/block/device.rs
alindima: I think the right naming is `writeback`. This is the term QEMU uses as well: https://documentation.suse.com/sles/11-SP4/html/SLES-kvm4zseries/cha-qemu-cachemodes.html#id-1.10.6.6.5

georgepisaltu: `writeback` is indeed the right name. I copy-pasted the wrong one the first time around 😆. Nice catch.

src/devices/src/virtio/block/device.rs
georgepisaltu: Only reason was that `ignore_flush` seemed like a perfect name for the field.

Changed it so that we use `cache_type` as a field in the API request JSON, all the way down to the device.

src/devices/src/virtio/block/persist.rs
georgepisaltu: I am leaning towards no because of the risk that users might shoot themselves in the foot, but I can see how we can enable this and print a warning or something and let them handle it.

I guess we can do either, I'm on the fence about this. Other opinions?

src/api_server/swagger/firecracker.yaml
georgepisaltu: Changed it so that it uses `cache_type`.

src/devices/src/virtio/block/persist.rs
acatangiu: @firecracker-microvm/compute-capsule

georgepisaltu: I went ahead and changed it so that it's only a warning now. Also added an integ test for this case.

@firecracker-microvm/compute-capsule please raise concerns if you have any.

alindima:
This PR also partially fixes: #2207.
In order to fully fix it, we should also tackle in this PR this related issue: https://github.com/firecracker-microvm/firecracker/issues/2188

The block device caching strategy could also influence the decision of calling `fsync` on `DiskProperties::drop`, so that we make sure that all data has been flushed before closing the backing file, WDYT @acatangiu @georgepisaltu ?

acatangiu:
> This PR also partially fixes: #2207.
> In order to fully fix it, we should also tackle in this PR this related issue: #2188
> 
> The block device caching strategy could also influence the decision of calling `fsync` on `DiskProperties::drop`, so that we make sure that all data has been flushed before closing the backing file, WDYT @acatangiu @georgepisaltu ?

@2207 is fully fixed by this PR in my view. We have seen no customer usecase for `O_SYNC/O_DSYNC` so we can update the issue and close it as fixed by this PR.

Indeed, for completeness, we should add the `fsync` on `DiskProperties::drop` for disks where configured caching is `writeback`.

alindima:
> > This PR also partially fixes: #2207.
> > In order to fully fix it, we should also tackle in this PR this related issue: #2188
> > The block device caching strategy could also influence the decision of calling `fsync` on `DiskProperties::drop`, so that we make sure that all data has been flushed before closing the backing file, WDYT @acatangiu @georgepisaltu ?
> 
> @2207 is fully fixed by this PR in my view. We have seen no customer usecase for `O_SYNC/O_DSYNC` so we can update the issue and close it as fixed by this PR.
> 
> Indeed, for completeness, we should add the `fsync` on `DiskProperties::drop` for disks where configured caching is `writeback`.

Yes, I agree. I meant that we can close #2207 if we also add the `drop` implementation, not necessarily `O_SYNC` support

georgepisaltu:
Added a `fsync` call on `DiskProperties::drop`.

src/vmm/src/version_map.rs
```diff
@@ -17,6 +18,7 @@ lazy_static! {
     pub static ref VERSION_MAP: VersionMap = {
         let mut version_map = VersionMap::new();
         version_map.new_version().set_type_version(DeviceStates::type_id(), 2);
+        version_map.new_version().set_type_version(BlockState::type_id(), 3);
```
acatangiu, line 21: I think this commit should be squashed together with the one changing block state struct.

src/devices/src/virtio/block/persist.rs
```diff
@@ -19,17 +20,63 @@ use super::*;
 use crate::virtio::persist::VirtioDeviceState;
 use crate::virtio::{DeviceState, TYPE_BLOCK};
 
+#[derive(Clone, Copy, Versionize, PartialEq)]
+// NOTICE: Any changes to this structure require a snapshot version bump.
+pub enum CacheTypeState {
+    Unsafe,
+    Writeback,
+}
```
acatangiu, line 28: nitish: this being a simple enum, we could have just used the original instead of creating a shadow _state_ one.

src/devices/src/virtio/block/persist.rs
```diff
@@ -19,17 +20,63 @@ use super::*;
...
+            CacheType::Writeback => CacheTypeState::Writeback,
+        }
+    }
+
+    fn create_cache_type(&self) -> CacheType {
+        match self {
+            CacheTypeState::Unsafe => CacheType::Unsafe,
+            CacheTypeState::Writeback => CacheType::Writeback,
+        }
+    }
```
acatangiu, line 43: nit: idiomatically better to implement [std::convert::From](https://doc.rust-lang.org/std/convert/trait.From.html).

src/devices/src/virtio/block/persist.rs
```diff
@@ -19,17 +20,63 @@ use super::*;
...
+
 #[derive(Clone, Versionize)]
 // NOTICE: Any changes to this structure require a snapshot version bump.
 pub struct BlockState {
     id: String,
     partuuid: Option<String>,
+    #[version(
+        start = 3,
+        ser_fn = "block_cache_type_ser",
+        default_fn = "default_cache_type_flush"
```
acatangiu, line 54: do we need a default_fn ?

src/devices/src/virtio/block/persist.rs
```diff
@@ -43,6 +90,7 @@ impl Persist<'_> for Block {
         BlockState {
             id: self.id.clone(),
             partuuid: self.partuuid.clone(),
+            cache_type: CacheTypeState::from_cache_type(self.cache_type()),
```
acatangiu, line 93: nit: could have been
```diff
- cache_type: CacheTypeState::from_cache_type(self.cache_type()),
+ cache_type: self.cache_type().into(),
```

tests/integration_tests/build/test_coverage.py
```diff
@@ -23,7 +23,7 @@
 # this contains the frequency while on AMD it does not.
 # Checkout the cpuid crate. In the future other
 # differences may appear.
-COVERAGE_DICT = {"Intel": 85.44, "AMD": 84.77, "ARM": 83.42}
+COVERAGE_DICT = {"Intel": 85.44, "AMD": 84.68, "ARM": 83.52}
```
acatangiu, line 26: pls coverage commit to integ tests commit (or some other)

tests/integration_tests/performance/test_versioned_serialization_benchmark.py
```diff
@@ -28,7 +28,7 @@
             },
             "crc": {
                 "target": 0.213,  # milliseconds
-                "delta": 0.020  # milliseconds
+                "delta": 0.025  # milliseconds
```
acatangiu, line 31: pls explain in commit why we're doing a 25% increase in delta here.

For sanity, please run the benchmark (can even do it locally) with and without these changes and document in commit description if/what are the changes' impact if any.

acatangiu (commented):
Code looks good, only nits.

PR is missing documentation. The newly available caching strategies need to be included in block device documentation.

src/devices/src/virtio/block/persist.rs
acatangiu: yes, we need it, nvm

src/devices/src/virtio/block/persist.rs
georgepisaltu: My understanding is that we should not include any `persist` logic in our devices, so it didn't make sense to add a `#[derive(Versionize)]` on `CacheType`.

Furthermore, `CacheType` is a part of the block device and has little to do with any snapshot logic, so it didn't make sense to have it in `block/persist.rs` either.

This was the reasoning behind creating a shadow struct for `CacheType`.

src/devices/src/virtio/block/persist.rs
georgepisaltu: Done.

src/devices/src/virtio/block/persist.rs
georgepisaltu: Done.

src/vmm/src/version_map.rs
georgepisaltu: Done.

tests/integration_tests/build/test_coverage.py
georgepisaltu: Done.

src/devices/src/virtio/block/persist.rs
```diff
@@ -19,17 +20,63 @@ use super::*;
...
+    }
+}
+
 #[derive(Clone, Versionize)]
 // NOTICE: Any changes to this structure require a snapshot version bump.
 pub struct BlockState {
     id: String,
     partuuid: Option<String>,
+    #[version(
+        start = 3,
```
sandreim, line 52: This is actually version 2 as it refers to the struct version, not the app version.

src/devices/src/virtio/block/persist.rs
```diff
@@ -19,17 +20,63 @@ use super::*;
...
+    cache_type: CacheTypeState,
     root_device: bool,
     disk_path: String,
     virtio_state: VirtioDeviceState,
     rate_limiter_state: RateLimiterState,
 }
 
+impl BlockState {
+    fn block_cache_type_ser(&mut self, target_version: u16) -> VersionizeResult<()> {
+        if target_version < 3 && self.cache_type != CacheTypeState::Unsafe {
```
sandreim, line 65: This needs coverage in unit test.

src/vmm/src/version_map.rs
sandreim: BlockState has version 2.

src/devices/src/virtio/block/persist.rs
```diff
@@ -19,17 +20,63 @@ use super::*;
...
+                "Target version does not implement the current cache type. \
+                Defaulting to \"unsafe\" mode."
+            );
+        }
+
+        Ok(())
+    }
+
+    fn default_cache_type_flush(_source_version: u16) -> CacheTypeState {
+        CacheTypeState::Unsafe
```
sandreim, line 76: This also needs coverage.

tests/integration_tests/performance/test_versioned_serialization_benchmark.py
georgepisaltu: I ran it and in 23 tries there was one failure on serializing with CRC on Intel `x86_64`. The test was off by `0.00001`, or about 0.05% of the original delta.

Nevertheless, the fail happened and this change is to ensure we don't have any intermittent failures on this in the future.

While adding a bit of extra logic in `block/persist.rs` certainly won't speed ser/deser up, it doesn't seem to degrade the averages too much. From my measurements, there seems to be a difference of about `0.0001` on average between the means before and after the change, which is less than 0.1%. I believe this failure could have happened without the change simply because the bench parameters were a bit too strict, or it may have just been a one time thing where the machine I was testing on had a lazy CPU cycle or two :grinning: .

georgepisaltu (commented):
Also added the documentation in `docs/api_requests/` as the block device doesn't have its own doc. I will create the documentation for the block device and move the cache documentation in a subsequent PR to not block this one as they are not related.

src/devices/src/virtio/block/persist.rs
georgepisaltu: Changed it.

src/devices/src/virtio/block/persist.rs
georgepisaltu: It already has coverage in a unit test.

src/devices/src/virtio/block/persist.rs
georgepisaltu: Added a unit test for this.

src/vmm/src/version_map.rs
georgepisaltu: Changed it.

tests/integration_tests/functional/test_drives.py
```diff
@@ -337,6 +338,134 @@ def test_patch_drive(test_microvm_with_ssh, network_config):
...
+    metrics_fifo_path = os.path.join(test_microvm.path, 'metrics_fifo')
+    metrics_fifo = log_tools.Fifo(metrics_fifo_path)
+    response = test_microvm.metrics.put(
+        metrics_path=test_microvm.create_jailed_resource(metrics_fifo.path)
+    )
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+
+    test_microvm.start()
+
+    # Verify no flush commands were generated during boot.
```
alindima, line 370: actually, the commands are generated but they are ignored

src/devices/src/virtio/block/device.rs
```diff
@@ -146,6 +197,7 @@ pub struct Block {
     pub(crate) partuuid: Option<String>,
     pub(crate) root_device: bool,
     pub(crate) rate_limiter: RateLimiter,
+    pub(crate) cache_type: CacheType,
```
acatangiu, line 200: this is already part of `DiskProperties`, why duplicate here?

src/devices/src/virtio/block/persist.rs
acatangiu: > My understanding is that we should not include any persist logic in our devices

I don't think that's accurate. Plain-old-data structs can be versioned directly I think. @sandreim can confirm/disprove this, I might be missing something.

nit: In current PR `CacheType` already has versionize.

docs/api_requests/block-caching.md
```diff
@@ -0,0 +1,70 @@
...
+- `Writeback`
+
+### Unsafe mode (default)
+
+When configuring the block caching strategy to `Unsafe`, the device will
+advertise the VirtIO `flush` feature to the guest driver. If negotiated when
+activating the device, the guest driver will be able to send flush requests
+to the device, but the device will just acknowledge the request without
+actually performing any flushing on the host side. The data which was not
+yet committed to disk will continue to reside in the host page cache.
```
alindima, line 23: Maybe note here that the data may is only committed to the physical disk when the host driver decides to

src/devices/src/virtio/block/persist.rs
sandreim: My take on this is to keep our devices clean of any serialization logic. This is mostly forward looking towards ingesting upstream rust-vmm devices more easily.

docs/api_requests/block-caching.md
georgepisaltu: I think an explanation of how caching works on Linux systems is not appropriate here, especially since whatever we write in this section highly depends on how the host OS and FS are set up. The fact that IO is cached is the default behavior in Linux systems and I don't think we need to explain this here.

We should, however, touch on this in the general block device documentation and provide our recommendations regarding host OS and FS setup. In that section we should also mention what host configuration we are running our tests on, and that should be enough.

src/api_server/src/request/drive.rs
```diff
@@ -216,20 +216,33 @@ mod tests {
...
+        assert!(parse_put_drive(&Body::new(body), Some(&"foo")).is_err());
+
+        // PUT with the complete configuration.
         let body = r#"{
                 "drive_id": "1000",
                 "path_on_host": "dummy",
                 "is_root_device": true,
                 "partuuid": "string",
                 "is_read_only": true,
+                "cache_type": "Unsafe",
```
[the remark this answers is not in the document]
georgepisaltu: This is how `serde::Serialize` works. The enum variants are `Unsafe` and `Writeback` due to Rust style and `serde` just picks this up and converts it to a string JSON.

There is also precedence regarding upper-case JSON fields when using `serde`, the most common example being `snapshot_type: "Full"`.

src/devices/src/virtio/block/device.rs
```diff
@@ -121,6 +147,31 @@ impl DiskProperties {
...
+                    error!("Failed to flush block data on drop.");
+                }
+                // Sync data out to physical media on host.
+                if self.file.sync_all().is_err() {
+                    error!("Failed to sync block data on drop.")
+                }
+                METRICS.block.flush_count.inc();
+            }
+            CacheType::Unsafe => {
+                // This is a noop.
```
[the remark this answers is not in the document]
georgepisaltu: There is no need to include `file.flush()` because `std::fs::File` is not buffered and that is by design. It works the same way an FD would work in C when using `open`, `read` and `write`.

For buffered IO in C, one would use `fopen`, `fread` and `fwrite`. In Rust, this is achieved by using `BufReader` and `BufWriter`.

This means that all data written with a `file.write()` will reach the cache as long as the file is not opened with `O_DIRECT` or `O_DSYNC` (which is not the case in Firecracker), and this is exactly the reason the Rust `file.flush()` is a noop. You can check this and the non-buffered behavior of `file.flush()` using `strace`.

Moreover, I don't think the argument that the `File` implementation might change supports adding `file.flush()`. If anything, it prohibits it. Right now, `file.flush()` is a noop, which means the behviour is the same as not having it. If it does _anything_ in the future and we add it now in the implementation, it would create another bug.

src/devices/src/virtio/block/device.rs
georgepisaltu: You're right, I removed the duplicate.

src/devices/src/virtio/block/device.rs
```diff
@@ -380,6 +435,10 @@ impl Block {
     pub fn is_root_device(&self) -> bool {
         self.root_device
     }
+
+    pub fn cache_type(&self) -> CacheType {
+        self.cache_type
```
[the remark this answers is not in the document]
georgepisaltu: You're right, I removed the duplicate.

src/vmm/src/builder.rs
```diff
@@ -1256,6 +1288,26 @@ pub mod tests {
...
+
+        // Use case 7: root block device is rw with flush enabled
+        {
+            let drive_id = String::from("root");
+            let block_configs = vec![CustomBlockConfig::new(
+                drive_id.clone(),
+                true,
+                None,
+                false,
+                CacheType::Unsafe,
```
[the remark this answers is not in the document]
georgepisaltu: Should have been `Writeback` in the code, the comment was right. Changed it.

src/vmm/src/rpc_interface.rs
```diff
@@ -1065,6 +1067,7 @@ mod tests {
             path_on_host: String::new(),
             is_root_device: false,
             partuuid: None,
+            cache_type: CacheType::Unsafe,
```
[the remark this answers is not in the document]
georgepisaltu: Because the purpose of this test is to ensure a request to add a second root device will fail, it does not test any block capability.

The parsing side of this is tested in the `api_server`.

src/vmm/src/vmm_config/drive.rs
```diff
@@ -79,10 +81,21 @@ pub struct BlockDeviceConfig {
...
+    pub cache_type: CacheType,
     /// Rate Limiter for I/O operations.
     pub rate_limiter: Option<RateLimiterConfig>,
 }
 
+// Serde does not allow specifying a default value for a field
+// that is not required. The workaround is to specify a function
+// that returns the value.
+fn default_cache_type() -> CacheType {
+    CacheType::Unsafe
```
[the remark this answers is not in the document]
georgepisaltu: I thought we can't but it seems `serde` is smarter than I thought. Changed it.

tests/integration_tests/functional/test_drives.py
georgepisaltu: Indeed, changed it.

src/api_server/src/request/drive.rs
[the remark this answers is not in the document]
alindima: I was thinking that having case-sensitive values is more error-prone. and serde has a rename directive that can take lowercase as value: for example one can add this on the enum:

`#[serde(rename_all = "lowercase")]`

I'm fine with keeping it CamelCase though.

src/vmm/src/version_map.rs
```diff
@@ -26,6 +28,7 @@ lazy_static! {
         let mut mapping = HashMap::new();
         mapping.insert(String::from("0.23.0"), 1);
         mapping.insert(String::from("0.24.0"), 2);
+        mapping.insert(String::from("0.25.0"), 3);
```
sandreim, line 31: Please also include the binary artifacts for running snapshot integration tests across versions.

georgepisaltu: Once I get all approves, I will build and publish the artifacts so they are built from the same source as what is going to get merged.

src/devices/src/virtio/block/device.rs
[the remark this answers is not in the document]
alindima: My arguments were also rooted in the fact that for `CacheType::WriteBack`, we do a `file.flush()` before calling f`sync()`. why is that?

Even if this a noop, this just follows the semantic of the `flush` method of the `Write` trait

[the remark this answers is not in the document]
georgepisaltu: It is just future proofing. I thought about the semantics of `File::flush` possibly changing in the future and I wanted to be on the safe side. The rationale was that it has no performance cost now, as it is a noop, and could save us some trouble in the future in the unlikely event that flush semantics change.

I would agree to removing it, I have no strong preference here. Let me know what you think.

acatangiu (approved):

alindima (approved):

PR #2453, opened 2021-02-10
Fix non-compliant checks inside the RTC PL031 device

dianpopa:
# Reason for This PR

This PR solves 2 problems:
* The RTC Pl031 implementation uses an u32 counter and our implementation would not check for a fixed 4-sized buffer.
* Removes unnecessary interrupt assertion

## Description of Changes

- unify appearance for read/write function
- cover corner cases in unit tests
- check for a fixed 4-sized buffer like the RTC PL031 specification asks for

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/legacy/rtc_pl031.rs
+++ b/src/devices/src/legacy/rtc_pl031.rs
@@ -199,6 +199,19 @@ mod tests {
         let mut rtc = RTC::new(EventFd::new(libc::EFD_NONBLOCK).unwrap());
         let mut data = [0; 4];
 
+        // Read and write to the DR register. Since this is a RO register, the write
+        // function should fail.
+        byte_order::write_le_u32(&mut data, 0);
+        let err_cnt_before_write = METRICS.rtc.error_count.count();
+        rtc.write(RTCDR, &data);
+        let err_cnt_after_write = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_write - err_cnt_before_write, 1);
+        rtc.read(RTCDR, &mut data);
+        let err_cnt_after_read = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_write, err_cnt_after_read);
+        let v_read = byte_order::read_le_u32(&data[..]);
+        assert_ne!(v_read, 0);
+
         // Read and write to the MR register.
         byte_order::write_le_u32(&mut data, 123);
         rtc.write(RTCMR, &data);
@@ -235,6 +248,30 @@ mod tests {
         rtc.read(RTCIMSC, &mut data);
         let v = byte_order::read_le_u32(&data[..]);
         assert_eq!(0, v);
+
+        // Read and write to the RIS RO register.
+        byte_order::write_le_u32(&mut data, 1);
+        let err_cnt_before_write = METRICS.rtc.error_count.count();
+        rtc.write(RTCRIS, &data);
+        let err_cnt_after_write = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_write - err_cnt_before_write, 1);
+        rtc.read(RTCRIS, &mut data);
+        let err_cnt_after_read = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_write, err_cnt_after_read);
+        let v_read = byte_order::read_le_u32(&data[..]);
+        assert_ne!(v_read, 1);
+
+        // Read and write to the MIS RO register.
+        byte_order::write_le_u32(&mut data, 1);
+        let err_cnt_before_write = METRICS.rtc.error_count.count();
+        rtc.write(RTCMIS, &data);
+        let err_cnt_after_write = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_write - err_cnt_before_write, 1);
+        rtc.read(RTCMIS, &mut data);
+        let err_cnt_after_read = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_write, err_cnt_after_read);
+        let v_read = byte_order::read_le_u32(&data[..]);
+        assert_ne!(v_read, 1);
 
         // Read and write to the ICR register.
         byte_order::write_le_u32(&mut data, 1);
--- a/src/devices/src/legacy/rtc_pl031.rs
+++ b/src/devices/src/legacy/rtc_pl031.rs
@@ -162,29 +162,35 @@ impl RTC {
 
 impl BusDevice for RTC {
     fn read(&mut self, offset: u64, data: &mut [u8]) {
-        if data.len() <= 4 {
+        if data.len() == 4 {
             match self.handle_read(offset) {
                 Ok(val) => byte_order::write_le_u32(data, val),
                 Err(e) => {
-                    warn!("Failed to read from the RTC PL031 device: {}", e);
+                    warn!("Failed to read from the RTC: {}", e);
                     METRICS.rtc.error_count.inc();
                 }
             }
         } else {
-            warn!("Invalid RTC PL031 data length {}", data.len());
+            warn!(
+                "Found invalid data length while trying to read from the RTC: {}",
+                data.len()
+            );
             METRICS.rtc.error_count.inc();
         }
     }
 
     fn write(&mut self, offset: u64, data: &[u8]) {
-        if data.len() <= 4 {
+        if data.len() == 4 {
             let v = byte_order::read_le_u32(&data[..]);
             if let Err(e) = self.handle_write(offset, v) {
-                warn!("Failed to write to the RTC PL031 device: {}", e);
+                warn!("Failed to write to the RTC: {}", e);
                 METRICS.rtc.error_count.inc();
             }
         } else {
-            warn!("Invalid RTC PL031 data length {}", data.len());
+            warn!(
+                "Found invalid data length while trying to write to the RTC: {}",
+                data.len()
+            );
             METRICS.rtc.error_count.inc();
         }
     }
@@ -308,5 +314,19 @@ mod tests {
         rtc.read(AMBA_ID_LOW, &mut data);
         let index = AMBA_ID_LOW + 3;
         assert_eq!(data[0], PL031_ID[((index - AMBA_ID_LOW) >> 2) as usize]);
+    }
+
+    #[test]
+    fn test_rtc_invalid_buf_len() {
+        let mut rtc = RTC::new(EventFd::new(libc::EFD_NONBLOCK).unwrap());
+        let mut data = [1; 2];
+
+        let err_cnt_before_write = METRICS.rtc.error_count.count();
+        rtc.write(RTCLR, &data);
+        let err_cnt_after_write = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_write - err_cnt_before_write, 1);
+        rtc.read(RTCLR, &mut data);
+        let err_cnt_after_read = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_read - err_cnt_after_write, 1);
     }
 }
--- a/src/devices/src/legacy/rtc_pl031.rs
+++ b/src/devices/src/legacy/rtc_pl031.rs
@@ -7,15 +7,11 @@
 //! This is achieved by generating an interrupt signal after counting for a programmed number of cycles of
 //! a real-time clock input.
 //!
-use std::fmt;
-use std::time::Instant;
-use std::{io, result};
+use std::{fmt, result, time::Instant};
 
 use crate::BusDevice;
 use logger::{warn, IncMetric, METRICS};
 use utils::byte_order;
-use utils::eventfd::EventFd;
-//use bus::Error;
 
 // As you can see in https://static.docs.arm.com/ddi0224/c/real_time_clock_pl031_r1p3_technical_reference_manual_DDI0224C.pdf
 // at section 3.2 Summary of RTC registers, the total size occupied by this device is 0x000 -> 0xFFC + 4 = 0x1000.
@@ -40,14 +36,12 @@ const AMBA_ID_HIGH: u64 = 0x1000;
 
 pub enum Error {
     BadOffset(u64, &'static str),
-    InterruptFailure(io::Error),
 }
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Error::BadOffset(offset, ops) => write!(f, "Bad {} offset: {}", ops, offset),
-            Error::InterruptFailure(e) => write!(f, "Failed to trigger interrupt: {}", e),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/devices/src/legacy/rtc_pl031.rs
```diff
@@ -171,26 +171,20 @@ impl BusDevice for RTC {
                 }
             }
         } else {
-            warn!(
-                "Invalid RTC PL031 data length {}",
-                data.len()
-            );
+            warn!("Invalid RTC PL031 data length {}", data.len());
```
georgepisaltu, line 174: Would it be worth it to state that this is a `read` warning in the message? Same for the `write`.

dianpopa: Good idea! I made the changes

src/devices/src/legacy/rtc_pl031.rs
```diff
@@ -236,6 +249,30 @@ mod tests {
...
+        rtc.write(RTCRIS, &data);
+        let err_cnt_after_write = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_write - err_cnt_before_write, 1);
+        rtc.read(RTCRIS, &mut data);
+        let err_cnt_after_read = METRICS.rtc.error_count.count();
+        assert_eq!(err_cnt_after_write, err_cnt_after_read);
+        let v_read = byte_order::read_le_u32(&data[..]);
+        assert_ne!(v_read, 1);
+
+        // Read and write to the RIS RO register.
```
alindima, line 270:
```suggestion
// Read and write to the MIS RO register
```

src/devices/src/legacy/rtc_pl031.rs
```diff
@@ -92,6 +91,33 @@ impl RTC {
...
+                off => {
+                    return Err(Error::BadOffset(off));
+                }
+            };
+        }
+        Ok(val)
+    }
+
     fn handle_write(&mut self, offset: u64, val: u32) -> Result<()> {
         match offset {
```
alindima, line 122: Not really tied to this PR, but the spec says that writing 1 to the RTCICR register clears the interrupt.
Writing 0 should not.
https://developer.arm.com/documentation/ddi0224/c/Programmers-model/General-registers/Interrupt-Clear-Register--RTCICR?lang=en

alindima: Also, register offsets are declared as u64, when I think they can even be of type `u8`

src/devices/src/legacy/rtc_pl031.rs
```diff
@@ -38,16 +38,15 @@ const PL031_ID: [u8; 8] = [0x31, 0x10, 0x04, 0x00, 0x0d, 0xf0, 0x05, 0xb1];
...
-    BadWriteOffset(u64),
+    BadOffset(u64),
     InterruptFailure(io::Error),
 }
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            Error::BadWriteOffset(offset) => write!(f, "Bad Write Offset: {}", offset),
+            Error::BadOffset(offset) => write!(f, "Bad Write Offset: {}", offset),
```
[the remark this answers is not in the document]
dianpopa: Done.

src/devices/src/legacy/rtc_pl031.rs
dianpopa: Done.

src/devices/src/legacy/rtc_pl031.rs
dianpopa: > Not really tied to this PR, but the spec says that writing 1 to the RTCICR register clears the interrupt.
> Writing 0 should not.
> https://developer.arm.com/documentation/ddi0224/c/Programmers-model/General-registers/Interrupt-Clear-Register--RTCICR?lang=en

> Not really tied to this PR, but the spec says that writing 1 to the RTCICR register clears the interrupt.
> Writing 0 should not.
> https://developer.arm.com/documentation/ddi0224/c/Programmers-model/General-registers/Interrupt-Clear-Register--RTCICR?lang=en

There is a problem with the RTCICR, but it is not related to how we set it. The RTCICR is used in conjunction with the RTCMR to implement a  time alarm function:
 ```
The Match Register is programmed by writing to RTCMR. The counter and match values are compared in a comparator. When both values are equal, the RTCINTR interrupt is asserted HIGH. An external processor can use the interrupt to implement a basic time alarm function. The interrupt is cleared by writing any data value to the Interrupt Clear Register, RTCICR. The value in the Match Register can be read at any time.
```
This is extra functionality that one can achieve with the RTC. As a consequence, our implementation does not need the RTCICR. It can easily be seen from the implementation that the value of self.ris never changes, so there is no need for the value to be cleared by the RTCICR as it will always be 0. I updated the code to be similar to how we proceed for the RTCMR register, take a look.

Comment (dianpopa) [reply src/devices/src/legacy/rtc_pl031.rs:122]: > Also, register offsets are declared as u64, when I think they can even be of type `u8`

Indeed the register offsets could be u8, but our interface works with u64. The offsets are u64 cause the `MmioRead(u64 /* address */, &'a mut [u8]),` reads it in an u64 and gives the value further to the `read` and `write` bus functions.

Comment (alindima) [reply src/devices/src/legacy/rtc_pl031.rs:122]: > > Not really tied to this PR, but the spec says that writing 1 to the RTCICR register clears the interrupt.
> > Writing 0 should not.
> > https://developer.arm.com/documentation/ddi0224/c/Programmers-model/General-registers/Interrupt-Clear-Register--RTCICR?lang=en
> 
> > Not really tied to this PR, but the spec says that writing 1 to the RTCICR register clears the interrupt.
> > Writing 0 should not.
> > https://developer.arm.com/documentation/ddi0224/c/Programmers-model/General-registers/Interrupt-Clear-Register--RTCICR?lang=en
> 
> There is a problem with the RTCICR, but it is not related to how we set it. The RTCICR is used in conjunction with the RTCMR to implement a time alarm function:
> 
> ```
> The Match Register is programmed by writing to RTCMR. The counter and match values are compared in a comparator. When both values are equal, the RTCINTR interrupt is asserted HIGH. An external processor can use the interrupt to implement a basic time alarm function. The interrupt is cleared by writing any data value to the Interrupt Clear Register, RTCICR. The value in the Match Register can be read at any time.
> ```
> 
> This is extra functionality that one can achieve with the RTC. As a consequence, our implementation does not need the RTCICR. It can easily be seen from the implementation that the value of self.ris never changes, so there is no need for the value to be cleared by the RTCICR as it will always be 0. I updated the code to be similar to how we proceed for the RTCMR register, take a look.


In the latest commit, I see that this was fixed.

Since we are not using the match functionality, can't we get rid of all interrupts?

Comment (alindima) [reply src/devices/src/legacy/rtc_pl031.rs:122]: > > Also, register offsets are declared as u64, when I think they can even be of type `u8`
> 
> Indeed the register offsets could be u8, but our interface works with u64. The offsets are u64 cause the `MmioRead(u64 /* address */, &'a mut [u8]),` reads it in an u64 and gives the value further to the `read` and `write` bus functions.

I see. I think a cast should do it though, right?

File: src/devices/src/legacy/rtc_pl031.rs [hunk 5]
```diff
@@ -142,10 +142,12 @@ impl RTC {
                 self.trigger_interrupt()?;
             }
             RTCICR => {
-                // As per above mentioned doc, the interrupt is cleared by writing any data value to
-                // the Interrupt Clear Register.
-                self.ris = 0;
-                self.trigger_interrupt()?;
+                // The RTCICR is used together with the RTCMR register to implement a basic time alarm function.
+                // As per the above comment, we do not implement this functionality in firecracker. It can
+                // be easily checked that the value of self.ris will never change. Just like we do
```
alindima, line 147: In this case, the value can actually change, if the driver does a write to RTCICR.

src/devices/src/legacy/rtc_pl031.rs
```diff
@@ -142,10 +142,12 @@ impl RTC {
...
+                // for the RTCMR, we set the expected value and we increment a missed metric.
+                self.ris &= !val;
```
alindima, line 149: what if the interrupt was already 0 and the driver wrote a 0, which should not do anything? this would actually set the interrupt status instead of clearing it

I think we should first do a check that the `val` is not 0, in which case just ignore it, as per the manual

src/devices/src/legacy/rtc_pl031.rs
dianpopa: > > > Not really tied to this PR, but the spec says that writing 1 to the RTCICR register clears the interrupt.
> > > Writing 0 should not.
> > > https://developer.arm.com/documentation/ddi0224/c/Programmers-model/General-registers/Interrupt-Clear-Register--RTCICR?lang=en
> > 
> > 
> > > Not really tied to this PR, but the spec says that writing 1 to the RTCICR register clears the interrupt.
> > > Writing 0 should not.
> > > https://developer.arm.com/documentation/ddi0224/c/Programmers-model/General-registers/Interrupt-Clear-Register--RTCICR?lang=en
> > 
> > 
> > There is a problem with the RTCICR, but it is not related to how we set it. The RTCICR is used in conjunction with the RTCMR to implement a time alarm function:
> > ```
> > The Match Register is programmed by writing to RTCMR. The counter and match values are compared in a comparator. When both values are equal, the RTCINTR interrupt is asserted HIGH. An external processor can use the interrupt to implement a basic time alarm function. The interrupt is cleared by writing any data value to the Interrupt Clear Register, RTCICR. The value in the Match Register can be read at any time.
> > ```
> > 
> > 
> > This is extra functionality that one can achieve with the RTC. As a consequence, our implementation does not need the RTCICR. It can easily be seen from the implementation that the value of self.ris never changes, so there is no need for the value to be cleared by the RTCICR as it will always be 0. I updated the code to be similar to how we proceed for the RTCMR register, take a look.
> 
> In the latest commit, I see that this was fixed.
> 
> Since we are not using the match functionality, can't we get rid of all interrupts?

Yup, I got rid of the interrupts. PTAL!

src/devices/src/legacy/rtc_pl031.rs
dianpopa: `self.ris` is never set to 1 by our code, so any value that would be written (0 or 1) it cannot change to anything other than 0, right?
In a scenario where we would make use of the match functionality, self.ris would be asserted when the match value would equal the one desired.

src/devices/src/legacy/rtc_pl031.rs
dianpopa: So:
self.ris = 0, RTCICR =0, => new self.ris = 0& 1 => 0, so self.ris keeps its value
self.ris = 0, RTCICR=1 => new self.ris = 0 & 0 => so self.ris remains deasserted. 
self.ris =1, RTCICR = 0, new self.ris = 1 & 1 => self.ris keeps its value
self.ris =1, RTCICR = 1, new self.ris = 1 & 0 => self.ris gets deasserted
The bitwise operation has exactly this scope, to keep the value of self.ris if RTCICR is 0 and change it to 0 if RTCICR is 1.

src/devices/src/legacy/rtc_pl031.rs
dianpopa: Regarding the offsets, u8 does not seem to be enough, the last address from the RTC spec is `0xFFC RTCPCellID3` which cannot be represented in an u8. Of course, we could use an u16, let me know if you have a strong preference for doing a cast to the smallest possible representation.

src/devices/src/legacy/rtc_pl031.rs
alindima: I was seeing AND but was reading OR, sorry 😄

src/devices/src/legacy/rtc_pl031.rs
alindima: I don't have a very strong preference. I think it probably only adds a tiny memory overhead

src/devices/src/legacy/rtc_pl031.rs
```diff
@@ -75,16 +71,14 @@ impl RTC {
             tick_offset: utils::time::get_time_ns(utils::time::ClockType::Real) as i64,
             match_value: 0,
             load: 0,
+            // The interrupt mask is initialised as not set.
             imsc: 0,
+            // The raw interrupt is initialised as not asserted.
             ris: 0,
             interrupt_evt,
```
alindima, line 78: does it even make sense to keep the interrupt eventfd in the code, since we are not using it anywhere?

dianpopa: No it does not but I was not sure whether to remove it or not. I guess I could remove it so that it does not become confusing, right?

alindima: sure, since it's not used at all, it's best to remove it. I see that you just did that 👍🏻

dianpopa:
@alindima @georgepisaltu PTAL and let me know your thoughts. I will update the coverage on aarch64 after your feedback.

alindima (dismissed):
🚀

georgepisaltu (dismissed):

src/vmm/src/device_manager/mmio.rs
```diff
@@ -285,19 +285,14 @@ impl MMIODeviceManager {
     /// otherwise allocate a new MMIO slot for it.
     pub fn register_mmio_rtc(
         &mut self,
-        vm: &VmFd,
         rtc: Arc<Mutex<devices::legacy::RTC>>,
         dev_info_opt: Option<MMIODeviceInfo>,
     ) -> Result<()> {
         // Create and attach a new RTC device.
+        // We allocate an IRQ even though we do not need it so that
+        // we do not break snapshot compatibility.
```
sandreim, line 293: I think we need to update changelog to mention the v0.24 RTC issue was fixed, how RTC works now and that we allocate a dummy IRQ starting with v0.25. Customers might wonder why there is a gap in allocated IRQs.

acatangiu: Do we need to keep allocating the dummy IRQ in v0.25 from-scratch-booted VMs?

Or are you only referring to microVMs loaded from snapshots generated with < v0.25 Firecracker versions?

acatangiu: I don't think we break compatibility if we remove the dummy IRQ.

dianpopa: Yes, I was referring to microVms loaded from snapshots generated with v0.24. I will test it out.

sandreim: v0.24 -> v0.25 should work fine, but the dummy IRQ is for v0.25 -> v0.24 which will actually register that irqfd.

dianpopa:
@alindima @georgepisaltu @sandreim i addressed all feedback. PTAL!

georgepisaltu (approved):

src/vmm/src/device_manager/mmio.rs
acatangiu: You're both right, resolving this.

alindima (approved):

Issue #2356, opened 2020-12-15
[Feature Request] Add capability to dump the Firecracker configuration


raduiliescu:
## Why is this feature request important?  What are the use cases?  Please describe.
When replicating an issue is important to start from the Firecracker configuration used by the customers. 
There is no way to dump the configuration of a running Firecracker microVM. Currently all API calls received by FC are logged, but this happens only on the info level which is not configured by default.

Either the customer has to reproduce the issue again with info log enabled, or have all commands recorded in some way.

## Describe the desired solution

Would be nice to add --dump-config API endpoint that would either dump a config file or a record of all the commands.

## Checks
- [x] Have you searched the Firecracker Issues database for similar requests?
- [x] Have you read all the existing relevant Firecracker documentation?
- [x] Have you read and understood Firecracker's core tenets?

Tobbeman:
I can take a look if maintainers want the feature in the code base.

An /export endpoint sounds like a good idea to me.

sandreim:
Thanks for the bias for action @Tobbeman. An `/export` endpoint also sounds good to me, but I would like to hear how the @firecracker-microvm/compute-capsule folks think about this as well.

georgepisaltu:
I think the endpoint solution proposed by @Tobbeman is a great way to do this, should also be straightforward to implement.

dianpopa:
It would be a great feature to have in Firecracker. An `/export` endpoint sounds good to me. @Tobbeman thanks for the interest in contributing!

sandreim:
@Tobbeman looks like `/export` is the way to go. We're happy to review the PR once it is ready 😄

Tobbeman:
Great, I'll look into it :+1:

acatangiu:
An `API` endpoint sounds good to me too. The URL name will depend on what information it will offer which in turn depends on the implementation.

There have been two options thrown around on this thread:
1. Exporting current configuration
2. Exporting a log of historic configuration commands

A quick brain-dump on approaching each of these two options:
1. Firecracker does not keep per-se configuration information other than internal intrinsic state. When a resource is configured, the _configuration description_ is used to build the resource, then the _description_ is discarded.
To provide _descriptions_ at runtime, new code spanning multiple components would be required to generate _configuration descriptions_ from internal state.
[...]

PR #2454, opened 2021-02-10
Feature/export config

Tobbeman:
# Reason for This PR
https://github.com/firecracker-microvm/firecracker/issues/2356

## Description of Changes
- Add vector of config changes to the api_server component
- Add endpoint exposing above changes

## License Acceptance

## PR Checklist

### DIFF
--- a/src/api_server/src/parsed_request.rs
+++ b/src/api_server/src/parsed_request.rs
@@ -7,6 +7,7 @@ use super::VmmData;
 use crate::request::actions::parse_put_actions;
 use crate::request::balloon::{parse_get_balloon, parse_patch_balloon, parse_put_balloon};
 use crate::request::boot_source::parse_put_boot_source;
+use crate::request::config_changes::parse_get_config_changes;
 use crate::request::drive::{parse_patch_drive, parse_put_drive};
 use crate::request::instance_info::parse_get_instance_info;
 use crate::request::logger::parse_put_logger;
@@ -28,6 +29,7 @@ use vmm::rpc_interface::{VmmAction, VmmActionError};
 pub(crate) enum ParsedRequest {
     GetInstanceInfo,
     GetMMDS,
+    GetConfigChanges,
     PatchMMDS(Value),
     PutMMDS(Value),
     Sync(Box<VmmAction>),
@@ -60,6 +62,7 @@ impl ParsedRequest {
             (Method::Get, "balloon", None) => parse_get_balloon(path_tokens.get(1)),
             (Method::Get, "machine-config", None) => parse_get_machine_config(),
             (Method::Get, "mmds", None) => parse_get_mmds(),
+            (Method::Get, "config-changes", None) => parse_get_config_changes(),
             (Method::Get, _, Some(_)) => method_to_error(Method::Get),
             (Method::Put, "actions", Some(body)) => parse_put_actions(body),
             (Method::Put, "balloon", Some(body)) => parse_put_balloon(body),
--- a/src/api_server/src/parsed_request.rs
+++ b/src/api_server/src/parsed_request.rs
@@ -153,7 +153,7 @@ fn log_received_api_request(api_description: String) {
 /// * `method` - one of `GET`, `PATCH`, `PUT`
 /// * `path` - path of the API request
 /// * `body` - body of the API request
-fn describe(method: Method, path: &str, body: Option<&Body>) -> String {
+pub fn describe(method: Method, path: &str, body: Option<&Body>) -> String {
     match (path, body) {
         ("/mmds", Some(_)) | (_, None) => format!("{:?} request on {:?}", method, path),
         (_, Some(value)) => format!(
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -14,7 +14,7 @@ use std::path::PathBuf;
 use std::sync::{mpsc, Arc, Mutex};
 use std::{fmt, io};
 
-use crate::parsed_request::ParsedRequest;
+use crate::parsed_request::{describe, ParsedRequest};
 use logger::{
     debug, error, info, update_metric_with_elapsed_time, IncMetric, StoreMetric, METRICS,
 };
@@ -81,6 +81,8 @@ pub struct ApiServer {
     /// If this flag is set, the process encountered a fatal error
     /// and it is going to exit once it sends any pending API response.
     vmm_fatal_error: bool,
+    /// Vector to hold vmm configuration change descriptions
+    config_changes: Vec<String>,
 }
 
 impl ApiServer {
@@ -101,6 +103,7 @@ impl ApiServer {
             vmm_response_receiver,
             to_vmm_fd,
             vmm_fatal_error: false,
+            config_changes: Vec::new(),
         }
     }
 
@@ -260,6 +263,15 @@ impl ApiServer {
         request: &Request,
         request_processing_start_us: u64,
     ) -> Response {
+        if request.method() == Method::Patch || request.method() == Method::Put {
+            let request_uri = request.uri().get_abs_path().to_string();
+            self.config_changes.push(describe(
+                request.method(),
+                request_uri.as_str(),
+                request.body.as_ref(),
+            ));
+        }
+
         match ParsedRequest::try_from_request(request) {
             Ok(ParsedRequest::Sync(vmm_action)) => {
                 self.serve_vmm_action_request(vmm_action, request_processing_start_us)
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -278,6 +278,7 @@ impl ApiServer {
             }
             Ok(ParsedRequest::GetInstanceInfo) => self.get_instance_info(),
             Ok(ParsedRequest::GetMMDS) => self.get_mmds(),
+            Ok(ParsedRequest::GetConfigChanges) => self.get_config_changes(),
             Ok(ParsedRequest::PatchMMDS(value)) => self.patch_mmds(value),
             Ok(ParsedRequest::PutMMDS(value)) => self.put_mmds(value),
             Err(e) => {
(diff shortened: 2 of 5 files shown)


src/api_server/src/parsed_request.rs
```diff
@@ -28,20 +29,29 @@ use vmm::rpc_interface::{VmmAction, VmmActionError};
...
             request_uri.as_str(),
             request.body.as_ref(),
         ));
 
+        if request.method() == Method::Patch || request.method() == Method::Put {
+            changes.push(describe(
+            request.method(),
+            request_uri.as_str(),
+            request.body.as_ref(),
+            ));
```
acatangiu, line 52: What do you think about compressing these? I think it would be interesting even if as an experiment?

I found https://docs.rs/smaz/0.1.0/smaz/ and https://lib.rs/crates/lz4_flex doing a quick search of simple pure Rust compression libs with no external dependencies.

sandreim: Are we talking about http compression or returing a compressed file ? If it is the latter I would expect the output in a format that one can decompress using common tools.

Tobbeman: Sure thing, I'll take a look!

sandreim: So, it is about reducing the memory footprint of the change log. Maybe a good idea is to make this as a feature opt-in via some cli param, but then we would loose on the debugability if this is not enabled. Even with compression, we are trading some CPU cycles for memory. Not sure how this plays at very high scale.

As an alternative, we could just log these to a separate file on disk.  @firecracker-microvm/compute-capsule WDYT?

acatangiu: > Even with compression, we are trading some CPU cycles for memory. Not sure how this plays at very high scale.

We only compress during configuration (this amount of cpu overhead should not matter) and decompress during API call to get config (again very seldom, only on some machines, etc).
Memory saved is minimal since the amount of saved bytes is pretty small, but I'd be curious to see numbers and think about it at scale.

> As an alternative, we could just log these to a separate file on disk and possibly skip logging API calls in the regular firecracker log. @firecracker-microvm/compute-capsule WDYT?

This is nice and clean, but what file? Would we add an API configuration option for a new `/your/path/to/config.log`?

> Also, one other idea is to depart from the log approach and store this in the same format expected by FC when using json config file startup.

Saving in memory has same problems, generating on demand would be somewhat difficult since it would need to touch all internal components (each component would have a `dump_config()` type method or trait impl). I'm not saying it's impossible, but I'm also not sure if complexity is worth it.

acatangiu: >>    As an alternative, we could just log these to a separate file on disk and possibly skip logging API calls in the regular firecracker log. @firecracker-microvm/compute-capsule WDYT?

> This is nice and clean, but what file? Would we add an API configuration option for a new /your/path/to/config.log?

Now that I think about it more, how is this much different than the normal log file?
And why isn't that enough? What is the real use case of this feature where log file doesn't cut it?

acatangiu: > > > As an alternative, we could just log these to a separate file on disk and possibly skip logging API calls in the regular firecracker log. @firecracker-microvm/compute-capsule WDYT?
> 
> > This is nice and clean, but what file? Would we add an API configuration option for a new /your/path/to/config.log?
> 
> Now that I think about it more, how is this much different than the normal log file?
> And why isn't that enough? What is the real use case of this feature where log file doesn't cut it?

Looking at the issue again I answered my own question: _config is only logged on `info` level_.

georgepisaltu: My gut tells me compression is not the way to go here. All requests should be a few KiB in text form when stored in memory, so storing them shouldn't impact our RSS significantly. But it is an interesting proposal, so I'm gonna roll with it.

An API call is processed in about 1 ms, from the time we get notified in the API server to the time we write the last byte of the response. Also, the average request is about 300 bytes of data and AFAIK Firecracker requests (aside from MMDS requests, which is data we shouldn't record) never exceed 1 KiB of data to be processed.

With this in mind, we should consider the time it takes to compress 300 bytes of data and check how much space is actually saved. If the compressing time is not significant and the amount of memory saved is, we should then measure the latency introduced by writing this to disk instead of keeping it in memory. With this information, I think we can make a data-driven decision.

Tobbeman: I made an attempt at gathering some data
Using a boot source configuration log entry as a sample I tested smaz and lz4_flex listed above.
Only tested the safe version of lz4_flex

The original length of the string as byte vector is 201.
smaz got it down to 167
lz4_flex got it down to 192

Running simple/naive bench I can see this:
```
running 3 tests
test tests::bench_lz4_flex_compress ... bench:         559 ns/iter (+/- 526)
test tests::bench_no_compress       ... bench:          65 ns/iter (+/- 2)
test tests::bench_smaz_compress     ... bench:      16,045 ns/iter (+/- 792)
```
I did not test decompression since I did not think the above numbers justified it.

In my mind, I don't think the saved bytes are worth the extra cycles, but I lack experience to say for sure.
The tests are in this [repository](https://github.com/Tobbeman/compression_benchmark). Please feel free to tinker and test.
If you feel like I should set something specific up I'd be happy to look at it too :)

Tobbeman: > So, it is about reducing the memory footprint of the change log. Maybe a good idea is to make this as a feature opt-in via some cli param, but then we would loose on the debugability if this is not enabled. Even with compression, we are trading some CPU cycles for memory. Not sure how this plays at very high scale.

Maybe the feature could be opt-out from the cli instead?
That would allow power users to shave of some request time at the cost of less debuggability.

Tobbeman: Fyi, moved some code around to make it more sensible.
(CONTRIBUTING.md does not mention not force pushing to forks, but perhaps it is better to create a new branch instead?)

sandreim: > Fyi, moved some code around to make it more sensible.
> (CONTRIBUTING.md does not mention not force pushing to forks, but perhaps it is better to create a new branch instead?)

Force pushing is fine. No need to create a new PR.

alindima: > > > > As an alternative, we could just log these to a separate file on disk and possibly skip logging API calls in the regular firecracker log. @firecracker-microvm/compute-capsule WDYT?
> > 
> > 
> > > This is nice and clean, but what file? Would we add an API configuration option for a new /your/path/to/config.log?
> > 
> > 
> > Now that I think about it more, how is this much different than the normal log file?
> > And why isn't that enough? What is the real use case of this feature where log file doesn't cut it?
> 
> Looking at the issue again I answered my own question: _config is only logged on `info` level_.

can't we log the config on all levels to solve the problem?

sandreim: @georgepisaltu Your comment makes a lot of sense and I agree we should measure CPU and memory usage before committing to any compression usage.

Recently, I've given this more thought and it makes more sense to have this feature enabled by default, otherwise we can just use the log level switch to expose the API calls in the log. Also, since we already have the capability to record the configuration at the`Info` log level it becomes redundant to have another way of doing it. 

With that said, I suggest to implement a mechanism that just dumps the JSON config that can be used as input to start Firecracker without any modification. This will result in a more complicated implementation, but I expect we can isolate all the code by defining a DumpConfig trait and implement it for all devices and VMM.

@acatangiu @georgepisaltu @alindima what is your thinking about this approach ?

georgepisaltu: I don't think we need a trait for this. It would be much simpler to just store a `Vec<String>`. It should be just a few KiB even when uncompressed.

acatangiu: Like I said from the beginning, having a trait implemented by every configurable resource adds quite a bit of code and is more complex than the scrappy approach in the current PR.

I am not against it if @Tobbeman wants to take a stab at it. Otherwise, I am also fine with current PR (compressed or otherwise).

If we do go the generating way: define `DumpConfig` trait in `vmm/src/vmm_config` and implement it there for all device objects as they are configured. This would keep config code/format isolated.

acatangiu: > can't we log the config on all levels to solve the problem?

Thinking more about it, I think this is actually a better solution than the current PR. What would be the downside here? Few KiB larger logfiles. Still better than same overhead in resident memory.
The purpose of the issue is to improve debug-ability and logs are critical for that, having config cmds in logs regardless of configured logging level seems to hit all points.
Logs also offer persistence; in case of Firecracker crash or any type of post-mortem investigation, logged info is still available while API endpoint is gone.

The other option to dynamically generate (trait `DumpConfig`) has the only cost of added code, but no other overhead and the generated config would be much prettier/easier to read than rebuilding config from log-entries.
With the disadvantage of only being available while firecracker is running.

@firecracker-microvm/compute-capsule WDYT?

alindima: I am +1 for logging this info all the time

sandreim: > I am +1 for logging this info all the time. The small downside is that one would have to dig through the logs instead of having a clear API endpoint, but I think this is reasonable

Before we commit to this approach we need to make sure it makes sense for:
- starting FC from config file, should the log also contain the JSON ?
- starting from snapshots + extra PATCHing of config ?

We also need to dive a little bit deeper into some numbers in the context of high density Firecracker deployments when the configuration has multiple network interfaces and disks. What is the actual I/O and storage footprint overhead ? Is it impactful for host level log processing in terms of CPU and maybe network I/O ? (if logs are being sent over the network). We should update the docs with this information so customers can account for this extra overhead when they deploy at high scale.

acatangiu: That's a very good call-out @sandreim! I guess the comprehensive solution to satisfy all usage scenarios would be a mix of the above:
- dynamically/on-demand describe resources (generate desc using trait)
- expose resource desc API endpoint
- log resource description on firecracker exit (clean or otherwise)

serban300:
Hi ! This issue is more complex then we initially thought and we would like to spend some more time discussing it within the team. We'll do that by next week.

Tobbeman:
> Hi ! This issue is more complex then we initially thought and we would like to spend some more time discussing it within the team. We'll do that by next week.

I was getting the same feeling, thanks for the update
Let me know when/if I can help :)

georgepisaltu:
Hi @Tobbeman ,

Sorry for the delayed response. In the past months we've been busy with other work and we haven't had the chance to create a design for this feature yet. We plan to pick this up in our next sprint cycle, which starts on April 19th.

Tobbeman:
Thanks for the update @georgepisaltu 

No need to rush for my sake, I'm looking forward to see what you cook up :) 
Let me know if you want someone more senior to implement it.

acatangiu:
@Tobbeman thanks for your patience. I have done a PoC of what we discussed to be the right path here. The PoC can be found here: https://github.com/acatangiu/firecracker/tree/poc-export-cfg

If you're interested, feel free to pick up that work and drive it to completion. If you do pick it up, feel free to rework any of the existing commits 😄 

Let us know your thoughts.

alindima:
Hi @Tobbeman, are you still interested in working on this?

acatangiu:
Closing in favor of #2642

Issue #2452, opened 2021-02-10
[Bug] logging api used inside the signal handler


KarthikNedunchezhiyan:
# Describe the bug

As logging API will use lock internally, I suspect using logging API inside the signal handler is not async safe. Let say logging API `error!` implementation is like below

```
error! => {
   get-lock
   write-to-sink
   release-lock
}
```

When the firecracker thread1's execution is at `get-lock` phase and suddenly signal handler is triggered, Signal handler will also try to get lock on `get-lock` phase (because the signal handler is using `error!`). Now both of them will be in a dead-lock. thread1 will wait for the signal handler to complete and the signal handler will wait for thread1 to release the lock.

## To Reproduce

It is hard to reproduce.

## Additional context

I can notice rust-sys-util blocked the other signals while the current signal handler is in use, but IMO this will not solve this problem.

I am newbie to this topic. feel free to close this if it is not valid 🙂.

## Checks

- [ ] Have you searched the Firecracker Issues database for similar problems?
- [ ] Have you read the existing relevant Firecracker documentation?
- [ ] Are you certain the bug being reported is a Firecracker issue?

alindima:
Hi. Thank you for logging the issue.
We've looked over the potential bug and it seems that it is indeed a very hard to reproduce bug.

This is also the case for metrics, not just for logging, and it seems to go beyond mutex locks (https://man7.org/linux/man-pages/man7/signal-safety.7.html)

The fix would be to use only async-signal-safe functions in the signal handler.

alindima:
Off the top of my head, a possible fix would be to remove the `exit` call from the signal handler, as well as the logging and metrics calls, and replace them with a notification mechanism (an eventfd maybe) that is checked regularly by the emulation thread.

KarthikNedunchezhiyan:
Using eventfd to offload the non-async signal task from the signal handler is a common way but one challenge in offloading the task to another handler is to pass sufficient information(like si_signo, si_code) to another handler.

alindima:
Fixed in #2459

alindima:
We've synced internally, and discussed the potential impact of this bug, as well as the alternatives for mitigation.

We converged on the fact that the bug is very hard to reproduce (and has no impact that we know of), and that the worst case scenario is a Firecracker process deadlock, with no security implications.
We think that the possible solutions we explored are too complicated for the moment.

[...]

PR #2459, opened 2021-02-16
Replace unsafe signal handlers with signalfd mechanism.

alindima:
# Reason for This PR

https://github.com/firecracker-microvm/firecracker/issues/2452 

Resources:
https://man7.org/linux/man-pages/man2/signalfd.2.html
https://man7.org/linux/man-pages/man7/signal.7.html

## Description of Changes

View commit messages

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/signal_handler.rs
+++ b/src/vmm/src/signal_handler.rs
@@ -81,38 +81,31 @@ pub struct SignalManager {
     signal_fd: File,
 }
 
-macro_rules! generate_handler {
-    ($fn_name:ident ,$signal_name:ident, $exit_code:ident, $signal_metric:expr, $body:ident) => {
-        fn $fn_name(info: signalfd_siginfo) {
-            let si_signo = info.ssi_signo as i32;
-            let si_code = info.ssi_code;
-
-            $signal_metric.inc();
-
-            $body(info);
-
-            error!(
-                "Shutting down VM after intercepting signal {}, code {}.",
-                si_signo, si_code
-            );
-            // Write the metrics before exiting.
-            if let Err(e) = METRICS.write() {
-                error!("Failed to write metrics while stopping: {}", e);
-            }
-
-            // Safe because we're terminating the process anyway. We don't actually do anything when
-            // running unit tests.
-            #[cfg(not(test))]
-            unsafe {
-                _exit(i32::from(match si_signo {
-                    $signal_name => super::$exit_code,
-                    _ => super::FC_EXIT_CODE_UNEXPECTED_ERROR,
-                }))
-            };
-        }
-    };
+#[inline]
+fn exit_unexpected() {
+    // Safe because we're terminating the process anyway.
+    unsafe { _exit(i32::from(super::FC_EXIT_CODE_UNEXPECTED_ERROR)) };
 }
 
+// Given a signal number, return the respective metric and exit code.
+fn get_metric_and_exitcode(signo: c_int) -> Option<(&'static dyn IncMetric, i32)> {
+    match signo {
+        SIGXFSZ => Some((&METRICS.signals.sigxfsz, super::FC_EXIT_CODE_SIGXFSZ as i32)),
+        SIGXCPU => Some((&METRICS.signals.sigxcpu, super::FC_EXIT_CODE_SIGXCPU as i32)),
+        SIGBUS => Some((&METRICS.signals.sigbus, super::FC_EXIT_CODE_SIGBUS as i32)),
+        SIGSEGV => Some((&METRICS.signals.sigsegv, super::FC_EXIT_CODE_SIGSEGV as i32)),
+        SIGPIPE => Some((&METRICS.signals.sigpipe, super::FC_EXIT_CODE_SIGPIPE as i32)),
+        SIGSYS => Some((
+            &METRICS.seccomp.num_faults,
+            super::FC_EXIT_CODE_BAD_SYSCALL as i32,
+        )),
+        SIGHUP => Some((&METRICS.signals.sighup, super::FC_EXIT_CODE_SIGHUP as i32)),
+        SIGILL => Some((&METRICS.signals.sigill, super::FC_EXIT_CODE_SIGILL as i32)),
+        _ => None,
+    }
+}
+
+// Special handling of logging for the SIGSYS signal.
 fn log_sigsys_err(info: signalfd_siginfo) {
     if info.ssi_code != SYS_SECCOMP_CODE {
         // We received a SIGSYS for a reason other than `bad syscall`.
@@ -120,8 +113,7 @@ fn log_sigsys_err(info: signalfd_siginfo) {
             "Shutting down VM after intercepting signal {}, code {}.",
             info.ssi_signo, info.ssi_code
         );
-        // Safe because we're terminating the process anyway.
-        unsafe { _exit(i32::from(super::FC_EXIT_CODE_UNEXPECTED_ERROR)) };
+        exit_unexpected();
     }
 
     let syscall = info.ssi_syscall;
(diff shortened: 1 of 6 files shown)


acatangiu:
This being an RFC I guess I should have added that I agree with the approach and the implementation and I suggest we promote this to a full PR 👍

src/vmm/src/signal_handler.rs
```diff
@@ -128,162 +156,95 @@ generate_handler!(
...
+    match siginfo.ssi_signo as i32 {
+        SIGXFSZ => sigxfsz_handler(siginfo),
+        SIGXCPU => sigxcpu_handler(siginfo),
+        SIGBUS => sigbus_handler(siginfo),
+        SIGSEGV => sigsegv_handler(siginfo),
+        SIGPIPE => sigpipe_handler(siginfo),
+        SIGSYS => sigsys_handler(siginfo),
+        SIGHUP => sighup_handler(siginfo),
+        SIGILL => sigill_handler(siginfo),
+        // This should never be reached since we'll only receive the signals in the signalfd mask.
```
sandreim, line 198: We should add an error! message here, to document the unexpected FC exit in the log.

src/vmm/src/signal_handler.rs
```diff
@@ -128,162 +156,95 @@ generate_handler!(
...
-        // way of either preventing this behaviour, or detecting for certain whether this test is
-        // run by kcov or not. The best we could do so far is to look at the perceived number of
-        // available CPUs. Kcov seems to make a single CPU available to the process running the
-        // tests, so we use this as an heuristic to decide if we check the assertion.
-        if cpu_count() > 1 {
-            // The signal handler should let the program continue during unit tests.
-            assert!(METRICS.seccomp.num_faults.count() >= 1);
+        // Even though we currently kill the process during signal handling, we should in theory
+        // read from the signalfd all the signals for future-proofing.
+        while self.signal_fd.read(&mut read_buf[..]).unwrap() == siginfo_size {
```
sandreim, line 238: Can the read fail ?

src/vmm/src/signal_handler.rs
```diff
@@ -128,162 +156,95 @@ generate_handler!(
...
+            let siginfo: signalfd_siginfo = unsafe { ptr::read(read_buf.as_ptr() as *const _) };
```
sandreim, line 239: Is this safe ?

tests/integration_tests/functional/test_signals.py
```diff
@@ -139,7 +142,7 @@ def test_handled_signals(test_microvm_with_ssh, network_config):
     assert int(stdout.read()) == 2
 
     # We have a handler installed for this signal.
-    os.kill(firecracker_pid, SIGRTMIN+1)
```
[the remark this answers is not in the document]
alindima: Hardcoded the 35 since the `SIGRTMIN` python reports is not correct for musl. It's likely the one for glibc, and we shouldn't rely on it.

alindima:
Implemented unit tests. PR is no longer RFC

src/vmm/src/vstate/vcpu/mod.rs
```diff
@@ -241,6 +242,8 @@ impl Vcpu {
         let vcpu_thread = thread::Builder::new()
             .name(format!("fc_vcpu {}", self.kvm_vcpu.index))
             .spawn(move || {
+                mask_handled_signals().expect("Unable to install signal mask on VCPU thread.");
```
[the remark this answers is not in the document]
alindima: Very nice catch. VCPU threads inherit the mask from the VMM thread, since they are created after the signalfd is registered.

The API thread is the one who inherits empty signal masks, since at the time it is launched, the `signalfd` was not yet created. And we don't want to mask the signals unless we have the signalfd mechanism in place (to not lose any signals data).

src/vmm/src/signal_handler.rs
```diff
@@ -1,38 +1,66 @@
...
+    Ok(())
+}
+
+/// Struct responsible for processing signalfd events.
+pub struct SignalManager {
+    /// The file associated to the current signalfd.
+    signal_fd: File,
+}
+
 macro_rules! generate_handler {
```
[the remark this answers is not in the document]
alindima: This is true. I will try to find a prettier way of defining these functions, since there is no longer a need for a macro.

src/vmm/src/signal_handler.rs
```diff
@@ -128,162 +156,95 @@ generate_handler!(
...
+    Ok(mask)
+}
 
-            // Call SIGSEGV signal handler.
-            assert_eq!(METRICS.signals.sigsegv.count(), 0);
-            unsafe {
-                syscall(libc::SYS_kill, process::id(), SIGSEGV);
-            }
+/// Handle the signal according to its signal number.
+fn handle_signal(siginfo: signalfd_siginfo) {
```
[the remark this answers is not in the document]
alindima: I was on the fence about this, since it does not need anything from the SignalManager state.
I agree however that it is cleaner to have it in the same scope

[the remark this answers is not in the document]
alindima: done

src/vmm/src/vstate/vcpu/mod.rs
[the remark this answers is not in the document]
alindima: Removed this

alindima:
Addressed all comments except for one which still needs work: replacing the `generate_handler` macro with something prettier

acatangiu (approved):

src/vmm/src/signal_handler.rs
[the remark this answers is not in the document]
alindima: Done. PTAL @serban300

src/vmm/src/signal_handler.rs
alindima: In the latest version, I replaced this unsafe call with a call to the `ByteValued::from_slice()` which returns an option, so we are now good in terms of error handling. If the read is not possible, we should get a `None` and we log the error.

alindima:
Replaced the unsafe call to `ptr::read` with a call to the `ByteValued` trait.
Also removed the signal handler generator macros and moved the handler logic in the `SignalManager::handle_signal()` method

src/vmm/src/signal_handler.rs
```diff
@@ -229,23 +136,60 @@ impl SignalManager {
...
+        let si_signo = info.ssi_signo as i32;
+        let si_code = info.ssi_code;
+
+        // For SIGSYS, we have some special logging.
+        if si_signo == SIGSYS {
+            log_sigsys_err(info);
+        }
+
+        // Increment the right metrics.
+        match si_signo {
```
serban300, line 149: I would rather have a function that does this match and returns a tuple with the metric and exit code associated with each signal.

src/vmm/src/signal_handler.rs
```diff
@@ -1,168 +1,254 @@
...
+
+        let siginfo_size = mem::size_of::<signalfd_siginfo>();
+        let mut read_buf: Vec<u8> = vec![0; siginfo_size];
+
+        // Even though we currently kill the process during signal handling, we should in theory
+        // read from the signalfd all the signals for future-proofing.
+        loop {
+            match self.signal_fd.read(&mut read_buf[..]) {
+                Ok(size) if size == siginfo_size => {
+                    match SignalInfoWrapper::from_slice(&read_buf) {
```
serban300, line 214: I don't think this is worth the added complexity introduced by `SignalInfoWrapper`. Personally I would just read. It should be safe.

alindima: As we discussed offline, there is not much added complexity and the `ByteValued` approach has some clear advantages:

 - does not copy the buffer, it uses the same memory as the byte slice, as opposed to the `ptr::read`
 - has some extra safety checks: checks that the data is correctly aligned and checks the length of the byte slice.

src/vmm/src/signal_handler.rs
alindima: done

alindima:
Addressed all comments. PTAL @acatangiu @sandreim

sandreim (dismissed):

serban300 (dismissed):

serban300 (approved):

sandreim (dismissed):

acatangiu (approved):

sandreim (approved):

georgepisaltu (approved):
🚀

PR #2474, opened 2021-02-26
[feature/file based seccomp] add devtool support, build script improvement & rename package

alindima:
# Reason for This PR

- Support for seccompiler in the various devtool commands was not yet implemented, like `build`, `strip`, `prepare_release`
- Seccompiler is run during the build process, being triggered by the `vmm/src/build.rs`. However, the build script was not triggered by a modification in the seccompiler source code. It was only triggered by a modification in the target-specific filter file.
- seccompiler library was named `seccomp`, which was the old crate, with another interface.

## Description of Changes

View commits

## License Acceptance

## PR Checklist


tools/devtool
```diff
@@ -675,6 +680,20 @@ cmd_build() {
...
+    # Build seccompiler
+    run_devctr \
+        --user "$(id -u):$(id -g)" \
+        --workdir "$CTR_FC_ROOT_DIR" \
+        ${extra_args} \
+        -- \
+        cargo build -p seccomp --bin seccompiler \
+            --target-dir "$CTR_CARGO_SECCOMPILER_TARGET_DIR" \
+            "${cargo_args[@]}"
+    ret=$?
```
acatangiu: would it make sense to only build this as a separate target? smth like `devtool build seccompiler`.

Otherwise I am afraid we'll keep building it unnecessarily.

acatangiu (commented):
Wording error/typo in commit msg 2:
> However, **was build script is not** triggered by a modification in the seccompiler source code.

I'm not sure commit 2 is needed at all, I would build seccompiler separately from firecracker since it is a helper tool. WDYT?

tools/devtool
alindima: Firecracker needs to build seccompiler anyway, it's run as part of its build script.
So only one of these build procedures will actually do the build. The next one will reuse the same build folder so will take 0.02s.

You think this is ok?

acatangiu: Hmm, I must have misunderstood then, I was under the impression that the seccompiler is a separate stand-alone binary.

alindima:
> I'm not sure commit 2 is needed at all, I would build seccompiler separately from firecracker since it is a helper tool. WDYT?

Let's say we make a change in seccompiler and we want to build release binaries. Firecracker should also use the new seccompiler binary. But it will not, by default. If people are aware of this, this is fine. This can be triggered artificially, by doing some hacks like `touch`-ing the corresponding target JSON filter (this would retrigger the build script).

I'm afraid this is counterintuitive and hidden behaviour. There is a risk that one may do a Firecracker release without compiling with the latest seccompiler binary. This can be very inconvenient if we want to fix a bug in seccompiler for example and make a new release.

It's even more inconvenient when testing/developing.

The long term plan would be to decouple the seccompiler crate from the Firecracker workspace. But since they are being developed and released at the same time, we have to somehow introduce this dependency, that the Firecracker binary would use the latest seccompiler binary during the build.

acatangiu:
Ok, that makes sense, thanks for clearing it up!

acatangiu (dismissed):

tools/devtool
gc-plp, line 693: If return value is relevant, we should check it.

tests/framework/utils.py
```diff
@@ -601,3 +602,28 @@ def writeln(self, data: str):
...
+    """
+    Run seccompiler.
+
+    :param bpf_path: path to the seccompiler output file
+    :param json_path: optional path to json file
+    """
+    cargo_target = '{}-unknown-linux-musl'.format(platform.machine())
+
+    # If no custom json filter, use the default one for the current target.
+    if json_path is None:
```
gc-plp, line 617: nit: implicit value for `json_path` could be `defs.SECCOMP_JSON_DIR`.

tests/framework/utils.py
```diff
@@ -601,3 +602,28 @@ def writeln(self, data: str):
         self._file.write(data)
         self._file.write("\n")
         self._file.flush()
+
+
+def run_seccompiler(bpf_path, json_path=None):
```
gc-plp, line 607: This function seems kind of misplaced to be located in the utils file.

Given that it's a requirement for tests, it could be a pytest fixture.

tools/devtool
```diff
@@ -706,6 +725,11 @@ cmd_build() {
     # accessible location.
     [ $ret -eq 0 ] && {
         cargo_bin_dir="$CARGO_TARGET_DIR/$target/$profile"
+
+        # Seccompiler has a different build folder, so copy its contents here.
+        seccompiler_bin_dir="$CARGO_SECCOMPILER_TARGET_DIR/$target/$profile"
```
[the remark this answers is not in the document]
dianpopa: I think we really need to think of another solution,  I will try to figure out something and come back.

tests/framework/utils.py
dianpopa: I guess we could move it to host_tools/cargo_build.py, I would avoid as much as possible adding fixtures.

gc-plp: Since it's a dependency for this test only, I think it makes sense to use it as a fixture. We don't have to call it explicitly or even import it.

tools/devtool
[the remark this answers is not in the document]
alindima: Nice call-outs. All these problems you encountered above are rooted in the fact that we copy the seccompiler build directory to Firecracker's default target directory. I thought that this would be more intuitive, since all binaries would be placed in the same location.

However, I think it causes more side-effects than benefits.

Removing the copy command should fix those problems, so I'll go ahead and do it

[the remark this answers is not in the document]
alindima: > Shouldn t the secompiler target folder get deleted after this?

not really, it is a cache mechanism to make sure that we don't rebuild it unnecessarily

tests/framework/utils.py
alindima: > Since it's a dependency for this test only, I think it makes sense to use it as a fixture. We don't have to call it explicitly or even import it.

I see your point @gc-plp. However, it might be personal bias, but I really don't like fixtures 😄 . They make for hard to understand, magical code.
This fixture would also need to be parametrised with the bpf and json file paths, and it would also return `void` (aka nothing). I find it confusing to have a fixture that does not return anything.

I think a plain helper function would work just fine here. I like @dianpopa's suggestion to move the function in cargo_build.py. Is this good enough?

alindima: moved it to `cargo_build.py`. WDYT?

alindima:
@gc-plp @dianpopa I addressed your comment, PTAL

tests/framework/utils.py
gc-plp: >I see your point @gc-plp. However, it might be personal bias, but I really don't like fixtures smile . They make for hard to understand, magical code.

Fixtures aren't great but they avoid the import mechanism built into python, thus treating tests as black boxes.

> This fixture would also need to be parametrised with the bpf and json file paths, and it would also return void (aka nothing). I find it confusing to have a fixture that does not return anything.

The fixture can return a function.

> I think a plain helper function would work just fine here. I like @dianpopa's suggestion to move the function in cargo_build.py. Is this good enough?

Sure, that works too.

gc-plp (approved):

dianpopa (approved):

Issue #2473, opened 2021-02-25
"for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done" is not working in container of Docker


yilmazsen:
Hello Dear Community of Firecracker,

Nowadays, I am working on firecracker microvm's. However, i saw that issue:
When I try to build new image from docker image this command **"for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done"** does not work. So I attempted to change this command like that **"for d in bin etc lib root sbin usr; do tar c "/$d" || tar x -C /my-rootfs; done"** it worked. After that when I run this image with firecracker, this image had not work.
Is there any problem with that command ? or Should I change docker version or firecracker version something like that ? 

if anyone could help me, I will be happy 
Best regards

twitchyliquid64:
What kernel, kernel command line etc did you use?

yilmazsen:
Thank you for your reply, I was using amazon' hello-vmlinux.bin (https://s3.amazonaws.com/spec.ccfc.min/img/hello/kernel/hello-vmlinux.bin).  By the way I was using Dockerfile with alpine image.  I solved my problem. I will type here what is my solution. @twitchyliquid64

twitchyliquid64:
Do you get any console messages?

And is your rootfs in the correct format? If you are `tar`'ing the files, I'm not sure if that works (the [example here](https://github.com/firecracker-microvm/firecracker/blob/master/docs/getting-started.md#running-firecracker) uses an ext4 filesystem)

gc-plp:
Hi @yilmazsen. I assume that you're referring to [this](https://github.com/firecracker-microvm/firecracker/blob/master/docs/rootfs-and-kernel-setup.md#creating-a-rootfs-image) example from the Firecracker repository, right?

Could you give us more details on what did not work when trying out the commands in the guide? What were the error messages?

yilmazsen:
I am explaining that problem:

when we create a rootfs (https://github.com/firecracker-microvm/firecracker/blob/master/docs/rootfs-and-kernel-setup.md#creating-a-rootfs-image) image for our app. Such as nginx, I saw that problem. I followed these instructions:

```bash
docker run -it --rm -v /tmp/my-rootfs:/my-rootfs alpine
apk add openrc
apk add util-linux
apk add nginx 

# Set up a login terminal on the serial console (ttyS0):
[...]

PR #2484, opened 2021-03-05
Docs: Suggestion against warning message

yilmazsen:
# Reason for This PR
I have added little detail about this command 'for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done' .  Some of the people might be confused about it. For example me :) I was confused about that command result and I created this issue (https://github.com/firecracker-microvm/firecracker/issues/2473). Thank you so much @gc-plp @twitchyliquid64

## Description of Changes

 1. An Added explanation and a warning message that is the result of 'for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done' command.

## License Acceptance

### DIFF
--- a/docs/rootfs-and-kernel-setup.md
+++ b/docs/rootfs-and-kernel-setup.md
@@ -119,9 +119,15 @@ Alpine Linux:
 
    # Then, copy the newly configured system to the rootfs image:
    for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done
+
+   # The above command may trigger the following message:
+   # tar: Removing leading "/" from member names
+   # However, this is just a warning, so you should be able to
+   # proceed with the setup process.
+
    for dir in dev proc run sys var; do mkdir /my-rootfs/${dir}; done
 
-   # All done, exit docker shell
+   # All done, exit docker shell.
    exit
    ```
 


docs/rootfs-and-kernel-setup.md
```diff
@@ -119,6 +119,11 @@ Alpine Linux:
 
    # Then, copy the newly configured system to the rootfs image:
    for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done
+   
+   # After this command, you can see these result
+   <img width="671" alt="Screen Shot 2021-03-02 at 21 19 49" src="https://user-images.githubusercontent.com/3442583/109695745-5456f900-7b9d-11eb-95b1-3720e1559292.png">
```
dianpopa, line 124: I suggest to use a simple plain text message instead of a screenshot, something of the kind:
This command may trigger the following message: `tar: removing leading '/' from member names`. However, this is just a warning and it should be ok to continue with the setup flow.

dianpopa (commented):
There are some things that need to be done in order for the style checks to pass:
1. Sign your commit with `git commit --amend -s`
2. The commit should follow the [50/72 rule](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule).
3. Remove the trailing spaces from the md document.
4. change the commit title to something more formal:
"docs: advise against warning message"

docs/rootfs-and-kernel-setup.md
yilmazsen: I changed my commit according to your suggestions

docs/rootfs-and-kernel-setup.md
```diff
@@ -119,8 +119,14 @@ Alpine Linux:
 
    # Then, copy the newly configured system to the rootfs image:
    for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done
+   
+    # This command may trigger the following messages
+   ```bash
```
alindima, line 124: this is already inside a code block.
Please reformat the message and make sure that the markdown rendering is easy to read

alindima (commented):
Hi! Please squash all your commits into 1 and make sure that the 1 commit is signed using `git commit -s`. Also, make sure that you also follow @dianpopa's suggestions and fix any errors from the CI runs

docs/rootfs-and-kernel-setup.md
```diff
@@ -119,9 +119,16 @@ Alpine Linux:
 
    # Then, copy the newly configured system to the rootfs image:
    for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done
+
+   # This command may trigger the following message:
+   tar: Removing leading "/" from member names
```
alindima, line 124: Let's also add this as a comment, since people might copy-paste it and it would be treated as a bash command

docs/rootfs-and-kernel-setup.md
```diff
@@ -119,9 +119,16 @@ Alpine Linux:
 
    # Then, copy the newly configured system to the rootfs image:
    for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done
+
+   # This command may trigger the following message:
```
alindima, line 123:
```suggestion
# The above command may trigger the following message:
```

yilmazsen: Thanks for your help @alindima

docs/rootfs-and-kernel-setup.md
yilmazsen: Thank you for your suggestion but we have already said that  "the following message" is a warning message. and when they use the 'for d in bin etc lib root sbin usr; do tar c "/$d" | tar x -C /my-rootfs; done' command , they will be see ' tar: Removing leading "/" from member names' which is warning message at the terminal. I just want to add this idea :) "Hey! this message is a warning not a error and skip it and continue"

alindima: I meant to simply add a `#` character to the line. Thanks Diana for helping here.

luminitavoicu (approved):

PR #2488, opened 2021-03-08
Add Jailer parameter to allow spawning the firecracker process into a new PID namespace

luminitavoicu:
# Reason for This PR

The [Jailer documentation](https://github.com/firecracker-microvm/firecracker/blob/master/docs/jailer.md) recommends that the user spawns the jailer process is into a new PID namespace, for extra resilience. This PR implements a mechanism which takes the burden off the users and allows the Jailer to spawn the Firecracker process into a new PID namespace.

## Description of Changes

Added a bool optional parameter to the Jailer: `--new-pid-ns`, which enables the jailer to exec into the provided binary in a new PID namespace.

## License Acceptance

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - Added devtool build `--ssh-keys` flag to support fetching from private
   git repositories.
 - Added option to configure block device flush.
+- Added `--new-pid-ns` flag to the Jailer in order to spawn the Firecracker
+  process in a new PID namespace.
 
 ### Fixed
 
--- a/docs/jailer.md
+++ b/docs/jailer.md
@@ -19,6 +19,7 @@ jailer --id <id> \
        [--chroot-base-dir <chroot_base>]
        [--netns <netns>]
        [--daemonize]
+       [--new-pid-ns]
        [--...extra arguments for Firecracker]
 ```
 
@@ -45,6 +46,10 @@ jailer --id <id> \
   jailer will use this to join the associated network namespace.
 - When present, the `--daemonize` flag causes the jailer to cal `setsid()` and
   redirect all three standard I/O file descriptors to `/dev/null`.
+- When present, the `--new-pid-ns` flag causes the jailer to `fork()` and then
+  exec the provided binary into a new PID namespace. As a result, the jailer and
+  the process running the exec file have different PIDs. The PID of the child
+  process is stored in the jail root directory inside `<exec_file_name>.pid`.
 - The jailer adheres to the "end of command options" convention, meaning
   all parameters specified after `--` are forwarded to Firecracker. For
   example, this can be paired with the `--config-file` Firecracker argument to
@@ -98,6 +103,13 @@ After starting, the Jailer goes through the following operations:
   namespace.
 - If `--daemonize` is specified, call `setsid()` and redirect `STDIN`,
   `STDOUT`, and `STDERR` to `/dev/null`.
+- If `--new-pid-ns` is specified, call `unshare()` into a new PID namespace.
+  This will not have any effect on the current process, but its first
+  child will assume the role of init(1) in the new namespace. Next, the
+  jailer is duplicated by a `fork()` call, so that the child process
+  belongs to the previously created PID namespace. The parent will store
+  child's PID inside `<exec_file_name>.pid`, while the child drops privileges
+  and`exec()`s into the `<exec_file_name>`, as described below.
 - Drop privileges via setting the provided `uid` and `gid`.
 - Exec into `<exec_file_name> --id=<id>
   --start-time-us=<opaque> --start-time-cpu-us=<opaque>` (and also forward
@@ -224,11 +236,10 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
   this involves registering handlers with the cgroup `notify_on_release`
   mechanism, while being wary about potential race conditions (the instance
   crashing before the subscription process is complete, for example).
-- For extra resilience, the jailer expects to be spawned by the user in a new
-  PID namespace, most likely via a combination of `clone()` with the
-  `CLONE_NEWPID` flag and `exec()`. A process must be created in a new PID
-  namespace in order to become a pseudo-init process, and the other option is
-  to use a `clone()` in the jailer, which seems unnecessary.
+- For extra resilience, the `--new-pid-ns` flag enables the Jailer to exec the
+  binary file in a new PID namespace, in order to become a pseudo-init process.
+  Alternatively, the user can spawn the jailer in a new PID namespace via a
+  combination of `clone()` with the `CLONE_NEWPID` flag and `exec()`.
 - When running with `--daemonize`, the jailer will fail to start if it's a
   process group leader, because `setsid()` returns an error in this case.
   Spawning the jailer via `clone()` and `exec()` also ensures it cannot be a
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -143,7 +143,7 @@ impl ApiServer {
     /// let (api_request_sender, _from_api) = channel();
     /// let (to_api, vmm_response_receiver) = channel();
     /// let mmds_info = MMDS.clone();
-    /// let time_reporter = ProcessTimeReporter::new(Some(1), Some(1));
+    /// let time_reporter = ProcessTimeReporter::new(Some(1), Some(1), Some(1));
     ///
     /// thread::Builder::new()
     ///     .name("fc_api_test".to_owned())
@@ -751,7 +751,7 @@ mod tests {
                 )
                 .bind_and_run(
                     PathBuf::from(api_thread_path_to_socket),
-                    ProcessTimeReporter::new(Some(1), Some(1)),
+                    ProcessTimeReporter::new(Some(1), Some(1), Some(1)),
                     SeccompFilter::empty().try_into().unwrap(),
                 )
                 .unwrap();
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -16,7 +16,7 @@ use std::{fmt, io};
 
 use crate::parsed_request::ParsedRequest;
 use logger::{
-    debug, error, info, update_metric_with_elapsed_time, IncMetric, StoreMetric, METRICS,
+    debug, error, info, update_metric_with_elapsed_time, IncMetric, ProcessTimeReporter, METRICS,
 };
 pub use micro_http::{
     Body, HttpServer, Method, Request, RequestError, Response, ServerError, ServerRequest,
@@ -126,6 +126,7 @@ impl ApiServer {
     /// };
     /// use utils::{eventfd::EventFd, tempfile::TempFile};
     /// use vmm::vmm_config::instance_info::InstanceInfo;
+    /// use logger::ProcessTimeReporter;
     ///
     /// let mut tmp_socket = TempFile::new().unwrap();
     /// tmp_socket.remove().unwrap();
@@ -142,6 +143,7 @@ impl ApiServer {
     /// let (api_request_sender, _from_api) = channel();
     /// let (to_api, vmm_response_receiver) = channel();
     /// let mmds_info = MMDS.clone();
+    /// let time_reporter = ProcessTimeReporter::new(Some(1), Some(1));
     ///
     /// thread::Builder::new()
     ///     .name("fc_api_test".to_owned())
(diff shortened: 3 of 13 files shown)


src/jailer/src/env.rs
```diff
@@ -184,6 +189,41 @@ impl Env {
...
+        SyscallReturnCode(unsafe { libc::unshare(libc::CLONE_NEWPID) })
+            .into_empty_result()
+            .map_err(Error::UnshareNewPID)?;
+
+        // Duplicate the current process. The child process will belong to the previously created
+        // PID namespace.
+        let pid = unsafe { libc::fork() };
+        match pid {
+            0 => Err(Error::Exec(self.exec_command(chroot_exec_file))),
+            child_pid => Env::wait_child_process(child_pid),
```
acatangiu, line 206: Why should we keep this process open waiting for the child? Why not just terminate the parent?

docs/jailer.md
```diff
@@ -45,6 +46,8 @@ jailer --id <id> \
   jailer will use this to join the associated network namespace.
 - When present, the `--daemonize` flag causes the jailer to cal `setsid()` and
   redirect all three standard I/O file descriptors to `/dev/null`.
+- When present, the `--new-pid-ns` flag causes the jailer to exec the provided
```
acatangiu, line 49: We should also mention that the jailer forks then execs, the result being `firecracker` process will have a different _PID_ than the current `jailer` process.

(Without the flag, `firecracker` runs under the same process/same pid as its jailer).

src/jailer/src/env.rs
```diff
@@ -184,6 +189,41 @@ impl Env {
...
+        // The current process will not be moved into the newly created namespace, but its first
+        // child will assume the role of init(1) in the new namespace.
+        // The call is safe because we're invoking a C library function with valid parameters.
+        SyscallReturnCode(unsafe { libc::unshare(libc::CLONE_NEWPID) })
+            .into_empty_result()
+            .map_err(Error::UnshareNewPID)?;
+
+        // Duplicate the current process. The child process will belong to the previously created
+        // PID namespace.
+        let pid = unsafe { libc::fork() };
```
acatangiu, line 203: `libc::fork()` might be a big hammer here, can we do it by calling the underlying `clone()` syscall with the right set of args ourselves?

acatangiu (commented):
PR also needs an integration test to validate new pid ns functionality

KarthikNedunchezhiyan:
@luminitavoicu I have the below doubts in this model.

We are totally dependent on the PID of the jailer to kill the firecracker process, This implementation creating the child process (via fork), and then the child process is `exec`ing to the firecracker binary and jailer is waiting for the child process to exit. Will our model of killing firecracker works in this new implementation too in all the cases?

luminitavoicu:
> @luminitavoicu I have the below doubts in this model.
> 
> We are totally dependent on the PID of the jailer to kill the firecracker process, This implementation creating the child process (via fork), and then the child process is `exec`ing to the firecracker binary and jailer is waiting for the child process to exit. Will our model of killing firecracker works in this new implementation too in all the cases?
> 
> Edit:
> If the firecracker PID is changing it will be helpful if we get the new PID somehow, It will be needed to apply the cgroup after launching the jailer (to add PID in the task list).

Hi @KarthikNedunchezhiyan, 

Thank you for your comment. Your concerns are valid. However, it is important to note that this PR does not change the current behavior. The new `--new-pid-ns` jailer parameter is optional and should be added if the users do not want to worry about the PIDs. Since your model requires more control, you could stick with the old set of jailer parameters.

Regarding your edit, do you have any suggestions on how we could provide firecracker's new PID?

KarthikNedunchezhiyan:
> > @luminitavoicu I have the below doubts in this model.
> > We are totally dependent on the PID of the jailer to kill the firecracker process, This implementation creating the child process (via fork), and then the child process is `exec`ing to the firecracker binary and jailer is waiting for the child process to exit. Will our model of killing firecracker works in this new implementation too in all the cases?
> > Edit:
> > If the firecracker PID is changing it will be helpful if we get the new PID somehow, It will be needed to apply the cgroup after launching the jailer (to add PID in the task list).
> 
> Hi @KarthikNedunchezhiyan,
> 
> Thank you for your comment. Your concerns are valid. However, it is important to note that this PR does not change the current behavior. The new `--new-pid-ns` jailer parameter is optional and should be added if the users do not want to worry about the PIDs. Since your model requires more control, you could stick with the old set of jailer parameters.
> 

It makes sense 👍🏼 

> Regarding your edit, do you have any suggestions on how we could provide firecracker's new PID?

`firecracker.pid` file on the jail root directory works for me 🙂.

src/jailer/src/env.rs
luminitavoicu: For testing purposes, mainly. Terminating the parent here would make it hard to test that the child process has been indeed spawned in a PID namespace that is different from its parent's.

KarthikNedunchezhiyan: we are currently spawning the jailer into different PID namespace. we are using `/proc/<pid>/status`'s `﻿NStgid` value in integration testing to find PID namespace is nested and its PID is 0. Hope it will be useful here.

acatangiu: Indeed, let's try it this way, or find another way. We can't keep all these processes around doing nothing, they can incur a cost at scale.

src/firecracker/src/main.rs
```diff
@@ -93,6 +93,11 @@ fn main() {
                 .takes_value(true)
                 .help("Process start CPU time (wall clock, microseconds)."),
         )
+        .arg(
+            Argument::new("jailer-time-cpu-us")
+                .takes_value(true)
+                .help("Jailer process CPU time (wall clock, microseconds)."),
+        )
```
acatangiu, line 100: Could this just be hidden inside process start time? I would avoid having jailer specific logic in firecracker.

src/api_server/src/lib.rs
```diff
@@ -191,8 +193,9 @@ impl ApiServer {
         }
 
         if let Some(cpu_start_time) = start_time_cpu_us {
-            let delta_us =
-                utils::time::get_time_us(utils::time::ClockType::ProcessCpu) - cpu_start_time;
+            let delta_us = utils::time::get_time_us(utils::time::ClockType::ProcessCpu)
+                - cpu_start_time
+                + jailer_time_cpu_us.unwrap_or_default();
```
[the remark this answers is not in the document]
luminitavoicu: I meant to add a commit description but I missed that. I will update it. So the problem here is that the process CPU time counter is reset in the child process after `fork()` (see https://man7.org/linux/man-pages/man2/fork.2.html). This leads incorrectly calculating the `process_startup_time_cpu_us` metric, when the Firecracker process is spawned in a new PID namespace. 

Currently, we compute the metric like so:

`process_startup_time_cpu_us = current_cpu_time - cpu_start_time`

In the case of jailer spawning the Firecracker process in a new pid ns, `current _time` represents only the time spent running the child process (right after fork). The time spent executing jailer instructions is not taken into account.  
`cpu_start_time` represents the start time in the jailer process, which is not significant to the child process.


We need to have a way of tracking the CPU time spent from the start of the jailer up until Firecracker is up and ready. So I added an additional parameter: `jailer_time_cpu_us` and tried to unify the calculation of the cpu start time for both cases (when jailer and Firecracker run within the same  process and when jailer spawns Firecracker in a new PID namespace).

In the normal case (jailer and Firecracker run within the same  process),  `jailer_time_cpu_us` does not apply - it is 0 and nothing changes.

If Firecracker belongs to a different PID namespace from Jailer, in order to compute the total start-up time, we must add the cpu time spent in the child process to the cpu time spent running jailer instructions up until `fork()`. In this case `cpu_start_time` becomes 0.

src/jailer/src/env.rs
```diff
@@ -184,6 +196,108 @@ impl Env {
...
+        SyscallReturnCode(unsafe { libc::unshare(libc::CLONE_NEWPID) })
+            .into_empty_result()
+            .map_err(Error::UnshareNewPID)?;
+
+        let type_size = std::mem::size_of::<u64>();
+        let mut fds = vec![0; 2];
+        // Set up a pipe for the parent-child inter-process communication.
+        SyscallReturnCode(unsafe { libc::pipe(fds.as_mut_ptr()) })
+            .into_empty_result()
+            .map_err(Error::Pipe)?;
```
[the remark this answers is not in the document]
luminitavoicu: As I mentioned above, the ProcessCPU clock is reset in the child process after `fork()`. So we need some kind of inter-process communication in order to share the parent cpu time with the child.

src/firecracker/src/main.rs
luminitavoicu: See [comment above](https://github.com/firecracker-microvm/firecracker/pull/2488#discussion_r604873051)

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - Added devtool build `--ssh-keys` flag to support fetching from private
   git repositories.
 - Added option to configure block device flush.
+- Added '--jailer-time-cpu-us' flag to Firecracker, which represents the total
+  CPU time of the jailer process in us.
```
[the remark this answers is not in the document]
acatangiu: Actually, these params (`start_time_X`) were only supposed to be used by the CI for latency regression testing. They were originally hidden and did not even show up in `--help` - since moving from `clap` to `arg_parser` we lost the _hidden_ property and they are now visible which is a different issue altogether.

Long term we should aim to deprecate and remove these params and move to a different measuring mechanism in CI (use `perf` for example). The current method of firecracker measuring itself while taking into account parents' time, then expose that through its metrics is weird and as can be seen here, does not scale/is not flexible with other usecases. For now, we're stuck with adding yet another param for these measurements 😢 

Since this (`--jailer-time-cpu-us`) is supposed to only be used internally, let's not add it to the CHANGELOG, it's only distracting without being useful.

src/firecracker/src/api_server_adapter.rs
```diff
@@ -120,6 +120,7 @@ impl Subscriber for ApiServerAdapter {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
luminitavoicu: Done!

luminitavoicu:
@acatangiu, I addressed your comments. @acatangiu and @dianpopa, please take a look.

acatangiu (dismissed):
LGTM!

CHANGELOG.md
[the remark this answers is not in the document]
luminitavoicu: Opened issue here: https://github.com/firecracker-microvm/firecracker/issues/2516

acatangiu (approved):

acatangiu (approved):

src/firecracker/src/main.rs
```diff
@@ -93,6 +93,11 @@ fn main() {
                 .takes_value(true)
                 .help("Process start CPU time (wall clock, microseconds). This parameter is optional."),
         )
+        .arg(
+            Argument::new("parent-cpu-time-us")
+                .takes_value(true)
+                .help("Parent process CPU time (wall clock, microseconds). This parameter is optional."),
+        )
```
serban300: This adds complexity. There's also the option to allow `start_time_cpu_us` to take negative values. It would be simpler, but maybe a bit strage. I'm not sure which option is the cleanest.

luminitavoicu: I did take into consideration allowing negative values for `start_time_cpu_us`, which, indeed, would be much simpler. However, I found it to be too hacky, given the fact that the process CPU time in essence should never take negative values. In my opinion, adding a new parameter made for a cleaner approach.

acatangiu: We should just prioritize https://github.com/firecracker-microvm/firecracker/issues/2516 and get rid of all these params.

serban300: Ok, sounds good to me

src/logger/src/metrics.rs
```diff
@@ -280,6 +280,53 @@ impl Serialize for SharedStoreMetric {
     }
 }
 
+/// Reporter object which computes the process and process CPU times
+/// and populates the metric with the results.
+pub struct ProcessTimeReporter {
```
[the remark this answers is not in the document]
luminitavoicu: We had an off-line discussion and decided that the `ProcessTimeReporter` struct fits best inside the `logger` crate.

src/logger/src/metrics.rs
```diff
@@ -280,6 +280,53 @@ impl Serialize for SharedStoreMetric {
     }
 }
 
+/// Reporter object which computes the process and process CPU times
```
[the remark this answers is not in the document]
luminitavoicu: Done.

serban300 (approved):

acatangiu (approved):

PR #2504, opened 2021-03-15
tests: sync FS before block IO performance test

acatangiu:
## Description of Changes

Before running block IO performance test we drop caches using
`/proc/sys/vm/drop_caches` but that only drops clean caches, dirty
ones are kept around, more details can be found at
https://www.kernel.org/doc/Documentation/sysctl/vm.txt

Before dropping caches, run a `sync` command which will flush
all dirty FS cache entries.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/integration_tests/performance/test_block_performance.py
+++ b/tests/integration_tests/performance/test_block_performance.py
@@ -98,12 +98,18 @@ def run_fio(env_id, basevm, ssh_conn, mode, bs):
     assert rc == 0, stderr.read()
     assert stderr.read() == ""
 
-    run_cmd("echo 3 > /proc/sys/vm/drop_caches")
-
+    # First, flush all guest cached data to host, then drop guest FS caches.
+    rc, _, stderr = ssh_conn.execute_command("sync")
+    assert rc == 0, stderr.read()
+    assert stderr.read() == ""
     rc, _, stderr = ssh_conn.execute_command(
         "echo 3 > /proc/sys/vm/drop_caches")
     assert rc == 0, stderr.read()
     assert stderr.read() == ""
+
+    # Then, flush all host cached data to hardware, also drop host FS caches.
+    run_cmd("sync")
+    run_cmd("echo 3 > /proc/sys/vm/drop_caches")
 
     # Start the CPU load monitor.
     with concurrent.futures.ThreadPoolExecutor() as executor:


tests/integration_tests/performance/test_block_performance.py
```diff
@@ -98,13 +98,19 @@ def run_fio(env_id, basevm, ssh_conn, mode, bs):
...
+    assert rc == 0, stderr.read()
+    assert stderr.read() == ""
     rc, _, stderr = ssh_conn.execute_command(
         "echo 3 > /proc/sys/vm/drop_caches")
     assert rc == 0, stderr.read()
     assert stderr.read() == ""
 
+    # Then, flush all host cached data to hardware, also drop host FS caches.
+    run_cmd("sync")
+    run_cmd("echo 3 > /proc/sys/vm/drop_caches")
```
serban300, line 112: I'm not sure this is enough. In my experiments I had to also do `blockdev --flushbufs` in in order to drop the caches. Also the order matters. This is what I did:

```
sync
blockdev --flushbufs
echo 3 > /proc/sys/vm/drop_caches
```

acatangiu: Indeed, but that doesn't directly work here since this function is not tied to using a hw blockdev on the host side. I think it generically works with any type of host backend file (actual nvme drive, ebs, nfs, etc).

Should we tie this test to a particular host backing storage type then make sure we also flush hw/driver buffers like Serban noted?

@serban300 what is the impact of having vs not having `blockdev --flushbufs` in your experiments?

serban300: Oh you're right, this only applies when using a hw block device on the host side. In my experiments, the difference between calling `blockdev --flushbufs` and not calling it is almost 10x when reading. It's basically the difference between reading from the page cache and reading from the actual hw device.

If we're using a ramdisk, all the data will be in the memory anyway. @acatangiu Did you notice any performance difference when dropping the caches in this scenario?

acatangiu: I'm not sure if this test uses ramdisk, it doesn't seem to be configurable at the test level, need to dive deep in the framework.

alindima: Devtool itself is configurable to use or to not use ramdisk, so the test can be parametrised with the `-r` parameter.
In our performance pipeline, we run it with a ramdisk.

Before any updates to the tests or to the baselines are merged, we need to run a number of tests on m5d instances with all available CPU models to validate that they work.

acatangiu: For ramdisk as backing medium it should make no difference, for ebs this PR is the correct method, for local nvme drives (that have extra device-level caching) this PR is a step in the right direction but not a complete fix.

@serban300 @alindima what do you say we merge as is once I validate it runs successfully on m5d?

alindima (approved):

georgepisaltu (approved):

serban300 (approved):

PR #2505, opened 2021-03-16
upgrade vmm-sys-util

alindima:
# Reason for This PR

New version of vmm-sys-util was available, with a couple of fixes.

## Description of Changes

- update vmm-sys-util dependency of `utils` crate and adapt code for slight modification in `epoll.wait` interface
- update dependencies that use vmm-sys-util with versions that depend on the same version as Firecracker (versionize, kvm-bindings, kvm-ioctls). Adapt code to account for upstream API changes.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/arch/Cargo.toml
+++ b/src/arch/Cargo.toml
@@ -5,11 +5,11 @@ authors = ["The Chromium OS Authors"]
 edition = "2018"
 
 [dependencies]
-kvm-bindings = { version = "0.3.0", features = ["fam-wrappers"] }
-kvm-ioctls = { version = "0.6.0" }
+kvm-bindings = { version = ">=0.4.0", features = ["fam-wrappers"] }
+kvm-ioctls = ">=0.8.0"
 libc = ">=0.2.39"
 vm-memory = { path = "../vm-memory" }
-versionize = ">=0.1.4"
+versionize = ">=0.1.6"
 versionize_derive = ">=0.1.3"
 
 arch_gen = { path = "../arch_gen" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,39 +4,40 @@
 name = "aho-corasick"
 version = "0.7.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7404febffaa47dac81aa44dba71523c9d069b1bdc50a77db41195149e17f68e5"
 dependencies = [
- "memchr 2.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr",
 ]
 
 [[package]]
 name = "api_server"
 version = "0.1.0"
 dependencies = [
- "libc 0.2.81 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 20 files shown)


acatangiu (approved):
LGTM

alindima:
I'm working on some updates for aarch64 code

src/arch/Cargo.toml
```diff
@@ -5,11 +5,11 @@ authors = ["The Chromium OS Authors"]
 edition = "2018"
 
 [dependencies]
-kvm-bindings = { version = "0.3.0", features = ["fam-wrappers"] }
-kvm-ioctls = { version = "0.6.0" }
+kvm-bindings = { version = "=0.4.0", features = ["fam-wrappers"] }
```
andreeaflorescu, line 8: Why is this a fixed dependency now? Should you instead use `>=0.4.0` so that it is accurate in terms of minimum required version, but it is also not fixed and thus can be updated using `cargo update`?

acatangiu: Good catch @andreeaflorescu!

It's weird that https://github.com/alindima/firecracker/blob/37af8903ae0aaf7241a36bc349d2f4330a328055/Cargo.lock#L855 shows that only `0.8.0` is used.

Maybe this is used only in `aarch64` and there are still some changes required there, this being one of them.

alindima: AFAIK, we have to use version `0.4.0` in this toml file, since we `[patch]` this at the workspace level to use our internal fork: https://github.com/firecracker-microvm/firecracker/blob/37af8903ae0aaf7241a36bc349d2f4330a328055/Cargo.toml#L13
If somebody will perform a `cargo update` in the future to get version 5.0, they would bypass our patch without knowing it. Am I missing something?

acatangiu: I think the cargo `[patch]` overrides everything. Unless that's not the case, keeping underlying dependencies flexible `>=` sounds like a good idea to me.

alindima: If we leave `>=0.4.0` and upstream rust-vmm releases `0.5.0`, but we don't create a internal vmm-sys-util release in our fork, the `cargo update` will bypass the patch and give a warning:

```
warning: Patch `kvm-bindings v0.4.0 (https://github.com/firecracker-microvm/kvm-bindings?tag=v0.4.0-1#9b1dd38a)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
```

However, I think this is ok. This would force us to create an internal version corresponding to the upstream release. Either this, or modify our toml files.

alindima: done

alindima:
@acatangiu @andreeaflorescu addressed feedback and also fixed the code for aarch64.
PTAL

andreeaflorescu (approved):

acatangiu (approved):

Issue #2425, opened 2021-01-25
[Improvement] Snapshotting on ARM64 works only for GICv3


dianpopa:
# Describe the bug

Currently the code saving and restoring the state  of the GIC registers works only on GICv3. However, all other functionality is supported on GICv2 too. We would like to add support for being able to save and restore the state of a microVM on ARM64 using GICv2.

## To Reproduce

1. Start Firecracker via the usual process on an GICv2 enabled host
2. Try to run following commands:

  ```
curl --unix-socket $socket -i      -X PATCH "http://localhost/vm"      -H  "accept: application/json"      -H  "Content-Type: application/json"      -d "{
            \"state\": \"Paused\"
         }"


curl --unix-socket $socket -i      -X PUT "http://localhost/snapshot/create"      -H  "accept: application/json"      -H  "Content-Type: application/json"      -d "{
            \"snapshot_path\": \"lola.snap\",
            \"mem_file_path\": \"mem.snap\"
         }"
  ```
An error is received while trying to save the redistributors of the GIC.

## Expected behaviour

1. The above command is successful.
2. The below commands are also successful:
  ```
curl --unix-socket $socket -i      -X PUT "http://localhost/snapshot/load"      -H  "accept: application/json"      -H  "Content-Type: application/json"      -d "{
  \"snapshot_path\": \"lola.snap\",
  \"mem_file_path\": \"mem.snap\"
}"

curl --unix-socket $socket -i      -X PATCH "http://localhost/vm"      -H  "accept: application/json"      -H  "Content-Type: application/json"      -d "{
  \"state\": \"Resumed\"
}"
  ```

## Environment

- Firecracker version: v0.24.0
- Host and guest kernel versions - GICv2 enabled host, guest kernel version 4.14.203
- Architecture - ARM64

PR #2510, opened 2021-03-24
ARM: add snapshotting support for GICV2

dianpopa:
# Reason for This PR

FIxes https://github.com/firecracker-microvm/firecracker/issues/2425.

## Description of Changes

* First part of the PR consists of the refactoring of the gic folder which now contains 2 folders for GICv2 and GICv3 implementations.
* Second part adds support for the actual saving of the registers of the GICv2 device.
* Last part adds support for being able to specify an arbitrary GIC version so that we can use it in unit tests.
* All unit tests work when run on a machine with gicv2 support
* Integration tests that are currently failing: test_cache_topology (which is expected since our tests are tailored for m6g instance) and the test_5_full_snapshots due to performance reasons (we are currently opening 50 vsocks connections for testing that the guest was successfully resumed which results in a timeout).

Check out documentation for GICv2 specification: https://developer.arm.com/documentation/ihi0048/latest/

## License Acceptance

## PR Checklist


src/arch/src/aarch64/gic/regs.rs
```diff
@@ -1,24 +1,45 @@
...
+use crate::aarch64::gic::{Error, Result};
+
 #[derive(Debug)]
 pub struct GicRegState<T: Versionize> {
     pub(crate) chunks: Vec<T>,
 }
 
+/// Structure for serializing the state of the Vgic ICC regs
+#[derive(Debug, Default, Versionize)]
+pub struct VgicSysRegsState {
```
alindima, line 22: I feel like this structure is GIC-version dependent and that it was created for GICv3 and then shoehorned for V2 as well.

src/arch/src/aarch64/gic/regs.rs
```diff
@@ -1,24 +1,45 @@
...
+    /// Temp doc
+    pub dist: Vec<GicRegState<u32>>,
+    /// Temp doc
+    pub gic_vcpu_states: Vec<GicVcpuState>,
+}
+
+/// Structure used for serializing the state of the GIC registers for a specific vCPU
+#[derive(Debug, Default, Versionize)]
+pub struct GicVcpuState {
+    pub rdist: Vec<GicRegState<u32>>,
```
alindima, line 39: This struct is also gicv3 specific, since there is no redistributors for v2

alindima: Can we write traits for these structs and define them for each GIC version?

alindima:
I think this PR also needs a changelog update

dianpopa (commented):
@alindima I addressed your feedback. PTAL!

docs/snapshotting/snapshot-support.md
```diff
@@ -121,6 +121,8 @@ The baseline for snapshot resume latency target on Intel is under **8ms** with
   Please see [Vsock device limitations](#vsock-device-limitations).
 - Poor entropy and replayable randomness when resuming multiple microvms which
   deal with cryptographic secrets. Please see [Snapshot security and uniqueness](#snapshot-security-and-uniqueness).
+- Snapshotting on arm64 works for both GICv2 and GICv3 enabled guests.
+  However, restoring between different GIC version is not possible.
```
acatangiu, line 125: nit:
```suggestion
  However, restoring between different GIC versions is not possible.
```

alindima (dismissed):

georgepisaltu (approved):
LGTM

src/arch/src/aarch64/gic/regs.rs
sandreim: From the top of my head: one idea that might work would be to use an enum for each state. This would require some more translation code as we would need to change the structure of the snapshot.

alindima: @dianpopa and I talked offline and this is indeed an option that she brought up. I am fine however with keeping things as they are, since I don't see that much benefit apart from some slightly increased code expressiveness, wdyt?

dianpopa: I did take into account using an [enum for each state](https://github.com/dianpopa/firecracker/tree/gicv2_enum), however as decided to keep the current version since it does not require a snapshot version bump for gicv3 and so there is no hidden cost with it. 
One practical reason for using an enum could be to make the transition smoother to GICv4 (in case we want to support that in the future). However, gicv4 has the same programmer model as Gicv3 so we would be fine with using the same GIcState. 
Additionally, the enum implementation can be done incrementally on the current one.
So @sandreim are you ok with merging the PR like it is now or you foresee problems with current implementation?

sandreim: I think we are fine with less changes in state for now, unless we have clear data that shows otherwise (Gicv4 doesn't).

alindima (approved):

sandreim (approved):

PR #2515, opened 2021-03-30
snapshots: sync host resources as part of create-snapshot API call

acatangiu:
# Reason for This PR

Depending on the host FS configuration, snapshot-related files (memory, vm-state, block devices backing files) could get corrupted if the backing mediums are moved while snapshot data is in-flight.

## Description of Changes

To provide a safe and consistent CreateSnapshot operation, Firecracker now flushes all snapshot related host-file as part of the create snapshot operation.

## License Acceptance

## PR Checklist

- [ ] ~All added/changed functionality is tested.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 - Fixed the SIGPIPE signal handler so Firecracker no longer exits. The signal
   is still recorded in metrics and logs.
+- Snapshot related host files (vm-state, memory, block backing files) are now
+  flushed to their backing mediums as part of the CreateSnapshot operation.
 
 ### Changed
 
--- a/src/devices/src/virtio/block/persist.rs
+++ b/src/devices/src/virtio/block/persist.rs
@@ -7,7 +7,7 @@ use std::io;
 use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
 
-use logger::warn;
+use logger::{error, warn};
 use rate_limiter::{persist::RateLimiterState, RateLimiter};
 use snapshot::Persist;
 use versionize::{VersionMap, Versionize, VersionizeError, VersionizeResult};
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,8 @@
 
 - Fixed the SIGPIPE signal handler so Firecracker no longer exits. The signal
   is still recorded in metrics and logs.
+- Snapshot related host files (vm-state, memory, block backing files) are now
+  flushed to their backing mediums as part of the CreateSnapshot operation.
```
sandreim, line 16: Do we need a knob here ? There are also usecases when it is safe to allow the pagecache to deal with this.

acatangiu: I am not sure the added knob complexity (and especially understanding when to use it correctly) is worth the gains. Is CreateSnapshot an operation that is time-critical/time-sensitive enough to start messing with knobs that can break consistency when misused?

Conceptually I see the `CreateSnapshot` operation as a way to snapshot a live VM to a persistent, stable medium. If users need this operation to be fast, they can just use a fast medium (like nvme or ramfs, tmpfs, etc) without needing tricky knobs in `firecracker`.

src/devices/src/virtio/block/persist.rs
```diff
@@ -89,6 +89,11 @@ impl Persist<'_> for Block {
     type Error = io::Error;
 
     fn save(&self) -> Self::State {
+        // Sync data out to backing file on host.
+        if self.disk.file().sync_all().is_err() {
```
serban300, line 93: Should we also flush first ? In order to force any rust cached data out to the OS caches ?

acatangiu: Good catch Serban! Currently `flush` is noop in rust stdlib, but we should definitely do it explicitly instead of relying on underlying mechanisms which could change in the future.

acatangiu:
I need to change the "create_snapshot_latency" tests to move target snapshot files to a ramfs so that we accurately measure firecracker latency cost and not underlying storage medium speed.

PR #2524, opened 2021-04-24
Refactor create snapshot invalid version error

### LINKED ISSUE #2520: Make errors more specific and user friendly
Some of the snapshot error messages are very vague like reported here: https://github.com/firecracker-microvm/firecracker/pull/2503#discussion_r600533833

We  need to identify and fix the API and log error messages that need improvement.

lyuyuan:
# Reason for This PR

Fix #2520 
Make snapshot error message more specific.

## Description of Changes

* Use regex to validate microVM version format, and return invalid format error message.
* Use the previous error message when version format is valid but doesn't support snapshot.

## License Acceptance

## PR Checklist


sandreim (commented):
Thanks for the PR draft @lyuyuan. I like the extra checks on input version string, but I do not like the extra dependency on regex 😄. We are working on reducing the dependencies we pull in to improve the security posture, so this would be a step back wrt this goal. 

In this specific case we are gaining a more fine grained error message.  Ultimately, if the user specifies an improperly formatted version string, the call would still fail as we have static mappings for versions we do support (which are properly formatted) and a message covering both cases is a more simple approach.

Please take a look at other error strings as there are many places where the messages do not really make sense from a UX perspective but more from a developer perspective. I think there are more code paths that can throw more user friendly and/or finer grained errors.

src/vmm/src/persist.rs
```diff
@@ -585,6 +597,11 @@ mod tests {
                 _ => assert!(res.is_ok()),
             }
         }
+
+        assert!(
```
dianpopa, line 601: add a test case for when:
* there are more than 2 dots in the string
* only the last part of the string is incorrect: e.g: 0.21.x

src/vmm/Cargo.toml
```diff
@@ -11,6 +11,7 @@ serde = { version = ">=1.0.27", features = ["derive"] }
 serde_json = ">=1.0.9"
 versionize = ">=0.1.6"
 versionize_derive = ">=0.1.3"
+regex = { version = ">=1.0.0", default-features = false, features = ["std"] }
```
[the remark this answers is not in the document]
lyuyuan: Removed `regex` and added validation code as you described, thanks!

src/vmm/src/persist.rs
lyuyuan: New test cases added.

dianpopa (dismissed):

acatangiu (approved):

sandreim (approved):

PR #2529, opened 2021-04-28
Add `describe-snapshot` parameter to fetch snapshot data version

luminitavoicu:

# Reason for This PR

As snapshot format versions evolve, we need a mechanism to fetch the data version of a snapshot state file.

## Description of Changes

Added `describe-snapshot` parameter to fetch data version of the snapshot state file provided. Example:

```
$ ./firecracker --describe-snapshot full_snap.image
v0.23.0
```

## License Acceptance

## PR Checklist

### DIFF
--- a/src/firecracker/src/main.rs
+++ b/src/firecracker/src/main.rs
@@ -13,13 +13,15 @@ use std::sync::{Arc, Mutex};
 use logger::{error, info, IncMetric, ProcessTimeReporter, LOGGER, METRICS};
 use polly::event_manager::EventManager;
 use seccomp::{BpfProgram, SeccompLevel};
+use snapshot::Snapshot;
+use std::fs::File;
 use utils::arg_parser::{ArgParser, Argument};
 use utils::terminal::Terminal;
 use utils::validators::validate_instance_id;
 use vmm::default_syscalls::get_seccomp_filter;
 use vmm::resources::VmResources;
 use vmm::signal_handler::{mask_handled_signals, SignalManager};
-use vmm::version_map::FC_VERSION_TO_SNAP_VERSION;
+use vmm::version_map::{FC_VERSION_TO_SNAP_VERSION, VERSION_MAP};
 use vmm::vmm_config::instance_info::InstanceInfo;
 use vmm::vmm_config::logger::{init_logger, LoggerConfig, LoggerLevel};
 
@@ -142,6 +144,11 @@ fn main() {
             Argument::new("version")
                 .takes_value(false)
                 .help("Print the binary version number and a list of supported snapshot data format versions.")
+        )
+        .arg(
+            Argument::new("describe-snapshot")
+                .takes_value(true)
+                .help("Print the data format version of the provided snapshot state file.")
         );
 
     let arguments = match arg_parser.parse_from_cmdline() {
@@ -166,6 +173,11 @@ fn main() {
                 process::exit(i32::from(vmm::FC_EXIT_CODE_OK));
             }
 
+            if let Some(snapshot_path) = arg_parser.arguments().single_value("describe-snapshot") {
+                print_snapshot_data_format(snapshot_path);
+                process::exit(i32::from(vmm::FC_EXIT_CODE_OK));
+            }
+
             arg_parser.arguments()
         }
     };
@@ -187,8 +199,10 @@ fn main() {
         // It's safe to unwrap here because the field's been provided with a default value.
         let level = arguments.single_value("level").unwrap().to_owned();
         let logger_level = LoggerLevel::from_string(level).unwrap_or_else(|err| {
-            error!("Invalid value for logger level: {}. Possible values: [Error, Warning, Info, Debug]", err);
-            process::exit(i32::from(vmm::FC_EXIT_CODE_GENERIC_ERROR));
+            process::exit(i32::from(generic_error_exit(&format!(
+                "Invalid value for logger level: {}. Possible values: [Error, Warning, Info, Debug]",
+                err
+            ))));
         });
         let show_level = arguments.flag_present("show-level");
         let show_log_origin = arguments.flag_present("show-log-origin");
@@ -200,8 +214,10 @@ fn main() {
             show_log_origin,
         );
         init_logger(logger_config, &instance_info).unwrap_or_else(|err| {
-            error!("Could not initialize logger: {}", err);
-            process::exit(i32::from(vmm::FC_EXIT_CODE_GENERIC_ERROR));
+            process::exit(i32::from(generic_error_exit(&format!(
+                "Could not initialize logger:: {}",
+                err
+            ))));
         });
     }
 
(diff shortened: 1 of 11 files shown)


src/firecracker/src/main.rs
```diff
@@ -281,6 +294,31 @@ fn print_supported_snapshot_versions() {
...
+        if val == data_format_version {
+            Some(key)
+        } else {
+            None
+        }
+    });
+    match firecracker_data_version {
+        Some(val) => println!("v{}", val),
+        None => {
+            error!("{}", InvalidVersion);
```
sandreim, line 316: Can we give more information in the error message ?

src/firecracker/src/main.rs
```diff
@@ -281,6 +294,31 @@ fn print_supported_snapshot_versions() {
...
+    let mut snapshot_reader =
+        File::open(snapshot_path).expect("Unable to open snapshot state file");
+    let format_version_map = Snapshot::format_version_map();
+    let data_format_version =
+        Snapshot::get_format_version(&mut snapshot_reader, &format_version_map)
+            .expect("Invalid data format version of snapshot file");
+
+    let firecracker_data_version = FC_VERSION_TO_SNAP_VERSION.iter().find_map(|(key, &val)| {
+        if val == data_format_version {
+            Some(key)
```
sandreim, line 308: We can make this fn shorter by printing version here, and error below.

sandreim (commented):
Looks good overall. Please also include integration tests.

src/firecracker/src/main.rs
luminitavoicu: Done

luminitavoicu:
@sandreim, I addressed your comments and added the integration test. PTAL

src/firecracker/src/main.rs
```diff
@@ -281,6 +293,30 @@ fn print_supported_snapshot_versions() {
     println!("{}\n", snapshot_versions_str);
 }
 
+// Print data format of provided snapshot state file.
+fn print_snapshot_data_format(snapshot_path: &str) {
+    let mut snapshot_reader =
```
dianpopa, line 298: I do not think we want to panic if the file containing the state is invalid. I suggest to add logic that exits gracefully. You can add an auxiliary function to avoid duplicated code:
```
fn generic_error_exit(msg: &str) -> u8 {
    error!("{}", msg);
    return vmm::FC_EXIT_CODE_GENERIC_ERROR;
}
```
and
```
let mut snapshot_reader = File::open(snapshot_path).unwrap_or_else(|err| {
        process::exit(i32::from(generic_error_exit(&format!(
            "Unable to open snapshot state file: {}",
            err
        ))));
    });
    let data_format_version = Snapshot::get_data_version(&mut snapshot_reader, &VERSION_MAP)
        .unwrap_or_else(|err| {
            process::exit(i32::from(generic_error_exit(&format!(
                "Invalid data format version of snapshot file: {:?}",
                err
            ))));
        });
```

src/firecracker/src/main.rs
dianpopa: There is no need for `find_map`, simple `find` should suffice:
```
    let (key, _) = FC_VERSION_TO_SNAP_VERSION
        .iter()
        .find(|(_, &val)| val == data_format_version)
        .unwrap_or_else(|| {
            error!(
                "Cannot translate snapshot data version {} to Firecracker microVM version",
                data_format_version
            );
            process::exit(i32::from(vmm::FC_EXIT_CODE_GENERIC_ERROR))
        });
    println!("v{}", key);
```

luminitavoicu: Done

src/utils/src/arg_parser.rs
```diff
@@ -365,6 +367,32 @@ impl<'a> Arguments<'a> {
             return Ok(());
         }
 
+        // If `--describe-snapshot` is provided as a parameter, we artificially skip the parsing of
+        // other command line arguments by adding just the version argument to the parsed list and
```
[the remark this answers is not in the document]
luminitavoicu: Missed that. You are right

src/utils/src/arg_parser.rs
```diff
@@ -365,6 +367,32 @@ impl<'a> Arguments<'a> {
...
+            // It is safe to unwrap because the argument's presence has been confirmed.
+            let value_idx = args
+                .iter()
+                .position(|x| x == DESCRIBE_SNAPSHOT_ARG)
+                .unwrap()
+                + 1;
+            match args.get(value_idx) {
+                Some(value) => {
+                    if value.starts_with(ARG_PREFIX) {
+                        return Err(MissingArgument(describe_snapshot_arg.name.to_string()));
```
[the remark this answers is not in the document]
luminitavoicu: Done

luminitavoicu:
Comments addressed. @sandreim & @dianpopa, please take another look.

src/utils/src/arg_parser.rs
```diff
@@ -365,6 +367,32 @@ impl<'a> Arguments<'a> {
             return Ok(());
         }
 
+        // If `--describe-snapshot` is provided as a parameter, we artificially skip the parsing of
```
alindima, line 370: I don't fully understand why any changes to the arg parser crate are needed. So far, we've managed to keep this crate quite standalone, without logic that is tied to a specific binary (firecracker, jailer, or the upcoming seccompiler).

As far as I understand, the only benefit of this modification is that the parsing of the arguments stops if it encounters the --describe-snapshot parameter.
I think the PR should still work as expected by leaving the arg parser unchanged, given the fact that Firecracker terminates after receiving this parameter.
If we however want to make sure that we stop parsing the rest of the arguments, maybe we could add a new parameter attribute (like `required()`), that stops parsing after is encountered. Maybe smth like `.stop_if_present()`, that could be used for such cases, maybe even for `--version` and `--help`.

The added value would be that we would keep the arg_parser crate more generic. WDYT?

dianpopa: This is a very good observation @alindima! Indeed, it would be preferable to keep the arg parser implementation as generic as possible. At a closer look,  
we could get rid of all the specific logic inside the arg-parser for `help`, `version` and `describe-snapshot` and we would be having the same behavior (that is if we also insert the `help` parameter in the `ArgParser` just like we do for all the other ones). @luminitavoicu is there something else about this logic that we are missing? If not, I am ok with deleting it.
@alindima I do not have any strong opinion about adding a `stop_if_present` flag, does not seem to add significant value.

luminitavoicu: I agree with you, @alindima and @dianpopa, the newly added `describe-snapshot` parameter should not make its way to the arg parser logic. I will move it.

I think it makes sense to also get rid of the extra logic for `help` and `version` inside the arg parser. Whether the parsing of the arguments stops when encountering one of these two parameters does not make a significant difference to me. I think adding a `stop_if_present` flag is not really necessary.

alindima: sounds good to me

alindima (dismissed):

dianpopa (approved):

alindima (approved):

dianpopa (approved):

PR #2532, opened 2021-04-30
[Balloon] Update `deflate_on_oom` documentation

georgepisaltu:
Signed-off-by: George Pisaltu <gpl@amazon.com>

# Reason for This PR

The ballooning documentation was misleading with regards to the `deflate_on_oom` setting.

## Description of Changes

Clarified the behavior of `deflate_on_oom` in `ballooning.md`.

## License Acceptance

## PR Checklist

### DIFF
--- a/docs/ballooning.md
+++ b/docs/ballooning.md
@@ -23,10 +23,20 @@ The device can be configured with the following options:
   allocate some memory which would make the guest enter an out-of-memory state,
   the kernel will take some pages from the balloon and give them to said
   process instead asking the OOM killer process to kill some processes to free
-  memory. Note that this applies to allocations from guest processes which would
-  make the system enter an OOM state. This does not apply to instances when the
-  kernel needs memory for its activities (i.e. constructing caches), or when the
-  user requests more memory than the amount available through an inflate.
+  memory. Note that this applies to physical page allocations in the kernel
+  which belong to guest processes. This does not apply to instances when the
+  kernel needs memory for its activities (i.e. constructing caches), when the
+  user requests more memory than the currently available to the balloon for
+  releasing, or when guest processes try to allocate large amounts of memory
+  that are refused by the guest memory manager, which is possible when the
+  guest runs with `vm.overcommit_memory=0` and the allocation does not pass
+  the MM basic checks. Setting `vm.memory_overcommit` to 1 would make the MM
+  approve all allocations, no matter how large, and using the memory mapped
+  for those allocations will always deflate the balloon instead of making the
+  guest enter an OOM state. Note: we do not recommend running with
+  `vm.overcommit_memory=1` because it requires complete control over what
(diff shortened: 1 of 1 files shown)


dianpopa (dismissed):

docs/ballooning.md
```diff
@@ -23,10 +23,14 @@ The device can be configured with the following options:
...
   the kernel will take some pages from the balloon and give them to said
   process instead asking the OOM killer process to kill some processes to free
-  memory. Note that this applies to allocations from guest processes which would
-  make the system enter an OOM state. This does not apply to instances when the
-  kernel needs memory for its activities (i.e. constructing caches), or when the
-  user requests more memory than the amount available through an inflate.
+  memory. Note that this applies to physical page allocations in the kernel
+  which belong to guest processes. This does not apply to instances when the
+  kernel needs memory for its activities (i.e. constructing caches), when the
+  user requests more memory than the amount available through an inflate, or
```
acatangiu, line 29: should it be
```suggestion
  user requests more memory than the amount available through a deflate, or
```
?

or clearer (from my pov):
```
user requests more memory than the currently _held_ by balloon, or
```
or
```
user requests more memory than the currently available to the balloon for releasing, or
```

docs/ballooning.md
```diff
@@ -23,10 +23,14 @@ The device can be configured with the following options:
...
+  when guest processes try to allocate large amounts of memory that are refused
+  by the guest memory manager (possible when the guest runs with
+  `vm.overcommit_memory=0` and the allocation does not pass the MM basic
+  checks).
```
acatangiu, line 33: I would end this paragraph with smth like:
```
In conclusion, `deflate_on_oom` is also a best-effort mechanism and does not guarantee that no OOM can happen in the guest.
```

docs/ballooning.md
georgepisaltu: I'll go with the second one.

docs/ballooning.md
georgepisaltu: This is partly true. If we say `deflate_on_oom` is best-effort, it would convey that its behavior is non-deterministic (that's how I would understand it at least), which is not the case. The behavior is clearly defined, it's just the conditions that are hard to define. I tried to capture this by mentioning `vm.memory_overcommit` and its behavior.

I could add something like:
```
Setting vm.memory_overcommit to 1 would make the MM approve all allocations, no matter how large, and using the memory mapped for those allocations will always deflate the balloon instead of making the guest enter an OOM state. Note: we do not recommend setting this to 1 because........
```
WDYT?

acatangiu: Sounds good.

acatangiu (approved):

luminitavoicu (approved):

acatangiu (approved):

PR #2535, opened 2021-05-01
Clean Shutdown on Debug Builds (Valgrind-friendly)

AE1020:
#119 expresses the desire to use Valgrind.  However, Firecracker's only modality of exiting the process was from within nested stacks using `unsafe { libc::_exit(...) }` from the main thread.  This would mean not running destructors for anything on other threads, or above the stack on the main thread.  It also means that there may be many shutdown paths which have faulty logic and are never being tested.  *(Such was the case with `Vmm's Drop` and `VcpuHandle's Drop`, which could never run due to reference cycles.)*

However, abrupt exits are faster in release build situations, where one is not looking for leaks.  Also, adding clean exit mechanics means creating new code paths which may deadlock and not quit as reliably.

This PR attempts to bring the best of both approaches <strike>by adding a `--cleanup-level` switch</strike> by running a Valgrind-friendly exit mode in debug builds.

The result of Valgrind on running the Alpine Linux image through boot and shutdown is:

    ==52247== LEAK SUMMARY:
    ==52247==    definitely lost: 0 bytes in 0 blocks
    ==52247==    indirectly lost: 0 bytes in 0 blocks
    ==52247==      possibly lost: 0 bytes in 0 blocks
    ==52247==    still reachable: 8,594 bytes in 10 blocks
    ==52247==         suppressed: 0 bytes in 0 blocks

[Here are backtraces of the still reachable allocations.](https://gist.github.com/AE1020/11f96dd8c7dd8a8d48790858a84ef1cf)  They may not represent leaks, and should be in a suppression list if they don't.  As examples get larger, having a baseline gives much better odds of discerning what's a leak and what is not.

## Description of Changes

The changes are broken into 4 parts, each with a commit description detailing what they do:

* Basic methodology for bubbling up an ExitCode all the way up from detecting an exit condition to main(), so that all destructors on the main thread can run.  (Although they *could* run in theory, some didn't in practice, due to reference cycles which needed resolving.)
* Graceful exit mechanics for the API server thread.
* Graceful exit mechanics for the VCPU threads.
* <strike>Addition of a `--cleanup-level` switch, reintroducing the original exit semantics as the default behavior.</strike>
* Limiting the clean exit behavior to the debug build via `#[cfg(debug_assertions)]` annotations.

Open questions and comments are inline in the code.

## License Acceptance

## PR Checklist

- [N/A] Any API changes are reflected in `firecracker/swagger.yaml`.

Regarding testing, my VirtualBox is very bad at that... sending PR for discussion and so that CI runs it.

### DIFF
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -198,69 +198,104 @@ impl ApiServer {
         }
 
         server.start_server().expect("Cannot start HTTP server");
+
         loop {
-            match server.requests() {
-                Ok(request_vec) => {
-                    for server_request in request_vec {
-                        let request_processing_start_us =
-                            utils::time::get_time_us(utils::time::ClockType::Monotonic);
-                        server
-                            .respond(
-                                // Use `self.handle_request()` as the processing callback.
-                                server_request.process(|request| {
-                                    self.handle_request(request, request_processing_start_us)
-                                }),
-                            )
-                            .or_else(|e| {
-                                error!("API Server encountered an error on response: {}", e);
-                                Ok(())
-                            })?;
-                        let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic)
-                            - request_processing_start_us;
-                        debug!("Total previous API call duration: {} us.", delta_us);
-                        if self.vmm_fatal_error {
-                            // Flush the remaining outgoing responses
-                            // and proceed to exit
-                            server.flush_outgoing_writes();
-                            error!(
-                                "Fatal error with exit code: {}",
-                                FC_EXIT_CODE_BAD_CONFIGURATION
-                            );
-                            unsafe {
-                                libc::_exit(i32::from(FC_EXIT_CODE_BAD_CONFIGURATION));
-                            }
-                        }
-                    }
-                }
-                Err(e) => {
+            let request_vec = match server.requests() {
+                Ok(vec) => vec,
+                Err(e) => {  // print request error, but keep server running
                     error!(
                         "API Server error on retrieving incoming request. Error: {}",
                         e
                     );
+                    continue
+                }
+            };
+
+            for server_request in request_vec {
+                let request_processing_start_us =
+                    utils::time::get_time_us(utils::time::ClockType::Monotonic);
+
+                // !!! ServerRequest::process() doesn't appear to do anything besides put a private
+                // id field into the response.  But since it takes an inner function it cannot
+                // return from this function without giving a response.  Review a better way of
+                // exiting cleanly than by reaching out of the closure to set a boolean.
+
+                let mut shutting_down = false;
+
+                server
+                    .respond(
+                        server_request.process(|request| {
+                            match self.handle_request_or_finishing(request, request_processing_start_us) {
+                                Some(response) => response,
+                                None => {
+                                    shutting_down = true;
+                                    Response::new(Version::Http11, StatusCode::NoContent)
+                                }
+                            }
+                        }),
+                    )
+                    .or_else(|e| {
+                        error!("API Server encountered an error on response: {}", e);
+                        Ok(())
+                    })?;
+
+                let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic)
+                    - request_processing_start_us;
+                debug!("Total previous API call duration: {} us.", delta_us);
+
+                if shutting_down {
+                    debug!("/shutdown-internal request received, API server thread now ending itself");
+                    return Ok(());
+                }
+
+                if self.vmm_fatal_error {
+                    // Flush the remaining outgoing responses
+                    // and proceed to exit
+                    server.flush_outgoing_writes();
+                    error!(
+                        "Fatal error with exit code: {}",
+                        FC_EXIT_CODE_BAD_CONFIGURATION
+                    );
+                    unsafe {
+                        libc::_exit(i32::from(FC_EXIT_CODE_BAD_CONFIGURATION));
+                    }
                 }
             }
         }
     }
 
     /// Handles an API request received through the associated socket.
+    /// If None is given back, it means a ShutdownInternal was requested and no response
+    /// is expected (should be requested by the main thread, not by clients)
+    pub fn handle_request_or_finishing(
+        &mut self,
+        request: &Request,
+        request_processing_start_us: u64,
+    ) -> Option<Response> {
+        match ParsedRequest::try_from_request(request) {
+            Ok(ParsedRequest::Sync(vmm_action)) => {
+                Some(self.serve_vmm_action_request(vmm_action, request_processing_start_us))
+            }
+            Ok(ParsedRequest::GetInstanceInfo) => Some(self.get_instance_info()),
+            Ok(ParsedRequest::GetMMDS) => Some(self.get_mmds()),
+            Ok(ParsedRequest::PatchMMDS(value)) => Some(self.patch_mmds(value)),
+            Ok(ParsedRequest::PutMMDS(value)) => Some(self.put_mmds(value)),
+            Ok(ParsedRequest::ShutdownInternal) => None,
+            Err(e) => {
+                error!("{}", e);
+                Some(e.into())
+            }
+        }
+    }
+
+    /// Variant of handle_request that is used by tests, and does not know about the
+    /// ShutdownInternal mechanism.  So the response is non-optional.
     pub fn handle_request(
         &mut self,
         request: &Request,
         request_processing_start_us: u64,
     ) -> Response {
-        match ParsedRequest::try_from_request(request) {
-            Ok(ParsedRequest::Sync(vmm_action)) => {
-                self.serve_vmm_action_request(vmm_action, request_processing_start_us)
-            }
-            Ok(ParsedRequest::GetInstanceInfo) => self.get_instance_info(),
-            Ok(ParsedRequest::GetMMDS) => self.get_mmds(),
-            Ok(ParsedRequest::PatchMMDS(value)) => self.patch_mmds(value),
-            Ok(ParsedRequest::PutMMDS(value)) => self.put_mmds(value),
-            Err(e) => {
-                error!("{}", e);
-                e.into()
-            }
-        }
+        self.handle_request_or_finishing(request, request_processing_start_us).unwrap()
     }
 
     fn serve_vmm_action_request(
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -280,7 +280,10 @@ impl ApiServer {
             Ok(ParsedRequest::GetMMDS) => Some(self.get_mmds()),
             Ok(ParsedRequest::PatchMMDS(value)) => Some(self.patch_mmds(value)),
             Ok(ParsedRequest::PutMMDS(value)) => Some(self.put_mmds(value)),
+
+            #[cfg(debug_assertions)]
             Ok(ParsedRequest::ShutdownInternal) => None,
+
             Err(e) => {
                 error!("{}", e);
                 Some(e.into())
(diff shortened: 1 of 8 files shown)


acatangiu:
I looked at the proposed changes and I like the consistent and structured teardown mechanisms added here 👍 

However, there is a part of the story which this PR doesn't consider - it is also the reason for the current _abrupt_ shutdown mechanism:

**Seccomp/syscall filtering**: Firecracker has multiple layers of security hardening and workload isolation [1] - one of these layers is *syscall filtering* through seccomp functionality. The default/recommended filter for production workloads is a tight whitelist that doesn't include some syscalls done by Rust during thread exit (I believe `rt_sigprocmask` is first to hit the filter) - meaning that when default seccomp filter is installed the _clean_ teardown will result in a _seccomp violation panic_.
The PR often references teardown _speed/latency_ as being a factor in the design - it is not, seccomp is - process teardown should not be on any critical path afaict.

Having runtime knob for (dis/en)abling the clean teardown mechanism introduces compatibility issues with the seccomp knob and filters, resulting in a complicated matrix of behavior depending on configuration.

Ideally we want to avoid that!

My recommendation on approaches to investigate (in order of preference):
1. Take another hard look at the missing syscalls from our default/recommended seccomp whitelist, that prevent proper thread cleanup and exiting. If we consider them _safe enough_ to be included, drop the _abrupt_ exit path and do clean exiting along the lines of this PR. (For example, `rt_sigprocmask` looks safe to me for whitelist inclusion - it only changes signal mask for current thread, does not affect system or other processes.)
2. Don't change seccomp filters, keep abrupt exit to be compatible with seccomp filters, add _clean_ exit as a **compile-time** option and only use it for `valgrind` and other testing tools. (This option might result in very ugly code.)
3. Even if we do none of the above and just forget about running `valgrind`, some of the cleanup work proposed by this PR (like fixing the `Vmm<->Vcpu` cyclic dependency) is still valuable and could go in regardless.

[1] - https://github.com/firecracker-microvm/firecracker/blob/main/docs/design.md#threat-containment

**P.S.** looks like unit-tests are hanging in current PR - do they pass for you locally?

AE1020:
> Even if we do none of the above and just forget about running valgrind, some of the cleanup work proposed by this PR (like fixing the Vmm<->Vcpu cyclic dependency) is still valuable and could go in regardless.

The Vmm<->Vcpu issue points to what I think is kind of a bigger point about the value of clean shutdown.  It doesn't just let you check Valgrind leaks, it's a "systemic check of assumptions".  Knowing you can unwind everything successfully helps build confidence that nothing got entangled while winding it up.  It also means you can shutdown the system fully and start it back up within the same process when you have a reason to do so.

> when default seccomp filter is installed the clean teardown will result in a seccomp violation panic.

Ah.  Valgrind cannot run *at all* with seccomp filtering.  So whenever I was running `--cleanup-level 2`, I happened to be also running it with `--seccomp-level 0`, thus did not get bit by this.

On the bright side: just changing that final moment during clean shutdown from `std::process::exit()` => `unsafe { libc::_exit() }`, only makes a tiny difference... 16 "still reachable" blocks instead of 10.  That's not bad.  This particular statistic will probably never go to zero, and hopefully won't grow on larger cases.  So once you've read through them and deemed them ok you can make a suppression list.

[Hence I've changed whether to call `std::process::exit()` based on whether seccomp filtering was enabled or not.](https://github.com/firecracker-microvm/firecracker/pull/2535/commits/314da6ce34ad648dcb793f108443e25349ef29c5#diff-b7b8b91b3aa8da3f6f9c0e172a280551452ab918ca41f5a5d2fda685f08ed607R295)

> Having runtime knob for (dis/en)abling the clean teardown mechanism introduces compatibility issues with the seccomp knob and filters, resulting in a complicated matrix of behavior depending on configuration.

There already was such a matrix...but in a way that created mechanics that only existed in tests, with no way of happening in the actual program.  See the [#[cfg(test)] exit() state for VCPU](https://github.com/firecracker-microvm/firecracker/blob/ffbc2efcf2dcffa02db6df3474a8ad2b0bad21f7/src/vmm/src/vstate/vcpu/mod.rs#L434) and the [#[cfg(not(test))] exit state for VCPU](https://github.com/firecracker-microvm/firecracker/blob/ffbc2efcf2dcffa02db6df3474a8ad2b0bad21f7/src/vmm/src/vstate/vcpu/mod.rs#L407).

Of course no matrix at all is even better.  But not defaulting to forcing an exit creates risks of running into deadlocks if there are bugs in the shutdown process...which could be bad in deployment.  I was trying to think of how to mitigate that.

It sounds like you would prefer this being a compile-time flag.  So I've removed `--cleanup-level` and changed it so that [whether exit() runs abruptly or not is controlled by #[cfg(debug_assertions)]](https://github.com/firecracker-microvm/firecracker/pull/2535/commits/314da6ce34ad648dcb793f108443e25349ef29c5#diff-142768394cb563e256dfe5b973511095fd91671c74af023a57a814afe69b0058R390).

What you lose in this exchange is an easy way to do "valgrind on release build".  Which could be a separate flag for those who want it.  But making all debug builds shut down cleanly regardless means the behavior is part of a release/debug distinction that's already tested.

> looks like unit-tests are hanging in current PR - do they pass for you locally?

I don't have a bare metal Linux machine at present, and am running inside of VirtualBox.  Hence tests have tons of failures.  It's also glacially slow compared to the CI--as in hours.  (My laptop may have been good when I got it, but it's 5 years old now, and it's a laptop running VMs inside of VMs).  Hence I needed the CI output generated by the PR to compare with.  I was able to fix compilation errors in the tests due to that.

But the hanging failure is a very generic message of "test runner failed" that points to the top level of something.  I don't know how to translate that into an invocation I can run myself and debug.  If you can provide a step-by-step of how to turn an error like that into something actionable, I could look at it.

acatangiu:
Let me take a look at this:

> Take another hard look at the missing syscalls from our default/recommended seccomp whitelist, that prevent proper thread cleanup and exiting. If we consider them safe enough to be included, drop the abrupt exit path and do clean exiting along the lines of this PR. (For example, rt_sigprocmask looks safe to me for whitelist inclusion - it only changes signal mask for current thread, does not affect system or other processes.)

Ideal state would be to have clean teardown on release builds as well and not juggle multiple binaries. Let's aim for ideal scenario first 😄 

In the meantime, @firecracker-microvm/compute-capsule please take a look at the proposed changes for bubbling up errors, retcodes, exit reasons, etc.

@andreeaflorescu if you get a chance, please take a look at the `polly` changes. Would they be compatible with https://github.com/rust-vmm/event-manager/ ?

acatangiu:
@AE1020 I've looked into filters and found that we can safely whitelist the missing few required for clean teardown.
Please see commit https://github.com/acatangiu/firecracker/commit/e40b1e2f7ef2dce799c149c2283d8fe41cba42a2 which is on top of your changes and allows clean teardown even with seccomp filtering enabled.

In this context, I suggest we keep your clean teardown changes but remove all branching between `debug` and `release` builds. All builds can just do clean teardown. Wdyt?

andreeaflorescu:
> @andreeaflorescu if you get a chance, please take a look at the `polly` changes. Would they be compatible with https://github.com/rust-vmm/event-manager/ ?

This PR introduces a big chance, and I would need more time to provide a helpful review.

A few points that I've observed looking quickly through the code:
- I noticed some asserts in the polly implementation; those should be avoided in library crates
- We're using ExitCode which is an u8; that is atypical for handling Rust exits though in library crates; I was expecting some sort of error to be returned instead

We're implementing clean exit in rust-vmm/vmm-reference (it still has some loose ends that we need to sort out), but we're doing that using the event-manager existing interface. We didn't encounter any problems so far. I am not sure if this is the same problem that you're trying to solve here, but we needed a way to exit from the event manager run loop. This was implemented through an `ExitHandler` which we check before re-entering the loop. The code is here: https://github.com/rust-vmm/vmm-reference/blob/master/src/vmm/src/lib.rs#L282

acatangiu:
@andreeaflorescu I didn't formulate my question properly, sorry.

Without going into detailed changes of `polly` in this PR, I am more interested in whether you think conceptually `event-manager` could have _user-provided Errors/break-conditions_. That way a `Subscriber` can also bubble up errors and or special return values originating in their event handling.

Doing so would avoid needing side-channels communication like this: https://github.com/rust-vmm/vmm-reference/blob/master/src/vmm/src/lib.rs#L282 ...

Anyway, I think we should move this conversation to a feature-request issue on https://github.com/rust-vmm/event-manager or similar.

FYI @AE1020 question above is not a blocker for current PR, it's us thinking ahead to when we switch `polly` for `rust-vmm/event-manager`.

andreeaflorescu:
It seems like a fair request to be able to break out of the event run loop. We can discuss about how to design this in rust-vmm/event-manager as you suggested @acatangiu.

AE1020:
I'm certainly fine with however you see fit to change it.  *(This was just my first attempt to program in Rust, which I wanted to do in an area where the security aspects were relevant.  And I wanted it to be something real instead of just make-work.  It has thus  served the educational purpose I had...I don't mind if you want to take over and do it a totally different way.)*

As the comments mention, [the HTTP "/shutdown-internal" request](https://github.com/firecracker-microvm/firecracker/pull/2535/files#diff-993b48801c3421206c29fe7f66a3862b1afd360150e53f07220b54fe0979b98cR64) I added is really just a hack because there was no other way to stop the API thread.  If there were some other channel of communication and shutdown could be negotiated without that, a real user-exposed endpoint to ask for shutdown seems a useful thing in general.

I'm actually in the middle of moving this week, and have some other coding commitments.  But if you have something specific you'd like me to look at let me know.  As I mentioned, I don't really know how to create a command line corresponding to a test failure to debug.  And it would all be different if the shutdown mechanism is changed.

sandreim:
I agree with @andreeaflorescu that breaking the eventloop is a fair feature request. Clean shutdown is something that has clear value outside the scope of Valgrind friendlyness, but I think requires more design work as it does touch a lot of components like EventManager. We should have that discussion in https://github.com/rust-vmm/event-manager as we plan to use it as a Polly replacement.

@AE1020 @acatangiu WDYT about a less invasive approach here like panicing! all threads so destructors are still called. We should check how memory usage is registered in Valgrind if we exit the VMM this way.

acatangiu:
> @AE1020 @acatangiu WDYT about a less invasive approach here like panicing! all threads so destructors are still called. We should check how memory usage is registered in Valgrind if we exit the VMM this way.

I am working on a variation of this PR that does clean shutdown [1]. The actual end result should be an improvement in firecracker code complexity not a degradation since things then naturally fit the RAII principle and are easier to mentally model.

Before publishing the PR I want to also add rust sanitizer tests which would be available for us to run when doing clean teardown; there is still some work left to do, and it's currently on the back-burner, I'm estimating to have an RFC end of next week.

[1] https://github.com/acatangiu/firecracker/tree/clean-shutdown

andreeaflorescu:
@acatangiu it looks like we are trying to achieve the same things in Firecracker, so I was wondering if you got the chance to look at how clean shut down is implemented in vmm-reference? At a first glance it looks like it is fairly easy to port the changes in Firecracker, and we wouldn't need to implement it from scratch. The vmm-reference is not using all the constructs that Firecracker does (i.e. it does not yet implement a way of sending messages between vCPUs), but we are planning to add this support when upstreaming vm-vcpu. Since it looks like clean shut down is something that we want to merge in Firecracker faster than we would consume something like vm-vcpu, I wonder if we could still align the implementations, or maybe clearly state why they're incompatible.

acatangiu:
Closing this in favor of #2599

acatangiu:
> Since it looks like clean shut down is something that we want to merge in Firecracker faster than we would consume something like vm-vcpu, I wonder if we could still align the implementations, or maybe clearly state why they're incompatible.

@andreeaflorescu please see #2599 : I have used the same model of breaking the main event loop as used by vmm-reference, so the two implementations are now aligned from that perspective. Regarding Vcpu exit handler, I've chosen to keep a simple EventFd instead of the complicated generic construct vmm-reference is using that allows customizable exit methods - even so, the two implementations are equivalent so we can easily consume the generic mechanism in upstream vm-vcpu when it will be available.

PR #2536, opened 2021-05-04
snapshots: sync host resources as part of create-snapshot API call

acatangiu:
# Reason for This PR

Depending on the host FS configuration, snapshot-related files (memory, vm-state, block devices backing files) could get corrupted if the backing mediums are moved while snapshot data is in-flight.

## Description of Changes

To provide a safe and consistent CreateSnapshot operation, Firecracker now flushes all snapshot related host-file as part of the create snapshot operation.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/persist.rs
+++ b/src/devices/src/virtio/block/persist.rs
@@ -3,11 +3,11 @@
 
 //! Defines the structures needed for saving/restoring block devices.
 
-use std::io;
+use std::io::{self, Write};
 use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
 
-use logger::warn;
+use logger::{error, warn};
 use rate_limiter::{persist::RateLimiterState, RateLimiter};
 use snapshot::Persist;
 use versionize::{VersionMap, Versionize, VersionizeError, VersionizeResult};
@@ -89,6 +89,14 @@ impl Persist<'_> for Block {
     type Error = io::Error;
 
     fn save(&self) -> Self::State {
+        if let Err(e) = self.disk.file().flush() {
+            error!("Failed to flush block data on serialization. Error: {}", e);
+        }
+        // Sync data out to backing file on host.
+        if let Err(e) = self.disk.file().sync_all() {
+            error!("Failed to sync block data on serialization. Error: {}", e);
+        }
+        // Save device state.
         BlockState {
             id: self.id.clone(),
             partuuid: self.partuuid.clone(),
--- a/src/vmm/src/persist.rs
+++ b/src/vmm/src/persist.rs
@@ -5,7 +5,7 @@
 
 use std::fmt::{Display, Formatter};
 use std::fs::{File, OpenOptions};
-use std::io;
+use std::io::{self, Write};
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 
@@ -116,12 +116,16 @@ pub enum CreateSnapshotError {
     Memory(memory_snapshot::Error),
     /// Failed to open memory backing file.
     MemoryBackingFile(io::Error),
+    /// Failed to flush memory backing file contents.
+    MemoryFileFlush(io::Error),
     /// Failed to save MicrovmState.
     MicrovmState(MicrovmStateError),
     /// Failed to serialize microVM state.
     SerializeMicrovmState(snapshot::Error),
     /// Failed to open the snapshot backing file.
     SnapshotBackingFile(io::Error),
+    /// Failed to flush the snapshot backing file.
+    SnapshotFileFlush(io::Error),
     #[cfg(target_arch = "x86_64")]
     /// Number of devices exceeds the maximum supported devices for the snapshot data version.
     TooManyDevices(usize),
@@ -139,9 +143,11 @@ impl Display for CreateSnapshotError {
             InvalidVmState(err) => write!(f, "Cannot save Vm state. Error: {:?}", err),
             Memory(err) => write!(f, "Cannot write memory file: {:?}", err),
             MemoryBackingFile(err) => write!(f, "Cannot open memory file: {:?}", err),
+            MemoryFileFlush(err) => write!(f, "Cannot flush memory file contents: {:?}", err),
             MicrovmState(err) => write!(f, "Cannot save microvm state: {}", err),
             SerializeMicrovmState(err) => write!(f, "Cannot serialize MicrovmState: {:?}", err),
             SnapshotBackingFile(err) => write!(f, "Cannot open snapshot file: {:?}", err),
+            SnapshotFileFlush(err) => write!(f, "Cannot flush snapshot file: {:?}", err),
             #[cfg(target_arch = "x86_64")]
             TooManyDevices(val) => write!(
                 f,
@@ -235,8 +241,8 @@ fn snapshot_state_to_file(
     snapshot
         .save(&mut snapshot_file, microvm_state)
         .map_err(SerializeMicrovmState)?;
-
-    Ok(())
+    snapshot_file.flush().map_err(SnapshotFileFlush)?;
+    snapshot_file.sync_all().map_err(SnapshotFileFlush)
 }
 
 fn snapshot_memory_to_file(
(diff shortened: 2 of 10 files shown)


src/vmm/src/persist.rs
```diff
@@ -235,8 +241,7 @@ fn snapshot_state_to_file(
     snapshot
         .save(&mut snapshot_file, microvm_state)
         .map_err(SerializeMicrovmState)?;
-
-    Ok(())
+    snapshot_file.sync_all().map_err(SnapshotFileFlush)
```
georgepisaltu, line 244: No `file.flush()` here like in the block implementation?

I understand it's a noop. Regardless of the reason for including it anyway, should we also have it here as well?

src/vmm/src/persist.rs
```diff
@@ -265,7 +270,8 @@ fn snapshot_memory_to_file(
                 .map_err(Memory)
         }
         SnapshotType::Full => vmm.guest_memory().dump(&mut file).map_err(Memory),
-    }
+    }?;
+    file.sync_all().map_err(MemoryFileFlush)
```
georgepisaltu, line 274: Same as before with the lack of `file.flush()`.

tests/integration_tests/performance/test_snapshot_perf.py
```diff
@@ -30,21 +30,21 @@
     'x86_64': {
         '2vcpu_256mb.json': {
             'FULL':  180,
-            'DIFF':  50,
```
georgepisaltu, line 33: What happened here (as well as the other benchmarks)? Does the ramdisk impact the performance tests this severely?

tests/framework/builder.py
```diff
@@ -55,17 +55,21 @@ def build(self,
               config: Artifact,
               net_ifaces=None,
               enable_diff_snapshots=False,
-              cpu_template=None):
+              cpu_template=None,
+              use_ramdisk=False):
         """Build a fresh microvm."""
         vm = init_microvm(self.root_path, self.bin_cloner_path,
```
alindima, line 61: To avoid having to create a ramdisk for every single jailer instance (and since most tests don't need it), I think we can pass the `use_ramdisk` parameter down to the `init_microvm` function and then to the `Microvm` instance as a constructor parameter.

tests/framework/builder.py
```diff
@@ -202,24 +220,44 @@ def create(self,
...
+               use_ramdisk=False):
         """Create a Snapshot object from a microvm and artifacts."""
+        if use_ramdisk:
+            snaps_dir = self._microvm.jailer.chroot_ramfs_path()
+            mem_full_path = os.path.join(snaps_dir, mem_file_name)
+            vmstate_full_path = os.path.join(snaps_dir, snapshot_name)
+
+            # Pre-allocate ram for memfile to eliminate allocation variability.
+            utils.run_cmd(
+                'dd if=/dev/zero of={} bs=1M count=512'.format(mem_full_path)
```
alindima, line 233: Why is the memfile size hardcoded here? It could change in the future

tests/framework/builder.py
```diff
@@ -202,24 +220,44 @@ def create(self,
...
         # Disable API timeout as the APIs for snapshot related procedures
         # take longer.
         self._microvm.api_session.untime()
-        snapshot_dir = self.create_snapshot_dir()
+
+        snaps_dir_name = os.path.basename(snaps_dir)
         self._microvm.pause_to_snapshot(
-            mem_file_path="/snapshot/"+mem_file_name,
-            snapshot_path="/snapshot/"+snapshot_name,
+            mem_file_path="/"+snaps_dir_name+"/"+mem_file_name,
```
alindima, line 252: nit: use `os.path` methods for concatenating

tests/framework/builder.py
```diff
@@ -202,24 +220,44 @@ def create(self,
...
+        if use_ramdisk:
+            snaps_dir = self._microvm.jailer.chroot_ramfs_path()
+            mem_full_path = os.path.join(snaps_dir, mem_file_name)
+            vmstate_full_path = os.path.join(snaps_dir, snapshot_name)
+
+            # Pre-allocate ram for memfile to eliminate allocation variability.
+            utils.run_cmd(
+                'dd if=/dev/zero of={} bs=1M count=512'.format(mem_full_path)
+            )
+            cmd = 'chown {}:{} {}'.format(
```
alindima, line 235: nit: `chown` calls identical to this one are sprinkled in multiple places over the test framework. We could have a method on the `Jailer` class for this called `chown_path(full_path)`

tests/integration_tests/functional/test_drives.py
```diff
@@ -449,6 +449,10 @@ def test_block_default_cache_old_version(test_microvm_with_ssh):
     response = test_microvm.vm.patch(state='Paused')
     assert test_microvm.api_session.is_status_no_content(response.status_code)
 
+    # Disable API timeout as the APIs for snapshot related procedures
+    # take longer.
+    test_microvm.api_session.untime()
```
alindima, line 454: The `snapshot.create()` call that follows already calls `untime()`, so this is redunant.
If we also add it to the `pause_to_snapshot` method, we can get rid of all the new `untime()` calls in the tests.

src/devices/src/virtio/block/persist.rs
```diff
@@ -89,6 +89,14 @@ impl Persist<'_> for Block {
     type Error = io::Error;
 
     fn save(&self) -> Self::State {
+        if self.disk.file().flush().is_err() {
+            error!("Failed to flush block data on serialization.");
```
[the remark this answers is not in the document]
georgepisaltu: nittier: explore doing the `error!("...")` in a `map_err` :grin:

src/vmm/src/persist.rs
```diff
@@ -618,6 +627,9 @@ mod tests {
         let err = SnapshotBackingFile(io::Error::from_raw_os_error(0));
         let _ = format!("{}{:?}", err, err);
 
+        let err = SnapshotFileFlush(io::Error::from_raw_os_error(0));
+        let _ = format!("{}{:?}", err, err);
```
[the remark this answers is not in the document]
georgepisaltu: This is standard stuff for not messing up the coverage.

tests/framework/builder.py
georgepisaltu: If this is just the preallocation, then the ramdisk can grow more than this, right?

alindima: I'm referriing to the size of the memfile, which here is allocated to 512M.
There are also microvms in the snapshot perf test that have a memory of 256M, and in the future we may add VMs with more memory maybe?

I'm saying that this should be configurable

tests/integration_tests/performance/test_snapshot_perf.py
acatangiu: Yes, well, not the ramdisk, but the flushing operation on all disks and all dirtied memory. To get an idea of the cost, without ramdisk the operation takes 2300ms, and under 50ms with ramdisk.

The flush still implies FS operations which add a bit of latency. I am not at all concerned about this value.

src/vmm/src/persist.rs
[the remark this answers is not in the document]
alindima: yeah, I was used to tests that were at least asserting the formatted output, but it doesn't really matter that much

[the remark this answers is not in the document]
acatangiu: not just coverage, it actually (compile-time) tests that the error implements `Display` and `Debug`.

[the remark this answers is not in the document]
alindima: That's also tested by the compiler, when/if the error gets printed and formatted (outside of the test)

tests/integration_tests/functional/test_drives.py
acatangiu: done

tests/framework/builder.py
alindima: sounds good to me

alindima (approved):
🚀

georgepisaltu (approved):

PR #2557, opened 2021-05-12
Merge feature/file_based_seccomp into main branch

alindima:
# Reason for This PR

Merge feature/file_based_seccomp into main branch

## Description of Changes

`[Author TODO: add description of changes.]`

- [ ] This functionality can be added in [`rust-vmm`](https://github.com/rust-vmm).

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria.]`
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [x] All commits in this PR are signed (`git commit -s`).
- [x] The reason for this PR is clearly provided (issue no. or explanation).
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] Any newly added `unsafe` code is properly documented.
- [x] Any API changes are reflected in `firecracker/swagger.yaml`.
- [x] Any user-facing changes are mentioned in `CHANGELOG.md`.
- [x] All added/changed functionality is tested.

PR #2563, opened 2021-05-14
make InstanceStart idempotent for seccomp filters

alindima:
# Reason for This PR

See comment: https://github.com/firecracker-microvm/firecracker/pull/2557#discussion_r631847141

Previously, the InstanceStart request, was using a mutable
reference of the seccomp filter collection, removing the
entries for each thread category, after they were launched.

If a recoverable error occured, that did not bring down
the process, because the BPF map was mutated, a subsequent
InstanceStart would fail because the filters would have
already been freed.

## Description of Changes

This commit fixes this, by changing the mutable reference
on the PrebootAPIController to be a regular reference,
kept valid for the entire lifetime of the preboot controller
(until the instance is started).

The vmm and vcpu filters are no longer removed from the
filter map, until the entire map is freed.

This way, a subsequent call would not fail.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -114,7 +114,7 @@ impl Display for StartMicrovmError {
                 "Cannot load initrd due to an invalid memory configuration."
             ),
             InitrdRead(err) => write!(f, "Cannot load initrd due to an invalid image: {}", err),
-            Internal(err) => write!(f, "Internal error while starting microVM: {:?}", err),
+            Internal(err) => write!(f, "Internal error while starting microVM: {}", err),
             KernelCmdline(err) => write!(f, "Invalid kernel command line: {}", err),
             KernelLoader(err) => {
                 let mut err_msg = format!("{}", err);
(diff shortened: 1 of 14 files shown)


src/vmm/src/builder.rs
```diff
@@ -365,9 +365,12 @@ pub fn build_microvm_for_boot(
...
     vmm.start_vcpus(
         vcpus,
-        seccomp_filters
-            .remove("vcpu")
-            .ok_or_else(|| MissingSeccompFilters("vcpu".to_string()))?,
+        Arc::new(
+            seccomp_filters
+                .get("vcpu")
+                .ok_or_else(|| MissingSeccompFilters("vcpu".to_string()))?
+                .clone(),
```
acatangiu, line 372: You could avoid having to `clone()` the entire _vcpu_ filter vec by changing `BpfThreadMap` definition:

```diff
- pub type BpfThreadMap = HashMap<String, BpfProgram>;
+ pub type BpfThreadMap = HashMap<String, Arc<BpfProgram>>;
```

That way any entry in the map can be easily shared among multiple users/consumers with no cloning necessary.

acatangiu (approved):

src/vmm/src/builder.rs
alindima: Indeed 👍🏻 . I'll add a commit that does this modification

alindima: Done, PTAL

acatangiu (approved):
👍

acatangiu (approved):

georgepisaltu (approved):

PR #2564, opened 2021-05-14
Use the evset passed as a parameter instead of the one from connection

### LINKED ISSUE #2539: [Bug] vsock event handling in muxer
Posting analysis by @alindima:

When forwarding an epoll event from the unix muxer to the targeted connection event handler, the eventset the connection registered is forwarded instead of the actual epoll operation (IN/OUT).

For example, if the connection was registered for EPOLLIN, and receives an EPOLLOUT, the connection will actually handle an EPOLLOUT. This is the root cause of previous bug, which caused the introduction of some workarounds (i.e: handling ewouldblock when reading after receiving EPOLLIN, which should never happen): https://github.com/firecracker-microvm/firecracker/issues/1572

Another example would be for receiving EPOLLHUP, an event for which all epoll fds are automatically registered and handling it as an EPOLLIN. EPOLLHUP may be received when the UDS was closed, and the connection should be terminated. This is however not a problem since reading/writing will error and the connection will still be terminated.
Here is the code problem: https://github.com/firecracker-microvm/firecracker/blob/ffbc2efcf2dcffa02db6df3474a8ad2b0bad21f7/src/devices/src/virtio/vsock/unix/muxer.rs#L344

When matching the connection, we retrieve and use the evset of the connection instead of the one passed as a parameter. The compiler does not complain for an unused variable because it is first logged, in the debug! statement. This is an unfortunate naming mistake that caused a lot of problems. The known side effects of this issue have been worked around in this PR: https://github.com/firecracker-microvm/firecracker/pull/1910.

This should be fixed to account for error cases that we may not have thought about, and to get rid of some fails metric increments. It could also bring a tiny performance improvement, since we wouldn't be handling some nonexistent epoll notifications.

halfzebra:
# Reason for This PR

Aims to close #2539

## Description of Changes

Renamed name-clashing variables to fix the bug and avoid the confusion in the future.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/vsock/unix/muxer.rs
+++ b/src/devices/src/virtio/vsock/unix/muxer.rs
@@ -273,7 +273,7 @@ impl VsockEpollListener for VsockMuxer {
     /// Get the epoll events to be polled upstream.
     ///
     /// Since the polled FD is a nested epoll FD, we're only interested in EPOLLIN events (i.e.
-    /// some event occured on one of the FDs registered under our epoll FD).
+    /// some event occurred on one of the FDs registered under our epoll FD).
     fn get_polled_evset(&self) -> EventSet {
         EventSet::IN
     }
@@ -327,30 +327,29 @@ impl VsockMuxer {
             local_port_set: HashSet::with_capacity(defs::MAX_CONNECTIONS),
         };
 
-        // Listen on the host initiated socket, for incomming connections.
+        // Listen on the host initiated socket, for incoming connections.
         muxer.add_listener(muxer.host_sock.as_raw_fd(), EpollListener::HostSock)?;
         Ok(muxer)
     }
 
     /// Handle/dispatch an epoll event to its listener.
-    fn handle_event(&mut self, fd: RawFd, evset: EventSet) {
+    fn handle_event(&mut self, fd: RawFd, event_set: EventSet) {
         debug!(
             "vsock: muxer processing event: fd={}, evset={:?}",
-            fd, evset
+            fd, event_set
         );
 
         match self.listener_map.get_mut(&fd) {
             // This event needs to be forwarded to a `MuxerConnection` that is listening for
             // it.
-            Some(EpollListener::Connection { key, evset }) => {
+            Some(EpollListener::Connection { key, evset: _ }) => {
                 let key_copy = *key;
-                let evset_copy = *evset;
                 // The handling of this event will most probably mutate the state of the
-                // receiving conection. We'll need to check for new pending RX, event set
+                // receiving connection. We'll need to check for new pending RX, event set
                 // mutation, and all that, so we're wrapping the event delivery inside those
                 // checks.
                 self.apply_conn_mutation(key_copy, |conn| {
-                    conn.notify(evset_copy);
+                    conn.notify(event_set);
                 });
             }
 
@@ -412,7 +411,10 @@ impl VsockMuxer {
             }
 
             _ => {
-                info!("vsock: unexpected event: fd={:?}, evset={:?}", fd, evset);
+                info!(
+                    "vsock: unexpected event: fd={:?}, evset={:?}",
+                    fd, event_set
+                );
                 METRICS.vsock.muxer_event_fails.inc();
             }
         }
--- a/src/devices/src/virtio/vsock/unix/muxer.rs
+++ b/src/devices/src/virtio/vsock/unix/muxer.rs
@@ -978,6 +978,24 @@ mod tests {
     }
 
     #[test]
+    fn test_muxer_epoll_listener_regression() {
+        let mut ctx = MuxerTestContext::new("muxer_epoll_listener");
+        ctx.local_connect(1025);
+
+        let (_, conn) = ctx.muxer.conn_map.iter().next().unwrap();
+
+        assert_eq!(conn.get_polled_evset(), EventSet::IN);
+
+        assert_eq!(METRICS.vsock.conn_event_fails.count(), 0);
+
+        let conn_eventfd = conn.as_raw_fd();
+
+        (&mut ctx.muxer).handle_event(conn_eventfd, EventSet::OUT);
+
+        assert_eq!(METRICS.vsock.conn_event_fails.count(), 1);
+    }
+
+    #[test]
     fn test_bad_peer_pkt() {
         const LOCAL_PORT: u32 = 1026;
         const PEER_PORT: u32 = 1025;


src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -326,30 +326,29 @@ impl VsockMuxer {
...
 
-        // Listen on the host initiated socket, for incomming connections.
+        // Listen on the host initiated socket, for incoming connections.
         muxer.add_listener(muxer.host_sock.as_raw_fd(), EpollListener::HostSock)?;
         Ok(muxer)
     }
 
     /// Handle/dispatch an epoll event to its listener.
-    fn handle_event(&mut self, fd: RawFd, evset: EventSet) {
+    fn handle_event(&mut self, fd: RawFd, handler_evset: EventSet) {
```
georgepisaltu, line 335: This is very subjective, but I don't like the naming `handler_evset`. Unfortunately, I couldn't think of a better one than `event_set`, which is basically just the longer version.

The problem is that it's not the event of the handler (`handler_evset`), it's the event set that the calling functions want to handle, so it would be something like `handled_evset`, which is weird and generally would be shortened to just `evset`.

I propose you name it `event_set` in order to avoid the shadowing of the other variable and just be done with it.

georgepisaltu (commented):
Please make sure your commits respect the 50/72 rule. Currently, your commit title length is over the maximum 50 character limit.

Also, run `cargo fmt --all` before making your commits. This is just the standard rust code format.

Also also, please update the coverage on `aarch64`.

src/devices/src/virtio/vsock/unix/muxer.rs
halfzebra: Thanks for the feedback, I'll address this asap! :+1:

alindima:
LGTM. Once you fix the styling issues and update the ARM coverage, we should be good 👍🏻 (don't worry about the x86_64 failure, it is an unrelated, intermittent one)

georgepisaltu (commented):
Everything looks good, only thing left to do is to update the code coverage on `aarch64`.

You can do this by opening the `tests/integration_tests/build/test_coverage.py` file and modifying the following line:
```diff
- COVERAGE_DICT = {"Intel": 84.85, "AMD": 84.15, "ARM": 83.13}
+ COVERAGE_DICT = {"Intel": 84.85, "AMD": 84.15, "ARM": 83.02}
```

andreeaflorescu:
Sorry to interject right at the end. This was a very interesting bug and it would be nice to be able to identify it when doing `git log`. Since you need to update the coverage, can you also add some more details in the commit message about what you are fixing and why?

alindima:
> Sorry to interject right at the end. This was a very interesting bug and it would be nice to be able to identify it when doing `git log`. Since you need to update the coverage, can you also add some more details in the commit message about what you are fixing and why?

Very good point. @halfzebra you can start by adapting some of the text from the issue: https://github.com/firecracker-microvm/firecracker/issues/2539

andreeaflorescu:
Since this PR is fixing a bug, we typically add a regression test as well.
The regression test is suppose to fail before the fix is added, and pass afterwards. @halfzebra Do you think you could also add a regression test for this fix?

alindima:
@halfzebra do you need any help with addressing the remaining comments?

halfzebra:
@alindima thanks for offering help!

Would having a Rust unit test in the same file be enough?
I've figured that I need to `rustup target add x86_64-unknown-linux-musl` to run the Rust tests, is this something that is worth mentioning somewhere in the docs for contributors?

alindima:
> Would having a Rust unit test in the same file be enough?
yes, this sounds good!

> I've figured that I need to `rustup target add x86_64-unknown-linux-musl` to run the Rust tests, is this something that is worth mentioning somewhere in the docs for contributors?

The recommended way of building Firecracker and running the unit tests is using `devtool`, which satisfies all the requirements and dependencies. See: https://github.com/firecracker-microvm/firecracker/blob/main/docs/getting-started.md#running-the-integration-test-suite.

You can run the unit tests like so: `tools/devtool test -- integration_tests/build/test_unittests.py`

If you think this extra info would help, you could open a PR on the getting started doc.

acatangiu (dismissed):
I've looked into adding a regression unit test for this and it's complicated because it'd require:
- configuring a `Muxer`
- constructing and sending a certain mix of vsock packets to open connections _over the muxer_ and get the muxer in a state where it is waiting for both `IN` and `OUT` events
- externally/artificially filling the consumer buffer so OUT is no longer generated/available
- then verifying that pushing a packet from host side (which generates IN) doesn't end up also trying to flush out data from connection (which can be verified through inspecting connection internal state).

This is a complex test that can't be done with current mock objects and/or test-context-objects.

I personally don't think investing in overhauling the vsock unit-test infrastructure (which is the best one of all our devices right now anyway - I'd rather invest in bringing the net/block/balloon test helpers to vsock level of flexibility) is worth doing for this corner-case regression test.

Also the actual fix is pending merger for more than a month and is currently blocked by this. I'm approving PR to merge the fix without a regression test.

alindima:
@acatangiu I actually managed to write a regression test for this corner case a while back. I still have the code and I can contribute it if @halfzebra thinks it's ok to force-push on his branch.

acatangiu:
@alindima feel free to add commits to the branch if doesn't have write-protection enabled. Do however try not to modify OP's original commits unless necessary.

alindima:
I added a regression test and changed the initial commit message to be more explicit.
@acatangiu PTAL

halfzebra:
@alindima thanks for helping out with the test! 👍

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -975,6 +977,24 @@ mod tests {
...
+
+        assert_eq!(conn.get_polled_evset(), EventSet::IN);
+
+        assert_eq!(METRICS.vsock.conn_event_fails.count(), 0);
+
+        let conn_eventfd = conn.as_raw_fd();
+
+        (&mut ctx.muxer).handle_event(conn_eventfd, EventSet::OUT);
+
+        assert_eq!(METRICS.vsock.conn_event_fails.count(), 1);
```
acatangiu, line 995: Oh, such a trickster test 😈! You're giving it bad input and testing an error branch. Seems legit to me! It punctually tests that the `EventSet` parameter is used instead of interest list. Nice job!

acatangiu (approved):

src/devices/src/virtio/vsock/unix/muxer.rs
```diff
@@ -977,6 +977,24 @@ mod tests {
...
```
georgepisaltu, line 995: This is a very cool way of testing this regression, but it works because `handle_event` calls `conn.notify()`, which will increment this particular metric when the event set is `EventSet::OUT` and the buffer is empty, which you set up to be. This being said, could we add a comment explaining this? I think it will be useful if we change anything about this in the future. This is kind of a nit, but I'll leave this here nonetheless.

georgepisaltu (approved):

PR #2566, opened 2021-05-14
Fix GetInstanceInfo when config-file is used

acatangiu:
This PR is based on https://github.com/firecracker-microvm/firecracker/pull/2489 and some of the commits herein are co-authored by `Franck Séhédic <franck.sehedic@ledger.fr>`

# Reason for This PR

When configured through `--config-file` the ApiServer improperly reports the VmState as "Not started" even when it actually is "Running".
(Note: This breaks integration with kata-runtime which does use `--config-file` and expects a "Running" VmState to be reported by the ApiServer).

## Description of Changes

`InstanceInfo` contains Vmm-instance-related information. As such, it makes more sense to be tied to a `Vmm` instance.
Embedded `InstanceInfo` in `struct Vmm` and exposed it through RPC all the way to HTTP API.

All Vmm state changes are naturally handled by Vmm internal logic.
API server no longer tracks Vmm state based on HTTP API calls responses, it now simply delegates all InstanceInfo-related requests to the Vmm RPC interface.

Because state tracking is done internally by the Vmm, `--config-file` use-case now also reports good state.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -15,9 +15,7 @@ use std::sync::{mpsc, Arc, Mutex};
 use std::{fmt, io};
 
 use crate::parsed_request::ParsedRequest;
-use logger::{
-    debug, error, info, update_metric_with_elapsed_time, IncMetric, ProcessTimeReporter, METRICS,
-};
+use logger::{debug, error, info, update_metric_with_elapsed_time, ProcessTimeReporter, METRICS};
 pub use micro_http::{
     Body, HttpServer, Method, Request, RequestError, Response, ServerError, ServerRequest,
     ServerResponse, StatusCode, Version,
@@ -27,7 +25,6 @@ use mmds::data_store::Mmds;
 use seccompiler::BpfProgramRef;
 use utils::eventfd::EventFd;
 use vmm::rpc_interface::{VmmAction, VmmActionError, VmmData};
-use vmm::vmm_config::instance_info::{InstanceInfo, VmState};
 use vmm::vmm_config::snapshot::SnapshotType;
 
 use vmm::FC_EXIT_CODE_BAD_CONFIGURATION;
@@ -69,8 +66,6 @@ type Result<T> = std::result::Result<T, Error>;
 pub struct ApiServer {
     /// MMDS info directly accessible from the API thread.
     mmds_info: Arc<Mutex<Mmds>>,
-    /// Firecracker instance info exposed through API.
-    instance_info: InstanceInfo,
     /// Sender which allows passing messages to the VMM.
     api_request_sender: mpsc::Sender<ApiRequest>,
     /// Receiver which collects messages from the VMM.
@@ -89,14 +84,12 @@ impl ApiServer {
     /// Returns the newly formed `ApiServer`.
     pub fn new(
         mmds_info: Arc<Mutex<Mmds>>,
-        instance_info: InstanceInfo,
         api_request_sender: mpsc::Sender<ApiRequest>,
         vmm_response_receiver: mpsc::Receiver<ApiResponse>,
         to_vmm_fd: EventFd,
     ) -> Self {
         ApiServer {
             mmds_info,
-            instance_info,
             api_request_sender,
             vmm_response_receiver,
             to_vmm_fd,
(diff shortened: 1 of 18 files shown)


src/vmm/src/lib.rs
```diff
@@ -307,13 +310,17 @@ impl Vmm {
     pub fn resume_vm(&mut self) -> Result<()> {
         self.mmio_device_manager.kick_devices();
         self.broadcast_vcpu_event(VcpuEvent::Resume, VcpuResponse::Resumed)
-            .map_err(|_| Error::VcpuResume)
+            .map_err(|_| Error::VcpuResume)?;
+        self.instance_info.state = VmState::Running;
```
alindima, line 319: this commit uses the `VmState`, which is introduced in the next commit, which makes the PR harder to review, and also uses both `state` types (enum and string)
I suggest we decouple the commits

src/api_server/src/lib.rs
```diff
@@ -707,11 +655,11 @@ mod tests {
             .unwrap();
 
         // Wait for the server to set itself up.
-        thread::sleep(Duration::new(0, 10_000_000));
+        thread::sleep(Duration::from_millis(10));
         let mut sock = UnixStream::connect(PathBuf::from(path_to_socket)).unwrap();
 
         // Send a GET instance-info request.
-        assert!(sock.write_all(b"GET / HTTP/1.1\r\n\r\n").is_ok());
+        assert!(sock.write_all(b"GET /mmds HTTP/1.1\r\n\r\n").is_ok());
```
alindima, line 662: is this a leftover?

src/logger/src/metrics.rs
```diff
@@ -349,16 +349,10 @@ pub struct ApiServerMetrics {
 pub struct GetRequestsMetrics {
     /// Number of GETs for getting information on the instance.
     pub instance_info_count: SharedIncMetric,
-    /// Number of failures when obtaining information on the current instance.
-    pub instance_info_fails: SharedIncMetric,
```
alindima, line 353: This one is actually used, here: https://github.com/firecracker-microvm/firecracker/blob/67a83f4ee3b3260c3fe2dcf29ff1c71d589c05da/src/api_server/src/lib.rs#L332
Is this a mistake or do you think we should remove it nevertheless?

gc-plp:
I think that we're missing the logic that checks the VM state in our integration tests, otherwise we would have caught this. Can we also modify the inner working of `microvm.py` to check the state?

src/vmm/src/rpc_interface.rs
```diff
@@ -468,6 +473,15 @@ impl RuntimeApiController {
             GetVmConfiguration => Ok(VmmData::MachineConfiguration(
                 self.vm_resources.vm_config().clone(),
             )),
+            GetVmInstanceInfo => Ok(VmmData::InstanceInformation(
+                // Clone is redundant only for test MockVmm.
+                #[allow(clippy::redundant_clone)]
```
georgepisaltu, line 478: Woldn't it make sense to have `Vmm::instance_info` return an `InstanceInfo` instead of a reference and have the clone inside that function?

src/api_server/src/lib.rs
acatangiu: no, this is on purpose. This test verifies `bind_and_run` functionality so it uses the simplest API endpoint - which is now mmds.

URL `/` is served by Vmm, so to continue using that in this test, we would have to also simulate that part of the communication which is out of scope for this test.

src/logger/src/metrics.rs
acatangiu: It was used before, this PR removes that function. With current code it cannot fail, therefore metric is useless.

src/vmm/src/lib.rs
acatangiu: yes, this was a commit fixup miss. will fix it

src/vmm/src/rpc_interface.rs
acatangiu: sure, that would work! I tried to adhere to the general practice of having a reference getter, but indeed we could just pass a copy and be done with it.

acatangiu:
> I think that we're missing the logic that checks the VM state in our integration tests, otherwise we would have caught this. Can we also modify the inner workings of `microvm.py` to check the state?

We do correctly check state in inner workings of `microvm.py`. But we are indeed missing validation/regression test on `--config-file` path.

I will add one.

src/api_server/src/lib.rs
alindima: makes sense, let's update the comment then

src/vmm/src/rpc_interface.rs
georgepisaltu: We took the same approach (passing a copy) with both the `GetBalloonStats` and `GetBalloonConfig` requests. The structs were small in size, so a `clone` didn't seem like a big deal at the time. I think it applies to this context as well.

acatangiu:
@alindima @gc-plp @gpl I've addressed all comments, PTAL.

alindima (approved):

georgepisaltu (approved):

gc-plp (approved):

PR #2576, opened 2021-05-21
Fix SSBD mitigation for aarch64

georgepisaltu:
# Reason for This PR

In our guide, we recommend enabling the SSBD mitiagtion through the kernel command line with `spec_store_bypass_disable=seccomp`. However, this only works for x86_64.

## Description of Changes

Added the fix in the jailer through the `prctl` interface. Updated the guide with the addition of the mitigation in the jailer through `prctl` and the alternative through the `ssbd=force-on` kernel command line parameter.

## License Acceptance

## PR Checklist

### DIFF
--- a/docs/prod-host-setup.md
+++ b/docs/prod-host-setup.md
@@ -220,13 +220,23 @@ See more details [here](https://www.kernel.org/doc/html/latest/admin-guide/hw-vu
 This will mitigate variants of Spectre side-channel issues such as
 Speculative Store Bypass and SpectreNG.
 
-It can be enabled by adding the following Linux kernel boot parameter:
+On x86_64 systems, it can be enabled by adding the following Linux kernel boot
+parameter:
 
 ```console
 spec_store_bypass_disable=seccomp
 ```
 
 which will apply SSB if seccomp is enabled by Firecracker.
+
+On aarch64 systems, it is enabled by Firecracker
+[using the `prctl` interface][3]. However, this is only availabe on host
+kernels Linux >=4.17 and also Amazon Linux 4.14. Alternatively, a global
+mitigation can be enabled by adding the following Linux kernel boot parameter:
+
+```console
+ssbd=force-on
+```
 
 Verification can be done by running:
 
@@ -329,3 +339,4 @@ to trap and control this in the hypervisor.
 
 [1]: https://elixir.free-electrons.com/linux/v4.14.203/source/virt/kvm/arm/hyp/timer-sr.c#L63
 [2]: https://lists.cs.columbia.edu/pipermail/kvmarm/2017-January/023323.html
+[3]: https://elixir.bootlin.com/linux/v4.17/source/include/uapi/linux/prctl.h#L212
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@
   to use "MiB" instead.
 - Snapshot related host files (vm-state, memory, block backing files) are now
   flushed to their backing mediums as part of the CreateSnapshot operation.
+- Fixed the SSBD mitigation not being enabled on `aarch64` with the provided
+  `prod-host-setup.md`.
 
 ## [0.24.0]
 
--- a/src/firecracker/src/main.rs
+++ b/src/firecracker/src/main.rs
@@ -31,10 +31,45 @@ const DEFAULT_API_SOCK_PATH: &str = "/run/firecracker.socket";
 const DEFAULT_INSTANCE_ID: &str = "anonymous-instance";
 const FIRECRACKER_VERSION: &str = env!("FIRECRACKER_VERSION");
 
+#[cfg(target_arch = "aarch64")]
+/// Enable SSBD mitigation through `prctl`.
+pub fn enable_ssbd_mitigation() {
+    // Parameters for `prctl`
+    // TODO: generate bindings for these from the kernel sources.
+    // https://elixir.bootlin.com/linux/v4.17/source/include/uapi/linux/prctl.h#L212
+    const PR_SET_SPECULATION_CTRL: i32 = 53;
+    const PR_SPEC_STORE_BYPASS: u64 = 0;
+    const PR_SPEC_FORCE_DISABLE: u64 = 1u64 << 3;
+
+    let ret = unsafe {
+        libc::prctl(
+            PR_SET_SPECULATION_CTRL,
+            PR_SPEC_STORE_BYPASS,
+            PR_SPEC_FORCE_DISABLE,
+            0,
+            0,
+        )
+    };
+
+    if ret < 0 {
+        let last_error = std::io::Error::last_os_error().raw_os_error().unwrap();
+        error!(
+            "Could not enable SSBD mitigation through prctl, error {}",
+            last_error
+        );
+        if last_error == libc::EINVAL {
+            error!("The host does not support SSBD mitigation through prctl.");
+        }
+    }
+}
+
 fn main() {
     LOGGER
         .configure(Some(DEFAULT_INSTANCE_ID.to_string()))
         .expect("Failed to register logger");
+
+    #[cfg(target_arch = "aarch64")]
+    enable_ssbd_mitigation();
 
     // We need this so that we can reset terminal to canonical mode if panic occurs.
     let stdin = io::stdin();
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -24,7 +24,7 @@ import host_tools.proc as proc
 # this contains the frequency while on AMD it does not.
 # Checkout the cpuid crate. In the future other
 # differences may appear.
-COVERAGE_DICT = {"Intel": 85.36, "AMD": 84.79, "ARM": 83.62}
+COVERAGE_DICT = {"Intel": 85.36, "AMD": 84.79, "ARM": 83.52}
 
 PROC_MODEL = proc.proc_type()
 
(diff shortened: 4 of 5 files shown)


src/jailer/src/main.rs
```diff
@@ -328,6 +328,33 @@ fn sanitize_process() {
     }
 }
 
+#[cfg(target_arch = "aarch64")]
+// Safe because this syscall takes static parameters and
+// we check the output.
+unsafe fn enable_ssbd_mitigation() {
```
alindima, line 334: I'd rather have the function safe, and only the `libc::prctl` wrapped in an `unsafe` block

src/jailer/src/main.rs
```diff
@@ -328,6 +328,33 @@ fn sanitize_process() {
...
+    let ret = libc::prctl(
+        PR_SET_SPECULATION_CTRL,
+        PR_SPEC_STORE_BYPASS,
+        PR_SPEC_FORCE_DISABLE,
+        0,
+        0,
+    );
+
+    if ret < 0 {
+        println!("Could not enable SSBD mitiagtion through the prctl interface.");
```
alindima, line 351: I'd print the last os error here and in the log line below. It can be useful for debugging

alindima (commented):
Can we add an integration test?

alindima (commented):
Can we add an integration test?

alindima:
Why are we doing this in the jailer? I think we can do this in the Firecracker binary, before we run any customer code and before the seccomp filters are enabled. This would be to account for any customers that don't run with the jailer.

Unless there's some other good reason not to.

src/jailer/src/main.rs
georgepisaltu: Yes, the `unsafe fn` is a remnant of an earlier implementation and I forgot to remove it, will do.

src/jailer/src/main.rs
georgepisaltu: Looking at the [`prctl` man page](https://man7.org/linux/man-pages/man2/prctl.2.html) and the[ kernel documentation](https://www.kernel.org/doc/html/latest/userspace-api/spec_ctrl.html), the only fail we are interested in is `EINVAL`, which means the architecture/host kernel do not support this mitigation.

I can add the error code after the message, but I don't personally see the added value.

georgepisaltu (commented):
> Can we add an integration test?

I will give it a go in the next iteration.

> Why are we doing this in the jailer?

My understanding is that the seccomp filter is applied in Firecracker and not the jailer because Firecracker does some setup which would be hindered by the filters, otherwise it would be in the jailer. As this has nothing to do with the actual Firecracker process and is just an external control mechanism, an extra safety feature, I believe it belongs in the jailer. Customers are advised to run with the jailer anyway and it would be good to decouple Firecracker and the jailer as much as possible.

src/jailer/src/main.rs
sandreim: Them we can make the error message sound like "SSBD mitigation not available on this architecture/kernel".

georgepisaltu: The problem is that SSBD mitigation might be available on that architecture/kernel, just not through the `prctl` interface but through the kernel command line parameter `ssbd=force-on`. The `prctl` interface supports SSBD mitigation since Linux 4.17 and it has also been ported to Amazon Linux 4.14 variants.

alindima:
> > Why are we doing this in the jailer?
> 
> My understanding is that the seccomp filter is applied in Firecracker and not the jailer because Firecracker does some setup which would be hindered by the filters, otherwise it would be in the jailer. As this has nothing to do with the actual Firecracker process and is just an external control mechanism, an extra safety feature, I believe it belongs in the jailer. Customers are advised to run with the jailer anyway and it would be good to decouple Firecracker and the jailer as much as possible.

The way I see it, on x86, seccomp enables SSBD, and since seccomp is enabled by the Firecracker process, I would expect that this would work the same way on arm. But since it doesn't, I think that for feature parity reasons, it'd be best if we call this `prctl` in the Firecracker code.

Let's see what others think: @firecracker-microvm/compute-capsule

src/jailer/src/main.rs
alindima: I think we can also get a `ENXIO` - Control of the selected speculation misfeature is not possible, not sure.

I'd really like to default to also printing the OS error. You never know when it might come handy when looking through the logs

src/jailer/src/main.rs
georgepisaltu: Removed.

src/jailer/src/main.rs
georgepisaltu: `ENXIO` is not possible because we only run this on `aarch64`, and the setting mentioned in the man page which could result in `ENXIO` is `x86` only.

Regardless, I appended the error code to the message.

src/jailer/src/main.rs
```diff
@@ -344,6 +377,12 @@ pub fn to_cstring<T: AsRef<Path>>(path: T) -> Result<CString> {
 fn main() {
     sanitize_process();
 
+    #[cfg(target_arch = "aarch64")]
+    // Safe because there are no variables that can crash this.
+    unsafe {
```
acatangiu, line 382: nit: `unsafe` no longer necessary

acatangiu (approved):
LGTM

sandreim:
> > > Why are we doing this in the jailer?
> > 
> > 
> > My understanding is that the seccomp filter is applied in Firecracker and not the jailer because Firecracker does some setup which would be hindered by the filters, otherwise it would be in the jailer. As this has nothing to do with the actual Firecracker process and is just an external control mechanism, an extra safety feature, I believe it belongs in the jailer. Customers are advised to run with the jailer anyway and it would be good to decouple Firecracker and the jailer as much as possible.
> 
> The way I see it, on x86, seccomp enables SSBD, and since seccomp is enabled by the Firecracker process, I would expect that this would work the same way on arm. But since it doesn't, I think that for feature parity reasons, it'd be best if we call this `prctl` in the Firecracker code.
> 
> Let's see what others think: @firecracker-microvm/compute-capsule

I propose for the short term to implement the `prctl` in the Firecracker codebase for parity reasons and avoiding adding extra concerns in the jailer. Long term, we should look at switching the prod host setup to spec_store_bypass_disable=prctl to provide more flexibility if needed and possibly move the prctl call to jailer as the current PR proposes.

WDYT ?

alindima:
> > > > Why are we doing this in the jailer?
> > > 
> > > 
> > > My understanding is that the seccomp filter is applied in Firecracker and not the jailer because Firecracker does some setup which would be hindered by the filters, otherwise it would be in the jailer. As this has nothing to do with the actual Firecracker process and is just an external control mechanism, an extra safety feature, I believe it belongs in the jailer. Customers are advised to run with the jailer anyway and it would be good to decouple Firecracker and the jailer as much as possible.
> > 
> > 
> > The way I see it, on x86, seccomp enables SSBD, and since seccomp is enabled by the Firecracker process, I would expect that this would work the same way on arm. But since it doesn't, I think that for feature parity reasons, it'd be best if we call this `prctl` in the Firecracker code.
> > Let's see what others think: @firecracker-microvm/compute-capsule
> 
> I propose for the short term to implement the `prctl` in the Firecracker codebase for parity reasons and avoiding adding extra concerns in the jailer. Long term, we should look at switching the x86 prod host setup to `spec_store_bypass_disable=prctl` to provide more flexibility if needed and possibly move the `prctl` call to jailer as the current PR proposes.
> 
> WDYT ?

Sounds ok to me. Let's see first if we need to call the prctl on a per thread basis. Based on the man page, it seems so.
If we want to do this, doing this in the jailer will no longer be an option

sandreim (commented):
Needs changelog update as well, otherwise looks good to me.

src/utils/src/mitigations.rs
```diff
@@ -0,0 +1,34 @@
...
+            PR_SPEC_STORE_BYPASS,
+            PR_SPEC_FORCE_DISABLE,
+            0,
+            0,
+        )
+    };
+
+    if ret < 0 {
+        let last_error = std::io::Error::last_os_error().raw_os_error().unwrap();
+        println!(
```
alindima, line 26: Since this is called from the Firecracker process, let's use the logger interface, i.e. `error!`

docs/prod-host-setup.md
```diff
@@ -220,14 +220,24 @@ See more details [here](https://www.kernel.org/doc/html/latest/admin-guide/hw-vu
...
+On x86_64 systems, it can be enabled by adding the following Linux kernel boot
+parameter:
 
 ```console
 spec_store_bypass_disable=seccomp
 ```
 
 which will apply SSB if seccomp is enabled by Firecracker.
 
+On aarch64 systems, it is done in the jailer [using the `prctl` interface][3].
```
alindima, line 232:
```suggestion
On aarch64 systems, it is enabled by Firecracker [using the `prctl` interface][3].
```

alindima: Please also update the commit message, which still mentions the jailer

georgepisaltu: Done.

src/utils/src/mitigations.rs
georgepisaltu: I had to move the function from the `utils` crate to the `firecracker` crate in order to not cause a cyclic dependency while using `logger::error`. I just dropped the function in `firecracker/main.rs`. While this isn't the best place to have it, I think it's fine for now as we need to release the fix. I will open an issue to refactor this.

sandreim (approved):

alindima (approved):

sandreim (approved):

Issue #2587, opened 2021-05-26
[Bug] Seccomp failure are not logged in logs or metrics in fc versions > 0.24


acatangiu:
Any seccomp violation results in the termination of the Firecracker process without any logging done regarding the termination reason.

`seccomp::num_faults` metric is not increased either.

This results in no visibility and no post-incident debug-ability for seccomp violations.

This does not affect Firecracker versions `v0.24.x` and lower/older.

acatangiu:
The root-cause has been identified as being the lack of full support for `SIGSYS` in `signalfd` kernel implementation.

This is fixed in newer kernels: https://patchwork.ozlabs.org/project/linux-um/patch/20180428140702.3925-2-ebiederm@xmission.com/
But 4.14 for example doesn't have the fix: https://elixir.bootlin.com/linux/v4.14/source/fs/signalfd.c#L100

Looks like we need to use a synchronous signal handler for `SIGSYS`. We can keep the async signalfd implementation for the other signals.

alindima:
Fixed by #2593

PR #2588, opened 2021-05-26
Fix seccomp logging

acatangiu:
# Reason for This PR

Fixes https://github.com/firecracker-microvm/firecracker/issues/2587

Only newer kernels have full support for SIGSYS over asynchronous signalfd signal handling:
https://patchwork.ozlabs.org/project/linux-um/patch/20180428140702.3925-2-ebiederm@xmission.com/

For example, 4.14 doesn't: https://elixir.bootlin.com/linux/v4.14/source/fs/signalfd.c#L100

Because of this, on seccomp violations our SIGSYS async handler was never being called resulting in no visibility and no post-incident debug-ability for seccomp violations.

## Description of Changes

This PR introduces a classic dedicated sync signal handler for SIGSYS as well as a regression test.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/signal_handler.rs
+++ b/src/vmm/src/signal_handler.rs
@@ -1,8 +1,8 @@
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 use libc::{
-    _exit, c_int, signalfd_siginfo, sigset_t, SFD_CLOEXEC, SFD_NONBLOCK, SIGBUS, SIGHUP, SIGILL,
-    SIGPIPE, SIGSEGV, SIGSYS, SIGXCPU, SIGXFSZ, SIG_BLOCK,
+    _exit, c_int, c_void, siginfo_t, signalfd_siginfo, sigset_t, SFD_CLOEXEC, SFD_NONBLOCK, SIGBUS,
+    SIGHUP, SIGILL, SIGPIPE, SIGSEGV, SIGSYS, SIGXCPU, SIGXFSZ, SIG_BLOCK,
 };
 use logger::{error, IncMetric, METRICS};
 use polly::event_manager::{EventManager, Subscriber};
@@ -13,9 +13,16 @@ use std::os::unix::io::{AsRawFd, FromRawFd};
 use std::ptr;
 use utils::epoll::{EpollEvent, EventSet};
 use utils::errno::Error as ErrnoError;
-use utils::signal::{create_sigset, validate_signal_num};
+use utils::signal::{create_sigset, register_signal_handler, validate_signal_num};
 use utils::syscall::SyscallReturnCode;
 use vm_memory::ByteValued;
+
+// The offset of `si_syscall` (offending syscall identifier) within the siginfo structure
+// expressed as an `(u)int*`.
+// Offset `6` for an `i32` field means that the needed information is located at `6 * sizeof(i32)`.
+// See /usr/include/linux/signal.h for the C struct definition.
+// See https://github.com/rust-lang/libc/issues/716 for why the offset is different in Rust.
+const SI_OFF_SYSCALL: isize = 6;
 
 // We need this wrapper in order to be able to implement the `ByteValued` trait,
 // since the `signalfd_siginfo` type is foreign.
@@ -36,9 +43,9 @@ impl Default for SignalInfoWrapper {
 
 unsafe impl ByteValued for SignalInfoWrapper {}
 
-const HANDLED_SIGNALS: [c_int; 8] = [
-    SIGBUS, SIGHUP, SIGILL, SIGPIPE, SIGSEGV, SIGSYS, SIGXCPU, SIGXFSZ,
-];
+// SIGSYS is handled separately because it doesn't have proper signalfd support in older kernels.
+// No SIL_SYS entry in https://elixir.bootlin.com/linux/v4.14/source/fs/signalfd.c#L100
+const HANDLED_SIGNALS: [c_int; 7] = [SIGBUS, SIGHUP, SIGILL, SIGPIPE, SIGSEGV, SIGXCPU, SIGXFSZ];
 
 const SYS_SECCOMP_CODE: i32 = 1;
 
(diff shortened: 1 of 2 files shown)


src/vmm/src/signal_handler.rs
```diff
@@ -166,6 +200,16 @@ impl SignalManager {
...
         .map_err(Error::SyscallError)?;
 
+        // SIGSYS is handled separately because it doesn't have proper
+        // signalfd support in older kernels.
+        //
+        // Call to register_signal_handler is considered unsafe because it will register a signal
+        // handler which will be called in the current thread and will interrupt whatever work
+        // is done on the current thread, so we have to keep in mind that the registered
+        // signal handler must only do async-signal-safe operations.
+        register_signal_handler(SIGSYS, sigsys_handler)
```
alindima, line 210: The `register_signal_handler` call is out of scope for this function, which should only return the mask used for blocking other signals.

I suggest we call this `register_signal_handler` function explicitly in the `main.rs`, before spawning any threads, to make sure that they inherit the handler

src/vmm/src/signal_handler.rs
```diff
@@ -534,13 +575,6 @@ mod tests {
 
         let child = thread::spawn(move || {
             extern "C" fn signal_handler(_: c_int, siginfo: *mut siginfo_t, _: *mut c_void) {
```
alindima, line 577: There's also the `test_sigsys_handler` test that needs modifications. We no longer need an additional function in the test for registering a classic signal handler and all that `siginfo_t` to `signalfd_siginfo` transformation logic.

That workaround was likely needed because the kernel did not support signalfd for SIGSYS.

tests/integration_tests/security/test_custom_seccomp.py
```diff
@@ -137,49 +137,43 @@ def test_working_filter(test_microvm_with_api):
...
     }""".encode('utf-8'))
 
-    test_microvm.spawn()
+    microvm.spawn()
 
     # give time for the process to get killed
     time.sleep(1)
 
+    # assert ioctl was trapped
+    assert "seccomp violation on syscall number 16" in microvm.log_data
```
alindima, line 174: nit: can we also check the metric?

tests/integration_tests/security/test_custom_seccomp.py
```diff
@@ -137,49 +137,43 @@ def test_working_filter(test_microvm_with_api):
...
     """Test --seccomp-filter, denying some needed syscalls."""
-    test_microvm = test_microvm_with_ssh
+    microvm = test_microvm_with_ssh
 
     # Configure VM from JSON. Otherwise, the test will error because
     # the process will be killed before configuring the API socket.
     _config_file_setup(test_microvm_with_ssh, vm_config_file)
 
-    _custom_filter_setup(test_microvm, """{
+    _custom_filter_setup(microvm, """{
```
alindima, line 146: if we'd like to be extra sure that this works on every Firecracker thread, we could somehow parametrize this test to run for each Firecracker thread, and making sure that it is killed by SIGSYS on each of these cases.
Do you think it's worth it?

alindima:
Having looked again over the [signalfd man page](https://man7.org/linux/man-pages/man2/signalfd.2.html), it says this in regards to multithreaded programs:

> The semantics of signalfd file descriptors in a multithreaded
       program mirror the standard semantics for signals.  In other
       words, when a thread reads from a signalfd file descriptor, it
       will read the signals that are directed to the thread itself and
       the signals that are directed to the process (i.e., the entire
       thread group).  (A thread will not be able to read signals that
       are directed to other threads in the process.)

This is a detail that I did not catch when implementing the signalfd mechanism in Firecracker. This would mean that if a program directs signals to a specific Firecracker thread, other than the vmm thread, they will be ignored.

There are two options I can think of, to fix this:

1. rollback to using regular signal handlers (and think of some other mechanism of getting the handlers to be async-signal-safe). as @acatangiu suggested, we could try to lock the metrics and logger mutexes for a discrete amount of time, after which we could bail without logging.
2. install dummy regular signal handlers for the vcpus and api threads, that raise the signal to the vmm thread. These actions (`getgid` and `kill`) [are async-signal-safe](https://docs.oracle.com/cd/E19455-01/806-5257/gen-26/index.html#:~:text=A%20concept%20similar%20to%20thread,operation%20that%20is%20being%20interrupted.)

acatangiu:
> Having looked again over the [signalfd man page](https://man7.org/linux/man-pages/man2/signalfd.2.html), it says this in regards to multithreaded programs:
> 
> > The semantics of signalfd file descriptors in a multithreaded
> > program mirror the standard semantics for signals.  In other
> > words, when a thread reads from a signalfd file descriptor, it
> > will read the signals that are directed to the thread itself and
> > the signals that are directed to the process (i.e., the entire
> > thread group).  (A thread will not be able to read signals that
> > are directed to other threads in the process.)
> 
> This is a detail that I did not catch when implementing the signalfd mechanism in Firecracker.
> This means that if a program directs signals to a specific Firecracker thread, other than the vmm thread, they will be ignored.
> 
> This is generally not doable from userspace, unless the caller is a thread in the same process group. However, this is the case for the SIGSYS seccomp trap, which is specifically directed to one thread, and I am afraid that this may happen for other signals as well (SIGILL, etc.).
> 
> There are two options I can think of, to fix this:
> 
>     1. rollback to using regular signal handlers (and think of some other mechanism of getting the handlers to be async-signal-safe). as @acatangiu suggested, we could try to lock the metrics and logger mutexes for a discrete amount of time, after which we could bail without logging, if the locks were not acquired.
> 
>     2. install dummy regular signal handlers for the vcpus and api threads, that raise the signal to the vmm thread. These actions (`getpid` and `kill`) [are async-signal-safe](https://docs.oracle.com/cd/E19455-01/806-5257/gen-26/index.html#:~:text=A%20concept%20similar%20to%20thread,operation%20that%20is%20being%20interrupted.) The downside is that we'd have to also forward the `siginfo_t` data of the original signal (which for SIGSYS includes the bad syscall number). This is difficult because it requires some IPC communication for sending the siginfo_t data to the vmm thread, which must also be async-signal-safe

Great dive deep Alin!
After wrapping my head around the above I believe that the current PR satisfies all our requirements (by unintentional luck 😆):

1. General purpose signals are handled through `signalfd`. This solves the potential deadlock on log/metrics mutex by handling these signals in a async-signal-safe way. However, by using `signalfd` we lose the ability to capture signals sent to specific threads instead of thread-group/process (for example using `pthread-kill()`). **I believe this is ok/acceptable**.
   - We know for a fact we're not doing any inter-thread communication using those specific signals.
   - Firecracker process should not share its process group with other processes in production scenarios.
   - Furthermore if there are setups in the wild with Firecracker sharing process group, I don't see any reason for (trusted/well-intentioned) sibling processes on the host to target specific Firecracker threads instead of the Firecracker process for these signals.
 2. For `SIGSYS` we're using a **async-signal-unsafe** handler which guarantees that each thread can handle their own (seccomp-originated) signal. The "problem" with this is that it can deadlock on logger or metrics mutexes.
    - But.. can it really though? If we think about it, the signal is generated for thread `X` when thread `X` calls into a not-allowed syscall. Following this logic, the only way for the deadlock to happen would be if seccomp triggers during a logger or metrics specific syscall (that's the only time the lock is taken). I believe we can say with a high degree of confidence that logger and metrics related syscalls are on the allow-list; therefore **I believe this scenario too is ok/acceptable**.

Generally Linux signals are hard enough to be used correctly and safely by their nature. Any simplifying assumptions that we can make are totally worth it in my opinion if they can help us keep the implementation "simple"... or at least simpleish.

acatangiu:
Actually,

```
Certain signals are synchronously generated in the sense that they are sent to a process as the direct result of an operation within a particular thread. The system is sending the process a signal because one of its threads tried to divide by zero (SIGFPE), touch forbidden memory in the wrong way (SIGSEGV), use a broken pipe (SIGPIPE), or do something else that triggered an exception. These signals are closely bound to the activities of a given thread, and it will be that thread, in its own context, that will handle the signal on behalf of the process as a whole.
```

So I think the `signalfd` approach doesn't work for these... We can investigate further or simply roll back to classic signal handlers.

alindima:
> Actually,
> 
> ```
> Certain signals are synchronously generated in the sense that they are sent to a process as
> the direct result of an operation within a particular thread. The system is sending the
> process a signal because one of its threads tried to divide by zero (SIGFPE), touch forbidden
> memory in the wrong way (SIGSEGV), use a broken pipe (SIGPIPE), or do something else
> that triggered an exception. These signals are closely bound to the activities of a given
> thread, and it will be that thread, in its own context, that will handle the signal on behalf of
> the process as a whole.
> ```
> 
> So I think the `signalfd` approach doesn't work for these... We can investigate further or simply roll back to classic signal handlers.

I agree, the most straightforward approach would be to rollback and fix the potential signal handler deadlock some other way. I'll work on it

PR #2590, opened 2021-05-26
Fix balloon statistics being broken after a snapshot restore event

georgepisaltu:
# Reason for This PR

When testing for another issue, I resumed a snapshot with a balloon device in Firecracker v0.25. After resuming, the following log message is printed every `stats_polling_interval_s` seconds:

```
2021-05-26T16:04:35.049659990 [anonymous-instance:main:ERROR:src/devices/src/virtio/balloon/device.rs:426] Failed to update balloon stats, missing descriptor.
```

After digging into this some more, I found the root cause.

For context, the way balloon statistics work is the following:
- The guest driver is always putting a descriptor with the latest stats on the stats queue, in the available ring.
- The host sees the statistics and, when it wants new ones, it places the descriptor in the used ring.
- The guest driver immediately sends a new descriptor in the available ring.

We save the descriptor head in a special field in the balloon device (`stats_desc_index`). When the stats timer expires and we want new stats, we take the descriptor and place it in the used ring, signaling the driver that we did so.

The problem is that when saving the device state, while we do save the descriptor, we do not configure the restored device to use it. This makes the interaction broken because the descriptor is essentially lost, and the driver will not send a new one until we place this one on the used ring.

## Description of Changes

Fixed the issue described above by actually restoring `stats_desc_index` in `Balloon::restore`.
Also added a negative test for this issue in the integration tests.

## License Acceptance

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@
   flushed to their backing mediums as part of the CreateSnapshot operation.
 - Fixed the SSBD mitigation not being enabled on `aarch64` with the provided
   `prod-host-setup.md`.
+- Fixed the balloon statistics not working after a snapshot restore event.
 
 ## [0.24.0]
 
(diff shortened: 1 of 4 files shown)


gc-plp (approved):
Nice find :)

AlexandruCihodaru (approved):
ship

PR #2596, opened 2021-05-27
Add snapshot compatibility checks for snapshot create/restore

georgepisaltu:
# Reason for This PR

VMs restored from a snapshot might have the TSC frequency misaligned with the original, which could lead to problems further down the line. We took this as an opportunity to add snapshot compatibility checks, which start with adding the original TSC in the state file and scaling it for resumed VMs.

## Description of Changes

Added compatibility checks for snapshot create/restore, starting with TSC scaling.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/cpuid/src/common.rs
+++ b/src/cpuid/src/common.rs
@@ -1,6 +1,7 @@
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::bit_helper::BitHelper;
 use crate::cpu_leaf::*;
 #[cfg(target_arch = "x86_64")]
 use kvm_bindings::CpuId;
@@ -91,6 +92,70 @@ pub fn get_vendor_id_from_cpuid(cpuid: &CpuId) -> Result<[u8; 12], Error> {
     Err(Error::NotSupported)
 }
 
+/// Validates that the provided CPUID belongs to a CPU of the same
+/// model as the host's.
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+pub fn is_same_model(cpuid: &CpuId) -> bool {
+    // Try to get the vendor IDs from the host and the CPUID struct.
+    if let (Ok(host_vendor_id), Ok(cpuid_vendor_id)) =
+        (get_vendor_id_from_host(), get_vendor_id_from_cpuid(cpuid))
+    {
+        // If the vendor IDs aren't the same, the CPUs are not identical.
+        if host_vendor_id != cpuid_vendor_id {
+            return false;
+        }
+    } else {
+        // This only fails when CPUID is not supported, in which case
+        // we can't tell if the CPUs are identical.
+        return false;
+    }
+
+    // Try to get the feature information leaf from the host CPUID.
+    let host_feature_info_leaf = get_cpuid(leaf_0x1::LEAF_NUM, 0);
+
+    // The relevant information for this comparison is in the EAX register.
+    let host_feature_info_leaf_eax = match host_feature_info_leaf {
+        Ok(leaf) => leaf.eax,
+        Err(_) => {
+            // If this fails, we can't tell if the CPUs are identical.
+            return false;
+        }
+    };
+
+    // Search for the entry for leaf0x1.
+    let feature_info_leaf = cpuid
+        .as_slice()
+        .iter()
+        .find(|entry| entry.function == leaf_0x1::LEAF_NUM);
+
+    // The relevant information is in EAX.
+    let feature_info_leaf_eax = match feature_info_leaf {
+        Some(leaf) => leaf.eax,
+        None => {
+            // Fail fast if we can't retrieve the relevant
+            // information from CPUID.
+            return false;
+        }
+    };
+
+    // Validate that all of these properties are the same.
+    for elem in &[
+        leaf_0x1::eax::EXTENDED_FAMILY_ID_BITRANGE,
+        leaf_0x1::eax::EXTENDED_PROCESSOR_MODEL_BITRANGE,
+        leaf_0x1::eax::PROCESSOR_FAMILY_BITRANGE,
+        leaf_0x1::eax::PROCESSOR_MODEL_BITRANGE,
+        leaf_0x1::eax::STEPPING_BITRANGE,
+    ] {
+        if feature_info_leaf_eax.read_bits_in_range(elem)
+            != host_feature_info_leaf_eax.read_bits_in_range(elem)
+        {
+            return false;
+        }
+    }
+
+    true
+}
+
 #[cfg(test)]
 pub mod tests {
     use crate::common::*;
@@ -146,5 +211,43 @@ pub mod tests {
         let vendor_id = get_vendor_id_from_host();
         assert!(vendor_id.is_ok());
         matches!(&vendor_id.ok().unwrap(), VENDOR_ID_INTEL | VENDOR_ID_AMD);
+    }
+
+    #[test]
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    fn test_is_same_model() {
+        let mut curr_cpuid = CpuId::new(2).unwrap();
+
+        // Add the vendor ID leaf.
+        let vendor = get_cpuid(0x0, 0).unwrap();
+        curr_cpuid.as_mut_slice()[0].function = 0x0;
+        curr_cpuid.as_mut_slice()[0].index = 0;
+        curr_cpuid.as_mut_slice()[0].ebx = vendor.ebx;
+        curr_cpuid.as_mut_slice()[0].ecx = vendor.ecx;
+        curr_cpuid.as_mut_slice()[0].edx = vendor.edx;
+
+        // Add the feature info leaf.
+        let feature_info = get_cpuid(0x1, 0).unwrap();
+        curr_cpuid.as_mut_slice()[1].function = 0x1;
+        curr_cpuid.as_mut_slice()[1].index = 0;
+        curr_cpuid.as_mut_slice()[1].eax = feature_info.eax;
+
+        assert!(is_same_model(&curr_cpuid));
+
+        let mut diff_vendor_cpuid = curr_cpuid.clone();
+        for mut entry in diff_vendor_cpuid.as_mut_slice() {
+            if entry.function == 0x0 && entry.index == 0 {
+                entry.ebx = 0xFFFF_FFFF;
+            }
+        }
+        assert!(!is_same_model(&diff_vendor_cpuid));
+
+        let mut diff_feature_cpuid = curr_cpuid;
+        for mut entry in diff_feature_cpuid.as_mut_slice() {
+            if entry.function == 0x1 && entry.index == 0 {
+                entry.eax ^= 0x1;
+            }
+        }
+        assert!(!is_same_model(&diff_feature_cpuid));
     }
 }
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -26,6 +26,8 @@ use crate::{device_manager, Error, EventManager, Vmm, VmmEventsObserver};
 
 use crate::vmm_config::instance_info::InstanceInfo;
 use arch::InitrdConfig;
+#[cfg(target_arch = "x86_64")]
+use cpuid::common::is_same_model;
 #[cfg(target_arch = "aarch64")]
 use devices::legacy::RTCDevice;
 use devices::legacy::Serial;
@@ -426,6 +428,41 @@ pub fn build_microvm_from_snapshot(
         track_dirty_pages,
         vcpu_count,
     )?;
+
+    #[cfg(target_arch = "x86_64")]
+    // Check if we need to scale the TSC.
+    // We start by checking if the CPU model in the snapshot is
+    // the same as this host's. If they are the same, we don't
+    // need to do anything else.
+    if !is_same_model(&microvm_state.vcpu_states[0].cpuid) {
+        // Extract the TSC freq from the state.
+        // No TSC freq in snapshot means we have to fail-fast.
+        let state_tsc = microvm_state.vcpu_states[0]
+            .tsc_khz
+            .ok_or_else(|| {
+                MicrovmStateError::IncompatibleState(
+                    "Error configuring the TSC, frequency not \
+                    present in snapshot."
+                        .to_string(),
+                )
+            })
+            .map_err(RestoreMicrovmState)?;
+
+        // Scale the TSC frequency for all VCPUs, if needed.
+        if vcpus[0]
+            .kvm_vcpu
+            .is_tsc_scaling_required(state_tsc)
+            .map_err(|e| MicrovmStateError::IncompatibleState(e.to_string()))
+            .map_err(RestoreMicrovmState)?
+        {
+            for vcpu in &vcpus {
+                vcpu.kvm_vcpu
+                    .set_tsc_khz(state_tsc)
+                    .map_err(|e| MicrovmStateError::IncompatibleState(e.to_string()))
+                    .map_err(RestoreMicrovmState)?;
+            }
+        }
+    }
 
     #[cfg(target_arch = "aarch64")]
     {
--- a/src/vmm/src/vstate/vcpu/x86_64.rs
+++ b/src/vmm/src/vstate/vcpu/x86_64.rs
@@ -21,10 +21,16 @@ use kvm_bindings::{
     kvm_xsave, CpuId, MsrList, Msrs,
 };
 use kvm_ioctls::{VcpuExit, VcpuFd};
-use logger::{error, IncMetric, METRICS};
+use logger::{error, warn, IncMetric, METRICS};
 use versionize::{VersionMap, Versionize, VersionizeResult};
 use versionize_derive::Versionize;
 use vm_memory::{Address, GuestAddress, GuestMemoryMmap};
+
+// Tolerance for TSC frequency expected variation.
+// The value of 250 parts per million is based on
+// the QEMU approach, more details here:
+// https://bugzilla.redhat.com/show_bug.cgi?id=1839095
+const TSC_KHZ_TOL: f64 = 250.0 / 1_000_000.0;
 
 /// Errors associated with the wrappers over KVM ioctls.
 #[derive(Debug)]
@@ -67,6 +73,8 @@ pub enum Error {
     VcpuGetXsave(kvm_ioctls::Error),
     /// Failed to get KVM vcpu cpuid.
     VcpuGetCpuid(kvm_ioctls::Error),
+    /// Failed to get KVM TSC freq.
+    VcpuGetTSC(kvm_ioctls::Error),
     /// Failed to set KVM vcpu cpuid.
     VcpuSetCpuid(kvm_ioctls::Error),
     /// Failed to set KVM vcpu debug regs.
@@ -87,6 +95,8 @@ pub enum Error {
     VcpuSetXcrs(kvm_ioctls::Error),
     /// Failed to set KVM vcpu xsave.
     VcpuSetXsave(kvm_ioctls::Error),
+    /// Failed to set KVM TSC freq.
+    VcpuSetTSC(kvm_ioctls::Error),
 }
 
 impl Display for Error {
@@ -125,6 +135,7 @@ impl Display for Error {
             VcpuGetXcrs(e) => write!(f, "Failed to get KVM vcpu xcrs: {}", e),
             VcpuGetXsave(e) => write!(f, "Failed to get KVM vcpu xsave: {}", e),
             VcpuGetCpuid(e) => write!(f, "Failed to get KVM vcpu cpuid: {}", e),
+            VcpuGetTSC(e) => write!(f, "Failed to get KVM TSC frequency: {}", e),
             VcpuSetCpuid(e) => write!(f, "Failed to set KVM vcpu cpuid: {}", e),
             VcpuSetDebugRegs(e) => write!(f, "Failed to set KVM vcpu debug regs: {}", e),
             VcpuSetLapic(e) => write!(f, "Failed to set KVM vcpu lapic: {}", e),
@@ -135,6 +146,7 @@ impl Display for Error {
             VcpuSetVcpuEvents(e) => write!(f, "Failed to set KVM vcpu event: {}", e),
             VcpuSetXcrs(e) => write!(f, "Failed to set KVM vcpu xcrs: {}", e),
             VcpuSetXsave(e) => write!(f, "Failed to set KVM vcpu xsave: {}", e),
+            VcpuSetTSC(e) => write!(f, "Failed to set KVM TSC frequency: {}", e),
         }
     }
 }
@@ -225,6 +237,11 @@ impl KvmVcpu {
         self.pio_bus = Some(pio_bus);
     }
 
+    /// Get the current TSC frequency for this vCPU.
+    pub fn get_tsc_khz(&self) -> Result<u32> {
+        self.fd.get_tsc_khz().map_err(Error::VcpuGetTSC)
+    }
+
     /// Save the KVM internal state.
     pub fn save_state(&self) -> Result<VcpuState> {
         /*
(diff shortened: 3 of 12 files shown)


src/api_server/swagger/firecracker.yaml
```diff
@@ -1018,6 +1022,10 @@ definitions:
         type: boolean
         description:
           When set to true, the vm is also resumed if the snapshot load is successful.
+      skip_compatibility_check:
+        type: boolean
+        description:
+          Optional field to skip the snapshot compatibility checks, which are on by default.
```
acatangiu, line 1028: Why do we need this knob? It adds a lot of complexity on the user/usage side as well Firecracker code complexity for handling all new cases introduced by it.

If the only production-worthy reason for it is to allow loading old snapshots, I suggest we look at some other approaches for migrating/upgrading those snaps, instead of adding and maintaining an extra knob in perpetuity..

acatangiu: We could offer a dedicated snapshot compatibility tool for example, or simply warn when coming from TSC-unaware snaps but do best-effort and load them.

dianpopa:
Should we close [this one](https://github.com/firecracker-microvm/firecracker/pull/2522/files)?

src/api_server/swagger/firecracker.yaml
berciuliviu: I agree with Adrian, we should take a simpler approach to this issue, without adding a new parameter into the account. I also believe that validating v0.25 snapshots for tsc_freq should be done by default, while loading an older-version snapshot should just warn of tsc_freq not available. If my understanding is correct, vcpu_state.tsc_freq is automatically set to None if an older version snapshot is loaded.

src/cpuid/src/common.rs
```diff
@@ -91,6 +92,39 @@ pub fn get_vendor_id_from_cpuid(cpuid: &CpuId) -> Result<[u8; 12], Error> {
     Err(Error::NotSupported)
 }
 
+/// Validates that a CPUID and another leaf 0x1 describe the same
+/// CPU in terms of manufacturer and model.
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+pub fn is_same_model(cpuid: &CpuId, other_leaf0x1_eax: u32) -> bool {
```
serban300, line 98: I wouldn't supply the other `0x1` leaf as a parameter. I would read it inside this method.

src/cpuid/src/common.rs
```diff
@@ -91,6 +92,39 @@ pub fn get_vendor_id_from_cpuid(cpuid: &CpuId) -> Result<[u8; 12], Error> {
...
+    // Search for the entry for leaf0x1.
+    let leaf0x1 = cpuid
```
serban300, line 100: The `0x1` leaf is also referred to as the feature information leaf. So I think we can name this something along the lines of `feature_info_leaf`.

src/cpuid/src/common.rs
```diff
@@ -91,6 +92,39 @@ pub fn get_vendor_id_from_cpuid(cpuid: &CpuId) -> Result<[u8; 12], Error> {
...
+
+    // Validate that all of these properties are the same.
+    for elem in &[
+        leaf_0x1::eax::EXTENDED_FAMILY_ID_BITRANGE,
+        leaf_0x1::eax::EXTENDED_PROCESSOR_MODEL_BITRANGE,
+        leaf_0x1::eax::PROCESSOR_FAMILY_BITRANGE,
+        leaf_0x1::eax::PROCESSOR_MODEL_BITRANGE,
+        leaf_0x1::eax::STEPPING_BITRANGE,
+    ] {
+        if leaf0x1_eax.read_bits_in_range(elem) != other_leaf0x1_eax.read_bits_in_range(elem) {
```
serban300, line 120: Are this values guaranteed to be unique even across different vendors ? Shouldn't we also compare the vendor just to make sure ? The vendor can be extracted from the brand string and we might already have a helper method for that.

src/cpuid/src/common.rs
```diff
@@ -91,6 +92,39 @@ pub fn get_vendor_id_from_cpuid(cpuid: &CpuId) -> Result<[u8; 12], Error> {
...
+        .as_slice()
+        .iter()
+        .find(|entry| entry.function == leaf_0x1::LEAF_NUM);
+    // Fail fast if we can't retrieve the relevant information from CPUID.
+    if leaf0x1.is_none() {
+        return false;
+    }
+
+    // The relevant information is in EAX.
+    let leaf0x1_eax = leaf0x1.unwrap().eax;
```
serban300, line 110: Does something like `let leaf0x1_eax = leaf0x1?.eax;` work ?

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -291,9 +307,55 @@ impl KvmVcpu {
...
             xsave,
+            tsc_khz,
         })
     }
 
+    // Scale the TSC frequency of this vCPU to the one provided as a parameter.
+    pub fn set_tsc_khz(&self, state_tsc: Option<u32>, cpuid: &CpuId) -> Result<()> {
+        // Check if CPU models are the same.
+        if let Ok(curr_cpuid) = cpuid::common::get_cpuid(0x1, 0) {
+            if is_same_model(cpuid, curr_cpuid.eax) {
```
sandreim, line 318: This allow snapshots to be loaded on the same CPU, even if there is no constant TSC or TSC scaling support. Are we sure the checks we do here guarantee that TSC freq is the same ?

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -291,9 +307,55 @@ impl KvmVcpu {
...
+        } else {
+            state_tsc.unwrap()
+        };
+
+        match curr_kvm_tsc {
+            Ok(curr_tsc_freq) => {
+                // Compare the current TSC freq to the one found
+                // in the state. If they are different, we need to
+                // scale the TSC to the freq found in the state.
+                if curr_tsc_freq != snapshot_tsc_freq {
```
sandreim, line 340: I think we should allow for some margin of error here, since even on the same CPU you might get different values during calibration.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -392,6 +457,30 @@ pub struct VcpuState {
...
+        None
+    }
+
+    fn ser_tsc(&mut self, target_version: u16) -> VersionizeResult<()> {
+        if target_version < 2 {
+            warn!("Saving to older snapshot version, TSC freq not included in snapshot.");
+        }
+
+        // Prevent v0.25 to create snapshots of uVMs running on non-invariant TSC host.
+        if self.tsc_khz.is_none() {
```
sandreim, line 476: We dont't allow this but we do load snaps from v0.24 (without TSC) given the CPUID model information matches. What happens if we drop this check ?

src/cpuid/src/common.rs
georgepisaltu: My research led me to believe they are unique across every CPU model, which means different vendors would make them different as well.

However, the vendor string check doesn't hurt in this case so I added it just to be safe.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: If we as a process never change the TSC frequency outside the scope of a snapshot save/restore, and the CPU model is the same, I expect the TSC frequency to be the same in the clone as in the original VM. If this is not true, then checking the CPU model is actually useless, since we still need to get/set the TSC frequency regardless.

On variant TSC systems, I don't think users even expect high precision from the TSC, but I might have a bad understanding of this whole story.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: Any margin of error here can lead to a great divergence in TSC over a period of time.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: Depends on whether you want to load the TSC-less snapshot on the same CPU model or not. I guess this ties in to the previous comment.

src/vmm/src/vstate/vcpu/x86_64.rs
sandreim: Some differences can be corrected by the guest NTP, more info in QEMU approach: https://patchew.org/QEMU/20200616165805.GA324612@fuller.cnet/

src/vmm/src/vstate/vcpu/x86_64.rs
sandreim: I think we should have a consistent behavior both for load and save wrt older snapshot versions.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: Updated it/

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: Dropped the check, now the failure path is on snapshot restore. We still warn users about this with a log.

georgepisaltu (commented):
@sandreim PTAL!

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -291,9 +309,57 @@ impl KvmVcpu {
...
     }
 
+    // Scale the TSC frequency of this vCPU to the one provided as a parameter.
+    pub fn set_tsc_khz(&self, state_tsc: Option<u32>, cpuid: &CpuId) -> Result<()> {
+        // Check if CPU models are the same.
+        if is_same_model(cpuid) {
+            // If they are the same, we don't need to do anything else.
+            return Ok(());
+        }
+
```
sandreim, line 323: Should we fail early here if KVM caps for get/set TSC freq are not available ?

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -250,6 +266,8 @@ impl KvmVcpu {
          * meaningful. For SET_MSRS it will then contain good data.
          */
 
+        let tsc_khz = self.get_tsc_khz()?;
```
sandreim, line 269: Do we still want to allow snapshots to be created if this fails ?

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -392,6 +461,34 @@ pub struct VcpuState {
...
+    fn default_tsc_khz(_: u16) -> Option<u32> {
+        warn!("CPU TSC freq not found in snapshot");
+        None
+    }
+
+    fn ser_tsc(&mut self, target_version: u16) -> VersionizeResult<()> {
+        if target_version < 2 {
+            // v0.24 and older versions do not support TSC scaling.
+            warn!(
+                "Saving to older snapshot version, TSC freq not included \
```
sandreim, line 478: It would be helpful to also include the TSC freq in the message.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -392,6 +461,34 @@ pub struct VcpuState {
...
+                in snapshot."
+            );
+        } else if self.tsc_khz.is_none() {
+            // v0.25 and newer snapshots without TSC will only work on
+            // the same CPU model as the host on which they were taken.
+            warn!(
+                "TSC freq not available. Snapshot will only work on the \
```
sandreim, line 485: I would rephrase: `Snapshot cannot be loaded on a different CPU model.`

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -291,9 +309,57 @@ impl KvmVcpu {
...
+    // Scale the TSC frequency of this vCPU to the one provided as a parameter.
+    pub fn set_tsc_khz(&self, state_tsc: Option<u32>, cpuid: &CpuId) -> Result<()> {
+        // Check if CPU models are the same.
+        if is_same_model(cpuid) {
+            // If they are the same, we don't need to do anything else.
+            return Ok(());
+        }
+
+        // Try to get a reading of the current TSC freq.
+        let curr_kvm_tsc = self.fd.get_tsc_khz();
```
sandreim, line 325: Can we refactor so the syscall and validation get called only once ? I expect the TSC freq to be the same across all vCPUs, so the load hotpath can be simplified.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: Yes, as we discussed earlier. Fixed it.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: It is a rather complex state where the `KVM_GET_TSC_KHZ` cap is always needed at this point, but the call is made just below and a failure does not tear the whole restore process down. Furthermore, we will only know if we need the `KVM_SET_TSC_KHZ` cap later, and we check it when we need it. So I don't really see the point of checking the capabilities here.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: I found it very difficult to do this design-wise. There is no good place to do this, as this is in the snap restore path and the "owners" of the VCPUs are, in order, the `builder.rs` function `build_microvm_from_snapshot` and then the VCPU threads themselves, which communicate with the VMM via events. It would be really awkward to do it by passing parameters through the event enum. The other option is to do it with the builder, but there is no good place to do it there either.

I tested this POC and there seems to be no performance degradation in snapshot restore times. If we do the validations once before the VCPUs are moved to their threads or if we do it on every VCPU thread, the total amount of time to restore completion should be roughly the same. Let me know if I missed anything and we need to revisit this.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: Done.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: Done.

georgepisaltu (commented):
@sandreim @serban300 I updated the PR, also added unit tests.

The next step is to add integration tests using snapshots generated with the final form of the binary after merging this.

PTAL!

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: I thought about it more and I attempted the refactor in the current iteration of the PR. Let me know if it makes sense. After all, doing the checks only once is the right way to go forward in the long run.

src/vmm/src/builder.rs
```diff
@@ -427,6 +427,27 @@ pub fn build_microvm_from_snapshot(
...
+    {
+        // Scale the TSC frequency for all VCPUs.
+        for vcpu in &vcpus {
+            vcpu.kvm_vcpu
+                .set_tsc_khz(microvm_state.vcpu_states[0].tsc_khz)
+                .map_err(|e| MicrovmStateError::IncompatibleState(e.to_string()))
+                .map_err(RestoreMicrovmState)?;
+        }
+    }
+
```
serban300, line 450: This approach leads to some developments that I personally find  both unsafe and hard to follow. For example:

```
    pub fn set_tsc_khz(&self, state_tsc: Option<u32>) -> Result<()> {
        // Extract the TSC freq from the state.
        let snapshot_tsc_freq = match state_tsc {
            Some(freq) => freq,
            None => {
                // The snapshot TSC was already validated.
                unreachable!();
            }
        };
        ...
    }
```
Here relying on the fact that the TSC was already validated is very risky since it's not in the control of the function and not even in the control of the class. `state_tsc` is a parameter that we receive and as far as we know here, it's completely out of our control.

I think this is all generated by the attempt to squeeze the `is_same_model()` check inside the `is_tsc_scaling_required()`, which definitely makes sense initially, but it is not worth the tradeoffs. As an incremental improvement here, just to make the code safer, I would at least separate `is_same_mode()` from `is_tsc_scaling_required()`, changing this code to something like:

```
        if !cpuid::common::is_same_model(&microvm_state.vcpu_states[0].cpuid) {
            let state_tsc = microvm_state.vcpu_states[0]
                .tsc_khz
                .ok_or(MicrovmStateError::IncompatibleState(
                    "TSC Error".to_string(),
                ))

            if vcpus[0].kvm_vcpu.is_tsc_scaling_required(state_tsc) {
                for vcpu in &vcpus {
                    vcpu.kvm_vcpu
                        .set_tsc_khz(state_tsc)
                        .map_err(|e| MicrovmStateError::IncompatibleState(e.to_string()))
                }
            }
        }
```

Where `is_tsc_scaling_required()` and `set_tsc_khz()` would also consequently change for the better since they would receive the `state_tsc_freq` as a u32 and not an option.

Cosmetically I guess we can go even further. It would be nice if we could do this operation as part of `KvmVcpu::restore_state()`. If it needs to be done earlier, I think we can find a clean way to do it. The implementation details would vary a lot depending on the requirements. We can discuss options offline.

src/cpuid/src/common.rs
```diff
@@ -147,4 +215,46 @@ pub mod tests {
...
     }
+
+    #[test]
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    fn test_is_same_model() {
+        let kvm = Kvm::new().unwrap();
+        let vm = kvm.create_vm().unwrap();
+        let vcpu = vm.create_vcpu(0).unwrap();
+
+        let mut curr_cpuid = vcpu.get_cpuid2(KVM_MAX_CPUID_ENTRIES).unwrap();
```
serban300, line 226: I would create the cpuid manually instead of calling `vcpu.get_cpuid2()`. You can find examples in other unit tests.

serban300 (changes requested):

src/cpuid/src/common.rs
georgepisaltu: Done.

src/vmm/src/builder.rs
georgepisaltu: While I don't necessarily see it as unsafe, I understand the concern. There is virtually no cost in refactoring this the way you described (in order to get rid of the `unreachable!` statement), so I tried to do it. Let me know if it's better.

Unfortunately, for the last part of the comment, the checks would ideally happen only once, so this means they can't be done inside `KcmVcpu::restore_state()`. This is what we've been trying to avoid. We considered workarounds for this, such as meddling with the TSC `Option<u32>` inside the state, but they feel clunky. @sandreim what do you think?

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -15,17 +15,19 @@ use crate::vstate::{
...
     kvm_xsave, CpuId, MsrList, Msrs,
 };
 use kvm_ioctls::{VcpuExit, VcpuFd};
-use logger::{error, IncMetric, METRICS};
+use logger::{error, warn, IncMetric, METRICS};
 use versionize::{VersionMap, Versionize, VersionizeResult};
 use versionize_derive::Versionize;
 use vm_memory::{Address, GuestAddress, GuestMemoryMmap};
 
+const TSC_KHZ_TOL: f64 = 250.0 / 1_000_000.0;
```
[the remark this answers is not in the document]
georgepisaltu: `TOL` is the tolerance for normal TSC frequency differences that appear after boot. The calibration process does not yield the same frequency every time. As for the second point, I misplaced the `.round()`. It was supposed to be rounded *after* multiplying with the current value. Fixed it now.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -291,9 +305,69 @@ impl KvmVcpu {
...
+                    Ok(false)
+                }
+            }
+            None => {
+                // We don't know the TSC freq for this machine,
+                // but we still have the one in the state. Try
+                // to set that regardless as a last effort.
+                Ok(true)
+            }
+        }
```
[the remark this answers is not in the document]
georgepisaltu: After the refactoring for the other comment, this is not needed anymore.

src/vmm/src/builder.rs
```diff
@@ -427,6 +429,36 @@ pub fn build_microvm_from_snapshot(
...
+        // Scale the TSC frequency for all VCPUs, if needed.
+        if vcpus[0].kvm_vcpu.is_tsc_scaling_required(state_tsc) {
+            for vcpu in &vcpus {
+                vcpu.kvm_vcpu
+                    .set_tsc_khz(state_tsc)
+                    .map_err(|e| MicrovmStateError::IncompatibleState(e.to_string()))
+                    .map_err(RestoreMicrovmState)?;
+            }
+        }
+    }
```
serban300, line 460: I would move this code to the beginning of `restore_vcpu_states()`

tests/integration_tests/functional/test_snapshot_advanced.py
```diff
@@ -137,6 +137,58 @@ def test_restore_old_version_all_devices(bin_cloner_path):
...
+        fc_binary=firecracker.local_path(),
+        jailer_binary=jailer.local_path(),
+        diff_snapshots=diff_snapshots,
+        balloon=True
+    )
+
+    # Resume microvm using current build of FC/Jailer.
+    # The resume should be successful because the CPU model
+    # in the snapshot state is the same as this host's.
+    microvm, _ = builder.build_from_snapshot(
```
sandreim, line 176: Please also add an additional issue to cover negative testing on different CPU models.

src/vmm/src/builder.rs
sandreim: Failing early and reducing resume cost are good reasons to have only one of the CPU states checked once. This works with the simplifying assumption that the host supports invariant TSC and freq is the same across all CPUs.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -392,6 +434,37 @@ pub struct VcpuState {
...
+}
+
+impl VcpuState {
+    fn default_tsc_khz(_: u16) -> Option<u32> {
+        warn!("CPU TSC freq not found in snapshot");
+        None
+    }
+
+    fn ser_tsc(&mut self, target_version: u16) -> VersionizeResult<()> {
+        if target_version < 2 {
```
sandreim, line 448: This can be simplified, no need to check the target_version here, please see https://docs.rs/versionize_derive/0.1.4/versionize_derive/derive.Versionize.html#ser_fn

Also I would add asserts in tests for the 2 warnings.

src/vmm/src/builder.rs
serban300: When Firecracker is used correctly, we shouldn't encounter issues related to setting the TSC. Failing should be more of an edge case. So I think that it's worth the move for readability.

georgepisaltu: The problem is that by the time `restore_vcpu_states` is called, the VCPUs are already moved to their own threads and we run into the same problem as before regarding the redundant checks. Furthermore, it would just add extra complexity to the code to try and send these check specific fields through the `VcpuHandle` interface. I lean towards keeping it as is.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -291,9 +306,35 @@ impl KvmVcpu {
...
+            // per million beacuse it is common for TSC frequency
+            // to differ due to calibration at boot time.
+            Some(curr_tsc_freq) => {
+                let diff = (curr_tsc_freq as i64 - state_tsc_freq as i64).abs();
+                diff > (state_tsc_freq as f64 * TSC_KHZ_TOL).round() as i64
+            }
+            // We don't know the TSC freq for this machine,
+            // but we still have the one in the state. Try
+            // to set that regardless as a last effort.
+            None => true,
```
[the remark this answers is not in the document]
georgepisaltu: Refactored to fail when `get_tsc_khz` fails.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -428,6 +503,31 @@ mod tests {
         (vm, vcpu, vm_mem)
     }
 
+    /*fn mutate_state_cpuid_vendor(orig_state: &VcpuState) -> VcpuState {
```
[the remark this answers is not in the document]
georgepisaltu: Done.

tests/integration_tests/functional/test_snapshot_advanced.py
georgepisaltu: Will do, after this PR is merged.

src/vmm/src/vstate/vcpu/x86_64.rs
georgepisaltu: I refactored this and added a negative test for the second message. The first one is really tough to get. I left a TODO there for when we decide to have mock structures for testing here, so we can easily cover the case when `get_tsc_khz` doesn't work.

georgepisaltu (commented):
@serban300 @sandreim PTAL!

serban300 (approved):

sandreim (approved):

Issue #2585, opened 2021-05-26
[Bug] "Deny All" Seccomp filter configuration not working properly


acatangiu:
# Describe the bug

It seems that following the `seccompiler` merge to `main`, seccomp filters are no longer working properly. Using empty allow-list doesn't kill firecracker as expected.

## Affected versions:

No released binaries affected, only binaries built from recent (have yet to pin-point commit) `main` branch.

## To Reproduce

1. Use a default **deny-all** filter in `resources/seccomp/x86_64-unknown-linux-musl.json`:
```
{
    "vmm": {
        "default_action": "trap",
        "filter_action": "allow",
        "filter": []
    },
    "api": {
        "default_action": "trap",
        "filter_action": "allow",
        "filter": []
    },
    "vcpu": {
        "default_action": "trap",
        "filter_action": "allow",
        "filter": []
    }
}
```
2. Compile Firecracker from sources
3. Run Firecracker and observe it runs successfully

## Expected behaviour

Since described allow-list contains no allowed syscalls Firecracker should be terminated by SIGSYS.

## Checks

- [x] Have you searched the Firecracker Issues database for similar problems?
- [x] Have you read the existing relevant Firecracker documentation?
- [x] Are you certain the bug being reported is a Firecracker issue?

acatangiu:
This seems to affect only empty lists, allow-lists that have at least one entry seem to work fine.

acatangiu:
@alindima found the root-cause: https://github.com/firecracker-microvm/firecracker/blob/e1ee11c575150cde23d05353d93c9c38dfb21a53/src/seccompiler/src/backend.rs#L766

Encountering an empty list shortcirtcuits logic to a blanket allow everything. This means that the only affected scenario is _trap-all_ filters which is not a useful configuration anyway.

Lowering priority based on the observation that this only affects a non-critical corner-case.

PR #2598, opened 2021-05-28
seccompiler: remove rule-level actions and other fixes

alindima:
# Reason for This PR

This commit fixes https://github.com/firecracker-microvm/firecracker/issues/2585, adds a check to deny filters that have the same on-match and default actions and removes support for rule-level actions.

See details on why we are removing rule-level actions below:

Until now, a filter consisted of a default action, a filter action and an array of rules.
Each of these rules can have an optional rule-level action, that, when it matches, overrides the filter-level action.
This can be useful sometimes, when debugging or for some advanced filters. For example, you could specifically kill the process when you get an exec() call, instead of trapping, which would be the filter-level action.

Firecracker does not use this option, and there are no real use cases I can think of, where this is neccessary.
Having thought of it some more, it can lead to some undefined results. Consider this filter:

```
{
    "vmm":
        "default_action": "trap",
        "filter_action": "allow",
        "filter": {
                "syscall": "futex",
                "comment": "Used for synchronization",
                "args": [
                    {
                        "arg_index": 1,
                        "arg_type": "dword",
                        "op": "eq",
                        "val": 128,
                        "comment": "FUTEX_WAIT_PRIVATE"
                    }
                ]
            },
            {
                "syscall": "futex",
                "comment": "Used for synchronization",
                "args": [
                    {
                        "arg_index": 1,
                        "arg_type": "dword",
                        "op": "eq",
                        "val": 129,
                        "comment": "FUTEX_WAKE_PRIVATE"
                    }
                ]
            }
            {
                "syscall": "futex",
                "comment": "Used for synchronization",
                "action": "kill",
                "args": [
                    {
                        "arg_index": 1,
                        "arg_type": "dword",
                        "op": "eq",
                        "val": 129,
                        "comment": "FUTEX_WAKE_PRIVATE"
                    }
                ]
            }
        }
    }
}
```

Note the `"action":"kill"` on the last rule.
For `FUTEX_WAKE_PRIVATE`, the result is undefined. Depending on how we compile the BPF and
the end order of the rules, this can either allow the syscall or kill the process.
This case would be difficult to validate, and for a use-case that Firecracker does not currently need.

Therefore, this PR removes it. It's a two way door. If, in the future, it turns out that we need it, we can build it.

## Description of Changes

## License Acceptance

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,8 @@
   statistics update immediately after processing the request.
 - Deprecated the `--seccomp-level parameter`. It will be removed  in a future
   release. Using it logs a runtime warning.
-- Experimental gnu libc builds no longer use a default seccomp filter.
+- Experimental gnu libc builds use empty default seccomp filters, allowing all
+  system calls.
 
 ### Fixed
 
--- a/docs/seccompiler.md
+++ b/docs/seccompiler.md
@@ -29,8 +29,8 @@ Example usage:
     --input-file "x86_64_musl.json" # File path of the JSON input.
     --output-file "bpf_x86_64_musl" # Optional path of the output file.
                                     # [default: "seccomp_binary_filter.out"]
-    --basic # Optional, creates basic filters, discarding rule-level actions
-            # and any parameter checks. Not recommended.
+    --basic # Optional, creates basic filters, discarding any parameter checks.
+            # (Not recommended).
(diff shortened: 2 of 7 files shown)


acatangiu (approved):
Looks good!

berciuliviu (approved):

Issue #2176, opened 2020-10-14
Problem on block device rate-limiting


namespaceXP:
I'm interested in I/O throttling. While testing the rate-limiter on block device, I use the following setting to set the average iops to 3000: 

"rate_limiter": {
  "ops": {
    "size": 3000,
    "one_time_burst": 0,
    "refill_time": 1000
    }
}

And use fio for testing the iops: 

`fio -rw=randwrite -ioengine=libaio -bs=4k -size=10G -numjobs=1 -runtime=1000 -group_reporting -filename=iotest -name=Rand_Write_Testing -direct=1 -iodepth=128`

However, I've found that the average iops is about 2765, somehow lower than expected(3000). I wonder the reason.

raduiliescu:
Before diving deep on the configuration, what performance number do you get without rate-limiter?

namespaceXP:
> Before diving deep on the configuration, what performance number do you get without rate-limiter?

The average IOPS without rate limiter is about 56000.

serban300:
This might be happening because the rate limiter also counts Flush and DeviceID requests. I will check it.

serban300:
I did some more research here. This is not caused by Flush and DeviceID requests. This seems to be the desired behavior for the rate limiter. 

The rate limiter allows 3000 OPS, after which it waits 1000 ms and then it receives a new budget of 3000 OPS. So the configuration that you used for the rate limiter doesn't translate to 3000 IOPS. it translates to cycles like the following:
```
+--------------------------+-------------+
| 3000 OPS                 | refill time |
+--------------------------+-------------+
| time needed for 3000 OPS | 1000 ms     |
+--------------------------+-------------+
```

In your case the time needed for 3000 OPS is approximately 1000 ms / 56000 * 3000 = 54ms . So for example in 1 minute there will be 60000 / 1054 = 57 such cycles. 57 * 3000 / 60 = 2850 IOPS which is close to the result that you are getting.

@namespaceXP does this answear your question ? Do you have any concerns related to this behavior ?

@acatangiu It seems that you implemented the block device rate limiting. Can you confirm if this is the desired behavior ?

acatangiu:
@namespaceXP The firecracker rate-limiter implementation leans more on efficiency at the expense of accuracy.

[...]

PR #2601, opened 2021-05-31
rate-limiter: increase limiter precision

acatangiu:
# Reason for This PR

Fixes https://github.com/firecracker-microvm/firecracker/issues/2176

The firecracker rate-limiter implementation leans more on efficiency at the expense of accuracy. Even so, the accuracy can be improved without performance penalties.

## Description of Changes

Firecracker rate-limiter is a passive implementation driven by the consumer. Each time `X` tokens go through it, the rate-limiter:
 1. calculates how much time has passed since last operation
 2. replenishes its budget based on elapsed time above
 3. attempts to allow `X` if it fits the budget

Because we don't want to incur the cost of floating point operations the operation in `step 2` is a non-floating point division which loses precision:
```rust
  // At each 'time_delta' nanoseconds the bucket should refill with:
  refill_amount = (time_delta * size) / (refill_time_ms * 1_000_000)
```

The smaller `time_delta * size` and the more of these operations we do, the higher our precision loss.

We can't control `size` since it's user-provided input, but we can slightly change our implementation to optimize on the other two dimensions: we want bigger `time_delta`s and fewer division operations.

These are complementary and can be achieved by doing bucket replenish operations only when needed (budget is low/empty) instead of every time the limiter sees traffic.

This small change results in increased precision (it should be cheaper too since we're doing fewer operations).

## Testing

Set up block rate limiter:
```json
"rate_limiter": {
  "ops": {
    "size": 3000,
    "one_time_burst": 0,
    "refill_time": 1000
  }
}
```

And tested _ops rate-limiting_ using `fio`:

```
fio -rw=randwrite -ioengine=libaio -bs=4k -size=1G -numjobs=1 -runtime=1000 -group_reporting -filename=iotest -name=Rand_Write_Testing -direct=1 -iodepth=128
```

### Before

```
iops        : min=  1980, max=13142, avg=2794.51
```

### After

```
iops        : min= 2518, max=12576, avg=2989.88
```

## License Acceptance

## PR Checklist

### DIFF
--- a/src/rate_limiter/src/lib.rs
+++ b/src/rate_limiter/src/lib.rs
@@ -21,11 +21,10 @@
 //! The contract is that the user shall also call the `event_handler()` method on
 //! receipt of such an event.
 //!
-//! The token buckets are replenished every time a `consume()` is called, before
-//! actually trying to consume the requested amount of tokens. The amount of tokens
-//! replenished is automatically calculated to respect the `complete_refill_time`
-//! configuration parameter provided by the user. The token buckets will never
-//! replenish above their respective `size`.
+//! The token buckets are replenished when a called `consume()` doesn't find enough
+//! tokens in the bucket. The amount of tokens replenished is automatically calculated
+//! to respect the `complete_refill_time` configuration parameter provided by the user.
+//! The token buckets will never replenish above their respective `size`.
 //!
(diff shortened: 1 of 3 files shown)


alindima (approved):
Great job!

luminitavoicu (approved):

PR #2605, opened 2021-06-02
Consume RTC crate from rust-vmm/vm-superio

AlexandruCihodaru:
# Reason for This PR

Use upstream https://github.com/rust-vmm components.

## Description of Changes

This patch series migrates Firecracker from using local RTC implementation, to using the one from rust-vmm/vm-superio crate.

## License Acceptance

## PR Checklist


src/devices/src/legacy/rtc_pl031.rs
```diff
@@ -1,336 +1,73 @@
...
-            match self.handle_read(offset) {
-                Ok(val) => byte_order::write_le_u32(data, val),
-                Err(e) => {
-                    warn!("Failed to read from the RTC: {}", e);
-                    METRICS.rtc.error_count.inc();
-                }
-            }
+            // read() expects a slice of len 4, and we just validated
+            // that this is the data length
+            self.read(offset as u16, data.try_into().unwrap())
```
sandreim, line 14: I would clarify the comments above: we are calling the RTC implementation. `RTC::read()`

src/vmm/src/builder.rs
```diff
@@ -634,9 +640,9 @@ pub fn setup_serial_device(
 
 #[cfg(target_arch = "aarch64")]
 /// Sets up the RTC device.
-pub fn setup_rtc_device() -> super::Result<Arc<Mutex<devices::legacy::Rtc>>> {
-    let rtc = Arc::new(Mutex::new(devices::legacy::Rtc::default()));
-    Ok(rtc)
+pub fn setup_rtc_device() -> super::Result<Arc<Mutex<RTCDevice>>> {
+    let rtc = RTC::with_events(METRICS.rtc.clone());
+    Ok(Arc::new(Mutex::new(rtc)))
```
sandreim, line 645: Can this function return an error ?

src/logger/src/metrics.rs
```diff
@@ -626,6 +630,19 @@ pub struct RTCDeviceMetrics {
     pub missed_write_count: SharedIncMetric,
 }
 
+#[cfg(target_arch = "aarch64")]
+impl RTCEvents for RTCDeviceMetrics {
+    fn invalid_read(&self) {
+        self.missed_read_count.inc();
+        warn!("Invalid read from RTC device.")
```
sandreim, line 637: I would make the error specific: `Guest read/write at invalid offset`. For debug purposes it would help to also have the offset here.

src/devices/src/legacy/rtc_pl031.rs
```diff
@@ -1,336 +1,73 @@
...
-    // This is used for implementing the RTC alarm. However, in Firecracker we do not need it.
-    match_value: u32,
-    // Writes to this register load an update value into the RTC.
-    load: u32,
-    imsc: u32,
-    ris: u32,
-}
-
-impl Rtc {
-    /// Constructs an AMBA PL031 RTC device.
```
sandreim, line 63: I would keep this comment in the Firecracker codebase, even if the implementation lives elsewhere.

src/vmm/src/builder.rs
AlexandruCihodaru: 🤔 not really, will remove Result

dianpopa (approved):

src/devices/src/legacy/rtc_pl031.rs
```diff
@@ -0,0 +1,84 @@
...
+pub type RTCDevice = vm_superio::RTC<Arc<RTCDeviceMetrics>>;
+
+// Implements Bus functions for AMBA PL031 RTC device
+#[cfg(target_arch = "aarch64")]
+impl BusDevice for RTCDevice {
+    fn read(&mut self, offset: u64, data: &mut [u8]) {
+        if data.len() == 4 {
+            // read() function from RTC implementation expects a slice of
+            // len 4, and we just validated that this is the data lengt
+            self.read(offset as u16, data.try_into().unwrap())
```
alindima, line 19: This cast can truncate the data. Should we use a try_from and log an error if the data does not fit?

src/logger/src/metrics.rs
```diff
@@ -778,8 +798,9 @@ pub struct FirecrackerMetrics {
     pub patch_api_requests: PatchRequestsMetrics,
     /// Metrics related to API PUT requests.
     pub put_api_requests: PutRequestsMetrics,
+    #[cfg(target_arch = "aarch64")]
     /// Metrics related to the RTC device.
-    pub rtc: RTCDeviceMetrics,
+    pub rtc: Arc<RTCDeviceMetrics>,
```
alindima, line 805: why do we require an Arc?

src/devices/src/legacy/rtc_pl031.rs
andreeaflorescu: If the guest sends an invalid offset or even an invalid data, it means that the driver is misbehaving. In this case, what is expected from the emulation is "undefined behavior" (i.e. in the emulation we choose what to do, the driver cannot have any expectations of what happens next). For this reason, it might be better to just not add additional checks and keep a cleaner code.

andreeaflorescu: The reason is also that these values are checked again when calling read.

src/logger/src/metrics.rs
AlexandruCihodaru: Because we METRICS.rtc is moved when we create a new RTC device. RTCDeviceMetrics does not implement Clone trait because SharedIncMetrics can not implement Cloine trait because AtomicUsize does not implement it, so the solution would be to use an arc

src/devices/src/legacy/rtc_pl031.rs
alindima: > If the guest sends an invalid offset or even an invalid data, it means that the driver is misbehaving. In this case, what is expected from the emulation is "undefined behavior" (i.e. in the emulation we choose what to do, the driver cannot have any expectations of what happens next). For this reason, it might be better to just not add additional checks and keep a cleaner code.

I understand that. The logs are however for Firecracker users, and it could help with debugging some obscure cases.
I agree though that the code is a little cleaner this way

alindima: > The reason is also that these values are checked again when calling read.

When calling read on the RTC from superio? The upstream read works on u16 so the truncation check cannot be done there.

src/logger/src/metrics.rs
alindima: I see. I just find it a little weird to have an Arc into a static variable. The static will live for the entire duration of the program.

To get rid of this, I would just implement the RTCEvents for a dummy struct, that doesn't contain any members, because we can still access the Metrics global variable.

We can consider this a nit

alindima (approved):

src/logger/src/metrics.rs
AlexandruCihodaru: Would be a little weird to increment a dummy metric on rtc implementation and then the real one.

alindima: There would not be any dummy metric increment. Just the trait implementation would be done on another struct. The function implementations would still use the real metrics, just not from `self`, but by accessing the global metrics object.

src/devices/src/legacy/rtc_pl031.rs
andreeaflorescu: What I meant with the check is something else :laughing: I'll try to explain it better.

So let's say that instead of an u16, the driver sends you an u64. We're truncating it, ending up with a u16, which is not going to represent any meaningful value. It's just going to be garbage. This garbage value cannot do anything bad tho because only "good" offsets are allowed into the emulation. That's what I mean with the value being checked in vm-superio.

We should also be mindful about logging too many things that come from untrusted users even if we rate limit the log file.

PR #2617, opened 2021-06-10
prod-host-setup: recommendation on async-unsafe signal handlers

alindima:
# Reason for This PR

https://github.com/firecracker-microvm/firecracker/issues/2452

## Description of Changes

- add a note and recommendation in the prod-host-setup.

## License Acceptance

## PR Checklist

### DIFF
--- a/docs/prod-host-setup.md
+++ b/docs/prod-host-setup.md
@@ -59,6 +59,22 @@ logging, which degraded the snapshot restore time from 3ms to 8.5ms on
 `aarch64`. In this case, creating the tap device for snapshot restore
 generated host kernel logs, which were very slow to write.
 
+### Logging and signal handlers
+
+Firecracker installs custom signal handlers for some of the POSIX signals, such
+as SIGSEGV, SIGSYS, etc.
+
+The custom signal handlers used by Firecracker are not async-signal-safe, since
+they write logs and flush the metrics, which use locks for synchronization.
+While very unlikely, it is possible that the handler will intercept a signal on
+a thread which is already holding a lock to the log or metrics buffer.
+This can result in a deadlock, where the specific Firecracker thread becomes
+unresponsive.
+
+While there is no security impact caused by the deadlock, we recommend that
+customers have an overwatcher process on the host, that periodically looks
+for Firecracker processes that are unresponsive, and kills them, by SIGKILL.
+
 ## Jailer Configuration
 
 Using Jailer in a production Firecracker deployment is highly recommended,


docs/prod-host-setup.md
```diff
@@ -59,6 +59,20 @@ logging, which degraded the snapshot restore time from 3ms to 8.5ms on
...
 generated host kernel logs, which were very slow to write.
 
+### Logging and signal handlers
+
+Firecracker installs custom signal handlers for some of the POSIX signals, such
+as SIGSEGV, SIGSYS, etc.
+The custom signal handlers used by Firecracker are not async-signal-safe, since
+they write logs and flush the metrics, which use locks for synchronization.
+There is a very low probability, that, when running the signal handler on a
+thread, that thread was already holding a lock to the log or metrics buffers.
```
georgepisaltu, line 69: This sentence is hard to read, I'll suggest this but feel free to rephrase however you see fit.

```diff
- There is a very low probability, that, when running the signal handler on a
- thread, that thread was already holding a lock to the log or metrics buffers.
+ It is possible but very unlikely that the handler will intercept a signal on a 
+ thread which is already holding a lock to the log or metrics buffer.
```

docs/prod-host-setup.md
```diff
@@ -59,6 +59,20 @@ logging, which degraded the snapshot restore time from 3ms to 8.5ms on
 `aarch64`. In this case, creating the tap device for snapshot restore
 generated host kernel logs, which were very slow to write.
 
+### Logging and signal handlers
+
+Firecracker installs custom signal handlers for some of the POSIX signals, such
+as SIGSEGV, SIGSYS, etc.
+The custom signal handlers used by Firecracker are not async-signal-safe, since
```
georgepisaltu, line 66: If you meant to add a newline here, you need to add one more, like

```diff
- as SIGSEGV, SIGSYS, etc.
+ as SIGSEGV, SIGSYS, etc.
+ 
```

docs/prod-host-setup.md
```diff
@@ -59,6 +59,20 @@ logging, which degraded the snapshot restore time from 3ms to 8.5ms on
...
+they write logs and flush the metrics, which use locks for synchronization.
+There is a very low probability, that, when running the signal handler on a
+thread, that thread was already holding a lock to the log or metrics buffers.
+This can result in a deadlock, where the specific Firecracker thread becomes
+unresponsive.
+While there is no security impact cause by the unlikely deadlock, we recommend
```
georgepisaltu, line 72: Same as above regarding the newline.

georgepisaltu, line 72:
```diff
- While there is no security impact cause by the unlikely deadlock, we recommend
+ While there is no security impact caused by the deadlock, we recommend
```

docs/prod-host-setup.md
```diff
@@ -59,6 +59,20 @@ logging, which degraded the snapshot restore time from 3ms to 8.5ms on
...
+that customers have an overwatcher process on the host, that periodically looks
+for Firecracker processes that are un-responsive, and kills them, by SIGKILL.
```
georgepisaltu, line 74: Do we have a recommendation for the specific method used to determine if the process is unresponsive? If we do, maybe it's worth mentioning here.

Also

```diff
- for Firecracker processes that are un-responsive, and kills them, by SIGKILL.
+ for Firecracker processes that are unresponsive and kills them with SIGKILL.
```

alindima: There are two cases to think about:

**1. a thread is deadlocked on flushing metrics**

If another thread than the VMM is deadlocked on metrics, the VMM thread will also get deadlocked in at most 1 minute (which is the metric flushing period).
We are now sure that the VMM thread is deadlocked as well, if one of the threads was deadlocked.
To detect that the VMM is deadlocked, one can either have an agent inside the VM that sends some data periodically to an overwatcher, or simply have an overwatcher poll the instance state periodically (which will trigger some code on the VMM thread, which is deadlocked, so it will timeout).

**2. a thread is deadlocked on writing a log**

if the VMM thread is blocked logging, any of the approaches in the previous point would work (polling instance state or having an agent that uses an emulated device, like net, vsock).
If the API thread is blocked while logging, the approach with polling the instance state also works.
Let's think what happens is a vcpu is deadlocked. It must have performed a VMEXIT, because the vcpu was interrupted by a signal while logging. If a vcpu is blocked in a VMEXIT, sooner or later, a lock-up will be detected by the guest (if it has more than one vcpus). If it's the only vcpu of the VM, having an agent inside the VM will help a customer detect this.

Hopefully my thinking makes sense.
To summarise, I would recommend that the overwatcher does this (assuming that Firecracker metrics are enabled and the guest has a lockup detector):
1. Poll instance state
2. if step 1 succeeded, proceed to ping the guest over network/vsock.

These steps are to be repeated periodically, on a sensible period.

How does this sound?

alindima: Note that the above makes sense to implement only if customers don't already have systems that already do this.

alindima: @firecracker-microvm/compute-capsule what are you thoughts?

georgepisaltu: > Note that the above makes sense to implement only if customers don't already have systems that already do this.

This is a very good point and I think we should start our recommendation with this if we decide to include the suggestion.

> if step 1 succeeded, proceed to ping the guest over network/vsock

I don't think forcing customers to use the net/vsock devices to make sure the VM is responsive is good, as they are both optional devices. I thought we could go a simpler route and just drop a quick suggestion, like for example trying to run any API request which interacts with the VMM emulation thread (a `VmmAction`, like a pause request). If there is an interval of time of up to 1 minute in which these requests might work even though the VM is doomed, I don't see how we present this as a reasonable option. Maybe just monitor the VMM thread with a tool for idle time? I have no clear solution, so I am fine with not including this.

acatangiu (approved):

georgepisaltu (approved):

PR #2629, opened 2021-06-18
Block device fixes and simplifications

serban300:
# Reason for This PR

Block device fixes and simplifications

## Description of Changes

Block device fixes and simplifications:
- fix the write_bytes metric
- fix the size of the block request status
- simplify the request status logic

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/request.rs
+++ b/src/devices/src/virtio/block/request.rs
@@ -49,14 +49,16 @@ impl Status {
         }
     }
 
-    pub fn virtio_blk_status(&self) -> u32 {
-        match self {
+    pub fn virtio_blk_status(&self) -> u8 {
+        let virtio_blk_status = match self {
             Status::Ok(_) => VIRTIO_BLK_S_OK,
             Status::Err(status) => match status {
                 ErrStatus::IoErr(_) => VIRTIO_BLK_S_IOERR,
                 ErrStatus::Unsupported(_) => VIRTIO_BLK_S_UNSUPP,
             },
-        }
+        };
+
+        virtio_blk_status as u8
     }
 
     pub fn num_used_bytes(&self) -> u32 {
@@ -367,7 +369,7 @@ mod tests {
     fn test_status() {
         {
             let status = Status::from_result(Ok(10));
-            assert_eq!(status.virtio_blk_status(), VIRTIO_BLK_S_OK);
+            assert_eq!(status.virtio_blk_status(), VIRTIO_BLK_S_OK as u8);
             assert_eq!(status.num_used_bytes(), 11);
         }
 
@@ -375,7 +377,7 @@ mod tests {
             let status = Status::from_result(Err(ErrStatus::IoErr(IoErrStatus::BadRequest(
                 Error::InvalidOffset,
             ))));
-            assert_eq!(status.virtio_blk_status(), VIRTIO_BLK_S_IOERR);
+            assert_eq!(status.virtio_blk_status(), VIRTIO_BLK_S_IOERR as u8);
             assert_eq!(status.num_used_bytes(), 1);
         }
 
@@ -383,7 +385,7 @@ mod tests {
             let status = Status::from_result(Err(ErrStatus::IoErr(IoErrStatus::Flush(
                 io::Error::from_raw_os_error(42),
             ))));
-            assert_eq!(status.virtio_blk_status(), VIRTIO_BLK_S_IOERR);
+            assert_eq!(status.virtio_blk_status(), VIRTIO_BLK_S_IOERR as u8);
             assert_eq!(status.num_used_bytes(), 1);
         }
 
@@ -392,7 +394,7 @@ mod tests {
                 0,
                 GuestMemoryError::InvalidBackendAddress,
             ))));
-            assert_eq!(status.virtio_blk_status(), VIRTIO_BLK_S_IOERR);
+            assert_eq!(status.virtio_blk_status(), VIRTIO_BLK_S_IOERR as u8);
             assert_eq!(status.num_used_bytes(), 1);
         }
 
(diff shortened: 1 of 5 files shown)


src/devices/src/virtio/block/request.rs
```diff
@@ -178,67 +178,72 @@ impl Request {
         Ok(req)
     }
 
+    fn execute_seek(&self, disk: &mut DiskProperties) -> result::Result<(), ExecuteError> {
+        // Check that the data length is a multiple of 512 as specified in the virtio standard.
+        // We could also check this at request parsing level.
+        if u64::from(self.data_len) & (SECTOR_SIZE - 1) != 0 {
```
alindima, line 217: same, I think this check is erroneous. It doesn't return true for 513 for example

alindima: I feel like these checks don't belong in the `execute_seek` method. I think they are more suited in the `Request` parsing logic, WDYT?

src/devices/src/virtio/block/request.rs
```diff
@@ -183,22 +183,27 @@ impl Request {
...
-        }
-        top = top
-            .checked_add(self.sector)
-            .ok_or(ExecuteError::BadRequest(Error::InvalidOffset))?;
-        if top > disk.nsectors() {
-            return Err(ExecuteError::BadRequest(Error::InvalidOffset));
+        if self.request_type == RequestType::In || self.request_type == RequestType::Out {
+            // Check that the data length is a multiple of 512 as specified in the virtio standard.
+            // We could also check this at request parsing level.
+            if u64::from(self.data_len) & (SECTOR_SIZE - 1) != 0 {
```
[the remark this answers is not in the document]
serban300: In theory `SECTOR_SIZE - 1 == b11111111` and any number that is not divisible with `512 (==b1000000000)` should be caught here since it must have one of the lower 8 bits set. I tested and `513_u64 & (SECTOR_SIZE - 1) == 1`

src/devices/src/virtio/block/request.rs
serban300: I agree that all these checks fit better inside `Request::Parse`. They even return an `ExecuteError::BadRequest` which should be specific to `Request::Parse`. It's just that it was easier to simply add this check here for the moment as an incremental improvement. It still simplifies the request status logic. And I would move them to `Request::Parse` in a separate PR. I'm not sure it's worth creating an issue for this. For the moment I just left a comment.

src/devices/src/virtio/block/request.rs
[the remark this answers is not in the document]
alindima: Sorry, I misread this as `SECTOR_SHIFT`, which is 9 😆

src/devices/src/virtio/block/request.rs
alindima: Sounds ok for now

alindima (dismissed):

src/devices/src/virtio/block/request.rs
```diff
@@ -18,28 +18,61 @@ use super::device::{CacheType, DiskProperties};
...
     Flush(io::Error),
-    Read(GuestMemoryError),
+    Read(u32, GuestMemoryError),
     Seek(io::Error),
     SyncAll(io::Error),
     Write(GuestMemoryError),
+}
+
+#[derive(Debug)]
+pub enum ErrStatus {
```
georgepisaltu, line 31: nit: I would rename this to something like `StatusError`. Actually all `Err*` stuff.

src/devices/src/virtio/block/request.rs
```diff
@@ -18,28 +18,61 @@ use super::device::{CacheType, DiskProperties};
...
-            ExecuteError::Read(_) => VIRTIO_BLK_S_IOERR,
-            ExecuteError::Seek(_) => VIRTIO_BLK_S_IOERR,
-            ExecuteError::SyncAll(_) => VIRTIO_BLK_S_IOERR,
-            ExecuteError::Write(_) => VIRTIO_BLK_S_IOERR,
-            ExecuteError::Unsupported(_) => VIRTIO_BLK_S_UNSUPP,
+#[derive(Debug)]
+pub enum Status {
+    Ok(u32),
+    Err(ErrStatus),
+}
```
georgepisaltu, line 40: nit: This looks like a glorified `Result<u32, ErrStatus>` with the upside that you can define your own methods on it. Is it worth it? Why not just process the contents of an actual `Result` or name the enum variants differently?

also nit: We could have a comment explaining that the `u32` is in fact `num_used_bytes`.

src/devices/src/virtio/block/request.rs
```diff
@@ -178,52 +211,85 @@ impl Request {
         Ok(req)
     }
 
+    fn execute_seek(&self, disk: &mut DiskProperties) -> result::Result<(), ErrStatus> {
+        // Check that the data length is a multiple of 512 as specified in the virtio standard.
+        // We could also check this at request parsing level.
+        if u64::from(self.data_len) & (SECTOR_SIZE - 1) != 0 {
```
georgepisaltu, line 217: nit: I believe the compiler optimizes the `%` operation anyway, so why not use that to make it more clear?

georgepisaltu (approved):

src/devices/src/virtio/block/request.rs
serban300: I think it's worth it since the possibility to define our own methods improves the readability. I didn't see a performance penalty in my tests.

I added a comment.

src/devices/src/virtio/block/request.rs
serban300: Yes, it doesn't seem to affect the performance. Done.

src/devices/src/virtio/block/request.rs
georgepisaltu: I commented this after reading `if let Status::Err(err_status) = status`, which was a bit confusing. When extracting an error, I expect something like `if let Err(e) = ...`. It's just a cosmetic thing at this point, so it's ok.

georgepisaltu (approved):

alindima (approved):

Issue #2635, opened 2021-06-25
[Perf] Use virtio event notification supression for the block device


serban300:
Subtask for #2478

serban300:
I posted a Darft PR: #2637 

There are more details and performance results on the PR

serban300:
Resolved by #2658

PR #2637, opened 2021-06-25
[RFC] [block] add support for notification suppression

serban300:
# Reason for This PR

#2635

### Important note

In the current RFC we try to batch requests and then send only 1 notification to the guest if needed. This is not fully compliant with the virtio standard which states that we should send a notification after processing each descriptor if `queue.needs_notification()` returns true: https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-400007 . But the current approach seems to work.

### Performance Results

**Test setup:** We run on a m5d.metal host with Ubuntu and kernel 5.5.19. We start a number of microVMs with 8vCPUs and 16GB of RAM running linux kernel v5.5.19 as a guest OS. Each microVMs is pinned to NUMA node 0. We mount a ramdisk. For each microVM we create an ext4 file on the ramdisk and we attach it to the microVM as /dev/vdb. We want to have the content of the drives in the page cache in order to asses the performance improvement brought by notification suppression. Otherwise we would be limited by the disk access speed. Inside the VM we run the following command:

```
fio --name=test --ioengine=libaio --rw=readwrite --randrepeat=1 --bs=4k --size=2G --numjobs=1 --iodepth=32 --runtime=30 --time_base=1 --end_fsync=1 --direct=1 --filename=/dev/vdb
```

Current implementation:
|description|IOPS|VMM thread CPU load|kworker threads CPU load|IOPS per CPU load|
|-|-|-|-|-|
|1 VMs|98.96|0.37|0.12|203.43||
|4 VMs|372.89|1.44|0.45|197.69||
|8 VMs|773.69|3.1|0.87|194.8||
|16 VMs|1400.15|6.6|2.03|162.33||
|32 VMs|2189.28|8.38|2.08|209.14||
|48 VMs|1377.53|7.0|1.96|153.83||

Notification suppression:
|description|IOPS|VMM thread CPU load|kworker threads CPU load|IOPS per CPU load|
|-|-|-|-|-|
|1 VMs|147.67|0.4|0.07|313.05||
|4 VMs|580.98|1.62|0.27|307.34||
|8 VMs|1163.95|3.42|0.55|292.82||
|16 VMs|2092.06|7.48|1.18|241.71||
|32 VMs|3587.84|10.45|1.13|309.81||
|48 VMs|1772.47|7.96|1.78|182.01||

If we look at the `IOPS per CPU load` metric, we observe an improvement of about 50% for the most cases when using notification suppression. There's a notable exception in the last case (48 VMs) where the performance improvement is only about 
20%. It would be worth investigating why this happens. If so, we'll do it in a different issue.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -16,7 +16,7 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 
 use logger::{error, warn, IncMetric, METRICS};
-use rate_limiter::{BucketUpdate, RateLimiter, TokenType};
+use rate_limiter::{BucketUpdate, RateLimiter};
 use utils::eventfd::EventFd;
 use virtio_gen::virtio_blk::*;
 use vm_memory::{Bytes, GuestMemoryMmap};
@@ -283,40 +283,24 @@ impl Block {
             // This should never happen, it's been already validated in the event handler.
             DeviceState::Inactive => unreachable!(),
         };
+
         let has_notification_suppression = self.has_feature(u64::from(VIRTIO_RING_F_EVENT_IDX));
-        let queue = &mut self.queues[queue_index];
-        let mut used_any = false;
         let mut needs_notification = false;
+        let queue = &mut self.queues[queue_index];
+        if queue.is_empty(mem) {
+            METRICS.block.no_avail_buffer.inc();
+            return needs_notification;
+        }
+
         while let Some(head) = queue.pop(mem) {
             let len = match Request::parse(&head, mem) {
                 Ok(request) => {
-                    // If limiter.consume() fails it means there is no more TokenType::Ops
-                    // budget and rate limiting is in effect.
-                    if !self.rate_limiter.consume(1, TokenType::Ops) {
+                    if request.rate_limit(&mut self.rate_limiter) {
                         // Stop processing the queue and return this descriptor chain to the
                         // avail ring, for later processing.
                         queue.undo_pop();
                         METRICS.block.rate_limiter_throttled_events.inc();
                         break;
-                    }
-                    // Exercise the rate limiter only if this request is of data transfer type.
-                    if request.request_type == RequestType::In
-                        || request.request_type == RequestType::Out
-                    {
-                        // If limiter.consume() fails it means there is no more TokenType::Bytes
-                        // budget and rate limiting is in effect.
-                        if !self
-                            .rate_limiter
-                            .consume(u64::from(request.data_len), TokenType::Bytes)
-                        {
-                            // Revert the OPS consume().
-                            self.rate_limiter.manual_replenish(1, TokenType::Ops);
-                            // Stop processing the queue and return this descriptor chain to the
-                            // avail ring, for later processing.
-                            queue.undo_pop();
-                            METRICS.block.rate_limiter_throttled_events.inc();
-                            break;
-                        }
                     }
 
                     let status = Status::from_result(request.execute(&mut self.disk, mem));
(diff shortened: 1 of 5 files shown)


src/devices/src/virtio/block/device.rs
```diff
@@ -344,17 +333,21 @@ impl Block {
...
-        }
 
-        if !used_any {
-            METRICS.block.no_avail_buffer.inc();
+            queue.set_avail_event(mem);
+            needs_notification |= queue.needs_notification(mem, has_notification_suppression);
         }
 
-        used_any
+        // We try to batch requests and then send only 1 notification to the guest.
```
sandreim, line 341: The purpose of the suppression mechanism is to reduce the overhead. The approach here basically does something similar on top it. Can this have any negative side effects like increased latency or stalls ?

src/devices/src/virtio/persist.rs
```diff
@@ -43,6 +43,10 @@ pub struct QueueState {
 
     next_avail: Wrapping<u16>,
     next_used: Wrapping<u16>,
+
+    /// The number of added used buffers since last guest kick
+    #[version(start = 2)]
+    num_added: Wrapping<u16>,
```
sandreim, line 49: You should also update the version map to account for this new QueueState version.

src/devices/src/virtio/queue.rs
```diff
@@ -393,6 +398,56 @@ impl Queue {
...
+
+    /// Check if we need to kick the guest.
+    /// This is similar to the `vring_need_event()` method implemented by the Linux kernel.
+    pub fn needs_notification(
+        &self,
+        mem: &GuestMemoryMmap,
+        has_notification_suppression: bool,
+    ) -> bool {
+        // If the device doesn't use notification suppression, always return true
+        if !has_notification_suppression {
```
sandreim, line 437: Can we promote `has_notification_suppression` to member of Queue, or otherwise perform this check at caller level ?

src/devices/src/virtio/queue.rs
```diff
@@ -393,6 +398,56 @@ impl Queue {
...
+        }
+
+        let new = self.next_used;
+        let old = self.next_used - self.num_added;
+        let used_event = self.used_event(mem);
+
+        new - used_event - Wrapping(1) < new - old
+    }
+
+    pub fn after_notification(&mut self) {
```
sandreim, line 448: This fn is missing documentation.

src/devices/src/virtio/block/device.rs
```diff
@@ -344,17 +333,21 @@ impl Block {
...
+        // This is not fully compliant with the virtio standard which states that we should
+        // send a notification after processing each descriptor if `queue.needs_notification()`
+        // returns true:
+        // https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-400007
+        // But the current approach seems to work.
```
acatangiu, line 346: What's the reasoning for deviating from the standard?

src/devices/src/virtio/persist.rs
serban300: Done.

src/devices/src/virtio/queue.rs
serban300: Done.

src/devices/src/virtio/queue.rs
serban300: Fixed.

serban300:
Closing in favour of #2658 . I removed batching for the moment in order to be fully virtio compliant.

PR #2638, opened 2021-06-25
Add snapshot restore performance tests for scalable configurations

georgepisaltu:
# Reason for This PR

The current snapshot restore tests are not enough to understand the impact on the restore latency of various variables in the VM configuration.

## Description of Changes

Added snapshot restore tests which measure the restore latency of VMs with variable configurations. The tests are scaling VCPU count, memory size, net device count and block device count, each individually over a base configuration of 1 VCPU 128MiB. We also test a configuration with the vsock and balloon devices enabled.

The tests are built on top of the existing performance test framework and come with a set of baselines, detailed in a config JSON file.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/framework/artifacts.py
+++ b/tests/framework/artifacts.py
@@ -480,6 +480,11 @@ class Snapshot:
         merge_memory_bitmaps(base.mem, self.mem)
         self._mem = base.mem
 
+    def cleanup(self):
+        """Delete the backing files from disk."""
+        os.remove(self._mem)
+        os.remove(self._vmstate)
+
     @property
     def mem(self):
         """Return the mem file path."""
--- a/tests/framework/matrix.py
+++ b/tests/framework/matrix.py
@@ -7,6 +7,8 @@ of the cartesian product of all artifact sets.
 """
 
 import os
+import sys
+import traceback
 from framework.artifacts import ARTIFACTS_LOCAL_ROOT
 from framework.utils import ExceptionAggregator
 
@@ -115,8 +117,10 @@ class TestMatrix:
         if len(self._sets) == len(cartesian_product):
             try:
                 self._run_test_fn(cartesian_product, test_fn)
-            except Exception as err:  # pylint: disable=W0703
-                self._failure_aggregator.add_row(err)
+            except Exception as _err:  # pylint: disable=W0703
+                self._failure_aggregator.add_row(
+                    "".join(traceback.format_exception(*sys.exc_info()))
+                )
             return
 
         current_set = self._sets[self._set_index]
--- a/tests/framework/stats/function.py
+++ b/tests/framework/stats/function.py
@@ -163,8 +163,9 @@ class Percentile(Function, ABC):
         length = len(result)
         result.sort()
         idx = length * self.k / 100
-        if idx is not int(idx):
-            return (result[int(idx)] + result[(int(idx) + 1)]) / 2
+        if not idx.is_integer():
+            return (result[int(idx)] + result[min((int(idx) + 1),
+                    length - 1)]) / 2
 
         return result[int(idx)]
 
--- a/tests/integration_tests/performance/configs/snap_restore_test_config.json
+++ b/tests/integration_tests/performance/configs/snap_restore_test_config.json
@@ -1,0 +1,537 @@
+{
+  "measurements": {
+    "restore_latency": {
+      "unit": "ms",
+      "statistics": [
+        {
+          "name": "P50",
+          "function": "Percentile50",
+          "criteria": "EqualWith"
+        },
+        {
+          "name": "P90",
+          "function": "Percentile90",
+          "criteria": "EqualWith"
+        }
+      ]
+    }
+  },
+  "hosts": {
+    "instances": {
+      "m5d.metal": {
+        "baselines": {
+          "restore_latency": {
+            "vmlinux-4.14.bin": {
+              "ubuntu-18.04.ext4": {
+                "1vcpu_128mb": {
+                  "P50": {
+                    "target": 5,
+                    "delta_percentage": 9
+                  },
+                  "P90": {
+                    "target": 6,
+                    "delta_percentage": 37
+                  }
+                },
+                "2vcpu_128mb": {
+                  "P50": {
+                    "target": 5,
+                    "delta_percentage": 9
+                  },
+                  "P90": {
+                    "target": 6,
+                    "delta_percentage": 68
+                  }
+                },
+                "3vcpu_128mb": {
+                  "P50": {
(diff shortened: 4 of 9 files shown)


tests/framework/stats/function.py
```diff
@@ -163,8 +163,9 @@ def __call__(self, result: List) -> Any:
         length = len(result)
         result.sort()
         idx = length * self.k / 100
-        if idx is not int(idx):
-            return (result[int(idx)] + result[(int(idx) + 1)]) / 2
+        if idx != float(int(idx)):
```
iulianbarbu, line 166: The `float` conversion is not necessary. 

nit: If find this logic more clear when using the `float` `is_integer` dedicated method. Multiple occurrences of `int(idx)` can be avoided as well by binding its result to a variable.

tests/framework/matrix.py
```diff
@@ -115,8 +117,10 @@ def _backtrack(self, test_fn, cartesian_product):
         if len(self._sets) == len(cartesian_product):
             try:
                 self._run_test_fn(cartesian_product, test_fn)
-            except Exception as err:  # pylint: disable=W0703
-                self._failure_aggregator.add_row(err)
+            except Exception as _err:  # pylint: disable=W0703
+                self._failure_aggregator.add_row(
+                    "".join(traceback.format_exception(*sys.exc_info()))
```
iulianbarbu, line 122: Very nice!

tests/integration_tests/performance/configs/snap_restore_test_config.json
```diff
@@ -0,0 +1,538 @@
...
+                    "P50": {
+                      "target": 67.457,
+                      "delta_percentage": 20
+                    },
+                    "P90": {
+                      "target": 69.69,
+                      "delta_percentage": 20
+                    }
+                  },
+                  "2net_dev": {
```
iulianbarbu, line 206: What about this identifier? Would we find useful the information about the vm config?

tests/integration_tests/performance/test_snapshot_restore_performance.py
```diff
@@ -0,0 +1,364 @@
...
+                             guest_ip="192.168.2.2",
+                             tap_name="tap2",
+                             dev_name="eth2"),
+              NetIfaceConfig(host_ip="192.168.3.1",
+                             guest_ip="192.168.3.2",
+                             tap_name="tap3",
+                             dev_name="eth3")]
+
+# We are using this as a global variable in order to only
+# have to call the constructor and destructor once.
```
iulianbarbu, line 54: Why shouldn't this be created alongside the test matrix, attached to its context and piped along to any scope that needs it?

iulianbarbu (commented):
Besides the nits I've covered in this review, I am wondering how long this test takes on a metal and if it has been run for multiple times, to gather some reliability data against the proposed baselines.

alindima (commented):
I think we should rename the `test_snapshot_restore_performance.py` to something else since it's quite confusing that we have another file named `test_snapshot_perf.py`.

Do you think it's worth having a separate folder: `integration_tests/nightly`? (in a future PR)
If we do this, I'm fine with keeping the name as it is

tests/framework/artifacts.py
```diff
@@ -480,6 +480,11 @@ def rebase_snapshot(self, base):
         merge_memory_bitmaps(base.mem, self.mem)
         self._mem = base.mem
 
+    def delete_files(self):
```
gc-plp, line 483: This could be called as part of the deconstructor to avoid explicit calls.

georgepisaltu (commented):
@iulianbarbu @gc-plp @alindima I've addressed your feedback, PTAL!

tests/integration_tests/performance/configs/snap_restore_test_config.json
```diff
@@ -0,0 +1,537 @@
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
alindima, line 537: still need a newline :)

georgepisaltu: My bad, commits on multiple branches got messed up.

tests/integration_tests/performance/test_snapshot_restore_performance.py
georgepisaltu: I had to revert this since no other option aside from the global variable seems to work. I tried putting it in the context and pass it along, and it would work for one kernel but on the second one the list would be unpopulated. I also tried generating it for every test that needed it without a global variable, but a local one, but then the destructor seems to get called before the test runs and the paths are invalid by the time the VM tries to use them. I must admit I don't actually know what exactly is causing this behavior, I am just guessing here. I did make the code that generates the mock drives prettier though :smile:

alindima (dismissed):
LGTM 🚢

tests/integration_tests/performance/test_snapshot_restore_performance.py
```diff
@@ -0,0 +1,360 @@
...
+    basevm.start()
+
+    ssh_connection = net_tools.SSHConnection(basevm.ssh_config)
+
+    # Create a snapshot builder from a microvm.
+    snapshot_builder = SnapshotBuilder(basevm)
+    snapshot = snapshot_builder.create(
+        [rw_disk.local_path()] + extra_disk_paths,
+        ssh_key,
+        SnapshotType.FULL,
```
dianpopa, line 171: nit: since this test is only for restoring full snapshots, would make sense to name the file or the function accordingly?

dianpopa (dismissed):

tests/framework/artifacts.py
georgepisaltu: Done.

tests/integration_tests/performance/test_snapshot_restore_performance.py
```diff
@@ -0,0 +1,360 @@
...
+# pylint: disable=C0103
+scratch_drives = []
+
+
+# pylint: disable=R0903
+class SnapRestoreBaselinesProvider(BaselineProvider):
+    """Baselines provider for snapshot restore latency."""
+
+    def __init__(self, env_id):
+        """Block baseline provider initialization."""
```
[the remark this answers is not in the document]
georgepisaltu: Done.

tests/integration_tests/performance/test_snapshot_restore_performance.py
georgepisaltu: Done.

alindima (approved):

dianpopa (approved):

PR #2642, opened 2021-07-01
Add capability to export full configuration through API endpoint

### LINKED ISSUE #2356: [Feature Request] Add capability to dump the Firecracker configuration

acatangiu:
# Reason for This PR

Fixes #2356 

## Description of Changes

Added capability to generate configuration JSON from actual resource/device/etc.
Added API endpoint to generate full VM configuration and provide as HTTP JSON response.

## License Acceptance

## PR Checklist


CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 - Added code logic to send `VIRTIO_VSOCK_EVENT_TRANSPORT_RESET` on snapshot
   creation, when the Vsock device is active. The event will close active
   connections on the guest.
+- Added `GET` request on `/export-vm-config` that provides full microVM
+  configuration as a JSON HTTP response.
```
sandreim, line 26: Can the config be used as JSON to start FC ?

src/api_server/src/parsed_request.rs
```diff
@@ -59,6 +59,9 @@ impl ParsedRequest {
         match (request.method(), path, request.body.as_ref()) {
             (Method::Get, "", None) => parse_get_instance_info(),
             (Method::Get, "balloon", None) => parse_get_balloon(path_tokens.get(1)),
+            (Method::Get, "export-vm-config", None) => {
+                Ok(ParsedRequest::new_sync(VmmAction::GetFullVmConfig))
```
sandreim, line 63: Nit: GetVmConfig

src/devices/src/virtio/net/device.rs
```diff
@@ -217,9 +218,31 @@ impl Net {
         self.guest_mac.as_ref()
     }
 
-    /// Provides a mutable reference to the `MmdsNetworkStack`.
-    pub fn mmds_ns_mut(&mut self) -> Option<&mut MmdsNetworkStack> {
-        self.mmds_ns.as_mut()
+    /// Provides the host IFACE name of this net device.
+    pub fn iface_name(&self) -> String {
```
sandreim, line 222: Instead of adding these many member fns, I would look at defining a trait like `ConfigDump` and implement it for all the objects we want to export the configuration as JSON. We have a similar pattern with the `Persist` trait.

tests/integration_tests/functional/test_api.py
```diff
@@ -980,6 +982,104 @@ def test_api_balloon(test_microvm_with_ssh_and_balloon):
     assert test_microvm.api_session.is_status_bad_request(response.status_code)
 
 
+def test_get_full_config(test_microvm_with_ssh_and_balloon):
```
sandreim, line 985: We should also add a test to cover that PATCH reflects changes in the exported config.

sandreim (commented):
Looks good overall. I would rename the endpoint to simply `config`.

CHANGELOG.md
acatangiu: Yes, it's using the same JSON serialized struct as the `--config-file` param uses 👍

src/devices/src/virtio/net/device.rs
acatangiu: This is what I initially went for, but it evolved into current form because:

Current Firecracker design/architecture has JSON/API configuration/description structs defined as a submodule of `vmm` crate as we considered them to be vmm-specific. The idea was to keep devices more generic and not tie them to the configuration model and methods.
So `devices` crate should not depend on these _configuration_ struct definitions.

We could define the trait in the config, then implement the trait for all devices, but I lean towards implementing `std::convert::From` to get _&device -> config_ instead of a custom trait in the same place with same functionality and usage.

The added member fns are simple immutable getters, I'm not aware of any disadvantage in having them.

tests/integration_tests/functional/test_api.py
acatangiu: Right! Will do.

src/devices/src/virtio/net/device.rs
sandreim: Gotcha. The idea was to reduce LOC by removing the getters.

CHANGELOG.md
sandreim: Then, we also need a test that ensure this -> new VM -> dump config -> new VM using dumped config.

acatangiu: I wouldn't say we _need_ that test, but sure, will add.

georgepisaltu: > Then, we also need a test that ensure this -> new VM -> dump config -> new VM using dumped config.

Did you just uncover a new way of starting VMs? 😁

src/devices/src/virtio/net/device.rs
georgepisaltu: I like `From` and the `fn`s better than a custom trait.

georgepisaltu (commented):
Looking good!

CHANGELOG.md
sandreim: I think so ... an easy way to convert from "curl" based configuration to one shot json based startup 😁

acatangiu:
> Current RFC proposes GET /export-vm-config, but I'm not really happy with that. Especially since we also have:
    GET / to get vm instance information
    PATCH /vm to set vm state (paused/running)
I'm hoping for some unifying idea 😄

What about `GET /vm/config` ?

src/api_server/src/parsed_request.rs
acatangiu: `src/vmm/src/vmm_config/machine_config.rs` defines `VmConfig` struct which is also what we call `MachineConfig` (cpu and mem config).

I defined the complete/full VM config retrieval action as `GetFullVmConfig` to avoid confusion between the two..

CHANGELOG.md
acatangiu: done! 👍

tests/integration_tests/functional/test_api.py
acatangiu: done! 👍

src/rate_limiter/src/lib.rs
```diff
@@ -93,13 +93,18 @@ pub enum BucketReduction {
 pub struct TokenBucket {
     // Bucket defining traits.
     size: u64,
-    // Initial burst size (number of free initial tokens, that can be consumed at no cost)
-    one_time_burst: u64,
+    // Initial burst size.
+    initial_one_time_burst: u64,
```
luminitavoicu, line 97: It is not clear to me why this was added and how it differs from the existing `one_time_burst`. Would you mind providing more detail?

luminitavoicu (commented):
Changes look good. I am in favor of `GET /vm/config`, it's explicit and cleaner than `GET /export-vm-config`.

src/rate_limiter/src/lib.rs
acatangiu: `one_time_burst` is an internal member that is mutated during operation - we need to store its initial config value (new member `initial_one_time_burst` to be able to generate config from object.

luminitavoicu: Got it, thanks!

tests/integration_tests/functional/test_api.py
```diff
@@ -980,6 +1020,104 @@ def test_api_balloon(test_microvm_with_ssh_and_balloon):
...
+    assert response.json() == expected_cfg
+
+    # Start the microvm.
+    test_microvm.start()
+
+    # Validate full vm configuration post-boot as well.
+    response = test_microvm.full_cfg.get()
+    assert test_microvm.api_session.is_status_ok(response.status_code)
+    assert response.json() == expected_cfg
+
```
sandreim, line 1119: We can start a microvm from the json as a final test.

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 - Added code logic to send `VIRTIO_VSOCK_EVENT_TRANSPORT_RESET` on snapshot
   creation, when the Vsock device is active. The event will close active
   connections on the guest.
+- Added `GET` request on `/vm/config` that provides full microVM configuration
```
sandreim, line 25: Should we also add an example in the docs ?

sandreim (commented):
Looks good!

tests/integration_tests/functional/test_api.py
acatangiu: It'd be redundant, microvm is started from file json at the beginning; then file json is checked to be `==` with generated json; starting another from generated is equivalent to the first operation, thus redundant

CHANGELOG.md
acatangiu: I don't see any good place for this, also it's a simple GET with no body, not sure what the example would achieve.

luminitavoicu (approved):
Great job!

CHANGELOG.md
georgepisaltu: I agree that the example would be trivial.

tests/integration_tests/functional/test_api.py
```diff
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests that ensure the correctness of the Firecracker API."""
 
+# Disable pylint C0302: Too many lines in module
```
georgepisaltu, line 5: :disappointed:

tests/integration_tests/functional/test_cmd_line_start.py
```diff
@@ -54,6 +55,12 @@ def test_config_start_with_api(test_microvm_with_ssh, vm_config_file):
     assert test_microvm.api_session.is_status_ok(response.status_code)
     assert test_microvm.state == "Running"
 
+    # Validate full vm configuration.
+    response = test_microvm.full_cfg.get()
+    assert test_microvm.api_session.is_status_ok(response.status_code)
+    with open(vm_config_file) as json_file:
+        assert response.json() == json.load(json_file)
```
georgepisaltu, line 62: I think this should address @sandreim 's concerns about starting a VM from a full config obtained from another VM.

georgepisaltu (approved):
LGTM!

Issue #2639, opened 2021-06-28
[Bug] Metrics "utc_timestamp_ms" is not a UTC timestamp


alindima:
# Describe the bug

The `utc_timestamp_ms` field that is reported by the metrics uses a `CLOCK_MONOTONIC` timestamp. The POSIX standard defines the monotonic clock as a clock with a fixed but undefined starting point.
For linux, the starting point is usually set at boot.

I believe we should be using a `CLOCK_REALTIME`, which gives the interval from the UTC UNIX epoch.

There is no actual documentation of this metric field (https://github.com/firecracker-microvm/firecracker/issues/2189) but the name suggests that it should be based to the UTC UNIX epoch.

## Checks

- [x] Have you searched the Firecracker Issues database for similar problems?
- [x] Have you read the existing relevant Firecracker documentation?
- [x] Are you certain the bug being reported is a Firecracker issue?

alindima:
Fixed by #2646

PR #2646, opened 2021-07-07
fix the metrics utc_timestamp_ms which was using a monotonic clock

alindima:
# Reason for This PR

https://github.com/firecracker-microvm/firecracker/issues/2639

## Description of Changes

- replace the monotonic clock with a realtime clock
- add a Changelog entry
- add a regression test
- slight refactor of test_metrics.py

## License Acceptance

## PR Checklist


src/logger/src/metrics.rs
```diff
@@ -769,7 +769,7 @@ struct SerializeToUtcTimestampMs;
 impl Serialize for SerializeToUtcTimestampMs {
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         serializer.serialize_i64(
-            utils::time::get_time_ns(utils::time::ClockType::Monotonic) as i64 / 1_000_000,
+            utils::time::get_time_ns(utils::time::ClockType::Real) as i64 / 1_000_000,
```
acatangiu, line 772: I'm not sure I agree here. There are PROs and CONs to each:

`ClockType::Monotonic` doesn't have a standardized base (on Linux though it's boot-time), so it might be hard to match timestamps in Firecracker logs to timestamps in other logs if other logs use other clocktype than monotonic.

`ClockType::Real` provides the machine's best guess as to current wall-clock, time-of-day time. The problem with this, is that it can jump forwards and backwards as the system time-of-day clock is changed, including by NTP and/or daylight savings. This means any time-jump event will make firecracker logs very confusing.

acatangiu: I'm not sure which is better for our case. Can you check other components on the system to see what timestamps they use in their logs?

For example `systemd` seems to log both: https://unix.stackexchange.com/a/453258

alindima: As discussed offline, one option would be to add both realtime and monotonic timestamps to the metric data. We'll further discuss in the team the options we have and add a conclusion soon.

acatangiu: After looking here in more detail I see this is only related to a specific metric named `utc_timestamp_ms`. The value for this metric uses `monotonic` as time source so it's not really UTC time.

Given that metrics publishing happens every 60s using a monotonic timer, I think the PR is actually correct in simply changing value of this metric to use `realtime` source to provide correct UTC time.

acatangiu (approved):

georgepisaltu (approved):

Issue #2602, opened 2021-05-31
[Bug] microVM cleanup not triggered in the CI tests using logger thread


dianpopa:
# Describe the bug

By default, the instantiation of a microVM object will also create and start a thread that receives [as reference the microVM object](https://github.com/firecracker-microvm/firecracker/blob/main/tests/framework/microvm.py#L805). This results in the microVM destructor never being called and thus cleanup is skipped.

## To Reproduce

```
./tools/devtool shell -p && cd tests/`
# check the contents of the /srv/ directory (default location for test resources)
du -sh /srv/
pytest integration_tests/functional/test_api.py::test_negative_api_lifecycle
du -sh /srv/*
# 387M	/srv/jailer
# the jailer directory has not gone through the cleanup process
```

## Expected behaviour

The `/srv/jailer/firecracker` exists but does not contain any microvm leftover resources.

## Environment

No specific environment needed.

## Additional context

Move the logging thread at the microVM level or one level above it so that we do not need to reference it in the function running the logging thread. This would also help us with being able to check the microVM logs after the destruction of the microVM.

## Checks

- [x] Have you searched the Firecracker Issues database for similar problems?
- [x] Have you read the existing relevant Firecracker documentation?
- [x] Are you certain the bug being reported is a Firecracker issue?

alindima:
I've taken a quick look over this. Indeed it appears that the microvm destructors in the test framework are not called, because firecracker pids still show up in `ps` after the test finished executing.

However, the microvm destructor doesn't have any logic for cleaning up the resources in `/srv/jailer/firecracker/`. Maybe this should be done, but I think the bigger problem is that the destructors are not being called.

alindima:
I don't think this has anything to do with the logger thread. I managed to pinpoint the fact that this only happens when we're not using pytest fixtures.
In the tests where we leverage fixtures for microvms, the destructor gets called.

In the test outlined above, `integration_tests/functional/test_api.py::test_negative_api_lifecycle`, we're using a `MicrovmBuilder`. We're using it in some more tests, especially the performance and snapshot tests.

I'll try to see why the destructor doesn't get called in this case.

dianpopa:
> I've taken a quick look over this. Indeed it appears that the microvm destructors in the test framework are not called, because firecracker pids still show up in `ps` after the test finished executing.

[...]

PR #2647, opened 2021-07-08
Add VM cleanup for tests using MicrovmBuilder

alindima:
# Reason for This PR

Fixes https://github.com/firecracker-microvm/firecracker/issues/2602

## Description of Changes

Explicitly calls `vm.kill()` on VmInstance destructor

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/framework/builder.py
+++ b/tests/framework/builder.py
@@ -27,7 +27,7 @@ class VmInstance:
         self._kernel = kernel
         self._disks = disks
         self._ssh_key = ssh_key
-        self._vm = vm
+        self.vm = vm
 
     @property
     def config(self):
@@ -49,10 +49,9 @@ class VmInstance:
         """Return ssh key artifact linked to the root block device."""
         return self._ssh_key
 
-    @property
-    def vm(self):
-        """Return the Microvm object instance."""
-        return self._vm
+    def __del__(self):
+        """Teardown the VM."""
+        self.vm.kill()
 
 
 class MicrovmBuilder:


tests/framework/builder.py
```diff
@@ -49,10 +49,9 @@ def ssh_key(self):
         """Return ssh key artifact linked to the root block device."""
         return self._ssh_key
 
-    @property
-    def vm(self):
-        """Return the Microvm object instance."""
-        return self._vm
+    def __del__(self):
+        """Teardown the VM."""
+        self._vm.kill()
```
sandreim, line 54:
```suggestion
        self.vm.kill()
```

acatangiu:
```suggestion
        self.vm.kill()
```

acatangiu (commented):
Please do manual run while _inside_ the container batches of each following tests:
- tests using fixture
- tests using MicrovmBuilder
- with and without the tests logging thread (which takes a reference to `microvm` object)

```
tools/devtool shell -p
cd tests
pytest integration_tests/functional/bla.py
```

While the tests are running, check for leftover resources by finished tests. I believe the microvm cleanup is not called when a tests finishes because its drop is not called because logging thread keeps it in scope.

The cleanup might be called when full batch of tests finishes during pytest teardown, but it should be called after each test for the resources used by that test.

tests/framework/builder.py
alindima: done, it seems that this wasn't caught by a test because python seems to ignore exceptions that happen inside destructors

alindima:
It seems that somehow, the VmInstance destructor is still getting called while the member Microvm object is referenced in the test.
This causes intermittent failures.

One option is to explicitly call vm.kill() at the end of all tests that use the MicrovmBuilder. Most of them already do this. WDYT?

alindima:
@dianpopa is right, this is an issue pertaining to the logger thread architecture, as the issue states. I'll close this PR and open a new one once I have the time to work on a better solution

PR #2655, opened 2021-07-12
use `--mount type=tmpfs` instead of mount ramdisk of the host

### LINKED ISSUE #2612: [Bug] devtool mount ramdisk does not fail when permission denied
Launching `./devtool/test -r` using a user that does not have the rights to mount filesystems, will not issue a fatal error:
```
[Firecracker devtool] Using ramdisk ...
mkdir: cannot create directory ‘/mnt/devtool-ramdisk’: Permission denied
mkdir: cannot create directory ‘/mnt/devtool-ramdisk’: Permission denied
mkdir: cannot create directory ‘/mnt/devtool-ramdisk’: Permission denied

```

Implementation here: https://github.com/firecracker-microvm/firecracker/blob/main/tools/devtool#L819

We need to do one of:
1. Err through devtool.
2. Create the ramdisk inside the container - this would make ramdisk creation transparent and not polluting the host environment.  Also it would work with multiple tests that use ramdisk.

bitoku:
# Reason for This PR

Close #2612 

The test suite fail to create a tmpfs when the user doesn't have appropriate permissions. As mentioned in #2612, docker has a feature of mounting tmpfs, so it doesn't have to create a tmpfs outside the container and mount it.
This replace the `mount_ramdisk` function to the docker option of mouting tmpfs, so it no longer needs a permission to do `mkdir`.

This is my first PR of this repository, so if you notice any mistakes, please let me know.

## Description of Changes

This PR removes mount_ramdisk function and adds `--mount type=tmpfs` option to the docker args.
The document of mounting tmpfs is below.
https://docs.docker.com/storage/tmpfs/

There is two options to mount tmpfs;
- `--mount type=tmpfs`
- `--tmpfs`

I used `--mount type=tmpfs` because `--tmpfs` doesn't allow us to use any other configuration such as `tmpfs-size`.

## License Acceptance

## PR Checklist

### DIFF
--- a/tools/devtool
+++ b/tools/devtool
@@ -816,21 +816,6 @@ cmd_strip() {
     return $ret
 }
 
-mount_ramdisk() {
-    say "Using ramdisk ..."
-    local ramdisk_size="$1"
-    umount_ramdisk
-    mkdir -p ${DEFAULT_RAMDISK_PATH} && \
-    mount -t tmpfs -o size=${ramdisk_size} tmpfs ${DEFAULT_RAMDISK_PATH}
-    mkdir -p ${DEFAULT_RAMDISK_PATH}/srv
-    mkdir -p  ${DEFAULT_RAMDISK_PATH}/tmp
-}
-
-umount_ramdisk() {
-    umount ${DEFAULT_RAMDISK_PATH} &>/dev/null
-    rmdir ${DEFAULT_RAMDISK_PATH} &>/dev/null
-}
-
 # `$0 test` - run integration tests
 # Please see `$0 help` for more information.
 #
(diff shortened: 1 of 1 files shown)


alindima (commented):
Changes look good, thank you for the contribution. Let me run some more tests using `devtool --ramdisk` argument to make sure that everything works just as before.
I'll get back to you soon

luminitavoicu:
Please make sure you sign your commit (have a line in the commit message that starts with `Signed-off-by`). You can achieve this by appending `-s` to the git commit command.

alindima:
@bitoku while the PR looks good, I've run the [long-running block performance tests](https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/performance/test_block_performance.py) we have in Firecracker, which use a ramdisk and the baselines we have are no longer correct. For some reason I can't pinpoint, the test results using your PR are different than the baselines.

There are two ways you can proceed here:
1. Investigate where the difference is coming from.
2. Use the alternative approach listed in the issue description: doing proper error-checking when creating the ramdisk on the host.

Given that this issue is quite small and doesn't have a big impact, we incline towards the second option. I'm afraid that a more thorough investigation is not worth it at this point.
What are your thoughts?

bitoku:
@alindima Thanks for checking this PR.
I completely agree with you.
Would it better to close this PR, and open another one that adds proper error-checking?

alindima:
> @alindima Thanks for checking this PR.
> I completely agree with you.
> Would it better to close this PR, and open another one that adds proper error-checking?

You can either do that or update this PR by force-pushing, whichever way you prefer

PR #2656, opened 2021-07-12
Add recipe for building kernel images used in the CI

dianpopa:
# Reason for This PR

Provide a tool for obtaining Firecracker tailored kernel images.

## Description of Changes

* the PR makes use of the tools present in [rust-vmm/vmm-reference repository](https://github.com/rust-vmm/vmm-reference/tree/master/resources/kernel) for building a kernel
* the way it uses the functionality is by downloading the make_kernel.sh script and then source it
* this PR is dependant on PR https://github.com/rust-vmm/vmm-reference/pull/149 getting merged

## License Acceptance

## PR Checklist


resources/microvm-kernel-5.2.1-x86_64
```diff
@@ -0,0 +1,2859 @@
+#
+# Automatically generated file; DO NOT EDIT.
+# Linux/x86 5.2.1 Kernel Configuration
```
serban300, line 3: Why do we need to add a config for kernel 5.2.1 ? It doesn't seem to be used.

tools/devtool
```diff
@@ -1432,6 +1448,70 @@ cmd_install() {
...
+
+    kernel_dir_host="$FC_BUILD_DIR"/kernel
+    kernel_dir_ctr="$CTR_FC_BUILD_DIR"/kernel
+    create_dir "$kernel_dir_host"
+
+    # Extract the kernel version from the config file provided as parameter.
+    KERNEL_VERSION=$(cat "$KERNEL_CFG" | grep -Po "^# Linux\/$cfg_pat (([0-9]+.){2}[0-9]+)" | cut -d ' ' -f 3)
+    validate_kernel_version "$KERNEL_VERSION"
+
+    recipe_url="https://raw.githubusercontent.com/dianpopa/vmm-reference/fc_down/resources/kernel/make_kernel.sh"
```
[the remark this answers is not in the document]
dianpopa: Indeed maybe I should have placed a comment here. The script that contains all the auxiliary methods for building a kernel is currently taken from my rust-vmm/vmm-reference fork since I did some changes there to work for aarch64. After [this PR is merged](https://github.com/rust-vmm/vmm-reference/pull/149), we will be able to download it directly from rust-vmm. The main goal here is to use as much as possible rust-vmm functionality.

dianpopa (commented):
@serban300 we do not need to add a 5.2.1 config. It is a leftover from when I was not able to make the balloon test work with a 4.14 config. I removed the commit adding it.

tools/devtool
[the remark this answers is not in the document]
serban300: I understand. Still, how do we plan to consume it from rust-vmm ? Will they publish it to an official repository ? If we could install it as a python package in our docker image it would also be awesome. 

I have concerns with just copying the script from the github project even if the project is trusted. Especially in this case where we only probably need 10-20 lines of code, I would rather copy the code to our project instead of downloading a script.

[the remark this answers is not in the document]
dianpopa: What do you mean by official repository? It is currently in vmm-reference. Do you mean like a repository that is only meant for building test resources? 

Making it available in our docker image is a really good idea indeed.

tools/devtool
```diff
@@ -1432,6 +1448,70 @@ cmd_install() {
...
+        die "Unsupported architecture!"
+    fi
+
+    kernel_dir_host="$FC_BUILD_DIR"/kernel
+    kernel_dir_ctr="$CTR_FC_BUILD_DIR"/kernel
+    create_dir "$kernel_dir_host"
+
+    # Extract the kernel version from the config file provided as parameter.
+    KERNEL_VERSION=$(cat "$KERNEL_CFG" | grep -Po "^# Linux\/$cfg_pat (([0-9]+.){2}[0-9]+)" | cut -d ' ' -f 3)
+    validate_kernel_version "$KERNEL_VERSION"
```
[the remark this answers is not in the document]
serban300: yes, `yes "" | make oldconfig` would generate a default config. Also, in case the provided config is fully compatible, it wouldn't alter it. In case it isn't fully compatible it would alter it, and we could check that and issue a warning. So we would reach a similar outcome as above. Seems more reliable to me then validating a comment that can be wrong or modified by the user. Also we would avoid regexes and pattern matching which I personally find hard to follow.

tools/devtool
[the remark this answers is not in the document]
serban300: Sorry, I mean a package registry. Something like `crates.io` for rust. I guess there must be something similar for python.

dianpopa:
[Changed needed for this PR](https://github.com/rust-vmm/vmm-reference/pull/149) were merged in rust-vmm. @firecracker-microvm/compute-capsule please review, thanks

tools/devtool
```diff
@@ -631,6 +631,17 @@ update_python_package_version() {
     docker rm -f "$dummy_container_name"
 }
 
+create_dir() {
```
gc-plp, line 634: Do we actually need to add this function? AFAIK we run the script with `set -e`.

tools/devtool
```diff
@@ -1637,6 +1653,73 @@ cmd_install() {
...
+
+    kernel_dir_host="$FC_BUILD_DIR"/kernel
+    kernel_dir_ctr="$CTR_FC_BUILD_DIR"/kernel
+    create_dir "$kernel_dir_host"
+
+    # Extract the kernel version from the config file provided as parameter.
+    KERNEL_VERSION=$(cat "$KERNEL_CFG" | grep -Po "^# Linux\/$cfg_pattern (([0-9]+.){2}[0-9]+)" | cut -d ' ' -f 3)
+    validate_kernel_version "$KERNEL_VERSION"
+
+    rust_vmm_recipe_url="https://raw.githubusercontent.com/rust-vmm/vmm-reference/fc_down/resources/kernel/make_kernel.sh"
```
gc-plp, line 1705: Do we have any mechanisms in place to make sure that we can still run this file if the rust-vmm upstream changes?

What about fixing it to a commit ID?

tools/devtool
dianpopa: This piece of code is used in multiple places in devtool: like [here](https://github.com/firecracker-microvm/firecracker/blob/main/tools/devtool#L285) or [here](https://github.com/firecracker-microvm/firecracker/blob/main/tools/devtool#L1546). My thinking was to get rid of this duplicate code so I will also replace those occurrences.  I do not see any `set -e` in devtool...
i am thinking of posting an update with the deduplication of `mkdir -p ...`. What do you think?

tools/devtool
dianpopa: Fixing it to a commit ID sounds good to me. @serban300 what are your thoughts on this solution?

serban300: Sorry, but personally in this particular case I still don't see a great value in reusing the code from rust-vmm and I would still just copy it since there are just a few lines from what I can tell:

```
get linux sources (with wget or clone)
copy the config
yes "" | make oldconfig
make -j [num_configed_threads] vmlinux (or make Image if we're on an aarch64 platform)
```

Also being so simple I don't see any fixes, improvements, or features that will make this script more appealing in the future.

It's true that fixing the script to a particular commit is more reliable, but it still has downsides. For example that commit can be deleted even if it's very unlikely.

andreeaflorescu: We are not deleting commits in rust-vmm, that is for sure :D.

Consuming it from the rust-vmm/vmm-reference crate is not the best outcome for the long term, but it looks like a good enough solution in the short term until we work towards the higher goal of having reusable scripts for building resources that can be reused by the multiple projects.

dianpopa: Cool, good to know commits are not going to be deleted. thanks @andreeaflorescu . Indeed, vmm-reference is not bound to remain the crate where these test resources are going to be kept. I do think that for the long term we it is best to put these in a separate dedicated repo. However, consuming this code from rust-vmm will ensure we will reach the best version of this script, one that will hopefully be consumed by as many rust powered vms out there. 

@firecracker-microvm/compute-capsule can I get your opinion on this too?

tools/devtool
dianpopa: I updated the code, let me know what you think

tools/devtool
dianpopa: @gc-plp I fixed the script to a hardcoded commit, PTAL!

gc-plp (dismissed):

tools/devtool
```diff
@@ -661,6 +667,11 @@ cmd_help() {
     echo "        Builds the development container from its Dockerfile."
     echo "        -n, --no-python-package-update  Do not update python packages."
     echo ""
+    echo "    build_kernel -c|--config [-n|--nproc]"
+    echo "        Builds a kernel version based on the config file provide as parameter."
```
alindima, line 671: Maybe it'd help to display the fact that this is used for the Firecracker CI, not for production.

alindima (commented):
I see that [this PR](https://github.com/rust-vmm/vmm-reference/pull/149) was merged, do we now want to  bundle the scripts in the docker container and use them from there?

docs/rootfs-and-kernel-setup.md
```diff
@@ -47,6 +49,23 @@ can boot:
...
+using the recipe inside devtool:
+
+```bash
+./tools/devtool build_kernel -c ./resources/microvm-kernel-x86.config -n 8
+```
+
+or
+
+```bash
+./tools/devtool build_kernel -c ./resources/microvm-kernel-arm64.config -n 8
```
alindima, line 64: nit: it'd be even easier to use if we added the paths to the configs as default parameters. wdyt?

tools/devtool
dianpopa: Ok, done.

docs/rootfs-and-kernel-setup.md
dianpopa: Indeed i thought about adding a default location for the config but I gave up the idea. My skepticism comes from the fact that we are not (as of yet) running this recipe with every PR and as a consequence, we run the risk of breaking it if the config file gets renamed or relocated. If you still think it would add value to define default paths to the config, let me know/

dianpopa (commented):
> I see that [this PR](https://github.com/rust-vmm/vmm-reference/pull/149) was merged, do we now want to bundle the scripts in the docker container and use them from there?

Indeed this idea popped up a while ago but it did not have much traction. My only fear regarding this is that baking it into the container will force us to  update the container image every time we need to change the script. Let me know if you have a strong preference for this.

alindima (approved):

tools/devtool
serban300: Anyway this is not a blocker for me and otherwise the PR looks good.

serban300 (approved):

PR #2658, opened 2021-07-13
[block] add support for notification suppression

serban300:
# Reason for This PR

#2635 

### Performance Results

**Test setup:** We run on a m5d.metal host with Ubuntu and kernel 5.8.18. We start a number of microVMs with 8vCPUs and 16GB of RAM running linux kernel v5.8.18 as a guest OS. Each microVMs is pinned to NUMA node 1. We mount a ramdisk and we pin it to NUMA node 1 as well. For each microVM we create an ext4 file on the ramdisk and we attach it to the microVM as /dev/vdb. We want to have the content of the drives in the page cache in order to asses the performance improvement brought by notification suppression. Otherwise we would be limited by the disk access speed. Inside the VM we run the following command:

```
fio --name=test --ioengine=libaio --rw=readwrite --randrepeat=1 --bs=4k --size=500m --numjobs=1 --iodepth=32 --runtime=30 --time_base=1 --end_fsync=1 --direct=1 --filename=/dev/vdb
```

Current implementation:
|description|IOPS|VMM thread CPU load|kworker threads CPU load|IOPS per CPU load|
|-|-|-|-|-|
|1 VMs|104.86|0.36|0.13|213.5|
|4 VMs|439.28|1.51|0.52|216.34|
|8 VMs|849.83|3.1|1.07|203.47|
|16 VMs|1524.45|6.76|2.24|169.51|
|32 VMs|2202.07|8.32|2.19|209.51|
|64 VMs|1182.12|5.76|1.77|156.89|

Notification suppression:
|description|IOPS|VMM thread CPU load|kworker threads CPU load|IOPS per CPU load|increase in efficiency|
|-|-|-|-|-|-|
|1 VMs|160.72|0.42|0.09|315.65|47.85%|
|4 VMs|599.36|1.64|0.36|299.79|38.57%|
|8 VMs|1175.92|3.33|0.81|283.79|39.48%|
|16 VMs|1872.87|7.07|2.12|203.68|20.16%|
|32 VMs|2551.24|8.67|2.61|226.2|7.97%|
|64 VMs|1369.84|6.07|1.79|174.26|11.07%|

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -30,6 +30,7 @@ use super::{
 use crate::virtio::{IrqTrigger, IrqType};
 
 use serde::{Deserialize, Serialize};
+use virtio_gen::virtio_ring::VIRTIO_RING_F_EVENT_IDX;
 
 /// Configuration options for disk caching.
 #[derive(Clone, Copy, Debug, Deserialize, PartialEq, Serialize)]
@@ -216,7 +217,7 @@ impl Block {
     ) -> io::Result<Block> {
         let disk_properties = DiskProperties::new(disk_image_path, is_disk_read_only, cache_type)?;
 
-        let mut avail_features = 1u64 << VIRTIO_F_VERSION_1;
+        let mut avail_features = (1u64 << VIRTIO_F_VERSION_1) | (1u64 << VIRTIO_RING_F_EVENT_IDX);
 
         if cache_type == CacheType::Writeback {
             avail_features |= 1u64 << VIRTIO_BLK_F_FLUSH;
@@ -279,7 +280,7 @@ impl Block {
 
         let queue = &mut self.queues[queue_index];
         let mut used_any = false;
-        while let Some(head) = queue.pop(mem) {
+        while let Some(head) = queue.pop_or_enable_notification(mem) {
             let len = match Request::parse(&head, mem, self.disk.nsectors()) {
                 Ok(request) => {
                     if request.rate_limit(&mut self.rate_limiter) {
@@ -320,16 +321,19 @@ impl Block {
                     head.index, e
                 )
             });
+
+            if queue.prepare_kick(mem) {
+                self.irq_trigger
+                    .trigger_irq(IrqType::Vring)
+                    .unwrap_or_else(|_| {
+                        METRICS.block.event_fails.inc();
+                    });
+            }
+
             used_any = true;
         }
 
-        if used_any {
-            self.irq_trigger
-                .trigger_irq(IrqType::Vring)
-                .unwrap_or_else(|_| {
-                    METRICS.block.event_fails.inc();
-                });
-        } else {
+        if !used_any {
             METRICS.block.no_avail_buffer.inc();
         }
     }
@@ -458,6 +462,13 @@ impl VirtioDevice for Block {
     }
 
     fn activate(&mut self, mem: GuestMemoryMmap) -> ActivateResult {
+        let event_idx = self.has_feature(u64::from(VIRTIO_RING_F_EVENT_IDX));
+        if event_idx {
+            for queue in &mut self.queues {
+                queue.enable_notif_suppression();
+            }
+        }
+
         if self.activate_evt.write(1).is_err() {
             error!("Block: Cannot write to activate_evt");
             return Err(super::super::ActivateError::BadActivate);
@@ -523,7 +534,7 @@ pub(crate) mod tests {
 
         assert_eq!(block.device_type(), TYPE_BLOCK);
 
-        let features: u64 = 1u64 << VIRTIO_F_VERSION_1;
+        let features: u64 = (1u64 << VIRTIO_F_VERSION_1) | (1u64 << VIRTIO_RING_F_EVENT_IDX);
 
         assert_eq!(block.avail_features_by_page(0), features as u32);
         assert_eq!(block.avail_features_by_page(1), (features >> 32) as u32);
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -16,7 +16,7 @@ use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
 
 use logger::{error, warn, IncMetric, METRICS};
-use rate_limiter::{BucketUpdate, RateLimiter, TokenType};
+use rate_limiter::{BucketUpdate, RateLimiter};
 use utils::eventfd::EventFd;
 use virtio_gen::virtio_blk::*;
 use vm_memory::{Bytes, GuestMemoryMmap};
@@ -282,33 +282,12 @@ impl Block {
         while let Some(head) = queue.pop(mem) {
             let len = match Request::parse(&head, mem, self.disk.nsectors()) {
                 Ok(request) => {
-                    // If limiter.consume() fails it means there is no more TokenType::Ops
-                    // budget and rate limiting is in effect.
-                    if !self.rate_limiter.consume(1, TokenType::Ops) {
+                    if request.rate_limit(&mut self.rate_limiter) {
                         // Stop processing the queue and return this descriptor chain to the
                         // avail ring, for later processing.
                         queue.undo_pop();
                         METRICS.block.rate_limiter_throttled_events.inc();
                         break;
-                    }
-                    // Exercise the rate limiter only if this request is of data transfer type.
-                    if request.request_type == RequestType::In
-                        || request.request_type == RequestType::Out
-                    {
-                        // If limiter.consume() fails it means there is no more TokenType::Bytes
-                        // budget and rate limiting is in effect.
-                        if !self
-                            .rate_limiter
-                            .consume(u64::from(request.data_len), TokenType::Bytes)
-                        {
-                            // Revert the OPS consume().
-                            self.rate_limiter.manual_replenish(1, TokenType::Ops);
-                            // Stop processing the queue and return this descriptor chain to the
-                            // avail ring, for later processing.
-                            queue.undo_pop();
-                            METRICS.block.rate_limiter_throttled_events.inc();
-                            break;
-                        }
                     }
 
                     let status = Status::from_result(request.execute(&mut self.disk, mem));
@@ -499,6 +478,7 @@ pub(crate) mod tests {
 
     use super::*;
     use crate::virtio::queue::tests::*;
+    use rate_limiter::TokenType;
     use utils::tempfile::TempFile;
     use vm_memory::GuestAddress;
 
--- a/src/devices/src/virtio/queue.rs
+++ b/src/devices/src/virtio/queue.rs
@@ -303,6 +303,35 @@ impl Queue {
             return None;
         }
 
+        self.do_pop_unchecked(mem)
+    }
+
+    /// Try to pop the first available descriptor chain from the avail ring.
+    /// If no descriptor is available, enable notifications.
+    pub fn pop_or_enable_notification<'a, 'b>(
+        &'a mut self,
+        mem: &'b GuestMemoryMmap,
+    ) -> Option<DescriptorChain<'b>> {
+        if !self.uses_notif_suppression {
+            return self.pop(mem);
+        }
+
+        if self.try_enable_notification(mem) {
+            return None;
+        }
+
+        self.do_pop_unchecked(mem)
+    }
+
+    /// Pop the first available descriptor chain from the avail ring.
+    ///
+    /// # Important
+    /// This is an internal method that ASSUMES THAT THERE ARE AVAILABLE DESCRIPTORS. Otherwise it
+    /// will retrieve a descriptor that contains garbage data (obsolete/empty).
+    fn do_pop_unchecked<'a, 'b>(
+        &'a mut self,
+        mem: &'b GuestMemoryMmap,
+    ) -> Option<DescriptorChain<'b>> {
         // This fence ensures all subsequent reads see the updated driver writes.
         fence(Ordering::Acquire);
 
@@ -683,6 +712,33 @@ pub(crate) mod tests {
             .unwrap();
         assert!(!d.has_next());
         assert!(d.next_descriptor().is_none());
+
+        // Undoing the last pop should let us walk the last chain again.
+        q.undo_pop();
+        assert_eq!(q.len(m), 1);
+
+        // Walk the last chain again (three descriptors) using pop_or_enable_notification().
+        let d = q
+            .pop_or_enable_notification(m)
+            .unwrap()
+            .next_descriptor()
+            .unwrap()
+            .next_descriptor()
+            .unwrap();
+        assert!(!d.has_next());
+        assert!(d.next_descriptor().is_none());
+
+        // There are no more descriptors, but notification suppression is disabled.
+        // Calling pop_or_enable_notification() should simply return None.
+        assert_eq!(q.avail_event(m), 0);
+        assert!(q.pop_or_enable_notification(m).is_none());
+        assert_eq!(q.avail_event(m), 0);
+
+        // There are no more descriptors and notification suppression is enabled. Calling
+        // pop_or_enable_notification() should enable notifications.
+        q.enable_notif_suppression();
+        assert!(q.pop_or_enable_notification(m).is_none());
+        assert_eq!(q.avail_event(m), 2);
     }
 
     #[test]
--- a/src/devices/src/virtio/queue.rs
+++ b/src/devices/src/virtio/queue.rs
@@ -186,6 +186,11 @@ pub struct Queue {
 
     pub(crate) next_avail: Wrapping<u16>,
     pub(crate) next_used: Wrapping<u16>,
+
+    /// VIRTIO_F_RING_EVENT_IDX negotiated (notification suppression enabled)
+    pub(crate) uses_notif_suppression: bool,
+    /// The number of added used buffers since last guest kick
+    pub(crate) num_added: Wrapping<u16>,
 }
 
 #[allow(clippy::len_without_is_empty)]
@@ -201,6 +206,8 @@ impl Queue {
             used_ring: GuestAddress(0),
             next_avail: Wrapping(0),
             next_used: Wrapping(0),
+            uses_notif_suppression: false,
+            num_added: Wrapping(0),
         }
     }
 
@@ -371,6 +378,7 @@ impl Queue {
         mem.write_obj(len as u32, len_addr)
             .map_err(QueueError::UsedRing)?;
 
+        self.num_added += Wrapping(1);
         self.next_used += Wrapping(1);
 
         // This fence ensures all descriptor writes are visible before the index update is.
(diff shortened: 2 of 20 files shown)


serban300:
I made some small adjustments to the notification suppression functionality and also implemented the logic that breaks snapshot compatibility with previous versions of Firecracker. The PR is now ready for review. @acatangiu @sandreim @firecracker-microvm/compute-capsule PTAL when you have time !

src/devices/src/virtio/queue.rs
```diff
@@ -296,6 +303,32 @@ impl Queue {
             return None;
         }
 
+        self.do_pop(mem)
+    }
+
+    /// Try to pop the first available descriptor chain from the avail ring.
+    /// If no descriptor is available, enable notifications.
+    pub fn pop_or_enable_notification<'a, 'b>(
```
sandreim, line 311: Do we want to have a future proof approach here so that we end up in an ideal state of being able to quickly switch to rust-vmm implementation ? https://github.com/rust-vmm/vm-virtio/blob/master/crates/virtio-queue/src/lib.rs

sandreim (commented):
These changes are properly structured and look good. I've left some comments. I have only one concern related to diverging from the current rust-vmm state of virtio queue implementation. We should try to be in sync with the upstream design so that it easier to switch. For example I think the Iterators semantics are more appropriate and easier to read or get wrong than queue.pop(). 

What's your thinking around the effort of sync this PR to rust-vmm ?

src/devices/src/virtio/queue.rs
serban300: Yes, of course. Actually this implementation is my interpretation of an idea suggested by @alexandruag who works closely with the rust-vmm community. I'm not sure how easy it would be to implement this in an iterator, but this way of enabling notifications simplifies the devices implementation by a lot. Otherwise in the devices we would have something like:

```
outter': loop {
    while let Some(chain) = queue.pop() {
        if rate_limit(chain) {
            break outter;
        }

        process(chain);
    }

    if !queue.enable_notification() {
        break;
    }
}
``` 

Especially in the net device this kind of implementation would be really hard to follow and I think it would lead to complications.

src/devices/src/virtio/queue.rs
```diff
@@ -393,6 +401,64 @@ impl Queue {
...
+    /// Fetch the used event (`virtq_avail->used_event`) from guest memory.
+    /// This field is used when notification suppression is enabled. It is written by the driver,
+    /// in order to specify how far the device can progress before a notification is required.
+    #[inline(always)]
+    pub fn used_event(&self, mem: &GuestMemoryMmap) -> Wrapping<u16> {
+        // We need to find the `used_event` field from the avail ring.
+        let used_event_addr = 4 + 2 * self.actual_size();
+
+        // Complete all the writes in add_used() before reading the event.
+        fence(Ordering::SeqCst);
```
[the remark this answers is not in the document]
serban300: AFAIK `Ordering::Acquire` is not enough. Here we need a store-load barrier (we need to make sure that the store to `used_idx` was completed before the following load. So we need to actually stop the current CPU until all the stores are visible to all other CPUs. This can be done only with a SeqCst barrier as far as i know. `Ordering::Acquire` only provides a load-load barrier when coupled with a  `Ordering::Release` fence.

src/devices/src/virtio/queue.rs
serban300: As we discussed offline, this shouldn't break compatibility with rust-vmm.

tests/integration_tests/functional/test_snapshot_version.py
```diff
@@ -186,3 +153,48 @@ def test_create_invalid_version(bin_cloner_path):
...
+    _snapshot_dir = snapshot_builder.create_snapshot_dir()
+
+    # Pause microVM for snapshot.
+    response = test_microvm.vm.patch(state='Paused')
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+
+    # We try to create a snapshot to a target version < 0.26.0.
+    # This should fail because Fc versions < 0.26.0 don't support
+    # virtio notification suppression.
+    response = test_microvm.snapshot.create(
```
sandreim, line 185: We could test v0.23 - v0.25 here to have full coverage.

src/devices/src/virtio/queue.rs
```diff
@@ -393,6 +427,82 @@ impl Queue {
...
+
+    /// Try to enable notification events from the guest driver. Returns true if notifications were
+    /// successfully enabled. Otherwise it means that one or more descriptors can still be consumed
+    /// from the available ring and we can't guarantee that there will be a notification. In this
+    /// case the caller might want to consume the mentioned descriptors and call this method again.
+    pub fn try_enable_notification(&mut self, mem: &GuestMemoryMmap) -> bool {
+        // If the device doesn't use notification suppression, we'll continue to get notifications
+        // no matter what.
+        if !self.uses_notif_suppression {
+            return true;
```
sandreim, line 459: Why are we double checking this ? In pop_or_enable_notification() we are already doing this.

src/devices/src/virtio/queue.rs
```diff
@@ -393,6 +427,82 @@ impl Queue {
...
+        // Make sure all subsequent reads are performed after `set_avail_event`.
+        fence(Ordering::SeqCst);
+
+        // If the actual avail_idx is different than next_avail one or more descriptors can still
+        // be consumed from the available ring.
+        self.next_avail.0 == self.avail_idx(mem).0
+    }
+
+    /// Enable/disable notification suppression.
+    pub fn toggle_notif_suppression(&mut self, enabled: bool) {
```
sandreim, line 478: Is there any usecase where we would want to disable ?

src/devices/src/virtio/queue.rs
```diff
@@ -393,6 +427,82 @@ impl Queue {
...
+    }
+
+    /// Check if we need to kick the guest.
+    ///
+    /// Please note this method has side effects: once it returns `true`, it considers the
+    /// driver will actually be notified, and won't return `true` again until the driver
+    /// updates `used_event` and/or the notification conditions hold once more.
+    ///
+    /// This is similar to the `vring_need_event()` method implemented by the Linux kernel.
+    pub fn prepare_kick(&mut self, mem: &GuestMemoryMmap) -> bool {
```
sandreim, line 489: Nit: should we rename to `needs_kick` ?

src/devices/src/virtio/queue.rs
serban300: Both methods are public so they can be used independently, and they need to treat this case differently.

src/devices/src/virtio/queue.rs
serban300: No. I changed the method to `enable_notification()`.

src/devices/src/virtio/queue.rs
serban300: IMO `needs_kick` might imply that we are performing only a check, but we are actually also changing the state here. That's why I named it `prepare_kick`. But if you have strong preferences I can rename it.

src/vmm/src/persist.rs
```diff
@@ -297,23 +306,39 @@ fn snapshot_memory_to_file(
...
+
+    if data_version < FC_V0_26_SNAP_VERSION {
+        vmm.mmio_device_manager
+            .for_each_virtio_device(|_id, _info, dev| {
+                if dev
+                    .lock()
+                    .expect("Poisoned lock")
+                    .has_feature(u64::from(VIRTIO_RING_F_EVENT_IDX))
+                {
+                    return Err(CreateSnapshotError::IncompatibleDeviceFeatures);
```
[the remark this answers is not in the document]
serban300: Done. I added a string to it in order to mention the feature that is causing the problem.

tests/integration_tests/functional/test_snapshot_version.py
serban300: Done.

sandreim (dismissed):

src/devices/src/virtio/queue.rs
```diff
@@ -393,6 +401,82 @@ impl Queue {
...
+
+        // Make sure all subsequent reads are performed after `set_avail_event`.
+        fence(Ordering::SeqCst);
+
+        // If the actual avail_idx is different than next_avail one or more descriptors can still
+        // be consumed from the available ring.
+        self.next_avail.0 == self.avail_idx(mem).0
+    }
+
+    /// Enable/disable notification suppression.
```
acatangiu, line 477: nit: incorrect doc: this is not a toggle, only enables never disables

src/devices/src/virtio/queue.rs
```diff
@@ -393,6 +401,82 @@ impl Queue {
...
+        }
+
+        // We need to expose used array entries before checking the used_event.
+        fence(Ordering::SeqCst);
+
+        let new = self.next_used;
+        let old = self.next_used - self.num_added;
+        let used_event = self.used_event(mem);
+
+        self.num_added = Wrapping(0);
```
acatangiu, line 502: just an idea: since this is the only state change (side effect) of this function, would it make sense to change this fn to `needs_kick()` which only does checking (with no side effects) and do `self.num_added = 0` in actual `do_kick()` logic?

src/devices/src/virtio/queue.rs
```diff
@@ -303,6 +303,32 @@ impl Queue {
...
+
+        self.do_pop(mem)
+    }
+
+    /// Pop the first available descriptor chain from the avail ring.
+    ///
+    /// # Important
+    /// This is an internal method that ASSUMES THAT THERE ARE AVAILABLE DESCRIPTORS. Otherwise it
+    /// will retrieve a descriptor that contains garbage data (obsolete/empty).
+    fn do_pop<'a, 'b>(&'a mut self, mem: &'b GuestMemoryMmap) -> Option<DescriptorChain<'b>> {
```
acatangiu, line 331: nit: rename `do_pop_unchecked()`

src/devices/src/virtio/block/device.rs
```diff
@@ -320,16 +321,19 @@ impl Block {
...
                 )
             });
+
+            if queue.prepare_kick(mem) {
+                self.irq_trigger
+                    .trigger_irq(IrqType::Vring)
+                    .unwrap_or_else(|_| {
+                        METRICS.block.event_fails.inc();
+                    });
+            }
```
acatangiu, line 331: This introduces a regression for cases where notif suppression is not enabled/negotiated: with this new code, IRQs are being generated _on each descriptor_ instead of just one IRQ after the full chunk of currently avail descriptors.

I am expecting this regression to be actually visible in perf testing with notif suppres disabled/unavailable in the guest.

src/vmm/src/device_manager/mmio.rs
```diff
@@ -360,6 +361,32 @@ impl MMIODeviceManager {
         Ok(())
     }
 
+    /// Run fn for each registered device.
```
acatangiu, line 364:
```suggestion
    /// Run fn for each registered virtio device.
```

src/vmm/src/device_manager/mmio.rs
```diff
@@ -360,6 +361,32 @@ impl MMIODeviceManager {
...
+    pub fn for_each_virtio_device<F, E>(&self, mut f: F) -> std::result::Result<(), E>
```
acatangiu, line 365: You can use this newly added specialized fn to simplify code in `fn kick_devices()` below and possibly even `fn with_virtio_device_with_id()`

src/vmm/src/persist.rs
```diff
@@ -32,10 +32,12 @@ use seccompiler::BpfThreadMap;
 use snapshot::Snapshot;
 use versionize::{VersionMap, Versionize, VersionizeResult};
 use versionize_derive::Versionize;
+use virtio_gen::virtio_ring::VIRTIO_RING_F_EVENT_IDX;
 use vm_memory::GuestMemoryMmap;
 
 #[cfg(target_arch = "x86_64")]
 const FC_V0_23_SNAP_VERSION: u16 = 1;
+const FC_V0_26_SNAP_VERSION: u16 = 4;
```
acatangiu, line 40: how does the version encoding numbering work? just looking at this code here it looks like we're jumping from version `1` to version `4`.. What am I missing, and how can we make this clearer for future PRs?

(this comment can also be addressed in a separate PR with its own issue)

src/vmm/src/version_map.rs
```diff
@@ -41,6 +45,7 @@ lazy_static! {
 
         mapping.insert(String::from("0.24.0"), 2);
         mapping.insert(String::from("0.25.0"), 3);
+        mapping.insert(String::from("0.26.0"), 4);
```
acatangiu, line 48: I see here the full mapping. Since we're using also constants (`FC_V0_23_SNAP_VERSION`), we should centralize their definitions and also use them here.

acatangiu (commented):
Great stuff!

src/devices/src/virtio/block/device.rs
acatangiu: I believe we can structure the code so as to avoid regressions on the _legacy_ path.

src/devices/src/virtio/queue.rs
serban300: Fixed.

src/devices/src/virtio/queue.rs
serban300: This is better from a naming point of view, but it introduces the possibility of forgetting to call `do_kick()` after `needds_kick()`. Actually the code was similar to this in an earlier version of the PR and when I tried to use this logic to implement notif suppression on net also in order to do some experiments I forgot to call `do_kick()` and it lead to freezes. I think it's safer to keep it as it is now.

src/vmm/src/version_map.rs
serban300: Nice. Fixed.

src/devices/src/virtio/block/device.rs
serban300: I was unsure if we should treat this case. It complicates the code a bit, and notification suppression is present in all the kernels that we support AFAIK. Anyway I tried to fix this. I added a new commit. PTAL.

src/devices/src/virtio/queue.rs
serban300: Nice. Done.

src/vmm/src/device_manager/mmio.rs
serban300: Fixed.

src/vmm/src/device_manager/mmio.rs
serban300: Nice. I used it in `kick_devices()`. `with_virtio_device_with_id` uses `get_device` which is more efficient.

src/vmm/src/persist.rs
serban300: Fixed.

src/devices/src/virtio/queue.rs
acatangiu: I still think current code is not ideal since it does only half of something. My preference was that we either:
- have single function which checks if IRQ needed then also kicks
- cleanly split checking and kicking

But I see that it's not easy to do that since part of the logic is in the queue while irq logic is in the device. I will not block the PR on this since it's a two-way door implementation detail.

src/vmm/src/persist.rs
```diff
@@ -32,10 +34,9 @@ use seccompiler::BpfThreadMap;
 use snapshot::Snapshot;
 use versionize::{VersionMap, Versionize, VersionizeResult};
 use versionize_derive::Versionize;
+use virtio_gen::virtio_ring::VIRTIO_RING_F_EVENT_IDX;
 use vm_memory::GuestMemoryMmap;
 
-#[cfg(target_arch = "x86_64")]
-const FC_V0_23_SNAP_VERSION: u16 = 1;
 #[cfg(target_arch = "x86_64")]
 const FC_V0_23_MAX_DEVICES: u32 = 11;
```
acatangiu, line 41: nit: should this also go in `version_map.rs` const definitions?

src/devices/src/virtio/block/device.rs
```diff
@@ -323,16 +359,14 @@ impl Block {
             });
 
             if queue.prepare_kick(mem) {
-                self.irq_trigger
-                    .trigger_irq(IrqType::Vring)
-                    .unwrap_or_else(|_| {
-                        METRICS.block.event_fails.inc();
-                    });
+                self.irq_trigger.trigger_vring_irq(batch_irqs);
```
acatangiu, line 362: To keep the code simple, could we "delay" the IRQ here for both interrupt suppression and no-suppression cases?

Basically "batch" IRQs in both cases? (might introduce extra latency for suppression case, but not sure if that's all or if it messes with the mechanism in other ways)

src/devices/src/virtio/block/device.rs
```diff
@@ -177,6 +178,41 @@ impl Drop for DiskProperties {
     }
 }
 
+pub(crate) struct BlockIrqTrigger {
+    pub(crate) inner: IrqTrigger,
+    has_batched_vring_irqs: bool,
```
acatangiu, line 183:
```suggestion
    pending_batched_vring_irqs: bool,
```

src/devices/src/virtio/block/device.rs
```diff
@@ -177,6 +178,41 @@ impl Drop for DiskProperties {
...
+        } else {
+            self.do_trigger_vring_irq();
+        }
+    }
+
+    fn flush_vring_irqs(&mut self) {
+        if self.has_batched_vring_irqs {
+            self.do_trigger_vring_irq();
+        }
+    }
```
acatangiu, line 213: For better clarity, I would try something like:
```suggestion
    pub fn trigger_vring_irq(&mut self) {
        self.inner.trigger_irq(IrqType::Vring).unwrap_or_else(|_| {
            METRICS.block.event_fails.inc();
        });
    }

    pub fn enqueue_vring_irq(&mut self) {
        self.has_batched_vring_irqs = true;
    }

    pub fn flush_vring_irqs(&mut self) {
        if self.has_batched_vring_irqs {
            self.trigger_vring_irq();
            self.has_batched_vring_irqs = false;
        }
    }
}
```

Then call `enqueue_vring_irq()` or directly `trigger_vring_irq()` in block based on batching flag, instead of passing down batching flag to this object.

src/devices/src/virtio/block/device.rs
acatangiu: If we can't delay, then I suggest we call `enqueue_irq()` or `trigger_irq()` here based on `batch_irqs` flag to keep logic clear and centered:

```suggestion
                if batch_irqs {
                    self.irq_trigger.enqueue_vring_irq();
                } else {
                    self.irq_trigger.trigger_vring_irq();
                }
```

src/devices/src/virtio/block/device.rs
serban300: Removed this code as discussed offline.

src/devices/src/virtio/block/device.rs
serban300: Removed this code as discussed offline.

src/vmm/src/persist.rs
serban300: I'm not sure. Since the check is implemented in this file I would keep it here. But if you have strong preferences I can move it.

src/devices/src/virtio/block/device.rs
serban300: As we discussed offline we decided to drop batching for both cases (with and wo notif suppression). This will help us keep the code simple.

Since all kernels >=4.14 support notification suppression, there are only few corner cases where the performance regression would show:
- very old or custom kernel that doesn't support notif suppression
- new kernel with notif suppression disabled (for snapshot compatibility for example)

So for the moment we'll drop batching and revisit it later if needed.

src/devices/src/virtio/queue.rs
serban300: Yes, exactly. This is the first half of a kick as specified [here](https://elixir.bootlin.com/linux/v5.13/source/drivers/virtio/virtio_ring.c#L1814). It's also used in the same way in the [kernel block device](https://elixir.bootlin.com/linux/v5.13/source/drivers/block/virtio_blk.c#L210) btw. 

Another option is to have a method `Queue::kick(&self, irq_trigger: IrqTrigger)`. This way it could check if a notification is needed and call `irq_trigger.trigger_irq()` if so. But it would probably raise question related to rust-vmm upstream compatibility. Also the IrqTrigger is not mature enough and I'm not sure how it will evolve and if it will be so easy to pass it to the `Queue::kick()` in the future.

Persoanlly for the moment I think it would be safest to keep it as it is and maybe revisit this after migrating to the Queue implementation in rust-vmm.

acatangiu (approved):
🎆

sandreim (approved):
🚀

PR #2659, opened 2021-07-13
Abort test when failing to mount ramdisk

### LINKED ISSUE #2612: [Bug] devtool mount ramdisk does not fail when permission denied

bitoku:
# Reason for This PR

When you don't have a proper permission to mount ramdisk, the test suite won't issue a fatal error as mentioned in #2612. 
This PR adds check the status code of creating and mounting ramdisk so that test will die if you fail to mount it.

Before this PR, I tried to use docker `--mount type=tmpfs` option instead of adding the status check, but it had some performance regression (See #2655 ).

Close #2612 

## Description of Changes

It checks the status code of this command
```bash
    mkdir -p ${DEFAULT_RAMDISK_PATH} && \
    mount -t tmpfs -o size=${ramdisk_size} tmpfs ${DEFAULT_RAMDISK_PATH}
```

In order to check the status code I used `ok_or_die` command.

## License Acceptance

## PR Checklist

### DIFF
--- a/tools/devtool
+++ b/tools/devtool
@@ -822,6 +822,7 @@ mount_ramdisk() {
     umount_ramdisk
     mkdir -p ${DEFAULT_RAMDISK_PATH} && \
     mount -t tmpfs -o size=${ramdisk_size} tmpfs ${DEFAULT_RAMDISK_PATH}
+    ok_or_die "Failed to mount ramdisk to ${DEFAULT_RAMDISK_PATH}. Check the permission."
     mkdir -p ${DEFAULT_RAMDISK_PATH}/srv
     mkdir -p  ${DEFAULT_RAMDISK_PATH}/tmp
 }
--- a/tools/devtool
+++ b/tools/devtool
@@ -828,6 +828,15 @@ mount_ramdisk() {
 }
 
 umount_ramdisk() {
+    if [ ! -e "${DEFAULT_RAMDISK_PATH}" ]; then
+        return 0
+    fi
+    if [ ! -d "${DEFAULT_RAMDISK_PATH}" ]; then
+        die "${DEFAULT_RAMDISK_PATH} is not a directory."
+    fi
+    if [ ! -w "${DEFAULT_RAMDISK_PATH}" ]; then
+        die "Failed to unmount ${DEFAULT_RAMDISK_PATH}. Check the permission."
+    fi
     umount ${DEFAULT_RAMDISK_PATH} &>/dev/null
     rmdir ${DEFAULT_RAMDISK_PATH} &>/dev/null
 }


tools/devtool
```diff
@@ -822,6 +822,7 @@ mount_ramdisk() {
     umount_ramdisk
```
alindima, line 822: I'd also add a check for the `umount` call in `umount_ramdisk`, which would require the same privileges as mount

bitoku: I added a check in `umount_ramdisk`!

tools/devtool
```diff
@@ -822,11 +822,21 @@ mount_ramdisk() {
...
+    if [ ! -e "${DEFAULT_RAMDISK_PATH}" ]; then
+        return 0
+    fi
+    if [ ! -d "${DEFAULT_RAMDISK_PATH}" ]; then
+        die "${DEFAULT_RAMDISK_PATH} is not a directory."
+    fi
+    if [ ! -w "${DEFAULT_RAMDISK_PATH}" ]; then
+        die "Failed to unmount ${DEFAULT_RAMDISK_PATH}. Check the permission."
+    fi
     umount ${DEFAULT_RAMDISK_PATH} &>/dev/null
```
alindima, line 840: I suggest we simply do a check with `ok_or_die` after the call to `umount`, just like we do with `mount`

tools/devtool
```diff
@@ -822,11 +822,21 @@ mount_ramdisk() {
     umount_ramdisk
     mkdir -p ${DEFAULT_RAMDISK_PATH} && \
     mount -t tmpfs -o size=${ramdisk_size} tmpfs ${DEFAULT_RAMDISK_PATH}
+    ok_or_die "Failed to mount ramdisk to ${DEFAULT_RAMDISK_PATH}. Check the permission."
     mkdir -p ${DEFAULT_RAMDISK_PATH}/srv
     mkdir -p  ${DEFAULT_RAMDISK_PATH}/tmp
 }
 
 umount_ramdisk() {
+    if [ ! -e "${DEFAULT_RAMDISK_PATH}" ]; then
```
alindima, line 831: this check is not needed if we later check that `[ -d "${DEFAULT_RAMDISK_PATH}" ]`

tools/devtool
bitoku: It might not be a good idea. If the directory does not exist, `umount` returns non-zero, and if the directory is not mounted, `umount` returns non-zero as well.
`umount_ramdisk` must have idempotency, so I didn't use `ok_or_die`.

tools/devtool
bitoku: There three patterns of `DEFAULT_RAMDISK_PATH`.
- not exist
- file
- directory

Obviously, if the path does not exist, we must not do nothing, just return 0.
If the path is a file, `DEFAULT_RAMDISK_PATH` might be wrong, so it had better die.
If the directory is a directory, it should unmount directory and remove it.

This if clause is for checking the existence of the path.
The second if clause is for checking whether the path is a file or a directory.

tools/devtool
alindima: makes sense

tools/devtool
alindima: thanks for the explanation

alindima (approved):
Looks good, thanks for the contribution

AlexandruCihodaru (approved):
🚢

PR #2663, opened 2021-07-15
snapshot create - extend target version checks

serban300:
# Reason for This PR

In the future we might want to forbid the creation of backwards
compatible snapshots based on numeric comparisons
(e.g. target_version <= 0.25.0). The current commit adds this
capability.

Other improvements:
- simplify get_snapshot_data_version()
- extend the unit tests in order to check the returned error type

## Description of Changes

See above

## License Acceptance

## PR Checklist


src/utils/src/fc_version.rs
```diff
@@ -0,0 +1,98 @@
...
+                .ok_or(Error::InvalidFormat)?
+                .parse()
+                .map_err(|_| Error::InvalidFormat)
+        }
+        let major = parse_next_part(&mut parts)?;
+        let minor = parse_next_part(&mut parts)?;
+        let revision = parse_next_part(&mut parts)?;
+        if parts.next() != None {
+            return Err(Error::InvalidFormat);
+        }
```
luminitavoicu, line 49: nit:

```suggestion
        let mut parts = version_str.splitn(3, '.');
        fn parse_next_part(parts: &mut Split<char>) -> Result<u32, Error> {
            parts
                .next()
                .ok_or(Error::InvalidFormat)?
                .parse()
                .map_err(|_| Error::InvalidFormat)
        }
        let major = parse_next_part(&mut parts)?;
        let minor = parse_next_part(&mut parts)?;
        let revision = parse_next_part(&mut parts)?;
```

serban300: Looks a bit better. Thanks ! Done.

src/vmm/src/persist.rs
```diff
@@ -297,23 +297,28 @@ fn snapshot_memory_to_file(
 
 /// Validate the microVM version and translate it to its corresponding snapshot data format.
 pub fn get_snapshot_data_version(
-    version: &Option<String>,
+    maybe_fc_version_string: &Option<String>,
```
luminitavoicu, line 300: I don't really like the name of this argument. Since its type already is `Option`, couldn't we just call it `fc_version_string`?

src/vmm/src/persist.rs
```diff
@@ -297,23 +297,28 @@ fn snapshot_memory_to_file(
...
     version_map: &VersionMap,
     _vmm: &Vmm,
```
luminitavoicu, line 302: Nit:
```suggestion
    vmm: &Vmm,
```

luminitavoicu (commented):
Looks good! I just have a couple of nits.

src/vmm/src/persist.rs
serban300: This is a naming pattern that I saw in the rust standard library. The variables that are of type `Option` are prefixed with `maybe_`. It also make sense, because after that you might want to unwrap it and you would do something like 

```
maybe_fc_version_string: &Option<String>
...
fc_version_string = maybe_fc_version_string.unwrap()
```

Personally I would keep it like this.

src/vmm/src/persist.rs
serban300: This will lead to a build error on aarch64. Because this variable is not needed on aarch64.

src/vmm/src/persist.rs
sandreim: why not `optional_fc_version_string` ?

src/vmm/src/version_map.rs
```diff
@@ -34,13 +42,14 @@ lazy_static! {
 
     /// Static instance used for creating a 1:1 mapping between Firecracker release version
     /// and snapshot data format version.
-    pub static ref FC_VERSION_TO_SNAP_VERSION: HashMap<String, u16> = {
+    pub static ref FC_VERSION_TO_SNAP_VERSION: HashMap<FcVersion, u16> = {
         let mut mapping = HashMap::new();
+        // aarch64 snapshotting support was introduced in v0.24.0
         #[cfg(not(target_arch = "aarch64"))]
-        mapping.insert(String::from("0.23.0"), 1);
+        mapping.insert(FC_VERSION_0_23_0, 1);
```
sandreim, line 49: What problem is replacing the string with FcVersion ? The extended version checks can be easily done against the data version. IMO this looks a bit over engineered or I am missing something.

serban300: 1. It looks like the natural thing is to perform the checks based on Fc version. In the end it is a Firecracker release that adds support for different features, and not a data version. Moreover, if you base the checks on the data version, you also have to store the reverse mapping. For example as it is done right now: `const FC_V0_23_SNAP_VERSION: u16 = 1;`. 
2. The `FcVersion` struct is very light and some of the logic included there is done anyway (for example checking if the fc_version string is valid). 

3. Consider the following situation:

```
    pub static ref FC_VERSION_TO_SNAP_VERSION: HashMap<String, u16> = {
        let mut mapping = HashMap::new();
        mapping.insert(String::from("0.23.0"), 1);
        mapping.insert(String::from("0.25.0"), 3);
        mapping.insert(String::from("0.26.0"), 4);
        mapping.insert(String::from("0.27.0"), 5);

        mapping
    };
```

Now let's say that you want to add version `0.23.1`. You will probably do `mapping.insert(String::from("0.23.1"), 6);`.

And let's say that you want to perform some logic for Fc versions <= 0.25.0. You can do `data_version <= 3 && version != 6`, but these kind of checks can be hard to follow.

src/vmm/src/persist.rs
serban300: Personally I prefer `maybe`. It's shorter, and it's used in the rust standard library. We also use it in Firecracker in different places. I would keep it for all these reasons, but it's just a personal preference.

luminitavoicu: I'm okay with keeping it like this if it's also common in the rust standard library.

src/vmm/src/version_map.rs
sandreim: The theoretical example makes sense and the yes, the solution is logical in that context. The complexity of managing state changes for patch releases makes it something we want to avoid entirely - in your example, Firecracker v0.24.0 would basically need to know how to load the v0.23.1 state which was mutated after v0.24.0 is released. My expectation and suggestion is to only do state changes as part of new release and never otherwise. This a

serban300: It's true that managing state changes for patch releases is quite complex, but I wouldn't completely dismiss the possibility. I would postpone the decision until we face a situation where we need to take this into consideration.

In any case personally I think it would be better to perform snapshot creation checks on Fc version and not on the data version. Also the snapshotting API (`snapshot/create`) uses the Fc version and not the data version.

But if you have strong concerns about this, I can remove the `FcVersion` struct and try to redo the PR, keeping any simplifications that make sense without it.

serban300:
Closing since I will include asimpler approach in https://github.com/firecracker-microvm/firecracker/pull/2658

PR #2665, opened 2021-07-16
[v0.24] Allow syscalls used by Rust panic

georgepisaltu:
# Reason for This PR

Whenever a panic happens, `libc::abort` is called. This function uses syscalls that are not allowed by the current seccomp filters.

## Description of Changes

Allowed these syscalls in order to allow the panics to finish cleanly.

## License Acceptance

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed seccomp blocking syscalls necessary for Rust panics.
+
 ## [0.24.4]
 
 ### Fixed
--- a/src/vmm/src/default_syscalls/filters.rs
+++ b/src/vmm/src/default_syscalls/filters.rs
@@ -62,6 +62,7 @@ pub fn default_filter() -> Result<SeccompFilter, Error> {
             allow_syscall_if(
                 libc::SYS_futex,
                 or![
+                    and![Cond::new(1, ArgLen::DWORD, Eq, super::FUTEX_WAIT)?],
                     and![Cond::new(1, ArgLen::DWORD, Eq, super::FUTEX_WAIT_PRIVATE)?],
                     and![Cond::new(1, ArgLen::DWORD, Eq, super::FUTEX_WAKE_PRIVATE)?],
                     #[cfg(target_env = "gnu")]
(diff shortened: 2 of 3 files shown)


acatangiu:
Can you add a seccomp integ/regression test with a simple binary that just does `panic!()` with the default seccomp filter set and verify it doesn't get `SIGSYS` anymore?

serban300 (commented):
The allowed syscalls seem ok. Shouldn't we fist merge this into main and then backport it to older versions ?

georgepisaltu:
> Can you add a seccomp integ/regression test with a simple binary that just does `panic!()` with the default seccomp filter set and verify it doesn't get `SIGSYS` anymore?

I was thinking of adding the test when merging on `main`, as it's much easier with the new seccompiler.

> The allowed syscalls seem ok. Shouldn't we fist merge this into main and then backport it to older versions ?

The idea was to merge this fast on currently released versions, but at this point it's better to just do it on `main` first I think.

acatangiu:
> I was thinking of adding the test when merging on main, as it's much easier with the new seccompiler.

Fine by me. But without a test, for this PR at least add a section in PR description or a comment describing the dev-testing done for these changes. Are we sure the current whitelist changes are enough to allow panics/aborts to run to completion?

georgepisaltu:
> > I was thinking of adding the test when merging on main, as it's much easier with the new seccompiler.
> 
> Fine by me. But without a test, for this PR at least add a section in PR description or a comment describing the dev-testing done for these changes. Are we sure the current whitelist changes are enough to allow panics/aborts to run to completion?

The testing was:

- having a program fork and run a modified Firecracker version where we introduced a panic after building the VMM (and installing the seccomp filters), then check in the parent process that the signal which terminated Firecracker is SIGABRT (6);
- creating a small program which installs the Firecracker seccomp filters with the seccompiler from v0.25 before panicking, then checking the output to make sure there is no "Bad system call" and we see the "Aborted" message, then check the exit code of the process and run it through a `WTERMSIG` to make sure the signal that terminated it is indeed SIGABRT (6).

Anyway, I think I will close this PR and open a new one on main where I also add the test, then we do the backport. If it works on v0.25 with the added syscalls, it should work on v0.24 as well, since we have the information about what syscalls are made during `abort` and they are the same across Firecracker versions, since it's a `musl` thing.

PR #2686, opened 2021-08-05
Avoid overflow on computing vsock packet used length

luminitavoicu:
# Reason for This PR

Avoid overflow during computing vsock packet used length.

## Description of Changes

Check package length before commiting the header to guest memory. If len is too big, we return `InvalidPktLen` error. 

## License Acceptance

## PR Checklist

- [X] ~Any required documentation changes (code and docs) are included in this PR.~
- [X] ~Any newly added `unsafe` code is properly documented.~
- [X] ~Any API changes are reflected in `firecracker/swagger.yaml`.~
- [X] ~Any user-facing changes are mentioned in `CHANGELOG.md`.~

### DIFF
--- a/src/devices/src/virtio/vsock/device.rs
+++ b/src/devices/src/virtio/vsock/device.rs
@@ -148,10 +148,14 @@ where
                 Ok(mut pkt) => {
                     if self.backend.recv_pkt(&mut pkt, mem).is_ok() {
                         match pkt.commit_hdr(mem) {
+                            // This addition cannot overflow, because packet length
+                            // is previously validated against `MAX_PKT_BUF_SIZE`
+                            // bound as part of `commit_hdr()`.
                             Ok(()) => VSOCK_PKT_HDR_SIZE as u32 + pkt.len(),
                             Err(e) => {
                                 warn!(
-                                    "vsock: Error writing packet header to guest memory: {:?}",
+                                    "vsock: Error writing packet header to guest memory: {:?}.\
+                                     Discarding the package.",
                                     e
                                 );
                                 0
@@ -165,7 +169,7 @@ where
                     }
                 }
                 Err(e) => {
-                    warn!("vsock: RX queue error: {:?}", e);
+                    warn!("vsock: RX queue error: {:?}. Discarding the package.", e);
                     0
                 }
             };
--- a/src/devices/src/virtio/vsock/packet.rs
+++ b/src/devices/src/virtio/vsock/packet.rs
@@ -176,10 +176,7 @@ impl VsockPacket {
         }
 
         // Reject weirdly-sized packets.
-        //
-        if pkt.len() > defs::MAX_PKT_BUF_SIZE as u32 {
-            return Err(VsockError::InvalidPktLen(pkt.len()));
-        }
+        pkt.check_len()?;
 
         pkt.init_buf(hdr_desc, false)?;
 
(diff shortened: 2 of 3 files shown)


src/devices/src/virtio/vsock/device.rs
```diff
@@ -148,7 +148,14 @@ where
                 Ok(mut pkt) => {
                     if self.backend.recv_pkt(&mut pkt, mem).is_ok() {
                         match pkt.commit_hdr(mem) {
-                            Ok(()) => VSOCK_PKT_HDR_SIZE as u32 + pkt.len(),
+                            Ok(()) => pkt.len().checked_add(VSOCK_PKT_HDR_SIZE as u32).unwrap_or_else(|| {
+                                warn!(
+                                    "vsock: packet length greater than maximum expected size: {} > {}",
```
serban300, line 153: I would also say that we're discarding the package.

georgepisaltu: And I would turn the `warn!` into `error!` since we're actually dropping the package.

luminitavoicu: I am not sure if `error!` would be the right level here. All other package drops have been signaled as `warn!` so far.

serban300 (approved):

src/devices/src/virtio/vsock/device.rs
```diff
@@ -148,10 +148,19 @@ where
...
                         match pkt.commit_hdr(mem) {
-                            Ok(()) => VSOCK_PKT_HDR_SIZE as u32 + pkt.len(),
+                            Ok(()) => pkt.len().checked_add(VSOCK_PKT_HDR_SIZE as u32).unwrap_or_else(|| {
+                                warn!(
+                                    "vsock: Packet length greater than maximum expected size: {} > {}.\
+                                     Discarding the package.",
+                                    pkt.len(),
+                                    u32::MAX - VSOCK_PKT_HDR_SIZE as u32
+                                );
+                                0
```
dianpopa, line 158: I have some concerns here related to value we are returning (i.e 0) given that we write the header to memory (commit_hdr) even though the packet length + header size exceeds u32::MAX. Let me wrap my head around it and come back with a conclusion.

luminitavoicu: How about we check that addition would not overflow (i.e. `pkt.len() < u32::MAX - VSOCK_PKT_HDR_SIZE`) before we commit the header to memory. If true, then we can write the packet to guest. Otherwise, we return 0 and drop the package. WDYT?

dianpopa: Yes, so it is actually correct to return 0 but there is still a problem with the code. We need to do the overflow check before committing the header (i.e writing the header to memory) since if we are in the scenario where we have an overflow this might mean we do not have enough space to write the header. 
So, my suggestion is to move the check in `commit_hdr` function and return an error if there is overflow.

dianpopa: > How about we check that addition would not overflow (i.e. `pkt.len() < u32::MAX - VSOCK_PKT_HDR_SIZE`) before we commit the header to memory. If true, then we can write the packet to guest. Otherwise, we return 0 and drop the package. WDYT?

Indeed, this was my suggestion also, saw your comment after replying.

src/devices/src/virtio/vsock/packet.rs
```diff
@@ -222,8 +222,14 @@ impl VsockPacket {
 
     /// Writes the local copy of the packet header to the guest memory.
     pub fn commit_hdr(&self, mem: &GuestMemoryMmap) -> Result<()> {
-        mem.write_obj(self.hdr, self.hdr_addr)
-            .map_err(VsockError::GuestMemoryMmap)
+        if self.len() < u32::MAX - VSOCK_PKT_HDR_SIZE as u32 {
```
serban300, line 225: On a second thought, isn't this already checked [here](https://github.com/firecracker-microvm/firecracker/blob/afe9c0623f12156b22c814e86f7b4420906c71c9/src/devices/src/virtio/vsock/packet.rs#L180) ?

luminitavoicu: This is inside `from_tx_virtq_head` function, it is not part of our code path.

dianpopa: That is nice, did not know there is a MAX_PKT_BUF_SIZE defined for this. So seeing this I suggest to create an auxiliary function in Vsockpacket with [this check](https://github.com/firecracker-microvm/firecracker/blob/afe9c0623f12156b22c814e86f7b4420906c71c9/src/devices/src/virtio/vsock/packet.rs#L180) and then use it from there. We can call it from `commit_hdr` before trying to `write_obj`.

luminitavoicu: Done

dianpopa (approved):

georgepisaltu (approved):

Issue #841, opened 2019-01-11
Jailer lacks support for cgroups v2


acipu-aws:
The regex in Cgroup::new only matches cgroups v1. Per the [man page](http://man7.org/linux/man-pages/man7/cgroups.7.html), cgroups v2 has a single hyerarchy, so specifying the type of controller in the options is no longer feasible. From the controllers used, cpu and pids are both supported in v2.

The jailer should be aware of both cgroups v1 and v2 and adapt the creation code accordingly.

acipu-aws:
Extract from a /proc/mounts that fails:

`none /cgroups/cpu cgroup2 rw,relatime 0 0`
`none /cgroups/cpuset cgroup rw,relatime,cpuset 0 0`

jdoss:
Are there any plans to support cgroups v2? I see there was a PR to add support for both v1 and v2 but it never got merged. :(

jdoss:
@alindima I am looking to use Firecracker on Fedora CoreOS which is moving to cgroups v2 by default on the Stable channel.

https://lists.fedoraproject.org/archives/list/coreos-status@lists.fedoraproject.org/thread/V6OSIUOL5ICEP7WMXLJ2NI47EFNQ3MYF/

alindima:
> @alindima I am looking to use Firecracker on Fedora CoreOS which is moving to cgroups v2 by default on the Stable channel.
> 
> https://lists.fedoraproject.org/archives/list/coreos-status@lists.fedoraproject.org/thread/V6OSIUOL5ICEP7WMXLJ2NI47EFNQ3MYF/

Thanks for the info! From what I quickly read in the link you pasted, it seems that Fedora CoreOS is enabling cgroups v2 by default, but preserves the v1 compatibility, is that right?

I'm trying to gauge if this would be a blocker for your use case or more of a nice to have.

jdoss:
It is a blocker for my use case. Currently right now on FCOS next and testing streams which runs as cgroups v2 you cannot launch the jailer command with the below example: 

```
$ fgrep VERSION /etc/os-release
VERSION="34.20210518.1.0 (CoreOS)"
$ cat /proc/mounts |grep cgroup
cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime,nsdelegate,memory_recursiveprot 0 0
$ grep cgroup /proc/filesystems
nodev   cgroup
nodev   cgroup2
[...]

PR #2687, opened 2021-08-05
Add support for cgroups v2

alsrdn:
# Reason for This PR

Addresses #841 

## Description of Changes

This patch set adds the ability to use the jailer on systems that use cgroup-v2.
A new parameter named `--cgroup-version` was introduced in the jailer that enables the user to swap between using Cgroup-v1 and Cgroup-v2.

Hybrid mode is not supported as it is impractical for production use because controllers cannot be added to both v1 and v2 hierarchies at the same time. Typically on systems that use hybrid cgroups, all controllers are mounted to v1 hierarchies by default preventing controllers to be added to v2.

The default behaviour of the jailer (if `--cgroup-version` is not specified) is to use Cgroup-v1.
Using the jailer with `--cgroup-version 1` on a system that does not have Cgroup-v1 hierarchies mounted will result in an error. The same happens if `--cgroup-version 2` on a system that does not have Cgroup-v2

## License Acceptance

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 ### Added
 
+- Added jailer option `--cgroup-version <1|2>` to support running the jailer
+  on systems that have cgroup-v2. Default value is `1` which means that if
+  `--cgroup-version` is not specified, the jailer will try to create cgroups
+  on cgroup-v1 hierarchies only.
 - Added devtool build `--ssh-keys` flag to support fetching from private
   git repositories.
 - Added option to configure block device flush.
--- a/src/jailer/src/cgroup.rs
+++ b/src/jailer/src/cgroup.rs
@@ -1,10 +1,12 @@
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::hash_map::Entry::{Occupied, Vacant};
+use std::collections::HashMap;
 use std::ffi::OsStr;
 use std::fs::{self, File};
 use std::io::{BufRead, BufReader};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process;
 
 use regex::Regex;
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Added
 
+- Added jailer option `--cgroup-version <1|2>` to support running the jailer
```
sandreim, line 7: We should also mention the default value.

src/jailer/src/cgroup.rs
```diff
@@ -1,25 +1,224 @@
...
+} else {
+    "/proc/mounts"
+};
+
 const NODE_TO_CPULIST: &str = "/sys/devices/system/node/node"; // This constant should be removed once the `--node` argument is removed.
 
-pub struct Cgroup {
+// Enum identify cgroup version to be used
+#[derive(PartialEq, Copy, Clone)]
+pub enum CgroupVersion {
```
sandreim, line 27: Does an u8 work here instead of the enum ?

tests/integration_tests/security/test_jail.py
```diff
@@ -104,8 +107,66 @@ def test_arbitrary_usocket_location(test_microvm_with_initrd):
...
+@functools.lru_cache(maxsize=None)
+def cgroup_v1_available():
+    """Check if cgroup-v1 is disabled on the system."""
+    with open("/proc/cmdline") as cmdline_file:
+        cmdline = cmdline_file.readline()
+        return bool("cgroup_no_v1=all" not in cmdline)
+
+
+@pytest.fixture
+def sys_setup_cgroups():
```
sandreim, line 119: Can this be a helper function instead of a fixture ?

src/jailer/src/cgroup.rs
```diff
@@ -1,25 +1,224 @@
...
+                    break;
+                } else if ver == CgroupVersion::V1 && capture["ver"].is_empty() {
+                    // Found a cgroupv1 mountpoint; with cgroupsv1 we can have multiple hiearachies.
+                    // Since we don't know which one will be used, we cache the mountpoints now,
+                    // and will create the hierachies on demand when a cgroup is built.
+                    b.mount_points.push(CgroupMountPoint {
+                        dir: String::from(&capture["dir"]),
+                        options: String::from(&capture["options"]),
+                    });
+                    found = true;
```
sandreim, line 100: Should we break loop here ?

src/jailer/src/cgroup.rs
```diff
@@ -1,25 +1,224 @@
...
+            File::open(PROC_MOUNTS).map_err(|e| Error::FileOpen(PathBuf::from(PROC_MOUNTS), e))?;
+
+        // Regex courtesy of Filippo.
+        // This will match on both v1 and v2 mount points.
+        // For v1, the "ver" capture group will be empty (len = 0)
+        let re = Regex::new(
+            r"^([a-z2]*)[[:space:]](?P<dir>.*)[[:space:]]cgroup(?P<ver>2?)[[:space:]](?P<options>.*)[[:space:]]0[[:space:]]0$",
+        ).map_err(Error::RegEx)?;
+
+        let mut found = false;
```
sandreim, line 80: Since we insert into mount points or hierachies we can use that information instead of an extra `found` var.

src/jailer/src/cgroup.rs
```diff
@@ -1,25 +1,224 @@
...
+struct CgroupBase {
     file: String,      // file representing the cgroup (e.g cpuset.mems).
     value: String,     // value that will be written into the file.
     location: PathBuf, // microVM cgroup location for the specific controller.
 }
 
+pub struct CgroupV1(CgroupBase);
+pub struct CgroupV2(CgroupBase);
+
+pub trait Cgroup {
```
sandreim, line 214: Is this really necessary ? It is highly unlinkely there's gonna be a v3 soon :)

src/jailer/src/cgroup.rs
```diff
@@ -1,25 +1,224 @@
...
+    }
+
+    // Creates a new cggroup and returns it
+    pub fn new_cgroup(
+        &mut self,
+        file: String,
+        value: String,
+        id: &str,
+        exec_file_name: &OsStr,
+    ) -> Result<Box<dyn Cgroup>> {
```
sandreim, line 120: If we drop the Cgroup trait, I think the code becomes more simple to read, as I see there is not much shared between the 2 impls. We could have new_v1_cgroup and new_v2_cgroup.

sandreim (commented):
Looks good overall. I've added some comments and questions.

src/jailer/src/cgroup.rs
alsrdn: It can work but enum would be much safer IMO because it forces handling of both variants when needed.
Also it reduces the need for error handling for values which are not 1 or 2.

tests/integration_tests/security/test_jail.py
alsrdn: It can be a helper function but why is a fixture not appropriate here?
The purpose of this code is to set-up the croups environment on the system so that the tests can run successfully.
Maybe a more suggestive name is needed for the fixture like `cgroups_env`. What do you think?

src/jailer/src/cgroup.rs
alsrdn: No. Different cgroup-v1 hierarchies can be mounted at different locations. This is actually the standard approach.
We break the loop in case of Cgroup-v2 because there can be a single mount point in the system and there is no point in searching further.

src/jailer/src/cgroup.rs
alsrdn: Indeed, will change.

src/jailer/src/cgroup.rs
alsrdn: This was my initial thought also but while it would make code in cgroups.rs shorter, the code in env.rs would become more ugly for a few reasons:

- While impls for cgroup v1 and v2 differ, the interface for interacting with cgroups remains the same.

- Dropping the trait would require the Env struct to have 2 vectors to hold Cgroup{v1/v2} objects out of which only one would be used during the entire lifetime of the jailer.

- Because Env is the one that parses the cgroup-version argument it would have to have 2 branches that build objects that look the same for each version (same arguments to build). This would be true also for every usage of the objects (writing the value and attaching pids) which looks like a lot of boiler plate.

tests/framework/jailer.py
```diff
@@ -49,6 +50,7 @@ def __init__(
             daemonize=True,
             new_pid_ns=False,
             cgroups=None,
+            cgroup_ver=None,
```
gc-plp, line 53: Would it make sense to set `cgroup_ver` to v1, as this is the default implementation?

tests/framework/jailer.py
```diff
@@ -110,6 +113,10 @@ def construct_param_list(self):
             jailer_param_list.append('--daemonize')
         if self.new_pid_ns:
             jailer_param_list.append('--new-pid-ns')
+        if self.cgroup_ver:
```
gc-plp, line 116: Should we test the jailer with v1 and then with `cgroup_ver` set to None? It seems that we're only exercising the v2 path.

tests/integration_tests/security/test_jail.py
```diff
@@ -152,14 +254,21 @@ def test_cgroups(test_microvm_with_initrd):
...
 
-def test_node_cgroups(test_microvm_with_initrd):
+def test_node_cgroups(test_microvm_with_initrd, sys_setup_cgroups):
     """Test the cgroups are correctly set by the jailer."""
+    # pylint: disable=redefined-outer-name
     test_microvm = test_microvm_with_initrd
     test_microvm.jailer.cgroups = None
     test_microvm.jailer.numa_node = 0
+    cgroup_version = sys_setup_cgroups
+    if cgroup_version == 2:
```
gc-plp, line 270: Does anything break if we just `test_microvm.jailer.cgroup_ver = sys_setup_cgroups`?

raduiliescu:
With how is defined not no one will use this with "cgroup-version 1".  Why not use a parameter like "--cgroupsv2" or when specifying the cgroup have a different parameter for v2 groups?
How do other maintainers think about this?

tests/integration_tests/security/test_jail.py
sandreim: One issue I have with fixtures is that it makes it harder to read and to reason about how the env is being setup. I would prefer to have it done imperatively in the test code vs declarative as you do with fixtures.

src/jailer/src/cgroup.rs
sandreim: I agree there is benefit generally when using enums - like making it easy to change the value in the future (and not the code), easier to read code (when dealing with 0x123131 constants) and avoid bugs due to typos. Also, you still validate non 1|2 values in FromStr trait.

I don't have a strong preference on this, but I do have related to increasing LoC :)

src/jailer/src/cgroup.rs
sandreim: I agree, makes sense.

alindima:
> How do other maintainers think about this?

Can we really not avoid adding a new parameter? Since we don't support hybrid mounts, where both v1 and v2 are used, can't we use the version that is mounted by default on the host?

tests/framework/jailer.py
alsrdn: Yes, makes sense to test with cgroup_ver set to 1 and 2 and also without the parameter set. I'll leave the default to None and add another test.

tests/integration_tests/security/test_jail.py
alsrdn: Nope, makes sense to assign it directly for sure.

alsrdn:
> How it is defined now, no one will use the "cgroup-version 1" option. Why not use a parameter like "--cgroupsv2" or when specifying the cgroup have a different parameter for v2 groups?

I wouldn't say that no one will use the v1 option. The default is put in place so that current uses of the jailer don't brake.
IMO having an explicit version switch helps with automation when some user wants to use either version. 
A `--cgroupsv2` like parameter feels more ambigous to me because it looks like the functionality is additive rather than mutually exclusive.
If these parameters would be translated to a REST API, then I think it would be a better design to have a field called cgroup_version that can be either 1 or 2 instead of having 2 fields cgroupsv2 and cgroupsv1 that can be true or false in 4 different combinations out of which only 2 are valid based on the implementation.

> > How do other maintainers think about this?
> 
> Can we really not avoid adding a new parameter? Since we don't support hybrid mounts, where both v1 and v2 are used, can't we use the version that is mounted by default on the host?

We could, but it would add a lot of complexity in the implementation because the decision cannot be made on the availability of the mount points alone. 
Most of the current distros mount both cgroup-v2 and cgroup-v1 hierarchies by default. Usually cgroup-v1 will hijack all controllers which would make croups-v2 unusable for distributing resources until v1 controllers are unmounted.
To make matters worse, some controller config files have the same name, while others are changed or removed completely. For example choosing the version based on the value of `--cgroup cpuset.mems` would be ambiguous.

src/jailer/src/cgroup.rs
alsrdn: Replaced with u8.

tests/framework/jailer.py
alsrdn: Left the default to None and added a test that does not set the parameter and expects the jailer to use cgroup-v1.

raduiliescu:
> 
> 
> > How it is defined now, no one will use the "cgroup-version 1" option. Why not use a parameter like "--cgroupsv2" or when specifying the cgroup have a different parameter for v2 groups?
> 
> I wouldn't say that no one will use the v1 option. The default is put in place so that current uses of the jailer don't break.
> IMO having an explicit version switch helps with automation when some user wants to use either version.
> A `--cgroupsv2` like parameter feels more ambigous to me because it looks like the functionality is additive rather than mutually exclusive.
> If these parameters would be translated to a REST API, then I think it would be a better design to have a field called cgroup_version that can be either 1 or 2 instead of having 2 fields cgroupsv2 and cgroupsv1 that can be true or false in 4 different combinations out of which only 2 are valid based on the implementation.

I agree having the version specified directly would be better when translated into a REST API. So let's stick with the initial proposed approach.

AlexandruCihodaru (approved):

src/jailer/src/cgroup.rs
```diff
@@ -113,102 +123,104 @@ pub fn cgroups_from_numa_node(
...
-impl Cgroup {
+// Return the path of the cgroup subfolder for a specific controller.
+// (<mountpoint>/<controller>/<exec_file_name>/<id>).
+fn get_location(file: &str, exec_file_name: &OsStr, id: &str) -> Result<PathBuf> {
+    let controller = get_controller_from_filename(file)?;
+    let f =
+        File::open(PROC_MOUNTS).map_err(|e| Error::FileOpen(PathBuf::from(PROC_MOUNTS), e))?;
+
+    // Regex courtesy of Filippo.
+    let re = Regex::new(
```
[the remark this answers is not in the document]
alsrdn: Added some more details about what the Regex does.

AlexandruCihodaru (approved):
🚢

src/jailer/src/cgroup.rs
```diff
@@ -1,25 +1,217 @@
...
+    // Creates the builder object
+    // It will discover cgroup mount points and hierarchies configured
+    // on the system and cache the info required to create cgroups later
+    // within this hierarchies
+    pub fn new(ver: u8) -> Result<Self> {
+        if ver != 1 && ver != 2 {
+            return Err(Error::CgroupInvalidVersion(ver.to_string()));
+        }
+
+        let mut b = CgroupBuilder {
```
sandreim, line 47: Nit: Use more precise variable names. builder ?

src/jailer/src/cgroup.rs
```diff
@@ -1,25 +1,217 @@
...
+            version: ver,
+            hierarchies: HashMap::new(),
+            mount_points: Vec::new(),
+        };
+
+        // search PROC_MOUNTS for cgroup mount points
+        let f =
```
sandreim, line 54: Nit: Same as above.

src/jailer/src/cgroup.rs
```diff
@@ -1,25 +1,217 @@
...
+                // name was specified. Return an error in this case
+                match path {
+                    Some(p) => Ok(e.insert(p)),
+                    None => Err(Error::CgroupControllerUnavailable(controller.to_string())),
+                }
+            }
+        }
+    }
+
+    // This function should be removed once the `--node` argument is removed.
```
sandreim, line 164: When is that going to happen ? Do we track it in a ticket ?

src/jailer/src/cgroup.rs
```diff
@@ -1,25 +1,217 @@
...
+            numa_node.to_string(),
+            microvm_id,
+            exec_file_name,
+        )?;
+
+        Ok(vec![cpuset_cpus, cpuset_mems])
+    }
+}
+
+struct CgroupBase {
```
sandreim, line 198: Why do we call it base ?

sandreim (approved):

AlexandruCihodaru (approved):
🚢

PR #2689, opened 2021-08-11
Allow mmap with MAP_PRIVATE on api_server thread

AlexandruCihodaru:
# Reason for This PR

When there is a large enough buffer there will be an attempt to map it using MAP_PRIVATE|MAP_ANONYMOUS. As mmap was not allowed with this flags on api_server thread the vmm was killed by seccomp filter.

## Description of Changes

-> Added rule to allow mmap with MAP_PRIVATE|MAP_ANONYMOUS
-> Added integration test.

## License Acceptance

## PR Checklist


resources/seccomp/x86_64-unknown-linux-musl.json
```diff
@@ -519,6 +519,19 @@
                     }
                 ]
             },
+{
+            "syscall": "mmap",
+            "comment": "Used for smth",
```
georgepisaltu, line 524: Update the comment.

tests/framework/defs.py
```diff
@@ -26,7 +26,7 @@
 SECCOMP_JSON_DIR = FC_WORKSPACE_DIR / "resources/seccomp"
 
 # Maximum accepted duration of an API call, in milliseconds
-MAX_API_CALL_DURATION_MS = 300
+MAX_API_CALL_DURATION_MS = 1200
```
georgepisaltu, line 29: Is there a way to increase this limit for that specific test only? This limit is actually pretty important.

resources/seccomp/x86_64-unknown-linux-musl.json
```diff
@@ -519,6 +519,19 @@
                     }
                 ]
             },
+{
```
alindima, line 522: what about the aarch64 filter? we should test if this is needed and add it

tests/integration_tests/functional/test_mmds.py
```diff
@@ -116,6 +118,31 @@ def test_custom_ipv4(test_microvm_with_ssh, network_config):
     _, stdout, _ = ssh_connection.execute_command(cmd)
     assert json.load(stdout) == data_store['latest']['meta-data']
 
+@pytest.mark.timeout(1200)
+def test_big_buffer(test_microvm_with_ssh, network_config):
+    """
+    Test send large buffer.
```
alindima, line 124: we should have a more descriptive name of the function and description. It'll appear in the test report and the test scenario is quite ambiguous

tests/integration_tests/functional/test_mmds.py
```diff
@@ -116,6 +118,31 @@ def test_custom_ipv4(test_microvm_with_ssh, network_config):
...
+        'latest': {
+            'meta-data': {
+            }
+        }
+    }
+
+    chars = ''.join(['b' for i in range(190*1024*1024)])
+    data_store["latest"]["meta-data"]["ami-id"] = chars
+    response = test_microvm.mmds.put(json=data_store)
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
```
alindima, line 145: this doesn't test that the data is actually available from the guest. I guess this is fine is we update the description to explicitly say that we are checking for seccomp errors as a regression test

georgepisaltu: This test should eventually be removed/replaced when we actually fix the non-seccomp side of the issue and we have functional tests for MMDS, so I don't think it's a problem in this PR.

alindima: hmm but could this be exercised for any API endpoint? Maybe this would be done for all huge JSON requests (even invalid?) I assume the mmap comes from a `malloc` or something, after the `recvfrom()`, but not sure

tests/framework/defs.py
gc-plp: I don't think that we should increase the API call duration this long for all the framework. We should check if we can do it temporarily for a test as @georgepisaltu suggested.

tests/integration_tests/functional/test_mmds.py
```diff
@@ -116,6 +118,31 @@ def test_custom_ipv4(test_microvm_with_ssh, network_config):
...
+    assert response.json() == {}
+
+    data_store = {
+        'latest': {
+            'meta-data': {
+            }
+        }
+    }
+
+    chars = ''.join(['b' for i in range(190*1024*1024)])
```
gc-plp, line 142: You don't necessarily need to go over the whole range. It's inefficient and takes about 5 seconds to do it - just tested this.

You can break it down in multiple loops:
```python
slice_1mb = list(itertools.repeat('x', 1024 * 1024))

slice_190mb = []
[slice_190mb.extend(slice_1mb) for _ in range(190))
```
This is like 10x faster.

We could also avoid creating a 190mb buffer in memory :)

tests/framework/defs.py
acatangiu: agree, one thing which is definitely supported by our framework is to disable API call measurements for specific tests or calls (see snapshot creation api calls).

src/devices/src/virtio/balloon/mod.rs
```diff
@@ -4,7 +4,6 @@
 pub mod device;
 pub mod event_handler;
 pub mod persist;
-#[cfg(test)]
```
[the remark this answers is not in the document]
AlexandruCihodaru: yes

tests/integration_tests/functional/test_mmds.py
AlexandruCihodaru: it is indeed correct the issue can be triggered for any api endpoint. Maybe I should move it in test_api.py? wdyt @alindima  @georgepisaltu

alindima: @AlexandruCihodaru I think that's the right thing to do. Please also update the name and description.

raduiliescu:
I am not sure if this is the best decision. Why we need a large buffer for MMDS, since is MMDS just metadata? 100MB of text is a bit too much.
My proposal would be to add information in the documentation for this, limit the response size and return 507 Insufficient Storage.

If a client really wants a large requests, than an alternative would be to first the HEAD method, and then make the GET request with the RANGES parameter. Not sure if microHTTP supports that.

PR #2697, opened 2021-08-13
Add Jailer ENV cleanup

berciuliviu:
# Reason for This PR

This PR provides a solution to cleansing the Jailer environment of environment variables received from the parent process. This is done to prevent any leaks (e.g. secrets etc.) from the host inside the Jailer environment.

## Description of Changes

When Jailer starts all `env variables` are parsed and removed. In this way, Jailer has a clean environment and no host env data will be present.

## License Acceptance

## PR Checklist


alindima (commented):
The PR looks good, but are we sure we want to remove all environment variables?
For example, there is the $TZ environment variable that is checked by libc when calling `localtime()` (which we are callling in Firecracker), to retrieve the local timezone.

Are we blocking users from leveraging valid environment variables? Should this be a switch? Should we have an allowlist of variables that are valid? (though not sure how that would look).

berciuliviu:
> The PR looks good, but are we sure we want to remove all environment variables?
> For example, there is the $TZ environment variable that is checked by libc when calling `localtime()` (which we are callling in Firecracker), to retrieve the local timezone.
> 
> Are we blocking users from leveraging valid environment variables? Should this be a switch? Should we have an allowlist of variables that are valid? (though not sure how that would look).

I am also thinking we could use a separate, configuration file, that would play the role of an allowlist for ENV variables. That way, if someone wants to add ENV variables to the jailer environment, they could be added in that file and the Jailer could load them. Or it could just ignore them by name when performing the removal. The file could be optional and the cost would be an extra file to manage in case someone wants env vars.

alindima:
We've discussed internally and decided that the approach of this PR is the right one. We will begin by clearing up all environment variables and state this in the Changelog. This should generally not pose problems, since these are variables that are only used by the Firecracker process, not injected in the VM in any way (and is also a default for other jailers. like runc).
If it turns out that users need to retain some of the host env vars, we may revisit this decision and add an allowlist/configurable list.

So let's get a second reviewer @firecracker-microvm/compute-capsule

PR #2702, opened 2021-08-20
[block] simplify unsafe flush behavior

serban300:
# Reason for This PR

[block] simplify unsafe flush behavior

## Description of Changes

Leverage the VIRTIO_BLK_F_FLUSH flag in order to avoid receiving flush
requests when the block device uses CacheType::Unsafe. This has 2
advantages:
1. It simplifies the request execution logic a bit
2. It improves efficiency when the block device uses CacheType::Unsafe

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/request.rs
+++ b/src/devices/src/virtio/block/request.rs
@@ -14,7 +14,7 @@ use virtio_gen::virtio_blk::*;
 use vm_memory::{ByteValued, Bytes, GuestAddress, GuestMemoryError, GuestMemoryMmap};
 
 use super::super::DescriptorChain;
-use super::device::{CacheType, DiskProperties};
+use super::device::DiskProperties;
 use super::{Error, SECTOR_SHIFT, SECTOR_SIZE};
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


src/devices/src/virtio/block/request.rs
```diff
@@ -283,22 +281,15 @@ impl Request {
                     })
             }
             RequestType::Flush => {
-                match cache_type {
```
georgepisaltu, line 286: Don't we want to still check if a flush request comes even if we didn't negotiate it? Drivers are part of the guest, which is not trusted. We should still check the cache strategy for flush requests and return an error if they don't match.

georgepisaltu (commented):
Do we have some measurement for the improvement in performance? I expect this change to speed things up as there would normally be no flush requests to process, but I was wondering if we have some numbers.

serban300:
> Do we have some measurement for the improvement in performance? I expect this change to speed things up as there would normally be no flush requests to process, but I was wondering if we have some numbers.

I did some manual tests and we get an improvement of roughly 2-3%

src/devices/src/virtio/block/request.rs
serban300: We should still rely on the fact that the guest is compliant with the virtio standard. Otherwise a lot of things wouldn't work. In this case if the guest doesn't behave correctly it would just slow itself down. Personally I don't think there's a risk here.

dianpopa (dismissed):

georgepisaltu (approved):

dianpopa (approved):

georgepisaltu (approved):

PR #2470, opened 2021-02-23
Linux loader integration

AlexandruCihodaru:
# Reason for This PR

Replaced kernel crate with rust-vmm/linux-loader
See https://github.com/firecracker-microvm/firecracker/issues/1182.

## Description of Changes

Added github dependency for rust-vmm/linux-loader
Deleted kernel crate from Firecracker
Use linux-loader to write bootparams
Integrated rust-vmm/linux-loader in vmm crate

As for right now the snapshot tests will not pass since the rust compiler uses upstream implementations of traits from vm-memory.

- [ ] This functionality can be added in [`rust-vmm`](https://github.com/rust-vmm).

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

`[Author TODO: Meet these criteria.]`
`[Reviewer TODO: Verify that these criteria are met. Request changes if not]`

- [x] All commits in this PR are signed (`git commit -s`).
- [x] The reason for this PR is clearly provided (issue no. or explanation).
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] Any newly added `unsafe` code is properly documented.
- [x] Any API changes are reflected in `firecracker/swagger.yaml`.
- [x] Any user-facing changes are mentioned in `CHANGELOG.md`.
- [x] All added/changed functionality is tested.

PR #2716, opened 2021-09-07
kernel: fix bug in cmdline setup

### LINKED ISSUE #2709: [Bug] virtio-mmio kernel cmdline append breaks init arguments separated by `--`
# Describe the bug

Currently, it is not possible to pass arguments to init using `--` ([ref](https://www.kernel.org/doc/html/latest/admin-guide/kernel-parameters.html)) as extra arguments are appended after the user provided command line [here](https://github.com/firecracker-microvm/firecracker/blob/cf46cb18dbc0a5dad6d19a69e44a80e21d3e43e3/src/vmm/src/device_manager/mmio.rs#L228-L233). When a user uses `--`, the behaviour causes essential arguments to be ignored as they will be passed to init.

In a bit of a contrived example, but for illustrations sake, let's imagine my init was to `/bin/cat` a file.
When specifying kernel `boot_args` such as `console=ttyS0 nomodules random.trust_cpu=on reboot=k panic=1 pci=off ip=:::::eth0:dhcp init=/bin/cat -- /etc/password`, the  aforementioned code block will append extra arguments which technically get discarded because of the the `--`. The resultant command line would therefore be:

```text
[    0.000000] Command line: console=ttyS0 nomodules random.trust_cpu=on reboot=k panic=1 pci=off ip=:::::eth0:dhcp init=/bin/cat -- /etc/password root=/dev/vda rw virtio_mmio.device=4K@0xd0000000:5 virtio_mmio.device=4K@0xd0001000:6 virtio_mmio.device=4K@0xd0002000:7
```

## To Reproduce

1. Start firecracker with `firecracker --api-sock /tmp/firecracker.socket`.
2. Configure the kernel with the following JSON body:

```json
{
   "kernel_image_path":"/path/to/kern",
   "boot_args":"console=ttyS0 nomodules random.trust_cpu=on reboot=k panic=1 pci=off ip=:::::eth0:dhcp init=/bin/cat -- /etc/password"
}
```
3. Configure other parts to get a VM to boot.
4. Start the instance and observe the `Command line:` output.

## Expected behaviour

Kernel arguments such as `root=` and `virtio_mmio.device=` should appear before user supplied arguments.

## Environment

1. Arch : x86_64
2. OS: Ubuntu 21.04
3. Firecracker 0.24.6

## Checks

dianpopa:
# Reason for This PR

Fixes #2709.
As per the documentation, everything that is
passed after "--" to the commandline is considered
an argument to init.
However, in Firecracker we are dynamically adjusting the
commandline by appending the virtio mmio devices and the
root device. As such, using "--" with Firecracker would result
in faulty behavior.

## Description of Changes

## License Acceptance

## PR Checklist


src/kernel/src/cmdline/mod.rs
```diff
@@ -131,24 +131,31 @@ impl Cmdline {
...
+            self.line.push_str(&format!("{}={}", k, v))
+        }
         self.end_push();
 
         Ok(())
     }
 
-    /// Validates and inserts a string to the end of the current command line.
+    /// Validates and inserts a string into the commandline. The position of the
+    /// string depends on the presence of  "--".
```
luminitavoicu, line 146: I would also specify here that string is inserted before `--`, if present. It's not clear from the description.

luminitavoicu (commented):
Looks good!

src/kernel/src/cmdline/mod.rs
```diff
@@ -131,24 +131,31 @@ impl Cmdline {
...
-        self.start_push();
-        self.line.push_str(k);
-        self.line.push('=');
-        self.line.push_str(v);
+        if let Some(index) = self.line.find("--") {
+            self.line.insert_str(index, &format!("{}={} ", k, v));
+        } else {
+            self.start_push();
+            self.line.push_str(&format!("{}={}", k, v));
+        }
```
serban300, line 139: In order to make this logic simpler and more generic, could we just rename this method from `insert` to `prepend` and insert these custom strings at the beginning of the cmdline ? Would this work? This way they would always be insreted before `--`.

dianpopa: I like this suggestion a lot. Take a look.

dianpopa (commented):
@luminitavoicu @serban300 take a look at this new version. Thanks!

src/vmm/src/vmm_config/boot_source.rs
```diff
@@ -102,7 +102,7 @@ impl BootConfig {
             Some(str) => str.as_str(),
         };
         cmdline
-            .insert_str(boot_args)
+            .prepend_str(boot_args)
```
serban300, line 105: Nitish: We might get away with this because this step is peformed before adding the virtio devices. But in order to make absolutelty certain that everything is ok we would need to make sure that the  `boot_args` are actually appended and everything else is prepended. So do you think we should also have an `append_str()` method ?

andreeaflorescu:
Since in #2470 we are preparing to consume the upstream linux-loader, should we fix this in upstream rust-vmm?

src/vmm/src/vmm_config/boot_source.rs
dianpopa: The cmdline is empty when the boot args are "prepended". This is actualy a workaround to the fact the Cmdline does not have a way of getting initialized to a certain value. 
However, I just gave it a second thought and it might be that the prepend method is not a very nice choice. Take a look at this: https://github.com/firecracker-microvm/firecracker/blob/200c2db7055e406125729c40e14acf47024c1420/src/vmm/src/builder.rs#L849.
We are appending the ro/rw to the string specifying the root device. With the new prepend method, we would need to first create the entire string root=dev/vda ro/rw and after that call into the Cmdline. IMO this makes things a little counterintuitive. 
So I am thinking of reverting to the previous insert_str version. WDYT?

dianpopa:
> Since in #2470 we are preparing to consume the upstream linux-loader, should we fix this in upstream rust-vmm?
> 
> LE: here is the cmdline functionality: https://github.com/rust-vmm/linux-loader/blob/main/src/cmdline/mod.rs

I am thinking of fixing it in both repos in order to unblock things [here](https://github.com/firecracker-microvm/firecracker/issues/2709).

src/vmm/src/vmm_config/boot_source.rs
serban300: I see. There are other options. We can have 2 "regions" in the string:
- one for the firecracker custom args (virtio devices, etc) at the beginning
- one for the boot_args at the end
We can keep track of them by keeping the end index for each one. And have 2 groups of methods: 
- some that insert firecracker custom args. These would go in the first region.
- some that insert boot_args. These would go in the second region

Or there might be simpler methods to keep track of the regions. Or we can have 2 strings and concat them when `as_string` is called. It shouldn't add a significant overhead.

What do you think ?

andreeaflorescu:
> I am thinking of fixing it in both repos in order to unblock things [here](https://github.com/firecracker-microvm/firecracker/issues/2709).

Nice, thanks for opening the issue in rust-vmm as well. I'll have a look at the PR as well in this case so we can keep both implementations aligned.

src/vmm/src/vmm_config/boot_source.rs
andreeaflorescu: In rust-vmm we already started adding helper methods on top of `insert*` functions. For example, we have a function for [`add_virtio_mmio_device`](https://github.com/rust-vmm/linux-loader/blob/ff3ae8bd44a8f6b33713e4eeae1781751ff8c959/src/cmdline/mod.rs#L292).

Having a new function for `insert_boot_args` sounds like a good idea to me, as it is inline with existing helpers. This would come with the advantage that the existing `Cmdline` interface stays the same, so we do not introduce breaking changes for existing customers. With this, most of the changes will be in the implementation details of `Cmdline`.

In the future we would like to look at other improvements to `Cmdline` as well such as having it implemented with the builder pattern and allow even more custom/commonly used parameters as specified here: https://github.com/rust-vmm/linux-loader/issues/25.

dianpopa: Nice info! ok, so since the direction in which rust-vmm's linux-loader has already shifted in the same way we want to go with this PR, I will open a PR there, merge it there and then bring the changes here. Thanks @andreeaflorescu for the heads up

AlexandruCihodaru:
Considering that merging of  #2661  unlocked #2470 should we considering adding needed support just in rust-vmm/linux-loader and then cherry-picking your commits on top of #2470?

dianpopa:
> Considering that merging of #2661 unlocked #2470 should we consider adding needed support just in rust-vmm/linux-loader and then cherry-picking your commits on top of #2470?

Indeed, since we are now close to moving to rust-vmm's linux-loader, I will do the fix there.

dianpopa:
As per the latest [comment in rust-vmm's linux-loade](https://github.com/rust-vmm/linux-loader/pull/94#issuecomment-929339137)r, we agreed that the cmdline implementation needs to be refactored in order to cleanly accommodate the changes required for this fix. As such we decided to implement a temporary fix in Firecracker. The fix was provided by @andreeaflorescu, thanks a lot for this!
@firecracker-microvm/compute-capsule PTAL!

PR #2725, opened 2021-09-29
Limit data store

AlexandruCihodaru:
# Reason for This PR

- Make sure that MMDS data store size is less than the payload limit of requests.

## Description of Changes
- Limit MMDS data store size.

## License Acceptance

## PR Checklist

### DIFF
--- a/docs/mmds/mmds-design.md
+++ b/docs/mmds/mmds-design.md
@@ -58,6 +58,10 @@ the MMDS contents. It leverages the recursive
 store supports at the moment storing and retrieving JSON values. Data store
 contents can be retrieved using the Firecracker API server from host and using
 the embedded MMDS HTTP/TCP/IPv4 network stack from guest.
+MMDS data store is upper bounded by a default maximum size of 51200 bytes. This
+limit can be modified using `--http_api_max_payload_size` command line parameter.
+All PUT and PATCH requests with Content-Length bigger than the imposed limit will
+receive HTTP 413 Payload Too Large response status code.
 
 ## Dumbo
 
--- a/src/mmds/src/data_store.rs
+++ b/src/mmds/src/data_store.rs
@@ -1,7 +1,8 @@
 // Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use serde_json::Value;
+use crate::MAX_DATA_STORE_SIZE;
+use serde_json::{to_vec, Value};
 use std::fmt;
 
 /// The Mmds is the Microvm Metadata Service represented as an untyped json.
@@ -9,6 +10,7 @@ use std::fmt;
 pub struct Mmds {
     data_store: Value,
     is_initialized: bool,
+    data_store_limit: usize,
 }
 
 /// MMDS possible outputs.
@@ -22,6 +24,7 @@ pub enum Error {
     NotFound,
     NotInitialized,
     UnsupportedValueType,
+    DataStoreLimitExceeded,
 }
 
 impl fmt::Display for Error {
@@ -33,6 +36,7 @@ impl fmt::Display for Error {
                 f,
                 "Cannot retrieve value. The value has an unsupported type."
             ),
+            Error::DataStoreLimitExceeded => write!(f, "The MMDS patch request doesn't fit."),
         }
     }
 }
@@ -42,6 +46,7 @@ impl Default for Mmds {
         Mmds {
             data_store: Value::default(),
             is_initialized: false,
+            data_store_limit: MAX_DATA_STORE_SIZE,
         }
     }
 }
@@ -58,6 +63,13 @@ impl Mmds {
         }
     }
 
+    pub fn set_data_store_limit(&mut self, data_store_limit: usize) {
+        self.data_store_limit = data_store_limit;
+    }
+
+    // We do not check data_store size here because a request with a body
+    // bigger than the imposed limit will be stoped by micro_http before
+    // reaching here.
     pub fn put_data(&mut self, data: Value) -> Result<(), Error> {
         self.data_store = data;
         self.is_initialized = true;
@@ -66,10 +78,21 @@ impl Mmds {
 
     pub fn patch_data(&mut self, patch_data: Value) -> Result<(), Error> {
         self.check_data_store_initialized()?;
-        super::json_patch(&mut self.data_store, &patch_data);
+        let mut data_store_clone = self.data_store.clone();
+
+        super::json_patch(&mut data_store_clone, &patch_data);
+        // It is safe to unwrap because our data store keys are all strings and
+        // we are using default serializer which does not return error.
+        if to_vec(&data_store_clone).unwrap().len() > self.data_store_limit {
+            return Err(Error::DataStoreLimitExceeded);
+        }
+        self.data_store = data_store_clone;
         Ok(())
     }
 
+    // We do not check size of data_store before returning a result because due
+    // to limit from put/patch the data_store can not be bigger than the limit
+    // imposed by the server.
     pub fn get_data_str(&self) -> String {
         if self.data_store.is_null() {
             return String::from("{}");
@@ -346,6 +369,7 @@ mod tests {
     #[test]
     fn test_update_data_store() {
         let mut mmds = Mmds::default();
+        mmds.set_data_store_limit(MAX_DATA_STORE_SIZE);
 
         let data = r#"{
             "name": {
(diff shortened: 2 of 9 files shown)


src/mmds/src/data_store.rs
```diff
@@ -66,7 +76,12 @@ impl Mmds {
 
     pub fn patch_data(&mut self, patch_data: Value) -> Result<(), Error> {
         self.check_data_store_initialized()?;
-        super::json_patch(&mut self.data_store, &patch_data);
+        let mut aux = self.data_store.clone();
```
luminitavoicu, line 79: Let's rename this `data_store_clone` or something more specific.

src/mmds/src/data_store.rs
```diff
@@ -386,5 +402,31 @@ mod tests {
...
+        let data_store: Value = serde_json::from_str(&data).unwrap();
+        assert!(mmds.patch_data(data_store).is_ok());
+        assert!(mmds.get_data_str().contains("smth"));
+        assert_eq!(mmds.get_data_str().len(), 53);
+
+        let data = "{\"new_key2\" : \"smth2\"}";
+        let data_store: Value = serde_json::from_str(&data).unwrap();
+        assert!(mmds.patch_data(data_store).is_ok());
+        assert!(mmds.get_data_str().contains("smth2"));
+        assert_eq!(mmds.get_data_str().len(), 72);
```
luminitavoicu, line 430: I think we can get rid of this, as it doesn't test a different case than the part above and I find it redundant.

tests/integration_tests/functional/test_mmds.py
```diff
@@ -478,3 +478,92 @@ def test_guest_mmds_hang(test_microvm_with_api, network_config):
 
     _, stdout, _ = ssh_connection.execute_command(cmd)
     assert 'Invalid request' in stdout.read()
+
+
+def test_patch_dos_scenario(test_microvm_with_api):
+    """
+    Test the MMDS json endpoint when data store size reach the limit.
```
luminitavoicu, line 485:
```suggestion
    Test the MMDS json endpoint when data store size reaches the limit.
```

tests/integration_tests/functional/test_mmds.py
```diff
@@ -478,3 +478,92 @@ def test_guest_mmds_hang(test_microvm_with_api, network_config):
...
+
+    @type: functional
```
luminitavoicu, line 487: This should be:
```suggestion
    @type: negative
```

tests/integration_tests/functional/test_mmds.py
```diff
@@ -478,3 +478,92 @@ def test_guest_mmds_hang(test_microvm_with_api, network_config):
...
+                'secret_key2': aux
+            }
+        }
+    }
+    response = test_microvm.mmds.patch(json=dummy_json)
+    assert test_microvm.api_session.\
+        is_status_payload_too_large(response.status_code)
+
+    # Check that the patch actually failed and we can not find the string that
+    # we tried to add.
```
luminitavoicu, line 534: nit:
```suggestion
    # Check that the patch actually failed and the contents of the data store 
    has not changed.
```

tests/integration_tests/functional/test_mmds.py
```diff
@@ -478,3 +478,92 @@ def test_guest_mmds_hang(test_microvm_with_api, network_config):
...
+        }
+    }
+    response = test_microvm.mmds.patch(json=dummy_json)
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+
+    # Check that the size has shrunk.
+    response = test_microvm.mmds.get()
+    assert len(str(response.json()).replace(" ", "")) == 59
+
+    # Try to send a new patch, this time the request should succed.
```
luminitavoicu, line 554:
```suggestion
    # Try to send a new patch, this time the request should succeed.
```

src/mmds/src/data_store.rs
AlexandruCihodaru: it is not redundant, it actually tests that you can add new data once there is enough space in data_store, in this case particularly after we delete some data with a patch we are able to add new data.

src/mmds/src/data_store.rs
```diff
@@ -42,6 +45,7 @@ impl Default for Mmds {
         Mmds {
             data_store: Value::default(),
             is_initialized: false,
+            data_store_limit: 51200_usize,
```
georgepisaltu, line 48: Is the `_usize` necessary? Can't the compiler infer the type?

src/mmds/src/data_store.rs
```diff
@@ -58,6 +62,12 @@ impl Mmds {
         }
     }
 
+    pub fn set_data_store_limit(&mut self, maybe_data_store_limit: Option<usize>) {
```
georgepisaltu, line 65: This should not take an `Option<usize>` but an `usize`. When someone is calling a set, they should have already validated that they indeed have a limit they want to enforce.

The check here should be done by the caller of this function.

src/mmds/src/data_store.rs
```diff
@@ -66,7 +76,12 @@ impl Mmds {
 
     pub fn patch_data(&mut self, patch_data: Value) -> Result<(), Error> {
         self.check_data_store_initialized()?;
-        super::json_patch(&mut self.data_store, &patch_data);
+        let mut data_store_clone = self.data_store.clone();
+        super::json_patch(&mut data_store_clone, &patch_data);
+        if data_store_clone.to_string().len() > self.data_store_limit {
+            return Err(Error::DataStoreLimitExceeded);
+        }
+        self.data_store = data_store_clone;
```
georgepisaltu, line 84: I think we can avoid the `to_string` here by checking the current length of the data store. As it is a `Value`, we can do `data_store.as_str().unwrap().as_bytes().len()` (unsure if we should unwrap or do error handling). It avoids extra allocation and deallocation of the `String` object.

georgepisaltu (commented):
Overall looks good.

I think we should also check MMDS Puts like we're doing for Patches.

src/mmds/src/data_store.rs
AlexandruCihodaru: Your observation is a valid one but according to `Value` documentation as_str will return some("a string") only when Value is a String, otherwise the result will be None.

In our use case this does not really makes sense because we usually store maps.

AlexandruCihodaru:
> Overall looks good.
> 
> I think we should also check MMDS Puts like we're doing for Patches.

I think this is not necessary, since for a PUT requests the size of new json is not added to the old MMDS but it replaces it. Thus, for PUT requests there is already the payload limit in place. Hence, it will stop the request before reaching the function that handles puts on server side.

georgepisaltu:
> > Overall looks good.
> > I think we should also check MMDS Puts like we're doing for Patches.
> 
> I think this is not necessary, since for a PUT requests the size of new json is not added to the old MMDS but it replaces it. Thus, for PUT requests there is already the payload limit in place. Hence, it will stop the request before reaching the function that handles puts on server side.

That seems fine. Let's just add comments for the Put and Get methods stating why we don't check the size here like we do for patches and we're good to go.

src/mmds/src/data_store.rs
luminitavoicu: I see your point. This sounds good.

luminitavoicu (dismissed):
Looks good to me!

georgepisaltu (approved):

tests/integration_tests/functional/test_mmds.py
```diff
@@ -478,3 +478,91 @@ def test_guest_mmds_hang(test_microvm_with_api, network_config):
...
+                'secret_key': aux
+            }
+        }
+    }
+    response = test_microvm.mmds.patch(json=dummy_json)
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+
+    # Check that the size grew as expected.
+    response = test_microvm.mmds.get()
+    assert len(str(response.json()).replace(" ", "")) == 158
```
dianpopa, line 568: We should also test this use case when updating the data store from inside the guest? You can find an example in the other tests inside test_mmds.py

dianpopa (commented):
The changes look good, however I think we should:
1. update the documentation; we are not mentioning anywhere in our docs this newly added feature: https://github.com/firecracker-microvm/firecracker/blob/main/docs/mmds/mmds-design.md
2. test the limit scenario when updating the data store from inside the guest

tests/integration_tests/functional/test_mmds.py
dianpopa: Ignore this, PUT requests are only accepted on v2 not on v1.

docs/mmds/mmds-design.md
```diff
@@ -58,6 +58,10 @@ the MMDS contents. It leverages the recursive
 store supports at the moment storing and retrieving JSON values. Data store
 contents can be retrieved using the Firecracker API server from host and using
 the embedded MMDS HTTP/TCP/IPv4 network stack from guest.
+MMDS data store is upper bounded, its default maximum size is 51200 bytes but
+can be updated using `--http_api_max_payload_size`. It can not be updated using
+PATCH requests with Content-Length bigger than the imposed limit. PUT requests
+body have to respect the same limit.
```
luminitavoicu, line 64: I feel the current using of `it` pronoun is a bit confusing here. Maybe try something like this:

```suggestion
MMDS data store is upper bounded by a default maximum size of 51200 bytes. This
limit can be modified using `--http_api_max_payload_size` command line parameter.
All PUT and PATCH requests with Content-Length bigger than the imposed limit will
receive HTTP 413 Payload Too Large response status code.
```

luminitavoicu (approved):
:ship:

georgepisaltu (approved):

PR #2787, opened 2021-11-15
block: expose io_engine in the API and add Async option

alindima:
# Reason for This PR

We need to expose the io_uring block device IO engine in the API

## Description of Changes

View commit messages

## License Acceptance

## PR Checklist


docs/api_requests/block-io-engine.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+## Host requirements
+
+Firecracker requires a minimum host kernel version of 5.10.51 for the `Async`
+IO engine.
+
+This requirement is based on the availability of the `io_uring` subsystem, as
+well as a couple of features and bugfixes that were added in newer kernel
+versions.
```
raduiliescu, line 45: Maybe some recommendations when to use one or the other.
Also you should mention what happens when one uses io_uring with kernel version lower than 5.10.

src/utils/src/kernel_version.rs
```diff
@@ -72,10 +72,23 @@ impl KernelVersion {
...
+    () => {
+        if KernelVersion::get().unwrap() < min_kernel_version_for_io_uring() {
+            return;
+        }
+    };
+}
+
 #[macro_export]
-macro_rules! skip_if_kernel_lt_5_10 {
+macro_rules! skip_if_io_uring_supported {
```
raduiliescu, line 89: I haven't seen a test to check that you are defaulting to the sync version when kernel version is lower than 5.10.

alindima: Added this to the `test_drive_io_engine `

docs/api_requests/block-io-engine.md
alindima: I added a couple of recommendations, which we will have to revisit once we complete the snapshotting support. I'd also like to get Serban's input on them.
I also mentioned what happens when you try to use io_uring on a kernel older than 5.10

tests/integration_tests/functional/test_cmd_line_parameters.py
```diff
@@ -24,8 +24,12 @@ def test_describe_snapshot_all_versions(bin_cloner_path):
     # Fetch all firecracker binaries.
     # For each binary create a snapshot and verify the data version
     # of the snapshot state file.
+
+    # TODO: FIXME: Once we upload 1.0.0 binaries in
+    # S3 that include io_uring, replace the
+    # max_version with get_version_from_toml()
     firecracker_artifacts = artifacts.firecrackers(
-        max_version=get_firecracker_version_from_toml())
+        max_version="0.25.0")
```
serban300, line 32: This can create problems. We might forget to do this or not understand why the tests are failing before the release. Could we add a binary build from this branch instead of the current 1.0.0 binary ? Would the tests pass on the main branch ?

tests/integration_tests/functional/test_api.py
```diff
@@ -37,6 +38,67 @@ def test_api_happy_start(test_microvm_with_api):
...
+        path_on_host=test_microvm.create_jailed_resource(
+            test_microvm.rootfs_file),
+        is_root_device=True,
+        is_read_only=False,
+        # Set the opposite of the default backend type.
+        io_engine="Sync" if compare_versions(
+            get_kernel_version(), "5.10.0") > 0 else "Async"
+    )
+
+    if compare_versions(get_kernel_version(), "5.10.0") < 0:
```
serban300: This is increasingly more used lately. I think helper methods would be useful, similar to what we have in the rust code. Something like `is_io_uring_supported`.

src/devices/src/virtio/block/io/mod.rs
```diff
@@ -49,20 +54,19 @@ pub enum FileEngine<T> {
...
-        Ok(FileEngine::Async(
-            AsyncFileEngine::from_file(file).map_err(Error::Async)?,
-        ))
+    pub fn from_file(file: File, engine_type: FileEngineType) -> Result<FileEngine<T>, Error> {
+        if !engine_type
+            .is_supported()
+            .map_err(Error::GetKernelVersion)?
+        {
+            return Err(Error::UnsupportedEngine(engine_type));
+        }
```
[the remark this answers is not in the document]
serban300: > I think the restore() implementation of the Block should check for the UnsupportedEngine error and fallback to Sync if it's unsupported. I've done this, WDYT?

Sounds good. 

> we should also add an integration test for this, that resumes a snapshot created with the Async engine on an older kernel.

I think such a test would be very hard to write with our current infrastructure. Only if we could write a unit test in which to somehow mock the `KernelVersion`. I wouldn't add an issue for this.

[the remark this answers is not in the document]
alindima: I was able to write a unit test in `persist.rs`, which verifies this on kernels that don't support io_uring.
It overwrites the `BlockState` struct after it's saved.

tests/integration_tests/functional/test_cmd_line_parameters.py
serban300: Another folder would complicate things. It's true that uploading a binary build from this branch wouldn't be ideal either. But as long as it doesn't break the functionality from the main branch personally I would prefer the latter. Anyway I'm curious what others think too.

I will try to post a PR for adding snapshotting support as soon as possible. Anyway I ran some manual tests and I didn't encounter problems. In general it should work.

Apart from this the PR LGTM.

alindima: ok I suggest we merge the PR as is and upload binaries and re-enable the tests after the snapshotting support is finalised, since we'll have to re-upload the artifacts anyway. sounds good?

serban300: yes, sounds good to me.

tests/framework/resources.py
```diff
@@ -192,16 +194,45 @@ class Drive():
...
+        return self._api_session.put(
+            "{}/{}".format(self._drive_cfg_url, args['drive_id']),
+            json=datax
+        )
+
+    def put_with_default_io_engine(self, **args):
+        """
+        Attach a block device or update the details of a previous one, using...
+
+        ...the Firecracker default for the io_engine.
```
luminitavoicu, line 234: nit
```suggestion
        Attach a block device or update the details of a previous one, using
        the Firecracker default for the io_engine.
```

tests/integration_tests/functional/test_api.py
```diff
@@ -37,6 +38,67 @@ def test_api_happy_start(test_microvm_with_api):
...
+        # The Async engine is not supported for older kernels.
+        assert test_microvm.api_session.is_status_bad_request(
+            response.status_code)
+
+        # time.sleep(0.1)
```
luminitavoicu, line 73: leftover

docs/api_requests/block-io-engine.md
```diff
@@ -0,0 +1,66 @@
+# Block device IO engine
+
+Prior to Firecracker 1.0.0, the emulated block device uses a synchronous IO
```
luminitavoicu, line 3:
```suggestion
Prior to Firecracker 1.0.0, the emulated block device used a synchronous IO
```

docs/api_requests/block-io-engine.md
```diff
@@ -0,0 +1,66 @@
...
+engine for executing the device requests, based on blocking system calls.
+
+Firecracker 1.0.0 adds support for a new, asynchronous block device IO engine.
```
luminitavoicu, line 6: I wouldn't add a comma here.
```suggestion
Firecracker 1.0.0 adds support for a new asynchronous block device IO engine.
```

docs/api_requests/block-io-engine.md
```diff
@@ -0,0 +1,66 @@
...
+
+The `Async` engine leverages [`io_uring`](https://kernel.dk/io_uring.pdf) for
+executing requests in an async manner, therefore getting overall higher
+throughput by taking better advantage of the block device hardware which
```
luminitavoicu, line 10: nit:
```suggestion
throughput by taking better advantage of the block device hardware, which
```

tests/framework/utils.py
```diff
@@ -615,3 +617,14 @@ def get_kernel_version():
             break
 
     return linux_version
+
+
+def is_io_uring_supported():
+    """
+    Return whether Firecracker supports io_uring for the running kernel ...
+
+    ...version.
```
luminitavoicu, line 626: nit: 
```suggestion
    Return whether Firecracker supports io_uring for the running kernel
    version.
```

tests/framework/resources.py
```diff
@@ -192,16 +194,31 @@ class Drive():
...
         self._drive_cfg_url = api_url + self.DRIVE_CFG_RESOURCE
         self._api_session = api_session
+        self._firecracker_version = firecracker_version
 
     def put(self, **args):
         """Attach a block device or update the details of a previous one."""
+        # Default the io engine to Async on kernels > 5.10 so that we
+        # make sure to exercise both Sync and Async behaviour in the CI.
+        # Also check the FC version to make sure that it has support for
+        # configurable io_engine.
```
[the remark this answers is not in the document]
alindima: Yes, there are a couple of tests that do:
`test_drive_io_engine`
`test_get_full_config`
`test_drive_patch`

tests/framework/utils.py
luminitavoicu: I see, that's fine then.

docs/api_requests/block-io-engine.md
luminitavoicu: I get your reasoning, but it just sounds a bit weird to me to keep present tense here. How about rephrasing to something like:

```suggestion
For all Firecracker versions prior to v1.0.0, the emulated block device uses a synchronous IO 
engine for executing the device requests, based on blocking system calls for.
```
so the it would be clear that firecracker supports Sync for all versions and the next paragraph specifies that Async is only available starting with Firecracker v1.0.0.

alindima: done 👍🏻

luminitavoicu (approved):
:+1:

serban300 (approved):

PR #2790, opened 2021-11-18
Add custom headers for MMDS requests

luminitavoicu:
# Reason for This PR

In order to be able to add IMDSv2 capabilities to Firecracker's MMDS, the backend needs to be able to process custom mmds-related headers. 

IMDSv2 uses session-oriented requests. With session-oriented requests, the guest creates a session token that defines the session duration and then uses that token for subsequent requests.

## Description of Changes

- Make use of micro-http's custom headers hash-map to fetch custom headers when processing the header of a request
- Add MMDS v2 headers:
  `X-metadata-ttl-seconds` - used when fetching the session token to provide token lifetime in seconds
  `X-metadata-token` - used to provide session token

## License Acceptance

## PR Checklist

- ~[] The issue which led to this PR has a clear conclusion.~

### DIFF
--- a/src/mmds/src/lib.rs
+++ b/src/mmds/src/lib.rs
@@ -4,9 +4,12 @@
 pub mod data_store;
 pub mod ns;
 pub mod persist;
+pub mod token_headers;
 
 use serde_json::{Map, Value};
 use std::sync::{Arc, Mutex};
+
+use token_headers::TokenHeaders;
 
 use crate::data_store::{Error as MmdsError, Mmds, OutputFormat};
 use lazy_static::lazy_static;
@@ -96,6 +99,17 @@ fn convert_to_response(request: Request) -> Response {
         response.allow_method(Method::Get);
         return response;
     }
+
+    let _token_headers = match TokenHeaders::try_from(request.headers.custom_entries()) {
+        Ok(token_headers) => token_headers,
+        Err(err) => {
+            return build_response(
+                request.http_version(),
+                StatusCode::BadRequest,
+                Body::new(err.to_string()),
+            )
+        }
+    };
 
     // The data store expects a strict json path, so we need to
     // sanitize the URI.
@@ -217,6 +231,30 @@ mod tests {
         let request = Request::try_from(request_bytes, None).unwrap();
         let mut expected_response = Response::new(Version::Http10, StatusCode::BadRequest);
         expected_response.set_body(Body::new("Invalid URI.".to_string()));
+        let actual_response = convert_to_response(request);
+        assert_eq!(actual_response, expected_response);
+
+        // Test invalid value for custom header.
+        let request_bytes = b"GET http://169.254.169.254/ HTTP/1.0\r\n\
+                                    Accept: application/json\r\n
+                                    X-metadata-token-ttl-seconds: application/json\r\n\r\n";
+        let request = Request::try_from(request_bytes, None).unwrap();
+        let mut expected_response = Response::new(Version::Http10, StatusCode::BadRequest);
+        expected_response.set_body(Body::new(
+            "Invalid header. Reason: Invalid value. \
+            Key:X-metadata-token-ttl-seconds; Value:application/json"
+                .to_string(),
+        ));
+        let actual_response = convert_to_response(request);
+        assert_eq!(actual_response, expected_response);
+
+        // Test valid values for custom headers.
+        let request_bytes = b"GET http://169.254.169.254/name/first HTTP/1.0\r\n\
+                                    X-metadata-token: application/json\r\n
+                                    X-metadata-token-ttl-seconds: 100\r\n\r\n";
+        let request = Request::try_from(request_bytes, None).unwrap();
+        let mut expected_response = Response::new(Version::Http10, StatusCode::OK);
+        expected_response.set_body(Body::new("John"));
         let actual_response = convert_to_response(request);
         assert_eq!(actual_response, expected_response);
 
--- a/src/mmds/src/lib.rs
+++ b/src/mmds/src/lib.rs
@@ -9,11 +9,12 @@ pub mod token_headers;
 use serde_json::{Map, Value};
 use std::sync::{Arc, Mutex};
 
-use token_headers::TokenHeaders;
+use crate::data_store::{Error as MmdsError, Mmds, MmdsVersion, OutputFormat};
 
-use crate::data_store::{Error as MmdsError, Mmds, OutputFormat};
 use lazy_static::lazy_static;
 use micro_http::{Body, MediaType, Method, Request, Response, StatusCode, Version};
+use token_headers::TokenHeaders;
+
 pub const MAX_DATA_STORE_SIZE: usize = 51200;
 
 lazy_static! {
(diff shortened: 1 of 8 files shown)


src/mmds/src/token_headers.rs
```diff
@@ -0,0 +1,141 @@
...
+        let mut headers = Self::default();
+
+        if let Some(token) = map.get(&*TokenHeader::XMetadataToken.to_string()) {
+            headers.x_metadata_token = Some(token.to_string());
+        }
+
+        if let Some(value) = map.get(&*TokenHeader::XMetadataTokenTtlSeconds.to_string()) {
+            match value.parse::<u32>() {
+                Ok(seconds) => {
+                    headers.x_metadata_token_ttl_seconds = Some(seconds);
```
serban300, line 63: Should we validate if `seconds <= MAX_TTL` ?

src/mmds/src/token_headers.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+}
+
+impl TokenHeaders {
+    /// Return `TokenHeaders` from headers map.
+    pub fn try_from(map: &HashMap<String, String>) -> Result<TokenHeaders, RequestError> {
+        let mut headers = Self::default();
+
+        if let Some(token) = map.get(&*TokenHeader::XMetadataToken.to_string()) {
```
serban300, line 56: is the `*` needed ? The same question for the next `if`.

src/mmds/src/token_headers.rs
```diff
@@ -0,0 +1,141 @@
...
+impl Display for TokenHeader {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match *self {
+            TokenHeader::XMetadataToken => write!(f, "X-metadata-token"),
+            TokenHeader::XMetadataTokenTtlSeconds => {
+                write!(f, "X-metadata-token-ttl-seconds")
+            }
+        }
+    }
+}
```
serban300, line 28: We could also define these values as constants associated with the `TokenHeaders` struct:
```
impl TokenHeaders {
    const X_METADATA_TOKEN: &'static str = "X-metadata-token";
    const X_METADATA_TOKEN_TTL_SECONDS: &'static str = "X-metadata-token-ttl-seconds";
    ...
}
```

Not sure if it's necessarily better, but I have a slight preference towards this option.

src/mmds/src/token_headers.rs
luminitavoicu: We do validate this as part of the token generation process.

src/mmds/src/token_headers.rs
luminitavoicu: I think it's a good idea. Done.

src/mmds/src/token_headers.rs
luminitavoicu: It was a leftover. Good catch!

src/mmds/src/token_headers.rs
```diff
@@ -0,0 +1,144 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use micro_http::{HttpHeaderError, RequestError};
+use std::collections::HashMap;
+use std::fmt;
+use std::fmt::{Display, Formatter};
+use std::result::Result;
+
+const X_METADATA_TOKEN: &str = "X-metadata-token";
+const X_METADATA_TOKEN_TTL_SECONDS: &str = "X-metadata-token-ttl-seconds";
```
serban300, line 11: 1. Can we define these consts in the `TokenHeaders` implementation as suggested in the previous comment ?
2. Do we still need `TokenHeader` if we define these consts ?

luminitavoicu: Seems like I misunderstood your initial comment. So you are suggesting to use constants instead of enums for the headers variants, right?
I don't have a strong preference to keep the initial implementation and it might make the code easier to understand if ditch the `TokenHeader` struct. I made the changes.

serban300: Yes, that's what I meant. Looks good.

serban300 (dismissed):

src/mmds/src/token_headers.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// `X-metadata-token-ttl-seconds` header.
+    const X_METADATA_TOKEN_TTL_SECONDS: &'static str = "X-metadata-token-ttl-seconds";
+
+    /// Return `TokenHeaders` from headers map.
+    pub fn try_from(map: &HashMap<String, String>) -> Result<TokenHeaders, RequestError> {
+        let mut headers = Self::default();
+
+        if let Some(token) = map.get(TokenHeaders::X_METADATA_TOKEN) {
+            headers.x_metadata_token = Some(token.to_string());
+        }
```
alindima, line 41: slightly more rustacean:

```suggestion
        map.get(TokenHeaders::X_METADATA_TOKEN).map(|token| {
            headers.x_metadata_token = Some(token.to_string());
        });
```

src/mmds/src/token_headers.rs
```diff
@@ -0,0 +1,130 @@
...
+                    headers.x_metadata_token_ttl_seconds = Some(seconds);
+                }
+                Err(_) => {
+                    return Err(RequestError::HeaderError(HttpHeaderError::InvalidValue(
+                        TokenHeaders::X_METADATA_TOKEN_TTL_SECONDS.to_string(),
+                        value.to_string(),
+                    )));
+                }
+            }
+        }
```
alindima, line 55: slightly more rustacean, your call if you think it looks better:
```suggestion
        map.get(TokenHeaders::X_METADATA_TOKEN_TTL_SECONDS)
            .and_then(|value| match value.parse::<u32>() {
                Ok(seconds) => Some(Ok(headers.x_metadata_token_ttl_seconds = Some(seconds))),
                Err(_) => Some(Err(RequestError::HeaderError(
                    HttpHeaderError::InvalidValue(
                        TokenHeaders::X_METADATA_TOKEN_TTL_SECONDS.to_string(),
                        value.to_string(),
                    ),
                ))),
            })
            .unwrap()?;
```

src/mmds/src/lib.rs
```diff
@@ -90,15 +91,40 @@ fn convert_to_response(request: Request) -> Response {
         );
     }
 
-    if request.method() != Method::Get {
-        let mut response = build_response(
+    respond_to_request(request)
+}
+
+fn respond_to_request(request: Request) -> Response {
```
alindima, line 97: shouldn't we have integration tests for querying the API?
or are they coming in a future PR, once v2 is added?

also unit tests for convert_to_response given an MMDS Get version request

alindima (commented):
Looks good

src/mmds/src/token_headers.rs
luminitavoicu: I agree that your proposal it's more rustacean, but I think it's harder to read. My personal preference would be to leave this as is.

src/mmds/src/token_headers.rs
luminitavoicu: Given https://github.com/firecracker-microvm/firecracker/pull/2790#discussion_r753137956, I would prefer to keep the `if` construction for the `X_METADATA_TOKEN` header as well, for consistency reasons. WDYT?

src/mmds/src/lib.rs
luminitavoicu: Integration tests for querying the API will come in a future PR, once MMDSv2 integration is complete. As for unit tests for `convert_to_response`, these, too, will be part of a later PR that also includes logic for responding to MMDSv2 requests. Does that sound alright to you?

alindima: okay, that sounds alright

src/mmds/src/token_headers.rs
alindima: personally I prefer the functional approach, but it's only a nit. looks good either way 👍🏻

alindima (approved):

serban300 (approved):

PR #2791, opened 2021-11-18
Add token generation logic for MMDS version 2

luminitavoicu:

# Reason for This PR

Need a logic to generate session tokens for requests to MMDS version 2.

## Description of Changes

Added token generation and validation logic.

### Token manipulation:
The core information around which the session token is generated is the token time to live value in seconds, provided by the user. Note that this value cannot be greater than 21600 (6 hours), similar to EC2's IMDSv2. The token is a base64 encoded object which essentially includes four values:

* Token version - currently hard coded as 1
* IV (Initialization Vector or nonce) - a random 12-byte/96-bit value 
* Payload length - length of encrypted expiry time (see below)
* Expire time - encrypted using AES-GCM block cipher. The expire time is computed at the time the PUT request is received by adding the ttl value to the current time in seconds.

The Token struct encapsulates all information required for token generation:

```
/// Structure for the session Token used by MMDSv2.
#[repr(packed)]
struct Token {
    iv: [u8; 12], // random bytes 
    payload: [u8; 8], // encrypted expire time
    tag: [u8; 16], // tag returned after encryption
}
```

### Token generation:
* populate IV with 12 random bytes
* compute expire time
* encrypt expire time using  AES-GCM block cipher and save this inside the payload field
* save tag generated post encryption inside the Token struct
* token is obtained by encoding Token struct into base64 

### Token validation flow

* decode base64 token string into a buffer
* decrypt expire time from Token struct using block cipher decryption algorithm
* compare expiry against current time; if expiry is in the past, token is not valid. 
If any of the above steps fail (i.e. decoding the token or decrypting the expiry time), token is considered invalid.

## License Acceptance

## PR Checklist

- ~[] The issue which led to this PR has a clear conclusion.~
- ~[ ] Any user-facing changes are mentioned in `CHANGELOG.md`~.


src/mmds/src/token.rs
```diff
@@ -0,0 +1,547 @@
...
+        // Create Nonce object from initialization vector.
+        let nonce = Nonce::from_slice(iv);
+        // Decrypt expiry as vector of bytes from ciphertext.
+        let expiry_as_bytes = self
+            .cipher
+            .decrypt(nonce, ciphertext)
+            .map_err(|_| Error::ExpiryExtraction)?;
+        // Get expiry value in seconds from bytes.
+        let expiry: u64 =
+            bincode::deserialize(&expiry_as_bytes).map_err(|_| Error::ExpiryExtraction)?;
```
alindima, line 214: bincode has support for limiting the memory usage used for the deserialization. This is what we use in seccompiler for example, since the input comes from untrusted sources and deserialization is known to be a target for resource-exhaust attempts.
Example: https://github.com/firecracker-microvm/firecracker/blob/9718aa6a924d151abf8b09b2226fc76c31b51c49/src/seccompiler/src/lib.rs#L90

alindima (commented):
not a full review, but had a comment

src/mmds/src/token.rs
```diff
@@ -0,0 +1,547 @@
...
+
+    /// Encode token structure into a string using base64 encoding.
+    fn base64_encode_token_struct(token: &Token) -> Result<String, Error> {
+        let token_bytes: Vec<u8> = bincode::serialize(token).map_err(|_| Error::TokenGeneration)?;
+        let mut encoded_token = Vec::new();
+        // Ensure vector has enough space for base64 encoding.
+        encoded_token.resize(get_size_of_base64_encoding(token_bytes.len()), 0);
+        // Encode token structure bytes into base64.
+        let size = base64::encode_config_slice(token_bytes, base64::STANDARD, &mut encoded_token);
+        encoded_token.resize(size, 0);
```
serban300, line 160: Is there any advantage in performing these resizing logic ? Instead of simply calling `base64::encode_config()` ?

src/mmds/src/token.rs
```diff
@@ -0,0 +1,547 @@
...
+        let tag_bytes: [u8; TAG_LEN] = tag
+            .as_slice()
+            .try_into()
+            .map_err(|_| Error::TokenGeneration)?;
+
+        Ok((payload, tag_bytes))
+    }
+
+    /// Encode token structure into a string using base64 encoding.
+    fn base64_encode_token_struct(token: &Token) -> Result<String, Error> {
```
serban300, line 153: I would move this method inside the `Token` struct if possible. Same for base64_decode .

src/mmds/src/token.rs
luminitavoicu: That is a great call-out!

I added a size check to the beginning of the token string validation, to ensure we don't bother decoding the tokens that are not the expected size. Since we expect a string of a certain length, I think it makes more sense to ensure the input in this manner. WDYT?

src/mmds/src/token.rs
luminitavoicu: I updated the code for both the encoding and decoding functions.

luminitavoicu:
@serban300, @alindima , I addressed your comments. Please take another look!

src/mmds/src/token.rs
```diff
@@ -0,0 +1,547 @@
...
+    /// Generate encoded token string using the token time to live provided.
+    pub fn generate_token_secret(&mut self, ttl_seconds: u32) -> Result<String, Error> {
+        // Check number of tokens encrypted under the current key. We need to
+        // make sure no more than 2^32 tokens are encrypted with the same key.
+        // If this number is reached, we need to reinitialize the cipher entity.
+        self.check_encryption_count()?;
+        // Create token structure containing the encrypted expiry value.
+        let token_struct = self.create_token_struct(ttl_seconds)?;
+        // Encode struct into base64 in order to obtain token string.
+        let token = TokenAuthority::base64_encode_token_struct(&token_struct)?;
```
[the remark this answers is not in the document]
serban300: Can you also change the name of all the methods accordingly please ?

src/mmds/src/token.rs
alindima: That's a good idea, but this bincode memory limit is actually defending against potential cases in the implementation where the memory taken up for the deserialisation is not proportional to the length of the bytes array, so I suggest we use it to guard against this.

It also applies to the other bincode deser use-case (for the token struct).

In this specific case of expiry_as_bytes, we may actually remove the bincode usage by splitting the u64 into an array of 4 u8s based using bitwise shifts, if that makes sense

src/mmds/src/token.rs
```diff
@@ -0,0 +1,583 @@
...
+    /// Attempts to decrypt expiry value within token sequence. Returns false if expiry
+    /// cannot be decrypted. If decryption succeeds, returns true if token has not expired
+    /// (i.e. current time is greater than expiry) and false otherwise.
+    pub fn is_valid(&self, token: &str) -> bool {
+        let mut ciphertext = Vec::new();
+
+        // Check size of encoded token struct.
+        if token.len() != Token::size_of_base64_encoding() {
+            return false;
+        }
```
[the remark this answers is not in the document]
luminitavoicu: I relaxed the check: it's no longer an equality condition, but a bound check. LMKWYT.

src/mmds/src/token.rs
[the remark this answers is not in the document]
luminitavoicu: Done

src/mmds/src/token.rs
luminitavoicu: I removed serialization/deserialization for expiry value. We now only use (de)serialize for Token struct operations. 

I also included a limit the memory usage used for the deserialization operation. PTAL.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,574 @@
...
+pub const PATH_TO_TOKEN: &str = "/latest/api/token";
+/// Randomness pool file path.
+const RANDOMNESS_POOL: &str = "/dev/urandom";
+
+/// Token length limit to ensure we don't bother decrypting huge character
+/// sequences. Tokens larger than this are automatically rejected.
+const TOKEN_LENGTH_LIMIT: usize = 100;
+/// Byte limit passed to `bincode` to guard against allocating
+/// too much memory when deserializing tokens.
+const DESERIALIZATION_BYTES_LIMIT: u64 = 50;
```
serban300, line 45: I would add some details about how these values were chosen.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,583 @@
...
+
+        expiry_duration.as_secs()
+    }
+}
+
+/// Structure for token information.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+struct Token {
+    version: u16,
+    payload_size: u16,
```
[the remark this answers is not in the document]
luminitavoicu: This was initially added for validation purposes, to make sure the token was generated using the token authority. However, since the `Token` struct also contains a `version` constant field, we can ditch the additional `payload_size` field. I removed it.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,583 @@
...
+    /// Validate constant fields inside the token structure.
+    fn sanity_check(&self) -> bool {
+        self.version == 1 && self.payload_size == PAYLOAD_LEN as u16
+    }
+
+    /// Return size of base64 encoding of the token struct because encodings
+    /// require more space than the number of bytes to encode.
+    fn size_of_base64_encoding() -> usize {
+        std::mem::size_of::<Self>() / 3 * 4 + 4
+    }
```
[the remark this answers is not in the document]
luminitavoicu: This is no longer part of the code. Resolving this.

src/mmds/src/token.rs
luminitavoicu: I added some additional comments to explain the logic behind these constants. LMKWYT.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,562 @@
...
+
+    /// Validate constant fields inside the token structure.
+    fn sanity_check(&self) -> bool {
+        self.version == 1
+    }
+}
+
+fn read_into_buf_from_entropy_pool(buf: &mut [u8]) -> Result<(), io::Error> {
+    let mut random_file = File::open(Path::new(RANDOMNESS_POOL))?;
```
serban300, line 326: Would it make sense to define this `fd` as member of `TokenAuthority` instead of opening and closing it each time we read ?

src/mmds/src/token.rs
serban300:
```
The value is computed
/// based on the size of Token struct, including a small deviation.
```
Can you specify the `size_of(Token)` and maybe provide an approximate formula ?

src/mmds/src/token.rs
```diff
@@ -0,0 +1,562 @@
...
+    }
+
+    /// Make sure to reinitialize the cipher under a new key before reaching
+    /// a count of 2^32 encrypted tokens under the same cipher entity.
+    fn check_encryption_count(&mut self) -> Result<(), Error> {
+        // Make sure no more than 2^32 - 1 tokens are encrypted under
+        // the same encryption key.
+        if self.num_encrypted_tokens == u32::MAX {
+            // Reinitialize the cipher entity under a new key.
+            self.cipher = TokenAuthority::create_cipher()?;
```
serban300, line 233: I would specify somewhere the fact that by design we don't retain the old cipher because this case shouldn't happen and if it happens we expect the customer code to implement a retry when the token is expired.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,562 @@
...
+        Token {
+            version: 1,
+            iv,
+            payload,
+            tag,
+        }
+    }
+
+    /// Encode token structure into a string using base64 encoding.
+    fn base64_encode_token(&self) -> Result<String, Error> {
```
alindima, line 307: I'd rename these functions to remove the `_token` part, as it's embedded in the struct name already

src/mmds/src/token.rs
luminitavoicu: Done.

src/mmds/src/token.rs
luminitavoicu: Good idea!

src/mmds/src/token.rs
```diff
@@ -0,0 +1,563 @@
...
+        }
+
+        // Decrypt ttl using AES-GCM block cipher.
+        let expiry = match self.decrypt_expiry(&mut token.payload, &token.tag, &token.iv) {
+            Ok(expiry) => expiry,
+            Err(_) => return false,
+        };
+
+        // Get current time.
+        let time_now = SystemTime::now()
```
alindima, line 193: This uses the realtime clock of the system, a clocksource that be reconfigured by other users in the system and that may be affected by drifts (both forward and backward).

Since we are interested in the absolute difference here, I suggest we use a monotonic clock. This is a clock type that is initialised when the system is booted and is guaranteed to only go forwards and never be adjusted by external actors.

There are helpers in `utils/src/time.rs` for working with clocks.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,563 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
alindima, line 1: I'd like to have the implementation details documented in: https://github.com/firecracker-microvm/firecracker/blob/main/docs/mmds/mmds-design.md.

Will this come in a future PR?

alindima (commented):
Nice job!

src/mmds/src/token.rs
luminitavoicu: Yes, documentation updates regarding the implementation, as well as usage and examples will come in a future PR.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,562 @@
...
+/// Length of encryption tag.
+pub const TAG_LEN: usize = 16;
+
+/// Minimum lifetime of token.
+pub const MIN_TOKEN_TTL_SECONDS: u32 = 1;
+/// Maximum lifetime of token.
+pub const MAX_TOKEN_TTL_SECONDS: u32 = 21600;
+
+/// Path to token.
+pub const PATH_TO_TOKEN: &str = "/latest/api/token";
```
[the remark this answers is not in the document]
luminitavoicu: It is unused for now, but will be used in the MMDS internal logic, as the path for accepted `PUT` requests. After we merge this PR, the one that follows tackles this.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,562 @@
...
+            .with_fixint_encoding()
+            .allow_trailing_bytes()
+            .with_limit(DESERIALIZATION_BYTES_LIMIT)
+            .deserialize(&token_bytes)
+            .map_err(|_| Error::ExpiryExtraction)?;
+        Ok(token)
+    }
+
+    /// Validate constant fields inside the token structure.
+    fn sanity_check(&self) -> bool {
```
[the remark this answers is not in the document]
luminitavoicu: The sanity check is not used on the token creation path, but on the validation path. Since the token is created by MMDS itself, and we explicitly set the `version` field as 1, there is no need to verify this during token creation. 

The check is, however, required on the token validation path: when responding to the `GET` request, after decoding the `Token` struct from the base64 string, the `version` field is verified. 

The `sanity_check` function is not public and is only used by the validation function: `is_valid()`.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,563 @@
...
+        let expiry_duration = now_as_duration.add(ttl_duration);
+
+        expiry_duration.as_secs()
+    }
+}
+
+/// Structure for token information.
+#[derive(Debug, Deserialize, PartialEq, Serialize)]
+struct Token {
+    version: u16,
```
alindima, line 286: Why do we need the token version though?
Do we envision adding new fields to this token struct and supporting older versions of the struct in newer Firecracker versions?

src/mmds/src/token.rs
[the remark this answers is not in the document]
alindima: I see. It can be moved to `base64_decode_token` then. But it's fine keeping it here as well

src/mmds/src/token.rs
```diff
@@ -0,0 +1,562 @@
...
+#[derive(Debug)]
+pub enum Error {
+    /// Failed to extract entropy from pool.
+    EntropyPool(io::Error),
+    /// Failed to extract expiry from token sequence.
+    ExpiryExtraction,
+    /// Time to live value for token is invalid.
+    InvalidTtlValue(u32),
+    /// Failed to generate token.
+    TokenGeneration,
```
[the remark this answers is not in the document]
luminitavoicu: I added more explicit error variants. LMKWYT.

serban300 (dismissed):

src/mmds/src/token.rs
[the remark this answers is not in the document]
alindima: Looks good!

src/mmds/src/token.rs
alindima: As we've chatted offline, this would only be useful only for loading snapshots older guests in a newer FC versions (which have a new version of the MMDS token). But this wouldn't work since the `TokenAuthority` will be reinitialised and all tokens invalidated.
The only other possible use case I can think of would be for live migration (but we should tackle this when we have some clear plans for it)

src/mmds/src/token.rs
alindima: Looks good 👍🏻

src/mmds/src/token.rs
luminitavoicu: I removed the `version` field inside the Token struct, as we discussed.

src/mmds/src/token.rs
```diff
@@ -0,0 +1,562 @@
...
+    /// a count of 2^32 encrypted tokens under the same cipher entity.
+    fn check_encryption_count(&mut self) -> Result<(), Error> {
+        // Make sure no more than 2^32 - 1 tokens are encrypted under
+        // the same encryption key.
+        if self.num_encrypted_tokens == u32::MAX {
+            // Reinitialize the cipher entity under a new key.
+            self.cipher = TokenAuthority::create_cipher()?;
+            // Reset encrypted tokens count.
+            self.num_encrypted_tokens = 0;
+            warn!(
```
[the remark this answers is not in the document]
luminitavoicu: I changed the behavior of MMDS when the maximum number of encrypted tokens (2^32) is reached. The cipher is reseeded, as before, but MMDS now sends a response body specifying that the token authority has been reinitialized and all tokens have been invalidated, so the guest should retry the request. This change is required in order to have a mechanism to warn the user that the limit of generated tokens has been reached. If the user retries the PUT request, they will receive a valid token, encrypted under a new key.

alindima (approved):

src/mmds/src/token.rs
```diff
@@ -0,0 +1,541 @@
...
+        })
+    }
+
+    /// Generate encoded token string using the token time to live provided.
+    pub fn generate_token_secret(&mut self, ttl_seconds: u32) -> Result<String, Error> {
+        // Check number of tokens encrypted under the current key. We need to
+        // make sure no more than 2^32 tokens are encrypted with the same key.
+        // If this number is reached, we need to reinitialize the cipher entity
+        // and inform the guest that all previously valid tokens have been invalidated.
+        self.check_encryption_count()?;
```
serban300, line 122: I think the previous approach (returning a token generated with the new cipher) was better. The new approach introduces a new corner case for the customer to handle.

alindima: Previously the guest had no idea that the tokens were invalidated.
I agree that this is a new corner case to handle.
Would it be better to add a custom response header that indicates this instead? Or a new field in the body of the response that notifies the guest.

luminitavoicu: I wouldn't add a new field in the body, as it might break guest integration. I think the response body should only contain the token if the request is successful.

serban300: The guest as a whole won't find out that the cipher was regenerated anyway. Only the application that issues the 2^32rd request to MMDS. I don't think this makes a big difference.

alindima: > I wouldn't add a new field in the body, as it might break guest integration. I think the response body should only contain the token if the request is successful.


Well there is no customer integration yet. Adding a JSON response would be more extensible for the future.

I agree however that this not make such a big difference, granted that we document this somewhere. Otherwise guests will not know why their tokens have been invalidated.

luminitavoicu: Every detail regarding the behavior of MMDS will be mentioned in the documentation.

alindima (dismissed):

alindima (approved):

serban300 (approved):

PR #2792, opened 2021-11-19
[block][io_uring] Complete pending requests on snapshotting 

serban300:
# Reason for This PR

[block][io_uring] Complete pending requests on snapshotting 

## Description of Changes

[block][io_uring] Complete pending requests on snapshotting 

## License Acceptance

## PR Checklist

### DIFF
--- a/src/devices/src/virtio/block/io/async_io.rs
+++ b/src/devices/src/virtio/block/io/async_io.rs
@@ -89,6 +89,7 @@ impl<T> AsyncFileEngine<T> {
         })
     }
 
+    #[cfg(test)]
     pub fn file(&self) -> &File {
         &self.file
     }
--- a/src/devices/src/virtio/block/io/async_io.rs
+++ b/src/devices/src/virtio/block/io/async_io.rs
@@ -180,27 +180,29 @@ impl<T> AsyncFileEngine<T> {
         self.ring.submit().map(|_| ()).map_err(Error::IoUring)
     }
 
-    pub fn drain(&mut self, flush: bool) -> Result<(), Error> {
-        self.drain_submission_queue()?;
-
-        // Drain the completion queue so that we may deallocate the user_data fields.
-        while self.do_pop()?.is_some() {}
-
-        if flush {
-            // Sync data out to physical media on host.
-            // We don't need to call flush first since all the ops are performed through io_uring
-            // and Rust shouldn't manage any data in its internal buffers.
-            self.file.sync_all().map_err(Error::SyncAll)
-        } else {
-            Ok(())
-        }
-    }
-
-    pub fn drain_submission_queue(&mut self) -> Result<(), Error> {
+    pub fn drain(&mut self, discard_cqes: bool) -> Result<(), Error> {
         self.ring
             .submit_and_wait_all()
             .map(|_| ())
-            .map_err(Error::IoUring)
+            .map_err(Error::IoUring)?;
+
+        if discard_cqes {
+            // Drain the completion queue so that we may deallocate the user_data fields.
+            while self.do_pop()?.is_some() {}
+        }
+
+        Ok(())
+    }
+
+    pub fn drain_and_flush(&mut self, discard_cqes: bool) -> Result<(), Error> {
+        self.drain(discard_cqes)?;
+
+        // Sync data out to physical media on host.
+        // We don't need to call flush first since all the ops are performed through io_uring
+        // and Rust shouldn't manage any data in its internal buffers.
+        self.file.sync_all().map_err(Error::SyncAll)?;
+
+        Ok(())
     }
 
     fn do_pop(&mut self) -> Result<Option<Cqe<WrappedUserData<T>>>, Error> {
--- a/src/vmm/src/device_manager/persist.rs
+++ b/src/vmm/src/device_manager/persist.rs
@@ -203,14 +203,11 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                     });
                 }
                 TYPE_BLOCK => {
-                    let block_state = locked_device
-                        .as_any()
-                        .downcast_ref::<Block>()
-                        .unwrap()
-                        .save();
+                    let block = locked_device.as_mut_any().downcast_mut::<Block>().unwrap();
+                    block.prepare_save();
                     states.block_devices.push(ConnectedBlockState {
                         device_id: devid.clone(),
-                        device_state: block_state,
+                        device_state: block.save(),
                         transport_state,
                         mmio_slot: devinfo.clone(),
                     });
--- a/tests/framework/utils_vsock.py
+++ b/tests/framework/utils_vsock.py
@@ -8,6 +8,7 @@ from select import select
 from socket import socket, AF_UNIX, SOCK_STREAM
 from threading import Thread, Event
 import re
+from framework import utils
 
 from host_tools.network import SSHConnection
 
@@ -30,6 +31,7 @@ class HostEchoServer(Thread):
         """."""
         super().__init__()
         self.vm = vm
+        self.path = path
         self.sock = socket(AF_UNIX, SOCK_STREAM)
         self.sock.bind(path)
         self.sock.listen(SERVER_ACCEPT_BACKLOG)
@@ -83,6 +85,8 @@ class HostEchoServer(Thread):
         This method can be called from any thread. Upon returning, the
         echo server will have shut down.
         """
+        self.sock.close()
+        utils.run_cmd("rm -f {}".format(self.path))
         self.exit_evt.set()
         self.join()
 
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -599,13 +599,18 @@ impl VirtioDevice for Block {
 
 impl Drop for Block {
     fn drop(&mut self) {
-        let flush = match self.disk.cache_type {
-            CacheType::Unsafe => false,
-            CacheType::Writeback => true,
+        match self.disk.cache_type {
+            CacheType::Unsafe => {
+                if let Err(e) = self.disk.file_engine_mut().drain(true) {
+                    error!("Failed to drain ops on drop: {:?}", e);
+                }
+            }
+            CacheType::Writeback => {
+                if let Err(e) = self.disk.file_engine_mut().drain_and_flush(true) {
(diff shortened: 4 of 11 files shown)


src/devices/src/virtio/block/io/async_io.rs
```diff
@@ -180,27 +180,31 @@ impl<T> AsyncFileEngine<T> {
...
+            .map_err(Error::IoUring)?;
 
         if flush {
             // Sync data out to physical media on host.
             // We don't need to call flush first since all the ops are performed through io_uring
             // and Rust shouldn't manage any data in its internal buffers.
-            self.file.sync_all().map_err(Error::SyncAll)
-        } else {
-            Ok(())
+            self.file.sync_all().map_err(Error::SyncAll)?;
```
alindima, line 205: I really think we should take this flush option out of the `drain_submission_queue` function to get better separation of concerns

src/devices/src/virtio/block/io/async_io.rs
```diff
@@ -180,27 +180,31 @@ impl<T> AsyncFileEngine<T> {
         self.ring.submit().map(|_| ()).map_err(Error::IoUring)
     }
 
-    pub fn drain(&mut self, flush: bool) -> Result<(), Error> {
-        self.drain_submission_queue()?;
+    pub fn drain(&mut self, flush: bool, discard: bool) -> Result<(), Error> {
```
alindima, line 184: I'd rename this param to `discard_cqes`

src/vmm/src/device_manager/persist.rs
```diff
@@ -203,14 +203,11 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                     });
                 }
                 TYPE_BLOCK => {
-                    let block_state = locked_device
-                        .as_any()
-                        .downcast_ref::<Block>()
-                        .unwrap()
-                        .save();
+                    let block = locked_device.as_mut_any().downcast_mut::<Block>().unwrap();
+                    block.prepare_save();
```
alindima, line 207: we could move this logic to the `save()` impl of the Block device, so that we get better encapsulation (and make prepare_save() private).
This way, users of the block device would not forget to call `prepare_save` before `save`

alindima (commented):
Looks good! Have you done performance measurements to see how long would it take to drain the submission and completion rings?
AFAICT the baselines in test_snapshot_perf.py are the same, but could it be that we are not using disk-intensive workloads there?

Would be nice to see a worst-case scenario where the submission ring is pretty much full at the time of snapshot (but it might be difficult/impossible to synchronize).

src/vmm/src/device_manager/persist.rs
serban300: The problem is that save takes an immutable reference to self and we need a mutable one. That's why I called `prepare_save` here. I don't think it's worth trying to modify the signature of `save`.

serban300:
> Looks good! Have you done performance measurements to see how long would it take to drain the submission and completion rings? AFAICT the baselines in test_snapshot_perf.py are the same, but could it be that we are not using disk-intensive workloads there?
> 
> Would be nice to see a worst-case scenario where the submission ring is pretty much full at the time of snapshot (but it might be difficult/impossible to synchronize).

Yes, our snapshotting tests don't use disk-intensive tasks. Also I didn't have time to test this, but the added latency shouldn't be significant compared to saving a large amount of memory. I can try to save a snapshot while running fio and do some measurements.

src/devices/src/virtio/block/io/async_io.rs
serban300: If we separate flush, it will not work in any conditions. A separate flush will work correctly only after a drain. So I would keep it as it is in order to avoid the possibility of using it incorrectly.

src/vmm/src/device_manager/persist.rs
alindima: oh I see. In this case it's not worth it indeed

src/devices/src/virtio/block/io/async_io.rs
alindima: It's just that there is nothing in the naming of the function that suggests it also flushes the file. That's why I find it confusing.
I am however not super keen on this, it's only a minor thing

alindima:
> I can try to save a snapshot while running fio and do some measurements.

Yes, that would be interesting to see. I agree that it should be insignificant compared to saving a large amount of memory, but what about when having a diff snapshot or a small memory size. However, correctness in this case is more important and I think that snapshot creation is not usually in the hot-path.

So I won't block this PR, but let me know the results once you have some

alindima (approved):

alindima:
I see that there are some snapshot tests failing, could they be related to these changes?

luminitavoicu (commented):
Looks good to me! Could you please take a look at the failing test?

src/devices/src/virtio/block/io/async_io.rs
acatangiu: You could have a `flush` function that does drain+flush, and this `drain()` fn only does drain.

then they can be called depending on needs and none could be incorrectly used.

Just sayin :smiley:

serban300:
> > I can try to save a snapshot while running fio and do some measurements.
> 
> Yes, that would be interesting to see. I agree that it should be insignificant compared to saving a large amount of memory, but what about when having a diff snapshot or a small memory size. However, correctness in this case is more important and I think that snapshot creation is not usually in the hot-path.
> 
> So I won't block this PR, but let me know the results once you have some

> I see that there are some snapshot tests failing, could they be related to these changes?

I did some measurements, both while running the `test_5_full_snapshots` and also manually while executing a fio test on a microvm. I never got more than 20ms. For example this output is from a run of `test_5_full_snapshots`:

```
2021-11-25T10:56:06.082896732 [5277d156-3c57-49b7-8780-8314be5e5f97:main] prepare_save() time: 17497
2021-11-25T10:56:07.235134051 [5277d156-3c57-49b7-8780-8314be5e5f97:main] 'create full snapshot' VMM action took 1172019 us.
```

As you can see, `prepare_save()` takes 17ms while the entire snapshotting process (probably dominated by the memory saving operation) takes 1172 - 17 ms which is huge. This is on my local development machine.

So the `test_5_full_snapshots` failures shouldn't be related to `prepare_save`.

src/devices/src/virtio/block/io/async_io.rs
serban300: Done. I defined 2 methods: `drain()` and `drain_and_flush()`.

alindima:
> I did some measurements, both while running the `test_5_full_snapshots` and also manually while executing a fio test on a microvm. I never got more than 20ms. For example this output is from a run of `test_5_full_snapshots`:
> 
> ```
> 2021-11-25T10:56:06.082896732 [5277d156-3c57-49b7-8780-8314be5e5f97:main] prepare_save() time: 17497
> 2021-11-25T10:56:07.235134051 [5277d156-3c57-49b7-8780-8314be5e5f97:main] 'create full snapshot' VMM action took 1172019 us.
> ```
> 
> As you can see, `prepare_save()` takes 17ms while the entire snapshotting process (probably dominated by the memory saving operation) takes 1172 ms which is huge. This is on my local development machine.
> 
> So the `test_5_full_snapshots` failures shouldn't be related to `prepare_save`.

Looking good, thanks for the update. While the timing looks good, the CI failures don't seem related to the performance of saving a snapshot. But it does seem very coincidental that they have failed a number of times on the snapshotting PR. could it be that there's a bug somewhere? Are the tests also failing on local runs?

src/devices/src/virtio/block/io/async_io.rs
alindima: Hm I think you forgot to rename this

acatangiu (approved):

serban300:
> Looking good, thanks for the update. While the timing looks good, the CI failures don't seem related to the performance of saving a snapshot. But it does seem very coincidental that they have failed a number of times on the snapshotting PR. could it be that there's a bug somewhere? Are the tests also failing on local runs?

Could be a bug. I'm investigating the issue. I can't reproduce the failure on any other machine for the moment. It only fails in the CI.

serban300:
I fixed the failing integration tests. I noticed that the problem was occurring after the OOM killer was running. My theory is that before introducing the current changes the fio processes didn't have time to process enough data before snapshotting, not causing any issues. Now, because we're waiting for all the pending ops to complete, the fio processes advance more, leading eventually to the OOM killer run.

So I increased the memory of the VM from 256MB to 512MB. Judging by the comments this was the desired mem size anyway.

Also to be fair, after running the tests so many times, with many fio processes I see that sometimes `prepare_save()` can take up to 2-3 seconds. 

@alindima @luminitavoicu PTAL !

src/devices/src/virtio/block/io/async_io.rs
serban300: Should be renamed: https://github.com/firecracker-microvm/firecracker/blob/81930e4a5599956a1c45a4c4f1e429a7831d6ae3/src/devices/src/virtio/block/io/async_io.rs#L184

tests/framework/utils_vsock.py
```diff
@@ -83,6 +85,8 @@ def exit(self):
         This method can be called from any thread. Upon returning, the
         echo server will have shut down.
         """
+        self.sock.close()
+        utils.run_cmd("rm -f {}".format(self.path))
```
alindima, line 89: nit: I think you could use `sock.getsockname()` to avoid storing an extra variable

alindima (approved):

alindima:
> I fixed the failing integration tests. I noticed that the problem was occurring after the OOM killer was running. My theory is that before introducing the current changes the fio processes didn't have time to process enough data before snapshotting, not causing any issues. Now, because we're waiting for all the pending ops to complete, the fio processes advance more, leading eventually to the OOM killer run.
> So I increased the memory of the VM from 256MB to 512MB. Judging by the comments this was the desired mem size anyway.

Interesting find!

> Also to be fair, after running the tests so many times, with many fio processes I see that sometimes prepare_save() can take up to 2-3 seconds.

As we discussed offline, the 2-3 second overhead is out of a total of ~7 seconds, so most of the snapshot overhead is not caused by io_uring. I assume it's caused by the heavy fio workload, since the VMM thread must be busy emulating block requests before getting the snapshot event from the API thread.

I think this is perfectly acceptable 👍🏻

luminitavoicu (approved):

PR #2804, opened 2021-12-07
Integrate MMDS token generation 

luminitavoicu:
# Reason for This PR

Integrate MMDS token generation logic.

## Description of Changes

- Added `TokenAuthority` entity inside `MMDS` and API to generate and validate session tokens.
- Mknod random generator device in jailer
- integration tests 

## License Acceptance

## PR Checklist


src/api_server/src/lib.rs
```diff
@@ -524,7 +525,7 @@ mod tests {
         let to_vmm_fd = EventFd::new(libc::EFD_NONBLOCK).unwrap();
         let (api_request_sender, _from_api) = channel();
         let (_to_api, vmm_response_receiver) = channel();
-        let mmds_info = Arc::new(Mutex::new(Mmds::default()));
+        let mmds_info = Arc::new(Mutex::new(Mmds::new().unwrap()));
```
AlexandruCihodaru, line 528: Add a comment stating why the unwrap is safe

resources/seccomp/aarch64-unknown-linux-musl.json
```diff
@@ -459,6 +459,19 @@
                     }
                 ]
             },
+            {
+                "syscall": "fcntl",
+                "comment": "Used by MMDS",
```
AlexandruCihodaru, line 464: Maybe write a more verbose comment here, explain why MMDS needs it

src/mmds/src/lib.rs
```diff
@@ -169,6 +200,46 @@ fn respond_to_get_request(request: Request) -> Response {
 mod tests {
     use super::*;
 
+    fn populate_mmds() -> Arc<Mutex<Mmds>> {
```
AlexandruCihodaru, line 271: nice :D

src/api_server/src/lib.rs
AlexandruCihodaru: At a second 👀  this is not safe, as `Mmds::new` will return an error if TokenAuthority fails

AlexandruCihodaru: This unwrap is in a test module I was talking about this one https://github.com/firecracker-microvm/firecracker/blob/32c12c82472189330d4f61d7b62c70feecd63b5f/src/mmds/src/lib.rs#L25. But I think expect is OK here so will resolve this

src/mmds/src/data_store.rs
```diff
@@ -27,20 +20,22 @@ pub enum OutputFormat {
...
     NotFound,
     NotInitialized,
+    TokenAuthority(TokenError),
     UnsupportedValueType,
 }
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
+        match &*self {
```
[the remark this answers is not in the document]
luminitavoicu: We can, this was unintentional. Solved this.

src/mmds/src/lib.rs
```diff
@@ -22,7 +22,7 @@ lazy_static! {
     // A static reference to a global Mmds instance. We currently use this for ease of access during
     // prototyping. We'll consider something like passing Arc<Mutex<Mmds>> references to the
     // appropriate threads in the future.
-    pub static ref MMDS: Arc<Mutex<Mmds>> = Arc::new(Mutex::new(Mmds::default()));
+    pub static ref MMDS: Arc<Mutex<Mmds>> = Arc::new(Mutex::new(Mmds::new().expect("Unable to create MMDS.")));
```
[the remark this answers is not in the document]
luminitavoicu: It makes sense to have the `MMDS` instance reference to be an `Option` in order to make sure customers that do not care about MMDS are not affected by MMDS not being successfully created. I logged a warning in case MMDS creation fails. Let me know what you think about this.

It is true that we do not have a way to start without MMDS at the moment - the instance is always created, even if there are no net devices configured to accept mmds requests.

src/mmds/src/lib.rs
```diff
@@ -96,58 +118,88 @@ fn convert_to_response(request: Request) -> Response {
...
+        false => {
+            let err_msg = "Mmds token not valid.";
+            build_response(
+                request.http_version(),
+                StatusCode::Unauthorized,
+                Body::new(err_msg),
+            )
+        }
+    }
+}
```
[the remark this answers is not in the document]
georgepisaltu: After taking another look, there is no easy way to merge accesses such that we only have one lock acquire and I don't believe this optimization is worth the effort. Resolving this comment.

tests/integration_tests/functional/test_mmds.py
```diff
@@ -89,7 +112,18 @@ def test_custom_ipv4(test_microvm_with_api, network_config):
...
+        ssh_connection,
+        ipv4_address="169.254.169.250",
+        token_ttl=60
+    )
+
+    pre = 'curl -m 2 -s'
+    pre += ' -X GET'
+    pre += ' -H  "X-metadata-token: {}"'.format(token)
+    pre += ' -H "Accept: application/json"'
+    pre += ' http://169.254.169.250/'
```
georgepisaltu, line 126: I see this repeated throughout this file. Can we have a function for this so we reuse the code?

georgepisaltu (commented):
PR is looking great!

src/firecracker/src/main.rs
```diff
@@ -311,8 +311,8 @@ fn main_exitable() -> ExitCode {
         .map(fs::read_to_string)
         .map(|x| x.expect("Unable to open or read from the mmds content file"));
 
-    if let Some(data) = metadata_json {
-        MMDS.lock()
+    if let (Some(data), Some(mmds)) = (metadata_json, MMDS.as_ref()) {
```
georgepisaltu, line 314: This change is good, but we should also disallow the usage of `--metadata` when we run with `--no-api`, since MMDS will be disabled.

georgepisaltu (commented):
PR is looking good! I didn't get a chance to take a look at the tests as well, I'll do that a bit later. In the meantime, I'll leave my feedback on the implementation.

src/mmds/src/lib.rs
```diff
@@ -11,18 +11,36 @@ use serde_json::{Map, Value};
...
+pub const NO_TTL_PROVIDED_ERROR: &str = "Token time to live value not found. Use \
+                                         `X-metadata-token-ttl_seconds` header to specify \
+                                         the token's lifetime.";
 
 lazy_static! {
     // A static reference to a global Mmds instance. We currently use this for ease of access during
     // prototyping. We'll consider something like passing Arc<Mutex<Mmds>> references to the
     // appropriate threads in the future.
-    pub static ref MMDS: Arc<Mutex<Mmds>> = Arc::new(Mutex::new(Mmds::default()));
+    pub static ref MMDS: Option<Arc<Mutex<Mmds>>> = Mmds::new().map_or_else(
```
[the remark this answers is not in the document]
luminitavoicu: I propose to remove the `Option` here. I was thinking that since we do support both MMDSv1 and MMDSv2 for now, customers that still prefer to use v1 should not be affected by MMDS default (v2) not being able to be initialized (for some reason or another). We could have MMDSv2 to be the default (as we currently do) and MMDSv1 the fallback mechanism in case MMDSv2 can't be initialized. We could log a warning if this happens. The user would be aware of the version configured because for MMDSv1 we do not allow `PUT` requests, so there would be no way to generate the token required for MMDSv2. WDYT?

src/mmds/src/lib.rs
```diff
@@ -174,14 +256,123 @@ fn respond_to_get_request(request: Request) -> Response {
...
+        Some(mmds) => mmds
+            .lock()
+            .expect("Poisoned lock")
+            .generate_token(ttl_seconds),
+        // This path is unreachable because MMDS entity has been previously
+        // checked.
+        None => unreachable!(),
+    };
+    match response {
+        Ok(token) => build_response(request.http_version(), StatusCode::OK, Body::new(token)),
```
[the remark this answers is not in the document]
luminitavoicu: Indeed there is no JSON logic in the code, the `PUT` request does return only a String. This is consistent to EC2's IMDSv2 behavior and I personally believe it's easier to operate with, since one could have something like:
```bash
TOKEN=curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-metadata-token-ttl-seconds: 21600"
curl -H "X-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/
```

src/mmds/src/lib.rs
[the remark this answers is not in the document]
georgepisaltu: The `Option` is the result of [this comment](https://github.com/firecracker-microvm/firecracker/pull/2804#discussion_r765807994) and I'd keep it. The question is what do we do with it.

I am actually against a fallback route in this case. It's cumbersome to support both versions in a client. Users which start the MMDS expect to get MMDSv2 (either they explicitly set the version or the default version kicked in) and a fallback mechanism to MMDSv1 is bad CX in my opinion, as the user wants the v2 features. If they didn't care about the v2 features, they would use v1 from the start. It also complicates the whole setup by adding a "negotiation" phase.

If we decide to not start with MMDS by default, in my opinion it makes sense to crash the process early if we can't create it. If we don't change the MMDS behavior now (i.e. we start Firecracker with MMDS enabled by default) we can keep the current approach in this PR. It is highly unlikely that we won't be able to create the MMDS anyway, so this behavior would only affect the CX in the unlikely case of an error.

src/mmds/src/lib.rs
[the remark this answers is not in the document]
georgepisaltu: In this case, we need to make sure we set the `Content-Type` header in the response to `text/plain`.

src/mmds/src/data_store.rs
```diff
@@ -112,21 +123,53 @@ impl Mmds {
...
+        match version {
+            MmdsVersion::V1 => {
+                self.token_authority = None;
+                Ok(())
+            }
+            MmdsVersion::V2 => {
+                if self.token_authority.is_none() {
+                    self.token_authority =
+                        Some(TokenAuthority::new().map_err(Error::TokenAuthority)?);
+                }
```
georgepisaltu, line 136: Can we do this with `Option::or`?

georgepisaltu (commented):
PR looking good!

resources/seccomp/aarch64-unknown-linux-musl.json
luminitavoicu: Update: since we now allow both MMDS versions, `fcntl` is required when setting V2 after Firecracker has started. I added comments to mention why MMDS needs to have this syscall on the allow list.

src/mmds/src/data_store.rs
luminitavoicu: We can’t use `Option::or` to replace here. `Token_authority` will be moved because it does not derive the `Copy` trait. This is due to the fact that `TokenAuthority` struct contains a `File` field, which also does not derive `Copy`. I propose to leave this as is.

src/mmds/src/data_store.rs
```diff
@@ -112,21 +123,53 @@ impl Mmds {
...
+        } else {
+            MmdsVersion::V2
+        }
+    }
+
+    /// Checks if the provided token has not expired.
+    pub fn is_valid_token(&self, token: &str) -> bool {
+        match &self.token_authority {
+            Some(ta) => ta.is_valid(token),
+            None => unreachable!(),
```
georgepisaltu, line 155: I missed this in the previous reviews. We should not have `unreachable!()` statements in the public interface. All public functions must offer panic-free functionality. I saw that the code logic will never actually call this without a token authority, but we should make the check at a higher level and offer an error here instead (like `InvalidState` or something like that). Another option would be to make this function private and create a safe public wrapper, but it feels clunky since it's not an internal function and it just calls `TokenAuthority::is_valid()`.

src/mmds/src/data_store.rs
```diff
@@ -112,21 +123,53 @@ impl Mmds {
...
+        }
+    }
+
+    /// Generate a new Mmds token using the token authority.
+    pub fn generate_token(&mut self, ttl_seconds: u32) -> Result<String, TokenError> {
+        match self.token_authority.as_mut() {
+            Some(ta) => ta.generate_token_secret(ttl_seconds),
+            None => unreachable!(),
```
georgepisaltu, line 163: Same as above regarding `unreachable!()` statements.

src/mmds/src/lib.rs
```diff
@@ -138,16 +195,70 @@ fn respond_to_get_request(request: Request) -> Response {
...
+                request.http_version(),
+                StatusCode::Unauthorized,
+                Body::new(error_msg),
+            );
+        }
+    };
+
+    // Validate MMDS token.
+    let is_valid = match MMDS.lock().expect("Poisoned lock").as_ref() {
+        Some(mmds) => mmds.is_valid_token(token),
```
georgepisaltu, line 231: We can have the check I mentioned in the other comment (regarding `unreachable!()`) here, especially since this is an internal function and we can better control the state machine of the MMDS object from here to actually ensure we never call `is_valid_token` in a context in which it would panic.

src/mmds/src/lib.rs
```diff
@@ -275,9 +287,60 @@ fn respond_to_get_request_unchecked(request: Request) -> Response {
     }
 }
 
+fn respond_to_put_request(request: Request, token_headers: TokenHeaders) -> Response {
+    let uri = request.uri().get_abs_path();
+    // Sanitize the URI into a strict json path.
+    let json_pointer = sanitize_uri(uri.to_string());
```
AlexandruCihodaru, line 313: can we rename it to json_path or smth like this?

src/api_server/src/lib.rs
```diff
@@ -317,37 +317,60 @@ impl ApiServer {
...
+        {
+            Some(mmds) => {
+                let version = mmds.version();
+                ParsedRequest::success_response_with_data(&MmdsVersionType::from(version))
+            }
+            None => ApiServer::json_response(
+                StatusCode::BadRequest,
+                ApiServer::json_fault_message(MmdsError::NoMmds.to_string()),
+            ),
+        }
```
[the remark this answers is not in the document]
georgepisaltu: While I agree the `mmds_unlock_and_do` call would look nice with a callback, I don't think the extra complexity added by a dynamic dispatched callback is worth the benefit here. If we really want to deduplicate code, I'd move towards a simpler, more modular solution where we replace just the MMDS unlocking part, which is actually the common denominator here. Moreover, the callback definition is not really extensible, because it has to take all the possible arguments for the "do" part, but in reality only one or none of them is used in a "do".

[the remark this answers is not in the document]
AlexandruCihodaru: Agree with you @georgepisaltu , although the part where we are returning the `NoMMDS` answer is also common, I think a macro with variable number of arguments (similar to println!) will work just right here. For now I am fine with returning just the unlocked part but we can try to simplify it afterwards.

luminitavoicu:
@AlexandruCihodaru and @georgepisaltu, I addressed your comments. Please take another look!

georgepisaltu (approved):

AlexandruCihodaru (approved):

PR #2805, opened 2021-12-07
Document Firecracker's defense in depth approach

### LINKED ISSUE #2785: [Feature Request] Please remove the implied "roadmap" to remove the ability ro run Firecracker without jailer
# Feature Request
Please remove the implied "roadmap" to remove the ability to run Firecracker without jailer.

In the design docs: https://github.com/firecracker-microvm/firecracker/blob/main/docs/design.md and possibly elsewhere there is a comment:
```
Firecracker should be started only via the jailer binary. The firecracker binary can also be executed directly, but this will no longer be possible in the future.
```
Although starting Firecracker from jailer might be a good approach in some circumstances, there are legitimate use cases where this might not be ideal or even be sub-optimal.

In my situation I am launching firecracker in a (Docker) container and by doing that there is a similar level of additional containment to that provided by jailer - e.g. it is running in isolated mount, pid, network, etc. namespaces, and the cgroups (cpu/memory etc.) that I specify when starting the container will resource limit the firecracker executable in a similar way to when using jailer.

Whilst it is possible to run jailer from the Docker container a) It is not clear that it is actually providing any meaningful additional security benefits in that situation b) in order to run jailer in a container one has to actively *reduce* the container's security by providing elevated privileges. To run jailer the container needs to be run with CAP_SYS_ADMIN, CAP_CHOWN, CAP_DAC_OVERRIDE, CAP_MKNOD, CAP_SETGID, CAP_SETUID, CAP_SYS_CHROOT and to run jailer as a non-root user in the container I need to run:
```
setcap cap_chown,cap_dac_override,cap_mknod,cap_setgid,cap_setuid,cap_sys_admin,cap_sys_chroot=eip jailer
```
Moreover I need a custom seccomp profile to add pivot_root to the seccomp allow list (CAP_SYS_ADMIN ungates unshare, but that capability is way more privileged than I'd like) also a custom apparmor profile is required.

I'm aware that jailer will drop these capabilities when it starts firecracker, but in general I'd prefer to not allow then in the first place, especially given the marginal, if any, additional benefits that jailer gives in addition to the container isolation.

Please consider removing the statement (and thus implied roadmap) that at some point it might not be possible to run Firecracker without jailer - the current position offers the flexibility to choose to use jailer or to take another approach and whilst it is totally sensible to provide guidance on how best to run Firecracker in production it is unhelpful to have an implicit roadmap to remove that option as I believe that running without jailer is a legitimate use case.  

If it became mandatory to use jailer it would be possible, but at the cost of launching a container with far more privileges than I'd like.

## Additional context
N/A

raduiliescu:
# Reason for This PR

Fixes  #2785

## Description of Changes

Clearly mention that secure isolation and defense in depth can be
guaranteed only if prod-host-setup recommendations are implemented.

Remove the mention that Firecracker won't start without
the jailer in the future.

## License Acceptance

## PR Checklist

### DIFF
--- a/docs/prod-host-setup.md
+++ b/docs/prod-host-setup.md
@@ -1,5 +1,10 @@
 # Production Host Setup Recommendations
 
+Firecracker relies on KVM and on the processor virtualization features
+for workload isolation. Security guarantees and defense in depth can only be
+upheld, if the following list of recommendations are implemented in
+production.
+
 ## Firecracker Configuration
 
 ### Seccomp
@@ -77,8 +82,12 @@ for Firecracker processes that are unresponsive, and kills them, by SIGKILL.
 
 ## Jailer Configuration
 
-Using Jailer in a production Firecracker deployment is highly recommended,
-as it provides additional security boundaries for the microVM.
+For assuring secure isolation in production deployments, Firecracker should
+must be started using the `jailer` binary that's part of each Firecracker
+release, or executed under process constraints equal or more restrictive than
+those in the jailer. For more about Firecracker sandboxing please see
+[Firecracker design](design.md)
+
 The Jailer process applies
 [cgroup](https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt),
 namespace isolation and drops privileges of the Firecracker process.
--- a/docs/design.md
+++ b/docs/design.md
@@ -49,10 +49,10 @@ images/firecracker_host_integration.png?raw=true
 Firecracker runs on Linux hosts with 4.14 or newer kernels and with Linux
 guest OSs (from this point on, referred to as guests). In production
 environments, Firecracker should be started only via the `jailer` binary.
-The `firecracker` binary can also be executed directly, but this will no longer
-be possible in the future. After launching the process, users interact with
-the Firecracker API to configure the microVM, before issuing the
-`InstanceStart` command.
+See [Sandboxing](#Sandboxing) for more details.
+
+After launching the process, users interact with the Firecracker API to
+configure the microVM, before issuing the `InstanceStart` command.
 
 ### Host Networking Integration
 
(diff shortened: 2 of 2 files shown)


docs/prod-host-setup.md
```diff
@@ -1,5 +1,10 @@
 # Production Host Setup Recommendations
 
+Firecracker relies on KVM and on the processor virtualization features
+for assuring workload isolation. Security guarantees and defense in deepth can
```
georgepisaltu, line 4:
```suggestion
for workload isolation. Security guarantees and defense in depth can
```

docs/prod-host-setup.md
```diff
@@ -1,5 +1,10 @@
...
+only be upheald, if the following list of recommendations are implemented in
```
georgepisaltu, line 5:
```suggestion
only be upheld if the following list of recommendations are implemented in
```

docs/prod-host-setup.md
```diff
@@ -77,8 +82,13 @@ for Firecracker processes that are unresponsive, and kills them, by SIGKILL.
 
 ## Jailer Configuration
 
-Using Jailer in a production Firecracker deployment is highly recommended,
-as it provides additional security boundaries for the microVM.
+For assuring secure isolation in production deployments, Firecracker should
+only be started only using the `jailer` binary that's part of each Firecracker
+release, or executed under process constraints equal or more restrictive than
+those in the included jailer. For more about Firecracker sandboxing please see
```
georgepisaltu, line 88:
```suggestion
those in the jailer. For more about Firecracker sandboxing please see
```

georgepisaltu (commented):
Looks good! Please fix the style tests as well.

alindima (approved):

docs/prod-host-setup.md
```diff
@@ -77,8 +82,12 @@ for Firecracker processes that are unresponsive, and kills them, by SIGKILL.
 
 ## Jailer Configuration
 
-Using Jailer in a production Firecracker deployment is highly recommended,
-as it provides additional security boundaries for the microVM.
+For assuring secure isolation in production deployments, Firecracker should
+only be started only using the `jailer` binary that's part of each Firecracker
```
dianpopa, line 86:
```suggestion
must be started using the `jailer` binary that's part of each Firecracker
```

raduiliescu: Thanks for catching this typo. Now is resolved in a new iteration.

dianpopa (approved):

alindima (approved):

Issue #2783, opened 2021-11-12
[Bug] GET `/vm/config` returns empty config data for VMs resumed from a snapshot


alindima:
# Describe the bug

A `GET` request on `/vm/config` should return the full VM configuration.
For a VM built from requests or a JSON, this works as expected.
However, for a VM created from a snapshot, this only includes an incorrect machine-config (the default one), with the rest of the fields left blank:

Sample output:

```
{'balloon': None, 'drives': [], 'boot-source': {'kernel_image_path': '', 'initrd_path': None}, 'logger': None, 'machine-config': {'vcpu_count': 1, 'mem_size_mib': 128, 'ht_enabled': False, 'track_dirty_pages': False}, 'metrics': None, 'mmds-config': None, 'network-interfaces': [], 'vsock': None}
```

## To Reproduce

1. Start a Firecracker VM.
2. Pause and create a snapshot.
3. Launch a new Firecracker process and load the snapshot.
4. Issue a GET request on `/vm/config` to get the full config.

## Expected behaviour

Should return the correct configuration

## Checks

- [x] Have you searched the Firecracker Issues database for similar problems?
- [x] Have you read the existing relevant Firecracker documentation?
- [x] Are you certain the bug being reported is a Firecracker issue?

summerlinryan:
Hey I'm pretty new to Rust, Firecracker, and OSS in general but I recently have been doing some OS development in Rust and though this was a great project for me to try to contribute to. I'd like to take a shot at solving this bug if that's alright with the maintainers. I'm able to reproduce it and think I see the underlying root cause. I'm developing a solution now and plan to submit a PR as soon as I've got it tested/documented.

alindima:
Hi! Great, thank you and looking forward to your PRs!

summerlinryan:
@alindima 

[...]

PR #2806, opened 2021-12-08
Partial fix for restore snapshot GET /vm/config bug

summerlinryan:
Fixed the issue where GET /vm/config would return the default config for a VM restored from a snapshot. The config now updates after restoring from a snapshot.

# Reason for This PR

This PR exists to fix the bug described in issue #2783. 

## Description of Changes

Changed restoring a snapshot to include updating the `VmResources` so that the correct config would be returned for the GET /vm/config endpoint. The only parts of the config that are still using default values after restoring a snapshot are the following:  boot-source (which is reasonable since there is no boot-source in the case of a restore), machine-config.smt and machine-config.cpu_template. 

## License Acceptance

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,10 @@
 
 ### Fixed
 
+- GET `/vm/config` was returning a default config object after restoring from a
+  snapshot. It now correctly returns the config of the original microVM, except
+  for boot_config and the cpu_template and smt fields of the machine config,
+  which are currently lost.
 - Fixed incorrect propagation of init parameters in kernel commandline.
   Related to:
   [#2709](https://github.com/firecracker-microvm/firecracker/issues/2709).
--- a/src/api_server/swagger/firecracker.yaml
+++ b/src/api_server/swagger/firecracker.yaml
@@ -612,7 +612,8 @@ paths:
     get:
       summary: Gets the full VM configuration.
       description:
-        Gets configuration for all VM resources.
+        Gets configuration for all VM resources. If the VM is restored from a snapshot, the boot-source,
+        machine-config.smt and machine-config.cpu_template will be empty.
       operationId: getExportVmConfig
       responses:
         200:
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -36,9 +36,11 @@ use crate::vstate::{
     vcpu::{Vcpu, VcpuConfig},
     vm::Vm,
 };
-use crate::{device_manager, Error, EventManager, Vmm, VmmEventsObserver};
+use crate::{device_manager, mem_size_mib, Error, EventManager, Vmm, VmmEventsObserver};
 
+use crate::resources::VmResources;
 use crate::vmm_config::instance_info::InstanceInfo;
+use crate::vmm_config::machine_config::{VmConfigError, VmUpdateConfig};
 use arch::InitrdConfig;
 #[cfg(target_arch = "x86_64")]
 use cpuid::common::is_same_model;
@@ -97,6 +99,8 @@ pub enum StartMicrovmError {
     RegisterMmioDevice(device_manager::mmio::Error),
     /// Cannot restore microvm state.
     RestoreMicrovmState(MicrovmStateError),
+    /// Unable to set VmResources.
+    SetVmResources(VmConfigError),
 }
 
 /// It's convenient to automatically convert `linux_loader::cmdline::Error`s
@@ -177,7 +181,8 @@ impl Display for StartMicrovmError {
                     err_msg
                 )
             }
-            RestoreMicrovmState(err) => write!(f, "Cannot restore microvm state: {}", err),
+            RestoreMicrovmState(err) => write!(f, "Cannot restore microvm state. Error: {}", err),
+            SetVmResources(err) => write!(f, "Cannot set vm resources. Error: {}", err),
         }
     }
 }
@@ -446,6 +451,7 @@ pub fn build_microvm_from_snapshot(
     guest_memory: GuestMemoryMmap,
     track_dirty_pages: bool,
     seccomp_filters: &BpfThreadMap,
+    vm_resources: &mut VmResources,
 ) -> std::result::Result<Arc<Mutex<Vmm>>, StartMicrovmError> {
     use self::StartMicrovmError::*;
     let vcpu_count = u8::try_from(microvm_state.vcpu_states.len())
@@ -513,12 +519,25 @@ pub fn build_microvm_from_snapshot(
         .map_err(MicrovmStateError::RestoreVmState)
         .map_err(RestoreMicrovmState)?;
 
+    vm_resources
+        .update_vm_config(&VmUpdateConfig {
+            vcpu_count: Some(vcpu_count),
+            mem_size_mib: Some(mem_size_mib(&guest_memory) as usize),
+            smt: Some(false),
+            cpu_template: None,
+            track_dirty_pages: Some(track_dirty_pages),
+        })
+        .map_err(SetVmResources)?;
+
     // Restore devices states.
     let mmio_ctor_args = MMIODevManagerConstructorArgs {
         mem: guest_memory,
         vm: vmm.vm.fd(),
         event_manager,
+        for_each_restored_device: VmResources::update_from_restored_device,
+        vm_resources,
     };
+
     vmm.mmio_device_manager =
         MMIODeviceManager::restore(mmio_ctor_args, &microvm_state.device_states)
             .map_err(MicrovmStateError::RestoreDevices)
--- a/src/vmm/src/device_manager/persist.rs
+++ b/src/vmm/src/device_manager/persist.rs
@@ -373,6 +373,7 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                 constructor_args.event_manager,
             )?;
         }
+
         for net_state in &state.net_devices {
             let device = Arc::new(Mutex::new(
                 Net::restore(
@@ -396,6 +397,7 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                 constructor_args.event_manager,
             )?;
         }
+
         if let Some(vsock_state) = &state.vsock_device {
             let ctor_args = VsockUdsConstructorArgs {
                 cid: vsock_state.device_state.frontend.cid,
@@ -436,6 +438,7 @@ impl<'a> Persist<'a> for MMIODeviceManager {
 mod tests {
     use super::*;
     use crate::builder::tests::*;
+    use crate::resources::VmmConfig;
     use crate::vmm_config::balloon::BalloonDeviceConfig;
     use crate::vmm_config::net::NetworkInterfaceConfig;
     use crate::vmm_config::vsock::VsockDeviceConfig;
@@ -643,16 +646,77 @@ mod tests {
         let vmm = default_vmm();
         let device_states: DeviceStates =
             DeviceStates::deserialize(&mut buf.as_slice(), &version_map, 2).unwrap();
+        let vm_resources = &mut VmResources::default();
         let restore_args = MMIODevManagerConstructorArgs {
             mem: vmm.guest_memory().clone(),
             vm: vmm.vm.fd(),
             event_manager: &mut event_manager,
             for_each_restored_device: VmResources::update_from_restored_device,
-            vm_resources: &mut Default::default(),
+            vm_resources,
         };
         let restored_dev_manager =
             MMIODeviceManager::restore(restore_args, &device_states).unwrap();
 
+        let expected_vm_resources = format!(
+            r#"{{
+  "balloon": {{
+    "amount_mib": 123,
+    "deflate_on_oom": false,
+    "stats_polling_interval_s": 1
+  }},
+  "drives": [
+    {{
+      "drive_id": "root",
+      "path_on_host": "{}",
+      "is_root_device": true,
+      "partuuid": null,
+      "is_read_only": true,
+      "cache_type": "Unsafe",
+      "rate_limiter": null,
+      "io_engine": "Sync"
+    }}
+  ],
+  "boot-source": {{
+    "kernel_image_path": "",
+    "initrd_path": null
+  }},
+  "logger": null,
+  "machine-config": {{
+    "vcpu_count": 1,
+    "mem_size_mib": 128,
+    "smt": false,
+    "track_dirty_pages": false
+  }},
+  "metrics": null,
+  "mmds-config": null,
+  "network-interfaces": [
+    {{
+      "iface_id": "netif",
+      "host_dev_name": "hostname",
+      "guest_mac": "00:00:00:00:00:00",
+      "rx_rate_limiter": null,
+      "tx_rate_limiter": null
+    }}
+  ],
+  "vsock": {{
+    "guest_cid": 3,
+    "uds_path": "{}"
+  }}
+}}"#,
+            _block_files
+                .last()
+                .unwrap()
+                .as_path()
+                .to_str()
+                .unwrap()
+                .to_string(),
+            tmp_sock_file.as_path().to_str().unwrap().to_string()
+        );
+
         assert_eq!(restored_dev_manager, original_mmio_device_manager);
+        assert_eq!(
+            expected_vm_resources,
+            serde_json::to_string_pretty(&VmmConfig::from(&*vm_resources)).unwrap()
+        );
     }
 }
--- a/src/vmm/src/device_manager/persist.rs
+++ b/src/vmm/src/device_manager/persist.rs
@@ -10,6 +10,7 @@ use super::mmio::*;
 use crate::EventManager;
 use logger::error;
 
+use crate::resources::VmResources;
 #[cfg(target_arch = "aarch64")]
 use arch::DeviceType;
 use devices::virtio::balloon::persist::{BalloonConstructorArgs, BalloonState};
@@ -129,6 +130,15 @@ pub struct DeviceStates {
     pub balloon_device: Option<ConnectedBalloonState>,
 }
 
+/// A type used to extract the concrete Arc<Mutex<T>> for each of the device types when restoring
+/// from a snapshot.
+pub enum SharedDeviceType {
+    SharedBlock(Arc<Mutex<Block>>),
+    SharedNetwork(Arc<Mutex<Net>>),
+    SharedBalloon(Arc<Mutex<Balloon>>),
+    SharedVsock(Arc<Mutex<Vsock<VsockUnixBackend>>>),
+}
+
 impl DeviceStates {
     fn balloon_serialize(&mut self, target_version: u16) -> VersionizeResult<()> {
         if target_version < 2 && self.balloon_device.is_some() {
@@ -145,6 +155,8 @@ pub struct MMIODevManagerConstructorArgs<'a> {
     pub mem: GuestMemoryMmap,
     pub vm: &'a VmFd,
     pub event_manager: &'a mut EventManager,
+    pub for_each_restored_device: fn(&mut VmResources, SharedDeviceType),
+    pub vm_resources: &'a mut VmResources,
 }
 
 impl<'a> Persist<'a> for MMIODeviceManager {
@@ -323,6 +335,11 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                 .map_err(Error::Balloon)?,
             ));
 
+            (constructor_args.for_each_restored_device)(
+                constructor_args.vm_resources,
+                SharedDeviceType::SharedBalloon(device.clone()),
+            );
+
             restore_helper(
                 device.clone(),
                 device,
@@ -342,6 +359,11 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                 .map_err(Error::Block)?,
             ));
 
+            (constructor_args.for_each_restored_device)(
+                constructor_args.vm_resources,
+                SharedDeviceType::SharedBlock(device.clone()),
+            );
+
             restore_helper(
                 device.clone(),
                 device,
@@ -359,6 +381,11 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                 )
                 .map_err(Error::Net)?,
             ));
+
+            (constructor_args.for_each_restored_device)(
+                constructor_args.vm_resources,
+                SharedDeviceType::SharedNetwork(device.clone()),
+            );
 
             restore_helper(
                 device.clone(),
@@ -385,6 +412,11 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                 )
                 .map_err(Error::Vsock)?,
             ));
+
+            (constructor_args.for_each_restored_device)(
+                constructor_args.vm_resources,
+                SharedDeviceType::SharedVsock(device.clone()),
+            );
 
             restore_helper(
                 device.clone(),
@@ -615,6 +647,8 @@ mod tests {
             mem: vmm.guest_memory().clone(),
             vm: vmm.vm.fd(),
             event_manager: &mut event_manager,
+            for_each_restored_device: VmResources::update_from_restored_device,
+            vm_resources: &mut Default::default(),
         };
         let restored_dev_manager =
             MMIODeviceManager::restore(restore_args, &device_states).unwrap();
--- a/src/vmm/src/resources.rs
+++ b/src/vmm/src/resources.rs
@@ -15,6 +15,7 @@ use crate::vstate::vcpu::VcpuConfig;
 use mmds::ns::MmdsNetworkStack;
 use utils::net::ipv4addr::is_link_local_valid;
 
+use crate::device_manager::persist::SharedDeviceType;
 use mmds::data_store::MmdsVersion;
 use mmds::MMDS;
 use serde::{Deserialize, Serialize};
@@ -169,6 +170,45 @@ impl VmResources {
         }
 
         Ok(resources)
+    }
+
+    /// Updates the resources from a restored device (used for configuring resources when
+    /// restoring from a snapshot).
+    pub fn update_from_restored_device(&mut self, device: SharedDeviceType) {
+        match device {
+            SharedDeviceType::SharedBlock(block) => {
+                self.block.add_device(block);
+            }
+
+            SharedDeviceType::SharedNetwork(network) => {
+                {
+                    let net = network.lock().expect("Poisoned lock");
+
+                    if let Some(mmds_ns) = &net.mmds_ns {
+                        if let Some(mut_mmds_config) = self.mmds_config.as_mut() {
+                            mut_mmds_config.network_interfaces.push(net.id().to_owned());
+                        } else {
+                            // The MMDS Version isn't saved in the snapshot, we set it to its default.
+                            self.mmds_config = Some(MmdsConfig {
+                                version: Default::default(),
+                                network_interfaces: vec![net.id().to_owned()],
+                                ipv4_address: Some(mmds_ns.ipv4_addr),
+                            });
+                        }
+                    }
+                }
+
+                self.net_builder.add_device(network);
+            }
+
+            SharedDeviceType::SharedBalloon(balloon) => {
+                self.balloon.set_device(balloon);
+            }
+
+            SharedDeviceType::SharedVsock(vsock) => {
+                self.vsock.set_device(vsock);
+            }
+        }
     }
 
     /// Returns a VcpuConfig based on the vm config.
@@ -389,6 +429,7 @@ impl From<&VmResources> for VmmConfig {
             .as_ref()
             .map(BootSourceConfig::from)
             .unwrap_or_default();
+
         VmmConfig {
             balloon_device: resources.balloon.get_config().ok(),
             block_devices: resources.block.configs(),
--- a/src/vmm/src/rpc_interface.rs
+++ b/src/vmm/src/rpc_interface.rs
@@ -31,7 +31,7 @@ use crate::vmm_config::net::{
 use crate::vmm_config::snapshot::{CreateSnapshotParams, LoadSnapshotParams, SnapshotType};
 use crate::vmm_config::vsock::{VsockConfigError, VsockDeviceConfig};
 use crate::vmm_config::{self, RateLimiterUpdate};
-use crate::{builder::StartMicrovmError, EventManager};
+use crate::{builder::StartMicrovmError, warn, EventManager};
 use crate::{ExitCode, FC_EXIT_CODE_BAD_CONFIGURATION};
 use logger::{info, update_metric_with_elapsed_time, METRICS};
 use seccompiler::BpfThreadMap;
@@ -318,7 +318,10 @@ impl<'a> PrebootApiController<'a> {
                 .map(|()| VmmData::Empty)
                 .map_err(VmmActionError::Metrics),
             GetBalloonConfig => self.balloon_config(),
-            GetFullVmConfig => Ok(VmmData::FullVmConfig((&*self.vm_resources).into())),
+            GetFullVmConfig => {
+                warn!("If the VM was restored from snapshot, boot-source, machine-config.smt, and machine-config.cpu_template will all be empty.");
+                Ok(VmmData::FullVmConfig((&*self.vm_resources).into()))
+            }
             GetVmMachineConfig => Ok(VmmData::MachineConfiguration(
                 self.vm_resources.vm_config().clone(),
             )),
@@ -448,6 +451,7 @@ impl<'a> PrebootApiController<'a> {
             self.seccomp_filters,
             load_params,
             VERSION_MAP.clone(),
+            self.vm_resources,
         )
         .and_then(|vmm| {
             let ret = if load_params.resume_vm {
@@ -455,6 +459,7 @@ impl<'a> PrebootApiController<'a> {
             } else {
                 Ok(())
             };
+
             ret.map(|()| {
                 self.built_vmm = Some(vmm);
                 VmmData::Empty
@@ -1011,6 +1016,7 @@ mod tests {
         _: &BpfThreadMap,
         _: &LoadSnapshotParams,
         _: versionize::VersionMap,
+        _: &mut MockVmRes,
     ) -> Result<Arc<Mutex<Vmm>>, LoadSnapshotError> {
         Ok(Arc::new(Mutex::new(MockVmm::default())))
     }
--- a/src/vmm/src/vmm_config/net.rs
+++ b/src/vmm/src/vmm_config/net.rs
@@ -132,6 +132,11 @@ impl NetBuilder {
         self.net_devices.iter_mut()
     }
 
+    /// Adds an existing network device in the builder.
+    pub fn add_device(&mut self, device: Arc<Mutex<Net>>) {
+        self.net_devices.push(device);
+    }
+
     /// Builds a network device based on a network interface config. Keeps a device reference
     /// in the builder's internal list.
     pub fn build(&mut self, netif_config: NetworkInterfaceConfig) -> Result<Arc<Mutex<Net>>> {
--- a/src/vmm/src/vmm_config/net.rs
+++ b/src/vmm/src/vmm_config/net.rs
@@ -209,6 +209,7 @@ impl NetBuilder {
 
 #[cfg(test)]
 mod tests {
+    use rate_limiter::RateLimiter;
     use std::str;
 
     use super::*;
@@ -379,5 +380,37 @@ mod tests {
         let configs = net_builder.configs();
         assert_eq!(configs.len(), 1);
         assert_eq!(configs.first().unwrap(), &net_if_cfg);
+    }
+
+    #[test]
+    fn test_add_device() {
+        let mut net_builder = NetBuilder::new();
+        let net_id = "test_id";
+        let host_dev_name = "dev";
+        let guest_mac = "01:23:45:67:89:0b";
+
+        let net = Net::new_with_tap(
+            net_id.to_string(),
+            host_dev_name.to_string(),
+            Some(&MacAddr::parse_str(guest_mac).unwrap()),
+            RateLimiter::default(),
+            RateLimiter::default(),
+            false,
+        )
+        .unwrap();
+
+        net_builder.add_device(Arc::new(Mutex::new(net)));
+        assert_eq!(net_builder.net_devices.len(), 1);
+        assert_eq!(
+            net_builder
+                .net_devices
+                .pop()
+                .unwrap()
+                .lock()
+                .unwrap()
+                .deref()
+                .id(),
+            net_id
+        );
     }
 }
--- a/src/vmm/src/vmm_config/vsock.rs
+++ b/src/vmm/src/vmm_config/vsock.rs
@@ -192,4 +192,24 @@ pub(crate) mod tests {
         ));
         let _ = format!("{}{:?}", err, err);
     }
+
+    #[test]
+    fn test_set_device() {
+        let mut vsock_builder = VsockBuilder::new();
+        let mut tmp_sock_file = TempFile::new().unwrap();
+        tmp_sock_file.remove().unwrap();
+        let vsock = Vsock::new(
+            0,
+            VsockUnixBackend::new(1, tmp_sock_file.as_path().to_str().unwrap().to_string())
+                .unwrap(),
+        )
+        .unwrap();
+
+        vsock_builder.set_device(Arc::new(Mutex::new(vsock)));
+        assert!(vsock_builder.inner.is_some());
+        assert_eq!(
+            vsock_builder.inner.unwrap().uds_path,
+            tmp_sock_file.as_path().to_str().unwrap().to_string()
+        )
+    }
 }
--- a/src/vmm/src/vmm_config/vsock.rs
+++ b/src/vmm/src/vmm_config/vsock.rs
@@ -77,6 +77,19 @@ impl VsockBuilder {
         Self { inner: None }
     }
 
+    /// Inserts an existing vsock device.
+    pub fn set_device(&mut self, device: Arc<Mutex<Vsock<VsockUnixBackend>>>) {
+        self.inner = Some(VsockAndUnixPath {
+            uds_path: device
+                .lock()
+                .expect("Poisoned lock")
+                .backend()
+                .host_sock_path()
+                .to_owned(),
+            vsock: device.clone(),
+        });
+    }
+
     /// Inserts a Unix backend Vsock in the store.
     /// If an entry already exists, it will overwrite it.
     pub fn insert(&mut self, cfg: VsockDeviceConfig) -> Result<()> {
--- a/tests/integration_tests/functional/test_api.py
+++ b/tests/integration_tests/functional/test_api.py
@@ -19,8 +19,9 @@ import host_tools.network as net_tools
 from conftest import _test_images_s3_bucket
 
 from framework.utils import is_io_uring_supported
-from framework.artifacts import ArtifactCollection
-from framework.builder import MicrovmBuilder
+from framework.artifacts import ArtifactCollection, SnapshotType, \
+    NetIfaceConfig, DEFAULT_DEV_NAME, DEFAULT_TAP_NAME
+from framework.builder import MicrovmBuilder, SnapshotBuilder
 
 MEM_LIMIT = 1000000000
 
@@ -554,7 +555,7 @@ def test_api_machine_config(test_microvm_with_api):
         assert "CPU templates are not supported on aarch64" in response.text
 
     response = test_microvm.actions.put(action_type='InstanceStart')
-    if utils.get_cpu_vendor() != utils.CpuVendor.INTEL:
+    if utils.get_cpu_vendor() == utils.CpuVendor.AMD:
         # We shouldn't be able to apply Intel templates on AMD hosts
         fail_msg = "Internal error while starting microVM: Error configuring" \
                    " the vcpu for boot: Cpuid error: InvalidVendor"
@@ -1359,6 +1360,141 @@ def test_api_balloon(test_microvm_with_api):
     # requesting u32::MAX / 128.
     response = test_microvm.balloon.patch(amount_mib=33554432)
     assert test_microvm.api_session.is_status_bad_request(response.status_code)
+
+
+def test_get_full_config_after_restoring_snapshot(bin_cloner_path):
+    """
+    Test the configuration of a microVM after restoring from a snapshot.
+
+    @type: functional
+    """
+    microvm_builder = MicrovmBuilder(bin_cloner_path)
+    net_iface = NetIfaceConfig()
+    vm_instance = microvm_builder.build_vm_nano(
+        net_ifaces=[net_iface],
+        io_engine='Sync')
+    test_microvm = vm_instance.vm
+    root_disk = vm_instance.disks[0]
+    ssh_key = vm_instance.ssh_key
+    cpu_vendor = utils.get_cpu_vendor()
+
+    setup_cfg = {}
+
+    # Basic config also implies a root block device.
+    setup_cfg['machine-config'] = {
+        'vcpu_count': 2,
+        'mem_size_mib': 256,
+        'smt': True,
+        'track_dirty_pages': False
+    }
+
+    if cpu_vendor == utils.CpuVendor.INTEL:
+        setup_cfg['machine-config']['cpu_template'] = 'C3'
+
+    test_microvm.machine_cfg.patch(**setup_cfg['machine-config'])
+
+    setup_cfg['drives'] = [{
+        'drive_id': 'rootfs',
+        'path_on_host': f"/{os.path.basename(root_disk.local_path())}",
+        'is_root_device': True,
+        'partuuid': None,
+        'is_read_only': False,
+        'cache_type': 'Unsafe',
+        'rate_limiter': None,
+        'io_engine': 'Sync'
+    }]
+
+    # Add a memory balloon device.
+    response = test_microvm.balloon.put(amount_mib=1, deflate_on_oom=True)
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+    setup_cfg['balloon'] = {
+        'amount_mib': 1,
+        'deflate_on_oom': True,
+        'stats_polling_interval_s': 0
+    }
+
+    # Add a vsock device.
+    response = test_microvm.vsock.put(
+        guest_cid=15,
+        uds_path='vsock.sock'
+    )
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+    setup_cfg['vsock'] = {
+        'guest_cid': 15,
+        'uds_path': 'vsock.sock'
+    }
+
+    setup_cfg['logger'] = None
+    setup_cfg['metrics'] = None
+    setup_cfg['mmds-config'] = {
+        'version': "V1",
+        'network_interfaces': [DEFAULT_DEV_NAME]
+    }
+
+    response = test_microvm.mmds.put_config(json=setup_cfg['mmds-config'])
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+
+    # Start the microvm.
+    test_microvm.start()
+
+    # Add a tx rate limiter to the net device.
+    tx_rl = {
+        'bandwidth': {
+            'size': 1000000,
+            'refill_time': 100,
+            'one_time_burst': None
+        },
+        'ops': None
+    }
+
+    response = test_microvm.network.patch(
+        iface_id=DEFAULT_DEV_NAME,
+        tx_rate_limiter=tx_rl
+    )
+    assert test_microvm.api_session.is_status_no_content(response.status_code)
+    setup_cfg['network-interfaces'] = [{
+        'guest_mac': net_tools.mac_from_ip(net_iface.guest_ip),
+        'iface_id': DEFAULT_DEV_NAME,
+        'host_dev_name': DEFAULT_TAP_NAME,
+        'rx_rate_limiter': None,
+        'tx_rate_limiter': tx_rl
+    }]
+    # Create a snapshot builder from a microvm.
+    snapshot_builder = SnapshotBuilder(test_microvm)
+
+    # Create base snapshot.
+    snapshot = snapshot_builder.create([root_disk.local_path()],
+                                       ssh_key,
+                                       SnapshotType.FULL)
+
+    microvm, _ = microvm_builder.build_from_snapshot(snapshot, True, False)
+
+    expected_cfg = setup_cfg.copy()
+
+    # We expect boot-source, machine-config.smt, and
+    # machine-config.cpu_template to all be empty/default after restoring
+    # from a snapshot.
+    expected_cfg['boot-source'] = {
+        'kernel_image_path': '',
+        'initrd_path': None
+    }
+    expected_cfg['machine-config']['smt'] = False
+
+    if cpu_vendor == utils.CpuVendor.INTEL:
+        expected_cfg['machine-config'].pop('cpu_template')
+
+    # no ipv4 specified during PUT /mmds/config so we expect the default
+    expected_cfg['mmds-config'] = {
+        'version': 'V1',
+        'ipv4_address': '169.254.169.254',
+        'network_interfaces': [DEFAULT_DEV_NAME]
+    }
+
+    # Validate full vm configuration post-restore.
+    response = microvm.full_cfg.get()
+    assert microvm.api_session.is_status_ok(response.status_code)
+    assert response.json() != setup_cfg
+    assert response.json() == expected_cfg
 
 
 def test_get_full_config(test_microvm_with_api):
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -122,7 +122,7 @@ pub struct Net {
     pub(crate) device_state: DeviceState,
     pub(crate) activate_evt: EventFd,
 
-    pub(crate) mmds_ns: Option<MmdsNetworkStack>,
+    pub mmds_ns: Option<MmdsNetworkStack>,
 
     #[cfg(test)]
     pub(crate) mocks: Mocks,
--- a/src/devices/src/virtio/vsock/unix/muxer.rs
+++ b/src/devices/src/virtio/vsock/unix/muxer.rs
@@ -332,6 +332,10 @@ impl VsockMuxer {
         Ok(muxer)
     }
 
+    pub fn host_sock_path(&self) -> &str {
+        &self.host_sock_path
+    }
+
     /// Handle/dispatch an epoll event to its listener.
     fn handle_event(&mut self, fd: RawFd, event_set: EventSet) {
         debug!(
--- a/src/mmds/src/ns.rs
+++ b/src/mmds/src/ns.rs
@@ -61,7 +61,7 @@ pub struct MmdsNetworkStack {
     // The Ethernet MAC address of the MMDS server.
     pub(crate) mac_addr: MacAddr,
     // MMDS server IPv4 address.
-    pub(crate) ipv4_addr: Ipv4Addr,
+    pub ipv4_addr: Ipv4Addr,
     // ARP reply destination IPv4 address (requester of address resolution reply).
     // It is the Ipv4Addr of the network interface for which the MmdsNetworkStack
     // routes the packets.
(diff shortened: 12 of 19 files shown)


summerlinryan:
@acatangiu @alindima I've left TODOs that describe the parts that I'm still confused on as of now. Sorry there are more TODOs than I was hoping for, just trying to learn and make more sense of how Firecracker all works together.

alindima:
@summerlinryan sorry, I haven't found the time to look over the PR yet, but will do it this week! Thanks!

src/vmm/src/resources.rs
```diff
@@ -153,6 +159,67 @@ impl VmResources {
...
+        let mut resources = Self::default();
+        resources.boot_config = None;
+
+        if let Ok(balloon_config) = vmm.balloon_config() {
+            resources
+                .set_balloon_device(BalloonDeviceConfig::from(balloon_config))
+                .map_err(Error::BalloonDevice)?;
+        }
+
+        vmm.mmio_device_manager.for_each_virtio_device(|virtio_type, _device_id, _device_info, virtio_device| {
```
alindima, line 173: We can limit the calls to `for_each_virtio_device` to just one, by `match`-ing on the `virtio_type` inside the closure

src/vmm/src/resources.rs
```diff
@@ -153,6 +159,67 @@ impl VmResources {
...
+            resources.set_block_device(BlockDeviceConfig::from(
+                virtio_device
+                    .lock()
+                    .expect("Poisoned lock")
+                    .as_any()
+                    .downcast_ref::<Block>()
+                    .unwrap()
+            ))
+            .map_err(Error::BlockDevice)
+        })?;
```
alindima, line 187: The methods here are actually reconstructing the devices. We only need to push the existing ones to the `Builder` objects. You will need to add some helper methods on the Builder objects for this

src/vmm/src/resources.rs
```diff
@@ -153,6 +159,67 @@ impl VmResources {
...
+
+        resources.set_vm_config(&VmConfig {
+            vcpu_count: Some(vmm.vcpus_handles.len() as u8),
+            mem_size_mib: Some(mem_size_mib(vmm.guest_memory()) as usize),
+            ht_enabled: Some(false), // TODO: how to determine if HT enabled?
+            cpu_template: None, // TODO: how to determine cpu_template?
+            track_dirty_pages: vmm
+                .guest_memory()
+                .iter()
+                .any(|region| (region.flags() as u32) & KVM_MEM_LOG_DIRTY_PAGES != 0)
```
alindima, line 197: a nicer way of doing this is to: `.all(|region| (region.bitmap().is_some())`

src/vmm/src/rpc_interface.rs
```diff
@@ -441,20 +441,23 @@ impl<'a> PrebootApiController<'a> {
...
             &self.instance_info,
             &mut self.event_manager,
             self.seccomp_filters,
             load_params,
             VERSION_MAP.clone(),
-        )
-        .and_then(|vmm| {
+        ).and_then(|(vm_resources, vmm)| {
+            println!("{}", serde_json::to_string(&VmmConfig::from(&vm_resources)).unwrap());
+            // TODO: how to deal with lifetime issue here
```
alindima, line 453: The problem here is that you want to take a reference to an owned vmm variable, but the vmm variable is local to the closure. So the reference `self.vm_resources` would live longer than the actual object.

Instead of returning the `VmResources` from `restore_from_snapshot`, you should pass it in there as a mutable reference and modify that variable to contain the device information.

src/vmm/src/resources.rs
```diff
@@ -153,6 +159,67 @@ impl VmResources {
...
+                    .expect("Poisoned lock")
+                    .as_any()
+                    .downcast_ref::<Net>()
+                    .unwrap()
+            )).map_err(Error::NetDevice)?;
+
+            Ok(())
+        })?;
+
+        // TODO: how to determine vsock & mmds config (seems like should be able to use for_each_virtio_device
```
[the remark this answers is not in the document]
alindima: I also noticed that currently (even if not resumed from a snapshot), the outputted MMDS config is `None` if a custom IP address has not been specified but a net device was configured with `allow_mmds_requests: true`. We should actually output the default MMDS config in this case.
You could include this fix in the PR.

summerlinryan:
Thanks for the feedback! Just wanted to let you know that I've started working through implementing the suggestions and hope to have some more code pushed within the next day or two.

resources/seccomp/x86_64-unknown-linux-musl.json
```diff
@@ -46,6 +46,18 @@
...
+                "syscall": "stat",
+                "comment": "Used for restoring from snapshot (setting vm resources of api controller)"
+            },
+            {
+                "syscall": "timerfd_create",
+                "comment": "Used for restoring from snapshot (setting vm resources of api controller)"
+            },
+            {
+                "syscall": "eventfd2",
+                "comment": "Used for restoring from snapshot (setting vm resources of api controller)"
```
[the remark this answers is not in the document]
summerlinryan: I'm still trying to make full sense of what's happening with this but if I remove the above syscalls then I get the errors related to "intercepting signal 31, code 1 and a bad syscall(4)" when trying to load the snapshot with the PUT /snapshot/load route.

src/vmm/src/resources.rs
```diff
@@ -153,6 +159,67 @@ impl VmResources {
...
+                    .unwrap()
+            ))
+            .map_err(Error::BlockDevice)
+        })?;
+
+        resources.set_vm_config(&VmConfig {
+            vcpu_count: Some(vmm.vcpus_handles.len() as u8),
+            mem_size_mib: Some(mem_size_mib(vmm.guest_memory()) as usize),
+            ht_enabled: Some(false), // TODO: how to determine if HT enabled?
+            cpu_template: None, // TODO: how to determine cpu_template?
```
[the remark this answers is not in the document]
summerlinryan: Sounds good, I'll update the swagger endpoint description to indicate that those pieces of information will be inaccurate when using GET /vm/config with a restored VM. For the log message, are you referring to just using `println!` to log some information whenever the `GET /vm/config` endpoint is hit or are you referring to updating some documentation?

src/vmm/src/resources.rs
summerlinryan: That makes perfect sense. I'll definitely do that :+1:

resources/seccomp/x86_64-unknown-linux-musl.json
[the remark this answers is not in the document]
alindima: hm I'm fairly sure they were caused by the fact that the code was reconstructing the devices when populating the `VmResources` struct. If we no longer do this we should be able to remove them

src/vmm/src/resources.rs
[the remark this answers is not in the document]
alindima: similarly to a `println!`, use the `warn!` macro which is used for logging warnings.
(There are other macros in the logger crate as well, for `error!`, `info!`, `debug!`)

alindima:
> Thanks for the feedback! Just wanted to let you know that I've started working through implementing the suggestions and hope to have some more code pushed within the next day or two.

Great! Thank you for the effort! Looking forward to the updates

src/vmm/src/resources.rs
summerlinryan: So this piece of feedback is the one that I'm having the hardest time implementing. For the `BlockBuilder` I was thinking about adding a new method like this:

```
pub fn insert_block(&mut self, block: &Block) {
    if block.is_root_device() {
        self.list.push_front(Arc::new(Mutex::new(block.clone())));
    } else {
        self.list.push_back(Arc::new(Mutex::new(block.clone())));
    }
}
```
which would allow using something like the following when setting `VmResources` from a snapshot

```
self.set_block_device(
    virtio_device
        .lock()
        .expect("Poisoned lock")
        .as_any()
        .downcast_ref::<Block>()
        .unwrap()
        .to_owned()
);
```

where `self` is the `VmResources` and the `virtio_device` is one of the arguments of the closure in the `for_each_virtio_device`. The `set_block_device` above would be a method that just calls the `insert_block` on the `BlockBuilder` (the existing `set_block_device` on `VmResources` would be renamed to `set_block_device_from_config` since it uses a `BlockDeviceConfig`). However, this doesn't work because it'd either require that Block satisfies the Clone trait would require a ton of other types inside of Block satisfying the Clone trait which I was thinking wouldn't be the correct approach.

acatangiu: What you need to do get a hold of the `Arc<>` part of `Arc<Mutex<Device>>` so that you can clone the reference without also cloning the device.

The problem is that the `MMIODeviceManager::restore` builds devices internally and you never see the `Arc`. Furthermore the `MMIODeviceManager` interface doesn't offer any API to expose the `Arc`: there is only https://github.com/firecracker-microvm/firecracker/blob/83ec483d4254936e1e2fd4adfe94761611a74a93/src/devices/src/bus.rs#L98 which provides only plain reference

So I see two options:
1. Change `MMIODeviceManager::get_device()` and `Bus::get_device()` APIs to return `Option<(u64, Arc<Mutex<dyn BusDevice>>)>`, the you can change the passed function signature in `MMIODeviceManager::for_each_device()` to also take `Arc<Mutex<dyn BusDevice>>` so that within the closure you can clone the `Arc` reference to the device and thus share ownership of it.
2. When restoring devices from states using the `MMIODeviceManager::restore` here https://github.com/firecracker-microvm/firecracker/blob/83ec483d4254936e1e2fd4adfe94761611a74a93/src/vmm/src/builder.rs#L528 you can also provide a function/closure parameter to `MMIODeviceManager::restore` to do auxiliary work on the restored devices (like in your case, also adding them to VmResources). The new signature could look like:
    ```rust
    fn restore(
        constructor_args: Self::ConstructorArgs,
        state: &Self::State,
        for_each_device: Fn(Arc<Mutex<dyn BusDevice>>),
    ) -> Result<Self, Self::Error> {
    ```

I think first option has less cognitive complexity.

summerlinryan: Sorry again for being slow to respond I got busy over the holidays.

TL;DR: How do I cast a `Arc<Mutex<dyn VirtioDevice>>` into a `Arc<Mutex<Block>>`?

I'm starting to wonder if I just don't have a deep enough understanding of Rust to work on a project as sophisticated as Firecracker. If helping guide me through this issue becomes too burdensome, just let me know if it's best if someone else takes over this bug and I'll completely understand (I can try to contribute again in the future to a simpler problem or when I have a better understanding of Rust/Firecracker).

If you guys are able to continue helping, I'm at a point where I have modified the `MMIODeviceManager::get_device()` and `Bus::get_device()` APIs to return the actual `Arc<Mutex<dyn BusDevice>`, I've modified the closure of `MMIODeviceManager::for_each_virtio_device` to take a `Arc<Mutex<dyn VirtioDevice>>` and I've modified the closure of  `MMIODeviceManager::for_each_device` to take a `Arc<Mutex<dyn BusDevice>>`. The problem I'm having now is that I have a method that looks like 

```rust
pub fn set_block_device(&mut self, block_device: Arc<Mutex<Block>>>)
```
on the `struct VmResources` but the when I clone the `Arc<Mutex<dyn VirtioDevice>>` inside of the `MMIODeviceManager::for_each_device` closure, it can't be passed to `set_block_device` as a `Arc<Mutex<Block>>` since the types don't match. I've tried downcasting the `dyn VirtioDevice` into the `Block` inside of the `Arc<Mutex<dyn VirtioDevice>>` but that seems to cast the entire expression to a `&Block` rather than a `Arc<Mutex<Block>>` which is what I need.

acatangiu: Hi @summerlinryan,

> I'm starting to wonder if I just don't have a deep enough understanding of Rust to work on a project as sophisticated as Firecracker.

Don't beat yourself up, this issue is not a straightforward issue. But it is a good learning opportunity! :+1: 

You are right, I don't know if it's possible to clone `Arc<Mutex<dyn Trait>>` into `Arc<Mutex<T: Trait>>`.

The other way around is easily allowed through type coercion (`let t1 = Arc::clone(&struct) as Arc<Mutex<Trait>>;`) because correctness can be easily checked at compile time.

I believe it could be done by implementing `Any` trait for `Mutex<dyn BusDevice>`, `Mutex<Block>`, etc. Then you can use [Arc::downcast](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.downcast) to get what you want.

Having said that, you could also do the simpler thing and go with option 2 from [here](https://github.com/firecracker-microvm/firecracker/pull/2806#discussion_r774466266) with a simple `enum` that gets rid of all dynamic casting:
```rust
enum SharedDeviceTypeWrapper {
    SharedBlock(Arc<Mutex<Block>>),
    SharedNetwork(Arc<Mutex<Net>>),
    SharedBalloon(Arc<Mutex<Balloon>>),
    SharedEtc(Arc<Mutex<Etc>>),
}

fn MMIODeviceManager::restore(
    constructor_args: Self::ConstructorArgs,
    state: &Self::State,
    for_each_restored_device: FnOnce(enum SharedDeviceTypeWrapper),
) -> Result<Self, Self::Error>
{
    // ...
    if let Some(balloon_state) = &state.balloon_device {
        let device = Arc::new(Mutex::new(
            Balloon::restore(
                BalloonConstructorArgs { mem: mem.clone() },
                &balloon_state.device_state,
            )
            .map_err(Error::Balloon)?,
        ));

        // CALL NEW FN HERE:
        for_each_restored_device(SharedDeviceTypeWrapper::SharedBalloon(device.clone()));
        // same code as before after this ^

        restore_helper(
            device.clone(),
            device,
            &balloon_state.device_id,
            &balloon_state.transport_state,
            &balloon_state.mmio_slot,
            constructor_args.event_manager,
        )?;
    }
    // ...
}
```

`pub fn build_microvm_from_snapshot(` could take a `vm_resources: &super::resources::VmResources` reference, and in `build_microvm_from_snapshot` have smth like:
```rust
    // ...
    let register_to_resources = |shared_dev: enum SharedDeviceTypeWrapper| {
        match shared_dev {
            SharedDeviceTypeWrapper::SharedBalloon(b) => vm_resources.register_existing_balloon_device(b),
            ...
        };
    };
    // ...
    vmm.mmio_device_manager =
        MMIODeviceManager::restore(mmio_ctor_args, &microvm_state.device_states, register_to_resources)
            .map_err(MicrovmStateError::RestoreDevices)
```

`register_to_resources()` closure in example above could even be a member function of `VmResources` instead of a closure defined in `builder.rs`.

summerlinryan: I started adding the `for_each_restored_device` function to `MMIODeviceManager::restore` but I noticed that requires adding it to the `Persist` trait which is implemented by a lot of different structs across many different crates. To do so would require either adding a ton of different crate dependencies to allow all of those different crates to know about the new `SharedDeviceTypeWrapper` enum. It would also require adding the new closure argument to all places that currently call the `Persist::restore` method which is quite a few and most of them don't really need to specify the closure anyway since it's primarily just used for setting resources when restoring from a snapshot in `build_microvm_from_snapshot`

Unless I'm misunderstanding the above, then I'm thinking that maybe the next approach to try is to implement Any for `Mutex<dyn VirtioDevice>` like you mentioned, so I'll try that unless you have a workaround to the above problems and still prefer the `for_each_restored_device` approach.

acatangiu: I think you can add it to `MMIODevManagerConstructorArgs` [here](https://github.com/firecracker-microvm/firecracker/blob/73ccdfd643fd60b7c22f06fd05765470de3102b2/src/vmm/src/builder.rs#L522) so that you don't have to change the trait (definitely don't want to change that generic trait with this specific param).

src/vmm/src/resources.rs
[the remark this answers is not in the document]
summerlinryan: > the outputted MMDS config is None if a custom IP address has not been specified but a net device was configured with allow_mmds_requests: true. We should actually output the default MMDS config in this case.

I'm a little confused by this; could you elaborate? I see the mmds-config appears to be always be null when hitting the `GET /vm/config` endpoint which I guess is what you're saying is a problem. Specifically the part I'm confused on is what you mean by "if a custom IP address has not been specified but a net device was configured with allow_mmds_requests: true". I'm reading through the MMDS documentation again to see if I can make more sense of what you mean by that.

summerlinryan:
Well, finally some progress has been made from an implementation standpoint! :+1:  It looks like all of the devices are showing correctly in the config in `GET /vm/config` after restoring from a snapshot. I'm sure there is still plenty of room for feedback but it's good to at least see things working. The only thing that isn't working is the mmds config which I've left another comment for in response to what Alin was saying about the mmds config. 

@acatangiu I ended up using just a plain function pointer instead of the FnOnce trait for the `for_each_restored_device` function because I was running into some issues with FnOnce not being `Sized` and therefore the `MMIODevManagerConstructorArgs` wasn't a valid parameter for the `restore` method. Not sure if a plain function pointer is acceptable here or not. If not, I can revisit trying to get it to work with FnOnce, maybe by making the constructor args generic like `MMIODevManagerConstructorArgs<F: FnOnce(&mut self VmResources, SharedDeviceType)`?

The next thing I'm moving onto is adding some tests and making all existing tests pass. I'm thinking unit tests won't be enough and I'll probably need to integration tests for this since it involves a VM being restored from a snapshot.

src/vmm/src/resources.rs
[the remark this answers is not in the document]
alindima: > I'm a little confused by this; could you elaborate? I see the mmds-config appears to be always be null when hitting the GET /vm/config endpoint which I guess is what you're saying is a problem. Specifically the part I'm confused on is what you mean by "if a custom IP address has not been specified but a net device was configured with allow_mmds_requests: true". I'm reading through the MMDS documentation again to see if I can make more sense of what you mean by that.

If a user does not customise the default IP address used for MMDS, there is no MMDS IP address outputted when dumping the vm config (while there is actually a default one).

We could open a separate issue for this, you don't need to tackle it in this PR. Let's take them one step at a time. WDYT?

src/api_server/swagger/firecracker.yaml
```diff
@@ -604,7 +604,8 @@ paths:
     get:
       summary: Gets the full VM configuration.
       description:
-        Gets configuration for all VM resources.
+        Gets configuration for all VM resources. If the VM is restored from a snapshot the boot-source,
+        machine-config.ht_enabled, and machine-config.cpu_template will all have default values.
```
alindima, line 608:
```suggestion
        Gets configuration for all VM resources. If the VM is restored from a snapshot, the boot-source,
        machine-config.ht_enabled and machine-config.cpu_template will be empty.
```

They should be empty, since they don't really have default values

resources/seccomp/x86_64-unknown-linux-musl.json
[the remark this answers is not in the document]
alindima: let's now remove these as well

src/vmm/src/builder.rs
```diff
@@ -556,7 +565,22 @@ pub fn build_microvm_from_snapshot(
...
+            vcpu_count: Some(vmm.vcpus_handles.len() as u8),
+            mem_size_mib: Some(mem_size_mib(vmm.guest_memory()) as usize),
+            ht_enabled: None,
+            cpu_template: None,
+            track_dirty_pages: vmm
+                .guest_memory()
+                .iter()
+                .any(|region| (region.bitmap().is_some()))
+        }).map_err(SetVmResources)?;
+    }
```
alindima, line 581: I'd move this a bit higher up in the function body, and remove the extra code block. We can also use the `vmm` variable that is already in scope for this function

src/vmm/src/builder.rs
```diff
@@ -556,7 +565,22 @@ pub fn build_microvm_from_snapshot(
     .map_err(Error::SeccompFilters)
     .map_err(StartMicrovmError::Internal)?;
 
-    Ok(vmm)
+    {
+        let vmm = vmm_reference.lock().expect("Poisoned lock");
+
+        vm_resources.set_vm_config(&VmConfig {
+            vcpu_count: Some(vmm.vcpus_handles.len() as u8),
```
alindima, line 572: There is already a `vcpu_count` variable in scope that we may use. Same for `track_dirty_pages` and `guest_memory`

src/vmm/src/resources.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-#![deny(warnings)]
+// #![deny(warnings)]
```
alindima, line 4: let's remove this if there aren't any warnings to suppress

src/vmm/src/vmm_config/net.rs
```diff
@@ -147,6 +147,21 @@ impl NetBuilder {
...
+    pub fn add_device(&mut self, device: Arc<Mutex<Net>>) {
+        if let Some(index) = self
+            .net_devices
+            .iter()
+            .position(|net| {
+                net.lock().expect("Poisoned lock").id() == &device.lock().expect("Poisoned lock").id
+            })
+        {
+            self.net_devices.swap_remove(index);
+        }
```
alindima, line 160: When restoring from a snapshot there will never be two net devices with the same id. The Firecracker VM on which the snapshot was taken would have prevented this at net device creation time.

alindima (commented):
Nice work!

alindima:
> Not sure if a plain function pointer is acceptable here or not. If not, I can revisit trying to get it to work with FnOnce, maybe by making the constructor args generic like MMIODevManagerConstructorArgs<F: FnOnce(&mut self VmResources, SharedDeviceType)?

I think this is perfectly fine. This is not a closure that captures variables from the outside scope so a plain function pointer works just right. @acatangiu agree?

> The next thing I'm moving onto is adding some tests and making all existing tests pass. I'm thinking unit tests won't be enough and I'll probably need to integration tests for this since it involves a VM being restored from a snapshot.

Yes, try to add unit tests that cover all the new code. Python integration tests will also be needed to ensure that everything works end-to-end.

Thanks for the hard work! This is no easy issue and I think we're on the right track.

summerlinryan:
A few questions: 

1. I noticed a coverage test passing and I feel like I recall reading somewhere in the documentation a while back that mentioned a need to update some coverage numbers in order to get some tests to pass. I can't seem to find any detailed information in the docs about updating the coverage numbers. Any idea there is some documentation describing what coverage numbers need to be updated and how?
2. For integration tests, I've seen Python integration tests and Rust integration tests: should I be adding both, or just python integration tests? 
3. Any good articles/documentation that you know of for signing commits with an already provided public GPG key? Everything online I seem to find about signing commits involves generating a key pair but that doesn't seem right to me since this repo already provides a public GPG key. I added the public GPG key in this repo to my GitHub account but I'm having a difficult time telling Git locally to use the GPG key to sign.

alindima:
> Oh no. I think I might've messed things up pretty badly while trying to get this develop branch in my fork rebased onto the latest main. I checked out main in my fork, did a pull from the upstream main branch, then checked out my local feature branch, and ran `git rebase main`. There were a few small conflicts during the rebase that I resolved. After that I tried pushing those changes to the feature branch of my fork but that had diverged with my local feature branch so I had to do a pull. After doing that and addressing the merge conflicts, things look pretty messed up. I'm still trying to figure out what happened and how to correct it so please bear with me until I get that figured out.

no worries. you can also create a new branch from latest main and cherry-pick the commits you've added on this branch.
You can then `reset --hard` this snapshot_vm_config branch to be identical to this new branch and force-push it.

alindima:
> A few questions:
> 
> 1. I noticed a coverage test failing and I feel like I recall reading somewhere in the documentation a while back that mentioned a need to update some coverage numbers in order to get some tests to pass. I can't seem to find any detailed information in the docs about updating the coverage numbers. Any idea there is some documentation describing what coverage numbers need to be updated and how?

Here is an example commit that updates the numbers: https://github.com/firecracker-microvm/firecracker/pull/2835/commits/cf72ba578da426acad5a3a4faa58c27a8271de6d

You need to update them if you see the coverage test reported as failed in the CI. The test will tell you what the current coverage is.
Note that the coverage dictionary is per-CPU type (ARM, Intel, AMD) and also kernel-version dependent. (So you may need to update a total of 6 values).

If you need help with this, I can show you how to update them once all other things are done and the PR is ready for merge.

As a side note, you need one of the maintainers to unblock the CI run. Feel free to ping me on slack if you need to have the CI unblocked and we don't notice it.

> 2. For integration tests, I've seen Python integration tests and Rust integration tests: should I be adding both, or just python integration tests?

In this specific case, python integration tests are what we're after. We want to test the API end-to-end. The python integration tests basically compile and run firecracker as a regular user would do.
Rust integration tests on the other hand test the Rust libraries/crates and cannot test the binary itself.

> 3. Any good articles/documentation that you know of for signing commits with an already provided public GPG key? Everything online I seem to find about signing commits involves generating a key pair but that doesn't seem right to me since this repo already provides a public GPG key. I added the public GPG key in this repo to my GitHub account but I'm having a difficult time telling Git locally to use the GPG key to sign.

Don't bother with GPG-signing your commits. Running `git commit -s` will add a `Signed-off-by` line to each commit, which is what our tests verify.

src/api_server/swagger/firecracker.yaml
summerlinryan: I figured this comment meant that boot-source should be null when a snapshot is restored instead of a default full object value. Is that what you meant? I made boot-source null in this [commit](https://github.com/firecracker-microvm/firecracker/pull/2806/commits/3d652cc2d8af7b396e0140812172277976bb7ad6).

summerlinryan:
I'm working on getting this passing for existing tests and adding new tests but I have a question. How do you all debug the integration tests? I attempted to run `pytest --pdb` to get debugging but that didn't work because it required sudo.  Running `sudo pytest` doesn't work though because `pytest` isn't recognized as a command for sudo. I tried using `su` to temporarily get me in a root environment but then I no longer have access to the repo since the repo is stored in a folder in my users home directory. I must be overlooking something obvious here.

alindima:
> I'm working on getting this passing for existing tests and adding new tests but I have a question. How do you all debug the integration tests? I attempted to run `pytest --pdb` to get debugging but that didn't work because it required sudo. Running `sudo pytest` doesn't work though because `pytest` isn't recognized as a command for sudo. I tried using `su` to temporarily get me in a root environment but then I no longer have access to the repo since the repo is stored in a folder in my users home directory. I must be overlooking something obvious here.

We don't have a go-to way for debugging integration tests. Usually the output of the failing test is enough. You may output additional things using `print()` inside the test and capture the output when running with `pytest -s`.

Now I see that the tests that are failing on this PR are build and style tests - mostly unit tests.
In this case the output is pretty verbose and contains a lot of information.

Note that some performance tests will likely fail on your machine, because they assume more powerful machines (like AWS m5d.metal), but this is no issue as long as they pass in the CI.

We recommend running pytest inside devtool. You can get a root shell inside the devtool container like so:
`tools/devtool shell -p`

summerlinryan:
So I noticed that I accidentally violated some of the contribution quality standards by not ensuring all commits were formatted & passing unit tests. I've corrected that now in my most recent commits but I'm wondering if I should do a interactive rebase to ensure that each and every commit is formatted & passing unit tests. Also, I'm thinking I'll need to interactive rebase & force push to adjust my commit messages because of them are slightly exceeding the title length requirement of gitlint and missing signoff messages. Do both of those adjustments sound correct to you? 

Lastly, this is a small question, do you think the testing of the config when restoring a snapshot should go in the test_api.py test file or the test_snapshot_[basic/advanced].py files? I'm leaning towards the test_api.py but I could see it either way.

src/vmm/src/device_manager/persist.rs
```diff
@@ -616,6 +650,8 @@ mod tests {
             mem: vmm.guest_memory().clone(),
             vm: vmm.vm.fd(),
             event_manager: &mut event_manager,
+            for_each_restored_device: VmResources::update_from_restored_device,
+            vm_resources: &mut Default::default(),
```
alindima, line 654: In this test we can also assert the expected `vm_resources` configuration

alindima (commented):
We should also add a changelog entry, unit tests for the add_device/set_device methods and python integration tests.

Note that we have recently changed the MMDS API so it's best to rebase your branch again. In a nutshell, the net device no longer has the allow_mmds_requests.
It's best to do this periodically and to fix eventual conflicts.

alindima:
> So I noticed that I accidentally violated some of the contribution quality standards by not ensuring all commits were formatted & passing unit tests. I've corrected that now in my most recent commits but I'm wondering if I should do a interactive rebase to ensure that each and every commit is formatted & passing unit tests. Also, I'm thinking I'll need to interactive rebase & force push to adjust my commit messages because of them are slightly exceeding the title length requirement of gitlint and missing signoff messages. Do both of those adjustments sound correct to you?

Yes, you'll need to do an interactive rebase and modify the commit messages.
This PR has a lot of commits and most of them should be squashed.
We usually make sure that every commit successfully builds on its own and we aim for commits that each serve a defined purpose (for example, one commit for doc updates, one commit for tests, one commit for a specific unitary code change).
Commits that modify a piece of code introduced by the same PR should usually be squashed.

> Lastly, this is a small question, do you think the testing of the config when restoring a snapshot should go in the test_api.py test file or the test_snapshot_[basic/advanced].py files? I'm leaning towards the test_api.py but I could see it either way.

Let's have it in `test_api.py`, since that's where `test_get_full_config` is. You could actually edit that test to perform the validation both pre and post snapshot (to reuse most of the code)

summerlinryan:
@alindima I've addressed all of the recent feedback regarding the commits, unit & integration tests, and CHANGELOG entry. Let me know how it all looks when you get the chance, thanks!

src/vmm/src/device_manager/persist.rs
```diff
@@ -645,16 +646,74 @@ mod tests {
...
+    "ht_enabled": false,
+    "track_dirty_pages": false
+  }},
+  "metrics": null,
+  "mmds-config": null,
+  "network-interfaces": [
+    {{
+      "iface_id": "netif",
+      "host_dev_name": "hostname",
+      "guest_mac": "00:00:00:00:00:00",
```
summerlinryan, line 693: @alindima  I'm unsure what to think about this guest_mac being all 0s. It feels like it should actually be null because the guest_mac is None on line 604 but then I saw some code in Net::new_with_tap that seemed to use a ConfigSpace::default() which has all 0s for the mac address so then I figured maybe the all 0s is right.

alindima:
> @alindima I've addressed all of the recent feedback regarding the commits, unit & integration tests, and CHANGELOG entry. Let me know how it all looks when you get the chance, thanks!

great! I'll have a look probably by early next week

summerlinryan:
Although I thought I verified that all tests passed locally before pushing, I must've made a mistake because I see that the python/markdown linting are failing. I'll fix those small issues when I get home from work tonight.

src/vmm/src/device_manager/persist.rs
alindima: Nice catch! I've done some digging to fond out why that is.
This is all zeros because indeed the guest mac in the config space is initialised with zeros. When the Net device is then saved to snapshot, we don't specifically save the guest mac as configured by the user and we then initialise it from the config space value, which in this case is all zeros.

I think this is a bug so we should open a new issue for it after we merge the PR. It is confusing to have a zeroed MAC when it was not really configured.

src/vmm/src/device_manager/persist.rs
```diff
@@ -611,14 +646,74 @@ mod tests {
...
             event_manager: &mut event_manager,
+            for_each_restored_device: VmResources::update_from_restored_device,
+            vm_resources,
         };
         let restored_dev_manager =
             MMIODeviceManager::restore(restore_args, &device_states).unwrap();
 
+        let expected_vm_resources = format!(
+            r#"{{
+  "balloon": {{
```
alindima, line 662: the indentation is a bit off

src/vmm/src/resources.rs
```diff
@@ -72,7 +73,7 @@ pub struct VmmConfig {
     #[serde(rename = "drives")]
     block_devices: Vec<BlockDeviceConfig>,
     #[serde(rename = "boot-source")]
-    boot_source: BootSourceConfig,
+    boot_source: Option<BootSourceConfig>,
```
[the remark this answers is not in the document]
summerlinryan: I'm having a hard time finding it now but there was some documentation edit suggestion in the past that used the word "empty" when referring to a couple of properties (one of which was `boot_source`). That made me think that in the case of restored snapshots, we'd want the `boot_source` to appear as null (since there isn't actually a boot source in that case). If we'd rather keep it as the default `BootSourceConfig` that's no problem, just let me know and I'll undo this change of making it `Optional`.

src/vmm/src/resources.rs
```diff
@@ -94,7 +95,7 @@ pub struct VmResources {
     /// The vCpu and memory configuration for this microVM.
     vm_config: VmConfig,
     /// The boot configuration for this microVM.
-    boot_config: Option<BootConfig>,
+    pub boot_config: Option<BootConfig>,
```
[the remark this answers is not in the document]
summerlinryan: No, it actually doesn't. This was from a previous failed approach of fixing this bug and accidentally got left in. Sorry for leaving it in and nice job spotting it.

tests/integration_tests/functional/test_api.py
```diff
@@ -1407,175 +1407,18 @@ def test_get_full_config(test_microvm_with_api):
...
-#
-#     # Validate full vm configuration post-boot as well.
-#     response = test_microvm.full_cfg.get()
-#     assert test_microvm.api_session.is_status_ok(response.status_code)
-#     assert response.json() == expected_cfg
+    test_microvm.pause_to_snapshot(
+        mem_file_path='memfile',
+        snapshot_path='statefile',
+        diff=False,
+        version='0.24.0'
```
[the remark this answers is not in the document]
summerlinryan: Honestly, I can't recall why I explicitly set the version to 0.24.0. I must've just been confused. I'll update it to to be unspecified (using latest).

src/vmm/src/device_manager/persist.rs
summerlinryan: If I change the indentation here, the test will fail since it's a raw string and the indentation of the raw string JSON will no longer match the pretty printed JSON from serde serialization. Is there another way to change the indentation formatting but still have the assertion pass?

src/vmm/src/vmm_config/vsock.rs
```diff
@@ -192,4 +192,20 @@ pub(crate) mod tests {
...
+        tmp_sock_file.remove().unwrap();
+        let vsock = Vsock::new(
+            0,
+            VsockUnixBackend::new(1, tmp_sock_file.as_path().to_str().unwrap().to_string())
+                .unwrap(),
+        )
+        .unwrap();
+
+        vsock_builder.set_device(Arc::new(Mutex::new(vsock)));
+        assert!(vsock_builder.inner.is_some());
```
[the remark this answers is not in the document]
summerlinryan: I added an assertion for the `uds_path` of the `vsock_builder`. I have assertions for the ids of all of the `test_add_device` tests that involve builders with collections. Is there something else I can assert to confirm the identity of the device for the builders with collections (e.g. `BlockBuilder`)? The only other test that I can think of that you might be referring to would be the `BalloonBuilder.test_set_device` but I'm not sure what I can check in terms of identity on that because the `Balloon.id()` always returns `BALLOON_DEV_ID` so checking the `id` of the device wouldn't be very useful in that case.

summerlinryan:
I made most of the recent feedback changes locally and have rebased on to main so I went ahead and marked some of your feedback comments as resolved. I'm going to wait to push the changes since I might be making a change to the Optional `boot_source` depending on your feedback to my other comment.

tests/integration_tests/functional/test_api.py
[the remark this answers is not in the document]
alindima: sounds good, let's remove it

alindima:
> I made most of the recent feedback changes locally and have rebased on to main so I went ahead and marked some of your feedback comments as resolved. I'm going to wait to push the changes since I might be making a change to the Optional `boot_source` depending on your feedback to my other comment.

Great, thanks! I've responded to the comments.
One other outstanding item is the integration test. View this comment: https://github.com/firecracker-microvm/firecracker/pull/2806#discussion_r799571125

tests/integration_tests/functional/test_api.py
```diff
@@ -1342,6 +1342,142 @@ def test_api_balloon(test_microvm_with_api):
...
+    microvm_builder = MicrovmBuilder(bin_cloner_path)
+
+    root_disk = test_microvm.vm.disks[0]
+    ssh_key = test_microvm.ssh_key
+
+    # Create a snapshot builder from a microvm.
+    snapshot_builder = SnapshotBuilder(test_microvm)
+
+    # Create base snapshot.
+    snapshot = snapshot_builder.create([root_disk.local_path()],
```
summerlinryan, line 1457: @alindima I'm a bit confused on this final test. I've been trying to find tests to reference for creating a microvm from a snapshot (e.g. the tests in `test_snapshot_basic.py`) and they mostly seem to use context to get the `ssh_key`, `root_disk`, and `snapshot_type` but when I try to pass `context` into the new test I made, I get an error that I can't get past. This approach that I'm trying is using the `MicrovmBuilder` and `SnapshotBuilder` but I also tried another approach where I was trying to use `SnapshotHelper.load` but that failed as well. Any tips on which approach would be the best? Thanks!

alindima: This is a little tricky to get right. Try applying this patch:
```
diff --git a/tests/integration_tests/functional/test_api.py b/tests/integration_tests/functional/test_api.py
index be039c84..a0b66a84 100644
--- a/tests/integration_tests/functional/test_api.py
+++ b/tests/integration_tests/functional/test_api.py
@@ -19,7 +19,7 @@ import host_tools.network as net_tools
 from conftest import _test_images_s3_bucket
 
 from framework.utils import is_io_uring_supported
-from framework.artifacts import ArtifactCollection, SnapshotType
+from framework.artifacts import ArtifactCollection, SnapshotType, NetIfaceConfig, DEFAULT_DEV_NAME, DEFAULT_TAP_NAME
 from framework.builder import MicrovmBuilder, SnapshotBuilder
 
 MEM_LIMIT = 1000000000
@@ -1342,19 +1342,24 @@ def test_api_balloon(test_microvm_with_api):
 
 
-def test_get_full_config_after_restoring_snapshot(test_microvm_with_api, bin_cloner_path):
+def test_get_full_config_after_restoring_snapshot(bin_cloner_path):
     """
     Test the reported configuration of a microVM configured with all resources.
 
     @type: functional
     """
-    test_microvm = test_microvm_with_api
+    microvm_builder = MicrovmBuilder(bin_cloner_path)
+    test_microvm = microvm_builder
+
+    net_iface = NetIfaceConfig()
+    vm_instance = microvm_builder.build_vm_nano(net_ifaces=[net_iface])
+    test_microvm = vm_instance.vm
+    root_disk = vm_instance.disks[0]
+    ssh_key = vm_instance.ssh_key
 
     setup_cfg = {}
 
-    test_microvm.spawn()
     # Basic config also implies a root block device.
-    test_microvm.basic_config(rootfs_io_engine="Sync")
     setup_cfg['machine-config'] = {
         'vcpu_count': 2,
         'mem_size_mib': 256,
@@ -1362,13 +1367,10 @@ def test_get_full_config_after_restoring_snapshot(test_microvm_with_api, bin_clo
         'track_dirty_pages': False,
         'cpu_template': 'C3',
     }
-    setup_cfg['boot-source'] = {
-        'kernel_image_path': '/vmlinux.bin',
-        'initrd_path': None
-    }
+    test_microvm.machine_cfg.patch(**setup_cfg['machine-config'])
     setup_cfg['drives'] = [{
         'drive_id': 'rootfs',
-        'path_on_host': '/bionic.rootfs.ext4',
+        'path_on_host': f"/{os.path.basename(root_disk.local_path())}",
         'is_root_device': True,
         'partuuid': None,
         'is_read_only': False,
@@ -1397,11 +1399,20 @@ def test_get_full_config_after_restoring_snapshot(test_microvm_with_api, bin_clo
         'uds_path': 'vsock.sock'
     }
 
-    # Add a net device.
-    iface_id = '1'
-    tapname = test_microvm.id[:8] + 'tap' + iface_id
-    tap1 = net_tools.Tap(tapname, test_microvm.jailer.netns)
-    guest_mac = '06:00:00:00:00:01'
+    setup_cfg['logger'] = None
+    setup_cfg['metrics'] = None
+    setup_cfg['mmds-config'] = {
+        'version': "V1",
+        'network_interfaces': [DEFAULT_DEV_NAME]
+    }
+
+    response = test_microvm.mmds.put_config(json=setup_cfg['mmds-config'])
+
+    # Start the microvm.
+    test_microvm.start()
+
+    # Add a tx rate limiter to the net device.
     tx_rl = {
         'bandwidth': {
             'size': 1000000,
@@ -1410,46 +1421,19 @@ def test_get_full_config_after_restoring_snapshot(test_microvm_with_api, bin_clo
         },
         'ops': None
     }
-    response = test_microvm.network.put(
-        iface_id=iface_id,
-        guest_mac=guest_mac,
-        host_dev_name=tap1.name,
+    response = test_microvm.network.patch(
+        iface_id=DEFAULT_DEV_NAME,
         tx_rate_limiter=tx_rl
     )
     setup_cfg['network-interfaces'] = [{
-        'iface_id': iface_id,
-        'host_dev_name': tap1.name,
-        'guest_mac': '06:00:00:00:00:01',
+        'guest_mac': net_tools.mac_from_ip(net_iface.guest_ip),
+        'iface_id': DEFAULT_DEV_NAME,
+        'host_dev_name': DEFAULT_TAP_NAME,
         'rx_rate_limiter': None,
         'tx_rate_limiter': tx_rl
     }]
 
-    # Update MMDS config.
-    mmds_config = {
-        'version': 'V2',
-        'ipv4_address': '169.254.169.250',
-        'network_interfaces': ['1']
-    }
-    response = test_microvm.mmds.put_config(json=mmds_config)
-    assert test_microvm.api_session.is_status_no_content(response.status_code)
-
-    setup_cfg['logger'] = None
-    setup_cfg['metrics'] = None
-    setup_cfg['mmds-config'] = {
-        'version': 'V2',
-        'ipv4_address': '169.254.169.250',
-        'network_interfaces': ['1']
-    }
-
-    # Start the microvm.
-    test_microvm.start()
-
-    microvm_builder = MicrovmBuilder(bin_cloner_path)
-
-    root_disk = test_microvm.vm.disks[0]
-    ssh_key = test_microvm.ssh_key
-
     # Create a snapshot builder from a microvm.
     snapshot_builder = SnapshotBuilder(test_microvm)
 
@@ -1465,11 +1449,12 @@ def test_get_full_config_after_restoring_snapshot(test_microvm_with_api, bin_clo
     # We expect boot-source, machine-config.smt, and machine-config.cpu_template to all
     # be empty/default after restoring from a snapshot.
     expected_cfg['boot-source'] = {
-        'kernel_image_path': None,
+        'kernel_image_path': '',
         'initrd_path': None
     }
     expected_cfg['machine-config']['smt'] = False
-    expected_cfg['machine-config']['cpu_template'] = None
+    expected_cfg['machine-config'].pop('cpu_template')
+    expected_cfg['mmds-config'] = None
 
     # Validate full vm configuration post-restore.
     response = microvm.full_cfg.get()
```

alindima: Notice that the mmds_config is still reported as empty after snapshotting, you need to also hook that in when re-adding the network device to VmResources.
There is a known limitation that MMDS version 2 is lost after snapshot, so use the default value in that case.

summerlinryan: Unfortunately, some stuff came up in my personal life. I'll look back into this soon (should be within a few days).

alindima: Sure, no worries!

summerlinryan: Sorry I'm struggling so much on this final test. I'm not sure what you mean with the following:

> you need to also hook that in when re-adding the network device to VmResources.

> There is a known limitation that MMDS version 2 is lost after snapshot, so use the default value in that case.

It seems that the MMDS config always shows as null (which is maybe what you're saying)? If it should be null after snapshot, then I'm thinking what I have currently, is right.

Also, any idea what I can use to test the `cpu_template` of the `machine-config`? I had it set to `C3` because that's what some other tests were using but I was getting an `InvalidVendor` that was caused by the C3 `cpu_template` so I ended up getting rid of setting the `cpu_template` (at least temporarily). 

As a side not, I'll merge the 2 final tests commits together this is worked through. I just wanted to push what I had for review.

alindima: > Sorry I'm struggling so much on this final test.

No worries. Don't get discouraged, you've done the more difficult part 😄 .

> It seems that the MMDS config always shows as null (which is maybe what you're saying)?

Yes, that what I was referring to. It should not be empty after restoring from snapshot. The `NetState` struct contains the optional `MmdsNetworkStackState` which can be used to extract the Mmds config. This should be done in the `update_from_restored_device` function, for each of the net devices.

> There is a known limitation that MMDS version 2 is lost after snapshot, so use the default value in that case.

What I meant by this is that the used MMDS version is not yet saved in the snapshot (so it should always be V1, the default).

alindima: > Also, any idea what I can use to test the cpu_template of the machine-config? I had it set to C3 because that's what some other tests were using but I was getting an InvalidVendor that was caused by the C3 cpu_template so I ended up getting rid of setting the cpu_template (at least temporarily).

Hmm, are you using an AMD host by any chance?

alindima: You should use the `utils.get_cpu_vendor()` in the test, and only set that CPU template for intel hosts

summerlinryan: > Hmm, are you using an AMD host by any chance?

Yes I am actually. I'll use your recommendation of using `utils.get_cpu_vendor` to conditionally set the `cpu_template`. Thanks!

After rereading a bunch of MMDS documentation, reading through the code for a while, and experimenting with the API, I feel like I'm understanding the MMDS change more than I was before. A few implementation questions I still have though are:

1. You mentioned using `MmdsNetworkStackState` in the `NetState`. I noticed that property is currently not public; are we comfortable making it public? On a related note though, the `update_from_restored_device` currently doesn't have access to the `NetState` but it does have access to the `Net` device which seems to have the MMDS network stack stored on it so I'm thinking about using that instead of the `NetState`. Does that sound reasonable or is there a reason I should still make use `NetState`? If it sounds reasonable to use `Net` instead of `NetState`, then ignore my first question about making the `NetState` MMDS state public. 
2. There is only one `ipv4_address` that exists for the `MmdsConfig`, however, each `Net` device has a `ipv4_addr` that seems to refer to the MMDS server. Presumably, the `ipv4_addr` for all net devices will always be the same (since there is only a single MMDS server) but correct me if that presumption is wrong. What I'm thinking about doing is setting the `MmdsConfig` ipv4 address to the default (`169.254.169.254`) in `build_microvm_from_snapshot` and then updating it to the value of the `ipv4_addr` of the net device in `update_from_restored_device`. Does that sound reasonable? (see the code snippet below for an example of how I might go about the update in `update_from_restored_device`)

```rs
 SharedDeviceType::SharedNetwork(network) => {
    if let Some(mmds_ns) = &network.lock().unwrap().mmds_ns {
        let mmds_config = self.mmds_config.as_mut().unwrap();
        mmds_config.ipv4_address = Some(mmds_ns.ipv4_addr);
        mmds_config
            .network_interfaces
            .push(network.lock().expect("Poisoned lock").id().to_owned());
    }

    self.net_builder.add_device(network);
}
```

alindima: > You mentioned using MmdsNetworkStackState in the NetState. I noticed that property is currently not public; are we comfortable making it public? On a related note though, the update_from_restored_device currently doesn't have access to the NetState but it does have access to the Net device which seems to have the MMDS network stack stored on it so I'm thinking about using that instead of the NetState. Does that sound reasonable or is there a reason I should still make use NetState? If it sounds reasonable to use Net instead of NetState, then ignore my first question about making the NetState MMDS state public.

This was a mistake from me, I was indeed referring to the `Net` struct.

> There is only one ipv4_address that exists for the MmdsConfig, however, each Net device has a ipv4_addr that seems to refer to the MMDS server. Presumably, the ipv4_addr for all net devices will always be the same (since there is only a single MMDS server) but correct me if that presumption is wrong.

This is true 👍🏻 

> What I'm thinking about doing is setting the MmdsConfig ipv4 address to the default (169.254.169.254) in build_microvm_from_snapshot and then updating it to the value of the ipv4_addr of the net device in update_from_restored_device. Does that sound reasonable? (see the code snippet below for an example of how I might go about the update in update_from_restored_device)

No need to set the ipv4 address to default before `update_from_restored_device` (since the `ipv4_addr` field on the `MmdsNetworkStack` is not optional; if there is no network interface with MMDS configured, then there is no MMDS configuration at all). The code snippet you pasted looks good to me overall.

Note that `self.mmds_config.as_mut().unwrap()` may panic as it's written in the snippet.

src/vmm/src/resources.rs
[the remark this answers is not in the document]
alindima: This will be solved when implementing the proposal from this comment: https://github.com/firecracker-microvm/firecracker/pull/2806#discussion_r812532289

summerlinryan:
@alindima I think I've got the MMDS config working as expected now and I've updated the tests. I've also updated test to conditionally set the cpu template only for Intel hosts. Unfortunately, there were 2 fields I had to change from `pub(crate)` to `pub` but it seemed necessary in order to restore the MMDS config. Let me know what you think about everything, thanks!

src/vmm/src/resources.rs
```diff
@@ -171,6 +172,44 @@ impl VmResources {
...
+        match device {
+            SharedDeviceType::SharedBlock(block) => {
+                self.block.add_device(block);
+            }
+
+            SharedDeviceType::SharedNetwork(network) => {
+                if let Ok(net) = network.lock() {
+                    if let Some(mmds_ns) = &net.mmds_ns {
+                        if let Some(mut_mmds_config) = self.mmds_config.as_mut() {
+                            mut_mmds_config.ipv4_address = Some(mmds_ns.ipv4_addr);
```
alindima, line 187: we may skip this since they should all have the same ipv4 adddress

src/vmm/src/resources.rs
```diff
@@ -171,6 +172,44 @@ impl VmResources {
...
+    /// Updates the resources from a restored device (used for configuring resources when
+    /// restoring from a snapshot).
+    pub fn update_from_restored_device(&mut self, device: SharedDeviceType) {
+        match device {
+            SharedDeviceType::SharedBlock(block) => {
+                self.block.add_device(block);
+            }
+
+            SharedDeviceType::SharedNetwork(network) => {
+                if let Ok(net) = network.lock() {
```
alindima, line 184: let's do an `expect("Poisoned lock")` here, since this is an irrecoverable error

CHANGELOG.md
```diff
@@ -99,6 +99,10 @@
 
 ### Fixed
 
+- GET `/vm/config` was returning a default config object after restoring from a snapshot.
+  It now correctly returns the config of the original microVM, except for boot_config and the
+  cpu_template and smt fields of the machine config,
+  which are currently lost.
```
alindima, line 105:
```suggestion
  cpu_template and smt fields of the machine config, which are currently lost.
```

src/api_server/swagger/firecracker.yaml
```diff
@@ -612,7 +612,8 @@ paths:
     get:
       summary: Gets the full VM configuration.
       description:
-        Gets configuration for all VM resources.
+        Gets configuration for all VM resources. If the VM is restored from a snapshot, the boot-source,
+        machine-config.ht_enabled and machine-config.cpu_template will be empty.
```
alindima, line 616:
```suggestion
        machine-config.smt and machine-config.cpu_template will be empty.
```

alindima:
> @alindima I think I've got the MMDS config working as expected now and I've updated the tests. I've also updated test to conditionally set the cpu template only for Intel hosts. Unfortunately, there were 2 fields I had to change from `pub(crate)` to `pub` but it seemed necessary in order to restore the MMDS config. Let me know what you think about everything, thanks!

Looks good @summerlinryan . Please have a look at the failing CI tests and try to fix them. Let me know if you need help

dianpopa:
I think we should do a full fix for this issue and implement the right solution which would be to save all the info related to the vm config in a separate structure and store that in the snapshot. We would need to be careful about not breaking snapshot compatibility by checking whether the vm config info is available in the snapshot. Sorry if this was discussed eariler, there are lots of comments and I might have missed some of them. 

WDYT about the solution proposed?

alindima:
> I think we should do a full fix for this issue and implement the right solution which would be to save all the info related to the vm config in a separate structure and store that in the snapshot. We would need to be careful about not breaking snapshot compatibility by checking whether the vm config info is available in the snapshot. Sorry if this was discussed eariler, there are lots of comments and I might have missed some of them. 
> 
> WDYT about the solution proposed?

We would only need to do such things for smt and cpu template (and potentially boot source, but I don't really think it adds any value)
We decided to move on with this PR and skip those for now.

We can definitely add those as a followup

dianpopa:
> > I think we should do a full fix for this issue and implement the right solution which would be to save all the info related to the vm config in a separate structure and store that in the snapshot. We would need to be careful about not breaking snapshot compatibility by checking whether the vm config info is available in the snapshot. Sorry if this was discussed eariler, there are lots of comments and I might have missed some of them.
> > WDYT about the solution proposed?
> 
> We would only need to do such things for smt and cpu template (and potentially boot source, but I don't really think it adds any value) We decided to move on with this PR and skip those for now.
> 
> We can definitely add those as a followup

It would be great to add them as a follow up but is there a way to obtain those info  after resuming from snapshot (smt maybe, but for cpu template I do not see a way...)? This is my main concern actually, that we would be left with a partial solution. 
Sorry for missing the decision discussion.

alindima:
> > > I think we should do a full fix for this issue and implement the right solution which would be to save all the info related to the vm config in a separate structure and store that in the snapshot. We would need to be careful about not breaking snapshot compatibility by checking whether the vm config info is available in the snapshot. Sorry if this was discussed eariler, there are lots of comments and I might have missed some of them.
> > > WDYT about the solution proposed?
> > 
> > 
> > We would only need to do such things for smt and cpu template (and potentially boot source, but I don't really think it adds any value) We decided to move on with this PR and skip those for now.
> > We can definitely add those as a followup
> 
> It would be great to add them as a follow up but is there a way to obtain those info after resuming from snapshot (smt maybe, but for cpu template I do not see a way...)? This is my main concern actually, that we would be left with a partial solution. Sorry for missing the decision discussion.

Here is where we discussed this topic: https://github.com/firecracker-microvm/firecracker/pull/2806#discussion_r769688210

In the current form of this PR, this would indeed be a partial solution (which fixes more than 90% let's say of the issue).
The followup would take the smt information and the cpu template likely from the snapshot state (which currently does not persist them).
Do you have a concern with merging such a partial solution and then coming back to it as a follow-up? Just as a clarification, the follow-up would build on the solution from this PR (the solutions are conflicting here AFAICT).

I believe that a partial solution is better than trying to fix everything in one go (given that it's a pretty complicated one as well)

dianpopa:
> > > > I think we should do a full fix for this issue and implement the right solution which would be to save all the info related to the vm config in a separate structure and store that in the snapshot. We would need to be careful about not breaking snapshot compatibility by checking whether the vm config info is available in the snapshot. Sorry if this was discussed eariler, there are lots of comments and I might have missed some of them.
> > > > WDYT about the solution proposed?
> > > 
> > > 
> > > We would only need to do such things for smt and cpu template (and potentially boot source, but I don't really think it adds any value) We decided to move on with this PR and skip those for now.
> > > We can definitely add those as a followup
> > 
> > 
> > It would be great to add them as a follow up but is there a way to obtain those info after resuming from snapshot (smt maybe, but for cpu template I do not see a way...)? This is my main concern actually, that we would be left with a partial solution. Sorry for missing the decision discussion.
> 
> Here is where we discussed this topic: [#2806 (comment)](https://github.com/firecracker-microvm/firecracker/pull/2806#discussion_r769688210)
> 
> In the current form of this PR, this would indeed be a partial solution (which fixes more than 90% let's say of the issue). The followup would take the smt information and the cpu template likely from the snapshot state (which currently does not persist them). Do you have a concern with merging such a partial solution and then coming back to it as a follow-up? Just as a clarification, the follow-up would build on the solution from this PR (the solutions are not conflicting here AFAICT).
> 
> I believe that a partial solution is better than trying to fix everything in one go (given that it's a pretty complicated one as well)

Alin and me synced offline and decided to go ahead with the solution proposed in this PR. @summerlinryan let us know if you need help with the feedback.

summerlinryan:
@dianpopa When you say "the solution proposed in this PR", are you referring to the work that's already been done or the new approach you suggested including the saving of the full config? If you're referring to the work that's already been done, then I should be working on getting the tests passing tonight so within the next day or so I should have the final commit pushed. 👍 Thanks for the feedback!

dianpopa:
> @dianpopa When you say "the solution proposed in this PR", are you referring to the work that's already been done or the new approach you suggested including the saving of the full config? If you're referring to the work that's already been done, then I should be working on getting the tests passing tonight so within the next day or so I should have the final commit pushed. 👍 Thanks for the feedback!

Oh sorry for the misunderstanding. I meant that we should go ahead with the work that's already been done.

summerlinryan:
> Please rebase over latest main

Sorry, I thought I had. I must've either forgot to or made a mistake while doing the rebase.

Nice catch on `setup_cfg['machine-config']['cpu_template']: 'C3'`! That was definitely a problem that I was completely overlooking. 

I updated the coverage and got all other tests passing as far I can tell locally. It seems that the `test_python` test doesn't work for me locally. I end up getting a ton of 

```
Exception ignored when trying to write to the signal wakeup fd:
BlockingIOError: [Errno 11] Resource temporarily unavailable
```

errors when running it locally and it seems to timeout. I'm assuming that's happening from how many processes it's spawning. 

Also, I wasn't sure how to test the `aarm64` `platform.machine()` change locally since I'm running on a `x86_64` host but it seemed like a simple change so hopefully I'll wake up in the morning to a bunch of passing tests! :crossed_fingers:

alindima:
Most of the tests are passing now, great! 👍🏻 

I see that on ARM there's a test that fails: `test_api_machine_config`.

This is because this test was relying on the logic that on ARM `get_cpu_vendor()` used to return Intel (this logic was wrong).

AFAICT, this change would fix things (in test_api.py):

`if utils.get_cpu_vendor() != utils.CpuVendor.INTEL:` => `if utils.get_cpu_vendor() == utils.CpuVendor.AMD:`

I believe you can also mark this PR as ready for review now (it's still a draft)

alindima:
Tests are all green! 😄 
@summerlinryan there are a couple of comments from me left unaddressed. @dianpopa will also review the code.
Looking good!

src/vmm/src/device_manager/persist.rs
summerlinryan: @alindima You want to wait until after this PR is merged before we create the new issue related to the zeroed out mac address, right? Just going through and making sure I've addressed all outstanding comments.

src/vmm/src/resources.rs
summerlinryan: I changed it to use the `expect("Poisoned lock")` but doing that introduced a borrow checker issue since `network.lock().expect()` results in a borrow but then `network` gets moved with `self.net_builder.add_device(network);`. A workaround I added was a putting the `if` in its own scope but I feel like there must be a better , more idiomatic Rust way? I tried cloning the `network` device but that also had a separate borrow checker issue.

src/vmm/src/device_manager/persist.rs
alindima: Let's wait until we merge this PR and leave a comment on the same issue with the unaddressed items, like this MAC issue and the fact that the cpu_template, boot source and smt are currently not persisted.

src/vmm/src/resources.rs
alindima: you may call `drop(net)` before calling `self.net_builder.add_device(network);`, but I think the current approach is fine

alindima (approved):
Looks great to me, thanks for the hard work! 😄

dianpopa (approved):

georgepisaltu:
Awesome work @summerlinryan 🎉

PR #2819, opened 2021-12-13
seccomp: fix large diff snapshot failure

alindima:
# Reason for This PR

https://github.com/firecracker-microvm/firecracker/discussions/2811

## Description of Changes

- seccomp: allow mmap(MAP_PRIVATE|MAP_ANONYMOUS) on the vmm thread.
- add regression test for it
- fix error from commit https://github.com/firecracker-microvm/firecracker/pull/2701/commits/a3f5d47b063599e2eb73a116e537988de53b2929. The mmap call on aarch64 should have been allowed on the api thread, but was instead added to the vmm

## License Acceptance

## PR Checklist

### DIFF
--- a/resources/seccomp/aarch64-unknown-linux-musl.json
+++ b/resources/seccomp/aarch64-unknown-linux-musl.json
@@ -191,19 +191,6 @@
             },
             {
                 "syscall": "mmap",
-                "comment": "Used for large buffers sent to api_server",
-                "args": [
-                    {
-                        "index": 3,
-                        "type": "dword",
-                        "op": "eq",
-                        "val": 34,
-                        "comment": " libc::MAP_ANONYMOUS | libc::MAP_PRIVATE"
-                    }
-                ]
-            },
-            {
-                "syscall": "mmap",
                 "comment": "Used for reading the timezone in LocalTime::now()",
                 "args": [
                     {
@@ -534,6 +521,19 @@
                         "op": "eq",
                         "val": 1,
                         "comment": "libc::MAP_SHARED"
+                    }
+                ]
+            },
+            {
+                "syscall": "mmap",
+                "comment": "Used for large buffers sent to api_server",
+                "args": [
+                    {
+                        "index": 3,
+                        "type": "dword",
+                        "op": "eq",
+                        "val": 34,
+                        "comment": " libc::MAP_ANONYMOUS | libc::MAP_PRIVATE"
                     }
                 ]
             },
--- a/resources/seccomp/aarch64-unknown-linux-musl.json
+++ b/resources/seccomp/aarch64-unknown-linux-musl.json
@@ -203,6 +203,19 @@
                 ]
             },
             {
+                "syscall": "mmap",
+                "comment": "Used by rust's stdlib, particularly when creating a diff snapshot of a VM with ~16 GB of memory",
+                "args": [
+                    {
+                        "index": 3,
+                        "type": "dword",
+                        "op": "eq",
+                        "val": 34,
+                        "comment": "libc::MAP_ANONYMOUS | libc::MAP_PRIVATE"
+                    }
+                ]
+            },
+            {
                 "syscall": "rt_sigaction",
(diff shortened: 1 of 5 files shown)


luminitavoicu (approved):
Looks good to me! Should we add a changelog entry for this as well?

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -601,3 +601,32 @@ def test_negative_snapshot_create(bin_cloner_path):
     assert not os.path.exists('memfile')
 
     vm.kill()
+
+
+def test_mmap_seccomp_regression(test_microvm_with_api):
```
georgepisaltu, line 606: I don't think this test will age well in this form. I'd like us to integrate this test case into an already existing test, maybe functional snapshot tests, and we can mention there what the regression that we're testing for was. I think it will make more sense long term.

resources/seccomp/aarch64-unknown-linux-musl.json
```diff
@@ -537,6 +524,19 @@
                     }
                 ]
             },
+            {
```
georgepisaltu, line 540: Very good catch. However, I'm worried that we had to rely on a person finding this rather than our test. Did we just not cover this in any kind of test on `aarch64`? Maybe we can open an issue about this, might even be a "Good First Contribution" 😄 .

georgepisaltu (commented):
Overall looks great! Just a couple of comments.

alindima:
> Looks good to me! Should we add a changelog entry for this as well?

Personally, I don't think it's worth adding a changelog entry

resources/seccomp/aarch64-unknown-linux-musl.json
alindima: We have a regression test for this in `test_api.py::test_map_private_seccomp_regression`, but for some reason this mmap call must not have been triggered on ARM.
It's very likely a platform-specific implementation detail of either the Global Allocator or the rust stdlib.
I'll try to see if sending a larger payload than the current 1 mib would trigger the regression on ARM

tests/integration_tests/functional/test_snapshot_basic.py
alindima: I personally like the separation between regression tests and other types of tests (functional, negative, etc.). They are also easier to spot given the `@type: regression` adnotation.

Why do you think having this separate would be better in the longer term?

resources/seccomp/aarch64-unknown-linux-musl.json
```diff
@@ -191,27 +191,27 @@
...
+                "comment": "Used for reading the timezone in LocalTime::now()",
                 "args": [
                     {
                         "index": 3,
                         "type": "dword",
                         "op": "eq",
-                        "val": 34,
-                        "comment": " libc::MAP_ANONYMOUS | libc::MAP_PRIVATE"
+                        "val": 1,
+                        "comment": "libc::MAP_SHARED"
```
dianpopa, line 201: is the comment correct? MAP_SHARED being needed for reading the timezone?

alindima: yes it is. I've checked with strace and in the musl implementation of `localtime_r`
It first checks for the TZ environment variable. If that's not available and if the `/etc/localtime` file is present, it opens the file and mmaps it.

resources/seccomp/aarch64-unknown-linux-musl.json
alindima: Turns out that the regression test in `test_api.py` is not failing even if I remove the mmap from the allowlist.
Looking with strace, there are no such mmap calls issued.
I think it's a bug in the test, I'm looking into it

tests/integration_tests/functional/test_snapshot_basic.py
alindima: As we chatted offline, we renamed the regression test to `test_create_large_diff_snapshot`

luminitavoicu (approved):

georgepisaltu (approved):

Issue #2802, opened 2021-11-30
Stability issues when resuming from diff snapshots


vadimberezniker:
# Describe the bug

We have been experimenting with snapshot support in Firecracker, but have been running into some stability issues when booting from diff snapshots. We have not experienced any issues when using only full snapshots.

The "stability" issues appear to be symptoms of memory corruption, for example unexpected kernel panics, weird application segfaults and otherwise strange application behavior. 

The machine I'm testing on:
`Linux aorus 5.11.0-40-generic #44~20.04.2-Ubuntu SMP Tue Oct 26 18:07:44 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`
I've also tested this on different Intel-based machine. 

## To Reproduce

1. Start a VM configured with a vsock device, networking and a block device (not sure if any of these datails matter, but including them for completeness)
2. Run the following commands:
```
git clone https://github.com/buildbuddy-io/buildbuddy.git
cd buildbuddy
bazelisk build //flag
```
note: the bazelisk command will ultimately error out because there's no //flag target but that's expected. 
2. Pause VM
3. Take a full snapshot
4. Remove VM
5. Create new VM from full snapshot
6. Run `bazelisk build //flag`
7. Pause VM
8. Take diff snapshot
9. Merge diff snapshot into full snapshot from step (3). I've tried using both custom code and a `dd bs=4096 if=diff of=full conv=sparse,notrunc` command with same results. 
10. Remove VM
11. Create new VM from full memory snapshot made in step (9) and state snapshot made in step (8)
12. Run `bazelisk build //flag`

## Expected behavior

Output of step 12 should be the same as output of step 6 and should complete relatively quickly. 
What we actually see here is strange behaviors such as kernel panics, strange segfaults and otherwise unexpected guest behavior. 
Everything is fine if we replace step 8-9 with another full memory snapshot instead of a diff snapshot.

It's worth noting I have also done a test after step 7 where i take a diff snapshot followed by full snapshot to compare the full snapshot and the generated merged snapshot and verified that they are identical. 

## Environment

`Firecracker v0.25.1`
`Host kernel: 5.11.0-40-generic, guest kernel: 5.4.0+`
`Arch: x86_64`
`CPU: AMD Ryzen 9 3950X 16-Core Processor`
(Note I've also tested on an Intel-based machine.)

FS information:
[...]

PR #2820, opened 2021-12-14
[io_uring] Test PR

serban300:
Test PR to investigate `test_5_inc_snapshots` failures


sandreim:
Is this only happening with `io_uring` ? Asking because of this other similar issue: https://github.com/firecracker-microvm/firecracker/issues/2802

serban300:
> @serban300 Is this only happening with `io_uring` ? Asking because of this other similar issue: #2802

Yes, this failure only occurs with `io_uring`.

alindima:
I've seen this same failure without io_uring as well

sandreim:
I would look at any dirty bitmap changes or vm-memory changes that might have been merged.

PR #2846, opened 2022-01-11
docs: kernel policy support

dianpopa:
Added the policy behind validating host/guest
kernel versions with Firecracker.

# Reason for This PR

`[Author TODO: add issue #.]`
`[Open the PR after the related issue has a clear conclusion.]`
`[If there is no issue which states the need for this PR, create one first.]`
Fixes #

## Description of Changes

## License Acceptance

## PR Checklist


alindima:
I think we should we have a link to this document somewhere in the README, WDYT?
I'd also add a Changelog update to specifically state that we added a support policy and a link to it.

raduiliescu:
I wouldn't add this in the changelog since is not a per release thing. In the changelog I would add something like add support and verification for kernel 5.10.
Beside the readme, I would review the docs and add links to this in the places where kernel and kernel operations are mentioned, if it makes sense.

docs/kernel-policy.md
```diff
@@ -0,0 +1,85 @@
+# Firecracker's Kernel Support Policy
+
+Being a virtual machine monitor, Firecracker represents a component in a
+larger stack, one in which it is tightly coupled with the guest and host
+kernels on which it is run.
+
+As of right now, Firecracker is being validated
```
raduiliescu, line 7: I would say "Firecracker code from main branch is being validated"

docs/kernel-policy.md
```diff
@@ -0,0 +1,85 @@
...
+continuously on *4.14* and *5.10* host and guest kernels. While other
+versions and other kernel configs might work, they are not periodically
+validated in our test suite, and using them might result in unexpected behaviour.
+Once enabled, a kernel version is supported for a **minimum of 2 years**.
+
+We are validating the currently supported Firecracker releases as per
+[Firecracker’s release policy](../docs/RELEASE_POLICY.md). Adding support
```
raduiliescu, line 14: I would add before adding support... Starting with release v1.0 each major and minor release will specify the supported kernel versions. Subsequent patch releases will continue to support the same kernel versions as the minor release.

docs/kernel-policy.md
```diff
@@ -0,0 +1,85 @@
+# Firecracker's Kernel Support Policy
+
+Being a virtual machine monitor, Firecracker represents a component in a
+larger stack, one in which it is tightly coupled with the guest and host
+kernels on which it is run.
```
raduiliescu, line 5: I  would add that With this kernel support policy we want to offer customers predictability into future kernel related changes.

docs/kernel-policy.md
georgepisaltu: Shouldn't this be mentioned in the release policy rather than here?

docs/kernel-policy.md
```diff
@@ -0,0 +1,85 @@
...
+
+Below is a per-functionality breakdown of guest kernel modules
+relevant to Firecracker for all platforms:
+
+* serial console - `CONFIG_SERIAL_8250_CONSOLE`, `CONFIG_PRINTK`
+* initrd support - `CONFIG_BLK_DEV_INITRD`
+* virtio devices - `CONFIG_VIRTIO_MMIO`
+  * balloon - `CONFIG_MEMORY_BALLOON`, `CONFIG_VIRTIO_BALLOON`
+  * block - `CONFIG_VIRTIO_BLK`
+    * partuuuid support - `CONFIG_MSDOS_PARTITION`
```
georgepisaltu, line 51:
```suggestion
    * partuuid support - `CONFIG_MSDOS_PARTITION`
```

docs/kernel-policy.md
raduiliescu: Is a good idea. I propose to mention it there.

dianpopa:
I added references to the kernel support policy throughout our doc.
Also mentioned it in the CHANGELOG.
@raduiliescu @alindima @georgepisaltu let me know what you think!

FAQ.md
```diff
@@ -70,10 +70,10 @@ and guest.
 
 ### What operating systems are supported by Firecracker?
 
-Firecracker supports Linux host and guest operating systems with kernel versions
-4.14 and above, as well as
+Firecracker supports Linux host and guest operating systems as well as
 [OSv](http://blog.osv.io/blog/2019/04/19/making-OSv-run-on-firecraker/) guests.
-The long-term support plan is still under discussion.
+Currently supported host/guest kernel versions can be found in the
```
raduiliescu, line 75: I suggest the following change: Firecracker supports Linux host and guest operating systems with kernel version 4.14 and 5.10. More information supported kernel versions can be found here. Firecracker also runs OSv guests

README.md
```diff
@@ -141,6 +141,11 @@ Firecracker currently only supports little-endian platforms. Firecracker will
 not compile for big-endian architectures, and will not work correctly with
 big-endian configured guests.
 
+## Supported kernels
```
raduiliescu, line 144: I would name the version explicitly and then link for more details our support policy.

docs/getting-started.md
```diff
@@ -23,11 +23,10 @@ sudo setfacl -m u:${USER}:rw /dev/kvm
 
 The generic requirements are explained below:
 
-- **Linux 4.14+**
+- **Linux**
 
   Firecracker currently supports physical Linux **x86_64** and **aarch64**
```
raduiliescu, line 28: I would put again the kernel versions explicitly, and *maybe* a reference to the support policy.

raduiliescu (commented):
The policy looks good. I think it is better to be explicit in the FAQ, README and Getting started on the supported kernels. For example in the FAQ we should answer the question spot on, rather than providing a link. In design I am fine with leaving just a reference.

dianpopa:
> The policy looks good. I think it is better to be explicit in the FAQ, README and Getting started on the supported kernels. For example in the FAQ we should answer the question spot on, rather than providing a link. In design I am fine with leaving just a reference.

I understand your suggestion but we need to be aware that every time we modify our support policy (adding or removing support support for some kernel) we would need to modify this 3 files. This is prone to errors. @firecracker-microvm/compute-capsule what is your take on this?

georgepisaltu:
> > The policy looks good. I think it is better to be explicit in the FAQ, README and Getting started on the supported kernels. For example in the FAQ we should answer the question spot on, rather than providing a link. In design I am fine with leaving just a reference.
> 
> I understand your suggestion but we need to be aware that every time we modify our support policy (adding or removing support support for some kernel) we would need to modify this 3 files. This is prone to errors. @firecracker-microvm/compute-capsule what is your take on this?

Our documentation is constantly growing. I think referencing other parts of the documentation scales better than duplicating information throughout multiple files. I quite like the link approach in the FAQ for example, especially because kernel support is a complex topic and if a user is interested in it, they should in fact read the full documentation. The link not only makes the codebase more maintainable, but it also provides users who are interested in this an easier way of finding relevant documentation.

alindima:
> > > The policy looks good. I think it is better to be explicit in the FAQ, README and Getting started on the supported kernels. For example in the FAQ we should answer the question spot on, rather than providing a link. In design I am fine with leaving just a reference.
> > 
> > 
> > I understand your suggestion but we need to be aware that every time we modify our support policy (adding or removing support support for some kernel) we would need to modify this 3 files. This is prone to errors. @firecracker-microvm/compute-capsule what is your take on this?
> 
> Our documentation is constantly growing. I think referencing other parts of the documentation scales better than duplicating information throughout multiple files. I quite like the link approach in the FAQ for example, especially because kernel support is a complex topic and if a user is interested in it, they should in fact read the full documentation. The link not only makes the codebase more maintainable, but it also provides users who are interested in this an easier way of finding relevant documentation.

I agree with Diana and George. I think this approach scales better and we'll most surely forget about this in the long run.
There are examples of such outdated documentation even on our Readme: https://github.com/firecracker-microvm/firecracker#features--capabilities

raduiliescu:
Ok, let's push it like that.

raduiliescu (approved):

PR #2872, opened 2022-01-20
Specify network interfaces that accept MMDS requests through `/mmds/config` API

### LINKED ISSUE #2174: Align MMDS configuration and MMDS operation device requirements
## Describe the bug

Per RFC issue #1866 / PR #1883: 

> ... the APIs for MMDS data store creation or MMDS configuration do not output any error if there is no network device configured. However, in terms of its operation, MMDS is not reachable if there is no network device which allows MMDS requests

and 

> The misalignment which might cause customer confusion comes from the fact that the MMDS configuration requirements does not correspond with MMDS operation requirements. I can not say this is a required design choice or simply an arbitrary happening due to unclear requirements. My understanding is that it is not a blocker of any sort for Firecracker usage and configuration since we haven't had any issues with it so far. I also believe that with the advent of the Device-API matrix it will be easier to think for the future in terms of API refinement in this area. Feel free to open any issue on this topic.

## To Reproduce

Issue is identified per @iulianbarbu comments in PR #1883

## Expected behaviour

Configuring MMDS for a uVM without a network device should return an error to the effect that MMDS **operation** requires a network device.

## Environment

- Firecracker version.

```bash
$ firecracker --version
Firecracker v0.21.1
```
- Host and kernel version.
```bash
$ uname -a
Linux desktop.local.lan 5.4.0-47-generic #51~18.04.1-Ubuntu SMP Sat Sep 5 14:35:50 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

- Rootfs used: Not relevant
- Architecture: ` x86_64`

## Additional context

**How has this bug affected you?** Consuming Firecracker errors requires different logic depending on the API-device pair. 

**What are you trying to achieve?** Robust error handling.

**Do you have any idea of what the solution might be?** Single code path for Firecracker API-device pair error handling?

## Checks

luminitavoicu:
# Reason for This PR

Provide a cleaner API for allowing MMDS requests on network interfaces. 
Also fixes #2174

## Description of Changes

- Add a `network_interfaces` field that takes a list of the network interface IDs capable of forwarding packets to the MMDS. Network interface IDs mentioned must be valid at the time of this request. The net device model will reply to HTTP GET requests sent to the MMDS address via the interfaces mentioned. In this case, both ARP requests and TCP segments heading to `ipv4_address` are intercepted by the device model, and do not reach the associated TAP device.
- Remove `allow_mmds_requests` from the body of the request that creates network interfaces.

## License Acceptance

## PR Checklist

- [X] The issue which led to this PR has a clear conclusion.
- [X] This PR follows the solution outlined in the related issue.
- [X] Any API changes follow the [Runbook for Firecracker API changes][2].

### DIFF
--- a/src/vmm/src/resources.rs
+++ b/src/vmm/src/resources.rs
@@ -301,18 +301,8 @@ impl VmResources {
         &mut self,
         body: NetworkInterfaceConfig,
     ) -> Result<NetworkInterfaceError> {
-        self.net_builder.build(body).map(|net_device| {
-            // Update `Net` device `MmdsNetworkStack` IPv4 address.
-            match &self.mmds_config {
-                Some(cfg) => cfg.ipv4_addr().map_or((), |ipv4_addr| {
-                    net_device
-                        .lock()
-                        .expect("Poisoned lock")
-                        .set_mmds_ipv4_addr(ipv4_addr);
-                }),
-                None => (),
-            };
-        })
+        let _ = self.net_builder.build(body)?;
+        Ok(())
     }
 
     /// Sets a vsock device to be attached when the VM starts.
@@ -326,21 +316,7 @@ impl VmResources {
         config: MmdsConfig,
         instance_id: &str,
     ) -> Result<MmdsConfigError> {
-        // Check IPv4 address validity.
-        let ipv4_addr = match config.ipv4_addr() {
-            Some(ipv4_addr) if is_link_local_valid(ipv4_addr) => Ok(ipv4_addr),
-            None => Ok(MmdsNetworkStack::default_ipv4_addr()),
-            _ => Err(MmdsConfigError::InvalidIpv4Addr),
-        }?;
-
-        // Update existing built network device `MmdsNetworkStack` IPv4 address.
-        for net_device in self.net_builder.iter_mut() {
-            net_device
-                .lock()
-                .expect("Poisoned lock")
-                .set_mmds_ipv4_addr(ipv4_addr);
-        }
-
+        self.set_mmds_network_stack_config(&config)?;
         self.set_mmds_version(config.version, instance_id)?;
 
         self.mmds_config = Some(MmdsConfig {
@@ -348,6 +324,7 @@ impl VmResources {
             ipv4_address: config
                 .ipv4_addr()
                 .or_else(|| Some(MmdsNetworkStack::default_ipv4_addr())),
+            network_interfaces: config.network_interfaces,
         });
 
         Ok(())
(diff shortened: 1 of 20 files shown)


src/vmm/src/resources.rs
```diff
@@ -301,18 +301,8 @@ impl VmResources {
...
-                    net_device
-                        .lock()
-                        .expect("Poisoned lock")
-                        .set_mmds_ipv4_addr(ipv4_addr);
-                }),
-                None => (),
-            };
-        })
+        let _ = self.net_builder.build(body)?;
+        Ok(())
```
dianpopa, line 305: nit:
```suggestion
        let _ = self.net_builder.build(body)
```

dianpopa (commented):
There are some passages from the [mmds doc](../../tree/main/docs/mmds) that need updating, for example this one: https://github.com/firecracker-microvm/firecracker/blob/main/docs/mmds/mmds-user-guide.md#example.

luminitavoicu:
> There are some passages from the [mmds doc](../../tree/main/docs/mmds) that need updating, for example this one: https://github.com/firecracker-microvm/firecracker/blob/main/docs/mmds/mmds-user-guide.md#example.

Indeed, there are many updates needed to the docs. I would prefer to include them in the PR that updates MMDS's behavior, since these changes are tightly coupled with MMDS. WDYT?

dianpopa (approved):

tests/integration_tests/functional/test_api.py
```diff
@@ -319,6 +298,101 @@ def test_net_api_put_update_pre_boot(test_microvm_with_api):
     assert test_microvm.api_session.is_status_no_content(response.status_code)
 
 
+def test_api_mmds_config(test_microvm_with_api):
```
raduiliescu, line 301: What happens when an interface is configured via /network-interfaces, mmds is enabled on it, and then the interface deleted from /network-interfaces?

luminitavoicu:
```rust
        // Create `MmdsNetworkStack` and configure the IPv4 address for
        // existing built network devices whose names are defined in the
        // network interface ID list.
        for net_device in self.net_builder.iter_mut() {
            let mut net_device_lock = net_device.lock().expect("Poisoned lock");
            if network_interfaces.contains(net_device_lock.id()) {
                net_device_lock.configure_mmds_network_stack(ipv4_addr);
            } else {
                net_device_lock.disable_mmds_network_stack();
            }
        }
```

This piece of code from `set_mmds_config()` handles this case. We iterate through all network devices attached and if the ID of the net iface is present in the `network_interfaces` list provided through `MmdsConfig`, then we configure it to allow MMDS requests. Otherwise, we disable the MMDS Network Stack from that interface. This ensures that only the interfaces specified in the list accept forwarding requests to MMDS.

dianpopa: I think Radu was referring to the case where an user appends an interface and then deletes it. We do not offer a delete api endpoint in Firecracker. Also, the patch request on network interface allows updating of the rate limiters only.

AlexandruCihodaru (approved):

PR #2876, opened 2022-01-25
api: rename `ht_enabled` to `smt`, make it optional and forbid setting it to True on ARM

alindima:
# Reason for This PR

We plan on renaming/removing ht_enabled in a future release and this is a first step towards this without introducing a breaking change in this release.
Also, making it required did not make sense for ARM, where it actually does nothing if configured. Added an error for when it's configured as True on ARM.

## Description of Changes

View commits

## License Acceptance

## PR Checklist

### DIFF
--- a/src/api_server/src/request/machine_configuration.rs
+++ b/src/api_server/src/request/machine_configuration.rs
@@ -22,13 +22,17 @@ pub(crate) fn parse_put_machine_config(body: &Body) -> Result<ParsedRequest, Err
     #[cfg(target_arch = "aarch64")]
     check_unsupported_fields(&vm_config)?;
 
-    if vm_config.vcpu_count.is_none()
-        || vm_config.mem_size_mib.is_none()
-        || vm_config.ht_enabled.is_none()
-    {
+    if vm_config.vcpu_count.is_none() {
         return Err(Error::Generic(
             StatusCode::BadRequest,
-            "Missing mandatory fields.".to_string(),
+            "Missing mandatory field: `vcpu_count`.".to_string(),
+        ));
+    }
+
+    if vm_config.mem_size_mib.is_none() {
+        return Err(Error::Generic(
+            StatusCode::BadRequest,
+            "Missing mandatory field: `mem_size_mib`.".to_string(),
         ));
     }
 
@@ -99,30 +103,39 @@ mod tests {
 
         // 2. Test case for mandatory fields.
         let body = r#"{
-                "mem_size_mib": 1024,
-                "ht_enabled": false
-              }"#;
-        assert!(parse_put_machine_config(&Body::new(body)).is_err());
-
-        let body = r#"{
-                "vcpu_count": 8,
-                "ht_enabled": false
-                }"#;
-        assert!(parse_put_machine_config(&Body::new(body)).is_err());
-
-        let body = r#"{
-                "vcpu_count": 8,
                 "mem_size_mib": 1024
               }"#;
         assert!(parse_put_machine_config(&Body::new(body)).is_err());
 
-        // 3. Test case a success scenario for both architectures.
+        let body = r#"{
+                "vcpu_count": 8
+                }"#;
+        assert!(parse_put_machine_config(&Body::new(body)).is_err());
+
+        // 3. Test case for success scenarios for both architectures.
+        let body = r#"{
+                "vcpu_count": 8,
+                "mem_size_mib": 1024
+              }"#;
+        let expected_config = VmConfig {
+            vcpu_count: Some(8),
+            mem_size_mib: Some(1024),
+            ht_enabled: None,
+            cpu_template: None,
+            track_dirty_pages: false,
+        };
+
+        match vmm_action_from_request(parse_put_machine_config(&Body::new(body)).unwrap()) {
+            VmmAction::SetVmConfiguration(config) => assert_eq!(config, expected_config),
+            _ => panic!("Test failed."),
+        }
+
         let body = r#"{
                 "vcpu_count": 8,
                 "mem_size_mib": 1024,
                 "ht_enabled": false,
                 "track_dirty_pages": true
-              }"#;
+            }"#;
         let expected_config = VmConfig {
             vcpu_count: Some(8),
             mem_size_mib: Some(1024),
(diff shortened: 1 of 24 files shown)


alindima:
There are some expected integration test failures caused by outdated artifacts in S3.
I will regenerate and upload them once I get some approvals on this PR

src/api_server/src/request/machine_configuration.rs
```diff
@@ -157,6 +166,35 @@ mod tests {
         {
             assert!(parse_put_machine_config(&Body::new(body)).is_err());
         }
+
+        // 5. Test that setting `ht_enabled: true` is successful on x86_64 while on aarch64, it is not.
```
raduiliescu, line 183: Can we also test that both true and false fail on aarch64?

src/api_server/src/request/machine_configuration.rs
```diff
@@ -91,13 +100,13 @@ mod tests {
         // 2. Test case for mandatory fields.
         let body = r#"{
                 "mem_size_mib": 1024,
-                "ht_enabled": true
+                "ht_enabled": false
```
raduiliescu: Not clear why this were changed from true to false?

alindima: So that those unit tests don't fail on ARM

src/api_server/src/request/machine_configuration.rs
alindima: Actually only `ht_enabled=True` fails on ARM. `ht_enabled` defaults to False on all platforms.

acatangiu:
Making `"ht_enabled": false` fail on ARM _is a breaking change_. Make sure to go through the versioned api support runbook :smile_cat:

alindima:
> Making `"ht_enabled": false` fail on ARM _is a breaking change_. Make sure to go through the versioned api support runbook 😸

This PR only makes `ht_enabled: true` fail on ARM and makes this field optional in the machine-config.
I don't believe we should go down the deprecation path since this was actually a bug. (this field does nothing on ARM if set to True).
Moreover, we are including this in the v1.0 release, which allows for breaking changes since it's a major version bump.

src/api_server/src/request/machine_configuration.rs
```diff
@@ -57,17 +63,24 @@ pub(crate) fn parse_patch_machine_config(body: &Body) -> Result<ParsedRequest, E
...
+fn check_unsupported_fields(vm_config: &VmConfig) -> Result<(), Error> {
+    if vm_config.cpu_template.is_some() {
+        // cpu_template is not supported on aarch64
+        return Err(Error::Generic(
+            StatusCode::BadRequest,
+            "CPU templates are not supported on aarch64".to_string(),
+        ));
     }
+
+    if let Some(true) = vm_config.ht_enabled {
```
georgepisaltu, line 76: I would disallow using this field on ARM altogether. What purpose does it serve to allow users to set it to false? It's just confusing.

```suggestion
    if vm_config.ht_enabled.is_some() {
```

If we go down this path, we could even have the `ht_enabled` field in the `VmConfig` struct marked with a `#[cfg(target_arch = "aarch64")]` so that we never mistakenly deserialize/serialize it.

alindima: I generally agree, but I was trying to minimise the breaking change impact. Let's discuss this some more

acatangiu: A breaking change is a breaking change is a breaking change.

Why do half the job to have only a breaking change in some cases? (setting `ht true on arm` will break, but `ht false on arm` will still work albeit it being useless/confusing)

alindima: We've discussed this topic inside the team and reached the following path forward:

keep the behaviour of this PR but additionally rename the `ht_enabled` param to `smt` (it will be optional and return error if set to True on ARM). This enables us to add support for 'emulated' smt in ARM as well if we need to in the future. We think this is the most extensible and cleanest option.

WDYT @acatangiu ?

alindima: I've implemented this in the latest commit

acatangiu: Sounds reasonable, but I don't see why you're not disabling/removing it altogether on ARM. You could seamlessly add it later with default value `false` when/if you ever implement smt emulation for arm. And in the meantime you're not left with a stub optional param which allows only false or otherwise fails the request.

You're basically keeping a button saying "this button might be useful later, but don't push it now or it will error" :smile:

alindima: > L.E.: an argument for keeping the stub is if that way the api requests would be identical between x86_64 and aarch64 -> more uniform code on customer side

This is one of the reasons I thought about.
It's also that otherwise, it may also be confusing for users to see that there is no ht_enabled config on arm. One may then wonder if it's emulated or not. The answer would be clearer if there was an ht_enabled option that defaults to `false` and that only allows False on Arm.

tests/integration_tests/functional/test_api.py
```diff
@@ -181,7 +181,7 @@ def test_api_put_update_pre_boot(test_microvm_with_api):
     # The machine configuration has a default value, so all PUTs are updates.
     microvm_config_json = {
         'vcpu_count': 4,
-        'ht_enabled': True,
+        'smt': not platform.machine() == 'aarch64',
```
georgepisaltu, line 184: Why not `platform.machine() == 'x86_64'`?

tests/integration_tests/functional/test_api.py
```diff
@@ -393,6 +393,7 @@ def test_api_mmds_config(test_microvm_with_api):
     )['mmds-config']['version'] == "V2"
 
 
+# pylint: disable=too-many-statements
```
georgepisaltu, line 396: Maybe we should split this test into multiple parts so that we get rid of this error. The test is indeed pretty long now.

georgepisaltu (approved):
PR looks great! Just a couple of minor comments.

luminitavoicu (approved):
Great job!

tests/integration_tests/functional/test_api.py
alindima: done

tests/integration_tests/functional/test_api.py
alindima: It's fairly long but I can't think of a way to purposefully split this test into multiple ones. It's a pretty cohesive one.

alindima:
I've regenerated S3 artifacts and changed microvm config files to no longer include `ht_enabled` flag.
I've also performed a couple of changes in the test framework to work with the older firecracker artifacts.
The PR is currently using a temporary folder in the S3 bucket but once I see all tests are passing I'll move the artifacts and switch back the folder.

georgepisaltu (approved):

AlexandruCihodaru (approved):

luminitavoicu (approved):

PR #2878, opened 2022-01-26
tests: ensure MMDS works as expected when configuring microVM from file

luminitavoicu:
Ensure MMDS works as expected (both default and V2) when spawning
microVM from a configuration file.

# Reason for This PR

Enhance MMDS tests.

## Description of Changes

Add tests to ensure that MMDS works as expected when configuring microVM from file.

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/framework/jailer.py
+++ b/tests/framework/jailer.py
@@ -247,7 +247,7 @@ class JailerContext:
             )
             utils.run_cmd(cmd)
 
-        if self.netns:
+        if self.netns and self.netns not in utils.run_cmd('ip netns list')[1]:
             utils.run_cmd('ip netns add {}'.format(self.netns))
 
     def cleanup(self):
--- a/tests/integration_tests/functional/test_cmd_line_start.py
+++ b/tests/integration_tests/functional/test_cmd_line_start.py
@@ -12,10 +12,15 @@ from retry.api import retry_call
 import pytest
 
 from framework import utils
+from framework.artifacts import NetIfaceConfig
+from framework.utils import generate_mmds_session_token, \
+    generate_mmds_v2_get_request
+import host_tools.network as net_tools
 
 
 def _configure_vm_from_json(test_microvm, vm_config_file):
-    """Configure a microvm using a file sent as command line parameter.
+    """
+    Configure a microvm using a file sent as command line parameter.
 
     Create resources needed for the configuration of the microvm and
     set as configuration file a copy of the file that was passed as
@@ -42,7 +47,8 @@ def _configure_vm_from_json(test_microvm, vm_config_file):
 
 
 def _add_metadata_file(test_microvm, metadata_file):
-    """Configure the microvm using a metadata file.
+    """
+    Configure the microvm using a metadata file.
 
     Given a test metadata file this creates a copy of the file and
     uses the copy to configure the microvm.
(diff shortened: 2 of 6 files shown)


tests/integration_tests/functional/test_cmd_line_start.py
```diff
@@ -206,6 +225,73 @@ def test_config_with_default_limit(test_microvm_with_api, vm_config_file):
...
+
+    @type: functional
+    """
+    test_microvm = test_microvm_with_api
+
+    _configure_vm_from_json(test_microvm, vm_config_file)
+    _configure_network_interface(test_microvm)
+
+    # Network namespace has already been created.
+    test_microvm.spawn(with_netns=False)
```
serban300, line 244: Why do we need to avoid using a netns ?

tests/integration_tests/functional/test_cmd_line_start.py
```diff
@@ -206,6 +225,73 @@ def test_config_with_default_limit(test_microvm_with_api, vm_config_file):
     assert stdout.encode("utf-8") == response_err.encode("utf-8")
 
 
+@pytest.mark.parametrize(
+    "vm_config_file",
+    ["framework/vm_config_with_mmdsv1.json"]
+)
+def test_config_start_with_mmds_default(test_microvm_with_api, vm_config_file):
```
serban300, line 232: I would try to rename this to make it more consistent with `test_with_config_and_metadata_no_api`.

tests/integration_tests/functional/test_cmd_line_start.py
luminitavoicu: Since we start the microVM from config file, we need to create the tap device before calling `spawn()` ( so we need to add the network namespace before that). That flag ensures that we don't try to add the netns twice.

serban300: I see. Makes sense. But can we modify the spawning logic in order to create the netns before starting the VM instead ? I think it would be simpler.

luminitavoicu: `spawn()` already creates the nets before starting the jailer, but this is still problematic when configuring a VM from file. Here is why:
When configuring the VM from API, we call `spawn()`, which creates the netns and starts the jailer. Then, if we want to attach a net device, we first need to create the associated tap, then call the FC API to create the network interface. 
When configuring a VM from a json file, if we want to be able to attach a net device, we need to create the network namespace AND THEN create the tap, ALL BEFORE starting the jailer.

The current approach adds a flag to `spawn()` to disallow the creation of the network namespace, so that we can have better control over it. Instead of letting `spawn()` handle the netns, we create the network namespace and tap, then call `spawn()`.

Another approach would be to embed tap creation into `spawn()`, but I feel like this would be a much more rigid approach. It would also burden the `spawn()` function in case we mean to start a microVM from json file with multiple network devices. WDYT?

tests/integration_tests/functional/test_cmd_line_start.py
alindima: well this one actually has an api. should we add tests for both scenarios? checking config start with mmds default both with and without an api

tests/integration_tests/functional/test_cmd_line_start.py
serban300: Ok, got it. It makes sense to keep it as it is.

luminitavoicu: Done.

tests/framework/jailer.py
```diff
@@ -247,7 +247,7 @@ def setup(self, use_ramdisk=False):
             )
             utils.run_cmd(cmd)
 
-        if self.netns:
+        if self.netns and self.netns not in utils.run_cmd('ip netns list')[1]:
```
serban300, line 250: `utils.run_cmd('ip netns list')[1]` Contains the entire list of net namespaces ? Or just the first one ?

tests/integration_tests/functional/test_cmd_line_start.py
```diff
@@ -55,6 +60,67 @@ def _add_metadata_file(test_microvm, metadata_file):
     test_microvm.metadata_file = vm_metadata_path
 
 
+def _configure_network_interface(test_microvm):
```
serban300, line 63: Will these netns and tap be cleaned up ? We should make sure that they are deleted after the test ends or at any assert error.

tests/framework/jailer.py
luminitavoicu: From the [man page](https://man7.org/linux/man-pages/man8/ip-netns.8.html): 

> ip netns list - show all of the named network namespaces

              This command displays all of the network namespaces in
              /var/run/netns

`run_cmd()` returns a tuple of (return code, stdout, stderr). This is why we compare against result[1], which is `stdout`.

tests/integration_tests/functional/test_cmd_line_start.py
alindima: I see that the ns is removed on the jailer destructor. as far as cleanup for the tap device, I couldn't find it

luminitavoicu: Indeed, there is no destructor for the tap device. I found it very odd, especially since the `__init__()` for `class Tap` said:

> The tap will
   stay on the host as long as the object obtained by instantiating this
   class will be in scope. Once it goes out of scope, its destructor will
   get called and the tap interface will get removed.

Indeed, there was a clean-up mechanism that was removed as part of [this commit](https://github.com/firecracker-microvm/firecracker/commit/c239013d0714bc5d8a1bbf69128a801b901b0bc0). Putting it back cause some tests to fail so how about we unblock this PR, since is a dependency for v1.0 and the clean-up problem is more of a general framework issue. I will work on adding a clean-up handle for the `Tap` as part of a separate PR. @alindima @serban300, WDYT?

serban300: Ok, sounds good to me, but please log an issue about this. Also I think that for the netns we should do the same. Have an object and have it cleaned up at the end of scope. Instead of cleaning it up as part of the jailer destruction. Since we can use netns without the jailer as we do here.

serban300 (approved):

alindima (approved):

PR #2879, opened 2022-01-27
Set RDA bit in serial after restore

AlexandruCihodaru:
# Reason for This PR
 After resuming from snapshot, the console is unresponsive.

## Description of Changes

* Set the RDA bit on serial after restore to kick the dive
* Added integration test for console functionality after restore

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -528,6 +528,8 @@ pub fn build_microvm_from_snapshot(
         MMIODeviceManager::restore(mmio_ctor_args, &microvm_state.device_states)
             .map_err(MicrovmStateError::RestoreDevices)
             .map_err(RestoreMicrovmState)?;
+    vmm.emulate_serial_init()
+        .map_err(StartMicrovmError::Internal)?;
 
     // Move vcpus to their own threads and start their state machine in the 'Paused' state.
     vmm.start_vcpus(
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -354,12 +354,28 @@ impl Vmm {
 
     /// Sets RDA bit in serial console
     pub fn emulate_serial_init(&self) -> Result<()> {
+        #[cfg(target_arch = "aarch64")]
+        use devices::legacy::SerialDevice;
         #[cfg(target_arch = "x86_64")]
         let mut serial = self
             .pio_device_manager
             .stdio_serial
             .lock()
             .expect("Poisoned lock");
+
+        #[cfg(target_arch = "aarch64")]
+        let serial_bus_device = self.get_bus_device(DeviceType::Serial, "Serial");
+        #[cfg(target_arch = "aarch64")]
+        if serial_bus_device.is_none() {
+            return Ok(());
+        }
+        #[cfg(target_arch = "aarch64")]
+        let mut serial_device_locked = serial_bus_device.unwrap().lock().expect("Poisoned lock");
+        #[cfg(target_arch = "aarch64")]
+        let serial = serial_device_locked
+            .as_mut_any()
+            .downcast_mut::<SerialDevice>()
+            .expect("Unexpected BusDeviceType");
 
         // When restoring from a previously saved state, there is no serial
         // driver initialization, therefore the RDA (Received Data Available)
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -52,6 +52,7 @@ use crate::vstate::{
     vm::Vm,
 };
 use arch::DeviceType;
+use devices::legacy::serial::{IER_RDA_BIT, IER_RDA_OFFSET};
 use devices::virtio::balloon::Error as BalloonError;
 use devices::virtio::{
     Balloon, BalloonConfig, BalloonStats, Block, MmioTransport, Net, BALLOON_DEV_ID, TYPE_BALLOON,
@@ -349,6 +350,28 @@ impl Vmm {
     /// Returns a reference to the inner `GuestMemoryMmap` object if present, or `None` otherwise.
     pub fn guest_memory(&self) -> &GuestMemoryMmap {
         &self.guest_memory
+    }
+
+    /// Sets RDA bit in serial console
+    pub fn emulate_serial_init(&self) -> Result<()> {
+        #[cfg(target_arch = "x86_64")]
+        let mut serial = self
+            .pio_device_manager
+            .stdio_serial
+            .lock()
+            .expect("Poisoned lock");
+
+        // When restoring from a previously saved state, there is no serial
+        // driver initialization, therefore the RDA (Received Data Available)
+        // interrupt is not enabled. Because of that, the driver won't get
+        // notified of any bytes that we send to the guest. The clean solution
+        // would be to save the whole serial device state when we do the vm
+        // serialization. For now we set that bit manually
+        serial
+            .serial
+            .write(IER_RDA_OFFSET, IER_RDA_BIT)
+            .map_err(|_| Error::Serial(std::io::Error::last_os_error()))?;
+        Ok(())
     }
 
     /// Injects CTRL+ALT+DEL keystroke combo in the i8042 device.
--- a/tests/integration_tests/functional/test_serial_io.py
+++ b/tests/integration_tests/functional/test_serial_io.py
@@ -4,15 +4,18 @@
 
 import fcntl
 import os
+import platform
 import subprocess
 import termios
 import time
 
 from framework.microvm import Serial
 from framework.state_machine import TestState
+from framework.builder import MicrovmBuilder, SnapshotBuilder
+from framework.artifacts import SnapshotType
 from framework import utils
 import host_tools.logging as log_tools
-import host_tools.network as net_tools  # pylint: disable=import-error
+import host_tools.network as net_tools
 
 
 class WaitTerminal(TestState):  # pylint: disable=too-few-public-methods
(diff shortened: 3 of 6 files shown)


tests/integration_tests/functional/test_serial_io.py
```diff
@@ -44,6 +46,56 @@ def handle_input(self, unused_serial, _) -> TestState:
...
 
 
+def test_serial_after_snapshot(bin_cloner_path):
+    """
+    Test that writing to stdout works after restore.
+
+    @type: functional
+    """
+    vm_builder = MicrovmBuilder(bin_cloner_path)
+    net_iface = NetIfaceConfig()
```
dianpopa, line 56: I do not think we need this.

tests/integration_tests/functional/test_serial_io.py
```diff
@@ -44,6 +46,56 @@ def handle_input(self, unused_serial, _) -> TestState:
         return self
 
 
+def test_serial_after_snapshot(bin_cloner_path):
+    """
+    Test that writing to stdout works after restore.
```
alindima, line 51: you cannot write to stdout 😄 
I suggest:
```suggestion
    Test serial I/O after restoring from a snapshot.
```

tests/integration_tests/functional/test_serial_io.py
```diff
@@ -44,6 +46,56 @@ def handle_input(self, unused_serial, _) -> TestState:
...
+    serial.tx("root")
+
+    snapshot_builder = SnapshotBuilder(microvm)
+    disks = [root_disk.local_path()]
+    # Create diff snapshot.
+    snapshot = snapshot_builder.create(disks,
+                                       ssh_key,
+                                       SnapshotType.FULL)
+
+    # Resume microVM and ensure session token is still valid on the base.
```
alindima, line 81: leftover comment?

dianpopa (dismissed):

src/vmm/src/lib.rs
```diff
@@ -334,6 +334,20 @@ impl Vmm {
...
+        // would be to save the whole serial device state when we do the vm
+        // serialization. For now we set that bit manually
+        #[cfg(target_arch = "x86_64")]
+        self.pio_device_manager
+            .stdio_serial
+            .lock()
+            .expect("Poisoned lock")
+            .serial
+            .write(1, 0b0000_0001)
+            .map_err(|_| Error::Serial(std::io::Error::last_os_error()))?;
```
acatangiu, line 350: Why not set it for `x86_64` in `MMIODeviceManager::restore()` just like you do for `aarch64`? It's weird to separate them like this..

src/vmm/src/device_manager/persist.rs
```diff
@@ -277,8 +277,23 @@ impl<'a> Persist<'a> for MMIODeviceManager {
...
+                    // driver initialization, therefore the RDA (Received Data Available)
+                    // interrupt is not enabled. Because of that, the driver won't get
+                    // notified of any bytes that we send to the guest. The clean solution
+                    // would be to save the whole serial device state when we do the vm
+                    // serialization. For now we set that bit manually
+                    serial
+                        .lock()
+                        .expect("Poisoned lock")
+                        .serial
+                        .write(1, 0b0000_0001)
```
alindima, line 293: I'd add a constant to represent that this 1 value is the interrupt enable register. Seems quite magical otherwise

src/vmm/src/lib.rs
alindima: I agree. It's also that resume is also called when resuming a paused VM (not neccessarily a snapshotted one).

acatangiu: Maybe in `builder::build_from_snapshot`? After building the PIO Serial on x86 and MMIO Serial on Arm you can call an `emulate_serial_driver_init()` specialized fn. wdyt?

AlexandruCihodaru: how is this different from how we kick all devices after resume?

acatangiu: > how is this different from how we kick all devices after resume?

well.. that was a _necessary_ (ugly) hack coming from the fact that `EventManager` events are not saved to snapshot.

I am hoping there is a nicer way to solve this for the Serial :pray:

AlexandruCihodaru: I guess we can put it in `build_from_snapshot` does it seem better?

src/vmm/src/builder.rs
```diff
@@ -465,6 +465,8 @@ pub fn build_microvm_from_snapshot(
         track_dirty_pages,
         vcpu_count,
     )?;
+    vmm.set_rda_pio_serial()
```
acatangiu, line 468: This would better be `builder::serial_set_rda()`, rather than a `vmm` member function since it's only called during building from snapshot and not useful anywhere else in the lifetime of vmm.

It should also be the same function called in the same place for all archs since it's needed on all archs. I suggest L#533 in this file (in `build_microvm_from_snapshot()` just after `MMIODeviceManager::restore`).

src/vmm/src/device_manager/persist.rs
```diff
@@ -277,8 +277,23 @@ impl<'a> Persist<'a> for MMIODeviceManager {
...
+                    // would be to save the whole serial device state when we do the vm
+                    // serialization. For now we set that bit manually
+                    serial
+                        .lock()
+                        .expect("Poisoned lock")
+                        .serial
+                        .write(1, crate::IER_RDA_BIT)
+                        .map_err(|_| {
+                            Error::Legacy(crate::Error::Serial(std::io::Error::last_os_error()))
+                        })?;
```
acatangiu, line 296: This can be removed from here and added to the specialized function you created (`serial_set_rda()`).
```suggestion
```

src/vmm/src/lib.rs
```diff
@@ -98,6 +98,9 @@ pub const FC_EXIT_CODE_SIGILL: ExitCode = 157;
 pub const FC_EXIT_CODE_BAD_CONFIGURATION: ExitCode = 152;
 /// Command line arguments parsing error.
 pub const FC_EXIT_CODE_ARG_PARSING: ExitCode = 153;
+// Received Data Available interrupt - for letting the driver know that
+// there is some pending data to be processed.
+const IER_RDA_BIT: u8 = 0b0000_0001;
```
acatangiu, line 103: This would be more suited in a serial-specific file (maybe `src/devices/src/legacy/serial.rs`). Anyone looking at the main `vmm` file would have no idea what `IER_RDA_BIT` means.

src/vmm/src/lib.rs
```diff
@@ -351,6 +354,25 @@ impl Vmm {
         &self.guest_memory
     }
 
+    /// Sets RDA bit in serial console
+    pub fn set_rda_pio_serial(&self) -> Result<()> {
```
acatangiu, line 358: This function should do the same for `aarch64` as well not just `x86_64`.

Also the function doesn't make much sense to live on the vmm and be called at any point in the vmm lifetime, but that's less important.

AlexandruCihodaru: I agree with you, but this is a quick fix in order to unblock the v1.0 release, afterthat we will consume serial state from upstream

src/vmm/src/lib.rs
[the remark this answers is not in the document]
AlexandruCihodaru: Good suggestion, but I think that we should not import that many constants from [vm-superio/Serial ](https://github.com/rust-vmm/vm-superio/blob/main/crates/vm-superio/src/serial.rs) and actually making them public there and consuming from there. WDYT?

acatangiu (approved):
:100:

src/vmm/src/builder.rs
```diff
@@ -528,6 +528,8 @@ pub fn build_microvm_from_snapshot(
         MMIODeviceManager::restore(mmio_ctor_args, &microvm_state.device_states)
             .map_err(MicrovmStateError::RestoreDevices)
             .map_err(RestoreMicrovmState)?;
+    vmm.emulate_serial_init()
+        .map_err(|_| StartMicrovmError::Internal(Error::Serial(std::io::Error::last_os_error())))?;
```
alindima, line 532: why do we disregard the error here? we should propagate it instead of calling `last_os_error` twice

src/vmm/src/lib.rs
```diff
@@ -354,13 +354,29 @@ impl Vmm {
...
+        #[cfg(target_arch = "aarch64")]
+        use devices::legacy::SerialDevice;
         #[cfg(target_arch = "x86_64")]
         let mut serial = self
             .pio_device_manager
             .stdio_serial
             .lock()
             .expect("Poisoned lock");
 
+        #[cfg(target_arch = "aarch64")]
```
alindima, line 366: we can have a cfg-ed block, like so:
```
{
... code here
}
```

alindima (commented):
Looks good

tests/integration_tests/functional/test_serial_io.py
```diff
@@ -44,6 +47,56 @@ def handle_input(self, unused_serial, _) -> TestState:
...
+
+    snapshot_builder = SnapshotBuilder(microvm)
+    disks = [root_disk.local_path()]
+    # Create diff snapshot.
+    snapshot = snapshot_builder.create(disks,
+                                       ssh_key,
+                                       SnapshotType.FULL)
+
+    response = microvm.vm.patch(state='Resumed')
+    assert microvm.api_session.is_status_no_content(response.status_code)
```
luminitavoicu, line 83: Do we need to resume the base microVM here if the next thing we do is kill it?

luminitavoicu (commented):
LGTM. I only have one nit:

src/vmm/src/lib.rs
alindima: turns out it's very tricky to get this right due to lifetime constraints. let's keep things as they are

alindima (approved):
Nice work!

luminitavoicu (approved):

PR #2893, opened 2022-02-10
Fix signal handler race condition on metrics.write()

alindima:
# Reason for This PR

FIxes the root cause of intermittent failure of `test_custom_seccomp.py::test_failing_fiter`
View commit messages.

## Description of Changes

View commit messages.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/logger/src/metrics.rs
+++ b/src/logger/src/metrics.rs
@@ -134,10 +134,19 @@ impl<T: Serialize> Metrics<T> {
     /// Upon success, the function will return `True` (if metrics system was initialized and metrics
     /// were successfully written to disk) or `False` (if metrics system was not yet initialized).
     ///
-    /// This function is supposed to be called only from a single thread and
+    /// This function is usually supposed to be called only from a single thread and
     /// is not meant to be used in a multithreaded scenario. The reason
     /// `metrics_buf` is enclosed in a `Mutex` is that `lazy_static` enforces
     /// thread-safety on all its members.
+    /// The only exception is for signal handlers that result in process exit, which may be run on
+    /// any thread. To prevent the race condition present in the serialisation step of
+    /// SharedIncMetrics, deadly signals use SharedStoreMetrics instead (which have a thread-safe
+    /// serialise implementation).
+    /// The only known caveat is that other metrics may not be properly written before exiting from
+    /// a signal handler. We make this compromise since the process will be killed anyway and the
+    /// important metric in this case is the signal one.
+    /// The alternative is to hold a Mutex over the entire function call, but this increases the
+    /// known deadlock potential.
     pub fn write(&self) -> Result<bool, MetricsError> {
         if self.is_initialized.load(Ordering::Relaxed) {
             match serde_json::to_string(&self.app_metrics) {
@@ -674,7 +683,7 @@ impl RtcEvents for RTCDeviceMetrics {
 #[derive(Default, Serialize)]
 pub struct SeccompMetrics {
     /// Number of errors inside the seccomp filtering.
-    pub num_faults: SharedIncMetric,
+    pub num_faults: SharedStoreMetric,
 }
 
 /// Metrics specific to the UART device.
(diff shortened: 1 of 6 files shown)


raduiliescu:
For the second commit I doubt this is the best approach. I am hardly familiar with RUST, but adding acquire/release a lock, on operations done on the hotpath is not OK. Maybe you can take the second commit out, and we fix just the issue with intermittent test.
One other approach with metrics on multiple threads is to have that metric split per thread, and when dumping just do an add.

alindima:
> I am hardly familiar with RUST, but adding acquire/release a lock, on operations done on the hotpath is not OK.

There are no locks added. These are just which atomic instruction flags that prevent the CPU from reordering loads of a value before stores on the same atomic.
On x86 this should have no effect because of the stronger memory model. On ARM it should just prevent the CPU from reordering the atomic instructions in ways that violate the load-after-store rule across different threads.

src/logger/src/metrics.rs
```diff
@@ -251,48 +251,45 @@ pub struct SharedIncMetric(AtomicUsize, AtomicUsize);
...
     fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
         // There's no serializer.serialize_usize() for some reason :(
-        let snapshot = self.0.load(Ordering::Relaxed);
-        let res = serializer.serialize_u64(snapshot as u64 - self.1.load(Ordering::Relaxed) as u64);
+        let snapshot = self.count();
+        let res = serializer.serialize_u64(snapshot as u64 - self.1.load(Ordering::Acquire) as u64);
 
         if res.is_ok() {
-            self.1.store(snapshot, Ordering::Relaxed);
+            self.1.store(snapshot, Ordering::Release);
```
serban300, line 284: I'm not sure if this is correct. If we can serialize from multiple threads, I think `snapshot` can be different on each one and the value that ends up being assigned to `self.1` in the end (`self.1.store(snapshot, Ordering::Release);`) won't necessarily be the most recent one. So we can end up in the situation where self.0 < self.1.

A solution would be to have another atomic bool that is marking whether there is a serialization in progress, and if there is, don't do anything. Although this wouldn't guarantee the serialization of the latest values.

alindima: `self.0` is incremented atomically from any thread. If we have Relaxed semantics, it may have different values on multiple threads when serialising.
If using Acquire semantics, like we do in `self.count()`, we can be sure that any writes to `self.0` with `Release` semantics (like we do in add()) are going to be executed and visible before the load.

I believe that the scenario you are describing is possible prior to this PR, due to the Relaxed ordering.

> A solution would be to have another atomic bool that is marking whether there is a serialization in progress, and if there is, don't do anything. Although this wouldn't guarantee the serialization of the latest values.

This is essentially simulating a mutex

alindima: anyway, I believe that the usage of the metrics system in signal handlers needs a redesign with thread-safety in mind.
Since this second commit is pretty controversial and may be introducing some performance degradation on ARM, I will remove it from this PR to unblock it and I will open a new issue for it

serban300: Would something like the following be possible ?
```
+-----------------------------------+-----------------------------------+
| Thread 1                          | Thread 2                          |
+-----------------------------------+-----------------------------------+
| let snapshot = self.count(); -> 5 | ...                               |
+-----------------------------------+-----------------------------------+
| ...                               | inc metric -> 6                   |
+-----------------------------------+-----------------------------------+
| ...                               | let snapshot = self.count(); -> 6 |
+-----------------------------------+-----------------------------------+
| ...                               | store snapshot in self.1 -> 6     |
+-----------------------------------+-----------------------------------+
| store snapshot in self.1 -> 5     |                                   |
+-----------------------------------+-----------------------------------+
```

alindima: Actually this is not possible since we only write the metrics from the VMM thread or the signal handler.
In the signal handler we now only use StoreMetrics so this would not be possible

alindima: The only known issue with the current form of this PR is that the other (non-signal) SharedIncMetrics may be slightly outdated when written from the signal handler. But I think this is a fine compromise since the process is going to be killed anyway

serban300: Ok, makes sense. Anyway this is hard to reason about and we need to make sure that `IncMetrics` won't be added to the signal handler thread. So should we mark this method as unsafe since it's not thread safe ? Or what would be the best way to do this ?

alindima: I can't think of a good way of preventing this, except for code reviews.
Rust's unsafe keyword is not a good fit in this case because it does not express thread-unsafety.
For example `Rc` is not thread safe but is not marked as unsafe.

The one way of showing that a type is not thread safe is by not implementing the `Sync` trait, but this is required since the type is a `lazy_static`.

This race condition is furthermore detailed in the comment above `Metrics.write()` function. Do you have any other ideas or do you think this is sufficient?

serban300: It doesn't seem sufficient to me because if I had to add a new metric I can imagine that it would be very easy to miss these comments. I wouldn't check the comments for the `Metrics.write()` or `serialize()` methods. It would be the same even if the `serialize()` method would be unsafe.

But I don't have a solution. I'm still thinking.

andreeaflorescu: If the type is not thread safe than `Sync` should not be implemented for it. Otherwise we're breaking the Rust safety principles and Rust cannot protect and benefit of the "fearless concurrency" anymore. The problem seems to be that we're doing more things than we should be doing in a context of the signal handler, so maybe a solution would be to simplify the signal handler. This will come at the expense of maybe metrics being missed, but it is arguably better than opening potentially other safety problems.

alindima: We have discussed removing logging and metrics systems from the signal handlers.
One idea is to use specialised files for dumping abrupt exit information from signal and panic handlers (much like a core dump).

This is a separate feature and will be worked on in the future. It requires careful thought and more investment.

This PR addresses the issue surgically, since it is creating much trouble for us with intermittent test failures.
We are not introducing problems with this PR, we are applying a patch over the known occurrence of this problem.

alindima: Created an issue for it: https://github.com/firecracker-microvm/firecracker/issues/2899

alindima: > It doesn't seem sufficient to me because if I had to add a new metric I can imagine that it would be very easy to miss these comments. I wouldn't check the comments for the Metrics.write() or serialize() methods. It would be the same even if the serialize() method would be unsafe.

I realised that the signal handler generation macro is hardcoded to use the `store` function, which is only defined for `StoreMetrics`. It would require a separate function in order to add an IncMetric to a signal handler, like it is for SIGPIPE. This should be even more evident in code reviews

serban300 (approved):

PR #2915, opened 2022-02-28
Fix a panic in cpuid/brand_string.rs

### LINKED ISSUE #2914: [Bug] Zero-length CPUID (on Chrome OS) causes panic: 'attempt to subtract with overflow'
# Describe the bug

The offending code is this:
```
        let mut len = Self::MAX_LEN;
        {
            let this_bytes = this.as_bytes();
            while this_bytes[len - 1] == 0 && len > 0 {
                len -= 1;
            }
```

You can get to a case where
```
len is 1
len -=1 -> len becomes 0
while this_bytes[unsigned 0 - 1]
```
The len check should come first to short-circuit the bad case.

## To Reproduce

On ChromeOS, download Firecracker main, go to `src/cpuid` and run `cargo test`. This gives the "attempt to subtract with overflow" error.

Alternatively, in prod, I ran through the quickstart guide and hit the much-less-helpful error
```
2022-02-28T13:38:25.634380343 [anonymous-instance:main:ERROR:src/firecracker/src/main.rs:91] Firecracker panicked at 'index out of bounds: the len is 48 but the index is 18446744073709551615', src/cpuid/src/brand_string.rs:124:19
```

## Expected behaviour

`[Author TODO: A clear and concise description of what you expected to happen.]`

## Environment

`[ - Firecracker version.]`
9d8bb4c8fff7d5c082e6405f4b842979881c7a42

`[ - Architecture.]`
x86-64

--

## Additional context

I can't even try to run firecracker

Run firecracker

Yeah, I might submit a PR.

## Checks

wimax-grapl:
# Reason for This PR

`[Author TODO: add issue #.]`
`[Open the PR after the related issue has a clear conclusion.]`
`[If there is no issue which states the need for this PR, create one first.]`
Fixes #2914

## Description of Changes
- Fix a length check that accesses `some_slice[unsigned 0 - 1]` on systems without a Brand String
- Refactor that length check so we can write a test around it that doesn't depend on the system it's run on

## License Acceptance

Yes

## PR Checklist

### DIFF
--- a/src/cpuid/src/brand_string.rs
+++ b/src/cpuid/src/brand_string.rs
@@ -118,14 +118,7 @@ impl BrandString {
             this.set_reg_for_leaf(leaf, Reg::Edx, cpuid_regs.edx);
         }
 
-        let mut len = Self::MAX_LEN;
-        {
-            let this_bytes = this.as_bytes();
-            while this_bytes[len - 1] == 0 && len > 0 {
-                len -= 1;
-            }
-        }
-        this.len = len;
+        this.len = null_terminator_index(this.as_bytes());
 
         Ok(this)
     }
@@ -300,6 +293,16 @@ impl BrandString {
     }
 }
 
+/// Find the position of the first null-terminator (\b'0') in a u8 slice,
+/// or return 0 if they're all null-terminators.
+/// This function is roughly equivalent to libc strlen.
+fn null_terminator_index(slice: &[u8]) -> usize {
+    slice
+        .iter()
+        .rposition(|&b| b != b'\0')
+        .map_or(0, |idx| idx + 1)
+}
+
 #[cfg(test)]
 mod tests {
     use std::iter::repeat;
@@ -410,6 +413,19 @@ mod tests {
         assert!(bstr.as_bytes().starts_with(BRAND_STRING_AMD));
         let bstr = BrandString::from_vendor_id(b"............");
         assert!(bstr.as_bytes() == vec![b'\0'; 48].as_slice());
+    }
+
+    #[test]
+    /// Prevent against https://github.com/firecracker-microvm/firecracker/issues/2914
+    fn test_null_terminator_index() {
+        let bytes = vec![b'\0'; 48];
+        assert_eq!(null_terminator_index(bytes.as_slice()), 0);
+
+        let bytes = vec![b'h', b'i', b'\0'];
+        assert_eq!(null_terminator_index(bytes.as_slice()), 2);
+
+        let bytes = vec![b'h', b'i', b'\0', b'\0'];
+        assert_eq!(null_terminator_index(bytes.as_slice()), 2);
     }
 
     #[test]


src/cpuid/src/brand_string.rs
```diff
@@ -300,6 +293,21 @@ impl BrandString {
...
 }
 
+trait NullTerminatorIndex {
+    fn null_terminator_index(&self) -> usize;
+}
+impl NullTerminatorIndex for &[u8] {
+    fn null_terminator_index(&self) -> usize {
+        let mut len = self.len();
+        {
+            while len > 0 && self[len - 1] == 0 {
```
wimax-grapl, line 303: The major change here is moving the `len > 0` check before the `[]` accessor.

Ideally I'd do this as an .iter().rev().position() check but I assume there's a reason it's coded C style :)

src/cpuid/src/brand_string.rs
```diff
@@ -118,14 +118,7 @@ impl BrandString {
...
 
-        let mut len = Self::MAX_LEN;
-        {
-            let this_bytes = this.as_bytes();
-            while this_bytes[len - 1] == 0 && len > 0 {
-                len -= 1;
-            }
-        }
-        this.len = len;
+        this.len = this.as_bytes().null_terminator_index();
```
serban300, line 121: Could we use `this.as_bytes().trim_end_with().len()` ?

serban300: Sorry, I didn't realize this.

src/cpuid/src/brand_string.rs
```diff
@@ -298,13 +298,10 @@ trait NullTerminatorIndex {
...
 impl NullTerminatorIndex for &[u8] {
     fn null_terminator_index(&self) -> usize {
-        let mut len = self.len();
-        {
-            while len > 0 && self[len - 1] == 0 {
-                len -= 1;
-            }
+        match self.iter().rposition(|&b| b != b'\0') {
+            Some(idx) => idx + 1,
+            None => 0,
```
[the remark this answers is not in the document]
wimax-grapl: The problem is that `from_host_cpuid` is seemingly not unit-testable against arbitrary data, just real honest-to-goodness memory addresses that are different per-CPU :/

src/cpuid/src/brand_string.rs
dianpopa: We could use `iter().reposition()`. Something along the lines:
```
this.len = {
            if let Some(index) = this.as_bytes().iter().rposition(|&x| x != 0) {
                index + 1
            } else {
                0
            }
        };
```
WDYT?

dianpopa: Oh, sorry I see that you are doing exactly that in your next commit. Cool
Make sure you squash the changes into one commit before merging.
Also use the 50/72 format for the commit message in order for the tests to pass.

src/cpuid/src/brand_string.rs
```diff
@@ -300,6 +293,13 @@ impl BrandString {
     }
 }
 
+fn null_terminator_index(slice: &[u8]) -> usize {
```
serban300, line 296: Nit: Another option would be to call this method `strlen` since I think it's equivalent to `strlen` from C.

wimax-grapl: I'm a fan of overly-descriptive function names, but I can leave a comment about this if that's a fair compromise?
i'd argue that naming it strlen could have people unfamiliar with libc thinking they could swap this out with str::len

src/cpuid/src/brand_string.rs
```diff
@@ -300,6 +293,13 @@ impl BrandString {
...
+    match slice.iter().rposition(|&b| b != b'\0') {
```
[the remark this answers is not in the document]
wimax-grapl: good point

src/cpuid/src/brand_string.rs
serban300: Agree. I think a comment describing more precisely what the function is doing would help.

wimax-grapl:
(Also, just to understand Firecracker's flow, I've just been `git commit --amend`ing due to the above comment:
> Make sure you squash the changes into one commit before merging.
Do you just not do squashes on the Github PR page, instead preferring to do it manually once it's approved?)

wimax-grapl:
Okay, commit message is below 50 (though Github seems to be changing my `#2914` into a `firecracker-microvm#2914`) on display - not at `git log` though.

This previously was failing some integration test about snapshots on 4.x - let's see if that still fails - otherwise, I think I'm good to go?

serban300 (approved):

serban300:
Yes, looks good ! The tests have passed also.

georgepisaltu (approved):
Thank you for the contribution!

PR #2920, opened 2022-03-07
Refactor MMDS usage and remove lazy_static global state

alindima:
# Reason for This PR

MMDS is currently used as a `lazy_static`, accessible from all threads.
This has the disadvantage that the data store is initialised without the user configuring MMDS. In a followup PR, we will make the Mmds data store optional.

## Description of Changes

This PR refactors the Vmm and Api code using MMDS so that we no longer have a global variable for the data store.
For more details, view commit messages

## License Acceptance

## PR Checklist

~~- [] The issue which led to this PR has a clear conclusion.~~
~~- [] This PR follows the solution outlined in the related issue.~~


src/devices/src/virtio/net/device.rs
```diff
@@ -167,17 +166,12 @@ impl Net {
         }
 
         let mut queue_evts = Vec::new();
-        for _ in QUEUE_SIZES.iter() {
+        let mut queues = Vec::new();
+        for size in QUEUE_SIZES.iter() {
```
georgepisaltu, line 171: We can get rid of the `.iter()`. Also, we could do `&size` to not deref later.

src/api_server/src/lib.rs
```diff
@@ -250,11 +237,10 @@ impl ApiServer {
             Ok((req_action, mut parsing_info)) => {
                 let mut response = match req_action {
                     RequestAction::Sync(vmm_action) => {
-                        self.serve_vmm_action_request(vmm_action, request_processing_start_us)
+                        let res =
+                            self.serve_vmm_action_request(vmm_action, request_processing_start_us);
+                        return res;
```
georgepisaltu, line 242: Isn't this functionally the same but a bit uglier 😆 ?

src/mmds/src/data_store.rs
```diff
@@ -190,16 +190,21 @@ impl Mmds {
...
         if self.data_store.is_null() {
             return String::from("{}");
         }
         self.data_store.to_string()
     }
 
+    // We do not check size of data_store before returning a result because due
+    // to limit from put/patch the data_store can not be bigger than the limit
+    // imposed by the server.
+    pub fn data_store(&self) -> Value {
```
georgepisaltu, line 196: nit: I would rename this to `data_store_value` or something like this. `data_store` implies you get a reference to the data store.

src/devices/src/virtio/net/device.rs
```diff
@@ -214,13 +215,18 @@ impl Net {
...
 
+    /// Provides the MmdsNetworkStack of this net device.
+    pub fn mmds_ns(&self) -> Option<&MmdsNetworkStack> {
+        self.mmds_ns.as_ref()
+    }
+
     /// Configures the `MmdsNetworkStack` to allow device to forward MMDS requests.
     /// If the device already supports MMDS, updates the IPv4 address.
-    pub fn configure_mmds_network_stack(&mut self, ipv4_addr: Ipv4Addr) {
+    pub fn configure_mmds_network_stack(&mut self, ipv4_addr: Ipv4Addr, mmds: Arc<Mutex<Mmds>>) {
```
georgepisaltu, line 225: The `mmds` parameter feels like it should be an option, since we don't actually use it for all possible paths. It seems to be used only when we don't have an MMDS ns, so we shouldn't have to provide it for the other case. Alternatively we could split this in 2 functions.

Note, this will make more sense if we ever decide to not have MMDS on by default, so I'm fine with it being like this as of right now.

src/firecracker/src/main.rs
```diff
@@ -448,11 +440,15 @@ fn build_microvm_from_json(
     config_json: String,
     instance_info: InstanceInfo,
     boot_timer_enabled: bool,
+    mmds_byte_limit: Option<usize>,
```
georgepisaltu, line 443: nit: I'd rename this to `mmds_max_size` or something like that.

src/api_server/src/lib.rs
georgepisaltu: Yes, didn't realize from the code snippet that the function goes on for longer than the `match`.

georgepisaltu (dismissed):

georgepisaltu (approved):

src/mmds/src/lib.rs
```diff
@@ -129,22 +121,18 @@ pub fn convert_to_response(request: Request) -> Response {
         );
     }
 
-    respond_to_request(request)
-}
-
-fn respond_to_request(request: Request) -> Response {
-    let mmds_version = MMDS.lock().expect("Poisoned lock").version();
+    let mut mmds_guard = mmds.lock().unwrap();
```
luminitavoicu, line 124: Any particular reason for removing the unwrap() error here?

src/vmm/src/rpc_interface.rs
```diff
@@ -908,11 +928,18 @@ mod tests {
...
+        pub fn set_mmds_config(
+            &mut self,
+            mmds_config: MmdsConfig,
+            _: &str,
+        ) -> Result<(), MmdsConfigError> {
             if self.force_errors {
                 return Err(MmdsConfigError::InvalidIpv4Addr);
             }
-            self.mmds_set = true;
+            let mut mmds_guard = self.mmds.lock().unwrap();
```
luminitavoicu, line 939: Same as above, I think an `expect()` with error message would be better here instead of plain `unwrap()`.

luminitavoicu (commented):
Really nice PR, great job!

src/mmds/src/lib.rs
alindima: I think you mean removing the expect, right?
Good point, reverted to an expect

src/vmm/src/rpc_interface.rs
alindima: since this is defined in the tests module, doesn't really matter. I did however replace the unwrap with expect in the other places 👍🏻

src/api_server/src/lib.rs
```diff
@@ -623,43 +462,42 @@ mod tests {
     }
 
     #[test]
-    fn test_bind_and_run() {
+    fn test_bind_and_run_this() {
```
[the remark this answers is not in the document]
alindima: forgot about this 😄  removed

src/firecracker/src/api_server_adapter.rs
```diff
@@ -125,6 +124,7 @@ pub(crate) fn run_with_api(
     process_time_reporter: ProcessTimeReporter,
     boot_timer_enabled: bool,
     payload_limit: Option<usize>,
+    metadata_json: Option<&str>,
```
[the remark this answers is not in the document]
alindima: I thought about this as well, but I think that it'd a pretty artificial grouping of parameters. I'm not sure that it wouldn't be harder to read in this case. I'd leave this as is, WDYT?

georgepisaltu (approved):

PR #2921, opened 2022-03-07
Add option to handle memory page-faults through UFFD when restoring from a snapshot

luminitavoicu:
# Reason for This PR

Add option to handle memory page-faults through UFFD when restoring from a snapshot. 

## Description of Changes

- Deprecate `mem_file_path` field for `/snapshot/load` API.
- Add new optional`mem_backend` field for `/snapshot/load` API. This is an
  object `MemoryBackend` that defines the configuration of the backend 
  responsible for handling memory loading during snapshot restore:

```.yaml
MemoryBackend:
    type: object
    required:
      - backend_type
      - backend_path
```

  - `backend_type` is an enum that can take either `File` or `Uffd` as
     value. 
  -  interpretation of `backend_path` field depends on the value of
    `backend_type`. If `File`, then the user must provide the path to file that
    contains the guest memory to be loaded. Otherwise, if `backend_type` is `Uffd`,
    then `backend_path` is the path to a unix domain socket where a custom page
    fault handler process is listening and expecting a UFFD to be sent by
    Firecracker. The UFFD is used to handle the guest memory page faults in the
    separate process.

When /snapshot/load specifies memory backend type as 'Uffd', Firecracker doesn't handle the memory file itself anymore and expects an external process to handle its guest memory page faults.

To do this, anonymous memory is mmap()ed as guest memory while keeping the original memory regions shape. Then a Uffd is created and each guest memory range is registered with the Uffd so that any page faults won't be handled by the kernel, but will come up as events on the Uffd.

Firecracker then sends the memory ranges descriptions/mappings along with the Uffd over a UnixDomainSocket specified in 'backend_path' parameter on the API call.

It is expected that on the other side there is already a process listening for incoming connections. Once Firecracker's connection is accepted, Firecracker sends the mappings and Uffd. The receiving process is from now responsible for handling any pagefaults on the Uffd.

The communication medium is a UDS, the protocol is SOCK_STREAM and the encoding is JSON.

Integration tests for UFFD are still in progress.

## License Acceptance

## PR Checklist


acatangiu (commented):
:partying_face:

CHANGELOG.md
```diff
@@ -14,6 +14,18 @@
 
 - The API `PATCH` method for `/machine-config` can be now used to change
   `track_dirty_pages` on aarch64.
+- Replaced `mem_file_path` parameter for `/snapshot/load` API by `mem_backend`.
```
georgepisaltu, line 17: Before going any deeper here, we need to ensure the changes that we make to the API are backwards compatible. In this regard, just replacing `mem_file_path` with another parameter is not enough. We need to first make `mem_file_path` optional, deprecate it, then add this new `mem_backend` parameter as optional. In the absence of the new parameter, the legacy one must be present, as per the old specification where it was required.

luminitavoicu: Snapshots are currently GA, so it's not a `requirement` to ensure backwards compatibility. I agree, however, that it would be better for our users to be compatible with the old functionality.

dianpopa (approved):
LGTM!

src/api_server/src/request/snapshot.rs
```diff
@@ -43,10 +54,65 @@ pub(crate) fn parse_patch_vm_state(body: &Body) -> Result<ParsedRequest, Error>
...
+        enable_diff_snapshots: snapshot_config.enable_diff_snapshots,
+        resume_vm: snapshot_config.resume_vm,
+    };
+
+    // Construct the `ParsedRequest` object.
+    let mut parsed_req = ParsedRequest::new_sync(VmmAction::LoadSnapshot(snapshot_params));
+
+    // If `mem_file_path` was present, set the deprecation message in `parsing_info`.
+    if let Some(msg) = deprecation_message {
+        parsed_req.parsing_info().append_deprecation_message(msg);
```
georgepisaltu, line 105: I think we could do this with `Option::map`.

src/api_server/swagger/firecracker.yaml
```diff
@@ -1097,7 +1116,6 @@ definitions:
   SnapshotLoadParams:
```
georgepisaltu, line 1116: Should add here that exactly one of the two `mem_*` parameters must be present in the body of the request.

georgepisaltu (dismissed):
Looks good overall, just a couple of nits.

src/api_server/src/request/snapshot.rs
luminitavoicu: Good suggestion! Done.

src/api_server/src/request/snapshot.rs
```diff
@@ -5,9 +5,23 @@ use super::super::VmmAction;
...
 
+/// Deprecation message for the `mem_file_path` field.
+const LOAD_DEPRECATION_MESSAGE: &str = "PUT /snapshot/load: mem_file_path field is deprecated.";
+/// None of the `mem_backend` or `mem_file_path` fields has been specified.
+pub const MISSING_FIELD: &str =
+    "missing field: either `mem_backend` or `mem_file_path` is required";
+/// Both the `mem_backend` and `mem_file_path` fields have been specified.
+/// Only specifying one of them is allowed.
+pub const TOO_MANY_FIELDS: &str =
+    "too many fields: either `mem_backend` or `mem_file_path` is required";
```
georgepisaltu, line 23: nit: Seems a bit unclear right now for someone reading, might also add "exclusively" here to suggest the exclusive or we're trying to impose.

```suggestion
    "too many fields: either `mem_backend` or `mem_file_path` exclusively is required";
```

georgepisaltu (commented):
Looks good!

src/vmm/src/persist.rs
```diff
@@ -498,11 +541,63 @@ fn guest_memory_from_file(
...
+            backend_mappings.as_bytes(),
+            // In the happy case we can close the fd since the other process has it open and is
+            // using it to serve us pages.
+            // The problem is that if other process crashes/exits, firecracker guest memory
+            // will simply revert to anon-mem behavior which would lead to silent errors and
+            // undefined behavior.
+            //
+            // If the fault serving process exits while we hold a copy of the FD as well, the
+            // uffd will still be alive but with no one to serve faults leading to guest freeze
+            // (which is an unfortunate, but explicit and defined behavior).
```
serban300, line 591: Do we plan to address this in the future ? Is there anything we can do to handle this scenario ?

src/vmm/src/memory_snapshot.rs
```diff
@@ -189,7 +190,7 @@ impl SnapshotMemory for GuestMemoryMmap {
                 .iter()
                 .map(|r| {
                     (
-                        Some(FileOffset::new(file.try_clone().unwrap(), r.offset)),
+                        file.map(|f| FileOffset::new(f.try_clone().unwrap(), r.offset)),
```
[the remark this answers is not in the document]
luminitavoicu: Done

src/vmm/src/persist.rs
```diff
@@ -67,6 +71,23 @@ pub struct MicrovmState {
...
 }
 
+/// This describes the mapping between Firecracker base virtual address and offset in the
+/// buffer or file backend for a guest memory region. It is used to tell an external
+/// process/thread where to populate the guest memory data for this range.
+///
+/// E.g. Guest memory contents for a region of `size` bytes can be found in the backend
+/// at `offset` bytes from the beginning, and should be copied/populated into `base_host_address`.
+#[derive(Clone, Debug, Serialize)]
+pub struct RegionBackendMapping {
```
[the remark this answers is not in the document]
serban300: `RegionBackendMapping` seems too generic to me because the backend is actually very specific here. Will every backend need this info ? For example the file backend doesn't need this structure.

src/vmm/src/persist.rs
dianpopa: Not really, no. What we are recommending is for the page fault handler process to make sure it kills the Firecracker process in case any error ensues while trying to handle page faults. In that respect, we are thinking of including the Firecracker pid in the UDS payload, so that the page fault handler can bring down the faulting thread with him.

src/vmm/src/vmm_config/snapshot.rs
```diff
@@ -41,23 +54,52 @@ pub struct CreateSnapshotParams {
...
+#[derive(Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct LoadSnapshotConfig {
+    /// Path to the file that contains the microVM state to be loaded.
+    pub snapshot_path: PathBuf,
+    /// Path to the file that contains the guest memory to be loaded. To be used only if
+    /// `mem_backend` is not specified.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub mem_file_path: Option<PathBuf>,
+    /// Guest memory backend configuration. Is not to be used in conjunction with `mem_file_path`.
```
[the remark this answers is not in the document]
dianpopa: We have a couple of validations here: https://github.com/firecracker-microvm/firecracker/pull/2921/files#diff-207be1a8aebc7e071e107acc561986d603444d4b51a595da0c5739fd861e0308R63.

src/vmm/src/persist.rs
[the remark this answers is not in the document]
serban300: Sounds good to me

src/vmm/src/persist.rs
```diff
@@ -487,7 +536,68 @@ fn guest_memory_from_file(
...
+            backend_mappings.as_bytes(),
+            // In the happy case we can close the fd since the other process has it open and is
+            // using it to serve us pages.
+            // The problem is that if other process crashes/exits, firecracker guest memory
+            // will simply revert to anon-mem behavior which would lead to silent errors and
+            // undefined behavior.
+            //
+            // If the fault serving process exits while we hold a copy of the FD as well, the
+            // uffd will still be alive but with no one to serve faults leading to guest freeze
+            // (which is an unfortunate, but explicit and defined behavior).
```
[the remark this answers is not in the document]
alsrdn: Yup I think this solves most of the issues. We will need to document this a little bit but I'm ok with the solution.

serban300 (approved):

PR #2925, opened 2022-03-14
Acpi poweroff feature

andreitraistaru:
# Reason for This PR

Support for ACPI shutdown added in Firecracker.

## Description of Changes

In order to implement a graceful shutdown of the guest, two features were implemented:

1. Send ACPI signal to the guest using fixed hardware method (according to the [spec](https://uefi.org/specs/ACPI/6.4/04_ACPI_Hardware_Specification/ACPI_Hardware_Specification.html?highlight=fixed%20hardware%20method#acpi-hardware-features)) when HTTP request is received by the API thread
2. Detection of the guest's transition to soft-off state (S5) in order to stop Firecracker process gracefully

## License Acceptance

## PR Checklist


src/devices/src/legacy/acpi_device.rs
```diff
@@ -0,0 +1,166 @@
...
+        const PWRBTN_STS_OFFSET: u64 = 8;
+        const PWRBTN_EN: u64 = 24;
+
+        // Set bits according to the spec in order to trigger the shutdown
+        // https://uefi.org/specs/ACPI/6.4/04_ACPI_Hardware_Specification/ACPI_Hardware_Specification.html#pm1-event-grouping
+        self.data = self.data | (1 << PWRBTN_STS_OFFSET) | (1 << PWRBTN_EN);
+
+        // Make OSPM read ACPI registers and generate the ACPI event by setting the interrupt on ACPI Device
+        self.interrupt
+            .write(9) // SCI_INT value from ACPI FADT table
```
acatangiu, line 95: the idiomatic way to trigger an eventfd is `write(1)` which can be easily mentally matched to `write(true)`.
While `write(9)` also works, it is confusing and can be misconstrued as sending 9 separate interrupts.

The irq number (9) for the raised IRQ tied to this eventfd is set during `register_irqfd()` and is not related to the amount written to the eventfd. That's why `write(1)` is enough, it says this eventfd got 1 event => Guest IRQ `#9` tied to it also gets 1 event.

(Also, `ACPI_SCI_INT` constant is defined in acpi crate, better to use constants than hardcode numbers)

src/devices/src/legacy/acpi_device.rs
```diff
@@ -0,0 +1,166 @@
...
+            value = value & (!(1 << SLP_EN_OFFSET));
+
+            // need to read bytes from given offset; bytes from 0 to offset - 1 should be skipped
+            value = value >> (8 * offset);
+
+            // read each byte from AcpiDevice
+            for i in 0..(data.len()) {
+                data[i] = (value % 256) as u8;
+                value = value >> 8;
+            }
```
acatangiu, line 126: This looks very specific to x86_64 endianness. While this is fine for current purpose of the PR, I would leave a warning comment for future readers of the code. Smth like
```
// warning: hardcoded little endian logic ahead.
```

(same for `fn write`)

src/devices/src/legacy/acpi_device.rs
```diff
@@ -0,0 +1,166 @@
...
+pub struct AcpiDevice {
+    /// Data Register
+    data: u64,
+    /// Interrupt to be triggered for OSPM to read ACPI registers
+    interrupt: EventFd,
+    /// CPU reset eventfd. We will set this event to stop Firecracker.
+    reset_evt: EventFd,
+}
+
+impl AcpiDevice {
```
acatangiu, line 75: I think device unit tests should be fairly easy to add for this device. Right now there's no tests.

acatangiu (commented):
Glad to see this PR up, very nice job!

Overall looking good, but I also left some ideas.

src/api_server/swagger/firecracker.yaml
```diff
@@ -874,6 +874,7 @@ definitions:
           - FlushMetrics
           - InstanceStart
           - SendCtrlAltDel
+          - SendACPIShutdown
```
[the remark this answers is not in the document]
andreitraistaru: Good to know. Thanks!

andreitraistaru:
Thank you so much for the comments you left above! I'll take each of them and improve the code accordingly!

src/devices/src/legacy/acpi_device.rs
andreitraistaru: It makes sense to be so... At the beginning, it was a bit unclear for me how interrupt triggering works ... and then I discovered register_irqfd() function that did the mapping between the irq number and eventfd. But I also kept write(9) thinking that it is still required, but the mapping was already done in register_irqfd(). Thank you so much for the explanation!

src/devices/src/legacy/acpi_device.rs
andreitraistaru: I improved this piece of code by using functions like from_le_bytes() and to_le_bytes() that deals with this transformation involving endianness in a more standardized way. Added a comment regarding this, as well.

src/acpi_tables/src/sdt.rs
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    pub fn update_checksum(&mut self) {
+        self.data[9] = 0;
+        let checksum = super::generate_checksum(self.data.as_slice());
+        self.data[9] = checksum
+    }
+
+    pub fn as_slice(&self) -> &[u8] {
+        &self.data.as_slice()
```
serban300, line 107: Maybe it would be better to call `update_checksum()` here instead of having to call it each time after `append`, `write` or any other function.  If we're sure that `as_slice()` will always be called when writing the table to memory. Anyway it would be nice to call `update_checksum()` from a single place.

src/acpi_tables/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+pub mod aml;
+pub mod rsdp;
+pub mod sdt;
+
+fn generate_checksum(data: &[u8]) -> u8 {
+    (255 - data.iter().fold(0u8, |acc, x| acc.wrapping_add(*x))).wrapping_add(1)
```
serban300, line 14: I would rename `acc` to `sum` and `x` to `data_byte` maybe.

src/vmm/src/acpi.rs
```diff
@@ -0,0 +1,172 @@
...
+fn create_dsdt_table(_cpu_count: u8) -> SDT {
+    // DSDT
+    let dsdt = SDT::new(*b"DSDT", 36, 6, *b"FIRECR", *b"FCDSDT  ", 1);
+
+    // TODO: add CPU devices
+
+    dsdt
+}
+
+fn create_madt(_cpu_count: u8) -> SDT {
```
serban300: I would create a separate structure for `Madt` and define this method on the struct if possible. Or at least I would create helper methods on that structure in order to avoid working directly with `SDT::append` and other low-level functions that also seem quite usafe.

The same comment for the other tables as well.

src/arch/src/x86_64/mptable.rs
```diff
@@ -97,8 +97,10 @@ const MPC_SPEC: i8 = 4;
 const MPC_OEM: [c_char; 8] = char_array!(c_char; 'F', 'C', ' ', ' ', ' ', ' ', ' ', ' ');
 const MPC_PRODUCT_ID: [c_char; 12] = ['0' as c_char; 12];
 const BUS_TYPE_ISA: [u8; 6] = char_array!(u8; 'I', 'S', 'A', ' ', ' ', ' ');
-const IO_APIC_DEFAULT_PHYS_BASE: u32 = 0xfec0_0000; // source: linux/arch/x86/include/asm/apicdef.h
-const APIC_DEFAULT_PHYS_BASE: u32 = 0xfee0_0000; // source: linux/arch/x86/include/asm/apicdef.h
+/// IOAPIC default phys addr; source: linux/arch/x86/include/asm/apicdef.h
```
serban300, line 100: Do we still need `mptable.rs` now ? After switching to ACPI ?

src/acpi_tables/src/aml.rs
```diff
@@ -0,0 +1,1988 @@
+// Copyright © 2019 Intel Corporation
```
serban300: Right now if I understand correctly we need very few lines of aml language. Less than 5 maybe ? Considering this I don't know how much it's worth having `aml.rs` which has 2000 lines defining helper functions in order to work with AML. Could we just hardcode a couple of lines of AML and remove this file ? Do we expect to use much more AML code in the future ?

andreitraistaru:
Closing this in favor of https://github.com/firecracker-microvm/firecracker/pull/3190

PR #2947, opened 2022-04-04
Don't explicitly send FC's PID to the page fault handler

luminitavoicu:

# Reason for This PR

The page fault handler process requires FC's PID in order
to be able to notify FC of any crashes/exists. The way
we initially intended to make the handler process aware
of FC's PID was to have FC sent its PID explicitly through
the UDS. But there is a problem with this approach:

Firecracker's PID depends on the way the process was started.
When normally started through the jailer, Firecracker will
see its global PID. However, if using `--new-pid-ns` flag
when running the jailer, the jailer will spawn FC in a new
PID namespace, which would cause Firecracker to not be
aware of its PID and see it as being 1.

Moreover, when not using `--new-pid-ns` flag, we recommend
to the users to spawn the jailer process into a new PID
namespace when starting it.

This is why sending FC's PID from whithin the FC process is
not reliable.

To tackle this, we no longer send FC's PID explicitly,
but advise the page fault handler process to use
`getsockopt` call with `SO_PEERCRED` option in order
to fetch the global Firecracker PID at the moment of
connecting. Per [pid_namespaces man page](https://man7.org/linux/man-pages/man7/pid_namespaces.7.html):

> When a process ID is passed over a UNIX domain socket to a
   process in a different PID namespace (see the description of
   SCM_CREDENTIALS in [unix(7)](https://man7.org/linux/man-pages/man7/unix.7.html)), it is translated into the
   corresponding PID value in the receiving process's PID namespace.

## Description of Changes

Don't explicitly send FC's PID to the page fault handler process.

## License Acceptance

## PR Checklist

- ~[ ] The issue which led to this PR has a clear conclusion.~
- ~[ ] Any user-facing changes are mentioned in `CHANGELOG.md`.~

### DIFF
--- a/src/vmm/src/persist.rs
+++ b/src/vmm/src/persist.rs
@@ -8,7 +8,6 @@ use std::fs::{File, OpenOptions};
 use std::io::{self, Write};
 use std::os::unix::{io::AsRawFd, net::UnixStream};
 use std::path::Path;
-use std::process;
 use std::sync::{Arc, Mutex};
 
 use crate::builder::{self, StartMicrovmError};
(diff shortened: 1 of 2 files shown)


dianpopa (approved):
This is awesome! We need to make sure we include the example from the comment in the upcoming handler example

alindima (approved):

PR #2949, opened 2022-04-07
Uffd integration tests

luminitavoicu:
# Reason for This PR

UFFD-related logic lacks integration testing.

## Description of Changes

Modify existing test framework to reflect the new snapshot load API and add tests for user fault fd functionality.

## License Acceptance

## PR Checklist


tests/framework/builder.py
```diff
@@ -210,10 +216,28 @@ def build_from_snapshot(self,
                                          guest_ip=iface.guest_ip,
                                          netmask_len=iface.netmask,
                                          tapname=iface.tap_name)
-        response = vm.snapshot.load(mem_file_path=jailed_mem,
-                                    snapshot_path=jailed_vmstate,
-                                    diff=diff_snapshots,
-                                    resume=resume)
+        if utils.compare_versions(vm.firecracker_version, '1.0.0') >= 0:
```
dianpopa, line 221: Here we need to do `utils.compare_versions(vm.firecracker_version, '1.0.0') > 0`, right?
Now one possible solution would be to tag Firecracker to point to the next release which has not proven to be a very accurate way (see v.0.26.0).
One other option would be to modify the `compare_versions` to be able to check dirty versions (i.e the ones that are not actual releases).
Then we could use `test_microvm.version.get_from_api()` to get the full firecracker version (including the dirty info).
Something like:
```
    is_first_dirty = False; is_second_dirty = False

    # code to see if any of the versions contain dirty info  (i.e if anything follows X.Y.Z)
    # we should not allow for both versions to be dirty; not sure if the dirty info reveals order info
    .....
    if is_first_dirty and is_second_dirty:
        raise ValueError
    if is_first_dirty:
        return 1
    if is_second_dirty:
        return -1
    return 0
```

dianpopa: Cool, we can go like this

tests/host_tools/uffd/src/bin/valid_handler.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    creds
+}
+
+fn create_mem_regions(mappings: &Vec<GuestRegionUffdMapping>) -> Vec<MemRegion> {
+    let page_size = get_page_size().unwrap();
+    let mut mem_regions: Vec<MemRegion> = Vec::with_capacity(mappings.len());
+
+    for (_, r) in mappings.iter().enumerate() {
```
alindima, line 242: if we ignore the key, just call `iter()`

tests/host_tools/uffd/src/bin/valid_handler.rs
```diff
@@ -0,0 +1,289 @@
...
+        let mapping = r.clone();
+        let mut addr = r.base_host_virt_addr;
+        let end_addr = r.base_host_virt_addr + r.size as u64;
+        let mut page_states = HashMap::new();
+
+        while addr < end_addr {
+            page_states.insert(addr, MemPageState::Uninitialized);
+            addr += page_size as u64;
+        }
```
alindima, line 251: nit:
could use smth like:

```
(addr..(end_addr + 1)).step_by(page_size).for_each(|addr| page_states.insert(addr, MemPageState::Uninitialized))
```

tests/host_tools/uffd/src/bin/valid_handler.rs
```diff
@@ -0,0 +1,289 @@
...
+}
+
+fn main() {
+    let uffd_sock_path = std::env::args().nth(1).expect("No socket path given");
+    let mem_file_path = std::env::args().nth(2).expect("No memory file given");
+
+    let file = File::open(mem_file_path).expect("Cannot open memfile");
+    let size = file.metadata().unwrap().len() as usize;
+
+    // Create a page that will be copied into the faulting region
```
alindima, line 268: is it an outdated comment?

tests/framework/builder.py
```diff
@@ -218,11 +220,9 @@ def build_from_snapshot(self,
                                          tapname=iface.tap_name)
         if utils.compare_versions(vm.firecracker_version, '1.0.0') > 0:
             if uffd_path:
-                jailed_uffd = vm.copy_to_jail_ramfs(uffd_path) if use_ramdisk \
-                    else vm.create_jailed_resource(uffd_path)
```
alindima: why did we remove this? don't we need the uds path to be inside the jail?

tests/integration_tests/functional/test_uffd.py
```diff
@@ -100,3 +105,70 @@ def test_unbinded_socket(bin_cloner_path, test_microvm_with_api):
...
+    os.chroot(".")
+
+    vm, _ = vm_builder.build_from_snapshot(snapshot, vm=vm,
+                                           resume=True,
+                                           uffd_path=socket_path)
+
+    ssh_connection = net_tools.SSHConnection(vm.ssh_config)
+
+    # Verify if guest can run commands.
+    exit_code, _, _ = ssh_connection.execute_command("sync")
```
alindima, line 173: should we also test for the ballooning scenario?

tests/integration_tests/functional/test_uffd.py
```diff
@@ -100,3 +105,70 @@ def test_unbinded_socket(bin_cloner_path, test_microvm_with_api):
...
+
+    real_root = os.open("/", os.O_RDONLY)
+
+    os.chroot(vm.chroot())
+    os.chdir('/')
+    st = os.stat("valid_handler")
+    os.chmod("valid_handler", st.st_mode | stat.S_IEXEC)
+    socket_path = "/firecracker-uffd.sock"
+
+    proc = Popen(
```
alindima, line 146: do we ever kill this process?

tests/integration_tests/functional/test_uffd.py
```diff
@@ -100,3 +105,70 @@ def test_unbinded_socket(bin_cloner_path, test_microvm_with_api):
...
+        stdout=PIPE,
+        stderr=PIPE,
+        bufsize=1)
+    try:
+        outs, errs = proc.communicate(timeout=1)
+        print(outs)
+        print(errs)
+        assert False, "Could not start PF handler!"
+    except TimeoutExpired:
+        print("This is the good case!")
```
alindima, line 157: I'd like to have a helper function doing this setup and launching the handler process, including chroot-ing in and out of the vm jailed root

tests/integration_tests/functional/test_uffd.py
```diff
@@ -100,3 +105,70 @@ def test_unbinded_socket(bin_cloner_path, test_microvm_with_api):
...
+
+    jailed_mem = vm.create_jailed_resource(snapshot.mem)
+
+    # Copy page fault binary into chroot of microVM.
+    jailed_handler = os.path.join(vm.chroot(), "valid_handler")
+    shutil.copyfile(uffd_handler_paths['valid_handler'], jailed_handler)
+    # Change its permissions to the UID and GID of the jail.
+    cmd = 'chown {}:{} {}'.format(
+        vm.jailer.uid, vm.jailer.gid, jailed_handler)
+    run_cmd(cmd)
```
alindima, line 136: wouldn't this be done by the removed `vm.create_jailed_resource(uffd_path)`?

tests/framework/microvm.py
```diff
@@ -552,52 +551,21 @@ def spawn(self, create_logger=True,
...
             # This file will collect any output from 'screen'ed Firecracker.
             self._screen_log = self.SCREEN_LOGFILE.format(self._session_name)
-            start_cmd = 'screen -L -Logfile {logfile} '\
-                        '-dmS {session} {binary} {params}'
-            start_cmd = start_cmd.format(
-                logfile=self.screen_log,
-                session=self._session_name,
-                binary=self._jailer_binary_path,
-                params=' '.join(jailer_param_list)
+            screen_pid, binary_pid = utils.start_screen_process(
```
alindima, line 562: 👍🏻

tests/conftest.py
```diff
@@ -394,6 +394,40 @@ def bin_seccomp_paths(test_fc_session_root_path):
...
+
+    valid_handler = os.path.normpath(
+        os.path.join(
+            release_binaries_path,
+            'valid_handler'
+        )
+    )
+
+    yield {
+        'valid_handler': valid_handler,
```
alindima, line 427: do we expect to add invalid/other handlers?

tests/host_tools/uffd/src/bin/valid_handler.rs
```diff
@@ -0,0 +1,289 @@
...
+                // 3. Removed - MADV_DONTNEED was called due to balloon inflation
+                // 4. Anonymous - page was zeroed out -> this implies that more than one page fault
+                //    event was received. This can be a consequence of guest reclaiming back its
+                //    memory from the host (through balloon device)
+                Some(MemPageState::Uninitialized) => {
+                    let (start, end) = self.populate_from_file(region);
+                    self.update_mem_state_mappings(start, end, &MemPageState::FromFile);
+                    return;
+                }
+                Some(MemPageState::FromFile)
```
alindima, line 157: so if we get a page fault for a page that we've already faulted in from the file, we will zero it?
shouldn't it be a no-op?

tests/host_tools/uffd/src/bin/valid_handler.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn zero_out(&mut self, addr: u64) -> (u64, u64) {
+        let page_size = get_page_size().unwrap();
+
+        let ret = unsafe {
+            self.uffd
+                .zeropage(addr as *mut _, page_size, true)
+                .expect("Uffd zeropage failed")
+        };
+        // Make sure the UFFD zeroed out some bytes.
+        assert!(ret > 0);
```
alindima, line 129: what if it was already all zeroed out?

tests/host_tools/uffd/src/bin/valid_handler.rs
```diff
@@ -0,0 +1,289 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
alindima, line 3: I'd add a general comment about what this handler does (servicing pages one-by-one as requested)

tests/host_tools/uffd/src/bin/valid_handler.rs
```diff
@@ -0,0 +1,289 @@
...
+        for region in self.mem_regions.iter() {
+            match region.page_states.get(&fault_page_addr) {
+                // Our simple PF handler has a simple strategy:
+                // There exist 4 states in which a memory page can be in:
+                // 1. Uninitialized - page was never touched
+                // 2. FromFile - the page is populated with content from snapshotted memory file
+                // 3. Removed - MADV_DONTNEED was called due to balloon inflation
+                // 4. Anonymous - page was zeroed out -> this implies that more than one page fault
+                //    event was received. This can be a consequence of guest reclaiming back its
+                //    memory from the host (through balloon device)
```
alindima, line 151: If I read the code correctly, for Anonymous we simply zero out the page instead of re-populating it from the file. Is that the expected behaviour?

tests/host_tools/uffd/src/bin/valid_handler.rs
luminitavoicu: Looking at the [uffd kernel code](https://elixir.bootlin.com/linux/v4.8/source/fs/userfaultfd.c#L1101), the kernel does not care for the contents of the page before zeroing. The `zeropage()` function returns 1 if operation succeeded (`pagesize` zeros were written at the specified address) and -1 in case an error occurs.

tests/host_tools/uffd/src/bin/valid_handler.rs
luminitavoicu: Yes. So at first, all pages are marked as `Uninitialized` when fetching the guest region mappings. When a page fault occurs, the page is brought into memory only when not initialized. 

The pages that are in the `Anonymous` state are zeroed out and have been page faulted before. This happens if the balloon device is enabled and the memory area has been claimed by the host. When having the balloon device present, the PF handler can also receive `Remove` events, alongside page faults. This is because when the balloon device asks for removal of a memory range, Firecracker will call the `madvise(MADV_DONTNEED)` system call in order to let the kernel know that it can free up memory found in that specific range. On such a system call, the UFFD interface sends the `Remove` event, but the area is still monitored by the UFFD. We handle such an event by setting the affected pages to `Removed` state. This way we know when we receive a subsequent page fault on a page that has been claimed by the host, so that we can zero it out (instead of bringing it again from the file) - and mark it as `Anonymous`. 

This is why page faults on `Anonymous` pages should not be brought into memory from the file.

This proposed mitigation is also covered by the UFFD documentation (https://www.kernel.org/doc/html/latest/admin-guide/mm/userfaultfd.html#non-cooperative-userfaultfd): 
> Although the `UFFD_FEATURE_EVENT_REMOVE` and `UFFD_FEATURE_EVENT_UNMAP `are pretty similar, they quite differ in the action expected from the userfaultfd manager. In the former case, the virtual memory is removed, but the area is not, the area remains monitored by the userfaultfd, and if a page fault occurs in that area it will be delivered to the manager. The proper resolution for such page fault is to zeromap the faulting address.

tests/framework/builder.py
luminitavoicu: Yes, we do need to have the UDS inside the jail, but it must be already present in the jail before issuing the command to build the microvm from snapshot, because the page fault handler needs to bind it and accept incoming connections.

tests/integration_tests/functional/test_uffd.py
luminitavoicu: See comment above. Also, this snippet copies the page fault handler binary into the jailer, not the UDS.

tests/integration_tests/functional/test_uffd.py
luminitavoicu: To be honest, I am not sure how easy it would be to make the reproduce the balloon device scenario in our testing framework.

tests/integration_tests/functional/test_uffd.py
luminitavoicu: Done and good catch!

tests/integration_tests/functional/test_uffd.py
luminitavoicu: Done

tests/conftest.py
luminitavoicu: Yes, added malicious handler.

tests/host_tools/uffd/src/bin/valid_handler.rs
luminitavoicu: I don't think it matters much how we handle `MemPageState::FromFile` pages in our case, since our test does not trigger this path (page faulting a memory region that has been previously loaded into memory).

tests/framework/builder.py
luminitavoicu: I ended up adding this in this PR after all, because when using binaries built from the local firecracker version, `git describe` is not aware of the version bump (it looks at the latest tag).

tests/integration_tests/functional/test_uffd.py
alindima: wouldn't it simply be a balloon inflate -> wait -> balloon deflate?
just to see that the uffd events are handled correctly

tests/integration_tests/functional/test_uffd.py
```diff
@@ -0,0 +1,238 @@
...
+                                           uffd_path=SOCKET_PATH)
+
+    ssh_connection = net_tools.SSHConnection(vm.ssh_config)
+
+    # Verify if guest can run commands.
+    exit_code, _, _ = ssh_connection.execute_command("sync")
+    assert exit_code == 0
+
+    # Kill page fault handler process.
+    pf_handler_process.kill()
```
alindima, line 187: If any previous assert fails, it will not be killed. We could have a PfHandler class that kills the process when going out of scope.

tests/integration_tests/functional/test_uffd.py
```diff
@@ -0,0 +1,238 @@
...
+    vm_builder = MicrovmBuilder(bin_cloner_path)
+    vm = test_microvm_with_api
+    vm.spawn()
+
+    # Copy snapshot memory file into chroot of microVM.
+    jailed_mem = vm.create_jailed_resource(snapshot.mem)
+    # Copy the valid page fault binary into chroot of microVM.
+    jailed_handler = vm.create_jailed_resource(
+        uffd_handler_paths['valid_handler']
+    )
```
alindima, line 171: could we also include this in `spawn_pf_handler`? Just so that it's less boilerplate code when adding a new test

tests/integration_tests/functional/test_uffd.py
```diff
@@ -0,0 +1,238 @@
...
+    # Copy the valid page fault binary into chroot of microVM.
+    jailed_handler = vm.create_jailed_resource(
+        uffd_handler_paths['malicious_handler']
+    )
+
+    # Spawn page fault handler process.
+    _ = spawn_pf_handler(vm, jailed_handler, jailed_mem)
+
+    # We expect Firecracker to freeze while resuming from a snapshot
+    # due tu the malicious handler's unavailability.
```
alindima, line 225:
```suggestion
    # due to the malicious handler's unavailability.
```

tests/host_tools/uffd/src/bin/valid_handler.rs
alindima: what if we page in a region from a file, and the guest OS swaps it to disk. The guest then wants to access it again and it's not in memory. How will that be handled?

alindima:
I'm afraid you need to have this commit on the uffd branch in order for the AMD tests to pass: https://github.com/firecracker-microvm/firecracker/commit/b4c0ffd9c2375ca2ecc00afb4553320afabb5254

tests/host_tools/uffd/src/bin/valid_handler.rs
alindima: I see, so if we deflate the balloon and get some pages back to the guest, those pages should be zeroed

tests/framework/builder.py
```diff
@@ -210,10 +219,31 @@ def build_from_snapshot(self,
...
+                    'type': SnapshotMemBackendType.FILE,
+                    'path': jailed_mem
+                }
+            response = vm.snapshot.load(mem_backend=mem_backend,
+                                        snapshot_path=jailed_vmstate,
+                                        diff=diff_snapshots,
+                                        resume=resume,
+                                        timeout=timeout)
+        else:
+            response = vm.snapshot.load(mem_file_path=jailed_mem,
```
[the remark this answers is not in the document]
luminitavoicu: Yes.

tests/host_tools/uffd/src/bin/valid_handler.rs
luminitavoicu: No-op would not help in this scenario. If the guest swaps a page that is in the `FromFile` state, and then a page fault occurs on that page, the UFFD PF handler should load the page into memory again.

Our test case does not trigger this scenario, but I will update the handler accordingly for completeness.

tests/integration_tests/functional/test_uffd.py
luminitavoicu: We chatted offline. 

It's tricky to fully control what memory regions to touch in order to trigger page faults for the pages that have been claimed by the host after a balloon inflate (and ensure they are zeroed out as expected). However, we could try to identify receiving a `UFFD_EVENT_REMOVE` after a balloon inflate. 

I will update the test.

tests/framework/artifacts.py
```diff
@@ -304,7 +304,7 @@ class ArtifactCollection:
     PLATFORM = platform.machine()
 
     # S3 bucket structure.
-    ARTIFACTS_ROOT = 'ci-artifacts'
+    ARTIFACTS_ROOT = 'ci-artifacts-dev'
```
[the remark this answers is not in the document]
alindima: I know that @bchalios also plans on modifying the binaries on the main branch. Should we have a `ci-artifacts-uffd` folder on this branch and revert to `ci-artifacts` when merging on main?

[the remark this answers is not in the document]
alindima: That's what we've done for io_uring in the past

alindima (approved):
LGTM 🚀

dianpopa (approved):

Issue #2908, opened 2022-02-19
Firecracker panicked at 'Error creating the Kvm object: Error(13)', src/vmm/src/vstate/system.rs:53:30


bootrino:
I'm following this tutorial but I am also trying to do networking:
https://aws.amazon.com/blogs/aws/firecracker-lightweight-virtualization-for-serverless-computing/

To do networking, I added the script below: **3_config_guest_network.sh**

When I follow the tutorial without script **3_config_guest_network.sh** the virtual machine starts OK.

When I run the script below **4_start_guest_machine.sh** firecracker crashes with:
```
@homecloud:~$ ./firecracker --api-sock /tmp/firecracker.socket
2022-02-19T16:10:21.775601118 [anonymous-instance:main:ERROR:src/firecracker/src/main.rs:91] Firecracker panicked at 'Error creating the Kvm object: Error(13)', src/vmm/src/vstate/system.rs:53:30
Aborted
@homecloud:~$
```

**1_set_guest_kernel.sh**
```
arch=`uname -m`
kernel_path=$(pwd)"/hello-vmlinux.bin"

if [ ${arch} = "x86_64" ]; then
    curl --unix-socket /tmp/firecracker.socket -i \
      -X PUT 'http://localhost/boot-source'   \
      -H 'Accept: application/json'           \
      -H 'Content-Type: application/json'     \
      -d "{
            \"kernel_image_path\": \"${kernel_path}\",
            \"boot_args\": \"console=ttyS0 reboot=k panic=1 pci=off\"
       }"
elif [ ${arch} = "aarch64" ]; then
    curl --unix-socket /tmp/firecracker.socket -i \
      -X PUT 'http://localhost/boot-source'   \
      -H 'Accept: application/json'           \
      -H 'Content-Type: application/json'     \
      -d "{
            \"kernel_image_path\": \"${kernel_path}\",
            \"boot_args\": \"keep_bootcon console=ttyS0 reboot=k panic=1 pci=off\"
       }"
else
    echo "Cannot run firecracker on $arch architecture!"
    exit 1
fi
```

**2_set_root_guestfs.sh**
```
rootfs_path=$(pwd)"/hello-rootfs.ext4"
curl --unix-socket /tmp/firecracker.socket -i \
  -X PUT 'http://localhost/drives/rootfs' \
  -H 'Accept: application/json'           \
  -H 'Content-Type: application/json'     \
  -d "{
        \"drive_id\": \"rootfs\",
[...]

PR #2983, opened 2022-05-02
Clarify FC panic error message when missing user ACL for KVM device

mattschlebusch:
# Reason for This PR

[Issue #2908](https://github.com/firecracker-microvm/firecracker/issues/2908) shows an example where configuring a network tap for a microVM run by a user without access to KVM will cause the firecracker process to close and returns an error that is quite vague.

Some potential options: 
1) Validate the user's access to the `/dev/kvm` device prior to attempting to fulfill the `InstanceStart` request. This would leave the firecracker process running and the API call could be re-attempted after configuring the file's ACL.
1) Add some detail to the error logged in response to the failed initialization of the `/dev/kvm` device.

Unfortunately the first option would likely require a system call which be executed for every `InstanceStart` API call (or possibly on the firecracker process' launch). The value of not having to restart the process to accommodate a configuration issue does not seem to warrant slowing down the launch process.

### Before
```
$ firecracker  --api-sock /tmp/firecracker.socket
2022-05-02T14:00:53.524156161 [anonymous-instance:main:ERROR:src/firecracker/src/main.rs:90] Firecracker panicked at 'Error creating the Kvm object: Error(13)', src/vmm/src/vstate/system.rs:53:30
Aborted
```

### After
```
$ firecracker --api-sock /tmp/firecracker.socket
2022-05-05T12:41:34.181549756 [anonymous-instance:main:ERROR:src/firecracker/src/main.rs:90] Firecracker panicked at 'Error creating KVM object. [Permission denied (os error 13)]
Make sure the user launching the firecracker process is configured on the /dev/kvm file's ACL.', src/vmm/src/vstate/system.rs:69:13
./launch-fc-process.sh: line 6: 36845 Aborted                 /home/ec2-user/workspace/fc-mattschlebusch/build/cargo_target/x86_64-unknown-linux-musl/debug/firecracker --api-sock /tmp/firecracker.socket
[ec2-user@ip-172-31-15-233 fc-sandbox]$ ./launch-fc-process.sh 
+ rm -rf /tmp/firecracker.socket
+ /home/ec2-user/workspace/fc-mattschlebusch/build/cargo_target/x86_64-unknown-linux-musl/debug/firecracker --api-sock /tmp/firecracker.socket
2022-05-05T12:52:20.941678838 [anonymous-instance:fc_api:ERROR:src/api_server/src/parsed_request.rs:192] Received Error. Status code: 400 Bad Request. Message: Internal error while starting microVM: Failed to validate KVM support: Error creating KVM object. [Permission denied (os error 13)]
Make sure the user launching the firecracker process is configured on the /dev/kvm file's ACL.

```
(Take note that the firecracker process does not terminate)

## License Acceptance

### DIFF
--- a/src/api_server/src/request/net.rs
+++ b/src/api_server/src/request/net.rs
@@ -27,7 +27,11 @@ pub(crate) fn parse_put_net(
         METRICS.put_api_requests.network_fails.inc();
         return Err(Error::Generic(
             StatusCode::BadRequest,
-            "The id from the path does not match the id from the body!".to_string(),
+            format!(
+                "The id from the path [{}] does not match the id from the body [{}]!",
+                id,
+                netif.iface_id.as_str()
+            ),
         ));
     }
     Ok(ParsedRequest::new_sync(VmmAction::InsertNetworkDevice(
@@ -56,7 +60,11 @@ pub(crate) fn parse_patch_net(
         METRICS.patch_api_requests.network_count.inc();
         return Err(Error::Generic(
             StatusCode::BadRequest,
-            "The id from the path does not match the id from the body!".to_string(),
+            format!(
+                "The id from the path [{}] does not match the id from the body [{}]!",
+                id,
+                netif.iface_id.as_str()
+            ),
         ));
     }
     Ok(ParsedRequest::new_sync(VmmAction::UpdateNetworkInterface(
--- a/src/vmm/src/vstate/system.rs
+++ b/src/vmm/src/vstate/system.rs
@@ -11,7 +11,7 @@ use std::{
 };
 
 use kvm_bindings::KVM_API_VERSION;
-use kvm_ioctls::Kvm;
+use kvm_ioctls::{Error as KvmIoctlsError, Kvm};
 
 /// Errors associated with the wrappers over KVM ioctls.
 #[derive(Debug)]
@@ -20,6 +20,8 @@ pub enum Error {
     KvmApiVersion(i32),
     /// Cannot initialize the KVM context due to missing capabilities.
     KvmCap(kvm_ioctls::Cap),
+    /// Cannot initialize the KVM context.
+    KvmInit(KvmIoctlsError),
 }
 
 impl Display for Error {
@@ -33,6 +35,19 @@ impl Display for Error {
                 v
             ),
             KvmCap(cap) => write!(f, "Missing KVM capabilities: {:?}", cap),
+            KvmInit(err) => {
+                if err.errno() == libc::EACCES {
+                    write!(
+                        f,
+                        "Error creating KVM object. [{}]\n\
+                         Make sure the user launching the firecracker process \
+                         is configured on the /dev/kvm file's ACL.",
+                        err
+                    )
+                } else {
+                    write!(f, "Error creating KVM object. [{}]", err)
+                }
+            }
         }
     }
 }
@@ -50,7 +65,7 @@ pub struct KvmContext {
 impl KvmContext {
     pub fn new() -> Result<Self> {
         use kvm_ioctls::Cap::*;
-        let kvm = Kvm::new().expect("Error creating the Kvm object");
+        let kvm = Kvm::new().map_err(Error::KvmInit)?;
 
         // Check that KVM has the correct version.
         if kvm.get_api_version() != KVM_API_VERSION as i32 {
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -29,9 +29,9 @@ from host_tools import proc
 # Checkout the cpuid crate. In the future other
 # differences may appear.
 if utils.is_io_uring_supported():
-    COVERAGE_DICT = {"Intel": 84.89, "AMD": 84.38, "ARM": 83.96}
+    COVERAGE_DICT = {"Intel": 84.82, "AMD": 84.32, "ARM": 83.91}
 else:
-    COVERAGE_DICT = {"Intel": 81.94, "AMD": 81.43, "ARM": 80.96}
+    COVERAGE_DICT = {"Intel": 81.86, "AMD": 81.36, "ARM": 80.90}
 
 PROC_MODEL = proc.proc_type()
 


raduiliescu:
Can you please rename this, so that we can get an idea of the changes from the title? Maybe something like Better log message for KVM insertion error due to permissions. This way when one looks on the PR history can quickly understand the changes.

src/vmm/src/vstate/system.rs
```diff
@@ -50,7 +50,12 @@ pub struct KvmContext {
 impl KvmContext {
     pub fn new() -> Result<Self> {
         use kvm_ioctls::Cap::*;
-        let kvm = Kvm::new().expect("Error creating the Kvm object");
+        let kvm = Kvm::new().expect(
+            "Critical error creating the KVM object. \
```
dianpopa, line 54: Not sure why we expect here and not gracefully handle the error (not having the right capabilities isn t less bad than this). So I suggest doing a `map_err()?`instead of `expect`. 
If we do that, we can also pretty print the `Error (13)` cryptical message.

mattschlebusch: Thanks for the suggestion. I will make the change but clarify this would permit firecracker to be used with limited features with the user configuration as is.

For addressing the immediate issue raised this change seems like a reasonable improvement. But it isn't clear to me yet how KVM permits memory and vCPU usage but it fails with network configuration, even without the ACL configured for `/dev/kvm`. Are there other spots where this missing ACL could potentially cause a "panic"?

dianpopa: I am not sure I understand what you mean by KVM failing with network configuration? I agree that this is a reasonable improvement and we can go ahead with current version

src/api_server/src/request/net.rs
```diff
@@ -27,7 +27,14 @@ pub(crate) fn parse_put_net(
...
             StatusCode::BadRequest,
-            "The id from the path does not match the id from the body!".to_string(),
+            format!(
+                "{}{}{}{}{}",
+                "The id from the path [",
+                id,
+                "] does not match the id from the body [",
+                netif.iface_id.as_str(),
+                "]!"
+            ),
```
serban300, line 37: Nit: I would write this as:

```
            format!(
                "The id from the path [{}] does not match the id from the body [{}] !",
                id,
                netif.iface_id.as_str()
            )
```

src/vmm/src/vstate/system.rs
mattschlebusch: > I am not sure I understand what you mean by KVM failing with network configuration?

To reproduce the [issue](https://github.com/firecracker-microvm/firecracker/issues/2908), I had to (via the API):
1) Configure the kernel
1) Configure the file system
1) *Configure the network (tap) device for the guest*
1) Launch the guest instance

If the step to configure the network device is omitted, the instance will launch successfully. (at least in my initial attempt to duplicate the issue).

Unfortunately using a `map_err()` will leave the kvm reference that is supposed to be created will be null. I'm inclined to use `express()` for now to force the exit of the process at this stage. At the very least the user will get a clearer message as to what is wrong.

alindima: > If the step to configure the network device is omitted, the instance will launch successfully. (at least in my initial attempt to duplicate the issue).

Can you retry this? It doesn't have something to do with the network device.
I remember I bumped into this error even without a tap device configured

alindima: I suggest you add a new variant for the `Error` enum in `src/vmm/vstate/system.rs`, like `KvmCreateVm(errno::Error)` and implement the `Display` trait for it such that you output the desired error message, together with the wrapped error.
Also do a check to see whether the `Debug` or `Display` implementations are used higher up in the calling chain

dianpopa: The issue is not related to network in any way. You can replicate it by running Firecracker with a user that does not have access to `/dev/kvm`. The `Error(13)` is debug code for "Permission Denied" error.

dianpopa: > I suggest you add a new variant for the `Error` enum in `src/vmm/vstate/system.rs`, like `KvmCreateVm(errno::Error)` and implement the `Display` trait for it such that you output the desired error message, together with the wrapped error. Also do a check to see whether the `Debug` or `Display` implementations are used higher up in the calling chain

Since we also want to force process exit and also pretty print the kvm_ioctls error, we can use:
`option.unwrap_or_else(|e| panic!("Cannot create KVM object: {}", e))`

mattschlebusch: > If the step to configure the network device is omitted, the instance will launch successfully. (at least in my initial attempt to duplicate the issue).
>> Can you retry this? It doesn't have something to do with the network device.
You are correct. While I have got two results on two machines, I have confirmed at least once to be a problem that was not net config specific.

While the forced process exit works, I think defining a new error enum entry with a tidy formatted display is reasonable in the context of this PR.

alindima: > Since we also want to force process exit and also pretty print the kvm_ioctls error, we can use:
option.unwrap_or_else(|e| panic!("Cannot create KVM object: {}", e))

Do we neccessarily want to force process exits?
If we were to use regular error propagation using `?`, it would either result in an API error for the `InstanceStart` call or a process exit if Firecracker was started without an API thread.

I think this is perfectly ok (and preferred actually), since a user may then add the right permission to `/dev/kvm` and retry the API call.

@dianpopa @mattschlebusch WDYT?

mattschlebusch: My current thinking here is that it is possible that permissions to KVM are organized by a user group instead of the ACL in which case the FC process would need to get restarted anyway when the user starts a new session for the group's membership to take effect. An argument could be made either way though.

src/vmm/src/vstate/system.rs
```diff
@@ -11,7 +11,7 @@ use std::{
 };
 
 use kvm_bindings::KVM_API_VERSION;
-use kvm_ioctls::Kvm;
+use kvm_ioctls::{Error as kvm_ioctls_err, Kvm};
```
alindima, line 14: The Rust convention is to have data types in CamelCase, like: `KvmIoctlsError`

src/vmm/src/vstate/system.rs
```diff
@@ -50,7 +65,9 @@ pub struct KvmContext {
 impl KvmContext {
     pub fn new() -> Result<Self> {
         use kvm_ioctls::Cap::*;
-        let kvm = Kvm::new().expect("Error creating the Kvm object");
+        let kvm = Kvm::new().unwrap_or_else(|e| {
```
alindima, line 68: as discussed, let's do a: `Kvm::new().map_err(Error::KvmInit)?;`

Please also confirm by running manually that the right error message is outputted

src/vmm/src/vstate/system.rs
mattschlebusch: If I can interpret the principal that you state about not closing the process and the customer is in control of the process makes sense. I'm happy to use that as the metric for the decision.

src/vmm/src/vstate/system.rs
mattschlebusch: I have indeed run it and update the overview for the PR as well. I will need to re-test once as soon as I make the discussed changes.

src/vmm/src/vstate/system.rs
mattschlebusch: Good to know thanks.

dianpopa (approved):

dianpopa (dismissed):

serban300 (approved):
Looks good to me !

dianpopa (approved):

PR #2996, opened 2022-05-13
Fixing loosely define behaviour within `write_log`

JonathanWoollett-Light:
# Reason for This PR

The function `Logger::write_log`:

```rust
/// The `write_log` method takes care of the common logic involved in writing
/// regular log messages.
fn write_log(&self, mut msg: String, msg_level: Level) {
    let mut guard;
    let mut dest: Box<dyn Write + Send> = if self.init.is_initialized() {
        guard = extract_guard(self.log_buf.lock());
        Box::new(guard.as_mut())
    } else {
        match msg_level {
            Level::Error | Level::Warn => Box::new(stderr()),
            _ => Box::new(stdout()),
        }
    };

    // No need to explicitly call flush because the underlying LineWriter flushes
    // automatically whenever a newline is detected (and we always end with a
    // newline the current write).
    msg.push('\n');
    if dest.write_all(msg.as_bytes()).is_err() {
        // No reason to log the error to stderr here, just increment the metric.
        METRICS.logger.missed_log_count.inc();
    }
}
```
This allows `dest` to be any type which implements Write and Send notably a type implementing Write which does not necessarily call .flush() when it detects a newline. Thus in an instance where `dest` is not `LineWriter` the comment and assumptions that:
```
// No need to explicitly call flush because the underlying LineWriter flushes
// automatically whenever a newline is detected (and we always end with a
// newline the current write).
```
Are false and could lead to unexpected behavior. Possibly a system may call  Logger::write_log then crash/exit and not log the messages near this crash/exit.

Possible fixes may be:

1. Restrict the type of `dest` to `Box<LineWriter>`
2. Simply call `flush()`.
3. Implement a new trait extending Write which defines a writer which flushes on newline:
```rust
pub trait WriteLine: Write {
    /// Functions similar to [std::io::Write::write_all] except calls 
    /// [std::io::Write::flush] when a b'/n' is present in the given buffer.
    fn writeln(&mut self, buf: &[u8]) -> std::io::Result<()>;
}
```
Given my assumptions about usage I would suggest the 2nd option here to be the best approach.

## Description of Changes

Added `.flush()` call and reworked to best integrate this.

```rust
/// Handles the common logic of writing regular log messages.
///
/// Writes `msg` followed by a newline to the destination, flushing afterwards.
fn write_log(&self, msg: String, msg_level: Level) {
    let mut guard;
    let mut writer: Box<dyn Write> = if self.init.is_initialized() {
        guard = extract_guard(self.log_buf.lock());
        Box::new(guard.as_mut())
    } else {
        match msg_level {
            Level::Error | Level::Warn => Box::new(stderr()),
            _ => Box::new(stdout()),
        }
    };
    // Writes `msg` followed by newline and flushes, if either operation returns an error,
    // increment missed log count.
    // This approach is preferable over `Result::and` as if `write!` returns  an error it then
    // does not attempt to flush.
    if write!(writer, "{}\n", msg)
        .and_then(|_| writer.flush())
        .is_err()
    {
        // No reason to log the error to stderr here, just increment the metric.
        METRICS.logger.missed_log_count.inc();
    }
}
```

## License Acceptance

## PR Checklist


mattschlebusch:
It's quite difficult to tell/confirm the value being added at this stage unfortunately.

Some of my thoughts in two parts, one from a code change point of view and the other from a PR/process hygiene point of view.

-------------------------------------------------------

* From a code/change standpoint, my main consideration is that the value added isn't made clear. So in this particular case, following the PR's first comment (effectively the overview) implies the change aims to fix a bug. Reproducing the bug manually at a minimum will be due diligence, but there should always be a strong preference to add a unit test (or an integration test if a unit test is not do-able) demonstrating the bug. Then the src change being made is vetted against the unit test that demonstrated the bug.
* When drafting the overview or description of a pull request, try to imagine what the path of least resistance will be for a reviewer to safely approve a proposed change.
  * Avoid raw code in the overview if at all possible. Whatever you try to convey in code, should rather be communicated in words as a narrative.
  * When a PR is being used to illustrate and discuss ideas, rather put the PR into draft to make clear what your expectations are from the reviewer when you reach out to them.
  * If you are using the PR for discussion, make it clear what your open questions are, and if possible what it would take to answer them.
  * The overview should have at least two sections at a minimum:
    * `Description`: What/Why/How - Depending on the complexity of the change, elaborate on these questions within reason.
    * `Testing`: How am I sure the pull request's changes are safe?

src/logger/src/logger.rs
```diff
@@ -526,6 +530,40 @@ mod tests {
         assert_eq!(l.show_level(), true);
     }
 
+    #[test]
+    fn test_write_log() {
+        const TEST_HEADER: &str = "test_log";
+        const TEST_STR: &str = "testing flushing";
+        const TEST_FILE: &str = "./test.txt";
```
mattschlebusch, line 537: Is it at all possible to have the buffered writer write to the vector directly instead of using a file? If not, I think this path should move to `/tmp` and if we could clarify the file name as well. My vote would be for `/tmp/firecracker-unit-tests-.log`.

mattschlebusch:
I think the code is fairly close to equivalent in terms of logic and error handling aside from the `flush()` (which is preferred). I think my open question at the moment is that while the current implementation using the `LineWriter` is effectively synchronous, forcing the flush as proposed removes the possibility of using an asynchronous writer in the future without updating/reverting this change. This may not be a big deal, I would like to get a take from the team on how this may impact Firecracker consumers.

src/logger/src/logger.rs
JonathanWoollett-Light: I believe I need to use a [`std::fs::File`](https://doc.rust-lang.org/stable/std/fs/struct.File.html) in this instance since when using an in-memory [`std::io::Cursor`](https://doc.rust-lang.org/std/io/struct.Cursor.html) since read/write times are so fast even when using a [`BufWriter`](https://doc.rust-lang.org/std/io/struct.BufWriter.html) it will not buffer writes. I agree `/tmp` would be a good idea, I will make that change.

mattschlebusch:
> I think is very unlikely to want to add async logging support in the future. I rather see this adding value for heavy logging services but in the context of Firecracker which has a small number of threads and also intends to stay minimalist in features and components, I do not see us wanting to do async logging in the near future.

Thanks for your input!

mattschlebusch:
Given a second opinion, no blockers from me 👍🏻

src/logger/src/logger.rs
```diff
@@ -526,6 +530,48 @@ mod tests {
...
+        let logger = Logger::new();
+        logger.init(String::from(TEST_HEADER), writer).unwrap();
+        // Log some generic data
+        logger.write_log(String::from(TEST_STR), Level::Info);
+        // To drop the logger without calling its destructor, or to `forget` it
+        // (https://doc.rust-lang.org/stable/std/mem/fn.forget.html) will lead
+        // to a memory leak, so for this test I do not do this.
+        // As such this test simply illustrate the `write_log` function will
+        // always flush such that in the occurrence of the crash the expected
+        // behavior that all temporally distant logs from a crash are flushed.
```
luminitavoicu, line 563:
```suggestion
        // As such this test simply illustrates the `write_log` function will
        // always flush such that in the occurrence of the crash the expected
        // behavior that all temporally distant logs from a crash are flushed.
```

luminitavoicu (commented):
Looks good! Just a general nit regarding structuring clean commits: I would rather have the logger fix in a different commit than the one that re-formats imports and comments. It makes for a cleaner separation of purpose and easier to review or come back to in the future.

JonathanWoollett-Light:
@luminitavoicu This occurred as I was fixing imports. I think the reordering here is reasonably within the scope of this commit, in fact as a separate commit it may appear notably weird to have a commit `re-ordering logging imports` which re-orders imports in 3 seemingly random files. 

Do you think it is important to have this in another commit (as splitting the commit will take me some time)?

luminitavoicu (dismissed):
> @luminitavoicu This occurred as I was fixing imports.

Okay, that is fine. It makes more sense to keep it as is.

luminitavoicu (approved):

Issue #2948, opened 2022-04-05
[Bug] rebase-snap is slow due to SEEK_DATA


ColinChartier:
# Describe the bug

Trying to merge some snapshots takes abysmally long, ~30ms per 20,000 bytes => 1ms per KB, 1s per MB, 1000s per GB

## To Reproduce

1. Host is ubuntu 18.04, filesystem is btrfs
2. Command is strace -ff -T rebase-snap --base-file base --diff-file diff
3. Result looks like this:

```
sendfile(3, 4, [20406272] => [20410368], 4096) = 4096 <0.000030>
lseek(4, 20410368, SEEK_DATA)           = 20414464 <0.023237>
lseek(4, 20414464, SEEK_HOLE)           = 20418560 <0.026512>
lseek(3, 20414464, SEEK_SET)            = 20414464 <0.000010>
sendfile(3, 4, [20414464] => [20418560], 4096) = 4096 <0.000031>
lseek(4, 20418560, SEEK_DATA)           = 20439040 <0.026877>
lseek(4, 20439040, SEEK_HOLE)           = 20447232 <0.027989>
lseek(3, 20439040, SEEK_SET)            = 20439040 <0.000010>
```

## Expected behaviour

Ideally rebase-snap should run on the order of seconds, not hours

## Environment

`[Author TODO: Please supply the following information):]`
x86_64 linux 5.4.0-107-generic #121-Ubuntu SMP
5.4.0-107-generic

## Additional context

It seems like any syscall taking 30ms is long, but SEEK_HOLE is especially slow.

## Checks

- [ ] Have you searched the Firecracker Issues database for similar problems?
- [ ] Have you read the existing relevant Firecracker documentation?
- [ ] Are you certain the bug being reported is a Firecracker issue?

serban300:
I tried to reproduce the issue locally and these are the results I am geting:

```
lseek(4, 901419008, SEEK_DATA)          = 901423104 <0.000005>
lseek(4, 901423104, SEEK_HOLE)          = 901427200 <0.000005>
lseek(3, 901423104, SEEK_SET)           = 901423104 <0.000004>
sendfile(3, 4, [901423104] => [901427200], 4096) = 4096 <0.000008>
lseek(4, 901427200, SEEK_DATA)          = 901431296 <0.000005>
lseek(4, 901431296, SEEK_HOLE)          = 901435392 <0.000005>
[...]

PR #3002, opened 2022-05-25
Allow directly writing snapshots on top of existing files

CompuIves:
# Reason for This PR

We're exploring diff snapshots, but found out that our underlying filesystem (btrfs) doesn't perform well with the `rebase-snap` tool due to performance issues with finding holes in files (related: https://github.com/firecracker-microvm/firecracker/issues/2948). Initially we tried using another syscall (`fiemap`), but that also was too slow (10s for 4GB). Because of this, we cannot apply diff snapshots quickly on top of a base snapshot.

Instead, we're now taking another approach where we directly save the diff snapshot on top of the base memory. Firecracker already supports this, because it only writes changed blocks. This has worked out well for us, and this might be interesting to merge into upstream.

There's no issue opened for this right now, and it might well be that this change cannot be merged into Firecracker, but if you're interested I'd be happy to make the required changes to make it fit for the repo!

## Description of Changes

Instead of truncating the memory file we write to, we write on top of it.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/persist.rs
+++ b/src/vmm/src/persist.rs
@@ -327,14 +327,20 @@ fn snapshot_memory_to_file(
     let mut file = OpenOptions::new()
         .write(true)
         .create(true)
-        .truncate(true)
         .open(mem_file_path)
         .map_err(|e| MemoryBackingFile("open", e))?;
 
-    // Set the length of the file to the full size of the memory area.
     let mem_size_mib = mem_size_mib(vmm.guest_memory());
-    file.set_len((mem_size_mib * 1024 * 1024) as u64)
-        .map_err(|e| MemoryBackingFile("set_length", e))?;
+    let expected_size = (mem_size_mib * 1024 * 1024) as u64;
+    let file_size = file
+        .metadata()
+        .map_err(|e| MemoryBackingFile("get_metadata", e))?
+        .len();
+    if file_size != expected_size {
+        // Set the length of the file to the full size of the memory area.
+        file.set_len(expected_size)
+            .map_err(|e| MemoryBackingFile("set_length", e))?;
+    }
 
     match snapshot_type {
         SnapshotType::Diff => {


acatangiu:
My 2c is this is a valid use-case that `firecracker` could support _in addition_ to existing behavior of creating _new_ snapshot containing just _the diff bits_.

There are two different use-cases here:
1. (existing) the user is interested in the _diff pages_ of memory, the _diff snapshot_ - for example, multiple layers of "diffs" can used in a distributed system to build snapshots at varying stages in the VM's lifetime, while minimizing the amount of "data" to distribute across the system.
2. (what this PR proposes) the user is interested in a single snapshot instance, but wants to accelerate mutations to it - take snapshots as the guest workload makes progress (but only interested in latest snapshot), so instead of creating slow(er) full snapshots, directly create fast(er) diff snapshots on top of each other.

I think there's value in both use-cases, and the proposed changes in this PR seem to support both of them :+1:

CompuIves:
It seems like the tests are failing on test coverage, should I change a value here?

luminitavoicu:
@CompuIves, thanks for your contribution! Your changes increase ARM coverage. Please rebase your PR and update the coverage value for the failing tests.

src/vmm/src/persist.rs
```diff
@@ -327,14 +327,20 @@ fn snapshot_memory_to_file(
...
         .create(true)
-        .truncate(true)
         .open(mem_file_path)
         .map_err(|e| MemoryBackingFile("open", e))?;
 
-    // Set the length of the file to the full size of the memory area.
     let mem_size_mib = mem_size_mib(vmm.guest_memory());
-    file.set_len((mem_size_mib * 1024 * 1024) as u64)
-        .map_err(|e| MemoryBackingFile("set_length", e))?;
+    let expected_size = (mem_size_mib * 1024 * 1024) as u64;
```
luminitavoicu, line 334: Not necessarily in scope of this PR, but I would like us to have a constant for `1024 * 1024`.

luminitavoicu (commented):
Looks good, I like the idea! Could you add an integration test for this, as well?

CompuIves:
Rebased the PR, but I'm not sure what the code coverage is now with the changes that came from the rebase. Should I wait for a test run?

CompuIves:
Added integration test as well!

luminitavoicu (commented):
Looks good! Please make sure that all of your commits are signed. 
The coverage test is failing, you will have to update the numbers once again. Also take a look at the errors output from the python style test.

CompuIves:
Addressed the feedback, squashed the commits, signed it and updated coverage! I think the tests should now pass, though I did see other perf tests fail, so that's something we should check.

dianpopa:
Hi!

I do like the proposal! However, I am concerned about the possibility of a breaking change since I do expect that users of the `rebase-snap` tool to have the expectation to obtain diff layers containing dirty memory since *last* retrieval.
Shouldn't we let rebase-snap tool users know that from now in order to obtain a diff layer they need to make sure to remove any previous (name conflicting) snapshots?

roypat:
> Hi!
> 
> I do like the proposal! However, I am concerned about the possibility of a breaking change since I do expect that users of the `rebase-snap` tool to have the expectation to obtain diff layers containing dirty memory since _last_ retrieval. Shouldn't we let rebase-snap tool users know that from now in order to obtain a diff layer they need to make sure to remove any previous (name conflicting) snapshots?

Mh, yes, for successive diff snapshots, this essentially does "rebase-snap" functionality at snapshot-taking-time, we might want to add a note about that somewhere. For full snapshots however, this "fixes" the scenario where currently if a full snapshot from which the microvm is loaded is being used as the target of a full/diff snapshot, it just gets zeroed out.

PR #3011, opened 2022-05-30
Upgrade devtool container to use Python 3.10

mattschlebusch:
# Reason for This PR

Python 3.6 is reaching end of life and needs to be upgraded to version 3.7+.

Upgrading to 3.10.4 was started in [PR/3003](https://github.com/firecracker-microvm/firecracker/pull/3003) which upgraded the Ubuntu version of the container from 18.04 to 22.04. This resulted in a change to the Firecracker binary size that needs to be better understood before that upgrade can proceed. This change targets upgrading Python only and builds the runtime from source which is not ideal. This can ultimately be removed once the upgrade to 22.04 can be completed safely.

## Description of Changes

* Build and install Python 3.10.4 runtime
* Configure safe.directory to satiate git (via gitlint) file-system ownership checks
* Bump the vm-memory crate's version. vm-memory is now flagged as a side-effect of the update due to [RUSTSEC-2020-0157](https://rustsec.org/advisories/RUSTSEC-2020-0157). The advisory is referring to the [external crate's](https://crates.io/crates/vm-memory) version which has already been updated, however cargo-audit is not able to differentiate between the crates.

## License Acceptance

### DIFF
--- a/tests/framework/utils.py
+++ b/tests/framework/utils.py
@@ -14,9 +14,11 @@ import time
 import platform
 
 from collections import namedtuple, defaultdict
+from pathlib import Path
 import psutil
 from retry import retry
 from retry.api import retry_call
+from framework import utils
 from framework.defs import MIN_KERNEL_VERSION_FOR_IO_URING
 
 CommandReturn = namedtuple("CommandReturn", "returncode stdout stderr")
@@ -526,6 +528,31 @@ def run_cmd_list_async(cmd_list):
             *cmds
         )
     )
+
+
+def configure_git_safe_directory():
+    """
+    Add the root firecracker git folder to safe.directory config.
+
+    Firecracker root git folder in the container is
+    bind-mounted to a folder on the host which is mapped to a
+    user that is different from the user which runs the integ tests.
+    This difference in ownership is validated against by git.
+    https://github.blog/2022-04-12-git-security-vulnerability-announced/
+
+    :return: none
+    """
+    # devtool script will set the working directory to FC_ROOT/tests.
+    # We will need the parent for safe.directory configuration
+    working_dir = Path(os.getcwd())
+
+    try:
+        utils.run_cmd('git config --global '
+                      '--add safe.directory {}'.format(working_dir.parent))
+    except ChildProcessError as error:
+        assert False, "Failure to set the safe.directory " \
+                      "git config to [{}] required for gitlint tests: " \
+                      "{}".format(working_dir.parent, error)
 
 
 def run_cmd(cmd, ignore_return_code=False, no_shell=False, cwd=None):
--- a/tests/framework/utils.py
+++ b/tests/framework/utils.py
@@ -550,9 +550,9 @@ def configure_git_safe_directory():
         utils.run_cmd('git config --global '
                       '--add safe.directory {}'.format(working_dir.parent))
     except ChildProcessError as error:
-        assert False, "Failure to set the safe.directory " \
-                      "git config to [{}] required for gitlint tests: " \
-                      "{}".format(working_dir.parent, error)
+        raise Exception("Failure to set the safe.directory "
+                        "git config to [{}] required for gitlint tests"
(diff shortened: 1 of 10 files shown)


dianpopa:
It is not clear why we need to build from sources? is it a workaround for the increase in binary size (sorry if there was already a conversation on this that I missed)

dianpopa:
Another alternative I see here would be to try to upgrade to using Ubuntu 20.04 and Python version 3.8.x. WDYT?

mattschlebusch:
> It is not clear why we need to build from sources? is it a workaround for the increase in binary size (sorry if there was already a conversation on this that I missed)

Yea the binary size was a rabbit hole I wasn't willing to go done at the time, and building from source I consider to be temporary as upgrading Ubuntu is a good idea, there are a number of fixes and updates we get (more specifically iperf3 which the current version has some documented bugs in).

> Another alternative I see here would be to try to upgrade to using Ubuntu 20.04 and Python version 3.8.x. WDYT?

I will try this briefly to see if the build goes through without the binary size issue.

tests/framework/utils.py
```diff
@@ -528,6 +530,31 @@ def run_cmd_list_async(cmd_list):
...
+    # We will need the parent for safe.directory configuration
+    working_dir = Path(os.getcwd())
+
+    try:
+        utils.run_cmd('git config --global '
+                      '--add safe.directory {}'.format(working_dir.parent))
+    except ChildProcessError as error:
+        assert False, "Failure to set the safe.directory " \
+                      "git config to [{}] required for gitlint tests: " \
+                      "{}".format(working_dir.parent, error)
```
JonathanWoollett-Light, line 555:
```suggestion
        raise Exception("Failure to set...") from error
```
This allows chaining the errors for giving more infomation, see https://docs.python.org/3/tutorial/errors.html#exception-chaining.

tests/integration_tests/functional/test_api.py
```diff
@@ -1184,10 +1185,26 @@ def test_api_version(test_microvm_with_api):
...
+                                   '--dirty', '--abbrev=0']).decode()
+
+    # Skip the "v" at the start
+    tag_version = out[1:]
+    # Strip the metadata appended to the tag
+    if out.find('-') > -1:
+        tag_version = tag_version[:tag_version.index('-')]
+    else:
+        # Just strip potential newlines
+        tag_version = tag_version.strip()
```
JonathanWoollett-Light, line 1202:
```suggestion
    [_, *tag_version] = out
    if i := tag_version.find('-') != -1:
        tag_version = tag_version[:i]
    else:
        tag_version = tag_version.strip()
```
This presumes we can always simply skip the first character.
https://blog.teclado.com/destructuring-in-python/ and https://realpython.com/python-walrus-operator/ might be useful.

dianpopa (approved):

tests/integration_tests/functional/test_api.py
mattschlebusch: I don't think the comparisons count here is critical for test code, and the readability is a bit easier laid out more in my opinion.

JonathanWoollett-Light: I can understand that `i := tag_version.find('-') != -1:` is a little esoteric (perhaps we should record this decision to abstain from walrus operators in a style guide).

If you want to avoid this I would suggest instead:
```pyth
    [_v, *tag_version] = out
    i = tag_version.find('-')
    if i != -1:
        tag_version = tag_version[:i]
    else:
        tag_version = tag_version.strip()
```
Even though I have seen it before I personally find the `find` into `index` to be somewhat misleading from a readability perspective.

I also understand `[_v, *tag_version] = out` may raise some contention. Here I would strongly advocate for it as it helps clarifies we are dropping and not using the respective first character `_v` (I think de-structuring syntax can appear initially odd but helps readability (and safety) when you get used to it).

luminitavoicu (approved):

dianpopa (approved):

PR #3015, opened 2022-05-30
Simplify ApiServer thread shutdown

acatangiu:
## Description of Changes

Replace hacky "/shutdown-internal" HTTP request with clean signal+flag based mechanism.

`SIGRTMAX` noop handler is installed and is used on the API thread to break the internal micro-http `epoll` loop. A simple `AtomicBool` flag is then used to signal the ApiServer it should shut down.

## License Acceptance

## PR Checklist

- [x] ~The issue which led to this PR has a clear conclusion.~
- [x] ~This PR follows the solution outlined in the related issue.~
- [x] ~Any API changes follow the [Runbook for Firecracker API changes][2].~

### DIFF
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -9,10 +9,15 @@
 mod parsed_request;
 mod request;
 
+use std::os::unix::net::UnixStream;
 use std::path::PathBuf;
-use std::sync::mpsc;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{mpsc, Arc};
+use std::thread::JoinHandle;
+use std::time::Duration;
 use std::{fmt, io};
 
+use libc::{c_int, c_void, siginfo_t};
 use logger::{
     debug, error, info, update_metric_with_elapsed_time, warn, ProcessTimeReporter, METRICS,
 };
@@ -20,9 +25,10 @@ pub use micro_http::{
     Body, HttpServer, Method, Request, RequestError, Response, ServerError, ServerRequest,
     ServerResponse, StatusCode, Version,
 };
-use seccompiler::BpfProgramRef;
+use seccompiler::{BpfProgram, BpfProgramRef};
 use serde_json::json;
 use utils::eventfd::EventFd;
+use utils::signal::{register_signal_handler, sigrtmax, Killable};
 use vmm::rpc_interface::{VmmAction, VmmActionError, VmmData};
 use vmm::vmm_config::snapshot::SnapshotType;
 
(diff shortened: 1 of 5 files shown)


src/api_server/src/lib.rs
```diff
@@ -486,31 +640,33 @@ mod tests {
...
-        // Wait for the server to set itself up.
-        socket_ready_receiver.recv().unwrap();
         to_api
             .send(Box::new(Ok(VmmData::InstanceInformation(
                 InstanceInfo::default(),
             ))))
             .unwrap();
-        let mut sock = UnixStream::connect(PathBuf::from(path_to_socket)).unwrap();
+        let mut sock =
+            try_connect(PathBuf::from(path_to_socket), 5, Duration::from_millis(10)).unwrap();
```
mattschlebusch, line 669: Interesting tweak, was there previously a risk of this locking up the test until the test framework forcefully timed out?

src/api_server/src/lib.rs
```diff
@@ -60,6 +66,49 @@ impl fmt::Debug for Error {
...
+}
+
+/// An owned handle to an ApiServer thread, used to stop and join the thread.
+pub struct ThreadedApiServerHandle {
+    stop_signum: c_int,
+    stop_flag: Arc<AtomicBool>,
+    thread_handle: JoinHandle<()>,
+}
+
+impl ThreadedApiServerHandle {
```
mattschlebusch, line 96: nitpick: If another revision of this PR is made, consider renaming this type from `ThreadedApiServerHandle` to `ApiServerThreadHandle`. The former implies it is a handle to the whole server (all the threads included). Really a nitpick though.

mattschlebusch (dismissed):
LGTM.

src/api_server/src/lib.rs
acatangiu: Not locking up, `UnixStream::connect` would sometimes just return error (test fail) because the socket doesn't exist at provided path.

This would happen when the test thread is faster than the spawned API thread.

---

It didn't happen before because `socket_ready_receiver` channel was used as a barrier to wait for the API thread to set up its socket. This "barrier" was used outside of tests as well when the main thread would send "/shutdown-internal" to the API thread.

Now that we removed the public facing "/shutdown-internal" HTTP command, and we shut down using actual _internal mechanisms_, I don't see the point in keeping extra `socket_ready_receiver` channel just for tests.

In tests we can just retry the connection a few times and it will be available if all is well, it will correctly fail if smth is wrong.

src/api_server/src/lib.rs
acatangiu: sure, can do once reviews are done; also feel free to just push this on top of the PR branch yourself.

src/api_server/src/lib.rs
mattschlebusch: Thanks for explaining your thinking there. Certainly a leaner approach in my opinion as well.

luminitavoicu (approved):
Looking good!

acatangiu:
I don't think this works as expected.. for some reason my local run of `test_shutdown.py` fails - the API server epoll syscall is not being interrupted although the signal handler does get called on the API thread...

Let's see what the CI says. I will come back to this later when I have some time to burn.

PR #3018, opened 2022-05-31
Better error handling

JonathanWoollett-Light:
# Reason for This PR

1. Error propagation features excessive usage of `map_err` to wrap types before returning them.
    Some error types implement some `From` implementations, but most don't.
2. Usage of the `e` identifier is ambiguous, internally inconsistent (`e` and `err` can both be found in the code base) and externally inconsistent with typical Rust patterns (where `err` is typically used for `impl Error` types).

## Description of Changes

1. Used [`derive_more::from`](https://jeltef.github.io/derive_more/derive_more/from.html) to create many more `From` implementations for error types to allow cleaner error propagation.
   I have taken a conservative approach here only using these implementations when they can be the most simple, this leaves `map_err` still being used extensively throughout the code base. In many of these remaining instances `map_err` is used for clarity with errors where multiple variants contain the same type/s or where error variants contain multiple types.
2. Disambiguated the `e` identifier:
    - Types that implemented `Error` now use the identifier `err`, inline with the standard library and general approach
    - `Endpoint` types now use the identifier `endpoint`
    - `WriteEvent` types now use the identifier `event`

Added [`derive_more`](https://crates.io/crates/derive_more) as a dependency. It is a small (~55kb) and widely used (~17m downloads) library in the Rust ecosystem for minimizing code duplication by deriving simple implementations. Here I also restrict this further by disabling default features and only including the feature for the [`derive_more::from`](https://jeltef.github.io/derive_more/derive_more/from.html) macro.

## License Acceptance

## PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -41,6 +41,7 @@ dependencies = [
 name = "api_server"
 version = "0.1.0"
 dependencies = [
+ "derive_more",
  "libc",
  "logger",
  "micro_http",
@@ -58,6 +59,7 @@ name = "arch"
 version = "0.1.0"
 dependencies = [
  "arch_gen",
+ "derive_more",
  "device_tree",
  "kvm-bindings",
  "kvm-ioctls",
@@ -236,6 +238,7 @@ dependencies = [
 name = "cpuid"
 version = "0.1.0"
 dependencies = [
+ "derive_more",
  "kvm-bindings",
  "kvm-ioctls",
  "utils",
@@ -359,6 +362,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derive_more"
+version = "0.99.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4fb810d30a7c1953f91334de7244731fc3f3c10d7fe163338a35b9f640960321"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "device_tree"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -368,6 +382,7 @@ checksum = "f18f717c5c7c2e3483feb64cccebd077245ad6d19007c2db0fd341d38595353c"
 name = "devices"
 version = "0.1.0"
 dependencies = [
+ "derive_more",
  "dumbo",
  "event-manager",
  "io_uring",
@@ -393,6 +408,7 @@ name = "dumbo"
 version = "0.1.0"
 dependencies = [
  "bitflags",
+ "derive_more",
  "logger",
  "micro_http",
  "serde_json",
@@ -488,6 +504,7 @@ dependencies = [
 name = "io_uring"
 version = "0.1.0"
 dependencies = [
+ "derive_more",
  "libc",
  "proptest",
  "utils",
@@ -650,6 +667,7 @@ dependencies = [
  "aes-gcm",
  "base64",
  "bincode",
+ "derive_more",
  "dumbo",
  "logger",
  "micro_http",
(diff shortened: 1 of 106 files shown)


mattschlebusch:
No objections to the objectives of the PR. Some thoughts in summary:

* Could we consider renaming `WriteEvent` to `writeEvent`. I think we lose some useful context by dropping `write` entirely from the reference name.
* The PR is already quite large. Addressing multiple standards under one review makes the investment for review larger (even if the diff size doesn't change, there is still more logic to vet for each delta).
* The code coverage delta shouldn't have happened for a coding standards change (ideally). This is likely due to the `derive_from`/`map_err` change which is the only explanation I can think of that would change the behaviour of the binary. If that's the case this is not really just a coding standards change anymore. With customer experience implications here that aren't clear (to me). I doubt it's serious, but I do think due diligence suggests checking with more experienced Firecracker Rustaceans that this is not a problem.

JonathanWoollett-Light:
> Could we consider renaming WriteEvent to writeEvent. I think we lose some useful context by dropping write entirely from the reference name.

Rust structs typically use pascal case, however I think you mean to consider renaming the values of type `WriteEvent` to use the identifier `writeEvent`? In this case I would suggest instead `write_event` for snake case here, would you agree?

> The code coverage delta shouldn't have happened for a coding standards change (ideally). This is likely due to the derive_from/map_err change which is the only explanation I can think of that would change the behaviour of the binary. If that's the case this is not really just a coding standards change anymore. With customer experience implications here that aren't clear (to me). I doubt it's serious, but I do think due diligence suggests checking with more experienced Firecracker Rustaceans that this is not a problem.

This is what is causing the code coverage degradation here, since `derive_from` derives [`From`](https://doc.rust-lang.org/std/convert/trait.From.html) implementations for all the variants it can (some of which we may not currently use). I would note however this would not `change the behavior of the binary` as these unused implementations would not be present in the binary. Although it is not as clear as I would like (given our test suite) I do not believe this would have any affect on the Firecracker binary.

JonathanWoollett-Light:
> With regards to the question of behaviour, is there a way we can be certain? Tests are supposed to be vetting behaviour, if the code test coverage changes surely that does indicate a possible change in behaviour?

Our tests at-least guarantee no change in most behavior we depend on, with no test failures that seems to be the case with these changes (however underlying behavior can change without necessarily changing the behavior we test).

I believe you are correct that there is some change to the binary introduced here, at the moment even combing through the PR I cannot see where this is introduced. My initial guess was new unused derived impls but I think these should be optimized out of the binary. I could look at making a diff of the assembly to see where this difference arises however this is considerable work and I do not think it is necessary given all tests pass.

I think this highlights the issue with our current code coverage testing and approach that it does not highlight the specific areas of uncovered code (we use the first approach [mentioned here](https://doc.rust-lang.org/rustc/instrument-coverage.html) rather than the second which I would consider preferable). I suggest this raises a separate issue relating to improving our code coverage process.

mattschlebusch:
tl;dr - I'm inclined to set aside the possible coverage shortfalls in the context of this PR to move forward.

-------------------------

At the risk of being needlessly verbose, here are my passing comments:

* Code coverage tooling is lacking here and would benefit from some investment in the future. The sky is the limit in an idealistic sense, and we can consider binary-code and human-code path coverage (likely not the same). Binary coverage may be useful for vetting changes in the automated context, but it's the code coverage delta (a diff between covered lines on the base branch and covered lines of the developer branch) that would have been the most useful here for humans to scrutinize possible behavioural differences. I am also inclined to think that the `llvm-cov` implementation give the options to produce this coverage delta report.
* It's rare that code consistency is not appreciated by developers working on said code. I am still not sure though that a change this size is worthwhile given entropy in the future without programmatic enforcement of a given standard.

PR #3021, opened 2022-06-01
Enhanced code coverage error message

JonathanWoollett-Light:
# Reason for This PR

Current code coverage error message is slightly inaccurate e.g.
```
E   AssertionError: Current code coverage (83.77%) is below the target (84.06%).
E   assert 83.76920450665756 >= 84.01
```
We are asserting we are not below `84.01` (`>= 84.01`) while the error message tells us we are asserting we are not below `84.06` (`>= 84.06`).

(same applies with above the target)

## Description of Changes

Enhanced error message to give:
```
E   AssertionError: Current code coverage (83.77%) is >0.05% below the target (84.06%).
E   assert 83.76920450665756 >= 84.01
```

(applied same change with above the target)

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -116,26 +116,29 @@ def test_coverage(test_fc_session_root_path, test_session_tmp_path):
     print("Thus, coverage is: {:.2f}%".format(coverage))
 
     coverage_low_msg = (
-        'Current code coverage ({:.2f}%) is below the target ({}%).'
-        .format(coverage, coverage_target_pct)
+        'Current code coverage ({:.2f}%) is >{:.2f}% below the target ({}%).'
+        .format(coverage, COVERAGE_MAX_DELTA, coverage_target_pct)
     )
 
-    min_coverage = coverage_target_pct - COVERAGE_MAX_DELTA
-    assert coverage >= min_coverage, coverage_low_msg
+    assert coverage >= coverage_target_pct - COVERAGE_MAX_DELTA, \
+        coverage_low_msg
 
     # Get the name of the variable that needs updating.
     namespace = globals()
-    cov_target_name = [name for name in namespace if namespace[name]
-                       is COVERAGE_DICT][0]
+    cov_target_name = \
+        [name for name in namespace if namespace[name] is COVERAGE_DICT][0]
 
     coverage_high_msg = (
-        'Current code coverage ({:.2f}%) is above the target ({}%).\n'
-        'Please update the value of {}.'
-        .format(coverage, coverage_target_pct, cov_target_name)
+        'Current code coverage ({:.2f}%) is >{:.2f}% above the target ({}%).\n'
+        'Please update the value of {}.'.format(
+            coverage, COVERAGE_MAX_DELTA, coverage_target_pct, cov_target_name
+        )
     )
 
-    assert coverage - coverage_target_pct <= COVERAGE_MAX_DELTA,\
-        coverage_high_msg
+    assert \
+        coverage <= coverage_target_pct + COVERAGE_MAX_DELTA, coverage_high_msg
 
-    return f"{coverage}%", \
+    return (
+        f"{coverage}%",
         f"{coverage_target_pct}% +/- {COVERAGE_MAX_DELTA * 100}%"
+    )


tests/integration_tests/build/test_coverage.py
```diff
@@ -116,26 +116,26 @@ def test_coverage(test_fc_session_root_path, test_session_tmp_path):
...
-    assert coverage >= min_coverage, coverage_low_msg
+    assert coverage >= coverage_target_pct - COVERAGE_MAX_DELTA, coverage_low_msg
 
     # Get the name of the variable that needs updating.
     namespace = globals()
-    cov_target_name = [name for name in namespace if namespace[name]
-                       is COVERAGE_DICT][0]
+    cov_target_name = [name for name in namespace if namespace[name] is COVERAGE_DICT][
+        0
+    ]
```
luminitavoicu, line 130: I am not loving this reformatting.

tests/integration_tests/build/test_coverage.py
```diff
@@ -64,35 +64,35 @@ def test_coverage(test_fc_session_root_path, test_session_tmp_path):
...
-        '_gen'
+        "bootparam.rs,"
+        "elf.rs,"
+        "mpspec.rs,"
+        "msr_index.rs,"
+        "bindings.rs,"
+        "_gen"
     )
-    exclude_region = '\'mod tests {\''
+    exclude_region = "'mod tests {'"
```
luminitavoicu, line 81: In this case changing `'` into `"` makes sense because it saves us from needing to escape `'`. But is there any particular reason for changing `'` enclosed strings into `"` all over the file? In Python, single-quoted strings and double-quoted strings are the same. 
We should pick one and stick with it (apart from cases when the string contains single or double quotes that need to be escaped).

luminitavoicu (commented):
Changes look good. Just a couple of general comments.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: While it doesn't matter in Python, using `"` conforms to the more universal pattern of `"` for strings and `'` for chars.

Also the popular python formatter [`black`](https://github.com/psf/black) uses `"` for strings and this is what I just used out of habit which changed all the strings to use it. I would argue in favor of both, using `"` for strings and `black` for formatting (e.g. `black .`) as I think  is it easier and simpler than `pycodestyle`.

For now I will undo these stylistic changes to be more inline with the current style.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: Adjusted this [here](https://github.com/firecracker-microvm/firecracker/pull/3021/commits/06160dfda2d6097760f3e5ee56a5f63f86863a01#diff-58497b264df840dfe4603fa26477d633d32c0f2173bf4525265afb814b3d30eaR130-R131). Opinion?

tests/integration_tests/build/test_coverage.py
luminitavoicu: `pycodestyle` is what we currently use for python formatting (see [here](https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/style/test_python.py#L26)). Using `black` would require a full reformat for all python files, and a change in our python-style test, as well. I am not against switching to `black`, but we should first see the differences between the two first and have a discussion about the benefits.

> While it doesn't matter in Python, using " conforms to the more universal pattern of " for strings and ' for chars

I agree. We should settle on one, make the changes in all of the files and maybe document it somewhere as a rule for Firecracker contributing.

luminitavoicu (approved):

tests/integration_tests/build/test_coverage.py
luminitavoicu: Looks good!

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: > Using black would require a full reformat for all python files, and a change in our python-style test, as well.

This would essentially just be running `black .` then skimming and re-testing to check if it broke anything.

luminitavoicu: Yes, I am not saying that it's hard work, I am just saying that complying with one formatter should be done for all files at once and that our python style test should reflect the preferred formatter.

dianpopa (approved):

PR #3032, opened 2022-06-08
Fix intermittent fail on AMD when getting cpu count

### LINKED ISSUE #2216: [Bug] SIGILL handler not called when running with kcov on arm64
## Describe the bug

While investigating why this test fails when running coverage test for arm64, I came to the conclusion that SIGILL handler does not get called at all. Investigate why this happens.

## To Reproduce
```
./tools/devtool shell -p
cd src/vmm
RUSTFLAGS="-C link-arg=-lgcc -C link-arg=-lfdt " cargo kcov --target aarch64-unknown-linux-musl

```

dianpopa:
# Reason for This PR

We have been dealing with intermittent fails for `vmm::signal_handler::test_signal_handler` on AMD.
This happens in the cpu_count() function due. We are computing the number of cpus by getting the affinity mask of current process and then going through CPU_SETSIZE to count the number of cpus that hit the mask. 
Problem is that `libc:musl` has a maximum of CPU_SETSIZE equal 128 and  on AMD the maximum number of CPUS with SMT enabled is 192.

Fixes #2216 

## Description of Changes

The cpu_count was used as an trick for detecting whether or not the unit tests are run under `cargo test` or `cargo kcov` since cargo kcov would make it seem that there is only one CPU online.

We remove this workaround and introduce a environment variable CARGO_WRAPPER that can tell us what is the wrapper currently run by cargo.

While implementing the solution, also found the reason why this happens: https://github.com/firecracker-microvm/firecracker/issues/2216. Comment in code.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/vmm/src/signal_handler.rs
+++ b/src/vmm/src/signal_handler.rs
@@ -172,37 +172,16 @@ pub fn register_signal_handlers() -> utils::errno::Result<()> {
 mod tests {
     use super::*;
 
-    use libc::{cpu_set_t, syscall};
-    use std::{mem, process, thread};
+    use std::env;
+    use std::{process, thread};
 
+    use libc::syscall;
     use seccompiler::sock_filter;
-
-    // This function is used when running unit tests, so all the unsafes are safe.
-    fn cpu_count() -> usize {
-        let mut cpuset: cpu_set_t = unsafe { mem::zeroed() };
-        unsafe {
-            libc::CPU_ZERO(&mut cpuset);
-        }
-        let ret = unsafe {
-            libc::sched_getaffinity(
-                0,
-                mem::size_of::<cpu_set_t>(),
-                &mut cpuset as *mut cpu_set_t,
-            )
-        };
-        assert_eq!(ret, 0);
-
-        let mut num = 0;
-        for i in 0..libc::CPU_SETSIZE as usize {
-            if unsafe { libc::CPU_ISSET(i, &cpuset) } {
-                num += 1;
-            }
-        }
-        num
-    }
 
     #[test]
     fn test_signal_handler() {
+        let run_with_kcov = env::var("CARGO_WRAPPER").unwrap_or_else(|_| "".to_string()) == *"kcov";
+
         let child = thread::spawn(move || {
             assert!(register_signal_handlers().is_ok());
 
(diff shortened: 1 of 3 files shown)


src/vmm/src/signal_handler.rs
```diff
@@ -172,37 +172,15 @@ pub fn register_signal_handlers() -> utils::errno::Result<()> {
...
-            if unsafe { libc::CPU_ISSET(i, &cpuset) } {
-                num += 1;
-            }
-        }
-        num
-    }
-
     #[test]
     fn test_signal_handler() {
+        let run_with_kcov = env::var("CARGO_WRAPPER").is_ok();
```
bchalios, line 183: nit: Maybe explicitly check for "kcov", just in case we need to check against other wrappers in the future

src/vmm/src/signal_handler.rs
```diff
@@ -258,27 +236,26 @@ mod tests {
...
-            assert!(METRICS.seccomp.num_faults.fetch() >= 1);
-        }
+        // SIGSYS, which is raised whenever a bad syscall is caught will be intercepted by kcov on x86_64
+        // and thus not reach Firecracker:
+        // https://github.com/SimonKagstrom/kcov/blob/a8b60c43fb33f56553a2bb20633e3b59a08abae1/src/engines/ptrace.cc#L187
+        // So, we are not checking for the `num_faults` metrics which gets incremented on each bad syscall
+        // if we run with kcov and we are on x86_64.
+        assert!(
+            METRICS.seccomp.num_faults.fetch()
+                >= (!run_with_kcov || !cfg!(target_arch = "x86_64")) as usize
```
bchalios, line 246: Do we actually need to check for the `target_arch` here? There wasn't there in the original code.

luminitavoicu (commented):
Apart from the already existing comments, looks good to me!

src/vmm/src/signal_handler.rs
dianpopa: Done.

src/vmm/src/signal_handler.rs
dianpopa: Yes, so I wanted to be really strict about this in order to reflect on the reason behind why this happens. The num_faults is 0 only on x86_64 since kcov will intercept the SIGSYS while on ARM, SIGILL is the one intercepted which motivates the next change.
The explanation is here: https://github.com/SimonKagstrom/kcov/blob/a8b60c43fb33f56553a2bb20633e3b59a08abae1/src/engines/ptrace.cc#L187

bchalios (approved):
LGTM

src/vmm/src/signal_handler.rs
```diff
@@ -258,27 +236,26 @@ mod tests {
...
         assert!(METRICS.signals.sigxfsz.fetch() >= 1);
         assert!(METRICS.signals.sigxcpu.fetch() >= 1);
         assert!(METRICS.signals.sigpipe.count() >= 1);
         assert!(METRICS.signals.sighup.fetch() >= 1);
-        // Workaround to GitHub issue 2216.
-        #[cfg(not(target_arch = "aarch64"))]
-        assert!(METRICS.signals.sigill.fetch() >= 1);
+        assert!(
+            METRICS.signals.sigill.fetch()
+                >= (!run_with_kcov || !cfg!(target_arch = "aarch64")) as usize
```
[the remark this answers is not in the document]
bchalios: ok, it makes sense! thanks!

luminitavoicu (approved):

PR #3036, opened 2022-06-17
Updating recommendations and checking mitigations regarding Spectre-like attacks

JonathanWoollett-Light:
# Reason for This PR

For better security recommendations regarding Spectre-like attacks should be improved and there should be testing regarding the presence of these mitigations.

## Description of Changes

1. Added an integration test checking for the presence of necessary SPECTREv2 mitigations.
2. Updated recommendations to better cover Spectre-like attacks.

## License Acceptance

## PR Checklist

### DIFF
--- a/docs/prod-host-setup.md
+++ b/docs/prod-host-setup.md
@@ -293,70 +293,54 @@ echo "KSM: ENABLED (Recommendation: DISABLED)"
 
 #### Check for mitigations against Spectre Side Channels
 
-##### Branch Target Injection mitigation (Spectre V2)
+##### Branch Target Injection mitigation (Spectre V2, including Spectre-BHB)
 
-**Intel and AMD**
-Where available, Intel recommends using Enhanced Indirect Branch Restricted
-Speculation (eIBRS) together with microcode supporting conditional
-Indirect Branch Prediction Barriers (IBPB).
+###### Intel and AMD
 
-If eIBRS is not available, use a kernel compiled with retpoline and run on
-hardware with microcode supporting IBPB and
-Indirect Branch Restricted Speculation (IBRS).
+We recommend using a kernel compiled with eIBRS or `RETPOLINE`, together with
+microcode supporting conditional Indirect Branch Prediction Barriers (IBPB).
 
 Verification can be done by running:
 
 ```bash
-(grep -Eq '^Mitigation: Full [[:alpha:]]+ retpoline,
-IBPB: conditional, IBRS_FW' \
-/sys/devices/system/cpu/vulnerabilities/spectre_v2 && \
-echo "retpoline, IBPB, IBRS: ENABLED (OK)") \
-|| (grep -Eq '^Mitigation: Enhanced IBRS,
-IBPB: conditional' \
-/sys/devices/system/cpu/vulnerabilities/spectre_v2 && \
-echo "eIBRS, IBPB: ENABLED (OK)") \
-|| echo "eIBRS, IBPB: DISABLED (Recommendation: ENABLED)"
+cat /sys/devices/system/cpu/vulnerabilities/spectre_v2
 ```
 
-**ARM** The mitigations for ARM systems are patched in all linux stable versions
-starting with 4.16. More information on the processors vulnerable to this type
+The output should mention the following mitigations being in use:
+
+- `Enhanced IBRS` or `Retpolines`
+- `IBPB` at least `conditional`
+
+###### ARM64
+
+We recommend using a kernel compiled with `MITIGATE_SPECTRE_BRANCH_HISTORY`.
+
+More information on the processors vulnerable to this type
 of attack and detailed information on the mitigations can be found in the
 [ARM security documentation](https://developer.arm.com/support/arm-security-updates/speculative-processor-vulnerability).
 
 Verification can be done by running:
 
 ```bash
-(grep -q "^Mitigation:" /sys/devices/system/cpu/vulnerabilities/spectre_v2 || \
-grep -q "^Not affected$" /sys/devices/system/cpu/vulnerabilities/spectre_v2) && \
+grep -q "^(Mitigation: CSV2, BHB|Not affected)$" \
+/sys/devices/system/cpu/vulnerabilities/spectre_v2 && \
 echo "SPECTRE V2 -> OK" || echo "SPECTRE V2 -> NOT OK"
 ```
-
-##### Spectre-BHB
-
-**All hosts** Use latest default SpectreV2 mitigations by keeping kernels up to
-date. Additionally, use a host kernel that has unprivileged BPF disabled by
-enabling `BPF_UNPRIV_DEFAULT_OFF` in the kernel config or by writing `1` or `2`
-to `/proc/sys/kernel/unprivileged_bpf_disabled`.
-
-**Intel** . For an extra layer of security, while trading off performance, you
-can add `spectrev2=eibrs,lfence` or more strictly, `spectrev2=eibrs,retpoline`
-to the kernel commandline of the host.
 
 ##### Bounds Check Bypass Store (Spectre V1)
 
 Verification for mitigation against Spectre V1 can be done:
 
 ```bash
-(grep -q "^Mitigation:" /sys/devices/system/cpu/vulnerabilities/spectre_v1 || \
-grep -q "^Not affected$" /sys/devices/system/cpu/vulnerabilities/spectre_v1) && \
+grep -q "^(Mitigation:|Not affected)$" \
+/sys/devices/system/cpu/vulnerabilities/spectre_v1 && \
 echo "SPECTRE V1 -> OK" || echo "SPECTRE V1 -> NOT OK"
 ```
 
-#### [Intel only] Apply L1 Terminal Fault (L1TF) mitigation
+##### [Intel only] Apply L1 Terminal Fault (L1TF) mitigation
 
 These features provide mitigation for Foreshadow/L1TF side-channel issue on
 affected hardware.
-
 They can be enabled by adding the following Linux kernel boot parameter:
 
 ```console
@@ -365,7 +349,6 @@ l1tf=full,force
 
 which will also implicitly disable SMT.  This will apply the mitigation when
 execution context switches into microVMs.
-
 Verification can be done by running:
 
 ```bash
@@ -378,30 +361,44 @@ echo "$cond: DISABLED (Recommendation: ENABLED)"; done
 
 See more details [here](https://www.kernel.org/doc/html/latest/admin-guide/hw-vuln/l1tf.html#guest-mitigation-mechanisms).
 
-#### Apply Speculative Store Bypass (SSBD) mitigation
+##### Apply Speculative Store Bypass (SSBD) mitigation
 
 This will mitigate variants of Spectre side-channel issues such as
-Speculative Store Bypass and SpectreNG.
+Speculative Store Bypass (Spectre v4) and SpectreNG.
 
-On x86_64 systems, it can be enabled by adding the following Linux kernel boot
+We recommend applying SSBD to Firecracker and the host kernel.
+
+###### X86_64
+
+On x86_64 systems, this can be done using the following kernel cmdline
 parameter:
 
 ```console
-spec_store_bypass_disable=seccomp
+spec_store_bypass_disable=on
 ```
 
-which will apply SSB if seccomp is enabled by Firecracker.
+Unfortunately, this applies SSBD to all the other processes running on the
+host as well.
 
-On aarch64 systems, it is enabled by Firecracker
-[using the `prctl` interface][3]. However, this is only available on host
-kernels Linux >=4.17 and also Amazon Linux 4.14. Alternatively, a global
-mitigation can be enabled by adding the following Linux kernel boot parameter:
+###### ARM64
+
+On aarch64 systems, SSBD can be applied to the kernel by using the following
+kernel cmdline parameter:
+
+```console
+ssbd=kernel
+```
+
+SSBD is applied to Firecracker by [using the `prctl` interface][3].
+However, this is only available on host kernels Linux >=4.17 and also Amazon
+Linux 4.14. Alternatively, a global mitigation can be enabled by adding the
+following Linux kernel cmdline parameter:
 
 ```console
 ssbd=force-on
 ```
 
-Verification can be done by running:
+The following command can be used to check if SSBD is applied to Firecracker:
 
 ```bash
 cat /proc/$(pgrep firecracker | head -n1)/status | grep Speculation_Store_Bypass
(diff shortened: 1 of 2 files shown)


tests/integration_tests/security/spectre-mitigations
```diff
@@ -0,0 +1,30 @@
...
+    if [ $? != 0 ]; then
+        echo "SPECTRE V2 mitigated"
+    else
+        echo "FATAL: SPECTRE V2 not mitigated"
+        exit 1
+    fi
+else
+    echo "Unsupported arch $arch."
+    exit 1
+fi
```
pb8o, line 30: nit: newline at the end

tests/integration_tests/security/spectre-mitigations
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
```
pb8o, line 2: I would add `set -u` here to check for uninitialized variables.

tests/integration_tests/security/spectre-mitigations
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
pb8o, line 1: can we rename so it has a .sh extension?

I am thinking if we want to run shellcheck in the future that would be an easy way to collect all shell scripts: `find . -name "*.sh"`

docs/prod-host-setup.md
```diff
@@ -293,81 +293,56 @@ echo "KSM: ENABLED (Recommendation: DISABLED)"
...
+More information on the processors vulnerable to this type
 of attack and detailed information on the mitigations can be found in the
 [ARM security documentation](https://developer.arm.com/support/arm-security-updates/speculative-processor-vulnerability).
 
 Verification can be done by running:
-
 ```bash
-(grep -q "^Mitigation:" /sys/devices/system/cpu/vulnerabilities/spectre_v2 || \
-grep -q "^Not affected$" /sys/devices/system/cpu/vulnerabilities/spectre_v2) && \
+grep -q "^(Mitigation: CSV2, BHB|Not affected)$" /sys/devices/system/cpu/vulnerabilities/spectre_v2
```
bchalios, line 323: missing a `&& \` in the EOL here.

docs/prod-host-setup.md
```diff
@@ -293,81 +293,56 @@ echo "KSM: ENABLED (Recommendation: DISABLED)"
...
-to the kernel commandline of the host.
-
 ##### Bounds Check Bypass Store (Spectre V1)
 
 Verification for mitigation against Spectre V1 can be done:
-
 ```bash
-(grep -q "^Mitigation:" /sys/devices/system/cpu/vulnerabilities/spectre_v1 || \
-grep -q "^Not affected$" /sys/devices/system/cpu/vulnerabilities/spectre_v1) && \
+grep -q "^(Mitigation:|Not affected)$" /sys/devices/system/cpu/vulnerabilities/spectre_v1
```
bchalios, line 331: same here: `&&  \' in the end of line

tests/integration_tests/security/test_spectre.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pb8o, line 1: 2022 :)

tests/integration_tests/security/test_spectre.py
```diff
@@ -0,0 +1,19 @@
...
+
+@pytest.mark.timeout(100)
+def test_spectre_mitigations():
+    """
+    Run `spectre-mitigations` to check the kernel is compiled with SPECTREv2 mitigations.
+
+    @type: security
+    """
+    # Run command and raise exception if non-zero return code
+    utils.run_cmd("chmod a+x tests/integration_tests/security/spectre-mitigations", cwd=defs.FC_WORKSPACE_DIR)
```
bchalios, line 19: you sure this does what you want? It looks like you're just running `chmod a+x` on the `spectre-mitigations` script?

tests/integration_tests/security/spectre-mitigations
```diff
@@ -0,0 +1,30 @@
...
+# Checks SPECTREv2 mitigations are present.
+
+arch=`uname -m`
+
+if [ ${arch} = "x86_64" ]; then
+    # Validate SPECTRE V2 mitigations are enabled:
+    # - `IBPB: conditional` 
+    # - `eIBRS` (`Enhanced IBRS` intel, `Retpolines` AMD).
+    BODY=$(sudo cat /sys/devices/system/cpu/vulnerabilities/spectre_v2)
+    if [[ $BODY =~ "IBPB: conditional" ]] && [[ $BODY =~ "Enhanced IBRS" || $BODY =~ "Retpolines" ]]; then
```
bchalios, line 12: A comment above was saying that IBPB needs to be at least conditional. Here, we only check *only* for conditional.

tests/integration_tests/security/test_spectre.py
pb8o, line 19: So this will run the chmod but not the script.

The script should be +x in the repo, and as part of this PR. To change the permissions on Windows https://stackoverflow.com/questions/21691202/how-to-create-file-execute-mode-permissions-in-git-on-windows

tests/integration_tests/security/spectre-mitigations
JonathanWoollett-Light: As we see with `devtool` a few other scripts throughout the repo the standard is to not use the `.sh` extension.

bchalios, line 1: nit: why bash and not python?

bchalios (changes requested):

tests/integration_tests/security/spectre-mitigations
JonathanWoollett-Light: Updated to `if [[ $BODY =~ "IBPB: conditional" || $BODY =~ "IBPB: always-on" ]] && [[ $BODY =~ "Enhanced IBRS" || $BODY =~ "Retpolines" ]]`

tests/integration_tests/security/spectre-mitigations
JonathanWoollett-Light: It is a relatively short and simple script and simplifies many of the operations such as `grep -q "^(Mitigation: CSV2, BHB|Not affected)$" /sys/devices/system/cpu/vulnerabilities/spectre_v2`.

I think Python would also be suitable here, but due to these reasons and existing use of bash in the project (e.g. `devtool`) I went with this approach. I do not lean heavily either direction though so if you think it would be better suited in Python I could rewrite it as such.

tests/integration_tests/security/test_spectre.py
JonathanWoollett-Light: Fixed as mentioned under  https://github.com/firecracker-microvm/firecracker/pull/3036#discussion_r900064268

bchalios: If we go with the bash script, I' d go with @pb8o 's approach. Add the script in the repo with +x permissions and avoid doing `chmod` here.

pb8o: yeah I think it's clearer that way

JonathanWoollett-Light: I've updated it to only use python.

JonathanWoollett-Light: See https://github.com/firecracker-microvm/firecracker/pull/3036#discussion_r900152307

tests/integration_tests/security/spectre-mitigations
JonathanWoollett-Light: See https://github.com/firecracker-microvm/firecracker/pull/3036#discussion_r900152307

tests/integration_tests/security/test_spectre.py
bchalios: Ok, looks good to me now.

tests/integration_tests/security/spectre-mitigations
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
JonathanWoollett-Light: This test is useful for users verifying their integration meets our recommendations and for verifying our own testing/benchmarking environments meet these recommendations.

About this I would first ask, do we not use the same machines for CI as for performance tests? and secondly, do we not already run integration tests before beginning these performance tests?

I believe for these longer running performance tests it would be advisable to run all integration tests at the start. Taking this approach I believe is preferable to manual selection of specific tests and guarantees we are benchmarking a fully working executable as we intend it. If we do not already run integration tests before long running performance tests I would suggest this as another PR.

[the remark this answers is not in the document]
JonathanWoollett-Light: I do not think it is necessary to run these test on our CI AMIs as we already know these settings are enabled by default (I believe we also do not validate other default system settings we recommend). But I'm not very familiar here so would also appreciate additional input.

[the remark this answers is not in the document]
JonathanWoollett-Light: > I disagree with the notion that users should run our integration test in a production machine (where the security recommendations matter). Very often users just run in production pre-built binaries from a release, no need to build Firecracker there.
> 
> Having a tool to check production machines compliance to our recommendations is useful but I don't think our integration tests is the right tool to do so.

Relating to these points I would suggest that as we offer a pre-built binaries Firecracker we could also offer pre-built binaries for the Rust tests. This way a user could simply run this in the same environment as the Firecracker application binary. Among additional benefits believe this also efficiently fulfills the task of a tool to check machine compliance to our recommendations. Notably this could not  (or I am not aware of a simple way to) include Python tests, another motivation perhaps to move towards Rust for these integration tests.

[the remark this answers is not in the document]
pb8o: So the goal as I interpret it is that whenever we make performance tests, we should be running SPECTRE mitigations. Ideally we would check everything is enabled when we create the testing images, but right now we don't have a way to do this, so the next best thing is to run them every time we run integration tests.

I agree that this test is a bit of a kludge, as we know the images are already well configured, so in a sense this test is working as a canary, making sure we will notice if something changes. In the future when we can run validation during image creation we should move this test there and remove it from here.

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -25,6 +25,7 @@
 import framework.stats as st
 from integration_tests.performance.configs import defs
 from integration_tests.performance.utils import handle_failure
+from integration_tests.security.test_spectre import test_spectre_mitigations
```
pb8o, line 28: If you move it to conftest.py, then it does not need to get explicitly imported.

tests/integration_tests/security/test_spectre.py
```diff
@@ -0,0 +1,34 @@
...
+
+import pytest
+import platform
+
+from framework import utils
+from framework import defs
+
+
+@pytest.mark.timeout(50)
+@pytest.fixture(autouse=True)
```
pb8o, line 13: The fixture should have session scope. We only need to run it once per test session. 

```
@pytest.fixture(autouse=True, scope="session")
```

With those 2 things, and if you move it to `tests/integration_tests/performance/conftest.py`, I think you then don't need to make any changes anywhere else (no need to request the fixture explicitly), but then it will run for sessions. And it would only run once **iff** the performance tests are run.

pb8o (changes requested):

pb8o (dismissed):

tests/conftest.py
```diff
@@ -473,6 +473,28 @@ def test_multiple_microvms(test_fc_session_root_path, context, bin_cloner_path):
         shutil.rmtree(os.path.join(test_fc_session_root_path, microvms[i].id))
 
 
+@pytest.fixture(autouse=True, scope="session")
```
bchalios, line 476: As it is this right now, will it run every time tests are ran or only when we tests under the `performance` folder?

JonathanWoollett-Light: Tl;dr: The test is run once per each execution of pytest.

Looking at https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request it notes:

> “Autouse” fixtures are a convenient way to make all tests automatically request them

Meaning all tests become `my_test(..., test_spectre_mitigations)`.

Looking at https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session it notes:

> we can add a scope="module" parameter to the [@pytest.fixture](https://docs.pytest.org/en/6.2.x/reference.html#pytest.fixture) invocation to cause a smtp_connection fixture function, responsible to create a connection to a preexisting SMTP server, to only be invoked once per test module

and:

> Fixtures are created when first requested by a test, and are destroyed based on their scope:
> - `function`: the default scope, the fixture is destroyed at the end of the test.
> - `class`: the fixture is destroyed during teardown of the last test in the class.
> - `module`: the fixture is destroyed during teardown of the last test in the module.
> - `package`: the fixture is destroyed during teardown of the last test in the package.
> - `session`: the fixture is destroyed at the end of the test session.

Meaning this fixture (`test_spectre_mitigations`) is run once at the beginning of each test `session`, meaning each execution of pytest.

bchalios: so, if I'm developing on a machine that does not implement these mitigations and run the tests, will the tests fail, right? I thought this is something we don't want

pb8o: If we want to only run when a performance test is run, you can just move the fixture to `tests/integration_tests/performance/conftest.py`

pb8o (approved):

tests/conftest.py
JonathanWoollett-Light: @bchalios 

> so, if I'm developing on a machine that does not implement these mitigations and run the tests, will the tests fail, right? I thought this is something we don't want

Yes in this instance the tests would fail. This instance is extremely rare, I do not expect it to ever occur given all modern machines implement spectrev2 mitigations.

My proffered approach here is to use the simplest solution of one `conftest.py`. In my opinion it would not be worth adding the additional complexity of another `contest.py` for this issue as I expect what you mention will never occur.

However if you have a stronger opinion here I can make this change.

bchalios: It just seems weird to me to have as a requirement to implement production-level requirements on a dev machine. The way I see it is that the test is meant to make sure that our performance tests (and baselines) are being executed consistently on a production-like machine.

JonathanWoollett-Light: `./tools/devtool test` runs performance tests which require these mitigations to be implemented to be accurate. So in this case most of the time when running tests on a dev machine you will be running some performance tests (the tests required to merge a PR).

Notably adding a second `conftest.py` also introduces errors needing to be resolved around importing/re-exporting items, this may require some restructuring making it a non-trivial change.

bchalios: We were having discussions before to add messages for the users to let them know that performance testing is not reliable in non-supported platforms.

Anyways, we can add this now and re-visit that when/if the need arises.

bchalios (approved):

PR #3060, opened 2022-07-19
Update docs to provide better clarity of supported features

### LINKED ISSUE #3059: [Doc] Improve the documentation to reflect the max amount of vCPU supported
# Feature Request

Improve the documentation and give better clarity of how many vCPU are supported into Firecracker.

Documentation should state the max number of vCPU supported

N/A

## Additional context

As result of the [survey](https://github.com/firecracker-microvm/firecracker/discussions/2987), some of our customers were not aware that Firecracker support already up to 32 vCPU. As part of this issue we would like to improve our documentation to make such feature more explicit.

## Checks

- [X] Have you searched the Firecracker Issues database for similar requests?
- [X] Have you read all the existing relevant Firecracker documentation?
- [X] Have you read and understood Firecracker's core tenets?

xmarcalx:
Signed-off-by: Marco Cali <xmarcalx@amazon.co.uk>

# Reason for This PR

Modify the documentation to include explicitly the max number of vCPU supported and other minor grammatical changes.

Fixes #3059

## Description of Changes

Updated the "Features" section of the design.md file to state the max vCPU number. 

## License Acceptance

## PR Checklist
- [X] The issue which led to this PR has a clear conclusion.
- [X] This PR follows the solution outlined in the related issue.

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -18,8 +18,8 @@ Firecracker virtual machine manager (VMM).
 
 1. Firecracker can safely run workloads from different customers on the same
    machine.
-1. Customers can create microVMs with any combination of vCPU and memory to
-   match their application requirements.
+1. Customers can create microVMs with any combination of vCPU (up to 32)
+   and memory to match their application requirements.
 1. Firecracker microVMs can oversubscribe host CPU and memory. The degree of
    oversubscription is controlled by customers, who may factor in workload
    correlation and load in order to ensure smooth host system operation.
--- a/docs/prod-host-setup.md
+++ b/docs/prod-host-setup.md
@@ -82,10 +82,10 @@ for Firecracker processes that are unresponsive, and kills them, by SIGKILL.
 
 ## Jailer Configuration
 
-For assuring secure isolation in production deployments, Firecracker should
-must be started using the `jailer` binary that's part of each Firecracker
-release, or executed under process constraints equal or more restrictive than
-those in the jailer. For more about Firecracker sandboxing please see
+For assuring secure isolation in production deployments, Firecracker should be
+started using the `jailer` binary that's part of each Firecracker release, or
+executed under process constraints equal or more restrictive than those in the jailer.
+For more about Firecracker sandboxing please see
 [Firecracker design](design.md)
 
 The Jailer process applies
@@ -127,7 +127,7 @@ Here are some recommendations on how to limit the process's resources:
 
 - Jailer's `resource-limit` provides control on the disk usage through:
   - `fsize` - limits the size in bytes for files created by the process
-  - `no-file` - specifies a value one greater than the maximum file
+  - `no-file` - specifies a value greater than the maximum file
     descriptor number that can be opened by the process. If not specified,
     it defaults to 4096.
 


docs/design.md
```diff
@@ -18,7 +18,7 @@ Firecracker virtual machine manager (VMM).
 
 1. Firecracker can safely run workloads from different customers on the same
    machine.
-1. Customers can create microVMs with any combination of vCPU and memory to
+1. Customers can create microVMs with any combination of vCPU (up to 32) and memory to
```
mattschlebusch, line 21: I'm not sure a clarification on operational limitations/constraints should be put in the design document unless its a constraint that exists *because* of the design. My 2c.

xmarcalx: While I agree that having this information together with the design may be convoluted to be found, in this document we are specifying many other features/capabilities which we support because we do not have at the moment even any other better place where to do so (SPECIFICATION.md in reality state our promises in terms of performance more than features/capability).  Currently the limit comes from our design in Firecracker or better we imposed a soft limit of 32 [here](https://github.com/firecracker-microvm/firecracker/blob/1a61cbafa9e1dda6aef92b94f5beaffcd0c296ff/src/vmm/src/vmm_config/machine_config.rs#L11)

luminitavoicu (approved):

docs/prod-host-setup.md
```diff
@@ -82,10 +82,10 @@ for Firecracker processes that are unresponsive, and kills them, by SIGKILL.
 
 ## Jailer Configuration
 
-For assuring secure isolation in production deployments, Firecracker should
-must be started using the `jailer` binary that's part of each Firecracker
-release, or executed under process constraints equal or more restrictive than
-those in the jailer. For more about Firecracker sandboxing please see
+For assuring secure isolation in production deployments, Firecracker should be
+started using the `jailer` binary that's part of each Firecracker release, or 
+executed under process constraints equal or more restrictive than those in the jailer.
```
sandreim, line 87: This would make more sense if we refer to the `Threat Containment` section which defines the threat model. Using the `Jailer` is an integral part o that. Users do have some leeway, but this documentation should place such situations in the "I know what I am doing" basket.

dianpopa: The threat containment section is found in the design doc which we are referring to below. Maybe we can focus the attention of readers by changing the line  to
```
For more info about Firecracker sandboxing and threat containment please see [Firecracker design](design.md)
```
WDYT @sandreim?

docs/design.md
mattschlebusch: >  Currently the limit comes from our design in Firecracker
I don't agree that the limit comes from the design, but from the implementation.

> ... in this document we are specifying many other features/capabilities which we support because we do not have at the moment even any other better place ...

This is not a great justification for placing it here. The documentation is quite tough to navigate, better/correct sensical placement can make a difference. If that place doesn't exist yet, it can still be created.

xmarcalx: > I don't agree that the limit comes from the design, but from the implementation.
That is what i meant but our implementation reflect the limitation which we have in Firecracker because with bigger vCPU numbers we have degradation which will make Firecraker not usable. At the same time, this simple modification of that sentence is not claiming that a customer can use up to 32 vCPU due to the design and i can make it maybe explicit changing the bracket into "(with a soft limit up to 32)". design.md is the name of the file but in that file there is much more. Just read the first sentence "_The scope of this document is to describe the features and architecture of the Firecracker virtual machine manager (VMM)._"

> The documentation is quite tough to navigate, better/correct sensical placement can make a difference.
I do not think the documentation is tough to navigate or this change will affect the readability. Actually this file is well structured and present a section on Features that we intended to include in Firecracker. We can of course in future think about a separated file with just a series of features which we want to highlight and their corresponding values, but this to me still seem a valid incremental step in the direction to provide the max amount of information to our users now while may undertake in future the task of creating a specific page (which i guess is your current suggestion?)  and which need even more thoughts in order to finalize which other parameters need to be included.

docs/design.md
```diff
@@ -18,8 +18,8 @@ Firecracker virtual machine manager (VMM).
 
 1. Firecracker can safely run workloads from different customers on the same
    machine.
-1. Customers can create microVMs with any combination of vCPU and memory to
-   match their application requirements.
+1. Customers can create microVMs with any combination of vCPU (up to 32)
```
bchalios, line 21: where does this restriction come from? Maybe we should add a comment?

luminitavoicu (approved):

docs/design.md
dianpopa: The reason for this is mentioned [here](https://github.com/firecracker-microvm/firecracker/blob/firecracker-v1.1/src/vmm/src/vmm_config/machine_config.rs#L9).

I am not sure how to rephrase this to also include that especially because the first paragraph already mentions the lightweight microVM story. I am opened to suggestions.

bchalios: Ok, I see. To me, it's not a clear the distinction `lightweight VM <=> up to 32 vCPUs`, but I don't see a way out of this now.

bchalios (approved):

PR #3066, opened 2022-07-27
cpu templates: add T2S templates

alsrdn:
# Reason for This PR

Fix Guest CPU consistency issues when snapshotting and resuming uVMs across hosts with different CPUs.

## Description of Changes

The T2 template lets KVM populate the ARCH_CAPABILITIES MSR by using a
copy of the Host information. We sometimes want to force the information
contained in that MSR in order to make it look like a less capable
processor WRT speculative execution vulnerabilities and mitigations.
This is helpful when we want to snapshot a uVM on a newer processor and
resume on an older processor.
In this case, the software should use only the capabilities that are
available in both the newer and older host.

Create a new CPU template called T2S which is similar to the T2 template
but also sets additional information in the ARCH_CAPABILITIES MSR.
We basically set the MSR to expose to the guest a CPU that has the
security capabilities of a Skylake CPU, regardless of the host.
Also set some bits that were missing from the T2 CPUID config.

Adding the MSR information to the cpuid crate is not ideal based on
the current implementation. In the end the whole cpuid crate should
be refactored to remove the templates implementation into another crate.
The assumption that CPUID is the only place where information about the
processor is stored proves to be false so we need to extend the templates
to include at least the MSRs.

## License Acceptance

## PR Checklist

### DIFF
--- a/src/cpuid/src/template/intel/t2.rs
+++ b/src/cpuid/src/template/intel/t2.rs
@@ -8,7 +8,10 @@ use crate::cpu_leaf::*;
 use crate::template::intel::validate_vendor_id;
 use crate::transformer::*;
 
-fn update_feature_info_entry(entry: &mut kvm_cpuid_entry2, _vm_spec: &VmSpec) -> Result<(), Error> {
+pub(crate) fn update_feature_info_entry(
+    entry: &mut kvm_cpuid_entry2,
+    _vm_spec: &VmSpec,
+) -> Result<(), Error> {
     use crate::cpu_leaf::leaf_0x1::*;
 
     entry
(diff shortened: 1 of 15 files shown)


src/cpuid/src/template/intel/t2.rs
```diff
@@ -33,26 +36,31 @@ fn update_feature_info_entry(entry: &mut kvm_cpuid_entry2, _vm_spec: &VmSpec) ->
         .write_bit(ecx::MONITOR_BITINDEX, false)
         .write_bit(ecx::DS_CPL_SHIFT, false)
         .write_bit(ecx::VMX_BITINDEX, false)
+        .write_bit(ecx::SMX_BITINDEX, false)
```
mattschlebusch, line 39: Is it worth calling out the changes to the `t2` template in the changelog?

mattschlebusch (approved):
LGTM.

bchalios (dismissed):

src/cpuid/src/template/intel/t2.rs
alsrdn: Yes, I'll add a callout to the Changelog.

bchalios (approved):

mattschlebusch (approved):

PR #3079, opened 2022-08-03
Update project structure

JonathanWoollett-Light:
# Changes

Re-structured the the workspace.

This **significantly reduces the overall Jailer and Firecracker binary size (by ~250kb or ~7%)**, slightly reduces build time and integrates better with tooling (by using the standard layout and build directory).

  - Moved workspace from `/src/` to `/` and moved many non-member crates into modules (to help avoid a large number of crates in the root) within the `firecracker` crate, this reduces the number of top level items but more importantly allows the compiler to better understand the context which allows better identification of dead code.
    The intended usage of a workspace is to contain multiple crates which are 
    1. Published libraries.
    2. Shared between crates.
    3. Binaries.
    4. Technically required to be separate (e.g. proc-macro crates).
    
    Many of the previously separate crates did not meet these conditions. By moving them into modules we restrict their usage and allow better compiler inference about their behavior (which allows better linting by detecting unused code). In this case this raises quite a few warnings for unused code, in most of these instance here I have resolved these quickly and dirtily with `#[allow(dead_code)]` for now this silences the warning. I do this as these cases are often tricky with parts being used in `cfg(not(test))` and other parts in `cfg(test)` and for now want a relatively quick solution (this has no affect on the binary).
    This also:
    - Allows easier interaction with tooling by using a default layout and avoiding some otherwise necessary command line arguments.
    - Speeds up the build process by producing less build artifacts/clutter.
    - Centralizes dependency management allowing easier maintenance of dependencies (no need to update multiple links)
    - Centralizes documentation and configuration by requiring less duplicate files like `Cargo.toml` and giving a clear place to put documentation relating to specific user facing functionality (it would be preferable to have documentation under the specific crates rather than in the repository root).
  - Deleted `./cargo/config`: This can be done as all the options it sets are negative or unnecessary:
    - `[build.target]`: This would previously often break `cargo build` as it restricted usage to `x86_64-unknown-linux-musl` (subsequently this would break rust-analyzer) as most of the time the default Rust toolchain would not be `x86_64-unknown-linux-musl`. If we wanted to select this toolchain the better approach would be [The toolchain file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file). When creating a release we can specify the toolchain when building (e.g. `cargo +1.52.1-x86_64-unknown-linux-musl build`).
    - `[build.target-dir]`: The choice of build directory is arbitrary, it should not make a different whether `build/cargo_target` or the default directory is used. It seems this is set to match with `devtool`, both here and in `devtool` we gain nothing by changing this. Using the default here is the best option (for simplicity and tool interoperability).
    - `[net.git-fetch-with-cli]`: This is often set as a workaround for network failure when pulling crates, in a generic Linux environment this is not necessary (I do not know why this was used, my best guess is a historical workaround relating to docker, regardless this should not be set globally for anyone using Firecraker).

### Binary size

Package | Before | After | Change (% 2sf)
---|---|---|---
Firecracker | 2,495,800 | 2,012,944 | -19
Jailer |  850,216 | 1,087,192 | +22
**Sum** | **3,346,016** | **3,100,136** | **-7.3**

### Compile time

There is small reduction in compile time from ~37.7s to 36.6s on a AWS EC2 c5.metal instance.

```
sudo ./tools/devtool shell -p
cargo clean
cargo build  --all --release --target x86_64-unknown-linux-musl
```

# Reason

Significantly reduces the overall Jailer and Firecracker binary size (by ~250kb or ~7%), slightly reduces build time and integrates better with tooling.

## License Acceptance

## PR Checklist

---


dianpopa:
Did you remove the binding manually or re-run rust bindgen?

dianpopa:
Do we maybe want to combine this rust update with ubuntu update? I am thinking about saving some devctr entries?

JonathanWoollett-Light:
> Did you remove the binding manually or re-run rust bindgen?

I manually removed some of the assertions within some of the tests which where flagged as UB.

> Do we maybe want to combine this rust update with ubuntu update? I am thinking about saving some devctr entries?

I think in this case for making changes more incremental it would be better to keep these separate.

JonathanWoollett-Light:
Many tests require `sudo` and many require `--test-threads`.

In the case of the `--test-threads-1` it is likely these tests could be adjusted to not require this, this may however require significant work and take significant time, so for now I simply run the tests with this argument.

E.g. :
```
jonathan@jonathan-pc:~/Projects/firecracker$ cargo test --no-run
warning: patch for `kvm-bindings` uses the features mechanism. default-features and features will not take effect because the patch dependency does not support this mechanism
warning: No default seccomp policy for target: x86_64-unknown-linux-gnu. Defaulting to `resources/seccomp/unimplemented.json`.
   Compiling firecracker v1.1.0 (/home/jonathan/Projects/firecracker/firecracker)
    Finished test [unoptimized + debuginfo] target(s) in 5.49s
  Executable unittests src/lib.rs (target/debug/deps/firecracker-b81536796be881b8)
  Executable unittests src/main.rs (target/debug/deps/firecracker-78ec52f9d8f007a4)
  Executable unittests src/main.rs (target/debug/deps/jailer-73cd82bda598bde6)
  Executable unittests src/lib.rs (target/debug/deps/net_gen-a88b69c85993db6c)
  Executable unittests src/main.rs (target/debug/deps/rebase_snap-25099416fd78b37c)
  Executable unittests src/lib.rs (target/debug/deps/seccompiler-846b62e75929cdc8)
  Executable unittests src/seccompiler_bin.rs (target/debug/deps/seccompiler_bin-400ebfb834494ac1)
  Executable unittests src/lib.rs (target/debug/deps/utils-a9fe913f87ee0a11)
jonathan@jonathan-pc:~/Projects/firecracker$ sudo ./target/debug/deps/firecracker-b81536796be881b8 --test-threads=1
...
```
At present with Firecracker we get:
```
failures:
    vm_memory_ext::tests::test_create_guest_memory
    vmm::tests::test_dirty_bitmap_success
    vmm::vmm_config::logger::tests::test_init_logger
    vmm::vmm_config::metrics::tests::test_init_metrics
    vmm::vstate::vcpu::tests::test_vcpu_pause_resume
    vmm::vstate::vcpu::tests::test_vcpu_save_restore_state_events
```
Of these tests, only 3 of them fail when run on their own (e.g. `sudo ./target/debug/deps/firecracker-b81536796be881b8 test_create_guest_memory -- --test-threads=1`), these 3 are:

1. `vmm::tests::test_dirty_bitmap_success`
2. `vmm::vstate::vcpu::tests::test_vcpu_pause_resume`
3. `vmm::vstate::vcpu::tests::test_vcpu_save_restore_state_events`

JonathanWoollett-Light:
## Build status

Changes have not been made to  `devtool` although are required. It would be useful to discuss and possibly work with someone more familiar with `devtool` here.

dianpopa:
Since the plan is to do this incrementally, would it make sense to close this?

Thanks!

PR #3095, opened 2022-08-17
CI: enhance allowed CPU features on ARM guests

luminitavoicu:

# Reason for This PR

CI: enhance allowed CPU features on ARM guests

## Description of Changes

## License Acceptance

## PR Checklist

### DIFF
--- a/tests/integration_tests/functional/test_cpu_features.py
+++ b/tests/integration_tests/functional/test_cpu_features.py
@@ -26,10 +26,17 @@ def _check_cpuid_x86(test_microvm, expected_cpu_count, expected_htt):
 
 
 def _check_cpu_features_arm(test_microvm):
-    expected_cpu_features = {
-        "Flags": "fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp "
-        "asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs",
-    }
+    if utils.get_instance_type() == "m6g.metal":
+        expected_cpu_features = {
+            "Flags": "fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp "
+            "asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs",
+        }
+    else:
+        expected_cpu_features = {
+            "Flags": "fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp "
+            "asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp "
+            "sha512 asimdfhm dit uscat ilrcpc flagm ssbs",
+        }
 
     utils.check_guest_cpuid_output(
         test_microvm, "lscpu", None, ":", expected_cpu_features


tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -26,10 +26,20 @@ def _check_cpuid_x86(test_microvm, expected_cpu_count, expected_htt):
...
 
 def _check_cpu_features_arm(test_microvm):
-    expected_cpu_features = {
-        "Flags": "fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp "
-        "asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs",
-    }
+    _, stdout, stderr = utils.run_cmd("lscpu")
+    assert stderr == ""
+
+    if "sve" in stdout:
```
dianpopa, line 32: I d rather use the instance type to compute expected guest cpu features rather than the presence of sve. It is more future proof IMO.

luminitavoicu: You are right! Done.

dianpopa (approved):

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -26,10 +26,17 @@ def _check_cpuid_x86(test_microvm, expected_cpu_count, expected_htt):
 
 
 def _check_cpu_features_arm(test_microvm):
-    expected_cpu_features = {
-        "Flags": "fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp "
-        "asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs",
-    }
+    if utils.get_instance_type() == "m6g.metal":
+        expected_cpu_features = {
+            "Flags": "fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp "
```
mattschlebusch, line 31: Not a blocker at all, but I am quite curious how these feature flags were arrived at. Any points on requirements and what the constraints are would be  useful retrospectively.

mattschlebusch (approved):

bchalios (approved):
Obviously, I don't want to block this now, but maybe in the future it would be worth looking into something like this: https://github.com/robbmcleod/cpufeature

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -26,10 +26,17 @@ def _check_cpuid_x86(test_microvm, expected_cpu_count, expected_htt):
...
+            "asimdhp cpuid asimdrdm lrcpc dcpop asimddp ssbs",
+        }
+    else:
```
pb8o, line 34: Maybe we can do sthg like:

```python
if m6g.metal:
   ...
if m7g.metal:
   ...
else:
   raise Exception("Unexpected instance")
```

that way it is explicit which flags belong to which instance type

mattschlebusch:
> Obviously, I don't want to block this now, but maybe in the future it would be worth looking into something like this: https://github.com/robbmcleod/cpufeature

This looks like a good find.. We are using Intel's [msr-tools](https://github.com/intel/msr-tools) at the moment but it is deprecated and limited.

dianpopa:
> Obviously, I don't want to block this now, but maybe in the future it would be worth looking into something like this: https://github.com/robbmcleod/cpufeature

This is nice but it might only work on x86 (at least that is what their doc says)

tests/integration_tests/functional/test_cpu_features.py
luminitavoicu: These are the CPU features that KVM exposes to the guest. For `m6g.metal` the list is the same on the host and the guest, but this is not the case for all platforms because sometimes KVM disables some features by default.

PR #3096, opened 2022-08-18
Simplify sh shebang and syntax for pre-commit

mattschlebusch:
# Reason for This PR

Pre-commit hook can currently fail due to differences in how bash/sh and a POSIX environment handles boolean expressions and string handling.

## Description of Changes

* Drop quotes from `./tests/fmt.toml` to simplify string handling in `pre-commit` script.
* Homogenize script environment to `sh` (`pre-commit` was configured for **bash`** but `rusty-hook.toml` was boostrapping the script with `sh` which override the shebang)

## License Acceptance


JonathanWoollett-Light (approved):
LGTM

pre-commit
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
```
pb8o, line 1: FWIW, all the other scripts are bash, so shouldn't be an issue.

pb8o (approved):

pre-commit
pb8o: nevermind, I read the motivation now :)

pb8o (approved):

PR #3097, opened 2022-08-19
Comply with TOML string requirements

mattschlebusch:
## Reason for this Change

The pre-commit script used to manually parse the key-value pairs in the file, which required strings as unquoted when passed in manually to the `rustmt --config ...` command. This meant string manipulation when parsing parameters from the `tests/fmt.toml` file. With the `--config-path` parameter for `rustfmt`, it is possible to leave all the string values in the TOML file as quoted.

## Description of Changes

* Remove string handling from code that make use of `fmt.toml`. This requires the use of rustfmt's --config-path option to be able to use the string values as quoted as per the TOML standard.
  * This required the addition of the Rust *nightly* toolchain in addition to the *stable* toolchain in the development container.

* Drops the license-template-path
formatting parameter as it is no longer supported by the
rust toolchain. See github discussion:
https://github.com/rust-lang/rustfmt/issues/3352

## License Acceptance

### DIFF
--- a/pre-commit
+++ b/pre-commit
@@ -23,10 +23,10 @@ for i in $(git diff --name-only --cached); do
     extension="${filename##*.}"
     if [ "$extension" = "rs" ]; then
         # We first do a check run, this will fail when it finds a non-matching license.
-        rustfmt $i +nightly --check --config-path ./tests/fmt.toml
+        rustfmt $i --check --config-path ./tests/fmt.toml
 
         # Run `cargo fmt` for this source file using the fmt.toml config
-        rustfmt $i +nightly --config-path ./tests/fmt.toml
+        rustfmt $i --config-path ./tests/fmt.toml
     fi
     if [ "$extension" = "py" ]; then
         # Run `black` for this file
--- a/pre-commit
+++ b/pre-commit
@@ -22,12 +22,11 @@ for i in $(git diff --name-only --cached); do
     filename=$(basename -- "$i")
     extension="${filename##*.}"
     if [ "$extension" = "rs" ]; then
-        # Read rustfmt config, replace '\n' with ','
-        rustfmt_config="$(sed -z "s/\n/,/g;s/,$/\n/" ./tests/fmt.toml)"
         # We first do a check run, this will fail when it finds a non-matching license.
-        rustfmt $i --check --config $rustfmt_config
-        # Run `cargo fmt` for this file
-        rustfmt $i --config $rustfmt_config
+        rustfmt $i +nightly --check --config-path ./tests/fmt.toml
+
+        # Run `cargo fmt` for this source file using the fmt.toml config
+        rustfmt $i +nightly --config-path ./tests/fmt.toml
     fi
     if [ "$extension" = "py" ]; then
         # Run `black` for this file
--- a/tests/fmt.toml
+++ b/tests/fmt.toml
@@ -1,9 +1,12 @@
+# This file can be used as a whole TOML file via rustfmt's --config-path option
+# The individual key-pairs can also used with the rustfmt --config option,
+# but requires removing the quotes to work
+
 comment_width=100
 wrap_comments=true
 format_code_in_doc_comments=true
 format_strings=true
-license_template_path=./tests/license.txt
-imports_granularity=Module
+imports_granularity="Module"
 normalize_comments=true
 normalize_doc_attributes=true
-group_imports=StdExternalCrate
+group_imports="StdExternalCrate"
--- a/tests/integration_tests/style/test_rust.py
+++ b/tests/integration_tests/style/test_rust.py
@@ -12,10 +12,8 @@ def test_rust_style():
     @type: style
     """
 
-    #  ../src/io_uring/src/bindings.rs
-    config = open("fmt.toml", encoding="utf-8").read().replace("\n", ",")
     # Check that the output is empty.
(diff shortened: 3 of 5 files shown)


pb8o (approved):

tests/fmt.toml
```diff
@@ -1,9 +1,12 @@
+# This file can be used as a whole TOML file via rustfmt's --config-path option
```
pb8o, line 1: Shouldn't this file be named `rustfmt.toml`? That way it could automatically be picked up. https://github.com/rust-lang/rustfmt#configuring-rustfmt Would the nightly requirement apply as well?

tools/devctr/Dockerfile.aarch64
```diff
@@ -99,6 +99,8 @@ RUN cd "$TMP_POETRY_DIR" \
 RUN mkdir "$TMP_BUILD_DIR" \
     && curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --default-toolchain "$RUST_TOOLCHAIN" \
         && rustup target add aarch64-unknown-linux-musl \
+        && rustup install nightly \
```
pb8o, line 102: hmm this adds a third Rust runtime to the Docker image

mattschlebusch: That's correct. For whatever reason, `rust-lang` puts a number of the formatter options in the *nightly* toolchain. Using *stable* yields errors such as the following:

```
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `format_code_in_doc_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `comment_width = 100`, unstable features are only available in nightly channel.
Warning: can't set `normalize_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `normalize_doc_attributes = true`, unstable features are only available in nightly channel.
Warning: can't set `format_strings = true`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
```

tests/fmt.toml
mattschlebusch: This is probably a better option. 

This file contains config that is only available in the nightly toolchain which someone using rustfmt may not want to apply or check with?

pb8o: Good point... dev laptops running stable may get constant warnings of unstable options...

tools/devctr/Dockerfile.x86_64
```diff
@@ -108,6 +108,7 @@ RUN mkdir "$TMP_BUILD_DIR" \
     && curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --default-toolchain "$RUST_TOOLCHAIN" \
         && rustup target add x86_64-unknown-linux-musl \
         && rustup component add rustfmt clippy clippy-preview \
+        && rustup install nightly \
```
pb8o: If we add it, can we do it with `--profile minimal`?

tests/fmt.toml
```diff
@@ -1,9 +1,12 @@
...
+# The individual key-pairs can also used with the rustfmt --config option,
+# but requires removing the quotes to work
+
 comment_width=100
```
pb8o, line 5: Maybe add a comment saying all those options are unstable and require nightly channel

tests/integration_tests/style/test_rust.py
```diff
@@ -12,10 +12,8 @@ def test_rust_style():
     @type: style
     """
 
-    #  ../src/io_uring/src/bindings.rs
-    config = open("fmt.toml", encoding="utf-8").read().replace("\n", ",")
     # Check that the output is empty.
-    _, stdout, _ = utils.run_cmd(f"cargo fmt --all -- --check --config {config}")
+    _, stdout, _ = utils.run_cmd("cargo fmt --all -- --check --config-path fmt.toml")
```
pb8o: Shouldn't this also need to be run with `+nightly`?

mattschlebusch: This would have, but I have dropped nightly from the toolchain here. This generates warnings, but the test doesn't vet or fail for warnings.

pre-commit
```diff
@@ -22,12 +22,11 @@ for i in $(git diff --name-only --cached); do
...
-        # Read rustfmt config, replace '\n' with ','
-        rustfmt_config="$(sed -z "s/\n/,/g;s/,$/\n/" ./tests/fmt.toml)"
         # We first do a check run, this will fail when it finds a non-matching license.
-        rustfmt $i --check --config $rustfmt_config
-        # Run `cargo fmt` for this file
-        rustfmt $i --config $rustfmt_config
+        rustfmt $i --check --config-path ./tests/fmt.toml
+
+        # Run `cargo fmt` for this source file using the fmt.toml config
+        rustfmt $i --config-path ./tests/fmt.toml
```
JonathanWoollett-Light, line 29: The primary reason I used this approach was essentially to circumvent needing to include nightly. If you are including nightly it might be best to simply use a `rustfmt.toml` file (see https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=).

JonathanWoollett-Light (approved):

mattschlebusch:
There is legitimate reasons around this change but a revision around how we consume the nightly features in the formatter are in order. This isn't urgent and doesn't warrant the time for it at this stage.

PR #3098, opened 2022-08-21
refactor: use slab for user_data instead of box

ihciah:
1. use slab for user_data instead of boxto avoid unsafe code.
2. move generic type from push/pop to IoUring struct
  Put generic type on push and pop may cause developers misuse it since current user_data is a pointer, and the type should be the same. Since we always use the same user_data type for an IoUring instance, I move the generic to the IoUring struct.

Signed-off-by: ihciah <ihciah@gmail.com>

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/io_uring/mod.rs
+++ b/src/vmm/src/io_uring/mod.rs
@@ -169,11 +169,11 @@ impl<T: Debug> IoUring<T> {
         // validate that we actually did register fds
         let fd = op.fd();
         match self.registered_fds_count {
-            0 => Err((IoUringError::NoRegisteredFds, op.user_data())),
-            len if fd >= len => Err((IoUringError::InvalidFixedFd(fd), op.user_data())),
+            0 => Err((IoUringError::NoRegisteredFds, op.user_data)),
+            len if fd >= len => Err((IoUringError::InvalidFixedFd(fd), op.user_data)),
             _ => {
                 if self.num_ops >= self.cqueue.count() {
-                    return Err((IoUringError::FullCQueue, op.user_data()));
+                    return Err((IoUringError::FullCQueue, op.user_data));
                 }
                 self.squeue
                     .push(op.into_sqe(&mut self.slab))
@@ -186,14 +186,14 @@ impl<T: Debug> IoUring<T> {
                         (
                             IoUringError::SQueue(sqe_err),
                             // We don't use slab.try_remove here for 2 reasons:
-                            // 1. user_data was insertde in slab with step `op.into_sqe` just
+                            // 1. user_data was inserted in slab with step `op.into_sqe` just
                             //    before the push op so the user_data key should be valid and if
                             //    key is valid then `slab.remove()` will not fail.
                             // 2. If we use `slab.try_remove()` we'll have to find a way to return
                             //    a default value for the generic type T which is difficult because
                             //    it expands to more crates which don't make it easy to define a
                             //    default/clone type for type T.
-                            // So beleiving that `slab.remove` won't fail we don't use
+                            // So believing that `slab.remove` won't fail we don't use
                             // the `slab.try_remove` method.
                             #[allow(clippy::cast_possible_truncation)]
                             self.slab.remove(user_data_key as usize),
--- a/src/vmm/src/io_uring/mod.rs
+++ b/src/vmm/src/io_uring/mod.rs
@@ -78,7 +78,7 @@ impl IoUringError {
 
 /// Main object representing an io_uring instance.
 #[derive(Debug)]
-pub struct IoUring {
+pub struct IoUring<T> {
     registered_fds_count: u32,
     squeue: SubmissionQueue,
     cqueue: CompletionQueue,
@@ -91,9 +91,10 @@ pub struct IoUring {
     // The total number of ops. These includes the ops on the submission queue, the in-flight ops
     // and the ops that are in the CQ, but haven't been popped yet.
     num_ops: u32,
+    slab: slab::Slab<T>,
 }
 
-impl IoUring {
+impl<T: Debug> IoUring<T> {
     /// Create a new instance.
     ///
     /// # Arguments
@@ -136,6 +137,8 @@ impl IoUring {
 
         let squeue = SubmissionQueue::new(fd, &params).map_err(IoUringError::SQueue)?;
         let cqueue = CompletionQueue::new(fd, &params).map_err(IoUringError::CQueue)?;
+        let slab =
+            slab::Slab::with_capacity(params.sq_entries as usize + params.cq_entries as usize);
 
         let mut instance = Self {
             squeue,
@@ -143,6 +146,7 @@ impl IoUring {
             fd: file,
             registered_fds_count: 0,
             num_ops: 0,
+            slab,
         };
 
         instance.check_operations()?;
@@ -161,11 +165,7 @@ impl IoUring {
     }
 
     /// Push an [`Operation`](operation/struct.Operation.html) onto the submission queue.
-    ///
-    /// # Safety
-    /// Unsafe because we pass a raw user_data pointer to the kernel.
-    /// It's up to the caller to make sure that this value is ever freed (not leaked).
-    pub unsafe fn push<T: Debug>(&mut self, op: Operation<T>) -> Result<(), (IoUringError, T)> {
+    pub fn push(&mut self, op: Operation<T>) -> Result<(), (IoUringError, T)> {
         // validate that we actually did register fds
         let fd = op.fd();
         match self.registered_fds_count {
@@ -176,14 +176,28 @@ impl IoUring {
                     return Err((IoUringError::FullCQueue, op.user_data()));
                 }
                 self.squeue
-                    .push(op.into_sqe())
+                    .push(op.into_sqe(&mut self.slab))
                     .map(|res| {
                         // This is safe since self.num_ops < IORING_MAX_CQ_ENTRIES (65536)
                         self.num_ops += 1;
                         res
                     })
-                    .map_err(|err_tuple: (SQueueError, T)| -> (IoUringError, T) {
-                        (IoUringError::SQueue(err_tuple.0), err_tuple.1)
+                    .map_err(|(sqe_err, user_data_key)| -> (IoUringError, T) {
+                        (
+                            IoUringError::SQueue(sqe_err),
+                            // We don't use slab.try_remove here for 2 reasons:
+                            // 1. user_data was insertde in slab with step `op.into_sqe` just
+                            //    before the push op so the user_data key should be valid and if
+                            //    key is valid then `slab.remove()` will not fail.
+                            // 2. If we use `slab.try_remove()` we'll have to find a way to return
+                            //    a default value for the generic type T which is difficult because
+                            //    it expands to more crates which don't make it easy to define a
+                            //    default/clone type for type T.
+                            // So beleiving that `slab.remove` won't fail we don't use
+                            // the `slab.try_remove` method.
+                            #[allow(clippy::cast_possible_truncation)]
+                            self.slab.remove(user_data_key as usize),
+                        )
                     })
             }
         }
@@ -191,14 +205,9 @@ impl IoUring {
 
     /// Pop a completed entry off the completion queue. Returns `Ok(None)` if there are no entries.
     /// The type `T` must be the same as the `user_data` type used for `push`-ing the operation.
-    ///
-    /// # Safety
-    /// Unsafe because we reconstruct the `user_data` from a raw pointer passed by the kernel.
-    /// It's up to the caller to make sure that `T` is the correct type of the `user_data`, that
-    /// the raw pointer is valid and that we have full ownership of that address.
-    pub unsafe fn pop<T: Debug>(&mut self) -> Result<Option<Cqe<T>>, IoUringError> {
+    pub fn pop(&mut self) -> Result<Option<Cqe<T>>, IoUringError> {
         self.cqueue
-            .pop()
+            .pop(&mut self.slab)
             .map(|maybe_cqe| {
                 maybe_cqe.map(|cqe| {
                     // This is safe since the pop-ed CQEs have been previously pushed. However
@@ -391,8 +400,8 @@ mod tests {
     use super::*;
     use crate::vstate::memory::{Bytes, MmapRegion};
 
-    fn drain_cqueue(ring: &mut IoUring) {
-        while let Some(entry) = unsafe { ring.pop::<u32>().unwrap() } {
+    fn drain_cqueue(ring: &mut IoUring<u32>) {
+        while let Some(entry) = ring.pop().unwrap() {
             entry.result().unwrap();
 
             // Assert that there were no partial writes.
@@ -581,9 +590,7 @@ mod tests {
                             ring.submit_and_wait_all().unwrap();
                             drain_cqueue(&mut ring);
                         }
-                        unsafe {
-                            ring.push(operation).unwrap();
-                        }
+                        ring.push(operation).unwrap();
                     }
 
                     // Submit any left async ops and wait.
--- a/src/vmm/src/io_uring/operation/cqe.rs
+++ b/src/vmm/src/io_uring/operation/cqe.rs
@@ -13,22 +13,13 @@ unsafe impl ByteValued for io_uring_cqe {}
 #[derive(Debug)]
 pub struct Cqe<T> {
     res: i32,
-    user_data: Box<T>,
+    user_data: T,
 }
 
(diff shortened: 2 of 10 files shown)


src/devices/src/virtio/block/io/async_io.rs
```diff
@@ -27,7 +27,7 @@ pub enum Error {
 
 pub struct AsyncFileEngine<T> {
     file: File,
-    ring: IoUring,
+    ring: IoUring<WrappedUserData<T>>,
     completion_evt: EventFd,
     phantom: PhantomData<T>,
```
alindima, line 32: Do we need this any more?

src/io_uring/src/lib.rs
```diff
@@ -142,13 +143,15 @@ impl IoUring {
 
         let squeue = SubmissionQueue::new(fd, &params).map_err(Error::SQueue)?;
         let cqueue = CompletionQueue::new(fd, &params).map_err(Error::CQueue)?;
+        let slab = slab::Slab::with_capacity((params.sq_entries + params.cq_entries) as usize);
```
alindima, line 146: check for overflow

src/io_uring/src/lib.rs
```diff
@@ -184,29 +183,27 @@ impl IoUring {
...
+                            self.slab.remove(err_tuple.1 as usize),
+                        )
                     })
             }
         }
     }
 
     /// Pop a completed entry off the completion queue. Returns `Ok(None)` if there are no entries.
     /// The type `T` must be the same as the `user_data` type used for `push`-ing the operation.
-    ///
```
alindima, line 202: We can't just remove all usage of `unsafe`. There could still be undefined behavior because we pass pointers of the user data entries to the kernel.

What if we free the data before the kernel passes us back the pointer (the kernel could do a use-after-free).
What if the kernel frees the data before passing us back the pointer (we could do a double-free).

alindima (commented):
I believe this only makes sense if there is a noticeable performance improvement. You'd need to to some fio testing comparing the results, ideally with file backed on an an NVME drive.

Here are the performance tests we have in Firecracker for block: https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/performance/test_block_performance.py

Note that the baselines are gathered on specific kernels and specific aws metals

ihciah:
Thanks for reviewing. It has been a while since this pr is blocked by code coverage threshold.
I will rebase this PR this weekend and try to solve the comments.

src/io_uring/src/lib.rs
ihciah: In this PR I put a slab index inside the entry's userdata, so there's no need to do pointer derefrence.

src/devices/src/virtio/block/io/async_io.rs
ihciah: Yes. I moved generic to IoUring struct it self. It helps removing unsafe code.

src/io_uring/src/lib.rs
ihciah: Maybe there is no need for checking overflow here.
We have do syscall with the same parameters and it succeed. So sq_entries + cq_entries will never overflow usize. Otherwise the previous syscall will return ENOMEM.
I added some comment about it.

ihciah:
> I believe this only makes sense if there is a noticeable performance improvement.

I think so. One alloc and dealloc cost is relativly small comparing to the whole io process. The performance may not improve significantly. But it also helps removing some unsafe code.

> You'd need to to some fio testing comparing the results, ideally with file backed on an an NVME drive.
> Note that the baselines are gathered on specific kernels and specific aws metals

Well it may be difficult, I don't have required bare-metal resources...
I tried some hack on `test_block_performance.py` to treat my current kernel as 5.10 and use a rewriten instance type, but it seems not enough to let the test run. Can you help me test it? It there a way to let me compare performance on my own hardware?

src/devices/src/virtio/block/io/async_io.rs
alindima: I meant that I think we can remove the Phantom now

src/io_uring/src/lib.rs
alindima: >We have do syscall with the same parameters and it succeed

It's not about overflowing usize in this case, because you only cast to `usize` after the addition, it's about overflowing `u32`.

So I suggest you either add overflow check or just cast both values to usize before doing the addition (with `into()` instead of `as`)

src/io_uring/src/lib.rs
alindima: you're right.

but another issue comes to mind now:

you are using `remove`, which can panic if for example there is a bug in the kernel. This brings down the entire VM instead of trying to limit the blast radius to only this device.
Maybe use `try_remove` instead and log an error if this returns `None`

alindima:
> I think so. One alloc and dealloc cost is relativly small comparing to the whole io process. The performance may not improve significantly. But it also helps removing some unsafe code.

it needs to be measured in practice, it scales with the number of block devices and allocates more memory upfront instead of allocating it when needed. Maybe the block device has little usage in terms of throughput but we allocate space for all user entries corresponding to a full virtio queue unnecessarily.

> Well it may be difficult, I don't have required bare-metal resources...
I tried some hack on test_block_performance.py to treat my current kernel as 5.10 and use a rewriten instance type, but it seems not enough to let the test run. Can you help me test it? It there a way to let me compare performance on my own hardware?

I think you'll need somebody from the core team to help in this case. But it's no easy feat, it takes time and effort.
I'm not part of the core team anymore, so you'd need @xmarcalx or somebody from the team to help with this.

I believe that if there isn't a significant increase in memory usage (for rarely used block devices) or boot time, this may be justified to help us get rid of unsafe code. But it'd still need performance measurements to make sure we don't introduce regressions

xmarcalx:
Hi @alindima  and @ihciah ,

sorry for the late reply, somehow i missed the notification.
First of all thanks a lot @ihciah  for the proposed PR and @alindima for helping with this review! 

We discussed this PR today with the team and somehow they share the same sentiment that removing unsafe code can be an improvement but yes we have concerns about performances. At the moment, io_uring is not GA and it is developer preview that is why was de-prioritized. 
I think we will not be able to support this benchmark activities at least till June, sorry.


I see btw there are still some conversation unresolved, like use the try_remove() to avoid to bring down the whole VMM thread. @ihciah are you planning to address those?

src/vmm/src/io_uring/mod.rs
```diff
@@ -176,29 +176,33 @@ impl IoUring {
...
+                    .map_err(|err_tuple: (SQueueError, u64)| -> (IoUringError, T) {
+                        (
+                            IoUringError::SQueue(err_tuple.0),
+                            #[allow(
+                                clippy::cast_possible_truncation,
+                                clippy::unnecessary_literal_unwrap
+                            )]
+                            self.slab
+                                .try_remove(err_tuple.1 as usize)
+                                .unwrap_or_else(|| None::<T>.unwrap()),
```
ShadowCurse, line 194: This is a very complicated way of unwrapping a value. Shouldn't an error be returned if value can not be removed from a slab? Maybe this whole `map` and `map_err` can be converted into simple `match` so that the code is a bit more clear?

src/vmm/src/io_uring/operation/mod.rs
```diff
@@ -51,7 +51,7 @@ pub struct Operation<T> {
     pub(crate) len: Option<u32>,
     flags: u8,
     pub(crate) offset: Option<u64>,
-    user_data: Box<T>,
+    user_data: T,
```
ShadowCurse, line 54: From what I understand, the `user_data` field here is just a key into a slab. I think it should be renamed into `data_key`(or smth similar). Also generic `<T>` can be removed from the `Operation<T>` as well, as no longer needed.

src/vmm/src/io_uring/operation/mod.rs
```diff
@@ -119,7 +119,7 @@ impl<T: Debug> Operation<T> {
 
     /// Consumes the operation and returns the associated `user_data`.
     pub fn user_data(self) -> T {
-        *self.user_data
+        self.user_data
```
ShadowCurse, line 122: I think this getter does not need to be here. Just make `user_data` a public field.

src/vmm/src/io_uring/queue/completion.rs
```diff
@@ -93,7 +92,13 @@ impl CompletionQueue {
             self.unmasked_head += Wrapping(1u32);
             ring.store(self.unmasked_head.0, self.head_off, Ordering::Release)?;
 
-            Ok(Some(Cqe::new(cqe)))
+            let res = cqe.res;
+            #[allow(clippy::cast_possible_truncation)]
+            let index = cqe.user_data as usize;
+            #[allow(clippy::unnecessary_literal_unwrap)]
+            let user_data = slab.try_remove(index).unwrap_or_else(|| None::<T>.unwrap());
```
ShadowCurse, line 99: Here as well the return of `try_remove` should either be converted into a error or unwrapped with `expect` if error can not be handled.

src/vmm/src/io_uring/operation/sqe.rs
```diff
@@ -24,13 +24,9 @@ impl Sqe {
...
     }
 
-    /// Consume the sqe and return the `user_data`.
-    ///
-    /// # Safety
-    /// Safe only if you guarantee that this is a valid pointer to some memory
-    /// where there is a value of type `T` created from a [`Box<T>`].
-    pub(crate) unsafe fn user_data<T: Debug>(self) -> T {
-        *Box::from_raw(self.0.user_data as *mut T)
+    /// Return the `user_data`.
```
ShadowCurse, line 27: It would be worth adding a comment that `user_data` represents a `key` into a slab.

src/vmm/src/io_uring/mod.rs
sudanl0: > Shouldn't an error be returned if value can not be removed from a slab?

Returning an error would mean that instead of returning
```
(IoUringError::SQueue(err_tuple.0), None::<T>.unwrap())
```
we would be returning something like 
```
(IoUringError::SQueueSlab(err_tuple.0), None::<T>.unwrap())
```

In both the cases, the main focus of investigation should be the SQueue errors and the second return value None is enough to identify that there was also an error while removing items from slab.
So I don't see a value in adding a new error type just for something that we already know from existing data.

src/vmm/src/io_uring/operation/mod.rs
sudanl0: `user_data` used with `Operation` is the actual data, the key is stored in `Sqe.0.user_data`. I can change that to make things clear.

src/vmm/src/io_uring/mod.rs
ShadowCurse: `None::<T>.unwrap()` is just a fancy way of saying `panic!`.  So if it is better to panic here, then use `panic!` with a message.

roypat: > `None::<T>.unwrap()` is just a fancy way of saying `panic!`. So if it is better to panic here, then use `panic!` with a message.

At this point, we don't need to use `try_remove` anymore, we can just use `.remove()` and have the panic message be generated by the `slab` crate.

But I don't quite understand why we would want to panic here. As already pointed out in https://github.com/firecracker-microvm/firecracker/pull/3098#discussion_r1126587198, bringing down the entire VM because the kernel decided to mess up something in io_uring doesn't seem right. Better to return an error and mark the device as failed, so the guest can try to reset it and potentially recover imo.

src/vmm/src/io_uring/queue/completion.rs
```diff
@@ -93,7 +92,13 @@ impl CompletionQueue {
             self.unmasked_head += Wrapping(1u32);
             ring.store(self.unmasked_head.0, self.head_off, Ordering::Release)?;
 
-            Ok(Some(Cqe::new(cqe)))
+            let res = cqe.res;
+            #[allow(clippy::cast_possible_truncation)]
+            let index = cqe.user_data as usize;
```
roypat, line 97: If the user data is always an index into a slab, and `slab` wants its indices to be `usize`, can we just define the type of `user_data` to be `usize` instead of `u64`?

src/vmm/src/io_uring/operation/mod.rs
ShadowCurse: Ok

src/vmm/src/io_uring/mod.rs
sudanl0: unwrap on None panics yes but, the intention was not to panic and return 0 as a data here and in one of my experiments I got the value return value as 0 instead of a panic hence the code. Let me double check that and change this to return a different error.

sudanl0: I have updated the code with more comments after discussing with Egor

src/vmm/src/io_uring/bindings.rs
```diff
@@ -89,7 +89,7 @@ pub struct io_uring_sqe {
     pub __bindgen_anon_2: io_uring_sqe__bindgen_ty_2,
     pub len: __u32,
     pub __bindgen_anon_3: io_uring_sqe__bindgen_ty_3,
-    pub user_data: __u64,
+    pub user_data_key: usize,
```
zulinx86, line 92: Is it fine to update bindings that are automatically generated? Doesn't this result in requiring manual change every time we update bindings in line with the newer kernel in the future?

src/vmm/src/io_uring/mod.rs
```diff
@@ -161,44 +165,48 @@ impl IoUring {
...
                         self.num_ops += 1;
                         res
                     })
-                    .map_err(|err_tuple: (SQueueError, T)| -> (IoUringError, T) {
-                        (IoUringError::SQueue(err_tuple.0), err_tuple.1)
+                    .map_err(|(sqe_err, user_data_key)| -> (IoUringError, T) {
+                        (
+                            IoUringError::SQueue(sqe_err),
+                            // We don't use slab.try_remove here for 2 reasons:
+                            // 1. user_data was insertde in slab with step `op.into_sqe` just
```
zulinx86, line 189: nit: typo
```suggestion
                            // 1. user_data was inserted in slab with step `op.into_sqe` just
```

src/vmm/src/io_uring/bindings.rs
sudanl0: oh I missed that this is autogenerated, I'll revert the change. Thanks for noticing it!

src/vmm/src/io_uring/queue/completion.rs
sudanl0: user_data is defined in [src/vmm/src/io_uring/bindings.rs](https://github.com/firecracker-microvm/firecracker/blob/49db07b32df5cc0dde0ff4a9e04a431f17340091/src/vmm/src/io_uring/bindings.rs#L92) which is autogenerated so changing it becomes difficult.

roypat: Ah, bindgen strikes again, sorry for not noticing that earlier

src/vmm/src/io_uring/mod.rs
roypat: Ah, I see, this is not on the return-from-kernel path, thanks for the clarification!

zulinx86 (approved):
LGTM
```
                                 d
                                d8
                               dP8
                              d8 8
                             d8P 8
                            d88  8
                           d88P  8
                          d888   8
                         d888P   8
                        d8888    8
                       d8888P    8
                      d88888     8
        oooood       d88888booooo8
       d8 8"""""Ybooooooooooooooo8oooooooooooooodP"""P""
       88 Y                                  cgmm .P"   
      d88  Y                                    .P"  
      888   YooooooooooooooooooooooooooooooooooP"  
```

src/vmm/src/io_uring/operation/cqe.rs
```diff
@@ -13,22 +13,13 @@ unsafe impl ByteValued for io_uring_cqe {}
...
-    /// to this function.
-    pub(crate) unsafe fn new(inner: io_uring_cqe) -> Self {
-        Self {
-            res: inner.res,
-            user_data: Box::from_raw(inner.user_data as *mut T),
-        }
+    /// Construct a Cqe object from a raw `io_uring_cqe`.
+    pub(crate) fn new(res: i32, user_data: T) -> Self {
+        Self { res, user_data }
     }
```
ShadowCurse, line 23: The comment is outdated. Also would be worth removing `(crate)` as all other methods are just `pub`.

zulinx86 (approved):

ShadowCurse (approved):

Issue #2943, opened 2022-04-01
Fix loose ends of GET /vm/config post-restore


alindima:
This issue is a follow up of https://github.com/firecracker-microvm/firecracker/issues/2783.
The bulk of the issue was solved in this PR: https://github.com/firecracker-microvm/firecracker/pull/2806

There are still a couple of loose ends that could be fixed in order to get the best user experience:

- net devices MAC addresses are reported as all zeros if left unconfigured. View https://github.com/firecracker-microvm/firecracker/pull/2806#discussion_r793271297
- the `machine_config.smt` and `machine_config.cpu_templates` and `boot_source` values are lost (we could persist them in the `VmState`).

dianpopa:
#3121 fixed the network part of this issue.
We still have some loose ends to tie (i.e cpu_template, smt and boot-source)

PR #3121, opened 2022-09-06
Snapshot-ga: Generate a MAC address if unspecified

vibharya:
Signed-off-by: Vibha Acharya <vibharya@amazon.co.uk>

## Changes

Code to generate and assign a MAC address when it is not explicitly specified when a network interface is added.

## Reason

When MAC address is unspecified vm config reports it as null. Post restore it is reported as 00:00:00:00:00:00.
In reality the MAC address is assigned by the driver.

Related to: #2943 

## License Acceptance

## PR Checklist

---

### DIFF
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -4,9 +4,6 @@
 // Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
-
-#[cfg(not(test))]
-use std::io;
 use std::io::{Read, Write};
 use std::net::Ipv4Addr;
 use std::sync::atomic::AtomicUsize;
@@ -21,6 +18,7 @@ use mmds::ns::MmdsNetworkStack;
 use rate_limiter::{BucketUpdate, RateLimiter, TokenType};
 use utils::eventfd::EventFd;
 use utils::net::mac::{MacAddr, MAC_ADDR_LEN};
+use utils::rand_bytes;
 use virtio_gen::virtio_net::{
     virtio_net_hdr_v1, VIRTIO_F_VERSION_1, VIRTIO_NET_F_CSUM, VIRTIO_NET_F_GUEST_CSUM,
     VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_UFO, VIRTIO_NET_F_HOST_TSO4, VIRTIO_NET_F_HOST_UFO,
@@ -48,6 +46,9 @@ enum FrontendError {
     GuestMemory(GuestMemoryError),
     ReadOnlyDescriptor,
 }
+
+// KVM OUI MAC address is 52:54:00:xx:xx:xx
+const KVM_OUI: [u8; 3] = [0x52, 0x54, 0x00];
 
 pub(crate) fn vnet_hdr_len() -> usize {
     mem::size_of::<virtio_net_hdr_v1>()
@@ -78,6 +79,14 @@ fn init_vnet_hdr(buf: &mut [u8]) {
     for i in &mut buf[0..vnet_hdr_len()] {
         *i = 0;
     }
+}
+
+fn generate_mac_address() -> MacAddr {
+    let mut random_mac: [u8; MAC_ADDR_LEN] = [0; MAC_ADDR_LEN];
+    let (left, right) = random_mac.split_at_mut(3);
+    left.copy_from_slice(&KVM_OUI);
+    right.copy_from_slice(rand_bytes(3).as_slice());
+    MacAddr::from_bytes_unchecked(&random_mac)
 }
 
 #[derive(Clone, Copy)]
@@ -120,7 +129,7 @@ pub struct Net {
     pub(crate) irq_trigger: IrqTrigger,
 
     pub(crate) config_space: ConfigSpace,
-    pub(crate) guest_mac: Option<MacAddr>,
+    pub(crate) guest_mac: MacAddr,
 
     pub(crate) device_state: DeviceState,
     pub(crate) activate_evt: EventFd,
@@ -162,12 +171,18 @@ impl Net {
             | 1 << VIRTIO_RING_F_EVENT_IDX;
 
         let mut config_space = ConfigSpace::default();
-        if let Some(mac) = guest_mac {
-            config_space.guest_mac.copy_from_slice(mac.get_bytes());
-            // When this feature isn't available, the driver generates a random MAC address.
-            // Otherwise, it should attempt to read the device MAC address from the config space.
-            avail_features |= 1 << VIRTIO_NET_F_MAC;
-        }
+        // Enable feature to configure a MAC address
+        // If not enabled, the driver generates a random MAC address.
+        avail_features |= 1 << VIRTIO_NET_F_MAC;
+        let mac_addr = {
+            if let Some(some_mac_adrr) = guest_mac {
+                *some_mac_adrr
+            } else {
+                // When the MAC address is not specified explicitly, generate one and assign it.
+                generate_mac_address()
+            }
+        };
+        config_space.guest_mac.copy_from_slice(mac_addr.get_bytes());
 
         let mut queue_evts = Vec::new();
         let mut queues = Vec::new();
@@ -195,7 +210,7 @@ impl Net {
             activate_evt: EventFd::new(libc::EFD_NONBLOCK).map_err(Error::EventFd)?,
             config_space,
             mmds_ns: None,
-            guest_mac: guest_mac.copied(),
+            guest_mac: mac_addr,
 
             #[cfg(test)]
             mocks: Mocks::default(),
@@ -208,8 +223,8 @@ impl Net {
     }
 
     /// Provides the MAC of this net device.
-    pub fn guest_mac(&self) -> Option<&MacAddr> {
-        self.guest_mac.as_ref()
+    pub fn guest_mac(&self) -> &MacAddr {
+        &self.guest_mac
     }
 
     /// Provides the host IFACE name of this net device.
@@ -413,7 +428,7 @@ impl Net {
         rate_limiter: &mut RateLimiter,
         frame_buf: &[u8],
         tap: &mut Tap,
-        guest_mac: Option<MacAddr>,
+        guest_mac: MacAddr,
     ) -> Result<bool> {
         let checked_frame = |frame_buf| {
             frame_bytes_from_buf(frame_buf).map_err(|err| {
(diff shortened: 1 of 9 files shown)


src/devices/src/virtio/net/device.rs
```diff
@@ -161,13 +188,20 @@ impl Net {
...
+        avail_features |= 1 << VIRTIO_NET_F_MAC;
+        if guest_mac.is_some() {
+            mac_addr = guest_mac.copied();
+        } else {
+            // When the MAC address is not specified explicitely, generate one and assign it.
+            mac_addr = generate_mac_address();
         }
+        config_space
+            .guest_mac
+            .copy_from_slice(mac_addr.unwrap().get_bytes());
```
luminitavoicu, line 204: Is this unwrap safe?

src/devices/src/virtio/net/device.rs
```diff
@@ -4,11 +4,12 @@
 // Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
-
+use std::fs::File;
```
pb8o, line 7: The rest of the imports seem to be alphabetical order. Not sure if the format checker will catch it.

src/devices/src/virtio/net/device.rs
```diff
@@ -80,6 +85,28 @@ fn init_vnet_hdr(buf: &mut [u8]) {
...
+fn generate_mac_address() -> Option<MacAddr> {
+    let mut random_mac: [u8; MAC_ADDR_LEN] = [0; MAC_ADDR_LEN];
+    let (left, right) = random_mac.split_at_mut(3);
+    left.copy_from_slice(&KVM_OUI);
+    let file_result = File::open(Path::new(RANDOM_GENERATOR));
+    let mut file = match file_result {
+        Ok(file) => file,
+        Err(e) => {
+            error!("Error opening file {}: {}", RANDOM_GENERATOR, e);
+            return MacAddr::from_bytes(&random_mac);
```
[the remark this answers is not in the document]
vibharya: The function was added in order to specifically not leave the MAC address as None (not configured)
I spent a little time wondering what would be the right approach if the device open or read fails, decided on this as `52 54 00 00 00 00` is still a valid MAC address.

src/devices/src/virtio/net/device.rs
vibharya: Yes, I believe so. As `mac_addr` is assigned with `guest_mac` or from the result of `generate_mac_address()` which will both be valid

src/devices/src/virtio/net/device.rs
```diff
@@ -161,13 +188,20 @@ impl Net {
...
 
+        let mac_addr: Option<MacAddr>;
         let mut config_space = ConfigSpace::default();
-        if let Some(mac) = guest_mac {
-            config_space.guest_mac.copy_from_slice(mac.get_bytes());
-            // When this feature isn't available, the driver generates a random MAC address.
-            // Otherwise, it should attempt to read the device MAC address from the config space.
-            avail_features |= 1 << VIRTIO_NET_F_MAC;
+        // When this feature isn't available, the driver generates a random MAC address.
+        // Otherwise, it should attempt to read the device MAC address from the config space.
```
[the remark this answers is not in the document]
vibharya: I agree, I'll rephrase the comment.

src/devices/src/virtio/net/device.rs
vibharya: The `cargo fmt --all` actually places it right there.

src/devices/src/virtio/net/device.rs
```diff
@@ -48,6 +49,10 @@ enum FrontendError {
     GuestMemory(GuestMemoryError),
     ReadOnlyDescriptor,
 }
+// KVM OUI MAC address is 52:54:00:xx:xx:xx
```
dianpopa, line 52: nit: newline

src/devices/src/virtio/net/device.rs
```diff
@@ -48,6 +49,10 @@ enum FrontendError {
...
+pub const KVM_OUI: [u8; 3] = [0x52, 0x54, 0x00];
```
dianpopa, line 53: Does not need to be public.

src/devices/src/virtio/net/device.rs
```diff
@@ -48,6 +49,10 @@ enum FrontendError {
...
+// Random number generator device file
+pub const RANDOM_GENERATOR: &str = "/dev/urandom";
```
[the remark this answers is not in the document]
dianpopa: also it does not need to be public

src/devices/src/virtio/net/device.rs
```diff
@@ -80,6 +85,28 @@ fn init_vnet_hdr(buf: &mut [u8]) {
...
+        }
+    };
+    match file.read_exact(right) {
+        Ok(_) => right,
+        Err(e) => {
+            error!("Error reading from file {}: {}", RANDOM_GENERATOR, e);
+            return MacAddr::from_bytes(&random_mac);
+        }
+    };
+    MacAddr::from_bytes(&random_mac)
```
dianpopa, line 107: Since we already know that random_mac has the proper length, we can use:

```
MacAddr::from_bytes_unchecked(&random_mac)
```
This would give you back a MacAddr instead of an Option<MacAddr>.

src/devices/src/virtio/net/device.rs
```diff
@@ -80,6 +85,28 @@ fn init_vnet_hdr(buf: &mut [u8]) {
     }
 }
 
+fn generate_mac_address() -> Option<MacAddr> {
+    let mut random_mac: [u8; MAC_ADDR_LEN] = [0; MAC_ADDR_LEN];
+    let (left, right) = random_mac.split_at_mut(3);
+    left.copy_from_slice(&KVM_OUI);
+    let file_result = File::open(Path::new(RANDOM_GENERATOR));
```
dianpopa, line 92: What about using the rand implementations from utils::rand to generate the needed random stuff?

We would just need to do
```
right.copy_from_slice(rand::rand_alphanumerics(3).as_bytes());
```
You would get rid of all the file manipulations.

src/devices/src/virtio/net/device.rs
```diff
@@ -161,13 +188,20 @@ impl Net {
             | 1 << VIRTIO_F_VERSION_1
             | 1 << VIRTIO_RING_F_EVENT_IDX;
 
+        let mac_addr: Option<MacAddr>;
```
dianpopa, line 191: We can do this with less code, something like:
```
let mac_addr = {
            if let Some(some_mac_adrr) = guest_mac {
                some_mac_adrr.clone()
            } else {
                // When the MAC address is not specified explicitly, generate one and assign it.
                generate_mac_address()
            }
        };
```

src/devices/src/virtio/net/device.rs
```diff
@@ -161,13 +188,20 @@ impl Net {
...
-            // When this feature isn't available, the driver generates a random MAC address.
-            // Otherwise, it should attempt to read the device MAC address from the config space.
-            avail_features |= 1 << VIRTIO_NET_F_MAC;
+        // When this feature isn't available, the driver generates a random MAC address.
+        // Otherwise, it should attempt to read the device MAC address from the config space.
+        avail_features |= 1 << VIRTIO_NET_F_MAC;
+        if guest_mac.is_some() {
+            mac_addr = guest_mac.copied();
+        } else {
+            // When the MAC address is not specified explicitely, generate one and assign it.
```
dianpopa, line 199: nit: typo in explicitly

src/devices/src/virtio/net/device.rs
```diff
@@ -195,7 +229,7 @@ impl Net {
             activate_evt: EventFd::new(libc::EFD_NONBLOCK).map_err(Error::EventFd)?,
             config_space,
             mmds_ns: None,
-            guest_mac: guest_mac.copied(),
+            guest_mac: mac_addr,
```
dianpopa, line 232: guest_mac no longer needs to be an option, right?

src/devices/src/virtio/net/device.rs
```diff
@@ -195,7 +229,7 @@ impl Net {
...
 
             #[cfg(test)]
             mocks: Mocks::default(),
```
dianpopa, line 235: I suggest you add a simple unit test where we just check that the result of generate_mac_address  gives back something different than KVM_OUI (which would mean that our rand code is working);

src/vmm/src/device_manager/persist.rs
```diff
@@ -33,6 +33,7 @@ use vm_memory::GuestMemoryMmap;
 use super::mmio::*;
 use crate::resources::VmResources;
 use crate::vmm_config::mmds::MmdsConfigError;
+// use crate::vmm_config::net::MacAddr;
```
dianpopa, line 36: nit: remove comment

src/devices/src/virtio/net/persist.rs
```diff
@@ -83,7 +83,10 @@ impl Persist<'_> for Net {
         let mut net = Net::new_with_tap(
             state.id.clone(),
             state.tap_if_name.clone(),
-            None,
+            Some(MacAddr::from_bytes_unchecked(
```
dianpopa, line 86: With this change, the initial bug should also be fixed?

Is there an integration test that we need to adjust?  Maybe test_get_full_config_after_restoring_snapshot

dianpopa (commented):
Overview:
1. we still need to adjust the changelog
2. add a little unit test in virito/net/devie
3. adjust integ test: test_get_full_config_after_restoring_snapshot

src/devices/src/virtio/net/device.rs
vibharya: I saw this one too while developing, the only issue is that it generates chars only between 0-9/a-z/A-Z which only covers a subset of 2^8 possible values per byte.
On the other hand since we are generating 3 bytes 62x62x62 options should be enough options for network devices.

src/devices/src/virtio/net/device.rs
vibharya: Done

src/devices/src/virtio/net/device.rs
vibharya: It's no longer an `Option` in `Net`

src/devices/src/virtio/net/device.rs
vibharya: Added a test in net.rs, that's where the unit test for adding a network device is.

src/vmm/src/resources.rs
```diff
@@ -1210,11 +1214,13 @@ mod tests {
...
+                            "guest_mac": "06:00:00:00:32:de"
                         }},
                         {{
                             "iface_id": "netif2",
-                            "host_dev_name": "hostname10"
+                            "host_dev_name": "hostname10",
+                            "guest_mac": "06:00:00:00:a2:de"
                         }}
                     ],
                     "machine-config": {{
```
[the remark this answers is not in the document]
vibharya: I've added `test_mac_address_generation` in net.rs to perform this test.

dianpopa (commented):
Changes look good! Good job!
Now we just need to push the doc changes.

src/devices/src/virtio/net/persist.rs
vibharya: Doesn't seem like it needs any change/adjustment.

dianpopa (dismissed):

src/utils/src/lib.rs
```diff
@@ -28,3 +28,32 @@ pub fn get_page_size() -> Result<usize, errno::Error> {
         ps => Ok(ps as usize),
     }
 }
+
+// The below fucntions will get merged in rust-vmm function once we will upstream it there
+fn xor_psuedo_rng_u8_bytes(rand_fn: &dyn Fn() -> u32) -> Vec<u8> {
```
pb8o, line 33: typo: psuedo -> pseudo

pb8o (dismissed):

dianpopa (approved):

src/utils/src/lib.rs
fabianpatras: Also `fucntions` -> `functions`

pb8o (approved):

PR #3124, opened 2022-09-06
[RFC] vmm: handle uniformly GSIs and memory of a guest

bchalios:
## Reason

Right now we have one pair of (IdAllocator, AddressAllocator) per device
manager (mmio and legacy for x86, mmio for aarch64) and we separate the
resources (GSIs and guest memory) these handle.

With ACPI we will add new types of devices that will need as well GSIs
and guest memory, which would mean that with the current way of things
we would need to create yet another pair of allocators.

This fragments the resources (we should divide the GSIs used by ACPI and
MMIO devices for example) and keeps information about them fragmented.

## Changes

This commit introduces a global resource manager for GSIs and guest
memory. Its scope is the whole VM rather than type of devices, so we can
avoid the aforementioned fragmentation.

## License Acceptance

## PR Checklist

---


src/vmm/src/builder.rs
```diff
@@ -244,15 +248,13 @@ fn create_vmm_and_vcpus(
...
     // Instantiate the MMIO device manager.
     // 'mmio_base' address has to be an address which is protected by the kernel
     // and is architectural specific.
-    let mmio_device_manager = MMIODeviceManager::new(
-        arch::MMIO_MEM_START,
-        arch::MMIO_MEM_SIZE,
-        (arch::IRQ_BASE, arch::IRQ_MAX),
-    )
-    .map_err(StartMicrovmError::RegisterMmioDevice)?;
+    let mmio_device_manager =
```
dianpopa, line 256: Trying to figure out why we do not associate a device manager with a resource manager. Based on the type of device manager we are dealing with, we could call into the right AddressAllocator, right (I might be missing some aspects)

bchalios: This is what we do currently, though without a resource manager. Each device manager has an `AddressAllocator` and (maybe) an `IdAllocator`.

This works mostly fine, so far, but it doesn't scale:

When we add the ACPI devices they will need GSIs as well, so if we follow the same approach, create an `IdAllocator` for the "ACPI device manager", we would need to statically split the available GSIs between the two device types, introducing fragmentation. Having a single resource manager (per VM) avoids this. 

Also, not sure if it is important, but I was thinking that having a specific type that handles these resources, would allow us to more easily include that information to the snapshot state, so on restore we can avoid trying to re-allocate things in the exact same order. Btw, atm we do not have a way to allocate the exact same GSI, which is problematic

src/vmm/src/device_manager/mmio.rs
```diff
@@ -102,39 +103,30 @@ pub struct MMIODeviceInfo {
...
             id_to_dev_info: HashMap::new(),
         })
     }
 
     /// Allocates resources for a new device to be added.
-    fn allocate_new_slot(&mut self, irq_count: u32) -> Result<MMIODeviceInfo> {
-        let irqs = (0..irq_count)
-            .map(|_| self.irq_allocator.allocate_id())
-            .collect::<vm_allocator::Result<_>>()
+    pub(crate) fn allocate_new_slot(
```
zulinx86, line 119: IMO, this function doesn't need to be public.
```suggestion
    fn allocate_new_slot(
```
Consumers of `MMIODeviceManager` can use device type-specific public methods (like `regsiter_mmio_virtio()`, `register_mmio_serial()`, etc.).
In addition, issue #2014 claims that the current implementation does IRQ allocation and registration in different place in the code and they should be done in one place.
Keeping this method private guarantees both IRQ allocation and registration are done via device type-specific public methods.

zulinx86 (changes requested):

bchalios:
Closing this, as it will be integrated as part of #3190

PR #3128, opened 2022-09-08
Fixes to allow using ramdisk to run tests

bchalios:
## Changes

This PR makes a few fixes to make it easier to use `devtool` with a ramdisk in order to run integration tests:

1. It prefixes with `sudo` commands used in `devtool` to create the ramdisk. Until now we need to run `devtool` with sudo which messes up with the build directory of the repo (makes it root-owned).
2. It makes TMPDIR point to the ramdisk
3. It uses `mktemp` for the directory of the ramdisk to avoid collisions when more than one test campaigns are running

## Reason

Make it easier to run tests using a ramdisk. This can be potentially faster and also reduce interference (for example disk I/O) when running multiple tests.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tools/devtool
+++ b/tools/devtool
@@ -1251,7 +1251,7 @@ cmd_test() {
 
     if [[ $ramdisk = true ]]; then
         mount_ramdisk ${ramdisk_size}
-        ramdisk_args="--volume ${DEFAULT_RAMDISK_PATH}:${DEFAULT_TEST_SESSION_ROOT_PATH}"
+        ramdisk_args="--env TMPDIR=${DEFAULT_TEST_SESSION_ROOT_PATH}/tmp --volume ${DEFAULT_RAMDISK_PATH}:${DEFAULT_TEST_SESSION_ROOT_PATH}"
     fi
 
     # Testing (running Firecracker via the jailer) needs root access,
--- a/tools/devtool
+++ b/tools/devtool
@@ -1186,14 +1186,17 @@ get_swagger_version() {
 }
 
 mount_ramdisk() {
-    say "Using ramdisk ..."
     local ramdisk_size="$1"
     umount_ramdisk
+    DEFAULT_RAMDISK_PATH=$(sudo mktemp -d /mnt/devtool-ramdisk.XXXXXX)
+    ok_or_die "Could not create ramdisk directory"
     sudo mkdir -p ${DEFAULT_RAMDISK_PATH} && \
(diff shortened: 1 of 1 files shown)


tests/integration_tests/functional/test_snapshot_advanced.py
```diff
@@ -232,7 +235,9 @@ def create_snapshot_helper(
     # Add disks.
     for scratch in test_drives:
         # Add a scratch 64MB RW non-root block device.
-        scratchdisk = drive_tools.FilesystemFile(tempfile.mktemp(), size=64)
+        scratchdisk = drive_tools.FilesystemFile(
+            tempfile.mktemp(dir=scratch_dir), size=64
```
pb8o, line 239: can we use `tmp_path` fixture instead of tempfile? https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html It has some advantages to `tempfile`. 

Alternatively, we could change `TMPDIR` envvar before calling pytest

bchalios: I will look into it, thanks!

tests/integration_tests/functional/test_cmd_line_parameters.py
```diff
@@ -180,7 +180,10 @@ def test_cli_metrics_if_resume_no_metrics(test_microvm_with_api, microvm_factory
 
     # When: restoring from the snapshot
     uvm2 = microvm_factory.build()
-    uvm2.spawn()
+    # Do not create a logger, otherwise this tries to make a hard link from
+    # somewhere under `/tmp` to somewhere under `/srv` and in case we use a
+    # ramdisk for `/srv` it fails.
+    uvm2.spawn(create_logger=False)
```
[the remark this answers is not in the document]
bchalios: Eventually, we might want to switch to using the ramdisk for running CI functional tests, so I 'd rather find a solution. I agree that current fix is just a "patch". I 'll try to look into it next week. If you have any ideas, I 'm all ears.

Thanks Pablo

tests/integration_tests/functional/test_snapshot_advanced.py
bchalios: In fact, you were right! I think the way to go here is putting $TMPDIR pointing to the ramdisk. Then everything is solved and we don't need this hacky solutions for the tests.

dianpopa (approved):

PR #3132, opened 2022-09-14
Add CPUID and MSR snapshot restore test

kalyazin:
## Changes

This adds 2-part tests to verify that CPUID and MSRs retain their values after restoring from a snapshot.
First parts of the tests dump CPUID or MSRs respectively to files and take a snapshot.
Second parts of the tests restore from the snapshot and compare the two CPUIDs or two sets of MSRs values respectively.

## Reason

We need to make sure that when a VM is restored from a snapshot on a different CPU model from the one where the snapshot was taken, CPUID and MSRs hold the same values.

## License Acceptance

## PR Checklist

---


tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -536,13 +536,37 @@ def _check_msr_values_are_equal(before_msrs_fname, after_msrs_fname, guest_kerne
...
+
+    # Arrange the diff as a side by side comparison of values
+    not_equal = (impl_before != impl_after).any(axis=1)
+    impl_before.columns = ["MSR_ADDR", "Before"]
+    impl_after.columns = ["MSR_ADDR", "After"]
+    diff = pd.merge(
+        impl_before[not_equal], impl_after[not_equal], on=["MSR_ADDR", "MSR_ADDR"]
+    ).to_string()
+
+    assert all_equal, f"\n{guest_kernel_name}:\n{diff}"
```
luminitavoicu, line 569: Same here, I would amend these improvements to the commit where you created this function.

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -153,6 +155,34 @@ def test_brand_string(test_microvm_with_api, network_config):
...
+#   * MSR_{C, L}STAR used for SYSCALL/SYSRET; can be different between guests.
+#   * MSR_IA32_SYSENTER_E{SP, IP} used for SYSENTER/SYSEXIT; same as above.
+#   * MSR_KVM_{WALL, SYSTEM}_CLOCK addresses for struct pvclock_* can be different.
+#
+# More detailed information about MSRs can be found in the Intel® 64 and IA-32
+# Architectures Software Developer’s Manual - Volume 4: Model-Specific Registers
+# Check `arch_gen/src/x86/msr_idex.rs` and `msr-index.h` in upstream Linux
+# for symbolic definitions.
+# fmt: off
+MSR_EXCEPTION_LIST = [
```
luminitavoicu, line 171: Nit: Can we move this to the beginning of the file for improved readability?

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +320,386 @@ def _test_cpu_rdmsr(context):
     assert val_diff.empty, f"\n {val_diff}"
 
 
+# These names need to be consistent across the two parts of the snapshot-restore test
+# that spans two instances (one that takes a snapshot and one that restores from it)
+# fmt: off
+SNAPSHOT_RESTORE_SHARED_NAMES = {
```
luminitavoicu, line 326: Nit: same comment as above.

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +344,287 @@ def _test_cpu_rdmsr(context):
...
+    Some MSRs are not inherently supposed to retain their values, so they
+    form an MSR exception list.
+
+    This part of the test is responsible for taking a snapshot and publishing
+    its files along with the `before` MSR dump.
+
+    @type: functional
+    """
+
+    shared_names = SNAPSHOT_RESTORE_SHARED_NAMES
```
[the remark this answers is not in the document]
kalyazin: My motivation was the global name needs to be longer to fully convey the context, however it is not practical to use the long name frequently for readability considerations, that's why I introduced an alias.

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +344,287 @@ def _test_cpu_rdmsr(context):
...
+        snapshot_artifacts_dir, shared_names["msrs_before_fname"]
+    )
+    with open(msrs_before_fname, "w", encoding="UTF-8") as file:
+        file.write(stdout.read())
+
+    # Take a snapshot
+    vm.pause_to_snapshot(
+        mem_file_path=shared_names["mem_fname"],
+        snapshot_path=shared_names["snapshot_fname"],
+        diff=False,
```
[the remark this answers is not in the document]
kalyazin: That's good to know, thanks.

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +344,287 @@ def _test_cpu_rdmsr(context):
...
+    Some MSRs are not inherently supposed to retain their values, so they
+    form an MSR exception list.
+
+    This part of the test is responsible for restoring from a snapshot and
+    comparing two sets of MSR values.
+
+    @type: functional
+    """
+
+    shared_names = SNAPSHOT_RESTORE_SHARED_NAMES
```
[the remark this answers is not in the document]
kalyazin: as above

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +344,287 @@ def _test_cpu_rdmsr(context):
...
+    _, stdout, stderr = ssh_connection.execute_command(
+        shared_names["msr_reader_guest_fname"]
+    )
+    assert stderr.read() == ""
+
+    msrs_after_fname = os.path.join(
+        snapshot_artifacts_dir, shared_names["msrs_after_fname"]
+    )
+    with open(msrs_after_fname, "w", encoding="UTF-8") as file:
+        file.write(stdout.read())
```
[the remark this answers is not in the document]
kalyazin: Good point. Refactored.

resources/tests/msr/wrmsr_list_some.txt
```diff
@@ -1 +1,89 @@
-0xc001102c 1
+0x1b 0xfee00c00
```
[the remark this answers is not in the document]
kalyazin: That was the idea. I didn't want to disturb any ongoing reviews by rebasing often, but now it looks like a good time to do.

tests/integration_tests/functional/test_cpu_features.py
kalyazin: same as above

tests/integration_tests/functional/test_cpu_features.py
kalyazin: I was thinking about this, but decided to leave it just before the first place where it is used. I can imagine that this file expands with new test functions that don't use the list, and the list definition will lack relevance to the neighbouring code.
Do we have a guideline that advises placing data at the top of the file?

tests/integration_tests/functional/test_cpu_features.py
kalyazin: same as above

tests/integration_tests/functional/test_cpu_features.py
luminitavoicu: That makes sense, we can leave this is as. Not a guidance, more like a personal preference.

luminitavoicu (commented):
Looks good!

luminitavoicu (approved):

kalyazin:
I made the following changes compared to the previous version:
- exclude MSR_IA32_TSX_CTRL from the test (limitation is documented in https://github.com/firecracker-microvm/firecracker/pull/3182)
- fix interference between MSR and CPUID tests by using separate artifact directories; they can now be run in the same pipeline
- remove write MSR test with an empty list of MSRs as it is a special case and would complicate the previous item
- rename CPUID tests for consistency

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +346,367 @@ def _test_cpu_rdmsr(context):
...
+
+def check_msr_values_are_equal(before_msrs_fname, after_msrs_fname, guest_kernel_name):
+    """
+    Checks that MSR statuses and values in the files are equal.
+    """
+    before = pd.read_csv(before_msrs_fname)
+    after = pd.read_csv(after_msrs_fname)
+
+    # pylint: disable=singleton-comparison
+    flt_before = before[before["MSR_ADDR"].isin(MSR_EXCEPTION_LIST) == False]
```
pb8o, line 530: I think in pandas it is more idiomatic to write `before[~before[].isin()]`, and should get rid of the pylint exception.

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +346,367 @@ def _test_cpu_rdmsr(context):
...
+    flt = (flt_before["STATUS"] == "implemented") & (
+        flt_after["STATUS"] == "implemented"
+    )
+    impl_before = flt_before.loc[flt]
+    impl_after = flt_after.loc[flt]
+
+    status_diff = diff_msrs(before, after, column_to_drop="VALUE")
+    value_diff = diff_msrs(impl_before, impl_after, column_to_drop="STATUS")
+
+    assert_expr = not status_diff and not value_diff
```
pb8o, line 544: Could be simpler to write it as 2 separate asserts?

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +346,367 @@ def _test_cpu_rdmsr(context):
...
+        shared_names["snapshot_artifacts_root_dir_wrmsr"],
+        context.kernel.base_name(),
+        cpu_template if cpu_template else "none",
+    )
+
+    # Bring snapshot files from the 1st part of the test into the jail
+    chroot_dir = vm.chroot()
+    tmp_snapshot_artifacts_dir = os.path.join(
+        chroot_dir, "tmp", context.kernel.base_name()
+    )
```
pb8o, line 587: Could use `Path` here and below:

```
tmp_snapshot_artifacts_dir = Path(chroot_dir) / tmp / context.kernel.base_name()
...
mem_fname_in_jail = tmp_snapshot_artifacts_dir / shared_names["mem_fname"]
```

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +346,367 @@ def _test_cpu_rdmsr(context):
...
+        os.path.join(chroot_dir, shared_names["snapshot_fname"]),
+        os.path.join(snapshot_artifacts_dir, shared_names["snapshot_fname"]),
+    )
+    shutil.copyfile(
+        root_disk.local_path(),
+        os.path.join(snapshot_artifacts_dir, shared_names["rootfs_fname"]),
+    )
+
+
+@pytest.mark.skipif(
```
pb8o, line 440: are all tests in this file x86 specific? We can skip the whole file at the top rather than one by one https://docs.pytest.org/en/7.1.x/reference/reference.html#globalvar-pytestmark

```
pytestmark = pytest.mark.skipif(...)
```

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -683,6 +685,272 @@ def test_cpu_wrmsr_restore(microvm_factory, cpu_template):
...
+    vm = vm_instance.vm
+    vm.start()
+
+    ssh_connection = net_tools.SSHConnection(vm.ssh_config)
+
+    # Dump CPUID to a file that will be published to S3 for the 2nd part of the test
+    snapshot_artifacts_dir = os.path.join(
+        shared_names["snapshot_artifacts_root_dir_cpuid"],
+        context.kernel.base_name(),
+        cpu_template if cpu_template else "none",
```
pb8o, line 723: Would "no_cpu_template" be more descriptive (though longer)?

mattschlebusch (commented):
I see you have enough reviewers, so just quick high-level comments:

* The commit "tests: cpu features: share msr exception list" can probably be merged into the same commit that defines the `MSR_EXCEPTION_LIST`.
* The `functional/test_cpu_features.py` file is quite large. A test utilities file with some of the ancilliary functions would be easier and more re-usable.
* As far as I can tell this PR focuses on `T2S` only. While there is a case to be made for deprecating and eventually removing `C3` and `T2`, we should probably be covering those templates as well.

tests/integration_tests/functional/test_cpu_features.py
kalyazin: Not quite see what you mean. Could you expand?

pb8o: When I saw `"none"` I immediately thought it could be a misspelling of Python's `None`, but I think it means `no_cpu_template`. No need to change, more of a nit really.

resources/tests/msr/msr_writer.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Helper script used to write MSR values from a text file
+
+set -e
```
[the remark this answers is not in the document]
kalyazin: done

tests/integration_tests/functional/test_cpu_features.py
kalyazin: done

tests/integration_tests/functional/test_cpu_features.py
kalyazin: This would not be ideal, because we'd like to see both diffs even if the first one fails, to get a bigger picture.

tests/integration_tests/functional/test_cpu_features.py
kalyazin: done

tests/integration_tests/functional/test_cpu_features.py
kalyazin: there is one old test that runs on Arm... I guess we could look into restructuring of this file when user-configurable templates are available and new tests are implemented.

tests/integration_tests/functional/test_cpu_features.py
kalyazin: This looks a bit ugly, because the test framework wants None when no template is used, but I need a string to be used as a directory. Not sure how it can be expressed simpler.

kalyazin:
> I see you have enough reviewers, so just quick high-level comments:
> 
> * The commit "tests: cpu features: share msr exception list" can probably be merged into the same commit that defines the `MSR_EXCEPTION_LIST`.
> * The `functional/test_cpu_features.py` file is quite large. A test utilities file with some of the ancilliary functions would be easier and more re-usable.
> * As far as I can tell this PR focuses on `T2S` only. While there is a case to be made for deprecating and eventually removing `C3` and `T2`, we should probably be covering those templates as well.

Thanks, @mattschlebusch .
I squashed the "tests: cpu features: share msr exception list".
I agree, the test file is large, but I refrained from splitting it at this point, as the user-configurable templates are coming and would require some tests. I think it will be more apparent how to refactor it, as more data will be available to extract common code.
I didn't want to test C3 and T2 templates as they are suboptimal with respect to T2S, and we'd have to maintain multiple MSR exception lists, which is in my opinion is not really justified. Customers should be discouraged to use C3 and T2.

mattschlebusch:
> > I see you have enough reviewers, so just quick high-level comments:
> > 
> > * The commit "tests: cpu features: share msr exception list" can probably be merged into the same commit that defines the `MSR_EXCEPTION_LIST`.
> > * The `functional/test_cpu_features.py` file is quite large. A test utilities file with some of the ancilliary functions would be easier and more re-usable.
> > * As far as I can tell this PR focuses on `T2S` only. While there is a case to be made for deprecating and eventually removing `C3` and `T2`, we should probably be covering those templates as well.
> 
> Thanks, @mattschlebusch . I squashed the "tests: cpu features: share msr exception list". I agree, the test file is large, but I refrained from splitting it at this point, as the user-configurable templates are coming and would require some tests. I think it will be more apparent how to refactor it, as more data will be available to extract common code. I didn't want to test C3 and T2 templates as they are suboptimal with respect to T2S, and we'd have to maintain multiple MSR exception lists, which is in my opinion is not really justified. Customers should be discouraged to use C3 and T2.

C3 and T2 haven't been deprecated yet. If they were to be we would need to plan and communicate that. Until then we should be matching the testing for all hardcoded templates. With that said, the testing of T2S added by the PR stands on its own and is not blocked by this point at all.

luminitavoicu (dismissed):
Looks good to me. I only have a couple of nits that I wouldn't block the PR on.

resources/tests/msr/wrmsr_list.txt
```diff
@@ -0,0 +1,88 @@
+0x1b 0xfee00c00
```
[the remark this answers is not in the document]
kalyazin: That is a semi-structured way. I was thinking about documenting the approach I took, but it was rather dirty and incomplete. I will probably create a task to document/enhance it.

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -314,6 +346,367 @@ def _test_cpu_rdmsr(context):
     assert val_diff.empty, f"\n {val_diff}"
 
 
+# These names need to be consistent across the two parts of the snapshot-restore test
+# that spans two instances (one that takes a snapshot and one that restores from it)
```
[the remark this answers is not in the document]
kalyazin: I did mean to say spans actually :)

resources/tests/msr/wrmsr_list.txt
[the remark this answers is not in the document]
luminitavoicu: That would be a great idea, it would be very useful to have the approach documented.

pb8o (approved):

PR #3139, opened 2022-09-22
Move `devtool release` functionality to a separate script

pb8o:
## Changes

Move the release functions of devtool into its own script. That way it can be built from within the Docker container too.

## Reason

- Make it simple to change the release workflow
- Simplify `devtool`
- Allow builds from within the devctr (just call `./tools/release.sh`)

## License Acceptance

## PR Checklist

- [X] New `unsafe` code is documented.
- [X] User-facing changes are mentioned in `CHANGELOG.md`.

---

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -16,5 +16,8 @@ Andrei Sandu <sandreim@amazon.com> <54316454+sandreim@users.noreply.github.com>
 Diana Popa <dpopa@amazon.com> <dpopa@38f9d3563971.ant.amazon.com>
 Alexandru Cihodaru <cihodar@amazon.com>
 Liviu Berciu <lberciu@amazon.com>
-Jonathan Woollett-Light <jonathanwoollettlight@gmail.com> <jonthanwoollettlight@gmail.com>
+Jonathan Woollett-Light <jcawl@amazon.co.uk> <jonathanwoollettlight@gmail.com>
+Jonathan Woollett-Light <jcawl@amazon.co.uk> <jonthanwoollettlight@gmail.com>
 karthik nedunchezhiyan <karthik.n@zohocorp.com>
+Babis Chalios <bchalios@amazon.es> <babis.chalios@gmail.com>
+Pablo Barbáchano <pablob@amazon.com>
--- a/tests/integration_tests/style/test_licenses.py
+++ b/tests/integration_tests/style/test_licenses.py
@@ -3,6 +3,7 @@
 """Tests checking against the existence of licenses in each file."""
 
 import datetime
+
 from framework import utils
 
 AMAZON_COPYRIGHT_YEARS = range(2018, datetime.datetime.now().year + 1)
@@ -105,3 +106,7 @@ def test_for_valid_licenses():
         if _validate_license(file) is False:
             error_msg.append("{}".format(str(file)))
     assert not error_msg, "Files {} have invalid licenses".format((error_msg))
+
+
+if __name__ == "__main__":
+    test_for_valid_licenses()
--- a/tools/devtool
+++ b/tools/devtool
@@ -79,11 +79,9 @@ DEVCTR_IMAGE_TAG="v44"
 # (Yet another step on our way to reproducible builds.)
 DEVCTR_IMAGE="${DEVCTR_IMAGE_NO_TAG}:${DEVCTR_IMAGE_TAG}"
 
-# Naming things is hard
-MY_NAME="Firecracker $(basename "$0")"
-
 # Full path to the Firecracker tools dir on the host.
 FC_TOOLS_DIR=$(cd "$(dirname "$0")" && pwd)
+source "$FC_TOOLS_DIR/functions"
 
 # Full path to the Firecracker sources dir on the host.
 FC_ROOT_DIR=$(cd "${FC_TOOLS_DIR}/.." && pwd)
@@ -159,57 +157,6 @@ DEFAULT_TEST_SESSION_ROOT_PATH=/srv
 DEFAULT_RAMDISK_PATH=/mnt/devtool-ramdisk
 
 
-# Send a decorated message to stdout, followed by a new line
-#
-say() {
-    [ -t 1 ] && [ -n "$TERM" ] \
-        && echo "$(tput setaf 2)[$MY_NAME]$(tput sgr0) $*" \
-        || echo "[$MY_NAME] $*"
-}
-
-# Send a decorated message to stdout, without a trailing new line
-#
-say_noln() {
-    [ -t 1 ] && [ -n "$TERM" ] \
-        && echo -n "$(tput setaf 2)[$MY_NAME]$(tput sgr0) $*" \
-        || echo "[$MY_NAME] $*"
-}
-
-# Send a text message to stderr
-#
-say_err() {
-    [ -t 2 ] && [ -n "$TERM" ] \
-        && echo -e "$(tput setaf 1)[$MY_NAME] $*$(tput sgr0)" 1>&2 \
-        || echo -e "[$MY_NAME] $*" 1>&2
-}
-
-# Send a warning-highlighted text to stdout
-say_warn() {
-    [ -t 1 ] && [ -n "$TERM" ] \
-        && echo "$(tput setaf 3)[$MY_NAME] $*$(tput sgr0)" \
-        || echo "[$MY_NAME] $*"
-}
-
-# Exit with an error message and (optional) code
-# Usage: die [-c <error code>] <error message>
-#
-die() {
-    code=1
-    [[ "$1" = "-c" ]] && {
-        code="$2"
-        shift 2
-    }
-    say_err "$@"
-    exit $code
-}
-
-# Exit with an error message if the last exit code is not 0
-#
-ok_or_die() {
-    code=$?
-    [[ $code -eq 0 ]] || die -c $code "$@"
-}
-
 # Check if Docker is available and exit if it's not.
 # Upon returning from this call, the caller can be certain Docker is available.
 #
@@ -276,14 +223,6 @@ ensure_devctr() {
 
         ok_or_die "Error pulling docker image. Aborting."
     }
-}
-
-# Check if /dev/kvm exists. Exit if it doesn't.
-# Upon returning from this call, the caller can be certain /dev/kvm is
-# available.
-#
-ensure_kvm() {
-    [[ -c /dev/kvm ]] || die "/dev/kvm not found. Aborting."
 }
 
 # Make sure the build/ dirs are available. Exit if we can't create them.
@@ -378,50 +317,6 @@ cmd_build_devctr() {
     # Copy back the lockfile, since a new dependency or version would have
     # updated it.
     copy_poetry_lockfile
-}
-
-# Prompt the user for confirmation before proceeding.
-# Args:
-#   $1  prompt text.
-#       Default: Continue? (y/n)
-#   $2  confirmation input.
-#       Default: y
-# Return:
-#   exit code 0 for successful confirmation
-#   exit code != 0 if the user declined
-#
-get_user_confirmation() {
-
-    # Pass if running unattended
-    [[ "$OPT_UNATTENDED" = true ]] && return 0
-
-    # Fail if STDIN is not a terminal (there's no user to confirm anything)
-    [[ -t 0 ]] || return 1
-
-    # Otherwise, ask the user
-    #
-    msg=$([ -n "$1" ] && echo -n "$1" || echo -n "Continue? (y/n) ")
-    yes=$([ -n "$2" ] && echo -n "$2" || echo -n "y")
-    say_noln "$msg"
-    read c && [ "$c" = "$yes" ] && return 0
-    return 1
-}
-
-# Validate the user supplied version number.
-# It must start with 3 groups of integers separated by dot and
-# must not contain `wip` or `dirty`.
-validate_version() {
-    declare version_regex="^([0-9]+\.){2}[0-9]+"
-    version="$1"
-
-    if [ -z "$version" ]; then
-        die "Version cannot be empty."
-    elif [[ ! "$version" =~ $version_regex ]]; then
-        die "Invalid version number: $version. Version should start with \$Major.\$Minor.\$Build, see
-        https://github.com/firecracker-microvm/firecracker/blob/main/docs/RELEASE_POLICY.md for more information."
-    elif [[ "$version" == *"wip"* ]] || [[ "$version" == *"dirty"* ]]; then
-        die "Invalid version number: $version. Version should not contain \`wip\` or \`dirty\`."
-    fi
 }
 
 # Validate that the repo targetted for a release exists.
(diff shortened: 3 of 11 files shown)


.mailmap
```diff
@@ -16,5 +16,8 @@ Andrei Sandu <sandreim@amazon.com> <54316454+sandreim@users.noreply.github.com>
 Diana Popa <dpopa@amazon.com> <dpopa@38f9d3563971.ant.amazon.com>
 Alexandru Cihodaru <cihodar@amazon.com>
 Liviu Berciu <lberciu@amazon.com>
-Jonathan Woollett-Light <jonathanwoollettlight@gmail.com> <jonthanwoollettlight@gmail.com>
+Jonathan Woollett-Light <jcawl@amazon.co.uk> <jonathanwoollettlight@gmail.com>
+Jonathan Woollett-Light <jcawl@amazon.co.uk> <jonthanwoollettlight@gmail.com>
```
JonathanWoollett-Light, line 20: Duplicated line?

JonathanWoollett-Light:
Nit: Could `tools/prepare-release.sh` and `release.sh` be python scripts instead? (I tend to think any non-trivial script should be python over bash)

.mailmap
pb8o: Not quite, the 2nd one fixes a typo. I think it cannot be done all in one line. See https://git-scm.com/docs/gitmailmap

pb8o:
> Nit: Could `tools/prepare-release.sh` and `release.sh` be python scripts instead?
> 
> I would tend to argue that any non-trivial script should be python over bash.

Yep, I agree overly complicated/long shell scripts should be rewritten. Why I kept them as bash scripts in this case:
1. It's mostly moving working functionality out of `devtool`
2. The scripts _mostly_ call other programs like `git`, `cargo`, etc. That's where shell scripts shine, and doing it in Python would be more verbose.
3. After the cleanup this would be a net reduction of shell script code

One thing we could do would be to drop the `.sh` suffix, to make it easy to rewrite them in the future.

JonathanWoollett-Light (approved):

tests/integration_tests/style/test_licenses.py
```diff
@@ -105,3 +106,7 @@ def test_for_valid_licenses():
         if _validate_license(file) is False:
             error_msg.append("{}".format(str(file)))
     assert not error_msg, "Files {} have invalid licenses".format((error_msg))
+
+
+if __name__ == "__main__":
```
dianpopa, line 111: Is this changed used within this PR?

tools/devtool
```diff
@@ -972,6 +972,16 @@ cmd_build() {
     return $ret
 }
 
+function cmd_make_release {
```
dianpopa, line 975: As far as I understand it, one would firstly call prepare_release from outside devtool and then have the choice to do the actual release by calling the ./tools/release.sh from either devtool or from outside?

tools/prepare-release.sh
```diff
@@ -0,0 +1,232 @@
+#!/usr/bin/env bash
+# -*- shell-script[bash] -*-
```
dianpopa, line 2: What s the purpose of this?

tools/prepare-release.sh
```diff
@@ -0,0 +1,232 @@
...
+say "Creating local tag..."
+create_local_tag "$version" "$prev_ver" "$LOCAL_BRANCH" || die "Could not create local tag v$version."
+
+
+# https://en.wikipedia.org/wiki/ANSI_escape_code
+function SGR {
+    local codes=$*
+    printf "\e[%sm" "${codes// /;}"
+}
+# function SGR {
```
dianpopa, line 184: nit: remove?

tools/prepare-release.sh
```diff
@@ -0,0 +1,232 @@
...
+# Update changelog.
+say "Updating changelog..."
+sed -i "s/\[Unreleased\]/\[$version\]/g" "$FC_ROOT_DIR/CHANGELOG.md"
+
+git add "${files_to_change[@]}" Cargo.lock CREDITS.md CHANGELOG.md
+git commit -s -m "Releasing v$version"
+
+# Start by creating a local tag and associate to it a description.
+say "Creating local tag..."
+create_local_tag "$version" "$prev_ver" "$LOCAL_BRANCH" || die "Could not create local tag v$version."
```
dianpopa, line 176: I am not sure we should do this step here. 
We want to do this after the PR with the changelog update has been merged (i.e as part of the actual `make_release` step) since the changelog may suffer changes after PR review and the tag description would not have them, right?

tools/prepare-release.sh
```diff
@@ -0,0 +1,232 @@
...
+done
+
+# Run `cargo check` to update firecracker and jailer versions in
+# `Cargo.lock`.
+say "Updating lockfile..."
+cargo check
+
+# Update credits.
+say "Updating credits..."
+$FC_TOOLS_DIR/update-credits.sh
```
dianpopa, line 165: nit: double quote this

tools/prepare-release.sh
```diff
@@ -0,0 +1,232 @@
...
+    prev_ver="$2"
+    branch="$3"
+
+    # Create tag.
+    say "Obtaining tag description for local tag v$version..."
+    tag_text=$(compose_release_text "$version" "$prev_ver") || die "Could not compose tag description."
+    say "Tag description for v$version:"
+    echo "$tag_text"
+    git tag -a v"$version" "$branch" -m "$tag_text"
+    say "Local tag v$version created."
```
dianpopa, line 82: nit:  `v"$version"`

tools/prepare-release.sh
```diff
@@ -0,0 +1,232 @@
...
+
+function warn {
+    # reset reverse yellow
+    echo "$(SGR 0 7 33)WARNING: $*$(SGR 0)"
+}
+
+cat <<EOF
+🎉 Almost done!
+
+$(li 1. Check the changes made to the repo:)
```
dianpopa, line 208: nit: missed some quotes in v$version

tools/release.sh
```diff
@@ -0,0 +1,166 @@
...
+    *)
+      echo "Unknown option $1"
+      exit 1
+      ;;
+  esac
+done
+
+
+ARCH=$(uname -m)
+VERSION=$(git describe --tags --dirty)
```
dianpopa, line 101: for the moment this script only builds the artifacts, right?

The old functionality of cmd_release would also post the actual release to github.
Is there another component that will do that?

tools/prepare-release.sh
```diff
@@ -0,0 +1,232 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+set -eu -o pipefail
+# set -x
+
+MY_NAME="Firecracker $(basename "$0")"
+FC_TOOLS_DIR=$(dirname $(realpath $0))
+FC_ROOT_DIR=$FC_TOOLS_DIR/..
+
+function say {
```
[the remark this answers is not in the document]
pb8o: Do you mean `say` and `die`? I can move those, but they are just a few lines, so it won't save much. For the rest of the functions I have copied them over but also slightly modified them with simplifications and fixes. The plan is I will remove them from `devtool` after the new release is done. I am doing it in 2 steps so that we have a back-up plan in case the new release process does not work.

tools/release.sh
```diff
@@ -48,6 +48,14 @@ function check_bin_artifact {
     fi
 }
 
+function strip-and-split-debuginfo {
+    local bin=$1
+    echo "STRIP $bin"
```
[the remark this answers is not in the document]
pb8o: done

tools/prepare-release.sh
```diff
@@ -0,0 +1,232 @@
...
+}
+
+function get_swagger_version {
+    grep -oP 'version: \K.*' "$1"
+}
+
+# Validate the user supplied version number.
+# It must start with 3 groups of integers separated by dot and
+# must not contain `wip` or `dirty`.
+function validate_version {
```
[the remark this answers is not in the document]
pb8o: yep, that's the plan

tests/integration_tests/style/test_licenses.py
pb8o: Actually no, I used it to check locally that the licenses were fine.

tools/devtool
pb8o: The process would be to just call `prepare-release.sh`, do the PR, tag, push etc and then the release automation will take it from there. I.e. there's no need to call `release.sh` directly.

tools/prepare-release.sh
pb8o: done

tools/prepare-release.sh
pb8o: hmm true. We could:

1. Separate that step as `release-tag.sh`?
2. Provide instructions to do the tag by hand

I would probably go with 1). 2) seems too error prone.

tools/prepare-release.sh
```diff
@@ -0,0 +1,232 @@
...
+
+# Update credits.
+say "Updating credits..."
+$FC_TOOLS_DIR/update-credits.sh
+
+# Update changelog.
+say "Updating changelog..."
+sed -i "s/\[Unreleased\]/\[$version\]/g" "$FC_ROOT_DIR/CHANGELOG.md"
+
+git add "${files_to_change[@]}" Cargo.lock CREDITS.md CHANGELOG.md
```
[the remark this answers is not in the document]
pb8o: It will just print a URL to make the PR, but won't build it directly... that would be cool.

I will add it to the help.

tools/prepare-release.sh
pb8o: that's a "Hey Emacs" line. I think it's not needed as we have a she-bang. I can remove those.

tools/prepare-release.sh
pb8o: fixing

tools/release.sh
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env bash
+# -*- shell-script[bash] -*-
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# fail if we encounter an error, uninitialized variable or a pipe breaks
+set -eu -o pipefail
```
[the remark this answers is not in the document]
pb8o: I like the idea of keeping related things together, but I'm not a big fan of subdirectories with just a few files :). I.e.

```
tools/release/release.sh
tools/release/prepare-release.sh
tools/release/update-credits.sh
```

Another alternative could be to prefix the scripts so it's clear they are related:

```
tools/release-prepare.sh
tools/release.sh
tools/release-update-credits.sh
```

tools/prepare-release.sh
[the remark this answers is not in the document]
dianpopa: Yes, I like it a lot this way.

tools/release.sh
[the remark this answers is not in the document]
dianpopa: 💯

tests/integration_tests/style/test_licenses.py
dianpopa: Seems that it is still there

tools/prepare-release.sh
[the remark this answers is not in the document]
dianpopa: Oh sorry, I said inaccurate things. Old functionality did not post a PR, I got confused.

tools/prepare-release.sh
dianpopa: Yes, let s go with 1.

tools/release-prepare.sh
```diff
@@ -0,0 +1,124 @@
...
+    3. Generate CREDITS.md and CHANGELOG.md
+    4. Commit the result
+    5. Create a link to PR the changes
+EOF
+    exit 1
+fi
+version=$1
+validate_version "$version"
+
+check_local_branch_is_release_branch
```
dianpopa, line 30: I am thinking about this...when we prepare the release, do we really need to be on a release branch? We would need to be on a branch that is a fork of the actual upstream release branch, right?

tools/release-tag.sh
```diff
@@ -0,0 +1,69 @@
...
+RELEASE_BRANCH=firecracker-v$(echo "$version" |cut -d. -f-2)
+UPSTREAM=upstream
+UPSTREAM_URL=$(git remote get-url $UPSTREAM)
+check_local_branch_is_release_branch
+
+# Start by creating a local tag and associate to it a description.
+say "Creating local tag..."
+create_local_tag "$version" "$prev_ver" "$LOCAL_BRANCH"
+
+get_user_confirmation
```
dianpopa, line 57: What confirmation are we expecting here? There is not question asked

tools/release.sh
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# fail if we encounter an error, uninitialized variable or a pipe breaks
+set -eu -o pipefail
+
+function die {
```
dianpopa, line 9: Shouldn t we be using the "die" from "functions" file?

tools/release.sh
```diff
@@ -0,0 +1,159 @@
...
+ARTIFACTS=(firecracker jailer seccompiler-bin rebase-snap)
+
+if [ "$LIBC" == "gnu" ]; then
+    # Don't build jailer. See commit 3bf285c8f
+    echo "Not building jailer because glibc selected instead of musl"
+    CARGO_OPTS+=" --exclude jailer"
+    ARTIFACTS=(firecracker seccompiler-bin rebase-snap)
+fi
+
+echo "Building version=$VERSION, profile=$PROFILE, target=$CARGO_TARGET..."
```
dianpopa, line 132: say?

tools/release.sh
```diff
@@ -0,0 +1,159 @@
...
+# shellcheck disable=SC2086
+cargo build --target "$CARGO_TARGET" $CARGO_OPTS --workspace
+
+echo "Binaries placed under $CARGO_TARGET_DIR"
+
+# # # # Make a release
+if [ -z "$MAKE_RELEASE" ]; then
```
dianpopa, line 139: How is this used? If MAKE_RELEASE is not provided, we would jstop at just building the binaries?

dianpopa (commented):
Additional review comments

tools/release.sh
pb8o: Yes. The idea is the script can be used to do a normal build or a full release build.

tools/release-tag.sh
pb8o: fixed

tools/release.sh
pb8o: fixed

tools/release.sh
pb8o: fixed

tools/release-prepare.sh
pb8o: Yeah, it doesn't matter except for patch releases, since those are done in a specific branch.

dianpopa (dismissed):

JonathanWoollett-Light (approved):

dianpopa (approved):

Issue #1082, opened 2019-05-03
Jailer exiting with odd start-time-us flag


xibz:
When executing the jailer like so
```bash
sudo jailer --id my-test-id --uid 123 --gid 100 --exec-file /usr/local/bin/jailer --node 0 --chroot-base-dir ./jails --seccomp-level 0
``` 
will return 

```bash
error: Found argument '--start-time-us' which wasn't expected, or isn't valid in this context

USAGE:
    jailer --chroot-base-dir <chroot_base> --exec-file <exec_file> --gid <gid> --id <id> --node <numa_node> --seccomp-level <seccomp-level> --uid <uid>

For more information try --help
```

I was expecting this to run properly, however it seems to crash with an error that doesn't relate to any of the flags I passed in.

raduweiss:
@xibz I think the value of `--exec-file` should be 

> the path to the Firecracker binary that will be exec-ed by the jailer. The user can provide a path to any binary, but the interaction with the jailer is mostly Firecracker specific.

(taken from the [jailer docs](https://github.com/firecracker-microvm/firecracker/blob/master/docs/jailer.md#jailer-usage)).

I'm wondering if we should return a meaningful error, like "it appears that `exec_file` is not a Firecracker binary. Are you sure?"

xibz:
@raduweiss - Ah that was it! Thanks, and yes, I think a useful error message would be nice here especially if it has nothing to do with `start-time-us` flag.

dianpopa:
hi, @xibz I am not able to replicate your problem with the latest firecracker release. Can you confirm that it still happens with the [release 0.17.0](https://github.com/firecracker-microvm/firecracker/releases/tag/v0.17.0)?

xibz:
@dianpopa - Just built from source v0.17.0 and still get the weird error of 
`error: Found argument '--start-time-us' which wasn't expected, or isn't valid in this context`

andreeaflorescu:
@xibz the only idea I have here is to check that the ```--exec-file``` parameter contains the "firecracker"  string. As of now we don't have a good way of checking that the binary send via ```--exec-file``` is indeed firecracker.

lauralt:
With the new implementation for arguments parsing, we still receive this error for that input, but I agree we should think of some check for the '--exec-file'.

zulinx86:
I've confirmed that this issue still happens with the latest version (v1.1.1).
```
[...]

PR #3143, opened 2022-09-24
[jailer] add check of --exec-file parameter

zulinx86:
## Changes

- check that the given `--exec-file` parameter contains `firecracker`.
- create `/tmp/pseudo_firecracker_exec_file` before tests with arguments.

## Reason

Issue: #1082
As the doc states, the `--exec-file` parameter is intended to be specified a firecracker binary. 
> `exec_file` is the path to the Firecracker binary that will be exec-ed by the jailer. The user can provide a path to any binary, but the interaction with the jailer is mostly Firecracker specific.

But jailer didn't check if the given executable file as `--exec-file` is firecracker.
If jailer run it without this check, it might output nothing or uninterpretable messages.
This situation made it difficult for users to identify the root cause.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@
   network interfaces. This address can be obtained as part of the GET
   `/vm/config`.
 
+### Changed
+
+- Changed the jailer option `--exec-file` to fail if the filename does not
+  contain the string `firecracker` to prevent from running non-firecracker
+  binaries.
+
 ### Fixed
 
 - Make the `T2` template more robust by explicitly disabling additional
--- a/docs/jailer.md
+++ b/docs/jailer.md
@@ -32,8 +32,8 @@ jailer --id <id> \
 - `id` is the unique VM identification string, which may contain alphanumeric
   characters and hyphens. The maximum `id` length is currently 64 characters.
 - `exec_file` is the path to the Firecracker binary that will be exec-ed by the
-  jailer. The user can provide a path to any binary, but the interaction with
-  the jailer is mostly Firecracker specific.
+  jailer. The filename must include the string `firecracker`. This is enforced
+  because the interaction with the jailer is Firecracker specific.
 - `uid` and `gid` are the uid and gid the jailer switches to as it execs the
   target binary.
 - `parent-cgroup` is used to allow the placement of microvm cgroups in custom
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -120,16 +120,7 @@ impl Env {
         let exec_file = arguments
             .single_value("exec-file")
             .ok_or_else(|| Error::ArgumentParsing(MissingValue("exec-file".to_string())))?;
-        let exec_file_path = canonicalize(&exec_file)
-            .map_err(|err| Error::Canonicalize(PathBuf::from(&exec_file), err))?;
-
-        if !exec_file_path.is_file() {
-            return Err(Error::NotAFile(exec_file_path));
-        }
-
-        let exec_file_name = exec_file_path
-            .file_name()
-            .ok_or_else(|| Error::FileName(exec_file_path.clone()))?;
+        let (exec_file_path, exec_file_name) = Env::validate_exec_file(exec_file)?;
 
         let chroot_base = arguments
             .single_value("chroot-base-dir")
@@ -169,7 +160,7 @@ impl Env {
         let mut cgroups: Vec<Box<dyn Cgroup>> = Vec::new();
         let parent_cgroup = match arguments.single_value("parent-cgroup") {
             Some(parent_cg) => Path::new(parent_cg),
-            None => Path::new(exec_file_name),
+            None => Path::new(&exec_file_name),
         };
         if parent_cgroup
             .components()
@@ -247,6 +238,29 @@ impl Env {
         self.uid
     }
 
+    fn validate_exec_file(exec_file: &str) -> Result<(PathBuf, String)> {
+        let exec_file_path = canonicalize(exec_file)
+            .map_err(|err| Error::Canonicalize(PathBuf::from(exec_file), err))?;
+
+        if !exec_file_path.is_file() {
+            return Err(Error::NotAFile(exec_file_path));
+        }
+
+        let exec_file_name = exec_file_path
+            .file_name()
+            .ok_or_else(|| Error::ExtractFileName(exec_file_path.clone()))?
+            .to_str()
+            // Safe to unwrap as the original `exec_file` is `String`.
+            .unwrap()
+            .to_string();
+
+        if !exec_file_name.contains("firecracker") {
+            return Err(Error::ExecFileName(exec_file_name));
+        }
+
+        Ok((exec_file_path, exec_file_name))
+    }
+
     fn parse_resource_limits(resource_limits: &mut ResourceLimits, args: &[String]) -> Result<()> {
         for arg in args {
             let (name, value) = arg
@@ -293,7 +307,7 @@ impl Env {
     fn save_exec_file_pid(&mut self, pid: i32, chroot_exec_file: PathBuf) -> Result<()> {
         let chroot_exec_file_str = chroot_exec_file
             .to_str()
-            .ok_or_else(|| Error::FileName(chroot_exec_file.clone()))?;
+            .ok_or_else(|| Error::ExtractFileName(chroot_exec_file.clone()))?;
         let pid_file_path =
             PathBuf::from(format!("{}{}", chroot_exec_file_str, PID_FILE_EXTENSION));
         let mut pid_file = OpenOptions::new()
@@ -363,7 +377,7 @@ impl Env {
         let exec_file_name = self
             .exec_file_path
             .file_name()
-            .ok_or_else(|| Error::FileName(self.exec_file_path.clone()))?;
+            .ok_or_else(|| Error::ExtractFileName(self.exec_file_path.clone()))?;
         // We do a quick push here to get the global path of the executable inside the chroot,
         // without having to create a new PathBuf. We'll then do a pop to revert to the actual
         // chroot_dir right after the copy.
@@ -620,6 +634,8 @@ mod tests {
     use crate::build_arg_parser;
     use crate::cgroup::test_util::MockCgroupFs;
 
+    const PSEUDO_EXEC_FILE_PATH: &str = "/tmp/pseudo_firecracker_exec_file";
+
     #[derive(Clone)]
     struct ArgVals<'a> {
         pub id: &'a str,
@@ -637,9 +653,10 @@ mod tests {
 
     impl ArgVals<'_> {
         pub fn new() -> ArgVals<'static> {
+            File::create(PSEUDO_EXEC_FILE_PATH).unwrap();
             ArgVals {
                 id: "bd65600d-8669-4903-8a14-af88203add38",
-                exec_file: "/proc/cpuinfo",
+                exec_file: PSEUDO_EXEC_FILE_PATH,
                 uid: "1001",
                 gid: "1002",
                 chroot_base: "/",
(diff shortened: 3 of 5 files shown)


src/jailer/src/env.rs
```diff
@@ -130,6 +130,14 @@ impl Env {
...
             .file_name()
             .ok_or_else(|| Error::FileName(exec_file_path.clone()))?;
+        let exec_file_name_as_string = exec_file_name
+            .to_str()
+            // Safe to unwrap as the original `exec_file` is `String`.
+            .unwrap()
+            .to_string();
+        if !exec_file_name_as_string.contains("firecracker") {
+            return Err(Error::ExecFileName(exec_file_name_as_string));
+        }
```
luminitavoicu, line 140: Nit: I would put all checks around this (from line 123 to 140) to a separate function, to improve readability.

src/jailer/src/env.rs
```diff
@@ -645,9 +647,10 @@ mod tests {
 
     impl ArgVals<'_> {
         pub fn new() -> ArgVals<'static> {
+            File::create(PSEUDO_EXEC_FILE_PATH).unwrap();
             ArgVals {
                 id: "bd65600d-8669-4903-8a14-af88203add38",
-                exec_file: "/proc/cpuinfo",
+                exec_file: PSEUDO_EXEC_FILE_PATH,
```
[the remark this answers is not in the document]
luminitavoicu: We should also add a negative integration test inside `test_jail.py` that sets `test_microvm.jailer.exec_file` to a non-firecracker file and verifies that the error is the expected one.

src/jailer/src/main.rs
```diff
@@ -140,6 +141,12 @@ impl fmt::Display for Error {
             CStringParsing(_) => write!(f, "Encountered interior \\0 while parsing a string"),
             Dup2(ref err) => write!(f, "Failed to duplicate fd: {}", err),
             Exec(ref err) => write!(f, "Failed to exec into Firecracker: {}", err),
+            ExecFileName(ref filename) => write!(
+                f,
+                "Invalid filename. The filename of `--exec-file` option must contain \
+                 \"firecracker\": {}",
+                filename
+            ),
             FileName(ref path) => write!(
```
luminitavoicu, line 150: We should also add an test for the new error format inside [test_error_display](https://github.com/firecracker-microvm/firecracker/blob/main/src/jailer/src/main.rs#L486) unit test.

luminitavoicu (commented):
Looks good! I left a few comments. Please also squash the two commits into one. Per our [contributing guideline](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md?plain=1#L71), each commit needs to pass unit tests. Also please add some documentation around this change to the jailer docs and a Changelog.md entry.

src/jailer/src/env.rs
JonathanWoollett-Light: If this is only done here it might be easiest to use a scope.
```rust
// Validates file
{
    // your checks here
    // ...
}
```

luminitavoicu: Even better. Let's use scope!

zulinx86: OK. I'll revise it using scope.

luminitavoicu: Sorry, @zulinx86, your comment wasn't yet visible when I commented :sweat_smile:

zulinx86: No worries, @luminitavoicu !
Currently, I'm thinking of coding as follows:
```
@@ -130,14 +130,7 @@ impl Env {
         let exec_file_name = exec_file_path
             .file_name()
             .ok_or_else(|| Error::FileName(exec_file_path.clone()))?;
-        let exec_file_name_as_string = exec_file_name
-            .to_str()
-            // Safe to unwrap as the original `exec_file` is `String`.
-            .unwrap()
-            .to_string();
-        if !exec_file_name_as_string.contains("firecracker") {
-            return Err(Error::ExecFileName(exec_file_name_as_string));
-        }
+        Env::validate_exec_file(exec_file_name)?;
 
         let chroot_base = arguments
             .single_value("chroot-base-dir")
@@ -255,6 +248,18 @@ impl Env {
         self.uid
     }
 
+    fn validate_exec_file(exec_file_name: &OsStr) -> Result<()> {
+        let exec_file_name_as_string = exec_file_name
+            .to_str()
+            // Safe to unwrap as the original `exec_file` is `String`.
+            .unwrap()
+            .to_string();
+        match exec_file_name_as_string.contains("firecracker") {
+            true => Ok(()),
+            false => Err(Error::ExecFileName(exec_file_name_as_string)),
+        }
+    }
+
     fn parse_resource_limits(resource_limits: &mut ResourceLimits, args: &[String]) -> Result<()> {
         for arg in args {
             let (name, value) = arg
@@ -887,6 +892,15 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_validate_exec_file() {
+        // Success case.
+        assert!(Env::validate_exec_file(OsStr::new("firecracker_xyz")).is_ok());
+
+        // Error case: filename without "firecracker".
+        assert!(Env::validate_exec_file(OsStr::new("foobarbaz")).is_err());
+    }
+
     #[test]
     fn test_setup_jailed_folder() {
         let mut mock_cgroups = MockCgroupFs::new().unwrap();
```
Splitting function looks easier to test.
How do you think about this?

zulinx86: After discussing with @luminitavoicu , I decided to adopt splitting it into a function.
- Pros of using scope.
  - It's simple.
  - We can avoid adding a new function which leads to jump to it when reading.
- Pros of splitting it into a function.
  - It's easier to test the functionality that I added this time.
  - We can keep `Env::new()` small for readability.
  - We can include the canonicalization part of `exec_file` into the function, which results in comprehensive tests.

Thanks for advice!

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@
   network interfaces. This address can be obtained as part of the GET
   `/vm/config`.
 
+### Changed
+
+- Changed the jailer option `--exec-file` to reject the filename without the
```
pb8o, line 22: .. to fail if the executable does not contain the string firecracker to prevent...

docs/jailer.md
```diff
@@ -32,8 +32,7 @@ jailer --id <id> \
 - `id` is the unique VM identification string, which may contain alphanumeric
   characters and hyphens. The maximum `id` length is currently 64 characters.
 - `exec_file` is the path to the Firecracker binary that will be exec-ed by the
-  jailer. The user can provide a path to any binary, but the interaction with
-  the jailer is mostly Firecracker specific.
+  jailer. The filename of the path must include the string `firecracker`.
```
pb8o, line 35: The filename ~~of the path~~ must include the string firecracker.

src/jailer/src/env.rs
```diff
@@ -876,6 +893,31 @@ mod tests {
...
 
+    #[test]
+    fn test_validate_exec_file() {
+        // Success case
+        File::create(PSEUDO_EXEC_FILE_PATH).unwrap();
+        assert!(Env::validate_exec_file(PSEUDO_EXEC_FILE_PATH).is_ok());
+
+        // Error case 1: No such file exists
+        std::fs::remove_file(PSEUDO_EXEC_FILE_PATH).unwrap();
+        assert!(Env::validate_exec_file(PSEUDO_EXEC_FILE_PATH).is_err());
```
pb8o, line 904: Can we be specific here (and below) about the error returned?

src/jailer/src/env.rs
```diff
@@ -876,6 +893,31 @@ mod tests {
...
+
+        // Error case 2: Not a file
```
pb8o, line 906: IMO each test function should test only one thing, so if there are 3 cases, this should be 3 separate functions. Not sure what the convention is in Rust though.

tests/integration_tests/security/test_jail.py
```diff
@@ -104,6 +104,82 @@ def test_empty_jailer_id(test_microvm_with_api):
...
+    Test the jailer option `--exec-file`
+
+    @type: security
+    """
+    test_microvm = test_microvm_with_api
+
+    # Error case 1: No such file exists
+    pseudo_exec_file_path = "/tmp/pseudo_firecracker_exec_file"
+    if os.path.exists(pseudo_exec_file_path):
+        os.remove(pseudo_exec_file_path)
```
pb8o, line 118: Use the `tmp_path` fixture as that would guarantee you an empty directory, plus you also benefit from the `pathlib` API. Then

```
pseudo_exec_file_path = tmp_path / "pseudo_firecracker_exec_file"
```

Also no need to clean up at the end, as pytest cleans up tmp_path fixtures automatically

tests/integration_tests/security/test_jail.py
```diff
@@ -104,6 +104,82 @@ def test_empty_jailer_id(test_microvm_with_api):
...
+        # a non-existing file successfully.
+        assert False
+    except Exception as err:
+        expected_err = (
+            "Jailer error: Failed to canonicalize path {}: "
+            "No such file or directory (os error 2)".format(pseudo_exec_file_path)
+        )
+        assert expected_err in str(err)
+
+    # Error case 2: Not a file
```
pb8o, line 137: I would say separate in another test.

docs/jailer.md
luminitavoicu: I would also keep the clarification that we had before, something like:
```suggestion
  jailer. The filename must include the string `firecracker`. This is enforced
  because the interaction with the jailer is Firecracker specific.
```

src/jailer/src/env.rs
```diff
@@ -247,6 +238,29 @@ impl Env {
...
+        let exec_file_path = canonicalize(exec_file)
+            .map_err(|err| Error::Canonicalize(PathBuf::from(exec_file), err))?;
+
+        if !exec_file_path.is_file() {
+            return Err(Error::NotAFile(exec_file_path));
+        }
+
+        let exec_file_name = exec_file_path
+            .file_name()
+            .ok_or_else(|| Error::FileName(exec_file_path.clone()))?
```
luminitavoicu, line 251: I think we should rename `FileName` error to `ExtractFileName` or something similar. It's confusing to keep it as is because we introduced a second error for the file name: `ExecFileName`. `FileName` it's too general and it's hard to tell what the error is about.

src/jailer/src/env.rs
```diff
@@ -876,6 +893,31 @@ mod tests {
...
+
+        // Error case 2: Not a file
+        std::fs::create_dir_all("/tmp/firecracker_test_dir").unwrap();
+        assert!(Env::validate_exec_file("/tmp/firecracker_test_dir").is_err());
+
+        // Error case 3: Filename without "firecracker"
+        File::create("/tmp/foobarbaz").unwrap();
+        File::create("/tmp/firecracker_test_dir/foobarbaz").unwrap();
+        assert!(Env::validate_exec_file("/tmp/foobarbaz").is_err());
+        assert!(Env::validate_exec_file("/tmp/firecracker_test_dir/foobarbaz").is_err());
```
luminitavoicu, line 914: Nit: Only one would suffice in my opinion. I would keep the one that has `firecracker` in directory name.

tests/integration_tests/security/test_jail.py
```diff
@@ -104,6 +104,82 @@ def test_empty_jailer_id(test_microvm_with_api):
...
+    test_microvm = test_microvm_with_api
+
+    # Error case 1: No such file exists
+    pseudo_exec_file_path = "/tmp/pseudo_firecracker_exec_file"
+    if os.path.exists(pseudo_exec_file_path):
+        os.remove(pseudo_exec_file_path)
+    test_microvm.jailer = JailerContext(
+        jailer_id="bd65600d-8669-4903-8a14-af88203add38",
+        exec_file=pseudo_exec_file_path,
+    )
```
luminitavoicu, line 122: Doing just:
```suggestion
    test_microvm.jailer.exec_file=pseudo_exec_file_path
```
would work here.

tests/integration_tests/security/test_jail.py
luminitavoicu: Same comment as above, since each of the cases is really lightweight and doesn't clutter the test.

src/jailer/src/env.rs
zulinx86: I can agree both:
- Each test function should test only one thing.
- Tiny and simple tests that are related to each other should be included in one test function.

IMO, this depends on whether we look at "only one thing" at the granularity of functionality or the granularity of test case.
When I see other test functions, those are separated at the granularity of functionality.
Thus, I'll try to follow this policy.

tests/integration_tests/security/test_jail.py
zulinx86: Same comment as above.
I'll put together multiple co-related tiny tests at the granularity of functionality.

tests/integration_tests/security/test_jail.py
```diff
@@ -104,6 +104,82 @@ def test_empty_jailer_id(test_microvm_with_api):
...
+            "No such file or directory (os error 2)".format(pseudo_exec_file_path)
+        )
+        assert expected_err in str(err)
+
+    # Error case 2: Not a file
+    pseudo_exec_dir_path = "/tmp/firecracker_test_dir"
+    if not os.path.exists(pseudo_exec_dir_path):
+        os.mkdir(pseudo_exec_dir_path)
+    test_microvm.jailer = JailerContext(
+        jailer_id="bd65600d-8669-4903-8a14-af88203add38",
```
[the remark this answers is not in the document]
zulinx86: I think that this does not matter because the spawned thread fails with invalid `exec_file` input every time.
But this is redundant, so I'll swap this as follows, as @luminitavoicu commented above:
```
test_microvm.jailer.exec_file=pseudo_exec_file_path
```

pb8o (approved):

pb8o (approved):

luminitavoicu (dismissed):

tests/integration_tests/security/test_jail.py
```diff
@@ -104,6 +104,52 @@ def test_empty_jailer_id(test_microvm_with_api):
...
+    @type: security
+    """
+    test_microvm = test_microvm_with_api
+
+    # Error case 1: No such file exists
+    pseudo_exec_file_path = tmp_path / "pseudo_firecracker_exec_file"
+    test_microvm.jailer = JailerContext(
+        jailer_id="bd65600d-8669-4903-8a14-af88203add38",
+        exec_file=pseudo_exec_file_path,
+    )
```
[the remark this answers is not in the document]
zulinx86: Oh, I didn't notice that `test_micro.jailer.jailer_id` has already been set up at this part.
Thanks!

luminitavoicu (approved):

pb8o (approved):

PR #3150, opened 2022-09-29
x86_64: save extra msrs (mtrr and mce)

kalyazin:
## Changes

This adds a new set of MSRs to be saved in x86_64: MTRR and MCE.

## Reason

Both MTRR and MCE MSRs are not advertised by KVM via KVM_GET_MSR_INDEX_LIST, however are emulated by it. This means that they have to be saved/restored when taking a snapshot.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/src/cpuid/src/cpu_leaf.rs
+++ b/src/cpuid/src/cpu_leaf.rs
@@ -61,6 +61,8 @@ pub mod leaf_0x1 {
     }
 
     pub mod edx {
+        pub const MCE_BITINDEX: u32 = 7; // Memory Check Exception
+        pub const MTRR_BITINDEX: u32 = 12; // Memory Type Range Registers
         pub const PSN_BITINDEX: u32 = 18; // Processor Serial Number
         pub const SSE42_BITINDEX: u32 = 20; // SSE 4.2
         pub const DS_BITINDEX: u32 = 21; // Debug Store.
--- a/src/cpuid/src/template/intel/mod.rs
+++ b/src/cpuid/src/template/intel/mod.rs
@@ -52,3 +52,47 @@ pub(crate) fn msrs_to_save_by_cpuid(cpuid: &CpuId) -> HashSet<u32> {
     );
     msrs
 }
+
+#[cfg(test)]
+pub mod tests {
+    use super::*;
+
+    #[test]
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    fn test_msrs_to_save_by_cpuid_empty() {
+        // No CPUID entries are provided.
+        let entrs = [];
+        let cpuid = CpuId::from_entries(&entrs).unwrap();
+
+        let msrs = msrs_to_save_by_cpuid(&cpuid);
+
+        // No MSRs are expected to match the CPUID features.
+        assert!(msrs.is_empty());
+    }
+
+    #[test]
+    #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+    fn test_msrs_to_save_by_cpuid_one() {
+        use kvm_bindings::kvm_cpuid_entry2;
+
+        use crate::cpu_leaf::leaf_0x7;
+
+        // One CPUID entry with MPX feature flag is provided
+        // that causes MSR_IA32_BNDCFGS to be pulled in.
+        let entrs = [kvm_cpuid_entry2 {
+            function: 0x7,
+            index: 0x0,
+            ebx: 1 << leaf_0x7::index0::ebx::MPX_BITINDEX,
+            ..Default::default()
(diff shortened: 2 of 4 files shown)


src/cpuid/src/template/intel/mod.rs
```diff
@@ -40,6 +40,11 @@ pub(crate) fn msrs_to_save_by_cpuid(cpuid: &CpuId) -> HashSet<u32> {
                 msrs.extend(&[$($msr),*]);
             }
         };
+        ($leaf:expr, $index:expr, $reg:tt, $feature_bit:expr, $msr:expr) => {
```
alsrdn, line 43: No need to keep both patterns here. We can just keep the last one with `$msrs:expr` that would match for both ranges and slices or other iterable expressions in this case.

src/cpuid/src/template/intel/mod.rs
```diff
@@ -50,5 +55,48 @@ pub(crate) fn msrs_to_save_by_cpuid(cpuid: &CpuId) -> HashSet<u32> {
...
     );
+
+    // MTRR
+    cpuid_msr_dep!(
+        0x1,
+        0,
+        edx,
+        leaf_0x1::edx::MTRR_BITINDEX,
+        [
+            0x200, // IA32_MTRR_PHYSBASE0
```
alsrdn, line 66: nit: you can use a range here as well.

src/cpuid/src/template/intel/mod.rs
```diff
@@ -50,5 +55,48 @@ pub(crate) fn msrs_to_save_by_cpuid(cpuid: &CpuId) -> HashSet<u32> {
...
+            0x26c, // IA32_MTRR_FIX4K_E0000
+            0x26d, // IA32_MTRR_FIX4K_E8000
+            0x26e, // IA32_MTRR_FIX4K_F0000
+            0x26f, // IA32_MTRR_FIX4K_F8000
+            0x277, // IA32_PAT
+            0x2ff  // IA32_MTRR_DEF_TYPE
+        ]
+    );
+
+    // MCE
```
alsrdn, line 98: It would be good to expand a bit on this comment. We want to save MCE MSRs that we define statically here. But the number of MCE banks is variable and depends on the CPU. Maybe add a note here that this will need to be changed in the future so that the information is not lost.

src/cpuid/src/cpu_leaf.rs
```diff
@@ -61,6 +61,8 @@ pub mod leaf_0x1 {
     }
 
     pub mod edx {
+        pub const MCE_BITINDEX: u32 = 7; // Memory Check Exception
```
alsrdn, line 64: Since we added these definitions, it would be good to have the bits set in the templates as well since that was the intention.

alsrdn (changes requested):

src/cpuid/src/template/intel/mod.rs
kalyazin: Not quite. The structure is not regular. 0x200-0x20f are contiguous, but then there are holes. We could encode 0x200-0x20f as a range and the rest as individual values, but then the macro needs to be more clever to understand it. Any suggestions?

kalyazin: Ah, maybe you meant splitting into 2 macro calls, one with a range, one with separate MSRs?

src/cpuid/src/template/intel/mod.rs
kalyazin: done

src/cpuid/src/template/intel/mod.rs
kalyazin: done

src/cpuid/src/template/intel/mod.rs
kalyazin: done

src/cpuid/src/cpu_leaf.rs
kalyazin: done

src/cpuid/src/template/intel/mod.rs
alsrdn: yup, that's what I meant. Looks good!

alsrdn (approved):
LGTM!

kalyazin:
Added unit tests for msrs_to_save_by_cpuid to fix coverage.

dianpopa:
Can you also please add a GPG signature, commits appear as Unverififed?

kalyazin:
@dianpopa Updated the GPG signature, thanks!

dianpopa (approved):

alsrdn (approved):

PR #3156, opened 2022-10-03
virtio-queue: Error out on invalid available ring index

### LINKED ISSUE #3149: [Bug] Backport fix for a self-DOS scenario from rust-vmm's vm-virtio
# Describe the bug

Backport PR https://github.com/rust-vmm/vm-virtio/pull/196 which fixes a self-DOS scenario in the virtio queue logic.
The PR solves an issue where a malitious guest driver could force the virtio device to process some descriptors more than once. This in turn could lead to a self-DOS on the thread processing the queue.

More details on the aformentioned PR.

## Environment

`[ - Firecracker version.]`
`[ - Host and guest kernel versions.]`
`[ - Rootfs used.]`
`[ - Architecture.]`

## Additional context

## Checks

- [ ] Have you searched the Firecracker Issues database for similar problems?
- [ ] Have you read the existing relevant Firecracker documentation?
- [ ] Are you certain the bug being reported is a Firecracker issue?

dianpopa:
The number of descriptor chain heads to process should always be smaller or equal to the queue size, as the driver should never ask the VMM to process a available ring entry more than once. Checking and reporting such incorrect driver behavior can prevent potential hanging and Denial-of-Service from happening on the VMM side.

Issue reported in rust-vmm/vm-virtio and fixed in
https://github.com/rust-vmm/vm-virtio/pull/196.

Co-authored-by: Bo Chen <chen.bo@intel.com>

Fixes  #3149.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/src/devices/src/virtio/queue.rs
+++ b/src/devices/src/virtio/queue.rs
@@ -49,7 +49,7 @@ impl fmt::Display for QueueError {
     }
 }
 
-/// A virtio descriptor constraints with C representive.
+/// A virtio descriptor constraints with C representative.
 #[repr(C)]
 #[derive(Default, Clone, Copy)]
 struct Descriptor {
@@ -104,9 +104,9 @@ impl<'a> DescriptorChain<'a> {
         // These reads can't fail unless Guest memory is hopelessly broken.
         let desc = match mem.read_obj::<Descriptor>(desc_head) {
             Ok(ret) => ret,
-            Err(_) => {
+            Err(err) => {
                 // TODO log address
-                error!("Failed to read from memory");
+                error!("Failed to read virtio descriptor from memory: {}", err);
                 return None;
             }
         };
@@ -289,7 +289,7 @@ impl Queue {
     }
 
     /// Returns the number of yet-to-be-popped descriptor chains in the avail ring.
-    pub fn len(&self, mem: &GuestMemoryMmap) -> u16 {
+    fn len(&self, mem: &GuestMemoryMmap) -> u16 {
         (self.avail_idx(mem) - self.next_avail).0
     }
 
@@ -300,7 +300,21 @@ impl Queue {
 
     /// Pop the first available descriptor chain from the avail ring.
     pub fn pop<'a, 'b>(&'a mut self, mem: &'b GuestMemoryMmap) -> Option<DescriptorChain<'b>> {
-        if self.len(mem) == 0 {
+        let len = self.len(mem);
+        // The number of descriptor chain heads to process should always
+        // be smaller or equal to the queue size, as the driver should
+        // never ask the VMM to process a available ring entry more than
+        // once. Checking and reporting such incorrect driver behavior
+        // can prevent potential hanging and Denial-of-Service from
+        // happening on the VMM side.
+        if len > self.actual_size() {
+            // We are choosing to interrupt execution since this could be a potential malicious
+            // driver scenario. This way we also eliminate the risk of repeatedly
+            // logging and potentially clogging the microVM through the log system.
+            panic!("The number of available virtio descriptors is greater than queue size!");
+        }
+
+        if len == 0 {
             return None;
         }
 
@@ -425,7 +439,7 @@ impl Queue {
         // via `self.is_valid()`, so it's safe to unwrap and use unchecked offsets here.
         // Note: the `MmioTransport` code ensures that queue addresses cannot be changed by the
         // guest       after device activation, so we can be certain that no change has
-        // occured since       the last `self.is_valid()` check.
+        // occurred since the last `self.is_valid()` check.
         let addr = self.avail_ring.unchecked_add(2);
         Wrapping(mem.read_obj::<u16>(addr).unwrap())
     }
@@ -461,7 +475,17 @@ impl Queue {
             return true;
         }
 
-        if self.len(mem) != 0 {
+        let len = self.len(mem);
+        if len != 0 {
+            // The number of descriptor chain heads to process should always
+            // be smaller or equal to the queue size.
+            if len > self.actual_size() {
+                // We are choosing to interrupt execution since this could be a potential malicious
+                // driver scenario. This way we also eliminate the risk of
+                // repeatedly logging and potentially clogging the microVM through
+                // the log system.
+                panic!("The number of available virtio descriptors is greater than queue size!");
+            }
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


src/devices/src/virtio/queue.rs
```diff
@@ -300,7 +300,21 @@ impl Queue {
...
+        // never ask the VMM to process a available ring entry more than
+        // once. Checking and reporting such incorrect driver behavior
+        // can prevent potential hanging and Denial-of-Service from
+        // happening on the VMM side.
+        if len > self.actual_size() {
+            // We are choosing to interrupt execution since this could be a potential malicious driver
+            // scenario. This way we also eliminate the risk of repeatedly logging and potentially
+            // clogging the microVM through the log system.
+            panic!("The number of available virtio descriptors is greater than queue size!");
+        }
```
JonathanWoollett-Light, line 315:
```suggestion
        // We are choosing to interrupt execution since this could be a potential malicious driver
        // scenario. This way we also eliminate the risk of repeatedly logging and potentially
        // clogging the microVM through the log system.
        assert!(len > self.actual_size(), "The number of available virtio descriptors is greater than queue size!");
```

src/devices/src/virtio/queue.rs
```diff
@@ -461,7 +475,16 @@ impl Queue {
...
+        let len = self.len(mem);
+        if len != 0 {
+            // The number of descriptor chain heads to process should always
+            // be smaller or equal to the queue size.
+            if len > self.actual_size() {
+                // We are choosing to interrupt execution since this could be a potential malicious driver
+                // scenario. This way we also eliminate the risk of repeatedly logging and potentially
+                // clogging the microVM through the log system.
+                panic!("The number of available virtio descriptors is greater than queue size!");
+            }
```
JonathanWoollett-Light, line 487:
```suggestion
            // We are choosing to interrupt execution since this could be a potential malicious driver
            // scenario. This way we also eliminate the risk of repeatedly logging and potentially
            // clogging the microVM through the log system.
            assert!(len > self.actual_size(), "The number of available virtio descriptors is greater than queue size!");
```

src/devices/src/virtio/queue.rs
```diff
@@ -741,6 +764,87 @@ pub(crate) mod tests {
...
+        vq.avail.ring[0].set(0);
+        vq.avail.ring[1].set(2);
+        vq.avail.idx.set(2);
+
+        // We've just set up two chains.
+        assert_eq!(q.len(m), 2);
+
+        // We process the first descriptor.
+        let d = q.pop(m).unwrap().next_descriptor();
+        assert!(d.is_some() && !d.unwrap().has_next());
```
JonathanWoollett-Light, line 802:
```suggestion
        let d = q.pop(m).unwrap().next_descriptor();
        assert!(matches!(d, Some(x) if !x.has_next()));
```

src/devices/src/virtio/queue.rs
```diff
@@ -104,9 +104,9 @@ impl<'a> DescriptorChain<'a> {
         // These reads can't fail unless Guest memory is hopelessly broken.
         let desc = match mem.read_obj::<Descriptor>(desc_head) {
             Ok(ret) => ret,
-            Err(_) => {
+            Err(e) => {
                 // TODO log address
-                error!("Failed to read from memory");
+                error!("Failed to read virtio descriptor from memory: {}", e);
                 return None;
             }
```
JonathanWoollett-Light, line 111: Prefer `err` over `e`.
```suggestion
            Err(err) => {
                // TODO log address
                error!("Failed to read from memory");
                error!("Failed to read virtio descriptor from memory: {}", err);
                return None;
            }
```

src/devices/src/virtio/queue.rs
```diff
@@ -741,6 +765,91 @@ pub(crate) mod tests {
...
+        vq.dtable[3].flags.set(0);
+        vq.avail.ring[0].set(0);
+        vq.avail.ring[1].set(2);
+        vq.avail.idx.set(2);
+
+        // We've just set up two chains.
+        assert_eq!(q.len(m), 2);
+
+        // We process the first descriptor.
+        let mut d = q.pop(m).unwrap().next_descriptor();
```
bchalios, line 804: no need to be `mut` here.

src/devices/src/virtio/queue.rs
```diff
@@ -741,6 +765,91 @@ pub(crate) mod tests {
...
 
+    #[test]
+    #[should_panic(
+        expected = "The number of available virtio descriptors is greater than queue size!"
+    )]
+    fn test_invalid_avail_idx_no_notification() {
+        // This test ensures constructing a descriptor chain succeeds
+        // with valid available ring indexes while it produces an error with invalid
+        // indexes.
+        let m = &create_anon_guest_memory(&[(GuestAddress(0), 0x1000)], false).unwrap();
```
bchalios, line 776: doesn't really really matter for the test, but shouldn't the size here be `0x4000`?

bchalios: same in the other test.

src/devices/src/virtio/queue.rs
```diff
@@ -741,6 +765,91 @@ pub(crate) mod tests {
...
+
+        // We set up a queue of size 4.
+        let vq = VirtQueue::new(GuestAddress(0), m, 4);
+        let mut q = vq.create_queue();
+
+        q.ready = true;
```
bchalios, line 782: not needed. `create_queue` marks queue as ready

src/devices/src/virtio/queue.rs
```diff
@@ -741,6 +765,91 @@ pub(crate) mod tests {
...
```
bchalios, line 832: same here. not needed

src/devices/src/virtio/queue.rs
```diff
@@ -741,6 +765,91 @@ pub(crate) mod tests {
...
+        q.uses_notif_suppression = true;
+
+        // Create 1 descriptor chain of 4.
+        for j in 0..4 {
+            vq.dtable[j].set(
+                0x1000 * (j + 1) as u64,
```
bchalios, line 838: should that be `0x100 * j as u64`?

src/devices/src/virtio/queue.rs
```diff
@@ -741,6 +765,91 @@ pub(crate) mod tests {
...
+        q.ready = true;
+
+        for j in 0..4 {
+            vq.dtable[j].set(
+                0x1000 * (j + 1) as u64,
+                0x1000,
+                VIRTQ_DESC_F_NEXT,
+                (j + 1) as u16,
+            );
+        }
```
JonathanWoollett-Light, line 791: Nit: It is always safe to cast a `u64` into a `usize` but it is not always safe to do the reverse (with stricter linting this would currently raise a warning).
```suggestion
        for j in 0u64..4u64 {
            vq.dtable[usize::from(j)].set(
                0x1000 * (j + 1),
                0x1000,
                VIRTQ_DESC_F_NEXT,
                (j + 1),
            );
        }
```

src/devices/src/virtio/queue.rs
```diff
@@ -741,6 +765,91 @@ pub(crate) mod tests {
...
+
+        // Create 1 descriptor chain of 4.
+        for j in 0..4 {
+            vq.dtable[j].set(
+                0x1000 * (j + 1) as u64,
+                0x1000,
+                VIRTQ_DESC_F_NEXT,
+                (j + 1) as u16,
+            );
+        }
```
JonathanWoollett-Light, line 843: Nit: It is always safe to cast a u64 into a usize but it is not always safe to do the reverse (with stricter linting this would currently raise a warning).
```suggestion
for j in 0u64..4u64 {
            vq.dtable[usize::from(j)].set(
                0x1000 * (j + 1),
                0x1000,
                VIRTQ_DESC_F_NEXT,
                (u16::from(j) + 1),
            );
        }
```

src/devices/src/virtio/queue.rs
dianpopa: This is a good point @JonathanWoollett-Light! However, this applies to all queue tests in this file. Can you please open an issue for this? Thanks!

src/devices/src/virtio/queue.rs
dianpopa: Uh actually it needs to be at least 0x4000 since we also need to reserve place for the contents of the queue, right?

src/devices/src/virtio/queue.rs
dianpopa: No since the first 0-0x1000 will be reserved for the virtqueue (desc table, avail ring and used ring).
The contents to which the desc table points to will start from 0x1000.

src/devices/src/virtio/queue.rs
dianpopa: Applies to all virtio queue unit tests. We need to open an issue for this.

src/devices/src/virtio/queue.rs
bchalios: yes, it looks like it should be at least 0x5000 actually (4 descriptors plus the queue)

bchalios (approved):

src/devices/src/virtio/queue.rs
JonathanWoollett-Light: > At least, this is what we have targeted to do until now so that it is easier to track the panic branches we have in production code.

Could you expand, how do we currently track panic branches?

> Also, we have a panic handler installed when starting Firecracker (for flushing metrics and whatnot) which I am not sure would get called if we use asssert.

The docs for [`std::assert`](https://doc.rust-lang.org/std/macro.assert.html) specify:

> This will invoke the [panic!](https://doc.rust-lang.org/core/macro.panic.html) macro if the provided expression cannot be evaluated to true at runtime.

src/devices/src/virtio/queue.rs
JonathanWoollett-Light: https://github.com/firecracker-microvm/firecracker/pull/3156#discussion_r985864587

src/devices/src/virtio/queue.rs
JonathanWoollett-Light: With a wider context, maybe an issue should relate to adding `#![warn(clippy::as_conversions)]` (https://rust-lang.github.io/rust-clippy/master/)?

src/devices/src/virtio/queue.rs
dianpopa: >  At least, this is what we have targeted to do until now so that it is easier to track the panic branches we have in production code.
> 
>> Could you expand, how do we currently track panic branches?
>
We do not have an automated process or something of the kind. But if we were to use `assert`s in production code (which are extensively used throughout unit tests) is going to be quite challenging to uncover those if the need arises. That would be one of the practical aspects of it.
> > Also, we have a panic handler installed when starting Firecracker (for flushing metrics and whatnot) which I am not sure would get called if we use asssert.
> 
> The docs for [`std::assert`](https://doc.rust-lang.org/std/macro.assert.html) specify:
> 
> > This will invoke the [panic!](https://doc.rust-lang.org/core/macro.panic.html) macro if the provided expression cannot be evaluated to true at runtime.

Nice, thanks for looking this up!
Can you tell me is there is a strong reason for using assert over panic here?






> I prefer to use panic over assert in production code. At least, this is what we have targeted to do until now so that it is easier to track the panic branches we have in production code. Also, we have a panic handler installed when starting Firecracker (for flushing metrics and whatnot) which I am not sure would get called if we use asssert.

src/devices/src/virtio/queue.rs
dianpopa: Or maybe modify add this restriction in the `cargo clippy` command directly if possible.

mattschlebusch (approved):

src/devices/src/virtio/queue.rs
JonathanWoollett-Light: It is simpler, but beyond that, no strong reasons.

src/devices/src/virtio/queue.rs
JonathanWoollett-Light: Created basic issue https://github.com/firecracker-microvm/firecracker/issues/3161

PR #3158, opened 2022-10-04
Add tool to check bindgen structs

bchalios:
## Changes

Add a small python script that uses pahole (man 1 pahole) to test if the bindings created between two firecracker binaries match. Currently, this compares for the size of all structs created through bindgen.

## Reason

When creating new bindings through bindgen we want to have visibility into possible changes in the binary layout of the new structs.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tools/test_bindings.py
+++ b/tools/test_bindings.py
@@ -1,0 +1,165 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Script used to check if bindgen-generated code creates structs that differ from previously created
+onces.
+
+The script uses `pahole` (man 1 pahole) to gather debug information from two firecracker binaries
+(script's arguments). It parses pahole output and gathers struct information in a dictionary of the
+form:
+
+```
+{
+    "struct_name": {"size": size_in_bytes, "alignment": alignment_in_bytes},
+    ...
+}
+```
+
+It also, filters structure names using the "bindings" filter for keeping only bindgen related
+structs.
+
+*NOTE*: this assumes that all bindgen-related structs live under a crate or module name with
+"bindings" in it. At the moment, this is true.
+
+It then iterates through the structs of the firecracker binary built from the older version and
+checks if there are mismatches with the struct info from the second binary (newer version)
+
+### Usage
+
+1. Create the two binaries
+
+```
+# First create the binary with existing bindings
+$ git checkout main
+$ ./tools/devtool build
+$ cp ./build/cargo_target/x86_64-unknown-linux-musl/debug/firecracker firecracker_old
+
+# Second create the binary with new bindings
+$ git checkout new_bindings
+$ ./tools/devtool build
+$ cp ./build/cargo_target/x86_64-unknown-linux-musl/debug/firecracker firecracker_new
+
+# Run the script
+$ python3 ./tools/test_bindings.py firecracker_old firecracker_new
+```
+"""
+
+import sys
+import re
+import logging
+import subprocess
+import argparse
+
+logging.basicConfig(level=logging.DEBUG)
+log = logging.getLogger(__name__)
+
+
+def parse_pahole(pahole_output):
+    """Gather bindings related structs from pahole output
+
+    Parse pahole output and gather struct information filtering for the 'bindings' keyword.
+    The information gathered is the struct size and its alignment.
+
+    @param fname: File including pahole output
+    @return: A dictionary where keys are struct names and values struct size and alignment
+    """
+    ret = {}
+
+    # regular expression matches the name of the struct, its size and alignment
+    structs = re.findall(
+        rb"struct (.*?)\{.*?/\* size: (\d+).*?\*/.*?\n\} "
+        rb"__attribute__\(\(__aligned__\((\d+)\)\)\)\;",
+        pahole_output,
+        flags=re.DOTALL,
+    )
+
+    for struct in structs:
+        struct_name = str(struct[0])
+        size = int(struct[1])
+        alignment = int(struct[2])
+
+        if "bindings" in struct_name:
+            ret[struct_name] = {"size": size, "alignment": alignment}
+
+    return ret
+
+
+def pahole(binary: str) -> str:
+    """Runs pahole on a binary and returns its output as a str
+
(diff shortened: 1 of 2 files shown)


pb8o (dismissed):

tools/test_bindings.py
```diff
@@ -0,0 +1,142 @@
...
+            log.error("new: %s", prop_2['size'])
+            return True
+
+        # Alignment mismatches just cause warnings
+        if prop_1['alignment'] != prop_2['alignment']:
+            log.warning("alignment of '%s' does not match in two binaries", name)
+            log.warning("old: %s", prop_1['alignment'])
+            log.warning("new: %s", prop_2['alignment'])
+        else:
+            log.info("struct '%s' matches", name)
```
zulinx86, line 131: This does not check for bindings that are in the second but not in the first, which means that only backward compatibility is assured.
Is this intentional? Let me check if this is enough to fulfill our objective just in case.

zulinx86 (changes requested):

tools/test_bindings.py
```diff
@@ -0,0 +1,142 @@
...
+$ ./tools/devtool build
+$ cp ./build/cargo_target/x86_64-unknown-linux-musl/debug/firecracker firecracker_old
+
+# Second create the binary with new bindings
+$ git checkout new_bindings
+$ ./tools/devtool build
+$ cp ./build/cargo_target/x86_64-unknown-linux-musl/debug/firecracker firecracker_new
+
+# Run the script
+$ ./tools/test_bindings.py firecracker_old firecracker_new
```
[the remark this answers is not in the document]
bchalios: you are right. I fixed it. Actually now I'm requiring that you run `python3 ./tools/test_bindings.py firecracker_old firecracker_new`.

tools/test_bindings.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
bchalios: Fixed due to removing shebang (see https://github.com/firecracker-microvm/firecracker/pull/3158#discussion_r987635463)

tools/test_bindings.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Parse pahole output and gather struct information filtering for the 'bindings' keyword.
+    The information gathered is the struct size and its alignment.
+
+    @param fname: File including pahole output
+    @return: A dictionary where keys are struct names and values struct size and alignment
+    """
+    ret = {}
+
+    #regural expression matches the name of the struct, its size and alignment
```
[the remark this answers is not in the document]
bchalios: Fixed! Thanks

tools/test_bindings.py
bchalios: It is intentional. The purpose of this test is that it checks whether there are bindings we are already using (`firecracker_old`) that have changed in the new version. New bindings cannot break ABI compatibility.

I 'll add a comment explaining that.

bchalios: Ok, added comments in the code and also used `argparse` to give to the arguments passed to the script a bit more meaning. Hopefully, this is more clear now?

zulinx86: That comment is very helpful and clear! Thanks!

zulinx86 (dismissed):

tools/test_bindings.py
```diff
@@ -0,0 +1,147 @@
...
+                        help="Firecracker binary with new bindings")
+    args = parser.parse_args()
+
+    if check_pahole_mismatches(args.firecracker_old, args.firecracker_new):
+        log.error("pahole information mismatch")
+        sys.exit(1)
+    else:
+        log.info("pahole information matches")
+
+    sys.exit(0)
```
JonathanWoollett-Light, line 147: Could be simplified:
```suggestion
    if check_pahole_mismatches(args.firecracker_old, args.firecracker_new):
        log.error("pahole information mismatch")
        sys.exit(1)
    else:
        log.info("pahole information matches")
```
Also may be worth using `raise Exception("pahole infomation mismatch")` to replace:
```python
log.error("pahole information mismatch")
sys.exit(1)
```
But I'm not sure here.

tools/test_bindings.py
```diff
@@ -0,0 +1,147 @@
...
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Check bindings ABI compatibility for Firecracker')
+    parser.add_argument("firecracker_old", type=str, metavar="old-firecracker-binary",
+                        help="Firecracker binary with old bindings")
+    parser.add_argument("firecracker_new", type=str, metavar="new-firecracker-binary",
+                        help="Firecracker binary with new bindings")
+    args = parser.parse_args()
+
+    if check_pahole_mismatches(args.firecracker_old, args.firecracker_new):
+        log.error("pahole information mismatch")
```
dianpopa, line 142: I would not use `pahole` in log message (I personally would not know what it refers too :D)
I would just say "structure layout mismatch" and print (if possible) the name of the structure in question.

bchalios: Updated comment. The name of the struct that has the mismatch will already be logged.

zulinx86 (dismissed):

tools/test_bindings.py
```diff
@@ -0,0 +1,147 @@
...
+$ git checkout new_bindings
+$ ./tools/devtool build
+$ cp ./build/cargo_target/x86_64-unknown-linux-musl/debug/firecracker firecracker_new
+
+# Run the script
+$ python3 ./tools/test_bindings.py firecracker_old firecracker_new
+```
+"""
+
+import sys
```
pb8o, line 48: nit: you may want to run it through black / isort / flake8

pb8o (approved):

tools/test_bindings.py
bchalios: hmmm, the pylint test from the CI doesn't complain. Shouldn't it?

pb8o: Maybe it should, but I think we only do checks within tests due to this: https://github.com/firecracker-microvm/firecracker/blob/main/tools/devtool#L1270

bchalios: Ok, passed it through `black`. Thanks for the suggestion.

pb8o (approved):

zulinx86 (approved):

PR #3164, opened 2022-10-05
snap/restore: Save MSRs based on CPUID

alsrdn:
## Changes

This PR brings a few changes to the way we save MSRs in snapshots. 
First, it fixes a bug where Firecracker didn't check the number of MSRs that were actually successfully set by KVM.
Secondly it introduces a way to define dependencies between CPUID feature bits and MSRs that need to be saved.
Finally it extends the number of MSRs that can be saved in the snapshot.

## Reason

Firecracker uses the `KVM_GET_MSR_INDEX_LIST` KVM ioctl to get the list of MSRs that it needs to save in a snapshot. This list does not always reflect which MSRs are actually available by the Guest. The correct list of MSRs has a dependency of what CPUID features are set for the Guest vCPU. Also this list returned by KVM excludes some MSRs that depend on KVM features or that can be retrieved by other ioctl (e.g. MCE MSRs).
We implement a way to define extra MSRs to be saved depending on the CPUID features to fix the issues above.

Also the number of MSRs we save is limited to `KVM_MAX_MSR_ENTRIES` which may not be large enough to accommodate all MSRs that need to be saved. Starting with this patch we bump the version of the vCPU state to accommodate larger numbers of MSRs in the snapshot.

We always used to save the `MSR_IA32_BNDCFGS` MSR if returned by KVM in the list. Now we don't save that MSR by default if it's not used by the Guest to allow snapshots taken on machines that have MPX to be resumed on other machines that don't have MPX if the feature is not used. This is just one example, and we probably need to do this for more MSRs as well.

## License Acceptance

## PR Checklist

---


src/cpuid/src/template/mod.rs
```diff
@@ -3,3 +3,22 @@
...
+use std::collections::HashSet;
+
+use kvm_bindings::CpuId;
+
+use crate::common::{get_vendor_id_from_cpuid, VENDOR_ID_INTEL};
+use crate::transformer::Error;
+
+/// Returns MSRs to be saved based on CPUID features that are enabled.
+pub fn msrs_to_save_by_cpuid(cpuid: &CpuId) -> Result<HashSet<u32>, Error> {
+    let vendor_id = get_vendor_id_from_cpuid(cpuid).map_err(|_| Error::InvalidVendor)?;
```
kalyazin, line 16: nit: not sure what the convention is, but could be just (I guess)
```rust
.map_err(Error::InvalidVendor)
```

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -529,6 +575,31 @@ impl VcpuState {
...
+
+    fn de_saved_msrs(&mut self, source_version: u16) -> VersionizeResult<()> {
+        if source_version < 3 {
+            self.saved_msrs.push(self.msrs.clone());
+        }
+        Ok(())
+    }
+
+    fn ser_saved_msrs(&mut self, target_version: u16) -> VersionizeResult<()> {
+        if target_version < 3 && self.saved_msrs.len() == 1 {
```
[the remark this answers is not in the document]
alsrdn: Yes, actually having no MSRs to save would mean something went wrong so I'll add an error here.

src/cpuid/src/template/mod.rs
alsrdn: It is this way because we map `cpuid::common::Error` to `cpuid::transformer::Error`.

src/cpuid/src/template/intel/mod.rs
```diff
@@ -20,3 +26,29 @@ pub fn validate_vendor_id() -> Result<(), Error> {
 
     Ok(())
 }
+
+/// Returns MSRs to be saved based on the Intel CPUID features that are enabled.
+pub(crate) fn msrs_to_save_by_cpuid(cpuid: &CpuId) -> HashSet<u32> {
```
kalyazin, line 31: Now I started wondering, is this approach only applicable to a templated execution? I guess we should consult CPUID when choosing which MSRs to save regardless if a template is used or not in general. There will be Intel/AMD specific features/MSRs, but there is a considerable overlap.

alsrdn: It's not only applicable to templates and it is used even if we don't use templates. Only the higher level function is exported and at some point we might want to add common things there for X86 general registers.
The reason why I added this functionality in the `template` module is just convenience. Ideally we would want to move all the template stuff in another crate that is separate from CPUID. Not sure if we should move this functionality outside of the template module since it will be used by it too and, in a way, the passthrough CPU configuration is like a template by itself.

But I'm open to suggestions if you think we need to move it somewhere else.

kalyazin: Indeed, it is used in non-templated version. I was misled in my analysis. I think the function could stay were it is until we have a better idea of how to structure the MSR/CPUID dependency code.

kalyazin (approved):
Is this still in draft for a reason?

pb8o (dismissed):

kalyazin (approved):

aghecenco (approved):
RSLGTM

PR #3167, opened 2022-10-06
[Snapshot] Performance baselines for Intel 4.14

dianpopa:
## Changes

* gathered 4.14 baselines on Intel

## Reason

Snapshot GA

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/performance/configs/test_snap_restore_performance_config_4.14.json
+++ b/tests/integration_tests/performance/configs/test_snap_restore_performance_config_4.14.json
@@ -28,252 +28,352 @@
                                     "ubuntu-18.04.ext4": {
                                         "1vcpu_128mb": {
                                             "P50": {
-                                                "target": 3.819,
-                                                "delta_percentage": 9
+                                                "restore": {
+                                                    "target": 3.623,
+                                                    "delta_percentage": 18
+                                                }
                                             },
                                             "P90": {
-                                                "target": 4.764,
-                                                "delta_percentage": 37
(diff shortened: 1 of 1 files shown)


tools/parse_baselines/providers/snapshot_restore.py
```diff
@@ -33,8 +33,14 @@ def __init__(self, data_provider: Iterator):
...
+        max_ = max(data)
+        print(len(data))
+        min_delta = 100 * abs(avg - min_) / avg
+        max_delta = 100 * abs(avg - max_) / avg
+        delta = max(max_delta, min_delta)
+
         return {
             "target": round(avg, 3),
-            "delta_percentage": math.ceil(3 * stddev / avg * 100) + DELTA_EXTRA_MARGIN,
+            "delta_percentage": math.ceil(delta) + DELTA_EXTRA_MARGIN,
```
bchalios, line 45: Doesn't this logic mean that we should re-run performance baselines for everything?

Also, do you see any value in keeping in the data format both the delta as in (max - min) and the standard deviation? Even though we will only use the delta for signaling problems with performance the standard deviation could be useful information when we plot data.

dianpopa: Not necessarily since the current logic should include the old values (actually this is a more relaxed way to compute values -> what it does is that it will include the entire range covered by a number of build).

bchalios (approved):

mattschlebusch (approved):

PR #3105, opened 2022-08-24
CPUID (2/3): Bit fields

JonathanWoollett-Light:
### https://github.com/firecracker-microvm/firecracker/pull/3104 must be merged first

# Reason for This PR

Working with the large number of values smaller than `u8` required for CPUID quickly becomes untenable. With thousands of constants denoting offsets.

[bitflags](https://crates.io/crates/bitflags) does not support fields which are required for CPUID. To use [bitflags](https://crates.io/crates/bitflags) here would require high hundreds (possibly 1000+) of additional loosely associated constants to reference these fields within the `bitflag` structures. This would form a large and difficult to read implementation.

[bitfield](https://crates.io/crates/bitfield) does not support exotically sized bit ranges (e.g. bits 2-7). To use this would require heavy adaption.

## Description of Changes

Introduces `bit-fields` and `bit-fields-macros` crates to support handling these values. These function as one crate[^one] inacting a similar role to `bitflags` except with additional support for fields within bytes used for unsigned integer values.

[^one]: A crates cannot export both procedural-macros and non-procedural-macros, so it requires 2 crates acting as 1, like `serde` and `serde-derive`

### Linting

`bit-fields` and `bit-fields-macros` raise the bar in safety adhering to much stricter linting. This is defined with:
```
#![warn(clippy::pedantic, clippy::restriction)]
#![allow(
    clippy::blanket_clippy_restriction_lints,
    clippy::implicit_return,
    clippy::pattern_type_mismatch,
    clippy::std_instead_of_alloc,
    clippy::std_instead_of_core,
    clippy::pub_use,
    clippy::non_ascii_literal,
    clippy::single_char_lifetime_names,
    clippy::exhaustive_enums,
    clippy::exhaustive_structs,
    clippy::unseparated_literal_suffix,
    clippy::mod_module_files
)]
```
Beginning from the strictest possible linting `#![warn(clippy::pedantic, clippy::restriction)]` then disabling some of the lints within `clippy::restriction` to make it reasonable.

### Testing & Coverage

**Coverage = ~97% or ~75% with grcov or ~93% with cargo-llvm-cov**

Both crates feature extensive testing. Due to #3206 when running tests within the container and in CI coverage is greatly reduced to ~40% ([recent-coverage.zip](https://github.com/firecracker-microvm/firecracker/files/9858985/recent-coverage.zip)).

When running with doc tests:

```
[...]

PR #3173, opened 2022-10-10
Update code coverage

JonathanWoollett-Light:
## Changes

Changed code coverage to use [grcov](https://github.com/mozilla/grcov).

### Coverage value changes

grcov includes the test code as covered lines (where as kcov does not appear to do so). This moderately increases our line coverage.

### Unit test and profiling tool chains

Previously we ran tests with the `aarch64-unknown-linux-musl` and `x86_64-unkown-linux-musl` tool chains to generate code coverage data then ran the unit tests under the `x86_64-unknown-linux-gnu` and `aarch64-unknown-linux-gnu` tool chains. This accomplished testing both both combinations of tool chains covering both our primary `musl` targets and the commonly used `gnu` target.

At present `aarch64-unknown-linux-musl` is not supported with profiling by Rust lang, I opened an issue regarding this https://github.com/rust-lang/rustup/issues/3095 (interestingly `x86_64-unknown-linux-musl` is supported https://github.com/rust-lang/rust/issues/79556). To accommodate this we need to generate code coverage using the respective `gnu` tool chains of `x86_64-unknown-linux-gnu` and `aarch64-unknown-linux-gnu` and then run unit tests with the `musl` tool chains of `aarch64-unknown-linux-musl` and `x86_64-unkown-linux-musl`. This does shift our code coverage metric generation from `musl` to `gnu` although maintains the same actual code coverage for both. I do not believe this represents a decrease in security or safety. Actual test coverage across all tool chains is identical as before and it is extremely unlikely the test coverage will diverge between the `gnu` and `musl` tool chains in the future (or until profiling is support by `aarch64-unknown-linux-musl` at which point we can use the `musl` tool chains for code coverage generation).

## Reason

We currently use [cargo_kcov](https://crates.io/crates/cargo-kcov) (a wrapper around [kcov](https://github.com/SimonKagstrom/kcov)). [cargo_kcov](https://crates.io/crates/cargo-kcov) appears unmaintained with the last commit >2 years ago.

[grcov](https://github.com/mozilla/grcov) is a popular coverage tool for Rust. It does not produce the same error as [kcov](https://github.com/SimonKagstrom/kcov) with https://github.com/firecracker-microvm/firecracker/pull/3105. It offers a better solution in the long term.

The primary advantages of [grcov](https://github.com/mozilla/grcov) over [kcov](https://github.com/SimonKagstrom/kcov) are:
- Better documentation,
- Higher contributions and maintenance,
- Better integration with Rust (using the default coverage mechanism https://doc.rust-lang.org/rustc/instrument-coverage.html and avoiding the need for a wrapper).
- Allow multiple output targets, including lcov allowing better integration with tooling (e.g. https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters).

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -8,15 +8,10 @@
   target should be put in `s3://spec.firecracker` and automatically updated.
 """
 
-
 import os
-import platform
-import re
-import shutil
 import pytest
 
 from framework import utils
-import host_tools.cargo_build as host  # pylint: disable=import-error
 from host_tools import proc
 
 # We have different coverages based on the host kernel version. This is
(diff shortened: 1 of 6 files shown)


dianpopa:
I do not agree with some of the supporting arguments for switching to `grcov`:
1. cargo-kcov is just a wrapper over [kcov](https://github.com/SimonKagstrom/kcov); I am skeptical the issue steams from there. You can actually try running plain `kcov` and see if you still receive the error; As far as maintainance goes, `kcov` has supported us in adding arm support a while ago: https://github.com/SimonKagstrom/kcov/pull/316.
2. `proc-macro2` on its own cannot be the root cause of the problem; we are using it in [versionize-derive](https://github.com/firecracker-microvm/versionize_derive/blob/main/src/lib.rs#L212) and we are able to run kcov successfuly:  https://buildkite.com/firecracker/versionize-derive-ci/builds/23. 

As far as my concern, we do not have enough datapints to affirm that kcov is the problem and we need to identify what part of the code triggers the error.

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,114 +24,96 @@
...
-        "bootparam.rs,"
-        "elf.rs,"
-        "mpspec.rs,"
-        "msr_index.rs,"
-        "bindings.rs,"
-        "_gen"
+    # Get coverage target.
+    processor_model = [item for item in COVERAGE_DICT if item in PROC_MODEL]
+    assert len(processor_model) == 1, "Could not get processor model!"
+    coverage_target = COVERAGE_DICT[processor_model[0]]
```
pb8o, line 65: I think this is a bit obfuscated. It can be replaced with 

```
assert PROC_MODEL in COVERAGE_DICT, "Could not get processor model!"
coverage_target = COVERAGE_DICT[PROC_MODEL]
```

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,114 +24,96 @@
...
+
+    # Add `llvm-tools-preview` component
+    utils.run_cmd("rustup component add llvm-tools-preview")
+
+    # Generate test profiles.
+    utils.run_cmd(
+        f'\
+        export RUSTFLAGS="-Cinstrument-coverage" \
+        && export LLVM_PROFILE_FILE="your_name-%p-%m.profraw" \
+        && cargo test --all --target={TARGET} -- --test-threads=1 \
```
pb8o, line 78: Maybe better as `env RUSTFLAGS="-Cinstrument-coverage" LLVM_PROFILE_FILE="your_name-%p-%m.profraw" cargo test --all --target={TARGET} -- --test-threads=1 ...`

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,114 +24,96 @@
...
+    # Get coverage target.
+    processor_model = [item for item in COVERAGE_DICT if item in PROC_MODEL]
+    assert len(processor_model) == 1, "Could not get processor model!"
+    coverage_target = COVERAGE_DICT[processor_model[0]]
+
+    # Re-direct to repository root.
+    os.chdir("..")
+
+    # Add `llvm-tools-preview` component
+    utils.run_cmd("rustup component add llvm-tools-preview")
```
pb8o, line 71: Should this move to the Dockerfile? How big is it?

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,114 +24,96 @@
...
+    # Extract coverage from html report.
+    # When we update grcov to 0.8.* we can update this to pull the coverage from a generated .json
+    # file.
+    index = open(
+        f"./build/cargo_target/{TARGET}/debug/coverage/index.html", encoding="utf-8"
+    )
+    index_contents = index.read()
+    end = index_contents.find(" %</abbr></p>")
+    start = index_contents[:end].rfind(">")
+    coverage_str = index_contents[start + 1 : end]
```
pb8o, line 106: Maybe put an example of a line containing the coverage. Can we use XPath here? https://docs.python.org/3/library/xml.etree.elementtree.html#xpath-support

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,114 +24,96 @@
...
-            coverage, COVERAGE_MAX_DELTA, coverage_target_pct, cov_target_name
-        )
+    # Generate coverage report.
+    utils.run_cmd(
+        f'\
+        cargo install --version {GRCOV_VERSION} grcov \
+        && grcov . \
+            -s . \
+            --binary-path ./build/cargo_target/{TARGET}/debug/ \
+            --ignore "build/*" \
```
pb8o, line 89: Should we also exclude what the old coverage excluded? Is that why the coverage % bumps so high?

tests/integration_tests/build/test_coverage.py
pb8o: also could replace `your_name` with `coverage` or sthg

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: Not excluding bindings and some generated code is causing a coverage bump.

I think these should be included in our coverage, if its code we rely on whether auto-generated or not I think unless we have a strong argument to exclude it, it is better to include by default. I think including this change in this PR is fair, but I don't have a strong opinion on that.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: Added an example line. I'm not sure about XPath will need to look at this.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: This doesn't work since `PROC_MODEL` may be `Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz` such that `COVERAGE_DICT[PROC_MODEL]` will be invalid.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: This step is quite fast, doing the code:
```
...
# Re-direct to repository root.
os.chdir("..")

import time
start = time.process_time()

# Add `llvm-tools-preview` component
utils.run_cmd("rustup component add llvm-tools-preview")

print("time to add llvm-tools-preview:",time.process_time() - start)
assert False

utils.run_cmd(
...
```
prints:
```
time to add llvm-tools-preview: 0.0008735560000000002
```
I would prefer having it here over docker as it makes its purpose clearer.

bchalios: I think it's not a good practice to install stuff on someone's machine while running a set of tests. It should be either mentioned somewhere as a dependency, or (ideally) added in the Docker image

JonathanWoollett-Light: It can be moved to the docker container (for now I'd like to leave it here, then test and resolve this conversation immediately before merging so as to avoid creating a docker container too early).

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: Looking into it a bit more, the html generated seems to result in errors with `xml.etree.ElementTree.parse()`. I think using a library to parse the html here will be more complicated than the current approach (since we only need to extract a small amount of data).

roypat:
> grcov includes the test code as covered lines (where as kcov does not appear to do so). This moderately increases our line coverage.

I think this is because we explicitly exclude test modules for kcov:
https://github.com/firecracker-microvm/firecracker/blob/87e0613b8be73657e53745902a070b446cad03ef/tests/integration_tests/build/test_coverage.py#L81

tests/integration_tests/build/test_coverage.py
roypat: HTML is generally not valid XML, as XML is a stricter standard (there were some attempts to make HTML conform to those stricter standard in the form of [XHTML](https://www.w3schools.com/html/html_xhtml.asp), but it never caught on). That's probably why you see those errors

JonathanWoollett-Light:
@roypat Yes this is why. Excluding tests is a non-optimal solution here. I would propose to evaluate test code coverage and production code coverage separately. With this approach we can get the benefits of evaluating test coverage without the negative of this giving us false confidence in our production code coverage.

I will make the change to exclude test code.

tests/integration_tests/build/test_coverage.py
pb8o: Ah I see it now.

tests/integration_tests/build/test_coverage.py
pb8o: hmm I seem to remember there was sthg in Python stdlib to parse html... maybe I got confused with BeautifulSoup, but that's another dependency.

pb8o (approved):

aghecenco (approved):

PR #3175, opened 2022-10-10
Add CPUID snapshot restore test

kalyazin:
## Changes

The 2-part test checks if the CPUID retains its value after a VM is restored from a snapshot.
It dumps CPUID before taking snapshot, dumps CPUID after restoring from the snapshot and then compares the two.

pylint's too-many-lines is disabled in the test file as there is no good way to split the file at the moment.  As it grows further, there should be an opportunity to reenable the check, for example, by refactoring.

## Reason

CPU template feature allows exposing predictable CPUID and MSRs to the guest regardless of the host CPU type. There is a need for a test that would check that CPUID remains the same after restoring from a snapshot on the CPU of a different type.

## License Acceptance

## PR Checklist

---


kalyazin:
Going to merge this to https://github.com/firecracker-microvm/firecracker/pull/3132 as the changes are related.

PR #3182, opened 2022-10-14
doc: add MSR_IA32_TSX_CTRL snapshot limitation

kalyazin:
## Changes

Document MSR_IA32_TSX_CTRL limitation when taking snapshot.

## Reason

Inform users of a known limitation.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,8 @@ plans, check out our [kernel support policy](docs/kernel-policy.md).
 - The [SendCtrlAltDel](docs/api_requests/actions.md#sendctrlaltdel) API request
   is not supported for aarch64 enabled microVMs.
 - Configuring CPU templates is only supported for Intel enabled microVMs.
+- If a CPU template is not used on x86_64, overwrites of `MSR_IA32_TSX_CTRL` MSR
+  value will not be preserved after restoring from a snapshot.
 - The `pl031` RTC device on aarch64 does not support interrupts, so guest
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -151,6 +151,8 @@ plans, check out our [kernel support policy](docs/kernel-policy.md).
 - The [SendCtrlAltDel](docs/api_requests/actions.md#sendctrlaltdel) API request
   is not supported for aarch64 enabled microVMs.
 - Configuring CPU templates is only supported for Intel enabled microVMs.
+- If a CPU template is not used in x86_64, MSR_IA32_TSX_CTRL MSR may not retain
```
mattschlebusch, line 154: Would it be possible to be more specific when the register would not retain its value?

kalyazin: Well, we don't save/restore it, but if the guest never changes it, the restored value will match the original value. It only affects those who override the default. Not sure how to put it better.

luminitavoicu: Maybe it would be useful to phrase it something like: 

``` suggestion
- If a CPU template is not used in x86_64, overwrites of `MSR_IA32_TSX_CTRL` MSR 
  will not be preserved after restoring from a snapshot.
```
?

kalyazin: Yes, looks good to me, thanks for the suggestion. Updated.

luminitavoicu (approved):

mattschlebusch (approved):

PR #3209, opened 2022-10-24
test/versioned_serialization: update vmm baselines

kalyazin:
## Changes

Increase baseline values in the VMM serialisation test.

## Reason

Serialisation takes longer due to addition of MTRR and MCE MSRs to the snapshot.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/performance/test_versioned_serialization_benchmark.py
+++ b/tests/integration_tests/performance/test_versioned_serialization_benchmark.py
@@ -23,11 +23,11 @@ BASELINES = {
     "Intel": {
         "serialize": {
             "no-crc": {"target": 0.150, "delta": 0.036},  # milliseconds  # milliseconds
-            "crc": {"target": 0.205, "delta": 0.04},  # milliseconds  # milliseconds
+            "crc": {"target": 0.244, "delta": 0.44},  # milliseconds  # milliseconds
         },
         "deserialize": {
-            "no-crc": {"target": 0.034, "delta": 0.02},  # milliseconds  # milliseconds
-            "crc": {"target": 0.042, "delta": 0.025},  # milliseconds  # milliseconds
+            "no-crc": {"target": 0.056, "delta": 0.02},  # milliseconds  # milliseconds
+            "crc": {"target": 0.046, "delta": 0.035},  # milliseconds  # milliseconds
         },
     },
     "AMD": {


JonathanWoollett-Light (approved):

aghecenco (approved):

tests/integration_tests/performance/test_versioned_serialization_benchmark.py
```diff
@@ -23,11 +23,11 @@
     "Intel": {
         "serialize": {
             "no-crc": {"target": 0.150, "delta": 0.036},  # milliseconds  # milliseconds
-            "crc": {"target": 0.205, "delta": 0.04},  # milliseconds  # milliseconds
+            "crc": {"target": 0.244, "delta": 0.44},  # milliseconds  # milliseconds
```
mattschlebusch, line 26: This is a large increase in the delta? Is this intended?

kalyazin: It is large. What I have been seeing recently is a very high variance of the samples (I saw the test hitting both high and low bounds). This change is certainly a tactical one to allow other PRs to be merged.
We need to do a proper investigation and understand why the variance is so high and whether the baselines are sensible.

mattschlebusch: I'm not convinced this increase is related to recent MSR changes introduced into main. This pipeline runs on an environment from a multitude of dependencies, especially Linux Kernel changes. I'm inclined to move forward, but polishing our reporting and tracking here is a must.

JonathanWoollett-Light (approved):

tests/integration_tests/performance/test_versioned_serialization_benchmark.py
JonathanWoollett-Light: I would agree, it may be worth creating an issue to track this.

kalyazin: The way I see it now is we have two separate factors:
- new MSRs in the snapshot increasing the mean time. This can be addressed via proportional increase of the target.
- unknown factor increasing variance. We need to investigate it to understand better.

kalyazin: The latter, in my opinion, is not even a single issue to solve. We should approach it in a systematic way and possibly reassess what baselines we collect and how, as well as how we monitor them.

JonathanWoollett-Light: > The latter, in my opinion, is not even a single issue to solve. We should approach it in a systematic way and possibly reassess what baselines we collect and how, as well as how we monitor them.

I think this larger item (which is also worth tracking) links to this specific case, but it is also worth having a specific issue for why this test specifically illustrates more problematic behavior.

mattschlebusch (approved):

Issue #3188, opened 2022-10-18
[Bug] Build - Cross Compilation Failure


jwinarske:
# Describe the bug

When cross compiling firecracker the secompiler-bin execution step fails.

## To Reproduce

Cross compile using Yocto `kirkstone` build system using this recipe
```
SUMMARY = "Firecracker"
HOMEPAGE = "https://github.com/firecracker-microvm/firecracker"
ISSUES = "https://github.com/firecracker-microvm/firecracker/issues"
LICENSE = "Apache-2.0"

LIC_FILES_CHKSUM = "\
    file://${COREBASE}/meta/files/common-licenses/Apache-2.0;md5=89aea4e17d99a7cacdbeed46a0096b10 \
"

DEPENDS += "\
    compiler-rt \
    libcxx \
    seccompiler-native \
    "

S = "${WORKDIR}/git"

SRC_URI = "git://github.com/firecracker-microvm/firecracker.git;protocol=https;nobranch=1"
SRCREV = "df69983f6d57abd5df986fea11de2952a1beae2c"

COMPATIBLE_MACHINE = "(-)"
COMPATIBLE_MACHINE:aarch64 = "(.*)"
COMPATIBLE_MACHINE:x86_64 = "(.*)"

do_compile[network] = "1"

SRC_URI += " \
    git://github.com/firecracker-microvm/kvm-bindings;lfs=0;nobranch=1;protocol=https;destsuffix=git/kvm-bindings>
    git://github.com/firecracker-microvm/micro-http;lfs=0;nobranch=1;protocol=https;destsuffix=git/micro-http;nam>
    crate://crates.io/aead/0.4.3 \
    crate://crates.io/aes/0.7.5 \
    crate://crates.io/aes-gcm/0.9.4 \
    crate://crates.io/atty/0.2.14 \
    crate://crates.io/autocfg/1.0.1 \
    crate://crates.io/base64/0.13.0 \
    crate://crates.io/bincode/1.3.3 \
    crate://crates.io/bindgen/0.59.2 \
    crate://crates.io/bitflags/1.3.2 \
    crate://crates.io/bstr/0.2.17 \
    crate://crates.io/bumpalo/3.9.1 \
    crate://crates.io/byteorder/1.4.3 \
    crate://crates.io/cast/0.2.7 \
    crate://crates.io/cc/1.0.73 \
    crate://crates.io/cexpr/0.6.0 \
    crate://crates.io/cfg-if/0.1.10 \
[...]

PR #3185, opened 2022-10-17
Deny undocumented unsafe blocks using clippy

Gilnaa:
## Changes
Test that all unsafe blocks must be documented; resulting in a test failure otherwise.
PR template is also modified to drop this bullet-point.

It looks like the pre-commit hook's `black` invocation changed `test_clippy.py` a tad bit; is that acceptable?

I've added some missing comments to satisfy clippy and modified existing comments to conform to clippy's formatting.
Test modules that used unsafe are excluded from the lint, as well as auto-generated bindgen code.

## Reason

Closes #3177

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

- [x] All commits in this PR are signed (`git commit -s`).
- [x] If a specific issue led to this PR, this PR closes the issue.
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] ~API changes follow the [Runbook for Firecracker API changes][2].~ (No API was changed)
- [x] ~User-facing changes are mentioned in `CHANGELOG.md`.~ (There are no user-facing changes)
- [x] All added/changed functionality is tested.

---

- [ ] This functionality can be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md

PR #3213, opened 2022-10-27
Fix cross compiling firecracker for different architecture

roypat:

## Changes

Remove the `--target` flag, meaning seccompiler-bin will always be built for the host architecture on which it is run. This fixes the issue encountered in https://github.com/firecracker-microvm/firecracker/pull/3185 and might also fix https://github.com/firecracker-microvm/firecracker/issues/3188 (although this has not been tested).

This PR also removes the global `target` directive in `.cargo/config`, which was added 5 years ago before firecracker supported any non-x86_64 architectures (I assume at least). This directive forced the `x86_64-unknown-linux-musl` target whenever no `--target` flag was provided (I assume this is why we originally passed `--target` everywhere). Without this the CI would fail on ARM as it would try to compile seccompiler-bin for x86_64.

## Reason

When cross compiling rust binaries, some components nevertheless have to be compiled for the host architecture. These are proc-macros and build scripts. In vmm(build.rs) we use cargo run to execute seccompiler-bin. However, we pass the `--target` flag, meaning seccompiler-bin is not built for the architecture it is being run on.

The reason we do not run into this issue when passing a `musl` target via the `--target` flag is that both musl and glibc are available on the host. 

## License Acceptance

## PR Checklist

---


pb8o (approved):

aghecenco (approved):

dianpopa:
The reason the global `--target` directive does not work on non-x86 is because cargo config did not have support for specifying a default target per arch. I do not think the approach of removing it all together is the right one since it might break client who expect to obtain a statically built (musl enabled) binary on x86.

roypat:
You are right, I did not consider that this would be a publicly facing change. I have introduced #3219 to revert the change of `cargo build` behavior while keeping cross-compilation working

PR #3216, opened 2022-10-27
Add clippy::cast_lossless

### LINKED ISSUE #3196: Use `clippy::cast_lossless`
Implement the `clippy::cast_lossless`.

StemCll:
## Changes
Adds `clippy::cast_lossless` warning.

## Reason

Closes #3196

## License Acceptance

## PR Checklist

---

### DIFF
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -4,7 +4,7 @@
 #![deny(missing_docs)]
 #![warn(clippy::ptr_as_ptr)]
 #![warn(clippy::undocumented_unsafe_blocks)]
-
+#![warn(clippy::cast_lossless)]
 //! Implements the interface for intercepting API requests, forwarding them to the VMM
 //! and responding to the user.
 //! It is constructed on top of an HTTP Server that uses Unix Domain Sockets and `EPOLL` to
--- a/src/io_uring/src/lib.rs
+++ b/src/io_uring/src/lib.rs
@@ -4,7 +4,7 @@
 #![deny(missing_docs)]
 #![warn(clippy::ptr_as_ptr)]
 #![warn(clippy::undocumented_unsafe_blocks)]
-
+#![warn(clippy::cast_lossless)]
 //! High-level interface over Linux io_uring.
 //!
 //! Aims to provide an easy-to-use interface, while making some Firecracker-specific simplifying
@@ -368,7 +368,7 @@ impl IoUring {
         let supported_opcodes: HashSet<u8> = probes
             .as_slice()
             .iter()
-            .filter(|op| ((op.flags as u32) & bindings::IO_URING_OP_SUPPORTED) != 0)
+            .filter(|op| ((u32::from(op.flags)) & bindings::IO_URING_OP_SUPPORTED) != 0)
             .map(|op| op.op)
             .collect();
 
--- a/src/arch/src/aarch64/fdt.rs
+++ b/src/arch/src/aarch64/fdt.rs
@@ -146,10 +146,10 @@ fn create_cpu_nodes(fdt: &mut FdtWriter, vcpu_mpidr: &[u64]) -> Result<()> {
                 fdt.property_u32(cache.type_.of_cache_size(), size as u32)?;
             }
             if let Some(line_size) = cache.line_size {
-                fdt.property_u32(cache.type_.of_cache_line_size(), line_size as u32)?;
+                fdt.property_u32(cache.type_.of_cache_line_size(), u32::from(line_size))?;
             }
             if let Some(number_of_sets) = cache.number_of_sets {
-                fdt.property_u32(cache.type_.of_cache_sets(), number_of_sets as u32)?;
+                fdt.property_u32(cache.type_.of_cache_sets(), u32::from(number_of_sets))?;
             }
         }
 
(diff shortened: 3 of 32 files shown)


StemCll:
Hey @JonathanWoollett-Light  I've fixed warnings for aarch64 CI

JonathanWoollett-Light (changes requested):
Great work!

On ARM you will need to update the code coverage values (see the outputs given in the failing tests (`82.41` for 5.10 and `79.60` for 4.14) then paste these value as the new targets under `tests/integration_tests/build/test_coverage.py`).

I see some changes to `tests/integration_tests/build/test_clippy.py` are included, it would best to avoid changes to this in this PR.

StemCll:
I've pushed the updated values in `test_coverage.py`. Regarding the `test_clippy.py` - as far as I can see, the black is executed in pre-commit script. What should I do in this case? Add this file without executing black? :smile:

JonathanWoollett-Light:
> the black is executed in pre-commit script. What should I do in this case? Add this file without executing black? 😄

Yep, using the `--no-verify` git commit argument should skip it.

It'd be worth opening a separate PR for the black reformat here.

StemCll:
Separate PR for black reformat is ready here - https://github.com/firecracker-microvm/firecracker/pull/3232

I've reverted the formatting here.

Also - should this clippy lint be added to bindgen? Same as `clippy::ptr_as_ptr`?

JonathanWoollett-Light:
> Also - should this clippy lint be added to bindgen? Same as clippy::ptr_as_ptr?

Yes.

StemCll:
Done :heavy_check_mark:

JonathanWoollett-Light:
Just the git message left.

https://buildkite.com/firecracker/firecracker-x86-64-intel-kernel-5-dot-10-ci/builds/1969#018438f9-9ffe-4ced-82ff-1739fbe50f26:

> Title exceeds max length (86>50): "Update current code coverage for ARM in tests/integration_tests/build/test_coverage.py"

StemCll:
done :heavy_check_mark:

JonathanWoollett-Light (dismissed):

src/api_server/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 
 #![deny(missing_docs)]
 #![warn(clippy::ptr_as_ptr)]
+#![warn(clippy::cast_lossless)]
```
dianpopa, line 6: Warning on this error from inside each crate is polluting the code too much (even more so if we decide to warn on multiple cast related errors).
Apart from polluting the code, this method of putting the warn cfg inside each crate is not enforcing it on future crates, right?

I see you can directly specify to the `cargo clippy` command the config we want to warn on. Can you give it a try?

dianpopa (changes requested):

src/api_server/src/lib.rs
JonathanWoollett-Light: I think for testing this approach makes sense, however in development it can become awkward to need to enter many command line arguments rather than just run `cargo clippy`. It might be worth including the command line arguments in the testing, but I would still argue for these tags in the code.

What do you think?

dianpopa: Let s try out with a clippy.toml, at least the ones we are interested in are there: https://rust-lang.github.io/rust-clippy/master/index.html.

JonathanWoollett-Light: @StemCll In this case when using a `clippy.toml` file it may make sense to open a new issue and pr on the topic of replacing all the crate level `#![warn()]` statements with `clippy.toml`.

StemCll: Unfortunately `clippy.toml` is not a place where one can deny / allow lints ([source](https://github.com/rust-lang/rust-clippy#configure-the-behavior-of-some-lints). Here's an issue relating to the exact case as we have here -> https://github.com/rust-lang/rust-clippy/issues/1313

Reading through your conversation, I would keep warns in one place - allow only specific functions / mods where needed. One of the reasons here is that there are more lints to come (waiting for your comments [here](https://github.com/firecracker-microvm/firecracker/pull/3237) :smile: ) and in each lib.rs file we will have more and more warns. It's not bad, but it would be easier to maintain to keep all warns in one place.

In terms of development - I don't know how you work with code, but I'd just extend the capabilities of the `devtool` script to run clippy check without Pytest. Pytest output is less readable for me, thus I prefer to run clippy without it.

dianpopa: another alternative is to specify them in the clippy command; I posted an example of this here: https://github.com/firecracker-microvm/firecracker/pull/3258

StemCll: Alright, this's some solution. What should I do then?

dianpopa: hi! I synced offline with @JonathanWoollett-Light and decided to go with the current form (of specifying clippy in source code) for the time being!

You can go ahead and rebase your PR, thanks!

JonathanWoollett-Light: @dianpopa consider that for the moment it is fair to go with an implementation here using `#[warn(clippy::)]` rather than the command line.

So if you can rebase with this approach that be would great.

StemCll: done :heavy_check_mark:

JonathanWoollett-Light:
You can rename the 2nd commit ti simply `Updated code coverage for Arm`.

StemCll:
@JonathanWoollett-Light done :heavy_check_mark:

JonathanWoollett-Light:
Getting failures for Arm coverage being higher, the 2nd commit needs to be adjusted to match (`79.66` and `82.47` need to be set on Arm for 4.14 and 5.10 respectively).

JonathanWoollett-Light (dismissed):

src/io_uring/src/lib.rs
```diff
@@ -368,7 +368,8 @@ impl IoUring {
         let supported_opcodes: HashSet<u8> = probes
             .as_slice()
             .iter()
-            .filter(|op| ((op.flags as u32) & bindings::IO_URING_OP_SUPPORTED) != 0)
+            //.filter(|op| ((op.flags as u32) & bindings::IO_URING_OP_SUPPORTED) != 0)
```
dianpopa, line 371: remove commented code

StemCll: done :heavy_check_mark:

JonathanWoollett-Light (approved):

dianpopa (approved):

PR #3219, opened 2022-10-28
Revert default behavior of cargo build

roypat:

## Changes

This commit restored the old behavior while still fixing cross-compilation. For this, instead of relying on the default target matching the compilation host, we instead explicitly read the host target from the HOST variable that cargo sets for build scripts.

## Reason

My PR #3213 carelessly changed the default behavior of cargo build.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,6 @@
 [build]
 target-dir = "build/cargo_target"
+target = "x86_64-unknown-linux-musl"
 
 [net]
 git-fetch-with-cli = true
--- a/src/vmm/build.rs
+++ b/src/vmm/build.rs
@@ -51,6 +51,11 @@ fn main() {
 
 // Run seccompiler with the given arguments.
 fn run_seccompiler_bin(json_path: &str, out_path: &str) {
+    // We have a global `target` directive in our .cargo/config file specifying x86_64 architecture.
+    // However, seccompiler-bin has to be compiled for the host architecture. Without this, cargo
+    // would produce a x86_64 binary on aarch64 host, causing this compilation step to fail as such
+    // a binary would not be executable.
+    let host_arch = env::var("HOST").expect("Could not determine compilation host");
     let target_arch = env::var("CARGO_CFG_TARGET_ARCH").expect("Missing target arch.");
 
     // Command for running seccompiler-bin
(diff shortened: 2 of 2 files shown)


JonathanWoollett-Light (approved):

JonathanWoollett-Light (commented):
w

bchalios:
Could we detect in `build.rs` the architecture and unless a `--target` is defined, we set `--target ${ARCH}-unknown-linux-musl`?

roypat:
> Could we detect in `build.rs` the architecture and unless a `--target` is defined, we set `--target ${ARCH}-unknown-linux-musl`?

cargo generally ignored whatever `target` you specify for build scripts and proc-macro crates, and instead uses the host triple, because they have to be executed on the host machine (they are steps of the compilations process and compilation cannot successfully finish unless they run). The only reason we run into issues here is because we explicitly call `cargo run` inside a build-script, which breaks out of cargos assurances. If `seccompiler-bin` were to be inlined into the build-script, it would be compiled for the compilation host triple. 

I am not entirely sure what you mean by "unless a `--target` is specified", since we always specify a target. However, we can ensure that seccompiler-bin is compiled with the `musl` abi for the relevant architecture, instead of the host triple. Is this what you mean?

roypat:
> Can we also check if cargo config now supports specifying a default target per arch?

Babis and me just checked this. It seems to not be the case. While there is the option to [pass an array of triples](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget) to the `build.target` configuration option, this apparently causes cargo to build for all of them simultaneously (which in the case of `cargo run` results in the error "error: only one `--target` argument is supported"), instead of picking the most appropriate one. There does not seem to be the option to define a compilation target based on what the compilation host is (e.g a way to say "if we're on x86_64 then use `x86_64-unknown-linux-musl`, if `aarch64`...)

JonathanWoollett-Light (approved):
I think this is a good solution at the moment.

However non-ideal in the long term. the issue https://github.com/firecracker-microvm/firecracker/issues/3221 may be a good place to propose and discuss removing the build target restriction (`target = "x86_64-unknown-linux-musl"`) altogether.

luminitavoicu (approved):

PR #3235, opened 2022-11-03
CI: fix 5_inc_snapshots intermittent failure

dianpopa:
## Reason

We have stumbled upon an OOM scenario in one of our integration tests (5_inc and 5_full snapshots) due to the fact that we spawn FIO in the guest in every create/resume snapshot sequence. Doing it upon each invocation of snap/resume might possibly lead to multiple FIO instances on the guest (the FIO processes from the previous guest might get snapshotted). This can result in an OOM and thus lead to an intermittent failure in the CI.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/functional/test_snapshot_basic.py
+++ b/tests/integration_tests/functional/test_snapshot_basic.py
@@ -12,7 +12,10 @@ from conftest import _test_images_s3_bucket
 from framework.artifacts import ArtifactCollection, ArtifactSet
 from framework.builder import MicrovmBuilder, SnapshotBuilder, SnapshotType
 from framework.matrix import TestMatrix, TestContext
-from framework.utils import wait_process_termination
+from framework.utils import (
+    wait_process_termination,
+    check_filesystem,
+)
 from framework.utils_vsock import (
     make_blob,
     check_host_connections,
@@ -25,17 +28,6 @@ from framework.utils_vsock import (
 
 import host_tools.network as net_tools  # pylint: disable=import-error
 import host_tools.drive as drive_tools
-
-
-def _guest_run_fio_iteration(ssh_connection, iteration):
-    fio = """fio --filename=/dev/vda --direct=1 --rw=randread --bs=4k \
-        --ioengine=libaio --iodepth=16 --runtime=10 --numjobs=4 --time_based \
-        --group_reporting --name=iops-test-job --eta-newline=1 --readonly \
-        --output /tmp/fio{} > /dev/null &""".format(
-        iteration
-    )
-    exit_code, _, _ = ssh_connection.execute_command(fio)
-    assert exit_code == 0
 
 
 def _get_guest_drive_size(ssh_connection, guest_dev_name="/dev/vdb"):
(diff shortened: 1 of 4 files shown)


tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -124,9 +110,6 @@ def _test_seq_snapshots(context):
         path = os.path.join(microvm.jailer.chroot_path(), VSOCK_UDS_PATH)
         check_host_connections(microvm, path, blob_path, blob_hash)
 
-        # Start a new instance of fio on each iteration.
-        _guest_run_fio_iteration(ssh_connection, i)
```
bchalios, line 128: should we completely remove it? I guess the point of this being here is to test block functionality after restoring from a snapshot. Could we not just make sure the fio process is finished before getting the next snapshot?

tests/integration_tests/performance/test_snapshot_perf.py
```diff
@@ -42,21 +43,32 @@
 # TODO: Update baseline values after fix.
 CREATE_LATENCY_BASELINES = {
```
zulinx86, line 44: Need to change L167
```
-                CREATE_LATENCY_BASELINES[platform.machine()][vm_type][
+                CREATE_LATENCY_BASELINES[platform.machine()][get_instance_type()][vm_type][
```

tests/integration_tests/performance/test_snapshot_perf.py
```diff
@@ -42,21 +43,32 @@
...
-        "2vcpu_512mb.json": {
-            "FULL": {"target": 280},
-            "DIFF": {"target": 75},
+        "m5d.metal": {
+            "2vcpu_256mb.json": {"FULL": {"target": 180}, "DIFF": {"target": 70}},
+            "2vcpu_512mb.json": {
+                "FULL": {"target": 280},
+                "DIFF": {"target": 75},
+            },
+        },
```
zulinx86, line 52: Aren't values for m6i.metal also needed?
```suggestion
        },
        {
            "2vcpu_256mb.json": {"FULL": {"target": 180}, "DIFF": {"target": 70}},
            "2vcpu_512mb.json": {
                "FULL": {"target": 280},
                "DIFF": {"target": 75},
            },
        },
```

zulinx86 (changes requested):

tests/integration_tests/performance/test_snapshot_perf.py
dianpopa: I added a TODO to also gather actual numbers for m6i.
I made this change so that I am able to modify changes per processor type.
It is not in the scope of this PR to update m6i values.

dianpopa:
@bchalios @zulinx86 please take another look.

zulinx86 (approved):

tests/integration_tests/functional/test_snapshot_basic.py
bchalios: @dianpopa any comments on this?

bchalios (approved):

tests/integration_tests/functional/test_snapshot_basic.py
dianpopa: The reason I deleted it completely is because I see this as a superfluous/wrong check; given that FIO is run in the background and we do not check for its status, if the block device malfunctions we would not receive a clear error for it; it would manifest (probably) in hangs;

Also, my assumption is that this test was introduced to check for file corruptions (that is why is run on the root device).

Now, the test in its current form runs for 10s; I do not think is a good idea to run it for that long;
Should be enough to just read the root device file once, right?

PR #3238, opened 2022-11-04
CI: generate PR pipeline dynamically

pb8o:
## Changes

- Made a script that can generate the build script we run for every PR
- Made the license check a bit more robust by finding the copyright line instead of expecting it to be at an exact location.

## Reason

We have more control to decide what steps run and on what agents. Also there is less clicking when trying to find a failed job. 

## License Acceptance

## PR Checklist

- [X] New `unsafe` code is documented.

---

### DIFF
--- a/.buildkite/pipeline_pr.py
+++ b/.buildkite/pipeline_pr.py
@@ -1,0 +1,105 @@
+#!/usr/bin/env python3
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Generate Buildkite pipelines dynamically"""
+
+import json
+
+INSTANCES = [
+    "m5d.metal",
+    "m6i.metal",
+    "m6a.metal",
+    "m6gd.metal",
+]
+
+KERNELS = ["4.14", "5.10"]
+
+
+def group(group_name, command, agent_tags=None, priority=0, timeout=30):
+    """
+    Generate a group step with specified parameters, for each instance+kernel
+    combination
+
+    https://buildkite.com/docs/pipelines/group-step
+    """
+    if agent_tags is None:
+        agent_tags = []
+    # Use the 1st character of the group name (should be an emoji)
+    label1 = group_name[0]
+    steps = []
+    for instance in INSTANCES:
+        for kv in KERNELS:
+            agents = [
+                f"type={instance}",
+                f"kv={kv}",
+            ]
+            agents.extend(agent_tags)
+            step = {
+                "command": command,
+                "label": f"{label1} {instance} kv={kv}",
+                "priority": priority,
+                "timeout": timeout,
+                "agents": agents,
+            }
+            steps.append(step)
+
+    return {"group": group_name, "steps": steps}
+
+
+step_block_unless_maintainer = {
+    "block": "Waiting for approval to run",
+    "if": '(build.creator.teams includes "prod-compute-capsule") == false',
+}
+
+step_style = {
+    "command": "./tools/devtool -y test -- ../tests/integration_tests/style/",
+    "label": "🪶 Style",
+    # no agent tags, it doesn't matter where this runs
+}
+
+build_grp = group(
+    "📦 Build",
+    "./tools/devtool -y test -- ../tests/integration_tests/build/",
+    priority=1,
+)
+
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+"""
```
zulinx86, line 6: nit: Shall we put some comments here for this script?

zulinx86 (commented):
LGTM! Commented just one nit.

zulinx86 (dismissed):

.buildkite/pipeline.py
```diff
@@ -0,0 +1,108 @@
...
+    "./tools/devtool -y test -- ../tests/integration_tests/performance/",
+    priority=1,
+    agent_tags=["ag=1"],
+)
+
+pipeline = {
+    "agents": {"queue": "default"},
+    "steps": [
+        step_block_unless_maintainer,
+        step_style,
```
bchalios, line 99: would this make the style step to be present in all pipelines? We only need it in one. Currently, it runs in Intel/5.10

bchalios: *m5d/5.10

.buildkite/pipeline.py
pb8o: yep, the CI complained about it.

.buildkite/pipeline.py
pb8o: No, the style is just a single step, not a group of steps crossed with instances and kernels.

bchalios (dismissed):

.buildkite/pipeline.py
```diff
@@ -0,0 +1,108 @@
...
+import json
+
+INSTANCES = [
+    "m5d.metal",
+    "m6i.metal",
+    "m6a.metal",
+    "m6gd.metal",
+]
+
+KERNELS = ["4.14", "5.10"]
```
dianpopa, line 16: You can use here the SUPPORTED_KERNELS variable from  framework.defs.

.buildkite/pipeline.py
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env python3
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
dianpopa, line 2: If we also wanted to do the same for the performance pipelines, we would need to create another pipeline file?
Maybe would be worth to rename this to ci-pipeline.py?

pb8o: Yeah, not sure how to best do that at the moment... We could have different arguments.. `pipeline.py --pr`, `pipeline.py --perf`, or as you say a different script, using some common functionality.

.buildkite/pipeline.py
pb8o: I think it may be too much trouble for what's it's worth. Would have to set up `PYTHONPATH` etc, and we may want to use the constants for different purposes.

.buildkite/pipeline.py
pb8o: PyLint complains about about `pipeline-pr.py` .. will do `pipeline_pr.py` to avoid the error.

zulinx86 (dismissed):

bchalios (dismissed):

.buildkite/pipeline_pr.py
```diff
@@ -0,0 +1,105 @@
...
+)
+
+functional_1_grp = group(
+    "⚙️ Functional [a-n]",
+    "./tools/devtool -y test -- `cd tests; ls integration_tests/functional/test_[a-n]*.py`",
+    priority=1,
+)
+
+functional_2_grp = group(
+    "⚙️ Functional [o-z]",
```
roypat, line 74: see above

dianpopa (dismissed):
nit: commits are unverified

pb8o:
> nit: commits are unverified

Weird, they are signed on my workspace. Could it be because I rebased here with the UI?

luminitavoicu (approved):

zulinx86 (approved):

bchalios (approved):

PR #3241, opened 2022-11-06
Custom minimal event manager

JonathanWoollett-Light:
## Changes

Replaces [`rust-vmm/event-manager`](https://github.com/rust-vmm/event-manager) dependency with a custom minimal solution.

## Reason

The implementation of [`rust-vmm/event-manager`](https://github.com/rust-vmm/event-manager) is non-ideal for Firecracker's narrow and sparse usage. While a large number of code paths rely upon it we rarely interact with the library directly (overall there are few `add_subscriber` or `impl EventSubscriber` blocks), in this sense it motivates the question of whether it is needed as an external dependency.

[`rust-vmm/event-manager`](https://github.com/rust-vmm/event-manager) also leans towards a more object orient and abstract approach, this results in large `<T as EventSubscriber>::process` functions which often handle multiple different file descriptors leading to unneeded `if`s and `match`s within Firecracker. By directly associating functions with events rather than using trait implementations this also reduces indirection and aids readability.

An old implementation like:
```rust
fn process(&mut self, event: Events, ops: &mut EventOps) {
    let source = event.fd();
    let event_set = event.event_set();
    let supported_events = EventSet::IN;

    if !supported_events.contains(event_set) {
        warn!(
            "Received unknown event: {:?} from source: {:?}",
            event_set, source
        );
        return;
    }

    if self.is_activated() {
        let virtq_inflate_ev_fd = self.queue_evts[INFLATE_INDEX].as_raw_fd();
        let virtq_deflate_ev_fd = self.queue_evts[DEFLATE_INDEX].as_raw_fd();
        let virtq_stats_ev_fd = self.queue_evts[STATS_INDEX].as_raw_fd();
        let stats_timer_fd = self.stats_timer.as_raw_fd();
        let activate_fd = self.activate_evt.as_raw_fd();

        // Looks better than C style if/else if/else.
        match source {
            _ if source == virtq_inflate_ev_fd => self
                .process_inflate_queue_event()
                .unwrap_or_else(report_balloon_event_fail),
            _ if source == virtq_deflate_ev_fd => self
                .process_deflate_queue_event()
                .unwrap_or_else(report_balloon_event_fail),
            _ if source == virtq_stats_ev_fd => self
                .process_stats_queue_event()
                .unwrap_or_else(report_balloon_event_fail),
            _ if source == stats_timer_fd => self
                .process_stats_timer_event()
                .unwrap_or_else(report_balloon_event_fail),
            _ if activate_fd == source => self.process_activate_event(ops),
            _ => {
                warn!("Balloon: Spurious event received: {:?}", source);
            }
        };
    } else {
        warn!(
            "Balloon: The device is not yet activated. Spurious event received: {:?}",
            source
        );
    }
}
```
Becomes multiple simpler functions (one for each file descriptor) like:
`balloon.activate_evt`
```rust
Box::new(move |event_manager: &mut EventManager, _: EventSet| {
    if balloon_clone.lock().unwrap().is_activated() {
        Balloon::process_activate_event(balloon_clone, event_manager);
    } else {
        warn!("Balloon: The device is not yet activated. Spurious event received.");
    }
}),
```
`balloon.queue_evts[STATS_INDEX]`
```rust
Box::new(move |_: &mut EventManager, _: EventSet| {
    let mut b = balloon_clone.lock().unwrap();
    if b.is_activated() {
        b.process_stats_queue_event()
            .unwrap_or_else(report_balloon_event_fail)
    } else {
        warn!(
            "Balloon: The device is not yet activated. Spurious event \
                received."
        );
    }
}),
```
`balloon.queue_evts[DEFLATE_INDEX]`
```rust
Box::new(move |_: &mut EventManager, _: EventSet| {
    let mut b = balloon_clone.lock().unwrap();
    if b.is_activated() {
        b.process_deflate_queue_event()
            .unwrap_or_else(report_balloon_event_fail)
    } else {
        warn!("Balloon: The device is not yet activated. Spurious event received.");
    }
}),
```
`balloon.queue_evts[INFLATE_INDEX]`
```rust
Box::new(move |_: &mut EventManager, _: EventSet| {
    let mut b = balloon_clone.lock().unwrap();
    if b.is_activated() {
        b.process_inflate_queue_event()
            .unwrap_or_else(report_balloon_event_fail)
    } else {
        warn!("Balloon: The device is not yet activated. Spurious event received.");
    }
}),
```
This provides a more readable implementation as events can be more easily followed. This is also more efficient implementation as this behavior is closer to that of the underlying `epoll` and requires less operations on top.

[`rust-vmm/event-manager`](https://github.com/rust-vmm/event-manager):
```
--------------------------------------------------------------------------------
 Language             Files        Lines        Blank      Comment         Code
--------------------------------------------------------------------------------
 Rust                    15         2435          340          583         1512
 Markdown                 5          725          148            0          577
 Python                   6          522           93           68          361
 Toml                     2           37            7            0           30
 JSON                     3           15            0            0           15
--------------------------------------------------------------------------------
 Total                   31         3734          588          651         2495
--------------------------------------------------------------------------------
```
`firecracker/src/event-manager`:
```
--------------------------------------------------------------------------------
 Language             Files        Lines        Blank      Comment         Code
--------------------------------------------------------------------------------
 Rust                     1          204           12           45          147
 Toml                     1            9            2            1            6
--------------------------------------------------------------------------------
 Total                    2          213           14           46          153
--------------------------------------------------------------------------------
```

## License Acceptance

## PR Checklist

---


src/devices/src/legacy/serial.rs
```diff
@@ -96,29 +96,7 @@ pub struct SerialWrapper<T: Trigger, EV: SerialEvents, W: Write> {
     pub input: Option<Box<dyn ReadableFd + Send>>,
 }
 
-impl<W: Write> SerialWrapper<EventFdTrigger, SerialEventsWrapper, W> {
-    fn handle_ewouldblock(&self, ops: &mut EventOps) {
```
acatangiu, line 100: Make sure you're not throwing away corner-case handling. Some of this code is a direct result of operational issues, study its history and make sure you don't end up repeating it. There be dragons here :dragon_face:

acatangiu (commented):
While I agree `rust-vmm/event-manager` is over-engineered and could be simpler, I don't think this PR is an overall improvement for `Firecracker` for the following reasons (in order of importance):
1. The new design **forces** inner objects (devices for example) to be wrapped in `Arc<Mutex<>>` which is bad for two reasons:
    - it may negatively impact performance at scale (run thousands of active firecrackers and see) because of extra interlocked memory operations,
    - cannot use inner object from multiple threads without lock contention - potentially limiting in future features;
2. This new design doesn't actually dispose of dynamic dispatching, it just replaces it with explicit look-up code to call custom closures, ergo I think it's actually slower even without taking the extra required mutex operation into consideration,
3.  Even if the minimal `event-manager` proposed here is so much smaller in terms of LoC compared to the `rust-vmm` version, overall this PR **adds** many LoC to `firecracker`, to critical emulation code: `+998 −572`. Unless there are safety concerns with (the battle-tested) existing dependency, adding more attack surface and code to maintain to firecracker might be counter-productive.

JonathanWoollett-Light:
> 1. The new design **forces** inner objects (devices for example) to be wrapped in `Arc<Mutex<>>` which is bad for two reasons:
>     - it may negatively impact performance at scale (run thousands of active firecrackers and see) because of extra interlocked memory operations,
>     - cannot use inner object from multiple threads without lock contention - potentially limiting in future features;

The new design doesn't introduce any new restrictions here, it just makes them clearer.

Consider that the previous pattern of:
```rust
let x = Arc::new(Mutex::new(...));
event_manager.add_subscriber(x);
```
Has the same lock as in the new implementation.
```rust
let x = Arc::new(Mutex::new(...));
event_manager.add(...,|..|{
    x
});
```
In both implementations you can change `x` to any type (except the first requires the type implement the subscriber trait).

There is no functional or performance impact here.

> 2. This new design doesn't actually dispose of dynamic dispatching, it just replaces it with explicit look-up code to call custom closures, ergo I think it's actually slower even without taking the extra required mutex operation into consideration,

Of my previous point there is not an extra required mutex. On your main point here the goal was not to dispose of dynamic dispatch. On lookup, consider that `rust-vmm/event-manager` passes through 2 hash maps to call the `procedure` function while this only passes through 1.

> 3. Even if the minimal `event-manager` proposed here is so much smaller in terms of LoC compared to the `rust-vmm` version, overall this PR **adds** many LoC to `firecracker`, to critical emulation code: `+998 −572`. Unless there are safety concerns with (the battle-tested) existing dependency, adding more attack surface and code to maintain to firecracker might be counter-productive.

In total overall LoC involved in the process are reduced, although this is a bad metric for sure. In my view this improves readability of the code within Firecracker even if it adds lines.

On a final performance point I would note that in every `procedure` function previously we do multiple comparisons for safety checking and to detect which signal occurred which can be skipped with the new implementation.

acatangiu:
You are right that even if `rust-vmm/event-manager` API doesn't **require** `Arc<Mutex<x>>`, the usage pattern in `firecracker` is to wrap devices in `Arc<Mutex<>>` for other reasons anyway.

In this case I agree that the main concern:
> The new design forces inner objects (devices for example) to be wrapped in Arc<Mutex<>>

has no impact in current code :+1: 

Regarding performance, it would be best to see some benchmark results and go for a data-driven decision.

bchalios:
This is in the hot path!

This comment:

> Regarding performance, it would be best to see some benchmark results and go for a data-driven decision.

is spot on.

If we decide this is worth merging we should first run a set of performance benchmarks which also include high-density scenarios.

roypat:
If we decide to replace event manager, it will need prior investigation before we can move ahead with any implementation, so closing this PR.

PR #3247, opened 2022-11-07
seccompiler: remove some unneeded unsafe blocks

petrosagg:
## Changes

Any `AsRef<[u8]>` value can be converted into something implementing `Read` using `std::io::Cursor` so the json parsing can be performed directly on the static strings.

## Reason

Remove unneeded unsafe blocks

## License Acceptance

## PR Checklist

---


src/seccompiler/src/seccompiler_bin.rs
```diff
@@ -532,70 +529,56 @@ mod tests {
...
-            assert!(parse_json(&mut json_input.as_ref()).is_err());
+            assert!(parse_json(std::io::empty()).is_err());
 
             // not json
-            let mut json_input = "hjkln".to_string();
-            let json_input = unsafe { json_input.as_bytes_mut() };
-
-            assert!(parse_json(&mut json_input.as_ref()).is_err());
+            let json_input = Cursor::new("hjkln");
+            assert!(parse_json(json_input).is_err());
```
roypat, line 541:
```suggestion
            let json_input = "hjkln";
            assert!(parse_json(json_input.as_ref()).is_err());
```

roypat (changes requested):
Hi Petros,
Very nice catch, thanks for contributing a fix! It turns out that even a cursor is overkill here, as we have [`impl Read for &'_ [u8]`](https://doc.rust-lang.org/std/io/trait.Read.html#impl-Read-for-%26%5Bu8%5D). You can use that as below

petrosagg:
@roypat good point! changed all of them

src/seccompiler/src/seccompiler_bin.rs
```diff
@@ -231,9 +231,8 @@ mod tests {
     use crate::backend::SeccompCmpOp::{Le, *};
     use crate::backend::{SeccompAction, SeccompCondition as Cond, TargetArch, TargetArchError};
 
-    // test helper for generating correct JSON input data
-    fn get_correct_json_input() -> String {
-        r#"
+    // Correct JSON input data
+    static CORRECT_JSON_INPUT: &str = r#"
```
JonathanWoollett-Light, line 235:
```suggestion
    static CORRECT_JSON_INPUT: &str = br#"
```
Since we only use `CORRECT_JSON_INPUT` with `CORRECT_JSON_INPUT.as_bytes()` we can use a bytes string such that we can replace `assert_eq!(CORRECT_JSON_INPUT.as_bytes(), /* ... */);` with `assert_eq!(CORRECT_JSON_INPUT, /* ... */);`

petrosagg: The compiler already embeds a const string in the binary so the two are equivalent in function. But if the type is set to a const array of some length `[u8; N]` not only you don't gain anything but instead also make it annoying to edit the string as you'll have to constantly update the length there.

JonathanWoollett-Light: The compiler will tell you the length required for byte array, this is a small inconvenience. But I don't have strong opinion here.

If not `const`, I would suggest:

```
    static CORRECT_JSON_INPUT: &[u8]= br#"
```

petrosagg: Not strong opinions either. My argument is that semantically the correct type for that value is a `&str`. The fact that it is being parsed with a particular parser that requires an `impl Read` is incidental. You can imagine some other parsing function (e.g https://docs.rs/serde_json/latest/serde_json/fn.from_str.html) that doesn't require that. The type of value should ideally be dictated by its structure and not its use.

petrosagg: What?!, it does work! But `&str` is not Read! https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=65e51ba094f4c61ac25c28a7f67c7d0a

The compiler must be doing some magic here.

You couldn't implement `Read` on `&str` because it must always be a valid utf8 byte sequence. The `Read` impl on `&[u8]` relies on the fact that `Read::read` has a `&mut self` receiver (so `&mut &[u8]` in this case) and so it can take a subslice and replace the original `self` with the smaller one. https://doc.rust-lang.org/src/std/io/impls.rs.html#247

But you can't do that with `&str`, if you take say one byte off from the beginning the remaining sequence might not be valid utf8 so you'd have UB.

petrosagg: Wait no, it doesn't. I checked with `cargo check` which doesn't include the tests, oops. Ok the compiler isn't doing weird magic. Phew!

petrosagg: CI is almost done so I would merge as is if that's ok :)

roypat (approved):

bchalios (approved):

PR #3248, opened 2022-11-08
Use caret instead of comparison requirements

### LINKED ISSUE #3226: [Bug] `cargo update` results in un-buildable workspace
# Describe the bug

Running `cargo update` or deleting `Cargo.lock` results in an un-buildable workspace in the sense that `cargo build` fails with compilation errors due to incompatible dependencies. 

## To Reproduce

1. Run `cargo update` or delete `Cargo.lock`
2. Run `cargo build` (or `./tools/devtools build`)

## Expected behaviour

The dependencies specified in the various `Cargo.toml` files are compatible and the build succeeds

## Additional context

This is due to us using `>=` instead of `=` to specify dependency versions for crates that do not follow semver (e.g. crates prior to version one, especially `rust-vmm`). For example, `vmm-sys-utils=0.10.0` does not export `ioctl_expr` anymore, but since `utils/Cargo.toml` specifies `vmm-sys-utils>=0.8.0`, this version is downloaded by `cargo update`.

To fix this, we need to go through `Cargo.lock` and lock our dependencies to versions that are proven to compile.

## Checks

roypat:

## Changes

This commit puts all our comparisons requirements to the versions Cargo.lock currently specifies.

Closes #3226 

## Reason

When regenerating `Cargo.lock` (or when adding firecracker as a library component), the comparison requirements in Cargo.toml files (`>=`) can cause silent major version upgrades, causing firecracker to fail building due to either incompatible changes, or because of using multiple versions of the same dependency. To see what I mean with the last point, consider the following scenario:

We have 2 dependencies outside our org, lets call them `A` and `B`. We specify `A >= 1.0.0` and `B>=1.0.0`. Assume `B` has a dependency `A=^1.0.0` and exposes some public API taking structs defined by `A`. Also assume `A` pushes a backward incompatible version `2.0.0`. If we now run `cargo update` we update to `A=2.0.0` but `B` keeps using `A=1.0.0`. After this, we will depend on two separate versions of `A`, and if we try passing `A=2.0.0` objects to `B`, we will get a compiler error.  

Note that a similar issue can happen even if we only use caret requirements, but one of our dependencies uses comparison requirements. This is outside of our control, and will require manual updates of dependencies. To catch this I propose a nightly CI that runs `cargo update && cargo build` and notifies us of dependency issues.

## License Acceptance

## PR Checklist

- [x] If a specific issue led to this PR, this PR closes the issue

---


roypat:
Draft because I will add a integration test running on each PR to deny any newly introduced comparision requirements. However, as this PR does not actually fix https://github.com/firecracker-microvm/firecracker/issues/3226 I will work on that fix first (which involved upgrading `vmm-sys-util` to `0.10.0`

JonathanWoollett-Light:
From https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements

> **Caret requirements** are an alternative syntax for the default strategy, ^1.2.3 is exactly equivalent to 1.2.3.

So it would seem simpler to just use `=` to produce the same behavior?

roypat:
I have added an integration test that verifies that no dependencies are added with comparison requirements, however the placement of the test is a bit awkward: 
Currently it lives in `src/firecracker`, but it checks the `Cargo.toml` files in the entire workspace. This is due to there being no option to have a rust test at the workspace level. I went with rust for this test because of the ease of parsing Cargo.toml files with rust vs python (existence of a crate for it) to keep the code short. I talked to @JonathanWoollett-Light about potential alternatives, but we could not come up with a solution that has no drawbacks. The cleanest would be to have the integration test in each individual crate (and have the test only check the Cargo.toml of the crate it is in), but this would result in a lot of code duplication. Alternatively, we could add a [special crate just for unit tests](https://stackoverflow.com/questions/71460402/use-rust-cargo-to-run-tests-in-workspace-root) to the workspace.

src/firecracker/tests/verify_dependencies.rs
```diff
@@ -0,0 +1,70 @@
...
+                Dependency::Simple(version) => Some((name, version)),
+                Dependency::Detailed(DependencyDetail {
+                    version: Some(version),
+                    ..
+                }) => Some((name, version)),
+                _ => None, // doesn't tell us about version, e.g. if a direct git dependency is specified
+            }
+        })
+        .filter(|(_, version)| version.starts_with(">"))
+}
```
JonathanWoollett-Light, line 70:
```suggestion
fn violating_dependencies_of_depsset(depsset: DepsSet) -> impl Iterator<Item = (String, String)> {
    // Matches version numbers.
    let re = regex::Regex::new(r"[0-9]*\.[0-9]*\.[0-9]*");
    depsset.into_iter().filter_map(|(name, dependency)| match dependency {
            Dependency::Simple(version) | Dependency::Detailed(DependencyDetail { version: Some(version), ..}) 
                if !re.is_match(version) => Some((name, version)),
            _ => None, // doesn't tell us about version, e.g. if a direct git dependency is specified
    })
}
```

src/firecracker/tests/verify_dependencies.rs
```diff
@@ -0,0 +1,70 @@
...
+    let mut violating = HashMap::new();
+
+    violating.extend(violating_dependencies_of_depsset(manifest.dependencies));
+    violating.extend(violating_dependencies_of_depsset(manifest.dev_dependencies));
+    violating.extend(violating_dependencies_of_depsset(
+        manifest.build_dependencies,
+    ));
+
+    violating
+}
```
JonathanWoollett-Light, line 51:
```suggestion
fn violating_dependencies_of_cargo_toml(path: impl AsRef<Path>) -> HashMap<String, String> {
    let manifest = Manifest::from_path(path).unwrap();
    violating_dependencies_of_depsset(manifest.dependencies)
        .chain(violating_dependencies_of_depsset(manifest.dev_dependencies))
        .chain(violating_dependencies_of_depsset(manifest.build_dependencies))
        .collect()
}
```

src/firecracker/tests/verify_dependencies.rs
roypat: I don't want to restrain non-comparison requirements, which is why I specifically match for those. Although I agree with your suggestions for simplifying the iterator :)

src/api_server/Cargo.toml
```diff
@@ -6,10 +6,10 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
-libc = ">=0.2.39"
-serde = { version = ">=1.0.27", features = ["derive"] }
-serde_derive = ">=1.0.27"
-serde_json = ">=1.0.9"
+libc = "0.2.117"
```
pb8o, line 9: Should we just specify `major.minor` here and below to avoid churn?

pb8o: Though it may not work if we /really/ require `>=0.2.39`. Wouln't work for all cases.

roypat: The difference between specifying `libc = "0.2"` and `libc = "0.2.117"` is that the latter disallows patch versions less than 117, whereas the former allows all patch releases of 0.2. 

I do not know if Firecracker requires any minimal patch releases (e.g. something like performance improvements could be done in patch releases, and firecracker might implicitly depend on those through our performance tests), the versions specified in this PR are pulled from `Cargo.lock` (e.g. are exactly the versions that `cargo build` was already using prior to this PR). 

I agree that the currently specified patch versions are somewhat arbitrary (in the end, they are just the newest patch versions that happened to be available whenever `Cargo.lock` got last updated). I'd be in favor of removing them for crates where firecracker does not depend on a specific minimal patch version, but since I do not know for which crates this is the case, I think that change if outside the scope of this PR.

JonathanWoollett-Light (changes requested):

src/firecracker/tests/verify_dependencies.rs
```diff
@@ -0,0 +1,76 @@
...
+///
+/// The iterator produces tuples of the form (violating dependency, specified version)
+fn violating_dependencies_of_depsset(depsset: DepsSet) -> impl Iterator<Item = (String, String)> {
+    depsset.into_iter().filter_map(|(name, dependency)| {
+        match dependency {
+            Dependency::Simple(version) // dependencies specified as `libc = "0.2.117"`
+            | Dependency::Detailed(DependencyDetail {  // dependencies specified as `libc = {version = "0.2.117",...}
+                version: Some(version),
+                ..
+            }) if version.starts_with('>') => Some((name, version)),
```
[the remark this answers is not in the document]
roypat: I think there are valid reasons to for example not specify a patch version (e.g. `libc = 0.2`), or to specify that we do not want to upgrade _past_ a specific patch version (e.g. `libc = "0.2, <0.2.118"`). Especially for crates that are past version 1.0, caret requirements allow to automatically drag in minor version upgrades (e.g. "1.2.3" means ">=1.2.3, <2.0.0", and to prohibit that, we would need `1.2.*` or similar. 

To cover all of these cases with a regex, if we really only want to disallow the comparison requirement, seems unergonomic to me. Besides, we do not have control over our dependencies' versioning, so we actually _need_ these additional tools in case they ever do not adhere to semver.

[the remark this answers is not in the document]
JonathanWoollett-Light: I think these reasons only apply in the case we depend on crates which do not adhere to SemVer, this is something we should actively try to avoid.

In the case it is unavoidable I think we should use the exact version e.g. `libc = "=0.2.117"`.

If we think this is a likely we can adjust the regex to be `Regex::new(r"(=|)[0-9]*\.[0-9]*\.[0-9]*")` to allow either SemVer versioning or pinning to an exact version.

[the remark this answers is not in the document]
roypat: Talking with jcawl@ offline, I agree that my solution is insufficient for ensuring proper dependency management, as one could still specify something as `<2340382` (ridiculously large number that will never be reached) or using wildcards. Instead of trying to catch all of these cases on a deny list, I will simply go ahead with the regex posted above. This is overly restrictive, but in praxis I have not seen use for ways to specify dependency versions beyond this. If we ever find a need to allow more requirement-types, this is always an easy change to make.

roypat:
Do not merge until #3209 is merged and I had the chance to update this PR accordingly

roypat:
This is blocked on getting updates to all rust-vmm crates (on our side, aghecen@ and me are working on this), as almost every rust-vmm crate will get a major version bump

tests/integration_tests/security/demo_seccomp/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Amazon Firecracker team <firecracker-devel@amazon.com>"]
 edition = "2018"
 
 [dependencies]
-libc = ">=0.2.39"
+libc = "0.2.66"
```
dianpopa, line 8: Can t we use the same as in prod? (i.e 0.2.121)?

tests/framework/dependencies.txt
```diff
@@ -1,6 +1,7 @@
 ['aead v0.4.3',
  'aes v0.7.5',
  'aes-gcm v0.9.4',
+ 'aho-corasick v0.7.19',
```
luminitavoicu, line 4: > For instance, vmm-sys-utils got updated in https://github.com/firecracker-microvm/firecracker/pull/3248. I do not know
why the test did not fail on that PR.
We should investigate this.

tests/integration_tests/build/test_dependencies.py
```diff
@@ -42,17 +42,17 @@ def test_num_dependencies(dep_file):
...
+    # appear multiple times).
     for line in deps:
         if line and "(*)" not in line:
             current_deps.add(line)
 
     # Use the code below to update the expected dependencies.
     # from pprint import pprint
     # with open(dep_file, "w", encoding='utf-8') as prev_deps:
-    #     pprint(sorted(current_deps), stream=prev_deps)
+    #      pprint(sorted(current_deps), stream=prev_deps)
```
luminitavoicu, line 55: Intentional?

src/firecracker/tests/verify_dependencies.rs
```diff
@@ -0,0 +1,77 @@
...
+                violating_dependencies.insert(
+                    fc_crate.file_name().into_string().unwrap(),
+                    violating_in_crate,
+                );
+            }
+        }
+    }
+
+    assert_eq!(
+        violating_dependencies,
```
dianpopa, line 39: Is this meant to check that `violating_dependencies` is empty? Can t we say `assert_eq(violating_dependencies.keys.len(), 0)`? -> just seems more straightforward

dianpopa (approved):
LGTM, just a couple of nits

luminitavoicu:
> Attempt to disable changes requested flag. The changes requested were discussed offline and the conclusion was to keep code snippet as is.
Correction of the above: the conclusion was to modify the code and changes were implemented as discussed.

src/firecracker/tests/verify_dependencies.rs
roypat: Yes, this just checks that the list is empty. The reason I chose to write the test this way instead of `assert!(violating_deps.is_empty())` is that this way, if the list is not empty, the assert message will contain every violating dependency, whereas otherwise it will just tell you _that_ the assert failed, without telling you what dependencies caused it to fail.

tests/host_tools/uffd/Cargo.toml
```diff
@@ -7,11 +7,11 @@ edition = "2018"
 [dependencies]
 utils = { path = "../../../src/utils" }
 
-libc = ">=0.2.39"
+libc = "0.2.121"
```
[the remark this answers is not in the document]
roypat: Generally my philosophy here was to not update any dependencies, so all versions are taken verbatim from the respective `Cargo.lock` files. Some crates had their libc dependencies added earlier than others, so their `Cargo.lock` files happen to be different.

src/firecracker/tests/verify_dependencies.rs
```diff
@@ -0,0 +1,77 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::HashMap;
+use std::path::Path;
+
+use cargo_toml::{Dependency, DependencyDetail, DepsSet, Manifest};
+use regex::Regex;
+
+#[test]
+fn test_no_comparison_requirements() {
```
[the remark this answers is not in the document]
roypat: This test is run by `cargo test`, so as part of `https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/build/test_unittests.py` (in the CI)

tests/integration_tests/build/test_dependencies.py
roypat: Not really (I think it was me on auto-pilot added "hash"-"space" to every line but the final line used to not have a space?)

tests/integration_tests/security/demo_seccomp/Cargo.toml
roypat: See https://github.com/firecracker-microvm/firecracker/pull/3248#discussion_r1035015892

src/firecracker/tests/verify_dependencies.rs
dianpopa: Nice, did not think of that!

luminitavoicu (dismissed):

dianpopa (approved):

luminitavoicu (approved):

PR #3249, opened 2022-11-08
Update rust-vmm/vmm-sys-util to 0.10.0

roypat:
Note that I had to delete my `build` directory to get the update to go through, otherwise `kvm-bindings` refused to pull in the updated `vmm-sys-util`

## Changes

Update vmm-sys-util to 0.10.0

## Reason

Due to unlucky dependency management coinciding with a breaking change, cargo update caused an unbuildable workspace, and due to external factors, #3248 does not fix this.

## License Acceptance

## PR Checklist

---


mattschlebusch (approved):

tests/integration_tests/security/demo_seccomp/Cargo.toml
```diff
@@ -5,7 +5,7 @@ authors = ["Amazon Firecracker team <firecracker-devel@amazon.com>"]
 edition = "2018"
 
 [dependencies]
-libc = ">=0.2.39"
+libc = "0.2.66"
```
aghecenco, line 8: `libc` is a dependency in a lot of crates, maybe add a separate commit before these 2 that updates it everywhere to 0.2.66? What's more, does the dependency need to be `=0.2.66` or can it be `>=`?

roypat: This PR does not update any dependencies, all of the versions I specify in this PR are taken directly from the corresponding Cargo.lock files (which is why those files are unchanged in this PR). Operationally, `cargo build` ran before and after this PR should produce bit-identical binaries. 

For your second question, there's a difference between `libc = "0.2.66"`, `libc = ">=0.2.66"` and `libc = "=0.2.66"`. The first one fixes the minor version but allows patch updates, the second fixes nothing (e.g. potentially breaking minor version updates would be allows) and the last fixes even the patch version. See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html for more information.

roypat: I am so sorry, I just realized that I somehow managed to put a change from #3248 into this PR. My comment above refers to the intended change of #3248, not this PR

This file should not have been touched, thanks for catching this!

aghecenco (approved):

mattschlebusch (approved):

PR #3251, opened 2022-11-09
rename error on block device update

dianpopa:

## Changes

* be consistent when referring to the block device (sometimes it would be block device, sometimes drive)
* remove unused error entry

## Reason

...

## License Acceptance

## PR Checklist

---

### DIFF
--- a/src/devices/src/virtio/queue.rs
+++ b/src/devices/src/virtio/queue.rs
@@ -6,7 +6,6 @@
 // found in the THIRD-PARTY file.
 
 use std::cmp::min;
-use std::fmt;
 use std::num::Wrapping;
 use std::sync::atomic::{fence, Ordering};
 
@@ -26,26 +25,14 @@ pub(super) const VIRTQ_DESC_F_WRITE: u16 = 0x2;
 // The Virtio Spec 1.0 defines the alignment of VirtIO descriptor is 16 bytes,
 // which fulfills the explicit constraint of GuestMemoryMmap::read_obj_from_addr().
 
-#[derive(Debug, derive_more::From)]
+#[derive(Debug, thiserror::Error)]
 pub enum QueueError {
     /// Descriptor index out of bounds.
+    #[error("Descriptor index out of bounds: {0}.")]
     DescIndexOutOfBounds(u16),
     /// Attempted an invalid write into the used ring.
-    UsedRing(GuestMemoryError),
-}
-
-impl fmt::Display for QueueError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        use self::QueueError::*;
-
-        match self {
-            DescIndexOutOfBounds(val) => write!(f, "Descriptor index out of bounds: {val}"),
-            UsedRing(err) => write!(
-                f,
-                "Failed to write value into the virtio queue used ring: {err}"
-            ),
-        }
-    }
+    #[error("Failed to write value into the virtio queue used ring: {0}")]
+    UsedRing(#[from] GuestMemoryError),
 }
 
 /// A virtio descriptor constraints with C representative.
--- a/src/devices/src/virtio/queue.rs
+++ b/src/devices/src/virtio/queue.rs
@@ -39,11 +39,10 @@ impl fmt::Display for QueueError {
         use self::QueueError::*;
 
         match self {
-            DescIndexOutOfBounds(val) => write!(f, "Descriptor index out of bounds: {}", val),
+            DescIndexOutOfBounds(val) => write!(f, "Descriptor index out of bounds: {val}"),
             UsedRing(err) => write!(
                 f,
-                "Failed to write value into the virtio queue used ring: {}",
-                err
+                "Failed to write value into the virtio queue used ring: {err}"
             ),
         }
     }
--- a/src/vmm/src/device_manager/mmio.rs
+++ b/src/vmm/src/device_manager/mmio.rs
@@ -6,7 +6,6 @@
 // found in the THIRD-PARTY file.
 
 use std::collections::HashMap;
-use std::fmt;
 use std::sync::{Arc, Mutex};
 
 #[cfg(target_arch = "aarch64")]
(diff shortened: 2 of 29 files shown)


src/vmm/src/vmm_config/drive.rs
```diff
@@ -44,10 +42,11 @@ impl Display for DriveError {
     fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
         use self::DriveError::*;
         match self {
-            CreateBlockDevice(err) => write!(f, "Unable to create the block device {:?}", err),
-            BlockDeviceUpdateFailed(err) => write!(f, "The update operation failed: {}", err),
+            BlockDeviceUpdate(err) => {
+                write!(f, "Unable to patch the block device: {}", err)
+            }
+            CreateBlockDevice(err) => write!(f, "Unable to create the block device: {:?}", err),
```
JonathanWoollett-Light, line 48:
```suggestion
            BlockDeviceUpdate(err) => write!(f, "Unable to patch the block device: {err}"),
            CreateBlockDevice(err) => write!(f, "Unable to create the block device: {err:?}"),
```

src/vmm/src/device_manager/mmio.rs
```diff
@@ -68,7 +68,7 @@ impl fmt::Display for Error {
             }
             Error::EventFd(err) => write!(f, "failed to create or clone event descriptor: {}", err),
             Error::IncorrectDeviceType => write!(f, "incorrect device type"),
-            Error::InternalDeviceError(err) => write!(f, "device error: {}", err),
+            Error::InternalDeviceError(err) => write!(f, "{}", err),
```
JonathanWoollett-Light, line 71: It is generally better to inline format! arguments (https://rust-lang.github.io/rust-clippy/master/#uninlined_format_args)
```suggestion
            Error::InternalDeviceError(err) => write!(f, "{err}"),
```

aghecenco (dismissed):

dianpopa:
@JonathanWoollett-Light @aghecenco tests have passed, can you give it another look? Thx

src/devices/src/virtio/queue.rs
```diff
@@ -39,11 +39,10 @@ impl fmt::Display for QueueError {
         use self::QueueError::*;
```
JonathanWoollett-Light, line 39: Could `thiserror::Error` be used on `QueueError` to avoid needing to write this manually?

src/vmm/src/device_manager/mmio.rs
```diff
@@ -35,46 +35,40 @@ use vm_memory::GuestAddress;
 /// Errors for MMIO device manager.
 #[derive(Debug)]
 pub enum Error {
```
JonathanWoollett-Light, line 37: Could `thiserror::Error` be used on `Error` here to avoid needing to write the `Display` implementation manually?

src/vmm/src/device_manager/mmio.rs
```diff
@@ -671,31 +665,27 @@ mod tests {
...
+                Error::InvalidDeviceType => format!("{}{:?}", err, err),
                 Error::InternalDeviceError(_) => format!("{}{:?}", err, err),
-                Error::InvalidInput => format!("{}{:?}", err, err),
+                Error::InvalidIrqConfig => format!("{}{:?}", err, err),
                 Error::RegisterIoEvent(_) => format!("{}{:?}", err, err),
                 Error::RegisterIrqFd(_) => format!("{}{:?}", err, err),
-                Error::UpdateFailed => format!("{}{:?}", err, err),
-                Error::AllocatorError(_) => format!("{}{:?}", err, err),
+                Error::Allocator(_) => format!("{}{:?}", err, err),
             };
```
JonathanWoollett-Light, line 677: I don't see a reason for this match, can this not be replaced like:
```suggestion
            let msg = format!("{err}{err:?}");
```

src/vmm/src/vmm_config/drive.rs
```diff
@@ -23,19 +23,14 @@ type Result<T> = result::Result<T, DriveError>;
 /// Errors associated with the operations allowed on a drive.
 #[derive(Debug)]
 pub enum DriveError {
```
JonathanWoollett-Light, line 25: Could `thiserror::Error` be used on `DriveError` here to avoid needing to write the `Display` implementation manually?

JonathanWoollett-Light (approved):

src/vmm/src/vmm_config/drive.rs
dianpopa: Yes,  I created a separate commit for that.

src/vmm/src/device_manager/mmio.rs
dianpopa: Done, created a separate commit.

src/vmm/src/device_manager/mmio.rs
dianpopa: Reason is mentioned in comment "// Use an exhaustive 'match' to make sure we cover all error variants.".

src/devices/src/virtio/queue.rs
dianpopa: Done in a separate commit.

dianpopa (commented):
@JonathanWoollett-Light  addressed your comments, PTAL!

src/vmm/src/device_manager/mmio.rs
JonathanWoollett-Light: I don't see the value in this, it doesn't add coverage.

src/devices/src/virtio/queue.rs
```diff
@@ -26,29 +25,16 @@ pub(super) const VIRTQ_DESC_F_WRITE: u16 = 0x2;
 // The Virtio Spec 1.0 defines the alignment of VirtIO descriptor is 16 bytes,
 // which fulfills the explicit constraint of GuestMemoryMmap::read_obj_from_addr().
 
-#[derive(Debug, derive_more::From)]
+#[derive(Debug, derive_more::From, thiserror::Error)]
```
JonathanWoollett-Light, line 28: Can `#[from]` from `thiserror::Error` be used? (then `derive_more::From` could be removed).

JonathanWoollett-Light (dismissed):

src/vmm/src/device_manager/mmio.rs
ShadowCurse: So this basically tests that error messages are not empty. Not so convinced in usefulness of this.
Also relying on the comment is not so great idea. It can be ignored easily.
Do we check that error message is not zero anywhere else? If someone will add new error with empty message it will be obvious without this test.

Anyway you can collapse this match to use "or":
```rust
    let msg = match err {
        Error::Bus(_)
        | Error::BusInsert(_)
        | Error::Cmdline(_)
        | Error::DeviceNotFound
        | Error::EventFd(_)
        | Error::IncorrectDeviceType
        | Error::InvalidDeviceType
        | Error::InternalDeviceError(_)
        | Error::InvalidInput
        | Error::InvalidIrqConfig
        | Error::RegisterIoEvent(_)
        | Error::RegisterIrqFd(_)
        | Error::UpdateFailed
        | Error::AllocatorError(_)
        | Error::Allocator(_) => format!("{err}{err:?}"),
    };

```

dianpopa: This indeed might not bring a lot of value. We have lots of unit tests this way, unfortunately. If you feel like it, propose a change.
This PR is not related to this.

roypat: I disagree that the PR is not related to removing this specific test. I tried tracing down the reason for these kind of tests, and they were indeed for code coverage reason: If we did not have these tests, then a _manually implemented_ `Display` or `Debug` implementation would not be covered. This is also why this match has to be exhaustive: If we added a match arm to the `Display` implementation, then we'd need to also add it to the test, otherwise the new arm would be uncovered. 

_However_ we have since switched to use `thiserror`, which removes manually implemented `Display` implementations. So the need for these tests that "game the metric" is no longer present. For this reason, I would argue that removing them should happen in the same commit that introduces a `thiserror` implementation. Sadly this hasn't happened so far :(

dianpopa: TBH I do not know how `thiserror` works behind the scenes, but was wondering why you say that this test was only relevant for manual implementation and is not for `thiserror`? Will the lines adding the "#[error("....")]" display message get covered automatically with `thiserror`?

roypat: The lines marked as `#[error(...)]` are not really lines of code, they are instructions for the `thiserror` proc-macro. They are treated by code coverage tooling just like comments: neither covered nor uncovered. See the below screenshot of a coverage report from an enum using `thiserror` from the api_server crate:
image

dianpopa: thanks @roypat for clarifying. I will remove the test.

src/devices/src/virtio/queue.rs
dianpopa: Done.

dianpopa (commented):
Addressed outstanding feedback, PTAL!

ShadowCurse (approved):

PR #3245, opened 2022-11-07
make tests fail when criteria not found

zulinx86:
## Reason

When criteria is not available, performance tests pass without any errors.
By making them fail in such a situation, it lets us know that performance baselines are missing.

## Changes

Makes tests fail when criteria is not found. 

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

- [x] All commits in this PR are signed (`git commit -s`).
- [x] If a specific issue led to this PR, this PR closes the issue.
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] New `unsafe` code is documented.
- [x] API changes follow the [Runbook for Firecracker API changes][2].
- [x] User-facing changes are mentioned in `CHANGELOG.md`.
- [x] All added/changed functionality is tested.

---

- [ ] This functionality can be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md

PR #3225, opened 2022-10-30
Add performance baseline for m6a.metal

zulinx86:
## Changes

Adds missing baselines for long-running performance tests for m6a.metal.

## Reason

Currently, the following baselines for long-running performance tests are missing for m6a.metal
- vsock_throughput
- network_tcp_throughput
- block_performance
- network_latency

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

- [x] All commits in this PR are signed (`git commit -s`).
- [x] If a specific issue led to this PR, this PR closes the issue.
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] New `unsafe` code is documented.
- [x] API changes follow the [Runbook for Firecracker API changes][2].
- [x] User-facing changes are mentioned in `CHANGELOG.md`.
- [x] All added/changed functionality is tested.

---

- [ ] This functionality can be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md

PR #3253, opened 2022-11-10
Revert "make tests fail when criteria not found"

mattschlebusch:
## Changes

This reverts commit 082e58fbd931040a777da06f963eab71b6e8b8ff ([PR/3245](https://github.com/firecracker-microvm/firecracker/pull/3245)).

## Description

A change was introduced in [PR/3245](https://github.com/firecracker-microvm/firecracker/pull/3245) to force performance tests to fail if a given baseline configuration was missing for a given CPU. While this seems to have had the desired affect, the commit did not include baseline configurations for actual missing configurations:

* Intel m6i (4.14) - Vsock Throughput
* Intel m6i (5.10) - Vsock Throughput
* Intel m6i (4.14) - Network TCP Throughput
* Intel m6i (5.10) - Network TCP Throughput
* Arm (4.14) - Vsock Throughput

While the change is good and should be introduced, results in failures of performance pipelines for missing performance baseline configuration. This makes it harder to tell if other regressions are introduced as the performance pipelines are consistently in alarm. This change needs to be re-introduced with the missing baselines before merging into firecracker:main.

## License Acceptance

### DIFF
--- a/tests/framework/stats/consumer.py
+++ b/tests/framework/stats/consumer.py
@@ -7,8 +7,7 @@ from abc import ABC, abstractmethod
 from numbers import Number
 from typing import Any, Callable
 from collections import defaultdict
-from framework.utils import ExceptionAggregator, get_kernel_version
-from framework.utils_cpuid import get_cpu_model_name, get_instance_type
+from framework.utils import ExceptionAggregator
 
 from .criteria import CriteriaException
 from .metadata import Provider as MetadataProvider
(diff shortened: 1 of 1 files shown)


zulinx86 (approved):
They failed basically due to the lack of "duration/Avg" values.
This metric looks not for performance result, but for performance test setting.
We may need to delete them from baseline files like this: https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/performance/configs/test_vsock_throughput_config_4.14.json#L39-L46
I'll take a look at it later.

zulinx86:
In addition, this issue is not archecture-dependent and happen for vsock throughput test and network TCP throughput test in any architecture.
The reason why we don't detect this on AMD (m6a.metal) is we don't have performance baselines for this now (I'm gathering this on PR #3225 )

andreitraistaru (approved):

PR #3254, opened 2022-11-10
Add support for virtio-rng

bchalios:
## Changes

Add support for the [VirtIO entropy device](https://docs.oasis-open.org/virtio/virtio/v1.2/cs01/virtio-v1.2-cs01.html#x1-3050004). This is a VirtIO device that provides the guest VM with entropy bytes.

This PR provides the basic functionality for the device:
1. Implement the VirtIO protocol.
2. Metrics for the device
3. Allow snapshotting of the device
4. Configure the device through a new API request: `PUT /entropy` (with a, currently, empty body).

We allow only one `entropy` device. The device gets random data from the `getrandom` system call, through the `OsRng` type of the `rand` crate.

Subsequent PRs will:
1. Add rate-limiting for the device.
2. Add integration tests for the device.

## Reason

We are investigating the possibility to implement snapsafety over VirtIO by extending the entropy device with VMGenID-like functionality.

## License Acceptance

## PR Checklist

---


src/devices/src/virtio/mod.rs
```diff
@@ -49,6 +51,7 @@ mod device_status {
 /// Type 0 is not used by virtio. Use it as wildcard for non-virtio devices
 pub const TYPE_NET: u32 = 1;
 pub const TYPE_BLOCK: u32 = 2;
+pub const TYPE_RNG: u32 = 4;
 pub const TYPE_BALLOON: u32 = 5;
```
JonathanWoollett-Light, line 55: We can use [the virtio bindings from rust-vmm](https://github.com/rust-vmm/vm-virtio/tree/main/crates/virtio-bindings) e.g. [`virtio_bindings::VIRTIO_ID_RNG`](https://docs.rs/virtio-bindings/0.1.0/virtio_bindings/bindings/virtio_blk/constant.VIRTIO_ID_RNG.html) etc.

bchalios: I was wondering why don't we use this, but I see the project is archived: https://github.com/rust-vmm/virtio-bindings

JonathanWoollett-Light: https://github.com/rust-vmm/virtio-bindings are the old bindings, what I linked https://github.com/rust-vmm/vm-virtio/tree/main/crates/virtio-bindings are the new bindings which are not archived.

src/vmm/src/builder.rs
```diff
@@ -936,6 +941,24 @@ pub(crate) fn attach_boot_timer_device(
...
+        devices::virtio::rng::Entropy::new().map_err(StartMicrovmError::CreateEntropyDevice)?,
+    ));
+    let id = entropy_device
+        .lock()
+        .expect("Poisoned lock")
+        .id()
+        .to_string();
+
+    attach_virtio_device(event_manager, vmm, id, entropy_device, cmdline)?;
+    Ok(())
```
JonathanWoollett-Light, line 962: nit
```suggestion
    attach_virtio_device(event_manager, vmm, id, entropy_device, cmdline)
```

src/api_server/src/parsed_request.rs
```diff
@@ -717,6 +719,19 @@ pub(crate) mod tests {
...
+    fn test_try_from_put_entropy() {
+        let (mut sender, receiver) = UnixStream::pair().unwrap();
+        let mut connection = HttpConnection::new(receiver);
+        let body = "{}";
+        sender
+            .write_all(http_request("PUT", "/entropy", Some(body)).as_bytes())
+            .unwrap();
+        assert!(connection.try_read().is_ok());
+        let req = connection.pop_parsed_request().unwrap();
+        assert!(ParsedRequest::try_from_request(&req).is_ok());
```
JonathanWoollett-Light, line 732: Nit it is usually better to match the specific value.
```suggestion
        assert_eq!(ParsedRequest::try_from_request(&req), ParsedRequest { .. });
```
or as close as possible e.g.
```
assert!(matches!(ParsedRequest::try_from_request(&req), Ok(...)))
```

src/api_server/src/request/entropy.rs
```diff
@@ -0,0 +1,33 @@
...
+    Ok(ParsedRequest::new_sync(VmmAction::SetEntropyDevice(cfg)))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_parse_put_entropy_request() {
+        assert!(parse_put_entropy(&Body::new("invalid_payload")).is_err());
```
JonathanWoollett-Light, line 21: Nit it is usually better to match the specific value e.g. `assert_eq!(..., Err(...))`.

src/api_server/src/request/entropy.rs
```diff
@@ -0,0 +1,33 @@
...
+
+        // PUT with invalid fields.
+        let body = r#"{
+            "some_id": 4
+        }"#;
+        assert!(parse_put_entropy(&Body::new(body)).is_err());
```
JonathanWoollett-Light, line 27: Nit it is usually better to match the specific value e.g. `assert_eq!(..., Err(...))`.

src/api_server/src/request/entropy.rs
```diff
@@ -0,0 +1,33 @@
...
+
+        // PUT with valid fields.
+        let body = r#"{}"#;
+        assert!(parse_put_entropy(&Body::new(body)).is_ok());
```
JonathanWoollett-Light, line 31: Nit it is usually better to match the specific value e.g. `assert_eq!(..., Ok(...))`.

src/vmm/src/builder.rs
```diff
@@ -943,19 +948,17 @@ pub(crate) fn attach_boot_timer_device(
 
 fn attach_entropy_device(
     vmm: &mut Vmm,
-    event_manager: &mut EventManager,
     cmdline: &mut LoaderKernelCmdline,
+    event_manager: &mut EventManager,
+    entropy_device: &Arc<Mutex<Entropy>>,
```
JonathanWoollett-Light, line 953: Would it make sense to replace `&Arc<Mutex<Entropy>>` with `Arc<Mutex<Entropy>>`?

src/vmm/src/builder.rs
```diff
@@ -1258,6 +1263,23 @@ pub mod tests {
...
+    pub(crate) fn insert_entropy_device(
+        vmm: &mut Vmm,
+        cmdline: &mut Cmdline,
+        event_manager: &mut EventManager,
+        entropy_config: EntropyDeviceConfig,
+    ) {
+        let mut builder = EntropyDeviceBuilder::new();
+        let entropy = builder.build(entropy_config).unwrap();
+
+        assert!(attach_entropy_device(vmm, cmdline, event_manager, &entropy).is_ok());
```
JonathanWoollett-Light, line 1275: Nit it is usually better to match the specific value e.g. assert_eq!(..., Ok(...)).

src/vmm/src/builder.rs
```diff
@@ -1258,6 +1263,23 @@ pub mod tests {
...
+
+        assert!(vmm
+            .mmio_device_manager
+            .get_device(DeviceType::Virtio(TYPE_RNG), ENTROPY_DEV_ID)
+            .is_some());
+    }
```
JonathanWoollett-Light, line 1281: Nit it is usually better to match the specific value e.g. assert_eq!(..., Some(...)).

src/vmm/src/device_manager/persist.rs
```diff
@@ -723,6 +724,9 @@ mod tests {
                 uds_path: tmp_sock_file.as_path().to_str().unwrap().to_string(),
             };
             insert_vsock_device(&mut vmm, &mut cmdline, &mut event_manager, vsock_config);
+            // Add an entropy device.
+            let entropy_config = EntropyDeviceConfig {};
```
JonathanWoollett-Light:
```suggestion
           let entropy_config = EntropyDeviceConfig;
```

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -0,0 +1,109 @@
...
+use std::sync::{Arc, Mutex};
+
+use devices::virtio::rng::{Entropy, Error as EntropyError};
+use serde::{Deserialize, Serialize};
+
+/// This struct represents the strongly typed equivalent of the json body from entropy device
+/// related requests.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct EntropyDeviceConfig {}
```
JonathanWoollett-Light, line 14:
```suggestion
/// This struct represents the strongly typed equivalent of the json body from entropy device
/// related requests.
#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
#[serde(deny_unknown_fields)]
pub struct EntropyDeviceConfig;
```

src/vmm/src/resources.rs
```diff
@@ -28,48 +29,44 @@ use crate::vstate::vcpu::VcpuConfig;
 type Result<E> = std::result::Result<(), E>;
 
 /// Errors encountered when configuring microVM resources.
-#[derive(Debug, derive_more::From)]
+#[derive(Debug, thiserror::Error, derive_more::From)]
```
JonathanWoollett-Light, line 32: You can use `#[from]` with `thiserror::Error` to cover the functionality of `derive_more::From`, so maybe it makes sense to remove `derive_more::From` here?

src/vmm/src/resources.rs
```diff
@@ -1368,6 +1385,21 @@ mod tests {
         assert!(vm_resources.set_balloon_device(new_balloon_cfg).is_err());
     }
 
+    #[test]
+    fn test_set_entropy_device() {
+        let mut vm_resources = default_vm_resources();
+        vm_resources.entropy = EntropyDeviceBuilder::new();
+        let entropy_device_cfg = EntropyDeviceConfig {};
+
+        assert!(vm_resources.entropy.get().is_none());
```
JonathanWoollett-Light, line 1397: Nit: I tend to prefer using `assert_eq` to be more inline with the other specific asserts, but this is very picky.
```suggestion
        assert_eq!(vm_resources.entropy.get(),None);
```

src/vmm/src/resources.rs
```diff
@@ -1579,5 +1611,17 @@ mod tests {
                 VsockConfigError::CreateVsockDevice(VsockError::BufDescTooSmall)
             )
         );
+        assert_eq!(
+            format!(
+                "{}",
+                Error::EntropyDevice(EntropyDeviceError::CreateDevice(
+                    EntropyError::ReadOnlyDescriptor
+                ))
+            ),
```
JonathanWoollett-Light, line 1623:
```suggestion
            Error::EntropyDevice(EntropyDeviceError::CreateDevice(
                EntropyError::ReadOnlyDescriptor
            )).to_string(),
```

src/vmm/src/rpc_interface.rs
```diff
@@ -126,88 +129,71 @@ pub enum VmmAction {
 }
 
 /// Wrapper for all errors associated with VMM actions.
-#[derive(Debug, derive_more::From)]
+#[derive(Debug, thiserror::Error, derive_more::From)]
```
JonathanWoollett-Light, line 132: You can remove `derive_more::From` as `thiserror::Error` covers its functionality with `#[from]`.

src/vmm/src/rpc_interface.rs
```diff
@@ -126,88 +129,71 @@ pub enum VmmAction {
...
 pub enum VmmActionError {
     /// The action `SetBalloonDevice` failed because of bad user input.
+    #[error("{0}")]
```
JonathanWoollett-Light, line 135: In these cases could we simply include the comment?
```suggestion
    /// The action `SetBalloonDevice` failed because of bad user input.
    #[error("The action `SetBalloonDevice` failed because of bad user input: {0}")]
```

src/vmm/src/rpc_interface.rs
```diff
@@ -1392,6 +1402,15 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_preboot_set_entropy_device() {
+        let req = VmmAction::SetEntropyDevice(EntropyDeviceConfig {});
```
JonathanWoollett-Light, line 1407:
```suggestion
        let req = VmmAction::SetEntropyDevice(EntropyDeviceConfig);
```

src/vmm/src/rpc_interface.rs
```diff
@@ -2082,6 +2101,10 @@ mod tests {
             }),
             VmmActionError::OperationNotSupportedPostBoot,
         );
+        check_runtime_request_err(
+            VmmAction::SetEntropyDevice(EntropyDeviceConfig {}),
```
JonathanWoollett-Light, line 2105:
```suggestion
            VmmAction::SetEntropyDevice(EntropyDeviceConfig),
```

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/// Errors that can occur while handling configuration for
+/// an entropy device
+#[derive(Debug, thiserror::Error)]
+pub enum EntropyDeviceError {
+    /// Error while create an entropy device
+    #[error("Could not create Entropy device: {0}")]
+    CreateDevice(#[from] EntropyError),
+}
```
JonathanWoollett-Light, line 29: Is it expected that additional variants will be added? If not this can be more simply represented as a newtype:
```suggestion
/// Errors that can occur while handling configuration for
/// an entropy device
#[derive(Debug, thiserror::Error)]
#[error("Could not create Entropy device: {0}")]
pub struct EntropyDeviceError(#[from] EntropyError);
```

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -0,0 +1,109 @@
...
+
+type Result<T> = std::result::Result<T, EntropyDeviceError>;
+
+/// A builder type used to construct an Entropy device
+#[derive(Default)]
+pub struct EntropyDeviceBuilder {
+    entropy_device: Option<Arc<Mutex<Entropy>>>,
+}
```
JonathanWoollett-Light, line 37:
```suggestion
/// A builder type used to construct an Entropy device
#[derive(Default)]
pub struct EntropyDeviceBuilder(Option<Arc<Mutex<Entropy>>>);
```

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    pub(crate) fn default_config() -> EntropyDeviceConfig {
+        EntropyDeviceConfig {}
+    }
+
+    #[test]
+    fn test_entropy_device_create() {
+        let config = default_config();
+        let mut builder = EntropyDeviceBuilder::new();
+        assert!(builder.get().is_none());
```
JonathanWoollett-Light, line 91: Nit: Prefer I tend to prefer `assert_eq!` here.

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -0,0 +1,109 @@
...
+
+        builder.insert(config.clone()).unwrap();
+        assert!(builder.get().is_some());
```
JonathanWoollett-Light, line 94: Nit: Prefer I tend to prefer `assert_eq!` here.

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -0,0 +1,109 @@
...
+        assert_eq!(builder.config().unwrap(), config);
+    }
+
+    #[test]
+    fn test_set_device() {
+        let mut builder = EntropyDeviceBuilder::new();
+        let device = Entropy::new().unwrap();
+        assert!(builder.entropy_device.is_none());
```
JonathanWoollett-Light, line 102: Nit: Prefer I tend to prefer `assert_eq!` here.

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -0,0 +1,109 @@
...
+        builder.set_device(Arc::new(Mutex::new(device)));
+        assert!(builder.entropy_device.is_some());
```
JonathanWoollett-Light, line 104: Nit: Prefer I tend to prefer `assert_eq!` here.

src/devices/src/virtio/rng/persist.rs
```diff
@@ -0,0 +1,102 @@
...
+use vm_memory::GuestMemoryMmap;
+
+use crate::virtio::persist::Error as VirtioStateError;
+use crate::virtio::rng::{Entropy, Error as EntropyError, NUM_QUEUES, QUEUE_SIZE};
+use crate::virtio::{VirtioDeviceState, TYPE_RNG};
+
+#[derive(Clone, Versionize)]
+pub struct EntropyState {
+    virtio_state: VirtioDeviceState,
+}
```
JonathanWoollett-Light, line 18:
```suggestion
#[derive(Clone, Versionize)]
pub struct EntropyState(VirtioDeviceState);
```

src/devices/src/virtio/rng/persist.rs
```diff
@@ -0,0 +1,102 @@
...
+
+pub struct EntropyConstructorArgs {
+    pub mem: GuestMemoryMmap,
+}
```
JonathanWoollett-Light, line 22:
```suggestion
pub struct EntropyConstructorArgs(pub GuestMemoryMmap)
```

src/vmm/src/device_manager/mmio.rs
```diff
@@ -448,6 +448,13 @@ impl MMIODeviceManager {
                     // Any in-flight packets or events are simply lost.
                     // Vsock is restored 'empty'.
                 }
+                TYPE_RNG => {
+                    let entropy = virtio.as_mut_any().downcast_mut::<Entropy>().unwrap();
+                    if entropy.is_activated() {
+                        info!("kick entropy {}.", id);
```
JonathanWoollett-Light, line 454:
```suggestion
                        info!("kick entropy {id}.");
```

src/vmm/src/device_manager/persist.rs
```diff
@@ -766,6 +824,37 @@ mod tests {
...
+                .set_type_version(DeviceStates::type_id(), 4)
+                .set_type_version(NetConfigSpaceState::type_id(), 2);
+
+            vmm.mmio_device_manager
+                .save()
+                .serialize(&mut buf.as_mut_slice(), &version_map, 4)
+                .unwrap();
+            let device_states: DeviceStates =
+                DeviceStates::deserialize(&mut buf.as_slice(), &version_map, 4).unwrap();
+            assert!(device_states.entropy_device.is_some());
```
JonathanWoollett-Light, line 856: Nit it is usually better to match the specific value e.g. `assert_eq!(..., Some(...))`.

src/devices/src/virtio/rng/device.rs
```diff
@@ -109,31 +113,41 @@ impl Entropy {
...
+                METRICS.entropy.entropy_event_fails.inc();
                 0
             });
 
             used_any = true;
             self.queues[0]
                 .add_used(mem, index, bytes)
                 .unwrap_or_else(|err| {
-                    error!("entropy: Could not add used descriptor to queue: {}", err)
+                    error!("entropy: Could not add used descriptor to queue: {}", err);
```
JonathanWoollett-Light, line 129:
```suggestion
                    error!("entropy: Could not add used descriptor to queue: {err}");
```

src/devices/src/virtio/rng/device.rs
```diff
@@ -109,31 +113,41 @@ impl Entropy {
...
                 });
+
+            METRICS.entropy.entropy_bytes.add(bytes as usize);
         }
 
         if used_any {
-            self.signal_used_queue()
-                .unwrap_or_else(|err| error!("entropy: {:?}", err));
+            self.signal_used_queue().unwrap_or_else(|err| {
+                error!("entropy: {:?}", err);
```
JonathanWoollett-Light, line 138:
```suggestion
                error!("entropy: {err:?}");
```

src/devices/src/virtio/rng/device.rs
```diff
@@ -0,0 +1,332 @@
...
+    ReadOnlyDescriptor,
+    #[error("Bad guest memory buffer: {0}")]
+    GuestMemory(#[from] GuestMemoryError),
+    #[error("Could not get random bytes: {0}")]
+    Random(#[from] rand::Error),
+}
+
+type Result<T> = std::result::Result<T, Error>;
+
+pub struct Entropy {
```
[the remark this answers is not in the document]
bchalios: This would essentially create dead code, since we don't print entropy anywhere.

src/devices/src/virtio/queue.rs
```diff
@@ -165,6 +165,30 @@ impl<'a> DescriptorChain<'a> {
     }
 }
 
+pub struct DescriptorIterator<'a> {
+    inner: Option<DescriptorChain<'a>>,
+}
```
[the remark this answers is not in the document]
bchalios: what is the added value of doing so?

src/devices/src/virtio/rng/event_handler.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        match source {
+            _ if source == self.queue_events()[RNG_QUEUE].as_raw_fd() => {
+                self.process_entropy_queue_event()
+            }
+            _ if source == self.activate_event().as_raw_fd() => self.process_activate_event(ops),
+            _ => {
+                warn!("rng: Unknown event received: {:?}", source);
+            }
+        }
```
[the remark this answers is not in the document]
bchalios: agreed. I have changed it

src/devices/src/virtio/rng/device.rs
```diff
@@ -0,0 +1,332 @@
...
+        ENTROPY_DEV_ID
+    }
+
+    fn signal_used_queue(&self) -> std::result::Result<(), DeviceError> {
+        debug!("entropy: raising IRQ");
+        self.irq_trigger
+            .trigger_irq(IrqType::Vring)
+            .map_err(DeviceError::FailedSignalingIrq)?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
bchalios: done!

src/devices/src/virtio/queue.rs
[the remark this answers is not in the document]
JonathanWoollett-Light: It is a common pattern and minimizes/simplifies the code.

src/vmm/src/device_manager/persist.rs
bchalios: Hmmm, interesting. It sounds counter-intuitive to me, since I don't really care about what's inside `Some`. Could you elaborate?

src/devices/src/virtio/rng/device.rs
bchalios: done

src/devices/src/virtio/rng/device.rs
bchalios: done!

src/devices/src/virtio/rng/persist.rs
bchalios: done

src/vmm/src/device_manager/mmio.rs
bchalios: Done! Actually, I 'm very surprised our rust-style test doesn't complain about these.

src/devices/src/virtio/rng/device.rs
```diff
@@ -0,0 +1,332 @@
...
+            // SAFETY: This is safe since we get `ptr` from `get_slice` which also checks the length
+            // of the descriptor
+            let slice = unsafe { std::slice::from_raw_parts_mut(ptr, desc.len as usize) };
+
+            OsRng.try_fill_bytes(slice)?;
+            bytes += desc.len;
+        }
+
+        Ok(bytes)
+    }
```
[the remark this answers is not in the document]
bchalios: The logic here will be changed a bit later with the PR that adds rate-limiting: we will need to first parse the whole chain to see how many bytes we have to consume in total. So, I 'd prefer to do this there, if that's fine with you.

src/vmm/src/device_manager/persist.rs
JonathanWoollett-Light: Take an example:
```
fn outer_fn() -> Option<i32> {
    Some(inner_fn())
}
fn inner_fn() -> i32 {
    4
}
```
If we assume `inner_fn()` is pre-existing and that we are adding `outer_fn()` we may test this by `assert!(outer_fn().is_some())`  but here we are missing something. `outer_fn()` could definitely mutate the the value from `inner_fn()` incorrectly and still return `Some` so it would be better to test `assert_eq!(outer_fn(),Some(4))` as we test the full behavior of `outer_fn()`.

Testing `assert!(outer_fn().is_some())` is like testing `assert!(inner_fn() > 0)` (in fact if `inner_fn()` returned `NonZeroI32` these would compile to the exact same) when it would be better to test the exact value.

bchalios: I understand that we should check the value contained in `Some` if we cared about it.
Since here we: (1) rely on `Versionize` to do the serialization/deserialization, (2) it's not quite straight-forward to get what's inside `Some` and we actually check for the expected VM configuration later on, I think it's not needed in this particular assertion.

Thoughts?

src/devices/src/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/device_manager/persist.rs
JonathanWoollett-Light: It is always better to check the value, but if its not worth the effort in this case I think that's reasonable.

src/devices/src/virtio/rng/device.rs
```diff
@@ -0,0 +1,332 @@
...
+
+    fn is_activated(&self) -> bool {
+        self.device_state.is_activated()
+    }
+
+    fn activate(&mut self, mem: GuestMemoryMmap) -> ActivateResult {
+        if self.activate_event.write(1).is_err() {
+            error!("entropy: Cannot write to activate_evt");
+            return Err(super::super::ActivateError::BadActivate);
+        }
```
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/rpc_interface.rs
bchalios: I had done this initially, but there are tests that depend on the error message. I don't want to fix all these in this PR.

src/devices/src/virtio/rng/test_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+            for _ in 0..num_queues {
+                let vqueue = VirtQueue::new(start_addr, mem, Self::QUEUE_SIZE);
+                // Address for the next virt queue will be the first aligned address after
+                // the end of this one.
+                start_addr = vqueue.end().unchecked_align_up(VirtqDesc::ALIGNMENT);
+                virtqueues.push(vqueue);
+            }
+
+            virtqueues
+        }
```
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/builder.rs
bchalios: done

src/api_server/src/parsed_request.rs
bchalios: I think that finding what `ParsedRequest` should look like is not quite straight-forward and I don't think it's worth the effort here.

src/api_server/src/request/entropy.rs
bchalios: same as before, I don't think the added value justifies the complexity.

src/api_server/src/request/entropy.rs
bchalios: same

src/api_server/src/request/entropy.rs
bchalios: same

src/vmm/src/builder.rs
bchalios: The only thing this `Arc` is used for is to be `clone`d later on. So if we were just passing an `Arc` instead of an `&Arc` we would need to perform two `clone`s. Does that make sense to you?

src/vmm/src/builder.rs
bchalios: Here it would be doable, if `StartMicrovmError` implemented `PartialEq`

src/vmm/src/builder.rs
bchalios: Not sure if it's doable but it might be just me not understanding really well trait objects. The compiler complains to me that `dyn devices::BusDevice + 'static` does not implement `Debug`

src/vmm/src/vmm_config/entropy.rs
bchalios: done

src/vmm/src/resources.rs
bchalios: I tried to. I need to go derive `thiserror::Error` in all variants. I don't think this is part of this PR.

src/vmm/src/resources.rs
bchalios: Can't do, because `EntropyDeviceBuilder` does not implement `Debug`. I tried to add it but then I need to derive `Debug` for `Entropy` which includes other objects that do not implement it and I'm not touching in this PR, so I'd like to avoid doing it.

bchalios: ... in this PR

src/vmm/src/rpc_interface.rs
bchalios: same as before, I'd like to but, I can't. I 'd need to change all the variants of the enum to derive `thiserror::Error`

src/vmm/src/rpc_interface.rs
bchalios: done

src/vmm/src/rpc_interface.rs
bchalios: done

src/vmm/src/vmm_config/entropy.rs
bchalios: yes, after this one gets merged I'll open another PR that will actually add more variants.

src/vmm/src/resources.rs
bchalios: done

src/devices/src/virtio/mod.rs
JonathanWoollett-Light: I'm okay with it not being a part of this PR, in this case I would think it useful to create an issue on transiting to using the rust-vmm bindings instead of our own.

src/devices/src/virtio/rng/device.rs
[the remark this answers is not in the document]
JonathanWoollett-Light: True, it is a relatively small matter of opinion. I don't mind either way.

src/vmm/src/device_manager/persist.rs
bchalios: Ok, so I applied this change (here and all other places you spotted) but it breaks some unit-tests. Apparently when you serialize to json `EntropyDeviceConfig` results to `None` whereas `EntropyDeviceConfig {}` to `{}`.

Since in the next PR I will populate this struct what I did was define the struct like this:

```
#[derive(Default)]
struct EntropyDeviceConfig {}
```

and the use `EntropyDeviceConfig::default()` which is slightly nicer than `EntropyDeviceConfig {}`

src/devices/src/virtio/rng/device.rs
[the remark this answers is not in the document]
JonathanWoollett-Light: Okay.

src/devices/src/virtio/rng/device.rs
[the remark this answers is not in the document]
bchalios: I actually, I ended up trying to add it (see later comment) but the `DeviceState` member does not include it, so I would need to go change structs that are not related with this PR.

src/api_server/src/request/entropy.rs
JonathanWoollett-Light: Resolving given https://github.com/firecracker-microvm/firecracker/pull/3254#discussion_r1060553675.

src/api_server/src/request/entropy.rs
JonathanWoollett-Light: Resolving given https://github.com/firecracker-microvm/firecracker/pull/3254#discussion_r1060553675.

src/devices/src/virtio/mod.rs
bchalios: Before opening an issue, shall we try to figure out whether there is any issue with versioning?

src/devices/src/virtio/rng/persist.rs
```diff
@@ -0,0 +1,106 @@
...
+        EntropyState {
+            virtio_state: VirtioDeviceState::from_device(self),
+        }
+    }
+
+    fn restore(
+        constructor_args: Self::ConstructorArgs,
+        state: &Self::State,
+    ) -> std::result::Result<Self, Self::Error> {
+        // It's ok to unwrap here. We are restoring from an already created
```
roypat, line 49: Nit: Outdated comment? (there's no unwrap I can see)

src/vmm/src/device_manager/mmio.rs
```diff
@@ -448,6 +448,13 @@ impl MMIODeviceManager {
                     // Any in-flight packets or events are simply lost.
                     // Vsock is restored 'empty'.
                 }
+                TYPE_RNG => {
+                    let entropy = virtio.as_mut_any().downcast_mut::<Entropy>().unwrap();
+                    if entropy.is_activated() {
+                        info!("kick entropy {id}.");
+                        entropy.process_virtio_queues();
+                    }
+                }
```
roypat, line 457: I feel like this entire `match` block can be replaced by proper dynamic dispatch if we just put `is_actived` and `process_virtio_queues` into the `VirtioDevice` trait, but that's outside the scope of this PR

src/vmm/src/device_manager/persist.rs
```diff
@@ -193,6 +214,16 @@ impl DeviceStates {
 
         Ok(())
     }
+
+    fn entropy_serialize(&mut self, target_version: u16) -> VersionizeResult<()> {
+        if target_version < 4 && self.entropy_device.is_some() {
```
roypat, line 219: If there some constant we can refer to for this magic number?

src/vmm/src/device_manager/persist.rs
```diff
@@ -313,6 +345,20 @@ impl<'a> Persist<'a> for MMIODeviceManager {
                         device_info: device_info.clone(),
                     });
                 }
+                TYPE_RNG => {
+                    let entropy = locked_device
+                        .as_mut_any()
+                        // Currently, VsockUnixBackend is the only implementation of VsockBackend.
```
roypat, line 351: Comment copied from vsock case xP

src/devices/src/virtio/mod.rs
JonathanWoollett-Light: I think questions/concerns like this might reasonably be discussed under such an issue.

src/api_server/src/parsed_request.rs
JonathanWoollett-Light: Okay.

JonathanWoollett-Light:
Commit titles can use conventional commit tags like:
- [test: virtio: add a test helper for virtio devices](https://github.com/firecracker-microvm/firecracker/pull/3254/commits/9311eeef8944b84b4aef043134469694c3d185b2)
- [build: entropy: bindgen for virtio_rng](https://github.com/firecracker-microvm/firecracker/pull/3254/commits/fa3d5611c44ef4c1205f34a69a6aa195433ceb43)
- [feat: entropy: add support for virtio-rng](https://github.com/firecracker-microvm/firecracker/pull/3254/commits/b28ad1098938740a1b1387f02a95758182740d9a)
- [feat: entropy: plug in with API](https://github.com/firecracker-microvm/firecracker/pull/3254/commits/ee1ba35330b456e049de93773771bbeb3a163872)
- [feat: entropy: snapshot the device](https://github.com/firecracker-microvm/firecracker/pull/3254/commits/f43ff99e9649b41b72439b46247994c987d39c8a)
- [feat: entropy: add metrics to device](https://github.com/firecracker-microvm/firecracker/pull/3254/commits/026c678a713b3d35bfb3c6abeabcefbde471d336)
- [test: entropy: fixes to existing integration integration](https://github.com/firecracker-microvm/firecracker/pull/3254/commits/56b590074a283304e653de8d608437e80fe24e3b)
- [test: entropy: adjust firecracker binary size for ARM64](https://github.com/firecracker-microvm/firecracker/pull/3254/commits/7cd92d32d9e0bc1e6d275a57a090c9023d02dde0)
- [test: entropy: update code coverage](https://github.com/firecracker-microvm/firecracker/pull/3254/commits/45ccdc7df803eccecb5a84f61f8b9c1375194e2c)

src/devices/src/virtio/rng/device.rs
```diff
@@ -110,19 +114,23 @@ impl Entropy {
         let mut used_any = false;
         while let Some(desc) = self.queues[RNG_QUEUE].pop(mem) {
             debug!("entropy: guest request for {} bytes of entropy", desc.len);
```
roypat, line 116: I should've left this comment on an earlier commit, but I don't think this debug message is correct. `desc` is the descriptor chain head, so `len` is the length of the buffer to which the head descriptor points. However, if the chain has length longer than one, then the total number of random bytes requested is higher than `desc.len`, as its will be the sum of all `.len` of all descriptors in the chain, according to your `handle_one` impl.

src/devices/src/virtio/rng/device.rs
```diff
@@ -110,19 +114,23 @@ impl Entropy {
...
             let bytes = self.handle_one(mem, desc).unwrap_or_else(|err| {
                 error!("entropy: {err}");
+                METRICS.entropy.entropy_event_fails.inc();
                 0
             });
 
             match self.queues[RNG_QUEUE].add_used(mem, index, bytes) {
                 Ok(_) => {
                     used_any = true;
+                    METRICS.entropy.entropy_bytes.add(bytes as usize);
```
roypat, line 129: To you see this metric as "total bytes of random data consumed from host" or "total number of random bytes offered to guest"? If its the former, then the metric should be incremented inside of `handle_one` due to the fallible nature of that method.

src/vmm/src/resources.rs
```diff
@@ -1579,5 +1611,15 @@ mod tests {
...
+        assert_eq!(
+            Error::EntropyDevice(EntropyDeviceError::CreateDevice(
+                EntropyError::ReadOnlyDescriptor
+            ))
+            .to_string(),
+            format!(
+                "Entropy device error: {}",
+                EntropyDeviceError::CreateDevice(EntropyError::ReadOnlyDescriptor)
+            )
+        );
```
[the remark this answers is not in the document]
bchalios: I _think_ that these have been implemented because when we used kcov, the report would show that the `Debug` or `Display` was not covered by tests. So this is a way to increase code coverage. Not sure if this is still the case with grcov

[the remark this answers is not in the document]
JonathanWoollett-Light: I believe this is still the case with `grcov`, this cannot be avoided until https://github.com/rust-lang/rust/issues/84605 is closed and `#[no_coverage]` is stabilized, and even then this would only help if these derived implementations used `#[no_coverage]` which would depend on `std`.

src/devices/src/virtio/test_utils.rs
```diff
@@ -329,3 +329,175 @@ impl<'a> VirtQueue<'a> {
...
+                desc.set(addr, len, flags, 0);
+                if let Some(&&(next_index, _, _)) = iter.peek() {
+                    desc.flags.set(flags | VIRTQ_DESC_F_NEXT);
+                    desc.next.set(next_index);
+                }
+
+                addr += u64::from(len);
+                // Add small random gaps between descriptor addresses in order to make sure we
+                // don't blindly read contiguous memory.
+                addr += u64::from(utils::rand::xor_pseudo_rng_u32()) % 10;
```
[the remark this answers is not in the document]
bchalios: Hmmm, I _think_ you are right, but we somehow need to be able to "emulate" the driver's behaviour. The spec allows buffers to not be in contiguous memory and it very often happens. So, I like the idea that are tests don't assume otherwise.

src/devices/src/virtio/rng/device.rs
```diff
@@ -0,0 +1,343 @@
...
+
+pub struct Entropy {
+    // VirtIO fields
+    avail_features: u64,
+    acked_features: u64,
+    activate_event: EventFd,
+
+    // Transport fields
+    device_state: DeviceState,
+    queues: Vec<Queue>,
```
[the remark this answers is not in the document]
bchalios: We discussed offline with @roypat that this will be extended with more queues.

src/vmm/src/device_manager/persist.rs
bchalios: It doesn't look like. The number is just +1 of what was the previous maximum. The whole versionizing thingy needs a better design I think.

src/devices/src/virtio/rng/device.rs
bchalios: You are right. I have fixed this by introducing a helper type that parses the `DescriptorChain` into a something that derefs into slice of slice[u8]. We already use that (for other reasons) in the net device. I think it fixes the issue.

src/devices/src/virtio/rng/device.rs
bchalios: see previous comment.

src/devices/src/virtio/test_utils.rs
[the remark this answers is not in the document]
roypat: Maybe we could seed a PRNG with some other value from the unit test?

Although I also saw the other day that this randomness is also present in other tests, so am happy to just address it outside of this PR

src/vmm/src/resources.rs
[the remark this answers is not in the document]
roypat: How many lines will be counted as uncovered? I thought it was only the actual `#[derive(Debug)]` that'd count as uncovered, not the entire implementation. Because in that case imo it'd be justifiable to remove these noisy tests and accept a slight drop in coverage (or, just pass `#[derive` to grcov's `--excl-line` option). 

Either way, also out of scope for this PR.

src/vmm/src/device_manager/persist.rs
roypat: Ah, yeah for some reason I thought it mapped to one of the snapshot versions. Forgot we had separate versions for each struct. Maybe those should've been tied to  each other, but that ship has long sailed I guess.

src/devices/src/virtio/rng/device.rs
roypat: "number of bytes offered to guest" it is 😆

tests/framework/dependencies.txt
```diff
@@ -55,8 +55,11 @@
  'opaque-debug',
  'peeking_take_while',
  'polyval',
+ 'ppv-lite86',
  'proc-macro2',
  'quote',
+ 'rand',
+ 'rand_chacha',
```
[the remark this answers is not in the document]
bchalios: Hmmm, I wouldn't mind switching but:

The `rand_core::Error` doesn't seem to implement `AsDynError` and `StdError` so it does error handling a tiny bit more awkward.

and looking in the crate's documentation:

> This crate is mainly of interest to crates publishing implementations of [RngCore](https://docs.rs/rand_core/latest /rand_core/trait.RngCore.html). 
> Other users are encouraged to use the [rand](https://docs.rs/rand) crate instead which re-exports the main traits > and error types.

so, maybe it's worth keep using `rand` instead?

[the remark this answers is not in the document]
roypat: Ah, I missed that we also use the Error type. Yeah, let's not inconvenience ourselves. Saw the other day that we'll seemingly pull in `rand` via https://github.com/firecracker-microvm/firecracker/tree/feature/cpu-templates anyway

src/vmm/src/resources.rs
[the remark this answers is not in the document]
roypat: After discussion with @dpopa in https://github.com/firecracker-microvm/firecracker/pull/3251#discussion_r1072328768 (and offline), we do want to remove these tests whenever we switch a manual Display implementation to a thiserror generated one. So this one should go with 904b612eb6975e19340b64c0ba6f923ce5d04c3d

src/vmm/src/resources.rs
JonathanWoollett-Light: We could avoid this in the future by using [`#![warn(missing_debug_implementations)]`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/builtin/static.MISSING_DEBUG_IMPLEMENTATIONS.html), maybe it is worth creating an issue for implementing this lint?

src/vmm/src/rpc_interface.rs
JonathanWoollett-Light: Resolving this as duplicate of https://github.com/firecracker-microvm/firecracker/pull/3254#discussion_r1087706349

src/vmm/src/rpc_interface.rs
JonathanWoollett-Light: Would it be worth creating an issue for this, maybe ``Ensure error types have descriptive `Display` implementations``?

src/vmm/src/resources.rs
bchalios: Not sure we want to commit to create debug implementations for all of our types.

src/devices/src/virtio/rng/device.rs
```diff
@@ -0,0 +1,343 @@
...
+        for desc in head {
+            // We should only receive write-only descriptors
+            if !desc.is_write_only() {
+                return Err(Error::ReadOnlyDescriptor);
+            }
+
+            let ptr = mem.get_slice(desc.addr, desc.len as usize)?.as_ptr();
+            // SAFETY: This is safe since we get `ptr` from `get_slice` which also checks the length
+            // of the descriptor
+            let slice = unsafe { std::slice::from_raw_parts_mut(ptr, desc.len as usize) };
```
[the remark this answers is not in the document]
bchalios: This is now being called only from the `IoVecBuffer` and `IoVecBufferMut` implementations so I added the missing documentation on the `src/device/src/virtio/iovec.rs` call sites.

src/vmm/src/resources.rs
JonathanWoollett-Light: I think we should commit to fixing this where we encounter it, or we need to commit such that don't encounter it. Otherwise I think every PR which touches these types will have the same reasoning that they don't think fixing them fits in their PR so will not test as thoroughly.

src/vmm/src/rpc_interface.rs
bchalios: I guess that this: https://github.com/firecracker-microvm/firecracker/issues/3278 would address it.

src/vmm/src/resources.rs
bchalios: I think that this https://github.com/firecracker-microvm/firecracker/issues/3278 will address it

src/vmm/src/rpc_interface.rs
JonathanWoollett-Light: I think https://github.com/firecracker-microvm/firecracker/issues/3278 is different, this type already implements `thiserror::Error` so the issue wouldn't entail any action here.

src/vmm/src/vmm_config/entropy.rs
bchalios: I find `assert_eq` needlessly verbose, so, if you don't mind, I'd rather keep it as is here.

src/vmm/src/vmm_config/entropy.rs
bchalios: I find `assert_eq` needlessly verbose, so, if you don't mind, I'd rather keep it as is here.

src/vmm/src/vmm_config/entropy.rs
bchalios: This would require to `Entropy` to be `PartialEq`, right? If that's the case, I don't think it's doable. There are various fields that are not `PartialEq`.

src/vmm/src/vmm_config/entropy.rs
bchalios: done!

src/vmm/src/vmm_config/entropy.rs
bchalios: Same here. I think that `Entropy` would need to be `PartialEq`.

src/devices/src/virtio/rng/persist.rs
bchalios: We can only derive `Versionize` on structs with named fields.

src/vmm/src/resources.rs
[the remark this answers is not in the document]
bchalios: Hmmm, should I drop the whole test?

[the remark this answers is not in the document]
roypat: Yeah, since the code it was "testing" isnt there anymore, I'd say so

roypat (approved):

src/vmm/src/rpc_interface.rs
bchalios: Ah sorry, I got confused with the other case, where included types did not implement `thiserror::Error`. So, if you check the message printed by the including type is already quite descriptive I think.

JonathanWoollett-Light (approved):

roypat (approved):

src/devices/src/virtio/iovec.rs
```diff
@@ -0,0 +1,535 @@
...
+            let buf_ptr = buf[buf_start..buf_end].as_mut_ptr();
+
+            // SAFETY:
+            // The call to `std::ptr::add` is safe because `seg.iov_base` is a valid pointer (it's
+            // the pointer to a valid guest memory region (`GuestMemoryMmap` implementation checked
+            // its boundaries) and `start` is less than `seg.iov_len`.
+            //
+            // The call to `copy_nonoverlapping` is safe because:
+            // 1. `buf_ptr` is a pointer valid for writing `buf_end - buf_start + 1` bytes.
+            // 2. `seg_ptr` is a pointer valid for reading `buf_end - buf_start + 1` bytes.
```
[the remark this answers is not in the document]
roypat: dealt with inhttps://github.com/firecracker-microvm/firecracker/pull/3406, feature branch will pick it up on next rebase

src/devices/src/virtio/iovec.rs
```diff
@@ -0,0 +1,535 @@
...
+            let buf_ptr = buf[buf_start..buf_end].as_ptr();
+
+            // SAFETY:
+            // The call to `std::ptr::add` is safe because `seg.iov_base` is a valid pointer (it's
+            // the pointer to a valid guest memory region (`GuestMemoryMmap` implementation checked
+            // its boundaries) and `start` is less than `seg.iov_len`.
+            //
+            // The call to `copy_nonoverlapping` is safe because:
+            // 1. `buf_ptr` is a pointer valid for writing `buf_end - buf_start + 1` bytes.
+            // 2. `seg_ptr` is a pointer valid for reading `buf_end - buf_start + 1` bytes.
```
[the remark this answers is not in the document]
roypat: see above

src/devices/src/virtio/iovec.rs
```diff
@@ -0,0 +1,535 @@
...
+    /// Writes a number of bytes into the `IoVecBufferMut` starting at a given offset.
+    ///
+    /// This will try to fill `IoVecBufferMut` writing bytes from the `buf` starting from
+    /// the given offset. It will write as many bytes from `buf` as they fit inside the
+    /// `IoVecBufferMut` starting from `offset`.
+    ///
+    /// # Returns
+    ///
+    /// The number of bytes written (if any)
+    pub fn write_at(&mut self, buf: &[u8], offset: usize) -> Option<usize> {
```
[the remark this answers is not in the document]
bchalios: Sorry, for not responding to this, earlier. I actually like the idea. I had a commit somewhere that actually did something like this but got lost at some rebase. I 'll adapt it and push it. Please don't merge.

roypat (approved):

src/vmm/src/rpc_interface.rs
```diff
@@ -512,6 +499,12 @@ impl<'a> PrebootApiController<'a> {
             .map_err(VmmActionError::VsockConfig)
     }
 
+    fn set_entropy_device(&mut self, cfg: EntropyDeviceConfig) -> ActionResult {
+        self.boot_path = true;
+        self.vm_resources.build_entropy_device(cfg)?;
+        Ok(VmmData::Empty)
+    }
```
JonathanWoollett-Light, line 506: Very minor nit:
```suggestion
    fn set_entropy_device(&mut self, cfg: EntropyDeviceConfig) -> ActionResult {
        self.boot_path = true;
        self.vm_resources.build_entropy_device(cfg).map(|_| VmmData::Empty)
    }
```

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// A builder type used to construct an Entropy device
+#[derive(Default)]
+pub struct EntropyDeviceBuilder(Option<Arc<Mutex<Entropy>>>);
+
+impl EntropyDeviceBuilder {
+    /// Create a new instance for the builder
+    pub fn new() -> Self {
+        Self(None)
+    }
```
JonathanWoollett-Light, line 41: This is the same as `EntropyDeviceBuilder::default()` would it make sense to remove this duplicated code?

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -0,0 +1,98 @@
...
+        self.0 = Some(dev.clone());
+
+        Ok(dev)
+    }
+
+    /// Insert a new entropy device from a configuration object
+    pub fn insert(&mut self, config: EntropyDeviceConfig) -> Result<()> {
+        let _ = self.build(config)?;
+        Ok(())
+    }
```
JonathanWoollett-Light, line 55: Nit:
```suggestion
    pub fn insert(&mut self, config: EntropyDeviceConfig) -> Result<()> {
        self.build(config).map(drop)
    }
```

JonathanWoollett-Light (approved):

src/vmm/src/vmm_config/entropy.rs
bchalios: As Patrick mentioned in one of the previous comments, I think that all of this infrastructure that parses and creates devices from config needs re-work, so I'd rather follow the same pattern as other devices and just fix all of this in a later PR. What do you think?

src/vmm/src/rpc_interface.rs
bchalios: I don't think that works, `build_entropy_device` does not return the same Error type, so it doesn't know how to handle it without the `?` thingy. Am I missing something?

src/vmm/src/vmm_config/entropy.rs
bchalios: same as before!

PR #3264, opened 2022-11-15
Extract duplicate code into a method for Cmdline substring checking

andreitraistaru:
Signed-off-by: Traistaru Andrei Cristian <atc@amazon.com>

## Reason

In the context of backporting linux-loader v0.7.0 consuming in Firecracker 1.0/1.1,
we enhanced the code in these older versions of Firecracker by applying this change. As
a result, it makes sense to apply it on main as well.

## License Acceptance

## PR Checklist

---


roypat:
I feel like there is something that can be done on rust-vmm's side to make things like this easier. Looking at the code for [`CmdLine`](https://github.com/rust-vmm/linux-loader/blob/main/src/cmdline/mod.rs), it seems that the parameter lists are stored as strings, so we shouldn't need to go through `CString` just to go back to `String`. 

Maybe we can propose `CmdLine::contains` in `linux-loader`?

src/vmm/src/builder.rs
```diff
@@ -1088,6 +1088,15 @@ pub mod tests {
...
 
+    fn cmdline_contains(cmdline: &Cmdline, slug: &str) -> bool {
+        cmdline
+            .as_cstring()
+            .unwrap()
+            .into_string()
+            .unwrap()
+            .contains(slug)
+    }
+
```
pb8o, line 1099: LGTM. Would there be value in adding this as a method in `Cmdline`?

pb8o (approved):

src/vmm/src/builder.rs
andreitraistaru: From my point of view there is no real value in adding this in `Cmdline` on rust-vmm side as long as there are only a couple of places where we use this. And in each of this places, we use this just for unit testing purposes.

andreitraistaru:
From my point of view, it's not necessarily the case to add this to rust-vmm as long as there is no other usecase for this. `Cmdline` is meant to be a builder for the kernel cmdline in which the user generally adds parameters and gets the result at the end. As far as I know, rust-vmm wants to keep its code as light as possible and adds new functionality only if there is a strong request for this.

src/vmm/src/builder.rs
andreitraistaru: Closing this for unblocking merge of the PR; the discussion about rust-vmm upstreaming of this method will continue on the main thread, however.

JonathanWoollett-Light, line 1099: We can remove the `unwrap()`s so if they where `Err` it would panic on the specific assertion line rather than in this function, giving more information on what broke.
```suggestion
    fn cmdline_contains(cmdline: &Cmdline, slug: &str) -> bool {
        if let Ok(cstring) = cmdline.as_cstring() {
            if let Ok(string) = cstring.into_string() {
                if string.contains(slig) {
                    return true;
                }
            }
        }
        false
    }
```

roypat (approved):

src/vmm/src/builder.rs
andreitraistaru: Yeah, I agree with @roypat . Closing this comment to unblock the PR merge.

JonathanWoollett-Light: In the case an unwrap will never fail it should be commented.

andreitraistaru: I will open a new PR for adding the comment as suggested! Sorry for hurry-up with the PR, it wasn't intentional.

Issue #3221, opened 2022-10-28
`cargo build` produces `x86_64` binary on ARM


roypat:
# Describe the bug

Since we globally specify the `x86_64-unknown-linux-musl` target, plain `cargo build` (e.g. when building firecracker _without_ the `devtool build` wrapper. `devtool build` will always set the correct target) will always produce a `x86_64` binary, even if that binary cannot be executed on the host platform (e.g. on ARM).

## To Reproduce

run `cargo build` on non-`x86_64` machine (you will have to install quite a few dependencies to get the cross-compilation to actually work)

## Expected behaviour

`cargo build` produces a binary that is compatible with the host architecture. Particularly, we want it to compile for `x86_64-unknown-linux-musl` on `x86_64` linux, and for `aarch64-unknown-linux-musl` on ARM linux. 

## Environment

N/A

## Additional context

This stems from the fact that cargo does not allow us to specify targets dependent on host architecture. On option would be to remove the `target` directive in `.cargo/config` completely, although this would also change the behavior on `x86_64` (it would produce a binary with glibc ABI instead of musl, which is not desirable). 

It might not be possible to achieve the desired behavior without changes to cargo 

## Checks

- [x] Have you searched the Firecracker Issues database for similar problems?
- [x] Have you read the existing relevant Firecracker documentation?
- [x] Are you certain the bug being reported is a Firecracker issue?

ls-ivan-kiselev:
> it would produce a binary with glibc ABI instead of musl, which is not desirable

Does it compile to `x86_64-unknown-linux-musl` by default just for the static linking and avoiding system dependencies or is there some other reason?

roypat:
This was fixed in https://github.com/firecracker-microvm/firecracker/pull/4121

PR #3265, opened 2022-11-15
Selecting targer arch for Firecracker based on the host arch

ShadowCurse:
## Changes
Modified `.cargo/config` to properly select build tartget depending on the host architecture.
So on `x86` -> `86_64-unknown-linux-musl`
and on `aarch64` -> `aarch64-unknown-linux-musl`
...

## Reason
Fixes: #3221
...

## License Acceptance

## PR Checklist

---

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,9 +1,10 @@
-[build.'cfg(target_arch = "x86")']
+[build]
 target-dir = "build/cargo_target"
+
+[build.'cfg(target_arch = "x86")']
 target = "x86_64-unknown-linux-musl"
 
 [build.'cfg(target_arch = "aarch64")']
-target-dir = "build/cargo_target"
 target = "aarch64-unknown-linux-musl"
 
 [net]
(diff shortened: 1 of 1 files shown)


roypat:
Wohooo, good job! I totally did not know you could use cfg inside `.cargo/config`. We should consider if we want to revert the fix introduced [here](https://github.com/firecracker-microvm/firecracker/commit/dd1a77ef5c7c00cf5cff60ddf469406aa2228e59) now that we can rely on the default target being appropriate.

Also, could you squash your commits into one?

ShadowCurse:
So after some investigation it appears that this approach does not work.
Cargo just gives warnings about unused keys:
```
warning: unused config key `build.cfg(target_arch = "x86_64")` in ...`
warning: unused config key `build.cfg(target_arch = "aarch64")` in ...`
```

It appears that `cfg(target_arch = "x86_64")` can not be used in `[build]` section of the config
So this is incorrect:
```toml
[build.'cfg(target_arch = "x86_64")']
```

Such specification can only be used in `[target]` section.

On this note it make sense to close this PR.

PR #3274, opened 2022-11-18
api_server: fix error propagating

dianpopa:
The old IOError and EventFdError in the api_server's enum Error are not actually used.
They have been at some point used but we later introduced the logci where bind_and_run would call std::proces::exit on errors on HttpServer creation so in reality the IOError and EventFD error would not be thrown ever.

There are still some things to figure out: bind_and_run will not return any type of error except for the one we add in this commit. All the error processing happens inside bind and run in reality.

## Changes

...

## Reason

...

## License Acceptance

## PR Checklist

---

### DIFF
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -14,7 +14,6 @@ mod request;
 
 use std::path::PathBuf;
 use std::sync::mpsc;
-use std::{fmt, io};
 
 use logger::{
     debug, error, info, update_metric_with_elapsed_time, warn, ProcessTimeReporter, METRICS,
@@ -30,6 +29,7 @@ use vmm::rpc_interface::{VmmAction, VmmActionError, VmmData};
 use vmm::vmm_config::snapshot::SnapshotType;
 
 use crate::parsed_request::{ParsedRequest, RequestAction};
+use crate::Error::ServerCreation;
 
 /// Shorthand type for a request containing a boxed VmmAction.
 pub type ApiRequest = Box<VmmAction>;
@@ -37,23 +37,10 @@ pub type ApiRequest = Box<VmmAction>;
 pub type ApiResponse = Box<std::result::Result<VmmData, VmmActionError>>;
 
 /// Errors thrown when binding the API server to the socket path.
-#[derive(thiserror::Error)]
+#[derive(Debug)]
 pub enum Error {
-    /// IO related error.
-    #[error("IO error: {0}")]
-    Io(io::Error),
-    /// EventFD related error.
-    #[error("EventFd error: {0}")]
-    Eventfd(io::Error),
-}
-
-impl fmt::Debug for Error {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
-            Error::Io(ref err) => write!(f, "IO error: {}", err),
-            Error::Eventfd(ref err) => write!(f, "EventFd error: {}", err),
-        }
-    }
+    /// HTTP Server creation related error.
+    ServerCreation(ServerError),
 }
 
 type Result<T> = std::result::Result<T, Error>;
@@ -137,7 +124,7 @@ impl ApiServer {
     ///     .spawn(move || {
     ///         ApiServer::new(api_request_sender, vmm_response_receiver, to_vmm_fd)
     ///             .bind_and_run(
-    ///                 PathBuf::from(api_thread_path_to_socket),
+    ///                 &PathBuf::from(api_thread_path_to_socket),
     ///                 time_reporter,
     ///                 seccomp_filters.get("api").unwrap(),
     ///                 payload_limit,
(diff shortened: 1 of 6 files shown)


src/api_server/src/lib.rs
```diff
@@ -30,30 +29,18 @@ use vmm::rpc_interface::{VmmAction, VmmActionError, VmmData};
...
 use crate::parsed_request::{ParsedRequest, RequestAction};
+use crate::Error::ServerCreation;
 
 /// Shorthand type for a request containing a boxed VmmAction.
 pub type ApiRequest = Box<VmmAction>;
 /// Shorthand type for a response containing a boxed Result.
 pub type ApiResponse = Box<std::result::Result<VmmData, VmmActionError>>;
 
 /// Errors thrown when binding the API server to the socket path.
-#[derive(thiserror::Error)]
```
roypat, line 40: Since this is still an error `Enum`, shouldn't it still derive `thiserror::Error`?

src/firecracker/src/api_server_adapter.rs
```diff
@@ -149,25 +149,35 @@ pub(crate) fn run_with_api(
...
-                        api_server::Error::Io(inner)
-                    ),
-                    _ => panic!(
-                        "Failed to communicate with the API socket: {:?}",
-                        api_server::Error::Io(inner)
-                    ),
-                },
-                Err(eventfd_err @ api_server::Error::Eventfd(_)) => {
-                    panic!("Failed to open the API socket: {:?}", eventfd_err)
+                Err(api_server::Error::ServerCreation(ServerError::IOError(inner))) => {
```
JonathanWoollett-Light, line 159: It would be better if the thread returned a result rather than calling `std::process::exit`. This can then be propagated up through `run_with_api` and eventually back to main, calling `std::process::exit` is non-ideal (https://rust-lang.github.io/rust-clippy/master/#exit).

src/firecracker/src/api_server_adapter.rs
```diff
@@ -149,25 +149,35 @@ pub(crate) fn run_with_api(
...
+                                "Failed to bind and run the HTTP server: {}",
+                                ServerError::IOError(inner)
+                            );
+                            std::process::exit(vmm::FcExitCode::GenericError as i32);
+                        }
+                    }
+                }
+                Err(api_server::Error::ServerCreation(err)) => {
+                    error!("Failed to bind and run the HTTP server: {err}");
+                    std::process::exit(vmm::FcExitCode::GenericError as i32);
```
roypat, line 180: nit: de-duplicate the match arms, at the expense of a slightly longer line
```suggestion
                Err(api_server::Error::ServerCreation(ServerError::IOError(inner))) if inner.kind() == std::io::ErrorKind::AddrInUse => {
                    let sock_path = api_bind_path.display().to_string();
                    error!(
                        "Failed to open the API socket at: {sock_path}. Check that it is \
                         not already used."
                    );
                    std::process::exit(vmm::FcExitCode::GenericError as i32);
                }
                Err(api_server::Error::ServerCreation(err)) => {
                    error!("Failed to bind and run the HTTP server: {err}");
                    std::process::exit(vmm::FcExitCode::GenericError as i32);
```

roypat (approved):

src/firecracker/src/api_server_adapter.rs
dianpopa: This is something we can do as an overall improvement for the Firecracker code (make the std::process::exit get propagated to main) if possible (the std::process::exit was there before this PR) -> you can open an issue for this.

For the moment, this PR actually solves a bug in which some code would never get called and the error message triggered when the API socket cannot be opened is not descriptive of the problem.

src/api_server/src/lib.rs
dianpopa: The only error this enum is used only to signal that an error during server creation happened, but its value is actually never printed so no need to implement Debug or Display on it.
It is only used as an intermediate to hold the error that triggered the failure of `Http::new()`

JonathanWoollett-Light (dismissed):

dianpopa (commented):
@JonathanWoollett-Light @roypat please take another one, I applied a little suggestion

roypat (approved):

src/api_server/src/lib.rs
```diff
@@ -30,30 +29,18 @@ use vmm::rpc_interface::{VmmAction, VmmActionError, VmmData};
...
-
-impl fmt::Debug for Error {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match *self {
-            Error::Io(ref err) => write!(f, "IO error: {}", err),
-            Error::Eventfd(ref err) => write!(f, "EventFd error: {}", err),
-        }
-    }
+    /// HTTP Server creation related error.
+    ServerCreation(ServerError),
```
JonathanWoollett-Light, line 43: Nit: Errors types with a single variant should ideally be newtypes.
```rust
/// HTTP Server creation related error.
#[derive(Debug)]
pub struct ServerCreationError(ServerError)
```

JonathanWoollett-Light (approved):

src/api_server/src/lib.rs
dianpopa: I do plan on adding more errors in the future (you can notice that there is a `panic` below  when trying to apply the seccomp filters -> I think we need to get rid of and  propagate and handle it the same we do for errors triggered by `ServerCreation`).

Issue #750, opened 2018-12-07
Explicit error message when multiqueue tap device used as backend


mcastelino:
If firecracker is passed a multiqueue tap device created using

```
ip tuntap add name tap0 mode tap multi_queue
```

It reports an error

```
[PUT /network-interfaces/{iface_id}][400] putGuestNetworkInterfaceByIdBadRequest  &{FaultMessage:Cannot open TAP device. Invalid name/permissions. CreateTap(
Os { code: 22, kind: InvalidInput, message: Invalid argument })}
```

Does firecracker not support multi-queue tap devices?

alexandruag:
Hello, you are right, we do not currently support multi-queue TAP devices. We might at some point, because it's a relatively low hanging fruit, but there are no clear plans right now.

egernst:
Hey @alexandruag - pardon me if I'm missing; is it feasible to return an error here so it is clear?  In our case, we determined there's an error because we noticed firecracker was creating its own tap instead of utilizing ours.  Being more explicit may be helpful?

raduweiss:
On multi-queue tap devices, I've just read a bit of the docs to understand what they do. Now I'm wondering what the use case for them is. @egernst , what is Kata using it for?

mcastelino:
@raduweiss 

> On multi-queue tap devices, I've just read a bit of the docs to understand what they do. Now I'm wondering what the use case for them is. @egernst , what is Kata using it for?

We use multiqueue to achieve higher throughput. We scale up the number of queues based on the number of vCPUs till we reach the point of diminished returns. Inside Kata we enable multiple queues also on the interface.

mcastelino:
> Yea, providing friendlier, non-misleading error messages is definitely something we'll iterate through (PRs welcome :D). The part where Firecracker starts creating interfaces by itself when run with the appropriate capabilities is tricker to get rid of, due to how open/create works for TAP interfaces. We could check for the existence of the specified interface beforehand, but that's kinda awkward and adds overhead. Btw, why are you running Firecracker with elevated privileges? Vsock, I guess?

@alexandruag we pre-create the network interface and send it into firecracker. Yes we need vsock today as we need a method to setup a communication channel between the host and the agent we run within the VM. We are open to alternate mechanisms too. Currently we found the vsock is the one that is easiest for us to use with firecracker as kata already has support for the same. We also support virtio-serial, but firecracker currently does not have support for it.

alexandruag:
[...]

PR #3276, opened 2022-11-21
Better error message when multi queue net device is in use

ShadowCurse:
So the task is to introduce a proper error message when passing a tap device with multi-queue feature enabled to firecracker. The error itself appears when we try to `TUNSETIFF` on the tap device [here](https://github.com/firecracker-microvm/firecracker/blob/1644b3cb4947b927f29da795492284b4b10f2405/src/devices/src/virtio/net/tap.rs#L126).

- One way to check if a tap device is configured with multi-queue flag is to read `/sys/class/net/{}/tun_flags` file and check it for the flag. The problem is that it does not work when firecracker is run inside the `jailer`, because it can not access that file. 
- So there is another way to check for the flag. We can just try to create tap device twice: one time with usual flags and second time with multi-queue flag set. If the second call successes then we know that device has multi-queue enabled and we can return a proper error. But this solution seems clunky. 
- The last option is just to split existing IoctlError error into separate errors and make messages more explicit.

This PR implements last option.

## Changes
- Split `IoctlError` in `tap.rs` into separate errors and added more descriptive error messages
- Added a note to the network documentation about usage of multi-queue devices
- Added `thiserror::Error` derive to some errors to reduce boilerplate `std::fmt::Display` implementations

## Reason
Fixes: #750

## License Acceptance

## PR Checklist

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+
+- Improved error message when invalid network backend provided
+
 ## [1.2.0]
 
 ### Added
--- a/src/devices/src/virtio/net/tap.rs
+++ b/src/devices/src/virtio/net/tap.rs
@@ -19,16 +19,26 @@ use utils::{ioctl_ioc_nr, ioctl_iow_nr};
 const IFACE_NAME_MAX_LEN: usize = 16;
 
 /// List of errors the tap implementation can throw.
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum Error {
-    /// Unable to create tap interface.
-    CreateTap(IoError),
-    /// Invalid interface name.
-    InvalidIfname,
-    /// ioctl failed.
-    IoctlError(IoError),
-    /// Couldn't open /dev/net/tun.
+    /// Couldn't open /dev/net/tun
+    #[error("Couldn't open /dev/net/tun: {0}")]
     OpenTun(IoError),
+    /// Invalid interface name
+    #[error("Invalid interface name")]
+    InvalidIfname,
+    /// Error while creating ifreq structure
+    #[error(
+        "Error while creating ifreq structure: {0}. Invalid TUN/TAP Backend provided by {1}. \
+         Check our documentation on setting up the network devices"
+    )]
+    IfreqExecuteError(IoError, String),
+    /// Error while setting the offload flags
+    #[error("Error while setting the offload flags: {0}")]
+    SetOffloadFlags(IoError),
+    /// Error while setting size of the vnet header
+    #[error("Error while setting size of the vnet header: {0}")]
+    SetSizeOfVnetHdr(IoError),
 }
 
 pub type Result<T> = ::std::result::Result<T, Error>;
@@ -66,6 +76,7 @@ fn build_terminated_if_name(if_name: &str) -> Result<[u8; IFACE_NAME_MAX_LEN]> {
     Ok(terminated_if_name)
 }
 
+#[derive(Copy, Clone)]
 pub struct IfReqBuilder(ifreq);
 
 impl IfReqBuilder {
@@ -87,11 +98,10 @@ impl IfReqBuilder {
         self
     }
 
-    pub(crate) fn execute<F: AsRawFd>(mut self, socket: &F, ioctl: u64) -> Result<ifreq> {
+    pub(crate) fn execute<F: AsRawFd>(mut self, socket: &F, ioctl: u64) -> std::io::Result<ifreq> {
         // SAFETY: ioctl is safe. Called with a valid socket fd, and we check the return.
-        let ret = unsafe { ioctl_with_mut_ref(socket, ioctl, &mut self.0) };
-        if ret < 0 {
-            return Err(Error::IoctlError(IoError::last_os_error()));
+        if unsafe { ioctl_with_mut_ref(socket, ioctl, &mut self.0) } < 0 {
+            return Err(IoError::last_os_error());
         }
 
         Ok(self.0)
@@ -104,8 +114,6 @@ impl Tap {
     ///
     /// * `if_name` - the name of the interface.
     pub fn open_named(if_name: &str) -> Result<Tap> {
-        let terminated_if_name = build_terminated_if_name(if_name)?;
-
         // SAFETY: Open calls are safe because we give a constant null-terminated
         // string and verify the result.
         let fd = unsafe {
@@ -117,13 +125,16 @@ impl Tap {
         if fd < 0 {
             return Err(Error::OpenTun(IoError::last_os_error()));
         }
+
         // SAFETY: We just checked that the fd is valid.
         let tuntap = unsafe { File::from_raw_fd(fd) };
 
+        let terminated_if_name = build_terminated_if_name(if_name)?;
         let ifreq = IfReqBuilder::new()
             .if_name(&terminated_if_name)
             .flags((net_gen::IFF_TAP | net_gen::IFF_NO_PI | net_gen::IFF_VNET_HDR) as i16)
-            .execute(&tuntap, TUNSETIFF())?;
+            .execute(&tuntap, TUNSETIFF())
+            .map_err(|io_error| Error::IfreqExecuteError(io_error, if_name.to_owned()))?;
 
         Ok(Tap {
             tap_file: tuntap,
@@ -144,9 +155,8 @@ impl Tap {
     /// Set the offload flags for the tap interface.
     pub fn set_offload(&self, flags: c_uint) -> Result<()> {
         // SAFETY: ioctl is safe. Called with a valid tap fd, and we check the return.
-        let ret = unsafe { ioctl_with_val(&self.tap_file, TUNSETOFFLOAD(), c_ulong::from(flags)) };
-        if ret < 0 {
-            return Err(Error::IoctlError(IoError::last_os_error()));
+        if unsafe { ioctl_with_val(&self.tap_file, TUNSETOFFLOAD(), c_ulong::from(flags)) } < 0 {
+            return Err(Error::SetOffloadFlags(IoError::last_os_error()));
         }
 
         Ok(())
@@ -155,9 +165,8 @@ impl Tap {
     /// Set the size of the vnet hdr.
     pub fn set_vnet_hdr_size(&self, size: c_int) -> Result<()> {
         // SAFETY: ioctl is safe. Called with a valid tap fd, and we check the return.
-        let ret = unsafe { ioctl_with_ref(&self.tap_file, TUNSETVNETHDRSZ(), &size) };
-        if ret < 0 {
-            return Err(Error::IoctlError(IoError::last_os_error()));
+        if unsafe { ioctl_with_ref(&self.tap_file, TUNSETVNETHDRSZ(), &size) } < 0 {
+            return Err(Error::SetSizeOfVnetHdr(IoError::last_os_error()));
         }
 
         Ok(())
@@ -247,8 +256,14 @@ pub mod tests {
             tap_file: unsafe { File::from_raw_fd(-2) },
             if_name: [0x01; 16],
         };
-        assert!(faulty_tap.set_vnet_hdr_size(16).is_err());
-        assert!(faulty_tap.set_offload(0).is_err());
+        assert_eq!(
+            faulty_tap.set_vnet_hdr_size(16).unwrap_err().to_string(),
+            Error::SetSizeOfVnetHdr(IoError::from_raw_os_error(9)).to_string()
+        );
+        assert_eq!(
+            faulty_tap.set_offload(0).unwrap_err().to_string(),
+            Error::SetOffloadFlags(IoError::from_raw_os_error(9)).to_string()
+        );
     }
 
     #[test]
--- a/src/vmm/src/vmm_config/net.rs
+++ b/src/vmm/src/vmm_config/net.rs
@@ -3,8 +3,8 @@
 
 use std::convert::TryInto;
 use std::ops::Deref;
+use std::result;
 use std::sync::{Arc, Mutex};
-use std::{fmt, result};
 
 use devices::virtio::net::TapError;
 use devices::virtio::Net;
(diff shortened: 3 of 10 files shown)


tests/integration_tests/functional/test_net_multi_queue_unsupported.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pb8o, line 1: 2022

src/devices/src/virtio/net/tap.rs
```diff
@@ -48,7 +48,6 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 
 const TUNTAP: ::std::os::raw::c_uint = 84;
 ioctl_iow_nr!(TUNSETIFF, TUNTAP, 202, ::std::os::raw::c_int);
-ioctl_ior_nr!(TUNGETFEATURES, TUNTAP, 207, ::std::os::raw::c_uint);
```
pb8o: ok this is later removed in another commit. Could this be rebased?

src/devices/src/virtio/net/tap.rs
```diff
@@ -162,6 +179,20 @@ impl Tap {
...
+        let tun_flags_str = std::fs::read_to_string(path).map_err(Error::ReadTunFlags)?;
+        let tun_flags = u32::from_str_radix(tun_flags_str.trim().trim_start_matches("0x"), 16)
+            .map_err(Error::ConvertStrToHex)?;
+        if tun_flags & net_gen::IFF_MULTI_QUEUE != 0 {
+            Err(Error::MultiQueueUnsupported(if_name.into()))
+        } else {
+            Ok(())
+        }
+    }
 }
```
pb8o, line 185: This is then removed on the 4th commit. Could this be rebased?

src/devices/src/virtio/net/tap.rs
```diff
@@ -66,6 +85,7 @@ fn build_terminated_if_name(if_name: &str) -> Result<[u8; IFACE_NAME_MAX_LEN]> {
     Ok(terminated_if_name)
 }
 
+#[derive(Copy, Clone)]
```
JonathanWoollett-Light, line 88: Nit: I tend to favor deriving`Debug` on all types.

src/vmm/src/vmm_config/net.rs
```diff
@@ -61,45 +61,25 @@ pub struct NetworkInterfaceUpdateConfig {
...
     GuestMacAddressInUse(String),
-    /// Error during interface update (patch).
+    /// Error during interface update (patch)
+    #[error("Error during interface update (patch): {0}")]
     DeviceUpdate(VmmError),
-    /// Cannot open/create tap device.
+    /// Cannot open/create tap device
+    #[error("Cannot open/create tap device: {0}")]
     OpenTap(TapError),
 }
```
JonathanWoollett-Light, line 81: If you are deriving `thiserror::Error` you don't need `derive_more::From` you can simply place `#[from]` in front of the variants to you want to derive from for.
```suggestion
#[derive(Debug, thiserror::Error)]
pub enum NetworkInterfaceError {
    /// Could not create Network Device.
    #[error("Could not create Network Device: {0}")]
    CreateNetworkDevice(#[from] devices::virtio::net::Error),
    /// Failed to create a `RateLimiter` object.
    #[error("Failed to create a `RateLimiter` object: {0}")]
    CreateRateLimiter(#[from] std::io::Error),
    /// The MAC address is already in use.
    #[error("The MAC address is already in use: {0}")]
    GuestMacAddressInUse(String),
    /// Error during interface update (patch).
    #[error("Error during interface update (patch): {0}")]
    DeviceUpdate(#[from] VmmError),
    /// Cannot open/create tap device.
    #[error("Cannot open/create tap device: {0}")]
    OpenTap(#[from] TapError),
}
```

src/devices/src/virtio/net/tap.rs
ShadowCurse: I'd like to, but `ifreq` does not implement `Debug`, so can't derive

src/vmm/src/vmm_config/net.rs
ShadowCurse: Yup, forgot about that feature, added now

dianpopa (commented):
We might be able to find out whether the tap backend has been set up with multiqueue support with the TUNGETFEATURES ioctl.

```
ioctl(fd, TUNGETFEATURES, &features)
if (features & net_gen:: IFF_MULTI_QUEUE) {
return "No multiqueue support"
}
```
Could you try this out?

ShadowCurse:
> We might be able to find out whether the tap backend has been set up with multiqueue support with the TUNGETFEATURES ioctl.
> 
> ```
> ioctl(fd, TUNGETFEATURES, &features)
> if (features & net_gen:: IFF_MULTI_QUEUE) {
> return "No multiqueue support"
> }
> ```
> 
> Could you try this out?

I have tried `TUNGETFEATURE` and `TUNGETIFF`
- `TUNGETFEATURE` just gives the capabilities of physical network device and not the flags of some concrete net device passed to firecracker
- `TUNGETIFF` gives needed info about net device (including flags that we need to check), but it seems it can only be called after `TUNSETIFF`, so it is useless because we would fail at that step.

tests/integration_tests/functional/test_api.py
```diff
@@ -254,7 +254,7 @@ def test_net_api_put_update_pre_boot(test_microvm_with_api):
     )
     assert test_microvm.api_session.is_status_bad_request(response.status_code)
     assert (
-        "The guest MAC address {} is already in use.".format(guest_mac) in response.text
+        "The MAC address is already in use: {}".format(guest_mac) in response.text
```
JonathanWoollett-Light, line 257:
```suggestion
       f"The MAC address is already in use: {guest_mac}" in response.
```

src/devices/src/virtio/net/tap.rs
```diff
@@ -91,7 +99,7 @@ impl IfReqBuilder {
         // SAFETY: ioctl is safe. Called with a valid socket fd, and we check the return.
         let ret = unsafe { ioctl_with_mut_ref(socket, ioctl, &mut self.0) };
         if ret < 0 {
-            return Err(Error::IoctlError(IoError::last_os_error()));
+            return Err(Error::IfreqExecuteError(IoError::last_os_error()));
         }
```
JonathanWoollett-Light, line 103: The return value should ideally be checked before we could mistakenly access it.

This can be done using a scope like:
```rust
        // SAFETY: ioctl is safe. Called with a valid socket fd, and we check the return.
        let ret = {
            let x = unsafe { ioctl_with_mut_ref(socket, ioctl, &mut self.0) };
            if x < 0 {
                return Err(Error::IfreqExecuteError(IoError::last_os_error()));
            }
            x
        };
```
or more simply, a match:
```rust
        // SAFETY: ioctl is safe. Called with a valid socket fd, and we check the return.
        let ret = match unsafe { ioctl_with_mut_ref(socket, ioctl, &mut self.0) } {
            x @ 0.. => x,
            _ => return Err(Error::IfreqExecuteError(IoError::last_os_error()));
        };
```

src/devices/src/virtio/net/tap.rs
```diff
@@ -146,7 +154,7 @@ impl Tap {
         // SAFETY: ioctl is safe. Called with a valid tap fd, and we check the return.
         let ret = unsafe { ioctl_with_val(&self.tap_file, TUNSETOFFLOAD(), c_ulong::from(flags)) };
         if ret < 0 {
-            return Err(Error::IoctlError(IoError::last_os_error()));
+            return Err(Error::SetOffloadFlags(IoError::last_os_error()));
```
JonathanWoollett-Light, line 157: See https://github.com/firecracker-microvm/firecracker/pull/3276/files#r1029065821

src/devices/src/virtio/net/tap.rs
```diff
@@ -157,7 +165,7 @@ impl Tap {
         // SAFETY: ioctl is safe. Called with a valid tap fd, and we check the return.
         let ret = unsafe { ioctl_with_ref(&self.tap_file, TUNSETVNETHDRSZ(), &size) };
         if ret < 0 {
-            return Err(Error::IoctlError(IoError::last_os_error()));
+            return Err(Error::SetSizeOfVnetHdr(IoError::last_os_error()));
```
JonathanWoollett-Light, line 168: See https://github.com/firecracker-microvm/firecracker/pull/3276/files#r1029065821

src/devices/src/virtio/net/tap.rs
ShadowCurse: We don't use `ret` anyway so according to your suggestion it should be 
```rust
        // SAFETY: ioctl is safe. Called with a valid socket fd, and we check the return.
        let _ = match unsafe { ioctl_with_mut_ref(socket, ioctl, &mut self.0) } {
            x @ 0.. => x,
            _ => return Err(Error::IfreqExecuteError(IoError::last_os_error()));
        };
```
but this just adds boilerplate code.

I think current check looks fine.

ShadowCurse: Ok, 1st variant looks much better, keeps `unsafe` as small as possible.

tests/integration_tests/functional/test_net_multi_queue_unsupported.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Tests for the net device."""
+
+from framework import utils
```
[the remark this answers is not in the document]
ShadowCurse: moved it into `test_net.py`

src/devices/src/virtio/net/tap.rs
```diff
@@ -89,9 +100,8 @@ impl IfReqBuilder {
 
     pub(crate) fn execute<F: AsRawFd>(mut self, socket: &F, ioctl: u64) -> Result<ifreq> {
         // SAFETY: ioctl is safe. Called with a valid socket fd, and we check the return.
-        let ret = unsafe { ioctl_with_mut_ref(socket, ioctl, &mut self.0) };
-        if ret < 0 {
-            return Err(Error::IoctlError(IoError::last_os_error()));
+        if unsafe { ioctl_with_mut_ref(socket, ioctl, &mut self.0) } < 0 {
+            return Err(Error::IfreqExecuteError(IoError::last_os_error()));
```
dianpopa, line 104: I think there is space for improvement here. The TUNSETIFF ioctl returns EINVAL when a multiqueue tap backend has been configured and it also can return other various error types like (operation not permitted, bad file descriptor etc). So having a single message for all (i.e "check that your network device is configured properly") might actually be misleading. 

We could filter the type of error that the IOCTL returnes:
```
let err = IoError::last_os_error();
if err.kind() == std::io::ErrorKind::InvalidInput {
    return Error(IfReqInvalidArgument(err));
}
```
and the description of `IfReqInvalidArgument` could be something even more explicit, like:
`Invalid TUN/TAP Backend provided. Firecracker supports only TUN/TAP network backend with no multi queue support`.

dianpopa (commented):
Make sure to squash your commits

I would make 2 commits out of the existent 4:

* One with the actual new error type
* One with coverage update

src/devices/src/virtio/net/tap.rs
dianpopa: > Well, I think that if `TUNSETIFF` fails then there is definitely an issue with the configuration of the network device, so trying to catch only this specific one is incorrect. 

I think you misunderstood what I was proposing. I am not suggesting to catch only `InvalidInput`. I  was saying that in case of `InvalidInput` to explicitly let the user know what type of backend they should be configuring. 

> `InvalidInput` can also be caused by something else (even though in our case it is unlikely). Because of that I think having more generic error is better. It would be even better if there was some section in the documentation that states all possible net device configurations and I could add link to that section in the error message.

The initial issue we are trying to solve is to throw a proper error message when user configures a multi-queue backend, right? So, "having a more generic error" goes in the opposite direction IMO. I understand that a proper check would require too much effort for the gains and that is why I think that the least we can do is try to guide the user to a proper configuration of the backend. 

Your proposal to guide the user to our documentation is also a valid option IMO.
(Although, putting  a  link is not good practice since links are ephemerals -> locations change).
We can just say `Invalid TUN/TAP Backend provided. Check our documentation on setting up the network device (i.e network-setup).`
However, we do not mention in our net setup we do not support multi-queue. Would make sense for this issue to also be explicit about that there too, so that we avoid future inquiries in that direction.

ShadowCurse: @dianpopa  By saying `more generic error is better` I basically meant that the error should look like your new proposal: 
`Invalid TUN/TAP Backend provided. Check our documentation on setting up the network device (i.e network-setup)` 
rather then:
`Invalid TUN/TAP Backend provided. Firecracker supports only TUN/TAP network backend with no multi queue support`

@mattschlebusch I think the problem is not the `IoError::last_os_error()` itself, but the lack of the proper description around this error. 
Also I don't think changing `jailer` for this issue is a great idea. Who knows what security nuances it would bring.

So it seems we all agree that guiding user to the documentation is a reasonable solution here.
I think the phrase: 
`Invalid TUN/TAP Backend provided. Check our documentation on setting up the network device (i.e network-setup)`
 that @dianpopa proposed is fine.
Now the question is where to put the documentation and how to reference it in the error message.

dianpopa: We already have documentation on network setup (is found in docs/nework-setup.md).
But putting the location inside the error message is risky since that might change in the future. So, i think it suffices to leave the error message just referring to the doc.

src/vmm/src/vmm_config/net.rs
```diff
@@ -310,44 +292,23 @@ mod tests {
...
             NetworkInterfaceError::CreateNetworkDevice(devices::virtio::net::Error::TapOpen(
-                TapError::IoctlError(std::io::Error::from_raw_os_error(16))
+                TapError::IfreqExecuteError(std::io::Error::from_raw_os_error(16))
             ))
             .to_string()
         );
     }
 
-    #[test]
-    fn test_error_display() {
```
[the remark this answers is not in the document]
ShadowCurse: This test only tests if `Display` and `Debug` traits are implemented for the error type. So basically the only way this test would fail is when these traits are not implemented for the error, but the test would never run in this case, because it would not compile. So the test does not `test` for anything during runtime and will always be successful. In my opinion such test is worthless.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### Unreleased
+
+- Improved error message when invalid network backend provided
```
dianpopa, line 5: Would need to be


```
## Unreleased

### Changed

- Improved error message when invalid network backend provided.

```

dianpopa (dismissed):

src/vmm/src/vmm_config/net.rs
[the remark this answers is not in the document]
dianpopa: yes, I agree with @ShadowCurse that these type of tests are kind of superfluous.

src/devices/src/virtio/net/tap.rs
```diff
@@ -19,16 +19,26 @@ use utils::{ioctl_ioc_nr, ioctl_iow_nr};
...
 pub enum Error {
-    /// Unable to create tap interface.
-    CreateTap(IoError),
-    /// Invalid interface name.
-    InvalidIfname,
-    /// ioctl failed.
-    IoctlError(IoError),
-    /// Couldn't open /dev/net/tun.
+    /// Couldn't open /dev/net/tun
+    #[error("Couldn't open /dev/net/tun: {0}")]
```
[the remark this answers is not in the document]
mattschlebusch: It generates the message based on the parameter passed in. This error can be used quite generically though, and could include issues with the `/sys/class` devices, in which case nothing would be wrong with the cloning device (`/dev/net/tun`). The issue here is that we are treating this as a catch all for all issues.

For a generic/catch-all error, it would be more appropriate to pass in parameters as:
```
"Unable to open network device [{}]: {}"
```

The first parameter being the network device target, in this case the tap name, followed by the error.

src/vmm/src/vmm_config/net.rs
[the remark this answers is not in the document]
mattschlebusch: I could have clarified my point here better.

Given there are new error types being introduced, I would have expected some test code to make use of them. If you have a concern with the completeness of the tests that is completely fine. I maintain that it would be a bit better to extend rather than just delete.

With that said, testing that Display and Debug have been implemented is not a bad thing, but it isn't the point that I think is more important around addressing the coverage.

mattschlebusch (commented):
The root of the initial issue still centres around clarifying why the network device is not usable. 

The change here effectively changes the error message from 
```
Os { code: 22, kind: InvalidInput, message: Invalid argument })
``` 
to 
```
Error while creating ifreq structure: Invalid argument (os error 22).
Invalid TUN/TAP Backend provided. Check our documentation on setting
up the network devices.
```

The original issue is asking for an explicit error message as to why the error is occurring. If we are saying that Firecracker  cannot operationally accommodate this request for whatever reason (start-time performance, or security), then we should close issue [#750](https://github.com/firecracker-microvm/firecracker/issues/750) (as unresolved).

With all that said, that is not to say that we can't look at shipping this change as well as it is still friendlier.

src/devices/src/virtio/net/tap.rs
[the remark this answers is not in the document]
ShadowCurse: Well, it is not like we open different files here. We always open `/dev/net/tun` and through it we configure net devices. So we can get this error regardless of the net device name.

src/vmm/src/vmm_config/net.rs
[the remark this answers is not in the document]
ShadowCurse: Well, you can say that there are new error variants, but in reality I just split one generic error into several more specific. Actually there are tests for `SetOffloadFlags` and `SetSizeOfVnetHdr` errors [here](https://github.com/firecracker-microvm/firecracker/blob/3dd605ad678613294369c492b3966a469239a758/src/devices/src/virtio/net/tap.rs#L250) for example, they are just not easily visible (and I never touched them). 
So I don't see how to expand testing on this errors, because tests for them already existed. 
In my opinion it is better to have a lower coverage with actually useful tests, rather then high coverage with `Debug`, `Display` tests.

src/devices/src/virtio/net/tap.rs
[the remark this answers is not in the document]
mattschlebusch: `/dev/net/tun` is the cloning device that tunnels are created from. However the individualistic properties of the tap device is associated with the `/sys/class/net/...` device path. I think this is emphasized by the limitation of needing to query this device to find out if it is a multi-queue device for example.

To be a bit more specific, it's the host device name that would be useful to provide here, see the parameter passed in to the network configuration [described in the docs](https://github.com/firecracker-microvm/firecracker/blob/main/docs/network-setup.md?plain=1#L59).

src/vmm/src/vmm_config/net.rs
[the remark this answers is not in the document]
mattschlebusch: Can you confirm which tests make use of the new error types?

mattschlebusch (approved):
This error with the doc tweak is certainly better over what we had. I think it will just be fair to clarify on the associated issue that we are not able to be explicit about the reason for the configuration error.

PR #3280, opened 2022-11-22
Update rust vmm dependencies

roypat:
## Changes

Bump all rust-vmm dependencies to the newest release and specify their versions using caret requirements (we update all our other dependencies in #3248). The only crate where the version bump also introduces API changes is kvm-ioctl,  but these are minimal and this PR adjusts for them. 

## Reason

Recently, all rust-vmm crates received major version bumps to fix the way intra-rust-vmm dependencies are specified. See https://github.com/rust-vmm/community/issues/136#issuecomment-1314184934 for details

## License Acceptance

## PR Checklist

---

### DIFF
--- a/src/arch/src/aarch64/regs.rs
+++ b/src/arch/src/aarch64/regs.rs
@@ -10,9 +10,28 @@ use std::{fmt, fs, mem, result, u32};
 
 use kvm_bindings::*;
 use kvm_ioctls::VcpuFd;
+use versionize::*;
+use versionize_derive::Versionize;
 use vm_memory::GuestMemoryMmap;
 
 use super::get_fdt_addr;
+
+/// Struct describing a saved aarch64 register.
+///
+/// Used for interacting with `KVM_GET/SET_ONE_REG`.
+#[derive(Debug, Clone, Versionize, PartialEq, Eq)]
+pub struct Aarch64Register {
+    /// The KVM register ID.
+    ///
+    /// See https://docs.kernel.org/virt/kvm/api.html?highlight=kvm_set_one_reg#kvm-set-one-reg
+    pub id: u64,
+
+    /// The value of the register.
+    ///
+    /// 128 bit wide, as we want to restore the V0-V31 FP SIMD registers,
+    /// which are this wide.
+    pub value: u128,
+}
 
 /// Errors thrown while setting aarch64 registers.
 #[derive(Debug)]
@@ -160,10 +179,10 @@ arm64_sys_reg!(MIDR_EL1, 3, 0, 0, 0, 0);
 /// # Arguments
 ///
 /// * `state` - Array slice of kvm_one_reg structures, representing the registers of a VCPU state.
-pub fn get_manufacturer_id_from_state(state: &[kvm_one_reg]) -> Result<u32> {
+pub fn get_manufacturer_id_from_state(state: &[Aarch64Register]) -> Result<u32> {
     let midr_el1 = state.iter().find(|reg| reg.id == MIDR_EL1);
     match midr_el1 {
-        Some(value) => Ok(value.addr as u32 >> 24),
+        Some(register) => Ok(register.value as u32 >> 24),
         None => Err(Error::GetMidrEl1(
             "Failed to find MIDR_EL1 in vCPU state!".to_string(),
         )),
@@ -206,7 +225,7 @@ pub fn setup_boot_regs(
     let pstate = offset__of!(user_pt_regs, pstate) + kreg_off;
     vcpu.set_one_reg(
         arm64_core_reg_id!(KVM_REG_SIZE_U64, pstate),
-        PSTATE_FAULT_BITS_64,
+        PSTATE_FAULT_BITS_64.into(),
     )
     .map_err(|err| Error::SetCoreRegister(err, "processor state".to_string()))?;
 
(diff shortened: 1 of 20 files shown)


roypat:
Okay I am not quite sure what to do about the ARM register thing, I'll look into what is going on there.

roypat:
Okay I have adapted to the changes in kvm_ioctls by just casting our u64 to u128 when we use `set_one_reg`, and to use `try_into()` for converting the result of `get_one_reg` from u128 to u64. This introduces a new error variant, but realistically it should _never_ get instantiated as firecracker only interacts with 64 bit registers on ARM. Just casting our u64s to u128 works, I believe, only on little endian architectures, as kvm_ioctls passes a [pointer to the u128](https://github.com/rust-vmm/kvm-ioctls/commit/e6739aeebcc56dfea62216b2cc4cddd13d0a1182) to the kernel, and the kernel then inteprets this as a point to u32/u64/u128 depending on the register that is being accessed. Since firecracker does not support big endian, this should be fine. 

Alternatively, changing all of our register handling to u128 instead of u64 would change the serialization format of snapshots. 

I would really appreciate it if someone could validate this logic as part of the review of this PR, as I am not overly familiar with KVM/this part of firecracker yet.

aghecenco:
@roypat I think your reasoning is fine. AFAICT the kernel deals with an `u64` address: https://elixir.bootlin.com/linux/v5.10/source/include/uapi/linux/kvm.h#L1221

```c
struct [kvm_one_reg](https://elixir.bootlin.com/linux/v5.10/C/ident/kvm_one_reg) {
	[__u64](https://elixir.bootlin.com/linux/v5.10/C/ident/__u64) id;
	[__u64](https://elixir.bootlin.com/linux/v5.10/C/ident/__u64) addr;
};
```
and further along in the 5.10 kernel code I can only find the value handled as `u64`. I haven't checked newer kernels to look for `u128` use cases but I think we're covered with this logic and the fact that we're little endian only.

src/arch/src/aarch64/regs.rs
```diff
@@ -256,7 +276,10 @@ pub fn is_system_register(regid: u64) -> bool {
 ///
 /// * `vcpu` - Structure for the VCPU that holds the VCPU's fd.
 pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
-    vcpu.get_one_reg(MPIDR_EL1).map_err(Error::GetSysRegister)
+    match vcpu.get_one_reg(MPIDR_EL1) {
+        Err(err) => Err(Error::GetSysRegister(err)),
+        Ok(val) => Ok(val.try_into().expect("MPIDR register to be 64 bit"))
```
luminitavoicu: Let's also provide a documentation to this expect, to explain that we don't expect it to happen. 
```suggestion
        // MPIDR register is 64-bit so this operation will never fail.
        Ok(val) => Ok(val.try_into().expect("MPIDR register to be 64 bit"))
```
or something similar

src/arch/src/aarch64/regs.rs
```diff
@@ -160,10 +180,10 @@ arm64_sys_reg!(MIDR_EL1, 3, 0, 0, 0, 0);
 /// # Arguments
 ///
 /// * `state` - Array slice of kvm_one_reg structures, representing the registers of a VCPU state.
-pub fn get_manufacturer_id_from_state(state: &[kvm_one_reg]) -> Result<u32> {
+pub fn get_manufacturer_id_from_state(state: &[Aarch64Register]) -> Result<u32> {
     let midr_el1 = state.iter().find(|reg| reg.id == MIDR_EL1);
     match midr_el1 {
-        Some(value) => Ok(value.addr as u32 >> 24),
+        Some(value) => Ok(value.value as u32 >> 24),
```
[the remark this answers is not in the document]
roypat: good point, `value.value` looks a bit silly haha

src/arch/src/aarch64/regs.rs
roypat: done!

luminitavoicu (dismissed):
LGTM

src/arch/src/aarch64/regs.rs
```diff
@@ -10,10 +10,30 @@ use std::{fmt, fs, mem, result, u32};
 
 use kvm_bindings::*;
 use kvm_ioctls::VcpuFd;
+use versionize::*;
+use versionize_derive::Versionize;
 use vm_memory::GuestMemoryMmap;
 
 use super::get_fdt_addr;
 
+/// Struct describing a saved aarch64 register
```
aghecenco, line 19: nit: add full stop

src/arch/src/aarch64/regs.rs
```diff
@@ -10,10 +10,30 @@ use std::{fmt, fs, mem, result, u32};
...
+///
+/// Used for interacting with `KVM_GET/SET_ONE_REG`
+#[derive(Debug, Clone, Versionize, PartialEq, Eq)]
+pub struct Aarch64Register {
+    // fields have to be public because of unittest in vmm crate
```
aghecenco, line 24: nit: extraneous comment

aghecenco (approved):

src/arch/src/aarch64/regs.rs
roypat: done

src/arch/src/aarch64/regs.rs
roypat: done

tests/framework/artifacts.py
```diff
@@ -306,7 +306,7 @@ class ArtifactCollection:
     PLATFORM = platform.machine()
 
     # S3 bucket structure.
-    ARTIFACTS_ROOT = "ci-artifacts"
+    ARTIFACTS_ROOT = "ci-artifacts-dev"
```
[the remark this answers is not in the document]
roypat: done

luminitavoicu (approved):

aghecenco (approved):

PR #3287, opened 2022-11-23
Removing MAC address random generation

### LINKED ISSUE #2943: Fix loose ends of GET /vm/config post-restore

ShadowCurse:
Signed-off-by: Egor Lazarchuk <yegorlz@amazon.co.uk>

This PR:
- Removes random MAC address generation introduced by [this](https://github.com/firecracker-microvm/firecracker/pull/3121) commit
- fixes an issue in `GET /vm/config` which returned invalid `00:00:00:00:00:00` MAC address if it was not previously specified by the user.

Additionally fixes snapshot files not being truncated when saved.

## Reason
Fixes #2943 second time

## License Acceptance

## PR Checklist

---

### DIFF
--- a/docs/network-setup.md
+++ b/docs/network-setup.md
@@ -76,10 +76,6 @@ to your configuration file like this:
 Alternatively, if you are using firectl, add
 --tap-device=tap0/AA:FC:00:00:00:01` to your command line.
 
-*Note:* When `guest_mac` is not explicitely specified, a MAC address is generated
-and assigned to the network interface. This address is included in the result of
-GET `/vm/config`.
-
 ## In The Guest
 
 Once you have booted the guest, bring up networking within the guest:
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -4,6 +4,7 @@
 // Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
+
 use std::io::{Read, Write};
 use std::net::Ipv4Addr;
 use std::sync::atomic::AtomicUsize;
@@ -17,8 +18,7 @@ use mmds::data_store::Mmds;
 use mmds::ns::MmdsNetworkStack;
 use rate_limiter::{BucketUpdate, RateLimiter, TokenType};
 use utils::eventfd::EventFd;
-use utils::net::mac::{MacAddr, MAC_ADDR_LEN};
-use utils::rand_bytes;
+use utils::net::mac::MacAddr;
 use virtio_gen::virtio_net::{
     virtio_net_hdr_v1, VIRTIO_F_VERSION_1, VIRTIO_NET_F_CSUM, VIRTIO_NET_F_GUEST_CSUM,
     VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_UFO, VIRTIO_NET_F_HOST_TSO4, VIRTIO_NET_F_HOST_UFO,
@@ -47,9 +47,6 @@ enum FrontendError {
     ReadOnlyDescriptor,
 }
 
-// KVM OUI MAC address is 52:54:00:xx:xx:xx
-const KVM_OUI: [u8; 3] = [0x52, 0x54, 0x00];
-
 pub(crate) fn vnet_hdr_len() -> usize {
     mem::size_of::<virtio_net_hdr_v1>()
 }
@@ -75,31 +72,12 @@ fn frame_bytes_from_buf_mut(buf: &mut [u8]) -> Result<&mut [u8]> {
 // This initializes to all 0 the VNET hdr part of a buf.
 fn init_vnet_hdr(buf: &mut [u8]) {
     // The buffer should be larger than vnet_hdr_len.
-    // TODO: any better way to set all these bytes to 0? Or is this optimized by the compiler?
-    for i in &mut buf[0..vnet_hdr_len()] {
-        *i = 0;
-    }
+    buf[0..vnet_hdr_len()].fill(0);
 }
 
-fn generate_mac_address() -> MacAddr {
-    let mut random_mac: [u8; MAC_ADDR_LEN] = [0; MAC_ADDR_LEN];
-    let (left, right) = random_mac.split_at_mut(3);
-    left.copy_from_slice(&KVM_OUI);
-    right.copy_from_slice(rand_bytes(3).as_slice());
-    MacAddr::from_bytes_unchecked(&random_mac)
-}
-
-#[derive(Clone, Copy)]
+#[derive(Debug, Default, Clone, Copy)]
 pub struct ConfigSpace {
-    pub guest_mac: [u8; MAC_ADDR_LEN],
-}
-
-impl Default for ConfigSpace {
-    fn default() -> ConfigSpace {
-        ConfigSpace {
-            guest_mac: [0; MAC_ADDR_LEN],
-        }
-    }
+    pub guest_mac: MacAddr,
 }
 
 // SAFETY: `ConfigSpace` contains only PODs.
@@ -130,7 +108,7 @@ pub struct Net {
     pub(crate) irq_trigger: IrqTrigger,
 
     pub(crate) config_space: ConfigSpace,
-    pub(crate) guest_mac: MacAddr,
+    pub(crate) guest_mac: Option<MacAddr>,
 
     pub(crate) device_state: DeviceState,
     pub(crate) activate_evt: EventFd,
@@ -146,7 +124,7 @@ impl Net {
     pub fn new_with_tap(
         id: String,
         tap_if_name: String,
-        guest_mac: Option<&MacAddr>,
+        guest_mac: Option<MacAddr>,
         rx_rate_limiter: RateLimiter,
         tx_rate_limiter: RateLimiter,
     ) -> Result<Self> {
@@ -172,18 +150,12 @@ impl Net {
             | 1 << VIRTIO_RING_F_EVENT_IDX;
 
         let mut config_space = ConfigSpace::default();
-        // Enable feature to configure a MAC address
-        // If not enabled, the driver generates a random MAC address.
-        avail_features |= 1 << VIRTIO_NET_F_MAC;
-        let mac_addr = {
-            if let Some(some_mac_adrr) = guest_mac {
-                *some_mac_adrr
-            } else {
-                // When the MAC address is not specified explicitly, generate one and assign it.
-                generate_mac_address()
-            }
-        };
-        config_space.guest_mac.copy_from_slice(mac_addr.get_bytes());
+        if let Some(mac) = guest_mac {
+            config_space.guest_mac = mac;
+            // Enabling feature for MAC address configuration
+            // If not set, the driver will generates a random MAC address
+            avail_features |= 1 << VIRTIO_NET_F_MAC;
+        }
 
         let mut queue_evts = Vec::new();
         let mut queues = Vec::new();
@@ -207,11 +179,11 @@ impl Net {
             tx_frame_buf: [0u8; MAX_BUFFER_SIZE],
             tx_iovec: Vec::with_capacity(QUEUE_SIZE as usize),
             irq_trigger: IrqTrigger::new().map_err(Error::EventFd)?,
+            config_space,
+            guest_mac,
             device_state: DeviceState::Inactive,
             activate_evt: EventFd::new(libc::EFD_NONBLOCK).map_err(Error::EventFd)?,
-            config_space,
             mmds_ns: None,
-            guest_mac: mac_addr,
 
             #[cfg(test)]
             mocks: Mocks::default(),
@@ -224,8 +196,8 @@ impl Net {
     }
 
     /// Provides the MAC of this net device.
-    pub fn guest_mac(&self) -> &MacAddr {
-        &self.guest_mac
+    pub fn guest_mac(&self) -> Option<&MacAddr> {
+        self.guest_mac.as_ref()
     }
 
     /// Provides the host IFACE name of this net device.
@@ -429,7 +401,7 @@ impl Net {
         rate_limiter: &mut RateLimiter,
         frame_buf: &[u8],
         tap: &mut Tap,
-        guest_mac: MacAddr,
+        guest_mac: Option<MacAddr>,
     ) -> Result<bool> {
         let checked_frame = |frame_buf| {
             frame_bytes_from_buf(frame_buf).map_err(|err| {
@@ -454,11 +426,13 @@ impl Net {
         // This frame goes to the TAP.
 
         // Check for guest MAC spoofing.
-        let _ = EthernetFrame::from_bytes(checked_frame(frame_buf)?).map(|eth_frame| {
-            if guest_mac != eth_frame.src_mac() {
-                METRICS.net.tx_spoofed_mac_count.inc();
-            }
-        });
+        if let Some(mac) = guest_mac {
+            let _ = EthernetFrame::from_bytes(checked_frame(frame_buf)?).map(|eth_frame| {
+                if mac != eth_frame.src_mac() {
+                    METRICS.net.tx_spoofed_mac_count.inc();
+                }
+            });
+        }
 
         match tap.write(frame_buf) {
             Ok(_) => {
@@ -844,8 +818,7 @@ impl VirtioDevice for Net {
         }
 
         config_space_bytes[offset as usize..(offset + data_len) as usize].copy_from_slice(data);
-        self.guest_mac =
-            MacAddr::from_bytes_unchecked(&self.config_space.guest_mac[..MAC_ADDR_LEN]);
+        self.guest_mac = Some(self.config_space.guest_mac);
         METRICS.net.mac_address_updates.inc();
     }
 
@@ -881,6 +854,7 @@ pub mod tests {
     use dumbo::pdu::ethernet::ETHERTYPE_ARP;
     use logger::{IncMetric, METRICS};
     use rate_limiter::{RateLimiter, TokenBucket, TokenType};
+    use utils::net::mac::MAC_ADDR_LEN;
     use virtio_gen::virtio_net::{
         virtio_net_hdr_v1, VIRTIO_F_VERSION_1, VIRTIO_NET_F_CSUM, VIRTIO_NET_F_GUEST_CSUM,
         VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_UFO, VIRTIO_NET_F_HOST_TSO4,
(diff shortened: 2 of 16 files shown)


src/devices/src/virtio/net/persist.rs
```diff
@@ -23,10 +24,39 @@ use super::{NUM_QUEUES, QUEUE_SIZE};
...
 
-#[derive(Clone, Versionize)]
+#[derive(Debug, Default, Clone, Versionize)]
 // NOTICE: Any changes to this structure require a snapshot version bump.
 pub struct NetConfigSpaceState {
-    guest_mac: [u8; MAC_ADDR_LEN],
+    #[version(end = 2, default_fn = "def_net_config_old")]
+    guest_mac_old: [u8; utils::net::mac::MAC_ADDR_LEN],
+    #[version(start = 2, de_fn = "de_net_config", ser_fn = "ser_net_config")]
+    guest_mac: Option<MacAddr>,
```
dianpopa, line 33: Why not `Option<[u8; utils::net::mac::MAC_ADDR_LEN]>`, I am asking because this way it would not be necessary to make the utils crate dependent on Versionize

ShadowCurse: We have `MacAddr` as type that represents MAC address and basically just a wrapper around `[u8; MAC_ADDR_LEN]`. 
I think it is better to use it, then just an array of bytes. It makes code easier to understand, without a bunch of `MacAddr::from_bytes_unchecked` calls.
(On this note I probably should make `guest_mac_old` also be `MacAddr`)

luminitavoicu (commented):
To make this easier to review, how about we have a commit where you undo changes in #3121 and another commit with fixes for the mac returned in `vm/config`?

src/devices/src/virtio/net/device.rs
```diff
@@ -172,18 +150,12 @@ impl Net {
...
-            } else {
-                // When the MAC address is not specified explicitly, generate one and assign it.
-                generate_mac_address()
-            }
-        };
-        config_space.guest_mac.copy_from_slice(mac_addr.get_bytes());
+        if let Some(mac) = guest_mac {
+            config_space.guest_mac = mac;
+            // When this feature isn't available, the driver generates a random MAC address.
+            // Otherwise, it should attempt to read the device MAC address from the config space.
```
vibharya, line 156: Can we reword the below comment better. A suggestion would be:
```suggestion
            // Enable feature to configure a MAC address
            // If not enabled, the driver generates an address
```

ShadowCurse: Updated

src/devices/src/virtio/net/persist.rs
```diff
@@ -47,7 +47,7 @@ impl NetConfigSpaceState {
         // v1.1 and older versions do not have optional MAC address.
         warn!("Saving to older snapshot version, optional MAC address will not be saved.");
         match self.guest_mac {
-            Some(mac) => self.guest_mac_old = mac.into(),
```
pb8o: We have a (soft) requirement that each commit should build and pass the tests. If this (and the others) fixes a clippy warning, it should go in the previous commit.

tests/integration_tests/performance/test_snapshot_perf.py
```diff
@@ -118,8 +117,8 @@
                     "2vcpu_512mb.json": {"target": 60},
                 },
                 "async": {
-                    "2vcpu_256mb.json": {"target": 190},
```
pb8o: This change seems unrelated to the MAC address? Or does the snapshot latency performance change with this commit?

src/devices/src/virtio/net/device.rs
```diff
@@ -152,8 +152,8 @@ impl Net {
         let mut config_space = ConfigSpace::default();
         if let Some(mac) = guest_mac {
             config_space.guest_mac = mac;
-            // When this feature isn't available, the driver generates a random MAC address.
```
pb8o: nit grammar: "will generate" or "generates"

src/utils/src/net/mac.rs
```diff
@@ -104,6 +118,7 @@ impl MacAddr {
     /// let mac = MacAddr::from_bytes(&[0x01, 0x02, 0x03, 0x04, 0x05, 0x06]).unwrap();
     /// println!("{}", mac.to_string());
     /// ```
+    /// TODO this is unused, do we need it?
```
pb8o, line 121: Good catch. I would say remove it as a separate commit.

src/vmm/src/persist.rs
```diff
@@ -174,73 +173,47 @@ pub enum MicrovmStateError {
...
     /// Failed to open the snapshot backing file.
+    #[error("Cannot perform {0} on the snapshot backing file: {1}")]
     SnapshotBackingFile(&'static str, io::Error),
-    #[cfg(target_arch = "x86_64")]
     /// Number of devices exceeds the maximum supported devices for the snapshot data version.
+    #[cfg(target_arch = "x86_64")]
+    #[error(
+        "Too many devices attached: {0}. The maximum number allowed for the snapshot data version \
+         requested is {FC_V0_23_MAX_DEVICES}."
+    )]
```
pb8o, line 213: This seems unrelated to the MAC change. We have a PR best practice to have one logical change per commit. Can this be split to a separate commit?

pb8o:
> > To make this easier to review, how about we have a commit where you undo changes in #3121 and another commit with fixes for the mac returned in `vm/config`?
> 
> @luminitavoicu I don't think I can `unsquash` commits. In addition, as I remember I did both changes at the same time, so they were in one commit anyway.

You can [unsquash](1) and reorder commits with `git rebase`. What I would do here: 

1. Start from a clean branch
2. `git revert` the relevant commits from #3121 
3. Cherry pick your 2 commits from this branch `git cherry-pick ...`
4. Do any other history manipulation as needed (merge commits, split commits) as mentioned in the comments
 
[1]: https://stackoverflow.com/questions/6217156/break-a-previous-commit-into-multiple-commits

docs/network-setup.md
```diff
@@ -76,10 +76,6 @@ to your configuration file like this:
 Alternatively, if you are using firectl, add
 --tap-device=tap0/AA:FC:00:00:00:01` to your command line.
 
-*Note:* When `guest_mac` is not explicitely specified, a MAC address is generated
-and assigned to the network interface. This address is included in the result of
-GET `/vm/config`.
-
```
vibharya, line 82: Do we need to add any information regarding what happens when the user does not specify a MAC address?

```
*Note:* When `guest_mac` is not explicitly specified, a MAC address is generated
by the driver and assigned to the network interface. This address cannot be
queried via Firecracker
```

WDYT @dianpopa  @pb8o ?

tests/integration_tests/performance/test_snapshot_perf.py
ShadowCurse: I am not sure why this is in one of the commits, this change is not shown in the `file changes`

ShadowCurse:
> > > To make this easier to review, how about we have a commit where you undo changes in #3121 and another commit with fixes for the mac returned in `vm/config`?
> > 
> > 
> > @luminitavoicu I don't think I can `unsquash` commits. In addition, as I remember I did both changes at the same time, so they were in one commit anyway.
> 
> You can [unsquash](1) and reorder commits with `git rebase`. What I would do here:
> 
>     1. Start from a clean branch
> 
>     2. `git revert` the relevant commits from [Snapshot-ga: Generate a MAC address if unspecified #3121](https://github.com/firecracker-microvm/firecracker/pull/3121)
> 
>     3. Cherry pick your 2 commits from this branch `git cherry-pick ...`
> 
>     4. Do any other history manipulation as needed (merge commits, split commits) as mentioned in the comments

I did both changes at the same time, so I don't think this is `unsquash`able

vibharya:
In addition to my other review comments, I believe changes need to be made in CHANGELOG.md as well.

Remove [this point](https://github.com/firecracker-microvm/firecracker/blob/main/CHANGELOG.md#:~:text=A%20MAC%20address%20is%20generated%20if%20one%20is%20not%20explicitly%20specified%20while%20adding%20network%20interfaces.%20This%20address%20can%20be%20obtained%20as%20part%20of%20the%20GET%20/vm/config.) and edit [this one](https://github.com/firecracker-microvm/firecracker/blob/main/CHANGELOG.md#:~:text=When%20MAC%20address%20was%20left%20unspecified%2C%20GET%20/vm/config%20would%20report%20the%20interface%20MAC%20address%20incorrectly%20as%20null%2C%20and%20post%20snapshot%20restore%20as%2000%3A00%3A00%3A00%3A00%3A00.%20It%20now%20correctly%20reports%20the%20code%20generated%20MAC%20address%20when%20queried%20in%20both%20cases.)

vibharya:
In addition to my other review comments, I believe changes need to be made in CHANGELOG.md as well.

Remove [this point](https://github.com/firecracker-microvm/firecracker/blob/main/CHANGELOG.md#:~:text=A%20MAC%20address%20is%20generated%20if%20one%20is%20not%20explicitly%20specified%20while%20adding%20network%20interfaces.%20This%20address%20can%20be%20obtained%20as%20part%20of%20the%20GET%20/vm/config.) and edit [this one](https://github.com/firecracker-microvm/firecracker/blob/main/CHANGELOG.md#:~:text=When%20MAC%20address%20was%20left%20unspecified%2C%20GET%20/vm/config%20would%20report%20the%20interface%20MAC%20address%20incorrectly%20as%20null%2C%20and%20post%20snapshot%20restore%20as%2000%3A00%3A00%3A00%3A00%3A00.%20It%20now%20correctly%20reports%20the%20code%20generated%20MAC%20address%20when%20queried%20in%20both%20cases.)

src/devices/src/virtio/net/persist.rs
```diff
@@ -67,7 +97,11 @@ impl Persist<'_> for Net {
             tx_rate_limiter_state: self.tx_rate_limiter.save(),
             mmds_ns: self.mmds_ns.as_ref().map(|mmds| mmds.save()),
             config_space: NetConfigSpaceState {
-                guest_mac: self.config_space.guest_mac,
+                guest_mac: self.guest_mac,
+                guest_mac_old: match self.guest_mac {
```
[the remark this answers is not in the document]
ShadowCurse: yup, seems like useless duplication, I will remove it

src/utils/src/net/mac.rs
ShadowCurse: If we don't want to check the length, then we have a `from_bytes_unchecked`

vibharya: Yes I understand that there's another function.
I'm trying to make sure that we have no use case where we'll need to verify length. 

If there are none, we can remove it.

src/devices/src/virtio/net/device.rs
ShadowCurse: Am I blind or there is not grammar error?

src/utils/src/net/mac.rs
ShadowCurse: well if we need to verify then I think it is better to implement `TryFrom` trait

docs/network-setup.md
dianpopa: @vibharya you mean let them know that the driver will assign one for them?

As far as the CHANGELOG, yes, we need to remove the entries you added.

src/utils/src/net/mac.rs
dianpopa: This was meant as an interface, i say we leave it as it is

docs/network-setup.md
vibharya: Yes, let them know one will be assigned for them. And that we cannot retrieve/report it using FC.

Yes, we can remove the entries I added and instead mention that they are now reported as null (or "None" or excluded - not sure exactly what/vm/config is returning now in this case) pre and post snapshot resume.
Hence fixing [this part](https://github.com/firecracker-microvm/firecracker/issues/2943#:~:text=net%20devices%20MAC%20addresses%20are%20reported%20as%20all%20zeros%20if%20left%20unconfigured.%20View%20Partial%20fix%20for%20restore%20snapshot%20GET%20/vm/config%20bug%C2%A0%232806%20(comment)) of the issue.

ShadowCurse: Btw, can't we retrieve the MAC with `write_config`?

ShadowCurse: @dianpopa @vibharya I changed `CHANGELOG`. Your thoughts?

src/vmm/src/persist.rs
```diff
@@ -277,6 +250,7 @@ fn snapshot_state_to_file(
     let mut snapshot_file = OpenOptions::new()
         .create(true)
         .write(true)
+        .truncate(true)
```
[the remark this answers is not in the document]
pb8o: Maybe add more detail to the body of the commit, what happens if this is not set.

pb8o (dismissed):

docs/network-setup.md
dianpopa: sounds good to me

vibharya: Have you been able to retrieve the MAC address using `write_config`?

Also, what do we get when the MAC address is left unspecified currently?

src/devices/src/virtio/net/persist.rs
```diff
@@ -23,10 +24,39 @@ use super::{NUM_QUEUES, QUEUE_SIZE};
 use crate::virtio::persist::{Error as VirtioStateError, VirtioDeviceState};
 use crate::virtio::{DeviceState, TYPE_NET};
 
-#[derive(Clone, Versionize)]
+#[derive(Debug, Default, Clone, Versionize)]
 // NOTICE: Any changes to this structure require a snapshot version bump.
 pub struct NetConfigSpaceState {
-    guest_mac: [u8; MAC_ADDR_LEN],
+    #[version(end = 2, default_fn = "def_guest_mac_old")]
+    guest_mac_old: MacAddr,
```
dianpopa, line 31: I am actually surprised that the deserialization still works after changing the type of `guest_mac_old`. 
I guess this is due to how bincode does the serialization/deserialization. I suppose that `MacAddr` which behind the scenes is a struct with only one field will get deserialized to exactly what that one field contains.
So that 
```
MacAddr {byest [u8; MAC_ADDR_LEN] 
```
is equivalent to `[u8; MAC_ADDR_LEN]`
To be honest, I do not feel comfortable with this change since this implies that we will depend on how bincode works  (which can very well change in future). 
Also, I feel that when doing such changes in structs that get versionized, we should keep the changes as suggestive  as possible (history wise) by keeping old fields exactly as they were before the addition and apply the freedom of naming and type only on new fields.
In  a nutshell, the way I would have done it:

```
NetConfigSpaceState {
guest_mac: [u8; MAC_ADDR_LEN], -> exaclty as before,
guest_mac_opt/guest_mac_new: Option<MacAddr>,
}
```

ShadowCurse: Well, in my opinion, `MacAddr` should not even be a struct (well it will be) but just: `pub struct`

docs/network-setup.md
dianpopa: When the MAC address is left unspecified, we get back `None`. Exactly what the user configured.

In oder to receive the MAC_ADDR from the driver through `write_config`, we need to negotiate the `VIRTIO_NET_F_CTRL_MAC_ADDR` which needs to have a control queue set up. However, since our initial intent with the `GetConfig` was to give back the config provided by the user, we decided that obtaining the actual mac addr set by the driver is not part of that feature. Also, if the user has a real need to know the mac address from inside the microVM, they always have the possibility to specify one.

src/devices/src/virtio/net/persist.rs
ShadowCurse: Well in my opnion `MacAddr` should be:
```rust
pub struct MacAddr([u8; MAC_ADDR_LEN])
```
instead of what it is not. Then it is clear that it will always be ser/deser as just 6 bytes. And I don't think where is any way this bytes would be ser/de in any other way.

Also about changing the name to `guset_mac` and `guest_mac_opt/_new`, I think this is confusing, because in new versions of FC only `guest_mac_opt/_new` would be used. Then if someone would work with this `NetConfig` it would be unclear what field to use.

docs/network-setup.md
vibharya: Just verified that the output of GET /vm/config (even for boot from snapshot) would show the network interface as something like
`{
      "iface_id":"1",
      "host_dev_name":"vmtap30",
      "guest_mac":null,
      "rx_rate_limiter":null,
      "tx_rate_limiter":null
}`
Thats good, was the case previously as well.


> When the MAC address is left unspecified, we get back `None`. Exactly what the user configured.
> 
> In oder to receive the MAC_ADDR from the driver through `write_config`, we need to negotiate the `VIRTIO_NET_F_CTRL_MAC_ADDR` which needs to have a control queue set up. However, since our initial intent with the `GetConfig` was to give back the config provided by the user, we decided that obtaining the actual mac addr set by the driver is not part of that feature. Also, if the user has a real need to know the mac address from inside the microVM, they always have the possibility to specify one.

Yes, I understand and agree with this part 100%.
I'm only wondering if we need a note in the snapshot_support doc regarding this or not.

src/devices/src/virtio/net/persist.rs
dianpopa: > Well in my opnion `MacAddr` should be:
> 
> ```rust
> pub struct MacAddr([u8; MAC_ADDR_LEN])
> ```
> 
> instead of what it is now. Then it is clear that it will always be ser/deser as just 6 bytes. And I don't think where is any way this bytes would be ser/de in any other way.
> 
> Also about changing the name to `guset_mac` and `guest_mac_opt/_new`, I think this is confusing, because in new versions of FC only `guest_mac_opt/_new` would be used. Then if someone would work with this `NetConfig` it would be unclear what field to use.

I can only agree that `MacAddr` should have been exactly that, but it s not right now, true? My personal take on this is that I would not choose to change an old field (which indeed was not really the best choice) on which snapshot compatibility depends for the sake of readability? (I guess).

I was saying in my previous comment that we should have the changes to these fields tell a story. Maybe the names suggested by me are not the happiest choice and we can go for some other options (`guest_mac_v2`
etc). Also, If we need to change the type of `guest_mac` in the future, what would we call it that time?

ShadowCurse: Ok, we can have a bit ugly, but very story telling naming like `guest_mac_pre_1_2`, `guest_mac_1_2` and so on.
Also to make old fields more explicit we can use `#[deprecated]`like this:
```rust
 #[deprecated(since = "1.2.0", note = "this field is deprecated")]
 #[version(end = 2, default_fn = "def_guest_mac_old")]
 guest_mac_pre_1_2: [u8; MAC_ADDR_LEN],
 #[version(start = 2, de_fn = "de_guest_mac", ser_fn = "ser_guest_mac")]
 guest_mac_1_2: Option<MacAddr>,
```

dianpopa: Indeed it is a bit ugly to add all the number transitions; 
I think that end = 2, start  = 2  suffice to correlate to the version map.
The deprecated thingy I like a lot, since it s a way to also correlate the field to the FC version.

ShadowCurse: Well, lets leave adding deprecation to our future selfs. It deserves a separate PR because there are several places where we updated fields this way.

dianpopa (dismissed):

vibharya (approved):
LGTM

roypat (approved):

PR #3292, opened 2022-11-26
format files by black inside integration_tests/build

### LINKED ISSUE #3166: [Bug] Black does not format files inside integration_tests/build
# Describe the bug

See this https://github.com/psf/black/issues/1193.

## To Reproduce

`./tools/devtool shell -p`
`cd tests/integration_tests/build`
`black .`
```
No Python files are present to be formatted. Nothing to do
```

## Expected behaviour

Files in `integration_tests/build directory` are Python formatted.

cm-iwata:
## Changes
Specify black configration for check `tests/integration_tests/build` directory

## Reason
Fixes #3166
By default black exclude `build` directory.

https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#command-line-options

We can override default config via pyproject.toml.
 https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file

## License Acceptance

## PR Checklist

---


pyproject.toml
```diff
@@ -1,3 +1,6 @@
+[tool.black]
+exclude = "/(\\.direnv|\\.eggs|\\.git|\\.hg|\\.mypy_cache|\\.nox|\\.tox|\\.venv|venv|\\.svn|\\.ipynb_checkpoints|_build|buck-out|dist|__pypackages__)/"
```
cm-iwata, line 2: Remove `build` from [default](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#command-line-options)

roypat, line 2: The files in `tools/devctr/` are only used during building of our docker containers and should probably stay in there. Could you instead create a new `pyproject.toml` in the root?

Also, what are the disadvantages to just doing `black . --check --diff --exclude ''` over in https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/style/test_python.py? Testing locally that seems to give the same behavior.

roypat (commented):
Hey, 
good catch noticing that the build tests weren't being formatted! And thanks for directly submitting a fix. I only left a small comment (which is mostly to me being unfamiliar with python tooling), but overall looks pretty good to me!

pyproject.toml
cm-iwata: @roypat 
Thank you for review.

>The files in tools/devctr/ are only used during building of our docker containers

Devcontainer is convenient for quick setup.
But, I prefer use host environment directly because of some IDE support.
Since there are cases like this, I think that `pyproject.toml` should be placed in the project root without dividing.

>what are the disadvantages to just doing black . --check --diff --exclude ''

If you modify the code in `test_python.py`, this will only affect when you run the test code.
But, according #3166 , She try to use `black` on shell.
By adding settings to `pyproject.toml`, it is possible to apply common settings to both when executing test code and executing black directly on the shell.

roypat: > If you modify the code in test_python.py, this will only affect when you run the test code.

Ah, I see, yes that makes sense!

pb8o: We already have some python config files in the `tests` directory, so it would be more consistent to place it there or (probably simpler) just create a new `pyproject.toml` there and leave the poetry setting close to the `Dockerfile`

pb8o: Maybe add a comment saying that.

cm-iwata: OK.
Added comment!

cm-iwata: Created new `pyproject.toml` under `tests` directory.

roypat (approved):

roypat:
Looks good to me now! Could you squash your commits into one?

cm-iwata:
@roypat 
squashed my commits!

roypat (approved):

pb8o:
Thank you for the contribution @cm-iwata !

PR #3296, opened 2022-11-28
Improve assert messages in mmds::token::test_compute_expiry

roypat:

## Changes

In case the assertions fail, print out the values

## Reason

We have observed spurious failures in this test, and this hopefully helps debugging them

## License Acceptance

## PR Checklist

---

### DIFF
--- a/src/mmds/src/token.rs
+++ b/src/mmds/src/token.rs
@@ -378,12 +378,16 @@ mod tests {
         // We allow a deviation of 20ms to account for the gap
         // between the two calls to `get_time_ms()`.
         let deviation = 20;
-        assert!(ttl >= MILLISECONDS_PER_SECOND - deviation && ttl <= MILLISECONDS_PER_SECOND);
+        assert!(
+            ttl >= MILLISECONDS_PER_SECOND && ttl <= MILLISECONDS_PER_SECOND + deviation,
+            "ttl={ttl} not within [{MILLISECONDS_PER_SECOND}, \
+             {MILLISECONDS_PER_SECOND}+{deviation}]",
+        );
 
         let time_now = get_time_ms(ClockType::Monotonic);
         let expiry = TokenAuthority::compute_expiry(0);
         let ttl = expiry - time_now;
-        assert!(ttl <= deviation);
+        assert!(ttl <= deviation, "ttl={ttl} is greater than {deviation}");
     }
 
     #[test]


src/mmds/src/token.rs
```diff
@@ -378,12 +378,23 @@ mod tests {
...
+            MILLISECONDS_PER_SECOND
+        );
 
         let time_now = get_time_ms(ClockType::Monotonic);
         let expiry = TokenAuthority::compute_expiry(0);
         let ttl = expiry - time_now;
-        assert!(ttl <= deviation);
+        assert!(
+            ttl <= deviation,
+            "ttl={} is greater than {}",
```
pb8o, line 394: nit: Use f-string here and above? Also seems like a perfect use of the `=` formatting:

```python
ttl = 100
limit = 50
print(f"{ttl=} is greater than {limit}")
ttl=100 is greater than 50
```

pb8o (approved):

src/mmds/src/token.rs
roypat: Ah, this is a rust test, not a python one, and it seems `=` formatting is not a thing in rust :(

pb8o: Doh! Sorry, somehow I assumed this was Python 🤦

pb8o (approved):

src/mmds/src/token.rs
```diff
@@ -378,12 +378,15 @@ mod tests {
         // We allow a deviation of 20ms to account for the gap
         // between the two calls to `get_time_ms()`.
         let deviation = 20;
-        assert!(ttl >= MILLISECONDS_PER_SECOND - deviation && ttl <= MILLISECONDS_PER_SECOND);
+        assert!(
+            ttl >= MILLISECONDS_PER_SECOND - deviation && ttl <= MILLISECONDS_PER_SECOND,
+            "ttl={ttl} not within +-{deviation} of {MILLISECONDS_PER_SECOND}",
```
zulinx86, line 383: This condition evaluates the `ttl` between `MILLISECONDS_PER_SECOND - deviation` and `MILLISECONDS_PER_SECOND`, not between `MILLISECONDS_PER_SECOND - deviation` and `MILLISECONDS_PER_SECOND + deviation`. So the message `+-{deviation}` looks not be appropriate or the condition itself might not be proper.

zulinx86 (changes requested):

src/mmds/src/token.rs
roypat: You are right, good catch! `ttl` can only every be larger than 1 second, not smaller, so the assert allows deviation in exactly the wrong direction. I would assume that this is indeed the reason we were observing the spurious failures.

zulinx86 (approved):

pb8o (approved):

PR #3262, opened 2022-11-15
chore(tests): add more logs to some intermittently failing tests

pb8o:
~It would be helpful to have more context on what commands were run and what they returned in the captured logs. This is helpful for troubleshooting.~

Turns out the logging can be megabytes in size, and can hide things like the traceback. So I am pulling that commit but including all the other little quality-of-life improvements I did to support it. 

Signed-off-by: Pablo Barbáchano <pablob@amazon.com>

## Changes

Increase the verbosity of the logger so that more messages appear in the logs.

## Reason

To help troubleshooting intermittent issues. 

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

## PR Checklist

- [x] All commits in this PR are signed (`git commit -s`).
- [x] If a specific issue led to this PR, this PR closes the issue.
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] New `unsafe` code is documented.
- [x] API changes follow the [Runbook for Firecracker API changes][2].
- [x] User-facing changes are mentioned in `CHANGELOG.md`.
- [x] All added/changed functionality is tested.

---

- [ ] This functionality can be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md

PR #3298, opened 2022-11-30
Fix test_dependencies only checking dependency count

roypat:

## Changes

This commit changes to the to
1. fix the above issues (do not use pprint and check the actual dependency set instead of just its len)
2. ignore dependency versions, and only care about their names (meaning updates to a dependency will not trigger the test)

## Reason

Only checking the number of dependencies allowed dependencies to be replaced without the test failing, as long as the overall number of lines in the dependencies.txt file was correct.

Note that due to the use of pprint automatically wrapping lines at 80 characters, the file contained more lines than dependencies, making the check even more inaccurate (dependencies with long paths/git urls werevbroken up into multiple lines, leading to a leniency of two dependencies that could be added without the test failing).

## License Acceptance

## PR Checklist

---


andreeaflorescu:
The version of dependencies though might still be valid because in most cases you don't want to end up using more than one version of the crate even if they're compatible.

roypat:
> The version of dependencies though might still be a valid check because in most cases you don't want to end up using more than one version of the crate even if they're compatible.

Oh this is a very good point, but should maybe go into a separate test that verifies that no different versions of the same dependency are used. From my understanding the point `test_dependencies` was solely to make the process of adding new dependencies explicitly go through review.

andreeaflorescu:
> Oh this is a very good point, but should maybe go into a separate test that verifies that no different versions of the same dependency are used. From my understanding the point `test_num_dependencies` was solely to make the process of adding new dependencies explicitly go through review.

That is correct, it was serving the purpose of not using multiple versions of the same crate indirectly though. From my understanding (which might be veeeeery outdated) if you have for example vm-memory v0.9.0 and vm-memory v0.10.0 then this will show up as a "new dependency" because vm-memory would be counted twice. In the end is better to make it explicit instead like you suggested.

roypat:
> > Oh this is a very good point, but should maybe go into a separate test that verifies that no different versions of the same dependency are used. From my understanding the point `test_num_dependencies` was solely to make the process of adding new dependencies explicitly go through review.
> 
> That is correct, it was serving the purpose of not using multiple versions of the same crate indirectly though. From my understanding (which might be veeeeery outdated) if you have for example vm-memory v0.9.0 and vm-memory v0.10.0 then this will show up as a "new dependency" because vm-memory would be counted twice. In the end is better to make it explicit instead like you suggested.

I just drafted such a test and interestingly enough it does not pass, even though we ran `cargo update` just 2 days ago. If we transitively depend on incompatible versions of the same crate (in our case, memoffset 0.6.5 and 0.7.1) there is nothing we can do about it, so I don't think we can add this as part of our CI (at least not easily)

tests/integration_tests/build/test_dependencies.py
```diff
@@ -47,23 +47,17 @@ def test_num_dependencies(dep_file):
...
-    elif len(current_deps) != len(prev_deps):
-        msg = "The build dependencies have changed." \
-              " Use the code above to modify the {} file.". \
-            format(dep_file)
-        assert False, msg
+
+    difference = current_deps - set(prev_deps)
+
+    if difference:
+        assert False, f"New build dependencies detected. Is this expected? New dependencies {difference}"
```
dianpopa, line 63: This line looks longer than 80, how come no linter complained?

dianpopa (dismissed):

tests/integration_tests/build/test_dependencies.py
zulinx86: The warning was disabled in this commit.
https://github.com/firecracker-microvm/firecracker/pull/3262/commits/f8f21d36f3ffc30e3ee971741664231c3f9fee20

roypat: I think that might be related to this: https://github.com/firecracker-microvm/firecracker/pull/3292

roypat: black chose some... interesting formatting for that line, lol

dianpopa (approved):

zulinx86 (approved):

PR #3303, opened 2022-12-01
fix(release): run cargo check more reliably

pb8o:
Signed-off-by: Pablo Barbáchano <pablob@amazon.com>

## Changes

Fix a bug found during the last release

## License Acceptance

## PR Checklist

---


kalyazin (dismissed):

tools/release-prepare.sh
```diff
@@ -66,16 +66,18 @@ for file in "${files_to_change[@]}"; do
...
 cargo check
-CHANGED=(Cargo.lock)
 
-cd tests/integration_tests/security/demo_seccomp
-cargo check
-cd -
-CHANGED+=(tests/integration_tests/security/demo_seccomp/Cargo.lock)
+# Run `cargo check` to update firecracker and jailer versions in all
+# `Cargo.lock`.
+for cargo_lock in $(find . -path ./build -prune -o -name Cargo.lock -print); do
```
roypat, line 74: This breaks if a Cargo.lock file is ever found on a path containing a space (e.g. if you create the file "little test/Cargo.lock") the output will be (I think using `-exec` instead would work here)

```bash
$ for cargo_lock in $(find . -path ./build -prune -o -name Cargo.lock -print); do echo $cargo_lock; done;
./tests/host_tools/uffd/Cargo.lock
./tests/integration_tests/security/demo_seccomp/Cargo.lock
./Cargo.lock
./little
test/Cargo.lock
```

kalyazin: How about prohibiting creation of spaced files/directories in the repo? :)

pb8o: The advantage of not dealing with this would be that __it will break__ during release preparation, at which point we can choose to either fix the offending path, or implement this fix.

pb8o: I will make a change to add a comment on this.

tools/release-prepare.sh
```diff
@@ -66,16 +67,17 @@ for file in "${files_to_change[@]}"; do
     sed -i "s/$prev_ver/$version/" "$file"
 done
 
-# Run `cargo check` to update firecracker and jailer versions in
-# `Cargo.lock`.
-say "Updating lockfile..."
+CHANGED=()
 cargo check
```
roypat, line 71: Do we need this `cargo check`? I think it is picked up by the loop below

tools/release-prepare.sh
```diff
@@ -66,16 +67,17 @@ for file in "${files_to_change[@]}"; do
...
-CHANGED=(Cargo.lock)
 
-cd tests/integration_tests/security/demo_seccomp
-cargo check
-cd -
-CHANGED+=(tests/integration_tests/security/demo_seccomp/Cargo.lock)
+# Run `cargo check` to update firecracker and jailer versions in all
+# `Cargo.lock`.
+# NOTE: This will break if it finds paths with spaces in them
```
roypat, line 75: Funnily enough, using the while loop instead of the for loop makes it work with spaces, so this comment isn't needed anymore 

```bash
> find . -path ./build -prune -o -name Cargo.lock -print |while read -r cargo_lock; do     echo "Updating $cargo_lock ..."; done
Updating ./tests/host_tools/uffd/Cargo.lock ...
Updating ./tests/integration_tests/security/demo_seccomp/Cargo.lock ...
Updating ./Cargo.lock ...
Updating ./little test/Cargo.lock ...
```

tools/release-prepare.sh
pb8o: Yeah, I think we could add a quick check for that. I can add it in this PR.

tools/release-prepare.sh
pb8o: hmm yeah... where did that come from? fixing

roypat (approved):

kalyazin (approved):

Issue #3215, opened 2022-10-27
False negatives of `clippy::unneccesary_cast`/`clippy::needless_borrow`


roypat:
Due to a bug in clippy that is fixed in nightly, the version of clippy we currently use does not warn for many instances of `clippy::unncessary_cast`: https://github.com/rust-lang/rust-clippy/issues/9562. Running `cargo +nightly clippy` on the firecracker workspace yields around 50 warnings of this type.

Something similar happens to a lesser extend with the `unnecessary_borrow` lint, however I could not track the change of behavior in clippy down to a specific issue.

This might cause issues if we ever decide to upgrade rust past 1.66 (once it is stable ofc), since we deny warnings.

Gilnaa:
Running with current nightly clippy (`clippy 0.1.66 (0da281b 2022-10-27)`) also uses `clippy::uninlined-format-args`:
```
error: variables can be used directly in the `format!` string
  --> src/vmm/build.rs:23:20
   |
23 |     json_path.push(format!("{}.json", target));
   |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
   = note: `-D clippy::uninlined-format-args` implied by `-D warnings`
help: change this to
   |
23 -     json_path.push(format!("{}.json", target));
23 +     json_path.push(format!("{target}.json"));
   |

error: variables can be used directly in the `format!` string
  --> src/vmm/build.rs:41:5
   |
41 |     println!("cargo:rerun-if-changed={}", json_path);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args
help: change this to
   |
41 -     println!("cargo:rerun-if-changed={}", json_path);
41 +     println!("cargo:rerun-if-changed={json_path}");
   |
```

This lint will also have to be disabled specifically in order to make this work on newer clippys. (It can be trivially fixed, of course, but it will affect the MSRV, which I'm not sure if firecracker has)

roypat:
We might want to specify a "minimum supported rust version" (https://doc.rust-lang.org/clippy/configuration.html#specifying-the-minimum-supported-rust-version) so prevent issues like this going forward.

Gilnaa:
[...]

PR #3305, opened 2022-12-02
Update Rust 1.65.0

JonathanWoollett-Light:

### Updated Rust

Updated Rust version to latest stable 1.65.0.

### Fixed newly detected clippy warnings.

Fixed a variety of minor clippy warnings that where newly detected with the update. Some of these overlap with #3215.

### Fixed `sanitize_process()` in `/src/jailer/src/main`

With the Rust update a [`debug_assert!`](https://github.com/rust-lang/rust/blob/1.64.0/library/std/src/sys/unix/fs.rs#L690) is changed to an [`assert!`](https://github.com/rust-lang/rust/blob/1.65.0/library/std/src/sys/unix/fs.rs#L690:L694). Since we run with a pre-compiled standard library this meant previously this assert was not being checking (as the standard library is compiled in release), now with the update it is being checked.

This now results in an error. 
```
thread 'main' panicked at 'unexpected error during closedir: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }', library/std/src/sys/unix/fs.rs:690:9
```
In our previous implementation we read the directory `/proc/self/fd` and close each attached file descriptor we read from this directory, the problem being one of the attached file descriptors is the directory reader. This means we close the file descriptor **then** the struct executes [its `Drop` implementation](https://github.com/rust-lang/rust/blob/1.65.0/library/std/src/sys/unix/fs.rs#L687:L696) which attempts to close the file descriptor, this results in a `-1` return from [`libc::closedir`](https://github.com/rust-lang/rust/blob/1.65.0/library/std/src/sys/unix/fs.rs#L689) and a errno of `EBADF` (since the file descriptor is no longer open) which fails [the `assert!`](https://github.com/rust-lang/rust/blob/1.65.0/library/std/src/sys/unix/fs.rs#L690:L694) resulting in a panic, which fails test.

Switching to the much more efficient and safe mechanism of [close_range(2)](https://man7.org/linux/man-pages/man2/close_range.2.html) avoids opening a file descriptor to then close all file descriptors.
```rust
let res = unsafe { libc::syscall(libc::SYS_close_range,3,libc::c_int::MAX,0) };
assert_eq!(res,0);
```
Unfortunately [close_range(2)](https://man7.org/linux/man-pages/man2/close_range.2.html) is not supported in the `4.14` version of the Linux kernel; being introduced in `5.9`.

A slightly worse but supported fix is:
```rust
let fd_limit = i32::try_from(unsafe { libc::sysconf(libc::_SC_OPEN_MAX) }).unwrap();
for fd in 3..fd_limit {
    unsafe {
        libc::close(fd);
    }
}
```

## Reason

Require [generic associated types](https://blog.rust-lang.org/2022/10/28/gats-stabilization.html) for work on https://github.com/firecracker-microvm/firecracker/pull/3076.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -7,7 +7,7 @@ use std::process::Command;
 // embedding the FIRECRACKER_VERSION directly in the resulting binary
 fn main() {
     let firecracker_version = Command::new("git")
-        .args(&["describe", "--dirty"])
+        .args(["describe", "--dirty"])
         .output()
         .ok()
         .and_then(|output| {
--- a/src/devices/src/virtio/balloon/device.rs
+++ b/src/devices/src/virtio/balloon/device.rs
@@ -686,7 +686,7 @@ pub(crate) mod tests {
                 assert_eq!(balloon.device_type(), TYPE_BALLOON);
 
                 let features: u64 = (1u64 << VIRTIO_F_VERSION_1)
-                    | ((if *deflate_on_oom { 1 } else { 0 }) << VIRTIO_BALLOON_F_DEFLATE_ON_OOM)
+                    | (u64::from(*deflate_on_oom) << VIRTIO_BALLOON_F_DEFLATE_ON_OOM)
                     | ((u64::from(*stats_interval)) << VIRTIO_BALLOON_F_STATS_VQ);
 
                 assert_eq!(balloon.avail_features_by_page(0), features as u32);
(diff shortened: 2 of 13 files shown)


roypat:
> LGTM. Maybe you could add some comments in the commit body on what changes this required. It looks like new clippy lints?

I think it's #3215, not new lints

JonathanWoollett-Light:
Closes #3215

pb8o (approved):

roypat (approved):

pb8o (approved):

roypat (approved):

pb8o (approved):

PR #3314, opened 2022-12-08
chore: Update snap/restore baseline

zulinx86:
## Changes

Updates baselines for snapshot/restore tests on kernel 4.14 + m6a.metal (AMD).

## Reason

The snapshot/restore performance test continued to fail since 2022-10-30. 

As a result of investigation, `ioctl()` for `KVM_CREATE_VM` and the first `KVM_CREATE_VCPU` takes much time and most of the elapsed time for `ioctl()` was spent on sending IPIs to all-but-self.

The reasons why we see this issue only on the combination of kernel 4.14 and m6a.metal (AMD) are:
- m6a.metal (AMD) uses xAPIC and needs to send 1 IPI for each CPU thread, while m5d.metal (Intel) uses 2xAPIC support cluster configuration and just need to send 1 IPI for each cluster.
- [This patch](https://lwn.net/Articles/793065/) is available on kernel 5.10 and it reduces the time to send IPIs to all-buf-self by using the destination shorthand feature.

As described above, the latency seems to come from the available APIC (xAPIC or x2APIC) and this is out of control from us. Thus, this commit just updates the performance baseline.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/performance/configs/test_snap_restore_performance_config_4.14.json
+++ b/tests/integration_tests/performance/configs/test_snap_restore_performance_config_4.14.json
@@ -2182,350 +2182,350 @@
                                         "1vcpu_128mb": {
                                             "P50": {
                                                 "restore": {
-                                                    "target": 7.179,
-                                                    "delta_percentage": 12
+                                                    "target": 12.847,
+                                                    "delta_percentage": 13
                                                 }
                                             },
                                             "P90": {
                                                 "restore": {
-                                                    "target": 7.681,
-                                                    "delta_percentage": 21
+                                                    "target": 13.303,
+                                                    "delta_percentage": 16
                                                 }
                                             }
                                         },
                                         "2vcpu_128mb": {
                                             "P50": {
                                                 "restore": {
-                                                    "target": 7.277,
-                                                    "delta_percentage": 12
+                                                    "target": 13.035,
+                                                    "delta_percentage": 14
                                                 }
                                             },
                                             "P90": {
                                                 "restore": {
-                                                    "target": 7.835,
-                                                    "delta_percentage": 24
+                                                    "target": 13.483,
+                                                    "delta_percentage": 19
                                                 }
                                             }
                                         },
                                         "3vcpu_128mb": {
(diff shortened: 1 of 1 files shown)


zulinx86:
The comparison result between before and after is here:
```
{
    "source": "e2541f75414e5923176456197f687c36ce3da12f/tests/integration_tests/performance/configs/",
    "target": "tests/integration_tests/performance/configs/",
    "test_snap_restore_performance_config_4.14.json": {
        "test": "snap_restore_performance",
        "kernel": "4.14",
        "cpus": [
            {
                "instance": "m6a.metal",
                "model": "AMD EPYC 7R13 48-Core Processor",
                "stats": {
                    "latency": {
                        "target_diff_percentage": {
                            "mean": 72.11491432543579,
                            "stdev": 9.794462497354829
                        },
                        "delta_percentage_diff": {
                            "mean": -7.82,
                            "stdev": 19.748100535177425
                        }
                    }
                }
            }
        ]
    }
}
```

This introduces +72% latency.

pb8o (approved):

mattschlebusch (commented):
The result here is very interesting. Am I understanding this correctly that KVM performance on AMD is several times slower than Intel? (Based on this file's proposed thresholds compared to the m5d.metal instance)

If the answer is yes I would expect this issue to be tracked upstream, either by AMD or by Linux. Can we track that somewhere?

zulinx86:
> The result here is very interesting. Am I understanding this correctly that KVM performance on AMD is several times slower than Intel? (Based on this file's proposed thresholds compared to the m5d.metal instance)
> 
> If the answer is yes I would expect this issue to be tracked upstream, either by AMD or by Linux. Can we track that somewhere?

This result doesn't depend on CPU type, but on available APIC (xAPIC or x2APIC).
This latency has been resolved in [this patch](https://lwn.net/Articles/793065/) by using IPI shorthands and this is not available in kernel 4.14.

mattschlebusch (approved):

zulinx86:
Updated the description of this PR for more clarity.

PR #3324, opened 2022-12-13
Feature parity CPU templates for Intel/AMD

kalyazin:
## Changes

- Add a test that the MPX feature flag is never set in CPU templates
- Disable feature flags that are unintentionally set in T2 CPU template on Ice Lake
- Add T2CL (Intel) and T2A (AMD) CPU templates to provide instruction set feature parity between Intel and AMD

## Reason

Provide consistent behaviour from instruction set point of view across Intel and AMD CPUs. As long as the user software respects CPUID/MSR feature flags related to instruction sets, the instructions sets between Intel and AMD should be the same if the new templates are used.

## License Acceptance

## PR Checklist

---


docs/snapshotting/versioning.md
```diff
@@ -193,7 +193,12 @@ of CPU features. The T2S template is designed to allow migrating snapshots
 between hosts with Intel SkyLake and Cascade Lake securely by further
 restricting CPU features for the guest, however this comes with a performance
 penalty. Users are encouraged to carry out a performance assessment if they wish
-to use the T2S template. There are no templates available for AMD or ARM64.
+to use the T2S template. T2CL template is mapped to be close to Intel Cascade Lake.
+It is not safe to use it on Intel CPUs older than Cascade Lake (e.g., Skylake).
```
JonathanWoollett-Light, line 197: Unnecessary comma.
```suggestion
It is not safe to use it on Intel CPUs older than Cascade Lake (e.g. Skylake).
```

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,407 @@
...
+    cpuid_utils.get_cpu_vendor()
+    not in [cpuid_utils.CpuVendor.INTEL, cpuid_utils.CpuVendor.AMD],
+    reason="CPU templates are only available on Intel and AMD.",
+)
+@pytest.mark.parametrize(
+    "cpu_template",
+    ["C3", "T2", "T2S", "T2CL"]
+    if cpuid_utils.get_cpu_vendor() == cpuid_utils.CpuVendor.INTEL
+    else ["T2A"],
+)
```
JonathanWoollett-Light, line 116: Could we use a constant to simplify this here (and in other cases)?
```suggestion
if cpuid_utils.get_cpu_vendor() == cpuid_utils.CpuVendor.INTEL:
    SUPPORTED_TEMPLATES = ["C3", "T2", "T2S", "T2CL"]
elif cpuid_utils.get_cpu_vendor() == cpuid_utils.CpuVendor.AMD:
    SUPPORTED_TEMPLATES = ["T2A"]
else:
    SUPPORTED_TEMPLATES = []
@pytest.mark.parametrize("cpu_template", SUPPORTED_TEMPLATES )
```

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+@pytest.mark.skipif(
+    platform.machine() != "x86_64", reason="CPU templates are only available on x86_64."
+)
+@pytest.mark.skipif(
+    cpuid_utils.get_cpu_vendor()
+    not in [cpuid_utils.CpuVendor.INTEL, cpuid_utils.CpuVendor.AMD],
+    reason="CPU templates are only available on Intel and AMD.",
+)
```
JonathanWoollett-Light, line 144: Can we use a constant to simplify these?
```suggestion
CPUID_MANUFACTURER_SUPPORTED = cpuid_utils.get_cpu_vendor() in [cpuid_utils.CpuVendor.INTEL, cpuid_utils.CpuVendor.AMD]
@pytest.mark.skipif(
    not CPUID_MANUFACTURER_SUPPORTED ,
    reason="CPU templates are only available on Intel and AMD.",
)
```

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,407 @@
...
+    )
+    vm = vm_instance.vm
+    vm.start()
+
+    check_arch_cap_msr(vm, cpu_template)
+
+
+@pytest.mark.skipif(
+    platform.machine() != "x86_64", reason="CPU templates are only available on x86_64."
+)
```
JonathanWoollett-Light, line 368: Maybe this can be simplified with:
```suggestion
X86_64 = platform.machine() == "X86_64"
@pytest.mark.skipif(
    not X86_64, reason="CPU templates are only available on x86_64."
)
```

tests/integration_tests/functional/test_feat_parity.py
zulinx86: I think we'll need to use the mapping between vendor and templates in other tests as well. So how about defining this mapping in `tests/framework/utils_cpuid.py` or `tests/framework/defs.py`?

kalyazin: I'm inclined to wait for https://github.com/firecracker-microvm/firecracker/pull/3326/files to get merged and then reuse the helper from utils_cpuid.py to simplify this.

tests/integration_tests/functional/test_feat_parity.py
kalyazin: done

tests/integration_tests/functional/test_feat_parity.py
kalyazin: done

src/cpuid/src/template/intel/t2.rs
```diff
@@ -101,7 +102,12 @@ pub(crate) fn update_structured_extended_entry(
             .write_bit(ecx::UMIP_BITINDEX, false)
             .write_bit(ecx::PKU_BITINDEX, false)
             .write_bit(ecx::OSPKE_BITINDEX, false)
+            .write_bit(ecx::AVX512_VBMI2, false)
+            .write_bit(ecx::GFNI, false)
```
JonathanWoollett-Light, line 106: This bit flag is in `ecx`.

kalyazin:
Disabled MWAIT_EXTENDED explicitly in T2/T2CL/T2A (although it was never enabled anyway).

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,402 @@
...
+@pytest.mark.skipif(
+    not CPU_VENDOR_SUPPORTED,
+    reason="CPU templates are only available on Intel and AMD.",
+)
+@pytest.mark.parametrize(
+    "cpu_template",
+    ["C3", "T2", "T2S", "T2CL"]
+    if cpuid_utils.get_cpu_vendor() == cpuid_utils.CpuVendor.INTEL
+    else ["T2A"],
+)
```
JonathanWoollett-Light, line 120: Nit:

When using `SUPPORTED_TEMPLATES` this can be defined like:
```suggestion
@pytest.mark.parametrize(
    "cpu_template",
    intersection(["C3", "T2", "T2S", "T2CL", "T2A"], SUPPORTED_TEMPLATES)
)
```
where:
```python
def intersection(lst1, lst2):
    lst3 = [value for value in lst1 if value in lst2]
    return lst3
```
this avoids checking the vendor repeatedly, and makes the checks:
```python
@pytest.mark.skipif(not X86_64, reason="CPU templates are only available on x86_64.")
@pytest.mark.skipif(
    not CPU_VENDOR_SUPPORTED,
    reason="CPU templates are only available on Intel and AMD.",
)
```
uneeded as in these cases `"cpu_template" = []` thus the test will be skipped.

JonathanWoollett-Light (dismissed):

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,391 @@
...
+from framework.matrix import TestMatrix, TestContext
+from framework.builder import MicrovmBuilder
+import framework.utils_cpuid as cpuid_utils
+import host_tools.network as net_tools
+
+
+INTEL_TEMPLATES = ["C3", "T2", "T2CL", "T2S"]
+AMD_TEMPLATES = ["T2A"]
+ALL_TEMPLATES = INTEL_TEMPLATES + AMD_TEMPLATES
+INST_SET_TEMPLATES = ["T2A", "T2CL"]
```
JonathanWoollett-Light, line 19: Would be good to have comments on these constants.

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,395 @@
...
+def get_supported_templates():
+    """
+    Returns the list of CPU templates supported by the platform.
+    """
+    vendor = cpuid_utils.get_cpu_vendor()
+    if vendor == cpuid_utils.CpuVendor.INTEL:
+        return INTEL_TEMPLATES
+    if vendor == cpuid_utils.CpuVendor.AMD:
+        return AMD_TEMPLATES
+    return []
```
JonathanWoollett-Light, line 35: Do we need to check if x86_64 here?

kalyazin:
Minor update to the doc (versioning.md).

kalyazin:
Moved out CPU template-related utils into a separate file so they can be reused by https://github.com/firecracker-microvm/firecracker/pull/3326 and others.

tests/integration_tests/functional/test_feat_parity.py
pb8o: You could make a fixture that parametrizes the test right templates. Look at `firecracker_release` for an example.

tests/framework/utils_cpuid.py
```diff
@@ -101,3 +102,35 @@ def check_guest_cpuid_output(
...
+
+
+def build_cpuid_dict(raw_cpuid_output):
+    """Build CPUID dict based on raw cpuid output"""
+    cpuid_dict = {}
+    ptrn = re.compile("^ *(.*) (.*): eax=(.*) ebx=(.*) ecx=(.*) edx=(.*)$")
+    for line in raw_cpuid_output:
+        res = ptrn.findall(line)
+        if len(res):
+            leaf, subleaf, eax, ebx, ecx, edx = res[0]
```
pb8o, line 114: Maybe 

```python
... = [int(x, 16) for x in res[0]]
```

so you don't have to convert below?

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,155 @@
...
+
+SUPPORTED_TEMPLATES = get_supported_templates()
+
+
+def intersection(lst1, lst2):
+    """
+    Return the list that is the intersection of two lists.
+    """
+    lst3 = [value for value in lst1 if value in lst2]
+    return lst3
```
pb8o: You could use `set()s` instead, which have an intersection method.

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,155 @@
...
+        "builder": MicrovmBuilder(bin_cloner_path),
+        "network_config": network_config,
+        "cpu_template": cpu_template,
+        "flags_must_be_set": flags_must_be_set,
+        "flags_must_be_unset": flags_must_be_unset,
+    }
+    test_matrix = TestMatrix(
+        context=test_context,
+        artifact_sets=[microvm_artifacts, kernel_artifacts, disk_artifacts],
+    )
```
pb8o, line 103: See https://github.com/firecracker-microvm/firecracker/commit/b646a3388c8ee5d89225e399c78d0571210512c0 for an example of of using fixtures instead of TestMatrix.

tests/framework/utils_cpu_templates.py
```diff
@@ -0,0 +1,44 @@
...
+    lst3 = [value for value in lst1 if value in lst2]
+    return lst3
+
+
+def select_supported_cpu_templates(templates):
+    """
+    Return an intersection between the supplied list of CPU templates
+    and all the supported CPU templates.
+    """
+    return intersection(SUPPORTED_CPU_TEMPLATES, templates)
```
pb8o, line 44: I think this could all be a fixture that does the right thing depending on the cpu.

See this one for an example: https://github.com/firecracker-microvm/firecracker/commit/b646a3388c8ee5d89225e399c78d0571210512c0

docs/snapshotting/versioning.md
```diff
@@ -185,15 +185,21 @@ are an invariant when saving and restoring the snapshot. The trivial scenario
...
 of CPU features. The T2S template is designed to allow migrating snapshots
 between hosts with Intel SkyLake and Cascade Lake securely by further
 restricting CPU features for the guest, however this comes with a performance
 penalty. Users are encouraged to carry out a performance assessment if they wish
-to use the T2S template. There are no templates available for AMD or ARM64.
+to use the T2S template.
+The T2CL template is mapped to be close to Intel Cascade Lake.
+It is not safe to use it on Intel CPUs older than Cascade Lake (such as Skylake).
+
+The only AMD template is T2A. It is considered safe to be used with AMD Milan.
```
zulinx86, line 200: How about providing more description of why `T2CL` and `T2A` template exists? (e.g., designed to provide instruction set parity and allow snapshots migration between Intel Cascade Lake and AMD Milan)

src/cpuid/src/cpu_leaf.rs
```diff
@@ -296,13 +296,19 @@ pub mod leaf_0x80000001 {
     pub const LEAF_NUM: u32 = 0x8000_0001;
 
     pub mod ecx {
-        pub const TOPOEXT_INDEX: u32 = 22;
+        pub const SSE4A: u32 = 6; // EXTRQ, INSERTQ, MOVNTSS, and MOVNTSD instruction support
+        pub const MISALIGN_SSE: u32 = 7; // Misaligned SEE mode
         pub const PREFETCH_BITINDEX: u32 = 8; // 3DNow! PREFETCH/PREFETCHW instructions
+        pub const TOPOEXT_INDEX: u32 = 22; // topology extensions support
         pub const LZCNT_BITINDEX: u32 = 5; // advanced bit manipulation
```
zulinx86, line 303: Fixes the order of bits (in ascending order).
```suggestion
        pub const LZCNT_BITINDEX: u32 = 5; // advanced bit manipulation
        pub const SSE4A: u32 = 6; // EXTRQ, INSERTQ, MOVNTSS, and MOVNTSD instruction support
        pub const MISALIGN_SSE: u32 = 7; // Misaligned SEE mode
        pub const PREFETCH_BITINDEX: u32 = 8; // 3DNow! PREFETCH/PREFETCHW instructions
        pub const TOPOEXT_INDEX: u32 = 22; // topology extensions support
```

zulinx86 (changes requested):

tests/integration_tests/functional/test_feat_parity.py
kalyazin: If I do that, I would need to repeat the following for each test:
```python
@pytest.mark.parametrize(
    "disk", ARTIFACTS.disks(keyword="bionic-msrtools"), ids=lambda disk: disk.name()
)
@pytest.mark.parametrize(
    "kernel", ARTIFACTS.kernels(), ids=lambda kernel: kernel.name()
)
@pytest.mark.parametrize(
    "microvm", ARTIFACTS.microvms(keyword="1vcpu_1024mb"), ids=lambda uvm: uvm.name()
)
@pytest.mark.parametrize(
    "cpu_template",
    cputmpl_utils.select_supported_cpu_templates(cputmpl_utils.ALL_TEMPLATES),
)
def test_feat_parity_cpuid_mpx(bin_cloner_path, network_config, microvm, kernel, disk, cpu_template):
```
Do you think there is a way to avoid that?

pb8o: I think we could re-use some common fixtures in conftest and add a few specific ones for this file. I haven't tested but I think something like this should work:

```python
# Define common fixtures

@pytest.fixture(
    params=ARTIFACTS.microvms(keyword="1vcpu_1024mb"), ids=lambda uvm: uvm.name()
)
def microvm(request):
    """Common fixture for tests in this file"""
    return request.param

@pytest.fixture(
    params=cputmpl_utils.select_supported_cpu_templates(cputmpl_utils.ALL_TEMPLATES),
)
def cpu_template(request):
    return request.param

# Then each test becomes

@pytest.mark.parametrize("rootfs", params=ARTIFACTS.disks("bionic-msrtools"), indirect=True)
def test_feat_parity_cpuid_mpx(bin_cloner_path, network_config, microvm, guest_kernel, rootfs, cpu_template):
    pass
```

tests/framework/utils_cpu_templates.py
pb8o: See my above comment, the idea is that the fixture takes care of running a test for each valid template.

tests/integration_tests/functional/test_feat_parity.py
kalyazin: done

tests/framework/utils_cpuid.py
kalyazin: done

tests/integration_tests/functional/test_feat_parity.py
kalyazin: done

tests/integration_tests/functional/test_feat_parity.py
kalyazin: done

tests/framework/utils_cpu_templates.py
kalyazin: done

src/cpuid/src/template/intel/t2.rs
```diff
@@ -74,6 +74,7 @@ pub(crate) fn update_structured_extended_entry(
             .write_bit(ebx::SGX_BITINDEX, false)
             .write_bit(ebx::HLE_BITINDEX, false)
             .write_bit(ebx::FPDP_BITINDEX, false)
+            .write_bit(ebx::ERMS, true)
```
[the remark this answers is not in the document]
kalyazin: ERMS is not set by AMD, so this is done explicitly. MWAIT_EXTENDED is set on AMD host, so can be safer to clear it. AVX512_VP2INTERSECT is not set by any CPU tested by us, but may be set by some newer ones in the future.

src/cpuid/src/template/intel/t2.rs
```diff
@@ -157,13 +165,28 @@ pub(crate) fn update_extended_feature_info_entry(
...
 
+pub(crate) fn update_extended_feature_extensions_entry(
+    entry: &mut kvm_cpuid_entry2,
+    _vm_spec: &VmSpec,
+) -> Result<(), Error> {
+    use crate::cpu_leaf::leaf_0x80000008::*;
+
+    if entry.index == 0 {
+        entry.ebx.write_bit(ebx::WBNOINVD, false);
+    }
```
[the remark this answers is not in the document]
kalyazin: done

src/cpuid/src/cpu_leaf.rs
```diff
@@ -126,7 +126,8 @@ pub mod leaf_0x7 {
             pub const FPDP_BITINDEX: u32 = 6;
             // 7 = SMEP (Supervisor-Mode Execution Prevention if 1)
             pub const BMI2_BITINDEX: u32 = 8;
-            // 9 = Enhanced REP MOVSB/STOSB if 1
+            // ERMS = Enhanced REP MOVSB/STOSB if 1
+            pub const ERMS: u32 = 9;
```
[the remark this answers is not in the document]
kalyazin: done

docs/snapshotting/versioning.md
kalyazin: done

src/cpuid/src/cpu_leaf.rs
kalyazin: done

tests/integration_tests/functional/test_feat_parity.py
JonathanWoollett-Light: I was wondering `cpuid_utils.CpuVendor` is valid on aarch64 for checking this. If it is I have no other concerns.

docs/snapshotting/versioning.md
```diff
@@ -185,15 +185,28 @@ are an invariant when saving and restoring the snapshot. The trivial scenario
 is creating and restoring snapshots on hosts that have the same CPU model.
 
 To make snapshots more portable across Intel CPUs Firecracker provides an API to
-select a CPU template which is only available for Intel - T2, T2S and C3.
-Firecracker CPU templates mask CPUID and some MSR values (in case of T2S) to
-restrict the exposed features to a common denominator of multiple CPU models. T2
-and C3 templates are mapped as close as possible to AWS T2/C3 instances in terms
+select an Intel CPU template: T2, T2CL, T2S and C3.
```
JonathanWoollett-Light, line 188: Nit.
```suggestion
select an Intel CPU template (T2, T2CL, T2S or C3).
```

docs/snapshotting/versioning.md
```diff
@@ -185,15 +185,28 @@ are an invariant when saving and restoring the snapshot. The trivial scenario
...
-to use the T2S template. There are no templates available for AMD or ARM64.
+to use the T2S template.
+The T2CL template is mapped to be close to Intel Cascade Lake.
+It is not safe to use it on Intel CPUs older than Cascade Lake (such as Skylake).
+
+The only AMD template is T2A. It is considered safe to be used with AMD Milan.
+
+Intel T2CL and AMD T2A templates together aim to provide instruction set feature
+parity between CPUs running them, so they can form a heterogeniuous fleet
+exposing the same instruction sets to the application.
```
JonathanWoollett-Light, line 204:
```suggestion
Intel T2CL and AMD T2A templates together aim to provide instruction set feature
parity between CPUs running them, so they can form a heterogeneous fleet
exposing the same instruction sets to the application.
```

docs/snapshotting/versioning.md
```diff
@@ -185,15 +185,28 @@ are an invariant when saving and restoring the snapshot. The trivial scenario
...
+
+Taking a snapshot on Intel and restoring from it on AMD (or vice versa) is not
+currently supported.
```
JonathanWoollett-Light, line 207:
```suggestion
Restoring from an Intel snapshot on AMD (or vice-versa) is not supported.
```

src/cpuid/src/cpu_leaf.rs
```diff
@@ -285,19 +296,39 @@ pub mod leaf_0x80000001 {
...
+        // FXSAVE, XSAVE, FXSAVEOPT, XSAVEC, XSAVES always save error pointers and
+        // FXRSTOR, XRSTOR, XRSTORS always restore error pointers is supported
+        pub const RSTR_FP_ERR_PTRS: u32 = 2;
+        // Write Back and Do Not Invalidate Cache instruction
+        pub const WBNOINVD: u32 = 9;
+        // IBRS is preferred over software solution
+        pub const IBRS_PREFERRED: u32 = 18;
+        // IBRS provides Same Mode Protection
+        pub const IBRS_PROVIDES_SAME_MODE_PROTECTION: u32 = 19;
+    }
```
JonathanWoollett-Light, line 331: Should these have the `BITINDEX` suffix?

src/cpuid/src/template/amd/t2a.rs
```diff
@@ -0,0 +1,74 @@
...
+            .write_bit(ebx::WBNOINVD, false)
+            .write_bit(ebx::IBRS_PREFERRED, true)
+            .write_bit(ebx::IBRS_PROVIDES_SAME_MODE_PROTECTION, true);
+    }
+
+    Ok(())
+}
+
+/// Sets up the cpuid entries for a given VCPU following a T2A template.
+struct T2ACpuidTransformer {}
```
JonathanWoollett-Light, line 55: Nit.
```suggestion
struct T2ACpuidTransformer;
```

src/cpuid/src/template/intel/t2cl.rs
```diff
@@ -0,0 +1,53 @@
...
+
+use arch::x86_64::msr::{ArchCapaMSRFlags, MSR_IA32_ARCH_CAPABILITIES};
+use kvm_bindings::{kvm_cpuid_entry2, kvm_msr_entry, CpuId};
+
+use crate::cpu_leaf::*;
+use crate::template::intel::validate_vendor_id;
+use crate::transformer::*;
+
+/// Sets up the CPUID entries for a given VCPU following the T2CL template.
+struct T2CLCpuidTransformer {}
```
JonathanWoollett-Light, line 12: Nit
```suggestion
struct T2CLCpuidTransformer;
```

src/cpuid/src/template/intel/t2cl.rs
```diff
@@ -0,0 +1,53 @@
...
+            leaf_0x80000008::LEAF_NUM => Some(crate::t2::update_extended_feature_extensions_entry),
+            _ => None,
+        }
+    }
+}
+
+/// Sets up the CPUID entries for a given VCPU following the T2CL template.
+pub fn set_cpuid_entries(kvm_cpuid: &mut CpuId, vm_spec: &VmSpec) -> Result<(), Error> {
+    validate_vendor_id()?;
+    T2CLCpuidTransformer {}.process_cpuid(kvm_cpuid, vm_spec)
```
JonathanWoollett-Light, line 30: With https://github.com/firecracker-microvm/firecracker/pull/3324/files#r1067218200 this would be
```suggestion
    T2CLCpuidTransformer.process_cpuid(kvm_cpuid, vm_spec)
```

src/cpuid/src/template/intel/t2cl.rs
```diff
@@ -0,0 +1,53 @@
...
+        ..Default::default()
+    });
+}
+
+static EXTRA_MSR_ENTRIES: &[u32] = &[MSR_IA32_ARCH_CAPABILITIES];
+
+/// Return a list of MSRs speciffic to this T2CL template.
+pub fn msr_entries_to_save() -> &'static [u32] {
+    EXTRA_MSR_ENTRIES
+}
```
JonathanWoollett-Light, line 53: These seems unnecessary. In both usages here and in `t2s` they return a single value. It would far simpler to me to just include this value in the place it is used.
```rust
        if vcpu_config.cpu_template == CpuFeaturesTemplate::T2S {
            self.msr_list.extend_one(MSR_IA32_ARCH_CAPABILITIES);
            t2s::update_msr_entries(&mut msr_boot_entries);
        }
        if vcpu_config.cpu_template == CpuFeaturesTemplate::T2CL {
            self.msr_list.extend_one(MSR_IA32_ARCH_CAPABILITIES);
            t2cl::update_msr_entries(&mut msr_boot_entries);
        }
```
replaces
```rust
        if vcpu_config.cpu_template == CpuFeaturesTemplate::T2S {
            self.msr_list.extend(t2s::msr_entries_to_save());
            t2s::update_msr_entries(&mut msr_boot_entries);
        }
        if vcpu_config.cpu_template == CpuFeaturesTemplate::T2CL {
            self.msr_list.extend(t2cl::msr_entries_to_save());
            t2cl::update_msr_entries(&mut msr_boot_entries);
        }
```
and we can remove:
```rust
static EXTRA_MSR_ENTRIES: &[u32] = &[MSR_IA32_ARCH_CAPABILITIES];

/// Return a list of MSRs speciffic to this T2S template.
pub fn msr_entries_to_save() -> &'static [u32] {
    EXTRA_MSR_ENTRIES
}
```
and
```rust
static EXTRA_MSR_ENTRIES: &[u32] = &[MSR_IA32_ARCH_CAPABILITIES];

/// Return a list of MSRs speciffic to this T2CL template.
pub fn msr_entries_to_save() -> &'static [u32] {
    EXTRA_MSR_ENTRIES
}
```
```

File: tests/integration_tests/functional/test_feat_parity.py [hunk 26]
```diff
@@ -0,0 +1,343 @@
...
+    ]
+    # fmt: on
+
+    vendor = cpuid_utils.get_cpu_vendor()
+    if vendor == cpuid_utils.CpuVendor.INTEL:
+        must_be_set = must_be_set_common + must_be_set_intel_only
+        must_be_unset = must_be_unset_common + must_be_unset_intel_only
+    elif vendor == cpuid_utils.CpuVendor.AMD:
+        must_be_set = must_be_set_common + must_be_set_amd_only
+        must_be_unset = must_be_unset_common + must_be_unset_amd_only
```
JonathanWoollett-Light, line 298: Nit
```suggestion
    if vendor == cpuid_utils.CpuVendor.INTEL:
        must_be_set = must_be_set_common + must_be_set_intel_only
        must_be_unset = must_be_unset_common + must_be_unset_intel_only
    elif vendor == cpuid_utils.CpuVendor.AMD:
        must_be_set = must_be_set_common + must_be_set_amd_only
        must_be_unset = must_be_unset_common + must_be_unset_amd_only
    else:
        raise Exception("Unsupported CPU vendor.")
```

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,343 @@
...
+        expected = (
+            (1 << 0) | # RDCL_NO
+            (1 << 1) | # IBRS_ALL
+            (1 << 3) | # SKIP_L1DFL_VMENTRY
+            (1 << 5) | # MDS_NO
+            (1 << 6) | # IF_PSCHANGE_MC_NO
+            (1 << 7) # TSX_CTRL
+        )
+        # fmt: on
+        assert actual == expected, f"{actual=:#x}, {expected=:#x}"
```
JonathanWoollett-Light, line 340: Nit
```suggestion
       assert actual == expected, f"{actual=:#x} != {expected=:#x}"
```

src/cpuid/src/template/intel/t2cl.rs
kalyazin: I would disagree with this. The list of extra entries (although it's just one MSR) is now located in a file that is specific to the template definition. It does not look natural to me for ```vstate/vcpu/x86_64.rs``` to know about details of the CPU template. If we ever need to add more MSRs (which is not totally impossible), it would be simpler to do with the current implementation.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -310,6 +314,10 @@ impl KvmVcpu {
             self.msr_list.extend(t2s::msr_entries_to_save());
             t2s::update_msr_entries(&mut msr_boot_entries);
         }
+        if vcpu_config.cpu_template == CpuFeaturesTemplate::T2CL {
+            self.msr_list.extend(t2cl::msr_entries_to_save());
+            t2cl::update_msr_entries(&mut msr_boot_entries);
+        }
```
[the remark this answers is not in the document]
kalyazin: same reply

tests/integration_tests/functional/test_feat_parity.py
kalyazin: I can see it used in other tests. I don't expect for it to be broken.

src/cpuid/src/template/intel/t2cl.rs
JonathanWoollett-Light: Okay

tests/integration_tests/functional/test_feat_parity.py
kalyazin: done

tests/integration_tests/functional/test_feat_parity.py
kalyazin: done

src/cpuid/src/template/mod.rs
```diff
@@ -3,19 +3,21 @@
...
-use crate::common::{get_vendor_id_from_cpuid, VENDOR_ID_INTEL};
+use crate::common::{get_vendor_id_from_cpuid, VENDOR_ID_AMD, VENDOR_ID_INTEL};
 use crate::transformer::Error;
 
 /// Returns MSRs to be saved based on CPUID features that are enabled.
 pub fn msrs_to_save_by_cpuid(cpuid: &CpuId) -> Result<HashSet<u32>, Error> {
     let vendor_id = get_vendor_id_from_cpuid(cpuid).map_err(|_| Error::InvalidVendor)?;
     match &vendor_id {
-        VENDOR_ID_INTEL => Ok(intel::msrs_to_save_by_cpuid(cpuid)),
+        VENDOR_ID_INTEL | VENDOR_ID_AMD => Ok(intel::msrs_to_save_by_cpuid(cpuid)),
```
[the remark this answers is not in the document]
kalyazin: done

src/cpuid/src/template/intel/t2cl.rs
kalyazin: done

src/cpuid/src/template/amd/t2a.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    if entry.index == 0 {
+        entry
+            .ebx
+            .write_bit(ebx::CLZERO, false)
+            .write_bit(ebx::RSTR_FP_ERR_PTRS, false)
+            .write_bit(ebx::WBNOINVD, false)
+            .write_bit(ebx::IBRS_PREFERRED, true)
+            .write_bit(ebx::IBRS_PROVIDES_SAME_MODE_PROTECTION, true);
+    }
```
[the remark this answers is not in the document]
kalyazin: done

src/cpuid/src/cpu_leaf.rs
kalyazin: done

docs/snapshotting/versioning.md
kalyazin: done

docs/snapshotting/versioning.md
kalyazin: done

src/cpuid/src/template/intel/t2.rs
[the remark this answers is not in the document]
zulinx86: I'm a little bit confused about the concept. Let me check my understanding. T2 template intends to provide the instruction set-wise parity with AWS T2 instances to uVMs on x86 CPUs. Previously we explicitly enabled/disabled bits only which are inconsistent on uVMs with no templates on currently supported CPUs. But with the implementation, every when we want to add new CPU support, we need to check if it has inconsistency and change templates as needed. Then, this commit explicitly enables/disables all the required bits regardless of whether those bits are enabled/disabled by default on currently supported CPUs, in order to make it consistent for new CPUs support in the future. Please correct me if I'm wrong. If correct, please resolve this conversation.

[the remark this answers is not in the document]
zulinx86: If the above understanding is correct, we still need to enable/disable explicitly some other bits (e.g. `SSE4A` and `MIS_ALIGN_SSE` on `CPUID.80000001H:ECX`, `MMX_EXT`, `MMX`, `FXSR` and `FFXSR` on `CPUID.8000008H:EDX`, etc...).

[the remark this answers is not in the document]
zulinx86: I’m bit confused. Let me check my understanding. T2 template intends to provide the instruction set-wise parity with AWS T2 instances to uVMs on x86 CPUs. Currently we explicitly enabled/disabled bits only which are inconsistent on uVMs with no templates on currently supported CPUs. But with the implementation, every when we want to add new CPU support, we need to check if it has inconsistency and change templates as needed. Then, I agree with you that it is safer to explicitly enables/disables all the required bits regardless of whether those bits are enabled/disabled by default on currently supported CPUs, in order to make it consistent for new CPUs support in the future. Please correct me if I’m wrong. If correct, we still need to enable/disable some other bits according to [your feature parity test](https://github.com/firecracker-microvm/firecracker/pull/3324/commits/7c40a2b0362dbd5a70e27b47c1820eda1ada7d73#diff-57a8deea5cf913bfbffc93f8e9dfa44c57a4e67963eb1ed00f8089f42351e7d7) (like `AVX2` in `CPUID.(EAX=07H,ECX=0):EBX`).

src/cpuid/src/template/mod.rs
```diff
@@ -6,19 +6,125 @@ pub mod intel;
 
 use std::collections::HashSet;
 
+use arch_gen::x86::msr_index::*;
 use kvm_bindings::CpuId;
 
-use crate::common::{get_vendor_id_from_cpuid, VENDOR_ID_INTEL};
+use crate::common::{get_vendor_id_from_cpuid, VENDOR_ID_AMD, VENDOR_ID_INTEL};
```
zulinx86, line 14: nit: `VENDOR_ID_AMD` is not still used in this commit c9b195eaaf34d21320d7568d5a3d97da7ac8fab3.

tests/framework/utils_cpuid.py
```diff
@@ -101,3 +102,35 @@ def check_guest_cpuid_output(
...
+            cpuid_dict[(int(leaf, 16), int(subleaf, 16), "eax")] = int(eax, 16)
+            cpuid_dict[(int(leaf, 16), int(subleaf, 16), "ebx")] = int(ebx, 16)
+            cpuid_dict[(int(leaf, 16), int(subleaf, 16), "ecx")] = int(ecx, 16)
+            cpuid_dict[(int(leaf, 16), int(subleaf, 16), "edx")] = int(edx, 16)
```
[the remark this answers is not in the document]
zulinx86: But the line ("CPU:") is removed since `get_guest_cpuid()` uses `cpuid -1 --raw | grep -v CPU`, right?

src/cpuid/src/template/intel/t2cl.rs
zulinx86: As Jonathan said, it would be better to apply this change to other `Transformer` structs (`C3CpuidTransformer`, `T2CpuidTransformer`, `T2SCpuidTransformer`) for consistency.

src/cpuid/src/template/intel/t2.rs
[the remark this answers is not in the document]
kalyazin: Yes, your understanding is right. The current implementation only fixes up the bits that do not match the expectations if no-template is used. I am tightening it a bit with this change, but I do not perform proactive setting of all the bits. Nor are the tests I implemented exhaustive. We can be stricter with the bit checking, but the return is diminishing with more we do it. The problem is going to be solved holistically by the new CPUID implementation from Jonathan. The change I'm making here is a trade off between our confidence and amount of throw-away effort. I totally agree, the line is very subjective. I have added the 0x80000001 leaf to the T2CL template for consistency.

tests/framework/utils_cpuid.py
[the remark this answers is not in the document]
kalyazin: you're right, fixed.

src/cpuid/src/template/intel/t2cl.rs
kalyazin: done

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,137 @@
...
+    return request.param
+
+
+@pytest.fixture(
+    name="inst_set_cpu_template_ext",
+    params=list(set(SUPPORTED_CPU_TEMPLATES).intersection(INST_SET_TEMPLATES + ["T2"])),
+)
+def inst_set_cpu_template_ext_fxt(request):
+    """CPU template fixture for instruction set feature parity templates plus T2"""
+    return request.param
```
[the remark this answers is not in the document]
kalyazin: done

tests/framework/utils_cpuid.py
```diff
@@ -101,3 +102,35 @@ def check_guest_cpuid_output(
...
+
+
+def build_cpuid_dict(raw_cpuid_output):
+    """Build CPUID dict based on raw cpuid output"""
+    cpuid_dict = {}
+    ptrn = re.compile("^ *(.*) (.*): eax=(.*) ebx=(.*) ecx=(.*) edx=(.*)$")
+    for line in raw_cpuid_output:
+        res = ptrn.findall(line)
+        if len(res):
+            leaf, subleaf, eax, ebx, ecx, edx = [int(x, 16) for x in res[0]]
```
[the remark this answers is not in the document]
kalyazin: done

src/cpuid/src/cpu_leaf.rs
```diff
@@ -208,6 +214,11 @@ pub mod leaf_0x7 {
             pub const AVX512_4VNNIW_BITINDEX: u32 = 2;
             // AVX-512 4-register Multiply Accumulation Single precision
             pub const AVX512_4FMAPS_BITINDEX: u32 = 3;
+            // Fast Short REP MOV
+            pub const FSRM_BITINDEX: u32 = 4;
+            // AVX-512 Compute Intersection Between DWORDS/QUADWORDS
+            pub const AVX512_VP2INTERSECT: u32 = 8;
```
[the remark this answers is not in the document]
kalyazin: done

src/cpuid/src/template/mod.rs
kalyazin: done

src/cpuid/src/template/intel/mod.rs
```diff
@@ -26,107 +20,3 @@ pub fn validate_vendor_id() -> Result<(), Error> {
...
-    // Macro used for easy definition of CPUID-MSR dependencies.
-    macro_rules! cpuid_msr_dep {
-        ($leaf:expr, $index:expr, $reg:tt, $feature_bit:expr, $msr:expr) => {
-            if cpuid_is_feature_set!(cpuid, $leaf, $index, $reg, $feature_bit) {
-                msrs.extend($msr)
-            }
-        };
-    }
-
-    // TODO: Add more dependencies.
```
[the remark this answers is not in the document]
kalyazin: I changed that when I moved the function. Should have not been there anymore.

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def create_vm(bin_cloner_path, cpu_template, microvm, kernel, disk):
+    """
+    Create a VM.
+    """
+    root_disk = disk.copy()
+
+    vm_builder = MicrovmBuilder(bin_cloner_path)
```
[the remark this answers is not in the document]
kalyazin: I created the builder fixture, however I had to leave it local to the test. Making it global turns out to be a non-neglibigle effort. Will be tracking separately.

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+# CPU templates designed to provide instruction set feature parity
+INST_SET_TEMPLATES = []
+
+ARTIFACTS = ArtifactCollection(_test_images_s3_bucket())
+
+
+@pytest.fixture(
+    name="microvm",
```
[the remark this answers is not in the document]
kalyazin: yes, that's right.

tests/integration_tests/functional/test_feat_parity.py
```diff
@@ -0,0 +1,137 @@
...
+def disk_fxt(request):
+    """Common disk fixture for tests in this file"""
+    disk = request.param
+    disk.download()
+    return disk
+
+
+@pytest.fixture(
+    name="inst_set_cpu_template",
+    params=list(set(SUPPORTED_CPU_TEMPLATES).intersection(INST_SET_TEMPLATES)),
```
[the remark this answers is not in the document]
kalyazin: it does indeed, thanks.

tests/integration_tests/functional/test_feat_parity.py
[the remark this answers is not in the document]
pb8o: Yeah I see it creates some circular dependencies.

JonathanWoollett-Light (approved):

zulinx86 (dismissed):

src/cpuid/src/template/intel/t2.rs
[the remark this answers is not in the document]
zulinx86: > The problem is going to be solved holistically by the new CPUID implementation from Jonathan.

Sounds super cool!

I got your point. Thanks for your explanation!

tests/framework/utils_cpu_templates.py
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""Utilities for CPU template related functionality."""
+
+import framework.utils_cpuid as cpuid_utils
+
+# All existing CPU templates available on Intel
+INTEL_TEMPLATES = ["C3", "T2", "T2CL", "T2S"]
+# All existing CPU templates available on AMD
+AMD_TEMPLATES = ["T2A"]
```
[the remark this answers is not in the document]
kalyazin: That's very true, thanks! Done.

pb8o (approved):

PR #3333, opened 2022-12-22
Increase snapshot target for 512mb VM on aarch64/amd64

mattschlebusch:
## Changes

* Increase the snapshot load test performance target on aarch64 hosts for Kernel 5.10 using the async file engine.
* Increase the snapshot load test performance target on amd hosts for Kernel 4.14 using the sync file engine.
* Adding rounding of some floating point values in the serialization benchmark test

## Reason

Test failure for continuous integration seen in Buildkite ([ARM test](https://buildkite.com/firecracker/firecracker-pr/builds/242#01853975-5022-42eb-8f76-cb78fff3fb7d), [AMD test](https://buildkite.com/firecracker/firecracker-pr/builds/257#01853ec2-3e6a-496f-be2a-bd04ccc47e88)). This follows recent tweaks to that target in [PR/3291](https://github.com/firecracker-microvm/firecracker/pull/3291). A slightly increased time between the 256MB guest and the 512MB guest was not maintained when the previous update was made, however it isn't unreasonable to expect a slightly slower snapshot load for a larger guest.

Regarding the rounding added to the serialization benchmark. This was added as a result of arithmetic inaccuracy in Python (see a [test failure as a result](https://buildkite.com/firecracker/firecracker-pr/builds/244#018539a9-287e-49da-91f4-44a91a3155cahttps://buildkite.com/firecracker/firecracker-pr/builds/244#018539a9-287e-49da-91f4-44a91a3155ca)) which results in incorrectly failing tests. This is due to a binary representation of floating point numbers which results in values that are slightly off what was configured to be tested against. See [this guide](https://floating-point-gui.de/basic/) for more context. Another alternative is to use the `Decimal` type explicitly but is more verbose and does not seem necessary given the floating point resolution that the test requires.

## License Acceptance

### DIFF
--- a/tests/integration_tests/performance/test_versioned_serialization_benchmark.py
+++ b/tests/integration_tests/performance/test_versioned_serialization_benchmark.py
@@ -68,8 +68,8 @@ def _check_statistics(directory, mean):
         attribute = "no-crc"
 
     measure = BASELINES[proc_model[0]][bench][attribute]
-    low = measure["target"] - measure["delta"]
-    high = measure["target"] + measure["delta"]
+    low = round(measure["target"] - measure["delta"], 3)
+    high = round(measure["target"] + measure["delta"], 3)
     assert low <= mean <= high, "Benchmark result {} has changed!".format(directory)
 
     return directory, f"{mean} ms", f"{low} <= result <= {high}"
--- a/tests/integration_tests/performance/test_snapshot_perf.py
+++ b/tests/integration_tests/performance/test_snapshot_perf.py
@@ -138,7 +138,7 @@ LOAD_LATENCY_BASELINES = {
                 },
                 "async": {
                     "2vcpu_256mb.json": {"target": 320},
-                    "2vcpu_512mb.json": {"target": 310},
+                    "2vcpu_512mb.json": {"target": 330},
                 },
             },
         }
--- a/tests/integration_tests/performance/test_snapshot_perf.py
+++ b/tests/integration_tests/performance/test_snapshot_perf.py
@@ -90,7 +90,7 @@ LOAD_LATENCY_BASELINES = {
             "4.14": {
                 "sync": {
                     "2vcpu_256mb.json": {"target": 15},
-                    "2vcpu_512mb.json": {"target": 15},
+                    "2vcpu_512mb.json": {"target": 19},
                 }
             },
             "5.10": {


roypat:
I'm a bit confused by how `round` in python works. I'd expect it to fail if you give it a number that cannot be represented as an IEEE754 float, but `round(1.0-0.8, 2)` actually prints `0.2` instead of `0.1999...`. On the other hand, `round(5.55, 1)` gives `5.5` instead of `5.6`. 

How do you feel about instead of using `round` to just multiply by 100, round to int, and use integer comparison? 😅 I'm just not 100% convinced that float weirdness won't bite us again in the future

mattschlebusch:
> I'm a bit confused by how `round` in python works. I'd expect it to fail if you give it a number that cannot be represented as an IEEE754 float, but `round(1.0-0.8, 2)` actually prints `0.2` instead of `0.1999...`. On the other hand, `round(5.55, 1)` gives `5.5` instead of `5.6`.
> 
> How do you feel about instead of using `round` to just multiply by 100, round to int, and use integer comparison? sweat_smile I'm just not 100% convinced that float weirdness won't bite us again in the future

I think extra arithmetic would do the trick, but is a bit messier to read personally. The examples you provided are consistent I think though.

`round(1.0-0.8, 2)` rounds as I (we?) would expect, but `round(5.55, 1)` is odd for sure. It's important to keep in mind that Python's default rounding configuration is rounding to halfway-even numbers, which is non-standard in my opinion and would need to be overridden manually via the `Decimal` type to correct for. See more [detail here](https://docs.python.org/3/library/decimal.html#decimal.DefaultContext). This *ROUND_HALF_EVEN* specific nuance could only affect the test for `.******5` values. However given the binary format is always a close representation at 28 digits of precision, I don't see `round` being a problem here.

roypat (approved):
> > I'm a bit confused by how `round` in python works. I'd expect it to fail if you give it a number that cannot be represented as an IEEE754 float, but `round(1.0-0.8, 2)` actually prints `0.2` instead of `0.1999...`. On the other hand, `round(5.55, 1)` gives `5.5` instead of `5.6`.
> > How do you feel about instead of using `round` to just multiply by 100, round to int, and use integer comparison? sweat_smile I'm just not 100% convinced that float weirdness won't bite us again in the future
> 
> I think extra arithmetic would do the trick, but is a bit messier to read personally. The examples you provided are consistent I think though.
> 
> `round(1.0-0.8, 2)` rounds as I (we?) would expect, but `round(5.55, 1)` is odd for sure. It's important to keep in mind that Python's default rounding configuration is rounding to halfway-even numbers, which is non-standard in my opinion and would need to be overridden manually via the `Decimal` type to correct for. See more [detail here](https://docs.python.org/3/library/decimal.html#decimal.DefaultContext). This _ROUND_HALF_EVEN_ specific nuance could only affect the test for `.******5` values. However given the binary format is always a close representation at 28 digits of precision, I don't see `round` being a problem here.

Huh, that sure is interesting behavior, didn't know about that, thanks! 

My main confusion about the `1.0-0.8` was that 0.2 is not representable as IEEE754 float (iirc), which is why just writing `1.0-0.8` gives you `0.19999999999999996`. However `round(1.0-0.8, 2)` does print `0.2` while still seemingly being a float (`type(round(1.0-0.8, 2)) == <class 'float'>`), which I found curious. 

Although, whatever is going on, it seems to work, sooo guess we don't have to worry

tests/integration_tests/performance/test_versioned_serialization_benchmark.py
```diff
@@ -68,8 +68,8 @@ def _check_statistics(directory, mean):
         attribute = "no-crc"
 
     measure = BASELINES[proc_model[0]][bench][attribute]
-    low = measure["target"] - measure["delta"]
-    high = measure["target"] + measure["delta"]
+    low = round(measure["target"] - measure["delta"], 3)
+    high = round(measure["target"] + measure["delta"], 3)
```
pb8o, line 72: `round` in Python is different to other languages. Are we sure we want `round` and not `math.trunc` (round towards 0)

PR #3340, opened 2022-12-28
Rename `vm-memory` to `vm-memory-wrapper`

JonathanWoollett-Light:
## Changes

Rename `vm-memory` to `vm-memory-wrapper` and remove renaming of `vm-memory` to `vm-memory-upstream`.

## Reason

Renaming dependencies negatively affects readability, one will typically assume when seeing `use some_crate` in the code base it refers to the external crate `some_crate` which could be looked up at `crates.io/crates/some_crate`.

By removing this alias and renaming the local crate, readability is improved.

This was previously a part of https://github.com/firecracker-microvm/firecracker/pull/3079.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/.buildkite/pipeline_pr.py
+++ b/.buildkite/pipeline_pr.py
@@ -51,7 +51,7 @@ step_style = {
     "command": "./tools/devtool -y test -- ../tests/integration_tests/style/",
     "label": "🪶 Style",
     # we only install the required dependencies in x86_64
-    "agents": ["platform=x86_64.metal"]
+    "agents": ["platform=x86_64.metal"],
 }
 
 build_grp = group(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,7 +88,7 @@ dependencies = [
  "versionize",
  "versionize_derive",
  "vm-fdt",
- "vm-memory 0.3.0",
+ "vm-memory-wrapper",
 ]
 
 [[package]]
@@ -436,7 +436,7 @@ dependencies = [
  "versionize",
  "versionize_derive",
  "virtio_gen",
- "vm-memory 0.3.0",
+ "vm-memory-wrapper",
  "vm-superio",
 ]
 
@@ -599,7 +599,7 @@ dependencies = [
  "libc",
  "proptest",
  "utils",
- "vm-memory 0.3.0",
+ "vm-memory-wrapper",
 ]
 
 [[package]]
@@ -691,7 +691,7 @@ version = "0.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9259ddbfbb52cc918f6bbc60390004ddd0228cf1d85f402009ff2b3d95de83f"
 dependencies = [
- "vm-memory 0.10.0",
+ "vm-memory",
 ]
 
 [[package]]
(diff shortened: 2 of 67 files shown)


ShadowCurse:
I guess this change should be mentioned in `CHANGELOG.md` as well.

dianpopa:
So with this change the local `vm-memory` crate is renamed to `vm-memory-ext`? What does `ext` stand for? To me it sounds like external which is the opposite to what we want to call it.

ShadowCurse:
@dianpopa I guess `ext` is for `extension`

JonathanWoollett-Light:
> So with this change the local `vm-memory` crate is renamed to `vm-memory-ext`? What does `ext` stand for? To me it sounds like external which is the opposite to what we want to call it.

As @ShadowCurse noted `ext` stands for `extension`, if this is not clear maybe `wrapper` is better?

JonathanWoollett-Light:
> So with this change the local `vm-memory` crate is renamed to `vm-memory-ext`? What does `ext` stand for? To me it sounds like external which is the opposite to what we want to call it.

As @ShadowCurse mentioned `ext` stands for `extension`. Maybe `vm-memory-wrapper` might be better or is there a different suffix you would suggest?

dianpopa:
> > So with this change the local `vm-memory` crate is renamed to `vm-memory-ext`? What does `ext` stand for? To me it sounds like external which is the opposite to what we want to call it.
> 
> As @ShadowCurse mentioned `ext` stands for `extension`. Maybe `vm-memory-wrapper` might be better or is there a different suffix you would suggest?

Sounds better, yes.

JonathanWoollett-Light:
> > > So with this change the local `vm-memory` crate is renamed to `vm-memory-ext`? What does `ext` stand for? To me it sounds like external which is the opposite to what we want to call it.
> > 
> > 
> > As @ShadowCurse mentioned `ext` stands for `extension`. Maybe `vm-memory-wrapper` might be better or is there a different suffix you would suggest?
> 
> Sounds better, yes.

Just pushed change to use `-wrapper`

nmanthey:
I tested this, and needed to rename 2 more locations in 1 file: 

```
sed -i 's:vm_memory:vm_memory_wrapper:g' src/devices/src/virtio/net/iovec.rs
```

I also did the renaming in a local vm-memory checkout; not sure whether that was required.

JonathanWoollett-Light:
> I tested this, and needed to rename 2 more locations in 1 file:
> 
> ```
> sed -i 's:vm_memory:vm_memory_wrapper:g' src/devices/src/virtio/net/iovec.rs
> ```
> 
> I also did the renaming in a local vm-memory checkout; not sure whether that was required.

Thankyou, fixed.

roypat:
I noticed that vm-memory only contains a lib.rs file, which only contains 3 (short) functions (the rest of the file is test code). Maybe we can eliminate the crate altogether and move these functions to `util`?

dianpopa:
> I noticed that vm-memory only contains a lib.rs file, which only contains 3 (short) functions (the rest of the file is test code). Maybe we can eliminate the crate altogether and move these functions to `util`?

Yup, I think it is a very good idea.

JonathanWoollett-Light:
Closed for: https://github.com/firecracker-microvm/firecracker/issues/3408

PR #3341, opened 2022-12-29
fix(create-snapshot): workaround tests directory

pb8o:
The test infra assumes CWD is 'ROOT/tests'. create_snapshot_artifact runs from the root '.', so some functions can fail.

Workaround by changing to tests while doing the imports.

Original error:

```
[12] Command:
cd ../src/firecracker && cargo pkgid | cut -d# -f2 | cut -d: -f2
[12] stdout:

[12] stderr:
/bin/sh: 1: cd: can't cd to ../src/firecracker

Returned error code: 2
stderr:
/bin/sh: 1: cd: can't cd to ../src/firecracker
```

Signed-off-by: Pablo Barbáchano <pablob@amazon.com>

## Changes

...

## Reason

...

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tools/create_snapshot_artifact/main.py
+++ b/tools/create_snapshot_artifact/main.py
@@ -15,6 +15,8 @@ from functools import partial
 sys.path.append(os.path.join(os.getcwd(), "tests"))  # noqa: E402
 
 # pylint: disable=wrong-import-position
+# The test infra assumes it is running from the `tests` directory.
+os.chdir("tests")
 from conftest import _test_images_s3_bucket, _gcc_compile, init_microvm
 from framework.artifacts import (
     ArtifactCollection,
@@ -32,6 +34,8 @@ from framework.utils import (
 from framework.utils_cpuid import CpuVendor, get_cpu_vendor
 from integration_tests.functional.test_cmd_line_start import _configure_vm_from_json
 import host_tools.network as net_tools  # pylint: disable=import-error
+# restore directory
+os.chdir("..")
 
 DEST_KERNEL_NAME = "vmlinux.bin"
 ROOTFS_KEY = "ubuntu-18.04"


tools/create_snapshot_artifact/main.py
```diff
@@ -15,6 +15,8 @@
 sys.path.append(os.path.join(os.getcwd(), "tests"))  # noqa: E402
```
roypat, line 15: Shouldn't this line already ensure that we can import relative to the tests directory?

andreitraistaru: I think this just adds a new path for the python interpreter to look for files from which to do the import but it doesn't actually change the current working directory. The method `get_firecracker_version_from_toml()` expects the working directory to be set to /firecracker/tests, it doesn't look to the paths interpreter uses

andreitraistaru (approved):

tools/create_snapshot_artifact/main.py
andreitraistaru: I think this just adds a new path for the python interpreter to look for files from which to do the import but it doesn't actually change the current working directory. The method get_firecracker_version_from_toml() expects the working directory to be set to /firecracker/tests, it doesn't look to the paths interpreter uses

pb8o: Yes, that allows importing, but it does not move into the `tests` directory. A lot of it just assumes we are in `tests`. A better fix would be to extract the functions that are not related to testing into proper Python modules, and/or to fix the test infra to not assume the test directory. But both are pretty large in scope.

roypat: Ah, I see, slightly misunderstood what the issue was (thought it was about the imports failing)

roypat (approved):

PR #3346, opened 2023-01-02
Clean up Pytest integration tests

pb8o:
## Changes

This is part 1 of a series of test fixes, simplifications and improvements with the final goal to make the tests run in parallel. As the fixes are sometimes disconnected, the best way to review would be to go commit by commit, in order.

This PR doesn't fix everything, but it fixes the more salient ones, related to conflicting tap devices or trying to compile/download the same file in parallel.

## Reason

Running the tests in parallel results in many errors. 
## License Acceptance

## PR Checklist

---


tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -535,7 +535,6 @@ def test_create_large_diff_snapshot(test_microvm_with_api):
     vm = test_microvm_with_api
```
pb8o, line 535: hmm I think this is empty because I cherry picked this commit into another series.

tests/framework/with_filelock.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Simple decorator so that only one process is running the decorated function
+at any one time.
+
+Caveat: two functions sharing the same name and using this decorator will use
+the same lock, which may be unintended, but safe.
+
+TBD disambiguate with the module name in that case.
```
bchalios, line 10: Not exactly, there's a `tempdir` included in the path of the lock file, right?

tests/host_tools/cargo_build.py
```diff
@@ -49,35 +51,30 @@ def cargo_test(path, extra_args=""):
...
+    target_dir = FC_WORKSPACE_TARGET_DIR
+    out_dir = Path(f"{target_dir}/{target}/release")
+    fc_bin_path = out_dir / FC_BINARY_NAME
+    jailer_bin_path = out_dir / JAILER_BINARY_NAME
+
+    if not fc_bin_path.exists():
+        cd_cmd = "cd {}".format(FC_WORKSPACE_DIR)
+        flags = 'RUSTFLAGS="{}"'.format(get_rustflags())
+        cargo_default_cmd = f"cargo build --release --target {target}"
+        cargo_jailer_cmd = f"cargo build -p jailer --release --target {target}"
```
bchalios, line 71: Why can't we use here the `cargo_build` function from above?

bchalios (commented):
There are quite a few commits here, so I will be reviewing this by commit

bchalios (commented):
So, this one does not actually changes the names of the devices being used. It just uses the `NetIfaceConfig` object fields instead of the `DEFAULT_*_NAME` counterparts, even though the `NetIfaceConfig` objects actually use the defaults.

I guess this will be useful later on (?) If that is the case, could you add that information in the commit message?

bchalios:
> So, this one does not actually changes the names of the devices being used. It just uses the `NetIfaceConfig` object fields instead of the `DEFAULT_*_NAME` counterparts, even though the `NetIfaceConfig` objects actually use the defaults.
> 
> I guess this will be useful later on (?) If that is the case, could you add that information in the commit message?

Ignore this, it is posted in the wrong place and I can't delete it :(.

tests/integration_tests/functional/test_api.py
```diff
@@ -1236,7 +1234,7 @@ def test_get_full_config_after_restoring_snapshot(bin_cloner_path):
     setup_cfg["metrics"] = None
     setup_cfg["mmds-config"] = {
         "version": "V1",
-        "network_interfaces": [DEFAULT_DEV_NAME],
+        "network_interfaces": [net_iface.dev_name],
```
bchalios, line 1237: So, this one does not actually changes the names of the devices being used. It just uses the `NetIfaceConfig` object fields instead of the `DEFAULT_*_NAME` counterparts, even though the `NetIfaceConfig` objects actually use the defaults.

I guess this will be useful later on (?) If that is the case, could you add that information in the commit message?

bchalios: (here's the correct place)

tests/host_tools/network.py
```diff
@@ -295,6 +297,12 @@ def get_guest_net_if_name(ssh_connection, guest_ip):
     return if_name if if_name != "" else None
 
 
+def random_str(k):
+    """Create a random string of length `k`."""
+    symbols = string.ascii_lowercase + string.digits
+    return "".join(random.choices(symbols, k=k))
+
+
 class Tap:
```
bchalios, line 170: Just for extra measure, could we add a destructor here that will remove the tap device (the one outside the namespace) when the object gets dropped?

tests/integration_tests/performance/test_snapshot_perf.py
```diff
@@ -436,23 +436,39 @@ def _test_snapshot_resume_latency(context):
...
-    jailer = firecracker.jailer()
-    jailer.download()
-    fc_version = firecracker.base_name()[1:]
+
+@pytest.mark.parametrize("io_engine", ENGINES)
+@pytest.mark.parametrize(
+    "microvm", ARTIFACTS.microvms(keyword="2vcpu_512mb"), ids=lambda uvm: uvm.name()
+)
+def test_older_snapshot_resume_latency(
+    bin_cloner_path, results_file_dumper, firecracker_release, io_engine, microvm
```
bchalios, line 447: oh, ok, I got confused at the beginning because at the original test we were gathering two sets of artifacts, one for the microvm configs and one for the firecracker versions, however here it appeared only the former.

I had to dig a bit to figure out that `firecracker_release` is a fixture.

Could you add this info in the commit body? (In general, I think it's good practice to put a small description in the commit body, even though the change is very simple)

tests/integration_tests/functional/test_snapshot_version.py
```diff
@@ -39,38 +39,6 @@ def _create_and_start_microvm_with_net_devices(
         assert exit_code == 0
 
 
-@pytest.mark.skipif(
```
bchalios, line 42: Again, it's a nit, but I think it adds clarity and it is useful if you could mention in the commit body why we removing this (it refers to a Firecracker version we do not support any more)

tests/integration_tests/build/test_binary_size.py
```diff
@@ -33,7 +34,7 @@
 
 
 @pytest.mark.timeout(500)
-def test_firecracker_binary_size():
+def test_firecracker_binary_size(record_property):
```
bchalios, line 37: So `record_property` records properties per test, right? My concern here is "recording" properties with the same key.

tests/framework/s3fetcher.py
```diff
@@ -143,48 +143,6 @@ def init_vm_resources(self, microvm_image_name, microvm):
                 microvm.ssh_config["ssh_key_path"] = microvm_dest_path
                 os.chmod(microvm_dest_path, 0o400)
 
-    def hardlink_vm_resources(self, microvm_image_name, from_microvm, to_microvm):
```
bchalios, line 146: Is this related to the change of this commit?

pb8o: It is not used by any test anymore.

tests/conftest.py
```diff
@@ -234,6 +235,32 @@ def pytest_collection_modifyitems(config, items):
                 item.add_marker(skip_marker)
 
 
+def pytest_runtest_makereport(item, call):
+    """Decorate test results with additional properties."""
+    if call.when == "setup":
```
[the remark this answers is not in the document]
pb8o: Big fan of that as well. I think I copy-pasted the skeleton from pytest.org. Will fix.

tests/integration_tests/functional/test_api.py
pb8o: Hmm I don't remember but I thought this fixed something. Or maybe the issue got fixed with a later fix.

Will add to the commit body.

tests/host_tools/network.py
pb8o: But when it's moved it disappears from the default namespace no? Actually the docs of the class say it will get removed when it goes out of scope but I don't see anything related to that.

tests/integration_tests/performance/test_snapshot_perf.py
pb8o: Yep, I agree with always having some commit body. Sometimes I am too lazy :D

pb8o: Done

tests/integration_tests/functional/test_snapshot_version.py
pb8o: Done!

tests/host_tools/ip_generator.py
```diff
@@ -0,0 +1,120 @@
...
+        self.network = ip_network(network_str)
+        netmask_len_diff = subnet_netmask_len - self.network.prefixlen
+        self._subnets = self.network.subnets(netmask_len_diff)
+        self._returned_subnets = []
+
+    def borrow_subnet(self):
+        """Borrow a subnet from the pool"""
+        if len(self._returned_subnets) > 0:
+            return self._returned_subnets.pop(0)
+        return next(self._subnets)
```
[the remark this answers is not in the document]
pb8o: A `StopIteration` will be raised I think, and the test will error out. In the old one, a `NoMoreIPsError` would be raised, with the same effect.

tests/conftest.py
```diff
@@ -3,6 +3,8 @@
```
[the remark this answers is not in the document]
pb8o: done

tests/conftest.py
```diff
@@ -3,6 +3,8 @@
 
 # Pytest fixtures and redefined-outer-name don't mix well. Disable it.
 # pylint:disable=redefined-outer-name
+# We import some fixtures that are unused. Disable that too.
+# pylint:disable=unused-import
```
JonathanWoollett-Light, line 7: Why do we import unused fixtures?

tests/framework/microvm.py
```diff
@@ -782,10 +782,12 @@ def ssh_network_config(
         """
         # Create tap before configuring interface.
         tapname = tapname or (self.id[:8] + "tap" + iface_id)
-        (host_ip, guest_ip) = network_config.get_next_available_ips(2)
-        tap = self.create_tap_and_ssh_config(
-            host_ip, guest_ip, network_config.get_netmask_len(), tapname
-        )
+        # The guest is hardcoded to expect an IP in a /30 network,
+        # so we request the IPs to be aligned on a /30 network
+        # See file:../../resources/tests/fcnet-setup.sh#L21
```
JonathanWoollett-Light, line 787: Nit: Maybe worth linking to the line https://github.com/firecracker-microvm/firecracker/blob/main/resources/tests/fcnet-setup.sh#L21?

tests/host_tools/ip_generator.py
```diff
@@ -0,0 +1,120 @@
...
+    Yield a SubnetGenerator per pytest worker
+
+    We use the 16-bit block 192.168.0.0/16 as it's (empirically) the least
+    likely to conflict with a cloud provider private IPs.
+    https://en.wikipedia.org/wiki/Private_network
+    """
+    # Example worker_id = gw4
+    worker_num = 0
+    if worker_id != "master":
+        worker_num = int(worker_id[2:])
```
JonathanWoollett-Light, line 92: Nit
```suggestion
    if worker_id == "master":
        worker_num = 0
    else:
        worker_num = int(worker_id[2:])
```

tests/integration_tests/build/test_dependencies.py
```diff
@@ -4,8 +4,16 @@
...
+
 import pytest
+
 from framework import utils
+from host_tools import proc
+
+
+pytestmark = pytest.mark.skipif(
+    "Intel" not in proc.proc_type(), reason="test only runs on Intel"
+)
```
JonathanWoollett-Light, line 16: It may be better to use [a marker](https://docs.pytest.org/en/7.1.x/how-to/mark.html) for tests which are architecture independent like `--nonci`.
```suggestion
pytestmark = [pytest.mark.architecture_independent]
```
I would also prefer to apply it directly to tests as I think this is clearer than `pytestmark ` e.g.
```python
@pytest.mark.architecture_independent
def test_licenses():
```

tests/conftest.py
```diff
@@ -234,6 +235,32 @@ def pytest_collection_modifyitems(config, items):
...
+        description = []
+        attributes = {}
+        for line in function_docstring.split("\n"):
+            # extract tags like @type, @issue, etc
+            match = re.match(r"\s*@(?P<attr>\w+):\s*(?P<value>\w+)", line)
+            if match is not None:
+                attr, value = match["attr"], match["value"]
+                attributes[attr] = value
+            else:
+                description.append(line)
```
JonathanWoollett-Light, line 258: Nit: Can flip cases to avoid not.
```suggestion
            if match is None:
                description.append(line)
            else:
                attr, value = match["attr"], match["value"]
                attributes[attr] = value
```

JonathanWoollett-Light (approved):
Looks good.

JonathanWoollett-Light:
Nit: In conventional commits, `feat` is used to represent a major version change.

https://www.conventionalcommits.org/en/v1.0.0/:

>2. feat: a commit of the type feat introduces a new feature to the codebase (this correlates with [MINOR](http://semver.org/#summary) in Semantic Versioning).

`test` is an alternative tag that can be used here.

tests/integration_tests/build/test_dependencies.py
bchalios: I like it, but would that stop them for running in all but one architectures though?

tests/host_tools/network.py
bchalios: Does it? I'm not saying it doesn't, just lack the knowledge. If it does, that's fine by me.

pb8o: I guess the netns gets destroyed when the test ends, and anything dangling will get removed after Docker exits.

tests/conftest.py
pb8o: `conftest.py` is where you can share fixtures with the rest of the tests. That way you don't have to import them manually in each test.

tests/integration_tests/build/test_dependencies.py
pb8o: I thought the meaning was that this test only runs in x86 because the list of dependencies in ARM may be different? 

If it can run in both, then we shouldn't skip it. Skipping this is not worth the complexity, IMO.

tests/host_tools/ip_generator.py
pb8o: I always found that a bit too surprising in Python.

How about this?

```python
    worker_num = 0 if worker_id == "master" else int(worker_id[2:])
```

JonathanWoollett-Light: This is good.

tests/integration_tests/build/test_dependencies.py
bchalios: I _think_ the meaning of this is that it doesn't matter where does it run **and** it only needs to run once.

JonathanWoollett-Light: I agree with @pb8o that these tests don't really warrant the complexity required to skip them given how quick they are.

bchalios: well, as it is there's no complexity. It works already :P

tests/host_tools/cargo_build.py
pb8o: I didn't try. Looking at it it's not immediately obvious as `cargo_build` uses `CARGO_TARGET_DIR` so that could have some side effects. I agree it would be good to clean that up, but I consider it out of scope for this change.

tests/conftest.py
pb8o: Can just do `if match` as `None` is falsy

tests/framework/microvm.py
pb8o: Done

bchalios (approved):

JonathanWoollett-Light (approved):

PR #3353, opened 2023-01-06
Update code coverage to grcov

JonathanWoollett-Light:
## Changes

Changes code coverage to [grcov](https://github.com/mozilla/grcov).

## Reason

kcov outputs inaccurate coverage.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -2,15 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests enforcing code coverage for production code."""
 
-
-import os
-import platform
-import re
-import shutil
 import pytest
 
 from framework import utils
-import host_tools.cargo_build as host  # pylint: disable=import-error
 from host_tools import proc
 
 # We have different coverages based on the host kernel version. This is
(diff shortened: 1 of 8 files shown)


tests/integration_tests/build/test_coverage.py
```diff
@@ -29,109 +23,92 @@
...
+    assert len(processor_model) == 1, "Could not get processor model!"
+    coverage_target = COVERAGE_DICT[processor_model[0]]
+
+    # Re-direct to repository root.
+    monkeypatch.chdir("..")
+
+    # Generate test profiles.
+    utils.run_cmd(
+        f'\
+        env RUSTFLAGS="-Cinstrument-coverage" \
```
roypat, line 66: What does `env` do here? We didn't have it before

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,109 +23,92 @@
...
-
-    coverage_high_msg = (
-        "Current code coverage ({:.2f}%) is >{:.2f}% above the target ({}%).\n"
-        "Please update the value of {}.".format(
-            coverage, COVERAGE_MAX_DELTA, coverage_target_pct, cov_target_name
-        )
+    # Generate coverage report.
+    utils.run_cmd(
+        f'\
+        cargo install --version {GRCOV_VERSION} grcov \
```
roypat, line 75: Leftover from when grcov wasnt in the devctr?

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,109 +23,92 @@
...
+        && grcov . \
+            -s . \
+            --binary-path ./build/cargo_target/{TARGET}/debug/ \
+            --excl-start "mod tests" \
+            --ignore "build/*" \
```
roypat, line 80: Does this exclude list match what we had before? The old list is 
```
      "${CARGO_HOME:-$HOME/.cargo/},"
        "build/,"
        "tests/,"
        "usr/lib/gcc,"
        "lib/x86_64-linux-gnu/,"
        "test_utils.rs,"
        # The following files/directories are auto-generated
        "bootparam.rs,"
        "elf.rs,"
        "mpspec.rs,"
        "msr_index.rs,"
        "bindings.rs,"
        "_gen"
```

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,109 +23,92 @@
...
-        "Current code coverage ({:.2f}%) is >{:.2f}% above the target ({}%).\n"
-        "Please update the value of {}.".format(
-            coverage, COVERAGE_MAX_DELTA, coverage_target_pct, cov_target_name
-        )
+    # Generate coverage report.
+    utils.run_cmd(
+        f'\
+        cargo install --version {GRCOV_VERSION} grcov \
+        && grcov . \
+            -s . \
```
roypat, line 77:
```suggestion
            -s firecracker/src \
```

this automatically excludes the build directory and dependencies and what not

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,109 +23,92 @@
...
+            --binary-path ./build/cargo_target/{TARGET}/debug/ \
+            --excl-start "mod tests" \
+            --ignore "build/*" \
+            -t html \
```
roypat, line 81: Maybe `-t covdir` works better here? It produces json output, so we wouldn't have to parse any HTML (the top level json object contains the total coverage percentage, so it should be easy to extract)

roypat: it also prints to stdout, so no need to mess with files :)

tools/devctr/Dockerfile.aarch64
```diff
@@ -105,10 +105,8 @@ RUN cd "$TMP_POETRY_DIR" \
 RUN mkdir "$TMP_BUILD_DIR" \
     && curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --default-toolchain "$RUST_TOOLCHAIN" \
         && rustup target add aarch64-unknown-linux-musl \
-        && rustup component add clippy \
+        && rustup component add clippy llvm-tools-preview \
         && cd "$TMP_BUILD_DIR" \
```
roypat, line 109: I don't think we need `$TMP_BUILD_DIR` at all in this `RUN` directive anymore, since we're no longer building anything from source here

tools/devctr/Dockerfile.x86_64
```diff
@@ -110,14 +110,12 @@ RUN (curl -sL https://deb.nodesource.com/setup_14.x | bash) \
 RUN mkdir "$TMP_BUILD_DIR" \
     && curl https://sh.rustup.rs -sSf | sh -s -- -y --profile minimal --default-toolchain "$RUST_TOOLCHAIN" \
         && rustup target add x86_64-unknown-linux-musl \
-        && rustup component add rustfmt clippy clippy-preview \
+        && rustup component add rustfmt clippy clippy-preview llvm-tools-preview \
         && rustup install --profile minimal "stable" \
         && cd "$TMP_BUILD_DIR" \
```
roypat, line 115: see above

JonathanWoollett-Light: `& cargo +"stable" install cargo-audit \` is building`cargo-audit` from src, although I wouldn't expect the directory is needed here.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: Ignoring `build/*` relates more to fixing awkward behavior resulting from our non-default build directory. Otherwise `grcov` only evaluates coverage within `src`.

We shouldn't ignore auto-generated rust files. I would argue this was a mistake originally.

tools/devctr/Dockerfile.aarch64
JonathanWoollett-Light: Removed this and more to clean up the Rust install.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: We would then need to parse the json, then read the coverage. Reading the coverage here is more difficult and I'm not sure how to do it accurately with this format in our use case.

tests/integration_tests/build/test_coverage.py
roypat: Oh I see. I don't have strong opinions on the auto generated files thing, but bindgen doesn't use `#[cfg(test)]` to gatekeep its tests, so the reports contain a lot of autogenerated tests at the moment. Maybe including the generated files in the coverage report would be a discussion for a separate PR?

Independently of that, the `test_utils.rs` files and `tests/` modules used to be excluded from reports, and are not anymore.

tests/integration_tests/build/test_coverage.py
roypat: I think it would just be `stdout, _, _ = utils.cmd(...); json.loads(stdout)['total_coverage']`, but I've actually changed my mind about this as its kinda nice to have access to the html coverage report in the build directory after running tests.

tests/integration_tests/build/test_coverage.py
roypat: (just for my own understanding, not requesting this be changed): Do we need it, though? Doesn't just doing `RUSTFLAGS='-Cinstrument-coverage cargo test ...` do the same?

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: I agree `tests/` should be ignored, I will add this.

I would argue the `test_utils.rs` files should be included. The `test_utils.rs` not `cfg(test)` so we can only consider them to be production code.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: You're right it is unneeded here, I've removed it.

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,109 +23,92 @@
...
+            --branch \
```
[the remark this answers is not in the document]
JonathanWoollett-Light: Removed it.

tests/integration_tests/build/test_coverage.py
roypat: `test_utils` are not `#[cfg(test)]` because they are meant to be usable from non-unittests (e.g. when cargo compiled integration tests, it does not pass `--cfg test` to the library under test). It would be a logic error if any of their code was actually executed in a production context. 

Let's get the second reviewers opinion on this

JonathanWoollett-Light: This I would argue is a design flaw, `test_utils.rs` should not exist as they presently do. But I do not think we can ignore coverage on code which is compiled when building a release binary.

roypat: I agree that long term these `test_utils.rs` files should probablity live in their own crate which is not linked into a release binary (although currently their code is removed by link time dead code elimination, as they are not used outside of testing code).

JonathanWoollett-Light: > (although currently their code is removed by link time dead code elimination, as they are not used outside of testing code).

I'm not sure this is guaranteed.

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,109 +23,86 @@
...
+    #
+    # The line looks like `<abbr title="44724 / 49237">90.83 %</abbr></p>` and is the first
+    # occurrence of the `<abbr>` element in the file.
+    #
+    # When we update grcov to 0.8.* we can update this to pull the coverage from a generated .json
+    # file.
+    index = open(
+        f"./build/cargo_target/{TARGET}/debug/coverage/index.html", encoding="utf-8"
     )
-    assert coverage <= coverage_target_pct + COVERAGE_MAX_DELTA, coverage_high_msg
```
pb8o, line 137: Can you keep the `record_property`? This is to report test results as part of a release.

tests/integration_tests/build/test_coverage.py
```diff
@@ -29,109 +23,86 @@
...
+    index_contents = index.read()
+    end = index_contents.find(" %</abbr></p>")
+    start = index_contents[:end].rfind(">")
+    coverage_str = index_contents[start + 1 : end]
+    coverage = float(coverage_str)
+
+    # Compare coverage.
+    high = coverage_target * (1.0 + COVERAGE_MAX_DELTA)
+    low = coverage_target * (1.0 - COVERAGE_MAX_DELTA)
```
pb8o, line 100: In the old one the delta was 0.05% so it should be `1+- COVERAGE_MAX_DELTA/100`

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: Added it back in.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: I think its simpler to change `COVERAGE_MAX_DELTA` to `0.0005`.
```rust
# We allow coverage to have a max difference of `COVERAGE_MAX_DELTA` as percentage before failing
# the test (currently 0.05%).
COVERAGE_MAX_DELTA = 0.0005
```

tests/integration_tests/build/test_coverage.py
dianpopa: > This I would argue is a design flaw, the `test_utils.rs` files should not exist as they presently do. But I do not think we can ignore coverage on code which is compiled when building a release binary.

I would argue to not include changes in this PR that are not in parity with what used to be before (I.e let's target only to update coverage tool).
If we feel there are other changes that need to be made, we want to put them in a separate commit/PR and properly motivate them.

tests/integration_tests/build/test_coverage.py
```diff
@@ -23,115 +17,98 @@
...
 
+    # Record coverage.
     record_property(
-        "coverage", f"{coverage}% {coverage_target_pct}% ±{COVERAGE_MAX_DELTA:.2}%"
+        "coverage", f"{coverage}% {coverage_target}% ±{COVERAGE_MAX_DELTA:.2}%"
```
pb8o, line 104: This will now print 0.0005% instead of 0.05% You can fix by moving the % inside and Python will convert it to a percent: `±{COVERAGE_MAX_DELTA:.2%}`

tools/devctr/Dockerfile.x86_64
```diff
@@ -106,24 +109,12 @@ RUN (curl -sL https://deb.nodesource.com/setup_14.x | bash) \
...
-        && rm -rf "$CARGO_HOME/git" \
-        && ln -s "$CARGO_GIT_REGISTRY_DIR" "$CARGO_HOME/git" \
-    && cd / \
-    && rm -rf "$TMP_BUILD_DIR"
+RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain "$RUST_TOOLCHAIN" \
+    && rustup target add x86_64-unknown-linux-musl \
+    && rustup component add llvm-tools-preview \
+    && cargo install cargo-audit \
+    && cargo install --locked cargo-deny \
+    && cargo install --version $GRCOV_VERSION grcov
```
pb8o, line 117: This includes two other unrelated changes:

- Switch from profile minimal to default.
- Drop the stable channel for cargo-audit.

I'm fine with those, but I think they deserve to be mentioned in the commit message, or even as separate commits.

JonathanWoollett-Light: I agree it makes sense for them to be separate commits I will split them out.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: Ah yes, fixed.

tests/integration_tests/build/test_coverage.py
JonathanWoollett-Light: Added `test_utils.rs` to ignore list.

pb8o (dismissed):

tests/integration_tests/build/test_coverage.py
```diff
@@ -23,115 +17,97 @@
...
+        grcov . \
+            -s . \
+            --binary-path ./build/cargo_target/{TARGET}/debug/ \
+            --excl-start \"mod tests\" \
+            --ignore \"['build/*', 'tests/*', 'test_utils.rs']\" \
+            -t html \
+            --ignore-not-existing \
+            -o ./build/cargo_target/{TARGET}/debug/coverage \
+    "
+    )
```
roypat, line 80: You have to pass multiple `--ignore` flags, grcov does not accept this syntax. I've verified that the below invocation correctly excludes all unwanted files.

```suggestion
utils.run_cmd(
        f"""
        grcov . \
            -s . \
            --binary-path ./build/cargo_target/{TARGET}/debug/ \
            --excl-start \"mod tests\" \
            --ignore "build/*" \
            --ignore "**/tests/*" \
            --ignore "**/test_utils.rs" \
            -t html \
            --ignore-not-existing \
            -o ./build/cargo_target/{TARGET}/debug/coverage"""
    )
```

JonathanWoollett-Light: Fixed.

roypat (dismissed):

roypat (approved):

PR #3370, opened 2023-01-16
fix: Rate limiter `TokenBucket::auto_replenish()`

JonathanWoollett-Light:

## Changes

Fixes bug in `TokenBucket::auto_replenish`. This change also yields a slight performance increase (from ~28ns to ~26ns).

## Reason

When a `TokenBucket` gets empty, the implementation sets a timer of 100ms for replenishing it.

The bucket replenishing function calculates the new tokens to add to the budget using the following
formula:

```
let tokens = (time_delta * self.processed_capacity) / self.processed_refill_time;
```

However, this formula can return `0` depending of the values of `self.processed_capacity` and `self.processed_refill_time`.
For a `TokenBucket` of `size` total capacity and `complete_refill_time_ns` refill period in nanoseconds, the above values are calculated as follows:

```
// Get the greatest common factor between `size` and `complete_refill_time_ns`.
let common_factor = gcd(size, complete_refill_time_ns);
let processed_capacity: u64 = size / common_factor;
let processed_refill_time: u64 = complete_refill_time_ns / common_factor;
```

So, for example if `size == 1` and `complete_refill_time_ns == 1_000_000_000` (equivalent to 1 token per second) the replenishing tokens will be

```
let tokens = (100_000_000 * 1) / 1_000_000_000; // which gives 0 in integer division
```

As a result, the bucket will never get replenished.

## License Acceptance

## PR Checklist

---


acatangiu (dismissed):
since all internals are `u64` (and 2^64 nanoseconds is > 580 years) the `gcd` is definitely not needed and this simplification makes total sense :+1:

src/rate_limiter/src/lib.rs
```diff
@@ -101,6 +87,7 @@ pub struct TokenBucket {
     initial_one_time_burst: u64,
     // Complete refill time in milliseconds.
     refill_time: u64,
+    refill_one_ns: u64,
```
pb8o, line 90: LGTM. Can you move the PR explanation into the commit body?

pb8o (approved):

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: I've moved the explanation to the commit body.

src/rate_limiter/src/lib.rs
```diff
@@ -130,45 +113,30 @@ impl TokenBucket {
...
-        // The division will be exact since `common_factor` is a factor of
-        // `complete_refill_time_ns`.
-        let processed_refill_time: u64 = complete_refill_time_ns / common_factor;
 
         Some(TokenBucket {
             size,
             one_time_burst,
             initial_one_time_burst: one_time_burst,
             refill_time: complete_refill_time_ms,
+            refill_one_ns: (1_000_000 * complete_refill_time_ms) / size,
```
roypat, line 122: Technically, this just moves the integer division problem from "bucket with very small size" to "bucket with very large size". If `size > 1_000_000 * complete_refill_time_ms`, then `refill_one_ns` will be zero and in `auto_replenish` we perform a division by zero (so instead of hanging we crash).

bchalios: Good catch!!
I think the solution here is to somehow adjust the 100 msec period over which we (gradually) replenish the bucket, e.g. Find the minimum period for which the new tokens would be > 0

src/rate_limiter/src/lib.rs
```diff
@@ -569,30 +529,13 @@ pub(crate) mod tests {
...
-
-        let thousand = 1000;
-        let tb = TokenBucket::new(3 * 7 * 11 * 19 * thousand, 0, 7 * 11 * 13 * 17).unwrap();
-        assert_eq!(tb.get_processed_capacity(), 3 * 19);
-        assert_eq!(
-            tb.get_processed_refill_time(),
-            13 * 17 * (NANOSEC_IN_ONE_MILLISEC / thousand)
-        );
-    }
-
```
bchalios, line 595: It would be nice to add unit tests that make sure we don't have the issue any more (or the issue that @roypat) mentions in his review.

bchalios (changes requested):

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: As @acatangiu noted to overflow here requires `complete_refill_time_ms >= 1.8446744e+13` which is `~213503` days or `~584` years. I think crashing in this case is not unreasonable.

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: Could you suggest an what you are looking for here?

bchalios: Something along the lines:

> Our implementation of `TokenBucket` can dead-lock because of...
>
> Change the implementation in such and such way to solve this.

src/rate_limiter/src/lib.rs
bchalios:
```
// A TokenBucket with 1GB/sec allowance
TokenBucket::new(2_000_000_000, 0, 1000)
```

would yield

```
refill_one_ns: (1_000_000 * 1000) / 2_000_000_000
```

which is 0. Am I missing something?

roypat: > I think the solution here is to somehow adjust the 100 msec period over which we (gradually) replenish the bucket, e.g. Find the minimum period for which the new tokens would be > 0

Yes, that sounds right. I _think_ if we go back to the old implementation, but only execute `last_update = Instant::now()` if more than 0 tokens were replenished would have the effect you propose.

JonathanWoollett-Light: When `complete_refill_time_ms = 1000` and `size = 2_000_000_000` we are restoring 1 token every 0.5ns. Do any I/O operations in Firecracker can occur near this frequency?

I don't think most rate limiters consider any case near this frequency, for example https://docs.rs/governor/0.5.1/governor/struct.Quota.html#method.per_second at max allows replacing 4 tokens every ns.

JonathanWoollett-Light: > Yes, that sounds right. I think if we go back to the old implementation, but only execute last_update = Instant::now() if more than 0 tokens were replenished would have the effect you propose.

I think this is the best solution here, I've changed it to this.

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: I've update the commit message to match the new implementation.

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: This solution still drops tokens, e.g.:
```rust
#[test]
fn test_token_bucket_auto_replenish() {
    const SIZE: u64 = 1000;
    const TIME: u64 = 1000;
    let time = Duration::from_millis(TIME);

    let mut tb = TokenBucket::new(SIZE, 0, TIME).unwrap();
    tb.reduce(SIZE);
    assert_eq!(tb.budget(), 0);

    let now = Instant::now();
    while now.elapsed() < time {
        tb.auto_replenish();
    }
    assert_eq!(tb.budget(), SIZE - 1);
}
```
But I think this is reasonable.

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: I've added a test: https://github.com/firecracker-microvm/firecracker/pull/3370#discussion_r1072184741

src/rate_limiter/src/lib.rs
```diff
@@ -161,13 +161,21 @@ impl TokenBucket {
...
+
+        // Frequently calling `auto_replenish` will reset `self.last_update` each
+        // time and `tokens` may be a fractional value (0 since it is a `u64`), in
+        // this case no tokens will replenished. To avoid this we only update
+        // `self.last_update` when `tokens > 0`.
+        if tokens > 0 {
+            self.last_update = Instant::now();
+        }
+
         self.budget = std::cmp::min(self.budget + tokens, self.size);
```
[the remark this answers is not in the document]
JonathanWoollett-Light: Fixed.

JonathanWoollett-Light:
@roypat @bchalios Could you please re-review.

src/rate_limiter/src/lib.rs
```diff
@@ -161,14 +161,21 @@ impl TokenBucket {
...
         // At each 'time_delta' nanoseconds the bucket should refill with:
         // refill_amount = (time_delta * size) / (complete_refill_time_ms * 1_000_000)
         // `processed_capacity` and `processed_refill_time` are the result of simplifying above
         // fraction formula with their greatest-common-factor.
         let tokens = (time_delta * self.processed_capacity) / self.processed_refill_time;
-        self.budget = std::cmp::min(self.budget + tokens, self.size);
+
+        // Frequently calling `auto_replenish` will reset `self.last_update` each
+        // time and `tokens` may be a fractional value (0 since it is a `u64`), in
+        // this case no tokens will replenished. To avoid this we only update
```
bchalios, line 173: nit
```suggestion
        // this case no tokens will be replenished. To avoid this we only update
```

src/rate_limiter/src/lib.rs
```diff
@@ -161,14 +161,21 @@ impl TokenBucket {
...
+        // `self.last_update` when `tokens > 0`.
+        if tokens > 0 {
```
bchalios, line 175: This seems right to me.

I had a thought.

What do you think about making `TIMER_REFILL_STATE` a variable instead of a `const`? As it is now, this is a hard-coded value of 100ms. If a time delta of 100 msec yields a `false` in this if statement we will have the emulation thread to wake up every 100 msec without reason.

If, instead, we had a TIMER_REFILL_STATE to be the maximum between 100 msec and the minimum time delta for replenishing at least one token, we would avoid this (small) overhead and reduce CPU utilization in these cases.

I understand this is optimization for a corner case, but maybe it's worth it (?) Cc @roypat @acatangiu

src/rate_limiter/src/lib.rs
```diff
@@ -560,6 +567,23 @@ pub(crate) mod tests {
         }
     }
 
+    #[test]
+    fn test_token_bucket_auto_replenish() {
+        const SIZE: u64 = 1000;
+        const TIME: u64 = 1000;
```
bchalios, line 573: With these values we get 1 new token every millisecond.

with:

```suggestion
        const SIZE: u64 = 1;
        const TIME: u64 = 1000;
```

we should have 1 new token every second, so it's easier to trigger the corner-case condition.

bchalios (changes requested):

src/rate_limiter/src/lib.rs
bchalios: It doesn't drop tokens. I think that if you had put a `tb.auto_replenish()` just after the while loop you would reach `SIZE`. However, see comment below.

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: I understand your point, but `autoreplenish()` takes `~45ns` so a `1ms` step is distant enough (`x22222`) from this while still a reasonable use case.

src/rate_limiter/src/lib.rs
```diff
@@ -560,6 +567,23 @@ pub(crate) mod tests {
...
+
+        let mut tb = TokenBucket::new(SIZE, 0, TIME).unwrap();
+        tb.reduce(SIZE);
+        assert_eq!(tb.budget(), 0);
+
+        let now = Instant::now();
+        while now.elapsed() < time {
+            tb.auto_replenish();
+        }
+        assert_eq!(tb.budget(), SIZE - 1);
```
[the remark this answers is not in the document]
JonathanWoollett-Light: This test should be more stable I agree, I've used this instead.

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: I think this is resolved by using your suggestion in https://github.com/firecracker-microvm/firecracker/pull/3370#discussion_r1072304501.

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: Fixed.

src/rate_limiter/src/lib.rs
bchalios: When I changed the value to the suggested ones the test was actually failing:

```
running 1 test
test tests::test_token_bucket_auto_replenish ... FAILED

failures:

---- tests::test_token_bucket_auto_replenish stdout ----
thread 'tests::test_token_bucket_auto_replenish' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', src/rate_limiter/src/lib.rs:584:9
```

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: I did test this, it stills drops tokens

bchalios: weird, with the new test you can't see that, but it's probably some weird timing issue.

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: This relates to https://github.com/firecracker-microvm/firecracker/pull/3370#discussion_r1072332075

src/rate_limiter/src/lib.rs
```diff
@@ -161,14 +161,21 @@ impl TokenBucket {
...
         // fraction formula with their greatest-common-factor.
         let tokens = (time_delta * self.processed_capacity) / self.processed_refill_time;
-        self.budget = std::cmp::min(self.budget + tokens, self.size);
+
+        // Frequently calling `auto_replenish` will reset `self.last_update` each
+        // time and `tokens` may be a fractional value (0 since it is a `u64`), in
+        // this case no tokens will be replenished. To avoid this we only update
+        // `self.last_update` when `tokens > 0`.
+        if tokens > 0 {
+            self.last_update = Instant::now();
```
roypat, line 176:
```suggestion
            self.last_update += Duration::from_nanos((tokens * self.processed_refill_time) / self.processed_capacity);
```

src/rate_limiter/src/lib.rs
bchalios: Ok that makes total sense. I don't think we have a way out of this.

roypat (changes requested):
As @JonathanWoollett-Light noted, the implementation still "leaks" tokens (but doesnt hang anymore). This is because if `auto_replenish` is called after enough time for, say, 1.9 tokens has passed, one token will be added, and `last_update` will be fast forwarded. The 0.9 tokens will effectively be discarded. 

We can fix this by incrementing `last_update` by the actual time it took to generate the tokens that have been added to the budget, instead of setting it to `Instant::now()`. I think my proposed suggestion implements this correctly, but I'd love to have someone else trace out the logic to verify it. 

Technically, this does not even need the "if tokens > 0" conditional anymore.

src/rate_limiter/src/lib.rs
bchalios: can't this division fall into the exact same problem as before?

src/rate_limiter/src/lib.rs
roypat: Yeah, this makes sense to me. Personally, I am not entirely sure why we need the time in the first place, though, so maybe we can get rid of it altogether?

src/rate_limiter/src/lib.rs
bchalios: oh, I see now. it doesn't matter. if this division yields zero it means we have replenished zero tokens, right? I think it makes sense.

src/rate_limiter/src/lib.rs
bchalios: Imagine we get a net frame to transmit but we're out of budget. We need to know when we have new budget, so we setup a timer file descriptor to periodically replenish tokens. Without it, the net device would be stuck for ever, unless a new net event arrived.

src/rate_limiter/src/lib.rs
JonathanWoollett-Light: This does fix it yes.

acatangiu: which is equivalent to 
```suggestion
            self.last_update += time_delta;
```
right?

bchalios: I think it doesn't really matter. Both cases will appear if we access the rate limiter with extremely high resolution, i.e. ~nanoseconds. This is not normally the case.

acatangiu: > This does however mean we are incurring a time debt on every call. self.last_update will gradually fall further behind Instant::now().

should not fall behind.. every call new `time_delta` is `last_update.elapsed()` where `elapsed()` will "cover" what you "lost" last call (I use "" because nothing is actually lost, and making use of `time_delta` is actually more accurate than calling `Instant::now()` again).

bchalios: Yes, I think Adrian is right. If every time we increase by `time_delta` we actually have the resolution of `.elapsed()` (we only miss whatever `.elapsed()` misses), which should be tiny and it should solve the problem.

JonathanWoollett-Light: As I noted in https://github.com/firecracker-microvm/firecracker/pull/3370#discussion_r1072359788

> I think something like https://crates.io/crates/governor (more specifically https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm) would help with this but this would be a more significant switch.

acatangiu (approved):

src/rate_limiter/src/lib.rs
bchalios: hmmm, but then, `time_delta` might also include fractions of a token. We would loose that as well. Patrick's calculation discounts the fraction, it takes only the integer part, but it looses some time. In any case:

> I think it doesn't really matter. Both cases will appear if we access the rate limiter with extremely high resolution, > i.e. ~nanoseconds. This is not normally the case.

bchalios (approved):

src/rate_limiter/src/lib.rs
bchalios: > As I noted in [#3370 (comment)](https://github.com/firecracker-microvm/firecracker/pull/3370#discussion_r1072359788)
> 
> > I think something like https://crates.io/crates/governor (more specifically https://en.wikipedia.org/wiki/Generic_cell_rate_algorithm) would help with this but this would be a more significant switch.

I don't know whether this solves our issue or not, we would need to see their implementation. In general, I would expect _any_ rate limiting implementation to not have infinite resolution.

JonathanWoollett-Light: > > This does however mean we are incurring a time debt on every call. self.last_update will gradually fall further behind Instant::now().
> 
> should not fall behind.. every call new `time_delta` is `last_update.elapsed()` where `elapsed()` will "cover" what you "lost" last call

It won't cover all the time loss, since time is lost between taking the time measurement and calculating the number of tokens. This should only be nanoseconds,

acatangiu: This "rate-limiter" is very dumb anyway, it works in spikes, trying to hit an **average limit over time**. All the available bucket tokens can be instantly expended (spike well over limit), but then it's forcefully cooldowned for 100ms, then another spike, & so on..

And while simple and stupid, it does the job with good normalized results at scale with very little complexity as opposed to other "smarter" mechanisms :laughing:

roypat (approved):
I share @JonathanWoollett-Light's concern regarding `last_updated` falling behind `Instant::now()`, so here’s a proof (I hope I got everything right) that it doesn’t (for simplicity, I use the non-reduced versions of size and refill time):

1. Note that integer division is just normal division, followed by rounding down (floor function)
2. For positive y, we have floor(x / y) * y >= x - y, as the expression on the LHS “rounds down” to the closest multiple of y
3. For integer z, we have `floor(x + z) = floor(x) + z`
4. Multiplication by (-1) flips inequalities, e.g. if `x <= y`, then `z - x >= z - y`.
5. We refer to the system time at the start of `auto_replenish` as `now_old`, and to the system time at the end of `auto_replenish` as `now_new`. With this, we have `time_delta = now_start - last_update_old`
With this in place, note how at the end of `auto_replenish` we set `last_update` to 

`last_update_new = last_update_old + floor(floor((now_start - last_update_old) * size / refill_time) * refill_time / size)`

To prove that `last_update` does not deviate from `now_end = Instant::now()` unboundedly, we want to find an upper bound for `now - last_update_new`. We have 

```
now_end - last_update_new = now_end - last_update_old - floor(floor((now_start - last_update_old) * size / refill_time) * refill_time / size)
<= now_end - last_update_old - floor(((now_start - last_update_old) * size - refill_time) / size)
= now_end - last_update_old - floor(now_start - last_update_old - refill_time / size)
= now_end - now_start - last_update_old + last_update_old + floor(refill_time / size)
= (now_end - now_start) + floor(refill_time / size)
```

This means that `last_updated` indeed does not fall behind more than the execution time of `auto_replenish` plus a constant dependent on the bucket configuration.

src/rate_limiter/src/lib.rs
bchalios: So, what do you think about this?

roypat: Ah, yeah, I suppose a timer is needed somewhere to get rate-limited devices to retry, makes sense now, thanks!

I think making TIMER_REFILL_STATE depending on how long it will actually take to replenish tokens (either one, or maybe even how many the device requested) is a change worth making, but I'd consider it outside the scope of this PR. There is already some logic related to overconsumption that might also need change, so I don't think its a simple +1/-1 commit.

bchalios: ok, agreed!

bchalios (approved):

PR #3373, opened 2023-01-17
Better guest kernel selection

ShadowCurse:
## Changes
Updates kernel selection mechanism in `artifacts.py`

## Reason
Currently guest kernels are selected by searching a sub-string in kernel name.
This can lead to incorrect selection. If we have kernels: [5.10, 5.10-a, 5.10-b] and supported kernel is 5.10 then all 3 kernels will be selected. With this PR only 5.10 kernel will be selected.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tests/framework/artifacts.py
+++ b/tests/framework/artifacts.py
@@ -463,11 +463,11 @@ class ArtifactCollection:
             keyword=keyword,
         )
 
-        valid_kernels = list(
-            filter(
-                lambda kernel: any(s in kernel.key for s in SUPPORTED_KERNELS), kernels
-            )
-        )
+        supported_kernels = {f"vmlinux-{sup}.bin" for sup in SUPPORTED_KERNELS}
+        valid_kernels = [
+            kernel for kernel in kernels if Path(kernel.key).name in supported_kernels
+        ]
+
         return valid_kernels
 
     def disks(self, keyword=None):


tests/framework/artifacts.py
```diff
@@ -463,11 +464,13 @@ def kernels(self, keyword=None):
...
-                lambda kernel: any(s in kernel.key for s in SUPPORTED_KERNELS), kernels
-            )
-        )
+        def check_kernel(kernel):
+            kernel_name = re.search("vmlinux-(.*).bin", kernel.key)
+            if kernel_name:
+                return kernel_name.group(1) in SUPPORTED_KERNELS
+            return False
+
+        valid_kernels = list(filter(check_kernel, kernels))
```
pb8o, line 473: This tranforms the check so that they have to be exact matches. In that case, this is simpler and doesn't require regexes:

```python
supported_kernels = {f"vmlinux-{sup}.bin" for sup in SUPPORTED_KERNELS}
valid_kernels = supported_kernels.intersection(kernels)
```

ShadowCurse: Well, first of all, `kernels` - is a list of objects, not strings. Second `kernel.key` is a path like `ci-artifacts/kernels/aarch64/vmlinux-4.14.bin`. So actually we can extract last item of the path with `os.path.basename(kernel.key)`.
So it would look like:
```python
supported_kernels = {f"vmlinux-{sup}.bin" for sup in SUPPORTED_KERNELS}
valid_kernels = list(filter(lambda kernel: os.path.basename(kernel.key) in supported_kernels, kernels))
```
Looks better?

pb8o: God point. But list comprehensions are more readable/idiomatic in Python than list+filter. Try

```python
valid_kernels = [kernel for kernel in kernels if Path(kernel.key).name in supported_kernels]
```

JonathanWoollett-Light:
Nit: commit can have `test:` conventional commit prefix

tests/framework/artifacts.py
ShadowCurse: I see... snakes in snake language

zulinx86 (approved):

pb8o (approved):

PR #3381, opened 2023-01-23
Graviton3 baselines

ShadowCurse:
## Changes
Adding graviton3 baselines based on c7g.metal instance
Updated baseline tools to support graviton3
Added guest kernel with SVE disabled
Added kernels selection in tests as work around the issue with grav3 with 4.14 host and 5.10 guest

## License Acceptance

## PR Checklist

---


docs/kernel-policy.md
```diff
@@ -41,6 +41,15 @@ The guest kernel configs used in our validation pipelines
 can be found [here](../resources/guest_configs/) while a breakdown
 of the relevant guest kernel modules can be found in the next section.
 
+**Note**
+Be sure to check feature compatability between host, guest kernels
+and the hardware.
+For the combination of 4.14 host and 5.10 guest kernels we are using
+different configuration of the guest kernel with SVE extension disabled.
+This is due to the intoduction of the SVE extension in Graviton3,
```
zulinx86, line 49: nit: typo
```suggestion
This is due to the introduction of the SVE extension in Graviton3,
```

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -20,12 +20,13 @@
...
     get_cpu_percent,
     get_kernel_version,
+    get_instance_type,
     is_io_uring_supported,
     CmdBuilder,
     DictQuery,
     run_cmd,
 )
-from framework.utils_cpuid import get_cpu_model_name, get_instance_type
+from framework.utils_cpuid import get_cpu_model_name
```
zulinx86, line 29: How about moving `get_cpu_model_name()` to `utils` along with `get_instance_type()`? It is also used out of context of CPUID-related tests.

tests/framework/utils_cpuid.py
```diff
@@ -11,7 +11,7 @@
 from framework.utils_imdsv2 import imdsv2_get
 import host_tools.network as net_tools
 
-ARM_CPU_DICT = {"0xd0c": "ARM_NEOVERSE_N1"}
```
pb8o, line 14: To keep the naming scheme, I think this should be `ARM_NEOVERSE_V1` https://en.wikipedia.org/wiki/AWS_Graviton

docs/kernel-policy.md
```diff
@@ -41,6 +41,15 @@ The guest kernel configs used in our validation pipelines
 can be found [here](../resources/guest_configs/) while a breakdown
 of the relevant guest kernel modules can be found in the next section.
 
+**Note**
+Be sure to check feature compatibility between host, guest kernels
+and the hardware.
+For the combination of 4.14 host and 5.10 guest kernels we are using
+different configuration of the guest kernel with SVE extension disabled.
+This is due to the introduction of the SVE extension in Graviton3,
```
pb8o, line 49: technically Neoverse V1

pb8o (changes requested):

tests/framework/utils_cpuid.py
```diff
@@ -11,7 +11,7 @@
...
+ARM_CPU_DICT = {"0xd0c": "ARM_NEOVERSE_N1", "0xd40": "Graviton3"}
```
pb8o, line 13: To keep the naming scheme, I think this should be ARM_NEOVERSE_V1 https://en.wikipedia.org/wiki/AWS_Graviton#Graviton3

pb8o (changes requested):

tests/framework/artifacts.py
```diff
@@ -463,7 +462,7 @@ def kernels(self, keyword=None):
             keyword=keyword,
         )
 
-        supported_kernels = {f"vmlinux-{sup}.bin" for sup in SUPPORTED_KERNELS}
+        supported_kernels = {f"vmlinux-{sup}.bin" for sup in select_supported_kernels()}
```
sudanl0, line 465: This seems to be the only place where you would need the select_supported_kernels() logic so do you think it would be good to move the logic here instead of defining the function in utils.py? or do you see any other use case where you would need select_supported_kernels()?

tests/framework/utils_cpuid.py
ShadowCurse: Ok, didn't know that

ShadowCurse: fixed

tests/framework/artifacts.py
sudanl0: When we drop 4.14 do you plan to keep the artifact.py code as is or will it switch to calling SUPPORTED_KERNELS again?
1. If you plan to keep the select_supported_kernels() and 4.14 host support is dropped then we won't need the 5.10_no_sve and select_supported_kernels() will be simply calling SUPPORTED_KERNELS, right? and in this case select_supported_kernels() becomes a redundant function.
2. if you plan to modify artifacts.py instead to switch from select_supported_kernels() to SUPPORTED_KERNELS when we drop 4.14 host then I think having the function locally seems like a good idea.

There is some refactoring done by moving get_instance_type to utils and this is just to define select_supported_kernels() in utils.py, and its not going to be used by any other code and also going to be redundant when we drop 4.14 so, I am trying to understand the pros/cons of defining this function in utils vs keeping it here locally.

tests/integration_tests/performance/test_block_performance.py
pb8o: Long-term I think we should have some "properties" fixture where we can pull common test environment information like this, rather than each test having to request it on its own. The metrics PRs will bring some of that.

tests/framework/artifacts.py
sudanl0: by moving select_supported_kernels() to artifacts.py you can avoid moving get_instance_type() from utils_cpuid.py to utils.py and also avoid all the other changes to the test files which are not directly related to this PR.

ShadowCurse: Ok, that makes a lot of sense. I'll do that.

sudanl0 (approved):

pb8o (approved):

PR #3382, opened 2023-01-23
fix: CPUID pass-through

JonathanWoollett-Light:
## Changes

Passes portions of the host CPUID through to the guest depending on the CPUID template.

## Reason

In setting CPUID a user will typically want to pass-through most values and only specify a few. These values can relate to cache topology and be very tricky to specify in a template. In this case we need CPUID to support setting specific leaves and passing-through others.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -17,9 +17,9 @@ from host_tools import proc
 # Checkout the cpuid crate. In the future other
 # differences may appear.
 if utils.is_io_uring_supported():
-    COVERAGE_DICT = {"Intel": 78.08, "AMD": 78.92, "ARM": 82.09}
+    COVERAGE_DICT = {"Intel": 78.14, "AMD": 78.92, "ARM": 82.09}
 else:
-    COVERAGE_DICT = {"Intel": 75.81, "AMD": 76.64, "ARM": 79.04}
+    COVERAGE_DICT = {"Intel": 75.87, "AMD": 76.64, "ARM": 79.04}
 
 PROC_MODEL = proc.proc_type()
 
--- a/src/vmm/src/cpuid/amd/mod.rs
+++ b/src/vmm/src/cpuid/amd/mod.rs
@@ -27,6 +27,24 @@ mod registers;
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct AmdCpuid(pub std::collections::BTreeMap<CpuidKey, CpuidEntry>);
 
+impl AmdCpuid {
+    /// Include leaves from `other` that are not present in `self`.
+    #[inline]
+    #[must_use]
+    pub fn include_leaves_from(mut self, other: Self) -> Self {
+        let leaves = self
+            .0
+            .iter()
+            .map(|x| x.0.leaf)
+            .collect::<std::collections::HashSet<_>>();
+
+        self.0
+            .extend(other.0.into_iter().filter(|x| !leaves.contains(&x.0.leaf)));
+
+        self
+    }
+}
+
 impl CpuidTrait for AmdCpuid {
     /// Gets a given sub-leaf.
     #[inline]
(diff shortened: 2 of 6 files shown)


tests/integration_tests/build/test_coverage.py
```diff
@@ -17,9 +17,9 @@
 # Checkout the cpuid crate. In the future other
 # differences may appear.
 if utils.is_io_uring_supported():
-    COVERAGE_DICT = {"Intel": 77.81, "AMD": 78.54, "ARM": 72.26}
+    COVERAGE_DICT = {"Intel": 77.81, "AMD": 78.54, "ARM": 72.17}
 else:
-    COVERAGE_DICT = {"Intel": 75.53, "AMD": 76.26, "ARM": 69.58}
+    COVERAGE_DICT = {"Intel": 75.53, "AMD": 76.26, "ARM": 69.49}
```
kalyazin, line 22: Is that because the newly added functions are not covered with tests? Is there a reason why we can't test them now?

JonathanWoollett-Light: Added additional tests to avoid this.

src/cpuid/src/amd/mod.rs
```diff
@@ -28,6 +28,24 @@ mod registers;
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub struct AmdCpuid(pub std::collections::BTreeMap<CpuidKey, CpuidEntry>);
 
+impl AmdCpuid {
+    /// Include leaves from `other` that are not present in `self`.
+    #[inline]
+    #[must_use]
+    pub fn include_leaves_from(mut self, other: Self) -> Self {
```
kalyazin, line 35: Started wondering, can this go to the CpuidTrait? The implementations look very similar.

JonathanWoollett-Light: `CpuidTrait` doesn't implement any setters and its also implemented for `RawCpuid` and `kvm_bindings::CpuId`.

kalyazin: Well, ```apply_brand_string()``` looks like a setter (updates some leaves).

> its also implemented for RawCpuid and kvm_bindings::CpuId

Not sure I understand, what is implemented?

JonathanWoollett-Light: `apply_brand_string()` uses `get_mut()` to get mutable references to exist leaves. There currently isn't a method in `CpuidTrait` to push new leaves.

kalyazin: ok, fair enough.

kalyazin (approved):

src/cpuid/src/amd/mod.rs
zulinx86, line 35: Don't we have to care about sub-leaves?

JonathanWoollett-Light: It would be better to also be able to pass through subleaves but this would require considerable extra logic to be done safely. For now this is safe and should suffice but it may be worth looking into subleaves moving forward.

zulinx86: OK, I see. Sorry, but I don't understand why it requires considerable extra logic. Is this not enough?
```
        let leaves = self
            .0
            .iter()
            .map(|x| x.0)
            .collect::<std::collections::HashSet<_>>();

        self.0
            .extend(other.0.into_iter().filter(|x| !leaves.contains(&x.0)));
```

JonathanWoollett-Light: The number of sub-leaves within a leaf are denoted using different mechanism. Some may have the length noted in sub-leaf 0, some require getting each sub-leaf until one has a specific value in a specific bit range in a specific register. It is non-trivial to append leaves and it likely requires modification of existing leaves.

zulinx86: It's fair. Thanks for detailed explanation!

zulinx86 (approved):

zulinx86 (approved):

kalyazin (approved):

PR #3394, opened 2023-01-25
fix(tests): test_balloon_snapshot: add delay after inflate

kalyazin:
## Changes

A 1 second delay is added after the inflation request to give the stats time to update.

## Reason

Since the configured stats polling interval is 1 second, it may take up to 1 second for the `available_memory` stat to get updated.
In some cases, the VM is snapshotted, restored and the `available_memory` stat is queried before the 1 second has passed after an inflation had been requested.

## License Acceptance

## PR Checklist

- ~~[ ] New `unsafe` code is documented.~~

---

### DIFF
--- a/tests/integration_tests/functional/test_balloon.py
+++ b/tests/integration_tests/functional/test_balloon.py
@@ -18,6 +18,7 @@ import host_tools.network as net_tools  # pylint: disable=import-error
 
 
 MB_TO_PAGES = 256
+STATS_POLLING_INTERVAL_S = 1
 
 
 @retry(delay=0.5, tries=10)
@@ -480,7 +481,9 @@ def test_stats_update(test_microvm_with_api, network_config):
 
     # Add a memory balloon with stats enabled.
     response = test_microvm.balloon.put(
-        amount_mib=0, deflate_on_oom=True, stats_polling_interval_s=1
+        amount_mib=0,
+        deflate_on_oom=True,
+        stats_polling_interval_s=STATS_POLLING_INTERVAL_S,
     )
     assert test_microvm.api_session.is_status_no_content(response.status_code)
 
@@ -505,7 +508,7 @@ def test_stats_update(test_microvm_with_api, network_config):
     assert test_microvm.api_session.is_status_no_content(response.status_code)
 
     # Wait out the polling interval, then get the updated stats.
-    time.sleep(1)
+    time.sleep(STATS_POLLING_INTERVAL_S)
     next_stats = test_microvm.balloon.get_stats().json()
     assert initial_stats["available_memory"] != next_stats["available_memory"]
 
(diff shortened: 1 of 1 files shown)


tests/integration_tests/functional/test_balloon.py
```diff
@@ -588,6 +588,7 @@ def _test_balloon_snapshot(context):
     # Now inflate the balloon with 20MB of pages.
     response = basevm.balloon.patch(amount_mib=20)
     assert basevm.api_session.is_status_no_content(response.status_code)
+    time.sleep(1)
```
pb8o, line 591: Maybe add a (short) comment here on why we sleep? The explanation in the commit is really good, I'm just not sure I will reach 

```
# sleep 1s to give MemoryMonitor thread time to update
```

or this is ugly but more explicit: `time.sleep(basevm._memory.MEMORY_SAMPLE_TIMEOUT_S)`

pb8o: I just thought about it and this may be even better: make a new method in the `Microvm` class, like so:

```python
def wait_for_memory_update(self):
    if self._memory_monitor:
        time.sleep(self._memory_monitor.MEMORY_SAMPLE_TIMEOUT_S)
```

(or maybe it can delegate to another method in `self._memory_monitor`, maybe that's cleaner but has the same effect).

and then in the balloon test:

```python
    basevm.wait_for_memory_update()
```

tests/integration_tests/functional/test_balloon.py
```diff
@@ -615,7 +615,8 @@ def _test_balloon_snapshot(context):
     # Get the firecracker from snapshot pid, and open an ssh connection.
     firecracker_pid = microvm.jailer_clone_pid
 
-    # Get the stats right after we take a snapshot.
+    # Wait out the polling interval, then get the updated stats.
+    time.sleep(1)
```
pb8o, line 619: Ah I see. Maybe define `STATS_POLLING_INTERVAL_S` at the top of the file and use that everywhere?

pb8o (approved):

tests/integration_tests/functional/test_balloon.py
kalyazin: That was my initial idea. However the test file uses different polling intervals in some cases, and having separate constants for all of them looks a bit too much to me. Moreover, some of the delays are not related to the stats polling.

pb8o: hmm maybe a test-local one? My main concern is we will have to change the `1` in two places vs one, if we ever change it.

kalyazin: Yes, that is very true.

kalyazin: Added ```STATS_POLLING_INTERVAL_S``` and used it where it matters, left others intact.

pb8o (approved):

dianpopa (approved):

PR #3397, opened 2023-01-26
chore: Remove `unsafe` checkpoint from PR template

JonathanWoollett-Light:
## Changes

Removed the ``New `unsafe` code is documented.`` checkpoint from the PR template

## Reason

This is covered by the `clippy::undocumented_unsafe_blocks` lint.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -17,11 +17,11 @@ the terms of the Apache 2.0 license.
 - [ ] If a specific issue led to this PR, this PR closes the issue.
 - [ ] The description of changes is clear and encompassing.
 - [ ] Any required documentation changes (code and docs) are included in this PR.
-- [ ] New `unsafe` code is documented.
 - [ ] API changes follow the [Runbook for Firecracker API changes][2].
 - [ ] User-facing changes are mentioned in `CHANGELOG.md`.
 - [ ] All added/changed functionality is tested.
 - [ ] New `TODO`s link to an issue.
+- [ ] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).
 
 ---
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -13,7 +13,6 @@ the terms of the Apache 2.0 license.
 
 ## PR Checklist
 
-- [ ] All commits in this PR are signed (`git commit -s`).
 - [ ] If a specific issue led to this PR, this PR closes the issue.
 - [ ] The description of changes is clear and encompassing.
 - [ ] Any required documentation changes (code and docs) are included in this PR.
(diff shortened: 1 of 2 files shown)


pb8o (dismissed):

bchalios (dismissed):

roypat (changes requested):
I disagree on removing this one, I think it should be rephrased instead. The clippy lint only checks for the presence of a `SAFETY:` comment, but that only covers the requirement that we argue why invariants are upheld (which is the most important part, I agree). However, to me, this checkbox also includes justifying _why_ `unsafe` was used in the first place. What safe alternatives were considered and why were they discarded? So maybe we can change it to that?

roypat:
The "all commits are signed" one of the other hand can go imo, we check that as part of our CI

JonathanWoollett-Light:
@roypat  The lint guarantees `New unsafe code is documented.` as the checkpoint notes.

I would be in favor of adding a note on `unsafe` in [the contribution guide](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md) like:
> Unsafe should be documented with a justification for its usage (over its safe alternative where one exists) and an explanation on why it cannot produce undefined behavior.
> E.g.
> ```
> // Test creating a resource.
> // JUSTIFICATION: This cannot be accomplished without unsafe as `external_function()` returns `RawFd`. 
> // An alternative here still uses unsafe e.g. `let _ = unsafe { OwnedFd::from_raw_fd(external_function()) }`.
> // SAFETY: `external_function()` returns a valid file descriptor.
> unsafe {
>     libc::close(external_function());
> }
> ```
```

roypat:
> @roypat The lint guarantees `New unsafe code is documented.` as the checkpoint notes.
> 
> I would be in favor of adding a note on `unsafe` in [the contribution guide](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md) like:
> 
> > Usage of `unsafe` should be documented with a justification (over its safe alternative where one exists) and an explanation on why it cannot produce undefined behavior.
> > E.g.
> > ```
> > // Test creating a resource.
> > // JUSTIFICATION: This cannot be accomplished without unsafe as `external_function()` returns `RawFd`. 
> > // An alternative here still uses unsafe e.g. `drop(unsafe { OwnedFd::from_raw_fd(external_function()) });`.
> > // SAFETY: `external_function()` returns a valid file descriptor.
> > unsafe {
> >     libc::close(external_function());
> > }
> > ```
> 
> Is this a reasonable solution?

I think that's a good idea, allows us to be more detailed on our requirements, too. Maybe we could link the contributor guide in the pull request template?

roypat (dismissed):

JonathanWoollett-Light:
> The "all commits are signed" one of the other hand can go imo, we check that as part of our CI

I agree, I have also removed this.

roypat (dismissed):

pb8o (approved):

roypat (approved):

aghecenco (approved):

Issue #3349, opened 2023-01-04
[Bug] Memory monitor logic does not support multiple memory regions layout


dianpopa:
# Describe the bug

Some intermittent failures can be noticed for tests that configure microVMs that exceed the one memory region layout on x86.
Like `test_create_large_diff_snapshot`.
The test fails only sometimes because most of the times the memory monitor does not get a chance to run a second time (we have a 1s gap between runs) before the test finishes.
 
## To Reproduce

* decrease MEMORY_SAMPLE_TIMEOUT_S to 0.5
* run `test_create_large_diff_snapshot`

## Expected behaviour

* aforementioned test passes all the time

PR #3399, opened 2023-01-26
Support multiple regions in memory monitor

kalyazin:
## Changes

This change adds support for multiple regions in the memory monitor and enables memory monitor for test_create_large_diff_snapshot test.

## Reason

Before the change, the memory monitor could not work with guests with large memory (>3328M) on x86_64.
Resolves: https://github.com/firecracker-microvm/firecracker/issues/3349

## License Acceptance

## PR Checklist

---

### DIFF
--- a/tests/host_tools/memory.py
+++ b/tests/host_tools/memory.py
@@ -3,7 +3,6 @@
 """Utilities for measuring memory utilization for a process."""
 from queue import Queue
 import time
-import platform
 from threading import Thread, Lock
 
 from framework import utils
@@ -36,7 +35,10 @@ class MemoryMonitor(Thread):
         Thread.__init__(self)
         self._pid = None
         self._guest_mem_mib = None
-        self._guest_mem_start = None
+        self._guest_mem_start_1 = None
+        self._guest_mem_end_1 = None
+        self._guest_mem_start_2 = None
+        self._guest_mem_end_2 = None
         self._exceeded_queue = Queue()
         self._pmap_out = None
         self._threshold = self.MEMORY_THRESHOLD
@@ -96,12 +98,6 @@ class MemoryMonitor(Thread):
         while not self._should_stop:
             mem_total = 0
             try:
-                if (
-                    platform.machine() == "x86_64"
-                    and (self.guest_mem_mib * 1024) > self.X86_MEMORY_GAP_START
-                ):
-                    # TODO Remove when <https://github.com/firecracker-microvm/firecracker/issues/3349> is closed.
-                    return
                 _, stdout, _ = utils.run_cmd(pmap_cmd)
                 pmap_out = stdout.split("\n")
 
@@ -118,14 +114,9 @@ class MemoryMonitor(Thread):
                 except ValueError:
                     # This line doesn't contain memory related information.
                     continue
-                if (
-                    self._guest_mem_start is None
-                    and total_size == self.guest_mem_mib * 1024
-                ):
-                    # This is the start of the guest's memory region.
-                    self._guest_mem_start = address
+                if self.update_guest_mem_regions(address, total_size):
                     continue
-                if self.is_in_guest_mem_region(address):
+                if self.is_in_guest_mem_regions(address):
                     continue
                 mem_total += rss
             with self._lock:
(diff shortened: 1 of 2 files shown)


tests/host_tools/memory.py
```diff
@@ -137,12 +126,42 @@ def run(self):
...
+                self._guest_mem_start_1 = address
+                return True
+            if not self._guest_mem_start_2:
+                self._guest_mem_start_2 = address
+                return True
+        return False
+
+    def is_in_guest_mem_regions(self, address):
+        """Check if the address is inside a guest memory region."""
+        if self._guest_mem_start_1:
```
pb8o, line 156: Could this ever be 0? It may be more clear to say:

```python
if self._guest_mem_start_1 is not None:
   ...
```

here and below

tests/host_tools/memory.py
```diff
@@ -137,12 +126,42 @@ def run(self):
...
+            guest_mem_end = self._guest_mem_start_1 + self.guest_mem_mib
+            if self._guest_mem_start_1 <= address < guest_mem_end:
+                return True
+        if self._guest_mem_start_2:
+            guest_mem_end = self._guest_mem_start_2 + self.guest_mem_mib
+            if self._guest_mem_start_2 <= address < guest_mem_end:
+                return True
+        return False
```
pb8o, line 164: Optional: consider using ranges here:

```
In [18]: r = range(1000, 10_000)

In [19]: 9_999 in r
Out[19]: True

In [20]: 10_000 in r
Out[20]: False
```

so it would look like:

```
for guest_range in [self._guest_range_1, self._guest_range_2]:
    if guest_range is not None and address in guest_range:
        return True
return False
```

One thing I don't understand is that both ranges are the same size (`self.guest_mem_mib`)? Shouldn't they have different sizes?

tests/host_tools/memory.py
```diff
@@ -137,12 +126,39 @@ def run(self):
...
+        cache it and return True, otherwise return False.
+        """
+
+        # If x86_64 guest memory exceeds 3328M, it will be split
+        # in 2 regions: 3328M and the rest. We have 3 cases here
+        # to recognise a guest memory region:
+        #  - its size matches the guest memory exactly
+        #  - its size is 3328M
+        #  - its size is guest memory minus 3328M.
+        if size_kib in (
```
bchalios, line 141: Is testing sizes correct? Couldn't there be more mappings that happen to have the same sizes as the ones we care about?

bchalios: However, I'm not sure there's an alternative (unless we export address info from Firecracker).

tests/host_tools/memory.py
```diff
@@ -137,12 +126,39 @@ def run(self):
...
+                return True
+        return False
+
+    def is_in_guest_mem_regions(self, address):
+        """Check if the address is inside a guest memory region."""
+        for guest_mem_start in [self._guest_mem_start_1, self._guest_mem_start_2]:
+            if guest_mem_start is not None:
+                guest_mem_end = guest_mem_start + self.guest_mem_mib
+                if guest_mem_start <= address < guest_mem_end:
+                    return True
```
bchalios, line 160: super nit, so feel free to ignore. I tend to prefer the following:

```suggestion
            if guest_mem_start is None:
                continue
            guest_mem_end = guest_mem_start + self.guest_mem_mib
            if guest_mem_start <= address < guest_mem_end:
                return True
```

bchalios (approved):
LGTM. I had just some minor comments/questions.

Also, if you change "Related: https://github.com/firecracker-microvm/firecracker/issues/3349" to "Resolves: https://github.com/firecracker-microvm/firecracker/issues/3349", it will automatically resolve the issue once this gets merged.

tests/host_tools/memory.py
kalyazin: done

tests/host_tools/memory.py
kalyazin: rewrote that anyway

tests/host_tools/memory.py
kalyazin: No, I am not aware of an alternative either. There certainly can be false negatives, but they are not very probable.

tests/host_tools/memory.py
kalyazin: > Also, if you change "Related: #3349" to "Resolves: #3349", it will automatically resolve the issue once this gets merged.

Thanks, didn't know that.

bchalios (approved):

pb8o (approved):

PR #3405, opened 2023-01-30
fix: Exclude CPUID templates on Arm

JonathanWoollett-Light:
## Changes

Apply `#[cfg(target_arch = "x86_64")]` to CPUID template to avoid compiling them on Arm.

## Reason

Minimizes negative impact on Arm coverage.

## License Acceptance

## PR Checklist

---

### DIFF
--- a/src/cpuid-templates/Cargo.toml
+++ b/src/cpuid-templates/Cargo.toml
@@ -1,9 +1,0 @@
-[package]
-name = "cpuid-templates"
-version = "0.1.0"
-authors = ["Amazon Firecracker team <firecracker-devel@amazon.com>"]
-edition = "2021"
-license = "Apache-2.0"
-
-[dependencies]
-cpuid = { path="../cpuid" }
--- a/src/vmm/Cargo.toml
+++ b/src/vmm/Cargo.toml
@@ -35,7 +35,6 @@ vm-memory = { path = "../vm-memory" }
 
 [target.'cfg(target_arch = "x86_64")'.dependencies]
 cpuid = { path = "../cpuid" }
-cpuid-templates = { path = "../cpuid-templates" }
 
 [dev-dependencies]
 criterion = "0.4.0"
--- a/src/vmm/src/vstate/vcpu/x86_64/cpuid_templates.rs
+++ b/src/vmm/src/vstate/vcpu/x86_64/cpuid_templates.rs
@@ -1418,7 +1418,6 @@ pub fn t2() -> Cpuid {
 ///    0x80860000 0x00: eax=0x00000000 ebx=0x00000000 ecx=0x00000000 edx=0x00000000
 ///    0xc0000000 0x00: eax=0x00000000 ebx=0x00000000 ecx=0x00000000 edx=0x00000000
 /// ```
-
 pub fn t2s() -> Cpuid {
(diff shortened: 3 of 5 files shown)


kalyazin (commented):
Is there a way to exclude the entire crate from non-x86 builds altogether?

JonathanWoollett-Light:
> Is there a way to exclude the entire crate from non-x86 builds altogether?

By default it builds every crate in the workspace. 

There is an `--exclude` argument on build so we could conditionally exclude this crate when on `aarch64` with some extra Python code in `test_coverage`. I personally dislike this approach.

kalyazin:
> > Is there a way to exclude the entire crate from non-x86 builds altogether?
> 
> By default it builds every crate in the workspace.
> 
> There is an `--exclude` argument on build so we could conditionally exclude this crate when on `aarch64` with some extra Python code in `test_coverage`. I personally dislike this approach that adds complexity to the test itself.

That's fair enough. My only concern is if we ever need to add something to the ```src/cpuid-templates/src/lib.rs```, we'll have to keep in mind adding ```#[cfg(target_arch = "x86_64")]```, which is not very ergonomic.

JonathanWoollett-Light:
> > > Is there a way to exclude the entire crate from non-x86 builds altogether?
> > 
> > 
> > By default it builds every crate in the workspace.
> > There is an `--exclude` argument on build so we could conditionally exclude this crate when on `aarch64` with some extra Python code in `test_coverage`. I personally dislike this approach that adds complexity to the test itself.
> 
> That's fair enough. My only concern is if we ever need to add something to the `src/cpuid-templates/src/lib.rs`, we'll have to keep in mind adding `#[cfg(target_arch = "x86_64")]`, which is not very ergonomic.

I agree this is non-optimal. I think this is the common case where many of our crates should really be modules.

At the time I was trying to segregate functionality to make it more readable. Perhaps now a better solution here would be moving `cpuid-templates` into a module under `src/vmm/src/vstate/vcpu/x86_64/cpuid_templates.rs` what are your thoughts on this?

kalyazin:
> > > > Is there a way to exclude the entire crate from non-x86 builds altogether?
> > > 
> > > 
> > > By default it builds every crate in the workspace.
> > > There is an `--exclude` argument on build so we could conditionally exclude this crate when on `aarch64` with some extra Python code in `test_coverage`. I personally dislike this approach that adds complexity to the test itself.
> > 
> > 
> > That's fair enough. My only concern is if we ever need to add something to the `src/cpuid-templates/src/lib.rs`, we'll have to keep in mind adding `#[cfg(target_arch = "x86_64")]`, which is not very ergonomic.
> 
> I agree this is non-optimal. I think this is the common case where many of our crates should really be modules.
> 
> At the time I was trying to segregate functionality to make it more readable. Perhaps now a better solution here would be moving `cpuid-templates` into a module under `src/vmm/src/vstate/vcpu/x86_64/cpuid_templates.rs` what are your thoughts on this?

That looks sensible to me. At some point we will probably rename to something like cpu_templates, when they also include MSRs. We don't seem to have enough information now on how it is going to be structured though. If you can't see any immediate problems in such evolution, I'm fine with moving like you proposed.

kalyazin (approved):

zulinx86 (approved):

PR #3407, opened 2023-01-31
Feat virtio rng rate limiter

bchalios:
## Changes

Add rate limiting for the virtio-rng device. The properties of the rate-limiter are defined in the `/entropy` API call through a [RateLimiter](https://github.com/firecracker-microvm/firecracker/blob/34570dc7912de2d49a99e2f995e763a90236202c/src/api_server/swagger/firecracker.yaml#L1081) optional object.

If this is not defined, then the device will be operated without a rate-limiter.

## Reason

Requests for entropy from the guest consume host resource (currently we perform a `getrandom` system call). In production environments users might want to restrict the amount of entropy provided to each guest.

## License Acceptance

## PR Checklist

---


bchalios:
This is currently rebased on top of #3254, which should be merged first!

bchalios:
> This is currently rebased on top of #3254, which should be merged first!

#3254 has now been merged!

bchalios:
API changes documentation and README entries will be provided with a follow-up PR when all virtio-rng related changes have been merged on the feature branch

src/devices/src/virtio/rng/device.rs
```diff
@@ -237,7 +292,10 @@ impl VirtioDevice for Entropy {
 
 #[cfg(test)]
 mod tests {
+    use std::time::Duration;
+
     use super::*;
+    use crate::check_metric_after_block;
```
roypat, line 298: everytime I see this function I have to do a double take because I think its related to the block device 😂

src/devices/src/virtio/rng/device.rs
```diff
@@ -404,9 +483,135 @@ mod tests {
...
+            METRICS.entropy.entropy_event_count.count(),
+            entropy_event_count + 1
+        );
+        assert_eq!(METRICS.entropy.entropy_bytes.count(), entropy_bytes + 512);
+        assert_eq!(METRICS.entropy.host_rng_fails.count(), host_rng_fails);
+    }
+
+    #[test]
+    fn test_bad_rate_limiter_event() {
+        let mem = create_virtio_mem();
```
roypat, line 505: Nit: Could we use `single_region_mem` from the same `test_utils` module instead here?

src/devices/src/virtio/rng/device.rs
```diff
@@ -404,9 +483,135 @@ mod tests {
...
+        th.add_desc_chain(RNG_QUEUE, 0, &[(0, 4000, VIRTQ_DESC_F_WRITE)]);
+        check_metric_after_block!(
+            METRICS.entropy.entropy_bytes,
+            4000,
+            th.device().process_entropy_queue()
+        );
+        assert!(!th.device().rate_limiter.is_blocked());
+
+        // Sleep for 1 second to completely replenish the rate limiter
+        std::thread::sleep(Duration::from_millis(1000));
```
roypat, line 538: Nit: Could we just call `RateLimiter::manual_replenish` here? Just in case there's some weird timing conditions that could cause spurious failures.

src/api_server/swagger/firecracker.yaml
```diff
@@ -464,6 +464,13 @@ paths:
       description:
         Enables an entropy device that provides high-quality random data to the guest.
       operationId: putEntropyDevice
+      parameters:
+        - name: body
+          in: body
+          description: Guest entropy device properties
+          required: false
```
roypat, line 471: I don't really know my way around the API server, but is the body of this PUT request really optional? This code 

```rs
pub(crate) fn parse_put_entropy(body: &Body) -> Result<ParsedRequest, Error> {
    let cfg = serde_json::from_slice::<EntropyDeviceConfig>(body.raw())?;
    Ok(ParsedRequest::new_sync(VmmAction::SetEntropyDevice(cfg)))
}
```

seems to imply that its not (e.g. I can't just put `null` in the body)

src/devices/src/virtio/rng/event_handler.rs
```diff
@@ -57,6 +60,7 @@ impl MutEventSubscriber for Entropy {
 
         if !event_set.contains(EventSet::IN) {
             warn!("entropy: Received unknowon event: {event_set:?} from source {source}");
```
roypat, line 62: nit: "unknowon" -> "unknown"

src/devices/src/virtio/rng/device.rs
bchalios: I dislike it too and I think in this particular case we can do it. But these things are all over the place and I'm not sure we can get rid of them. I 'm a bit worried about having the rate limiter replenishing itself and us doing `manual_replenish` will make this logic quite hard to follow.

src/api_server/swagger/firecracker.yaml
bchalios: I think that what does the trick here is that `EntropyDeviceConfig` includes a `Option<RateLimiterConfig>`.

src/devices/src/virtio/rng/event_handler.rs
bchalios: ty!

src/devices/src/virtio/rng/device.rs
roypat: Ah I see, yeah I didn't take that into consideration. Makes sense to keep it this way then

src/devices/src/virtio/rng/device.rs
roypat: Fair, I guess the prefered solution would then be to update all other tests to use `create_virtio_mem` and that's not in-scope here 👍

src/api_server/swagger/firecracker.yaml
roypat: Mh, I'm not sure, there's a difference between having `body = {rate_limiter: null}` and `body = null`

bchalios: hmmm, you are right. Taking a look.

bchalios: Ok fixed!

roypat (approved):

roypat (approved):

src/devices/src/virtio/rng/device.rs
```diff
@@ -43,15 +44,18 @@ pub struct Entropy {
     queues: Vec<Queue>,
     queue_events: Vec<EventFd>,
     irq_trigger: IrqTrigger,
+
+    // Device specific fields
+    rate_limiter: RateLimiter,
```
JonathanWoollett-Light, line 49: Nit: This could be a doc comment.
```suggestion
    /// Device specific fields
    rate_limiter: RateLimiter,
```

JonathanWoollett-Light (approved):

src/vmm/src/vmm_config/entropy.rs
```diff
@@ -90,7 +107,7 @@ mod tests {
     #[test]
     fn test_set_device() {
         let mut builder = EntropyDeviceBuilder::new();
-        let device = Entropy::new().unwrap();
+        let device = Entropy::new(RateLimiter::default()).unwrap();
```
JonathanWoollett-Light, line 110: Nit: Could `Entropy` implement `Default`?
```suggestion
        let device = Entropy::default().unwrap();
```

PR #3409, opened 2023-02-01
Doc updates for block device rate limiting and API usage

mattschlebusch:
## Changes

* Add description of block device resource management with an updated API call sample of how to configure the rate limiter on a block device.
* Add some pointers for troubleshooting API usage with `curl`

## License Acceptance

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -31,10 +31,10 @@ The generic requirements are explained below:
 - **KVM**
 
   Please make sure that:
-  1. you have KVM enabled in your Linux kernel, and
-  2. you have read/write access to `/dev/kvm`.
-     If you need help setting up access to `/dev/kvm`, you should check out
-     [Appendix A](#appendix-a-setting-up-kvm-access).
+    1. you have KVM enabled in your Linux kernel, and
+    2. you have read/write access to `/dev/kvm`.
+       If you need help setting up access to `/dev/kvm`, you should check out
+       [Appendix A](#appendix-a-setting-up-kvm-access).
 
 To check if your system meets the requirements to run Firecracker, clone
 the repository and execute `tools/devtool checkenv`.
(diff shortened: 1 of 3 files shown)


docs/getting-started.md
```diff
@@ -364,6 +364,23 @@ not run on an EC2 .metal instance. You can skip performance tests with:
  ./tools/devtool test -- --ignore integration_tests/performance
  ```
 
+## Troubleshooting
+
+#### Errors while using `curl` to access the API
+
+Points to check to confirm the API socket is running and accessible:
+* Check that the user running the Firecracker process and the user using `curl` have equivalent privileges. 
+  For example, if you run Firecracker with **sudo** that you run `curl` with **sudo** as well
```
pb8o: Shouldn't r/w access to the socket be sufficient? I.e. `[ -w SOCKET ] && echo "Good to go"`.

docs/getting-started.md
```diff
@@ -364,6 +364,23 @@ not run on an EC2 .metal instance. You can skip performance tests with:
...
+* **SELinux** can regulate access to sockets on RHEL based distributions. 
+  You can check the configuration of **SELinux** in `/etc/selinux/config`.
+* With the Firecracker process running using `--api-sock /tmp/firecracker.socket`, confirm that the socket is open:
+  * `netstat -a | grep '/tmp/firecracker.socket'`
```
pb8o: `netstat` is deprecated in favor of `ss`. It has similar options so `ss -a` should work.

docs/prod-host-setup.md
```diff
@@ -216,6 +216,28 @@ containing other qdiscs, most classful qdiscs perform rate control.
   - `connlimit` - restricts the number of connections for a destination IP
     address/from a source IP address, as well as limit the bandwidth
 
+### Mitigating Block Device Issues
```
pb8o, line 219: Maybe `Limiting Block Device Throughput` or `Mitigating Noisy Neighbours`

mattschlebusch: Sure, I can rephrase. How about `Mitigating Noisy-Neighbour Related Block Device Issues`?

docs/prod-host-setup.md
```diff
@@ -216,6 +216,28 @@ containing other qdiscs, most classful qdiscs perform rate control.
...
+These pages are written to their mapped storage devices asynchronously
+at the host operating system's discretion. As a result, high storage output
+can result in this cache being filled quickly resulting in a backlog which
+can slow down I/O of other guests on the host.
+
+To protect the resource access of the guests, make sure to tune each Firecracker
+process via the following tools:
+
+- Jailer: A wrapper environment designed to contain Firecracker and strictly
+          control what the process and its guest has access to.
```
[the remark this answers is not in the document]
bchalios: Actually, we have a whole subsection on [Jailer](https://github.com/firecracker-microvm/firecracker/blob/main/docs/prod-host-setup.md#jailer-configuration). It looks like we are just repeating info here. What do you think?

bchalios (commented):
These all look reasonable to me.

docs/prod-host-setup.md
mattschlebusch: How does `Configuring block device resource access for a multi-tenant environment` sound?

docs/prod-host-setup.md
```diff
@@ -216,6 +216,28 @@ containing other qdiscs, most classful qdiscs perform rate control.
   - `connlimit` - restricts the number of connections for a destination IP
     address/from a source IP address, as well as limit the bandwidth
 
+### Mitigating Block Device Issues
+
+Data written to storage devices is managed in Linux with a dirty page cache.
```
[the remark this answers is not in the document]
mattschlebusch: I want to emphasize that is the dirtying of pages not being written through that is causing the problem.

docs/prod-host-setup.md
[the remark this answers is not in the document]
mattschlebusch: I want to draw attention to the Jailer as an option for the issue itself. I can add a hyperlink here though.

docs/prod-host-setup.md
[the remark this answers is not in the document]
bchalios: it is not necessarily. Even when we read-only from disk, we will consume pages from the cache (which means we might evict some other process) which will probably have similar noisy-neighbour effects

PR #3430, opened 2023-02-11
fix(cpuid): add leaf 0xb / subleaf 1 as needed

zulinx86:
## Changes

To keep backward compatibility for guest VM CPU topology, the sub-leaf is added as needed.

## Reason

The following commit changed the bahavior of KVM_GET_SUPPORTED_CPUID to no longer include leaf 0xb / sub-leaf 1.
https://lore.kernel.org/all/20221027092036.2698180-1-pbonzini@redhat.com/ 

## License Acceptance

## PR Checklist

---

- ~~[ ] This functionality can be added in [`rust-vmm`][1].~~

### DIFF
--- a/src/cpuid/src/transformer/intel.rs
+++ b/src/cpuid/src/transformer/intel.rs
@@ -118,6 +118,30 @@ fn update_extended_topology_entry(
 pub struct IntelCpuidTransformer {}
 
 impl CpuidTransformer for IntelCpuidTransformer {
+    fn process_cpuid(&self, cpuid: &mut CpuId, vm_spec: &VmSpec) -> Result<(), Error> {
+        // The following commit changed the behavior of KVM_GET_SUPPORTED_CPUID to no longer
+        // include leaf 0xb / sub-leaf 1.
+        // https://lore.kernel.org/all/20221027092036.2698180-1-pbonzini@redhat.com/
+        // The registers within sub-leaves of 0xB are zeroed by update_extended_topology_entry()
+        // and then modified. So the value they are set to at this point doesn't matter.
+        if !cpuid
+            .as_slice()
+            .iter()
+            .any(|entry| entry.function == 0xb && entry.index == 1)
+        {
+            cpuid
+                .push(kvm_cpuid_entry2 {
+                    function: leaf_0xb::LEAF_NUM,
+                    index: 1,
+                    flags: kvm_bindings::KVM_CPUID_FLAG_SIGNIFCANT_INDEX,
+                    ..kvm_cpuid_entry2::default()
+                })
+                .map_err(Error::Fam)?;
+        }
+
+        self.process_entries(cpuid, vm_spec)
+    }
+
     fn entry_transformer_fn(&self, entry: &mut kvm_cpuid_entry2) -> Option<EntryTransformerFn> {
         match entry.function {
             leaf_0x1::LEAF_NUM => Some(common::update_feature_info_entry),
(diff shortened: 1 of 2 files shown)


src/cpuid/src/transformer/intel.rs
```diff
@@ -118,6 +118,28 @@ fn update_extended_topology_entry(
 pub struct IntelCpuidTransformer {}
 
 impl CpuidTransformer for IntelCpuidTransformer {
+    fn process_cpuid(&self, cpuid: &mut CpuId, vm_spec: &VmSpec) -> Result<(), Error> {
+        // The following commit changed the bahavior of KVM_GET_SUPPORTED_CPUID to no longer
```
pb8o, line 122: nit: typo in bahavior

pb8o (dismissed):

src/cpuid/src/transformer/intel.rs
zulinx86: Thanks! Fixed.

pb8o (dismissed):

src/cpuid/src/transformer/intel.rs
```diff
@@ -118,6 +118,28 @@ fn update_extended_topology_entry(
...
+            .any(|entry| entry.function == 0xb && entry.index == 1)
+        {
+            cpuid
+                .push(kvm_cpuid_entry2 {
+                    function: leaf_0xb::LEAF_NUM,
+                    index: 1,
+                    flags: kvm_bindings::KVM_CPUID_FLAG_SIGNIFCANT_INDEX,
+                    ..kvm_cpuid_entry2::default()
+                })
+                .map_err(Error::Fam)?;
```
JonathanWoollett-Light, line 137: I believe it is important this is inserted with zero values? In this case where we care about the value, we should specifically use 0 rather than the default (which could change in the future).
```suggestion
            cpuid
                .push(kvm_cpuid_entry2 {
                    function: leaf_0xb::LEAF_NUM,
                    index: 1,
                    flags: kvm_bindings::KVM_CPUID_FLAG_SIGNIFCANT_INDEX,
                    eax: 0,
                    ebx: 0,
                    ecx: 0,
                    edx: 0,
                    padding: Default::default()
                })
                .map_err(Error::Fam)?;
```

zulinx86: Since EAX through EDX values are overwritten in `update_extended_topology_entry()`, any values are acceptible and don't matter here, I think. What do you think?

JonathanWoollett-Light: You are right. It may be worth adding a comment on this to be clear, like:

> The registers within sub-leaves of `0xB` are zeroed by `update_extended_topology_entry()` and then modified. So the value they are set to at this point doesn't matter.

zulinx86: Looks better! Thanks!

JonathanWoollett-Light (dismissed):

src/cpuid/src/transformer/intel.rs
```diff
@@ -139,6 +163,32 @@ mod tests {
...
+            ecx: 0,
+            edx: 0,
+            padding: [0, 0, 0],
+        };
+        cpuid.push(entry).unwrap();
+
+        assert!(IntelCpuidTransformer {}
+            .process_cpuid(&mut cpuid, &vm_spec)
+            .is_ok());
+        assert!(!cpuid.as_slice().contains(&entry));
```
kalyazin, line 185: I'm not sure it's in line with the rest of our tests, but a brief description of what is being done in the test and what result is expected would help it read better IMHO.

pb8o (dismissed):

src/cpuid/src/transformer/intel.rs
zulinx86: Good point! Additional commit would be helpful and it seems to be able to be simplified.
Thanks!

kalyazin (approved):

JonathanWoollett-Light (approved):

PR #3435, opened 2023-02-13
Wire up guest_config crate

mattschlebusch:
## Changes

* Create `guest_config` crate
* Move Firecracker specific CPUID code from `vmm` to the `guest_config` crate.

## Reason

* Having a distinct crate to encapsulate types and logic for handling guest configuration manages the dependencies and provides a natural namespace for modules as sub-components of the crate. 
* Moving CPUID types outside of the `vmm` crate resolves circular dependencies that can be introduced when `vmm` needs to make use of other crates, that they themselves use. (e.g. `vmm` -> `guest_config` -> `vmm`)

## License Acceptance


Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["src/firecracker", "src/jailer", "src/seccompiler", "src/rebase-snap"]
+members = ["src/firecracker", "src/jailer", "src/seccompiler", "src/rebase-snap", "src/guest_config"]
```
pb8o, line 2: I think this does not need to be here, since it's a dependency? These are only for top-level binaries AFAIR.

mattschlebusch: I will look into this a bit more. It's possible I missed something when configuring the crate, cause clippy required this entry, but I do agree that it shouldn't be necessary.

mattschlebusch:
This move has been discussed offline. To simplify dependency management for now, we will me squashing all CPU configuration related code into the vmm crate for now.

PR #3438, opened 2023-02-13
fix: Move `cpuid` into module under `vmm`

JonathanWoollett-Light:

## Changes

Moved `cpuid` crate into a module under `vmm`.

## Reason

See https://github.com/firecracker-microvm/firecracker/pull/3079

## License Acceptance

## PR Checklist

---


kalyazin:
Looks like we need to find a consensus between this and https://github.com/firecracker-microvm/firecracker/pull/3435 .

ShadowCurse:
Why change this in main? Shouldn't we first finalize `cpu-template` branch? And then we can merge all at once.

src/vmm/src/cpuid/brand_string.rs
```diff
@@ -4,7 +4,7 @@
 use std::arch::x86_64::__cpuid as host_cpuid;
 use std::slice;
 
-use crate::common::{VENDOR_ID_AMD, VENDOR_ID_INTEL};
+use crate::cpuid::common::{VENDOR_ID_AMD, VENDOR_ID_INTEL};
```
ShadowCurse, line 7: Maybe it is worth converting all `crate::cpuid` into just `super`? That way all inner module paths are self contained. And if we ever need to move this module anywhere else, it would be much easier.

JonathanWoollett-Light: Some can be changed to `super` but most end up looking like `super::super::super` which I don't think is ideal.

ShadowCurse: Well.. we already have `super::super` somewhere as I recall. But I was thinking about always using the shortest path. So if module need access to itself it uses `super::something` instead of `crate::module::something`. Maybe it is worth having some verbal rule about it.

ShadowCurse (approved):

zulinx86 (approved):

PR #3442, opened 2023-02-15
Fix CPU Load Monitor silently crashing

roypat:
We used to get CI failures resulting from

```
File "/firecracker/tests/host_tools/cpu_load.py", line 70, in run
   cpu_load = utils.ProcessManager.get_cpu_percent(self._process_pid)["real"]
KeyError: 'real'
```

At some point, pytest started to only issue warnings on exception in non-main threads, instead of failing the test. This means that it looked like these spurious failures went away, even though they really did not.

The CPU load monitor in its current form was broken, as the `get_cpu_percent` function returns a dictionary mapping process names to CPU utilization of each individual thread (despite it being annotated to return a single float). However, CpuLoadMonitor expected that dictionary to contain the cpu utilization under a key named 'real'. I'm assuming at some point in the past, `get_cpu_percent` was updated from returning the `top`-columns of the firecracker process to its current implementation, but the load monitor was not updated.

This commit changes the CpuLoadMonitor implementation to use `get_cpu_percent` correctly. Additionally, the delay between subsequent samples is reduces to 50ms (from 1s), to match behavior of the memory monitor, see #3441.

Lastly, the test_rate_limiter::test_rx_rate_limiting_cpu_load test never actually checked the cpu utilization samples to not exceed the threshold, so this is fixed too.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] ~~If a specific issue led to this PR, this PR closes the issue.~~
- [ ] ~~API changes follow the [Runbook for Firecracker API changes][2].~~
- [ ] ~~User-facing changes are mentioned in `CHANGELOG.md`.~~
- [ ] ~~New `TODO`s link to an issue.~~

---

### DIFF
--- a/tests/framework/utils.py
+++ b/tests/framework/utils.py
@@ -13,6 +13,7 @@ import typing
 import time
 import platform
 
+from typing import Dict
 from collections import namedtuple, defaultdict
 import psutil
 from retry import retry
@@ -52,7 +53,7 @@ class ProcessManager:
         return psutil.Process(pid).cpu_affinity(real_cpulist)
 
     @staticmethod
-    def get_cpu_percent(pid: int) -> float:
+    def get_cpu_percent(pid: int) -> Dict[str, Dict[str, float]]:
         """Return the instant process CPU utilization percent."""
         _, stdout, _ = run_cmd(GET_CPU_LOAD.format(pid))
         cpu_percentages = {}
@@ -60,11 +61,15 @@ class ProcessManager:
         # Take all except the last line
         lines = stdout.strip().split(sep="\n")
         for line in lines:
+            # sometimes the firecracker process will have gone away, in which case top does not return anything
+            if not line:
+                continue
+
             info = line.strip().split()
             # We need at least CPU utilization and threads names cols (which
             # might be two cols e.g `fc_vcpu 0`).
             info_len = len(info)
-            assert info_len > 11
+            assert info_len > 11, line
 
             cpu_percent = float(info[8])
             task_id = info[0]
--- a/tests/integration_tests/functional/test_rate_limiter.py
+++ b/tests/integration_tests/functional/test_rate_limiter.py
@@ -165,7 +165,6 @@ def test_rx_rate_limiting_cpu_load(test_microvm_with_api, network_config):
     )
 
     test_microvm.start()
-
     # Start iperf server on guest.
     _start_iperf_on_guest(test_microvm, guest_ip)
 
--- a/tools/devtool
+++ b/tools/devtool
@@ -660,7 +660,7 @@ cmd_test() {
         -v /boot:/boot \
         ${ramdisk_args} \
         -- \
-        pytest "$@"
+        pytest -W error::pytest.PytestUnraisableExceptionWarning -W error::pytest.PytestUnhandledThreadExceptionWarning "$@"
 
     ret=$?
 
--- a/tests/host_tools/cpu_load.py
+++ b/tests/host_tools/cpu_load.py
@@ -67,10 +67,21 @@ class CpuLoadMonitor(Thread):
         It is up to the caller to check the queue.
         """
         while not self._should_stop:
-            cpu_load = utils.ProcessManager.get_cpu_percent(self._process_pid)["real"]
-            if cpu_load > self.threshold:
-                self.cpu_load_samples.append(cpu_load)
-            time.sleep(1)  # 1 second granularity.
+            cpus = utils.ProcessManager.get_cpu_percent(self._process_pid)
+
+            try:
+                fc_threads = cpus["firecracker"]
+
+                # There can be multiple "firecracker" threads sometimes, see #3429
+                assert len(fc_threads) > 0
+
+                for _, cpu_load in fc_threads.items():
+                    if cpu_load > self._threshold:
+                        self._cpu_load_samples.append(cpu_load)
+            except KeyError:
+                pass  # no firecracker process
+
+            time.sleep(0.05)  # 50 milliseconds granularity.
 
     def check_samples(self):
         """Check that there are no samples above the threshold."""


tests/integration_tests/functional/test_rate_limiter.py
```diff
@@ -178,6 +177,13 @@ def test_rx_rate_limiting_cpu_load(test_microvm_with_api, network_config):
     )
     _iperf_out = _run_local_iperf(iperf_cmd)
 
+    # Stop the CPU monitor thread - otherwise it will crash after the microvm goes away due to no firecracker process
+    # existing anymore
+    test_microvm.cpu_load_monitor.signal_stop()
```
sudanl0, line 182: should we call test_microvm.kill() here instead because the kill() is calling signal_stop/join/check_samples on the cpu_load_monitor as shown here:
https://github.com/firecracker-microvm/firecracker/pull/3442/files#diff-d8abdd84a323dd0369671d7aa8e0359578e641946c5b764a8b7c82cb36167486L217

pb8o, line 182: hmm the `microvm` fixture should kill the vm after it is done. It also does the sample check. Is it not doing it?

roypat: Yeah, I just saw that after Sudan's comment - that's an oversight on my part. However, it seems that the microvm dies before the sampler thread (because inside of `get_cpu_count` we now get an AssertionError about no `top` output being found - but this also happens right now if you look at the CI). I'll investigate this more later (will look at how the memory monitor does it and copy that approach), I'll put the PR back into draft mode for the time being. 

Aside,  do we really want to suppress exceptions on other threads? Was this a deliberate change, or should I also look into how to can revert that?

pb8o: Yeah if there is an easy way for pytest not to ignore those exceptions I think we would want that. Otherwise I fear it may be hiding bugs.

pb8o: Seems there is https://docs.pytest.org/en/6.2.x/usage.html#warning-about-unraisable-exceptions-and-unhandled-thread-exceptions

roypat: Amazing, thanks, I'll tag a commit with that onto this PR then :)

roypat: See https://github.com/firecracker-microvm/firecracker/pull/3442#discussion_r1108174457

tests/framework/utils.py
```diff
@@ -52,7 +53,7 @@ def set_cpu_affinity(pid: int, cpulist: list) -> list:
         return psutil.Process(pid).cpu_affinity(real_cpulist)
 
     @staticmethod
-    def get_cpu_percent(pid: int) -> float:
+    def get_cpu_percent(pid: int) -> Dict[str, Dict[str, float]]:
```
sudanl0, line 56: https://github.com/firecracker-microvm/firecracker/pull/2227/commits/d6961e77af73471b9cc189f06f1354dbc3eabd5f this seems to be the initial commit for this function which means the return type has been incorrect since the beginning. Since python3 doesn't check the return type annotations (AFAIK) should we have some static analyzer to find more of these incorrect return types?

tests/integration_tests/functional/test_rate_limiter.py
```diff
@@ -165,7 +165,6 @@ def test_rx_rate_limiting_cpu_load(test_microvm_with_api, network_config):
     )
 
     test_microvm.start()
-
```
sudanl0, line 168: nit: since there is no change in this file, can we remove this file from the pr?

tools/devtool
```diff
@@ -660,7 +660,7 @@ cmd_test() {
         -v /boot:/boot \
         ${ramdisk_args} \
         -- \
-        pytest "$@"
+        pytest -W error::pytest.PytestUnraisableExceptionWarning -W error::pytest.PytestUnhandledThreadExceptionWarning "$@"
```
sudanl0, line 663: did you see this work in your testing?

roypat: Yep, there's old buildkite runs where it triggered, e.g. https://buildkite.com/firecracker/firecracker-pr/builds/1048

tests/framework/utils.py
roypat: oh whoops, forgot to respond here earlier. 

I don't think there's something that works 100% for python, given its dynamic typing. I supposed there might be a lint along the lines "this function is annotated as returning `float`, but you're treating the return value as a dict`, but I'm not sure if even that is possible without abundant false-positives

tests/integration_tests/functional/test_rate_limiter.py
roypat: I'd be inclined to leave it in unless there's any non-nit things that'll have to be addressed anyway

tests/framework/utils.py
pb8o: There is a standard static type checker, mypy, but I think if we add it now we will get tons of errors.

tests/framework/utils.py
```diff
@@ -13,6 +13,7 @@
 import time
 import platform
 
+from typing import Dict
```
pb8o, line 16: I think since 3.7 on you can also just use dict https://stackoverflow.com/questions/37087457/difference-between-defining-typing-dict-and-dict

(not important)

pb8o (approved):

tests/framework/utils.py
roypat: Oh cool, good to know! My python knowledge is pretty much limited to what existed in 3.6 haha

sudanl0 (approved):

PR #3467, opened 2023-02-22
Move userfaultfd handler implementation to its own crate

luminitavoicu:
## Changes

Move userfaultfd handler example implementation to Firecracker crates.

This PR does the following:
- move uffd handler implementation to its own crate
- restructure code for a more modular approach
- decouple handler implementation from malicious handler code
- unit test functionality

## Reason
Enhance uffd handler.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/uffd_handler/src/common.rs
```diff
@@ -30,3 +30,53 @@ pub fn parse_unix_stream(stream: &UnixStream) -> Result<(File, String), StreamEr
...
+
+    use super::*;
+
+    #[test]
+    fn test_parse_unix_stream() {
+        let (mut sender, receiver) = UnixStream::pair().unwrap();
+        let expected_body = "test body".to_string();
+        let fd = sender.socket_fd();
+
+        // Test case: successfully receive and parse a message body and file descriptor from sender.
```
kalyazin, line 47: Here and below: would it be possible to have separate tests instead? That could be more readable/modifiable and will provide more granular feedback.

src/uffd_handler/src/handler.rs
```diff
@@ -256,3 +256,177 @@ fn get_peer_process_credentials(stream: UnixStream) -> Result<libc::ucred> {
...
+        }
+
+        // Simulate zeroing out the third page.
+        fn zero_out(&self, _addr: u64) -> Result<(u64, u64)> {
+            Ok((0x1000, 0x2000))
+        }
+    }
+
+    fn create_handler(start: u64, mem_size: usize) -> PageFaultHandler {
+        // // Create a memory mapping of one region with size and address specified.
```
kalyazin, line 295: nit: double //

src/uffd_handler/src/handler.rs
```diff
@@ -1,59 +1,159 @@
...
+    }
 
-        println!(
-            "poll() returns: nready = {}; POLLIN = {}; POLLERR = {}",
-            nready,
-            revents & libc::POLLIN,
-            revents & libc::POLLERR,
-        );
+    fn populate_from_file(&self, region: &MemRegion) -> (u64, u64) {
+        let src = self.backing_buffer as u64 + region.mapping.offset;
```
roypat: It seems like `self.backing_buffer` starts life as a `*mut c_void` returned from `mmap`, it casted to `*const u8` when stored in `PageFaultHandler::backing_buffer`, is converted to a `u64` here, just to be casted back to `*const c_void` below. Could we have it be a `*const c_void` throughout its entire lifecycle, and do the pointer arithmetic here via [`ptr::add`](https://doc.rust-lang.org/std/primitive.pointer.html#method.add)?

src/uffd_handler/src/handler.rs
```diff
@@ -1,59 +1,159 @@
...
-                &MemPageState::Removed,
-            ),
-            _ => panic!("Unexpected event on userfaultfd"),
+        // Make sure the UFFD copied some bytes.
+        assert!(ret > 0);
+
+        (start_addr, start_addr + len as u64)
+    }
+
+    fn zero_out(&mut self, addr: u64) -> (u64, u64) {
```
roypat: same as above, can we represent the address as an actual pointer (and if not, can we represent the address using `usize`, which I think would be the more appropriate numeric type)?

src/uffd_handler/Cargo.toml
```diff
@@ -10,7 +10,7 @@ license = "Apache-2.0"
```
kalyazin, line 10: nit: typo in the commit message description.
Also, consider using https://www.conventionalcommits.org/en/v1.0.0/#summary for being explicit about the purpose of each commit.

src/uffd_handler/src/handler.rs
```diff
@@ -0,0 +1,432 @@
...
+            revents: 0,
+        };
+
+        // Loop, handling incoming events on the userfaultfd file descriptor.
+        loop {
+            // SAFETY: Safe because fd, nfds and timeout are valid parameters.
+            let nready = unsafe { libc::poll(&mut pollfd, 1, POLL_TIMEOUT) };
+            // Poll has an infinite timeout, therefore in theory this case should never happen.
+            if nready == -1 {
+                unreachable!();
```
roypat, line 214: seems like a signal could cause this: https://man7.org/linux/man-pages/man2/poll.2.html#ERRORS - maybe better to just `continue` instead of panicking?

src/uffd_handler/src/handler.rs
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        Ok((addr, addr + ret as u64))
+    }
+}
+
+pub struct PageFaultHandler {
+    mem_regions: Vec<MemRegion>,
+    backing_buffer: *const u8,
+    pub uffd: Box<dyn UffdManager>,
```
roypat, line 112: nit: Instead of doing dynamic dispatch via trait objects, we can statically dispatch by parameterizing `PageFultHandler` by `U: UffdManager`, I think.

src/uffd_handler/src/memory_region.rs
```diff
@@ -0,0 +1,179 @@
...
+
+/// Convert the guest memory mappings received from the Firecracker process
+/// into a vector of `MemRegion`s containing the guest mappings and page
+/// state information.
+pub fn create_mem_regions(mappings: &Vec<GuestRegionUffdMapping>) -> Vec<MemRegion> {
+    let page_size = get_page_size().unwrap();
+    let mut mem_regions: Vec<MemRegion> = Vec::with_capacity(mappings.len());
+
+    for r in mappings.iter() {
+        let mapping = r.clone();
```
roypat, line 55: nit: `GuestRegionUffdMapping` can derive `Copy`

roypat (commented):
Looks pretty good - only thing is that the handling of pointers seems a bit adventurous. Lots of converting between `* c_void`, `* u8`, `u64` and `usize`. It'd be nice if we could use a pointer type consistently (e.g. `* c_void` or `* u8`) wherever possible, and only drop down to `usize` when we have to do some manipulation for which rust's ptr module does not offer functions (e.g. masking of lower bits to get page base address). `u64` I feel is wrong to use for addresses (although it works on all architectures we support at the moment)

src/uffd_handler/src/handler.rs
luminitavoicu: I think here it would be better to handle each case separately, for example for these error types we should terminate the process (retrying will most likely not solve the issue):
```
       EFAULT fds points outside the process's accessible address space.
              The array given as argument was not contained in the
              calling program's address space.

       EINVAL The nfds value exceeds the RLIMIT_NOFILE value.

       ENOMEM Unable to allocate memory for kernel data structures.
```

not 100% decided on this one, need to dive a little deeper:
```
       EINTR  A signal occurred before any requested event; see
              [signal(7)](https://man7.org/linux/man-pages/man7/signal.7.html).
```

src/uffd_handler/src/handler.rs
luminitavoicu: You are right. Done.

src/uffd_handler/src/handler.rs
luminitavoicu: I think here it would be better to handle each case separately, for example for these error types we should terminate the process (retrying will most likely not solve the issue):
```
       EFAULT fds points outside the process's accessible address space.
              The array given as argument was not contained in the
              calling program's address space.

       EINVAL The nfds value exceeds the RLIMIT_NOFILE value.

       ENOMEM Unable to allocate memory for kernel data structures.
```

not 100% decided on this one, need to dive a little deeper:
```
       EINTR  A signal occurred before any requested event; see
              [signal(7)](https://man7.org/linux/man-pages/man7/signal.7.html).
```

kalyazin (approved):

src/uffd_handler/src/handler.rs
roypat: Mh, the reason I'd be in favor of `ptr::add` is that even if we avoid it (and thus `unsafe`), we'd still have to uphold all the invariants it requires, but without `unsafe` we get lulled into a false sense of security in the sense that "oh, nothing can go wrong" - although I'm also not sure how much of those invariants even imply here, since we don't get the pointer from rust's allocator, so am happy to leave things the way they are :)

src/uffd_handler/src/handler.rs
roypat: a 32-bit hypervisor supporting a 64-bit sounds... wild, haha.

As mentioned above, am happy with the way things are now, as we have just one pointer type (`* c_void`) and one numeric type (`usize`) to deal with addresses

roypat (approved):

src/uffd_handler/src/handler.rs
luminitavoicu: I updated the code to exit in case of `EFAULT`, `EINVAL` and `ENOMEM`, because these are not recoverable and we might end up looping endlessly. For  `EINTR`, the handler retries the operation, because the signal triggered before the event can be harmless. In case it is not, `poll` should fail after the retry with one of the unrecoverable errors.

roypat (approved):

kalyazin (approved):

PR #3469, opened 2023-02-22
fix: skip restore if template dir not found

sudanl0:
It is possible that some X template is not supported on the instance where the snapshots were created and, snapshot is loaded on an instance where X is supported. This results in error since restore doesn't find the file to load. To avoid this we make an assumption that if template directory does not exist then snapshot was not created for that template and we skip the test.

## Changes

Assume that if snapshot_artifacts/wrmsr/vmlinux-4.14/$TEMPLATE  does not exist then snapshot was not taken for this template and skip restore for the same ($TEMPLATE could be T2CL, T2S, etc.).

The skipped line will be shown as below (T2CL is taken as an example here):
> integration_tests/functional/test_cpu_features.py::test_cpu_cpuid_restore[T2CL] SKIPPED

## Reason

Snapshot restore test could fail is snapshot was taken on Skylake and restore was done on Cascade lake so, we need this fix to avoid this issue before it happens.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

~~- [ ] If a specific issue led to this PR, this PR closes the issue.~~
~~- [ ] API changes follow the [Runbook for Firecracker API changes][2].~~
~~- [ ] User-facing changes are mentioned in `CHANGELOG.md`.~~
~~- [ ] New `TODO`s link to an issue.~~

---

~~- [ ] This functionality can be added in [`rust-vmm`][1].~~

### DIFF
--- a/tests/integration_tests/functional/test_cpu_features.py
+++ b/tests/integration_tests/functional/test_cpu_features.py
@@ -58,6 +58,37 @@ def _check_cpu_features_arm(test_microvm):
     )
 
 
+def get_cpu_template_dir(cpu_template):
+    """
+    Utility function to return a valid string which will be used as
+    name of the directory where snapshot artifacts are stored during
+    snapshot test and loaded from during restore test.
+
+    """
+    return cpu_template if cpu_template else "none"
+
+
+def skip_test_based_on_artifacts(snapshot_artifacts_dir):
+    """
+    It is possible that some X template is not supported on
+    the instance where the snapshots were created and,
+    snapshot is loaded on an instance where X is supported. This
+    results in error since restore doesn't find the file to load.
+    e.g. let's suppose snapshot is created on Skylake and restored
+    on Cascade Lake. So, the created artifacts could just be:
+    snapshot_artifacts/wrmsr/vmlinux-4.14/T2S
+    but the restore test would fail because the files in
+    snapshot_artifacts/wrmsr/vmlinux-4.14/T2CL won't be available.
+    To avoid this we make an assumption that if template directory
+    does not exist then snapshot was not created for that template
+    and we skip the test.
+    """
+    if not Path.exists(snapshot_artifacts_dir):
+        reason = f"\n Since {snapshot_artifacts_dir} does not exist \
+                we skip the test assuming that snapshot was not"
+        pytest.skip(re.sub(" +", " ", reason))
+
+
 @pytest.mark.skipif(PLATFORM != "x86_64", reason="CPUID is only supported on x86_64.")
 @pytest.mark.parametrize(
     "num_vcpus",
@@ -426,7 +457,7 @@ def _test_cpu_wrmsr_snapshot(context):
     snapshot_artifacts_dir = (
         Path(shared_names["snapshot_artifacts_root_dir_wrmsr"])
         / context.kernel.base_name()
-        / (cpu_template if cpu_template else "none")
+        / get_cpu_template_dir(cpu_template)
     )
     shutil.rmtree(snapshot_artifacts_dir, ignore_errors=True)
     os.makedirs(snapshot_artifacts_dir)
@@ -573,6 +604,15 @@ def _test_cpu_wrmsr_restore(context):
     microvm_factory = context.custom["microvm_factory"]
     cpu_template = context.custom["cpu_template"]
 
+    cpu_template_dir = get_cpu_template_dir(cpu_template)
+    snapshot_artifacts_dir = (
+        Path(shared_names["snapshot_artifacts_root_dir_wrmsr"])
+        / context.kernel.base_name()
+        / cpu_template_dir
+    )
+
+    skip_test_based_on_artifacts(snapshot_artifacts_dir)
+
     vm = microvm_factory.build()
     vm.spawn()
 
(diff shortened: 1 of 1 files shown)


tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -572,6 +572,32 @@ def _test_cpu_wrmsr_restore(context):
...
+        Path(shared_names["snapshot_artifacts_root_dir_wrmsr"])
+        / context.kernel.base_name()
+        / cpu_template_dir
+    )
+
+    # It is possible that some X template is not supported on
+    # the instance where the snapshots were created and,
+    # snapshot is loaded on an instance where X is supported. This
+    # results in error since restore doesn't find the file to load.
+    # for e.g. lets suppose snapshot is created on Skylake and restored
```
kalyazin, line 586: nit: "e.g." (exempli gratia) already means "for example", so "for" is superfluous.
s/lets/let's/

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -572,6 +572,32 @@ def _test_cpu_wrmsr_restore(context):
...
+    # but the restore test would fail because the files in
+    # snapshot_artifacts/wrmsr/vmlinux-4.14/T2CL won't be available.
+    # To avoid this we make an assumption that if template directory
+    # does not exist then snapshot was not created for that template
+    # and we skip the test.
+    if not Path.exists(snapshot_artifacts_dir):
+        reason = f"\n Since {snapshot_artifacts_dir} does not exist \
+                we skip the test assuming that snapshot was not     \
+                created for {cpu_template=}"
+        print(re.sub(" +", " ", reason))
```
kalyazin, line 598: nit: I don't think I saw something like that for a multiline string before. Could be there a more idiomatic way of doing that?
Also, my impression is we don't tend to use `print()` in tests, because they may not be visible, unless pytest is run with `-s`. Is using  logger an option?

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -817,6 +836,32 @@ def _test_cpu_cpuid_restore(context):
     microvm_factory = context.custom["microvm_factory"]
     cpu_template = context.custom["cpu_template"]
 
+    cpu_template_dir = cpu_template if cpu_template else "none"
```
kalyazin, line 839: This piece looks very similar to the previous one. Would it make sense to extract it into a function?

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -572,6 +572,32 @@ def _test_cpu_wrmsr_restore(context):
     microvm_factory = context.custom["microvm_factory"]
```
kalyazin, line 572: nit: commit message: I think we should mention in the subject that this fix is for a test, preferably pointing to a specific test (or file).

sudanl0: I have changed it to "fix:test_cpu_features:skip restore if artifacts not found", let me know if it sounds ok

tests/integration_tests/functional/test_cpu_features.py
sudanl0: there was 1 other place where we just had  (cpu_template if cpu_template else "none"), so I created get_cpu_template_dir() to replace them all

kalyazin: yeah, I meant also the comment with the explanation. If one updates that, they need to be aware of the second instance too.

tests/integration_tests/functional/test_cpu_features.py
kalyazin: nit: if you can fit spaces after colons, would be perfect

tests/integration_tests/functional/test_cpu_features.py
sudanl0: logger is added but to use it we would have to modify all the buildkite jobs to have --log-cli-level=INFO
for the multiline string comment, re.sub seems to be the standard way to remove multiple spaces from a string.

sudanl0: pytest seems to be the preferred way so going with that.

tests/integration_tests/functional/test_cpu_features.py
sudanl0: done, moved the common code to skip_test_based_on_artifacts()

kalyazin (dismissed):

kalyazin (approved):

mattschlebusch (approved):
As a quick fix, this is ok. There is a risk that simply missing artifacts would mean a test isn't run when it should have been, and the test would be incorrectly skipped when it should have reported a failure. A preferable way to my mind would be an explicitly defined path with known/expected artifacts which would be able to report on a failed test or a simply skipped test.

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -58,6 +58,37 @@ def _check_cpu_features_arm(test_microvm):
...
+    snapshot_artifacts/wrmsr/vmlinux-4.14/T2S
+    but the restore test would fail because the files in
+    snapshot_artifacts/wrmsr/vmlinux-4.14/T2CL won't be available.
+    To avoid this we make an assumption that if template directory
+    does not exist then snapshot was not created for that template
+    and we skip the test.
+    """
+    if not Path.exists(snapshot_artifacts_dir):
+        reason = f"\n Since {snapshot_artifacts_dir} does not exist \
+                we skip the test assuming that snapshot was not"
```
pb8o, line 88: maybe just `snapshot dir does not exist` to avoid the long string?

pb8o (approved):

tests/integration_tests/functional/test_cpu_features.py
sudanl0: I wanted to print the directory name to have more info in the logs.

PR #3475, opened 2023-02-24
Custom and static Aarch64 templates

ShadowCurse:
## Changes
- Updated the way custom templates are presented
- Updated the way vcpus are configured
- Updated Aarch64 cpu template and configuration
- Updated X86_64 cpu configuration 
- Added static Aarch64 template

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -103,6 +88,7 @@ impl KvmVcpu {
         &mut self,
         guest_mem: &GuestMemoryMmap,
         kernel_load_addr: GuestAddress,
+        vcpu_config: &VcpuConfig,
```
kalyazin, line 91: worth mentioning the new argument in the function's doc.

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -68,7 +50,9 @@ pub struct KvmVcpu {
     pub mmio_bus: Option<devices::Bus>,
 
     mpidr: u64,
+    template_registers_ids: Vec<u64>,
```
kalyazin, line 53: nit: may template_register_ids sounds better?

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -15,49 +14,32 @@ use versionize::{VersionMap, Versionize, VersionizeResult};
...
-    /// Error getting the Vcpu preferred target on Arm.
+    #[error("Error getting the Vcpu preferred target on Arm: {0}")]
     GetPreferredTarget(kvm_ioctls::Error),
-    /// Error doing Vcpu Init on Arm.
+    #[error("Error doing Vcpu Init on Arm: {0}")]
     Init(kvm_ioctls::Error),
-    /// Failed to set value for some arm specific register.
+    #[error("Error applying template: {0}")]
+    CpuTemplate(TemplateError),
+    #[error("Failed to set value for some arm specific register: {0}")]
```
kalyazin, line 37: nit: could be an opportunity to unify the error message language (aarch64 vs Arm vs arm).

src/vmm/src/arm_templates/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+pub fn apply_template(
+    vcpu: &VcpuFd,
+    template: &CpuTemplateAarch64,
+) -> std::result::Result<(), Error> {
+    for (reg_id, reg_value) in template.iter() {
+        let value = {
+            let curr_value = vcpu
+                .get_one_reg(*reg_id)
+                .map_err(|error| Error::GetReg(*reg_id, error))?;
+            (curr_value & reg_value.mask) | reg_value.value
```
kalyazin, line 48: still think we have to align with @mattschlebusch 's interpretation (clear+set vs mask+value semantics).

src/vmm/src/arm_templates/mod.rs
```diff
@@ -0,0 +1,54 @@
...
+impl Deref for CpuTemplateAarch64 {
+    type Target = BTreeMap<u64, RegisterValue>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl CpuTemplateAarch64 {
+    pub fn registers_ids(&self) -> Vec<u64> {
```
kalyazin, line 34: nit: register_ids?

src/arch/src/aarch64/regs.rs
```diff
@@ -443,6 +436,27 @@ pub fn save_system_registers(vcpu: &VcpuFd, state: &mut Vec<Aarch64Register>) ->
     Ok(())
 }
 
+/// Get the state of the additional registers.
+///
+/// # Arguments
+///
+/// * `vcpu` - Structure for the VCPU that holds the VCPU's fd.
+/// * `state` - Structure for returning the state of the system registers.
```
kalyazin, line 444: may be worth clarifying that this is both input and output argument.

src/arch/src/aarch64/regs.rs
```diff
@@ -6,7 +6,7 @@
 // found in the THIRD-PARTY file.
```
kalyazin, line 6: oh, can also mention in the commit message that we're adding a new feature (I'm not a fan of no-verb sentences).

src/vmm/src/vstate/vcpu/aarch64.rs
ShadowCurse: it seems all modules named `aarch64` so I can rename everything to use this naming style

src/vmm/src/vstate/vcpu/aarch64.rs
ShadowCurse: My logic is:
registers_ids -> there are many registers and each has an id
register_ids -> there is one register and it has many ids
...naming is hard

src/vmm/src/vstate/vcpu/aarch64.rs
ShadowCurse: done

zulinx86 (commented):
Could you squash the commit "renamed registers_ids to register_ids" into the commit "aarch64 cpu templates"?

src/vmm/src/guest_config/aarch64_templates/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+/// Module with static aarch64 template
+pub mod g3g2;
+
+/// Error enum for aarch64 cpu templates
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// Error getting register
+    #[error("Error getting register {0}: {1}")]
+    GetReg(u64, kvm_ioctls::Error),
```
zulinx86, line 17: nit: unifying `GetRegister` (like `arch::aarch64::regs::Error`)?
```suggestion
    GetRegister(u64, kvm_ioctls::Error),
```

src/vmm/src/guest_config/aarch64_templates/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+    #[test]
+    fn test_apply_template() {
+        let (_, vcpu, _) = setup_vcpu(0x10000);
+        let template = aarch64_cpu_template();
+
+        assert!(template.apply(&vcpu.fd).is_ok());
+
+        let new_reg_value = vcpu.fd.get_one_reg(MPIDR_EL1).unwrap();
+        // checking that 30th bit is set to 0
+        assert_eq!(new_reg_value & (1 << 30), 0)
```
zulinx86, line 115: How about checking all possible patterns `{(0, 0), (0, 1), (1, 0), (1, 1)}`?
```
                RegisterValue {
                    mask: 0b10011111111111111111111111111111,
                    value: 0b00110000000000000000000000000000,
                },
```

src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -13,44 +13,31 @@ pub const DEFAULT_MEM_SIZE_MIB: usize = 128;
...
+         enabled."
+    )]
     InvalidVcpuCount,
     /// Could not get the config of the balloon device from the VM resources, even though a
     /// balloon device was previously installed.
+    #[error(
+        "Could not get the configuration of the previously installed balloon device to validate \
+         the memory size."
+    )]
     InvalidVmState,
```
zulinx86, line 39: nit: might be better to split `thiserror`-related changes into a different commit as refactor.

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -99,10 +87,13 @@ impl KvmVcpu {
     ///
     /// * `guest_mem` - The guest memory used by this microvm.
     /// * `kernel_load_addr` - Offset from `guest_mem` at which the kernel is loaded.
+    /// * `kernel_load_addr` - Offset from `guest_mem` at which the kernel is loaded.
```
zulinx86, line 90: nit: duplicate line.
```suggestion
```

src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -217,27 +196,9 @@ where
     Ok(val)
 }
 
-/// Deserialization function for the `cpu_template` field in `VmConfig` and `VmUpdateConfig`.
-/// This is called only when `cpu_template` is present in the JSON configuration.
-fn deserialize_cpu_template<'de, D, T>(_d: D) -> std::result::Result<T, D::Error>
```
zulinx86, line 222: Could you give more context about why `deserialize_cpu_template()` no longer required?

kalyazin:
If this is not a work in progress, I suggest going through commit messages and making them more informative, by adding verbs and scopes of the changes, perhaps making use of https://www.conventionalcommits.org/en/v1.0.0/#summary .

src/arch/src/aarch64/regs.rs
```diff
@@ -284,12 +277,12 @@ pub fn read_mpidr(vcpu: &VcpuFd) -> Result<u64> {
     }
 }
 
-/// Get the state of the core registers.
+/// Saves the states of the core registers into [`state`].
```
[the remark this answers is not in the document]
ShadowCurse: removed

src/arch/src/aarch64/regs.rs
```diff
@@ -443,6 +436,28 @@ pub fn save_system_registers(vcpu: &VcpuFd, state: &mut Vec<Aarch64Register>) ->
     Ok(())
 }
 
+/// Saves states of registers into [`state`]
+///
+/// # Arguments
+///
+/// * `vcpu` - Structure for the VCPU that holds the VCPU's fd.
+/// * `ids` - Slice of registers ids to save
```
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/guest_config/aarch64_templates/mod.rs
ShadowCurse: decided to just use arch::aarch64::regs::Error instead of creating a new one

src/vmm/src/vstate/vcpu/aarch64.rs
ShadowCurse: done

src/vmm/src/vmm_config/machine_config.rs
ShadowCurse: Well.. I actually don't really understand why it was needed in the first place. It was supposed to be a barrier for aarch64, so if the snapshot with cpu template was restored on aarch64, it would error out. But on aarch64 the cpu templates did not exist anyway, and the field should always be None. And because it should always be None, it should be never serialized because of the `skip_serializing_if`. So the method should have never been invoked. (At least this applies to `VmConfig`)
Anyway, now that aarch64 has cpu templates this method has no purpose.

src/vmm/src/guest_config/aarch64_templates/mod.rs
ShadowCurse: what do you mean by all patterns? I change 1 bit and check that it is set to the value I need.

src/vmm/src/guest_config/aarch64_templates/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+use std::collections::BTreeMap;
+
+use arch::aarch64::regs::Error;
+use kvm_ioctls::VcpuFd;
+
+/// Module with static aarch64 template
+pub mod g3g2;
+
+/// Structure that represents change in the register
+/// We zero bits with zero mask and set with values mask
```
luminitavoicu, line 13: This is a bit confusing. Could we rephrase?

src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -254,13 +215,31 @@ pub enum CpuFeaturesTemplate {
...
+/// to EC2 instances.
+#[cfg(target_arch = "aarch64")]
+#[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize, Versionize)]
+pub enum CpuFeaturesTemplate {
+    // TODO REMOVE THIS WITH REFACTORING OF CpuFeaturesTemplate
+    /// Empty0
+    Empty0,
+    // TODO REMOVE THIS WITH REFACTORING OF CpuFeaturesTemplate
+    /// Empty1
+    Empty1,
```
luminitavoicu, line 228: Are these still needed?

src/vmm/src/guest_config/aarch64_templates/g3g2.rs
```diff
@@ -0,0 +1,64 @@
...
+                // disable sha3
+                //
+                // set sha2 to disable sha3
+                //
+                // Disable sm3 and sm4 CPU features.
+                // "This field (sm3) must have the same value as ID_AA64ISAR0_EL1.SM4."
+                //
+                // Disable asimdfhm (fhm) CPU feature.
+                //
+                // Disable flagm (ts) CPU feature.
```
luminitavoicu, line 35: These comments were initially accompanied by the specified operations. Let's rephrase these to make a little more sense.

src/vmm/src/guest_config/aarch64_templates/g3g2.rs
```diff
@@ -0,0 +1,64 @@
...
+use arch::aarch64::regs::{ID_AA64ISAR0_EL1, ID_AA64ISAR1_EL1, ID_AA64MMFR2_EL1, ID_AA64PFR0_EL1};
+
+use super::{CpuTemplateAarch64, RegisterValue};
+
+/// Aarch64 template to mask graviton3 as graviton2
+pub fn g3g2() -> CpuTemplateAarch64 {
+    CpuTemplateAarch64(
+        vec![
+            (
+                // Disable dit CPU feature.
```
luminitavoicu, line 13: It would be great if we could provide also link to the ARM developer spec, to make them easier to find.

luminitavoicu: Also let's specify for all of them the bits and the register they are placed in, the way you do below:
```
sha3 occupies bits [35:32] in ID_AA64ISAR0_EL1
```

src/arch/src/aarch64/regs.rs
```diff
@@ -34,51 +34,38 @@ pub struct Aarch64Register {
...
+    SetRegister(u64, kvm_ioctls::Error),
+    /// Failed to get the register list.
+    #[error("Failed to retrieve list of registers: {0}")]
+    GetRegList(kvm_ioctls::Error),
+    /// Failed to get multiprocessor state.
+    #[error("Failed to get multiprocessor state: {0}")]
+    GetMP(kvm_ioctls::Error),
     /// Failed to Set multiprocessor state.
+    #[error("Failed to set multiprocessor state: {0}")]
     SetMP(kvm_ioctls::Error),
```
zulinx86, line 59: nit: It might be better to use upper camel case on this occasion (like `Fam`, `GetMidrEl1`).
```suggestion
    GetMp(kvm_ioctls::Error),
    /// Failed to Set multiprocessor state.
    #[error("Failed to set multiprocessor state: {0}")]
    SetMp(kvm_ioctls::Error),
```

src/arch/src/aarch64/regs.rs
```diff
@@ -436,7 +429,33 @@ pub fn save_system_registers(vcpu: &VcpuFd, state: &mut Vec<Aarch64Register>) ->
...
+
+    Ok(())
+}
+
+/// Saves states of registers into `state`
+///
+/// # Arguments
+///
+/// * `vcpu` - Structure for the VCPU that holds the VCPU's fd.
+/// * `ids` - Slice of registers ids to save,
```
zulinx86, line 446: nit: `,` => `.`
```suggestion
/// * `ids` - Slice of registers ids to save.
```

src/vmm/src/guest_config/aarch64_templates/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+/// (register & mask) | value
```
zulinx86, line 14: I thought that if a bit of mask is 1 use cpu template, otherwise use host value. In other words, `(value & mask) | (register & !mask)`. Please correct me if i'm wrong.

src/vmm/src/guest_config/aarch64_templates/mod.rs
zulinx86: Sorry, I was wrong. Please forget this.

src/vmm/src/vmm_config/machine_config.rs
zulinx86: Indeed. Thanks for explanation.

src/vmm/src/guest_config/aarch64_templates/mod.rs
ShadowCurse: I can try to rephrase, but the problem is... we still can not agree on one way to use these `mask` and `value`. There are 3 or 4 proposals and I just use the one I thought of first. Also this `RegisterValue` struct technically should be in `common` types somewhere.

src/vmm/src/vmm_config/machine_config.rs
ShadowCurse: So the reason for these dummy enum values is that for some reason when we restore snapshot on aarch64, the `cpu_template` enum is present in the snapshot (even thought it should not) and this enum has value None. Because of that, we try to deserialize this enum. And this deserialization fails, because now on aarch64 we only have 2 values in the enum. But if you look at the `CpuFeatureTemplate` for x86, the None value is 4th (so the integer value is 3). So I had to add 2 dummy values to make None variant to be 4th.

src/arch/src/aarch64/regs.rs
```diff
@@ -173,6 +160,12 @@ macro_rules! arm64_sys_reg {
 arm64_sys_reg!(MPIDR_EL1, 3, 0, 0, 0, 5);
 arm64_sys_reg!(MIDR_EL1, 3, 0, 0, 0, 0);
 
+// ID registers
+arm64_sys_reg!(ID_AA64PFR0_EL1, 3, 0, 0, 4, 0);
+arm64_sys_reg!(ID_AA64ISAR0_EL1, 3, 0, 0, 6, 0);
+arm64_sys_reg!(ID_AA64ISAR1_EL1, 3, 0, 0, 6, 1);
+arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
+
```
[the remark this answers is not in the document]
ShadowCurse: done

src/arch/src/aarch64/regs.rs
```diff
@@ -436,7 +429,33 @@ pub fn save_system_registers(vcpu: &VcpuFd, state: &mut Vec<Aarch64Register>) ->
...
+            value: vcpu
+                .get_one_reg(*index)
+                .map_err(|e| Error::GetRegister(*index, e))?,
+        });
+    }
+
+    Ok(())
+}
+
+/// Saves states of registers into `state`
```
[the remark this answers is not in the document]
ShadowCurse: done

src/arch/src/aarch64/regs.rs
```diff
@@ -519,7 +538,7 @@ mod tests {
         assert!(res.is_err());
         assert_eq!(
             format!("{}", res.unwrap_err()),
-            "Failed to get system register: Exec format error (os error 8)"
+            "Failed to get register: 6931039826524487685: Exec format error (os error 8)"
```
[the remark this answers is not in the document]
ShadowCurse: Updated Error type to derive Eq and changed assertion to use that instead of comparing strings

src/vmm/src/guest_config/aarch64_templates/g3g2.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use arch::aarch64::regs::{ID_AA64ISAR0_EL1, ID_AA64ISAR1_EL1, ID_AA64MMFR2_EL1, ID_AA64PFR0_EL1};
+
+use super::{CpuTemplateAarch64, RegisterValue};
+
+/// Aarch64 template to mask graviton3 as graviton2
```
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/guest_config/aarch64_templates/g3g2.rs
ShadowCurse: done

src/vmm/src/guest_config/aarch64_templates/g3g2.rs
ShadowCurse: done

src/arch/src/aarch64/regs.rs
ShadowCurse: done

src/arch/src/aarch64/regs.rs
ShadowCurse: done

src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -254,13 +215,31 @@ pub enum CpuFeaturesTemplate {
...
+    /// G3G2 Template
```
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -115,6 +104,12 @@ impl KvmVcpu {
         self.mpidr =
             arch::aarch64::regs::read_mpidr(&self.fd).map_err(Error::ConfigureRegisters)?;
 
+        if vcpu_config.cpu_template == CpuFeaturesTemplate::V1N1 {
```
kalyazin, line 107: line 107: `V1N1`, line 108: `n1v1`. Let's choose one of them.

ShadowCurse: v1n1 it is

src/vmm/src/resources.rs
```diff
@@ -1283,7 +1283,10 @@ mod tests {
             vcpu_count: Some(32),
             mem_size_mib: Some(512),
             smt: Some(true),
+            #[cfg(target_arch = "x86_64")]
```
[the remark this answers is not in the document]
ShadowCurse: done

src/arch/src/aarch64/regs.rs
```diff
@@ -6,7 +6,7 @@
 // found in the THIRD-PARTY file.
 
 use std::path::PathBuf;
-use std::{fmt, fs, mem, result, u32};
+use std::{fs, mem, result, u32};
```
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/guest_config/aarch64_templates/mod.rs
zulinx86: Dicussed offline, we need more inputs for this part. So we'll leave as it is as of now and fix as follow-up if neded.

zulinx86 (approved):
Looks good to me. Can you add integration tests that checks if all bits of the template is applied correctly inside guests as a follow-up PR (like [this](https://github.com/firecracker-microvm/firecracker/pull/3510/commits/940cd04ec86c4a8c2b57c00f027475b014ed23fe))? I think it might no be feasible, because the access to those system registers are restricted somehow, but it might be able to be done with `getauxval()` or `lscpu`.

ShadowCurse:
The application of this template relies on the patched kernel to access some registers. Until this patch is included into the upstream kernel, there is no way to check its validity through integration tests, because CI does not run with patched kernel. But in the future, sure, I will add integration tests.

zulinx86:
> The application of this template relies on the patched kernel to access some registers. Until this patch is included into the upstream kernel, there is no way to check its validity through integration tests, because CI does not run with patched kernel. But in the future, sure, I will add integration tests.

Sounds great. It would be nice to create a ticket/issue not to forget.

src/vmm/src/builder.rs
```diff
@@ -425,7 +359,32 @@ pub fn build_microvm_for_boot(
...
+            CpuConfiguration::from_cpu_template(&vcpu.kvm_vcpu.fd, cpu_template)
+                .map_err(ApplyCpuTemplate)?,
+        )
+    } else {
+        None
+    };
+
+    #[cfg(target_arch = "x86_64")]
+    let custom_cpu_config = if let Some(ref cpu_template) = vm_resources.vm_config().cpu_template {
+        CpuConfiguration::from_cpu_template(&vmm.vm, cpu_template).map_err(ApplyCpuTemplate)?
```
mattschlebusch, line 379: Having a different API for the architectures here doesn't quite make sense from the caller's point of view. The configuration between the two is understandably different internally, but should require only the same information/references passed to it when it comes time to use the template. This makes sense for the caller of the function so there isn't differentiation at that level, but also makes the unit tests easier to manage as well.

tests/integration_tests/functional/test_api.py
```diff
@@ -469,13 +461,12 @@ def test_api_machine_config(test_microvm_with_api):
...
-    # Set the cpu template again
-    response = test_microvm.machine_cfg.patch(cpu_template="C3")
+    # Set the cpu template
     if platform.machine() == "x86_64":
-        assert test_microvm.api_session.is_status_no_content(response.status_code)
+        response = test_microvm.machine_cfg.patch(cpu_template="C3")
     else:
-        assert test_microvm.api_session.is_status_bad_request(response.status_code)
-        assert "CPU templates are not supported on aarch64" in response.text
+        response = test_microvm.machine_cfg.patch(cpu_template="None")
```
mattschlebusch, line 468: Could we test an aarch64 static template instead?

src/vmm/src/guest_config/x86_64/mod.rs
```diff
@@ -1,17 +1,376 @@
...
+                    CpuFeaturesTemplate::T2CL => cpuid_templates::t2cl::template(),
+                    CpuFeaturesTemplate::T2A => cpuid_templates::t2a::template(),
+                    CpuFeaturesTemplate::None | CpuFeaturesTemplate::Custom => {
+                        unreachable!("None and Custom states are invalid")
+                    }
+                };
+
+                // Include leaves from host that are not present in CPUID template.
+                supported_cpuid = config_cpuid
+                    .include_leaves_from(supported_cpuid)
```
mattschlebusch, line 102: I *suspect* the inclusion of the leaves from the supported config needs to happen after [normalization in vcpu.configure()](https://github.com/firecracker-microvm/firecracker/pull/3475/files/faafb85f665efc7dfdadf288fc23da91bd33b6c9#diff-e1ead8e51789d6db11155356a0b6c2b85f76f4648e5cd5ef882b7cca6b710a52L295). Have a quick look at [the build error](https://buildkite.com/firecracker/firecracker-pr/builds/1800#018708e6-ff0f-495b-9b88-b37328d17e88) to try reproduce.

mattschlebusch (commented):
Quick thoughts:

* Putting the static template inside of `guest_config/aarch64/cpu_templates` makes good sense. Currently the x86 equivalent is in `vstates/vcpu` because the guest_config module didn't exist at the time, but should probably be moved to `guest_config/x86_64/cpu_templates` in a similar way. (out of scope obviously)
* The movement of the template logic inside the impl of the *CpuConfiguration has its merits but convolutes the code in other ways, and doesn't have all the data it should need, or duplicates data that exists in other places of the Vmm type.
* I think the approach here is quite heavy to support the static template and has added a lot of extras to the change as well. The VmConfigNew type is potentially quite confusing for new readers.
* Would it be possible to move the static template use inside of the vcpu.configure() logic instead? The static templates for x86 is already here and represents a much smaller code change to follow this pattern as well. I think this also simplifies the testing around snapshotting as well which I think is implicated by these changes.

src/vmm/src/guest_config/x86_64/mod.rs
ShadowCurse: Well, `inclusion` only fails if the `config_cpuid` and `supported_cpuid` are from different vendors.
The test that fails expects the vm to fail to start.
```python
fail_msg = "Internal error while starting microVM: Error configuring the vcpu for boot:"
assert fail_msg in response.text
```
It is just the error message I need to fix

src/vmm/src/vstate/vcpu/x86_64/mod.rs
```diff
@@ -678,91 +640,75 @@ mod tests {
...
-                &vm_mem,
-                GuestAddress(0),
-                &vcpu_config,
-                vm.supported_cpuid().clone(),
-            ),
+            vcpu.configure(&vm_mem, GuestAddress(0), &vcpu_config,),
             Ok(())
         );
 
+        let try_configure = |vcpu: &mut KvmVcpu, template| -> bool {
```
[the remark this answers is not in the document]
ShadowCurse: Ok, I will update this. The problem was that now there are 2 points that can fail: creation of cpu config and the application of it

tests/integration_tests/functional/test_api.py
ShadowCurse: Not yet. At least no the one I have, because it requires patched 5.10 kernel. Maybe it would be worth creating some additional static template, but then what if someone will want to use it?

PR #3477, opened 2023-02-25
Add common types for CPU configuration

mattschlebusch:
## Changes

* Common types to be used by Firecracker for configuring guest vCPUs.
* Lowered test coverage pending implementation and usage of the common types.

## Reason

These common types will be subject to change, but will serve as a middle "tier" between Firecracker and KVM, as well as Firecracker interfaces and Firecracker internally (`vmm` crate).

## License Acceptance

[`CONTRIBUTING.md`][3].

### DIFF
--- a/src/vmm/src/guest_config/common/mod.rs
+++ b/src/vmm/src/guest_config/common/mod.rs
@@ -1,0 +1,158 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// TODO - Remove allow(unused) once types have been used
+
+use std::collections::HashMap;
+use std::fmt::Debug;
+use std::ops::Add;
+
+use serde::{Deserialize, Serialize};
+
+#[cfg(target_arch = "x86_64")]
+use crate::guest_config::cpuid::Cpuid;
+
+/// Type defined to represent registers across CPU vendors
+/// (pointers and values)
+pub trait Numeric: Copy + Add<Self, Output = Self> {}
+impl Numeric for u32 {}
+impl Numeric for u64 {}
+impl Numeric for u128 {}
+
+/// Map to find and define model-specific registers
+#[cfg(target_arch = "x86_64")]
+#[allow(unused)]
+pub struct MsrRegisterMap {
+    register_map: HashMap<RegisterPointer<u32>, RegisterValue<u32>>,
+}
+/// Map to find and define registers for 64-bit ARM CPUs
+#[cfg(target_arch = "aarch64")]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[allow(unused)]
+pub struct Aarch64RegisterMap {
+    register_map: HashMap<RegisterPointer<u64>, RegisterValue<u128>>,
+}
+
+/// A trait type to be customized for each CPU architecture
+/// to modify CPU configuration for the Firecracker guest
+pub trait ConfigurationModifier {
+    /// Applies the CPUTemplate against the CPUConfiguration
+    fn apply_template(&self, cpu_template: CustomCpuTemplate) -> Box<CpuConfiguration>;
+}
+
+/// CPU configuration that can be of the host or the guest.
+/// Expected to be populated in part or modified by
+/// types implementing ConfigurationModifier
+#[allow(unused)]
+pub struct CpuConfiguration {
+    /// Hashed set containing register configuration
+    /// divided up by type of register. See `RegisterSet`
+    pub register_config: Vec<RegisterSet>,
+}
+
+impl CpuConfiguration {
+    /// Factory method to instantiate CPUConfiguration
+    /// with the provided `RegisterSet`
+    pub fn new(register_config: Vec<RegisterSet>) -> CpuConfiguration {
+        CpuConfiguration { register_config }
+    }
+}
+
+/// Template indicating a modification of CPU configuration
+/// to be applied against the `CPUConfiguration` type
+#[allow(unused)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct CustomCpuTemplate {
+    /// Modifier masks with pointer data for the modifiers
+    /// intended registers
+    pub modifiers: Vec<RegisterModifierSet>,
+}
+
+/// Modifier for a CPU configuration containing
+/// the location of the value to be modified,
+/// and provides a value to be applied alongside
+/// a filter to be applied
+#[allow(unused)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct RegisterModifier<T: Numeric> {
+    /// Location of the register to be modified
+    /// by the mask value and filter
+    pub pointer: RegisterPointer<T>,
+    /// Mask value to be applied
+    pub value: T,
+    /// Mask value is to be applied according
+    /// to what is allowed through by the filter
+    pub filter: T,
+}
+
+/// Numeric wrapper type acting as a marker type
+/// for the register's pointer
+#[allow(unused)]
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub enum RegisterPointer<T: Numeric> {
+    /// Pointer(location) of a specific register
+    NumericPointer(T),
+    /// Register pointer for CPUID that is Leaf/Subleaf aware
+    CpuidPointer(CpuidRegisterPointer),
+}
+
+/// Numeric wrapper type acting as a marker type
+/// for the register's value
+#[allow(unused)]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct RegisterValue<T: Numeric> {
+    /// Value of a specific register
+    pub value: T,
+}
+
+/// CPUID register enumeration
+#[allow(unused)]
+#[allow(missing_docs)]
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub enum CpuidRegister {
+    Eax,
+    Ebx,
+    Ecx,
+    Edx,
+}
+
+/// Composite type that holistically provides
+/// the location of a specific register being used
+/// in the context of a CPUID tree
+#[allow(unused)]
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub struct CpuidRegisterPointer {
+    leaf: u32,
+    subleaf: u32,
+    pointer: CpuidRegister,
+}
+
+/// Enumerates the types of registers and contains
+/// a property capable of containing the data
+/// formatted for the register type
+#[allow(unused)]
+#[allow(missing_docs)]
+pub enum RegisterSet {
+    #[cfg(target_arch = "x86_64")]
+    CpuId(Cpuid),
+    #[cfg(target_arch = "x86_64")]
+    Msrs(MsrRegisterMap),
+    #[cfg(target_arch = "aarch64")]
+    Aarch64Registers(Aarch64RegisterMap),
+}
+
+/// Intended to correlate to the RegisterSet.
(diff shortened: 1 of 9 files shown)


JonathanWoollett-Light:
Nit: I think it makes sense to squash the coverage change into [feat: Add common types of CPU config](https://github.com/firecracker-microvm/firecracker/pull/3477/commits/f5bd085f0fe27698b031fd913bf3536aeec38fc4).

mattschlebusch:
> Nit: I think it makes sense to squash the coverage change into [feat: Add common types of CPU config](https://github.com/firecracker-microvm/firecracker/pull/3477/commits/f5bd085f0fe27698b031fd913bf3536aeec38fc4).

Sure can do. It was my intention to drop the commit eventually given that the coverage *will* change again, but the distinction on commits isn't useful here.

src/vmm/src/guest_config/common/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+impl Numeric for u32 {}
+impl Numeric for u64 {}
+impl Numeric for u128 {}
+
+/// Map to find and define model-specific registers
+#[cfg(target_arch = "x86_64")]
+pub type MsrRegisterMap = HashMap<RegisterPointer<u32>, RegisterValue<u32>>;
+/// Map to find and define registers for 64-bit ARM CPUs
+#[cfg(target_arch = "aarch64")]
+pub type Aarch64RegisterMap = HashMap<RegisterPointer<u64>, RegisterValue<u128>>;
```
kalyazin, line 24: Any reason why "pointer" was chosen over (arguably more common) "id"?

src/vmm/src/guest_config/common/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// A trait type to be customized for each CPU architecture
+/// to modify CPU configuration for the Firecracker guest
+pub trait ConfigurationModifier {
+    /// Applies the CPUTemplate against the CPUConfiguration
+    fn apply_template(&self, cpu_template: CPUTemplate) -> Box<CPUConfiguration>;
```
kalyazin, line 30: I like "CpuTemplate" and "CpuConfig" more.

src/vmm/src/guest_config/common/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/// CPU configuration that can be of the host or the guest
```
kalyazin, line 33: nit: a full stop would help separate the two sentences.

src/vmm/src/guest_config/common/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Value of a specific register
+    pub value: T,
+}
+
+/// CPUID register enumeration
+#[allow(unused)]
+#[allow(missing_docs)]
+#[cfg(target_arch = "x86_64")]
+pub enum CpuidRegister {
+    EAX,
```
kalyazin, line 97: nit: rust naming conventions suggest camel case for enum variants.

src/vmm/src/guest_config/common/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+/// registers
+#[allow(unused)]
+#[allow(missing_docs)]
+pub enum RegisterModifierSet {
+    #[cfg(target_arch = "x86_64")]
+    CpuIdModifierSet(HashSet<RegisterModifier<CpuidRegisterPointer, u32>>),
+    #[cfg(target_arch = "x86_64")]
+    MsrsModifierSet(HashSet<RegisterModifier<RegisterPointer<u32>, u64>>),
+    #[cfg(target_arch = "aarch64")]
+    Aarch64RegistersModifierSet(HashSet<RegisterModifier<RegisterPointer<u32>, u64>>),
```
kalyazin, line 140: should it be "u128>>"?
Would it be too much if we have named types for all pointers and values, eg
```
type MsrId = u32;
type MsrVal = u64;
```

src/vmm/src/guest_config/common/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// to what is allowed through by the filter
+    pub filter: T,
+}
+
+/// Numeric wrapper type acting as a marker type
+/// for the register's pointer
+#[allow(unused)]
+pub struct RegisterPointer<T: Numeric> {
+    /// Pointer(location) of a specific register
+    pub pointer: T,
```
kalyazin, line 81: Is there a specific reason we need a named field here?

src/vmm/src/guest_config/common/mod.rs
```diff
@@ -0,0 +1,141 @@
...
+/// Type defined to represent registers across CPU vendors
+/// (pointers and values)
+pub trait Numeric: Copy + Add<Self, Output = Self> {}
+impl Numeric for u32 {}
+impl Numeric for u64 {}
+impl Numeric for u128 {}
+
+/// Map to find and define model-specific registers
+#[cfg(target_arch = "x86_64")]
+pub type MsrRegisterMap = HashMap<RegisterPointer<u32>, RegisterValue<u32>>;
```
kalyazin, line 21: "reg" is used quite a lot in our repo. Is there an advantage in using "register" instead?

src/vmm/src/guest_config/mod.rs
```diff
@@ -1,5 +1,54 @@
...
+        fn apply_template(&self, cpu_template: CPUTemplate) -> Box<CPUConfiguration> {
+            // TODO - Apply template
+            get_host_cpu_configuration()
+        }
+    }
+
+    fn get_host_cpu_configuration() -> Box<CPUConfiguration> {
+        // TODO - Retrieve host configuration
+
+        Box::new(CPUConfiguration::new(HashSet::new()))
```
kalyazin, line 52: In ARM case, `get_host_cpu_configuration()` depends on the register IDs in the CPU template, as we can't query all possible registers. This isn't a part of the common type definitions though, as I understand.

src/vmm/src/guest_config/common/mod.rs
mattschlebusch: Pointer was the name agreed to for this type offline. No strong opinions.

src/vmm/src/guest_config/common/mod.rs
mattschlebusch: It has been changed.

src/vmm/src/guest_config/common/mod.rs
mattschlebusch: It should be 128. I have updated that. With regards to the named types, I'm not sure they add enough value to warrant it.

src/vmm/src/guest_config/mod.rs
mattschlebusch: This is indeed within the aarch64 specific module. This will require parameterization as well to filter on registers to be queried via KVM in any event, so I expect the definition to change somewhat anyway.

src/vmm/src/guest_config/common/mod.rs
mattschlebusch: My intention here is to add clarity for anyone reading the code. Reg is very vague.

src/vmm/src/resources.rs
```diff
@@ -225,6 +225,7 @@ impl VmResources {
             vcpu_count: self.vm_config().vcpu_count,
             smt: self.vm_config().smt,
             cpu_template: self.vm_config().cpu_template,
+            custom_cpu_template: self.vm_config().custom_cpu_template.clone(),
```
ShadowCurse, line 228: Why separate `cpu_tempate` and `custom_cpu_template`?

src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -240,6 +247,8 @@ where
 /// to EC2 instances.
 #[derive(Clone, Copy, Debug, Deserialize, Eq, PartialEq, Serialize, Versionize)]
 pub enum CpuFeaturesTemplate {
+    /// Custom template
+    CUSTOM,
```
ShadowCurse, line 251: By adding anything to this enum field you break existing snapshots.
This is a problem I faced earlier with aarch64 templates.
This enum needs total overhaul. Maybe in separate PR.

src/vmm/src/guest_config/common/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+
+/// CPU configuration that can be of the host or the guest.
+/// Expected to be populated in part or modified by
+/// types implementing ConfigurationModifier
+#[allow(unused)]
+pub struct CpuConfiguration {
+    /// Hashed set containing register configuration
+    /// divided up by type of register. See `RegisterSet`
+    pub register_config: Vec<RegisterSet>,
+}
```
ShadowCurse, line 51: So from what I can see this struct is not used anywhere. 
But I guess your idea is that it will be used in `vcpu/x86_64` and `vcpu/aarch64` modules.
But then why do we even need this type, if it has totaly different implementations based on the architecture and used only in those implementations. There is nothing in common between x86 and aarch for this struct to exist.

src/vmm/src/guest_config/common/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+/// CPUID register enumeration
+#[allow(unused)]
+#[allow(missing_docs)]
+#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq, Serialize)]
+pub enum CpuidRegister {
+    Eax,
+    Ebx,
+    Ecx,
+    Edx,
+}
```
ShadowCurse, line 117: Does this need to be a common type? Isn't it only used for x86?

src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -65,13 +67,17 @@ pub struct VmConfig {
...
+    /// CPU features exposed to the guest.
     #[serde(
         default,
         deserialize_with = "deserialize_cpu_template",
         skip_serializing_if = "CpuFeaturesTemplate::is_none"
     )]
     pub cpu_template: CpuFeaturesTemplate,
+    /// CPU template to be defined by the user
+    #[serde(skip_serializing)]
+    pub custom_cpu_template: Option<CustomCpuTemplate>,
```
ShadowCurse, line 80: Shouldn't we also always skip `cpu_template` field if all info is stored in the vcpu state?

src/vmm/src/vmm_config/machine_config.rs
zulinx86: Could you give more details? Why does it break existing snapshots?

src/vmm/src/resources.rs
mattschlebusch: CpuFeaturesTemplate requires versionization as it is a part of the snapshot VmInfo type. However the custom CPU template should not be serialized by the snapshot.

src/vmm/src/vmm_config/machine_config.rs
mattschlebusch: I agree. I have split it out for versionization reasons as well.

src/vmm/src/guest_config/common/mod.rs
mattschlebusch: It is not supposed to be used anywhere yet. The commonality is that they will be configured by the same API.

src/vmm/src/guest_config/common/mod.rs
mattschlebusch: It is also used for the API (or any other potential interface in the future like CLI).

src/vmm/src/vmm_config/machine_config.rs
mattschlebusch: That is up for debate. I think in the long term the hard coded CPU template types will be removed completely.

src/vmm/src/resources.rs
zulinx86: +1. The following implementation looks simpler to me.
```rs
pub enum CpuFeaturesTemplate {
    /// Custom template
    CUSTOM(CustomCpuTemplate),
...
}
```

mattschlebusch:
Some concerns of complexity has called for a simplified definition of structs to be used first. Closing this for a simpler approach.

PR #3445, opened 2023-02-15
Use a deterministic hashmap for reproducible builds

ilmanzo:
## Changes

Replace HashMap usage in `seccomp` with BTreeMap to preserve order of items.
This enable reproducible builds (see issue #3439) 

## Reason

fixes #3439 

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following
Developer Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [x] If a specific issue led to this PR, this PR closes the issue.
- [x] The description of changes is clear and encompassing.
- [x] Any required documentation changes (code and docs) are included in this PR.
- [x] API changes follow the [Runbook for Firecracker API changes][2].
- [x] User-facing changes are mentioned in `CHANGELOG.md`.
- [x] All added/changed functionality is tested.
- [x] New `TODO`s link to an issue.
- [x] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

---

- [x] This functionality can be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #3483, opened 2023-02-28
Define base CPU config types

mattschlebusch:
## Changes

* Common types to be used by Firecracker for configuring guest vCPUs.

## Reason

Common types for CPU templates to be applied by `vmm` crate to configure guest vCPUs.

## License Acceptance

[`CONTRIBUTING.md`][3].

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -3,6 +3,7 @@
 
 //! Enables pre-boot setup, instantiation and booting of a Firecracker VMM.
 
+use std::collections::HashMap;
 #[cfg(target_arch = "x86_64")]
 use std::convert::TryFrom;
 use std::fmt::{Display, Formatter};
@@ -43,6 +44,12 @@ use crate::construct_kvm_mpidrs;
 use crate::device_manager::legacy::PortIODeviceManager;
 use crate::device_manager::mmio::MMIODeviceManager;
 use crate::device_manager::persist::MMIODevManagerConstructorArgs;
+#[cfg(target_arch = "aarch64")]
+use crate::guest_config::aarch64::Aarch64CpuConfiguration;
+#[cfg(target_arch = "x86_64")]
+use crate::guest_config::cpuid::{Cpuid, RawCpuid};
+#[cfg(target_arch = "x86_64")]
+use crate::guest_config::x86_64::X86_64CpuConfiguration;
 use crate::persist::{MicrovmState, MicrovmStateError};
 use crate::resources::VmResources;
 use crate::vmm_config::boot_source::BootConfig;
@@ -833,7 +840,13 @@ pub fn configure_system_for_boot(
                     vmm.guest_memory(),
                     entry_addr,
                     &vcpu_config,
-                    vmm.vm.supported_cpuid().clone(),
+                    X86_64CpuConfiguration {
+                        cpuid: Cpuid::try_from(
+                            RawCpuid::from(vmm.vm.supported_cpuid().clone()).clone(),
+                        )
+                        .unwrap(),
+                        msrs: HashMap::new(),
+                    },
                 )
                 .map_err(Error::VcpuConfigure)
                 .map_err(Internal)?;
@@ -864,7 +877,13 @@ pub fn configure_system_for_boot(
     {
         for vcpu in vcpus.iter_mut() {
             vcpu.kvm_vcpu
-                .configure(vmm.guest_memory(), entry_addr)
+                .configure(
+                    vmm.guest_memory(),
+                    entry_addr,
+                    Aarch64CpuConfiguration {
+                        regs: HashMap::new(),
+                    },
+                )
                 .map_err(Error::VcpuConfigure)
                 .map_err(Internal)?;
         }
--- a/src/vmm/src/guest_config/aarch64/mod.rs
+++ b/src/vmm/src/guest_config/aarch64/mod.rs
@@ -4,6 +4,7 @@
 use std::collections::HashMap;
 
 /// CPU configuration for aarch64 CPUs
+#[derive(Default)]
 pub struct Aarch64CpuConfiguration {
     /// Register values as a key pair
     /// Key: Register pointer
--- a/src/vmm/src/guest_config/aarch64/mod.rs
+++ b/src/vmm/src/guest_config/aarch64/mod.rs
@@ -1,0 +1,12 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::HashMap;
+
+/// CPU configuration for aarch64 CPUs
+pub struct Aarch64CpuConfiguration {
+    /// Register values as a key pair
+    /// Key: Register pointer
+    /// Value: Register value
+    pub regs: HashMap<u64, u128>,
+}
--- a/src/vmm/src/guest_config/mod.rs
+++ b/src/vmm/src/guest_config/mod.rs
@@ -3,3 +3,11 @@
 
 #[cfg(cpuid)]
 pub mod cpuid;
+
+/// Module containing type implementations needed for aarch64 (ARM) CPU configuration
+#[cfg(target_arch = "aarch64")]
+pub mod aarch64;
+
+/// Module containing type implementations needed for x86 CPU configuration
+#[cfg(target_arch = "x86_64")]
+pub mod x86_64;
--- a/src/vmm/src/guest_config/x86_64/mod.rs
+++ b/src/vmm/src/guest_config/x86_64/mod.rs
@@ -1,0 +1,16 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::HashMap;
+
+use crate::guest_config::cpuid::Cpuid;
+
+/// CPU configuration for x86_64 CPUs
+pub struct X86_64CpuConfiguration {
+    /// CPUID configuration
+    pub cpuid: Cpuid,
+    /// Register values as a key pair for model specific registers
+    /// Key: MSR address
+    /// Value: MSR value
+    pub msrs: HashMap<u64, u64>,
+}
--- a/src/vmm/src/vstate/vcpu/aarch64.rs
+++ b/src/vmm/src/vstate/vcpu/aarch64.rs
@@ -15,6 +15,7 @@ use versionize::{VersionMap, Versionize, VersionizeResult};
 use versionize_derive::Versionize;
 use vm_memory::{Address, GuestAddress, GuestMemoryMmap};
 
+use crate::guest_config::aarch64::Aarch64CpuConfiguration;
 use crate::vstate::vcpu::VcpuEmulation;
 use crate::vstate::vm::Vm;
 
@@ -103,7 +104,9 @@ impl KvmVcpu {
         &mut self,
         guest_mem: &GuestMemoryMmap,
         kernel_load_addr: GuestAddress,
+        _cpu_config: Aarch64CpuConfiguration,
     ) -> std::result::Result<(), KvmVcpuConfigureError> {
+        // TODO - Apply CPU config
         arch::aarch64::regs::setup_boot_regs(
             &self.fd,
             self.index,
@@ -233,12 +236,20 @@ mod tests {
         let (_vm, mut vcpu, vm_mem) = setup_vcpu(0x10000);
 
         assert!(vcpu
-            .configure(&vm_mem, GuestAddress(arch::get_kernel_start()),)
+            .configure(
+                &vm_mem,
+                GuestAddress(arch::get_kernel_start()),
+                Aarch64CpuConfiguration::default(),
+            )
             .is_ok());
 
         unsafe { libc::close(vcpu.fd.as_raw_fd()) };
 
-        let err = vcpu.configure(&vm_mem, GuestAddress(arch::get_kernel_start()));
+        let err = vcpu.configure(
+            &vm_mem,
+            GuestAddress(arch::get_kernel_start()),
+            Aarch64CpuConfiguration::default(),
+        );
         assert!(err.is_err());
         assert_eq!(
             err.err().unwrap().to_string(),
@@ -249,7 +260,11 @@ mod tests {
 
         let (_vm, mut vcpu, vm_mem) = setup_vcpu(0x10000);
         unsafe { libc::close(vcpu.fd.as_raw_fd()) };
-        let err = vcpu.configure(&vm_mem, GuestAddress(arch::get_kernel_start()));
+        let err = vcpu.configure(
+            &vm_mem,
+            GuestAddress(arch::get_kernel_start()),
+            Aarch64CpuConfiguration::default(),
+        );
         assert!(err.is_err());
         assert_eq!(
             err.err().unwrap().to_string(),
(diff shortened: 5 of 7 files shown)


src/vmm/src/guest_config/mod.rs
```diff
@@ -1,5 +1,26 @@
...
+/// Bit-mask based wrapper to apply
+/// changes to a given register's value
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct Modifier {
+    /// Mask value is to be applied according
+    /// to what is allowed through by the filter
+    bit_mask_filter: u64,
+    /// Mask value to be applied
+    bit_mask_value: u64,
+}
```
ShadowCurse, line 26: Maybe rename to `RegisterValue`? The name `Modifier` is very generic, I think we need smth associated with registers.
Or we can think of this struct as a generic modifier for any future devices. Then maybe `ValueModifier`?
Also IMO `bit_mask_...` names are a bit too explicit. Just `filter` and `value` should be fine. In addition can you provide the `apply` method as a point of reference of how this masking will work, so I could do the same in my aarch64 templates?

src/vmm/src/guest_config/aarch/mod.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
```
ShadowCurse, line 2: Is there a reason to drop `64` from the names of the modules?

src/vmm/src/guest_config/aarch/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+use crate::guest_config::Modifier;
+
+/// Template providing bitmask modifier info
+/// for CPU register configuration.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
```
ShadowCurse, line 12: super nit: keep `Deserialize` and `Serialize` derives together.

src/vmm/src/guest_config/aarch/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+pub struct ArmCpuTemplate {
+    /// Map of Modifiers for CPU configuration
+    /// Key: Register pointer
+    /// Value: Register's bitmask based modifier
+    pub modifiers: HashMap<u64, Modifier>,
```
ShadowCurse, line 17: Maybe `registers_modifiers`?

src/vmm/src/guest_config/x86/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+/// Template providing bitmask modifier info
+/// for CPUID and MSRs.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct X86CpuTemplate {
+    cpuid_modifiers: HashMap<CpuidRegisterId, Modifier>,
+    /// Map of Modifiers for CPU configuration
+    /// Key: Register pointer
+    /// Value: Register's bitmask based modifier
+    msr_modifiers: HashMap<u64, Modifier>,
+}
```
JonathanWoollett-Light, line 41: Could we simplify `cpuid_modifiers` to `HashMap<CpuidKey, CpuidKey>` where
```rust
struct CpuidKey { leaf: u32, subleaf: u32 }
struct CpuidEntry { eax: u32, ebx: u32, ecx: u32, edx: u32 }
```

src/vmm/src/guest_config/aarch/mod.rs
kalyazin: +1

src/vmm/src/guest_config/mod.rs
```diff
@@ -1,5 +1,26 @@
...
+
+/// Bit-mask based wrapper to apply
+/// changes to a given register's value
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct Modifier {
+    /// Mask value is to be applied according
+    /// to what is allowed through by the filter
+    bit_mask_filter: u64,
+    /// Mask value to be applied
+    bit_mask_value: u64,
```
kalyazin, line 25: this isn't going to work well with arm, is it?

src/vmm/src/guest_config/mod.rs
kalyazin: `filter` or `mask` and `value` look simpler to me.
Having an apply method doesn't look wrong to me, as it will reduce a room for interpretation of how filter/value should be used.

src/vmm/src/guest_config/x86/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+    Ebx,
+    Ecx,
+    Edx,
+}
+
+/// Template providing bitmask modifier info
+/// for CPUID and MSRs.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct X86CpuTemplate {
+    cpuid_modifiers: HashMap<CpuidRegisterId, Modifier>,
```
kalyazin, line 36: Are we not going to suffer from issues similar to https://github.com/firecracker-microvm/firecracker/pull/3445 when using HashMap?

src/vmm/src/guest_config/x86/mod.rs
kalyazin: We shouldn't forget about the flags somehow (which are properties of subleaves).

JonathanWoollett-Light: Made edit to include the flags.

src/vmm/src/guest_config/mod.rs
mattschlebusch: > rename to RegisterValue

It is a modifier of a register's value, there is a difference conceptually within the code base. RegisterBitMask or RegisterModifier would be more accurate.

> filter or mask and value look simpler to me.

If the struct has a clarifying name then I can consider shortening. But the notion of being a bit mask needs to be incorporated into the naming.

src/vmm/src/guest_config/aarch/mod.rs
mattschlebusch: Just that the notion of supporting 32-bit is not relevant in the Firecracker context.

src/vmm/src/guest_config/aarch/mod.rs
mattschlebusch: Formatters are ordering alphabetically. Without automatic formatting here I am not going to make changes in this regard.

src/vmm/src/guest_config/aarch/mod.rs
mattschlebusch: There have been numerous opinions on long versus short naming. I am not going to make further changes in this regard.

src/vmm/src/guest_config/x86/mod.rs
mattschlebusch: > Are we not going to suffer from issues similar to #3445 when using HashMap?

Not quite. The reproducibility is related to ordering of entries in the map which we don't want or need. Although thinking about it a bit more, using a Set here is more desirable. There shouldn't be a need to look up the entries, just iterate through them. I will need to change this.

src/vmm/src/guest_config/aarch/mod.rs
kalyazin: The problem is there is no such thing as just `aarch` (like `arm`, for example). It must be either `aarch64` or `aarch32`, but not just `aarch`, otherwise it becomes rather confusing.

src/vmm/src/guest_config/mod.rs
kalyazin: What I mean is, some ARM registers are u128.

src/vmm/src/guest_config/mod.rs
kalyazin: > If the struct has a clarifying name then I can consider shortening. But the notion of being a bit mask needs to be incorporated into the naming.

If we use `mask`, I think it's rather clear that it is a bit mask.

src/vmm/src/guest_config/x86/mod.rs
kalyazin: Why is a set better than a vector here?

src/vmm/src/guest_config/aarch/mod.rs
ShadowCurse: Also we already have aarch64 naming in use

src/vmm/src/guest_config/x86/mod.rs
JonathanWoollett-Light: To apply the registers without this I will need to find all the elements in the hash map then iterate through them, then map on the register field. I think this helps reduce duplicating the leaf & subleaf key and bundles the data together as it is conceptually.

src/vmm/src/guest_config/mod.rs
ShadowCurse: > In short, not here no. How the CPU sub-modules make use of the data is discretionary.


If modules can use data however they want, then there is no consistency. It can be that x86 registers configured one way and aarch64 another. Isn't this the opposite of what we want? Also it should be generic over u32, u64,... because registers can have different sizes.

src/vmm/src/guest_config/aarch64/mod.rs
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::HashMap;
+
+/// CPU configuration for aarch64 CPUs
+pub struct Aarch64CpuConfiguration {
+    /// Register values as a key pair
+    /// Key: Register pointer
+    /// Value: Register value
+    pub regs: HashMap<u64, u128>,
```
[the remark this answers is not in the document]
mattschlebusch: There will be eventually. To make use of the template using bit masks to modify values. the host CPU configuration will need to be looked up so that the template can be applied to something. The process of applying the masks will be keyed as well so they can look up the register in this configuration map directly.

src/vmm/src/guest_config/x86/mod.rs
mattschlebusch: This is out of date, but In case this becomes relevant again, instead of using a map. The HashSet enforces that there are no duplicates in the collection. Not a big deal though, this will be validated against, but it is my preference to have defensive structures in place.

src/vmm/src/guest_config/aarch64/mod.rs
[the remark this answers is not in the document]
kalyazin: Ok, that makes sense indeed.

src/vmm/src/guest_config/x86_64/mod.rs
```diff
@@ -0,0 +1,16 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::HashMap;
+
+use crate::guest_config::cpuid::Cpuid;
+
+/// CPU configuration for x86_64 CPUs
+pub struct X86_64CpuConfiguration {
+    /// CPUID configuration
```
ShadowCurse, line 10: If this is a X86 cpu config struct then where is `smt`?

src/vmm/src/guest_config/mod.rs
```diff
@@ -3,3 +3,11 @@
```
ShadowCurse, line 3: Why did you drop the `RegisterModifier` struct? Shouldn't it be somewhere here?

mattschlebusch: It will be. But in the context of the PR, this is only after the configuration has been created. The modifiers are relevant as a part of the template to create the configuration. This PR is just to focusing on the structure of the configuration.

src/vmm/src/builder.rs
```diff
@@ -833,7 +840,13 @@ pub fn configure_system_for_boot(
...
                     entry_addr,
                     &vcpu_config,
-                    vmm.vm.supported_cpuid().clone(),
+                    X86_64CpuConfiguration {
+                        cpuid: Cpuid::try_from(
+                            RawCpuid::from(vmm.vm.supported_cpuid().clone()).clone(),
+                        )
+                        .unwrap(),
+                        msrs: HashMap::new(),
+                    },
```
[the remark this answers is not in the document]
mattschlebusch: That was my intention originally, however without creating polymorphic types to accommodate a hierarchy, this would require having an attribute for each architecture configured (Vcpu_Config is a shared type for x86_64 and aarch64).

src/vmm/src/guest_config/x86_64/mod.rs
ShadowCurse: So for the user to enable smt he would need to use cpu template? Well, maybe we can have a static template that does that, but seems inconvenient to me.

zulinx86: `VcpuConfig` has `smt` field.
https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/vstate/vcpu/mod.rs#L77

I'm not sure what if `VcpuConfig` and CPU template conflicts on SMT.

ShadowCurse (approved):

kalyazin (approved):

src/vmm/src/builder.rs
```diff
@@ -833,7 +840,13 @@ pub fn configure_system_for_boot(
                     vmm.guest_memory(),
                     entry_addr,
                     &vcpu_config,
-                    vmm.vm.supported_cpuid().clone(),
+                    X86_64CpuConfiguration {
+                        cpuid: Cpuid::try_from(
+                            RawCpuid::from(vmm.vm.supported_cpuid().clone()).clone(),
```
zulinx86, line 845: nit: unnecessary clone()?
```suggestion
                            RawCpuid::from(vmm.vm.supported_cpuid().clone()),
```

src/vmm/src/guest_config/x86_64/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+    pub cpuid: Cpuid,
+    /// Register values as a key pair for model specific registers
+    /// Key: MSR address
+    /// Value: MSR value
+    pub msrs: HashMap<u64, u64>,
```
zulinx86, line 15: The input address space of MSR 32-bit wide.

src/vmm/src/guest_config/x86_64/mod.rs
mattschlebusch: @zulinx86 Yep this is an important point. For that particular case, the custom template will override.

src/vmm/src/builder.rs
mattschlebusch: @zulinx86 You are right. I will fix in [PR/3511](https://github.com/firecracker-microvm/firecracker/pull/3511/).

src/vmm/src/guest_config/x86/mod.rs
mattschlebusch: A similar approach to what you propose has been taken in [PR/3511](https://github.com/firecracker-microvm/firecracker/pull/3511/) for resolving duplication. Iteration won't be necessary though as the keys for the map will be addresses provided by the template.

src/vmm/src/guest_config/x86_64/mod.rs
mattschlebusch: You are correct. For the feature branch this isn't a blocker, this can be fixed downstream.

PR #3484, opened 2023-02-28
docs: Update `getting-started.md`

JonathanWoollett-Light:
## Changes

Updates `getting-started.md`.

The example now involves setting up a microVM you can SSH into.

See https://github.com/JonathanWoollett-Light/firecracker/blob/getting-started/docs/getting-started.md.

## Reason

Some of the instructions are out-dated and it is missing a simple guide to ssh into a microVM.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -1,394 +1,42 @@
 # Getting Started with Firecracker
 
-## Contents
-
-- [Prerequisites](#prerequisites)
-- [Getting the Firecracker Binary](#getting-the-firecracker-binary)
-- [Running Firecracker](#running-firecracker)
-- [Building From Source](#building-from-source)
-- [Running the Integration Test Suite](#running-the-integration-test-suite)
-- [Appendix A: Setting Up KVM Access](#appendix-a-setting-up-kvm-access)
-- [Appendix B: Setting Up Docker](#appendix-b-setting-up-docker)
-
 ## Prerequisites
 
-If you need an opinionated way of running Firecracker, launch an `i3.metal`
-instance using Ubuntu 18.04 on EC2. Firecracker uses
-[KVM](https://www.linux-kvm.org) and needs read/write access that can be
-granted as shown below:
+You can check if your system meets the requirements by running
+`firecracker/tools/devtool checkenv`.
 
-```console
-sudo setfacl -m u:${USER}:rw /dev/kvm
-```
+An opinionated way to run Firecracker is to launch an
+[EC2](https://aws.amazon.com/ec2/) `i3.metal` instance with Ubuntu 18.04.
 
-The generic requirements are explained below:
+Nested virtualization is not supported on non-metal EC2 instances. This is why
+`.metal` instances are used.
 
-- **Linux**
+### Architecture & OS
 
-  Firecracker currently supports physical Linux **x86_64** and **aarch64**
-  hosts. The currently supported kernel versions can be found [here](kernel-policy.md).
+Firecracker supports **x86_64** and **aarch64** Linux, see
+[specific supported kernels](kernel-policy.md).
 
-- **KVM**
+### KVM
 
-  Please make sure that:
-    1. you have KVM enabled in your Linux kernel, and
-    2. you have read/write access to `/dev/kvm`.
-       If you need help setting up access to `/dev/kvm`, you should check out
-       [Appendix A](#appendix-a-setting-up-kvm-access).
+Firecracker requires [the KVM Linux kernel module](https://www.linux-kvm.org/).
 
-To check if your system meets the requirements to run Firecracker, clone
-the repository and execute `tools/devtool checkenv`.
-
-## Getting the Firecracker Binary
-
-Firecracker is linked statically against
-[musl](https://musl.libc.org/), having no library dependencies. You can
-just download the latest binary from our
-[release page](https://github.com/firecracker-microvm/firecracker/releases),
-and run it on your x86_64 or aarch64 Linux machine.
-
-On the EC2 instance, this binary can be downloaded as:
-
-```wrap
-release_url="https://github.com/firecracker-microvm/firecracker/releases"
-latest=$(basename $(curl -fsSLI -o /dev/null -w  %{url_effective} ${release_url}/latest))
-arch=`uname -m`
-curl -L ${release_url}/download/${latest}/firecracker-${latest}-${arch}.tgz \
-| tar -xz
-```
-
-Rename the binary to "firecracker":
-
-```console
-mv release-${latest}-$(uname -m)/firecracker-${latest}-$(uname -m) firecracker
-```
-
-If, instead, you'd like to build Firecracker yourself, you should check out
-the [Building From Source section](#building-from-source) in this doc.
-
-## Running Firecracker
-
-In production, Firecracker is designed to be run securely, inside
-an execution jail, carefully set up by the `jailer` binary. This is how
-our
-[integration test suite](#running-the-integration-test-suite) does it.
-However, if you just want to see Firecracker booting up a guest Linux
-machine, you can do that as well.
-
-First, make sure you have the `firecracker` binary available - either
-[downloaded from our release page](#getting-the-firecracker-binary), or
-[built from source](#building-from-source).
-
-Next, you will need an uncompressed Linux kernel binary, and an ext4
-file system image (to use as rootfs).
-
-1. To run an `x86_64` guest you can download such resources from:
-   [kernel](https://s3.amazonaws.com/spec.ccfc.min/img/quickstart_guide/x86_64/kernels/vmlinux.bin)
-   and [rootfs](https://s3.amazonaws.com/spec.ccfc.min/img/quickstart_guide/x86_64/rootfs/bionic.rootfs.ext4).
-1. To run an `aarch64` guest, download them from:
-   [kernel](https://s3.amazonaws.com/spec.ccfc.min/img/quickstart_guide/aarch64/kernels/vmlinux.bin)
-   and [rootfs](https://s3.amazonaws.com/spec.ccfc.min/img/quickstart_guide/aarch64/rootfs/bionic.rootfs.ext4).
-
-Now, let's open up two shell prompts: one to run Firecracker, and another one
-to control it (by writing to the API socket). For the purpose of this guide,
-**make sure the two shells run in the same directory where you placed the
-`firecracker` binary**.
-
-In your **first shell**:
-
-- make sure Firecracker can create its API socket:
+The presence of the KVM module can be checked with:
 
 ```bash
-rm -f /tmp/firecracker.socket
+lsmod | grep kvm
 ```
 
-- then, start Firecracker:
+An example output where it is enabled:
 
 ```bash
-./firecracker --api-sock /tmp/firecracker.socket
+kvm_intel             348160  0
+kvm                   970752  1 kvm_intel
+irqbypass              16384  1 kvm
 ```
-
-In your **second shell** prompt:
-
-- get the kernel and rootfs, if you don't have any available:
-
-  ```bash
-  arch=`uname -m`
-  dest_kernel="hello-vmlinux.bin"
-  dest_rootfs="hello-rootfs.ext4"
-  image_bucket_url="https://s3.amazonaws.com/spec.ccfc.min/img/quickstart_guide/$arch"
-
-  if [ ${arch} = "x86_64" ]; then
-      kernel="${image_bucket_url}/kernels/vmlinux.bin"
-      rootfs="${image_bucket_url}/rootfs/bionic.rootfs.ext4"
-  elif [ ${arch} = "aarch64" ]; then
-      kernel="${image_bucket_url}/kernels/vmlinux.bin"
-      rootfs="${image_bucket_url}/rootfs/bionic.rootfs.ext4"
-  else
-      echo "Cannot run firecracker on $arch architecture!"
-      exit 1
-  fi
-
-  echo "Downloading $kernel..."
-  curl -fsSL -o $dest_kernel $kernel
-
-  echo "Downloading $rootfs..."
-  curl -fsSL -o $dest_rootfs $rootfs
-
-  echo "Saved kernel file to $dest_kernel and root block device to $dest_rootfs."
-  ```
-
-- set the guest kernel (assuming you are in the same directory as the
-  above script was run):
-
-  ```bash
-  arch=`uname -m`
-  kernel_path=$(pwd)"/hello-vmlinux.bin"
-
-  if [ ${arch} = "x86_64" ]; then
-      curl --unix-socket /tmp/firecracker.socket -i \
-        -X PUT 'http://localhost/boot-source'   \
-        -H 'Accept: application/json'           \
-        -H 'Content-Type: application/json'     \
-        -d "{
-              \"kernel_image_path\": \"${kernel_path}\",
-              \"boot_args\": \"console=ttyS0 reboot=k panic=1 pci=off\"
-         }"
-  elif [ ${arch} = "aarch64" ]; then
-      curl --unix-socket /tmp/firecracker.socket -i \
-        -X PUT 'http://localhost/boot-source'   \
-        -H 'Accept: application/json'           \
-        -H 'Content-Type: application/json'     \
-        -d "{
-              \"kernel_image_path\": \"${kernel_path}\",
-              \"boot_args\": \"keep_bootcon console=ttyS0 reboot=k panic=1 pci=off\"
-         }"
-  else
-      echo "Cannot run firecracker on $arch architecture!"
-      exit 1
-  fi
-  ```
-
-- set the guest rootfs:
-
-  ```bash
-  rootfs_path=$(pwd)"/hello-rootfs.ext4"
-  curl --unix-socket /tmp/firecracker.socket -i \
-    -X PUT 'http://localhost/drives/rootfs' \
-    -H 'Accept: application/json'           \
-    -H 'Content-Type: application/json'     \
-    -d "{
-          \"drive_id\": \"rootfs\",
-          \"path_on_host\": \"${rootfs_path}\",
-          \"is_root_device\": true,
-          \"is_read_only\": false
-     }"
-  ```
-
-- start the guest machine:
-
-  ```bash
-  curl --unix-socket /tmp/firecracker.socket -i \
-    -X PUT 'http://localhost/actions'       \
-    -H  'Accept: application/json'          \
-    -H  'Content-Type: application/json'    \
-    -d '{
-        "action_type": "InstanceStart"
-     }'
-  ```
-
-Going back to your first shell, you should now see a serial TTY prompting you
-to log into the guest machine. If you used our `hello-rootfs.ext4` image,
-you can login as `root`, using the password `root`.
-
-When you're done, issuing a `reboot` command inside the guest will actually
-shutdown Firecracker gracefully. This is due to the fact that Firecracker
-doesn't implement guest power management.
-
-**Note**: the default microVM will have 1 vCPU and 128 MiB RAM. If you wish to
-customize that (say, 2 vCPUs and 1024MiB RAM), you can do so before issuing
-the `InstanceStart` call, via this API command:
-
-```bash
-curl --unix-socket /tmp/firecracker.socket -i  \
-  -X PUT 'http://localhost/machine-config' \
-  -H 'Accept: application/json'            \
-  -H 'Content-Type: application/json'      \
-  -d '{
-      "vcpu_count": 2,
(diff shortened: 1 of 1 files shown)


farazfazli:
Really appreciate the effort here in making the getting started guide easier to follow.

docs/getting-started.md
```diff
@@ -1,372 +1,273 @@
...
-- **Linux**
-
-  Firecracker currently supports physical Linux **x86_64** and **aarch64**
-  hosts. The currently supported kernel versions can be found [here](kernel-policy.md).
+You can check if your system meets the requirements by running
+`firecracker/tools/devtool checkenv`.
 
-- **KVM**
+An opinionated way to run Firecracker is to launch an
+[EC2](https://aws.amazon.com/ec2/) `i3.metal` instance with Ubuntu 18.04.
```
roypat, line 9: nit: maybe we should recommend ubuntu 22, since that's what our build container uses? (also why is this "opinionated"? :D)

docs/getting-started.md
```diff
@@ -1,372 +1,273 @@
...
 
-```bash
-git checkout tags/v0.10.1
-```
+# Download a rootfs
+wget https://s3.amazonaws.com/spec.ccfc.min/ci-artifacts/disks/${ARCH}/ubuntu-18.04.ext4
 
-### Building Firecracker
+# Download the ssh key for the rootfs
+wget https://s3.amazonaws.com/spec.ccfc.min/ci-artifacts/disks/${ARCH}/ubuntu-18.04.id_rsa
```
roypat, line 76: ah, didnt know this was something that needed to be done. explains all my difficulties ssh-ing into my microvms 😂

docs/getting-started.md
```diff
@@ -1,372 +1,273 @@
...
+# Set network interface
+curl -X PUT --unix-socket "${API_SOCKET}" \
+    --data "{
+        \"iface_id\": \"net1\",
+        \"guest_mac\": \"$FC_MAC\",
+        \"host_dev_name\": \"$TAP_DEV\"
+    }" \
+    "http://localhost/network-interfaces/net1"
+
+sleep 0.015s
```
roypat, line 191: mh, instead of sleeping, can we give a bit more context here? Something along the lines of "Wait until you see in your first terminal that the microvm has started. Once it has done so, you can SSH into it using the key downloaded above, e.g [...]`?

docs/getting-started.md
```diff
@@ -1,372 +1,273 @@
...
-to log into the guest machine. If you used our `hello-rootfs.ext4` image,
-you can login as `root`, using the password `root`.
-
-When you're done, issuing a `reboot` command inside the guest will actually
-shutdown Firecracker gracefully. This is due to the fact that Firecracker
-doesn't implement guest power management.
-
-**Note**: the default microVM will have 1 vCPU and 128 MiB RAM. If you wish to
-customize that (say, 2 vCPUs and 1024MiB RAM), you can do so before issuing
-the `InstanceStart` call, via this API command:
```
roypat, line 210: Are these defaults still mentioned anywhere? I think they're valuable information here

docs/getting-started.md
roypat: Maybe briefly mention that we use `.metal` because we don't support nested virtualization?

roypat (commented):
Looks good, as someone who's struggled through this guide multiple times these seem like improvements :D

docs/getting-started.md
```diff
@@ -383,77 +284,3 @@ Points to check to confirm the API socket is running and accessible:
...
-or run Firecracker as `root` (via `sudo`).
-
-You can check if you have access to `/dev/kvm` with:
-
-```bash
-[ -r /dev/kvm ] && [ -w /dev/kvm ] && echo "OK" || echo "FAIL"
-```
-
-**Note:** If you've just added your user to the `kvm` group via `usermod`, don't
-forget to log out and then back in, so this change takes effect.
```
zulinx86, line 415: Why is this part removed?

docs/getting-started.md
```diff
@@ -383,77 +284,3 @@ Points to check to confirm the API socket is running and accessible:
...
-
-## Appendix B: Setting Up Docker
```
zulinx86, line 417: We might need to discuss if we should remove this section.

zulinx86 (commented):
I'd prefer to preserve both options "build from source" and "download latest release binaries".

docs/getting-started.md
JonathanWoollett-Light: This relates to standard Linux permissions, not specific to Firecracker. Although I don't have a strong view against keeping it.

docs/getting-started.md
JonathanWoollett-Light: Added a following comment:

> Firecracker does not support nested virtualization. This is why `.metal`
> instances are used.

docs/getting-started.md
JonathanWoollett-Light: Added comments on the `sleep`s:

> # API requests are handled asynchronously, it is important the configuration is
> # set, before `InstanceStart`.

and

> # API requests are handled asynchronously, it is important the microVM has been
> # started before we attempt to SSH into it.

respectively.

docs/getting-started.md
JonathanWoollett-Light: I don't think there is value in documenting the docker install process here.

roypat (approved):

docs/getting-started.md
```diff
@@ -1,372 +1,282 @@
...
-curl -L ${release_url}/download/${latest}/firecracker-${latest}-${arch}.tgz \
-| tar -xz
-```
-
-Rename the binary to "firecracker":
-
-```console
-mv release-${latest}-$(uname -m)/firecracker-${latest}-$(uname -m) firecracker
+```bash
+lsmod | grep kvm
```
kalyazin, line 26: what if kvm is built in the kernel, not compiled as a module (I don't know how common is that)?

docs/getting-started.md
```diff
@@ -1,372 +1,282 @@
...
+    "http://localhost/logger"
+
+# Set machine config
+#
+# Without this the microVM will have the default values of 1 vCPU and 128MB of RAM.
+curl -X PUT --unix-socket "${API_SOCKET}" \
+    --data "{
+        \"vcpu_count\": 2,
+        \"mem_size_mib\": 128,
+        \"cpu_template\": \"C3\"
```
kalyazin, line 157: do we want our users to jump straight into a template?

JonathanWoollett-Light: Good point, do you think this should just be removed?

kalyazin: I think it's an advanced feature, I wouldn't encourage using it by default.

JonathanWoollett-Light: Okay, I've removed this API call.

docs/getting-started.md
JonathanWoollett-Light: I'm not aware of this case, do you have any documentation of this you can link?

kalyazin: I think, checking for the device file is ok (already mentioned below).
Other ways are hacky (like grepping `/proc/kallsyms`).
Maybe leave the `lsmod` check as is, but also add a note that kvm may be compiled in and still be used, but not visible via `lsmod`?

kalyazin: Looks good to me.
```
config KVM
        tristate "Kernel-based Virtual Machine (KVM) support"
```
Tristate usually means enabled, disabled or module.

docs/getting-started.md
zulinx86: To my understanding, the reason why we use `.metal` instances is that firecracker requires KVM access and non-metal instances do not support nested virtualization.

JonathanWoollett-Light: Updated comment to:

> Nested virtualization is not supported on non-metal EC2 instances. This is why
> `.metal` instances are used.

docs/getting-started.md
```diff
@@ -1,372 +1,271 @@
...
+sudo sh -c "echo 1 > /proc/sys/net/ipv4/ip_forward"
+sudo iptables -t nat -D POSTROUTING -o eth0 -j MASQUERADE || true
+sudo iptables -D FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT \
+    || true
+sudo iptables -D FORWARD -i tap0 -o eth0 -j ACCEPT || true
+
+# Enable microVM internert access
+sudo iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+sudo iptables -I FORWARD 1 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
+sudo iptables -I FORWARD 1 -i tap0 -o eth0 -j ACCEPT
```
zulinx86, line 132: The first three `iptables` command for deleting existing rules.
```suggestion
# Enable ip forwarding
sudo sh -c "echo 1 > /proc/sys/net/ipv4/ip_forward"

# Set up microVM internet access
sudo iptables -t nat -D POSTROUTING -o eth0 -j MASQUERADE || true
sudo iptables -D FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT \
    || true
sudo iptables -D FORWARD -i tap0 -o eth0 -j ACCEPT || true
sudo iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
sudo iptables -I FORWARD 1 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
sudo iptables -I FORWARD 1 -i tap0 -o eth0 -j ACCEPT
```

docs/getting-started.md
```diff
@@ -1,372 +1,271 @@
...
+KERNEL_BOOT_ARGS="console=ttyS0 reboot=k panic=1 ipv6.disable=1 i8042.nokbd
+ 8250.nr_uarts=1 random.trust_cpu=on"
+
+# Set boot source
+curl -X PUT --unix-socket "${API_SOCKET}" \
+    --data "{
+        \"kernel_image_path\": \"${KERNEL}\",
+        \"boot_args\": \"${KERNEL_BOOT_ARGS}\"
+    }" \
+    "http://localhost/boot-source"
```
zulinx86, line 160: Trying on AL2, I got this error.
```
$ curl -X PUT --unix-socket "${API_SOCKET}" \
>     --data "{
>         \"kernel_image_path\": \"${KERNEL}\",
>         \"boot_args\": \"${KERNEL_BOOT_ARGS}\"
>     }" \
>     "http://localhost/boot-source"
{"fault_message":"An error occurred when deserializing the json body of a request: control character (\\u0000-\\u001F) found while parsing a string at line 4 column 0."}
```

docs/getting-started.md
JonathanWoollett-Light: Fixed.

docs/getting-started.md
zulinx86: I could fix this by changing:

before
```
KERNEL_BOOT_ARGS="console=ttyS0 reboot=k panic=1 ipv6.disable=1 i8042.nokbd
 8250.nr_uarts=1 random.trust_cpu=on"
```

after
```
KERNEL_BOOT_ARGS="console=ttyS0 reboot=k panic=1 ipv6.disable=1 i8042.nokbd 8250.nr_uarts=1 random.trust_cpu=on"
```

JonathanWoollett-Light: Does the previous API request of :
```bash
curl -X PUT --unix-socket "${API_SOCKET}" \
    --data "{
        \"log_path\": \"${LOGFILE}\",
        \"level\": \"Debug\",
        \"show_level\": true,
        \"show_log_origin\": true
    }" \
    "http://localhost/logger"
```
succeed?

What does `echo $KERNEL` and `echo $KERNEL_BOOT_ARGS` give?

docs/getting-started.md
```diff
@@ -1,372 +1,271 @@
...
+
+# API requests are handled asynchronously, it is important the microVM has been
+# started before we attempt to SSH into it.
+sleep 0.015s
+
+# SSH into the microVM
+sudo ssh -i ./ubuntu-18.04.id_rsa 172.16.0.2
+
+# Use `root` for both the login and password.
+# Run `reboot` to exit.
```
zulinx86, line 204: Although this includes instructions for net configuration on host, but does not include instructions for net configuration inside guests. How about adding it or redirect to https://github.com/firecracker-microvm/firecracker/blob/main/docs/network-setup.md ?

docs/getting-started.md
zulinx86: > Does the previous API request of:
> 
> ```shell
> curl -X PUT --unix-socket "${API_SOCKET}" \
>     --data "{
>         \"log_path\": \"${LOGFILE}\",
>         \"level\": \"Debug\",
>         \"show_level\": true,
>         \"show_log_origin\": true
>     }" \
>     "http://localhost/logger"
> ```
> 
> succeed?
> 
> What does `echo $KERNEL` and `echo $KERNEL_BOOT_ARGS` give?


Yes.
Here is the log.
```
[ec2-user@ip-172-31-11-151 ~]$ curl -X PUT --unix-socket "${API_SOCKET}" \
>     --data "{
>         \"log_path\": \"${LOGFILE}\",
>         \"level\": \"Debug\",
>         \"show_level\": true,
>         \"show_log_origin\": true
>     }" \
>     "http://localhost/logger"
[ec2-user@ip-172-31-11-151 ~]$ KERNEL="./vmlinux.bin"
[ec2-user@ip-172-31-11-151 ~]$ KERNEL_BOOT_ARGS="console=ttyS0 reboot=k panic=1 ipv6.disable=1 i8042.nokbd
>  8250.nr_uarts=1 random.trust_cpu=on"
[ec2-user@ip-172-31-11-151 ~]$ curl -X PUT --unix-socket "${API_SOCKET}" \
>     --data "{
>         \"kernel_image_path\": \"${KERNEL}\",
>         \"boot_args\": \"${KERNEL_BOOT_ARGS}\"
>     }" \
>     "http://localhost/boot-source"
{"fault_message":"An error occurred when deserializing the json body of a request: control character (\\u0000-\\u001F) found while parsing a string at line 4 column 0."}
[ec2-user@ip-172-31-11-151 ~]$ echo $KERNEL
./vmlinux.bin
[ec2-user@ip-172-31-11-151 ~]$ echo $KERNEL_BOOT_ARGS
console=ttyS0 reboot=k panic=1 ipv6.disable=1 i8042.nokbd 8250.nr_uarts=1 random.trust_cpu=on
```

zulinx86: I'm suspecting that `KERNEL_BOOT_ARGS` includes invisible control character...

zulinx86: Or something happens when copying and pasting it. Any way, removing the new line works on my env.
```
[ec2-user@ip-172-31-11-151 ~]$ KERNEL_BOOT_ARGS="console=ttyS0 reboot=k panic=1 ipv6.disable=1 i8042.nokbd 8250.nr_uarts=1 random.trust_cpu=on"
[ec2-user@ip-172-31-11-151 ~]$ curl -X PUT --unix-socket "${API_SOCKET}"     --data "{
        \"kernel_image_path\": \"${KERNEL}\",
        \"boot_args\": \"${KERNEL_BOOT_ARGS}\"
    }"     "http://localhost/boot-source"
[ec2-user@ip-172-31-11-151 ~]$
```

docs/getting-started.md
zulinx86: Ah sorry, this seems to be caused by the conflict of ip address range for host and guest.
```
$ ip route
default via 172.31.0.1 dev eth0
169.254.169.254 dev eth0
172.16.0.0/30 dev tap0 proto kernel scope link src 172.16.0.1
172.17.0.0/16 dev docker0 proto kernel scope link src 172.17.0.1 linkdown
172.31.0.0/20 dev eth0 proto kernel scope link src 172.31.11.151
```

docs/getting-started.md
```diff
@@ -1,372 +1,270 @@
...
+sudo ip addr add "${TAP_IP}${MASK_SHORT}" dev "$TAP_DEV"
+sudo ip link set dev "$TAP_DEV" up
+
+# Enable ip forwarding
+sudo sh -c "echo 1 > /proc/sys/net/ipv4/ip_forward"
+
+# Set up microVM internet access
+sudo iptables -t nat -D POSTROUTING -o eth0 -j MASQUERADE || true
+sudo iptables -D FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT \
+    || true
```
zulinx86, line 128: nit: remove the new line as well?

JonathanWoollett-Light: The newline is needed to conform to [`mdl`](https://github.com/markdownlint/markdownlint) column max width.

docs/getting-started.md
zulinx86: The guest decides its ip address based on MAC address with [fcnet-setup.sh](https://github.com/firecracker-microvm/firecracker/blob/main/resources/tests/fcnet-setup.sh).

Like my case, in some customers' environment, address range conflicts can happen.
To make `TAP_IP` modification acceptable, it would be nice that `FC_MAC` is derived from `TAP_IP` by doing reverse calculation of `fcnet-setup.sh`.

zulinx86: Here is an example.
```
FC_MAC=$(printf "06:00:%02x:%02x:%02x:%02x" $(echo $TAP_IP | sed "s/\./ /g"))
```

JonathanWoollett-Light: Would replacing `FC_MAC` with:
```bash
# Derive FC_MAC from TAP_IP e.g. `172.16.0.1` becomes `06:00:AC:10:00:02`.
FC_MAC=$(printf "06:00:%02x:%02x:%02x:%02x" $(echo $TAP_IP | sed "s/\./ /g"))
```
fit this?

JonathanWoollett-Light: Included comment on this:
```bash
# The IP address of a guest is derived from its MAC address with
# `fcnet-setup.sh`, this has been pre-configured in the guest rootfs. It is
# important that `TAP_IP` and `FC_MAC` match this.
FC_MAC="06:00:AC:10:00:02"
```

docs/getting-started.md
zulinx86: talked with Jonathan and concluded to simpify it.

docs/getting-started.md
```diff
@@ -1,372 +1,276 @@
...
+        \"show_level\": true,
+        \"show_log_origin\": true
+    }" \
+    "http://localhost/logger"
+
+KERNEL="./vmlinux.bin"
+KERNEL_BOOT_ARGS="console=ttyS0 reboot=k panic=1 pci=off"
+
+if [ ${ARCH} = "aarch64" ]; then
+    KERNEL_BOOT_ARGS="keep_bootcon ${KERNEL_BOOT_ARGS}"
```
zulinx86, line 154: In this terminal, `$ARCH` is not set.

zulinx86:
```suggestion
ARCH=$(uname -m)
if [ ${ARCH} == "aarch64" ]; then
    KERNEL_BOOT_ARGS="keep_bootcon ${KERNEL_BOOT_ARGS}"
fi
```

JonathanWoollett-Light: Fixed.

docs/getting-started.md
```diff
@@ -1,372 +1,279 @@
...
+
+# API requests are handled asynchronously, it is important the microVM has been
+# started before we attempt to SSH into it.
+sleep 0.015s
+
+# SSH into the microVM
+sudo ssh -i ./ubuntu-18.04.id_rsa 172.16.0.2
+
+# Use `root` for both the login and password.
+# Run `reboot` to exit.
```
zulinx86, line 212: To get network access from guest, it requires to set up default gateway.
https://github.com/firecracker-microvm/firecracker/blob/main/docs/network-setup.md?plain=1#L89

zulinx86: And it would be better to have DNS setup as well.

docs/getting-started.md
zulinx86: I think it needs `==`.

JonathanWoollett-Light: It is `if [ ${arch} = "x86_64" ]; then` in the previous getting started documentation, and in bash comparison is typically `=` https://tldp.org/LDP/abs/html/comparison-ops.html. Could you maybe explain more why it would be `==` here?

zulinx86: You're right. My bad.

zulinx86 (approved):

docs/getting-started.md
zulinx86: Talking with Jonathan offline, the connectivity between guests and the internet can be considered as optional and can be handled in a follow-up PR. Resolving this conversation.

kalyazin (approved):

PR #3486, opened 2023-03-01
chore: enhance cgroupv2 checking

andreitraistaru:
## Reason

On Ubuntu 22 we have cgroupv2 enabled by default. Thus, we no longer need to add the `cgroup_no_v1=all` flag in the cmdline to enforce cgroupv2. This resulted in a wrong detection of cgroup version by just looking into the cmdline. According to docs, existence of /sys/fs/cgroup/cgroup.controllers should be a marker of cgroupsv2 enabled.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/security/test_jail.py
+++ b/tests/integration_tests/security/test_jail.py
@@ -218,11 +218,10 @@ def test_arbitrary_usocket_location(test_microvm_with_initrd):
 
 
 @functools.lru_cache(maxsize=None)
-def cgroup_v1_available():
-    """Check if cgroup-v1 is disabled on the system."""
-    with open("/proc/cmdline", encoding="utf-8") as cmdline_file:
-        cmdline = cmdline_file.readline()
-        return bool("cgroup_no_v1=all" not in cmdline)
+def cgroup_v2_available():
+    """Check if cgroup-v2 is enabled on the system."""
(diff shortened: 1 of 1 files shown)


tests/integration_tests/security/test_jail.py
```diff
@@ -234,7 +233,7 @@ def sys_setup_cgroups():
     This set-up is important to do when running from inside a Docker
     container while the system is using cgroup-v2.
     """
-    cgroup_version = 1 if cgroup_v1_available() else 2
+    cgroup_version = 2 if cgroup_v2_available() else 1
```
pb8o, line 236: Doesn't this flip the default to cgroupsv2 if both are available? Though IMO we should definitely prefer v2 if both are available.

andreitraistaru: Good point. However, I'm not quite sure that the behavior is flipped regarding what version we prefer because I don't think we were in the situation to have both versions enabled in parallel (correct me if I'm wrong since I don't have to much experience with this). Up until now, we had to provide the `cgroup_no_v1=all` flag to the cmdline in order to enable cgroupv2 (which keeps v1 disabled) so I don't think we were in the situation to have both versions enabled in parallel. This means that when we had v2 enabled we simply used it which is what should happen after this PR.

pb8o (approved):

luminitavoicu (approved):

zulinx86 (approved):

PR #3488, opened 2023-03-02
UFFD: Add sanity checks for information received through UDS

luminitavoicu:
## Changes

Add sanity checks for information received from Firecracker through UDS.

## Reason

Ensure the mappings received through UDS are not corrupted.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/uffd_handler/src/main.rs
+++ b/src/uffd_handler/src/main.rs
@@ -18,7 +18,7 @@ use userfaultfd::Uffd;
 
 use crate::common::{parse_unix_stream, StreamError};
 use crate::handler::{HandlerError, PageFaultHandler, UffdManager};
-use crate::memory_region::{create_mem_regions, deserialize_mappings};
+use crate::memory_region::{mem_regions_from_stream, MemRegionError};
 
 const EXIT_CODE_ERROR: i32 = 1;
 
@@ -30,11 +30,8 @@ pub enum Error {
     ArgumentParsing(String),
     #[error("Failed to bind userfaultfd socket path: {0}")]
     BindSocket(io::Error),
-    #[error(
-        "Size of snapshot memory file differs from the size of memory mappings.  Mappings \
-         received through uffd socket might be corrupted."
-    )]
-    CorruptedMemoryMappings,
+    #[error("Memory mappings received through uffd socket might be corrupted: {0}")]
+    CorruptedMemoryMappings(MemRegionError),
     #[error("Deserializing guest memory mappings failed: {0}")]
     DeserializeMemoryMappings(serde_json::Error),
     #[error("Snapshot memory file is empty.")]
@@ -77,14 +74,14 @@ where
         .nth(2)
         .ok_or_else(|| Error::ArgumentParsing(String::from("No memory file provided")))?;
 
-    let (file, size) = validate_snapshot_file(&mem_file_path)?;
+    let (file, mem_file_size) = validate_snapshot_file(&mem_file_path)?;
 
     // mmap() a memory area used to bring in the faulting regions.
     // SAFETY: Safe because the parameters are valid.
     let memfile_buffer = unsafe {
         libc::mmap(
             ptr::null_mut(),
-            size,
+            mem_file_size,
             libc::PROT_READ,
             libc::MAP_PRIVATE,
             file.as_raw_fd(),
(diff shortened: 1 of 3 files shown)


src/uffd_handler/src/memory_region.rs
```diff
@@ -22,6 +34,15 @@ pub enum MemPageState {
     Anonymous,
 }
 
+/// Construct memory regions from stream received from Firecracker process.
+pub fn mem_regions_from_stream(msg: &str, mem_file_size: usize) -> Result<Vec<MemRegion>, Error> {
+    let mappings = deserialize_mappings(msg)?;
+    // Perform sanity checks on memory mappings to ensure they are not corrupted.
```
andreitraistaru, line 40: super small nit (personal touch): looks kind of self explanatory, but it's not necessarily something that I would remove.

Also, `memory_mappings_sanity_checks` name for the `mappings_sanity_checks()` could definitely lead to the removal of that comment. (in case you choose to rename it, don't forget about the line 57, and probably lines ).

src/uffd_handler/src/memory_region.rs
```diff
@@ -48,18 +71,49 @@ pub fn create_mem_regions(mappings: Vec<GuestRegionUffdMapping>) -> Vec<MemRegio
...
+            .base_host_virt_addr
+            .checked_add(region.size)
+            .is_none()
+        {
+            // The region end overflows usize.
+            return Err(MemRegionError::EndRegionOverflow);
+        }
+    }
+
     let memsize: usize = mappings.iter().map(|r| r.size).sum();
```
andreitraistaru, line 104: Could we do the sum of the region sizes in the previous loop such that there is no need to pass through the vector twice? Or, do you know if the compiler generally optimize such things? And third question here (probably the most relevant): how many regions, on average, do we have in a snap file (a low number could be in favor to let it as it is for code visibility)?

andreitraistaru (approved):

src/uffd_handler/src/main.rs
```diff
@@ -53,6 +52,17 @@ pub enum Error {
 
 type Result<T> = std::result::Result<T, Error>;
 
+fn validate_snapshot_file(file_name: &String) -> Result<(File, usize)> {
```
roypat, line 55: Mh, I think something weird is going on with clippy. There's a lint about using `&str` instead of `&String` (`clippy::ptr_arg`). If I copy this function from `main.rs` to `lib.rs` then the lint fires 🤔.

src/uffd_handler/src/memory_region.rs
```diff
@@ -119,37 +173,73 @@ mod tests {
...
+
+        // Sanity checks fail because the specified file offset and length
+        // cause overflow when added.
+        mappings[0].offset = u64::MAX;
+        let res = mappings_sanity_checks(&mappings, PAGE_SIZE);
+        assert!(res.is_err());
+        assert_eq!(
+            MemRegionError::InvalidOffsetLength.to_string(),
+            res.err().unwrap().to_string()
+        );
```
roypat, line 233: nit: I would do 
```suggestion
        assert_eq!(res, Err(MemRegionError::InvalidOffsetLength));
```
or if the `PartialEq` bounds are not satisfied, `assert!(matches!(...))` (same thing in a lot of other places).

roypat (approved):

src/uffd_handler/src/main.rs
luminitavoicu: Hmm, I experienced the same behavior. I will take a look into why clippy doesn't catch all errors. In the meantime, I fixed the code.

src/uffd_handler/src/memory_region.rs
luminitavoicu: I agree that it's self explanatory, I removed the line and renamed the function as suggested. Also, thanks for the reminder!

src/uffd_handler/src/memory_region.rs
luminitavoicu: The regions are the actual guest memory regions, so that will be maximum 2 on x86_64 (because of the MMIO gap) and one on ARM. I *think* the compiler should optimize this, but I am not 100% sure.

src/uffd_handler/src/memory_region.rs
luminitavoicu: Done!

src/uffd_handler/src/memory_region.rs
andreitraistaru: I'm ok to let as it is in this case. Thanks for clarifications!

andreitraistaru (approved):

src/uffd_handler/src/memory_region.rs
roypat: ah, I meant `assert!(matches!(res, Err(MemRegionError::InvalidOffsetLength))`, sorry, wasn't clear. Anyway, no need for you to go through the PR again, its just a nit :)

(fyi, you can do `res.unwrap_err()` instead of `res.err().unwrap()`)

roypat (approved):

PR #3491, opened 2023-03-02
Uffd: implement configurable prefaulting

luminitavoicu:
## Changes

This PR adds the following:
- CLI argument parsing
- optional argument:
```
--pages-after is the number of pages to prefault after the
    faulting page
```
- logic based on the new optional argument that allows customization of prefaulting chunk sizes
- integration tests adjustments

## Reason

Customizable prefaulting adds more flexibility and makes it easier to do performance testing.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -20,7 +20,7 @@ from host_tools import proc
 if utils.is_io_uring_supported():
     COVERAGE_DICT = {"Intel": 82.67, "AMD": 81.88, "ARM": 82.17}
 else:
-    COVERAGE_DICT = {"Intel": 79.85, "AMD": 79.05, "ARM": 79.11}
+    COVERAGE_DICT = {"Intel": 79.9, "AMD": 79.05, "ARM": 79.11}
 
 PROC_MODEL = proc.proc_type()
 


src/uffd_handler/src/main.rs
```diff
@@ -95,6 +120,16 @@ fn build_arg_parser() -> ArgParser<'static> {
                 .takes_value(true)
                 .help("The path to the snapshot memory file to serve page faults from."),
         )
+        .arg(
+            Argument::new("before-kib")
+                .takes_value(true)
+                .help("Amount in KiB to prefault before the start address of the faulting page."),
```
andreitraistaru, line 147: Should we specify that default value is 0 in the help message?

src/uffd_handler/src/prefaulter.rs
```diff
@@ -0,0 +1,31 @@
...
+
+impl UffdPrefaulter {
+    pub fn new(before: usize, after: usize) -> Result<Self, PrefaulterError> {
+        Ok(UffdPrefaulter {
+            bytes_before: before
+                .checked_add(KIB_MULT)
+                .ok_or_else(|| InvalidAmount(String::from("before"), before))?,
+            bytes_after: after
+                .checked_add(KIB_MULT)
+                .ok_or_else(|| InvalidAmount(String::from("after"), before))?,
```
andreitraistaru: Shouldn't be `InvalidAmount(String::from("after"), after))?,` instead?

src/uffd_handler/src/prefaulter.rs
```diff
@@ -0,0 +1,31 @@
...
+pub struct UffdPrefaulter {
+    bytes_before: usize,
+    bytes_after: usize,
+}
+
+impl UffdPrefaulter {
+    pub fn new(before: usize, after: usize) -> Result<Self, PrefaulterError> {
+        Ok(UffdPrefaulter {
+            bytes_before: before
+                .checked_add(KIB_MULT)
```
andreitraistaru: Is this the check that verifies that the number of KiB provided for before/after prefaulting is not equivalent to a number of bytes that is still `usize` (and doesn't produce an overflow)? I mean, I'm not quite sure why is `.checked_add(KIB_MULT)` and not `.checked_mul(KIB_MULT)` in order to convert KiB (as parsed from cmdline) to bytes (as saved in the `struct UffdPrefaulter`)

andreitraistaru:
nit: Commit `feature: uffd: implement configurable prefault` has a small inconsistency in the 4th line of the commit message: instead of `[faulting_page_start - after_kib, faulting_page_end + before_kib]` I think it should have been something like `[faulting_page_start - before_kib, faulting_page_end + after_kib]`

src/uffd_handler/src/prefaulter.rs
andreitraistaru: Ah.. so you fixed that in the next commit (`feature: uffd: implement configurable prefault`). Sorry, I wasn't reading until the end before writing. However, it would be nice to change the second commit, also (`feature: uffd: add optional arguments `)

src/uffd_handler/src/prefaulter.rs
andreitraistaru: Ah.. so you fixed that in the next commit (feature: uffd: implement configurable prefault). Sorry, I wasn't reading until the end before writing. However, it would be nice to change the second commit, also (feature: uffd: add optional arguments )

src/uffd_handler/src/prefaulter.rs
```diff
@@ -21,11 +26,266 @@ impl UffdPrefaulter {
...
+                    break;
+                }
+            }
+        }
+
+        Ok(prefault_start_addr)
+    }
+
+    /// Return the end address of the chunk to prefault into memory. The address is obtained
+    /// by adding `after` amount of bytes to the end of the faulting page. If the end address
```
andreitraistaru, line 85: nit: /// by adding `bytes_after` amount of bytes to the end of the faulting page. If the end address

src/uffd_handler/src/prefaulter.rs
```diff
@@ -21,11 +26,266 @@ impl UffdPrefaulter {
...
-                .checked_add(KIB_MULT)
-                .ok_or_else(|| InvalidAmount(String::from("after"), before))?,
+                .checked_mul(KIB_MULT)
+                .ok_or_else(|| PrefaulterError::InvalidAmount(String::from("after"), after))?,
         })
     }
+
+    /// Return the start address of the chunk to prefault into memory.
+    /// When there is no customized prefaulting configured, the minimum chunk to copy to RAM
+    /// is the faulting page.
```
andreitraistaru, line 39: Because there are a couple of steps performed by this method, I would enhance the docs comment a little bit like this:

> /// Return the start address of the chunk to prefault into memory. The address is obtained
    /// by subtracting `bytes_before` amount of bytes from the start of the faulting page and
    /// rounding down to a page start. If the computed start address is smaller then the start of
    /// current region, the start of the current memory region is returned. Furthermore, a check for
    /// already existing pages in RAM is performed such that the first Uninitialized page beginning
    /// from the prefaulting start address (that was just computed) is returned eventually.
    /// When there is no customized prefaulting configured, the minimum chunk to copy to RAM
    /// is the faulting page.

src/uffd_handler/src/prefaulter.rs
andreitraistaru: nit(more like a personal taste): Maybe to keep it consisten with the docs comment from the previous method, we can say smth like:

> /// Return the end address of the chunk to prefault into memory. The address is obtained
    /// by adding `bytes_after` amount of bytes to the end of the faulting page. If the end address
    /// exceeds the current region, the chunk is truncated to fit into the memory region
    /// (the end of the current memory region is returned).
    /// When there is no customized prefaulting configured, the minimum chunk to copy to RAM
    /// is the faulting page.

src/uffd_handler/src/prefaulter.rs
```diff
@@ -21,11 +26,266 @@ impl UffdPrefaulter {
...
+        let page_size = get_page_size().map_err(HandlerError::PageSize)?;
+
+        // This will not overflow because it has been verified beforehand during mappings
+        // deserialization.
+        let region_end = region.mapping.base_host_virt_addr + region.mapping.size;
+        // We add `page_size` because we always want to at least bring into RAM the faulting page,
+        // even if customized prefaulting is not configured.
+        let prefault_chunk_end = self
+            .bytes_after
+            .checked_add(page_fault_start_addr + page_size)
```
andreitraistaru, line 104: Maybe we could add another comment here that `page_fault_start_addr + page_size` could not overflow (as we know that the maximum value is the size of the region)?

src/uffd_handler/src/prefaulter.rs
```diff
@@ -21,11 +26,266 @@ impl UffdPrefaulter {
...
+            while prefault_start_addr < page_fault_start_addr {
+                prefault_start_addr += page_size;
+                // Found the first Uninitialized occurrence, so prefaulting will start from there.
+                // This unwrap is safe because it is guaranteed that the key is the start of a page
+                // within the current region.
+                if *region.page_states.get(&prefault_start_addr).unwrap()
+                    == MemPageState::Uninitialized
+                {
+                    break;
+                }
```
andreitraistaru, line 77: Is there a possibility that the user, let's say, provides a `--before-kib 8` (which is 2 pages before) and those 2 pages before the actual page fault to be: the first one is `MemPageState::Uninitialized` but the following one is not `MemPageState::Uninitialized`? In this case, the `get_prefaulting_start_address()` will return the starting address of the `MemPageState::Uninitialized` page but the following one is not guaranteed to be also `MemPageState::Uninitialized`.  Would that be a concern thinking that we skipped the pages that are already in memory? Sorry, I'm not quite familiar with the code, yet.

andreitraistaru:
Sorry for being nit picking and letting you so many comments, maybe some of them are not so relevant, but feel free to adapt where needed. Thank you a lot for the patience to read them!

src/uffd_handler/src/main.rs
```diff
@@ -95,6 +120,16 @@ fn build_arg_parser() -> ArgParser<'static> {
                 .takes_value(true)
                 .help("The path to the snapshot memory file to serve page faults from."),
         )
+        .arg(
+            Argument::new("before-kib")
```
roypat, line 145: Mh, if this always needs to be a multiple of the page size, would it make sense to specify this argument in "number of pages to prefault" as opposed to "kb to prefault"?

src/uffd_handler/src/prefaulter.rs
roypat: I _think_ this cant happen because any not-uninitialized page will have been brought in by the uffd handler, and the invariant of this uffd handler is that it always prefaults (e.g. if `--before-kb 8` is specified, then pages are always brought in in sequences of 3 pages - the page requested + 2 prefaulted ones). This means "islands" of initialized pages of size < 3 are impossible, and that's why the case you outline cannot happen. 

But I am also not sure. If this is the case, then that should definitely be mentioned in this method

src/uffd_handler/src/prefaulter.rs
```diff
@@ -21,11 +26,266 @@ impl UffdPrefaulter {
...
+        let region_end = region.mapping.base_host_virt_addr + region.mapping.size;
+        // We add `page_size` because we always want to at least bring into RAM the faulting page,
+        // even if customized prefaulting is not configured.
+        let prefault_chunk_end = self
+            .bytes_after
+            .checked_add(page_fault_start_addr + page_size)
+            .unwrap_or(region_end);
+
+        // Make sure that the end of the chunk to prefault is within the current region.
+        Ok(min(prefault_chunk_end, region_end))
```
roypat, line 108: nit:

```suggestion
        Ok(self.bytes_after
            .checked_add(page_fault_start_addr + page_size)
            .unwrap_or(region_end)
            .min(region_end))
```

src/uffd_handler/src/prefaulter.rs
```diff
@@ -21,11 +26,266 @@ impl UffdPrefaulter {
...
+    fn test_init_prefaulter() {
+        let prefaulter = UffdPrefaulter::new(512, 1024).unwrap();
+        assert_eq!(prefaulter.bytes_before, 512 * KIB_MULT);
+        assert_eq!(prefaulter.bytes_after, 1024 * KIB_MULT);
+
+        let res = UffdPrefaulter::new(usize::MAX, 4);
+        assert!(res.is_err());
+        assert_eq!(
+            res.err().unwrap().to_string(),
+            PrefaulterError::InvalidAmount("before".to_string(), usize::MAX).to_string()
```
roypat, line 141: nit: Can we have these `"before"`/`"after"` strings either be `&'static str`, or even better imo, distinct enum variants?

src/uffd_handler/src/handler.rs
```diff
@@ -65,18 +67,31 @@ impl UffdManager for Uffd {
...
-        // Populate whole region from backing mem-file.
-        // This offers an example of how memory can be loaded in RAM,
-        // however this can be adjusted to accommodate use case needs.
+        let page_size = get_page_size().map_err(HandlerError::PageSize)?;
+        // The result will always be positive because both start_addr and end_addr are within the
+        // current region and the difference between them is always at least page size.
+        let mut len = end_addr - start_addr;
+        // Length to uffd copy must be multiple of page size.
+        if len % page_size != 0 {
+            len = page_size * (len / page_size + 1);
```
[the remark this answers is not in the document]
luminitavoicu: > we can avoid multiplication/division via len += page_size - remainder

You are right! I have no idea why I overly complicated this check, haha. 

> And since page_size is always a multiple of 2, we can even do something like len = (len | (page_size - 1)) + 1, but that's very obscure and probably shouldnt be done unless we really care about every last ns of performance here, haha.

I would go with your first suggestion, I agree that `len = (len | (page_size - 1)) + 1` looks obscure and it's harder to digest.

src/uffd_handler/src/prefaulter.rs
luminitavoicu: On a second thought, I need to think if my logic holds when we add the balloon device into the equation. I saw a test failing earlier and I think it might have something to do with the scenario outlined by @andreitraistaru above, coupled with the balloon device's work.

andreitraistaru: Regarding the initial scenario (without the balloon device) it looks like, yeah...., it's not actually possible to reach the situation I mentioned if keeping the number of pages to prefault a constant (in case we'll allow changing `pages-before` at runtime, the problem will become possible but that's not the case yet, right?).

Regarding the second scenario (when adding the balloon device in discussion) I think it could be a problem depending on the placement of the pages taken from the balloon (I'm not really familiar with how the balloon works, but reading from docs, looks like the kernel is responsible with this so, I'm not sure how much control we have over it). I'm talking about a scenario like this:

Initially (3) was a faulting page some time ago, and we have a prefault of 2 pages before and 0 pages after ):

|.....|_FromFile_||_FromFile_|_FromFile_|_Uninitialized_|......|
               (1)                (2)               (3)                  (4)

After a memory allocation from guest, the balloon gives kernel another page; if the new page (5) is placed like this, it could be a problem:

|.....|_FromFile_||_FromFile_|_Uninitialized_|_FromFile_|_Uninitialized_|......|
               (1)                (2)               (5)                     (3)                  (4)

Now, if we get a page fault at (4), the prefaulter will return (5) as the start address to bring into memory which is not ok as long as we have (3) already in memory in between the start address and the actual page fault. But I'm still not convinced that such a situation would make sense or not.

luminitavoicu: Yes, the scenario you described totally makes sense. We have no control over the pages that are claimed back by the host, so they could be placed anywhere really. 

Unfortunately, the balloon doesn't seem to work with UFFD for now. On balloon inflate, the removed pages are `madvise`'d with `MADV_DONTNEED`. When this happens, userfaultfd is supposed to receive also `Remove` event, apart from the `PageFault` events (see [here](https://man7.org/linux/man-pages/man2/ioctl_userfaultfd.2.html))
>If this feature is enabled, when the faulting process calls [madvise(2)](https://man7.org/linux/man-pages/man2/madvise.2.html) with the MADV_DONTNEED or MADV_REMOVE advice value to free a virtual memory area the userfaultfd monitor will receive an event of type UFFD_EVENT_REMOVE.

However, this doesn't happen, so uffd handler is unaware of the memory changes. The test that is failing our CI illustrates exactly this scenario: the handler fails because it tries to copy a page that has been claimed by the host, thus `UFFDIO_COPY` returns [`ENOENT`](https://man7.org/linux/man-pages/man2/ioctl_userfaultfd.2.html) and Firecracker freezes.

src/uffd_handler/src/prefaulter.rs
```diff
@@ -45,11 +46,108 @@ impl UffdPrefaulter {
...
+        let page_size = get_page_size().map_err(HandlerError::PageSize)?;
+
+        // This will not overflow because it has been verified beforehand during mappings
+        // deserialization.
+        let region_end = region.mapping.base_host_virt_addr + region.mapping.size;
+        // We add `page_size` because we always want to at least bring into RAM the faulting page,
+        // even if customized prefaulting is not configured. If the result exceeds the current
+        // region, the end of the region is returned instead.
+        Ok(self
+            .bytes_after
```
kalyazin, line 124: What makes it impossible for some pages being already prefaulted between the currently faulting page and `bytes_after` so we don't need a loop similar to the one in `get_prefaulting_start_address `?

luminitavoicu: The userfaultfd already takes care of this and stops copying when it encounters a page that has already been brought into memory. It doesn't error out, it just returns the number of bytes that it managed to bring into RAM. It is redundant for us to perform the same checks and verify each page specifically, as long as the faulting page is served.

src/uffd_handler/src/prefaulter.rs
luminitavoicu: I reduced the scope of this PR in order to unblock it, because it seems like we are going to need a different approach for ensuring that the faulting start is valid and that all pages in the computed range (at least up until the faulting page) are `Uninitialized`. My idea is to start checking from the faulting page down to the computed start and stop at the first occurrence of a non-`Uninitialized` page.

However, in order to unblock this PR, I removed the functionality of prefaulting pages before the faulting page. This way we always start from the faulting page and can only configure `pages-after`. Prefaulting pages before will be added in a subsequent PR. WDYT?

src/uffd_handler/src/main.rs
```diff
@@ -70,6 +78,10 @@ struct Arguments {
     socket: String,
     /// The path to the snapshot memory file to serve page faults from.
     mem_file: String,
+    /// The number of pages to prefault before the faulting page.
```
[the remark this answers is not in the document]
luminitavoicu: I am not sure that `vm-memory` has full support for huge pages, I will check. But there might be other changes needed in order for Firecracker to support huge pages. However, you have a good point for the future. I will update both the docs and the `help` message.

src/uffd_handler/src/prefaulter.rs
andreitraistaru: Yes, I agree with the plan to split the work in multiple PRs and keep this PR for prefaulting pages after the faulting page only.

src/uffd_handler/src/main.rs
```diff
@@ -79,10 +89,18 @@ fn get_argument_values(args: &ArgumentsBag) -> Result<Arguments> {
     let mem_file = args.single_value("mem-file").ok_or_else(|| {
         Error::ArgumentParsing(ArgumentError::MissingValue("mem-file".to_string()))
     })?;
+    let after = args
+        .single_value("pages-after")
+        .map(|kib| {
```
kalyazin, line 94: nit: this should be the number of pages, not kib, or am I wrong?

src/uffd_handler/src/handler.rs
kalyazin, line 1: nit: commit message: is `prefetch` used deliberately?

src/uffd_handler/src/handler.rs
```diff
@@ -65,18 +67,32 @@ impl UffdManager for Uffd {
...
+
+        // Compute source to prefault the memory chunk from.
+        let prefault_offset = start_addr - region.mapping.base_host_virt_addr;
+        let src = buff as u64 + region.mapping.offset + prefault_offset as u64;
+
         // SAFETY: this is safe because the parameters are valid.
         let ret = unsafe {
-            self.copy(src as *const _, start_addr as *mut _, len, true)
-                .map_err(HandlerError::UffdCopy)?
+            match self.copy(src as *const _, start_addr as *mut _, len, true) {
```
kalyazin, line 90: nit: not sure if it's idiomatic in Rust, but do we actually need to wrap the pattern matching in `unsafe` too? I guess it is only `copy()` that is unsafe here?

src/uffd_handler/src/main.rs
```diff
@@ -105,6 +123,10 @@ fn build_arg_parser() -> ArgParser<'static> {
                 .takes_value(true)
                 .help("The path to the snapshot memory file to serve page faults from."),
         )
+        .arg(Argument::new("pages-after").takes_value(true).help(
+            "The number of pages to prefault after the faulting page. If unspecified, the default \
+             value is 0.",
```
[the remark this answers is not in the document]
luminitavoicu: Good idea! I added a mention similar to what you requested. Let me know what you think.

src/uffd_handler/src/handler.rs
luminitavoicu: No, it was used as a synonym to `prefault`. My brain was tired of using `prefault` all the time :)). Fixed.

src/uffd_handler/src/handler.rs
luminitavoicu: That is true, there is no need to wrap the entire `match` into `unsafe`. Fixed.

kalyazin (approved):

PR #3499, opened 2023-03-03
fix: create clean artifact dir for taking snapshot

sudanl0:
While running snapshot test multiple times on the same instance, the snapshot test fails while trying to create artifacts directory if the old artifact directories are not deleted manually so, delete old artifacts directories and create clean directory before taking a new snasphot to avoid such failures.

To reproduce the issue run below command, twice back to back, manually (not through buildkite) on an instance :
> tools/devtool -y test -- -s --nonci integration_tests/functional/test_cpu_features.py -k 'test_cpu_cpuid_snapshot or test_cpu_wrmsr_snapshot'

## Changes

Use a new function to delete existing old artifacts directory and create clean directory before taking a snapshot.

## Reason

Snapshot test fails while creating an artifact directory if old artifact directory is not removed manually.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

~~- [ ] If a specific issue led to this PR, this PR closes the issue.~~
~~- [ ] API changes follow the [Runbook for Firecracker API changes][2].~~
-~~ [ ] User-facing changes are mentioned in `CHANGELOG.md`.~~
~~- [ ] New `TODO`s link to an issue.~~
- [x] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).~~

---

~~- [ ] This functionality can be added in [`rust-vmm`][1].~~

### DIFF
--- a/tests/integration_tests/functional/test_cpu_features.py
+++ b/tests/integration_tests/functional/test_cpu_features.py
@@ -27,6 +27,14 @@ from framework.utils_cpu_templates import SUPPORTED_CPU_TEMPLATES
 PLATFORM = platform.machine()
 
 
+def clean_and_mkdir(dir_path):
+    """
+    Create a clean directory
+    """
+    shutil.rmtree(dir_path, ignore_errors=True)
+    os.makedirs(dir_path)
+
+
 def _check_cpuid_x86(test_microvm, expected_cpu_count, expected_htt):
     expected_cpu_features = {
         "cpu count": "{} ({})".format(hex(expected_cpu_count), expected_cpu_count),
@@ -445,8 +453,7 @@ def _test_cpu_wrmsr_snapshot(context):
         / context.kernel.base_name()
         / get_cpu_template_dir(cpu_template)
     )
-    shutil.rmtree(snapshot_artifacts_dir, ignore_errors=True)
-    os.makedirs(snapshot_artifacts_dir)
+    clean_and_mkdir(snapshot_artifacts_dir)
 
     msrs_before_fname = Path(snapshot_artifacts_dir) / shared_names["msrs_before_fname"]
 
@@ -621,7 +628,7 @@ def _test_cpu_wrmsr_restore(context):
     tmp_snapshot_artifacts_dir = (
         Path() / chroot_dir / "tmp" / context.kernel.base_name()
     )
-    os.makedirs(tmp_snapshot_artifacts_dir)
+    clean_and_mkdir(tmp_snapshot_artifacts_dir)
 
     mem_fname_in_jail = Path(tmp_snapshot_artifacts_dir) / shared_names["mem_fname"]
     snapshot_fname_in_jail = (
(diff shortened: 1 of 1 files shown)


tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -436,8 +444,7 @@ def _test_cpu_wrmsr_snapshot(context):
         / context.kernel.base_name()
         / get_cpu_template_dir(cpu_template)
     )
-    shutil.rmtree(snapshot_artifacts_dir, ignore_errors=True)
-    os.makedirs(snapshot_artifacts_dir)
+    clean_and_mkdir(snapshot_artifacts_dir)
```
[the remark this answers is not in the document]
sudanl0: this is seen when you run the test manually on instances and not through buildkite.
e.g. I encountered the issue when I was updating wrmsr test and had to run it manually on my instances to test it.

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -612,7 +619,7 @@ def _test_cpu_wrmsr_restore(context):
     tmp_snapshot_artifacts_dir = (
         Path() / chroot_dir / "tmp" / context.kernel.base_name()
     )
-    os.makedirs(tmp_snapshot_artifacts_dir)
+    clean_and_mkdir(tmp_snapshot_artifacts_dir)
```
kalyazin, line 622: Could you explain a bit how it works if run locally (both snapshot and restore on the same machine). How are the artifacts transferred between the two parts? Is this not removing them before they are consumed?

mattschlebusch:
nitpick: I'm a fan of the temp directory approach, [TempDir](https://docs.rs/tempdir/latest/tempdir/). Clean up happens as a part of test scope, and is unique for each iteration as well.

tests/integration_tests/functional/test_cpu_features.py
sudanl0: The snapshots created the path specified by snapshot_artifacts_dir ("tests/snapshot_artifacts/wrmsr/").
tmp_snapshot_artifacts_dir (e.g. "/srv/jailer/firecracker/d16e2e7e-99b8-4519-8975-6f593b507801/root/tmp/vmlinux-4.14 ") is the temporary jailer path.
Files are copied from snapshot_artifacts_dir to tmp_snapshot_artifacts_dir and then used to restore vm.

So we are not removing the snapshot files before consuming them, we are just making sure the temp jailer directory is clean when we copy the artifacts to it. The same is done in the _test_cpu_cpuid_restore().

Also, the new code just replaces os.makedirs() with shutil.rmtree()+os.makedirs() which doesn't change the existing working behaviour. If a directory already existed then the old code would have crashed at os.makedirs(tmp_snapshot_artifacts_dir) anyways.

kalyazin: Clear, thank you!

kalyazin (approved):

sudanl0:
> nitpick: I'm a fan of the temp directory approach, [TempDir](https://docs.rs/tempdir/latest/tempdir/). Clean up happens as a part of test scope, and is unique for each iteration as well.

I think we cannot use a temp dir approach because snapshot and restore are 2 different tests and we need to share the artifacts between these two tests when running the test locally. We can upload and download the artifacts but I think its just less time consuming if the artifacts are available without the need to download them when we run the tests locally.

andreitraistaru (approved):

Issue #3233, opened 2022-11-02
Use `clippy::exit`


JonathanWoollett-Light:
Implement [`clippy::exit`](https://rust-lang.github.io/rust-clippy/master/#exit).

> ### What it does
> 
> exit() terminates the program and doesn’t provide a stack trace.
>
> ### Why is this bad?
> 
> Ideally a program is terminated by finishing the main function.

Without a stack trace and the program converging on main it can become difficult to track down errors and understand the flow of the program. 

[The standard library documentation](https://doc.rust-lang.org/stable/std/process/fn.exit.html) also notes:

> preferably, simply return a type implementing [Termination](https://doc.rust-lang.org/stable/std/process/trait.Termination.html) (such as [ExitCode](https://doc.rust-lang.org/stable/std/process/struct.ExitCode.html) or Result) from the main function and avoid this function altogether:

dhruvsethi227:
Hello, my name is Dhruv and I am a senior computer science student at UT Austin. My group and I were wondering if we could tackle this issue as part of our virtualization class assignment.

JonathanWoollett-Light:
If you could contribute on this issue that would be great! We always appreciate contribution so feel free to do so.

I would note that among the issues tagged with `Contribute: Good First Issue` this is likely to be one of the more difficult ones.

0xkelvin:
@JonathanWoollett-Light does anyone working on this, otherwise i can take it

JonathanWoollett-Light:
@0xkelvin Your contribution would be awesome, you should always feel free to contribute on open issues, you don't need to ask.

PR #3506, opened 2023-03-06
Uffd: inform Firecracker of crashes

luminitavoicu:
## Changes

Configure UFFD handler to send a SIGBUS signal to Firecracker in case something goes wrong while handling the page faults.

## Reason

Firecracker relies on the userfaultfd handler process to handle page faults. In case this userspace process crashes, the requested page will never be copied to RAM and Firecracker will freeze waiting for the unavailable memory. Sending a SIGBUS signal to Firecracker will prevent it to stale.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- [X The description of changes is clear and encompassing.

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1225,6 +1225,7 @@ name = "uffd_handler"
 version = "1.4.0"
 dependencies = [
  "libc",
+ "nix",
  "serde",
  "serde_json",
  "thiserror",
--- a/src/uffd_handler/Cargo.toml
+++ b/src/uffd_handler/Cargo.toml
@@ -14,6 +14,7 @@ path = "src/main.rs"
 
 [dependencies]
 libc = "0.2.117"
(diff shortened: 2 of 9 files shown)


roypat:
Mh, there are two things that come to mind when looking this over:
1. `process::exit` seems discouraged, and given that we have #3233, would it be feasible to instead return `ExitCode` from `main` (https://doc.rust-lang.org/stable/std/process/struct.ExitCode.html)?
2. This only deals with "foreseeable" errors. A more general solution could be to install a [panic hook](https://doc.rust-lang.org/std/panic/fn.set_hook.html) in which we send the `SIGBUS` to firecracker. This way, we can be sure that no matter why the uffd process might go away, it will take firecracker with it (as opposed to leading to a potentially hard-to-debug microvm hang with non-local cause).
What do you think?

PR #3511, opened 2023-03-08
CPU Template Wrapping and Deserialization

mattschlebusch:
## Changes

Changes extend [PR/3483](https://github.com/firecracker-microvm/firecracker/pull/3483/) that have the `*CpuConfiguration` types defined.

* CPU template types for x86_64 and aarch64.
* `apply_template()` function for x86_64 to use the template to create `X86_64CpuConfiguration`.

## Reason

* CPU template types to be used to store the payload for `/cpu-config` API.

## License Acceptance

[`CONTRIBUTING.md`][3].

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -3,7 +3,6 @@
 
 //! Enables pre-boot setup, instantiation and booting of a Firecracker VMM.
 
-#[cfg(target_arch = "aarch64")]
 use std::collections::HashMap;
 #[cfg(target_arch = "x86_64")]
 use std::convert::TryFrom;
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -48,6 +48,11 @@ use crate::device_manager::persist::MMIODevManagerConstructorArgs;
 use crate::guest_config::aarch64::Aarch64CpuConfiguration;
 #[cfg(target_arch = "x86_64")]
 use crate::guest_config::cpuid::{Cpuid, RawCpuid};
+use crate::guest_config::templates;
+#[cfg(target_arch = "aarch64")]
+use crate::guest_config::templates::aarch64::{create_guest_cpu_config, Aarch64CpuTemplate};
+#[cfg(target_arch = "x86_64")]
+use crate::guest_config::templates::x86_64::{create_guest_cpu_config, X86_64CpuTemplate};
 #[cfg(target_arch = "x86_64")]
 use crate::guest_config::x86_64::X86_64CpuConfiguration;
 use crate::persist::{MicrovmState, MicrovmStateError};
@@ -63,6 +68,8 @@ use crate::{device_manager, Error, EventManager, RestoreVcpusError, Vmm, VmmEven
 /// Errors associated with starting the instance.
 #[derive(Debug)]
 pub enum StartMicrovmError {
+    /// Error using CPU template to configure vCPUs
+    ApplyCpuTemplate(templates::Error),
     /// Unable to attach block device to Vmm.
     AttachBlockDevice(io::Error),
     /// This error is thrown by the minimal boot loader implementation.
@@ -115,6 +122,14 @@ impl Display for StartMicrovmError {
     fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
         use self::StartMicrovmError::*;
         match self {
+            ApplyCpuTemplate(err) => {
+                write!(
+                    f,
+                    "Unable to successfully apply custom CPU template to create configuration \
+                     usable for guest vCPUs: {:?}",
+                    err
+                )
+            }
             AttachBlockDevice(err) => {
                 write!(f, "Unable to attach block device to Vmm: {}", err)
             }
@@ -245,6 +260,7 @@ fn create_vmm_and_vcpus(
     use self::StartMicrovmError::*;
 
     // Set up Kvm Vm and register memory regions.
+    // Build custom CPU config if a custom template is provided.
     let mut vm = setup_kvm_vm(&guest_memory, track_dirty_pages)?;
 
     let vcpus_exit_evt = EventFd::new(libc::EFD_NONBLOCK)
@@ -268,6 +284,27 @@ fn create_vmm_and_vcpus(
     let pio_device_manager = {
         setup_interrupt_controller(&mut vm)?;
         vcpus = create_vcpus(&vm, vcpu_count, &vcpus_exit_evt).map_err(Internal)?;
+
+        // Apply CPU template to create vCPU custom config if available
+        if let Some(template) = &vm.guest_cpu_template {
+            if let Some(vcpu) = vcpus.get(0) {
+                vm.guest_cpu_config = Some(
+                    create_guest_cpu_config(
+                        template,
+                        &X86_64CpuConfiguration {
+                            cpuid: Cpuid::try_from(RawCpuid::from(vm.supported_cpuid().clone()))
+                                .unwrap(),
+                            msrs: get_msr_values(vcpu, template),
+                        },
+                    )
+                    .map_err(ApplyCpuTemplate)?,
+                );
+            } else {
+                return Err(ApplyCpuTemplate(templates::Error::Internal(
+                    "Unable to access a vCPU".to_string(),
+                )));
+            }
+        }
 
         // Make stdout non blocking.
         set_stdout_nonblocking();
@@ -295,6 +332,26 @@ fn create_vmm_and_vcpus(
     #[cfg(target_arch = "aarch64")]
     {
         vcpus = create_vcpus(&vm, vcpu_count, &vcpus_exit_evt).map_err(Internal)?;
+
+        // Apply CPU template to create vCPU custom config if available
+        if let Some(template) = &vm.guest_cpu_template {
+            if let Some(vcpu) = vcpus.get(0) {
+                vm.guest_cpu_config = Some(
+                    create_guest_cpu_config(
+                        template,
+                        &Aarch64CpuConfiguration {
+                            regs: get_reg_values(vcpu, template),
+                        },
+                    )
+                    .map_err(ApplyCpuTemplate)?,
+                );
+            } else {
+                return Err(ApplyCpuTemplate(templates::Error::Internal(
+                    "Unable to access a vCPU".to_string(),
+                )));
+            }
+        }
+
         setup_interrupt_controller(&mut vm, vcpu_count)?;
     }
 
@@ -313,6 +370,20 @@ fn create_vmm_and_vcpus(
     };
 
     Ok((vmm, vcpus))
+}
+
+#[cfg(target_arch = "x86_64")]
+#[allow(unused)]
+fn get_msr_values(_vcpu: &Vcpu, _template: &X86_64CpuTemplate) -> HashMap<u32, u64> {
+    // TODO - Use a created vCPU to retrieve MSR values
+    Default::default()
+}
+
+#[cfg(target_arch = "aarch64")]
+#[allow(unused)]
+fn get_reg_values(_vcpu: &Vcpu, _template: &Aarch64CpuTemplate) -> HashMap<u64, u128> {
+    // TODO - Use a created vCPU to retrieve register values
+    Default::default()
 }
 
 /// Builds and starts a microVM based on the current Firecracker VmResources configuration.
@@ -340,7 +411,6 @@ pub fn build_microvm_for_boot(
     let track_dirty_pages = vm_resources.track_dirty_pages();
     let guest_memory =
         create_guest_memory(vm_resources.vm_config().mem_size_mib, track_dirty_pages)?;
-    let vcpu_config = vm_resources.vcpu_config();
     let entry_addr = load_kernel(boot_config, &guest_memory)?;
     let initrd = load_initrd_from_config(boot_config, &guest_memory)?;
     // Clone the command-line so that a failed boot doesn't pollute the original.
@@ -353,8 +423,9 @@ pub fn build_microvm_for_boot(
         guest_memory,
         None,
         track_dirty_pages,
-        vcpu_config.vcpu_count,
+        vm_resources.vm_config().vcpu_count,
     )?;
+    let vcpu_config = vm_resources.custom_vcpu_config(vmm.vm.guest_cpu_config.clone());
 
     // The boot timer device needs to be the first device attached in order
     // to maintain the same MMIO address referenced in the documentation
@@ -700,7 +771,9 @@ pub(crate) fn setup_kvm_vm(
     let kvm = KvmContext::new()
         .map_err(Error::KvmContext)
         .map_err(Internal)?;
-    let mut vm = Vm::new(kvm.fd()).map_err(Error::Vm).map_err(Internal)?;
+    let mut vm = Vm::new(kvm.fd(), None)
+        .map_err(Error::Vm)
+        .map_err(Internal)?;
     vm.memory_init(guest_memory, kvm.max_memslots(), track_dirty_pages)
         .map_err(Error::Vm)
         .map_err(Internal)?;
@@ -840,13 +913,7 @@ pub fn configure_system_for_boot(
                     vmm.guest_memory(),
                     entry_addr,
                     &vcpu_config,
-                    X86_64CpuConfiguration {
-                        cpuid: Cpuid::try_from(
-                            RawCpuid::from(vmm.vm.supported_cpuid().clone()).clone(),
-                        )
-                        .unwrap(),
-                        msrs: HashMap::new(),
-                    },
+                    vmm.vm.supported_cpuid().clone(),
                 )
                 .map_err(Error::VcpuConfigure)
                 .map_err(Internal)?;
@@ -877,13 +944,7 @@ pub fn configure_system_for_boot(
     {
         for vcpu in vcpus.iter_mut() {
             vcpu.kvm_vcpu
-                .configure(
-                    vmm.guest_memory(),
-                    entry_addr,
-                    Aarch64CpuConfiguration {
-                        regs: HashMap::new(),
-                    },
-                )
+                .configure(vmm.guest_memory(), entry_addr, &vcpu_config)
                 .map_err(Error::VcpuConfigure)
                 .map_err(Internal)?;
         }
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -3,6 +3,7 @@
 
 //! Enables pre-boot setup, instantiation and booting of a Firecracker VMM.
 
+#[cfg(target_arch = "aarch64")]
 use std::collections::HashMap;
 #[cfg(target_arch = "x86_64")]
 use std::convert::TryFrom;
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -48,11 +48,11 @@ use crate::device_manager::persist::MMIODevManagerConstructorArgs;
 use crate::guest_config::aarch64::Aarch64CpuConfiguration;
 #[cfg(target_arch = "x86_64")]
 use crate::guest_config::cpuid::{Cpuid, RawCpuid};
-use crate::guest_config::templates;
 #[cfg(target_arch = "aarch64")]
 use crate::guest_config::templates::aarch64::{create_guest_cpu_config, Aarch64CpuTemplate};
 #[cfg(target_arch = "x86_64")]
 use crate::guest_config::templates::x86_64::{create_guest_cpu_config, X86_64CpuTemplate};
+use crate::guest_config::templates::Error as GuestConfigError;
 #[cfg(target_arch = "x86_64")]
 use crate::guest_config::x86_64::X86_64CpuConfiguration;
 use crate::persist::{MicrovmState, MicrovmStateError};
@@ -69,7 +69,7 @@ use crate::{device_manager, Error, EventManager, RestoreVcpusError, Vmm, VmmEven
 #[derive(Debug)]
 pub enum StartMicrovmError {
     /// Error using CPU template to configure vCPUs
-    ApplyCpuTemplate(templates::Error),
+    ApplyCpuTemplate(GuestConfigError),
     /// Unable to attach block device to Vmm.
     AttachBlockDevice(io::Error),
     /// This error is thrown by the minimal boot loader implementation.
@@ -300,7 +300,7 @@ fn create_vmm_and_vcpus(
                     .map_err(ApplyCpuTemplate)?,
                 );
             } else {
-                return Err(ApplyCpuTemplate(templates::Error::Internal(
+                return Err(ApplyCpuTemplate(GuestConfigError::Internal(
                     "Unable to access a vCPU".to_string(),
                 )));
             }
@@ -346,7 +346,7 @@ fn create_vmm_and_vcpus(
                     .map_err(ApplyCpuTemplate)?,
                 );
             } else {
-                return Err(ApplyCpuTemplate(templates::Error::Internal(
+                return Err(ApplyCpuTemplate(GuestConfigError::Internal(
                     "Unable to access a vCPU".to_string(),
                 )));
             }
--- a/src/vmm/src/guest_config/aarch64/mod.rs
+++ b/src/vmm/src/guest_config/aarch64/mod.rs
@@ -3,10 +3,8 @@
 
 use std::collections::HashMap;
 
-use serde::{Deserialize, Serialize};
-
 /// CPU configuration for aarch64 CPUs
-#[derive(Deserialize, Default, Debug, Serialize)]
+#[derive(Clone, Default, Debug, Eq, PartialEq)]
 pub struct Aarch64CpuConfiguration {
     /// Register values as a key pair
     /// Key: Register pointer
--- a/src/vmm/src/guest_config/aarch64/mod.rs
+++ b/src/vmm/src/guest_config/aarch64/mod.rs
@@ -3,8 +3,10 @@
 
 use std::collections::HashMap;
 
+use serde::{Deserialize, Serialize};
+
 /// CPU configuration for aarch64 CPUs
-#[derive(Default, Debug)]
+#[derive(Deserialize, Default, Debug, Serialize)]
 pub struct Aarch64CpuConfiguration {
     /// Register values as a key pair
     /// Key: Register pointer
--- a/src/vmm/src/guest_config/aarch64/mod.rs
+++ b/src/vmm/src/guest_config/aarch64/mod.rs
@@ -4,7 +4,7 @@
 use std::collections::HashMap;
 
 /// CPU configuration for aarch64 CPUs
-#[derive(Default)]
+#[derive(Default, Debug)]
 pub struct Aarch64CpuConfiguration {
     /// Register values as a key pair
     /// Key: Register pointer
--- a/src/vmm/src/guest_config/templates/mod.rs
+++ b/src/vmm/src/guest_config/templates/mod.rs
@@ -56,6 +56,7 @@ pub mod x86_64 {
     use std::collections::{BTreeMap, HashMap};
     use std::str::FromStr;
 
+    use log::debug;
     use serde::de::Error as SerdeError;
     use serde::{Deserialize, Deserializer};
 
@@ -84,6 +85,7 @@ pub mod x86_64 {
         pub register: CpuidRegister,
         /// Bit mapping to be applied as a modifier to the
         /// register's value at the address provided.
+        #[serde(deserialize_with = "deserialize_bitmap")]
         pub bitmap: RegisterValueFilter,
     }
 
@@ -122,7 +124,7 @@ pub mod x86_64 {
         /// Filter to be used when writing the value bits.
         #[serde(deserialize_with = "deserialize_u64_from_str")]
         pub filter: u64,
-        /// Value value to be applied.
+        /// Value to be applied.
         #[serde(deserialize_with = "deserialize_u64_from_str")]
         pub value: u64,
     }
@@ -136,6 +138,7 @@ pub mod x86_64 {
         pub addr: u32,
         /// Bit mapping to be applied as a modifier to the
         /// register's value at the address provided.
+        #[serde(deserialize_with = "deserialize_bitmap")]
         pub bitmap: RegisterValueFilter,
     }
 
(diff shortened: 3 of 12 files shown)


src/vmm/src/builder.rs
```diff
@@ -833,7 +840,13 @@ pub fn configure_system_for_boot(
                     vmm.guest_memory(),
                     entry_addr,
                     &vcpu_config,
-                    vmm.vm.supported_cpuid().clone(),
+                    X86_64CpuConfiguration {
+                        cpuid: Cpuid::try_from(
+                            RawCpuid::from(vmm.vm.supported_cpuid().clone()).clone(),
```
zulinx86, line 845: Does this need to do `clone()` multiple times?
```suggestion
                            RawCpuid::from(vmm.vm.supported_cpuid().clone()),
```

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,551 @@
...
+
+/// Types common for template definitions for both
+/// Aarch64 and x86_64.
+
+/// Bit-mask based wrapper to apply
+/// changes to a given register's value
+#[derive(Debug, Deserialize, Eq, PartialEq)]
+pub struct RegisterModifier {
+    /// Pointer of the location to be bit masked.
+    #[serde(deserialize_with = "deserialize_number_from_str")]
```
zulinx86, line 17: I prefer the type-explicit form `deserialize_u64_from_str()` (similar to `deserialize_u32_from_str()`).

mattschlebusch: Sure. It is my intention to change to a generic numeric type for a single function further down the road, but does make sense now.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,551 @@
...
+            let mut guest_cpuid_map: BTreeMap<CpuidKey, CpuidEntry>;
+
+            // Get the hash map of CPUID data
+            if host_config.cpuid.amd().is_some() {
+                guest_cpuid_map = host_config.cpuid.amd().unwrap().0.clone();
+            } else if host_config.cpuid.intel().is_some() {
+                guest_cpuid_map = host_config.cpuid.intel().unwrap().0.clone();
+            } else {
+                return Err(Error::Cpuid);
+            }
```
zulinx86, line 136: nit: might be simplified.
```suggestion
            guest_cpuid_map = match host_config.cpuid {
                Cpuid::Intel(cpuid) => cpuid.0.clone(),
                Cpuid::Amd(cpuid) => cpuid.0.clone(),
            };
```

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,551 @@
...
+
+    /// Wrapper type to containing x86_64 CPU config modifiers.
+    #[derive(Debug, Deserialize, Eq, PartialEq)]
+    pub struct X86_64CpuTemplate {
+        /// Modifiers for CPUID configuration.
+        #[serde(default)]
+        pub cpuid_modifiers: Vec<CpuidRegisterModifier>,
+        /// Modifiers for model specific registers.
+        #[serde(default)]
+        pub msr_modifiers: Vec<RegisterModifier>,
```
zulinx86, line 115: AFAIK, as rdmsr takes the ECX register (32-bit), while the return value is EDX:EAX (64-bit), this might not work.
https://www.felixcloutier.com/x86/rdmsr

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,551 @@
...
+                    msrs: guest_msrs_map,
+                })
+            } else if host_config.cpuid.intel().is_some() {
+                Ok(X86_64CpuConfiguration {
+                    cpuid: Cpuid::Intel(IntelCpuid(guest_cpuid_map)),
+                    msrs: guest_msrs_map,
+                })
+            } else {
+                Err(Error::Internal)
+            }
```
zulinx86, line 199: nit: same simplification could be applied here.
```suggestion
            match host_config.cpuid {
                Cpuid::Amd(_) => Ok(X86_64CpuConfiguration {
                    cpuid: Cpuid::Amd(AmdCpuid(guest_cpuid_map)),
                    msrs: guest_msrs_map,
                }),
                Cpuid::Intel(_) => Ok(X86_64CpuConfiguration {
                    cpuid: Cpuid::Intel(IntelCpuid(guest_cpuid_map)),
                    msrs: guest_msrs_map,
                }),
            }
```

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+    addr: u64,
+    /// Mask to be applied to register's value at the address
+    /// provided.
+    mask: BitMask,
+}
+
+/// Bit mask to adjust targeted bits to the value,
+/// applied by the filter only.
+#[derive(Debug, Deserialize, Eq, PartialEq)]
+pub struct BitMask {
```
kalyazin, line 29: I don't really like `BitMask` encapsulating `filter` and `value`. Can we place `filter` (or `mask`) and `value` directly to the `RegisterModifier`?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -66,21 +86,183 @@ pub mod x86_64 {
...
+            if host_config.cpuid.amd().is_some() {
+                guest_cpuid_map = host_config.cpuid.amd().unwrap().0.clone();
+            } else if host_config.cpuid.intel().is_some() {
+                guest_cpuid_map = host_config.cpuid.intel().unwrap().0.clone();
+            } else {
+                return Err(Error::Cpuid);
+            }
+
+            // Apply CPUID modifiers
+            for modifier in self.cpuid_modifiers {
```
kalyazin, line 140: I think it might be a good idea to extract this and the other loop into separate functions.

src/vmm/src/guest_config/templates/mod.rs
zulinx86: It might be difficult to do that, because it needs to associate a custom deserialization function with a field...
```
    /// Mask to be applied to register's value at the address
    /// provided.
    #[serde(deserialize_with = "deserialize_bitmask")]
    mask: BitMask,
```

kalyazin: I see, makes sense, `filter` and `value` need to go together in a struct.
The problem with naming arises because `RegisterModifier` is good for both (addr, filter, value) and (filter, value).
If `RegisterModifier` is used for (addr, filter, value), it's hard to come up with a sensible name for (filter, value).

zulinx86: What do you think about the following naming?
```
struct CpuidModifier {
    mask: u32,
    value: u32,
}

#[derive(Deserialize, Debug)]
struct CpuidTemplate {
    leaf: u32,
    subleaf: u32,
    #[serde(deserialize_with = "deserialize_cpuid_register")]
    register: CpuidRegister,
    #[serde(deserialize_with = "deserialize_u32_modifier")]
    modifier: CpuidModifier,
}

#[derive(Deserialize, Debug)]
struct X86_64CpuTemplate {
    cpuid_template: Vec<CpuidTemplate>,
    msr_template: Vec<MsrTemplate>,
}
```

src/vmm/src/builder.rs
```diff
@@ -864,7 +875,13 @@ pub fn configure_system_for_boot(
...
         for vcpu in vcpus.iter_mut() {
             vcpu.kvm_vcpu
-                .configure(vmm.guest_memory(), entry_addr)
+                .configure(
+                    vmm.guest_memory(),
+                    entry_addr,
+                    Aarch64CpuConfiguration {
+                        regs: HashMap::new(),
+                    },
+                )
```
ShadowCurse, line 884: You pass the CpuConfiguration, but where will the CpuTemplate come from?
Also this "query registers first, apply template later" approach really does not for arm, because you don't know what registers to query in the first place.

ShadowCurse:
I like how x86 module is 208 lines and aarch64 is 11

src/vmm/src/guest_config/templates/mod.rs
mattschlebusch: I plan to pull out the MSRs loop into its own function as the same logic will be applicable for aarch config.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,592 @@
...
+    use std::str::FromStr;
+
+    use serde::de::Error as SerdeError;
+    use serde::{Deserialize, Deserializer};
+
+    use crate::guest_config::cpuid::cpuid_ffi::KvmCpuidFlags;
+    use crate::guest_config::cpuid::{
+        AmdCpuid, Cpuid, CpuidEntry, CpuidKey, CpuidRegisters, IntelCpuid,
+    };
+    use crate::guest_config::templates::{deserialize_bitmask, BitMask, Error, RegisterModifier};
```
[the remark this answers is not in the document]
mattschlebusch: Why the strange question? :-) It is a personal preference to have it made more explicitly as well, and will make it a bit easier to refactor if need be.

mattschlebusch:
> I like how x86 module is 208 lines and aarch64 is 11

Yep, CPUID is quite a bit more complicated. Although apply_template() is not yet implemented.

> Why clone at all? You can disassemble host_config into parts

This instance of X86_64 config is not of the host config, but the modified config for the guest.

src/vmm/src/builder.rs
mattschlebusch: The registers to query will be based on the addresses in the template. The CPU template will come from the /cpu-config API call which is a different call path. This is a place holder for now to demonstrate where the CPU configuration will be passed into.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,592 @@
...
+            let start = Instant::now();
+            let cpu_config: X86_64CpuTemplate = serde_json::from_str(X86_64_TEMPLATE_JSON)
+                .expect("Failed to deserialize x86_64 CPU template.");
+            let end = Instant::now();
+            assert!(
+                end.duration_since(start).as_micros() < TEMPLATE_DESERIALIZATION_MAX_US,
+                "Deserialization time [{} microseconds] over the threshold [{} microseconds]",
+                end.duration_since(start).as_micros(),
+                TEMPLATE_DESERIALIZATION_MAX_US,
+            );
```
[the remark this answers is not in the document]
mattschlebusch: Hmmm, I can split this out. It is somewhat repetitive, but still good to have a separate test result dedicated to this.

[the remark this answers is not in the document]
ShadowCurse: If you want to have perf tests, then maybe use the `rust perf tests` like we do for snapshots? It is just that here you measure time to deserialize a very specific (and small) cpu template. If we going to measure performance then at least let's use something representative of the real world template.

ShadowCurse:
> > Why clone at all? You can disassemble host_config into parts
> 
> This instance of X86_64 config is not of the host config, but the modified config for the guest.

Mine main point is not to do unnecessary clones.

src/vmm/src/guest_config/templates/mod.rs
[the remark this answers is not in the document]
mattschlebusch: Using an integration test for testing a targeted segment of code does not seem practical or necessary. We can look at performance testing the entire API in the future, which we should be charting anyway, but for this code change, I don't think it is necessary.

mattschlebusch:
> > > Why clone at all? You can disassemble host_config into parts
> > 
> > 
> > This instance of X86_64 config is not of the host config, but the modified config for the guest.
> 
> Mine main point is not to do unnecessary clones.

I want to avoid moving or modifying the host's configuration in any way. I'm cloning the map here to have a dedicated map of memory for the guest configuration. The host's configuration must be borrowed only.

src/vmm/src/guest_config/templates/mod.rs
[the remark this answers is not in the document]
ShadowCurse: I was talking about benches, not integration tests (should have clarified): https://github.com/firecracker-microvm/firecracker/blob/main/src/snapshot/benches/version_map.rs

ShadowCurse:
> I want to avoid moving or modifying the host's configuration in any way. I'm cloning the map here to have a dedicated map of memory for the guest configuration. The host's configuration must be borrowed only.

Well.. the function `apply_template` takes CpuConfig by value, so it takes ownership of it. Also I did not see the use of the original `config.cpuid` hashmap value in the code. It is only used to be cloned into `guest_cpuid_map`. So why not use `config.cpuid` hashmap as a `guest_cpuid_map` without cloning?

Also if you intend to borrow CpuConfig in `apply_template`, what are use planing to use it for after `apply_template` is called?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+
+fn deserialize_u64_from_str<'de, D>(deserializer: D) -> Result<u64, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let number_str = String::deserialize(deserializer)?;
+    let deserialized_number: u64 = if number_str.len() > 2 {
+        match &number_str[0..2] {
+            "0b" => u64::from_str_radix(&number_str[2..], 2),
+            "0x" => u64::from_str_radix(&number_str[2..], 16),
```
kalyazin, line 427: what is the plan now? are we migrating to the tristate notation within this PR or separately?

src/vmm/src/guest_config/x86_64/mod.rs
```diff
@@ -3,12 +3,11 @@
...
 use std::collections::HashMap;
 
-use crate::guest_config::cpuid::Cpuid;
-
 /// CPU configuration for x86_64 CPUs
+#[derive(Clone, Debug)]
 pub struct X86_64CpuConfiguration {
     /// CPUID configuration
-    pub cpuid: Cpuid,
+    // pub cpuid: Cpuid,
```
kalyazin: is this deliberate?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -62,26 +80,197 @@ pub mod x86_64 {
...
+
+                let cpuid_entry_option = guest_cpuid_map.get(&cpuid_key);
+                let mut guest_cpuid_entry = if let Some(entry) = cpuid_entry_option {
+                    entry.clone()
+                } else {
+                    CpuidEntry::default()
+                };
+                guest_cpuid_entry.flags = leaf_mod.flags;
+
+                let (mut mod_eax, mut mod_ebx, mut mod_ecx, mut mod_edx) = (
```
kalyazin: Naming of `leaf_mod` and `mod_eax` doesn't look consistent. Is that on purpose?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -131,11 +320,124 @@ where
 
 #[cfg(test)]
 mod tests {
+    #[cfg(target_arch = "x86_64")]
```
kalyazin, line 459: Can possibly group together x86_64 and aarch64 stuff to avoid duplicating these for every line.

src/vmm/src/guest_config/templates/mod.rs
ShadowCurse: @mattschlebusch This is still relevant in your latest version.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,1043 @@
...
+    pub fn create_guest_cpu_config(
+        _template: &Aarch64CpuTemplate,
+        _host_config: &Aarch64CpuConfiguration,
+    ) -> Result<Aarch64CpuConfiguration, Error> {
+        // TODO
+        Ok(Aarch64CpuConfiguration::default())
+    }
+
+    fn deserialize_u128_from_str<'de, D>(deserializer: D) -> Result<u128, D::Error>
+    where
```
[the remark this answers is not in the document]
mattschlebusch: The numerical JSON ops are quite duplicated. They are split out in this way for the moment because each arch has its own bit-widths essentially. I wanted to introduce more generic types to galvanize the code, but it made the types quite hard to process for a first pass. A TODO for now seems reasonable.

src/vmm/src/vstate/vcpu/mod.rs
```diff
@@ -77,6 +82,11 @@ pub struct VcpuConfig {
     pub smt: bool,
     /// Hard-coded template to use.
     pub cpu_template: CpuFeaturesTemplate,
+    /// Custom configuration for vCPU,
+    #[cfg(target_arch = "x86_64")]
+    pub custom_cpu_config: Option<X86_64CpuConfiguration>,
+    #[cfg(target_arch = "aarch64")]
+    pub custom_cpu_config: Option<Aarch64CpuConfiguration>,
```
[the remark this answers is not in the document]
mattschlebusch: With regards to changing `cpu_template` -> `static_cpu_template`, maintaining this naming change across all the structs will make break the API unfortunately. But I do think in VcpuConfig specifically that this is a good idea. I will add this change now.

src/vmm/src/builder.rs
```diff
@@ -269,6 +285,21 @@ fn create_vmm_and_vcpus(
...
+        // Apply CPU template to create vCPU custom config if available
+        if let Some(template) = &vm.guest_cpu_template {
+            vm.guest_cpu_config = Some(
+                create_guest_cpu_config(
+                    template,
+                    &X86_64CpuConfiguration {
+                        cpuid: Cpuid::try_from(RawCpuid::from(vm.supported_cpuid().clone()))
+                            .unwrap(),
+                        msrs: get_msr_values(vcpus.get(0), template),
+                    },
```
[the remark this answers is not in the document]
mattschlebusch: The supported configuration of the host is maintained in memory (in the main branch), any considerations or changes to this is not in scope for now. I do believe that keeping the supported configuration in memory isn't necessary and is something we can update in future. As a result though, the X86_64CpuConfiguration here consists of the host's values. `create_guest_cpu_config` is only borrowing the values.

src/vmm/src/guest_config/aarch64/mod.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
mattschlebusch: I sort of agree with your point in that the term "wrapper" hints at not introducing new types. I'm not sure this really introduces a new type, given that `Cpuid` and register collections exist in kvmbindings already. I am just trying to provide a more concise type system for these various loose components.

src/vmm/src/guest_config/templates/mod.rs
mattschlebusch: I do have some tweaks downstream that aggregates this a bit better.

src/vmm/src/guest_config/templates/mod.rs
mattschlebusch: So the `deserialize_u64_from_str` in particular is for addresses, not the masks.

With regards to the tristate notation, I thought it was best to include it here. You can see an example of the tristate based template in the unit tests.

src/vmm/src/guest_config/templates/mod.rs
mattschlebusch: [Vm contains the supported configuration](https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/vstate/vm.rs#L146) which is where this data is passed from. Short of making another system call to retrieve a new instance of data, or moving the supported configuration out of Vm, the cloning will be necessary as far as I can tell?

ShadowCurse: From I can see in this PR you store the `guest_cpu_config` in `Vm`, not `host_cpu_config`. So why clone `host_cpu_config` if it is instantly thrown away? 
In any case we can leave this as it is and return to this question in final polishing steps.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -57,37 +76,182 @@ pub mod x86_64 {
...
+                cpuid: Cpuid::Amd(AmdCpuid(guest_cpuid_map)),
+                msrs: guest_msrs_map,
+            })
+        } else if host_config.cpuid.intel().is_some() {
+            Ok(X86_64CpuConfiguration {
+                cpuid: Cpuid::Intel(IntelCpuid(guest_cpuid_map)),
+                msrs: guest_msrs_map,
+            })
+        } else {
+            Err(Error::Internal)
```
kalyazin: can we be more descriptive here?

src/vmm/src/guest_config/templates/mod.rs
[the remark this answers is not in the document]
kalyazin: How will we be tracking this? A TODO in the code or something else?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -119,32 +283,212 @@ pub mod x86_64 {
...
     const X86_64_TEMPLATE_JSON: &str = r#"{
+        "cpuid_modifiers": [
+            {
+                "leaf": "0x80000001",
+                "subleaf": "0x0007",
+                "flags": 0,
+                "modifiers": [
+                    {
+                        "register": "eax",
+                        "mask": {
```
kalyazin: I see now what caused my confusion last time. Do we need this intermediate change that is removed in a later commit?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -217,4 +559,105 @@ mod tests {
...
+        let kvm_result = Kvm::new();
+        assert!(kvm_result.is_ok(), "Unable to access KVM");
+
+        // Create descriptor KVM resource's file descriptor
+        let vm_fd_result = kvm_result.as_ref().unwrap().create_vm();
+        assert!(vm_fd_result.is_ok(), "{}", vm_fd_result.unwrap_err());
+
+        let kvm_cpuid_result = kvm_result
+            .unwrap()
+            .get_supported_cpuid(KVM_MAX_CPUID_ENTRIES);
```
kalyazin, line 975: I'm not really a fan of relying on KVM in unit tests. Can we create `X86_64CpuConfiguration` by hand, so it is determininstic?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -545,17 +551,32 @@ mod tests {
     fn test_serialization_lifecycle() {
         #[cfg(target_arch = "x86_64")]
         {
+            let start = Instant::now();
             let cpu_config: X86_64CpuTemplate = serde_json::from_str(X86_64_TEMPLATE_JSON)
                 .expect("Failed to deserialize x86_64 CPU template.");
+            let end = Instant::now();
+            assert!(
+                end.duration_since(start).as_micros() < TEMPLATE_DESERIALIZATION_MAX_US,
```
kalyazin, line 785: Do you believe this template is representative for the duration test? Would it be a good idea to have some range (maybe 2 or 3 templates of different lengths)?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -394,6 +396,10 @@ mod tests {
     #[cfg(target_arch = "x86_64")]
     use crate::guest_config::x86_64::X86_64CpuConfiguration;
 
+    const TEMPLATE_DESERIALIZATION_MAX_US: u128 = 200;
```
kalyazin: where are these numbers coming from?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -369,6 +422,50 @@ pub mod aarch64 {
 
         Ok(deserialized_number)
     }
+
+    /// Deserialized a composite bitmask string into a value-mask pair
```
kalyazin, line 501: nit: s/Deserialized/Deserialize/ ?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -290,13 +292,64 @@ pub mod x86_64 {
             mask.value
         }
     }
+
+    /// Deserialized a composite bitmask string into a value-mask pair
```
kalyazin, line 289: nit: Related to my naming comment earlier: commet `value-mask` doesn't look consistent with `BitMask { filter, value )` in the code.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -718,6 +869,98 @@ mod tests {
...
+            let template_json = template_json_str_result.unwrap();
+
+            let deserialization_result = serde_json::from_str::<X86_64CpuTemplate>(&template_json);
+            assert!(deserialization_result.is_ok());
+            assert_eq!(template, deserialization_result.unwrap());
+        }
+    }
+
+    #[test]
+    fn test_bitmask_width() {
```
kalyazin, line 888: Could we add a description of what this test is supposed to do?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+            })?
+        };
+
+        Ok(deserialized_number)
+    }
+
+    /// Bit mask to adjust targeted bits to the value,
+    /// applied by the filter only.
+    #[derive(Debug, Deserialize, Eq, PartialEq)]
+    pub struct BitMask {
```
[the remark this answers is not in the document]
mattschlebusch: I think the `RegisterModifier` already describes what you propose akin to the operation? It contains a target address and a mask to be applied. Value seems appropriate to me because it is the value that we want the result to be, and the filter decides which bits from the value are relevant.

If you are not comfortable with the name `value` in this context we can change that?

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -57,37 +76,182 @@ pub mod x86_64 {
...
+    }
+
+    /// CPU template with modifiers to be applied on
+    /// top of an existing configuration to generate
+    /// the guest configuration to be used.
+    pub fn create_guest_cpu_config(
+        template: &X86_64CpuTemplate,
+        host_config: &X86_64CpuConfiguration,
+    ) -> Result<X86_64CpuConfiguration, Error> {
+        let mut guest_msrs_map: HashMap<u32, u64> = HashMap::new();
```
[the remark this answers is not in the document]
mattschlebusch: I don't believe a `BTreeMap` is required for Cpuid. Maybe there was consideration for a case involving iteration in which having deterministic ordering was useful, but I am not aware of it.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -57,37 +76,182 @@ pub mod x86_64 {
...
+            let cpuid_key = CpuidKey {
+                leaf: leaf_mod.leaf,
+                subleaf: leaf_mod.subleaf,
+            };
+
+            let cpuid_entry_option = guest_cpuid_map.get(&cpuid_key);
+            let mut guest_cpuid_entry = if let Some(entry) = cpuid_entry_option {
+                entry.clone()
+            } else {
+                CpuidEntry::default()
```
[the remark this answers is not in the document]
mattschlebusch: This aspect of the behaviour needs to be clear for sure. I have gone this route to be the most resilient to failure, but it might be more desirable to actually fail the start of the guest.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -57,37 +76,182 @@ pub mod x86_64 {
...
+            };
+            guest_cpuid_entry.flags = leaf_mod.flags;
+
+            let (mut mod_eax, mut mod_ebx, mut mod_ecx, mut mod_edx) = (
```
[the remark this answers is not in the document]
mattschlebusch: No particular reason. I didn't see it as an inconsistency.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -57,37 +76,182 @@ pub mod x86_64 {
...
+            };
+
+            guest_cpuid_map.insert(cpuid_key, guest_cpuid_entry);
+        }
+
+        // Apply MSR modifiers
+        for modifier in &template.msr_modifiers {
+            guest_msrs_map.insert(
+                modifier.addr,
+                apply_mask(host_config.msrs.get(&modifier.addr), &modifier.mask),
```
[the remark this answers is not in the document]
mattschlebusch: As things stand now, similar to CPUID, it will add the entry for the guest. I also think this is something we should consider synchronously validating against though.

src/vmm/src/guest_config/templates/mod.rs
mattschlebusch: Strange, this error was removed already. Can you refresh to confirm?

src/vmm/src/guest_config/templates/mod.rs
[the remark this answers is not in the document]
kalyazin: No, I'm good with the `value`. What I don't like is the `value` is a part of the `BitMask`. A modifier contains both a `value` and a `mask`. A `mask` should not contain a `value`, they are complementary to each other.

src/vmm/src/guest_config/templates/mod.rs
mattschlebusch: So the current `Cpuid` implementation is strongly tied to the vendor and building by hand is quite cumbersome and even trickier to debug if there is an issue. We do have hard-coded templates that I am fairly sure will do the job, as long as we don't try to use this for anything more than unit tests (the hard coded templates are Intel specific). I will swap this out to remove the KVM calls and see how it goes.

src/vmm/src/guest_config/templates/mod.rs
kalyazin: Indeed, it seems to have gone. Probably in later commits.

src/vmm/src/guest_config/templates/mod.rs
mattschlebusch: They are arbitrary, as long as they were not in the 100s of millseconds I was happy. This testing in the unit test won't be long term though.

src/vmm/src/guest_config/templates/mod.rs
kalyazin: Sounds like a good idea.

tests/integration_tests/build/test_coverage.py
```diff
@@ -17,9 +17,9 @@
 # Checkout the cpuid crate. In the future other
 # differences may appear.
 if utils.is_io_uring_supported():
-    COVERAGE_DICT = {"Intel": 78.20, "AMD": 79.09, "ARM": 82.19}
+    COVERAGE_DICT = {"Intel": 78.08, "AMD": 78.92, "ARM": 81.88}
```
kalyazin, line 20: is it clear to us what causes the drop and how to recover?

src/vmm/src/guest_config/templates/mod.rs
mattschlebusch: Yes, the changes are quite big, but yield more benefit when there are multiple crates involved.

src/vmm/src/guest_config/templates/mod.rs
```diff
@@ -0,0 +1,1030 @@
...
+            assert!(
+                guest_config_result.is_ok(),
+                "{}",
+                guest_config_result.unwrap_err()
+            );
+            assert_eq!(guest_config_result.unwrap(), host_configuration);
+        }
+
+        // TODO - Aarch64 test
+        // #[cfg(target_arch = "aarch64")]
```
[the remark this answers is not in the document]
mattschlebusch: For the empty template this should be fine. but without hardcoded templates, adding the tests balloons this PR further.

src/vmm/src/builder.rs
```diff
@@ -315,6 +360,20 @@ fn create_vmm_and_vcpus(
...
+    // TODO - Use a created vCPU to retrieve MSR values
+    Default::default()
+}
+
+#[cfg(target_arch = "aarch64")]
+#[allow(unused)]
+fn get_reg_values(_vcpu: Option<&Vcpu>, _template: &Aarch64CpuTemplate) -> HashMap<u64, u128> {
+    // TODO - Use a created vCPU to retrieve register values
+    Default::default()
+}
```
ShadowCurse, line 375: Why `Option<&Vcpu>`? How can you retrieve anything if you don't have vcpu?

mattschlebusch: This is a fair question. So the idea here was to be more flexible, in allowing use to just write whole register values via the template, in which case when the bitmap is applied we can just use the value from the map. I'm on the fence about being more rigorous in requiring the lookup for a bitmap to be usable, but I don't have a strong preference at this stage. Wdy?

kalyazin: @mattschlebusch could you elaborate on your point? Do you mean, in a very special case where the bitmap doesn't have any wildcards ("x"), we don't need any KVM calls and can build the output directly from the template? Do you think this warrants making Vcpu optional? Where would the logic deciding on pass vs not pass the real Vcpu reside?

ShadowCurse (approved):

tests/integration_tests/build/test_coverage.py
mattschlebusch: I have added tests to smooth these numbers out. The ARM thresholds still have a slight drop, some of this is related to error handling that is in place in case of system errors that are hard to replicate in a unit test environment. The drop is very small though, for x86, coverage has been increased.

kalyazin (approved):

PR #3513, opened 2023-03-08
fix: updated 5.10-no-sve guest kernel config

ShadowCurse:
## Changes
Added ftrace options to the config.

## Reason
Without them kernel hangs on g3 with 4.14 host kernel.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


resources/guest_configs/microvm-kernel-arm64-5.10-no-sve.config
```diff
@@ -2892,7 +2892,11 @@ CONFIG_HAVE_FTRACE_MCOUNT_RECORD=y
 CONFIG_HAVE_SYSCALL_TRACEPOINTS=y
 CONFIG_HAVE_C_RECORDMCOUNT=y
 CONFIG_TRACING_SUPPORT=y
-# CONFIG_FTRACE is not set
+CONFIG_FTRACE=y
```
kalyazin, line 2895: Any idea why this matters?

resources/guest_configs/microvm-kernel-arm64-5.10-no-sve.config
```diff
@@ -2892,7 +2892,11 @@ CONFIG_HAVE_FTRACE_MCOUNT_RECORD=y
...
+CONFIG_BOOTTIME_TRACING=y
+CONFIG_FUNCTION_TRACER=y
+CONFIG_FUNCTION_GRAPH_TRACER=y
+CONFIG_DYNAMIC_FTRACE=y
```
sudanl0, line 2899: do you need all the configs or just a few like CONFIG_FTRACE/CONFIG_FUNCTION_TRACER. It would be good to have more info on how you found these configs and if there is an external link which explains why/how these configs affect the bootup.

resources/guest_configs/microvm-kernel-arm64-5.10-no-sve.config
ShadowCurse: Not really. From what I know about ftrace is that it adds some `nop` instructions at the start of each function and those `nop` bytes can in the future be changed to represent a call to the tracer. I guess these `nop` instructions give time buffer for something to happend (initialize or smth) and because of this everything works. I can investigate further, but the amout of work in respect to usefulness of that work is questionable.

resources/guest_configs/microvm-kernel-arm64-5.10-no-sve.config
ShadowCurse: I was trying different options and this is the smallest set that I could find.

resources/guest_configs/microvm-kernel-arm64-5.10-no-sve.config
kalyazin: Fair enough. How did you end up disabling it?

resources/guest_configs/microvm-kernel-arm64-5.10-no-sve.config
sudanl0: ok, do you think if it would be a good idea to create a TODO somewhere to find out how these options helped? Maybe someone who has time could try to figure this out?

sudanl0:
Changes/Reason/Checklist need an update.

resources/guest_configs/microvm-kernel-arm64-5.10-no-sve.config
ShadowCurse: Disabling what? Ftrace? It is disabled by default i.e. the bytes are set `nop` at boot time.

kalyazin: I'm just trying to reconstruct how you first arrived at the config without ftrace (what was your default config) and why you decided to enable it.

resources/guest_configs/microvm-kernel-arm64-5.10-no-sve.config
ShadowCurse: Not sure. 4.14 is an ancient kernel and investigating the underlying reason may be not reasonable in terms of effort to result ratio. What deserves a TODO is a task of removing this kernel once we drop 4.14 from supported list of kernels. But who knows when that will happen.

resources/guest_configs/microvm-kernel-arm64-5.10-no-sve.config
ShadowCurse: The default was our default 5.10 config. This ftrace hack I found when I was investigation this issue in the first place (bach when I was gathering baselines for c7g)

kalyazin: Right, so we enable ftrace not because it was enabled in some other config that we use, but empirically, because it just solves the boot problem?

ShadowCurse: Not because it slows the boot process (even thought it does), but because it solves the problem. Most likely by slowing boot, but I am not a grey beard kernel engineer to tell for sure, not yet at least.

kalyazin: (I didn't say anything about slowing the boot process, btw)
By all my questions I tried to gauge the dirtiness of this hack. It is certainly fine to apply it if it serves the purpose. I would only suggest adding this in the commit message for future reference. Maybe something in line with "ftrace is a completely random thing, but for some reason it does something that makes the kernel boot successfully. We don't know what specifically, but that's what we need, and we can't afford investigate further at this point".

ShadowCurse: Sorry in my mind "solves the boot problem" became "slows the boot process"...
I will add this comment to commit message.

kalyazin (dismissed):

zulinx86 (approved):

PR #3523, opened 2023-03-11
Add error handling when screen/firecracker processes have not launched successfully

mattschlebusch:
## Changes

* Check to make sure the screen process in integration tests is started successfully.
* Check the parent process' status with a retry added to confirm the screen process is up and running.
* Try-and-except blocks added at multiple stages of the screen sessions startup for clearer error handling.

## Reason

Motivated by a [build error in Buildkite for Firecracker's CI pipeline](https://buildkite.com/firecracker/firecracker-pr/builds/1013#01865063-028d-473b-8ffa-346e1388e5c1).

The `screen` application is used to boot up Firecracker in a detached mode, and is used to launch the Firecracker process (or Jailer as a wrapper) for integration tests to be run against. This results in two processes being launched, the screen process, and the binary application's process (that is a child process of the screen process). It is rare that this boot up logic fails, but when it does can be an opaque issue.

Error handling for:
* If a runtime or system error occurs, due to the child process crashing very quickly, ending the parent screen process as well.
* If there is a race condition of the screen process launching, but the forking of the process for the child process has not completed yet.

Both cases can result in  a missing/empty `children` file (`/proc/{PPID}/task/{PPID}/children`) for the parent process. This is what can lead to the error seen in the CI pipeline, `ValueError: invalid literal for int() with base 10: ''`.

## License Acceptance

[`CONTRIBUTING.md`][3].

### DIFF
--- a/tests/framework/utils.py
+++ b/tests/framework/utils.py
@@ -21,6 +21,7 @@ from retry.api import retry_call
 
 from framework.defs import MIN_KERNEL_VERSION_FOR_IO_URING
 
+FLUSH_CMD = 'screen -S {session} -X colon "logfile flush 0^M"'
 CommandReturn = namedtuple("CommandReturn", "returncode stdout stderr")
 CMDLOG = logging.getLogger("commands")
 GET_CPU_LOAD = "top -bn1 -H -p {} -w512 | tail -n+8"
@@ -760,17 +761,22 @@ def start_screen_process(screen_log, session_name, binary_path, binary_params):
         delay=1,
     ).group(1)
 
-    binary_clone_pid = int(
-        open("/proc/{0}/task/{0}/children".format(screen_pid), encoding="utf-8")
-        .read()
-        .strip()
-    )
+    # Make sure the screen process launched successfully
+    # As the parent process for the binary.
+    screen_ps = psutil.Process(int(screen_pid))
+    wait_process_running(screen_ps)
 
     # Configure screen to flush stdout to file.
-    flush_cmd = 'screen -S {session} -X colon "logfile flush 0^M"'
-    run_cmd(flush_cmd.format(session=session_name))
+    run_cmd(FLUSH_CMD.format(session=session_name))
 
-    return screen_pid, binary_clone_pid
+    children_count = len(screen_ps.children())
+    if children_count != 1:
+        raise RuntimeError(
+            f"Failed to retrieve child process id for binary {binary_path}. "
+            f"screen session process had [{children_count}]"
+        )
+
+    return screen_pid, screen_ps.children()[0].pid
 
 
 def guest_run_fio_iteration(ssh_connection, iteration):
@@ -790,3 +796,13 @@ def check_filesystem(ssh_connection, disk_fmt, disk):
     cmd = "fsck.{} -n {}".format(disk_fmt, disk)
     exit_code, _, stderr = ssh_connection.execute_command(cmd)
     assert exit_code == 0, stderr.read()
+
+
+@retry(delay=0.5, tries=5)
+def wait_process_running(process):
+    """Wait for a process to run.
+
+    Will return successfully if the process is in
+    a running state and will otherwise raise an exception.
+    """
+    assert process.is_running()


tests/framework/utils.py
```diff
@@ -760,16 +762,38 @@ def start_screen_process(screen_log, session_name, binary_path, binary_params):
...
-
     # Configure screen to flush stdout to file.
     flush_cmd = 'screen -S {session} -X colon "logfile flush 0^M"'
     run_cmd(flush_cmd.format(session=session_name))
 
+    # Make sure the screen process launched successfully
+    # As the parent process for the binary.
+    try:
+        retry_call(
+            confirm_process_running,
```
pb8o, line 773: Could we use psutil here?

```python
screen_ps = psutil.Process(screen_pid)
screen_ps.is_running()
```

tests/framework/utils.py
```diff
@@ -760,16 +762,38 @@ def start_screen_process(screen_log, session_name, binary_path, binary_params):
...
+        ) from retry_exception
+
+    # While the process id was found, there may have been an error that caused
+    # the child process for the binary to fail, resulting in an empty children file.
+    try:
+        binary_clone_pid = int(
+            open("/proc/{0}/task/{0}/children".format(screen_pid), encoding="utf-8")
+            .read()
+            .strip()
+        )
```
pb8o, line 791: and then here:

```python
if len(screen_ps.children()) != 1:
```

tests/framework/utils.py
```diff
@@ -760,16 +762,38 @@ def start_screen_process(screen_log, session_name, binary_path, binary_params):
...
 
-    binary_clone_pid = int(
-        open("/proc/{0}/task/{0}/children".format(screen_pid), encoding="utf-8")
-        .read()
-        .strip()
-    )
-
     # Configure screen to flush stdout to file.
     flush_cmd = 'screen -S {session} -X colon "logfile flush 0^M"'
     run_cmd(flush_cmd.format(session=session_name))
```
pb8o, line 767: Maybe we could capture the output / error code here, and add it to the raised exceptions, to shed some light on why it fails.

mattschlebusch: I don't see why not. It is currently being deliberately ignored, I'm not 100% sure why, but I will tweak it and see how CI does.

tests/framework/utils.py
mattschlebusch: I'll give this a go, although there is no retry here in case the process hasn't come up yet. But I think this implementation will be much cleaner.

mattschlebusch:
Using psutils had some [test failures](https://buildkite.com/firecracker/firecracker-pr/builds/1641#0186dcc2-e3a3-47ce-8d95-2a65a540e797). I think it's possible that there is a retry with a delay to see if that passes tests.

tests/framework/utils.py
```diff
@@ -760,17 +761,29 @@ def start_screen_process(screen_log, session_name, binary_path, binary_params):
...
+            f"Failed to start screen process to launch binary {binary_path}"
+        )
+
+    if len(screen_ps.children()) != 1:
+        raise RuntimeError(
+            f"Failed to retrieve child process id for binary {binary_path}"
+        )
 
-    return screen_pid, binary_clone_pid
+    return screen_pid, screen_ps.children()[0]
```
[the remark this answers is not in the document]
mattschlebusch: I saw this a bit late :-)

tests/framework/utils.py
```diff
@@ -760,17 +761,29 @@ def start_screen_process(screen_log, session_name, binary_path, binary_params):
...
 
-    binary_clone_pid = int(
-        open("/proc/{0}/task/{0}/children".format(screen_pid), encoding="utf-8")
-        .read()
-        .strip()
-    )
+    # Make sure the screen process launched successfully
+    # As the parent process for the binary.
+    screen_ps = psutil.Process(int(screen_pid))
+    for _ in range(5):
```
dianpopa, line 767: Why 5? Since we are using the pid as a filter, we expect to only have 1 entry here always, right?

mattschlebusch: Yea, the loop here is just a retry of `screen_ps.is_running()` until the process reaches a running state. The reason the buildkite job crashed is because the `/proc/{}/task/{}/children` file existed but was empty. The retry here is to make sure the screen process process is up and running before checking for the child process.

tests/framework/utils.py
```diff
@@ -760,17 +761,29 @@ def start_screen_process(screen_log, session_name, binary_path, binary_params):
...
+        if not screen_ps.is_running():
+            time.sleep(1)
+        else:
+            break
```
pb8o, line 771: You could add an `else` here to print something 

```python
for ...
else:
    print("screen_ps still not running. bailing out after 5 tries and hoping for the best")

**Review by pb8o — APPROVED** [review]

File: tests/framework/utils.py (hunk 5 above)
Comment (dianpopa) [reply tests/framework/utils.py:767]: ah ok. For cases like this one, we have a heavy use in the python scripts for retry calls.

You can see here a bunch of them: https://github.com/firecracker-microvm/firecracker/blob/main/tests/framework/microvm.py#L499

You would want something like this:
```
@retry(delay=0.5, tries=5)
def wait_screen_ps(screen_ps):
  assert screen_ps.is_running()
```

tests/framework/utils.py
mattschlebusch: The approach of using the `@retry wait_process_running()` will log a warning for each attempt that fails.

tests/framework/utils.py
mattschlebusch: That is cleaner, thanks!

dianpopa (dismissed):

pb8o (approved):

tests/framework/utils.py
dianpopa: Indeed.

pb8o (approved):

PR #3543, opened 2023-03-20
Improve the robustness of closing open file descriptors in jailer

### LINKED ISSUE #3542: [Bug] Firecracker uses up 100% CPU after an upgrade from v1.1.2 to v1.3.1
# Describe the bug

After upgrading firecracker from v1.1.2 to v1.3.1 the VMM instantly uses up 100% of the CPU while not creating an unix api socket. Firecracker eventually manages to create the socket, but it takes a few minutes. After attaching to the faulty firecracker process with `strace` it is revealed that firecracker is stuck in a loop trying to close every pid in existence:
image

## To Reproduce

Run firecracker inside a privileged docker container based on `node:16-bullseye` with `/dev/kvm` mounted inside.  v1.1.2 works fine,  v1.3.1 doesn't work at all.

## Expected behaviour

The Firecracker process starts without using 100% of the CPU and creates an unix socket.

## Environment
 - Firecracker version: v1.3.1
- Host kernel: 6.1.12-arch1-1 Archlinux
- Guest kernel: 5.6
- Rootfs used: Ext4, ubuntu 22.04
- Architecture: x86_64, AMD Ryzen 9 5950X

## Additional context
I was working on my startup which is a project for creating and managing dev environments inside Firecracker Microvm's. The project works fine using firecracker v1.1.2, after I upgraded the firecracker version the project broke.

## Checks

gorbak25:
## Changes
Closes all open FDs in a single syscall. In case close_range is not available then fallback to the old approach. In case the fallback fails then discover open FDs by reading from /proc/self/fds.

## Reason
Fixes #3542 - a bug where firecracker would end up spending minutes starting which arises when running on systems with OPEN_MAX set to a very high number.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -587,6 +587,7 @@ name = "jailer"
 version = "1.3.0"
 dependencies = [
  "libc",
+ "nix 0.26.2",
  "regex",
  "thiserror",
  "utils",
@@ -765,6 +766,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "nix"
+version = "0.26.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bfdda3d196821d6af13126e40375cdf7da646a96114af134d5f417a9a1dc8e1a"
+dependencies = [
+ "bitflags",
+ "cfg-if",
+ "libc",
+ "memoffset 0.7.1",
+ "pin-utils",
+ "static_assertions",
+]
+
+[[package]]
 name = "nom"
 version = "7.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -822,6 +837,12 @@ name = "peeking_take_while"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "19b17cddbe7ec3f8bc800887bab5e717348c95ea2ca0b1bf0837fb964dc67099"
+
+[[package]]
+name = "pin-utils"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
 name = "plotters"
@@ -1111,6 +1132,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "static_assertions"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
+
+[[package]]
 name = "subtle"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1212,7 +1239,7 @@ dependencies = [
  "bitflags",
  "cfg-if",
  "libc",
- "nix",
+ "nix 0.23.1",
  "thiserror",
  "userfaultfd-sys",
 ]
--- a/src/jailer/Cargo.toml
+++ b/src/jailer/Cargo.toml
@@ -10,6 +10,7 @@ license = "Apache-2.0"
 
 [dependencies]
 libc = "0.2.117"
+nix = { version = "0.26.2", features = ["dir"] }
 regex = { version = "1.5.5", default-features = false, features = ["std"] }
 thiserror = "1.0.32"
 
--- a/src/jailer/src/main.rs
+++ b/src/jailer/src/main.rs
@@ -6,6 +6,7 @@ mod chroot;
 mod env;
 mod resource_limits;
 use std::ffi::{CString, NulError, OsString};
+use std::os::unix::prelude::AsRawFd;
 use std::path::{Path, PathBuf};
 use std::{env as p_env, fs, io, process, result};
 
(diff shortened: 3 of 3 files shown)


JonathanWoollett-Light:
Firecracker currently has full supported for 4.14 ([kernel-policy](https://github.com/firecracker-microvm/firecracker/blob/main/docs/kernel-policy.md)).

I believe the [`close_range`](https://man7.org/linux/man-pages/man2/close_range.2.html) syscall was introduced in 5.9 in https://github.com/torvalds/linux/blob/master/include/uapi/linux/close_range.h.

Please correct me if I'm wrong, but otherwise I think this was what prevented us using this approach previously.

gorbak25:
@JonathanWoollett-Light I just pushed a V2 of this patch. Changes compared to V1:
- Each of the fall backs has a separate test
- If OPEN_MAX is bellow 1024 then I iterate to 1024 anyway, I decreased the maximum threshold from  2**20 to 2**16
- I replaced `std::fs::read_dir` with `nix::dir::Dir::open` in order to not close the dirfd descriptor twice - otherwise the rust code would panic. Is using the nix crate appropriate here? I added it cause it is a peer dependency anyway.

gorbak25:
@JonathanWoollett-Light renamed the patch to a more appropriate name

JonathanWoollett-Light:
As separate Firecracker and Jailer versions are released for the 5.10 and 4.14 kernels respectively I think we can do this check at compile time.

If we include a `build.rs` in the `jailer` with:
```rust
fn main() {
    let process = std::process::Command::new("uname").arg("-r").output().unwrap();
    assert!(process.status.success());
    let version = std::str::from_utf8(&process.stdout).unwrap();
    let mut iter = version.split(".");
    let major = iter.next().unwrap().parse::<u32>().unwrap();
    let minor = iter.next().unwrap().parse::<u32>().unwrap();

    // Firecracker and the jailer officaily supports 2 kernel versions 5.10 and 4.14.
    //
    // Some improved functionality is supported in the 5.10 version that is not supported in 4.14,
    // to accmodate this we set conditionl compilation flags for these versions.
    //
    // In developing Firecracker and the jailer a newer kernel may be used, to support this we don't
    // match the versions exactly.
    match (major,minor) {
        (5.., 10..) => println!("cargo:rustc-cfg=kernel=\"5.10\""),
        (4, 14..) => println!("cargo:rustc-cfg=kernel=\"4.14\""),
        _ => unreachable!()
    }
}
```
and then change `sanitize_process` to:
```rust
fn sanitize_process() {
    // First thing to do is make sure we don't keep any inherited FDs
    // other that IN, OUT and ERR.
    #[cfg(kernel="5.10")]
    {
        // SAFETY: The values are constant and valid.
        unsafe {
            libc::syscall(
                libc::SYS_close_range,
                3,
                libc::c_uint::MAX,
                libc::CLOSE_RANGE_UNSHARE,
            )
        }
    }
    #[cfg(kernel="4.14")]
    {
        // SAFETY: Always safe.
        let fd_limit = i32::try_from(unsafe { libc::sysconf(libc::_SC_OPEN_MAX) }).unwrap();
        // Close all file descriptors excluding 0 (STDIN), 1 (STDOUT) and 2 (STDERR).
        for fd in 3..fd_limit {
            // SAFETY: Safe because close() cannot fail when passed a valid parameter.
            unsafe {
                libc::close(fd);
            }
        }
    }

    // Cleanup environment variables
    clean_env_vars();
}
```
it should fix this.

Does this solution appear good to you?

gorbak25:
@JonathanWoollett-Light any feedback on what approach this PR should take?

JonathanWoollett-Light:
I think the best approach here is to initially address the issue with the simplest solution.

This 1st commit could implement a change like:
```rust
fn sanitize_process() {
    // First thing to do is make sure we don't keep any inherited FDs
    // other that IN, OUT and ERR.

    // Try using the close_range syscall to close all open FDs in the range of 4..UINT_MAX
    // SAFETY: The parameters are valid constants.
    let res = unsafe { libc::syscall(
        libc::SYS_close_range,
        3,
        libc::c_uint::MAX,
        libc::CLOSE_RANGE_UNSHARE,
    ) };

    // If `SYS_close_range` fails as it is not supported (as is the case on a pre 5.9 kernel), we
    // instead close by iterating through all possible file descriptors.
    match res {
        0 => (),
        // SAFETY: Always safe.
        -1 => match unsafe { *libc::__errno_location } {
            0 => (),
            // The syscall failed due to not being supported.
            libc::ENOSYS => {
                // SAFETY: Always safe.
                let fd_limit = i32::try_from(unsafe { libc::sysconf(libc::_SC_OPEN_MAX) }).unwrap();
                // Close all file descriptors excluding 0 (STDIN), 1 (STDOUT) and 2 (STDERR).
                for fd in 3..fd_limit {
                    // SAFETY: Safe because close() cannot fail when passed a valid parameter.
                    unsafe {
                        libc::close(fd);
                    }
                }
            },
            _ => panic!()
        },
        _ => unreachable!()
    }

    // Cleanup environment variables
    clean_env_vars();
}
```

I am unsure about pulling in Nix with the fallback this will need to be considered. I think using Nix to iterate through `/proc` should be a separate commit, if we want to get the 1st commit merged more quickly it should be a separate PR.

gorbak25:
@JonathanWoollett-Light will come back to this PR on wednesday, I'm on a trip without my laptop right now.

dianpopa:
hello @gorbak25,

Thanks a lot for the proposed fix. The changes look reasonable to me in general, there are a couple of things I would like to adjust:
* firstly, I would like to drop the check that looks at the `_SC_OPEN_MAX ` variable and keep the one that calls into `close_range` and then fall back to using `proc/self/fd`. I tested a bit the performance of each method and for a system with `_SC_OPEN_MAX` 65535 I obtained around 47ms that are always spent. For kernels < 5.9, the behavior would be exactly the same as it was before the regression.
* also I would like to improve a bit on the error propagation of the code

Since this issue introduces a customer-facing regression, I would like to speed up the merging process so I opened a [separate PR](https://github.com/firecracker-microvm/firecracker/pull/3595) that starts from your commit and contains the proposed adjustments. Your contribution is mentioned there and kept you as the commit author! Please take a look at this alternate PR and review.

Thanks a lot again for looking into this!

dianpopa:
Closing in favor of https://github.com/firecracker-microvm/firecracker/pull/3595. The fix will be present in v1.4.0 (our next release).
Thanks @gorbak25 again for the contribution!

PR #3557, opened 2023-03-24
Warn on `clippy::cast_possible_truncation`

### LINKED ISSUE #3197: Use `clippy::cast_possible_truncation`
Implement the `clippy::cast_possible_truncation`.

### LINKED ISSUE #3198: `Use clippy::cast_possible_wrap`

JBYoshi:
Adds checks for `clippy::cast_possible_truncation` as per #3197.

## Changes

I added `-Dclippy::cast_possible_truncation` to the Rust configuration and wrapped most unchecked casts with `try_from(...).unwrap()`. In a few cases like tests or where the truncation behavior was intentional, I updated the types or added explicit allow markers.

## Reason

Fixes #3197.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/devices/src/virtio/block/device.rs
```diff
@@ -180,7 +180,8 @@ impl DiskProperties {
         // The config space is little endian.
         let mut config = Vec::with_capacity(CONFIG_SPACE_SIZE);
         for i in 0..CONFIG_SPACE_SIZE {
-            config.push(u8::try_from(self.nsectors >> (8 * i)).unwrap());
+            #[allow(clippy::cast_possible_truncation)]
+            config.push((self.nsectors >> (8 * i)) as u8);
         }
         config
```
JonathanWoollett-Light: `as` only wraps in release mode, in debug it will panic. How do you know this conversion needs to wrap?

JBYoshi: My understanding of this code is that it's writing `CONFIG_SPACE_SIZE` bytes of `self.nsectors` into `config`. When `i = 0`, only the least-significant byte is added; when `i=1`, the next byte is added.

The original version of this code also used `as u8` (I changed it in my original commit but undid it here), so I don't see any reason why there should be any new panics in this part of the code.

JBYoshi:
I've updated to the latest main commit. All tests are passing on my end, though the coverage for the case without io_uring has decreased due to the increased line count in io_uring. My dev machine uses an Intel chip so I updated the coverage for that; I don't have an AMD or ARM machine to test on, so those coverage numbers may still need to be fixed.

JBYoshi:
Updated to fix AMD test coverage and ARM compile errors.

JonathanWoollett-Light:
Please squash these changes into 1 commit.

JBYoshi:
Squashed, and I fixed a few more bugs in the build.

JBYoshi:
Fixed a few more styling points and updated coverage numbers.

I'm noticing that the Git commit message lint is failing because my sign-off line is longer than the maximum line length. Is there a way to work around this?

src/devices/src/virtio/block/device.rs
JonathanWoollett-Light: https://doc.rust-lang.org/reference/expressions/operator-expr.html#numeric-cast

> Casting from a larger integer to a smaller integer (e.g. u32 -> u8) will truncate

Apologies I was incorrect, for now this is fine.

JBYoshi: I only saw this just now; I've added that comment.

JBYoshi:
Updated to add comments to the other "allow" markers.

JBYoshi:
@JonathanWoollett-Light Updated.

JonathanWoollett-Light:
@JBYoshi  Could you please rebase and fix the clippy lints.

JBYoshi:
@JonathanWoollett-Light I've rebased to the latest changes as far as I can test. The coverage numbers may still need to be updated though.

JonathanWoollett-Light:
@JBYoshi Apologies for a slow response, could you please rebase and resolve the merge conflicts.

JBYoshi:
Rebased onto the latest changes.

JBYoshi:
Updated to readjust the coverage numbers and fix a few lint checks from the aarch64 code.

JonathanWoollett-Light (dismissed):

JBYoshi:
Updated to fix ARM build errors.

JonathanWoollett-Light (approved):

src/vmm/src/arch/x86_64/mptable.rs
```diff
@@ -433,12 +433,15 @@ mod tests {
     fn cpu_entry_count_max() {
         let cpus = MAX_SUPPORTED_CPUS + 1;
```
roypat, line 434: I think here we can change the type of the constant to `u8` to avoid casts altogether.

roypat (dismissed):
Thank you for your contribution, and sorry for taking so long to actually have a look.

I am a bit concerned about the approach taken here in parts of just replacing `as` casts with `try_from(...).unwrap()`, I think what we were hoping for when we opened https://github.com/firecracker-microvm/firecracker/issues/3197 was that only very few `.unwrap()`s would be introduced. There are some areas where panicking instead of truncating is worse (for example in the metric component, where a slightly wrong metric is annoying, but a panic escalates this into an availability issues). 

What makes this a bit hard to assess right now is that you also fixed the lint for our unittests. So far the approach we took with these is to just annotated test modules with `#![allow(clippy::*)}`. Could you maybe move all the unittest fixes into their own commit, so that its easier to review just the changes to production code? 

I think a few `.unwrap()` can be avoided by instead just declaring types correctly (I've commented on one such example). Some constants, such as `MAC_ADDR_LEN` it might make sense to change the type from `usize` to `u8` and instead of `try_from`ing the usize -> u8 conversion, using infallible conversions to for u8 -> usize. 

For places where we can't get around using `.unwrap()` (which to me would seem like maybe something bigger is wrong with our architecture? For example, why is `SyscallReturnCode` a wrapper around i32 when `libc::syscall` always returns i64?), I think we need some short comment for each, or err on the side of maintaining the status quo via targetted `#[allow(...)]` (for example in the metrics submodule.

Could you have a look at this?

JBYoshi:
I've started to look at some of those cases. So far, there are a few situations that I'm not fully settled on:

* In any struct marked `#[derive(Versionize)]`, [all of the size constants need to be`usize` constants with no type conversions](https://github.com/firecracker-microvm/versionize_derive/blob/f0a7194791c5cbeeb55aecfa6e60cca6f2c4168a/src/fields/struct_field.rs#L130-L142). So for example, if we were to change `MAC_ADDR_LEN` to `u8`, then the `bytes` field of the `MacAddr` struct would have to change from `[u8; MAC_ADDR_LEN]` to `[u8; MAC_ADDR_LEN as usize]`. However, putting `as` in there causes the `Versionize` derivation to fail.
* A lot of the 64-bit architecture-specific code uses `usize` and `u64` interchangeably. According to the [documentation on the `GuestAddress` type](https://github.com/rust-vmm/vm-memory/blob/main/src/guest_memory.rs#L102-L110) in the `vm-memory` crate, some ARM systems can virtualize 64-bit systems on 32-bit hypervisors, in which case a `usize` value wouldn't be able to fit a guest address. I'm considering prioritizing `u64` in more cases because of this.
* For the `SyscallReturnCode` case, I'm wondering if it's a design problem on the `vmm-sys-util` package's end. Most system calls return an `int`, but a few return pointers (such as the `mmap` system call). When `vmm-sys-util` does system calls, it still uses `libc::syscall`, but [it converts the returned `i64` value to an `i32` value using `as c_int`](https://github.com/rust-vmm/vmm-sys-util/blob/f7025a7d24f24f8b6ceacebf378df67b11c38be3/src/linux/aio.rs#L94), the same as what the old code here did.

roypat:
Heyo, thanks for looking into these!

> In any struct marked `#[derive(Versionize)]`, [all of the size constants need to be`usize` constants with no type conversions](https://github.com/firecracker-microvm/versionize_derive/blob/f0a7194791c5cbeeb55aecfa6e60cca6f2c4168a/src/fields/struct_field.rs#L130-L142). So for example, if we were to change `MAC_ADDR_LEN` to `u8`, then the `bytes` field of the `MacAddr` struct would have to change from `[u8; MAC_ADDR_LEN]` to `[u8; MAC_ADDR_LEN as usize]`. However, putting `as` in there causes the `Versionize` derivation to fail.

Mh, that is indeed unfortunate. Would you be interested in fixing this directly in `versionize_derive`? To me it seems that is no reason why the derive macro restricts array lengths to constants and integer literals, so a fix could be to change to `array_len` parameter of `generate_array_deserializer` to simply be `syn::Expr` should do the trick :thinking: 

> A lot of the 64-bit architecture-specific code uses `usize` and `u64` interchangeably. According to the [documentation on the `GuestAddress` type](https://github.com/rust-vmm/vm-memory/blob/main/src/guest_memory.rs#L102-L110) in the `vm-memory` crate, some ARM systems can virtualize 64-bit systems on 32-bit hypervisors, in which case a `usize` value wouldn't be able to fit a guest address. I'm considering prioritizing `u64` in more cases because of this.

Firecracker only supports 64 bit architectures, so 32bit architectures (either on the guest or host) is not something we need to consider here. I'd expect that a lot of these conversions will be hard to get rid of, so we can do [what rust-lang refuses to](https://users.rust-lang.org/t/cant-convert-usize-to-u64/6243) and write a `fn u64_to_usize(i: u64) -> usize` utility function (or the other way around) that's cfg'd for `pointer_width = 64` and annotated with an `allow` for this lint. 

>   For the `SyscallReturnCode` case, I'm wondering if it's a design problem on the `vmm-sys-util` package's end. Most system calls return an `int`, but a few return pointers (such as the `mmap` system call). When `vmm-sys-util` does system calls, it still uses `libc::syscall`, but [it converts the returned `i64` value to an `i32` value using `as c_int`](https://github.com/rust-vmm/vmm-sys-util/blob/f7025a7d24f24f8b6ceacebf378df67b11c38be3/src/linux/aio.rs#L94), the same as what the old code here did.

Mhh, yeah, it seems like the return values from libc are a hodgepodge of inconsistency. But I agree that maybe we can find a fix upstream here, if you're interested. Maybe generalize the current `pub struct SyscallReturnCode(pub c_int);` into `pub struct SyscallReturnCode<T>(pub T);` with the current `into_result` and `into_empty_result` impls macro-generated for all common types returned by libc functions? @JonathanWoollett-Light and me can help you with PRs for vmm-sys-util!

Let me know what you think! If you feels that's more than you signed up for when you started this work, no worries, we're always happy to merge as much of the work as you want to do here, and leave the rest for a future PR :)

JBYoshi:
> Would you be interested in fixing this directly in versionize_derive? To me it seems that is no reason why the derive macro restricts array lengths to constants and integer literals, so a fix could be to change to array_len parameter of generate_array_deserializer to simply be syn::Expr should do the trick 🤔

I'm setting up some PRs for that now.

> Firecracker only supports 64 bit architectures, so 32bit architectures (either on the guest or host) is not something we need to consider here. I'd expect that a lot of these conversions will be hard to get rid of, so we can do [what rust-lang refuses to](https://users.rust-lang.org/t/cant-convert-usize-to-u64/6243) and write a fn u64_to_usize(i: u64) -> usize utility function (or the other way around) that's cfg'd for pointer_width = 64 and annotated with an allow for this lint.

That sounds like a valid solution; I'll go with that.

> Mhh, yeah, it seems like the return values from libc are a hodgepodge of inconsistency. But I agree that maybe we can find a fix upstream here, if you're interested. Maybe generalize the current pub struct SyscallReturnCode(pub c_int); into pub struct SyscallReturnCode<T>(pub T); with the current into_result and into_empty_result impls macro-generated for all common types returned by libc functions? @JonathanWoollett-Light and me can help you with PRs for vmm-sys-util!

I'll look into that. Since it is an external crate, I'd want to make sure those changes don't break compatibility with previous versions.

---

Since this is getting to be a large PR, what would you think about breaking these changes up into separate, smaller PRs? Many of the lint fixes should be able to be merged even before we turn on the `clippy::cast_possible_truncation` warnings. I think that would also make the individual updates easier to review. The downside would be that someone could undo some of the lint fixes without realizing that it's a lint fix; if that's problematic, I could also group some of the fixes as separate commits within this PR.

roypat:
> Since this is getting to be a large PR, what would you think about breaking these changes up into separate, smaller PRs? Many of the lint fixes should be able to be merged even before we turn on the `clippy::cast_possible_truncation` warnings. I think that would also make the individual updates easier to review. The downside would be that someone could undo some of the lint fixes without realizing that it's a lint fix; if that's problematic, I could also group some of the fixes as separate commits within this PR.

Either approach is fine to me. If you notice you run into needing to do lots of manual rebases, feel free to split parts of this out into smaller PRs that we can merge incrementally. You can always enable the lint at module level in those PRs, so that unrelated code changes do not introduce regressions. 

Grouping this one into separate commits would be very useful for review though (especially to untangle the changes in test-code from the changes in production code, as we only need to scrutinize the latter, really).

roypat:
Merged and released the versionize patches!

JBYoshi:
I've mostly finished going back through everything. I just have one main question remaining.

VirtIO devices all have an `interrupt_status` field that's currently a `usize` value. However, whenever it's exposed to the VM, it's always done as a `u32` value ([read](https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/devices/virtio/mmio.rs#L241), [write](https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/devices/virtio/mmio.rs#L297)). I could change the type everywhere to `u32`, which would remove the need for any type conversions there. However, that would also involve changing the type in [the snapshot data format](https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/devices/virtio/persist.rs#L95-L109), which apparently would involve bumping the snapshot version.

Would changing that type be worth the effort of bumping the version? If so, is there a process for updating those versions?

JBYoshi:
I've pushed my updated changes, broken down by both whether they're production or test code as well as the type of change. While I'm waiting to hear back on the interrupt_status changes, I've switched the PR back to draft status, but feel free to leave other comments as well.

JBYoshi:
I've updated the code to fix some CI errors. There are still a few remaining, but most of those seem to be connected to the snapshot changes.

src/vmm/src/devices/virtio/device.rs
```diff
@@ -54,14 +54,14 @@ pub enum IrqType {
 /// Helper struct that is responsible for triggering guest IRQs
 #[derive(Debug)]
 pub struct IrqTrigger {
-    pub(crate) irq_status: Arc<AtomicUsize>,
+    pub(crate) irq_status: Arc<AtomicU32>,
```
roypat, line 57: According to the virtio spec (Section 4.2.2), this value is either 0 or 1, so I'm not sure why this was defined as `AtomicUsize` (a quick glance at the git history didnt tell me much either). I think this change is fine, and the require snapshot version bump here is not a problem. 

@bchalios do you have any idea what might've been the reason for making this a usize?

.cargo/config
```diff
@@ -5,6 +5,7 @@ rustflags = [
   "-Dclippy::ptr_as_ptr",
   "-Dclippy::undocumented_unsafe_blocks",
   "-Dclippy::cast_lossless",
+  "-Dclippy::cast_possible_truncation",
```
roypat, line 8:
```suggestion
  "-Wclippy::cast_possible_truncation",
```

roypat (commented):
I've gone through and left some preliminary comments. For the interrupt_status thing, a snapshot version bump is fine (we have to do one with each minor firecracker version anyway). The "process" for these is to modify the `VirtioDeviceState` structure and give it a new `u32` field for `interrupt_status`. The old field should be marked with `version(end=1)` and the new one with `version(start=2)`. Then you just need to have a translation function between version 1 and 2 (which will do a cast). You can look at vmm::devices::virtio::net::persist for a reference (there we have to do something similar with a `mac_addr` field). Lastly, you need to bump the version of the `VirtioDeviceState` structure in `vmm::version_map`.

For the generated constants, I personally don't see a problem with adjusting their types. It would only become slightly annoying if we ever wanted to regenerate the bindings (e.g. to add more constants). Maybe someone else can chime in here?

roypat:
I've also trigged the CI for you, seems like there is still some minor issues on ARM :(

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -480,7 +480,7 @@ impl KvmVcpu {
         // per million beacuse it is common for TSC frequency
         // to differ due to calibration at boot time.
         let diff = (i64::from(self.get_tsc_khz()?) - i64::from(state_tsc_freq)).abs();
-        Ok(diff > (f64::from(state_tsc_freq) * TSC_KHZ_TOL).round() as i64)
+        Ok(diff as f64 > (f64::from(state_tsc_freq) * TSC_KHZ_TOL).round())
```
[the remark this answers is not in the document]
kalyazin: I would share this concern. To be honest, I can't see why we need to use f64 here at all. It could be replaced with mul/div.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -390,7 +390,8 @@ impl Net {
             METRICS.net.rx_fails.inc();
             0
         } else {
-            self.rx_bytes_read as u32
+            // Safe to unwrap because a frame must be smaller than 2^16 bytes.
+            u32::try_from(self.rx_bytes_read).unwrap()
```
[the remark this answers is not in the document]
bchalios: We define the `MAX_BUFFER_SIZE` as 2^16, so this should be fine.

src/vmm/src/devices/virtio/device.rs
bchalios: Indeed, it's either 0 or 1 but what's its length? I couldn't find that in the spec.

kalyazin: >

Interrupt status
Reading from this register returns a bit mask of events that caused the device interrupt to be asserted. The following events are possible:
- Used Buffer Notification: bit 0 - the interrupt was asserted because the device has used a buffer in at least one of the active virtual queues.
- Configuration Change Notification: bit 1 - the interrupt was asserted because the configuration of the device has changed.

roypat: I think its implicit by virtue of being mapped to a single MMIO register, and each MMIO register is 32bits wide (e.g. the interrupt_status is at 0x60, and at 0x64 a different field starts, page 59 of the 1.1 spec, Table 4.2)

kalyazin: https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-1460002
search for 
```
InterruptStatus
0x60
R
```

roypat: Ah, you're right, I misread. Its not always 0 or 1, but instead only the bits 0 and 1 have meaning. So its a 2 bit field, thanks Nikita! :D

bchalios: I did some research. Qemu stores this as a `u8`: https://github.com/qemu/qemu/blob/005ad32358f12fe9313a4a01918a55e60d4f39e5/include/hw/virtio/virtio.h#L109

In any case, we write back to the guest a 32bit value: https://github.com/firecracker-microvm/firecracker/blob/e2a9ba42d8fe72590d9ca8858be3d40cfcf7e3ab/src/vmm/src/devices/virtio/mmio.rs#L249, so we should be fine with `AtomicU32`

roypat:
> For the generated constants, I personally don't see a problem with adjusting their types. It would only become slightly annoying if we ever wanted to regenerate the bindings (e.g. to add more constants). Maybe someone else can chime in here?

Alright, I've talked this through with the team, and we've decided we dont want to change the size of constants (and there's also no real way bindgen could get this "right" for us). So that part is good to go!

We've also released a new version of vmm-sys-util to crates.io

src/dumbo/src/pdu/udp.rs
```diff
@@ -143,7 +143,8 @@ impl<'a, T: NetworkBytesMut + Debug> UdpDatagram<'a, T> {
 
         packet.bytes.shrink_unchecked(len);
         packet.payload_mut().copy_from_slice(payload);
-        packet.set_len(len as u16);
+        // Safe to unwrap because of the size check above
+        packet.set_len(u16::try_from(len).unwrap());
```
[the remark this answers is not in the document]
JBYoshi: Done.

src/dumbo/src/tcp/connection.rs
```diff
@@ -879,7 +875,8 @@ impl Connection {
                 return Err(WriteNextError::PayloadBufTooLarge);
             }
 
-            let payload_end = payload_seq + Wrapping(read_buf.len() as u32);
+            // Safe to unwrap because we check the size.
+            let payload_end = payload_seq + Wrapping(u32::try_from(read_buf.len()).unwrap());
```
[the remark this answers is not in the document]
JBYoshi: Done.

src/vmm/src/rate_limiter/mod.rs
```diff
@@ -163,8 +163,12 @@ impl TokenBucket {
...
             // time_adjustment is at most time_delta, and since time_delta <= u64::MAX, this cast is
             // fine
-            self.last_update += Duration::from_nanos(time_adjustment as u64);
-            self.budget = std::cmp::min(self.budget.saturating_add(tokens as u64), self.size);
+            self.last_update += Duration::from_nanos(u64::try_from(time_adjustment).unwrap());
+            self.budget = std::cmp::min(
+                self.budget
+                    .saturating_add(u64::try_from(tokens).unwrap_or(u64::MAX)),
+                self.size,
+            );
```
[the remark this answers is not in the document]
JBYoshi: I've switched to use that.

src/vmm/src/arch/aarch64/cache_info.rs
```diff
@@ -248,11 +248,14 @@ fn mask_str2bit_count(mask_str: &str) -> Result<u16, CacheInfoError> {
...
-            .map_err(|err| {
-                CacheInfoError::InvalidCacheAttr("shared_cpu_map".to_string(), err.to_string())
-            })?
-            .count_ones() as u16;
+        bit_count += u16::try_from(
+            u32::from_str_radix(s_zero_free, 16)
+                .map_err(|err| {
+                    CacheInfoError::InvalidCacheAttr("shared_cpu_map".to_string(), err.to_string())
+                })?
+                .count_ones(),
```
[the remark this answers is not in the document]
JBYoshi: I think it's just because Rust typically uses `u32` as a de-facto default value for integers.

.cargo/config
JBYoshi: Done.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -901,7 +903,8 @@ mod tests {
         {
             // The frequency difference is within tolerance.
             let mut state = orig_state.clone();
-            state.tsc_khz = Some(state.tsc_khz.unwrap() + (TSC_KHZ_TOL / 2.0).round() as u32);
+            state.tsc_khz =
+                Some(state.tsc_khz.unwrap() + TSC_KHZ_TOL_NUMERATOR / TSC_KHZ_TOL_DENOMINATOR / 2);
```
JBYoshi, line 906: While rewriting this code, I noticed that `TSC_KHZ_TOL` is small enough that `TSC_KHZ_TOL_NUMERATOR / TSC_KHZ_TOL_DENOMINATOR / 2` will round to 0; I don't think that's intended behavior for the test code. This looks like it's been part of the existing code for some time; `(TSC_KHZ_TOL / 2.0).round()` would also round to 0. The same applies to two pieces of code below.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -910,19 +913,21 @@ mod tests {
...
 
     #[test]
-    #[allow(clippy::cast_sign_loss)] // always positive, no u32::try_from(f64)
+    #[allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)] // always positive, no u32::try_from(f64)
     fn test_set_tsc() {
         let (vm, vcpu, _) = setup_vcpu(0x1000);
         let mut state = vcpu.save_state().unwrap();
-        state.tsc_khz = Some(state.tsc_khz.unwrap() + (TSC_KHZ_TOL * 2.0).round() as u32);
+        state.tsc_khz =
+            Some(state.tsc_khz.unwrap() + TSC_KHZ_TOL_NUMERATOR / TSC_KHZ_TOL_DENOMINATOR * 2);
```
JBYoshi, line 928: Same as above, just `* 2` instead of `/ 2`.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -910,19 +912,20 @@ mod tests {
         {
             // The frequency difference is over the tolerance.
             let mut state = orig_state;
-            state.tsc_khz = Some(state.tsc_khz.unwrap() + (TSC_KHZ_TOL * 2.0).round() as u32);
+            state.tsc_khz =
+                Some(state.tsc_khz.unwrap() + TSC_KHZ_TOL_NUMERATOR / TSC_KHZ_TOL_DENOMINATOR * 2);
```
JBYoshi, line 916: Same as above, just `* 2` instead of `/ 2`.

JBYoshi (commented):
I've addressed this set of comments, including the remaining ARM build problems. I did find a likely bug in the test code while working through this; let me know if you'd like me to do anything about it.

JonathanWoollett-Light:
I've approved the CI, seems their are some test errors, mostly related to error messages.

src/vmm/src/vstate/vcpu/x86_64.rs
roypat: ... yeah, you're right, this makes very little sense to me, too.  The `* 2` case is actually even more baffling, because there's a comment that says "The frequency difference is over the tolerance.", but then the assertion still asserts that we're not above the tolerance (which we are indeed not, because we add 0 to our value). I'm assuming that these tests meant to do something like

```rs
state = Some(state.tsc_khz.unwrap() + state.tsc_khz.unwrap() * (TSC_KHZ_TOL / 2))
```

and that the second test is meant to have its assertion inverted. Nice catch!

cc @kalyazin

kalyazin: that is indeed a good finding. What Patrick proposes makes sense to me.

JBYoshi:
I've updated to fix the remaining errors - it was just changing `#[version(end = 1)]` to `#[version(end = 2)]` (looks like `end` is exclusive here), and then fixing one of the Git commit messages to match the styling for that.

src/vmm/src/vstate/vcpu/x86_64.rs
JonathanWoollett-Light: When I was refactoring this code I also encountered this, I thought it looked weird at the time.

roypat: I don't really think we need a new test, we can just fix this one in this PR

JonathanWoollett-Light:
A lot of the tests are currently failing see https://buildkite.com/firecracker/firecracker-pr/builds/6502#018af5ff-4f6c-44ff-a9f4-1e2833d0b94b.

JBYoshi:
I found the problem with the performance tests (I had mixed up `process_startup_time_us` and `process_startup_time_cpu_us` in some of the metrics) - should be fixed now. I also fixed the THC tolerance tests that were originally always rounding to 0.

roypat (approved):
Great work, thank you!

JonathanWoollett-Light (approved):

JonathanWoollett-Light:
Great work, this turned into a surprisingly herculean effort, thank you for all your work.

PR #3568, opened 2023-03-27
virtio-rng integration tests

bchalios:
## Changes

Add integration tests for `virtio-rng` device.

## Reason

Test end-to-end that virtio-rng is working. Also test rate limiting and functionality after resuming from a snapshot

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/functional/test_rng.py
+++ b/tests/integration_tests/functional/test_rng.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 """Tests for the virtio-rng device"""
 
+from pathlib import Path
+
 import pytest
 
 from framework.artifacts import NetIfaceConfig
@@ -76,3 +78,47 @@ def test_rng_present(test_microvm_with_rng):
     _start_vm_with_rng(vm)
 
     check_entropy(vm.ssh)
+
+
+@pytest.mark.skipif(
+    INSTANCE_TYPE == "c7g.metal" and HOST_KERNEL == "4.14",
+    reason="c7g requires no SVE 5.10 kernel",
+)
+def test_rng_snapshot(test_microvm_with_rng, microvm_factory):
+    """
+    Test that a virtio-rng device is functional after resuming from
+    a snapshot
+
+    @type: functional
+    """
+
+    vm = test_microvm_with_rng
+    _start_vm_with_rng(vm)
+
+    check_entropy(vm.ssh)
+
+    mem_path = Path(vm.jailer.chroot_path()) / "test.mem"
+    snapshot_path = Path(vm.jailer.chroot_path()) / "test.snap"
+    vm.pause_to_snapshot(
+        mem_file_path=mem_path.name,
+        snapshot_path=snapshot_path.name,
+    )
+    assert mem_path.exists()
+
+    new_vm = microvm_factory.build()
+    new_vm.spawn()
+    iface = NetIfaceConfig()
+    new_vm.create_tap_and_ssh_config(
+        host_ip=iface.host_ip,
+        guest_ip=iface.guest_ip,
+        netmask_len=iface.netmask,
+        tapname=iface.tap_name,
+    )
+    new_vm.ssh_config["ssh_key_path"] = vm.ssh_config["ssh_key_path"]
+    new_vm.restore_from_snapshot(
+        snapshot_vmstate=snapshot_path,
+        snapshot_mem=mem_path,
+        snapshot_disks=[vm.rootfs_file],
+    )
+
+    check_entropy(new_vm.ssh)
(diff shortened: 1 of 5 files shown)


tests/integration_tests/functional/test_rng.py
```diff
@@ -122,3 +123,111 @@ def test_rng_snapshot(test_microvm_with_rng, microvm_factory):
...
+    vm = test_microvm_with_rng
+    _start_vm_with_rng(vm, rate_limiter)
+
+    size = rate_limiter["bandwidth"]["size"]
+    expected_kbps = 10 * size / 1000.0
+    # Run the test asking twice the set limit, so that we make
+    # sure we exercise the rate limiter. Otherwise, i.e. if we ask
+    # less bytes than the size of the token bucket we will not ever
+    # get rate limited and achieve the maximum possible throughput
+    _check_entropy_rate_limited(vm.ssh, 2 * size * 10, expected_kbps)
```
roypat, line 233: Maybe add a comment about where the 10 comes from (I think its 1000/refill_time, in which case shouldn't be actually state it as that? Right now, if the parameterization of the test changes, it will break - also, same algebraic simplifcation as above is possible then :))

tests/integration_tests/functional/test_rng.py
```diff
@@ -122,3 +123,111 @@ def test_rng_snapshot(test_microvm_with_rng, microvm_factory):
...
+
+def _rate_limiter_id(rate_limiter):
+    """
+    Helper function to return a name for the rate_limiter to be
+    used as an id for parametrized tests.
+    """
+    size = rate_limiter["bandwidth"]["size"]
+    refill_time = rate_limiter["bandwidth"]["refill_time"]
+
+    return "{} KB/sec".format(size / (refill_time / 1000.0) / 1000.0)
```
roypat, line 202: nit: size / (refill_time / 1000) / 1000 = (size * 1000 / refill_time) / 1000 = size / refill_time (since size is in bytes and refill_time is in millseconds, both units are off by the same factor of 1000 and we get kb/s)

tests/integration_tests/functional/test_rng.py
```diff
@@ -0,0 +1,78 @@
...
+    # If the guest kernel has been built with the virtio-rng module
+    # the device should exist in the guest filesystem but we should
+    # not be able to get random numbers out of it.
+    cmd = "test -e /dev/hwrng"
+    ecode, _, _ = vm.ssh.execute_command(cmd)
+    assert ecode == 0
+
+    cmd = "dd if=/dev/hwrng of=/dev/null bs=10 count=1"
+    ecode, _, _ = vm.ssh.execute_command(cmd)
+    assert ecode == 1
```
roypat, line 63: nit: I wonder if just checking the exit code to be 1 might hide some other error condition (e.g. the exit code is 1 not because the device doesn't exist, but some other error). Although since this is the negative test, its not that important.

tests/integration_tests/functional/test_rng.py
bchalios: You are right! Adjusting

tests/integration_tests/functional/test_rng.py
bchalios: I re-worked all of this, because it was sloppy. I now just pass `size` and execute `dd` with that size and `count = 100`. My assumption is that running any request 100 times should average out any outliers of the rate limiter. WDYT?

tests/integration_tests/functional/test_rng.py
bchalios: Yes, I could check the `stderr` maybe, but the output would still be a standard libc string format, which again could come from anything :) Given this is a negative test I didn't think it was really worth it.

tests/integration_tests/functional/test_rng.py
```diff
@@ -122,3 +123,111 @@ def test_rng_snapshot(test_microvm_with_rng, microvm_factory):
...
+    raise Exception("Unknown units")
+
+
+def _process_dd_output(out):
+    """
+    Parse the output of `dd` and return the achieved throughput in
+    KB/sec
+    """
+    report = out.splitlines()[-1].split(" ")
+    (value, units) = (report[-2], report[-1])
```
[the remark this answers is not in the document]
roypat: Thank you!

tests/integration_tests/functional/test_rng.py
roypat: Yeah, I agree

tests/integration_tests/functional/test_rng.py
bchalios: You are right, I was just fixing in my head the value of `refill_time` in `100 msec`. Will update it.

roypat (approved):

kalyazin (approved):

PR #3576, opened 2023-03-29
Update rust toolchain to 1.68.2

roypat:
## Changes

- Update rust toolchain
- Appease new clippy lints wrt derivable `Default` implementations and elidable lifetimes
- Add `sched_yield` to the seccomp allow list, as with 1.67.0, the rust standard library changed the backend for `mpmc` channels (which we use through `std::sync::mpsc` for inter-thread communication) to use crossbeams implementatation, which uses `sched_yield` as part of its spin-locking logic [[1]]

[1]: https://github.com/rust-lang/rust/blame/cf32b9de1e8f66526c36ad2927458558d2e81093/library/std/src/sync/mpmc/utils.rs#L130

## Reason

backport of #3576

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/mmds/src/data_store.rs
```diff
@@ -19,18 +19,13 @@ pub struct Mmds {
 }
 
 /// MMDS version.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Deserialize, Serialize)]
 pub enum MmdsVersion {
+    #[default]
```
pb8o, line 24: out of scope, but shouldn't the default be V2?

JonathanWoollett-Light:
All the changes from these commits are required to pass CI, as a result of the rust update, I think it would make sense in this case for these to all be in 1 commit, so we minimize commits which don't pass CI.

resources/seccomp/aarch64-unknown-linux-musl.json
pb8o, line 1: Did we check it is used by all the threads?

src/mmds/src/data_store.rs
roypat: Yeah, this surprised me too, but the old manual `Default` implementation returned `V1`. The `default` impl seemingly is only used in test anyway though

resources/seccomp/aarch64-unknown-linux-musl.json
roypat: Yeah, all threads use channels to communicate. Api <-> VMM to send instructions to VMM thread, and VMM <-> Vcpu for things such as pausing, resuming, etc.

pb8o: Thanks!

bchalios:
> All the changes from these commits are required to pass CI, as a result of the rust update, I think it would make sense in this case for these to all be in 1 commit, so we minimize commits which don't pass CI.

I don't think we want to go that way. We currently do not ensure that all commits pass CI (see code coverage changes, binary size changes, etc.). Also, I see value in having these changes in different commits. It is much easier to review and I identify issues later on.

roypat:
> > All the changes from these commits are required to pass CI, as a result of the rust update, I think it would make sense in this case for these to all be in 1 commit, so we minimize commits which don't pass CI.
> 
> I don't think we want to go that way. We currently do not ensure that all commits pass CI (see code coverage changes, binary size changes, etc.). Also, I see value in having these changes in different commits. It is much easier to review and identify issues later on.

ah whoops, missed this earlier. I agree with Babis here. Having it broken into logical pieces makes cherry-picking/backporting a lot easier. For example, the coverage adjustment is different for every release branch, but the seccomp changes applied cleanly to all of them. With clippy, there was no knowing if I could cherry-pick from here (which I could), or whether old branches would need different patches.

JonathanWoollett-Light (approved):

PR #3595, opened 2023-04-06
jailer: fix perf regression when closing open FDs

### LINKED ISSUE #3542: [Bug] Firecracker uses up 100% CPU after an upgrade from v1.1.2 to v1.3.1

dianpopa:
## Changes

Fixes the mechanism for closing open FDs in jailer which was
based on the SC_OPEN_MAX system constant. Such an approach can lead
to bad performance when this value is very high.

The method for closing file descriptors is now chosen based on the environment:
 1. we try to call into the close_range syscall (available on kernels >=5.9)
 2. we fallback to reading from /proc/self/fd (for kernels <5.9)
    
 ## Reason

Fixes #3542 - a bug where firecracker would end up spending minutes starting which arises when running on systems with OPEN_MAX set to a very high number.

Thanks @gorbak25 for reporting the issue and proposing the fix!

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -522,6 +522,7 @@ name = "jailer"
 version = "1.3.0"
 dependencies = [
  "libc",
+ "nix 0.26.2",
  "regex",
  "thiserror",
  "utils",
@@ -679,6 +680,18 @@ dependencies = [
  "cfg-if",
  "libc",
  "memoffset",
+]
+
+[[package]]
+name = "nix"
+version = "0.26.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bfdda3d196821d6af13126e40375cdf7da646a96114af134d5f417a9a1dc8e1a"
+dependencies = [
+ "bitflags",
+ "cfg-if",
+ "libc",
+ "static_assertions",
 ]
 
 [[package]]
@@ -953,6 +966,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "static_assertions"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
+
+[[package]]
 name = "subtle"
 version = "2.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1054,7 +1073,7 @@ dependencies = [
  "bitflags",
  "cfg-if",
  "libc",
- "nix",
+ "nix 0.23.1",
  "thiserror",
  "userfaultfd-sys",
 ]
--- a/src/jailer/src/main.rs
+++ b/src/jailer/src/main.rs
@@ -6,15 +6,18 @@ mod chroot;
 mod env;
 mod resource_limits;
 use std::ffi::{CString, NulError, OsString};
+use std::os::unix::prelude::AsRawFd;
 use std::path::{Path, PathBuf};
 use std::{env as p_env, fs, io, process, result};
 
 use utils::arg_parser::{ArgParser, Argument, Error as ParsingError};
+use utils::syscall::SyscallReturnCode;
 use utils::validators;
 
 use crate::env::Env;
 
 const JAILER_VERSION: &str = env!("FIRECRACKER_VERSION");
+
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     #[error("Failed to parse arguments: {0}")]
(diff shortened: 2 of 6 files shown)


src/jailer/src/main.rs
```diff
@@ -235,21 +244,71 @@ pub fn readln_special<T: AsRef<Path>>(file_path: &T) -> Result<String> {
...
+    )
+    .map_err(|e| Error::DirOpen(path.to_string(), e.to_string()))?;
+
+    let dirfd = dir.as_raw_fd();
+    let mut c = dir.iter();
+
+    while let Some(Ok(path)) = c.next() {
+        let file_name = path.file_name();
+        let fd_str = file_name.to_str().map_err(Error::UTF8Parsing)?;
+        println!("fd_str is {}", fd_str);
```
luminitavoicu, line 280: Leftover?

src/jailer/src/main.rs
```diff
@@ -235,21 +244,71 @@ pub fn readln_special<T: AsRef<Path>>(file_path: &T) -> Result<String> {
...
+    SyscallReturnCode(unsafe {
+        libc::syscall(
+            libc::SYS_close_range,
+            3,
+            libc::c_uint::MAX,
+            libc::CLOSE_RANGE_UNSHARE,
+        )
+    } as libc::c_int)
+    .into_empty_result()
+    .map_err(Error::CloseRange)
```
luminitavoicu, line 259: Nit: do we need this dedicated error since we discard the error when this call fails and we have a fallback mechanism?

src/jailer/src/main.rs
```diff
@@ -235,21 +244,71 @@ pub fn readln_special<T: AsRef<Path>>(file_path: &T) -> Result<String> {
...
+
+// Closes all FDs other than 0 (STDIN), 1 (STDOUT) and 2 (STDERR)
+fn close_inherited_fds() -> Result<()> {
+    // The approach we take here is to firstly try to use the close_range syscall
+    // which is available on kernels > 5.9.
+    // We than fallback to using /proc/sef/fd to close open fds.
+    if close_fds_by_close_range().is_err() {
+        close_fds_by_reading_proc()?;
+    }
+    Ok(())
```
luminitavoicu, line 301: nit:
```suggestion
    if close_fds_by_close_range().is_err() {
        close_fds_by_reading_proc()
    }
```

src/jailer/src/main.rs
```diff
@@ -235,21 +244,71 @@ pub fn readln_special<T: AsRef<Path>>(file_path: &T) -> Result<String> {
...
         }
     }
+    Ok(())
+}
+
+// Closes all FDs other than 0 (STDIN), 1 (STDOUT) and 2 (STDERR)
+fn close_inherited_fds() -> Result<()> {
+    // The approach we take here is to firstly try to use the close_range syscall
+    // which is available on kernels > 5.9.
+    // We than fallback to using /proc/sef/fd to close open fds.
```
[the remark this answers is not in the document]
andreitraistaru:
```suggestion
    // We then fallback to using /proc/self/fd to close open fds.
```

Cargo.lock
```diff
@@ -681,6 +682,18 @@ dependencies = [
  "memoffset",
 ]
 
+[[package]]
+name = "nix"
+version = "0.26.2"
```
roypat, line 687: Mhh, the entry before this one is about nix 0.23.1 (transitive dependency). Can we do this with nix 0.23 so that we don't pull in the dependency twice??

dianpopa (commented):
@lumivo @atc @roypat I also updated the changelog, can you take another look?
@roypat addressed your comment around the nix dependency

Cargo.lock
roypat: Works for me to do that as a follow up then, no need to block this on it

roypat (approved):

src/jailer/src/main.rs
luminitavoicu: That makes sense. Let's keep it as is.

src/jailer/src/main.rs
```diff
@@ -235,21 +244,70 @@ pub fn readln_special<T: AsRef<Path>>(file_path: &T) -> Result<String> {
...
         }
     }
+    Ok(())
+}
+
+// Closes all FDs other than 0 (STDIN), 1 (STDOUT) and 2 (STDERR)
+fn close_inherited_fds() -> Result<()> {
+    // The approach we take here is to firstly try to use the close_range syscall
+    // which is available on kernels > 5.9.
+    // We then fallback to using /proc/sef/fd to close open fds.
```
andreitraistaru, line 296:
```suggestion
    // We then fallback to using /proc/self/fd to close open fds.
```

andreitraistaru: looks good to me apart from this very small nit

andreitraistaru (approved):

PR #3607, opened 2023-04-13
Fix spurious failure in api_server::tests::test_serve_vmm_action_request

roypat:
The test calls the `serve_vmm_action_request` function and tries to check whether it correctly updates specific metrices related to measuring how long a request took. However, since the VMM side is mocked out in this test (the receiver end of the channel through which the API server communicates is simply pre-filled with VmmData::Empty, and not passed to any implementation that handles the requests), the `serve_vmm_action_request` function only goes through the following steps:

1. Put something into a channel
2. Read a pre-submitted value from another channel
3. Update the metrics

This can happen very fast (below 1us, which is the resolution of our metrics), in which cases the unittests fails at the assertion that tries to check whether the metric of "how long did handling this request take" was correctly set (as this assertion only checks that the metric is not 0).

We fix this by subtracting 1 from the start time inside the test, so make sure that the difference "now - start time" which serve_vmm_action_request uses to compute the metric value is always at least 1.

Prior to this fix, empirical testing has shown a failure rate of around 1 in 3 on m6i.metal instances, with this patch that has gone done to 0 in 100. 

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/api_server/src/lib.rs
```diff
@@ -336,7 +336,12 @@ mod tests {
         let response = api_server.serve_vmm_action_request(Box::new(VmmAction::StartMicroVm), 0);
         assert_eq!(response.status(), StatusCode::BadRequest);
 
-        let start_time_us = utils::time::get_time_us(ClockType::Monotonic);
+        // Since the vmm side is mocked out in this test, the call to serve_vmm_action_request can
+        // complete very fast (under one 1ns, the resolution of our metrics). In these cases, the
```
zulinx86, line 340: nit:
```suggestion
        // complete very fast (under one 1us, the resolution of our metrics). In these cases, the
```

andreitraistaru:
Nice investigation! Looks good to me. Just for my understanding: do we have a reason for why this error seems to be more often reproducible in the last period of time? Have you discovered anything that could increase the reproducing frequency.

src/api_server/src/lib.rs
roypat: ah, very good catch! fixed

zulinx86 (approved):

andreitraistaru (approved):

PR #3618, opened 2023-04-18
Add CPU template benchmarking for JSON ops

mattschlebusch:
## Changes

Add criterion benchmarks for the to/from string JSON operations for the `CustomCpuTemplate` type.

## Reason

To keep a handle on the performance of serializing and deserializing CPU templates in JSON format.

## License Acceptance

[`CONTRIBUTING.md`][3].

### DIFF
--- a/src/vmm/Cargo.toml
+++ b/src/vmm/Cargo.toml
@@ -40,5 +40,9 @@ criterion = { version = "0.4.0", default-features = false }
 device_tree = "1.1.0"
 
 [[bench]]
-name = "main"
+name = "cpu_templates"
+harness = false
+
+[[bench]]
+name = "snapshots"
 harness = false
--- a/src/vmm/benches/cpu_templates.rs
+++ b/src/vmm/benches/cpu_templates.rs
@@ -1,0 +1,37 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Benchmarking cases:
+//   * `CustomCpuTemplate` Deserialization
+
+use std::path::Path;
+
+use criterion::{criterion_group, criterion_main, Criterion};
+use vmm::guest_config::templates::test_utils::TEST_TEMPLATE_JSON;
+use vmm::guest_config::templates::CustomCpuTemplate;
+
+#[inline]
+pub fn bench_deserialize_cpu_template(cpu_template_str: &str) {
+    serde_json::from_str::<CustomCpuTemplate>(cpu_template_str);
+}
+
+pub fn cpu_template_benchmark(c: &mut Criterion) {
+    println!(
+        "Template size (JSON string): [{}] bytes.",
+        TEST_TEMPLATE_JSON.len()
+    );
+
+    c.bench_function("Deserialize custom CPU Template", |b| {
+        b.iter(|| bench_deserialize_cpu_template(TEST_TEMPLATE_JSON))
+    });
+}
+
+criterion_group! {
+    name = cpu_template_benches;
+    config = Criterion::default().sample_size(200).output_directory(Path::new("../../build/vmm_benchmark/cpu_templates"));
+    targets = cpu_template_benchmark
+}
+
+criterion_main! {
+    cpu_template_benches
+}
--- a/src/vmm/benches/cpu_templates.rs
+++ b/src/vmm/benches/cpu_templates.rs
@@ -2,13 +2,20 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 // Benchmarking cases:
-//   * `CustomCpuTemplate` Deserialization
+//   * `CustomCpuTemplate` JSON deserialization
+//   * `CustomCpuTemplate` JSON serialization
 
+use std::mem::size_of_val;
 use std::path::Path;
 
 use criterion::{criterion_group, criterion_main, Criterion};
-use vmm::guest_config::templates::test_utils::TEST_TEMPLATE_JSON;
+use vmm::guest_config::templates::test_utils::{build_test_template, TEST_TEMPLATE_JSON};
 use vmm::guest_config::templates::CustomCpuTemplate;
+
+#[inline]
+pub fn bench_serialize_cpu_template(cpu_template: &CustomCpuTemplate) {
+    serde_json::to_string(cpu_template);
+}
 
 #[inline]
 pub fn bench_deserialize_cpu_template(cpu_template_str: &str) {
@@ -17,12 +24,22 @@ pub fn bench_deserialize_cpu_template(cpu_template_str: &str) {
 
 pub fn cpu_template_benchmark(c: &mut Criterion) {
     println!(
-        "Template size (JSON string): [{}] bytes.",
+        "Deserialization test - Template size (JSON string): [{}] bytes.",
         TEST_TEMPLATE_JSON.len()
     );
 
-    c.bench_function("Deserialize custom CPU Template", |b| {
+    let test_cpu_template = build_test_template();
+    println!(
+        "Serialization test - Template size: [{}] bytes.",
+        size_of_val(&test_cpu_template)
+    );
+
+    c.bench_function("deserialize_cpu_template", |b| {
         b.iter(|| bench_deserialize_cpu_template(TEST_TEMPLATE_JSON))
+    });
+
+    c.bench_function("serialize_cpu_template", |b| {
+        b.iter(|| bench_serialize_cpu_template(&test_cpu_template))
     });
 }
 
--- a/src/vmm/src/guest_config/templates/test_utils/mod.rs
+++ b/src/vmm/src/guest_config/templates/test_utils/mod.rs
@@ -1,0 +1,19 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+/// Aarch64 sub-module reference for "use" statements
+#[cfg(target_arch = "aarch64")]
+pub mod aarch64;
+/// x86_64 sub-module reference for "use" statements
+#[cfg(target_arch = "x86_64")]
+pub mod x86_64;
+
+/// The following wildcard imports enables consumers of the test_utils module
+/// to not be concerned with the architecture in question.
+
+/// Test data unique to aarch64 platforms
+#[cfg(target_arch = "aarch64")]
+pub use aarch64::*;
+/// Test data unique to x86_64 platforms
+#[cfg(target_arch = "x86_64")]
+pub use x86_64::*;
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -25,15 +25,15 @@ def is_on_skylake():
 # differences may appear.
 if utils.is_io_uring_supported():
     COVERAGE_DICT = {
-        "Intel": 81.71 if is_on_skylake() else 82.23,
-        "AMD": 80.81,
-        "ARM": 82.69,
+        "Intel": 81.67 if is_on_skylake() else 82.18,
+        "AMD": 80.77,
+        "ARM": 82.68,
     }
 else:
     COVERAGE_DICT = {
-        "Intel": 78.95 if is_on_skylake() else 79.46,
-        "AMD": 78.04,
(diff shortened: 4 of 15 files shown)


src/vmm/src/guest_config/templates/test_utils/mod.rs
```diff
@@ -0,0 +1,227 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kalyazin, line 1: Looks like this content was not removed from `guest_config/templates/*`. I expected this to be a move operation as the commit message suggests, not a copy.

src/vmm/src/guest_config/templates/test_utils/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+use crate::guest_config::templates::x86_64::{
+    CpuidLeafModifier, CpuidRegister, CpuidRegisterModifier, RegisterModifier, RegisterValueFilter,
+};
+use crate::guest_config::templates::CustomCpuTemplate;
+
+/// Create architecture agnostic type aliases
+
+/// Test CPU template in JSON format
+#[cfg(target_arch = "aarch64")]
+pub const TEST_TEMPLATE_JSON: &str = r#"{
```
kalyazin, line 18: would it be too much to separate these into arch-specific files?

src/vmm/benches/cpu_templates.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kalyazin, line 1: nit: s/2020/2023/ ?

tests/integration_tests/performance/test_cpu_template_benchmark.py
```diff
@@ -0,0 +1,90 @@
...
+BENCHMARK_DIRECTORY = "{}/src/vmm".format(FC_WORKSPACE_DIR)
+DEFAULT_BUILD_TARGET = "{}-unknown-linux-musl".format(platform.machine())
+
+PROC_MODEL = proc.proc_type()
+
+NSEC_IN_MSEC = 1000000
+
+BASELINES = {
+    "Intel": {
+        "deserialize": {"target": 0.025, "delta": 0.02}  # milliseconds
```
kalyazin: I presume targets are taken from actually measurements.
How were the deltas chosen?

src/vmm/Cargo.toml
```diff
@@ -40,5 +40,9 @@ criterion = { version = "0.4.0", default-features = false }
 device_tree = "1.1.0"
 
 [[bench]]
-name = "main"
+name = "cpu_templates"
 harness = false
+
+[[bench]]
+name = "snapshots"
+harness = false
```
kalyazin, line 48: nit: new line at the end.

tests/integration_tests/performance/test_cpu_template_benchmark.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def _check_statistics(directory, mean):
+    proc_model = [item for item in BASELINES if item in PROC_MODEL]
+    assert len(proc_model) == 1, "Could not get processor model!"
+
+    measure = BASELINES[proc_model[0]]["deserialize"]
+    target, delta = measure["target"], measure["delta"]
+    assert target == pytest.approx(
+        mean, abs=delta, rel=1e-6
```
kalyazin, line 53: why do we need both `abs` and `rel` here? Is `abs` not going to be always greater than `1e-6`?

src/vmm/benches/cpu_templates.rs
```diff
@@ -0,0 +1,34 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+// Benchmarking cases:
+//   * `CustomCpuTemplate` Deserialization
+
+use std::path::Path;
+
+use criterion::{criterion_group, criterion_main, Criterion};
+use vmm::guest_config::templates::test_utils::TEST_TEMPLATE_JSON;
```
kalyazin: The test json templates are rather short. I guess their size reflects a typical customer template size, but don't we want to test bigger templates to give us some margin of confidence?

src/vmm/benches/cpu_templates.rs
```diff
@@ -2,25 +2,44 @@
...
 
 pub fn cpu_template_benchmark(c: &mut Criterion) {
-    println!("Template size: [{}] bytes.", TEST_TEMPLATE_JSON.len());
+    println!(
+        "Template size for deserialization: [{}] bytes.",
+        TEST_TEMPLATE_JSON.len()
+    );
+    let test_cpu_template = build_test_template();
+    println!(
+        "Template size for serialization: [{}] bytes.",
```
kalyazin, line 32: So this is the size of the template in memory. Would it be a bit misleading to print the two values side by side, although they are different by nature?

tests/integration_tests/performance/test_cpu_template_benchmark.py
```diff
@@ -79,11 +87,12 @@ def test_cpu_template_benchmark(monkeypatch, record_property):
         estimates = json.loads(json_file.read_text())
 
         # Save the Mean measurement(nanoseconds) and transform it(milliseconds)
-        mean = estimates["mean"]["point_estimate"] / NSEC_IN_MSEC
-        logger.info("Mean: %f", mean)
+        mean_ns = estimates["mean"]["point_estimate"]
+        mean_ms = mean_ns / NSEC_IN_MSEC
```
kalyazin, line 91: I like this transparency.

src/vmm/src/guest_config/templates/test_utils/mod.rs
mattschlebusch: You are correct. I think the test data was introduced following a rebase that I didn't check through. I will remove from the non-test modules.

tests/integration_tests/performance/test_cpu_template_benchmark.py
mattschlebusch: Repeated runs to get a band of variance anecdotally. Given the performance margin being sub-millisecond, I think this is adequate for v1.

tests/integration_tests/performance/test_cpu_template_benchmark.py
mattschlebusch: Not necessarily. It depends on the data set size. I would like to see these tests expanded to check small versus large data sets. On very small data sets the relative margin is very small.

src/vmm/benches/cpu_templates.rs
mattschlebusch: I do want to expand these test for sure. Right now, I just want the benchmark to protect against changes that introduce order of magnitude regressions in the future. But we can and will be doing more.

tests/integration_tests/build/test_coverage.py
```diff
@@ -25,15 +25,15 @@ def is_on_skylake():
 # differences may appear.
 if utils.is_io_uring_supported():
     COVERAGE_DICT = {
-        "Intel": 81.71 if is_on_skylake() else 82.23,
-        "AMD": 80.81,
-        "ARM": 82.69,
+        "Intel": 81.49 if is_on_skylake() else 82.00,
```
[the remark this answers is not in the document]
mattschlebusch: I suspect that the benchmark code itself is categorized as prod code, not test code. So as the benchmark code grows, so does the code volume target. I have not confirmed this though, but given where this PR introduces code changes, I think it makes sense. The other possibility as to the amount of coverage dropped is in the duplication of the test data constants/builders. I am hoping this gap will drop a bit when I fix in the next push.

src/vmm/src/guest_config/templates/test_utils/mod.rs
mattschlebusch: Yea this is fair. I think this should be easy enough while maintaining the namespace path.

tests/integration_tests/performance/test_cpu_template_benchmark.py
kalyazin: Hm. I'm not sure I understand the influence of the dataset size. At this point the samples are already reduced to a scalar value (mean). We are just comparing 2 numbers here. How would the dataset size affect the significance of the `rel` value without being "noticed" by the `abs`?

src/vmm/benches/cpu_templates.rs
mattschlebusch: I can clarify the difference in the log statement a bit.

src/vmm/benches/cpu_templates.rs
kalyazin: That's fair enough.

src/vmm/benches/cpu_templates.rs
kalyazin: Yes, that would be good.

tests/integration_tests/build/test_coverage.py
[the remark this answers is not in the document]
kalyazin: Let's see what it looks like after removing the duplicate code. I doesn't look natural to me that benchmark code pulls our coverage down, to be honest. If it's the case, we should do something to fix it.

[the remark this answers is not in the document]
mattschlebusch: I agree it does not make sense. My current theory is that any code not under `#[test] ` counts as code that needs coverage. Configurability to fix this would improve our coverage accuracy/relevancy.

tests/integration_tests/performance/test_cpu_template_benchmark.py
mattschlebusch: Much larger data sets would have larger bands of variance, well beyond the delta defined here. This is a bit of a short-cut that `test_versioned_serialization_benchmark.py` takes as well, so that we don't have to configure deltas for every dataset size. I included the parameter here as it makes it convenient to add changing datasets. I can remove it if you prefer, and the relativity issue can be address when extra data sets are introduced?

tests/integration_tests/build/test_coverage.py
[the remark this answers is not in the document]
mattschlebusch: There is an improvement with the removed duplication [CI build after cleanup](https://buildkite.com/firecracker/firecracker-pr/builds/2355). There is still a slight drop though as I expected.

tests/integration_tests/performance/test_cpu_template_benchmark.py
kalyazin: I think it is fine is as it is, but could we document our reasoning in a code comment to aid decisions in the future?

tests/integration_tests/build/test_coverage.py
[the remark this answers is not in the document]
kalyazin: Yeah, the drop looks negligible now.

kalyazin (approved):

bchalios (commented):
My main comment is do we really need these as part of the PR CI tests?

If I understand correctly, this essentially is benchmarking the `serde` implementation. I remember we had a discussion regarding a similar situation with versionize.

In any case, my main concern is whether we might add another good candidate for intermittent failures in the CI, blocking PRs, for testing something that is not Firecracker specific.

tests/integration_tests/build/test_coverage.py
```diff
@@ -25,15 +25,15 @@ def is_on_skylake():
 # differences may appear.
 if utils.is_io_uring_supported():
     COVERAGE_DICT = {
-        "Intel": 81.71 if is_on_skylake() else 82.23,
-        "AMD": 80.81,
-        "ARM": 82.69,
+        "Intel": 81.67 if is_on_skylake() else 82.18,
```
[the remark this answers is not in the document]
mattschlebusch: My theory here is that there is more benchmark code, and it is not counted as test code but as prod code. I don't think there is too much more code in terms of the test_utils module itself.

mattschlebusch (commented):
> If I understand correctly, this essentially is benchmarking the serde implementation. I remember we had a discussion regarding a similar situation with versionize.

Not quite. There is quite a bit of custom serialization and deserialization code for the fields in the `CustomCpuTemplate` type.

> In any case, my main concern is whether we might add another good candidate for intermittent failures in the CI, blocking PRs, for testing something that is not Firecracker specific.

The format of the template and the custom processing we include for the fields in the template is not using default derivations unfortunately.

bchalios (approved):

PR #3631, opened 2023-04-26
Make dependency tests compatible with dependabot

roypat:
## Changes

Replaces the test_num_dependencies test with a GitHub action that fails whenever dependencies are updated (but doesn't block merging). See for an example ~https://github.com/roypat/firecracker/pull/19~ this very PR because the Cargo.lock files in seccomp_demo and uffd were dirty and needed updating. 

Also passes the `--locked` option to all our `cargo` invocations in the test suite, which makes them fail if `Cargo.lock` and `Cargo.toml` are out-of-sync.

## Reason

Our old test_num_dependencies test required a list of dependencies in a .txt file. This is incompatible with dependabot's update strategy. Since the test was only to prompt us toward dependency updates, this PR replaces it with a GitHub action that fails whenever a Cargo.lock file is changed in a PR. 

The alternative of only running a single command with `--locked` and turning that into a "cargo lock files in sync" test sadly does not work, as some other test might have previously updated the lockfile already via a command that was run without `--locked`. Our testing suite does not reset the repository between runs (and either can it, or we'd loose the ability to run it locally)

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


tests/host_tools/cargo_build.py
```diff
@@ -43,7 +45,7 @@ def cargo_test(path, extra_args=""):
     path = os.path.join(path, CARGO_UNITTEST_REL_PATH)
     cmd = (
         "CARGO_TARGET_DIR={} RUST_TEST_THREADS=1 RUST_BACKTRACE=1 "
-        'RUSTFLAGS="{}" cargo test {} --all --no-fail-fast'.format(
+        'RUSTFLAGS="{}" cargo test --locked {} --all --no-fail-fast'.format(
```
pb8o, line 48: hmm would it be possible to do it from a cargo wrapper?

```python
def cargo(cmd, env=None):
    return run_cmd("cargo --locked", cmd, env)
```

roypat: Mh, just checked that its indeed possible to run cargo subcommands as `cargo --locked check` for instance. I think this would be beneficial because it also reduces the probability of later added tests running without the `--locked`. I'll add it when I find some time to cycle back to this, sadly some higher priority things have come up right now :(

roypat: done, please have another look :)

tests/host_tools/cargo_build.py
```diff
@@ -33,9 +33,9 @@ def cargo_build(path, extra_args="", src_dir="", extra_env=""):
     """Trigger build depending on flags provided."""
     cmd = "CARGO_TARGET_DIR={} {} cargo build {}".format(path, extra_env, extra_args)
     if src_dir:
-        cmd = "cd {} && {}".format(src_dir, cmd)
-
-    utils.run_cmd(cmd)
+        utils.run_cmd(cmd, cwd=src_dir)
+    else:
+        utils.run_cmd(cmd)
```
pb8o: I think if we defaulted `src_dir =None` in the function signature, this else would not be needed

tests/host_tools/cargo_build.py
```diff
@@ -28,26 +28,41 @@
...
+    cargo_args: str = "",
+    subcommand_args: str = "",
+    *,
+    env: dict = None,
+    cwd: str = None,
+):
+    """Executes the specified cargo subcommand"""
+    env = env or {}
+
+    env_string = " ".join(f'{key}="{str(value)}"' for key, value in env.items())
```
pb8o, line 43: pity that run_cmd doesn't accept a `dict` for end. It could be passed straight to `subprocess.run`

tests/host_tools/cargo_build.py
```diff
@@ -28,26 +28,41 @@
...
+
+    cmd = f"{env_string} cargo {subcommand} {cargo_args} -- {subcommand_args}"
+
+    if cwd:
+        return utils.run_cmd(cmd, cwd=cwd)
+    return utils.run_cmd(cmd)
```
pb8o, line 52: I think if it's None it should work fine unconditionally

.github/workflows/dependency_modification_check.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: "Check Cargo.lock not in changeset" 
+        run: |
+          git fetch origin
+          git diff origin/$GITHUB_BASE_REF.. --name-only| ( ! grep "Cargo.lock")
```
pb8o, line 18: nit: may want to add the `-q` so grep doesn't print, but irrelevant if we never see the output

pb8o (approved):

tests/host_tools/cargo_build.py
roypat: oh, yeah, you're right! Not sure what I was thinking here, I even remember looking at `run_cmd_sync` thinking "wouldnt this be neat", lol

.github/workflows/dependency_modification_check.yml
roypat: Mh, you can see the output by clicking on "Details" below. It tells you what Cargo.lock file caused the failure, although I guess that's also clearly visible from the "Files Changed" view, lol.

tests/host_tools/cargo_build.py
roypat: haha, one step at a time, lets leave that for a follow up xP

tests/host_tools/cargo_build.py
roypat: Yup, but this code is yeeted in a follow up commit anyway :)

.github/workflows/dependency_modification_check.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  dependency_changed_check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: "Check Cargo.lock not in changeset"
```
JonathanWoollett-Light, line 15: It can occur where the `Cargo.lock` is included in the PR but `Cargo.toml` files are updated such that when built it updates the `Cargo.lock`. This test misses this case. This case is caught with https://github.com/firecracker-microvm/firecracker/pull/3631/commits/28129693a269c1f325665377ca396ff14afddb61 but this would mean that when we run CI if the lockfile is out-of-date it causes fails in many tests which don't need to care, it would be preferable to have only 1 failure in 1 test for this 1 error case (such that if other error cases occurred we could see them and fix them before needing another CI re-run).

This test can instead use `run: cargo generate-lockfile --locked`

roypat: Ahh, that works around the issue of not having control over test execution order, nice! Is there a way to mark a GitHub Action as required? I think we'll still want to block CI on out-of-date lockfiles, and the dependency_modification_check one is an optional action.

roypat: Ah, amazing, thanks! I'll drop 28129693a269c1f325665377ca396ff14afddb61 then and add the github action you propose then :) 

putting this in draft until that's done

.github/workflows/dependency_modification_check.yml
```diff
@@ -0,0 +1,18 @@
+name: Check no dependencies were modified
+
+on:
+  pull_request
```
JonathanWoollett-Light, line 4: Nit: The default trigger is `push` and I think this produces near identical behavior.
```suggestion
on:
  push
```
this will also run the action in the case we need to force-push to any branch like when rebasing a feature branch.

roypat: Seems like `push` doesnt trigger on pull_request. I've set it trigger on both now

.github/workflows/dependency_modification_check.yml
roypat: done!

pb8o (approved):

.github/workflows/deny_dirty_cargo_locks.yml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  no_dirty_cargo_locks_check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: "Check no Cargo.lock files are dirty"
```
JonathanWoollett-Light, line 15: This seems more complex than needed, 1 lock file should be generated for the whole workspace, so just doing `cargo generate-lockfile --locked` should be sufficient, am I missing something?

.github/workflows/deny_dirty_cargo_locks.yml
```diff
@@ -0,0 +1,32 @@
...
+  [push, pull_request]
+
+jobs:
+  no_dirty_cargo_locks_check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout repository"
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
JonathanWoollett-Light, line 13: By default it checks out the head of the PR I don't think this needs to be set manually.

.github/workflows/deny_dirty_cargo_locks.yml
roypat: There are multiple lockfiles (for example the ones updated in this PR for the uffd example and the seccomp demo). Also, `generate-lockfile` implicitly performs a `cargo update` since it generates a lockfile from scratch, which is why this Action is using `cargo metadata`.

.github/workflows/deny_dirty_cargo_locks.yml
roypat: It seems to checkout the branch on the fork by default, but to compare against main below we need it to checkout the head in _this_ repository.

JonathanWoollett-Light: I think this behavior is okay, especially since we don't actually allow merge commits.

.github/workflows/deny_dirty_cargo_locks.yml
JonathanWoollett-Light: Under https://doc.rust-lang.org/cargo/commands/cargo-generate-lockfile.html#manifest-options it notes:

> --frozen
> --locked
>
>  Either of these flags requires that the Cargo.lock file is up-to-date. If the lock file is missing, or it needs to be updated, Cargo will exit with an error. The --frozen flag also prevents Cargo from attempting to access the network to determine if it is out-of-date.
> 
>  These may be used in environments where you want to assert that the Cargo.lock file is up-to-date (such as a CI build) or want to avoid network access.

Specifically 

> If the lock file is missing, or it needs to be updated, Cargo will exit with an error.

Do we know what constitutes `needs to be updated`?

Is this when:

1. `Cargo.lock` is incompatible with the `Cargo.toml`s such that it must be updated to be made valid.
2. Dependencies in `Cargo.lock` can be updated (e.g. `cargo update` would produce a change).

My understand was that this related to the 1st point.

roypat: In this case, its the latter (generate-lockfile --locked fails against our main Cargo.lock for example, even though it is in-sync with Cargo.toml)

.github/workflows/deny_dirty_cargo_locks.yml
roypat: We have temporary merge commits due to people clicking "update branch"

PR #3689, opened 2023-05-15
fix: fixed bug in deserialization method

ShadowCurse:
## Changes
Fixed byte access with `strip_prefix` checks + added unit tests with unicode.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/cpu_config/templates_serde.rs
+++ b/src/vmm/src/cpu_config/templates_serde.rs
@@ -24,26 +24,19 @@ macro_rules! deserialize_from_str {
             D: Deserializer<'de>,
         {
             let number_str = String::deserialize(deserializer)?;
-            let deserialized_number = if number_str.len() > 2 {
-                match &number_str[0..2] {
-                    "0b" => $type::from_str_radix(&number_str[2..], 2),
-                    "0x" => $type::from_str_radix(&number_str[2..], 16),
-                    _ => $type::from_str(&number_str),
-                }
-                .map_err(|err| {
-                    D::Error::custom(format!(
-                        "Failed to parse string [{}] as a number for CPU template - {:?}",
-                        number_str, err
-                    ))
-                })?
+            let deserialized_number = if let Some(s) = number_str.strip_prefix("0b") {
+                $type::from_str_radix(s, 2)
+            } else if let Some(s) = number_str.strip_prefix("0x") {
+                $type::from_str_radix(s, 16)
             } else {
-                $type::from_str(&number_str).map_err(|err| {
-                    D::Error::custom(format!(
-                        "Failed to parse string [{}] as a decimal number for CPU template - {:?}",
-                        number_str, err
-                    ))
-                })?
-            };
+                $type::from_str(&number_str)
+            }
+            .map_err(|err| {
+                D::Error::custom(format!(
+                    "Failed to parse string [{}] as a number for CPU template - {:?}",
+                    number_str, err
+                ))
+            })?;
             Ok(deserialized_number)
         }
     };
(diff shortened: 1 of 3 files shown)


src/vmm/src/cpu_config/templates_serde.rs
```diff
@@ -24,26 +24,19 @@ macro_rules! deserialize_from_str {
...
+                $type::from_str_radix(s, 16)
             } else {
-                $type::from_str(&number_str).map_err(|err| {
-                    D::Error::custom(format!(
-                        "Failed to parse string [{}] as a decimal number for CPU template - {:?}",
-                        number_str, err
-                    ))
-                })?
-            };
+                $type::from_str(&number_str)
```
mattschlebusch, line 32: I think we can remove support for decimal numbers here.

mattschlebusch (commented):
Adding a unit test to check that strange characters are handleable is probably worthwhile, to duplicate any deserialization bug that was observed previously, that the new code fixes etc. Something similar to the following should be ok:
```
        // Malformed 64-bit address with invalid string encoding
        let cpu_template_result = serde_json::from_str::<CustomCpuTemplate>(
            r#"{
                    "msr_modifiers":  [
                        {
                            "addr": "xϽ0b0000000000000000000000000000000000000000000",
                            "bitmap": "0b0000000xxxxx11110000"
                        },
                    ]
                }"#,
        );
        assert!(cpu_template_result.is_err());
        assert!(cpu_template_result.unwrap_err().to_string().contains(
            "Failed to parse string [xϽ0b0000000000000000000000000000000000000000000]"
        ));
```

src/vmm/src/cpu_config/templates_serde.rs
ShadowCurse: Ok guys, you really don't like this 1 line of code, do you?

kalyazin: The reasons I don't like it are:
- I can't see a real use case for it to justify additional code
- It may catalyse a human error due to a high character overlap between dec and hex (eg `4000` instead of intended `0x4000`).

ShadowCurse: 1. The code size is same (or even smaller) with dec decode, rather then without it, because in the last `else` we need to create a `Err("some error string")`.
2. the more probable error is between "0b..." and "0x...". It is much more obvious when value is dec or not.

zulinx86: What actual use case are you imagining for using decimal values for CPUID leaf / MSR address / Aarch64 register IDs?
Minimizing attach surface gives us safety.

ShadowCurse: Well, initially I was defining aarch64 regs ids as decimals. But overall don't have strong use cases for or against decimal format. It is just feel correct to support all of them. The only reason to drop decimals will be if we would force a specific string length i.e for u32 it must be 32 + 2 chars long in binary and 8 + 2 in hex.

> Minimizing attack surface gives us safety.

If only we would have thought about this every time we add new features...

ShadowCurse: Just tried to remove decimal format format and it seems much more work then it should be. Decimal format is used for flags in all x86 templates. Is it really worth it to replace all 0 with 0x0 just to remove decimal format that does no hurt anyone and is even already used by us?

mattschlebusch: I am ok deferring the decimal support change personally. Not important for the case the PR is addressing.

mattschlebusch (approved):

kalyazin (approved):

PR #3693, opened 2023-05-15
Use `aws-lc-rs` for getting random bytes for `virtio-rng`

bchalios:
## Changes

Switch to using `aws-lc-rs` for getting random bytes for `virtio-rng`.

## Reason

Currently, we are using the `OsRng` type from the `rand` crate. `aws-lc-rs` is in the process of getting FIPS certification, so switching to this implementation might benefit us in the future for use-cases that require FIPS compliance.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


docs/entropy.md
```diff
@@ -53,10 +53,8 @@ be achieved by adding a section like this:
 }
 ```
 
-On the host side, firecracker uses [`OsRng`][2] to get random bytes from the
-host kernel. The [implementation][3] of `OsRng` on Linux uses the
-`getrandom(2)` system call when available, otherwise it falls back to
-`/dev/urandom` after successfully polling `/dev/random`.
+On the host side, Firecracker relies on [`aws-lc-rs`][2] to retrieve the random.
```
roypat, line 56: ...bytes?

resources/seccomp/aarch64-unknown-linux-musl.json
```diff
@@ -206,16 +206,7 @@
...
-                "args": [
-                    {
-                        "index": 2,
-                        "type": "dword",
-                        "op": "eq",
-                        "val": 4,
-                        "comment": "libc::MADV_DONTNEED"
-                    }
-                ]
+                "comment": "Used by the VirtIO balloon device and by musl for some customer workloads. It is also used by aws-lc during random number generation. They setup a memory page that mark with MADV_WIPEONFORK to be able to detect forks. They also call it with -1 to see if madvise is supported in certain platforms."
```
roypat, line 209: I'm not 100% sure on how the seccompiler/bpf thingy works, but I'm assuming it would just be tedious to specify the exact parameters with which madvise would be called, and the syscall isn't spicy enough for us to care about that? :o

bchalios: Problem is they are trying madvise with `-1` which as when unsigned is `0xFFFFFFFF`, so unfortunately we can't distinguish.

roypat (dismissed):

resources/seccomp/aarch64-unknown-linux-musl.json
mattschlebusch: Would it be possible to codify this use-case in a test of some sort, just so that the restriction isn't re-added to the seccomp rules accidentally?

mattschlebusch (commented):
Overall looks ok to me, just an integ test to make sure the change to the seccomp rule is vetted in the future (would be ideal). Is this easy enough to add?

tests/integration_tests/build/test_binary_size.py
```diff
@@ -14,7 +14,7 @@
 
 SIZES_DICT = {
     "x86_64": {
-        "FC_BINARY_SIZE_TARGET": 2649920,
+        "FC_BINARY_SIZE_TARGET": 2808184,
```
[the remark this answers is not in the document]
bchalios: Yup, we did on x86. I think it's to be expected. We are bringing in `aws-lc` which seems to be quite big a library.

resources/seccomp/aarch64-unknown-linux-musl.json
bchalios: I think that if the seccomp filter is changed to be stricter in the future the tests for `virtio-rng` will just break, because they will eventually exercise this particular `madvise` path. Do you have any suggestions for a more specific test?

bchalios:
Converting this temporarily to draft. Trying to find out if `aws-lc` is using `getrandom` or not. If not, we need to remove it from the allow list.

resources/seccomp/aarch64-unknown-linux-musl.json
mattschlebusch: Would the aws-lc related tests fail now if the `libc::MADV_DONTNEED` rule was present? If not, is there anything we can do in with aws-lc to force this madvise path?

bchalios:
I launched a pipeline with `getrandom` removed from the allow-list and we get failures for all architectures: https://buildkite.com/firecracker/firecracker-pr/builds/2965#_

So, I will keep the system call in the list.

roypat (approved):

andreitraistaru (approved):

resources/seccomp/aarch64-unknown-linux-musl.json
bchalios: they would! They always call `madvise` with -1: https://github.com/aws/aws-lc/blob/987bd7c672e2c3dd1cc76be3ce9eece93ba9bd88/crypto/fipsmodule/rand/fork_detect.c#L53, so the test would fail.

mattschlebusch (approved):

PR #3722, opened 2023-05-23
feat: upgrade of deserialization of RVF.

ShadowCurse:
- Rewrote deserialization method for `RegisterValueFilter`.
- Improved performance by ~75%.
- Added ability to use `_` in string representation.

running: `cargo bench --bench cpu_templates`
```bash
deserialize_cpu_template
                        time:   [5.5389 µs 5.5402 µs 5.5419 µs]
                        change: [-77.543% -77.536% -77.530%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 200 measurements (5.50%)
  2 (1.00%) low mild
  4 (2.00%) high mild
  5 (2.50%) high severe
```

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/cpu_config/templates.rs
+++ b/src/vmm/src/cpu_config/templates.rs
@@ -27,7 +27,6 @@ mod common_types {
 }
 
 use std::borrow::Cow;
-use std::num::ParseIntError;
 use std::result::Result;
 
 pub use common_types::*;
@@ -112,24 +111,39 @@ where
 }
 
 /// Trait for numeric types
-pub trait Numeric: Sized {
+pub trait Numeric:
+    Sized
+    + Copy
+    + PartialEq<Self>
+    + std::ops::Not<Output = Self>
+    + std::ops::BitAnd<Output = Self>
+    + std::ops::BitOr<Output = Self>
+    + std::ops::BitOrAssign<Self>
+    + std::ops::Shl<Output = Self>
+    + std::ops::AddAssign<Self>
+{
     /// Number of bits for type
     const BITS: u32;
-    /// Parse numeric type from string with given radix
-    fn from_str_radix(s: &str, radix: u32) -> Result<Self, ParseIntError>;
     /// Value of bit at pos
     fn bit(&self, pos: u32) -> bool;
+    /// Returns 0 of the type
+    fn zero() -> Self;
+    /// Returns 1 of the type
+    fn one() -> Self;
 }
 
 macro_rules! impl_numeric {
     ($type:tt) => {
         impl Numeric for $type {
             const BITS: u32 = $type::BITS;
-            fn from_str_radix(s: &str, radix: u32) -> Result<Self, ParseIntError> {
-                $type::from_str_radix(s, radix)
-            }
             fn bit(&self, pos: u32) -> bool {
                 (self & (1 << pos)) != 0
+            }
+            fn zero() -> Self {
+                0
+            }
+            fn one() -> Self {
+                1
             }
         }
     };
@@ -143,12 +157,7 @@ impl_numeric!(u128);
 
 impl<V> Serialize for RegisterValueFilter<V>
 where
-    V: Copy
-        + std::ops::Not<Output = V>
-        + std::ops::BitAnd<Output = V>
-        + std::ops::BitOr<Output = V>
-        + PartialEq<V>
-        + Numeric,
+    V: Numeric,
 {
     /// Serialize combination of value and filter into a single tri state string
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
@@ -179,11 +188,7 @@ where
 
 impl<'de, V> Deserialize<'de> for RegisterValueFilter<V>
 where
-    V: Copy
-        + std::ops::Not<Output = V>
-        + std::ops::BitAnd<Output = V>
-        + std::ops::BitOr<Output = V>
-        + Numeric,
+    V: Numeric,
 {
     /// Deserialize a composite bitmap string into a value pair
     /// input string: "010x"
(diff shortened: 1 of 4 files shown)


zulinx86:
> @zulinx86 Because perf tests pass I assume the delta is big enough for this improvement. I don't really think changing baselines here is necessary.

Hmm, 65% looks huge enough to me.
But this issue seems to be caused by the fact that the benchmark measure it in milliseconds granularity.
https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/performance/test_cpu_template_benchmark.py#L25-L38

For future reference, I'd appreciate if you could post the benchmark result before and after the change here.

src/vmm/src/cpu_config/templates.rs
```diff
@@ -195,26 +195,65 @@ where
...
+            if c == &b'0' {
+                *c = b'1';
+            }
+        });
+        // replace 'x' => '0'
+        filter.iter_mut().for_each(|c| {
+            if c == &b'x' {
+                *c = b'0';
+            }
+        });
```
roypat, line 232: if this is perf critical, maybe
```suggestion
        // replace '0' => '1'
        filter.iter_mut().for_each(|c| {
            *c = match *c {
                b'0' => b'1',
                b'x' => b'0',
                y => y
            };
        });
```
to avoid a loop? although rustc might be clever enough to figure this out itself

src/vmm/src/cpu_config/templates.rs
```diff
@@ -195,26 +195,65 @@ where
...
+
+        // # Safety:
+        // We know that vectors contain only valid ASCII characters
+        let value = unsafe { std::str::from_utf8_unchecked(&value) };
+        // # Safety:
+        // We know that vectors contain only valid ASCII characters
+        let filter = unsafe { std::str::from_utf8_unchecked(&filter) };
+
+        let filter = V::from_str_radix(filter, 2).map_err(|err| {
```
roypat, line 247: What is the exact performance improvement from foregoing the unicode checks here? Is it quantifiable? If not, I don't see why we shouldn't use `std::str::from_utf8(&...).unwrap()` here. 

Or maybe we can operate on a char array as opposed to a byte array? The check of "is this a unicode boundary" only checks a single bit afaik, so not sure how much impact skipping that has.

roypat (commented):
Mh, is this on the hot path? I'm not very familiar with y'all's work the past few months, but I thought that the json deserialization wasn't supposed to be on the hot path and instead some optimized binary forward was used? :o

src/vmm/src/cpu_config/templates.rs
ShadowCurse: I don't think rustc will optimize it but yeah, minimizing loc seems good.

zulinx86:
> Mh, is this on the hot path? I'm not very familiar with y'all's work the past few months, but I thought that the json deserialization wasn't supposed to be on the hot path and instead some optimized binary forward was used? :o

I don't have full context behind that, but the baseline values are ~45 us on x86 and ~7.5us on aarch64.
https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/performance/test_cpu_template_benchmark.py#L25-L38

Test templates are here:
- https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/cpu_config/x86_64/test_utils.rs
- https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/cpu_config/aarch64/test_utils.rs

IIUC, we don't support binary compiled CPU template in firecracker v1.4. If users need more performance, we can support binary ones in successive versions.

ShadowCurse:
> Mh, is this on the hot path? I'm not very familiar with y'all's work the past few months, but I thought that the json deserialization wasn't supposed to be on the hot path and instead some optimized binary forward was used? :o

So this whole PR was created, because I thought adding `_` as a separator would be nice in our trinary format.
The refactoring just came naturally. Why wouldn't I optimize code that I touch?

src/vmm/src/cpu_config/templates.rs
```diff
@@ -195,26 +195,62 @@ where
     where
         D: Deserializer<'de>,
     {
-        let mut bitmap_str = String::deserialize(deserializer)?;
+        let original_str = String::deserialize(deserializer)?;
```
JonathanWoollett-Light, line 198: This appears to be simply de-serializing 1 string into 2 integers, in this case I think this can be greatly simplified into something like: [playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=074c919a92fe2ed15682df87832baf68)
```rust
#[derive(Debug)]
struct UnknownCharacterInBitField;
impl std::fmt::Display for UnknownCharacterInBitField {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        write!(f, "Unknown character found in bitfield.")
    }
}
impl std::error::Error for UnknownCharacterInBitField {}

fn main() -> Result<(),Box<dyn std::error::Error>> {
    let original_str = String::from("0bx00100x0x1xxxx01xxx1xxxxxxxxxxx1");

    let stripped_str = original_str.strip_prefix("0b").unwrap_or(&original_str);

    let (mut filter,mut value) = (0u64,0u64);
    let mut i = 0;
    for s in stripped_str.as_bytes().iter().rev() {
        match s {
            b'_' => continue,
            b'x' => (),
            b'0' => {
                filter |= 1 << i;
            },
            b'1' => {
                value |= 1 << i;
            },
            _ => return Err(Box::new(UnknownCharacterInBitField)),
        }
        i += 1;
    }
    println!("filter: {filter:x}");
    println!("value:  {value:x}");
    Ok(())
}
```

src/vmm/src/cpu_config/templates.rs
```diff
@@ -195,29 +200,39 @@ where
...
 
-        if bitmap_str.starts_with("0b") {
-            bitmap_str = bitmap_str[2..].to_string();
+        if !original_str.is_ascii() {
+            return Err(D::Error::custom(format!(
+                "Failed to parse string [{}] as a bitmap - non ascii characters detected",
+                original_str,
+            )));
```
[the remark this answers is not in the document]
ShadowCurse: removed

src/vmm/src/cpu_config/templates.rs
```diff
@@ -195,29 +200,39 @@ where
...
-        let value = V::from_str_radix(value_str.as_str(), 2).map_err(|err| {
-            D::Error::custom(format!(
-                "Failed to parse string [{}] as a bitmap - {:?}",
-                bitmap_str, err
-            ))
-        })?;
+        let stripped_str = original_str.strip_prefix("0b").unwrap_or(&original_str);
 
+        let (mut filter, mut value) = (V::zero(), V::zero());
+        let mut i = V::zero();
```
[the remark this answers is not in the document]
ShadowCurse: Does this really matter if `i` is u32 or u64 or u128?

src/vmm/src/cpu_config/templates.rs
JonathanWoollett-Light: Version with improved generics https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c0a7e03edb76d40e62632678785ba1e6

JonathanWoollett-Light (dismissed):

src/vmm/src/cpu_config/templates.rs
```diff
@@ -112,25 +105,40 @@ where
...
+    + std::ops::Shl<Output = Self>
+    + std::ops::AddAssign<Self>
+{
     /// Number of bits for type
     const BITS: u32;
-    /// Parse numeric type from string with given radix
-    fn from_str_radix(s: &str, radix: u32) -> Result<Self, ParseIntError>;
     /// Value of bit at pos
     fn bit(&self, pos: u32) -> bool;
+    /// Returns 1 of the type
```
zulinx86, line 123: nit
```suggestion
    /// Returns 0 of the type
```

zulinx86 (approved):

JonathanWoollett-Light (approved):

roypat (approved):

PR #3777, opened 2023-06-02
Benchmark overhaul

roypat:
## Changes

Changes the way criterion benchmarks are executed by our CI. It removes all existing, baseline-based tests, and instead adds an new test that depends on criterions statistical analysis for determining regressions. The approach here is based on what rust-vmm is doing [[1]]: We first check out the target branch against which a pull request is made, run criterion, go back to the pull request branch, and run criterion again. This will cause criterion to compare the benchmark numbers from the pull request branch against those of the base branch. 

We set a noise threshold of 5%, as running the benchmarks in CI is still very unreliable, probably due to benchmark design. However, 5% is a significantly lower threshold than what our current test uses. 

Additionally, this PR introduces the concept of "optional" CI tests. These tests will be a run in a different buildkite pipeline- (which will not block merges on failure) that will be enabled for this repository after the PR is merged. For a example of how that pipeline works, see [[3]]. For this specific test, it will print the benchmark numbers even if the test passes, to allow us to keep track of them.

Note that the new pipeline can only be enabled after this PR is merged - for now it is running from my fork, for reviewers to verify its correctness.

[1]: https://github.com/rust-vmm/rust-vmm-ci/blob/main/integration_tests/test_benchmark.py
[3]: https://buildkite.com/firecracker/firecracker-pr-optional

## Reason

Baseline-based benchmarks are a frequent cause of intermittent failures, and updating the baseline is a churn that is causing to not add more benchmarks, even if they might be useful. 

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


.buildkite/pipeline_pr.py
```diff
@@ -25,7 +25,7 @@ def get_changed_files(branch):
 args = COMMON_PARSER.parse_args()
 
 step_style = {
-    "command": "./tools/devtool -y test -- ../tests/integration_tests/style/",
```
pb8o, line 28: Can we make it so that this is the default? i.e. If we want to run this tests, we would have to explicitly say `-m optional`. To see how to do that, look at `conftest.py::pytest_collection_modify_items`.

Maybe `benchmarks` is a more descriptive marker name, or, why not just re-using the `nonci` marker?

tests/conftest.py
```diff
@@ -126,32 +126,6 @@ def test_with_any_microvm(test_microvm_any):
...
-    for skip_marker_name in ["nonci"]:
-        if not config.getoption(f"--{skip_marker_name}"):
-            skip_markers[skip_marker_name] = pytest.mark.skip(
-                reason=f"Skipping {skip_marker_name} test"
-            )
-
-    for item in items:
-        for skip_marker_name, skip_marker in skip_markers.items():
-            if skip_marker_name in item.keywords:
-                item.add_marker(skip_marker)
```
pb8o, line 152: Nice simplification

roypat: sadly it doesn't work :sob: there's some other fixtures that explicitly check for the --nonci marker where I couldn't find a solution - so its back to this for now

.buildkite/pipeline_perf.py
```diff
@@ -46,7 +46,7 @@ def build_group(test):
     test_path = test.pop("test_path")
     return group(
         label=test.pop("label"),
-        command=f"./tools/devtool -y test {devtool_opts} -- --nonci {test_path}",
+        command=f"./tools/devtool -y test {devtool_opts} -- -m nonci {test_path}",
```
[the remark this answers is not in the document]
roypat: Yep, those make it quite a bit harder to change this - so for now I just switched back to the old solution, and combine "--optional" with "-m optional" in the new pipeline generator script

.buildkite/pipeline_pr.py
roypat: I was thinking that the marker should be general enough to allow for any kind of "optional" test we might want to add in the future (maybe some short running perf tests that we just use to collect numbers, or some of the other tests that currently rely on us adjusting numbers). That's why I didnt go with "benchmarks". As for nonci, since the optional pipeline runs all tests that have the marker, reusing it would cause it to also run all nightly perf tests :(

tests/conftest.py
```diff
@@ -126,21 +126,19 @@ def test_with_any_microvm(test_microvm_any):
...
-def pytest_configure(config):
-    """Pytest hook - initialization"""
-    config.addinivalue_line("markers", "nonci: mark test as nonci.")
-
-
 def pytest_addoption(parser):
     """Pytest hook. Add command line options."""
     parser.addoption("--nonci", action="store_true", help="run tests marked with nonci")
+    parser.addoption(
+        "--optional", action="store_true", help="run tests marked as optional"
```
bchalios, line 133: super nit: Shall we rename this to something like "--no-block-pr", so that we do not get confused with the typical connotations of an `--optional` cli flag?

.buildkite/pipeline_pr_optional.py
```diff
@@ -0,0 +1,36 @@
...
+    "priority": DEFAULT_PRIORITY + 1,
+    "agent_tags": ["ag=1"],
+}
+defaults.update(args.step_param)
+
+optional_grp = group(
+    "❓ Optional",
+    # We need to pass both --optional and -m optional here: The first causes our custom collection hook to
+    # not skip the optional tests, while the second causes pytest to select all non-optional tests.
+    "./tools/devtool -y test -- ../tests/integration_tests/ --optional -m optional",
```
bchalios: I would add here flags for cpusets and numa nodes, similar to `pipeline_perf.py`: `-c 1-10 -m 0` to minimize variability in the tests.

src/snapshot/benches/version_map.rs
```diff
@@ -126,7 +126,7 @@ criterion_group! {
     name = benches;
     // TODO: Fix the duration below to match the required time to collect enough samples.
     // A warning message is generated by criterion while the accuracy of the test will be lower.
```
bchalios, line 128: Not sure if this TODO is relevant any more (?)

bchalios (changes requested):

.buildkite/pipeline_pr_optional.py
roypat: done! I tried limiting the cpu range more, too, but it caused hangs. I suppose it makes sense because the benchmarking process isnt the only affected by this, but rather the entire docker container

pb8o (approved):

bchalios (approved):

bchalios (approved):

pb8o (approved):

PR #3787, opened 2023-06-05
feat(aarch64): ability to work with wide registers

ShadowCurse:
## Changes

- Updated `kvm-bindings` to 1.4 which changes `get/set_one_reg` methods.
- Updated internal representation of aarch64 registers:

Now we are able to work with register wider then 128 bits. All registers are stored in the `Aarch64RegisterVec` in one byte array. This way registers of different sizes can be efficiently stored and iterated on.
Supported sizes are: 8, 16, 32, 64, 128, 256, 512, 1024, 2048 bits (all sizes supported by KVM).
Register size is determined by the ID of the register.
Interactions with registers are happening using `Aarch64RegisterRef` or `Aarch64RegisterRefMut`.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/cpu-template-helper/src/template/dump/aarch64.rs
+++ b/src/cpu-template-helper/src/template/dump/aarch64.rs
@@ -1,6 +1,7 @@
 // Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use vmm::arch::aarch64::regs::RegSize;
 use vmm::cpu_config::aarch64::custom_cpu_template::RegisterModifier;
 use vmm::cpu_config::templates::{CpuConfiguration, CustomCpuTemplate, RegisterValueFilter};
 
@@ -10,7 +11,18 @@ pub fn config_to_template(cpu_config: &CpuConfiguration) -> CustomCpuTemplate {
     let mut reg_modifiers: Vec<RegisterModifier> = cpu_config
         .regs
         .iter()
-        .map(|reg| reg_modifier!(reg.id, reg.value))
+        .map(|reg| match reg.size() {
+            RegSize::U32 => {
+                reg_modifier!(reg.id, u128::from(reg.value::<u32, 4>()))
+            }
+            RegSize::U64 => {
+                reg_modifier!(reg.id, u128::from(reg.value::<u64, 8>()))
+            }
+            RegSize::U128 => {
+                reg_modifier!(reg.id, reg.value::<u128, 16>())
+            }
+            _ => unreachable!("Only 32, 64 and 128 bit wide registers are supported"),
+        })
         .collect();
     reg_modifiers.sort_by_key(|modifier| modifier.addr);
 
@@ -19,41 +31,38 @@ pub fn config_to_template(cpu_config: &CpuConfiguration) -> CustomCpuTemplate {
 
 #[cfg(test)]
 mod tests {
-    use vmm::arch::aarch64::regs::Aarch64Register;
+    use vmm::arch::aarch64::regs::{Aarch64RegisterRef, Aarch64RegisterVec};
 
     use super::*;
 
-    fn build_sample_regs() -> Vec<Aarch64Register> {
-        vec![
-            Aarch64Register {
-                id: 0x0,
-                value: 0xffff_ffff_ffff_ffff_ffff_ffff_ffff_ffff,
-            },
-            Aarch64Register {
-                id: 0xffff_ffff_ffff_ffff,
-                value: 0x0000_ffff_0000_ffff_0000_ffff_0000_ffff,
-            },
-            Aarch64Register {
-                id: 0x1,
-                value: 0x0000_ffff_0000_ffff_0000_ffff_0000_ffff,
-            },
-        ]
+    // These are used as IDs to satisfy requirenments
+    // of `Aarch64RegisterRef::new`
+    const KVM_REG_SIZE_U32: u64 = 0x0020000000000000;
+    const KVM_REG_SIZE_U64: u64 = 0x0030000000000000;
+    const KVM_REG_SIZE_U128: u64 = 0x0040000000000000;
+
+    fn build_sample_regs() -> Aarch64RegisterVec {
+        let mut v = Aarch64RegisterVec::default();
+        v.push(Aarch64RegisterRef::new(
+            KVM_REG_SIZE_U128,
+            &0xffff_ffff_ffff_ffff_ffff_ffff_ffff_ffff_u128.to_le_bytes(),
+        ));
+        v.push(Aarch64RegisterRef::new(
+            KVM_REG_SIZE_U32,
+            &0x0000_ffff_u32.to_le_bytes(),
+        ));
+        v.push(Aarch64RegisterRef::new(
+            KVM_REG_SIZE_U64,
+            &0x0000_ffff_0000_ffff_u64.to_le_bytes(),
+        ));
+        v
     }
 
     fn build_expected_reg_modifiers() -> Vec<RegisterModifier> {
         vec![
-            reg_modifier!(
-                0x0000_0000_0000_0000,
-                0xffff_ffff_ffff_ffff_ffff_ffff_ffff_ffff
-            ),
-            reg_modifier!(
-                0x0000_0000_0000_0001,
-                0x0000_ffff_0000_ffff_0000_ffff_0000_ffff
-            ),
-            reg_modifier!(
-                0xffff_ffff_ffff_ffff,
-                0x0000_ffff_0000_ffff_0000_ffff_0000_ffff
-            ),
+            reg_modifier!(KVM_REG_SIZE_U32, 0x0000_ffff),
+            reg_modifier!(KVM_REG_SIZE_U64, 0x0000_ffff_0000_ffff),
+            reg_modifier!(KVM_REG_SIZE_U128, 0xffff_ffff_ffff_ffff_ffff_ffff_ffff_ffff),
         ]
     }
 
--- a/src/vmm/src/arch/aarch64/regs.rs
+++ b/src/vmm/src/arch/aarch64/regs.rs
@@ -5,76 +5,27 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 
-use std::path::PathBuf;
-
 use kvm_bindings::*;
-use kvm_ioctls::VcpuFd;
-use utils::vm_memory::GuestMemoryMmap;
 use versionize::*;
 use versionize_derive::Versionize;
 
-use super::get_fdt_addr;
-
-/// Struct describing a saved aarch64 register.
-///
-/// Used for interacting with `KVM_GET/SET_ONE_REG`.
-#[derive(Debug, Clone, PartialEq, Eq, Versionize)]
-pub struct Aarch64Register {
-    /// The KVM register ID.
-    ///
-    /// See https://docs.kernel.org/virt/kvm/api.html?highlight=kvm_set_one_reg#kvm-set-one-reg
-    pub id: u64,
-
-    /// The value of the register.
-    ///
-    /// 128 bit wide, as we want to restore the V0-V31 FP SIMD registers,
-    /// which are this wide.
-    pub value: u128,
-}
-
-/// Errors thrown while setting aarch64 registers.
-#[derive(Debug, PartialEq, Eq, thiserror::Error)]
-pub enum Error {
-    /// Failed to get a register value.
-    #[error("Failed to get register {0}: {1}")]
-    GetOneReg(u64, kvm_ioctls::Error),
-    /// Failed to set a register value.
-    #[error("Failed to set register {0}: {1}")]
-    SetOneReg(u64, kvm_ioctls::Error),
-    /// Failed to get the register list.
-    #[error("Failed to retrieve list of registers: {0}")]
-    GetRegList(kvm_ioctls::Error),
-    /// Failed to get multiprocessor state.
-    #[error("Failed to get multiprocessor state: {0}")]
-    GetMp(kvm_ioctls::Error),
-    /// Failed to Set multiprocessor state.
-    #[error("Failed to set multiprocessor state: {0}")]
-    SetMp(kvm_ioctls::Error),
-    /// A FamStructWrapper operation has failed.
-    #[error("Failed FamStructWrapper operation: {0:?}")]
-    Fam(utils::fam::Error),
-    /// Failed to get midr_el1 from host.
-    #[error("{0}")]
-    GetMidrEl1(String),
-}
-
 #[allow(non_upper_case_globals)]
-// PSR (Processor State Register) bits.
-// Taken from arch/arm64/include/uapi/asm/ptrace.h.
+/// PSR (Processor State Register) bits.
+/// Taken from arch/arm64/include/uapi/asm/ptrace.h.
 const PSR_MODE_EL1h: u64 = 0x0000_0005;
 const PSR_F_BIT: u64 = 0x0000_0040;
 const PSR_I_BIT: u64 = 0x0000_0080;
 const PSR_A_BIT: u64 = 0x0000_0100;
 const PSR_D_BIT: u64 = 0x0000_0200;
-// Taken from arch/arm64/kvm/inject_fault.c.
-const PSTATE_FAULT_BITS_64: u64 = PSR_MODE_EL1h | PSR_A_BIT | PSR_F_BIT | PSR_I_BIT | PSR_D_BIT;
+/// Taken from arch/arm64/kvm/inject_fault.c.
+pub const PSTATE_FAULT_BITS_64: u64 = PSR_MODE_EL1h | PSR_A_BIT | PSR_F_BIT | PSR_I_BIT | PSR_D_BIT;
 
 // Following are macros that help with getting the ID of a aarch64 core register.
 // The core register are represented by the user_pt_regs structure. Look for it in
 // arch/arm64/include/uapi/asm/ptrace.h.
 
-// Gets offset of a member (`field`) within a struct (`container`).
-// Same as bindgen offset tests.
+/// Gets offset of a member (`field`) within a struct (`container`).
+/// Same as bindgen offset tests.
 macro_rules! offset__of {
     ($container:ty, $field:ident) => {
         // SAFETY: The implementation closely matches that of the memoffset crate,
@@ -86,6 +37,7 @@ macro_rules! offset__of {
         }
     };
 }
+pub(crate) use offset__of;
 
 /// Gets a core id.
 macro_rules! arm64_core_reg_id {
@@ -119,10 +71,11 @@ macro_rules! arm64_core_reg_id {
             | (($offset / std::mem::size_of::<u32>()) as u64)
     };
 }
+pub(crate) use arm64_core_reg_id;
 
-// This macro computes the ID of a specific ARM64 system register similar to how
-// the kernel C macro does.
-// https://elixir.bootlin.com/linux/v4.20.17/source/arch/arm64/include/uapi/asm/kvm.h#L203
+/// This macro computes the ID of a specific ARM64 system register similar to how
+/// the kernel C macro does.
+/// https://elixir.bootlin.com/linux/v4.20.17/source/arch/arm64/include/uapi/asm/kvm.h#L203
 macro_rules! arm64_sys_reg {
     ($name: tt, $op0: tt, $op1: tt, $crn: tt, $crm: tt, $op2: tt) => {
         /// System register constant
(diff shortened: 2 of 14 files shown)


src/cpu-template-helper/src/template/dump/aarch64.rs
```diff
@@ -20,38 +32,47 @@ pub fn config_to_template(cpu_config: &CpuConfiguration) -> CustomCpuTemplate {
...
-            Aarch64Register {
-                id: 0xffff_ffff_ffff_ffff,
-                value: 0x0000_ffff_0000_ffff_0000_ffff_0000_ffff,
-            },
-            Aarch64Register {
-                id: 0x1,
-                value: 0x0000_ffff_0000_ffff_0000_ffff_0000_ffff,
-            },
-        ]
+    const KVM_REG_SIZE_U128: u64 = 0x0040000000000000;
```
kalyazin, line 39: Is this the register ID where the u128 registers start? If so, we may need to name it a bit closer.

src/cpu-template-helper/src/template/dump/aarch64.rs
```diff
@@ -20,38 +32,47 @@ pub fn config_to_template(cpu_config: &CpuConfiguration) -> CustomCpuTemplate {
...
+
+    fn build_sample_regs() -> Aarch64RegisterVec {
```
kalyazin, line 41: will this now only exercise 128-bit regs?

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -1,80 +1,31 @@
-// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kalyazin, line 1: nit: I'm not 100% sure about the policy, but we tend to set it once when the file is created.

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,733 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+/// Struct describing a aarch64 register.
+#[derive(Debug, Clone, Copy, Versionize)]
+pub struct Aarch64RegisterOld {
+    /// ID of the register.
+    pub id: u64,
+    /// Register data.
+    pub data: u128,
+}
+
+impl Aarch64RegisterOld {
```
kalyazin, line 396: Would `Fixed` be a better name than `Old`? Old is a relative term.
Also, can we add a deprecation note that it needs not be used in new code?

src/vmm/src/arch/aarch64/vcpu.rs
```diff
@@ -0,0 +1,292 @@
...
+
+use kvm_bindings::*;
+use kvm_ioctls::VcpuFd;
+use utils::vm_memory::GuestMemoryMmap;
+
+use super::get_fdt_addr;
+use super::regs::*;
+
+/// Errors thrown while setting aarch64 registers.
+#[derive(Debug, PartialEq, Eq, thiserror::Error)]
```
kalyazin, line 18: It could ease the review process if the refactoring and code movements were extracted into separate commit(s).

src/vmm/src/cpu_config/aarch64/mod.rs
```diff
@@ -20,14 +22,27 @@ pub struct Error(#[from] pub ArchError);
...
+                        modifier.bitmap.apply(u128::from(*reg.value::<u32>())) as u32;
+                }
+                RegSize::U64 => {
+                    *reg.value_mut::<u64>() =
+                        modifier.bitmap.apply(u128::from(*reg.value::<u64>())) as u64;
+                }
+                RegSize::U128 => {
+                    *reg.value_mut::<u128>() = modifier.bitmap.apply(*reg.value::<u128>());
+                }
+                _ => unreachable!("Only 32, 64 and 128 bit wide registers are supported"),
```
kalyazin, line 44: are we going to support 256 when/if we add support for SVE?

src/vmm/src/version_map.rs
```diff
@@ -76,6 +75,8 @@ lazy_static! {
         // between firecracker minor versions and snapshot versions needs
         // to be 1-to-1 (see below)
         version_map.new_version();
+        #[cfg(target_arch = "aarch64")]
+        version_map.set_type_version(VcpuState::type_id(), 2);
```
kalyazin, line 79: nit: line 74 `v1.5 - no changes introduced` is no longer true?

src/vmm/src/cpu_config/aarch64/mod.rs
ShadowCurse: Well... we definitely can support it, but does it really make sense to change SVE registers? Anyway, currently we are limited by `u128` in cpu templates.

src/cpu-template-helper/src/template/dump/aarch64.rs
ShadowCurse: Basically to create a `Aarch64Register` I need an ID which corresponds to the 128 bit wide reg. 
`KVM_REG_SIZE_U128` is technically not a valid register ID.

src/vmm/src/cpu_config/aarch64/mod.rs
kalyazin: fair enough

src/cpu-template-helper/src/template/dump/aarch64.rs
ShadowCurse: Test was using 128 bit regs, so I did not change it. 
I can update it to use different sizes.

src/cpu-template-helper/src/template/dump/aarch64.rs
kalyazin: But it is still used as an ID later, isn't it? Can we name it to something explaining that? Maybe
```rust
const KVM_REG_SIZE_U128: u64 = 0x0040000000000000;
const U128_REG_ID_BASE: u64 = KVM_REG_SIZE_U128;
```
or something?

src/cpu-template-helper/src/template/dump/aarch64.rs
kalyazin: I think it would be a good idea. We will be confident that they work when we need them at least.

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,733 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+        assert_eq!(
+            std::mem::size_of::<T>(),
+            self.data.len(),
+            "Attempt to cast aarch64 register to incorrect type"
+        );
+
+        // # Safety:
+        // We checked that size of `T` is same as `data.len`
+        unsafe { &*(self.data.as_ptr().cast::<T>()) }
+    }
```
roypat, line 363: Mh, this needs to technically be marked as unsafe, as this is just an inlined version of `std::mem::transmute`, so all considerations of https://doc.rust-lang.org/std/mem/fn.transmute.html apply. 

There's a trait in rust-vmm/vmm-memory, `ByteValued` [[1]] which provides a safe interface for these kind of conversions. we should bound `T` by it and use its `from_slice` method to avoid unsoundness/unsafe code.

[[1]]: https://github.com/rust-vmm/vm-memory/blob/f93faaddc50933bc5f3739dbfbf8fec5dcd8e1d9/src/bytes.rs#L11

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,733 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+    /// Converts old representation of aarch64 register
+    /// into `Aarch64RegisterRef`.
+    pub fn as_reg_ref(&self) -> Aarch64RegisterRef {
+        // # Safety:
+        // Slice over known type
+        let data_ref = unsafe {
+            std::slice::from_raw_parts(
+                (&self.data as *const u128).cast::<u8>(),
+                std::mem::size_of::<u128>(),
+            )
```
roypat, line 406: https://doc.rust-lang.org/std/primitive.u32.html#method.from_ne_bytes is a safe alternative here

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,733 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+            // Registers data packed in the `vec.data`
+            // The offset for each register data is a
+            // sum of sizes of previous registers sizes.
+            let data = unsafe {
+                let data = self.vec.data.as_mut_ptr();
+                let reg_data = data.add(self.offset);
+                std::slice::from_raw_parts_mut(reg_data, reg_size)
+            };
+            self.index += 1;
+            self.offset += reg_size;
```
roypat, line 269: This should simply be `&reg_data[self.offset..self.offset + reg_size]`, or similar

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,733 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+            self.data.len(),
+            "Attempt to cast aarch64 register to incorrect type"
         );
 
-        let mut kvi: kvm_bindings::kvm_vcpu_init = kvm_bindings::kvm_vcpu_init::default();
-        vm.get_preferred_target(&mut kvi).unwrap();
-        vcpu.vcpu_init(&kvi).unwrap();
+        // # Safety:
+        // We checked that size of `T` is same as `data.len`
+        unsafe { &*(self.data.as_ptr().cast::<T>()) }
```
roypat, line 315: See my other comment about `transmute`

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,733 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+    pub fn value_mut<T>(&mut self) -> &mut T {
+        assert_eq!(
+            std::mem::size_of::<T>(),
+            self.data.len(),
+            "Attempt to cast aarch64 register to incorrect type"
+        );
+
+        // # Safety:
+        // We checked that size of `T` is same as `data.len`
+        unsafe { &mut *(self.data.as_mut_ptr().cast::<T>()) }
```
roypat, line 378: See my other comment about `transmute`

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,733 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+                unsafe {
+                    std::slice::from_raw_parts_mut(
+                        (&mut self.data as *mut $t).cast::<u8>(),
+                        std::mem::size_of::<Self::Type>(),
+                    )
+                }
+            }
+        }
+    };
+}
```
roypat, line 570: This looks to me like a reimplementation of `ByteValued` from vm-memory - can we use that trait?

roypat (changes requested):
Heyo, 

I've only had a brief look at the instances of `unsafe`, and I think there are safe alternatives for all of them. One of them can cause unsound behavior, and I think some parts of this duplicate code from vm-memory. Can you please take a look?

src/vmm/src/arch/aarch64/vcpu.rs
ShadowCurse: I can split it into 2 commits, but for now, while there are changes to be made, I would prefer not to do it.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: So you propose to add `vm-memory` as a dependency to use one trait out of it which literally does same things, but now code is not int FC, but in that dependency?  I guess we need third opinion on this.

roypat: Ah, whoops, forgot that we dont directly depend on vm-memory, but rather through the shim layer in utils (which is already a dependency of vmm) - sorry. We publically re-export the trait in `utils::vm_memory`.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: We need to return a mutable slice inside `Aarch64RegisterRefMut` and this slice is not tied to the iterator itself, but to the vec inside.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: We need a slice here, not a value

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Ok, so we can use `ByteValued` for `RegData` to replace `as_slice` and `as_mut_slice`. But for `value<T>` this will not work, because `from_slice` requires slice to be aligned, and this is not guarantied in our case.

src/vmm/src/arch/aarch64/vcpu.rs
kalyazin: I don't insist, but since some more review iterations are expected, every round would take less time if reviewers only have to look at functional changes.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Actually even using `ByteValued` for `RegData` is problematic, because it is not implemented for `u128` or any arrays that we use.

src/vmm/src/arch/aarch64/regs.rs
roypat: Ahh, right the rust iterator problem. ugh. And I guess since the registers all have difference sizes we also cant use standard lib methods that usually work around this, like chunks_mut?

src/vmm/src/arch/aarch64/vcpu.rs
ShadowCurse: I guess your biggest concern is that git shows that `regs.rs` has a bunch of changes, even thought it only contain new code, and `vcpu.rs` as new file, even though it is just old `regs.rs`, but renamed.

src/vmm/src/arch/aarch64/regs.rs
roypat: Mh, right, but taking unaligned references is undefined behavior: https://doc.rust-lang.org/reference/behavior-considered-undefined.html, you'd need to use `std::ptr::read_unaligned`.

ByteValued not being implemented for u128 and arrays is indeed annoying though - maybe we could submit a PR that implements it for u/i128 and also turns the up-to-32 array impl into one that uses const generics? (although maybe as a follow up)

src/cpu-template-helper/src/template/dump/aarch64.rs
ShadowCurse: change tests to use different sized register

src/vmm/src/arch/aarch64/vcpu.rs
kalyazin: yes, I guess so.

src/vmm/src/arch/aarch64/regs.rs
roypat: Actually, some of the unit tests are even failing because of the unaligned reads :(

ShadowCurse: Well, because all registers are stored in one continues memory, there is no way to avoid unaligned access.
Also `std::ptr::read_unaligned` is not what is needed here, because we need references.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: `ByteValued` is for values. Here we need to create a slice from a slice(vec).

ShadowCurse: Yeah... if it only was implemented on `u128`

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Nah, `chunks_mut` won't help here.

src/vmm/src/arch/aarch64/regs.rs
roypat: Mh, why do we definitely need a reference here? It looks like all use-sites of this function immediately dereference the returned reference, so returning an owned value here would not fundamentally break anything.

If you want to use a direct read from the memory inside the contiguous vec, you will have to use an unaligned read via `std::ptr::read_unaligned`. Otherwise, in debug mode you'll hit assertions, and in release mode you'll get a SIGILL because llvm will emit instructions that do not work for non-aligned accesses (iirc x86 processors do not care, but ARM does).

So the two possibility I see here are using `std::ptr::read_unaligned` and being very careful to never dereference pointers that are not aligned for their data types (and consequently dealing with raw pointers everywhere), or loosely a tiny bit of generality and using `{u16, u32, u64, u128}::from_ne_bytes` to go directly from slices to integral types (which I would honestly prefer, as it will eliminate all need for unsafe).

src/vmm/src/version_map.rs
ShadowCurse: fixed

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Replaced with `read_unaligned`.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Replaced with `read_unaligned`.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Simplified implementation. Now `RegData#` types just have a byte arrays inside.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Updated methods on `Aarch64RegisterRef` and `Aarch64RegisterRefMut` to return just a value with `read_unaligned` and added `set_value` for `Aarch64RegisterRefMut`. This does not eliminate `unsafe` fully, but guess it is better now. The remaining `unsafe` is mainly about casting types to slices and we can replace it with `ByteValued`, but it is would be literally same code.

src/cpu-template-helper/src/template/dump/aarch64.rs
ShadowCurse: I added a comment for this.

src/vmm/src/arch/aarch64/regs.rs
roypat: Actually, it seems that the mutable version of this iterator is not used anymore. Are there plans to use it in a follow up?

src/vmm/src/arch/aarch64/regs.rs
roypat: Mh, yes, this reduces the amount of unsafe and fixes the unaligned accesses. However, it is still essentially just https://doc.rust-lang.org/std/mem/fn.transmute_copy.html, so the function really should be marked as `unsafe`. Just checking for the size is not sufficient to make this safe. For example, one could try reading a u64 register into a `&mut u8` and create an invalid reference using this method, which is unsound.

Since this is only ever used with numeric types (and arrays I suppose, once we have registers wider than 128 bits), can this function instead be 

```rs
fn value<const N: usize>(&self) -> [u8; N] {
  assert_eq!(...);

  self.data.try_into().unwrap()
}
```
?

this way, its immediately compatible with https://doc.rust-lang.org/std/primitive.i128.html#method.from_le_bytes and related methods, and thus no unsafe is needed anywhere. I suppose if the extra function invocation at the call site is not desired, then this can be solved using some adhoc trait that realizes a `impl<const N: usize> TryFrom<[u8; N]> for {u8, u16, u32, u64, u128}`, which I think might be a better level of abstraction to choose here.

ShadowCurse: Hmm... actually, I think marking all these functions as `unsafe` is not a bad idea. We only use these structs/methods internally, so situations like `ref.value::<&mut u8>()` are not a real possibility. This will create more `unsafe` blocks, but again, because we control everything, it is not a real issue.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: it is used for the cpu templates in `vmm/src/cpu_config/aarch64/mod.rs`

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Ok, I adopted another approach, which does not need `unsafe`.

src/vmm/src/arch/aarch64/vcpu.rs
ShadowCurse: I renamed `regs.rs` into `registers.rs` and now git shows changes correctly.

src/vmm/src/arch/aarch64/registers.rs
```diff
@@ -0,0 +1,662 @@
...
+    /// Sets the register value.
+    /// Type `T` must be of the same length as an
+    /// underlying data slice. Otherwise this method
+    /// will panic.
+    pub fn set_value<T: Aarch64RegisterData<N>, const N: usize>(&mut self, value: T) {
+        self.data.copy_from_slice(&value.to_bytes())
+    }
+}
+
+/// Struct describing a aarch64 register.
```
kalyazin, line 364: I suggest adding a comment advising the type shouldn't be used in new code and explaining why.

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -182,21 +183,50 @@ impl KvmVcpu {
...
+    fn de_regs(&mut self, _source_version: u16) -> VersionizeResult<()> {
+        let mut regs = Aarch64RegisterVec::default();
+        for reg in self.old_regs.iter() {
+            regs.push(reg.as_reg_ref());
+        }
+        self.regs = regs;
+        Ok(())
+    }
+
+    fn ser_regs(&mut self, _target_versio: u16) -> VersionizeResult<()> {
```
kalyazin, line 210: nit: typo: _target_versio

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -182,21 +183,50 @@ impl KvmVcpu {
...
+        self.old_regs = self
+            .regs
+            .iter()
+            .map(Aarch64RegisterOld::from_reg_ref)
```
kalyazin, line 214: what's going to happen here if we feed registers bigger than 128 bit?

src/vmm/src/arch/aarch64/registers.rs
```diff
@@ -0,0 +1,662 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kalyazin, line 1: Nit: It looks like we name similar files `regs.rs`, not `registers.rs`.

src/vmm/src/arch/aarch64/registers.rs
```diff
@@ -0,0 +1,662 @@
...
+        v.push(reg9);
+
+        assert!(!v.is_empty());
+        assert_eq!(v.len(), 9);
+
+        // Test iter
+        {
+            let mut regs_iter = v.iter();
+
+            let reg1_ref = regs_iter.next().unwrap();
```
kalyazin, line 519: can we wrap the common part into a macro to avoid duplication?

src/vmm/src/arch/aarch64/registers.rs
```diff
@@ -0,0 +1,662 @@
...
+
+            assert!(regs_iter.next().is_none());
+        }
+
+        // Test iter mut
+        {
+            {
+                let mut regs_iter_mut = v.iter_mut();
+
+                let mut reg1_ref = regs_iter_mut.next().unwrap();
```
kalyazin, line 563: same as above.

src/vmm/src/arch/aarch64/registers.rs
```diff
@@ -0,0 +1,662 @@
...
+    fn test_reg_ref() {
+        let bytes = 69_u64.to_le_bytes();
+        let reg_ref = Aarch64RegisterRef::new(KVM_REG_SIZE_U64, &bytes);
+
+        assert_eq!(u64::from(reg_ref.size()), 8);
+        assert_eq!(reg_ref.value::<u64, 8>(), 69);
+    }
+
+    #[test]
+    #[should_panic]
```
kalyazin, line 610: please add comments to explain the failure cases.

src/cpu-template-helper/src/template/dump/aarch64.rs
```diff
@@ -1,6 +1,7 @@
 // Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use vmm::arch::aarch64::registers::RegSize;
```
kalyazin, line 4: nit: commit message: `Updated internal representation of aarch64 registers`
Could give a brief explanation of what the new internal representation is.

src/cpu-template-helper/src/template/dump/aarch64.rs
```diff
@@ -19,41 +31,38 @@ pub fn config_to_template(cpu_config: &CpuConfiguration) -> CustomCpuTemplate {
...
-                id: 0xffff_ffff_ffff_ffff,
-                value: 0x0000_ffff_0000_ffff_0000_ffff_0000_ffff,
-            },
-            Aarch64Register {
-                id: 0x1,
-                value: 0x0000_ffff_0000_ffff_0000_ffff_0000_ffff,
-            },
-        ]
+    // These are used as IDs to satisfy requirenments
+    // of `Aarch64Register::new`
```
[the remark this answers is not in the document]
ShadowCurse: fixed

src/vmm/src/arch/aarch64/vcpu.rs
```diff
@@ -58,116 +40,17 @@ pub enum Error {
...
-
-// EL0 Virtual Timer Registers
-arm64_sys_reg!(KVM_REG_ARM_TIMER_CNT, 3, 3, 14, 3, 2);
-
 /// Extract the Manufacturer ID from a VCPU state's registers.
 /// The ID is found between bits 24-31 of MIDR_EL1 register.
 ///
 /// # Arguments
 ///
 /// * `state` - Array slice of [`Aarch64Register`] structures, representing the registers of a VCPU
```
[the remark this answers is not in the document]
ShadowCurse: fixed

src/vmm/src/arch/aarch64/registers.rs
ShadowCurse: fixed

src/vmm/src/vstate/vcpu/aarch64.rs
ShadowCurse: fixed

src/vmm/src/vstate/vcpu/aarch64.rs
ShadowCurse: Changed `from_reg_ref` to return a error if conversion is not possible.

src/vmm/src/arch/aarch64/registers.rs
ShadowCurse: renamed it back to `regs.rs`

src/vmm/src/arch/aarch64/registers.rs
ShadowCurse: fixed

src/cpu-template-helper/src/template/dump/aarch64.rs
ShadowCurse: added additional explanation

src/vmm/src/arch/aarch64/registers.rs
ShadowCurse: fixed

kalyazin (dismissed):

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,551 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+            let reg_size = reg_size(id) as usize;
+            // # Safety:
+            // Registers data packed in the `vec.data`
+            // The offset for each register data is a
+            // sum of sizes of previous registers sizes.
+            let data = unsafe {
+                let data = self.vec.data.as_mut_ptr();
+                let reg_data = data.add(self.offset);
+                std::slice::from_raw_parts_mut(reg_data, reg_size)
+            };
```
roypat, line 267: Looking at this more closely, since this is deserialized data from a snapshot, and there is no (from what I can tell) validation in place that ensures that `vec.data` and `vec.ids` are in sync (e.g. someone might modify the snapshot file such that there are more reg ids than data in the packed register store), this can lead to an out of bounds memory access. I think we need a check somewhere (during deserialization, I suppose) that ensures that the packed data store has the correct length for the list of IDs given. 

Particularly, please extend the `SAFETY` comment to address all invariants of [`std::ptr::add`](https://doc.rust-lang.org/std/primitive.pointer.html#safety-7) and [`std::slice::from_raw_parts`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#safety)

ShadowCurse: Updated `Safety` comments.
To add verification during deserialization I added a wrapper type around `Aarch64RegisterVec` (now known as `Aarch64RegisterVecInner`).

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,685 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+    }
+
+    /// Returns an iterator over stored registers that allows register modifications.
+    fn iter_mut(&mut self) -> impl Iterator<Item = Aarch64RegisterRefMut> {
+        Aarch64RegisterVecIteratorMut {
+            index: 0,
+            offset: 0,
+            vec: self,
+        }
+    }
```
roypat, line 213: nit: I think the idiomatic way of doing this in rust is to implement [`IntoIterator`](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) for `&Aarch64RegisterVecInner` and `&mut Aarch64RegisterVecInner` respectively. That way you get for-syntax integration, too (e.g. `for item in &register_vec` instead of `for item in register_vec.iter()`)

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,685 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+                 registers data length"
+                    .to_string(),
+            ))
+        } else {
+            Ok(Self { inner })
+        }
+    }
+
+    fn version() -> u16 {
+        Aarch64RegisterVecInner::version()
```
roypat, line 283: Mh, I think using a versionize `de_fn` might be simpler here? The documentation says they can be used for "semantic validation" [[1]], so I think we can just do the check there instead of needing another wrapper.

[1]: https://github.com/firecracker-microvm/versionize_derive/blob/f0a7194791c5cbeeb55aecfa6e60cca6f2c4168a/src/lib.rs#L169

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,685 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+                data: reg_ref.value::<u128, 16>(),
+            },
+            _ => {
+                return Err(VersionizeError::Serialize(
+                    "Only 32, 64 and 128 bit wide registers are supported".to_string(),
+                ))
+            }
+        };
+        Ok(reg)
+    }
```
roypat, line 496: Mh, so the first function here panics and the second returns an error - can we unify this behavior? Also, these look like they could be `TryInto`/`TryFrom` implementations, maybe.

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,685 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+    /// Convert data type to array of bytes
+    fn to_bytes(&self) -> [u8; N];
+}
+
+macro_rules! reg_data {
+    ($t:ty, $bytes: expr) => {
+        impl Aarch64RegisterData<$bytes> for $t {
+            fn from_slice(slice: &[u8]) -> Self {
+                let mut bytes = [0_u8; $bytes];
+                bytes.clone_from_slice(slice);
```
roypat, line 513: can we use copy_from_slice here instead? copy* uses `memcpy` while clone* iterates the slice if the optimizer has a bad day [[1]]

[1]: https://github.com/rust-lang/rust/blob/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/core/src/slice/mod.rs#L2930-L2938

src/cpu-template-helper/src/template/dump/aarch64.rs
```diff
@@ -10,7 +11,18 @@ pub fn config_to_template(cpu_config: &CpuConfiguration) -> CustomCpuTemplate {
...
+            RegSize::U32 => {
+                reg_modifier!(reg.id, u128::from(reg.value::<u32, 4>()))
+            }
+            RegSize::U64 => {
+                reg_modifier!(reg.id, u128::from(reg.value::<u64, 8>()))
+            }
+            RegSize::U128 => {
+                reg_modifier!(reg.id, reg.value::<u128, 16>())
+            }
+            _ => unreachable!("Only 32, 64 and 128 bit wide registers are supported"),
```
roypat, line 24: Mh, why are these unreachable? Cant a user just edit a snapshot file to contain the id for a register with width > 128bits, and then this code gets triggers (maybe not here, but in one of the other places where panics are inserted). Can we use error handling here? I think there's some SLA we have to not panic (although we already break that in the virtio stack, I suppose)

roypat (commented):
Looks good overall - my main comments are regarding replacing the unsafe block (sorry, I only got the idea of solving the lifetime problem using mem::take an hours ago) and the error handling, the rest is pretty much just nits

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: `de_fn` is `the method is called if the field is skipped from deserialization because it does not exist in the source version of the serialized structure`. So if will not work for snapshots with latest version.

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,685 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+            // `reg_data` and `reg_size` are in bounds of `self.vec.data`.
+            let data = unsafe { std::slice::from_raw_parts_mut(reg_data, reg_size) };
+
+            self.index += 1;
+            self.offset += reg_size;
+            Some(Aarch64RegisterRefMut { id, data })
+        } else {
+            None
+        }
+    }
```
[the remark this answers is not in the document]
ShadowCurse: Nice. Updated with this version.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Converted to `TryFrom` traits.

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: Nice catch. I thought I used copy here. (even though for slices it would have used `memcopy` anyway)

src/cpu-template-helper/src/template/dump/aarch64.rs
ShadowCurse: This code is not run during snapshot restoration.

roypat (approved):
Look good! No more unsafe-golfing from me xP (sorry that review happened to piecemeal >.>)

src/vmm/src/arch/aarch64/regs.rs
roypat: ugh, versionize strikes again. nvm then

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,697 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+    /// Sets the register value.
+    /// Type `T` must be of the same length as an
+    /// underlying data slice. Otherwise this method
+    /// will panic.
+    pub fn set_value<T: Aarch64RegisterData<N>, const N: usize>(&mut self, value: T) {
+        self.data.copy_from_slice(&value.to_bytes())
+    }
+}
+
+/// Old definition of a struct describing a aarch64 register.
```
kalyazin, line 435: nit: "an aarch64"

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,697 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
-/// The ID is found between bits 24-31 of MIDR_EL1 register.
-pub fn get_manufacturer_id_from_host() -> Result<u32, Error> {
-    let midr_el1_path =
-        &PathBuf::from("/sys/devices/system/cpu/cpu0/regs/identification/midr_el1".to_string());
+/// Returns register size in bytes
+pub fn reg_size(reg_id: u64) -> u64 {
+    2_u64.pow(((reg_id & KVM_REG_SIZE_MASK) >> KVM_REG_SIZE_SHIFT) as u32)
+}
+
+/// Storage for aarch64 registers with different sizes.
```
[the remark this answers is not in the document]
ShadowCurse: Added a comment

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,697 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+        self.ids.is_empty()
+    }
 
-    let midr_el1 = std::fs::read_to_string(midr_el1_path).map_err(|err| {
-        Error::GetMidrEl1(format!("Failed to get MIDR_EL1 from host path: {err}"))
-    })?;
-    let midr_el1_trimmed = midr_el1.trim_end().trim_start_matches("0x");
-    let manufacturer_id = u32::from_str_radix(midr_el1_trimmed, 16)
-        .map_err(|err| Error::GetMidrEl1(format!("Invalid MIDR_EL1 found on host: {err}",)))?;
+    /// Appends a register to the vector, copping register data.
```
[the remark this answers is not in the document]
ShadowCurse: fixed

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,697 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
-        // not exceed 2 megabytes in size." -> https://www.kernel.org/doc/Documentation/arm64/booting.txt.
-        // We are choosing to place it the end of DRAM. See `get_fdt_addr`.
-        let regs0 = offset__of!(user_pt_regs, regs) + kreg_off;
-        let id = arm64_core_reg_id!(KVM_REG_SIZE_U64, regs0);
-        vcpufd
-            .set_one_reg(id, get_fdt_addr(mem).into())
-            .map_err(|err| Error::SetOneReg(id, err))?;
-    }
-    Ok(())
+/// Wrapper type around `Aarch64RegisterVec`.
```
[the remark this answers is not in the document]
ShadowCurse: fixed

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -157,265 +110,697 @@ arm64_sys_reg!(ID_AA64MMFR2_EL1, 3, 0, 0, 7, 2);
...
+    pub fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    /// Returns true if the vector contains no elements.
+    pub fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+
+    /// Appends a register to the vector, copping register data.
```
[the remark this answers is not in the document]
ShadowCurse: fixed

src/vmm/src/arch/aarch64/regs.rs
ShadowCurse: fixed

kalyazin (approved):

roypat (approved):

PR #3798, opened 2023-06-06
fix: Remove cpu-config benchmark lower threshold

mattschlebusch:
## Changes

Remove the lower threshold for the /cpu-config deserialization benchmarking.

The benchmark is useful in checking that  no regressions have been introduced into deserialization code for /cpu-config.

## Reason

The benchmark can cause CI pipelines to fail because the benchmark exceeds expectations, the benchmark running faster than expected is not a problem

## License Acceptance

[`CONTRIBUTING.md`][1].

---

[1]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

### DIFF
--- a/tests/integration_tests/performance/test_cpu_template_benchmark.py
+++ b/tests/integration_tests/performance/test_cpu_template_benchmark.py
@@ -9,8 +9,6 @@ import platform
 import shutil
 from pathlib import Path
 
-import pytest
-
 from framework import utils
 from framework.defs import FC_WORKSPACE_DIR
 from host_tools import proc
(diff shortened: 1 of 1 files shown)


tests/integration_tests/performance/test_cpu_template_benchmark.py
```diff
@@ -24,15 +24,15 @@
...
+        "deserialize": {"target": 0.000, "delta": 0.02},  # milliseconds
         "serialize": {"target": 0.015, "delta": 0.02},  # milliseconds
     },
     "AMD": {
-        "deserialize": {"target": 0.0037, "delta": 0.002},  # milliseconds
+        "deserialize": {"target": 0.000, "delta": 0.02},  # milliseconds
         "serialize": {"target": 0.015, "delta": 0.02},  # milliseconds
     },
     "ARM": {
-        "deserialize": {"target": 0.00037, "delta": 0.0006},  # milliseconds
```
pb8o, line 35: Would it be more clear to change delta to `"max": 0.02` and below the test would be 

```python
    assert mean < target_max, f"Benchmark result {directory} has changed!"
```

Also, we would need the same for the `serialize` part of the test no?

mattschlebusch: Sure, I'll make the tweak now.

zulinx86 (approved):

PR #3792, opened 2023-06-05
chore: Reduce cpu-config API log verbosity

mattschlebusch:
## Changes

`PUT /cpu-config` API will include the request payload which should contain the CPU config template only for DEBUG log levels or higher.

## Reason

CPU templates are potentially unbounded and could potentially make the logs very noisy. Rather opting for cleaning the logs

## Testing

Example firecracker log at `Info` log level:
```
2023-06-06T16:53:34.220881938 [anonymous-instance:fc_api] The API server received a Put request on "/cpu-config". To view the CPU template received by the API, configure log-level to DEBUG.
2023-06-06T16:53:34.221282345 [anonymous-instance:fc_api] The request was executed successfully. Status code: 204 No Content.
```

At `Debug` log level:
```
[...]

PR #3811, opened 2023-06-09
Use Debug trait formatter for API errors

mattschlebusch:
## Changes

Use `:?` to log API errors. This will use the Debug formatter when logging API errors when handling API requests.

## Reason

Instead of using the default Display formatter when logging the API request's error, use the Debug trait's formatter for a sanitized output. This will escape special characters to prevent multiline log records as follows.

### Error log with Debug trait's formatter used instead of Display
```
2023-06-09T12:22:45.009778687 [anonymous-instance:fc_api:INFO:src/api_server/src/parsed_request.rs:222] The API server received a Put request on "/cpu-config" with body "{  \"cpuid_modifiers\": [    {      \"leaf\": \"0x0\\n\\n\\n\\nTEST\\n\\n\\n\\n\",      \"subleaf\": \"0x0\",      \"flags\": 0,      \"modifiers\": [        {          \"register\": \"ecx\",          \"bitmap\": \"0b00\"        },        {          \"register\": \"edx\",          \"bitmap\": \"0b00\"        },        {          \"register\": \"ebx\",          \"bitmap\": \"0b00\"        }      ]    }  ],  \"msr_modifiers\": [    {      \"addr\": \"0x10a\",      \"bitmap\": \"0b0000000000000000000000000000000000000000000000000000000000000000\"    }  ]}".
2023-06-09T12:22:45.009970753 [anonymous-instance:fc_api:ERROR:src/api_server/src/lib.rs:246] SerdeJson(Error("Failed to parse string [0x0\n\n\n\nTEST\n\n\n\n] as a number for CPU template - ParseIntError { kind: InvalidDigit }", line: 1, column: 67))
2023-06-09T12:22:45.010068462 [anonymous-instance:fc_api:DEBUG:src/api_server/src/lib.rs:209] Total previous API call duration: 338 us.
```

### Error log without Debug trait's formatter used
```
2023-06-09T12:24:23.886189413 [anonymous-instance:fc_api:ERROR:src/api_server/src/lib.rs:247] An error occurred when deserializing the json body of a request: Failed to parse string [0x0



TEST



] as a number for CPU template - ParseIntError { kind: InvalidDigit } at line 1 column 67.
2023-06-09T12:24:23.886305141 [anonymous-instance:fc_api:DEBUG:src/api_server/src/lib.rs:209] Total previous API call duration: 390 us.
```

## License Acceptance

[`CONTRIBUTING.md`][1].

[1]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md


pb8o (dismissed):

src/api_server/src/lib.rs
```diff
@@ -243,7 +243,7 @@ impl ApiServer {
                 response
             }
             Err(err) => {
-                error!("{}", err);
+                error!("{:?}", err);
```
zulinx86, line 246: This change uses `:?` for all error types. Why didn't you use `:?` in `SerdeJson`?

kalyazin, line 246: Is it guaranteed that the Debug formatter will produce a sanitised output?

zulinx86: We might want to have a test for the error message.

mattschlebusch: > This change uses :? for all error types. Why didn't you use :? only in SerdeJson?

We can consider that, I think my justifications for preferring applying to all is based on the following:
* Technically `{}` requires that the display trait be implemented, which is not consistenty derived for all error types, but `Debug` should be.
* It is possible that there are other error types that do not involve JSON payload errors (Serde), and as a result may not be sanitized if we just focussed on `SerdeJson` errors.

zulinx86: > It is possible that there are other error types that do not involve JSON payload errors (Serde), and as a result may not be sanitized if we just focussed on SerdeJson errors.

Although I cannot see such error types other than `SerdeJson` in `api_server::parsed_request::Error`, I can agree that such error types might be added in the future.

src/api_server/src/lib.rs
```diff
@@ -314,12 +314,14 @@ mod tests {
     use utils::tempfile::TempFile;
     use utils::time::ClockType;
     use vmm::builder::StartMicrovmError;
+    use vmm::cpu_config::templates::test_utils::TEST_UNESCAPED_JSON_TEMPLATE;
```
zulinx86, line 317: nit: how about having the test input in `src/api_server/src/lib.rs` with `#[cfg(target_arch="<arch>")]`, since they are only used in the file?

zulinx86 (dismissed):

src/api_server/src/lib.rs
mattschlebusch: I need to rebase and merge commits in any event. I can add this. Although those methods are only used in this file for the moment, I have generally preferred to avoid using `target_arch` annotations, but this is my personal preference only. Another option would be to add the x86_64 template as a hard-coded string into the lib.rs file and then only run the test on x86_64. wdyt?

zulinx86: > Another option would be to add the x86_64 template as a hard-coded string into the lib.rs file and then only run the test on x86_64. wdyt?

This option also requires the `target_arch` annotation to run the test only on x86.

> Although those methods are only used in this file for the moment, I have generally preferred to avoid using target_arch annotations, but this is my personal preference only.

I also prefer to avoid using arch-dependent code, but having the test cases in different files looks worse for me. It's a matter of taste though.

zulinx86 (dismissed):

.cargo/config
```diff
@@ -1,6 +1,6 @@
 [build]
 target-dir = "build/cargo_target"
-target = "x86_64-unknown-linux-musl"
+target = "x86_64-unknown-linux-gnu"
```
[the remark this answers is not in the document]
mattschlebusch: This was a mistake. Not intended to be in scope.

src/api_server/Cargo.toml
```diff
@@ -23,4 +23,4 @@ utils = { path = "../utils" }
 vmm = { path = "../vmm" }
 
 [dev-dependencies]
-libc = "0.2.117"
+libc = "0.2.117"
```
kalyazin, line 26: nit: new line at the end of the file

kalyazin (approved):

src/api_server/Cargo.toml
mattschlebusch: I will check the other tomls to confirm, I can re-add it back in [#3792](https://github.com/firecracker-microvm/firecracker/pull/3792).

PR #3824, opened 2023-06-14
Merge vmm build script into firecracker build script

roypat:
The build script in question compiles our recommended seccomp filters by invoking seccompiler-bin. These are then included in the firecracker binary to be applied to the vmm/vcpu/api_server threads. However, only the firecracker binary crate needs this output. All usage of them outside the binary crate was in unittests. Since all of our _integration_ tests run with full production seccomp filters, there is really no point in applying them to a handful of unittests too, so those unittests now use empty seccomp filters. 

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/vmm/src/seccomp_filters/mod.rs
```diff
@@ -2,18 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
```
ShadowCurse, line 2: This is a 'one file module'. Can be renamed to `seccomp_filters.rs`.

Also nit: because this file is already heavily modified, maybe replace the `impl fmt::Display for FilterError` with `thiserror`?

src/firecracker/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "firecracker"
 version = "1.5.0-dev"
 authors = ["Amazon Firecracker team <firecracker-devel@amazon.com>"]
 edition = "2021"
-build = "../../build.rs"
+build = "build.rs"
```
pb8o, line 6: If we modify the top-level build.rs/, we would have to remember to modify this one as well?

pb8o: Since all we do is pass through `CARGO_PKG_VERSION`, maybe we should just rely on that and delete the toplevel build.rs? Though that is somewhat larger in scope.

roypat: Ah, right, I messed this up slightly, good catch! There is only supposed to be one build script anymore, but I got distracted by the seccomp changes and forgot to follow through with that

src/vmm/src/seccomp_filters/mod.rs
roypat: Good point, might as well :D

roypat: Mh, yeah a lot more can be moved to the binary crate, true - but moving `get_empty_filters` to `test_utils` doesn't "work" in the sense that then we'd have production code calling into a `test_utils` module, which is aesthetically unpleasing.

src/firecracker/src/seccomp.rs
```diff
@@ -0,0 +1,133 @@
...
+// allocation DOS caused by binary filters that are too large.
+// This limit can be safely determined since the maximum length of a BPF
+// filter is 4096 instructions and Firecracker has a finite number of threads.
+const DESERIALIZATION_BYTES_LIMIT: Option<u64> = Some(100_000);
+
+/// Retrieve the appropriate filters, based on the SeccompConfig.
+pub fn get_filters(config: SeccompConfig) -> Result<BpfThreadMap, FilterError> {
+    match config {
+        SeccompConfig::None => Ok(vmm::seccomp_filters::get_empty_filters()),
+        SeccompConfig::Advanced => get_default_filters(),
```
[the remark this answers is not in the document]
roypat: This is named "Advanced" in a lot of places (test suite, maybe even in some documentation/api requests?), so I'd rather not do that as part of this PR

src/vmm/src/seccomp_filters/mod.rs
bchalios: is the `get_empty_filters` used by production code?

roypat: When starting firecracker, theres a CLI option to disable seccomp, which causes `get_empty_filters` to be used

ShadowCurse (approved):

bchalios (approved):

PR #3832, opened 2023-06-15
test: change way api_server timings are determined

roypat:
## Changes

Now we instead directly utilize the timing information firecracker provides in its logs in our integration tests (a log message for this was previously present, and has been upgraded to info level so that we can access it in the integration tests).

## Reason

Previously, when checking that api server requests complete in under 700ms, we measured this client-side, which included whatever latency was added by python/the network stack. This also caused some intermittent failures where a request took ages to go through the requests module, but was processed within SLA by firecracker's server. 

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


tests/framework/microvm.py
```diff
@@ -205,6 +208,37 @@ def kill(self):
             self.expect_kill_by_signal = True
             utils.run_cmd("kill -9 {} || true".format(self.screen_pid))
 
+        if self.time_api_requests:
+            api_request_regex = re.compile(
+                r":fc_api\] The API server received a (?P<method>\w+) request on \"(?P<url>(/\w*)+)\" with body (?P<body>.*)\."
```
pb8o, line 213: LGTM, should we move it into its own function?

pb8o (approved):

tests/framework/microvm.py
roypat: Done :)

tests/framework/microvm.py
```diff
@@ -219,6 +225,47 @@ def kill(self):
...
+
+            match = api_request_times_regex.search(log_line)
+
+            if match and current_url != "/snapshot/create":
+                exec_time = float(match.group("execution_time")) / 1000.0
+
+                assert (
+                    exec_time <= MAX_API_CALL_DURATION_MS
+                ), f"{current_method} {current_url} API call exceeded maximum duration: {exec_time} ms. Body: {current_body}"
+
```
sudanl0, line 268: It would be good to check that we always get 1st and 2nd print in sequence so I would suggest clearing current_url here and have below 2 checks:
1.
```
if match:
    if current_url is not None:
        current_url = match.group("url")
    else:
        assert "something like got 1st pattern of prints twice"
```
2.
```
match = api_request_times_regex.search(log_line)
if match and current_url is None:
    assert "got 2nd pattern print without the 1st, so time did exceed but we don't know for which api since print having api name is missing from logs"
```

src/api_server/src/lib.rs
```diff
@@ -206,7 +206,7 @@ impl ApiServer {
 
                 let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic)
                     - request_processing_start_us;
-                debug!("Total previous API call duration: {} us.", delta_us);
+                info!("Total previous API call duration: {} us.", delta_us);
```
pb8o, line 209: Thinking more about this, since this would be a user-visible change, should we instead set the log level in our integration tests to debug?

tests/framework/microvm.py
```diff
@@ -219,6 +225,47 @@ def kill(self):
...
+    def _validate_api_response_times(self):
+        """
+        Parses the firecracker logs for information regarding api server request processing times, and asserts they
+        are within acceptable bounds.
+        """
+        # Log messages are either
+        # 2023-06-16T07:45:41.767987318 [fc44b23e-ce47-4635-9549-5779a6bd9cee:fc_api] The API server received a Get request on "/mmds".
+        # or
+        # 2023-06-16T07:47:31.204704732 [2f2427c7-e4de-4226-90e6-e3556402be84:fc_api] The API server received a Put request on "/actions" with body "{\"action_type\": \"InstanceStart\"}".
+        api_request_regex = re.compile(
```
sudanl0, line 237: nit: should we take into account "--show-level" which changes fc_api] to fc_api:INFO] or do you want to add a TODO to move to a better solution later if we are not using --show-level in CI?

src/api_server/src/lib.rs
roypat: will do!

tests/framework/microvm.py
roypat: sure

sudanl0 (approved):

sudanl0:
Please update the PR checklist before you close

tests/integration_tests/functional/test_logging.py
```diff
@@ -146,6 +148,8 @@ def test_api_requests_logs(test_microvm_with_api):
         show_level=True,
         show_log_origin=True,
     )
+    # only works if log level is Debug
+    microvm.time_api_requests = False
```
pb8o, line 152: Should we move this then to the microvm constructor? No point in timing if we know it's not going to work...

pb8o (approved):

PR #3874, opened 2023-06-26
fix(aarch64): removed warnings for missing cache info

ShadowCurse:
## Changes
- Removed unnecessary warnings when failing to get a `level` or `type` of cache for some index.
These errors were logged, but then ignored.
- Replaced error string construction when optional attributes are not available with separate `info` logs.

With these changes we don't create warning level logs during normal operation.

## Reason
Fixes warnings like this:
```
2023-06-26T10:38:56.776864578 [anonymous-instance:main:WARN:src/vmm/src/arch/aarch64/cache_info.rs:90] Could not read cache level for index 4: Failed to read cache information: No such file or directory (os error 2)
```
That appear during normal Firecracker startup and are causing confusion.
## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/arch/aarch64/cache_info.rs
+++ b/src/vmm/src/arch/aarch64/cache_info.rs
@@ -4,7 +4,7 @@
 use std::path::{Path, PathBuf};
 use std::{fs, io};
 
-use logger::warn;
+use log::warn;
 
 // Based on https://elixir.free-electrons.com/linux/v4.9.62/source/arch/arm64/kernel/cacheinfo.c#L29.
 const MAX_CACHE_LEVEL: u8 = 7;
@@ -15,8 +15,10 @@ pub(crate) enum CacheInfoError {
     FailedToReadCacheInfo(#[from] io::Error),
     #[error("Invalid cache configuration found for {0}: {1}")]
     InvalidCacheAttr(String, String),
-    #[error("Cannot proceed with reading cache info.")]
-    MissingCacheConfig,
+    #[error("Cannot read cache level.")]
+    MissingCacheLevel,
+    #[error("Cannot read cache type.")]
+    MissingCacheType,
     #[error("{0}")]
     MissingOptionalAttr(String, CacheEntry),
 }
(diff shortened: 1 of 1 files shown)


src/vmm/src/arch/aarch64/cache_info.rs
```diff
@@ -298,22 +288,20 @@ pub(crate) fn read_cache_config(
...
+            // do not propagate it upwards. We were prudent enough to log it.
             Err(Error::MissingOptionalAttr(msg, cache)) => {
                 let level = cache.level;
                 append_cache_level(cache_l1, cache_non_l1, cache);
                 if !msg.is_empty() && !logged_missing_attr {
-                    warn!(
-                        "{}",
-                        format!("Could not read the {} for cache level {}.", msg, level)
-                    );
+                    warn!("Could not read the {} for cache level {}.", msg, level);
```
pb8o, line 304: Rather than logging the negative case, can we log only the positives? So instead of this:

```
WARN Could not read ... for cache level 4.
```

we have this in the logs:

```
INFO Found cache level 1
INFO Found cache level 2
INFO Found cache level 3
```

ShadowCurse: This warning is about optional attributes and it is not an issue here. I think logging with `INFO` just to log is not a way. (no one really cares if caches are found successfully).

pb8o: If it's not important to log if we find a cache successfully or unsuccessfully, then why do we log at all?

ShadowCurse: The main change in this PR is in these lines: https://github.com/firecracker-microvm/firecracker/blob/9ab568b4892836841a358126f02f2d947e86578b/src/vmm/src/arch/aarch64/cache_info.rs#L90
Previously we were emitting warnings when we could not read level of the cache, but then we were just ignoring the error, because it is not critical.

pb8o: And what are we doing now? It seems we are still logging a warning?

ShadowCurse: The remaining warning is about missing attributes. I did not change the warning itself, just removed `format!` macro. I actually never seen this particular warning, but I don't mind converting it into `info!` as well.

pb8o: If we are making changes here, I think we should really consider not emitting a warning for something that is normal operations. One way may be to not log any warnings, at all. Another may be to log what caches we find, with Info or Debug.

Also include before/after logging in the commit so we understand what has changed.

ShadowCurse: Replaced this particular warn and replaced it with `info` level logs.

ShadowCurse: @kalyazin Added a more descriptive commit message.

kalyazin: > @kalyazin Added a more descriptive commit message.

Apart from couple typos in the message ("cahce", "they was"),
> The error handing of such situations did not change.

this is a bit ambiguous. Maybe provide examples of the errors to make it clear that they are separate from the warnings that have been removed.

ShadowCurse: @kalyazin Fixed the typos. Regarding examples of errors: there are no changes in error, the whole change is basically in just removing couple `warn!(...)` lines. I have no idea how to make it more clear.

kalyazin (approved):

src/vmm/src/arch/aarch64/cache_info.rs
```diff
@@ -437,10 +425,7 @@ mod tests {
         let res = CacheEntry::from_index(0, engine.store.as_ref());
         assert!(res.is_err());
         // We did create the level file but we still do not have the type file.
-        assert_eq!(
-            format!("{}", res.unwrap_err()),
-            "Cannot proceed with reading cache info."
-        );
+        matches!(res.unwrap_err(), CacheInfoError::MissingCacheType);
```
sudanl0, line 428: the original code would have panicked with the assert, is there any particular reason to change this behaviour?

ShadowCurse: Yeah, I forgot to add `assert!` around `matches!`. Fixed now.

sudanl0 (approved):

kalyazin (approved):

PR #3893, opened 2023-06-27
New CI artifacts: first batch of changes 

pb8o:
## Changes

Some cleanups done during the new CI artifacts work, that can be merged independently.

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/.buildkite/pipeline_cross.py
+++ b/.buildkite/pipeline_cross.py
@@ -39,7 +39,7 @@ def cross_steps():
     snap_instances = ["m5d.metal", "m6i.metal", "m6a.metal"]
     groups = []
     commands = [
-        "./tools/devtool -y create_snapshot_artifacts",
+        "./tools/devtool -y sh ./tools/create_snapshot_artifact/main.py",
         "mkdir -pv snapshots/{instance}_{kv}",
         "sudo chown -Rc $USER: snapshot_artifacts",
         "mv -v snapshot_artifacts/* snapshots/{instance}_{kv}",
--- a/tests/framework/stats/consumer.py
+++ b/tests/framework/stats/consumer.py
@@ -177,12 +177,9 @@ class LambdaConsumer(Consumer):
         """Initialize the LambdaConsumer."""
         super().__init__(metadata_provider)
         self._func = func
-        self._func_kwargs = func_kwargs
+        self._func_kwargs = func_kwargs or {}
 
     def ingest(self, iteration, raw_data):
-        """Execute the function with or without arguments."""
+        """Execute the function."""
         self._iteration = iteration
-        if self._func_kwargs:
-            self._func(self, raw_data, **self._func_kwargs)
-        else:
-            self._func(self, raw_data)
+        return self._func(self, raw_data, **self._func_kwargs)
--- a/tests/framework/stats/core.py
+++ b/tests/framework/stats/core.py
@@ -51,6 +51,8 @@ class Core:
             name=name, iterations=iterations, results={}, custom=custom
         )
         self._failure_aggregator = CoreException()
+        self.metrics_test = None
+        self.metrics = None
 
     def add_pipe(self, producer: Producer, consumer: Consumer, tag=None):
         """Add a new producer-consumer pipe."""
@@ -65,11 +67,19 @@ class Core:
         for tag, pipe in self._pipes.items():
             for iteration in range(iterations):
                 raw_data = pipe.producer.produce()
-                if isinstance(raw_data, types.GeneratorType):
-                    for data in raw_data:
-                        pipe.consumer.ingest(iteration, data)
-                else:
-                    pipe.consumer.ingest(iteration, raw_data)
+                if not isinstance(raw_data, types.GeneratorType):
+                    raw_data = [raw_data]
+                for data in raw_data:
+                    raws = pipe.consumer.ingest(iteration, data)
+                    if raws is not None:
+                        dimensions = self.custom.copy()
+                        dimensions["test"] = self.metrics_test
+                        self.metrics.set_dimensions(dimensions)
+                        for name, val, unit in raws:
+                            self.metrics.put_metric(name, val, unit)
+                            self.metrics.set_property("iteration", iteration)
+                        self.metrics.flush()
+
             try:
                 stats, custom = pipe.consumer.process(fail_fast)
             except (ProcessingException, AssertionError) as err:
--- a/tests/framework/utils_vsock.py
+++ b/tests/framework/utils_vsock.py
@@ -14,7 +14,7 @@ from threading import Thread
 ECHO_SERVER_PORT = 5252
 SERVER_ACCEPT_BACKLOG = 128
 TEST_CONNECTION_COUNT = 50
-BLOB_SIZE = 20 * 1024 * 1024
+BLOB_SIZE = 1 * 1024 * 1024
 BUF_SIZE = 64 * 1024
 VSOCK_UDS_PATH = "v.sock"
 
@@ -77,12 +77,12 @@ class HostEchoWorker(Thread):
             self.hash = hash_obj.hexdigest()
 
 
-def make_blob(dst_dir):
+def make_blob(dst_dir, size=BLOB_SIZE):
     """Generate a random data file."""
     blob_path = os.path.join(dst_dir, "vsock-test.blob")
 
     with open(blob_path, "wb") as blob_file:
-        left = BLOB_SIZE
+        left = size
         blob_hash = hashlib.md5()
         while left > 0:
             count = min(left, 4096)
@@ -205,10 +205,8 @@ def _vsock_connect_to_guest(uds_path, port):
 
 def _copy_vsock_data_to_guest(ssh_connection, blob_path, vm_blob_path, vsock_helper):
     # Copy the data file and a vsock helper to the guest.
+
     cmd = "mkdir -p /tmp/vsock"
-    cmd += " && mount -t tmpfs tmpfs -o size={} /tmp/vsock".format(
-        BLOB_SIZE + 1024 * 1024
-    )
     ecode, _, _ = ssh_connection.execute_command(cmd)
     assert ecode == 0, "Failed to set up tmpfs drive on the guest."
 
--- a/tests/integration_tests/performance/test_network_latency.py
+++ b/tests/integration_tests/performance/test_network_latency.py
@@ -91,16 +91,19 @@ def consume_ping_output(cons, raw_data, requests):
         assert len(time) == 1
         times.append(time[0])
 
-    times.sort()
+    sorted_times = sorted(times)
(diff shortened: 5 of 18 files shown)


.buildkite/pipeline_cross.py
```diff
@@ -39,7 +39,7 @@ def cross_steps():
     snap_instances = ["m5d.metal", "m6i.metal", "m6a.metal"]
     groups = []
     commands = [
-        "./tools/devtool -y create_snapshot_artifacts",
+        "./tools/devtool -y sh ./tools/create_snapshot_artifact/main.py",
```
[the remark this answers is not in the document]
pb8o: Yep, this part works.

tests/framework/utils_vsock.py
```diff
@@ -205,10 +205,8 @@ def _vsock_connect_to_guest(uds_path, port):
 
 def _copy_vsock_data_to_guest(ssh_connection, blob_path, vm_blob_path, vsock_helper):
     # Copy the data file and a vsock helper to the guest.
+
     cmd = "mkdir -p /tmp/vsock"
-    cmd += " && mount -t tmpfs tmpfs -o size={} /tmp/vsock".format(
-        BLOB_SIZE + 1024 * 1024
-    )
```
zulinx86, line 211: Do you have any context why we had this so far?

pb8o: No, as far as I can see tmpfs is already mounted in the guest with around 128MB

tests/integration_tests/performance/test_network_latency.py
```diff
@@ -113,7 +114,10 @@ def test_network_latency(
 
     Send a ping from the guest to the host.
     """
-    requests = 1000
+
+    # each iteration is 6 seconds
+    # * 30 iterations = 5 minutes
+    requests = 30
```
roypat, line 120: Instead of doing this, can we instead "backdate" the datapoints that we publish to cloudwatch? E.g. have the consumer return a 4-tuple that also include a timestamp? It seems that from a quick glance at the aws-embedded-metrics source, this should be possible by manually updating `metrics.meta['TimeStamp'] = ...`.

tests/framework/stats/core.py
```diff
@@ -65,11 +67,19 @@ def run_exercise(self, fail_fast=False) -> Result:
...
+                if not isinstance(raw_data, types.GeneratorType):
+                    raw_data = [raw_data]
+                for data in raw_data:
+                    raws = pipe.consumer.ingest(iteration, data)
+                    if raws is not None:
+                        dimensions = self.custom.copy()
+                        dimensions["test"] = self.metrics_test
+                        self.metrics.set_dimensions(dimensions)
+                        for name, val, unit in raws:
+                            self.metrics.put_metric(name, val, unit)
```
roypat, line 79: Since these are emitted at a high rate, we should set `storage_resolution = StorageResolution.HIGH`

tests/framework/stats/consumer.py
```diff
@@ -177,12 +177,9 @@ def __init__(
         """Initialize the LambdaConsumer."""
         super().__init__(metadata_provider)
         self._func = func
-        self._func_kwargs = func_kwargs
+        self._func_kwargs = func_kwargs if func_kwargs else {}
```
roypat, line 180: nit:

```suggestion
        self._func_kwargs = func_kwargs or {}
```

tests/integration_tests/performance/test_snapshot_restore_performance.py
```diff
@@ -209,10 +209,10 @@ def test_snapshot_scaling_vcpus(
     st_core.run_exercise()
 
 
-# exponent=8 takes around 400s seconds
+# mem_exponent=7 takes around 100s
 @pytest.mark.nonci
 @pytest.mark.timeout(10 * 60)
-@pytest.mark.parametrize("mem_exponent", range(1, 9))
+@pytest.mark.parametrize("mem_exponent", range(1, 8))
```
roypat, line 215: This should also remove the associated entries from the baseline files (unless that happens automatically the next time we collect?)

tests/integration_tests/performance/test_network_latency.py
roypat: This will also be easier to adapt for the remaining perf tests

tests/framework/stats/core.py
pb8o: Hmm I chose the numbers in the latency test to have around 1 datapoint per minute. It shouldn't hurt however.

roypat: huh, but the test is setup to ping 30 times with an interval of 0.2s between pings, so there'd be 5 data points per seconds (which is too much even for high resolution metrics, but oh well)

tests/integration_tests/performance/test_snapshot_restore_performance.py
pb8o: hmm right, I think it won't happen during baselines.

tests/framework/utils_vsock.py
zulinx86: Hmm, I cannot see the mounted tmpfs on /tmp inside the guest. Am I missing something?
```
 def _copy_vsock_data_to_guest(ssh_connection, blob_path, vm_blob_path, vsock_helper):
     # Copy the data file and a vsock helper to the guest.
+    cmd = "mount | grep tmpfs"
+    ecode, stdout, stderr = ssh_connection.execute_command(cmd)
+    print(stdout.read())
+    print(stderr.read())
```
```
$ tools/devtool test -- integration_tests/functional/test_vsock.py --capture=tee-sys
integration_tests/functional/test_vsock.py::test_vsock[ubuntu]
devtmpfs on /dev type devtmpfs (rw,relatime,size=119948k,nr_inodes=29987,mode=755)
tmpfs on /dev/shm type tmpfs (rw,nosuid,nodev)
tmpfs on /run type tmpfs (rw,nosuid,nodev,mode=755)
tmpfs on /run/lock type tmpfs (rw,nosuid,nodev,noexec,relatime,size=5120k)
tmpfs on /sys/fs/cgroup type tmpfs (ro,nosuid,nodev,noexec,mode=755)
```

tests/integration_tests/performance/test_network_latency.py
pb8o: It may be possible, but it may be using the API in a way that is not documented.

What would it buy us, beyond being able do the metrics outside of the "stats" code?

One downside if the test breaks we would lose all the datapoints, though this doesn't happen often.

tests/framework/utils_vsock.py
zulinx86: But yeah, as far as I see, having the file on tmpfs or root (/dev/vda) seems not to matter, since we don't have any performance-sensitive tests using it. In my understanding, the benefit of using tmpfs is just fast file access: https://man7.org/linux/man-pages/man5/tmpfs.5.html

tests/integration_tests/performance/test_network_latency.py
roypat: The underlying put-metrics-data seems to explicitly support backdating the metrics: https://docs.aws.amazon.com/cli/latest/reference/cloudwatch/put-metric-data.html.

If we don't backdate the timestamps, we'll just end up with a lot of data points at the same time, which will not really allow us to analyse/plot them usefully. And some of our other tests cannot really be split apart because they already only collect ~20 samples (vsock/tcp throughput for example). Reducing this even more would degrade the value of the data we collect (even if we allow each pass its own warm-up period, which does not seem like a good use of resources because we'd be spending more time warming up than actually collecting performance data). So for these, we'd just have to increase the overall duration, but at this point it makes more sense to me to shard the test across multiple instances, and have each pass happen on its own .metal instance to account for hardware variability. 

I suppose splitting this test, and assigning proper timestamps aren't mutually exclusive, but I don't really see why we _wouldn't_ want properly timestamped data. And for other test, I don't see how this splitting would work :(

tests/framework/stats/core.py
```diff
@@ -65,11 +67,19 @@ def run_exercise(self, fail_fast=False) -> Result:
...
+                            self.metrics.set_property("iteration", iteration)
+                        self.metrics.flush()
```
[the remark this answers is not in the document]
pb8o: We do, our metrics is a proxy class https://github.com/firecracker-microvm/firecracker/blob/main/tests/host_tools/metrics.py#L81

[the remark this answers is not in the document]
roypat: Ah, missed that, thanks!

tests/framework/stats/core.py
pb8o: Oh right, so it is 1 datapoint for each iteration, every 6 seconds, sorry. 

I agree it may not work for every test. If they cannot be split into iterations then it doesn't make sense for that test. In that case we can change the data format and return an optional timestamp. If the timestamp is present, we backdate, otherwise we use the regular timestamps.

pb8o: But yeah we can add that if we find this kind of data valuable

tests/framework/utils_vsock.py
pb8o: Oh my bad, I think I looked in the new rootfs... in the old rootfs there seems to be enough free space that it doesn't make a difference, but maybe for the perf tests it will?

tests/integration_tests/performance/test_network_latency.py
roypat: I've synced with Pablo offline, it seems that even if we backdate the timestamps, as long as they are high-frequency, cloudwatch will aggregate at the log level already (and combine everything at 1 second granularity). It does keep ordering though, so there's not really a point in doing this at this stage (maybe for other tests, but lets experiment with the output of this test for a bit first). Anything else can definitely be done later :)

tests/integration_tests/performance/test_network_latency.py
```diff
@@ -91,16 +91,19 @@ def consume_ping_output(cons, raw_data, requests):
...
     cons.consume_stat(
-        st_name="Percentile90", ms_name=LATENCY, value=times[int(requests * 0.9)]
+        st_name="Percentile90", ms_name=LATENCY, value=sorted_times[int(requests * 0.9)]
     )
     cons.consume_stat(
-        st_name="Percentile99", ms_name=LATENCY, value=times[int(requests * 0.99)]
+        st_name="Percentile99",
+        ms_name=LATENCY,
+        value=sorted_times[int(requests * 0.99)],
     )
```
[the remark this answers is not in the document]
pb8o: Yeah this looks unused...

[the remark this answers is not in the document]
pb8o: I will clean it up in a future PR

zulinx86 (approved):

PR #3736, opened 2023-05-25
Streamline performance testing setup

roypat:
## Changes

This PR streamlines our existing performance testing setup by
- Removing nonsensical values (e.g. pkt_loss for network latency test, which is always 0 in every baseline, and has never been non-zero)
- Removing highly correlated tests (e.g. block tests with mode read and randread had a correlation of r = 0.97 across all runs since February)
- For throughput tests, remove suboptimal test setups.

It also changes the way test parameterization is handled, by moving configuration from the baseline files to `@pytest.mark.parameterize` decorators. This will make it more obvious how far a running performance test run has progressed, as well as make it easier to see what exact configuration failed at a glance.

No changes in this PR require regathering of baseline - any changes that would involve new baselines will be made in follow ups. 

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following
Developer Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

---

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #3894, opened 2023-06-28
Cleanups of performance test code

roypat:
## Changes

This PR contains various clean ups to performance testing code which do not require baseline regathering. The only interesting change is that retries will not be handled at the pytest level instead of the buildkite level, which should significantly improve our compute resource utilization (as we no longer retry _all_ performance tests if a single one fails), and lead to less spurious failures (e.g. if a _different_ test fails on each whole-suite retry)

## Reason

Originally, the plan was to accumulate changes on a feature branch, so that baseline gathering can be batched, but it makes sense to keep the number of commits on the feature branch to a minimum (by only including those that actually depend on new baselines), to keep rebasing as painless as possible.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


tests/integration_tests/performance/test_block_performance.py
```diff
@@ -172,8 +172,8 @@ def read_values(cons, numjobs, env_id, mode, bs, measurement, logs_path):
             f"{logs_path}/{env_id}/{mode}{bs}/{mode}"
             f"{bs}_{measurement}.{job_id + 1}.log"
         )
-        file = open(file_path, encoding="utf-8")
-        lines = file.readlines()
+        with open(file_path, encoding="utf-8") as file:
+            lines = file.readlines()
```
pb8o, line 176: How about `file_path.read_text(encoding("utf-8").splitlines()`

pb8o (approved):

tests/integration_tests/performance/test_block_performance.py
roypat: ohh, nice, didn't know about that one! is there something similar we could use for the json files?

pb8o: I do `json.loads(file_path.read_text(encoding...))`

roypat: Alright, github doesnt display your comment for some reason, but I did the `json.loads(path.read_text())` thingy anyway

pb8o (approved):

pb8o (dismissed):

tests/framework/utils.py
```diff
@@ -510,6 +510,33 @@ def get_cpu_percent(pid: int, iterations: int, omit: int) -> dict:
     return cpu_percentages
 
 
+def summarize_cpu_percent(cpu_percentages: dict):
```
zulinx86, line 513: Please let me ask if it's fine to drop DEBUG feature (storing all data), just in case.

tests/framework/utils_iperf.py
```diff
@@ -0,0 +1,132 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
zulinx86, line 1: The year should be 2023.

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -279,7 +278,8 @@ def test_block_performance(
             )
             st_cons = st.consumer.LambdaConsumer(
                 metadata_provider=DictMetadataProvider(
-                    CONFIG["measurements"], BlockBaselinesProvider(env_id, fio_id)
+                    CONFIG["measurements"],
+                    BlockBaselinesProvider(env_id, fio_id, CONFIG),
```
zulinx86, line 284: Is there any reason why we use `CONFIG` here different from other perf tests?

tests/framework/utils_iperf.py
```diff
@@ -0,0 +1,132 @@
...
+from framework import utils
+from framework.utils import CmdBuilder, CpuMap, get_cpu_percent, summarize_cpu_percent
+
+DURATION = "duration"
+IPERF3_END_RESULTS_TAG = "end"
+THROUGHPUT = "throughput"
+CPU_UTILIZATION_VMM = "cpu_utilization_vmm"
+CPU_UTILIZATION_VCPUS_TOTAL = "cpu_utilization_vcpus_total"
+
+# Dictionary mapping modes (guest-to-host, host-to-guest, bidirectional) to arguments passed to the iperf3 clients spawned
```
zulinx86, line 18: nit: Reordering modes in comment looks more prettier:
```suggestion
# Dictionary mapping modes (bidirectional, guest-to-host, host-to-guest) to arguments passed to the iperf3 clients spawned
```

tests/integration_tests/performance/test_network_tcp_throughput.py
```diff
@@ -84,137 +59,38 @@ def get(self, metric_name: str, statistic_name: str) -> dict:
...
-    with concurrent.futures.ThreadPoolExecutor() as executor:
-        futures = []
-        cpu_load_future = executor.submit(
-            get_cpu_percent, basevm.jailer_clone_pid, cpu_load_runtime, omit
+
+    def guest_command(self, port_offset):
+        cmd = (
+            super()
+            .guest_command(port_offset)
+            .with_arg("--client", self._host_ip)
```
zulinx86, line 81: I think we can move this to the parent's `guest_command()`, since both iperf3 and iperf3-vsock commands can take either `--client <host>` or `-c <host>`:

https://iperf.fr/iperf-doc.php
> -c, --client host | Run iPerf in client mode, connecting to an iPerf server running on host.

https://github.com/stefano-garzarella/iperf-vsock/blob/master/docs/invoking.rst
>        -c, --client host
>              run  in  client  mode,  connecting  to the specified server.  By
>              default, a test consists of sending data from the client to  the
>              server, unless the -R flag is specified.

zulinx86 (commented):
Please let me ask if you've confirmed the existing pipelines work with this change, just in case.

tests/framework/utils_iperf.py
roypat: Good catch! I really need to start paying attention to that when I copy my copyright lines haha

tests/integration_tests/performance/test_block_performance.py
roypat: All other tests have been updated to use pytest-level parameterization instead of reading things such as "what numbers of vcpus to test" from the baseline file in https://github.com/firecracker-microvm/firecracker/pull/3736. For the block test, I couldn't do that, since it changes the performance characteristics in a way that'll require baseline regathering (since currently, subsequent block tests reuse the same VM, but with pytest parameterization, each test will get its own VM). That's why the change has been pushed back for the block test, and why we use the `CONFIG` dict differently here

tests/framework/utils_iperf.py
roypat: mh, I prefer having bidirectional last because that way you get know about what the two directions are first, before learning that we also support doing both at the same time.

zulinx86: I see. Then, how about changing `MODE_MAP`? It's a matter of taste, though.
```
MODE_MAP = {"g2h": [""], "h2g": ["-R"], "bd": ["", "-R"]}
```

tests/framework/utils.py
zulinx86: > We have plans to properly implement this in follow ups (@pb8o already did it for the network latency test).

Sounds great. Thanks.

tests/framework/utils_iperf.py
roypat: oh, yeah, matching they should

tests/integration_tests/performance/test_network_tcp_throughput.py
zulinx86: hmm, I'd prefer to change it to `host` rather than `client`, since the `-c, --client` means turning on the client mode and the passed arg means "host".

roypat: ah, yeah, you're right. What a confusingly named argument. Maybe `connect_to` for the instance variable?

tests/integration_tests/performance/test_block_performance.py
zulinx86: ah sorry, I missed the commit message. thanks for your explanation.

zulinx86 (approved):
All the reset comments are nitpicking.

tests/integration_tests/performance/test_network_tcp_throughput.py
zulinx86: That looks great to me!

pb8o (approved):

PR #3895, opened 2023-06-28
Remove check in integration test for balloon device trying to create a v0.23.0 snapshot

bchalios:
## Changes

Remove a check in integration test for balloon device that tries to create a snapshot targeting `v0.23.0`.

## Reason

`test_snapshot_compatibility` virtio-balloon integration test was trying to ensure we cannot create a snapshot from a microVM with a balloon device targeting Firecracker `v0.23.0`. The test expects to find an API response mentioning that the target Firecracker version does not support balloon. Instead, at times it receives an error due to the fact that the "target Firecracker version does not support notification suppression".

The latter error message is emitted when we try to create a snapshot targeting Firecracker version earlier `1.0.0`, since notification suppression support was introduced with `1.0.0`.

In fact, the virtio-balloon specific check should never be triggered, only the notification suppression one. The test was working was by chance. The notification suppression condition was not triggered because we were trying to take a snapshot immediately after booting the VM, before the block device was initialized.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


roypat (dismissed):
I suppose if we want to test the "cannot target version x if device y is loaded and didn't exist at version x" scenario, we could do a test that tries to snapshot a VM with a rng device at target version 1.3 or sth

pb8o (approved):

roypat (approved):

PR #3897, opened 2023-06-28
test: Disable Turbo Boost

roypat:
Turbo boost allows the CPU of the .metal instance running our test suite to temporarily operate at a higher frequency. Since some of our tests are performance tests, this add non-deterministic variability into the processor frequency. Therefore, disable it.

Also explicitly set the scaling governor to "performance" to prevent the kernel from adjusting CPU frequency for us.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tools/devtool
+++ b/tools/devtool
@@ -601,6 +601,19 @@ cmd_test() {
     # We need to run a privileged container to get that kind of access.
     env |grep -P "^(AWS_EMF_|BUILDKITE_)" > env.list
 
+    if [[ "$BUILDKITE" = "true" ]]; then
+      # Disable turbo boost. Some of our tests are performance tests, and we want minimum variability wrt processor frequency
+      # See also https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/processor_state_control.html
+      sudo sh -c "echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo" &> /dev/null
+
+      # The governor is a linux component that can adjust CPU frequency. "performance" tells it to always run CPUs at
+      # their maximum safe frequency. It seems to be the default for Amazon Linux, but it doesn't hurt to make this explicit.
+      # See also https://wiki.archlinux.org/title/CPU_frequency_scaling
+      sudo sh -c "echo "performance" | tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor" &> /dev/null
+
+      say "Detected CI, tuning CPU frequency scaling for reduced variability"
+    fi
+
     run_devctr \
         --privileged \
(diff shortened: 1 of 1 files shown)


zulinx86 (dismissed):

tools/devtool
```diff
@@ -601,6 +601,15 @@ cmd_test() {
     # We need to run a privileged container to get that kind of access.
     env |grep -P "^(AWS_EMF_|BUILDKITE_)" > env.list
 
+    # Disable turbo boost. Some of our tests are performance tests, and we want minimum variability wrt processor frequency
+    # See also https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/processor_state_control.html
+    sudo -n sh - c "echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo" 2>/dev/null
```
pb8o, line 606: Maybe save the value so we can restore it later

```sh
local saved_no_turbo=$(/sys/devices/system/cpu/intel_pstate/no_turbo
echo 1 | sudo tee /sys/devices/system/cpu/intel_pstate/no_turbo &>/dev/null
```

tools/devtool
```diff
@@ -618,6 +627,8 @@ cmd_test() {
 
     ret=$?
 
+    echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo
```
pb8o, line 630: and then

```sh
echo $saved_no_turbo | sudo tee /sys/devices/system/cpu/intel_pstate/no_turbo &>/dev/null
```

tools/devtool
zulinx86: Should we do the same thing for scaling_governor?

roypat: That might be a bit more complicated because different CPUs might have different configs. From my testing on EC2 instances and local machines, "performance" seems to be the default here anyway, so not sure if its worth it? Maybe only doing this if the test is running in the Infra would be better, if there's an easy way to check for this in the bash script (maybe presence of some buildkite env var)?

zulinx86: > Maybe only doing this if the test is running in the Infra would be better, if there's an easy way to check for this in the bash script (maybe presence of some buildkite env var)?

I agree with applying this only on our infra. `BUILDKITE="true"` seems to fit it.

https://buildkite.com/firecracker/firecracker-pr/builds/4268#018902ee-1b3c-485c-aae9-72e09809e719
> BUILDKITE="true"

roypat: Ahh, amazing, I'll use that then!

roypat: I thought about that, but as long as we also have perf tests in the normal CI, having it enabled always made sense to me

PR #3940, opened 2023-07-05
test(perf): Accumulated Block Test fixes + baselines

roypat:
This PR accumulates some fixes to the block performance test that all each require new baselines. These include:
- Pinning the fio workers to vcpus inside of the guest to prevent guest-scheduling to impact test results
- Change parameterization from baseline level to pytest level to improve readability of test progress and do not reuse microvms across multiple tests
- Introduce a write test, as the randrw fio mode tries to balance reads and writes, leading to its results being highly correlated with the read tests. 

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/performance/test_block_performance.py
+++ b/tests/integration_tests/performance/test_block_performance.py
@@ -27,8 +27,7 @@ from integration_tests.performance.configs import defs
 TEST_ID = "block_performance"
 kernel_version = get_kernel_version(level=1)
 CONFIG_NAME_REL = "test_{}_config_{}.json".format(TEST_ID, kernel_version)
-CONFIG_NAME_ABS = os.path.join(defs.CFG_LOCATION, CONFIG_NAME_REL)
-CONFIG = json.load(open(CONFIG_NAME_ABS, encoding="utf-8"))
+CONFIG_NAME_ABS = defs.CFG_LOCATION / CONFIG_NAME_REL
 
 FIO = "fio"
 
@@ -298,9 +297,13 @@ def test_block_performance(
             "bs": fio_block_size,
         },
     )
+
+    raw_baselines = json.loads(CONFIG_NAME_ABS.read_text("utf-8"))
+
     st_cons = st.consumer.LambdaConsumer(
         metadata_provider=DictMetadataProvider(
-            CONFIG["measurements"], BlockBaselinesProvider(env_id, fio_id)
+            raw_baselines["measurements"],
+            BlockBaselinesProvider(env_id, fio_id, raw_baselines),
         ),
         func=consume_fio_output,
         func_kwargs={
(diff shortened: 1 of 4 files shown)


tests/integration_tests/performance/test_block_performance.py
```diff
@@ -79,7 +81,11 @@ def run_fio(env_id, basevm, mode, bs):
         .with_arg("--ioengine=libaio")
         .with_arg("--iodepth=32")
         .with_arg(f"--ramp_time={CONFIG['omit']}")
-        .with_arg(f"--numjobs={CONFIG['load_factor'] * basevm.vcpus_count}")
+        .with_arg(f"--numjobs={num_jobs}")
+        # Set affinity of the entire fio process to a set of vCPUs equal in size to number of workers
+        .with_arg(f"--cpus_allowed={','.join(str(i) for i in range(num_jobs))}")
+        # Instruct fio to pin one worker per vcpu
+        .with_arg("--cpus_allowed_policy=split")
```
pb8o, line 88: Have we tested how does this perform? I.e. does it pass the block perf test, or would we need to gather baselines?

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -66,6 +66,8 @@ def run_fio(env_id, basevm, mode, bs):
     """Run a fio test in the specified mode with block size bs."""
     logs_path = f"{basevm.jailer.chroot_base_with_id()}/{env_id}/{mode}{bs}"
 
+    num_jobs = CONFIG["load_factor"] * basevm.vcpus_count
```
kalyazin, line 69: could be helpful to add a comment why `load_factor` is a part of this calculation

roypat: I would be, but I also don't know much more behind its intention beyond what the name would imply :sweat_smile:

tests/integration_tests/performance/test_block_performance.py
roypat: I don't expect it to change anything, but let me run a pipeline anyway

tests/integration_tests/performance/test_block_performance.py
kalyazin: I think even this is a valuable piece of info ;)

roypat: Well, the comment would be "load factor determines how many workers to spawn per vcpu", which only raises more questions: Why would we want to have more than 1 worker per vcpu? What's the impact of having it not be 1? I'll just... remove the variable I think :joy:

kalyazin: I was only able to grep this variable set to 1 in the code, which is currently a no-op. I would ideally look for a comment explaining why we believe it's a part of the equation and why it is going to hold when/if we introduce other values than 1.

roypat: It's been there (and set to 1) since the beginning of time, with no explanation.

tests/integration_tests/performance/test_block_performance.py
roypat: Interesting, it does make m5d.metal fail. Well, I'll combine this change with the rest of my changes that require new baselines then and just bite the bullet and recollect

pb8o (approved):

kalyazin (approved):

PR #3944, opened 2023-07-06
Set MAP_POPULATE on guest mmap

cperciva:
This directs the Linux kernel to set up page tables eagerly instead of the default lazy behaviour, reducing the FreeBSD kernel boot time from 29 ms to 24 ms.

Note that the total VM launch time improvement may be less than this since the initial mmap call will take longer; however mapping eagerly has the advantage of reducing context-switching overhead so as long as *most* pages end up being used it's probably still a net win.

## Changes

As above, sets the MAP_POPULATE flag on mmap.

## Reason

Speeding up boot times is good.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- [N/A] If a specific issue led to this PR, this PR closes the issue.
- [N/A] Any required documentation changes (code and docs) are included in this PR.
- [N/A] All added/changed functionality is tested.
- [N/A] New `TODO`s link to an issue.

---

### DIFF
--- a/src/utils/src/vm_memory.rs
+++ b/src/utils/src/vm_memory.rs
@@ -119,7 +119,7 @@ pub fn create_guest_memory(
 
     for region in regions {
         let flags = match region.0 {
-            None => libc::MAP_NORESERVE | libc::MAP_PRIVATE | libc::MAP_ANONYMOUS,
+            None => libc::MAP_NORESERVE | libc::MAP_PRIVATE | libc::MAP_ANONYMOUS | libc::MAP_POPULATE,
             Some(_) => libc::MAP_NORESERVE | libc::MAP_PRIVATE,
         };
 


cperciva:
As with all performance-related changes, this will need to be extensively tested in the right environments (and I'm guessing you'll care more about Linux guests too).  But it's a trivial change with a huge impact on FreeBSD so I figured I'd toss it over for your consideration.

roypat:
Hi @cperciva, 
Thanks for this proposal! 

Our CI boottime test only uses one microvm config (and a very small one at that, 1vpu/128mb), which is probably why it passes. @pb8o has some ongoing work wrt improving the test to cover more scenarios. Judging by some of the other test failures, it seems that the performance impact for larger VMs is quite significant (the test_create_large_diff_snapshot test uses a 16GB VM, which now seems to take 5 seconds to start, but previously started in under 700ms (the api request timeout in our tests)). I suppose it makes sense that prefauling the entire guest memory only helps with boottime if the guest kernel takes up most of guest memory. Maybe there's some heuristics we could apply here regarding microvm size? Or only prefault the memory region into which we plan on loading the guest kernel?

Another small thing that needs to be taken into consideration here is UFFD. When restoring a snapshot from UFFD, guest memory is setup before the UFFD handler is registered, so with this change everything is pre-faulted to nothingness instead of the snapshotted memory. That's why the uffd related integration tests fail. We probably want to disable this altogether for the snapshot restore path.

cperciva:
Thanks for taking a look at it!  Yes, if snapshot restores are using the same anonymous mmap then this would have to be disabled in that case -- I'm not familiar with the Linux userland faulting mechanism so I didn't realize the initial mmap setup was done the same way.

Re large VMs -- yes, there's definitely room for some heuristics, and "only prefault the pages we expect to be used early in the boot process" could scale well... at least, assuming we can predict which pages those will be.  My guess is that the first pages used would be the ones immediately after the end of the kernel image, but it's possible that someone would decide to grow down from the top of address space instead.

Another option might be to be more aggressive in faulting in pages -- if we're using userland fault handing anyway, we could say "for the first 5 seconds of VM runtime, if a 4k page gets faulted then populate its entire 1M region" for example.  That would need attention from someone familiar with Linux VM bits though.

roypat:
Alright, I finally got around to running some tests to compare end-to-end boottimes with this change and without it for various common microvm sizes (40MB being the smallest that managed to still boot): 

image

image

I got the data by having out "test_boottime" test print out the startup times on a m6i.metal with a 4.14 guest. 

This seems to show that generally the overhead of pre-populating exceeds startup time gains :(

Since even for tiny VMs the performance of MAP_POPULATE is worse than lazily paging in, I feel that even with heuristics there's not much to gain here (there might be a very very narrow sweet spot of pre-populating "exactly right", but I doubt its feasible to hit that).

cperciva:
Thanks for testing!  Agreed that it doesn't make sense to merge right now so I'll close the PR.  I might come back to you with a revised version if I can figure out how to make the lazy page table initialization do larger chunks at a time -- I suspect that during kernel boot the pages being touched are all contiguous so initializing multiple pages on a page fault instead of just one should work.

Can you tell me how you ran the test to generate those graphs so I can try things out without bothering you?

roypat:
> Can you tell me how you ran the test to generate those graphs so I can try things out without bothering you?

Sure! There's a [`boot-data` branch](https://github.com/roypat/firecracker/tree/boot-data) branch over on my fork, with is Pablo's ci refractoring + https://github.com/firecracker-microvm/firecracker/commit/49745a953517fd6fb5160a714c079fca7f0eb6cb + your MAP_POPULATE commit. Running `tools/devtool -y test -- integration_tests/performance/test_boottime.py::test_boottime -s` on that branch will print 10 boot time samples per test case. I ran the test both with and without the MAP_POPULATE commit, and then averaged the samples for the graphs above. I ran everything on a m6i.metal with host linux kernel 5.10 (mainly because my laptop does not have enough memory to start the bigger VMs - if you want to run it locally, just remove the 10GB/12GB test cases)

PR #3945, opened 2023-07-07
Use dependabot groups

roypat:
## Changes

Use dependabot groups beta [[1]] to get it to give us a single PR with all dependency updates accumulated. Also turn all of our test utility crates into examples so that we only have a single Cargo.lock to deal with. 

## Reason

Lots of small PRs were painful to manage

[1]: https://github.com/dependabot/dependabot-core/issues/1190#issuecomment-1623832701

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/firecracker/Cargo.toml
```diff
@@ -30,3 +30,31 @@ vmm = { path = "../vmm" }
...
 cargo_toml = "0.15.3"
 regex = { version = "1.8.4", default-features = false, features = ["std"] }
+
+# Dev-Dependencies for uffd examples
+serde = { version = "1.0.165", features = ["derive"] }
+userfaultfd = "0.6.0"
+
+[[example]]
+name = "uffd_malicious_handler"
+path = "../../examples/uffd/malicious_handler.rs"
```
kalyazin, line 40: is it idiomatic to use relative paths here?

.github/dependabot.yml
```diff
@@ -8,19 +8,7 @@ updates:
...
-    schedule:
-      interval: "weekly"
-      day: "monday"
-    allow:
-      - dependency-type: "all"
-    open-pull-requests-limit: 100
+    groups:
+      firecracker:
+        patterns:
+          - "*"
```
kalyazin, line 14: nit: can add a new line at the end.

.github/dependabot.yml
```diff
@@ -8,19 +8,7 @@ updates:
...
-    open-pull-requests-limit: 100
-  - package-ecosystem: "cargo"
-    directory: "/tests/integration_tests/security/demo_seccomp/"
-    schedule:
-      interval: "weekly"
-      day: "monday"
-    allow:
-      - dependency-type: "all"
-    open-pull-requests-limit: 100
+    groups:
```
kalyazin, line 11: just for my education, do we have to explicitly say we're using a beta somewhere in the config?

roypat: Mh, I couldnt find anything on that in the docs, so I don't think so? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

src/firecracker/Cargo.toml
JonathanWoollett-Light: If the examples where under `src/firecracker/examples` they would be linked automatically. I think this is the simplest approach instead of using paths.

roypat: They still wouldn't get picked up automatically since they are in subdirectories, and they'd only get picked up automatically if there was one main function per subdir. Can move them to src/firecracker though if that's preferred, then the paths wouldn't have "../" components.

.github/dependabot.yml
roypat: just out of curiosity, because I see requests like this a lot all over open source projects: Why?

src/firecracker/Cargo.toml
roypat: done

.github/dependabot.yml
kalyazin: I think it is an unwritten law. Everyone believes that there are tools out there that don't handle the absences properly and it is always safer to add them.

src/firecracker/Cargo.toml
kalyazin: I don't have a preference here, no much difference between `../` and `../../`. @JonathanWoollett-Light what do you think?

kalyazin (approved):

src/firecracker/Cargo.toml
roypat: It'd be just `examples/bla/bla/bla` if its inside firecracker. I actually kinda prefer it to be not a top level directory the more I think about it - its not like its "production grade" code and shouldn't be this prominent

JonathanWoollett-Light: I think it is unfortunate we need to specify example paths, I think I would prefer if we could use a prefix like `uffd_` or `seccomp_` to avoid needing to specify paths, but this is a nit.

JonathanWoollett-Light (approved):

PR #3946, opened 2023-07-07
docs: Use consistent path for firecracker.socket in documentation

sudanl0:
Its easier for users to copy paste the command and get started without modifying any of the commands from the examples so, use same/consistent path for firecracker.socket in documentation.

## Changes

Use /tmp/firecracker.socket instead of ./firecracker.socket in documentation

## Reason

To make it easier to use example commands from our documentation.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


kalyazin (commented):
Overall, certainly a positive change.

I also noticed that the default value is `/run/firecracker.socket`, and it is mentioned in the doc a couple of times. Why do we advise users to use a custom path? Is `/tmp/` better in any way than `/run/`? If so, would it make sense to update the default and remove all the `--api-sock` clutter from our docs?

sudanl0:
> Overall, certainly a positive change.
> 
> I also noticed that the default value is `/run/firecracker.socket`, and it is mentioned in the doc a couple of times. Why do we advise users to use a custom path? Is `/tmp/` better in any way than `/run/`? If so, would it make sense to update the default and remove all the `--api-sock` clutter from our docs?

I think they both have their place.
The permission on /tmp (777) are more relaxed than on /run (755) making it easier to create/delete files without any privilege. /run is also intended to be used for persistent files while /tmp is for temporary tasks so I see /tmp as a better fit for examples on how to use Firecracker.
Also, I think the reference to /run/firecracker.socket are linked to the usage of jailer.
If we are inside a jailer we know we are definitely there to start firecracker and so firecracker.socket file is expected to be present and so /run makes sense in that case.
However, without the jailer, we don't know when we would want to start firecracker so using /tmp is better option than /run so that its easier for everyone to cleanup the socket files when not needed.

kalyazin:
> /run is also intended to be used for persistent files

Does it actually make any difference? In most cases Firecracker runs on a bare metal anyway, meaning any metal restart will cause Firecracker shutdown.

> Also, I think the reference to /run/firecracker.socket are linked to the usage of jailer.

Also in the [balloon](https://github.com/firecracker-microvm/firecracker/blob/main/docs/ballooning.md?plain=1#L135).

> However, without the jailer, we don't know when we would want to start firecracker so using /tmp is better option than /run so that its easier for everyone to cleanup the socket files when not needed.

I'm wondering if we can make this explicit in our docs, although I don't know where a good place for it could be.

pb8o (approved):

kalyazin (approved):

sudanl0:
> > /run is also intended to be used for persistent files
> 
> Does it actually make any difference? In most cases Firecracker runs on a bare metal anyway, meaning any metal restart will cause Firecracker shutdown.
> 
My comment was about the general guideline of /run by linux users not related to Firecracker itself. But bare metal restart, shutting down Firecracker, doesn't change the usecase that not all users are expecting to run Firecracker in a Jailer.

> > Also, I think the reference to /run/firecracker.socket are linked to the usage of jailer.
> 
> Also in the [balloon](https://github.com/firecracker-microvm/firecracker/blob/main/docs/ballooning.md?plain=1#L135).
> 
The intention of this PR was to make it useful for users to just use the command and run it and the "/run/firecracker.socket" references in balloon and other documents don't fall in this category so I preferred not to change them.

> > However, without the jailer, we don't know when we would want to start firecracker so using /tmp is better option than /run so that its easier for everyone to cleanup the socket files when not needed.
> 
> I'm wondering if we can make this explicit in our docs, although I don't know where a good place for it could be.
Maybe in document related to Jailer but I don't see this as a problem right now so I would defer it.

PR #3575, opened 2023-03-29
fix: Warn on `clippy::exit`

ls-1801:
Sets clippy to warn on the lint `clippy::exit`. Introduces a new Error Type `FcExitError`, which implements the `Terminator` trait. Replaces uses of `Proccess::exit` by returning an error.

## Changes

Replaces `std::process::exit` with `Result::Err` propagated to the main function.

This PR currently uses multiple ExitError types, which could be combined into a single error Type.

## Reason

Enables the use of `clippy::exit`. Fixes #3233 

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following
Developer Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

---

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #3994, opened 2023-07-28
Adding `clippy::exit` lint

### LINKED ISSUE #3233: Use `clippy::exit`

ShadowCurse:
## Changes
This PR is build on top of another PR: #3575
Adds new lint: `clippy::exit` and makes necessary changes for it.
Updates `Error` and `Result` types for several crates.

## Reason
Enables the use of `clippy::exit`. Fixes #3233

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/api_server/src/lib.rs
+++ b/src/api_server/src/lib.rs
@@ -11,7 +11,6 @@ mod parsed_request;
 mod request;
 
 use std::fmt::Debug;
-use std::path::PathBuf;
 use std::sync::mpsc;
 
 use logger::{
@@ -28,16 +27,6 @@ use vmm::rpc_interface::{ApiRequest, ApiResponse, VmmAction, VmmData};
 use vmm::vmm_config::snapshot::SnapshotType;
 
 use crate::parsed_request::{ParsedRequest, RequestAction};
-use crate::Error::ServerCreation;
-
-/// Errors thrown when binding the API server to the socket path.
-#[derive(Debug)]
-pub enum Error {
-    /// HTTP Server creation related error.
-    ServerCreation(ServerError),
-}
-
-type Result<T> = std::result::Result<T, Error>;
 
 /// Structure associated with the API server implementation.
 #[derive(Debug)]
@@ -70,8 +59,7 @@ impl ApiServer {
         }
     }
 
-    /// Starts the HTTP Server by binding to the socket path provided as
-    /// an argument.
+    /// Runs the Api Server.
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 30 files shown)


src/firecracker/src/api_server_adapter.rs
```diff
@@ -151,36 +165,25 @@ pub(crate) fn run_with_api(
...
                 &api_bind_path,
                 process_time_reporter,
                 &api_seccomp_filter,
                 api_payload_limit,
                 socket_ready_sender,
-            );
-
-            let Err(err) = res else {
-                return;
-            };
```
JonathanWoollett-Light, line 164: If we change this to 
```rust
            let Err(err) = res else {
                return Ok(());
            };
```
we can keep the existing pattern (which I implemented and personally prefer although this is opinion).

src/firecracker/src/main.rs
```diff
@@ -35,51 +39,58 @@ const DEFAULT_INSTANCE_ID: &str = "anonymous-instance";
...
-        );
-        if last_error == libc::EINVAL {
-            error!("The host does not support SSBD mitigation through prctl.");
-        }
+        ExitCode::from(exit_code as u8)
     }
 }
 
-fn main_exitable() -> FcExitCode {
+fn main() -> ExitCode {
```
JonathanWoollett-Light, line 78: Can we change this to returns a `Result<(), MainError>` like the other binaries?

src/cpu-template-helper/src/main.rs
```diff
@@ -182,12 +180,14 @@ fn run(cli: Cli) -> Result<()> {
...
+fn main() -> Result<()> {
     let cli = Cli::parse();
-
-    if let Err(e) = run(cli) {
-        eprintln!("Error: {}", e);
-        std::process::exit(EXIT_CODE_ERROR);
+    let result = run(cli);
+    if let Err(e) = result {
+        eprintln!("{}", e);
+        Err(e)
```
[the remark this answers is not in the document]
ShadowCurse: The reason I added `eprintln!` is that by default Err is formatted with `Debug` trait and I want to use `Display`.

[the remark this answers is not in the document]
JonathanWoollett-Light: That seems fair, I agree.

src/firecracker/src/main.rs
```diff
@@ -408,28 +392,44 @@ fn main_exitable() -> FcExitCode {
...
-    // https://doc.rust-lang.org/std/process/fn.exit.html
-    //
-    // See process_exitable() method of Subscriber trait for what triggers the exit_code.
-    //
-    let exit_code = main_exitable();
-    std::process::exit(exit_code as i32);
-}
+#[cfg(target_arch = "aarch64")]
+/// Enable SSBD mitigation through `prctl`.
+pub fn enable_ssbd_mitigation() {
```
JonathanWoollett-Light, line 401: Nit
```suggestion
/// Enable SSBD mitigation through `prctl`.
#[cfg(target_arch = "aarch64")]
pub fn enable_ssbd_mitigation() {
```

src/firecracker/src/api_server_adapter.rs
ShadowCurse: I wonder why you prefer this way over usual match. (except part that you implemented it)
But ultimately yeah, I can change it back.

src/firecracker/src/api_server_adapter.rs
```diff
@@ -161,29 +161,30 @@ pub(crate) fn run_with_api(
...
-                    Err(ApiServerError::FailedToBindSocket(sock_path))
-                }
-                Err(api_server::Error::ServerCreation(err)) => {
-                    Err(ApiServerError::FailedToBindAndRunHttpServer(err))
-                }
-            }
-        });
+            );
+        })
+        .expect("API thread spawn failed.");
```
JonathanWoollett-Light, line 187: Nit: It would be better to also have an error variant for this instead of a possible panic here.

src/firecracker/src/api_server_adapter.rs
JonathanWoollett-Light: It simplifies the match (I don't really like `Ok(_) => Ok(())`).

src/firecracker/src/main.rs
JonathanWoollett-Light: [`std::process::Termination`](https://doc.rust-lang.org/std/process/trait.Termination.html) determines the exit code, so it can return error codes other than 1 if implemented for an error.

ShadowCurse: But what if we did not get an error? If we have `fn main() -> Error` then we can not exit normally without `Error` having some `success` case. But this does not make sense for me.

ShadowCurse: And `-> Result<(), Error>` will return 1 in case of any error. https://doc.rust-lang.org/src/std/process.rs.html#2272

JonathanWoollett-Light: It could be implemented like `struct MainReturn(Result<(),MainError>);`

```rust
impl Termination for MainReturn {
    fn report(self) -> ExitCode {
        match self.0 {
            Ok(()) => ExitCode::SUCCESS,
            Err(err) => err.report()
        }
    }
}
```

JonathanWoollett-Light (approved):

src/firecracker/src/main.rs
ShadowCurse: Well.. now you just moved code from main into separate struct, but in nutshell it is same as what I have right now.

JonathanWoollett-Light: Yeah, its practically the same.

src/firecracker/src/api_server_adapter.rs
roypat, line 1: In this file, there is no need for the (socket_ready_sender, socket_ready_receiver) channels anymore. They were used to indicate from the API server thread _to_ the VMM thread that creating the HttpServer worked (e.g. that binding to the socket went OK). With this change, this communication is no longer necessary (as we do not even spawn up the api server thread if things go wrong, so they can be removed.

roypat (dismissed):

src/firecracker/src/api_server_adapter.rs
ShadowCurse: removed

src/api_server/src/lib.rs
```diff
@@ -70,87 +59,22 @@ impl ApiServer {
...
-    ///         InstanceInfo::default(),
-    ///     ))))
-    ///     .unwrap();
-    /// let mut sock = UnixStream::connect(PathBuf::from(path_to_socket)).unwrap();
-    /// // Send a GET instance-info request.
-    /// assert!(sock.write_all(b"GET / HTTP/1.1\r\n\r\n").is_ok());
-    /// let mut buf: [u8; 100] = [0; 100];
-    /// assert!(sock.read(&mut buf[..]).unwrap() > 0);
-    /// ```
-    pub fn bind_and_run(
```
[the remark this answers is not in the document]
ShadowCurse: I think tests give a good enough usage example in this case

src/api_server/src/lib.rs
```diff
@@ -189,17 +113,12 @@ impl ApiServer {
...
-                        server_request.process(|request| {
-                            self.handle_request(request, request_processing_start_us)
-                        }),
-                    )
-                    .or_else(|err| {
-                        error!("API Server encountered an error on response: {}", err);
-                        Ok(())
-                    })?;
+                // Use `self.handle_request()` as the processing callback.
+                let responce = server_request
```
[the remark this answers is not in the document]
ShadowCurse: fixed

src/firecracker/src/main.rs
ShadowCurse: fixed

JonathanWoollett-Light (approved):

PR #4003, opened 2023-08-03
fix: Update `api_event_fd` in `PrebootApiController`

JonathanWoollett-Light:
## Changes

Updates and explicitly documents purpose of the `api_event_fd` in `PrebootApiController::build_microvm_from_requests`. The functional change should minimize latency between pre-boot API request handling as it removes 1 syscall that was previously being done between every request.

## Reason

The handling of `api_event_fd` could have previously given the false impression it played a functional role in the pre-boot API.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/rpc_interface.rs
+++ b/src/vmm/src/rpc_interface.rs
@@ -384,17 +384,16 @@ impl<'a> PrebootApiController<'a> {
         // Configure and start microVM through successive API calls.
         // Iterate through API calls to configure microVm.
         // The loop breaks when a microVM is successfully started, and a running Vmm is built.
+        let mut i = 0;
         while preboot_controller.built_vmm.is_none() {
             // Get request
             let req = from_api
                 .recv()
                 .expect("The channel's sending half was disconnected. Cannot receive data.");
 
-            // Also consume the API event along with the message. It is safe to unwrap()
-            // because this event_fd is blocking.
-            api_event_fd
-                .read()
-                .expect("VMM: Failed to read the API event_fd");
+            // Count the number writes to the eventfd so we know how many reads we need to do to
+            // zero it.
+            i += 1;
 
             // Process the request.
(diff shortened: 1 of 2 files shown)


src/vmm/src/rpc_interface.rs
```diff
@@ -408,6 +407,16 @@ impl<'a> PrebootApiController<'a> {
             }
         }
 
+        // TODO: This eventfd no affect on the pre-boot API, we only read on this eventfd such that
+        // when it is passed to the post-boot API it is zero (the API thread always writes to the
+        // eventfd and does not differentiate between pre-boot and post-boot). Here we are making
+        // unnecessary syscalls that slow down the API. This should be removed.
```
acatangiu, line 413: `write()` syscall on eventfd is really cheap and there are only a few calls during VM setup - the optimization you speak of here has zero practical impact on VM setup/boot latency - I will be very surprised if you could actually measure any difference that is not simple noise.

On the other hand, if I understand your comment, the cost of making this optimization is making `api_server` aware of which commands are before boot and which are after which is outside its logical scope and would brake the components' current decoupled design.

I agree this is all more complex than it could be, but the reason for it was to have a VMM that can be built without api server (see https://github.com/firecracker-microvm/firecracker/pull/1489 and https://github.com/firecracker-microvm/firecracker/pull/1490).

Now, if you want to tie back the components together and remove the generic rpc layer, I suggest you do just that, and completely remove this `rpc_interface` layer and just move the logic to api_server.

acatangiu: Maybe the diagrams in https://github.com/firecracker-microvm/firecracker/pull/1490 description would have also been useful to be included as docs in the codebase.

bchalios:
The performance improvement in this PR should be negligible (maybe not even noticeable). Moreover, it is not along a hot-path. Closing this.

PR #4017, opened 2023-08-09
feat: Simplify error logging

JonathanWoollett-Light:
## Changes

Simplifies errors to always log their `Display` implementation and
always write their `Debug` implementation to `stderr`, this also
simplifies general logging procedure when a log target has not been
initialized, now all logs are written to `stdout`. Previously `error`
and `warn` level logs where written to `stderr` while `info`, `debug`
and `trace` level logs where written to `stdout`.

## Reason

Simpler is better (it also helps with https://github.com/firecracker-microvm/firecracker/pull/3971).

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/firecracker/tests/test_api_server.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+const BINARY: &str = env!("CARGO_BIN_EXE_firecracker");
```
pb8o, line 4: Please don't make integration tests in Rust. We already have the Python integration tests for that. 

Furthermore, this test seems to duplicate https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/functional/test_api_server.py#L10

pb8o (changes requested):

src/firecracker/tests/test_api_server.rs
JonathanWoollett-Light: I don't think the existing tests look at stdout and stderr. They only check for the presence of the error in the log target (and check the exit-code is non-zero but not a specific value).

pb8o: > We have existing integration tests in https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/tests/integration_tests.rs. I found this the simplest test to specifically cover the change. I don't see a reason for not having this additional test?

To me that's still a unit test, as it does not use the firecracker binary. That's where I make the distinction.

src/firecracker/src/main.rs
```diff
@@ -75,16 +75,25 @@ impl From<MainError> for ExitCode {
...
+struct MainExit(Result<(), MainError>);
+impl std::process::Termination for MainExit {
+    fn report(self) -> ExitCode {
+        if let Err(err) = self.0 {
+            error!("{err}");
+            eprintln!("Error: {err:?}");
+            ExitCode::from(err)
+        } else {
+            ExitCode::SUCCESS
+        }
```
ShadowCurse, line 89: I don't really like adding an additional struct just to print errors. I think you can just modify the main itself:
```rust
fn main() -> ExitCode {
    if let Err(err) = main_exec() {
        error!("{err}");
        eprintln!("Error: {err:?}");
        ExitCode::from(err)
    } else {
        ExitCode::SUCCESS
    }
}

```

JonathanWoollett-Light: Fixed.

ShadowCurse (approved):

pb8o (approved):

PR #4018, opened 2023-08-09
test: add a test to measure Firecracker memory overhead

pb8o:
Our current binary size test measures space in disk, as a proxy for used memory. This is not accurate, as parts of the binary are not loaded into memory.

This commits adds a new test that measures memory overhead directly and published metrics for different guest configurations.

Also remove thresholds from test_binary_size.py test, as we measure memory overhead directly now.

## Changes

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/performance/test_memory_overhead.py
+++ b/tests/integration_tests/performance/test_memory_overhead.py
@@ -1,0 +1,92 @@
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Track Firecracker memory overhead
+
+Because Firecracker is a static binary, and is copied before execution, no
+memory is shared across many different processes. It is thus important to track
+how much memory overhead Firecracker adds.
+
+These tests output metrics to capture the memory overhead that Firecracker adds,
+both from the binary file (file-backed) and what Firecracker allocates during
+the process lifetime.
+
+The memory overhead of the jailer is not important as it is short-lived.
+"""
+
+from collections import defaultdict
+from pathlib import Path
+
+import psutil
+import pytest
+
+from framework.properties import global_props
+
+# If guest memory is >3328MB, it is split in a 2nd region
+X86_MEMORY_GAP_START = 3328 * 2**20
+
+
+@pytest.mark.parametrize(
+    "vcpu_count,mem_size_mib",
+    [(1, 128), (1, 1024), (2, 2024), (4, 4096)],
+)
+def test_memory_overhead(
+    microvm_factory, guest_kernel, rootfs, vcpu_count, mem_size_mib, metrics
+):
+    """Track Firecracker memory overhead.
+
+    We take a single measurement as it only varies by a few KiB each run.
+    """
+
+    microvm = microvm_factory.build(guest_kernel, rootfs)
+    microvm.spawn()
+    microvm.basic_config(vcpu_count=vcpu_count, mem_size_mib=mem_size_mib)
+    microvm.add_net_iface()
+    microvm.start()
+
+    # check that the vm is running
+    microvm.ssh.run("true")
+
+    guest_mem_bytes = mem_size_mib * 2**20
+    guest_mem_splits = {
+        guest_mem_bytes,
+        X86_MEMORY_GAP_START,
+    }
+    if guest_mem_bytes > X86_MEMORY_GAP_START:
+        guest_mem_splits.add(guest_mem_bytes - X86_MEMORY_GAP_START)
+
+    mem_stats = defaultdict(int)
+    mem_stats["guest_memory"] = guest_mem_bytes
+    ps = psutil.Process(microvm.jailer_clone_pid)
+
+    for pmmap in ps.memory_maps(grouped=False):
+        # We publish 'size' and 'rss' (resident). size would be the worst case,
+        # whereas rss is the current paged-in memory.
+
+        mem_stats["total_size"] += pmmap.size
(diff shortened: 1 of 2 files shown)


tests/integration_tests/performance/test_memory_overhead.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # check that the vm is running
+    microvm.ssh.run("true")
+
+    guest_mem_bytes = mem_size_mib * 2**20
+    guest_mem_splits = {
+        guest_mem_bytes,
+        X86_MEMORY_GAP_START,
+        guest_mem_bytes - X86_MEMORY_GAP_START,
+    }
```
bchalios, line 55: shoudn't this be conditioned to whether `guest_mem_bytes >= X86_MEMORY_GAP_START` or not?

tests/integration_tests/performance/test_memory_overhead.py
```diff
@@ -0,0 +1,90 @@
...
+    mem_stats = defaultdict(int)
+    mem_stats["guest_memory"] = guest_mem_bytes
+    ps = psutil.Process(microvm.jailer_clone_pid)
```
bchalios, line 58: So, this gathers statics for microVMs, that run inside the jailer. Maybe we should put that in the comment on the top of this file.

tests/integration_tests/performance/test_memory_overhead.py
```diff
@@ -0,0 +1,90 @@
...
+
+    for pmmap in ps.memory_maps(grouped=False):
+        # We publish 'size' and 'rss' (resident). size would be the worst case,
+        # whereas rss is the current paged-in memory.
+
+        mem_stats["total_size"] += pmmap.size
+        mem_stats["total_rss"] += pmmap.rss
```
bchalios, line 65: what is paged in or not depends on the guest's workload right? Should we publish that metric? it looks like it doesn't carry much information.

tests/integration_tests/performance/test_memory_overhead.py
```diff
@@ -0,0 +1,90 @@
...
+        "rootfs": rootfs.name,
+    }
+    metrics.set_dimensions(dimensions)
+    for key, value in mem_stats.items():
+        metrics.put_metric(key, value, unit="Bytes")
+
+    mem_info = ps.memory_full_info()
+    for metric in ["rss", "uss", "text"]:
+        val = getattr(mem_info, metric)
+        metrics.put_metric(metric, val, unit="Bytes")
```
bchalios, line 90: What's that publishing, exactly?

bchalios (commented):
LGTM. Just some small nits. Feel free to ignore.

tests/integration_tests/performance/test_memory_overhead.py
pb8o: `jailer_clone_pid` is the Firecracker PID. Since the jailer exec()s into Firecracker, the jailer doesn't play a role, I think?

tests/integration_tests/performance/test_memory_overhead.py
pb8o: hmm true... if it's not, it will end up with a negative number so it works almost by accident

tests/integration_tests/performance/test_memory_overhead.py
bchalios: oh, these are total values, isn't the RSS from here then the same as `mem_stats["total_rss"]`

pb8o: Yes, they should be equivalent. There's a small diff that I wouldn't know how to account for, but they are in line.

bchalios: I'm not sure. But I guess it would make sense to remove the one in line 88 (?) I guess the difference might be due to the fact that you poll the psutil info at different points in time(?)

tests/integration_tests/performance/test_memory_overhead.py
JonathanWoollett-Light: Would it make sense to have 2 tests?

1. Firecracker
2. Firecracker + Jailer

I think both metrics may be useful.

pb8o: I don't see how the jailer would play a role?

JonathanWoollett-Light: > I don't see how the jailer would play a role?

Could we then run the test without the jailer?

I think if we aren't concerned with testing it here maybe we can remove this complexity.

pb8o: Well, two reasons:
1. it is the default
2. the jailer it's the bit that makes sure different Firecracker processes don't share pages, buy copying the binary before executing it

JonathanWoollett-Light: I think in the case where the jailer does affect the memory size of Firecracker even as a result of side affects it makes sense to measure this.

Although I am okay to merge this without this.

JonathanWoollett-Light (dismissed):

JonathanWoollett-Light (approved):

tests/integration_tests/performance/test_memory_overhead.py
bchalios: Isn't that a concern here then?

pb8o: It's not a concern, but I will add the change to make it explicit.

JonathanWoollett-Light (approved):

PR #4037, opened 2023-08-15
fix: Only setup serial device if input is a terminal

roypat:
Currently, whenever firecracker starts with a daemonized jailer, it prints the following warning: "Failed to register serial input fd: event_manager: failed to manage epoll file descriptor: Operation not permitted (os error 1)". This is because if the jailer daemonizes the firecracker process, it will redirect stdin to /dev/null. However, when trying to register FILENO_STDIN to epoll via epoll_ctl, it will return EPERM if such a redirection has happened. However, if the jailer is daemonized, then this will result in the correct behavior, because we do not _want_ a serial device here.

The changes in this commit make firecracker only try to initialize the serial device if FILENO_STDIN is a terminal (as reported by isatty(3)). This fixes the spurious warnings.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


tests/integration_tests/functional/test_serial_io.py
```diff
@@ -219,3 +219,26 @@ def test_serial_block(test_microvm_with_api):
...
+    """
+
+    test_microvm = uvm_plain
+    test_microvm.spawn()
+    test_microvm.basic_config(
+        vcpu_count=1,
+        mem_size_mib=512,
+    )
+    test_microvm.start()
+
```
pb8o, line 243: I would add a `test_microvm.ssh.run("true")` here to make sure the guest has finished booting.

pb8o (dismissed):

tests/integration_tests/functional/test_serial_io.py
roypat: Ahh, I thought the .start() method will wait for it to finish booting. I'll add that, thanks! :)

pb8o (approved):

pb8o (approved):

zulinx86 (approved):

src/vmm/src/devices/legacy/serial.rs
```diff
@@ -277,6 +284,24 @@ impl<I: Read + AsRawFd + Send + Debug> MutEventSubscriber
...
+    if unsafe { libc::fstat(fd, stat.as_mut_ptr()) } < 0 {
+        return false;
+    }
+
+    // SAFETY: We can safely assume the libc::stat structure to be initialized, as libc::fstat
+    // returning 0 guarantees that the memory is now initialized with the requested file metadata.
+    let stat = unsafe { stat.assume_init() };
+
+    (stat.st_mode & libc::S_IFIFO) != 0
+}
```
JonathanWoollett-Light, line 303: Nit: I like to wrap `std::mem::MaybeUninit::uninit`  in the `unsafe` block so outside `unsafe` we don't need to deal wiht it.

This is a stylistic choice so feel free to ignore.

```rust
/// Checks whether the given file descriptor is a FIFO pipe.
fn is_fifo(fd: RawFd) -> bool {
    // SAFETY: We can safely assume the libc::stat structure to be initialized, as libc::fstat
    // returning 0 guarantees that the memory is now initialized with the requested file metadata.
    let stat = unsafe {
        let mut stat = std::mem::MaybeUninit::<libc::stat>::uninit();
        // No unsafety can be introduced by passing in an invalid file descriptor to fstat,
        // it will return -1 and set errno to EBADF. The pointer passed to fstat is valid for writing
        // a libc::stat structure.
        if libc::fstat(fd, stat.as_mut_ptr()) < 0 {
            return false;
        }
        stat.assume_init()
    };
    (stat.st_mode & libc::S_IFIFO) != 0
}
```

roypat: Mh, I like having each "unsafe-causing" call be its own block, although using scoping to make sure that the MaybeUninit does not escape is indeed nice! I'll keep it like this to not need to regather approvals, but thanks for the idea!

src/vmm/src/devices/legacy/serial.rs
```diff
@@ -340,4 +367,23 @@ mod tests {
...
+
+    #[test]
+    fn test_is_fifo() {
+        // invalid file descriptors arent fifos
+        let invalid = -1;
+        assert!(!is_fifo(invalid));
+
+        // Fifos are fifos
+        let mut fds: [libc::c_int; 2] = [0; 2];
+        let rc = unsafe { libc::pipe(fds.as_mut_ptr()) };
```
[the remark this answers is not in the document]
roypat: Mh, we dont have it as a direct dependency of firecracker yet, but it would be nice to have at slightly higher abstraction than just libc at hand. I guess the question is how much do we trust an external dependency to get the C-interop right (but since we already use uffd-rs, I guess maybe its not a question after all)

PR #4045, opened 2023-08-17
Snapshot editor

ShadowCurse:
## Changes
Added initial version of `snapshot-editor` tool.
Initial functionality is:
- (aarch64 only) remove registers from vcpu states
- (aarch64 only)  print all vcpu states from vmstate file

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Added
 
+- [#4045](https://github.com/firecracker-microvm/firecracker/pull/4045)
+  Added `snapshot-editor` tool for modifications of snapshot files.
 - [#3967](https://github.com/firecracker-microvm/firecracker/pull/3967/):
   Added new fields to the custom CPU templates. (aarch64 only) `vcpu_features`
   field allows modifications of vCPU features enabled during vCPU
--- a/docs/snapshotting/snapshot-editor.md
+++ b/docs/snapshotting/snapshot-editor.md
@@ -1,0 +1,91 @@
+# Snapshot editor
+
+The `snapshot-editor` is a program for modification of Firecracker snapshots.
+
+## Prior knowledge
+
+Firecracker snapshot consists of 2 files:
+
+- `vmstate` file: file with Firecracker internal data such as vcpu states,
+  devices states etc.
+- `memory` file: file with guest memory.
+
+`snapshot-editor` (currently) allows to modify only `vmstate` files only
+on aarch64 platform.
+
+## Usage
+
+### `edit-vmstate` command
+
+#### `remove-regs` subcommand (aarch64 only)
+
+This command is used to remove specified registers from vcpu states inside
+vmstate snapshot file.
+
+Arguments:
+
+- `VMSTATE_PATH` - path to the `vmstate` file
+- `OUTPUT_PATH` - path to the file where the output will be placed
+- `[REGS]` - set of u32 values representing registers ids as they are defined
+  in KVM. Can be both in decimal and in hex formats.
+
+Usage:
+
+```bash
+snapshot-editor edit-vmstate remove-regs \
+    --vmstate-path <VMSTATE_PATH> \
+    --output-path <OUTPUT_PATH> \
+    [REGS]...
+```
+
+Example:
+
+```bash
+./snapshot-editor edit-vmstate remove-regs \
+    --vmstate-path ./vmstate_file \
+    --output-path ./new_vmstate_file \
+    0x1 0x2
+```
+
+### `info-vmstate` command
+
+#### `version` subcommand
+
+This command is used to print version of the provided
+vmstate file.
+
+Arguments:
+
+- `VMSTATE_PATH` - path to the `vmstate` file
+
+Usage:
+
+```bash
+snapshot-editor info-vmstate version --vmstate-path <VMSTATE_PATH>
+```
+
+Example:
+
+```bash
+./snapshot-editor info-vmstate version --vmstate-path ./vmstate_file
+```
+
+#### `vcpu-states` subcommand (aarch64 only)
+
+This command is used to print the vCPU states inside vmstate snapshot file.
+
+Arguments:
+
+- `VMSTATE_PATH` - path to the `vmstate` file
+
+Usage:
+
+```bash
+snapshot-editor info-vmstate vcpu-states --vmstate-path <VMSTATE_PATH>
+```
+
+Example:
+
+```bash
+./snapshot-editor info-vmstate vcpu-states --vmstate-path ./vmstate_file
+```
--- a/src/snapshot-editor/src/edit_vmstate.rs
+++ b/src/snapshot-editor/src/edit_vmstate.rs
@@ -1,0 +1,195 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::path::PathBuf;
+
+use clap::Subcommand;
+use clap_num::maybe_hex;
+use vmm::arch::aarch64::regs::Aarch64RegisterVec;
+use vmm::persist::MicrovmState;
+
+use crate::utils::{open_vmstate, save_vmstate, UtilsError};
+
+#[derive(Debug, thiserror::Error)]
+pub enum EditVmStateError {
+    #[error("{0}")]
+    Utils(#[from] UtilsError),
+}
+
+#[derive(Debug, Subcommand)]
+pub enum EditVmStateSubCommand {
+    /// Remove registers from vcpu states.
+    RemoveRegs {
+        /// Set of registers to remove.
+        /// Values should be registers ids as the are defined in KVM.
+        #[arg(value_parser=maybe_hex::<u64>, num_args = 1.., value_delimiter = ' ')]
+        regs: Vec<u64>,
+        /// Path to the vmstate file.
+        #[arg(short, long)]
+        vmstate_path: PathBuf,
+        /// Path of output file.
+        #[arg(short, long)]
+        output_path: PathBuf,
+    },
+}
+
+pub fn edit_vmstate_command(command: EditVmStateSubCommand) -> Result<(), EditVmStateError> {
+    match command {
+        EditVmStateSubCommand::RemoveRegs {
+            regs,
+            vmstate_path,
+            output_path,
+        } => edit(&vmstate_path, &output_path, |state| {
+            remove_regs(state, &regs)
+        })?,
+    }
+    Ok(())
+}
+
+fn edit(
+    vmstate_path: &PathBuf,
+    output_path: &PathBuf,
+    f: impl Fn(MicrovmState) -> Result<MicrovmState, EditVmStateError>,
+) -> Result<(), EditVmStateError> {
+    let (microvm_state, version) = open_vmstate(vmstate_path)?;
+    let microvm_state = f(microvm_state)?;
+    save_vmstate(microvm_state, output_path, version)?;
+    Ok(())
+}
+
+fn remove_regs(
+    mut state: MicrovmState,
+    remove_regs: &[u64],
+) -> Result<MicrovmState, EditVmStateError> {
+    for (i, vcpu_state) in state.vcpu_states.iter_mut().enumerate() {
+        println!("Modifying state for vCPU {i}");
+
+        let mut removed = vec![false; remove_regs.len()];
+        let mut new_regs = Aarch64RegisterVec::default();
+        for reg in vcpu_state.regs.iter().filter(|reg| {
+            if let Some(pos) = remove_regs.iter().position(|r| r == &reg.id) {
+                removed[pos] = true;
+                false
+            } else {
+                true
+            }
+        }) {
+            new_regs.push(reg);
+        }
+        vcpu_state.regs = new_regs;
+        for (reg, removed) in remove_regs.iter().zip(removed.iter()) {
+            print!("Regsiter {reg:#x}: ");
+            match removed {
+                true => println!("removed"),
+                false => println!("not present"),
+            }
+        }
+    }
+    Ok(state)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_remove_regs() {
+        const KVM_REG_SIZE_U8: u64 = 0;
+        const KVM_REG_SIZE_U16: u64 = 0x10000000000000;
+        const KVM_REG_SIZE_U32: u64 = 0x20000000000000;
+
+        use vmm::arch::aarch64::regs::Aarch64RegisterRef;
+        use vmm::vstate::vcpu::aarch64::VcpuState;
+
+        let vcpu_state = VcpuState {
+            regs: {
+                let mut regs = Aarch64RegisterVec::default();
+                let reg_data: u8 = 69;
+                regs.push(Aarch64RegisterRef::new(
+                    KVM_REG_SIZE_U8,
+                    &reg_data.to_le_bytes(),
+                ));
+                let reg_data: u16 = 69;
+                regs.push(Aarch64RegisterRef::new(
+                    KVM_REG_SIZE_U16,
+                    &reg_data.to_le_bytes(),
+                ));
+                let reg_data: u32 = 69;
+                regs.push(Aarch64RegisterRef::new(
+                    KVM_REG_SIZE_U32,
+                    &reg_data.to_le_bytes(),
+                ));
+                regs
+            },
+            ..Default::default()
+        };
+        let state = MicrovmState {
+            vcpu_states: vec![vcpu_state],
+            ..Default::default()
+        };
+
+        let new_state = remove_regs(state, &[KVM_REG_SIZE_U32]).unwrap();
+
+        let expected_vcpu_state = VcpuState {
+            regs: {
+                let mut regs = Aarch64RegisterVec::default();
+                let reg_data: u8 = 69;
+                regs.push(Aarch64RegisterRef::new(
+                    KVM_REG_SIZE_U8,
+                    &reg_data.to_le_bytes(),
+                ));
+                let reg_data: u16 = 69;
+                regs.push(Aarch64RegisterRef::new(
+                    KVM_REG_SIZE_U16,
+                    &reg_data.to_le_bytes(),
+                ));
+                regs
+            },
+            ..Default::default()
+        };
+
+        assert_eq!(new_state.vcpu_states[0].regs, expected_vcpu_state.regs);
+    }
+
+    #[test]
+    fn test_remove_non_existed_regs() {
+        const KVM_REG_SIZE_U8: u64 = 0;
+        const KVM_REG_SIZE_U16: u64 = 0x10000000000000;
+        const KVM_REG_SIZE_U32: u64 = 0x20000000000000;
+
+        use vmm::arch::aarch64::regs::Aarch64RegisterRef;
+        use vmm::vstate::vcpu::aarch64::VcpuState;
+
+        let vcpu_state = VcpuState {
+            regs: {
+                let mut regs = Aarch64RegisterVec::default();
+                let reg_data: u8 = 69;
+                regs.push(Aarch64RegisterRef::new(
+                    KVM_REG_SIZE_U8,
+                    &reg_data.to_le_bytes(),
+                ));
+                let reg_data: u16 = 69;
+                regs.push(Aarch64RegisterRef::new(
+                    KVM_REG_SIZE_U16,
+                    &reg_data.to_le_bytes(),
+                ));
+                regs
+            },
+            ..Default::default()
+        };
+
+        let state_clone = MicrovmState {
+            vcpu_states: vec![vcpu_state.clone()],
+            ..Default::default()
+        };
+
+        let state = MicrovmState {
+            vcpu_states: vec![vcpu_state],
+            ..Default::default()
+        };
+
+        let new_state = remove_regs(state_clone, &[KVM_REG_SIZE_U32]).unwrap();
+
+        assert_eq!(new_state.vcpu_states[0].regs, state.vcpu_states[0].regs);
+    }
+}
--- a/src/snapshot-editor/src/info.rs
+++ b/src/snapshot-editor/src/info.rs
@@ -1,0 +1,89 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::path::PathBuf;
+
+use clap::Subcommand;
+use vmm::persist::MicrovmState;
+use vmm::version_map::FC_VERSION_TO_SNAP_VERSION;
+
+use crate::utils::*;
+
+#[derive(Debug, thiserror::Error)]
+pub enum InfoVmStateError {
+    #[error("Cannot translate snapshot data version {0} to Firecracker microVM version")]
+    InvalidVersion(u16),
+    #[error("{0}")]
+    Utils(#[from] UtilsError),
+}
+
+#[derive(Debug, Subcommand)]
+pub enum InfoVmStateSubCommand {
+    /// Print snapshot version.
+    Version {
+        /// Path to the vmstate file.
+        #[arg(short, long)]
+        vmstate_path: PathBuf,
+    },
+    /// Print info about vcpu states.
+    #[cfg(target_arch = "aarch64")]
+    VcpuStates {
+        /// Path to the vmstate file.
+        #[arg(short, long)]
+        vmstate_path: PathBuf,
+    },
+}
+
+pub fn info_vmstate_command(command: InfoVmStateSubCommand) -> Result<(), InfoVmStateError> {
+    match command {
+        InfoVmStateSubCommand::Version { vmstate_path } => info(&vmstate_path, info_version)?,
+        #[cfg(target_arch = "aarch64")]
+        InfoVmStateSubCommand::VcpuStates { vmstate_path } => {
+            info(&vmstate_path, info_vcpu_states)?
+        }
+    }
+    Ok(())
+}
+
+fn info(
+    vmstate_path: &PathBuf,
+    f: impl Fn(&MicrovmState, u16) -> Result<(), InfoVmStateError>,
+) -> Result<(), InfoVmStateError> {
+    let (vmstate, version) = open_vmstate(vmstate_path)?;
+    f(&vmstate, version)?;
+    Ok(())
+}
+
+fn info_version(_: &MicrovmState, version: u16) -> Result<(), InfoVmStateError> {
+    match FC_VERSION_TO_SNAP_VERSION
+        .iter()
+        .find(|(_, &v)| v == version)
+    {
+        Some((key, _)) => {
+            println!("v{key}");
+            Ok(())
+        }
+        None => Err(InfoVmStateError::InvalidVersion(version)),
+    }
+}
+
+#[cfg(target_arch = "aarch64")]
+fn info_vcpu_states(state: &MicrovmState, _: u16) -> Result<(), InfoVmStateError> {
+    for (i, state) in state.vcpu_states.iter().enumerate() {
+        println!("vcpu {i}:");
+        println!("kvm_mp_state: {:#x}", state.mp_state.mp_state);
+        println!("mpidr: {:#x}", state.mpidr);
+        for reg in state.regs.iter() {
+            println!(
+                "{:#x} 0x{}",
+                reg.id,
+                reg.as_slice()
+                    .iter()
+                    .rev()
+                    .map(|b| format!("{b:x}"))
+                    .collect::<String>()
+            );
+        }
+    }
+    Ok(())
+}
--- a/src/snapshot-editor/src/main.rs
+++ b/src/snapshot-editor/src/main.rs
@@ -1,0 +1,60 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use clap::{Parser, Subcommand};
+
+#[cfg(target_arch = "aarch64")]
+mod edit_vmstate;
+mod info;
+mod utils;
+
+#[cfg(target_arch = "aarch64")]
+use edit_vmstate::{edit_vmstate_command, EditVmStateError, EditVmStateSubCommand};
+use info::{info_vmstate_command, InfoVmStateError, InfoVmStateSubCommand};
+
+#[derive(Debug, thiserror::Error)]
+enum SnapEditorError {
+    #[cfg(target_arch = "aarch64")]
+    #[error("Error during editing vmstate file: {0}")]
+    EditVmState(#[from] EditVmStateError),
+    #[error("Error during getting info from a vmstate file: {0}")]
+    InfoVmState(#[from] InfoVmStateError),
+}
+
+#[derive(Debug, Parser)]
+#[command(version = format!("v{}", env!("FIRECRACKER_VERSION")))]
+struct Cli {
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(Debug, Subcommand)]
+enum Command {
+    #[cfg(target_arch = "aarch64")]
+    #[command(subcommand)]
+    EditVmstate(EditVmStateSubCommand),
+    #[command(subcommand)]
+    InfoVmstate(InfoVmStateSubCommand),
+}
+
+fn main_exec() -> Result<(), SnapEditorError> {
+    let cli = Cli::parse();
+
+    match cli.command {
+        #[cfg(target_arch = "aarch64")]
+        Command::EditVmstate(command) => edit_vmstate_command(command)?,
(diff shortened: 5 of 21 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 ### Added
 
+- [#4045](https://github.com/firecracker-microvm/firecracker/pull/4045)
+  Added `snapshot-editor` tool for modifications of snapshot files.
```
roypat, line 8: I think we can go a little bit more into details here. What exactly can it modify? Also, "snapshot files" is a bit ambiguous here, since it could refer to both the memory and the vmstate file. We should clarify that is only allows modification of the vmstate file.

roypat: Maybe it can also link to the documentation you add in this PR?

src/snapshot-editor/src/info.rs
```diff
@@ -0,0 +1,66 @@
...
+        InfoVmStateSubCommand::VcpuStates { vmstate_path } => {
+            info(&vmstate_path, info_vcpu_states)?
+        }
+    }
+    Ok(())
+}
+
+fn info(
+    vmstate_path: &PathBuf,
+    f: impl Fn(&MicrovmState, u16) -> Result<(), InfoVmStateError>,
```
roypat, line 40: Mh, why does this closure take a `u16` for the version, while the one of the `edit` function in `edit_vmstate.rs` does not? Especially since the field is just ignored below

tests/integration_tests/functional/test_snapshot_editor.py
```diff
@@ -0,0 +1,72 @@
...
+        "--vmstate-path",
+        str(snapshot.vmstate),
+    ]
+    _, stdout, _ = utils.run_cmd(cmd)
+    assert stdout.find(str(MIDR_EL1)) == -1
+
+    # test that we can restore from a snapshot
+    new_vm = microvm_factory.build()
+    new_vm.spawn()
+    new_vm.restore_from_snapshot(snapshot, resume=True)
```
roypat, line 72: Maybe here we'll want to execute a command inside the VM to check everything works? e.g. a `new_vm.ssh.execute("true")` (e.g. whatever the other snapshot related tests do)

roypat (commented):
Looks good to me, just the changelog could maybe explain a new tool a bit better

docs/snapshotting/snapshot-editor.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Prior knowledge
+
+Firecracker snapshot consists of 2 files:
+
+- vmstate file: file with Firecracker internal data such as vcpu states,
+  devices states etc.
+- memory file: file with guest memory.
+
+`snapshot-editor` allows to modify each of these files separately.
```
roypat, line 13: Right now it doesnt - we should write the documentation based on what currently is implemented (e.g. this tool deals with vmstate, and rebase-snap deals with memory file)

docs/snapshotting/snapshot-editor.md
```diff
@@ -0,0 +1,61 @@
...
+
+```bash
+snapshot-editor info-vmstate vcpu-states --vmstate-path <VMSTATE_PATH>
+```
+
+Example:
+
+```bash
+./snapshot-editor edit info-vmstate vcpu-states --vmstate-path ./vmstate_file
+```
```
roypat, line 61: Maybe we can add some example output here?

src/snapshot-editor/src/edit_vmstate.rs
```diff
@@ -0,0 +1,143 @@
...
+    for vcpu_state in state.vcpu_states.iter_mut() {
+        let mut new_regs = Aarch64RegisterVec::default();
+        for reg in vcpu_state
+            .regs
+            .iter()
+            .filter(|reg| !remove_regs.contains(&reg.id))
+        {
+            new_regs.push(reg);
+        }
+        vcpu_state.regs = new_regs;
```
roypat, line 75: Maybe we can print a warning if some register's removal was requested, but the register isnt present?

docs/snapshotting/snapshot-editor.md
```diff
@@ -0,0 +1,61 @@
...
+
+Example:
+
+```bash
+./snapshot-editor edit-vmstate remove-regs \
+    --vmstate-path ./vmstate_file \
+    --output-path ./new_vmstate_file \
+    0x1 0x2
+```
```
roypat, line 43: Maybe this can briefly explain what the tool does? e.g. reads `vmstate_file`, removes the registers 0x1 and 0x2 and then writes the modified file to `new_vmstate_file`

roypat (approved):

src/snapshot-editor/src/info.rs
ShadowCurse: initially the tool could also tell the snapshot/firecracker version, but I removed it to make PR smaller.
Removed the `u16` parameter.

tests/integration_tests/functional/test_snapshot_editor.py
ShadowCurse: Added a call to `ls` after restoration.

src/snapshot-editor/src/edit_vmstate.rs
ShadowCurse: Added printing of which registers were removed and which regs were missing/not present.

docs/snapshotting/snapshot-editor.md
ShadowCurse: There is a small description of the command. I think it is sufficient.

docs/snapshotting/snapshot-editor.md
roypat: I still we should just say "`snapshot-editor` (currently) allows to modify the `vmstate` file of aarch64 snapshots" instead of having multiple notes. Because this sentence here might be true in the future, but currently it is still false, so we should just merge it with the notes.

ShadowCurse: Merged

roypat (approved):

docs/snapshotting/snapshot-editor.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Prior knowledge
+
+Firecracker snapshot consists of 2 files:
+
+- `vmstate` file: file with Firecracker internal data such as vcpu states,
+  devices states etc.
+- `memory` file: file with guest memory.
+
+`snapshot-editor` (currently) allows to modify only `vmstate`files only
```
roypat, line 13: nit: 
```suggestion
`snapshot-editor` (currently) allows to modify only `vmstate` files only
```

src/snapshot-editor/src/main.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use clap::{Parser, Subcommand};
+
+#[cfg(target_arch = "aarch64")]
+mod edit_vmstate;
+#[cfg(target_arch = "aarch64")]
+mod info;
```
bchalios, line 9: You can use [std::compile_error](https://doc.rust-lang.org/std/macro.compile_error.html) to avoid doing the `#[cfg(target_arch = "aarch64)]` thing multiple times. It is quite awkward otherwise that we produce a binary that does nothing on x86.

src/snapshot-editor/src/info.rs
```diff
@@ -0,0 +1,66 @@
...
+        InfoVmStateSubCommand::VcpuStates { vmstate_path } => {
+            info(&vmstate_path, info_vcpu_states)?
+        }
+    }
+    Ok(())
+}
+
+fn info(
+    vmstate_path: &PathBuf,
+    f: impl Fn(&MicrovmState) -> Result<(), InfoVmStateError>,
```
bchalios, line 40: Same as before, I think it's much simpler to just have a single function here that opens the file and prints the related info.

src/snapshot-editor/src/edit_vmstate.rs
```diff
@@ -0,0 +1,156 @@
...
+    f: impl Fn(MicrovmState) -> Result<MicrovmState, EditVmStateError>,
+) -> Result<(), EditVmStateError> {
+    let (microvm_state, version) = open_vmstate(vmstate_path)?;
+    let microvm_state = f(microvm_state)?;
+    save_vmstate(microvm_state, output_path, version)?;
+    Ok(())
+}
+
+#[cfg(target_arch = "aarch64")]
+fn remove_regs(
```
bchalios, line 62: what's the point of passing this function pointer in `edit`? It looks to me it's a needless indirection.
Why not simply call `remove_regs` when you get the `EditVmStateSubCommand::RemoveRegs` subcommand?

src/snapshot-editor/src/edit_vmstate.rs
```diff
@@ -0,0 +1,156 @@
...
+    mut state: MicrovmState,
+    remove_regs: &[u64],
+) -> Result<MicrovmState, EditVmStateError> {
+    for (i, vcpu_state) in state.vcpu_states.iter_mut().enumerate() {
+        println!("Modifing vcpu state: {i}");
```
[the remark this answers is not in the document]
bchalios: Oh, is it meant to mention which vCPU you're currently editing? If so,

```
println!("Modifying state for vCPU {i}");
```

reads better IMHO.

src/snapshot-editor/src/edit_vmstate.rs
```diff
@@ -0,0 +1,156 @@
...
+    #[error("{0}")]
+    Utils(#[from] UtilsError),
+}
+
+#[derive(Debug, Subcommand)]
+pub enum EditVmStateSubCommand {
+    /// Remove registers from vcpu states.
+    #[cfg(target_arch = "aarch64")]
+    RemoveRegs {
+        /// Set of regs to remove.
```
bchalios, line 24: It'd be nice to mention what these `u64` values actually are. Are they indexes, some other name, or key?

src/snapshot-editor/src/edit_vmstate.rs
```diff
@@ -0,0 +1,156 @@
...
+        for reg in vcpu_state.regs.iter().filter(|reg| {
+            if let Some(pos) = remove_regs.iter().position(|r| r == &reg.id) {
+                removed[pos] = true;
+                false
+            } else {
+                true
+            }
+        }) {
+            new_regs.push(reg);
+        }
```
bchalios, line 80: This seems a bit difficult to read. Would something like that:

```rust
for reg in vcpu_state.regs.iter() {
   if !remove_regs.contains(&reg.id) {
      new_regs.push(reg);
   } else {
}
```

not work?

bchalios: oh sorry, I see you do that to see if the user requested to delete a register that is not present. Do we really care about that?

docs/snapshotting/snapshot-editor.md
```diff
@@ -0,0 +1,59 @@
...
+This command is used to remove specified registers from vcpu states inside
+vmstate snapshot file.
+
+Usage:
+
+```bash
+snapshot-editor edit-vmstate remove-regs \
+    --vmstate-path <VMSTATE_PATH> \
+    --output-path <OUTPUT_PATH> \
+    [REGS]...
```
bchalios, line 31: Here, as well, we should describe what type of things are `REGS`

tests/integration_tests/functional/test_snapshot_editor.py
```diff
@@ -0,0 +1,78 @@
...
+    # Test that MIDR_EL1 is not in the snapshot
+    cmd = [
+        str(snap_editor),
+        "info-vmstate",
+        "vcpu-states",
+        "--vmstate-path",
+        str(snapshot.vmstate),
+    ]
+    _, stdout, _ = utils.run_cmd(cmd)
+    assert stdout.find(str(MIDR_EL1)) == -1
```
bchalios, line 68: I think this test can be more generic. I can think of at least two cases to test here:

1. We should check that no other registers were touched or added.
2. We should check that if we ask the tool to remove a non-existent register it does the right thing.

tests/integration_tests/functional/test_snapshot_editor.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # test that we can restore from a snapshot
+    new_vm = microvm_factory.build()
+    new_vm.spawn()
+    new_vm.restore_from_snapshot(snapshot, resume=True)
+
+    # Attempt to connect to resumed microvm.
+    # Verify if guest can run commands.
+    exit_code, _, _ = new_vm.ssh.execute_command("ls")
+    assert exit_code == 0
```
bchalios, line 78: I guess these need to be adapted to the new style of checking errors, right?

src/snapshot/src/lib.rs
```diff
@@ -137,17 +137,19 @@ impl Snapshot {
     pub fn unchecked_load<T: Read + Debug, O: Versionize + Debug>(
         mut reader: &mut T,
         version_map: VersionMap,
-    ) -> Result<O, Error> {
+    ) -> Result<(O, u16), Error> {
         let data_version = Self::get_data_version(&mut reader, &version_map)?;
-        O::deserialize(&mut reader, &version_map, data_version).map_err(Error::Versionize)
+        let res =
+            O::deserialize(&mut reader, &version_map, data_version).map_err(Error::Versionize)?;
+        Ok((res, data_version))
```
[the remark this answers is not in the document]
ShadowCurse: it is part of the 'snapshot'. The `O::deserialize` in this case can be thought as `MicrovmState::deserialize`. And `MicrovmState` does not have version inside of it.

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -238,7 +238,7 @@ impl KvmVcpu {
 pub struct VcpuState {
     pub mp_state: kvm_bindings::kvm_mp_state,
     #[version(end = 2, default_fn = "default_old_regs")]
-    old_regs: Vec<Aarch64RegisterOld>,
+    pub old_regs: Vec<Aarch64RegisterOld>,
```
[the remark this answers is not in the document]
ShadowCurse: Actually opposite. I prefer to have fields to be public. The prevention of changes to these fields should be with usage of `const` or `mut` modifiers.

src/snapshot-editor/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+enum Command {
+    #[cfg(target_arch = "aarch64")]
+    #[command(subcommand)]
+    EditVmstate(EditVmStateSubCommand),
+    #[cfg(target_arch = "aarch64")]
+    #[command(subcommand)]
+    InfoVmstate(InfoVmStateSubCommand),
+}
+
+#[allow(unused)]
```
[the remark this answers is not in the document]
ShadowCurse: On x86_64 the tool actually has not functionality, so `#[allow(unused)]` is needed to allow `cli` variable to be unused. The follow up PRs should change this.

src/snapshot-editor/src/edit_vmstate.rs
ShadowCurse: The idea here is to have a common code for all `edit` methods in one method. So if we add new `edit` methods it can be used as such:
```rust
 match command {
        SubCommand::RemoveRegs => edit(&vmstate_path, &output_path, |state| {
            remove_regs(state, &regs)
        })?,
        StateSubCommand::SomeEditMethod2 => edit(&vmstate_path, &output_path, some_edit_method_2)?,
        StateSubCommand::SomeEditMethod3 => edit(&vmstate_path, &output_path, some_edit_method_3)?,
}
```

src/vmm/src/vstate/vcpu/aarch64.rs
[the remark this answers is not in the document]
bchalios: It's not about preventing modification of the field, it's about being able to change the implementation without breaking the anybody who consumes the API, e.g. if we ever change `old_regs` to be stored in different kind of container you can still have a getter that returns a `Vec<Aarch64RegisterOld>` so that anybody calling that function won't break.

src/snapshot-editor/src/main.rs
bchalios: yup, exactly. And then you can get rid of all the other directives (target_arch and unused)

src/snapshot-editor/src/edit_vmstate.rs
bchalios: I got it, I just think it's unnecessary indirection. What's the benefit of this over:

```rust
 match command {
        SubCommand::RemoveRegs => remove_regs(&vmstate_path, &output_path)?,
        StateSubCommand::SomeEditMethod2 => some_edit_method_2(&vmstate_path, &output_path)?,
        StateSubCommand::SomeEditMethod3 => some_edit_method_3(&vmstate_path, &output_path)?,
}
```

src/vmm/src/vstate/vcpu/aarch64.rs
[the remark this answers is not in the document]
ShadowCurse: Considering we are the only ones consuming our internal API, we can handle the changes. Getters/setters just increase indirection and complexity and even in our code we don't use them much.

src/snapshot-editor/src/edit_vmstate.rs
ShadowCurse: I guess it is just personal preference. Otherwise each edit method will start with ` let (microvm_state, version) = open_vmstate(vmstate_path)?;` and end with `save_vmstate(microvm_state, output_path, version)?;`. Considering how small the amount of code is, both versions can be correct.

tests/integration_tests/functional/test_snapshot_editor.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pb8o, line 1: 2023 :)

tests/integration_tests/functional/test_snapshot_editor.py
```diff
@@ -0,0 +1,78 @@
...
+    # Test that MIDR_EL1 is in the snapshot
+    cmd = [
+        str(snap_editor),
+        "info-vmstate",
+        "vcpu-states",
+        "--vmstate-path",
+        str(snapshot.vmstate),
+    ]
+    _, stdout, _ = utils.run_cmd(cmd)
+    assert stdout.find(str(MIDR_EL1)) != -1
```
pb8o, line 44: Simpler: `assert MIDR_EL1 in stdout`

tests/integration_tests/functional/test_snapshot_editor.py
```diff
@@ -0,0 +1,78 @@
...
+
+PLATFORM = platform.machine()
+MIDR_EL1 = hex(0x603000000013C000)
+
+
+@pytest.mark.skipif(
+    PLATFORM != "aarch64",
+    reason="This is aarch64 specific test.",
+)
+def test_remove_regs(uvm_nano, microvm_factory):
```
pb8o, line 20: Is there any value in testing with all guest kernels?

tests/integration_tests/functional/test_snapshot_editor.py
pb8o, line 68: Simpler: `assert MIDR_EL1 not in stdout`

src/vmm/src/vstate/vcpu/aarch64.rs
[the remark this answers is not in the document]
bchalios: Everybody can handle the changes. I'm just suggesting that we do it with the minimum amount of effort. While at the same time avoiding making public things that otherwise shouldn't be.

src/snapshot-editor/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+#[cfg(target_arch = "aarch64")]
+use info::{info_vmstate_command, InfoVmStateError, InfoVmStateSubCommand};
+
+#[derive(Debug, thiserror::Error)]
+enum SnapEditorError {
+    #[cfg(target_arch = "aarch64")]
+    #[error("Error during editing vmstate file: {0}")]
+    EditVmState(#[from] EditVmStateError),
+    #[cfg(target_arch = "aarch64")]
+    #[error("Error during gettinf info from a vmstate file: {0}")]
```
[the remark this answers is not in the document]
ShadowCurse: fixed

src/snapshot-editor/src/main.rs
ShadowCurse: wait, this will not work, because when we build with `./tool/devtool build` it build everything, including `snapshot-editor`. And on x86 this build will fail. And CI will fail.

src/snapshot-editor/src/info.rs
bchalios: Ok, the previous case is fine. But here, will we ever do something with print info command other than printing info?

ShadowCurse: hmm... printing more info? I thought of adding print command for each data type in vmstate: vm memory, device states... etc.

tests/integration_tests/functional/test_snapshot_editor.py
ShadowCurse: isn't uvm_nano 5.10 only?

tests/integration_tests/functional/test_snapshot_editor.py
ShadowCurse: fixed

tests/integration_tests/functional/test_snapshot_editor.py
ShadowCurse: fixed

tests/integration_tests/functional/test_snapshot_editor.py
ShadowCurse: fixed

tests/integration_tests/functional/test_snapshot_editor.py
ShadowCurse: what is new way?

docs/snapshotting/snapshot-editor.md
ShadowCurse: Added parameters descriptions

src/snapshot-editor/src/edit_vmstate.rs
ShadowCurse: Added another comment

src/snapshot-editor/src/main.rs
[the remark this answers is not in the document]
ShadowCurse: fixed

docs/snapshotting/snapshot-editor.md
ShadowCurse: fixed

tests/integration_tests/functional/test_snapshot_editor.py
bchalios: oh sorry forget that. I thought this was an API call that would trigger an exception.

tests/integration_tests/functional/test_snapshot_editor.py
pb8o: Yeah, that's why I am asking if running in 4.14 could have different behavior.

src/snapshot-editor/src/main.rs
ShadowCurse: I don't think this is worth it. There will be a follow up PR with additional functionality available on both x86 ant aarch64, so x86 version will be not useful only for short period of time.

docs/snapshotting/snapshot-editor.md
```diff
@@ -0,0 +1,70 @@
...
+
+This command is used to remove specified registers from vcpu states inside
+vmstate snapshot file.
+
+Arguments:
+
+- `VMSTATE_PATH` - path to the `vmstate` file
+- `OUTPUT_PATH` - path to the file where the output will be placed
+- `[REGS]` - set of u32 values representing registers ids as they are defined
+  in the KVM. Can be both in decimal and in hex formats.
```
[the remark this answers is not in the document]
ShadowCurse: fixed

tests/integration_tests/functional/test_snapshot_editor.py
ShadowCurse: The test itself won't change behavior. Set of registers in the snapshot is same for all kernels.

src/snapshot-editor/src/main.rs
bchalios: It doesn't look too complicated to me:

```diff
diff --git a/tools/release.sh b/tools/release.sh
index 1c84e510..2fb55039 100755
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -60,6 +60,7 @@ function get-firecracker-version {
 # defaults
 LIBC=musl
 PROFILE=dev
+ARCH=$(uname -m)
 MAKE_RELEASE=

 #### Option parsing
@@ -130,7 +131,12 @@ fi

 say "Building version=$VERSION, profile=$PROFILE, target=$CARGO_TARGET..."
 # shellcheck disable=SC2086
-cargo build --target "$CARGO_TARGET" $CARGO_OPTS --workspace
+if [[ $ARCH != "aarch64" ]]; then
+    cargo build --target "$CARGO_TARGET" $CARGO_OPTS --workspace --exclude snapshot-editor
+else
+    cargo build --target "$CARGO_TARGET" $CARGO_OPTS --workspace
+fi
+

 say "Binaries placed under $CARGO_TARGET_DIR"
```

tests/integration_tests/functional/test_snapshot_editor.py
bchalios: what about 6.1?

src/snapshot-editor/src/main.rs
ShadowCurse: The idea here is that this change will be reverted when x86 functionality will be added. Most `#[cfg(target_arch = "aarch64")]` will be removed anyway.

tests/integration_tests/functional/test_snapshot_editor.py
ShadowCurse: Same

src/vmm/src/vstate/vcpu/aarch64.rs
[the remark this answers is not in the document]
roypat: I think in Rust, generally direct field access is preferred over getters/setters, mainly because Rust's ownership rules mean that there is multiple ways of "getting" data (e.g. a borrow vs moving), and also because direct field access allows for pattern matching, which is more idiomatic than a sequence of `if`s on `.get_(...)` calls.

docs/snapshotting/snapshot-editor.md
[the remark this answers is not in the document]
roypat: _technically_, its "the kernel virtual machine" xP English is weird

src/snapshot-editor/src/main.rs
roypat: I think the change in devtool makes sense, because why would we ship a binary to our customers that does nothing. It would just be confusing. If we do the devtool change, I dont think we need to change anything about the conditional compilation: We dont ship it, so it doesnt matter if it builds and produces a binary that does nothing, or if it doesnt even build.

bchalios: I know. I 'd just like to have a way to explicitly not build the tool for x86 until it actually does something on x86.

ShadowCurse: Ok I added changes to the `release.sh`

src/vmm/src/vstate/vcpu/aarch64.rs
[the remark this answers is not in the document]
bchalios: Ok, we're going into philosophical questions now :P AFAIK, for the time being, Rust seems to have issues with getting/moving struct fields comparing to the whole struct, but in any case, my main issue is about hiding implementation from the API, not ownership.

In any case, this is not a roadblock for me (marked as nit). Feel free to resolve Egor.

tests/integration_tests/functional/test_snapshot_editor.py
bchalios: my point here is that, we cannot be sure that it will always be the same, so maybe it makes sense to add tests for the rest of the kernels as well? WDYT @pb8o @ShadowCurse ?

ShadowCurse: As I mentioned before, this test will behave same for any kernel, because MIDR_EL1 reg will be in the snapshot in any case. If we want to have a test, that verifies that the set of regs is the same on all kernels, it should be a separate test.

bchalios: You are right. That's fine by me then.

src/snapshot-editor/src/main.rs
ShadowCurse: Well.. CI still fails because it still tries to build x86 version of `snapshot-editor`. I would prefer to have
 `#[cfg(target_arch = "aarch64")]` rather then `std::compile_error!()` in this case.

ShadowCurse: Reverted some changes. Left `#[cfg(target_arch = "aarch64")]` only in `main.rs`

bchalios: I'd still like to not ship a binary that does something in the release for x86 (this implies fixing the tests), but I'd like to hear as well from others, @pb8o, @roypat

bchalios: This patch should fix the CI, btw: 

```diff
diff --git a/tests/host_tools/cargo_build.py b/tests/host_tools/cargo_build.py
index dffe0479..67af0e00 100644
--- a/tests/host_tools/cargo_build.py
+++ b/tests/host_tools/cargo_build.py
@@ -63,6 +63,8 @@ def cargo_test(path, extra_args=""):
         "RUST_BACKTRACE": 1,
         "RUSTFLAGS": get_rustflags(),
     }
+    if platform.machine() != "aarch64":
+        extra_args += "--exclude snapshot-editor"
     cargo("test", extra_args + " --all --no-fail-fast", env=env)


diff --git a/tests/integration_tests/build/test_clippy.py b/tests/integration_tests/build/test_clippy.py
index dc4ae415..7dafd797 100644
--- a/tests/integration_tests/build/test_clippy.py
+++ b/tests/integration_tests/build/test_clippy.py
@@ -22,4 +22,5 @@ def test_rust_clippy(target):
     """
     Test that clippy does not generate any errors/warnings.
     """
-    cargo("clippy", f"--target {target} --all --profile test", "-D warnings")
+    extra_args = "--exclude snapshot-editor" if platform.machine() != "aarch64" else ""
+    cargo("clippy", f"--target {target} --all --profile test {extra_args}", "-D warnings")
diff --git a/tests/integration_tests/build/test_coverage.py b/tests/integration_tests/build/test_coverage.py
index ac3af0b2..96714159 100644
--- a/tests/integration_tests/build/test_coverage.py
+++ b/tests/integration_tests/build/test_coverage.py
@@ -42,7 +42,7 @@ def test_coverage(monkeypatch):
     # Generate test profiles.
     cargo(
         "test",
-        f"--all --target {TARGET}",
+        f"--all --target {TARGET} --exclude snapshot-editor",
         "--test-threads=1",
         env={
             "RUSTFLAGS": "-Cinstrument-coverage",
diff --git a/tests/integration_tests/build/test_unittests.py b/tests/integration_tests/build/test_unittests.py
index 8bcb80cb..79f42bcd 100644
--- a/tests/integration_tests/build/test_unittests.py
+++ b/tests/integration_tests/build/test_unittests.py
@@ -21,10 +21,10 @@ def test_unittests(test_fc_session_root_path):
     Run unit and doc tests for all supported targets.
     """
     extra_args = "--target {} ".format(TARGET)
-
     host.cargo_test(test_fc_session_root_path, extra_args=extra_args)


 def test_benchmarks_compile():
     """Checks that all benchmarks compile"""
-    host.cargo("bench", f"--all --no-run --target {TARGET}")
+    extra_args = "--exclude snapshot-editor" if platform.machine() != "aarch64" else ""
+    host.cargo("bench", f"--all --no-run --target {TARGET} {extra_args}")
```

roypat: But I thought with the patch to release.sh, we are no longer shipping a binary for x86_64 as part of the release?

as an aside, i think the patch for the CI can do directly into `def cargo(...)` in host_tools.py, that'd make for a smaller change to later revert.

bchalios: true, we can keep the `release.sh` changes and we're fine with the release binary. But we ship source code as well, so someone could get the code, build the tool and wait for it to do something on x86. It just sounds weird (apart from it being ugly). The CI changes are small enough (good catch Patrick about the `host.cargo` command) to just do them and then revert a single line once we add x86 functionality.

bchalios: If everybody else is fine with it, I won't block the PR for this.

bchalios: @roypat @pb8o could you chime in? this is the last open blocker

roypat (dismissed):

bchalios (approved):

roypat (approved):

PR #4047, opened 2023-08-20
Update logger

JonathanWoollett-Light:
## Changes

- Adds support for reconfiguring the logger at runtime. You can now call the logger API root multiple times without error.
- Greatly simplifies the logger implementation removing ~800 LoC.
- Reduces the size of the Firecracker stripped release binary by ~20kb (~0.7%) (`2681136` -> `2664816`).

## Reason

See above.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/vmm/src/resources.rs
```diff
@@ -138,7 +138,7 @@ impl VmResources {
         let vmm_config = serde_json::from_str::<VmmConfig>(config_json)?;
 
         if let Some(logger) = vmm_config.logger {
-            init_logger(logger, instance_info)?;
+            logger.apply()?;
         }
```
ShadowCurse, line 142: I think directly using `LOGGER` to apply config is better:
```rust
if let Some(logger_config) = vmm_config.logger_config {
    LOGGER.apply(logger_config)?;
}
```

JonathanWoollett-Light: Updated to include this.

src/logger/src/logger.rs
```diff
@@ -1,776 +1,343 @@
...
+    /// Whether to show the log origin in the log.
+    pub show_log_origin: Option<bool>,
+}
 
-        fn flush(&mut self) -> std::io::Result<()> {
-            Ok(())
+/// This is required since we originally supported `Warning` and uppercase variants being used as
+/// the log level filter. It would be a breaking change to no longer support this. In the next
+/// breaking release this should be removed (replaced with `log::LevelFilter` and only supporting
+/// its default deserialization).
```
ShadowCurse, line 175: Can be marked as TODO

ShadowCurse (approved):

src/logger/src/logger.rs
JonathanWoollett-Light: Yes you are right, I will create an issue for this before the PR merges but when it has 2 approved reviews.

src/logger/src/logger.rs
```diff
@@ -1,776 +1,343 @@
...
-        self.show_level.load(Ordering::Relaxed)
-    }
-
-    fn show_file_path(&self) -> bool {
-        self.show_file_path.load(Ordering::Relaxed)
-    }
+        if let Some(log_path) = config.log_path {
+            let file = std::fs::OpenOptions::new()
+                .custom_flags(libc::O_NONBLOCK)
+                .read(true)
```
pb8o, line 53: why would we want to read from the log file?

tests/framework/microvm.py
```diff
@@ -512,8 +512,6 @@ def spawn(
         # and leave 0.2 delay between them.
         if "no-api" not in self.jailer.extra_args:
             self._wait_create()
-        if self.log_file:
-            self.check_log_message("Running Firecracker")
```
pb8o, line 516: Hmm, we expect this message to be in the log, to know that Firecracker hasn't crashed. Why do we remove it?

src/logger/src/logger.rs
JonathanWoollett-Light: This is ported from the original, it also enabled `read`. I do not know why.

tests/integration_tests/performance/test_boottime.py
```diff
@@ -126,7 +126,7 @@ def _get_microvm_boottime(vm):
     if timestamps:
         boot_time_us = int(timestamps[0])
 
-    assert boot_time_us > 0
+    assert boot_time_us >= 0
```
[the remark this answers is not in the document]
JonathanWoollett-Light: I don't remember why this change is here, I think this from rebasing. I've removed it.

tests/framework/microvm.py
pb8o: hmm making the message consistent sounds good. As for printing that every time we change log target, I would prefer if that message was only printed once at the beginning of FC. When changing the logger config could we instead print "Changed logger config to X".

JonathanWoollett-Light: I think there is also the problem that our default level is `Warn` so if its `info!("Running Firecracker")` it won't print.

pb8o: Well in the integration tests we configure the logger with "Debug", so we should be covered. But moving this log message to info sounds like a good idea.

JonathanWoollett-Light: The problem is this message happens before the user configuration is applied.

The logger is immediately initialized when starting Firecracker with this PR (before even parsing the command line input), but it is still re-configured with the user command line input at the same point. So it is possible and likely that warnings get printed before the `Running Firecracker` message. To ensure its the first message we need to either ensure no logs or possible errors happen before it (which I don't think is possible) or we set the default level to `info` so it can be put immediately after initialization and still print.

src/logger/Cargo.toml
```diff
@@ -17,3 +17,6 @@ serde_json = "1.0.78"
 thiserror = "1.0.32"
 vm-superio = "0.7.0"
 utils = { path = "../utils" }
+
+[dev-dependencies]
+uuid = { version = "1.4.1", features=["v4"] }
```
pb8o, line 22: Is this dependency used? I don't see it anywhere?

JonathanWoollett-Light: It's used in the test.

src/logger/src/logger.rs
```diff
@@ -1,776 +1,343 @@
...
-//! `ERROR`, `WARN`, `INFO`, `DEBUG`, `TRACE`.
-//! The file path and the line provides the exact location of where the call to the macro was made.
-//! ## Example of a log line:
-//! ```bash
-//! 2018-11-07T05:34:25.180751152 [anonymous-instance:ERROR:vmm/src/lib.rs:1173] Failed to write
-//! metrics: Failed to write logs. Error: operation would block
-//! ```
-//! # Limitations
-//! Logs can be flushed either to stdout/stderr or to a byte-oriented sink (File, FIFO, Ring Buffer
-//! etc).
```
roypat, line 77: Is this documentation moved somewhere else? Or is it not applicable anymore?

src/logger/src/logger.rs
```diff
@@ -1,776 +1,343 @@
...
+pub static LOGGER: Logger = Logger(Mutex::new(LoggerConfiguration {
+    target: None,
+    filter: LogFiler {
+        level: log::LevelFilter::Warn,
+    },
+    format: LogFormat {
+        show_level: false,
+        show_log_origin: false,
+    },
+}));
```
roypat, line 35: Mhh, this adds more locking that I don't think its necessary (but concurrency in Rust is also not my strongest suit). But currently we need to lock the entire logger if we check the level (in enabled below). This means that even if a log message does not get printed, we will do a `Mutex::lock`. Cant we instead only wrap the `target` into a `Mutex`, and have the log level be a `AtomicU8` and `show_level` and `show_log_origin` be `AtomicBool`? They allow interior mutability through  `store`, and atomics should be more performant than mutexes (at the very least they should match whatever performance we have currently, as the old logger uses atomics for these)

src/logger/src/logger.rs
```diff
@@ -1,776 +1,343 @@
...
-    /// Creates a new instance of the current logger.
-    fn new() -> Logger {
-        Logger {
-            init: Init::new(),
-            log_buf: Mutex::new(Box::new(sink())),
-            show_level: AtomicBool::new(true),
-            show_line_numbers: AtomicBool::new(true),
-            show_file_path: AtomicBool::new(true),
-            instance_id: RwLock::new(String::new()),
+    /// Applies the given logger configuration the logger.
```
roypat, line 43: This function has significantly less documentation than the functions is replaces. I think we should try to preserve the parts of the documentation that are still relevant.

src/logger/src/logger.rs
```diff
@@ -1,776 +1,343 @@
...
+        let mut guard = self.0.lock().unwrap();
 
-        if self.show_line_numbers() {
-            if let Some(line) = record.line() {
-                prefix.push(line.to_string());
+        // Check if the log message is enabled
+        {
+            let enabled = record.level() <= guard.filter.level;
+            if !enabled {
+                return;
```
roypat, line 104: This should already be checked by the `log!` macro, via a call to `enabled`, no?

roypat (commented):
My only real concern is the Mutex/locking story, but preserving the documentation would also be nice-to-have (you can't really claim it simplifies the code by hundreds of lines if you removed hundreds of lines of documentation xP)

src/firecracker/src/main.rs
```diff
@@ -87,9 +88,9 @@ fn main() -> ExitCode {
 }
 
 fn main_exec() -> Result<(), MainError> {
-    LOGGER
-        .configure(Some(DEFAULT_INSTANCE_ID.to_string()))
-        .expect("Failed to register logger");
+    // Initialize the logger.
+    log::set_logger(&LOGGER).map_err(MainError::SetLogger)?;
+    log::set_max_level(log::LevelFilter::Trace);
```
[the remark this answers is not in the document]
JonathanWoollett-Light: So the weird thing about how the `log` crate works is that `log::set_max_level(log::LevelFilter::Trace);` sets the filter that is in-lined and called within the `info!` (etc.) macros before even calling `Log::log()` or `Log::enabled()`.

```rust
fn main() {
    log::info!("check")
}
```
`cargo expand`
```rust
#![feature(prelude_import)]
#[prelude_import]
use std::prelude::rust_2021::*;      
#[macro_use]
extern crate std;
fn main() {
    {
        let lvl = ::log::Level::Info;
        if lvl <= ::log::STATIC_MAX_LEVEL && lvl <= ::log::max_level() {
            ::log::__private_api::log(
                format_args!("check"),
                lvl,
                &("test_bin", "test_bin", "src\\main.rs"),
                2u32,
                ::log::__private_api::Option::None,
            );
        }
    }
}
```
where `::log::max_level()` is getting the value set from [log](https://docs.rs/log/latest/log/index.html)::[set_max_level](https://docs.rs/log/latest/log/fn.set_max_level.html#).

So now thinking about this we are checking this twice. I will remove `level` from the `Logger` struct and just use the inbuilt check. Thankyou for reminding me.

src/logger/src/logger.rs
JonathanWoollett-Light: Almost all is outdated.

src/logger/src/logger.rs
JonathanWoollett-Light: There will still need to be a `Mutex` wrapping `target` by having a `Mutex` on `show_level` and `show_log_origin` it prevents these being updated mid logging some data (and is a bit simpler).

Moving this `Mutex` inner would only give a speedup if someone is very frequently updating the logger format. I don't think this case is common.

src/logger/src/logger.rs
JonathanWoollett-Light: Are there specific parts of the documentation you think are worth keeping?

src/logger/src/logger.rs
JonathanWoollett-Light: It is not via `enabled` but you are correct it is not needed, see https://github.com/firecracker-microvm/firecracker/pull/4047#discussion_r1304159168.

src/logger/src/logger.rs
roypat: You are right, I wrote that comment under the assumption that the mutex would also need to be held to determine whether a message should be printed in the first place. If that is not the case (e.g. mutex only needs to be held when a message is actually logged), then there's no big difference, and this does indeed make it slightly simpler

tests/integration_tests/functional/test_logging.py
```diff
@@ -6,69 +6,88 @@
...
+
+    # Assert the time in the logs is less than or equal to the current time
+    log_time = datetime.datetime(
+        year=int(year),
+        month=int(month),
+        day=int(day),
+        hour=int(hour),
+        minute=int(minute),
+        second=int(second),
+        microsecond=int(int(nanosecond) / 1000),
```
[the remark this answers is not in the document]
JonathanWoollett-Light: This works. Changed.

src/firecracker/src/main.rs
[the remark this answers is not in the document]
JonathanWoollett-Light: I've updated it considering this.

src/logger/src/logger.rs
```diff
@@ -1,429 +1,143 @@
...
-            self.create_prefix(record),
-            record.args()
-        );
-        self.write_log(&msg);
+        // Lock the logger.
+        let mut guard = self.0.lock().unwrap();
+
+        // Prints log message
+        {
+            let thread = thread::current().name().unwrap_or("-").to_string();
```
roypat, line 101:
```suggestion
            let thread = thread::current();
            let thread = thread.name().unwrap_or("-");
```
one less `to_string()` xP

src/logger/src/logger.rs
roypat: I think maybe the code example and the configuration explanation would be nice to keep

src/logger/src/logger.rs
roypat: Yeah, looking at it outside of GitHub's diff view I agree, the code is simpler with a lot less functions now, so this is fine by me :)

src/logger/src/logger.rs
JonathanWoollett-Light: Fixed (I used `thread_name` for the 2nd variable instead of `thread`)

src/logger/src/logger.rs
JonathanWoollett-Light: I've added some documentation to `LOGGER`.

src/logger/src/logger.rs
```diff
@@ -432,345 +173,186 @@ impl Log for Logger {
...
-                    .set_include_level(false)
-                    .set_include_origin(false, false);
-                logger.set_instance_id("".to_string());
-                logger.mock_log(Level::Info, "thread-msg");
-                validate_log(&mut Box::new(&mut reader), "[custom-thread] thread-msg\n");
-            })
+    fn logger() {
+        // Create temp file.
+        let id = uuid::Uuid::new_v4();
+        let path = format!("/tmp/{id}");
```
pb8o, line 310: Here you can use a temporary file instead of a UUID. See https://docs.rs/vmm-sys-util/latest/vmm_sys_util/tempfile/index.html

It should be accessible from `utils::tempfile::TempFile`

tests/integration_tests/functional/test_logging.py
```diff
@@ -6,69 +6,88 @@
...
+        return 0
+    if level == "WARN":
+        return 1
+    if level == "INFO":
+        return 2
+    if level == "DEBUG":
+        return 3
+    if level == "TRACE":
+        return 4
+    raise Exception("Invalid level")
```
pb8o, line 29: Simpler:

```python
return LOG_LEVELS.index(level)
```

tests/framework/microvm.py
JonathanWoollett-Light: > We could create a fake log message using println!() that is formatted just like the logger would format an info!() message using the default configuration

This could work but its a bit messy imo.

I would prefer if we could change the default log level to `Info`. Are there are problems with this?

tests/integration_tests/functional/test_logging.py
```diff
@@ -6,69 +6,88 @@
...
+    log_time = datetime.datetime(
+        year=int(year),
+        month=int(month),
+        day=int(day),
+        hour=int(hour),
+        minute=int(minute),
+        second=int(second),
+        microsecond=int(nanosecond) // 1000,
+    )
+    assert log_time <= now
```
pb8o, line 70: Why not keep using strptime? If you want to fix the time resolution, you can do it in a separate commit, like this:

```python
timestamp = "2023-12-21T13:12:23.123456789"
strptime(timestamp[:-3], "%Y-%m-%dT%H:%M:%S.%f")
time.struct_time(tm_year=2023, tm_mon=12, tm_mday=21, tm_hour=13, tm_min=12, tm_sec=23, tm_wday=3, tm_yday=355, tm_isdst=-1)
```

tests/integration_tests/functional/test_logging.py
```diff
@@ -6,69 +6,88 @@
...
 
-    mo = re.match(pattern, tag_and_msg)
-    assert (
-        mo is not None
-    ), f"Log message ({tag_and_msg}) does not match pattern ({pattern})."
+    # [fc_api:INFO:src\main.rs:18]
+    data = next(split)
+    data_split = iter(data[1:-1].split(":"))
```
pb8o, line 74: I'd rather we keep using the regex to extract information, this approach to parsing is less readable IMO

tests/integration_tests/functional/test_logging.py
```diff
@@ -6,69 +6,88 @@
...
+    # Thread names are not optional.
+    _thread_name = next(data_split)
 
     if show_level:
-        tag_level = mo.group(3)
-        tag_level_no = LOG_LEVELS.index(tag_level)
-        configured_level_no = LOG_LEVELS.index(to_formal_log_level(level))
-        assert tag_level_no <= configured_level_no
+        level = next(data_split)
+        assert level in ("ERROR", "WARN", "INFO", "DEBUG", "TRACE")
```
pb8o, line 84:
```
assert level in LOG_LEVELS
```

tests/integration_tests/functional/test_logging.py
```diff
@@ -223,8 +242,5 @@ def _test_log_config(microvm, log_level="Info", show_level=True, show_origin=Tru
     microvm.start()
 
     lines = microvm.log_data.splitlines()
-    for idx, line in enumerate(lines):
-        if idx == 0:
-            assert line.startswith("Running Firecracker")
-            continue
```
pb8o, line 229: I think this message should be preserved still.

tests/framework/microvm.py
pb8o: I'm ok with changing it to info. I hope nobody is depending in the exact message like we did. :crossed_fingers: 

On the other hand, I am also OK with this message just being printed in the logs with no formatting since that's how it has always been.

tests/integration_tests/functional/test_logging.py
JonathanWoollett-Light: I think this is less precise, it doesn't check for zero padding and `%f` doesn't check for the max value.

tests/integration_tests/functional/test_logging.py
JonathanWoollett-Light: I think how we check this depends on https://github.com/firecracker-microvm/firecracker/pull/4047#discussion_r1304524154

tests/framework/microvm.py
ShadowCurse: I would agree with order 2 -> 1 -> 3.

tests/integration_tests/functional/test_logging.py
pb8o: I think matching the zero-padding is overkill, but if you want to check for that, use a regex: 

```python
re.fullmatch(r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{9}", timestamp)
<re.Match object; span=(0, 29), match='2023-12-21T13:12:23.123456789'>
```

But for the rest, use `strptime` . It will do the right thing up to microseconds. If you want to also validate nanoseconds, you can just parse the remaining 3 characters: 

```python
assert int(timestamp[-3:]) >= 0
```

tests/integration_tests/functional/test_logging.py
pb8o: See above, it can be done in one line with a regex vs 10+ lines and manually incrementing an iterator.

src/logger/src/logger.rs
pb8o: I don't follow... since it's in scope, it should only delete itself at the end of the test no?

pb8o (changes requested):

src/logger/src/logger.rs
JonathanWoollett-Light: To pass the `temp_file` to the logger it needs to be a `File` [`TempFile::into_file`](https://docs.rs/vmm-sys-util/latest/vmm_sys_util/tempfile/struct.TempFile.html#method.into_file) removes the file from the file system making it unusable.

So to expand on my example it would need to be:
```rust
let temp_file = TempFile::new();
let path = temp_file.as_path();
std::mem::forget(temp_file); // This also leaks a small amount of memory for the rest of this test run.
let file = std::fs::OpenOptions::new().write(true).open(&path).unwrap();
// ...
std::fs::remove(path).unmwrap();
```

JonathanWoollett-Light: I haven't been able to get a version working quickly using `TempFIle` something that you might expect to work like:
```rust
        // Create temp file.
        let target = TempFile::new().unwrap();
        let mut file = target.into_file();

        // Create logger.
        let logger = Logger(Mutex::new(LoggerConfiguration {
            target: Some(file.try_clone().unwrap()),
            format: LogFormat {
                show_level: true,
                show_log_origin: true,
            },
        }));

        // Assert results of enabled given specific metadata.
        assert!(logger.enabled(&Metadata::builder().level(Level::Warn).build()));
        assert!(logger.enabled(&Metadata::builder().level(Level::Debug).build()));

        // Log
        let metadata = Metadata::builder().level(Level::Error).build();
        let record = Record::builder()
            .args(format_args!("Error!"))
            .metadata(metadata)
            .file(Some("dir/app.rs"))
            .line(Some(200))
            .module_path(Some("module::server"))
            .build();
        logger.log(&record);

        Assert calling flush panics.
        std::panic::catch_unwind(|| {
            logger.flush();
        })
        .unwrap_err();

        // Asserts result of log.
        let mut contents = String::new();
        file.read_to_string(&mut contents).unwrap();
        let (_time, rest) = contents.split_once(' ').unwrap();
        let thread = thread::current().name().unwrap_or("-").to_string();
        assert_eq!(
            rest,
            format!("[{DEFAULT_INSTANCE_ID}:{thread}:ERROR:dir/app.rs:200] Error!\n")
        );
```
doesn't.

pb8o: Yeah that could be improved, but there are existing alternatives, so let's use those rather than introducing a new pattern to creating temp files and a new dependency.

tests/integration_tests/functional/test_logging.py
JonathanWoollett-Light: > See above, it can be done in one line with a regex

I think the date can be done with 1 regex match but the rest of the log message requires conditionally composing the string.

I think comparing the original approach for matching the infomation outside the date:

```python
    pattern = "\\[(" + instance_id + ")"
    pattern += ":(.*)"
    if show_level:
        pattern += ":(" + "|".join(LOG_LEVELS) + ")"
    if show_origin:
        pattern += ":([^:]+/[^:]+):([0-9]+)"
    pattern += "\\].*"

    mo = re.match(pattern, tag_and_msg)
    assert (
        mo is not None
    ), f"Log message ({tag_and_msg}) does not match pattern ({pattern})."

    if show_level:
        tag_level = mo.group(3)
        tag_level_no = LOG_LEVELS.index(tag_level)
        configured_level_no = LOG_LEVELS.index(to_formal_log_level(level))
        assert tag_level_no <= configured_level_no
```

With the new one:

```python
    data = next(split)
    data_split = iter(data[1:-1].split(":"))

    log_instance_id = next(data_split)
    assert log_instance_id == instance_id

    # Thread names are not optional.
    _thread_name = next(data_split)

    if show_level:
        level = next(data_split)
        assert level in LOG_LEVELS
        assert LOG_LEVELS[level] <= LOG_LEVELS[log_level.upper()]

    if show_origin:
        _file = next(data_split)
        line = next(data_split)
        assert line.isnumeric()
```

The new approach is simpler.

An example would be that:

```python
if show_origin:
        pattern += ":([^:]+/[^:]+):([0-9]+)"
```
involves a lot more cognitive overhead than
```python
    if show_origin:
        _file = next(data_split)
        line = next(data_split)
        assert line.isnumeric()
```

JonathanWoollett-Light: I think if we still want to use the original approach it would make sense to drop the commit that changes the test.

src/logger/src/logger.rs
JonathanWoollett-Light: Okay.

I think there is also a race condition where inbetween `temp_file.remove()` and `OpenOptions::new().write(true).open(path).unwrap()` the system could create the file with `path` for another `TempFile`. But it is very unlikely we encounter this in unittests.

roypat: Why can't we do `let tmp_file = TempFile::new(); target: Some(tmp_file.as_file().clone())`?

JonathanWoollett-Light: > Why can't we do `let tmp_file = TempFile::new(); target: Some(tmp_file.as_file().clone())`? `TempFile::as_file` doesnt drop the `TempFile` and thus doesnt delete it, so it should work.

This is the approach I tried above with `as_file().try_clone()` but I couldn't get it working.

tests/framework/microvm.py
JonathanWoollett-Light: I've implemented approach 2.

tests/framework/microvm.py
```diff
@@ -316,7 +316,11 @@ def _validate_api_response_times(self):
     def firecracker_version(self):
         """Return the version of the Firecracker executable."""
         _, stdout, _ = utils.run_cmd(f"{self._fc_binary_path} --version")
-        return re.match(r"^Firecracker v(.+)", stdout.partition("\n")[0]).group(1)
+        # The 1st line will be `Running Firecracker vX.Y.Z`
```
pb8o, line 319: Do we log by default to the console? The current version of Firecracker does not output anything:

```sh
./build/cargo_target/x86_64-unknown-linux-musl/release/firecracker --version
Firecracker v1.5.0-dev

Supported snapshot data format versions:
...
```

If we log to the console by default, can we disable that behavior for --version / --describe-snapshot / --help?

JonathanWoollett-Light: So now when starting Firecracker it will always print (with the version changing):
```
Running FIrecracker v1.5.0-dev
```
This does make `--version` somewhat redundant.

I think for now to avoid a more breaking changing it is best to avoid changing `--version` but it may make sense to adjust this in the future.

I have a commit on top of my changes that removes the `Running Firecracker` log which I think is the best approach https://github.com/JonathanWoollett-Light/firecracker/commit/ee883c602c60b5b70ec6e46a95c6945e04ca006d

Cargo.lock
```diff
@@ -783,6 +785,7 @@ dependencies = [
  "serde_json",
  "thiserror",
  "utils",
+ "uuid",
```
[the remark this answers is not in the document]
JonathanWoollett-Light: Yes, I've removed it.

src/api_server/swagger/firecracker.yaml
```diff
@@ -981,8 +981,8 @@ definitions:
       level:
         type: string
         description: Set the level. The possible values are case-insensitive.
-        enum: [Error, Warning, Info, Debug]
-        default: Warning
+        enum: [Error, Warn, Info, Debug, Trace, Off]
```
pb8o, line 984: hmm this is a breaking change, no? Could we keep it as Warning (or provide an alias)?

src/logger/Cargo.toml
```diff
@@ -17,4 +18,4 @@ serde_json = "1.0.78"
 thiserror = "1.0.32"
 displaydoc = "0.2.4"
 vm-superio = "0.7.0"
-utils = { path = "../utils" }
+utils = { path = "../utils" }
```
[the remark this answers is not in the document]
roypat: (Off-Topic, but I just noticed the same problem in one of my own PRs): Can we add an integration test for this if we want to enforce it?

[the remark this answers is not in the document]
JonathanWoollett-Light: Fixed. I also agree with @roypat comment, we should automate this.

src/api_server/swagger/firecracker.yaml
pb8o: Yeah if it supports both, could we just keep Warning forever and don't deprecate it? Sounds like it's not too hard to maintain.

JonathanWoollett-Light: If we did remove `Warning` we could remove the code (see below), it would also make the API more consistent, although it may be low priority. I will remove this change to the `firecracker.yml` here as this can be done or discussed later.
```rust
/// This is required since we originally supported `Warning` and uppercase variants being used as
/// the log level filter. It would be a breaking change to no longer support this. In the next
/// breaking release this should be removed (replaced with `log::LevelFilter` and only supporting
/// its default deserialization).
#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
pub enum LevelFilter {
    /// [`log::LevelFilter:Off`]
    #[serde(alias = "OFF")]
    Off,
    /// [`log::LevelFilter:Trace`]
    #[serde(alias = "TRACE")]
    Trace,
    /// [`log::LevelFilter:Debug`]
    #[serde(alias = "DEBUG")]
    Debug,
    /// [`log::LevelFilter:Info`]
    #[serde(alias = "INFO")]
    Info,
    /// [`log::LevelFilter:Warn`]
    #[serde(alias = "WARN", alias = "WARNING", alias = "Warning")]
    Warn,
    /// [`log::LevelFilter:Error`]
    #[serde(alias = "ERROR")]
    Error,
}
impl From<LevelFilter> for log::LevelFilter {
    fn from(filter: LevelFilter) -> log::LevelFilter {
        match filter {
            LevelFilter::Off => log::LevelFilter::Off,
            LevelFilter::Trace => log::LevelFilter::Trace,
            LevelFilter::Debug => log::LevelFilter::Debug,
            LevelFilter::Info => log::LevelFilter::Info,
            LevelFilter::Warn => log::LevelFilter::Warn,
            LevelFilter::Error => log::LevelFilter::Error,
        }
    }
}

impl FromStr for LevelFilter {
    type Err = LevelFilterFromStrError;
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "Off" | "OFF" => Ok(Self::Off),
            "Trace" | "TRACE" => Ok(Self::Trace),
            "Debug" | "DEBUG" => Ok(Self::Debug),
            "Info" | "INFO" => Ok(Self::Info),
            "Warn" | "WARN" | "Warning" | "WARNING" => Ok(Self::Warn),
            "Error" | "ERROR" => Ok(Self::Error),
            _ => Err(LevelFilterFromStrError(String::from(s))),
        }
    }
}
```

src/logger/src/logger.rs
```diff
@@ -1,428 +1,155 @@
...
-                // finish init
-                true
-            })
-            .map_err(LoggerError::Init)?;
-
-        self.write_log(&header);
+        // Ensure we drop the guard before attempting to log, otherwise this
+        // would deadlock.
+        drop(guard);
+        if target_changed {
```
ShadowCurse, line 90: Maybe add a comment here with reasons we print `"Running Firecracker v{FIRECRACKER_VERSION}"` every time we change target.

ShadowCurse (dismissed):

src/firecracker/src/main.rs
```diff
@@ -149,124 +153,114 @@ fn main_exec() -> Result<(), MainError> {
...
-        )
-        .arg(
-            Argument::new("seccomp-filter")
-                .takes_value(true)
-                .forbids(vec!["no-seccomp"])
-                .help(
-                    "Optional parameter which allows specifying the path to a custom seccomp \
-                     filter. For advanced users.",
+    let mut arg_parser =
+        ArgParser::new()
```
pb8o, line 156: Could the formatting changes be moved to the "cargo fmt" commit?

[update: Hide whitespace in github makes it easier to see, but still would be best to split in 2]

src/firecracker/src/main.rs
```diff
@@ -99,9 +102,10 @@ fn main() -> ExitCode {
 }
 
 fn main_exec() -> Result<(), MainError> {
-    LOGGER
-        .configure(Some(DEFAULT_INSTANCE_ID.to_string()))
-        .expect("Failed to register logger");
+    // Initialize the logger.
+    log::set_logger(&LOGGER).map_err(MainError::SetLogger)?;
+    log::set_max_level(vmm::logger::DEFAULT_LEVEL);
+    info!("Running Firecracker v{FIRECRACKER_VERSION}");
```
pb8o, line 107: Can we move this after some args processing? For example after line 289?

pb8o: That avoid the double printing message of `--version`

src/firecracker/src/main.rs
```diff
@@ -149,124 +153,114 @@ fn main_exec() -> Result<(), MainError> {
...
-                     filter. For advanced users.",
+    let mut arg_parser =
+        ArgParser::new()
+            .arg(
+                Argument::new("api-sock")
+                    .takes_value(true)
+                    .default_value(DEFAULT_API_SOCK_PATH)
+                    .help("Path to unix domain socket used by the API."),
+            )
+            .arg(
```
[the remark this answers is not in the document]
JonathanWoollett-Light: I have run `cargo fmt` in every commit, it seems the tool only felt it necessary to do it in this commit.

src/firecracker/src/main.rs
```diff
@@ -113,7 +117,7 @@ fn main_exec() -> Result<(), MainError> {
             // These errors are non-critical: In the worst case we have worse snapshot restore
             // performance.
             ResizeFdTableError::GetRlimit | ResizeFdTableError::Dup2(_) => {
-                vmm::logger::debug!("Failed to resize fdtable: {err}")
+                debug!("Failed to resize debugfdtable: {err}")
```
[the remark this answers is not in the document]
JonathanWoollett-Light: No, good catch, my fault.

This does perhaps highlight an area where we don't have tests.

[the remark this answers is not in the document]
JonathanWoollett-Light: I've opened https://github.com/firecracker-microvm/firecracker/issues/4143

src/firecracker/src/main.rs
JonathanWoollett-Light: I think this is same as https://github.com/firecracker-microvm/firecracker/pull/4047#discussion_r1342429600 no?

src/firecracker/src/main.rs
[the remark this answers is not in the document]
pb8o: Thanks!

I think testing all log output is overkill, since we don't consider it part of the public API.

src/firecracker/src/main.rs
JonathanWoollett-Light: I would prefer leaving it at the start, its simplest to have it as a startup message.

JonathanWoollett-Light: I have a commit down line that removes this startup message https://github.com/JonathanWoollett-Light/firecracker/commit/485cbc3715aedc41fe65231f3bfc188cda915ed1 that would remove this duplication.

src/api_server/swagger/firecracker.yaml
```diff
@@ -981,8 +981,8 @@ definitions:
       level:
         type: string
         description: Set the level. The possible values are case-insensitive.
-        enum: [Error, Warning, Info, Debug]
-        default: Warning
+        enum: [Error, Warning, Info, Debug, Trace, Off]
+        default: Info
```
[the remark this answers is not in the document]
pb8o: What would be inaccurate?

src/firecracker/src/main.rs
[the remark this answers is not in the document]
pb8o: Yeah it can be worked around, but moving forward, it would be good to split formatting changes from functional changes. At least the ones that span more than a few lines.

src/firecracker/src/main.rs
pb8o: Yeah, don't know why GitHub posted it twice.

src/api_server/swagger/firecracker.yaml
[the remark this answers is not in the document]
JonathanWoollett-Light: - The new default is `Info` should if not changed to `Info` here documenting it as `Warning` would be inaccurate.
- The API can be provided with `Trace` and `Off` so not having them is inaccurate.

src/firecracker/src/main.rs
JonathanWoollett-Light: Added a commit that refactors it like we discussed offline https://github.com/firecracker-microvm/firecracker/pull/4047/commits/c75d6fb573e40a7fd573f5227db06de7044fba5e

pb8o: Looks good! Thanks!

src/api_server/swagger/firecracker.yaml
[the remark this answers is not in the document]
pb8o: Thanks!

src/jailer/src/env.rs
```diff
@@ -462,10 +462,9 @@ impl Env {
 
         // We do a copy instead of a hard-link for 2 reasons
         // 1. hard-linking is not possible if the file is in another device
-        // 2. while hardlinking would save up disk space and also memory by
-        //    sharing parts of the Firecracker binary (like the executable .text
-        //    section), this latter part is not desirable in Firecracker's
-        //    threat model. Copying prevents 2 Firecracker processes from
+        // 2. while hardlinking would save up disk space and also memory by sharing parts of the
```
pb8o, line 465: Nit: this commit title should be `style: cargo fmt`.

pb8o (approved):

src/jailer/src/env.rs
JonathanWoollett-Light: Updated.

src/vmm/src/logger/logging.rs
```diff
@@ -431,343 +170,189 @@ impl Log for Logger {
...
+            .line(Some(200))
+            .module_path(Some("module::server"))
+            .build();
+        logger.log(&record);
+
+        // Assert calling flush panics.
+        std::panic::catch_unwind(|| {
+            logger.flush();
+        })
+        .unwrap_err();
```
roypat, line 345: nit: could be a `#[should_panic]` test. Why do we need to test this at all though? :o

roypat (approved):

src/vmm/src/logger/logging.rs
JonathanWoollett-Light: This wouldn't work like it is currently, `#[should_panic]` would prevent testing anything after the panic occurs.

Calling `flush` would need to be moved to the end of the test.

> Why do we need to test this at all though?

It adds a small amount of coverage, there's a very small benefit and no notable downside.

PR #4080, opened 2023-08-30
A/B-Testing utilities for specifically testing changes introduced by a PR

roypat:
## Changes

Generalizes the A/B-testing approach used by the optional benchmark test (which does a dynamic comparison of PR HEAD vs main HEAD and determines pass state based on that). 

## Reason

We want to apply this approach to performance testing and some auditing test that can block PRs across the board due to hardware changes (for example the cpu-template-helper tool tests that ensure cpu fingerprint reported inside guests stay the same, but which can fail due to microcode updates)

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tests/framework/ab_test.py
+++ b/tests/framework/ab_test.py
@@ -1,0 +1,132 @@
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Defines utilities for performing A/B-tests.
+
+A/B-Tests are style of tests where we do not care what state a test is in, but only that this state does not change
+across a pull request. This is useful if
+1. Validating the state requires some baseline to be persisted in the repository, and maintaining this baseline
+   adds significant operational burden (for example, performance tests), or
+2. The state can change due to outside factors (e.g. Hardware changes), and such external changes would block all
+   pull requests until they are resolved.
+
+Consider for example a `cargo audit` tests, which is used to reject usage of dependency versinos that have known
+security vulnerabilities, or which have been yanked. The "state" here is "list of vulnerable dependencies". Clearly,
+this can change due to external action (a new vulnerability is discovered and published to RustSec). At this point,
+every PR would fail until this dependency is removed, blocking all development. Simply removing the test from PR CI
+is not an option, since we want to avoid the scenario where a PR adds a dependency with a known vulnerability (e.g.
+the PR itself changes the "list of vulnerable dependencies"). A/B-Testing allows us to not block PRs on the former case,
+while still preventing the latter: We run cargo audit twice, once on main HEAD, and once on the PR HEAD. If the output
+of both invocations is the same, the test passes (with us being alerted to this situtation via a special pipeline that
+does not block PRs). If not, it fails, preventing PRs from introducing new vulnerable dependencies.
+"""
+import contextlib
+import os
+from tempfile import TemporaryDirectory
+from typing import Callable, Optional, TypeVar
+
+from framework import utils
+
+# Locally, this will always compare against main, even if we try to merge into, say, a feature branch.
+# We might want to do a more sophisticated way to determine a "parent" branch here.
+DEFAULT_A_REVISION = os.environ.get("BUILDKITE_PULL_REQUEST_BASE_BRANCH", "main")
+
+
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+def default_comparator(ah: T, be: T) -> bool:
+    """Returns `true` iff that the two arguments are equal.
+
+    The default assertion for A/B-tests using `ab_test`.
+
+    Ridiculous variable names sponsored by pylint."""
+    return ah == be
+
+
+def git_ab_test(
+    test_runner: Callable[[bool], T],
+    comparator: Callable[[T, T], U] = default_comparator,
+    *,
+    a_revision: str = DEFAULT_A_REVISION,
+    b_revision: Optional[str] = None,
+) -> (T, T, U):
+    """
+    Performs an A/B-test using the given test runner between the specified revision, and the currently checked out revision.
+    The specified revisions will be checked out in temporary directories, with `test_runner` getting executed in the
+    repository root. If the test depends on firecracker binaries built from the requested revision, care has to be taken
+    that they are built from the sources in the temporary directory (which can be ensured via the `workspace` parameter
+    to `cargo_build.get_binary`).
+
+    Note that there are no guarantees on the order in which the two tests are run.
+
+    :param test_runner: A callable which when executed runs the test in the context of the current working directory. Its
+                        parameter is `true` if and only if it is currently running the "A" test.
+    :param comparator: A callable taking two outputs from `test_runner` and comparing them. Should return some value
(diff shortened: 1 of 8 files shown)


roypat:
I've realized that "A/B-testing" is more of a philosophy than something that can be captured in a framework, so this for now only contains a utility for doing an A/B-test across git revisions on the same instance. Something like an A/B-test of S3 artifacts (say because we want to test the same git revision across two AMIs, e.g. on different instances) would be captured by a separate `*_ab_test` function which would not setup a copy of the repository in a temporary file, but instead just downloads two files from S3, does some processing, and then calls the comparator with the result of that processing (and maybe we'll see some way to unify these in the future, but for now, lets not try to guess at abstractions).

tests/host_tools/cargo_build.py
```diff
@@ -7,7 +7,7 @@
 from pathlib import Path
 
 from framework import defs, utils
-from framework.defs import FC_WORKSPACE_DIR, FC_WORKSPACE_TARGET_DIR
+from framework.defs import FC_WORKSPACE_DIR
```
pb8o, line 10: I think the constant is not used anywhere else, could be dropped from defs.py

tests/framework/ab_test.py
```diff
@@ -0,0 +1,130 @@
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pb8o, line 1: 2023

tests/framework/ab_test.py
```diff
@@ -0,0 +1,130 @@
...
+import contextlib
+import os
+from tempfile import TemporaryDirectory
+from typing import Callable, Optional, TypeVar
+
+from framework import utils
+
+# Locally, this will always compare against main, even if we try to merge into, say, a feature branch.
+# We might want to do a more sophisticated way to determine a "parent" branch here.
+DEFAULT_A_REVISION = os.environ.get("BUILDKITE_PULL_REQUEST_BASE_BRANCH") or "main"
```
pb8o, line 33: Can include the default in the .get

```python
DEFAULT_A_REVISION = os.environ.get("BUILDKITE_PULL_REQUEST_BASE_BRANCH", "main")
```

tests/framework/ab_test.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+
+        with chdir(tmp_dir):
+            utils.run_cmd(f"git checkout {revision}")
+
+        yield tmp_dir
+
+
+@contextlib.contextmanager
+def chdir(to):
```
pb8o, line 122: In Python 3.11 this is in contextlib. We could add a note https://docs.python.org/3/library/contextlib.html#contextlib.chdir

tests/framework/ab_test.py
```diff
@@ -0,0 +1,130 @@
...
+def temporary_checkout(revision: str):
+    """
+    Context manager that checks out firecracker in a temporary directory and `chdir`s into it
+
+    Will change back to whatever was the current directory when the context manager was entered, even if exceptions
+    happen along the way.
+    """
+    with TemporaryDirectory() as tmp_dir:
+        utils.run_cmd(
+            f"git clone https://github.com/firecracker-microvm/firecracker {tmp_dir}"
```
pb8o, line 112: Are there any situations where we wouldn't want this hardcoded?

tests/integration_tests/performance/test_benchmarks.py
```diff
@@ -34,51 +49,44 @@ def cargo_bench():
...
-    for benchmark in os.listdir("../build/cargo_target/criterion"):
+def compare_results(location_a_baselines: Path, location_b_baselines: Path):
+    """Compares the two recorded criterion baselines for regressions, assuming that "A" is the baseline from main"""
+    for benchmark in location_b_baselines.glob("*"):
         with open(
-            f"../build/cargo_target/criterion/{benchmark}/new/estimates.json",
-            encoding="utf-8",
+            benchmark / "b_baseline" / "estimates.json", encoding="utf-8"
         ) as file:
             data = json.load(file)
```
pb8o, line 66:
```python
json.load(file.open(encoding...))
```

pb8o (dismissed):
Minor comments

tests/host_tools/cargo_build.py
roypat: Oh, neat! 

As an aside, since we synced offline to do the A/B-testing by compiling the binaries outside of pytest, this commit is actually not needed anymore, so we can just drop it altogether. I personally do not really see a use for it anymore.

tests/framework/ab_test.py
roypat: Yeah, for example if someone wants to run an A/B-test against a commit that's not in the firecracker repository (say, because its in a fork). Since right now this is only used for testing in PRs (where A = head of PR target inside firecracker repo and B = whatever is currently checked out), this is not a limitation we need to address just now though

tests/integration_tests/performance/test_benchmarks.py
roypat: Ah, right, I did this somewhere else and then promtly forgot about it for this commit :joy:

tests/framework/ab_test.py
```diff
@@ -0,0 +1,132 @@
...
+U = TypeVar("U")
+
+
+def default_comparator(ah: T, be: T) -> bool:
+    """Returns `true` iff that the two arguments are equal.
+
+    The default assertion for A/B-tests using `ab_test`.
+
+    Ridiculous variable names sponsored by pylint."""
+    return ah == be
```
zulinx86, line 46: nit:

> Ridiculous variable names sponsored by pylint.

This comment made me laugh :þ

How about `a_out` and `b_out`, since outputs assume to be given to them?
Or `a_result` and `b_result` like those in `git_ab_test()`?

tests/framework/ab_test.py
```diff
@@ -0,0 +1,132 @@
...
+    that they are built from the sources in the temporary directory (which can be ensured via the `workspace` parameter
+    to `cargo_build.get_binary`).
+
+    Note that there are no guarantees on the order in which the two tests are run.
+
+    :param test_runner: A callable which when executed runs the test in the context of the current working directory. Its
+                        parameter is `true` if and only if it is currently running the "A" test.
+    :param comparator: A callable taking two outputs from `test_runner` and comparing them. Should return some value
+                       indicating whether the test should pass or no, which will be returned by the `ab_test` functions,
+                       and on which the caller can then do an assertion.
```
zulinx86, line 69: Just out of curiosity, do we have any specific usecase to return non-bool value from a comparator?

zulinx86 (approved):

tests/framework/ab_test.py
roypat: Originally the idea was that for example a performance A/B test returns a p-value or similar. That allows the test to emit more detailed comparison output, without the comparator needing to know about the metrics component of our testing, which means they can be used more generally (say, imported into a Jupyter notebook)

zulinx86: Thanks. I was just thinking of whether it needs to be generalized much.

pb8o (approved):

PR #4086, opened 2023-09-01
Handle kernel-triggered page fault for 6.1 kernels

bchalios:
## Reason

In kernels >= 5.11 UFFD behaviour changes in that it distinguishes between page faults triggered in kernel and user space.
In order to create UFFDs that can handle kernel-triggered page faults, the process needs to have the `CAP_SYS_PTRACE` capability or, starting from kernel 6.1, use the `/dev/userfaultfd` device to create the file descriptor.

Right now, Firecracker opts-out from handling kernel triggered page faults. This does not affects us for kernels 4.14 and 5.10 but it will on 6.1.

## Changes

* Modify our example UFFD handler to serve a single page per time, so we exercise the problematic behaviour on 6.1. 
* Use a new version of `userfaultfd-rs` which, if present, uses `/dev/userfaultfd` to create userfault file descriptors.
* Also, modify jailer to create `/dev/userfaultd` inside the jail if the device exists in the host

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


tests/integration_tests/functional/test_uffd.py
```diff
@@ -61,14 +62,10 @@ def spawn_pf_handler(vm, handler_path, mem_path):
...
 
-    uffd_handler.spawn()
-    try:
-        outs, errs = uffd_handler.proc().communicate(timeout=1)
-        print(outs)
-        print(errs)
+    uffd_handler.spawn(Path("/uffd.log"))
+    time.sleep(1)
+    if not uffd_handler.is_running():
         assert False, "Could not start PF handler!"
```
pb8o, line 68: Could we `print(uffd.log_data)`?

src/jailer/src/env.rs
```diff
@@ -43,6 +43,17 @@ const DEV_URANDOM_WITH_NUL: &str = "/dev/urandom";
...
+// Userfault file descriptor device path. This is a misc character device
+// with a MISC_DYNAMIC_MINOR minor device:
+// https://elixir.bootlin.com/linux/v6.1.51/source/fs/userfaultfd.c#L2176.
+//
+// This means that its minor device number will be allocated at run time,
+// so we will have to find it at initialization time parsing /proc/misc.
+// What we do know is the major number for misc devices:
+// https://elixir.bootlin.com/linux/v6.1.51/source/Documentation/admin-guide/devices.txt
+const DEV_UFFD_PATH: &str = "/dev/userfaultfd";
+const DEV_UFFD_MAJOR: u32 = 10;
```
pb8o, line 55: Couldn't we just `stat("/dev/userfaultfd)` and copy the major and minor from there?

On second thought, since the others are hardcoded it's probably better to just keep them that way. Copying would be simpler but could potentially open the door to some attack.

tests/integration_tests/functional/test_uffd.py
bchalios: Yup, I can add it.

src/jailer/src/env.rs
pb8o: Yeah, just looking at the host's /dev/userfaultfd and copying whatever major/minor it has, instead of going through /proc/misc. But I think your approach is safer.

tests/framework/utils.py
```diff
@@ -96,20 +96,42 @@ def __init__(self, name, args):
...
 
-    def spawn(self):
+    def spawn(self, log_file: Path):
         """Spawn handler process using arguments provided."""
-        self._proc = subprocess.Popen(
-            self._args, stdout=subprocess.PIPE, stderr=subprocess.PIPE
-        )
 
+        self._log_file = log_file
+        with open(self._log_file, "w", encoding="utf-8") as logfile:
```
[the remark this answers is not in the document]
pb8o: Thanks, you are right... closing on the parent process has no effect on the child.

Cargo.toml
```diff
@@ -12,3 +12,4 @@ lto = true
 
 [patch.crates-io]
 kvm-bindings = { git = "https://github.com/firecracker-microvm/kvm-bindings", tag = "v0.6.0-2", features = ["fam-wrappers"] }
+userfaultfd = { git = "https://github.com/bchalios/userfaultfd-rs", branch = "kernel-6.1" }
```
ShadowCurse, line 15: I assume you will open a PR to upstream `userfaultfd-rs`. Otherwise it is OK to rely on the for for now, but then it might be worth adding `rev="..."` to fix the commit.

src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -99,23 +99,24 @@ impl UffdPfHandler {
...
-        // This offers an example of how memory can be loaded in RAM,
-        // however this can be adjusted to accommodate use case needs.
+    fn populate_from_file(&self, region: &MemRegion, dst: u64) -> (u64, u64) {
+        let offset = dst - region.mapping.base_host_virt_addr;
+        let src = self.backing_buffer as u64 + region.mapping.offset + offset;
+
+        // Populate a single page from backing memory file.
+        // This is just an example, probably, with the worst-case latency scenario,
+        // of how memory can be loaded in guest RAM.
+        let len = get_page_size().unwrap();
```
ShadowCurse, line 109: I understand this is an example, but doing a `sysconf` for each page fault to get a page size seems a bit too much. Maybe move page size outside of the function and pass it as a parameter?

Cargo.toml
bchalios: Yes, once we have the code upstream, I will drop this. That's why the PR is not Draft

src/firecracker/examples/uffd/uffd_utils.rs
bchalios: Fixed

src/jailer/src/env.rs
```diff
@@ -626,6 +658,12 @@ impl Env {
                 println!("MMDS version 2 will not be available to use.");
             });
 
+        // If we have a minor version for /dev/userfaultfd the device is present on the host.
+        // Expose the device in the jailed environment.
```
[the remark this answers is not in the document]
bchalios: It is unlikely, but the file might not exist but the device might be loaded (because for example nothing ran `mknod`).
I thought this is more generic.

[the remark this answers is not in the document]
ShadowCurse: In my mind checking `/dev/userfaultfd` and then moving it into the jail is more strait forward and makes code more simple (`std::path::Path::new("/dev/userfaultfd").exists()` vs opennig a file and parsing it's content).
Also I don't think there can be a scenario when `/dev/userfaultfd` does not exist, but the device is still available.

tests/integration_tests/functional/test_uffd.py
bchalios: done!

Cargo.toml
bchalios: updated to the new version of userfaultfd-rs

bchalios:
Converting back to Draft, so that I can add some Documentation and a CHANGELOG entry.

CHANGELOG.md
```diff
@@ -42,6 +42,16 @@
...
+- We use a new version of `userfaultfd-rs`, which changes the way that a
+  userfault file descriptor (UFFD) is created when running on (host) Linux
+  kernels >=6.1. In these kernels the library first tries to create the UFFD
+  through an `ioctl()` to the `/dev/userfaultfd` device. If the device is not
+  present, the library falls back to calling the `userfaultfd` system call.
+  Access to `/dev/userfaultfd` is managed through normal filesystem
+  permissions. That means that, when using the UFFD functionality on Linux
+  kernel 6.1, users need to ensure that the Firecracker process has sufficient
+  permissions to access `/dev/userfaultfd`. This change does not affect users
+  that use UFFD on host kernels 4.14 and 5.10.
```
kalyazin, line 54: would it be possible to use kernel version ranges instead? Customers may use anything in between 4.14 and 6.1 too.

bchalios: I wanted to explicitly use the kernel versions we support, so that we are explicit about that. Do you think we should add documentation for non-supported versions?

kalyazin: > Linux kernels >=6.1

This kind of already goes beyond that. I think at least we can be consistent with either saying what happens only on 4.14, 5.10, 6.1 or covering a range. This is more of a nit though.

kalyazin: Actually, the more I read this, the more I think it should better go to the doc folder, and the changelog summary could be more concise.

bchalios: > > Linux kernels >=6.1
> 
> This kind of already goes beyond that. I think at least we can be consistent with either saying what happens only on 4.14, 5.10, 6.1 or covering a range. This is more of a nit though.

>= 6.1 refers to what Linux does. When I refer to 4.14, 5.10 & 6.1 later, I refer to what Firecracker supports. I'd like to distinguish between the two. Mainly, because I'd like to avoid having to debug cases of users on the non-supported ranges.

bchalios: > Actually, the more I read this, the more I think it should better go to the doc folder, and the changelog summary could be more concise.

I can try to make this more concise, but I'd like to have _some_ info about the fact that users need to take care their `/dev/userfaultfd` permissions in the CHANGELOG. WDYT?

tests/framework/utils.py
```diff
@@ -91,25 +92,73 @@ def get_cpu_percent(pid: int) -> Dict[str, Dict[str, float]]:
...
 
+        # Need to keep these around so we can exit the chroot
+        real_root = os.open("/", os.O_RDONLY)
+        working_dir = os.getcwd()
+
+        # Jump in the chroot
+        os.chroot(self._chroot)
+        os.chdir("/")
+        st = os.stat(self._handler_name)
+        os.chmod(self._handler_name, st.st_mode | stat.S_IEXEC)
```
pb8o, line 115: We don't need to chroot to access files inside the chroot. We can just do:

```python
os.stat(self._chroot / self._handler_name)
```

tests/framework/utils.py
```diff
@@ -91,25 +92,73 @@ def get_cpu_percent(pid: int) -> Dict[str, Dict[str, float]]:
...
+            print(chroot_log_file.read_text(encoding="utf-8"))
+            assert False, "Could not start PF handler!"
+
+        # The page fault handler will create the socket path with root rights.
+        # Change rights to the jailer's.
+        os.chown(self._socket_path, uid, gid)
+
+        os.fchdir(real_root)
+        os.chroot(".")
+        os.chdir(working_dir)
```
pb8o, line 135: Could do a context manager for the chroot part:

```python
@contextmanager
def chroot(*args, **kwds):
    try:
        # Need to keep these around so we can exit the chroot
        real_root = os.open("/", os.O_RDONLY)
        working_dir = os.getcwd()

        # Jump in the chroot
        os.chroot(self._chroot)
        os.chdir("/")
        yield
    finally:
        os.fchdir(real_root)
        os.chroot(".")
        os.chdir(working_dir)
```

Then use it like:

```
with chroot("/chroot"):
    pass

```

tests/framework/utils.py
```diff
@@ -91,25 +92,73 @@ def get_cpu_percent(pid: int) -> Dict[str, Dict[str, float]]:
...
+        # Jump in the chroot
+        os.chroot(self._chroot)
+        os.chdir("/")
+        st = os.stat(self._handler_name)
+        os.chmod(self._handler_name, st.st_mode | stat.S_IEXEC)
+
+        chroot_log_file = Path("/") / self._log_file
+        with open(chroot_log_file, "w", encoding="utf-8") as logfile:
+            args = [f"/{self._handler_name}", self._socket_path, self._mem_file]
+            self._proc = subprocess.Popen(
```
pb8o, line 120: Could use preexec_fn to execute as uid/gid but not sure if it's worth the trouble: https://stackoverflow.com/questions/61302291/python-subprocess-popen-execute-as-different-user

docs/snapshotting/handling-page-faults-on-snapshot-resume.md
```diff
@@ -19,9 +19,35 @@ use case, doing this for every page can be time-consuming.
 
 Userfaultfd is a mechanism that passes that responsibility of handling page
 fault events from kernel space to user space. In order to be able to interact
-with this mechanism, userspace needs to firstly obtain an userfault object
-(i.e file descriptor) by calling into [`userfaultfd`
-syscall](https://man7.org/linux/man-pages/man2/userfaultfd.2.html).
+with this mechanism, userspace needs to firstly obtain an userfault file descriptor
+object (UFFD).
+
+### Creating UFFD object
```
pb8o, line 25: nit: "Creating an UFFD object"

tests/framework/utils.py
bchalios: That's awesome. I wanted to do something like that, I failed and then gave up, given that we only seem to use it in this particular test. I 'll give it a go.

tests/framework/utils.py
bchalios: True, I don't think it makes much of a difference though. Actually, doing it inside the chroot does is slightly simpler, because we don't have to concatenate paths (yes, I'm that lazy :P)

tests/framework/utils.py
bchalios: I tried that, but pylint complained that this function might not be safe in the presence of threads.

docs/snapshotting/handling-page-faults-on-snapshot-resume.md
bchalios: done

ShadowCurse (dismissed):

src/jailer/src/env.rs
```diff
@@ -355,6 +383,19 @@ impl Env {
         write!(pid_file, "{}", pid).map_err(|err| JailerError::Write(pid_file_path, err))
     }
 
+    fn get_userfaultfd_minor_dev_number() -> Result<u32, UserfaultfdParseError> {
+        let buf = read_to_string("/proc/misc")?;
+
+        for line in buf.lines() {
+            let dev: Vec<&str> = line.split(' ').collect();
+            if dev[1] == "userfaultfd" {
```
kalyazin, line 391: I know, not likely, but is the `/proc/misc` file format guaranteed? Is it possible that `dev[1]` doesn't exist?

src/jailer/src/env.rs
```diff
@@ -86,6 +97,16 @@ fn clone(child_stack: *mut libc::c_void, flags: libc::c_int) -> Result<libc::c_i
     .map_err(JailerError::Clone);
 }
 
+#[derive(Debug, thiserror::Error)]
+enum UserfaultfdParseError {
+    #[error("Could not read /proc/misc: {0}")]
+    ReadProcMisc(#[from] std::io::Error),
+    #[error("Could not parse minor number: {0}")]
+    ParseDevMinor(#[from] std::num::ParseIntError),
+    #[error("/dev/userfaultfd device not loaded")]
```
kalyazin, line 106: nit: we don't check for the device node, so maybe remove `/dev/` to avoid possible confusion for someone trying to create the device on their own via mknod to fix the error.

src/jailer/src/env.rs
bchalios: It will be there. If there is a line it will refer to the minor node (dev[0]) and device name (dev[1]).

CHANGELOG.md
```diff
@@ -42,6 +42,16 @@
 - Changed the dump feature of `cpu-template-helper` tool not to enumerate program
   counter (PC) on ARM because it is determined by the given kernel image and
   it is useless in the custom CPU template context.
+- We use a new version of `userfaultfd-rs`, which changes the way that a
+  userfault file descriptor (UFFD) is created when running on (host) Linux
+  kernels >=6.1. In these kernels the library first tries to create the UFFD
+  through an `ioctl()` to the `/dev/userfaultfd` device. If the device is not
+  present, the library falls back to calling the `userfaultfd` system call.
```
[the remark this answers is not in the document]
bchalios: I like it. Thanks for the suggestion.

src/jailer/src/env.rs
bchalios: However, Patrick had the same observation and it doesn't hurt checking, so I'm adding the check.

tests/framework/utils.py
```diff
@@ -88,28 +90,88 @@ def get_cpu_percent(pid: int) -> Dict[str, Dict[str, float]]:
...
+    # Jump in the chroot
+    os.chroot(path)
+    os.chdir("/")
+
+    yield
+
+    # Jump out of the chroot
+    os.fchdir(real_root)
+    os.chroot(".")
+    os.chdir(working_dir)
```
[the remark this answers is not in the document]
bchalios: Actually, I just realized we don't need to `chdir` back to the original directory. The `chdir` happens inside the `chroot` Once we exit the `chroot` we're back to where we should be.

CHANGELOG.md
bchalios: FYI, I addressed this according to a suggestion from Patrick: https://github.com/firecracker-microvm/firecracker/pull/4086#discussion_r1334560717

CHANGELOG.md
```diff
@@ -45,6 +45,9 @@
 - The ability to create snapshots for an older version of Firecracker is now
   deprecated. As a result, the `version` body field in `PUT` on
   `/snapshot/create` request in deprecated.
+- Added support for the /dev/userfaultfd device available on linux kernels >=
+  6.1. This is the default for creating UFFD handlers on these kernel versions.
+  If it is unavailable, firecracker falls back to the userfaultfd syscall.
```
kalyazin, line 50: nit: s/fire/Fire/ for consistency

bchalios: done!

ShadowCurse (approved):

kalyazin (approved):

Issue #4090, opened 2023-09-06
[Bug] Incorrect spelling of "nomodule" kernel parameter


bduffany:
The default kernel cmdline includes "nomodules" as an arg, but the actual argument name seems to be "nomodule" (as pointed out by @xmarcalx in slack - [thread](https://firecracker-microvm.slack.com/archives/CDL3FUR8B/p1694034000232299)). So the current arg will be incorrectly forwarded to the init binary, instead of being consumed by the kernel.

Reference: https://www.kernel.org/doc/html/v4.10/admin-guide/kernel-parameters.html

PR #4085, opened 2023-09-01
chore: Update vm-memory to 0.12.2

roypat:
With 0.12.1, .as_ptr() got deprecated because on Xen, some additional guards need to be around before VolatileSlices can be accessed (as memory is mapped in on-demand, and those guards call mmap and mmunmap). On KVM, these are no-ops, so just inserting .ptr_guard() at the correct spaces is okay.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following
Developer Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

---

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #4091, opened 2023-09-06
Fix spelling of 'nomodule' kernel param

bduffany:
## Changes

Changes "nomodules" in the default kernel cmdline to "nomodule", the correct spelling. The old argument was not recognized by the kernel and incorrectly forwarded to init binaries.

Reference: https://www.kernel.org/doc/html/v4.10/admin-guide/kernel-parameters.html

Note, **this is a breaking change** for anyone who is relying on the default kernel command line, and happens to be loading kernel modules at runtime - or if their init binary was inadvertently relying on `nomodules` to be present in argv.

Fixes https://github.com/firecracker-microvm/firecracker/issues/4090

## License Acceptance

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,13 @@
 - Fixed the T2S CPU template to set the GDS_NO bit of the IA32_ARCH_CAPABILITIES
   MSR to 1 in accordance with an Intel microcode update. To use the template
   securely, users should apply the latest microcode update on the host.
+- Fixed the spelling of the `nomodule` param passed in the default kernel
+  command line parameters. This is a **breaking change** for setups that
+  use the default kernel command line which also depend on being able to
+  load kernel modules at runtime. This may also break setups which use the
+  default kernel command line and which use an init binary that
+  inadvertently depends on the misspelled param ("nomodules") being
+  present at the command line, since this param will no longer be passed.
 
 ## [1.4.0]
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -134,7 +134,7 @@ line that enables the serial console (which goes in the `boot_args` field of
 the `/boot-source` Firecracker API resource):
 
 ```console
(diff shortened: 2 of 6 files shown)


xmarcalx:
Thanks a lot for contributing also the fix!

I enabled the tests to run and I think we will need to add a changelog entry for this fix, highlighting the possibility of be a breaking change like you correctly suggested in the body of the description of the issue.

For the rest looks good, pending result of the tests.

xmarcalx:
The issue on the firecracker-pr pipeline related to sec_audit is related to #4085, so does not require any action from you, we just need to rebase this branch on top of that.
Style of the commit (add a body description) need to be fixed which together with the previous message seems to be the only changes required.

bduffany:
> I think we will need to add a changelog entry for this fix, highlighting the possibility of be a breaking change

Done

> Style of the commit (add a body description) need to be fixed

Done

bchalios (approved):
LGTM

xmarcalx (approved):
LGTM thanks a lot for the contribution!

bchalios (commented):
On a second thought, this is not a breaking change, this is a bug in Firecracker exactly because of this:

> This may also break setups which use the
  default kernel command line and which use an init binary that
  inadvertently depends on the misspelled param ("nomodules") being
  present at the command line, since this param will no longer be passed.

Shall we update this entry to reflect that? WDYT @xmarcalx ?

bchalios (changes requested):
Discarding my previous approval until we decide what to do with that.

bchalios (approved):
Ok, I understand why this is a breaking change. Just for documenting it:

With this change we break users who were using the default command line **and** loading modules. In reality this should not work, but due to the bug (`nomodule` vs `nomodules`) it would. So fixing the bug, it would break these users.

I believe that this CHANGELOG entry describes well the situation and it should be enough.

PR #4095, opened 2023-09-08
docs: Add CODEOWNERS for markdown files

roypat:
To automatically request reviews

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


.github/CODEOWNERS
```diff
@@ -0,0 +1,14 @@
+# All markdown files
+*.md @xmarcalx @kalyazin @pb8o @sudanl0
+
+# But not the ones in docs/
+docs/*.md
+
+# Except these specific ones
+docs/RELEASE_POLICY.md @xmarcalx @kalyazin @pb8o @sudanl0
```
xmarcalx, line 8: I do not think we need to specify just this policy file but we should include any policy file in this rule
I am not sure if the rule is case sensitive but if it is case sensitive i will simply define a rule
*policy*.md
*POLICY*.md

or in case it is case insensitive just one of the two. 

At the moment you have a bug that a file like kernel-policy.md would be ignored.

roypat: Ahh, I thought we only had one policy file. Yes I think a wildcard rule like that will work (I don't think its case insensitive)

xmarcalx (changes requested):
The PR has a functional bug, please see comment for details.

Description of the PR is not exhaustive and should provide more context.

.github/CODEOWNERS
xmarcalx: NP, even if we had only one policy document at the moment, it would not have been very future proofing having such rule so we should think in case to something else.

.github/CODEOWNERS
```diff
@@ -0,0 +1,17 @@
...
+
+# But not the ones in docs/
+docs/*.md
+
+# Except these specific ones
+docs/getting-started.md @xmarcalx @kalyazin @pb8o @sudanl0
+docs/prod-host-setup.md @xmarcalx @kalyazin @pb8o @sudanl0
+
+# Also cover all "*-policy.md" documents
+**/*policy.md @xmarcalx @kalyazin @pb8o @sudanl0
```
xmarcalx, line 12: nit: I do not know if a * before .md would be safer as per previous comment (which unfortunately made it italic!) but I am ok also with this.

xmarcalx (approved):
LGTM thanks for the changes

kalyazin (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,17 @@
...
+**/*POLICY.md @xmarcalx @kalyazin @pb8o @sudanl0
+
+# Also these non-md files in the repository root
+THIRD_PARTY @xmarcalx @kalyazin @pb8o @sudanl0
+LICENSE @xmarcalx @kalyazin @pb8o @sudanl0
```
pb8o, line 17: Should we also add PGP-KEY.asc NOTICE ?

roypat: Sounds reasonable. On an unrelated note, our NOTICE has "Copyright 2017 - 2020", which seems... wrong?

.github/CODEOWNERS
roypat: Ah, sorry, done

xmarcalx (approved):

PR #4107, opened 2023-09-13
fix: Avoid reallocating fd table during snapshot restore

roypat:
Linux maintains a per-process file descriptor table. By default, this table has size 64. Whenever space in this table runs out, the file descriptor table gets reallocated, with its size increased to the next power of two.

Firecracker creates a lot of eventfds and timerfds for its devices. These are created in the hot path of snapshot restore. For medium to large microVMs, firecracker uses more than 64 file descriptors, meaning we reallocate the file descriptor table on the hot path. However, this reallocation uses a RCU, meaning the kernel needs to hold a lock. To acquire this lock, it needs to wait for all other accesses to the file descriptors to cease, which introduces a severe latency to snapshot-restore times (between 30ms and 70ms).

This commit avoids that latency by ensuring the file descriptor table is large enough to hold the jailer-defined limit of file descriptors at process start already. This avoids reallocating the file descriptor table at runtime (and the memory overhead is negligable, as each entry in the fdtable is simply a pointer).

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/jailer/src/resource_limits.rs
```diff
@@ -98,6 +98,17 @@ impl ResourceLimits {
...
         ResourceLimits::set_limit(Resource::RlimitNoFile, self.no_file)?;
 
+        // Resize the file descriptor table to its maximal possible size, to ensure that
+        // firecracker will not need to reallocate it later. If the file descriptor table
+        // needs to be reallocated (which by default happens once more than 64 fds exist,
+        // something that happens for reasonably complex microvms due to each device using
+        // a multitude of eventfds), this can incur a significant performance impact (it
+        // was responsible for a 30ms-70ms impact on snapshot restore times).
+        // SAFETY: Duplicating stdin is safe
+        if unsafe { libc::dup2(0, self.no_file as libc::c_int - 1) } < 0 {
```
ShadowCurse, line 108: Shouldn't new fd be closed after `dup2`? There is no point in keeping it after table is resized.

bchalios: I think @ShadowCurse is right. Unless we close the duped, we'll be keeping 2047 copies of stdin fd around for no reason.

roypat: Ahhh, you're right. I thought it would get auto-closed by `sanitize_process`, but that called before the resource limits are setup, so it does indeed leave it open (only a single copy though, `dup2` does the same as `dup`, only that we get to choose what number our duplicate should have)

roypat: Done :)

bchalios: That's wrong. We will not be keeping 2047 copies of the stdin, just one. But yes, we could close it.

src/jailer/src/resource_limits.rs
```diff
@@ -98,6 +98,24 @@ impl ResourceLimits {
         // Set limit on number of file descriptors.
         ResourceLimits::set_limit(Resource::RlimitNoFile, self.no_file)?;
 
+        // Resize the file descriptor table to its maximal possible size, to ensure that
+        // firecracker will not need to reallocate it later. If the file descriptor table
+        // needs to be reallocated (which by default happens once more than 64 fds exist,
+        // something that happens for reasonably complex microvms due to each device using
+        // a multitude of eventfds), this can incur a significant performance impact (it
+        // was responsible for a 30ms-70ms impact on snapshot restore times).
+        if self.no_file > 64 {
```
bchalios, line 107: 64 here is driven from a constant in the kernel. I think we should not rely to that in case it changes in the future. Do you see any issue with removing the `if` statement all together?

kalyazin: What's going to happen if ` self.no_file as libc::c_int - 1` already exists? I don't think we'd like to replace/close it.

bchalios: We could try to `dup2` `self.no_file as libc::c_int` instead. Nothing should be using that, right?

kalyazin: but that would be rejected by the jailer?

ShadowCurse: doesn't jailer limit the number of 'opened' fds? When we do the 'dup2' we just ask for 1 mode fd, but with big number. So this is a +1 -1 situation. Technically we can resize table to any size. It is just only `no_file` number of fds can be open at the same time.

bchalios: I think what Egor says is true. If it isn't, we could just apply this hack before setting the limit, right?

roypat: > 64 here is driven from a constant in the kernel. I think we should not rely to that in case it changes in the future. Do you see any issue with removing the `if` statement all together?

We need some check here because if `no_file = 3` then `dup2` would fail (as stdin, stdout, stderr are already 3 open fds). We actually have a test for this. So we can change the check to 3 (I picked 64 because it doesnt makes sense to do it if no_file=64, but I can see how hardcoding kernel details is not nice, so will remove)

bchalios (approved):

ShadowCurse (dismissed):

src/firecracker/src/main.rs
```diff
@@ -96,6 +108,19 @@ fn main_exec() -> Result<(), MainError> {
     #[cfg(target_arch = "aarch64")]
     enable_ssbd_mitigation();
 
+    if let Err(err) = resize_fdtable() {
+        match err {
+            // These errors are non-critical: In the worst case we have worse snapshot restore
+            // performance.
+            ResizeFdTableError::GetRlimit | ResizeFdTableError::Dup2(_) => {
+                logger::warn!("Failed to resize fdtable: {err}")
```
bchalios, line 116: IMO, printing a warning for an optimization is a bit too much. I'd keep this a debug message

src/firecracker/src/main.rs
```diff
@@ -397,6 +422,51 @@ fn main_exec() -> Result<(), MainError> {
     }
 }
 
+/// Attempts to resize the processes file descriptor table to match RLIMIT_NOFILE or 2048, whichever
```
bchalios, line 425: Where does 2048 come from?

bchalios (changes requested):

src/firecracker/src/main.rs
roypat: Fair, will change

src/firecracker/src/main.rs
roypat: Since we no longer are in the jailer, we cannot assume that a limit is actually set. 2048 is the default from the jailer

src/firecracker/src/main.rs
```diff
@@ -397,6 +422,54 @@ fn main_exec() -> Result<(), MainError> {
...
+        2048
+    } else {
+        rlimit.rlim_cur.try_into().unwrap_or(2048)
+    };
+
+    // Resize the file descriptor table to its maximal possible size, to ensure that
+    // firecracker will not need to reallocate it later. If the file descriptor table
+    // needs to be reallocated (which by default happens once more than 64 fds exist,
+    // something that happens for reasonably complex microvms due to each device using
+    // a multitude of eventfds), this can incur a significant performance impact (it
```
kalyazin, line 456: I started wondering if we want to mention that in the changelog.

src/firecracker/src/main.rs
bchalios: sure but we moved the change in Firecracker to avoid the assumption that we are spawned from the jailer. That makes me think that the 2048 thing is arbitrary and I wonder if there's a better limit that we should be using (for example `rlim_cur`.

roypat: Mh, so on my laptop I have `limit.rlim_max = 1048576`. I don't think we want to allocate that large of an fd table. Although maybe if we document it and just tell the users to set rlimits correctly when running firecracker? Not sure which solution I prefer here tbh

src/firecracker/src/main.rs
roypat: Yeah, that's fair. I'll get to that after my lunch break

roypat: Done!

src/firecracker/src/main.rs
roypat: Synced offline, since using rlim_max would result in a very significant memory overhead, we decided to leave it at the arbitrary default of 2048 (which matches the default the jailer uses for RLIMIT_NOFILE)

CHANGELOG.md
```diff
@@ -34,6 +34,11 @@
   apply the mitigation against MMIO stale data vulnerability when it is running
   on a processor that does not enumerate FBSDP_NO, PSDP_NO and SBDR_SSDP_NO on
   IA32_ARCH_CAPABILITIES MSR.
+- Made Firecracker resize its file descriptor table on process start. It now
+  preallocates the in-kernel fdtable to hold `RLIMIT_NOFILE` many fds (or 2048
+  if no limit is set). This avoids the kernel reallocating the fdtable during
+  firecracker operations, resulting in a 30ms to 70ms reduction of snapshot
```
kalyazin, line 40: nit: two different capitalisations of `firecracker` in this paragraph. If you push again, would be worth unifying.

ShadowCurse (approved):

bchalios (approved):
🚢

PR #4120, opened 2023-09-21
Moving seccomp trigger

ShadowCurse:
## Changes
Moved seccompiler step in firecracker build process from root `build.rs` into `build.rs` inside firecracker crate.
Instead of calling seccomp binary it now calls directly into seccomp lib to compile filters.
Also removed `FIRECRACKER_VERSION` definitions from `build.rs` because it had same value as `CARGO_PKG_VERSION`. Because of this `build.rs` in root now does not have any usage and it was removed.
Now all crates use `CARGO_PKG_VERSION` directly.

Also should solve #3188

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


roypat (dismissed):
:100:

src/firecracker/build.rs
```diff
@@ -0,0 +1,96 @@
...
+            println!(
+                "cargo:rerun-if-changed={}",
+                path.to_str().expect("Invalid unicode bytes.")
+            );
+        } else if metadata.is_dir() {
+            // If is a folder, recurse.
+            register_seccompiler_src_watchlist(&path);
+        }
+    }
+}
```
pb8o, line 90: From the docs https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed
> If the path points to a directory, it will scan the entire directory for any modifications.

Could we remove this and simplify line 51 to:

```rust 
            println!(
                "cargo:rerun-if-changed={}",
                SECCOMPILER_SRC_DIR.to_str().expect("Invalid unicode bytes.")
            );
```

Probably in a separate commit.

src/firecracker/Cargo.toml
```diff
@@ -36,6 +36,12 @@ regex = { version = "1.9.5", default-features = false, features = ["std", "unico
 serde = { version = "1.0.188", features = ["derive"] }
 userfaultfd = "0.6.1"
 
+[build-dependencies]
+bincode = "1.2.1"
+seccompiler = { path = "../seccompiler" }
+serde = { version = "1.0.188" }
+serde_json = "1.0.107"
```
pb8o, line 43: Cannot it take those transitively since they are defined in seccompiler?

build.rs
```diff
@@ -1,15 +0,0 @@
-// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pb8o, line 1: This is a nice cleanup! I would just adjust the commit message to refactor or fix like so:

`refactor: remove redundant crate version variables`

src/firecracker/build.rs
ShadowCurse: good catch, will do

ShadowCurse: done

build.rs
ShadowCurse: done

roypat (approved):

pb8o (approved):

src/firecracker/build.rs
alindima: This was purposefully implemented like this because at that time, cargo was not looking at all of the sources recursively: https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-changedpath.
The behaviour was OS-dependent and last time I checked, it did not detect changes in the filters unless I traversed every folder recursively.

PR #4121, opened 2023-09-21
fix: remove default target for cargo + check static linking of firecracker release binary

sudanl0:
## Changes

- Remove default target from .cargo/config
- check if Firecracker release binary is actually compiled using musl (static link) or gnu (dynamic link)
- test static linking of release binary
- create a dummy tests/test-report that release.sh can copy to avoid copy error

## Reason

- The global `target` directive in .cargo/build forced the x86_64-unknown-linux-musl target whenever `--target` flag was not provided and because of this `cargo build` produced x86_64 binaries even if we tried to build on ARM. Though `devtool build` is the documented way to build Firecracker and does not have this issue since it always specifies the right arch `--target`, we would like to fix issues like #3221 and be able to just run `cargo run` without specifying the target.
- We want to make sure Firecracker release binary is always compiled with static linking.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] ~If a specific issue led to this PR, this PR closes the issue.~
- [ ] ~API changes follow the [Runbook for Firecracker API changes][2].~
- [ ] ~User-facing changes are mentioned in `CHANGELOG.md`.~
- [ ] ~New `TODO`s link to an issue.~

---

### DIFF
--- a/.buildkite/pipeline_pr.py
+++ b/.buildkite/pipeline_pr.py
@@ -49,7 +49,7 @@ devctr_grp = group(
 
 release_grp = group(
     "📦 Release Sanity Build",
-    "touch ./tests/test-report.json && ./tools/devtool -y sh ./tools/release.sh --libc musl --profile release --make-release",
+    "mkdir -p ./test_results && touch ./test_results/test-report.json && ./tools/devtool -y sh ./tools/release.sh --libc musl --profile release --make-release",
     **defaults_once_per_architecture,
 )
 
--- a/.buildkite/pipeline_pr.py
+++ b/.buildkite/pipeline_pr.py
@@ -49,7 +49,7 @@ devctr_grp = group(
 
 release_grp = group(
     "📦 Release Sanity Build",
-    "./tools/devtool -y sh ./tools/release.sh --libc musl --profile release --make-release",
+    "touch ./tests/test-report.json && ./tools/devtool -y sh ./tools/release.sh --libc musl --profile release --make-release",
     **defaults_once_per_architecture,
 )
 
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,6 +1,5 @@
 [build]
 target-dir = "build/cargo_target"
-target = "x86_64-unknown-linux-musl"
 rustflags = [
   "-Wclippy::ptr_as_ptr",
   "-Wclippy::undocumented_unsafe_blocks",
--- a/tests/integration_tests/build/test_binary_static_linking.py
+++ b/tests/integration_tests/build/test_binary_static_linking.py
@@ -1,0 +1,22 @@
+# Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Tests to check if the release binary is statically linked.
+
+"""
+
+import pytest
+import host_tools.cargo_build as host
+from framework import utils
+
+@pytest.mark.timeout(500)
+def test_firecracker_binary_static_linking():
+    """
+    Test to make sure the firecracker binary is statically linked.
+    """
+    fc_binary_path = host.get_binary("firecracker")
+    _, stdout,stderr = utils.run_cmd(f"file {fc_binary_path}")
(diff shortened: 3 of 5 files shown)


.cargo/config
```diff
@@ -1,6 +1,5 @@
 [build]
 target-dir = "build/cargo_target"
-target = "x86_64-unknown-linux-musl"
```
pb8o, line 3: As part of testing can we run this in both ARM/x86? https://github.com/firecracker-microvm/firecracker/blob/main/.buildkite/pipeline_pr.py#L52

sudanl0: we could use `file` which is already available in the devctr :
```
root:/firecracker# file ./build/cargo_target/aarch64-unknown-linux-musl/debug/firecracker | grep "statically linked"
./build/cargo_target/aarch64-unknown-linux-musl/debug/firecracker: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), statically linked, BuildID[sha1]=f401b1c0598407f0a0ca9027c61b01fe1b699d23, with debug_info, not stripped
```

tests/integration_tests/build/test_binary_static_linking.py
```diff
@@ -0,0 +1,18 @@
...
+from framework import utils
+
+@pytest.mark.timeout(500)
+def test_firecracker_binary_static_linking():
+    """
+    Test to make sure the firecracker binary is statically linked.
+    """
+    fc_binary_path = host.get_binary("firecracker")
+    result = utils.run_cmd(f"file {fc_binary_path}")
+    assert "statically linked" not in result
```
roypat, line 18: The release binary is not build by `hot.get_binary()`, we probably also want to put this assertion at the end of `release.sh` (btw, shouldnt it be `"statically linked in results"`, instead of `not in`?)

sudanl0: as per https://github.com/firecracker-microvm/firecracker/blob/main/tests/host_tools/cargo_build.py#L79 it looks like host.get_binary() is returning the release binary, am I missing something?
and yeah the previous check were not correct, I have updated it thanks for pointing out.

roypat: > as per https://github.com/firecracker-microvm/firecracker/blob/main/tests/host_tools/cargo_build.py#L79 it looks like host.get_binary() is returning the release binary, am I missing something?

host.get_binary() is producing a release profile binary, but our _release script_ does not use host.get_binary() to get its release-profile binary :D But generally, "binary compiled with --release flag" and "binary is statically linked" are independent of each other, so even if we get a statically linked binary from host.get_binary() does not tell us anything about the binary that release.sh is producing
Sorry, should've been clearer.

.buildkite/pipeline_pr.py
```diff
@@ -49,7 +49,7 @@
 
 release_grp = group(
     "📦 Release Sanity Build",
-    "./tools/devtool -y sh ./tools/release.sh --libc musl --profile release --make-release",
+    "touch ./tests/test-report.json && ./tools/devtool -y sh ./tools/release.sh --libc musl --profile release --make-release",
```
roypat, line 52: Mhh, why? It should normally put this file into `./test_results/test-report.json` :thinking:

pb8o: Good catch.

Yep, we have to remove the option from here and fix the release.sh to match the new results location: https://github.com/firecracker-microvm/firecracker/blob/main/tools/devtool#L486

I forgot to do it during 41412c14d5d2697a95e121869d3d8c4352087564. @sudanl0 could you include that fix in this PR?

pb8o (commented):
LGTM, just needs a fix to release.sh

pb8o (approved):
Thanks!

tests/integration_tests/build/test_binary_static_linking.py
sudanl0: update the commit msg to reflect this.

roypat (approved):

PR #4136, opened 2023-09-27
test: Skip kani if not enough RAM

roypat:
Trying to run the kani integration test outside of the CI will almost certainly result in out of memory conditions in the best case, and in system hangs in the worst case. Therefore, do not attempt to run kani if running tests locally.

Also removes an unneeded assertion, as `run_cmd` already checks the exit code by default.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/docs/formal-verification.md
+++ b/docs/formal-verification.md
@@ -45,16 +45,17 @@ properties, **all Kani harnesses are run on every pull request in our CI.** To
 check whether the harnesses all work for your pull request, check out the
 “Kani” [Buildkite](https://buildkite.com/) step.
 
-You can also execute our harnesses locally through `./tools/devtool test --
-integration_tests/build/test_kani.py`. To only run individual harnesses, you
-will need to first [install
+To run our harnesses locally, you can either enter our CI docker container
+via `./tools/devtool shell -p`, or by [installing
 Kani](https://model-checking.github.io/kani/install-guide.html#installing-the-latest-version)
-locally.
-From there on, individual test can be run via `cargo kani` similarly to how
-`cargo test` can run individual unit tests, the only difference being that the
-harness needs to be specified via `--harness`. Note that the first invocation
+locally.  Note that the first invocation
 of Kani post-installation might take a while, due to it setting up some
 dependencies.
+
+Individual harnesses can then be executed using `cargo kani` similarly to how
+`cargo test` can run individual unit tests, the only difference being that the
+harness needs to be specified via `--harness`. Note, however, that many harnesses
+require significant memory, and might result in OOM conditions.
 
 ## An example harness
 
--- a/tests/integration_tests/test_kani.py
+++ b/tests/integration_tests/test_kani.py
@@ -4,6 +4,7 @@
 Proofs ensuring memory safety properites, user-defined assertions,
 absence of panics and some types of unexpected behavior (e.g., arithmetic overflows).
 """
+import os
 import platform
 
 import pytest
@@ -15,6 +16,10 @@ PLATFORM = platform.machine()
 
 @pytest.mark.timeout(1800)
 @pytest.mark.skipif(PLATFORM != "x86_64", reason="Kani proofs run only on x86_64.")
+@pytest.mark.skipif(
+    os.environ.get("BUILDKITE") != "true",
+    reason="Kani's memory requirements likely cannot be satisfied locally",
+)
 def test_kani(results_dir):
     """
     Test all Kani proof harnesses.
@@ -24,10 +29,8 @@ def test_kani(results_dir):
     # -j enables kani harnesses to be verified in parallel (required to keep CI time low)
     # --output-format terse is required by -j
     # --enable-unstable is needed for each of the above
-    rc, stdout, stderr = utils.run_cmd(
+    _, stdout, _ = utils.run_cmd(
         "cargo kani --enable-unstable --enable-stubbing --restrict-vtable -j --output-format terse"
     )
-
-    assert rc == 0, stderr
 
     (results_dir / "kani_log").write_text(stdout, encoding="utf-8")
--- a/.buildkite/pipeline_pr.py
+++ b/.buildkite/pipeline_pr.py
@@ -109,7 +109,9 @@ if any(x.name == "Dockerfile" for x in changed_files):
 if any(x.parent.name == "tools" and "release" in x.name for x in changed_files):
     steps.append(release_grp)
 
-if not changed_files or any(x.suffix == ".rs" for x in changed_files):
+if not changed_files or any(
+    x.suffix in [".rs", ".toml", ".lock"] for x in changed_files
+):
     steps.append(kani_grp)
 
 if run_all_tests(changed_files):
--- a/tools/devtool
+++ b/tools/devtool
@@ -574,7 +574,7 @@ cmd_test() {
     # in order to set-up the Firecracker jail (manipulating cgroups, net
     # namespaces, etc).
     # We need to run a privileged container to get that kind of access.
-    env |grep -P "^(AWS_EMF_|BUILDKITE_|CODECOV_)" > env.list
+    env |grep -P "^(AWS_EMF_|BUILDKITE|CODECOV_)" > env.list
 
     if [[ "$BUILDKITE" = "true" ]]; then
       # Disable turbo boost. Some of our tests are performance tests, and we want minimum variability wrt processor frequency


tests/integration_tests/test_kani.py
```diff
@@ -19,15 +21,23 @@ def test_kani(results_dir):
     """
     Test all Kani proof harnesses.
     """
+    # https://psutil.readthedocs.io/en/latest/#psutil.virtual_memory confusingly enough returns
+    # information about the physical memory. Limiting to 64GB should exclude most consumer hardware,
+    # preventing the test from locking up peoples computers, while still allowing them to force their
+    # way through it with enough swap.
+    if (
+        psutil.swap_memory().total + psutil.virtual_memory().total
+        <= 64 * 1000 * 1000 * 1000
```
JonathanWoollett-Light, line 30:
```suggestion
        <= 64 * 1024 * 1024 * 1024
```

roypat: Ah, whoops, I do mean `GB` here, forgot update the skip message

JonathanWoollett-Light: Its unusual to specifically use GB instead of GiB why would we do it in this case?

tests/integration_tests/test_kani.py
```diff
@@ -19,15 +21,23 @@ def test_kani(results_dir):
...
+    ):
+        pytest.skip("kani requires at least 64GB RAM")
```
JonathanWoollett-Light, line 32: Would it make sense to add this like how we skip other tests, e.g.

```python
@pytest.mark.timeout(1800)
@pytest.mark.skipif(PLATFORM != "x86_64", reason="Kani proofs run only on x86_64.")
@pytest.mark.skipif(
    psutil.swap_memory().total + psutil.virtual_memory().total < 64 * 2**30,
    reason="kani requires at least 64GB of memory"
)
def test_kani(results_dir):
```
```

Comment (JonathanWoollett-Light) [inline tests/integration_tests/test_kani.py:32]: Since this includes swap space, it should be `memory` instead of `RAM`.
```suggestion
kani requires at least 64GB of memory
```

tests/integration_tests/test_kani.py
bchalios: I agree with Jonathan here. I prefer GiB when it comes to memory.

tests/integration_tests/test_kani.py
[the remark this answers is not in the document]
bchalios: Why do we even consider swap memory for this? Allowing a test to run knowing in advance it will consume swap space, it will most certainly hang the system :)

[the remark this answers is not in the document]
JonathanWoollett-Light: It would be a bit slower. Why would it hang?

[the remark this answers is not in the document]
bchalios: Because there are other things in the system that need memory. Swap space is used for swapping in/out physical pages based on the active memory set of all the processes in the system. If you occupy all of that from a single process, nothing else will be able to run.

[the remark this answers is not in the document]
JonathanWoollett-Light: Considering swap space allows people to run the test on cheaper systems (using an ssd for swap space should not affect performance too drastically).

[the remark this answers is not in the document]
JonathanWoollett-Light: > Because there are other things in the system that need memory. Swap space is used for swapping in/out physical pages based on the active memory set of all the processes in the system. If you occupy all of that from a single process, nothing else will be able to run.

True but other things also need RAM, we could add some buffer to consider this but I think this is overkill.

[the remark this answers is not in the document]
bchalios: what do you mean a very small amount of swap space? It highly depends on the system's setup. Swap is not supposed to be memory. Swap is supposed to provide a buffer for the OS to juggle pages around when it runs out of RAM. A good-citizen program should not depend on the presence or not of swap space.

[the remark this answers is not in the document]
JonathanWoollett-Light: It is not ideal to use swap space, but it is likely the only way most contributors can run the test. I think it is reasonable to support this.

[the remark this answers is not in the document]
bchalios: Right, I agree with that. However, this should come with a big fat message saying that we are going to eat up all the machine's memory and swap space.

If I was a user and did not know about this requirement, I wouldn't be extremely happy that I ran a test which made my machine hang. So, I think we should add a warning and documentation on the `tests/README.md`

bchalios (changes requested):

tests/integration_tests/test_kani.py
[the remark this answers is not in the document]
JonathanWoollett-Light: So currently the test makes peoples machines hang (mine specifically is what triggered this PR), this check should help minimize this by checking if a user's machine has enough memory. But as you mention it can still hang with this check so it would be worth also adding a warning and documentation.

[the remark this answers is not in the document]
roypat: Okay lets just go back to "skip this test unless CI" :sweat_smile:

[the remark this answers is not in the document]
bchalios: In that case, could we add instructions on how someone may run the proofs on their own?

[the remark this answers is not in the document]
roypat: Ah, yeah, true, we do have https://github.com/firecracker-microvm/firecracker/blob/main/docs/formal-verification.md#how-to-run-kani-harnesses, but the first sentence needs to be updated

tests/integration_tests/test_kani.py
```diff
@@ -15,6 +16,10 @@
 
 @pytest.mark.timeout(1800)
 @pytest.mark.skipif(PLATFORM != "x86_64", reason="Kani proofs run only on x86_64.")
+@pytest.mark.skipif(
+    os.environ.get("BUILDKITE") != "true",
+    reason="Kani's memory requirements cannot be satisfied locally",
```
JonathanWoollett-Light, line 21: Nit

```suggestion
    reason="Kani's memory requirements likely cannot be satisfied locally",
```

tests/integration_tests/test_kani.py
JonathanWoollett-Light: The check based on the CI has been moved to use `@pytest.mark.skipif` which covers this point.

JonathanWoollett-Light (dismissed):

pb8o (approved):

docs/formal-verification.md
```diff
@@ -45,17 +45,18 @@ properties, **all Kani harnesses are run on every pull request in our CI.** To
...
-harness needs to be specified via `--harness`. Note that the first invocation
+locally.  Note that the first invocation
 of Kani post-installation might take a while, due to it setting up some
 dependencies.
 
+Individual harnesses can then be executed using `cargo kani` similarly to how
+`cargo test` can run individual unit tests, the only difference being that the
+harness needs to be specified via `--harness`. Note, however, that many harnesses
+require significant memory, and might result in OOM conditions.
+
```
bchalios, line 59: Should we mention something around memory requirements?

tests/integration_tests/test_kani.py
```diff
@@ -24,10 +29,8 @@ def test_kani(results_dir):
     # -j enables kani harnesses to be verified in parallel (required to keep CI time low)
     # --output-format terse is required by -j
     # --enable-unstable is needed for each of the above
-    rc, stdout, stderr = utils.run_cmd(
+    _, stdout, _ = utils.run_cmd(
         "cargo kani --enable-unstable --enable-stubbing --restrict-vtable -j --output-format terse"
     )
 
-    assert rc == 0, stderr
-
```
bchalios, line 32: Why remove that?

bchalios (approved):
LGTM, mainly questions for my education

tests/integration_tests/test_kani.py
roypat: It is already checked by `run_cmd`: 
https://github.com/firecracker-microvm/firecracker/blob/532677328480b7f4149192ec121ac076451ac098/tests/framework/utils.py#L482-L487

docs/formal-verification.md
roypat: I feel like any information we add here will be highly inaccurate. The tiniest changes to a harness can have tremendous effects on whether it passes in a few seconds, or gobbles up all memory available in a .metal and crashes.

tests/integration_tests/test_kani.py
bchalios: oh awesome. Thanks.

docs/formal-verification.md
bchalios: Sure, I wasn't thinking about concrete numbers. Mainly, a heads up that running this `cargo kani` thing might crash your system :)

roypat: Ah, I thought that would be implied by "might result in OOM conditions." Do you think we should be more specific than that?

bchalios: no, I think I should have another coffee. Just ignore me.

roypat: :D No worries haha

PR #4138, opened 2023-09-27
Memfd backed memory

ShadowCurse:
## Changes
Refactoring of the memory code.
Added memfd backed guest memory.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/vmm/Cargo.toml
```diff
@@ -30,6 +30,7 @@ versionize_derive = "0.1.5"
 vm-allocator = "0.1.0"
 vm-fdt = "0.2.0"
 vm-superio = "0.7.0"
+vm-memory = { version = "0.12.0", features = ["backend-mmap", "backend-bitmap"] }
```
kalyazin, line 34: Please provide the motivation for the move in the commit message.

src/vmm/src/volotile.rs
```diff
@@ -0,0 +1,434 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kalyazin, line 1: 1. typo in the filename
2. we probably need to update the copyright to 2023

src/vmm/src/builder.rs
```diff
@@ -84,7 +84,7 @@ pub enum StartMicrovmError {
     CreateLegacyDevice(device_manager::legacy::LegacyDeviceError),
     /// Memory regions are overlapping or mmap fails.
     #[error("Invalid Memory Configuration: {}", format!("{:?}", .0).replace('\"', ""))]
-    GuestMemoryMmap(crate::vstate::memory::Error),
+    GuestMemoryMmap(crate::vstate::memory::VmMemoryError),
```
kalyazin: Although this is a common practice for us to move away from `Error`, it is worth mentioning in the commit message.

src/vmm/src/vstate/memory.rs
```diff
@@ -25,6 +32,174 @@ pub type GuestMmapRegion = vm_memory::MmapRegion<Option<AtomicBitmap>>;
 
 const GUARD_PAGE_COUNT: usize = 1;
 
+/// Errors associated with dumping guest memory to file.
```
kalyazin, line 36: Would it not be worth keeping it in a separate file? What do we gain from merging into one?

src/vmm/src/vstate/memory.rs
```diff
@@ -53,12 +53,39 @@ pub enum MemoryError {
...
+    fn from_state(
+        file: Option<&File>,
+        state: &GuestMemoryState,
+        track_dirty_pages: bool,
+    ) -> Result<Self, MemoryError>;
+
     /// Describes GuestMemoryMmap through a GuestMemoryState struct.
     fn describe(&self) -> GuestMemoryState;
+
+    /// Mark memery range as dirty
```
kalyazin, line 100: s/memery/memory/

src/vmm/src/vstate/memory.rs
```diff
@@ -53,12 +53,39 @@ pub enum MemoryError {
 }
 
 /// Defines the interface for snapshotting memory.
-pub trait SnapshotMemory
+pub trait GuestMemoryExtension
```
kalyazin, line 60: I don't quite like the name `Extension`. It looks like we are adding all the stuff we are missing from `vm-memory::GuestMemory` here (`with_size()`, `from_raw_regions()`, `from_raw_regions_unguarded()`). Would it make sense to upstream it? @roypat wdyt?
There is still a method `from_state()` that refers back to the original name `SnapshotMemory`. This looks like a separate trait that should stay in FC as rust-vmm's vm-memory doesn't have a clue about the state.

src/vmm/src/vstate/memory.rs
```diff
@@ -79,6 +79,13 @@ where
         track_dirty_pages: bool,
     ) -> Result<Self, MemoryError>;
 
+    /// Creates a GuestMemoryMmap from raw regions with guard pages.
+    fn from_raw_regions_file(
```
kalyazin, line 83: can make the 1st line of the commit message more descriptive by referring to adding this method to the trait.

src/vmm/src/vstate/memory.rs
```diff
@@ -543,67 +521,50 @@ mod tests {
...
+            std::ptr::write(addr, val);
+        };
+
         let page_size = get_page_size().unwrap();
 
         // Check that the created range allows us to write inside it
-        let addr = region.as_ptr();
+        let region_first_byte = region.as_ptr();
+        let region_last_byte = unsafe { region_first_byte.add(region.size() - 1) };
+        // let region_size = region.size();
```
kalyazin, line 506: is that a leftover?

src/vmm/src/vstate/memory.rs
```diff
@@ -543,67 +521,50 @@ mod tests {
...
+        fork_and_run(&|| _ = read_mem(right_border_first_byte), true);
     }
 
     #[test]
     fn test_build_guarded_region() {
         // Create anonymous guarded region.
         {
             let page_size = get_page_size().unwrap();
-            let size = page_size * 10;
+            let region_size = page_size * 10;
```
kalyazin, line 522: since you're at it, how about splitting this test into 2?

src/vmm/src/vstate/memory.rs
```diff
@@ -694,7 +649,105 @@ mod tests {
     }
 
     #[test]
-    fn test_mark_dirty_mem() {
+    fn test_from_raw_regions_unguarded() {
+        // Check dirty page tracking is off.
```
kalyazin, line 615: can be 2 tests

src/vmm/src/vstate/memory.rs
```diff
@@ -800,119 +853,139 @@ mod tests {
     }
 
     #[test]
-    fn test_restore_memory() {
-        let page_size: usize = get_page_size().unwrap();
+    fn test_dump() {
```
kalyazin, line 818: I would really love to see this test being decomposed to smaller tests.

src/vmm/src/vstate/memory.rs
```diff
@@ -543,67 +521,50 @@ mod tests {
     }
 
     fn validate_guard_region(region: &GuestMmapRegion) {
+        let read_mem = |addr| unsafe { std::ptr::read_volatile::<u8>(addr) };
```
kalyazin, line 498: commit message `refactor(memory): memory module testst refactoring` typo (testst)

src/vmm/src/vstate/memory.rs
```diff
@@ -50,13 +50,20 @@ pub enum MemoryError {
     MmapRegionError(MmapRegionError),
     /// Cannot create guest memory: {0}
     VmMemoryError(VmMemoryError),
+    /// Cannot create memfd: {0:?}
+    Memfd(memfd::Error),
+    /// Cannot resize memfd file: {0:?}
+    MemefdSetLen(std::io::Error),
```
kalyazin, line 56: typo s/MemefdSetLen/MemfdSetLen/

tests/integration_tests/functional/test_api.py
```diff
@@ -389,8 +389,7 @@ def test_api_machine_config(test_microvm_with_api):
     test_microvm.api.machine_config.patch(mem_size_mib=bad_size)
 
     fail_msg = re.escape(
-        "Invalid Memory Configuration: MmapRegion(Mmap(Os { code: "
-        "12, kind: OutOfMemory, message: Out of memory }))"
+        "Invalid Memory Configuration: MemefdSetLen(Custom { kind: InvalidInput, error: TryFromIntError(()) })"
```
kalyazin, line 392: same typo as before s/Memefd/Memfd.
I actually think this change may go to the same commit where memfd was added, but no strong preference.

src/vmm/src/vstate/memory.rs
```diff
@@ -540,16 +512,6 @@ mod tests {
         // Write and read from the end of the region
         write_mem(region_last_byte, 0x69);
         assert_eq!(read_mem(region_last_byte), 0x69);
-
-        // Try a read/write operation against the left guard border of the range
```
kalyazin: Is that a temporary measure? Do we plan to replace these tests with something else?

src/vmm/src/vstate/memory.rs
roypat: Mh, I think the decision around these was that vm-memory has no interest in taking on firecracker's concept of "guard pages".

src/vmm/src/volotile.rs
ShadowCurse: fixed

src/vmm/src/builder.rs
ShadowCurse: fixed

src/vmm/src/vstate/memory.rs
ShadowCurse: I think keeping all memory related code in 1 file is better, considering there is not that much of it.

src/vmm/src/vstate/memory.rs
ShadowCurse: I also don't think we should upstream our FC specific memory creation process.
Regarding the `GuestMemoryExtension` name, the original idea was that this trait just 'extends' what `GuestMemoryMmap` already has. But I don't mind changing it to smth better sounding. Any ideas?

src/vmm/src/vstate/memory.rs
ShadowCurse: Added

src/vmm/src/vstate/memory.rs
ShadowCurse: yes, removed

src/vmm/src/vstate/memory.rs
ShadowCurse: done

src/vmm/src/vstate/memory.rs
ShadowCurse: I don't think splitting this is worth it. This test is small by itself, and it is almost same as `test_from_raw_regions` test.

src/vmm/src/vstate/memory.rs
ShadowCurse: This is already spit up. `test_dump` and `test_dump_dirty` were one test before.

src/vmm/src/vstate/memory.rs
ShadowCurse: fixed

src/vmm/src/vstate/memory.rs
ShadowCurse: fixed

tests/integration_tests/functional/test_api.py
ShadowCurse: fixed

src/vmm/src/vstate/memory.rs
ShadowCurse: I was trying to come up with other way to test guard pages, but there are not a lot of options here.
There is an easy way to do it with threads: basically make another thread panic when it receives SIGSEGV and then main thread can catch that panic and verify the cause. But this approach does not work with `musl` for some reason. (works with gnu)

src/vmm/src/vstate/memory.rs
kalyazin: What about the following:
- It looks like the purpose of guarding the guest memory has been lost over time. I think we should either document why we are doing that or remove it (and we won't need to think how to test it)
- I agree, the creation thing may be opinionated to FC (so don't upstream it), but then let's create a trait only for those methods with a meaningful name (`factory` or something like)
- We have a family of `from_raw_regions_*` methods which are either used internally by the implementation or in testing. Is it possible to remove the helpers from the trait as standalone private functions and extract the rest into a trait that is used specifically for testing? Also, does `raw` part of the name add any meaning? Can we drop it?
- There is a family of snapshot-related methods (`from_state`, `describe`, `make_dirty`, `dump`, `dump_dirty`). Let's extract them into a trait that is specific to snapshot. Plus maybe also include `get_dirty_bitmap`, which seems to partially belong here as well (I know it might not be worth it).

I hope thus we'll have a bit more structure than a toolbox trait.
Would that make sense?

ShadowCurse: > It looks like the purpose of guarding the guest memory has been lost over time. I think we should either document why we are doing that or remove it (and we won't need to think how to test it)

I will look into it

> I agree, the creation thing may be opinionated to FC (so don't upstream it), but then let's create a trait only for those methods with a meaningful name (factory or something like)

Please, no factories. We are not a manufacturing site. 

> We have a family of from_raw_regions_* methods which are either used internally by the implementation or in testing. Is it possible to remove the helpers from the trait as standalone private functions and extract the rest into a trait that is used specifically for testing? Also, does raw part of the name add any meaning? Can we drop it?
There is a family of snapshot-related methods (from_state, describe, make_dirty, dump, dump_dirty). Let's extract them into a trait that is specific to snapshot. Plus maybe also include get_dirty_bitmap, which seems to partially belong here as well (I know it might not be worth it).

The idea here is to have them in one trait for ease of use. It does not really matter if they are public or not, everything is used internally only anyway. The `raw` part is just to differentiate from `from_regions` method that `GuestMemoryMmap` already has.
The only alternative I see is to just remove trait and have a bunch of free standing functions. Personally I don't like this approach, but it is better than having multiple traits.

bchalios (commented):
Commit comment: Could you please format the commit messages? They include random new lines:

For example:

```
refactor(memory): moved vm_memory from utils

Moved `utils/vm_memory` crate into
`vmm/vstate/memory`.
Removed `utils/vm_memory`.

This module contains `vmm` specific code
and it make sense to have it in `vmm`.

Signed-off-by: Egor Lazarchuk <yegorlz@amazon.co.uk>
```

should read:

```
refactor(memory): moved vm_memory from utils

Moved `utils/vm_memory` crate into `vmm/vstate/memory`. Removed
`utils/vm_memory`.

This module contains `vmm` specific codeand it make sense to have it in
`vmm`.

Signed-off-by: Egor Lazarchuk <yegorlz@amazon.co.uk>
```

bchalios (commented):
Why move `volatile` types to a separate module not under `memory`?

bchalios (commented):
Ok, it generally looks ok to me functionality wise. I mainly have some style comments
and questions for me to understand better the changes. Also, some per commit comments here: 

### Commit ` refactor(memory): minor refactor of memory `

I think this can just be merged with the commit that introduces the `GuestMemoryExtension` trait.

### Commit ` refactor(memory): memory module tests refactoring `

Could you please add commit message that is more descriptive than "Updated tests"?
For example, how tests were updated, and why and what tests were added.

### Commit ` fix(tests): fixing test_api test`

It can be merged with the commit that introduced the change in the commit message.

bchalios:
> I moved volatile into a separate module, because it is not really memory related thing

`volatile` is actually tightly coupled with the memory thing.

> But considering we will remove volatile after updating vm_memory to new version.

What's blocking us from doing it already?

src/vmm/src/builder.rs
```diff
@@ -265,9 +265,10 @@ pub fn build_microvm_for_boot(
         .ok_or(MissingKernelConfig)?;
 
     let track_dirty_pages = vm_resources.track_dirty_pages();
-    let guest_memory =
-        GuestMemoryMmap::with_size(vm_resources.vm_config.mem_size_mib, track_dirty_pages)
-            .map_err(StartMicrovmError::GuestMemory)?;
+    let memfd = crate::vstate::memory::create_memfd(vm_resources.vm_config.mem_size_mib)
+        .map_err(StartMicrovmError::GuestMemory)?;
+    let guest_memory = GuestMemoryMmap::with_file(memfd.as_file(), track_dirty_pages)
+        .map_err(StartMicrovmError::GuestMemory)?;
```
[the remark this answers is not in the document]
ShadowCurse: It is easier to use memfd in all cases. No need to track if there are vhost-user devices attached. From functional point of view, memfd is same anonymous memory as before.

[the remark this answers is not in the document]
kalyazin: If we boot without vhost-user devices, it does not matter how we back the guest memory (memfd vs anon, private vs shared) as we never share it. Whatever is simpler from the implementation point of view.

tests/integration_tests/security/test_jail.py
```diff
@@ -514,42 +509,6 @@ def test_args_resource_limits(test_microvm_with_api):
     check_limits(pid, NOFILE, FSIZE)
 
 
-def test_negative_file_size_limit(uvm_plain):
```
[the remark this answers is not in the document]
ShadowCurse: There couple of issues with this test. First, there is no way to test `negative` case now. If VM uses 128MB of memory,
than limit needs to be 128MB. And when we create a snapshot it will create 128MB mem file. So there is no way for us to test this with FC. Second, what is the reason for this test overall? If FC somehow is compromized and asked to 
create a bunch of files, what prevents it to create a lot of small files and never hit the file size limit?

src/vmm/src/vstate/memory.rs
```diff
@@ -495,40 +495,12 @@ mod tests {
 
     use super::*;
 
-    fn fork_and_run(function: &dyn Fn(), expect_sigsegv: bool) {
```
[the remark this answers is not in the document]
ShadowCurse: From what I can understand, forking test harness when it runs all vmm crate tests causes test to hang. Test passes fine if it is run separately or in a small group of tests (like only tests in `memory.rs`) There is another way to test guard pages with treads (by causing separate thread to segfault and panic and than main thread can catch the panic). But this approach does not work with musl target.

src/vmm/src/builder.rs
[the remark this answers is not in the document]
bchalios: Thanks for the clarification

tests/integration_tests/security/test_jail.py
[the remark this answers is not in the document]
bchalios: Oh ok. I thought the rlimit was referring to the total amount of file sizes, not the size of each file individually.


> If FC somehow is compromized and asked to create a bunch of files, what prevents it to create a lot of small files and never hit the file size limit?

There is also a limit in the maximum amount of open file descriptors, so that covers that.

So, I think it's worth to keep the test and set the limit to something smaller than then memory of the guest. This should fail.

src/vmm/src/vstate/memory.rs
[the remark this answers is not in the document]
bchalios: Ok, I think I understand. Maybe this could be a good target for kani. WDYT @roypat?

tests/integration_tests/security/test_jail.py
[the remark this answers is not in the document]
ShadowCurse: Ok, redone the commit. Now test just tries to start VM to verify the limit.

src/vmm/src/vstate/memory.rs
```diff
@@ -119,15 +126,45 @@ pub struct GuestMemoryState {
 }
 
 impl GuestMemoryExtension for GuestMemoryMmap {
-    /// Creates a GuestMemoryMmap with `size` in MiB and guard pages.
+    /// Creates a GuestMemoryMmap with `size` in MiB and guard pages backed by memfd.
```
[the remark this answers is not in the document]
ShadowCurse: fixed

ShadowCurse:
> Commit refactor(memory): minor refactor of memory
I think this can just be merged with the commit that introduces the GuestMemoryExtension trait.

This refactoring only makes sense after introduction of memfd.

> Commit refactor(memory): memory module tests refactoring
Could you please add commit message that is more descriptive than "Updated tests"?
For example, how tests were updated, and why and what tests were added.

Updated the commit message.

> Commit  fix(tests): fixing test_api test
It can be merged with the commit that introduced the change in the commit message.

Merged

Regarding `volatile`, we will be able to remove it after updating rust-vmm deps, but this is out of scope here, plus from what I know, rust-vmm deps are not ready yet.

src/vmm/src/vstate/memory.rs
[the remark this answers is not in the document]
roypat: Mhh, I'm not sure. On the one hand, having kani harnesses for different parts of the firecracker codebase that interact with guest memory will ensure that the guard pages won't be triggered (as the proofs will include that no out of bound guest memory access will happen), but it doesnt exactly show that they work as intended (it only shows that they're not _needed_). But I can't directly see how kani could reason about that, because it doesn't really interact with signals and such.

On the other hand, we could make the test be gnu-only. It might not be out release target, but at least we'd be testing the functionality at all

roypat:
> > But considering we will remove volatile after updating vm_memory to new version.
> 
> What's blocking us from doing it already?

I opened #6763 for it, but its running into kani timeouts :unamused:

tests/integration_tests/security/test_jail.py
```diff
@@ -516,23 +516,18 @@ def test_args_resource_limits(test_microvm_with_api):
 
 def test_negative_file_size_limit(uvm_plain):
     """
-    Test creating snapshot file fails when size exceeds `fsize` limit.
+    Test creating vm fails when memory size exceeds `fsize` limit.
+    This is caused by the fact that we back guest memory by memfd.
     """
     test_microvm = uvm_plain
     # limit to 1MB, to account for logs and metrics
```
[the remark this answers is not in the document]
ShadowCurse: fixed

tests/integration_tests/security/test_jail.py
```diff
@@ -516,23 +516,18 @@ def test_args_resource_limits(test_microvm_with_api):
 
 def test_negative_file_size_limit(uvm_plain):
     """
-    Test creating snapshot file fails when size exceeds `fsize` limit.
+    Test creating vm fails when memory size exceeds `fsize` limit.
+    This is caused by the fact that we back guest memory by memfd.
```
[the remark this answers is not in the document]
ShadowCurse: updated test to bind vm mem size to the jailer limit + added new one to verify vm starts if limit is not hit

tests/integration_tests/security/test_jail.py
```diff
@@ -514,25 +514,41 @@ def test_args_resource_limits(test_microvm_with_api):
     check_limits(pid, NOFILE, FSIZE)
 
 
-def test_negative_file_size_limit(uvm_plain):
+def test_positive_file_size_limit(uvm_plain):
     """
-    Test creating snapshot file fails when size exceeds `fsize` limit.
+    Test creating vm succeeds when memory size is under `fsize` limit.
+    This is caused by the fact that we back guest memory by memfd.
```
kalyazin, line 520: looks like an out-of-date message in this test now

src/vmm/src/vstate/memory.rs
ShadowCurse: I don't think we can differentiate tests base on the gnu/musl compilation.

ShadowCurse: Actually found that `target_env` can be one of 
```
    ""
    "gnu"
    "msvc"
    "musl"
    "sgx"
```
But than we will run unit tests twice. (It can actually be a good thing)

tests/integration_tests/security/test_jail.py
ShadowCurse: removed

src/vmm/src/vstate/memory.rs
ShadowCurse: Ok, I have returned the guard pages tests and marked them as gnu only tests. I used the panicing thread approach to make it work.

bchalios (approved):

kalyazin (approved):

PR #4144, opened 2023-10-02
fix: add 640th KiB to the available memory

### LINKED ISSUE #3044: Memory map excludes 640th kB
# Describe the bug

The memory map provided by Firecracker is

```
[    0.000000] BIOS-e820: [mem 0x0000000000000000-0x000000000009fbff] usable
[    0.000000] BIOS-e820: [mem 0x0000000000100000-0x000000003fffffff] usable
```

In other words, the first 639 kB of address space exists, and then addresses from 1 MB upwards (in this case up to 1 GB) exist.

The MP table is placed at address 0x9fc00, i.e.. in the last kB of the "normal" 640 kB of base memory -- but according to the memory map, this doesn't exist on the system!

This is harmless for Linux, because Linux's MP table parsing has a bug -- it assumes that there's always 640 kB of base memory.  But it breaks FreeBSD, since FreeBSD looks at the memory map when following the Intel MP spec of "look in the *last kB of base memory*".

## To Reproduce

1. Start Firecracker
2. Look at the memory map printed by the guest kernel

## Expected behaviour

The memory map presented by Firecracker should go up to 640 kB, not 639 kB.

## Environment

Latest Firecracker version checked out from main.  Running on Ubuntu 22.04; guest kernel and rootfs from the "getting started" instructions.

ShadowCurse:
## Changes
Fixes #3044

As described in the issue FC was only announcing 639 KiB as the first memory region available to the guest.
But at the same time FC is placing MP tables in the 640th KiB of the memory.

This PR fixes it by adding an additional entry from `EBDA_START` to `EBDA_START` + 1 KiB to represent mp tables.

Old memory map:
```
[    0.000000] BIOS-e820: [mem 0x0000000000000000-0x000000000009fbff] usable
[    0.000000] BIOS-e820: [mem 0x0000000000100000-0x000000003fffffff] usable
```

New memory map:
```
[    0.000000] BIOS-e820: [mem 0x0000000000000000-0x000000000009fbff] usable
[    0.000000] BIOS-e820: [mem 0x000000000009fc00-0x000000000009ffff] reserved
[    0.000000] BIOS-e820: [mem 0x0000000000100000-0x0000000007ffffff] usable
```

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/arch/x86_64/mod.rs
+++ b/src/vmm/src/arch/x86_64/mod.rs
@@ -26,7 +26,10 @@ use utils::vm_memory::{Address, GuestAddress, GuestMemory, GuestMemoryMmap, Gues
 use crate::arch::InitrdConfig;
 
 // Value taken from https://elixir.bootlin.com/linux/v5.10.68/source/arch/x86/include/uapi/asm/e820.h#L31
+// Usable normal RAM
 const E820_RAM: u32 = 1;
+// Reserved area that should be avoided during memory allocations
+const E820_RESERVED: u32 = 2;
 
 /// Errors thrown while configuring x86_64 system.
 #[derive(Debug, PartialEq, Eq, derive_more::From)]
@@ -41,8 +44,12 @@ pub enum ConfigurationError {
     InitrdAddress,
 }
 
-// Where BIOS/VGA magic would live on a real PC.
+// EBDA is located in the last 1 KiB of the first 640KiB of memory, i.e in the range:
+// [0x9FC00, 0x9FFFF]
+// We mark first [0x0, EBDA_START] region as usable RAM
+// and [EBDA_START, (EBDA_START + EBDA_SIZE)] as reserved.
 const EBDA_START: u64 = 0x9fc00;
+const EBDA_SIZE: u64 = 1 << 10;
 const FIRST_ADDR_PAST_32BITS: u64 = 1 << 32;
 /// Size of MMIO gap at top of 32-bit address space.
 pub const MEM_32BIT_GAP_SIZE: u64 = 768 << 20;
@@ -135,6 +142,7 @@ pub fn configure_system(
     }
 
     add_e820_entry(&mut params, 0, EBDA_START, E820_RAM)?;
+    add_e820_entry(&mut params, EBDA_START, EBDA_SIZE, E820_RESERVED)?;
 
     let last_addr = guest_mem.last_addr();
     if last_addr < end_32bit_gap_start {


src/vmm/src/arch/x86_64/mod.rs
```diff
@@ -134,7 +136,7 @@ pub fn configure_system(
         params.hdr.ramdisk_size = initrd_config.size as u32;
     }
 
-    add_e820_entry(&mut params, 0, EBDA_START, E820_RAM)?;
+    add_e820_entry(&mut params, 0, EBDA_END, E820_RAM)?;
```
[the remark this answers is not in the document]
ShadowCurse: I did a test where I added another entry like this:
```rust 
add_e820_entry(&mut params, 0, EBDA_START, E820_RAM)?;
add_e820_entry(&mut params, EBDA_START, EBDA_END, E820_RAM)?;
```
And the result was same as adding 1 entry. (in kernel logs it was reporting as 1 range from 0 to EBDA_END)
I don't think there is a need for 2 entries when 1 covers the correct range anyway.

src/vmm/src/arch/x86_64/mod.rs
```diff
@@ -41,8 +42,11 @@ pub enum ConfigurationError {
     InitrdAddress,
 }
 
-// Where BIOS/VGA magic would live on a real PC.
+// EBDA is located in the last 1 KiB from 0x9FC00 - 0x9FFFF
```
bchalios, line 45:
```suggestion
// EBDA is located in the last 1 KiB of the first 640KiB of memory, i.e in the range: [0x9FC00, 0x9FFFF]
```

ShadowCurse: added

kalyazin:
In the map comparison (PR description), why are the last entries different (`0x000000003fffffff` vs `0x0000000007ffffff`)? Is this expected?

bchalios (approved):

bchalios:
> In the map comparison (PR description), why are the last entries different (`0x000000003fffffff` vs `0x0000000007ffffff`)? Is this expected?

Oh, good catch. I totally missed that. I guess the numbers were taken from microVMs with different memory. We should amend that.

ShadowCurse:
Yes, I was using uVM with default 128MB of RAM, while issue was with uVM with 1GB.

ShadowCurse:
> Not sure I get it. I think the issue should exist as well in a microVM with 128MB of memory. The memory map difference should be similar. Am I missing something?

Yes, the comment about different RAM sizes was about Nikita's comment regarding last entries being different.

PR #4145, opened 2023-10-02
feat: Add per net device metrics

sudanl0:
## Changes

- Move NetDeviceMetrics from logger to Net module.
- Use NET_DEV_METRICS_PVT instead of adding an entry of NetDeviceMetrics in Net so that metrics can be flushed even from signal handlers.
- use individual metrics instead of METRICS.net.
- Use "net_$iface_id" for the metrics name instead of "net_$tap_name" to be consistent with the net endpoint "/network-interfaces/{iface_id}".

## Reason

- To add more granularity in analysing network device metrics.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@
   capability checks that Firecracker performs during boot. If any of
   these fields are in use, minimal target snapshot version is
   restricted to 1.5.
+- [#4145](https://github.com/firecracker-microvm/firecracker/pull/4145):
+  Added support for per net device metrics. Now, along with aggregated
+  network device metrics, metrics per network device with also be emitted.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,10 @@
 ### Added
 
 - [#4145](https://github.com/firecracker-microvm/firecracker/pull/4145):
-  Added support for per net device metrics. Now, along with aggregated
-  network device metrics, metrics per network device with also be emitted.
-  Aggregate metrics will continue to be emitted with label "net" in the
-  metrics json object while each net device metrics will be emitted with
-  label "net_{iface_id}". E.g. the associated metrics for the endpoint
-  "/network-interfaces/eth0" will be available under "net_eth0:" in each
-  metrics json object.
+  Added support for per net device metrics. In addition to aggregate metrics `net`,
+  each individual net device will emit metrics under the label `"net_{iface_id}"`.
+  E.g. the associated metrics for the endpoint `"/network-interfaces/eth0"` will
+  be available under `"net_eth0"` in the metrics json object.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@
 
 ### Added
 
+- [#4145](https://github.com/firecracker-microvm/firecracker/pull/4145):
+  Added support for per net device metrics. Now, along with aggregated
+  network device metrics, metrics per network device with also be emitted.
+  Aggregate metrics will continue to be emitted with label "net" in the
+  metrics json object while each net device metrics will be emitted with
+  label "net_{iface_id}". E.g. the associated metrics for the endpoint
+  "/network-interfaces/eth0" will be available under "net_eth0:" in each
+  metrics json object.
+
 ### Changed
 
 - Simplified and clarified the removal policy of deprecated API elements
@@ -34,9 +43,6 @@
   capability checks that Firecracker performs during boot. If any of
   these fields are in use, minimal target snapshot version is
   restricted to 1.5.
-- [#4145](https://github.com/firecracker-microvm/firecracker/pull/4145):
-  Added support for per net device metrics. Now, along with aggregated
-  network device metrics, metrics per network device with also be emitted.
 
 ### Changed
 
--- a/src/vmm/src/devices/mod.rs
+++ b/src/vmm/src/devices/mod.rs
@@ -17,14 +17,15 @@ pub mod virtio;
 pub use bus::{Bus, BusDevice, BusError};
 use log::error;
 
+use crate::devices::virtio::metrics::NetDeviceMetricsIndex;
 use crate::devices::virtio::{QueueError, VsockError};
 use crate::logger::{IncMetric, METRICS};
 
 // Function used for reporting error in terms of logging
-// but also in terms of METRICS net event fails.
-pub(crate) fn report_net_event_fail(err: DeviceError) {
+// but also in terms of metrics of net event fails.
+pub(crate) fn report_net_event_fail(net_metrics: &NetDeviceMetricsIndex, err: DeviceError) {
     error!("{:?}", err);
-    METRICS.net.event_fails.inc();
+    net_metrics.get().event_fails.inc();
 }
 
 pub(crate) fn report_balloon_event_fail(err: virtio::balloon::BalloonError) {
--- a/src/vmm/src/devices/mod.rs
+++ b/src/vmm/src/devices/mod.rs
@@ -17,17 +17,17 @@ pub mod virtio;
 pub use bus::{Bus, BusDevice, BusError};
 use log::error;
 
+use crate::devices::virtio::metrics::NetDeviceMetrics;
 use crate::devices::virtio::{QueueError, VsockError};
 use crate::logger::{IncMetric, METRICS};
-use crate::net_metrics;
 
 // Function used for reporting error in terms of logging
 // but also in terms of metrics of net event fails.
-// network metrics is reported per device so we need `net_iface_id`
-// to identify which device the metrics and `err` should be reported for.
-pub(crate) fn report_net_event_fail(net_iface_id: &String, err: DeviceError) {
-    error!("{:?}:{:?}", net_iface_id, err);
-    net_metrics!(net_iface_id, event_fails.inc());
+// network metrics is reported per device so we need a handle to each net device's
+// metrics `net_iface_metrics` to report metrics for that device.
+pub(crate) fn report_net_event_fail(net_iface_metrics: &NetDeviceMetrics, err: DeviceError) {
+    error!("{:?}", err);
+    net_iface_metrics.event_fails.inc();
 }
 
 pub(crate) fn report_balloon_event_fail(err: virtio::balloon::BalloonError) {
--- a/src/vmm/src/devices/mod.rs
+++ b/src/vmm/src/devices/mod.rs
@@ -17,15 +17,20 @@ pub mod virtio;
 pub use bus::{Bus, BusDevice, BusError};
 use log::error;
 
-use crate::devices::virtio::metrics::NetDeviceMetricsIndex;
+// get_net_metrics and NetMetricsPerDevice are the functions called by the
+// macro NET_METRICS
+use crate::devices::virtio::metrics::{get_net_metrics, NetMetricsPerDevice};
 use crate::devices::virtio::{QueueError, VsockError};
 use crate::logger::{IncMetric, METRICS};
+use crate::NET_METRICS;
 
 // Function used for reporting error in terms of logging
 // but also in terms of metrics of net event fails.
-pub(crate) fn report_net_event_fail(net_metrics: &NetDeviceMetricsIndex, err: DeviceError) {
-    error!("{:?}", err);
-    net_metrics.get().write().unwrap().event_fails.inc();
+// network metrics is reported per device so we need `net_iface_id`
+// to identify which device the metrics and `err` should be reported for.
+pub(crate) fn report_net_event_fail(net_iface_id: &String, err: DeviceError) {
+    error!("{:?}:{:?}", net_iface_id, err);
+    NET_METRICS!(net_iface_id, event_fails.inc());
 }
 
 pub(crate) fn report_balloon_event_fail(err: virtio::balloon::BalloonError) {
--- a/src/vmm/src/devices/mod.rs
+++ b/src/vmm/src/devices/mod.rs
@@ -25,7 +25,7 @@ use crate::logger::{IncMetric, METRICS};
 // but also in terms of metrics of net event fails.
 pub(crate) fn report_net_event_fail(net_metrics: &NetDeviceMetricsIndex, err: DeviceError) {
     error!("{:?}", err);
-    net_metrics.get().event_fails.inc();
+    net_metrics.get().write().unwrap().event_fails.inc();
 }
 
 pub(crate) fn report_balloon_event_fail(err: virtio::balloon::BalloonError) {
--- a/src/vmm/src/devices/mod.rs
+++ b/src/vmm/src/devices/mod.rs
@@ -17,12 +17,9 @@ pub mod virtio;
 pub use bus::{Bus, BusDevice, BusError};
 use log::error;
 
-// get_net_metrics and NetMetricsPerDevice are the functions called by the
-// macro NET_METRICS
-use crate::devices::virtio::metrics::{get_net_metrics, NetMetricsPerDevice};
 use crate::devices::virtio::{QueueError, VsockError};
 use crate::logger::{IncMetric, METRICS};
-use crate::NET_METRICS;
+use crate::net_metrics;
 
 // Function used for reporting error in terms of logging
 // but also in terms of metrics of net event fails.
@@ -30,7 +27,7 @@ use crate::NET_METRICS;
 // to identify which device the metrics and `err` should be reported for.
 pub(crate) fn report_net_event_fail(net_iface_id: &String, err: DeviceError) {
     error!("{:?}:{:?}", net_iface_id, err);
-    NET_METRICS!(net_iface_id, event_fails.inc());
+    net_metrics!(net_iface_id, event_fails.inc());
 }
 
 pub(crate) fn report_balloon_event_fail(err: virtio::balloon::BalloonError) {
--- a/src/vmm/src/devices/virtio/net/metrics.rs
+++ b/src/vmm/src/devices/virtio/net/metrics.rs
@@ -1,0 +1,285 @@
+// Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Defines the metrics system for Network devices.
+//!
+//! # Metrics format
+//! The metrics are flushed in JSON when requested by vmm::logger::metrics::METRICS.write().
+//!
+//! ## JSON example with metrics:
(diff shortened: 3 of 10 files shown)


src/vmm/src/devices/virtio/net/metrics.rs
```diff
@@ -0,0 +1,308 @@
...
+            NET_DEV_METRICS_PVT.metrics.push(NetDeviceMetrics::new());
+            NetDeviceMetricsIndex(NET_DEV_METRICS_PVT.metrics.len() - 1)
+        }
+    }
+}
+
+/// Contains Network-related metrics per device.
+static mut NET_DEV_METRICS_PVT: NetDeviceMetricsPool = NetDeviceMetricsPool {
+    metrics: Vec::new(),
+};
```
roypat, line 113: I'm pretty sure this needs some sort of synchronization primitive (I would guess a `RwLock`, as writes are rare and only happen during VMM initialization, and then all the reads can essentially happen without any performance impact). Without this, concurrent reads and writes will cause undefined behavior (I havent looked at the code much, I think this can happen if Firecracker received a signal while initialization a new net device, for example. The `Vec` here will be extended without any locking, while the signal handler tries to iterate over it to dump the metrics, which will almost certainly result in a segfault). I think the safest way here is 

```
use lazy_static::lazy_static;   // we already have this dependency anyway
use std::sync::RwLock;

lazy_static! {
    /// This is an example for using doc comment attributes
    static ref METRICS: RwLock<Vec<u8>> = RwLock::new(Vec::new());
}
```
and then update all the unsafe code to instead call the correct functions on `RwLock` instead.

Although maybe I would go one step further here, because with this implementation we do not solve the unittest race condition, which is something I was really hoping we'd solve. This is just a sketch, but can we make `NET_DEV_METRICS_PVT` a `OnceCell<Vec<Arc<NetDeviceMetrics>>>` ( `NetDeviceMetrics` is `Sync` and `Send`, right? Its only atomics, so dont see why it shouldnt be). We initialize the `OnceCell` after all net devices have been created (say on `InstanceStart` or something). Then the code in the net devices will not access the global variable at all, instead each net device gets a `Arc<NetDeviceMetrics>` field (e.g. we use an `Arc` to share a reference to the `NetDeviceMetrics` object between the global and the individual net device). The signal handler and other global code that needs to dump metrics can do so by accessing the global variable, while all the devices access their own field, and do not need to ever reach out to the global state. This "fixes" our unittests, as in unittests we would simply not set up the global state at all, and only use the field that's stored in the actual device (meaning they can run in parallel).

roypat (dismissed):

src/vmm/src/devices/virtio/net/metrics.rs
```diff
@@ -0,0 +1,308 @@
...
+/// provides instance for net metrics
+#[derive(Debug)]
+pub struct NetDeviceMetricsPool {
+    /// used to access per net device metrics
+    metrics: Vec<NetDeviceMetrics>,
+}
+
+impl NetDeviceMetricsPool {
+    /// default construction
+    pub fn get() -> NetDeviceMetricsIndex {
```
ShadowCurse, line 99: Maybe the name `new_index` or `allocate` is better than `get`.

sudanl0: done

src/vmm/src/devices/virtio/net/metrics.rs
sudanl0: added the RwLock which also helped in removing all the "unsafe". Thanks for the feedback!

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
...
 ### Added
 
+- [#4145](https://github.com/firecracker-microvm/firecracker/pull/4145):
+  Added support for per net device metrics. Now, along with aggregated
+  network device metrics, metrics per network device with also be emitted.
+  Aggregate metrics will continue to be emitted with label "net" in the
+  metrics json object while each net device metrics will be emitted with
+  label "net_{iface_id}". E.g. the associated metrics for the endpoint
+  "/network-interfaces/eth0" will be available under "net_eth0:" in each
+  metrics json object.
```
roypat, line 14: This is out of scope here, but do we plan to do the same for the other devices? It'd be a bit awkward to have a release that has per-device metrics but only for a some of the devices, imo.

roypat, line 14: nit: This feels a bit verbose to me, I think we can cut out a sentence or two in the middle, e.g. something like

```suggestion
- [#4145](https://github.com/firecracker-microvm/firecracker/pull/4145):
  Added support for per net device metrics. In addition to aggregate metrics,
  each individual net device will emit metrics under the label "net_{iface_id}". 
  E.g. the associated metrics for the endpoint "/network-interfaces/eth0" will
  be available under "net_eth0" in the metrics json object.
```

src/vmm/src/logger/metrics.rs
```diff
@@ -1146,6 +1061,27 @@ impl Serialize for SerializeToUtcTimestampMs {
...
+// By using the below structure in FirecrackerMetrics it is easy
+// to serialise Firecracker app_metrics as a signle json object which
+// otherwise would have required extra string manipulation to pack
+// net as part of the same json object as FirecrackerMetrics.
+pub struct NetMetricsSerializer {}
+impl NetMetricsSerializer {
+    pub const fn new() -> Self {
+        Self {}
+    }
+}
```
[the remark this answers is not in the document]
sudanl0: do you mean we use Default instead?

src/vmm/src/logger/metrics.rs
```diff
@@ -1146,6 +1061,27 @@ impl Serialize for SerializeToUtcTimestampMs {
...
+
+impl Serialize for NetMetricsSerializer {
```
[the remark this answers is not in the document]
sudanl0: sure, NetMetricsProxy sounds better.

src/vmm/src/devices/virtio/net/metrics.rs
```diff
@@ -0,0 +1,300 @@
...
+            tx_fails: SharedIncMetric::new(),
+            tx_count: SharedIncMetric::new(),
+            tx_packets_count: SharedIncMetric::new(),
+            tx_partial_reads: SharedIncMetric::new(),
+            tx_queue_event_count: SharedIncMetric::new(),
+            tx_rate_limiter_event_count: SharedIncMetric::new(),
+            tx_rate_limiter_throttled: SharedIncMetric::new(),
+            tx_spoofed_mac_count: SharedIncMetric::new(),
+        }
+    }
```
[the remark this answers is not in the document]
sudanl0: Initially I used new() to be consistent with how other devices were initialised but then I think if default always provides a 0 initialised Atomicusize then we can just use that and avoid this.

src/vmm/src/devices/virtio/net/metrics.rs
```diff
@@ -0,0 +1,379 @@
...
+// Per device metrics are behind a ref lock and in a map which
+// makes it difficult to return the metrics pointer via a function
+// This macro is an attempt to make the code more readable and to
+// centralize the error handling.
+// The macro can be called in below ways where iface_id is String and
+// activate_fails is a metric from struct `NetDeviceMetrics`:
+//     NET_METRICS!(&iface_id, activate_fails.inc());
+//     NET_METRICS!(&iface_id, activate_fails.add(5));
+//     NET_METRICS!(&iface_id, activate_fails.count());
+macro_rules! NET_METRICS {
```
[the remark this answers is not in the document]
sudanl0: it looked similar to using METRICS.something so I kept it uppercase but if thats the convention I'll change it.

src/vmm/src/logger/metrics.rs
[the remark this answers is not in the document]
roypat: If you have a unit struct in rust (e.g. without fields) you can just initialize it by referring to its name, e.g. 

```rs
pub struct NetMetricsProxy;

fn main() {
    let proxy = NetMetricsProxy;
    proxy.serialize(...)
}
```

src/vmm/src/devices/virtio/net/metrics.rs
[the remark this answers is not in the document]
roypat: Yeah, all integer types in Rust have a default implementation that produces `0`. I think we can eliminate a bunch of explicit `new` functions this way actually, if we really don't need the const-ness :thinking: (although those are probably out of scope here)

src/vmm/src/devices/mod.rs
```diff
@@ -17,14 +17,20 @@ pub mod virtio;
 pub use bus::{Bus, BusDevice, BusError};
 use log::error;
 
+// get_net_metrics and NetMetricsPerDevice are the functions called by the
+// macro NET_METRICS
+use crate::devices::virtio::metrics::{get_net_metrics, NetMetricsPerDevice};
```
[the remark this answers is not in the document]
sudanl0: done.

src/vmm/src/logger/metrics.rs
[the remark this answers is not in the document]
sudanl0: I used NetMetricsSerializeProxy so that the name suggests this is to be used only as a proxy to serialize and nothing else should be added to this in the future.

src/vmm/src/logger/metrics.rs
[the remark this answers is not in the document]
sudanl0: Sorry but I didn't understand the comment.
If the suggestion is to remove the NetMetricsProxy::new() then, we need the new() because FirecrackerMetrics::new() needs a const initialization which Default doesn't seem to have.
If the suggestion is not to have the NetMetricsProxy at all in FirecrackerMetrics then, the reason why this was added is because it is easier this way to pack NetDeviceMetrics inside FirecrackerMetrics as a single json object.

src/vmm/src/devices/virtio/net/metrics.rs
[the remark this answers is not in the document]
sudanl0: ok so METRICS is static and so FirecrackerMetrics::new() had to be a const which made the rest of structs define a const new and so NetDeviceMetrics had a new() instead of default. But since we moved out of logger and don't use NetDeviceMetrics directly in a static initialization we can remove it and just use default. Once we are done with all the device related metrics we can revisit the const for FirecrackerMetrics as well.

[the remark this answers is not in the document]
roypat: What I meant is that you need neither `new` nor `default`, because in `FirecrackerMetrics` you can just use a struct initializer directly. Since its a unit struct, it can just be
```rs
impl FirecrackerMetrics {
    /// Const default construction.
    pub const fn new() -> Self {
        Self {
            ...
            net: NetDeviceMetrics {}, // or NetDeviceMetric,
            ...
        }
    }
}
```
definitely more of a nit though :)

src/vmm/src/logger/metrics.rs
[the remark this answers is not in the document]
roypat: What I meant is that you need neither new nor default, because in FirecrackerMetrics you can just use a struct initializer directly. Since its a unit struct, it can just be

impl FirecrackerMetrics {
    /// Const default construction.
    pub const fn new() -> Self {
        Self {
            ...
            net: NetDeviceMetrics {}, // or NetDeviceMetric,
            ...
        }
    }
}

definitely more of a nit though :)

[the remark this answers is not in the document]
sudanl0: oh, sorry totally forgot about that. fixed!

src/vmm/src/devices/virtio/net/metrics.rs
```diff
@@ -0,0 +1,349 @@
...
+macro_rules! net_metrics {
+    ($iface_id:expr,$metric:ident.$action:ident($($value:tt)?)) => {{
+        if $crate::devices::virtio::net::metrics::get_net_metrics().metrics.get($iface_id).is_some() {
+            $crate::devices::virtio::net::metrics::get_net_metrics().metrics.get($iface_id).unwrap().$metric.$action($($value)?)
+        }else{
+            $crate::devices::virtio::net::metrics::NetMetricsPerDevice::alloc($iface_id.to_string());
+            $crate::devices::virtio::net::metrics::get_net_metrics().metrics.get($iface_id).unwrap().$metric.$action($($value)?)
+        }
+    }}
+}
```
ShadowCurse, line 148: Can't this (and `get_net_metrics` from above) be replaced by:
```rust
fn get_net_metrics(id: &str) -> &NetDeviceMetrics {
  NET_DEV_METRICS_PVT.read().unwrap().get(id).unwrap()
}
```
Because there should never be a situation when we created a device, but did not create a metrics for it.

Or maybe just add `get_metrics` method to the `Net` itself:
```rust
fn get_metrics(&self) -> &NetDeviceMetrics {
  NET_DEV_METRICS_PVT.read().unwrap().get(self.id).unwrap()
}
```
Even though it needs `NET_DEV_METRICS_PVT` to  be public, it is not a big problem.

src/vmm/src/devices/virtio/net/metrics.rs
```diff
@@ -0,0 +1,349 @@
...
+
+/// Pool of Network-related metrics per device behind a lock to
+/// keep things thread safe. Since the lock is initialized here
+/// it is safe to unwrap it without any check.
+static NET_DEV_METRICS_PVT: RwLock<NetMetricsPerDevice> = RwLock::new(NetMetricsPerDevice {
+    metrics: BTreeMap::new(),
+});
```
ShadowCurse, line 155: nit: maybe rename `NetMetricsPerDevice` -> `NetMetrics`, `NET_DEV_METRICS_PVT` -> `GLOBAL_NET_METRICS` (because I have no idea what PVT means).

ShadowCurse (commented):
It seems the implementation jumps all over the place for this task, but I guess it is ok.

src/vmm/src/devices/virtio/net/metrics.rs
sudanl0: I am working on Patrick's suggestion of Arc which will remove the need for the macro net_metrics and get_net_metrics.

src/vmm/src/devices/virtio/net/metrics.rs
sudanl0: PVT stands for private :)

src/vmm/src/devices/virtio/net/metrics.rs
sudanl0: both get_metrics and the macro are removed.

src/vmm/src/devices/virtio/net/metrics.rs
sudanl0: typically _g or _global is added for global variables and _pvt or _private is added for private variables. Its used extensively in linux and c where adding visibility as suffix helps make the code more readable.
But assuming this is not something we do in rust I can remove the _pvt suffix.

sudanl0: But I prefer NetMetricsPerDevice to NetMetrics since it better describes that it is meant for per device metrics so it would have a map/vec in it.

src/vmm/src/devices/virtio/net/metrics.rs
```diff
@@ -0,0 +1,379 @@
...
+//     NET_METRICS!(&iface_id, activate_fails.count());
+macro_rules! NET_METRICS {
+    ($iface_id:expr,$metric:ident.$action:ident($($value:tt)?)) => {
+        if get_net_metrics().metrics.get($iface_id).is_some() {
+            get_net_metrics().metrics.get($iface_id).unwrap().$metric.$action($($value)?)
+        }else{
+            NetMetricsPerDevice::alloc($iface_id.to_string());
+            get_net_metrics().metrics.get($iface_id).unwrap().$metric.$action($($value)?)
+        }
+    }
```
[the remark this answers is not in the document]
sudanl0: this is done. Thx for the suggestion!

ShadowCurse (approved):

src/vmm/src/devices/virtio/net/metrics.rs
```diff
@@ -0,0 +1,435 @@
...
+                > 0,
+            "{}",
+            NET_METRICS
+                .read()
+                .unwrap()
+                .metrics
+                .get(devn)
+                .unwrap()
+                .activate_fails
+                .count()
```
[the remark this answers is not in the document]
sudanl0: in this particular case if the condition fails it didn't print the value of the metrics and so I added the extra print message.

src/vmm/src/devices/virtio/net/metrics.rs
```diff
@@ -0,0 +1,435 @@
...
+                .metrics
+                .insert(iface_id.clone(), Arc::new(NetDeviceMetrics::default()));
+        }
+        NET_METRICS
+            .read()
+            .unwrap()
+            .metrics
+            .get(&iface_id)
+            .unwrap()
+            .clone()
```
roypat, line 118: nit:
```suggestion
        Arc::clone(
            NET_METRICS
                .write()
                .unwrap()
                .metrics
                .entry(iface_id)
                .or_insert_with(|| Arc::new(NetDeviceMetrics::default())),
        )
```
avoids a clone.

roypat (approved):

CHANGELOG.md
```diff
@@ -34,6 +34,9 @@
   capability checks that Firecracker performs during boot. If any of
   these fields are in use, minimal target snapshot version is
   restricted to 1.5.
+- [#4145](https://github.com/firecracker-microvm/firecracker/pull/4145):
+  Added support for per net device metrics. Now, along with aggregated
```
kalyazin: nit: "net" and "network" are both used. Could we unify?

CHANGELOG.md
```diff
@@ -34,6 +34,9 @@
...
+  network device metrics, metrics per network device with also be emitted.
```
kalyazin: s/with/will/ ?

CHANGELOG.md
sudanl0: this was updated in the later commits.

CHANGELOG.md
sudanl0: fixed in later commits.

src/vmm/src/devices/virtio/net/metrics.rs
sudanl0: Since I have verified the PR with performance tests and changing this will mean I'll have to redo the tests again. In the interest of time to finalise this and since this is a nit, I would like to work on this later when I start adding per device metrics for other devices.
And thank you again for the feedback :)

kalyazin (approved):
changelog update looks good to me

PR #4164, opened 2023-10-10
Eliminate *_gen crates

roypat:
## Changes

Turns the `net_gen` and `virtio_gen` crates into modules in appropriate places in the `vmm::devices` module.

## Reason

Unifying our crate landscape.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -36,7 +36,7 @@ const FRAME_HEADER_MAX_LEN: usize = PAYLOAD_OFFSET + ETH_IPV4_FRAME_LEN;
 use crate::devices::virtio::iovec::IoVecBuffer;
 use crate::devices::virtio::net::tap::Tap;
 use crate::devices::virtio::net::{
-    NetError, NetQueue, MAX_BUFFER_SIZE, NET_QUEUE_SIZES, RX_INDEX, TX_INDEX,
+    gen as net_gen, NetError, NetQueue, MAX_BUFFER_SIZE, NET_QUEUE_SIZES, RX_INDEX, TX_INDEX,
```
JonathanWoollett-Light, line 39: Could the module just be named `net_gen`?

JonathanWoollett-Light: Ah, nevermind I understand. I think it would make sense to change the usages that use it as `net_gen` to using `gen`.

roypat: Ah, yeah, that's fair

bchalios: Does it make sense for us to follow typical conventions and call this `net-sys`?

roypat: Not sure, usually that's for when these are in a separate crate, while this PR moves in the opposite direction. I'd be open to rename the module to `sys` instead of `gen` if you feel strongly about it? I don't want to include the `net` prefix, because otherwise the import would `use crate::devices::virtio::net::net_sys`, e.g. with "net" twice.

bchalios: I was just thinking out loud. I really don't have any preference at all. I only mentioned it because this is the naming convention people typically use and I thought it might help to reduce mental cycles on understanding what's included in this module. Feel free to choose whatever you think makes more sense.

I agree to not use the `net` prefix in the module.

JonathanWoollett-Light (approved):

src/vmm/src/devices/virtio/net/device.rs
roypat: Let's just keep as is then :) that way it stays consistent with our other bindgen modules, and also with the convention used at rust-vmm

bchalios (approved):

PR #4166, opened 2023-10-11
Improve QoL for release process

roypat:
## Changes

Speeding up the running of tests, and automation of dependabot removal

## Reason

less friction

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tools/devtool
+++ b/tools/devtool
@@ -486,7 +486,7 @@ cmd_build() {
 }
 
 function cmd_make_release {
-    cmd_test -- --reruns 2 || die "Tests failed!"
+    cmd_test -- --reruns 2 -n 8 || die "Tests failed!"
 
     run_devctr \
         --user "$(id -u):$(id -g)" \
--- a/tools/update-credits.sh
+++ b/tools/update-credits.sh
@@ -31,5 +31,5 @@ written in Rust with a focus on safety and security. Thanks go to:
 Contributors to the Firecracker repository:
 EOH
     echo
-    git log --format='* %aN <%aE>' | LC_ALL=C.UTF-8 sort -uf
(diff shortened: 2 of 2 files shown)


tools/devtool
```diff
@@ -486,7 +486,7 @@ cmd_build() {
 }
 
 function cmd_make_release {
-    cmd_test -- --reruns 2 || die "Tests failed!"
+    cmd_test -- --reruns 2 -n 8 || die "Tests failed!"
```
pb8o, line 489: I really want to do this, but I think we need to be careful .Some of the tests cannot be run in parallel. Like the build ones. The performance ones may also fail since they assume they have the whole host to work with.

We could separate it in two runs.

```
    OPTS="--reruns=7"
    cmd_test -- $OPTS --json-report-file=test-report.json -n8 --dist=worksteal integration_tests/{functional,security,style} || die "Tests failed!"
    cmd_test -- $OPTS --json-report-file=test-report-perf.json integration_tests/{performance,build} || die "Tests failed!"
```

We end up with 2 files, so we would need to adapt the release script to account for the other file.

pb8o (changes requested):
See comment on running the build tests in parallel.

tools/devtool
zulinx86, line 489: When this parallelization is enabled, does the test work well?
I'm curious about why we haven't parallelize this so far.

roypat: Ahhh, I did not realize this also ran the build and performance test. Suffer more I shall then.

pb8o: Also if we do this we may be able to run the release sanity check without faking the testrun: https://github.com/firecracker-microvm/firecracker/blob/main/.buildkite/pipeline_pr.py#L52

PR #4175, opened 2023-10-13
chore: update to vm-memory 0.13.1

roypat:
As always, the linux-loader update has to biggy-back off of the vm-memory update. This allows us to remove the {Read,Write}Volatile traits, instead using the ones from upstream.

Also eliminates some more usages of now-deprecated vm-memory functions that I missed when I first wrote the volatile-io patch for firecracker.

Lastly, this PR also includes a lot of kani performance tweaking that became necessary with the new vm-memory update for some reason.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/vmm/src/devices/virtio/queue.rs
```diff
@@ -823,6 +823,14 @@ mod verification {
         }
```
pb8o, line 823: on the commit message: loose -> lose

src/utils/Cargo.toml
```diff
@@ -17,7 +17,7 @@ displaydoc = "0.2.4"
 versionize = "0.1.10"
```
pb8o, line 17: on the commit message: biggy-back -> piggy-back

pb8o (commented):
The code looks good to me. Just commented on a couple minor typos.

src/utils/Cargo.toml
roypat: Fixed

src/vmm/src/devices/virtio/queue.rs
roypat: Fixed

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -823,6 +823,14 @@ mod verification {
...
     }
 
+    mod stubs {
+        use super::*;
+
+        fn set_avail_event<M: GuestMemory>(_self: &mut Queue, _val: u16, _mem: &M) {
+            // do nothing
+        }
```
bchalios, line 831: Could you explain a bit better why this does not make us loose generality? IIUC:

`try_enable_notification` uses `set_avail_event` to set the index of the next available buffer to `Queue::next_avail`, but that value is only meant to be read by the guest. It does not affect the behaviour of `try_enable_notification`.

Please keep me true here :)

bchalios (changes requested):
Just a minor comment about a comment and we're good to go :)

src/vmm/src/devices/virtio/queue.rs
roypat: Yeah, that was exactly my thinking! I left a comment on the actual `#[kani::stub(...)]` where this is used further down, but maybe I should also put something here

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -562,25 +562,87 @@ impl Queue {
...
+    }
+
+    pub struct ProofContext(pub Queue, pub ProofGuestMemory);
 
     pub struct MmapRegionStub {
         addr: *mut u8,
         size: usize,
-        bitmap: Option<AtomicBitmap>,
+        bitmap: (),
```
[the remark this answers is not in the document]
roypat: Mh, let me check. If it doesnt affect performance then I don't see why we shouldn't add it back tbh. I think I removed it in prototyping and then just forget to add it back

src/vmm/src/devices/virtio/queue.rs
bchalios: I'd vote for adding the explanation in the commit message and the stub itself, rather than where we call it.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: If it doesn't add overhead, I would include it anyway. Even if we don't exercise related functionality today, we might do in the future.

src/vmm/src/devices/virtio/queue.rs
roypat: Done :)

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
roypat: Mh, seems it does add quite some overhead. I had some look at the code in question, and it contains some non-trivial arithmetic and loop that causes kani to go back to not terminating. Which now leaves me to wonder if maybe this is the real reason why performance improved instead of mocking out `GuestMemoryMmap` :thinking:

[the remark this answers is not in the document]
roypat: Nope, both optimizations are needed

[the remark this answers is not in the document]
bchalios: what about the question whether we need this?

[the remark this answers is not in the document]
roypat: I think its orthogonal to the properties we want to prove here: the bitmap tracking is vm-memory functionality and should be covered by kani proofs in vm-memory. Here we want to prove correctness properties of the virtio queue implementation, which are independent on the memory tracking functionality.

bchalios (approved):

pb8o (approved):

Issue #4176, opened 2023-10-15
[Bug] firecracker shuts down successfully with non-zero exit code without API


mkroening:
# Describe the bug

When running Firecracker without API, it prints `Error: RunWithoutApiError(Shutdown(Ok))` and returns with a non-zero exit code, although the kernel shut down cleanly, which is also indicated by the error message.

## To Reproduce

Launch any kernel without API. This has been tested with [Hermit](https://github.com/hermit-os). It should work the same with any other kernel.

1.  Create `firecracker-config.json`:
    
    ```
    {
      "boot-source": {
        "kernel_image_path": "rusty-loader-x86_64-fc",
        "initrd_path": "hello_world",
        "boot_args": ""
      },
      "drives": [],
      "machine-config": {
        "vcpu_count": 1,
        "mem_size_mib": 256,
        "smt": false
      }
    }
    ```

2.  Launch Firecracker:

    ```
    firecracker --no-api --config-file firecracker-config.json --log-path firecracker.log --level Info --show-level --show-log-origin
    ```

3.  Compare output and logs from 1.4.1 with 1.5.0:

    - 1.4.1 exit code: 0
    - 1.5.0 exit code: 1

    - 1.4.1 output:

      ```
      [kernel messages]
      ```

    - 1.5.0 output:

      ```
      2023-10-15T22:41:46.865613581 [anonymous-instance:main] Running Firecracker v1.5.0
      [kernel messages]
      Error: RunWithoutApiError(Shutdown(Ok))
      ```

    - 1.4.1 log:

      ```
      Running Firecracker v1.4.1
      2023-10-15T22:41:27.341970753 [anonymous-instance:main:INFO:src/vmm/src/device_manager/mmio.rs:393] Artificially kick devices.
      2023-10-15T22:41:27.342026016 [anonymous-instance:main:INFO:src/firecracker/src/main.rs:509] Successfully started microvm that was configured from one single json
      2023-10-15T22:41:27.367335451 [anonymous-instance:fc_vcpu 0:INFO:src/vmm/src/vstate/vcpu/mod.rs:497] Received KVM_EXIT_SHUTDOWN signal
      2023-10-15T22:41:27.367358340 [anonymous-instance:main:INFO:src/vmm/src/lib.rs:842] Vmm is stopping.
      2023-10-15T22:41:27.367426447 [anonymous-instance:main:INFO:src/vmm/src/lib.rs:842] Vmm is stopping.
      ```

[...]

PR #4178, opened 2023-10-16
fix: Pass through no-API shutdown exit code (main)

JonathanWoollett-Light:
## Changes

Pass through no-API shutdown exit code

## Reason

https://github.com/firecracker-microvm/firecracker/issues/4176

Without passing through the exit code a non-zero exit code was returned on a successful no-API shutdown.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/firecracker/src/main.rs
+++ b/src/firecracker/src/main.rs
@@ -83,6 +83,7 @@ impl From<MainError> for ExitCode {
             MainError::InvalidLogLevel(_) => FcExitCode::BadConfiguration,
             MainError::RunWithApi(ApiServerError::MicroVMStoppedWithoutError(code)) => code,
             MainError::RunWithApi(ApiServerError::MicroVMStoppedWithError(code)) => code,
+            MainError::RunWithoutApiError(RunWithoutApiError::Shutdown(code)) => code,
             _ => FcExitCode::GenericError,
         };
 
--- a/tests/framework/vm_config_network.json
+++ b/tests/framework/vm_config_network.json
@@ -1,0 +1,41 @@
+{
+  "boot-source": {
+    "kernel_image_path": "vmlinux.bin",
+    "boot_args": "console=ttyS0 reboot=k panic=1 pci=off",
+    "initrd_path": null
+  },
+  "drives": [
+    {
+      "drive_id": "rootfs",
+      "path_on_host": "bionic.rootfs.ext4",
+      "is_root_device": true,
+      "partuuid": null,
+      "is_read_only": false,
+      "cache_type": "Unsafe",
+      "io_engine": "Sync",
+      "rate_limiter": null
+    }
+  ],
+  "machine-config": {
+    "vcpu_count": 2,
+    "mem_size_mib": 1024,
+    "smt": false,
+    "track_dirty_pages": false
+  },
+  "cpu-config": null,
+  "balloon": null,
+  "network-interfaces": [
+    {
+      "iface_id": "eth0",
+      "host_dev_name": "tap0",
+      "guest_mac": "06:00:c0:a8:00:02",
+      "rx_rate_limiter": null,
+      "tx_rate_limiter": null
+    }
+  ],
+  "vsock": null,
+  "logger": null,
+  "metrics": null,
+  "mmds-config": null,
+  "entropy": null
+}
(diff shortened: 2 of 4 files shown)


mkroening:
Status code from https://github.com/firecracker-microvm/firecracker/issues/4176#issuecomment-1764347011 looks good, but it still prints `Error: RunWithoutApiError(Shutdown(Ok))`.

src/firecracker/src/main.rs
```diff
@@ -83,6 +83,7 @@ impl From<MainError> for ExitCode {
             MainError::InvalidLogLevel(_) => FcExitCode::BadConfiguration,
             MainError::RunWithApi(ApiServerError::MicroVMStoppedWithoutError(code)) => code,
             MainError::RunWithApi(ApiServerError::MicroVMStoppedWithError(code)) => code,
+            MainError::RunWithoutApiError(RunWithoutApiError::Shutdown(code)) => code,
```
pb8o, line 86: Can we add an integration test too? to the `test_api.py::test_happy` case maybe?

JonathanWoollett-Light: To get the behavior as seen in the issue requires the VCPU's to exit on startup, to do this requires a custom kernel/initrd that does this, as seen https://github.com/firecracker-microvm/firecracker/issues/4176#issuecomment-1764290192

tests/framework/vm_config.json
```diff
@@ -1,33 +1,34 @@
...
-  "cpu-config": null,
-  "balloon": null,
-  "network-interfaces": [],
-  "vsock": null,
-  "logger": null,
-  "metrics": null,
-  "mmds-config": null,
-  "entropy": null
-}
+    "boot-source": {
```
sudanl0, line 2: nit: the change seems to be just tabs so, do we actually need this?

JonathanWoollett-Light: You are right, this change should not be here,

tests/framework/vm_config_network.json
```diff
@@ -0,0 +1,42 @@
...
+        "vcpu_count": 2,
+        "mem_size_mib": 1024,
+        "smt": false,
+        "track_dirty_pages": false
+    },
+    "cpu-config": null,
+    "balloon": null,
+    "network-interfaces": [
+        {
+            "iface_id": "1",
```
sudanl0, line 29: if we use eth0 here we could later consider this file as a good reference for getting started guide to start Firecracker with network.

src/firecracker/src/main.rs
sudanl0: can we create a github issue to extract exit_code of Firecracker and see if someone from the community can help on it?

tests/framework/vm_config_network.json
JonathanWoollett-Light: This makes sense, I think this would be in a separate PR.

src/firecracker/src/main.rs
JonathanWoollett-Light: Yes I can add an issue for adding the ability to get the exit code with our test framework.

JonathanWoollett-Light: I've created #4186 .

pb8o (approved):

src/firecracker/src/main.rs
sudanl0: If main() is modified as below then we can actually verify the effect of this change:
```
fn main() -> ExitCode {
    let result = main_exec();
    if let Err(err) = result {
        error!("{err}");
        eprintln!("Error: {err:?}");
        let exit_code = ExitCode::from(err);
        error!("{exit_code:?}");
        exit_code
    } else {
        ExitCode::SUCCESS
    }
}
```
with the change the print is ExitCode(unix_exit_status(0))
while without this change the print is ExitCode(unix_exit_status(1))

sudanl0: ok so as per offline discussion:
- with or without the fix we still get the print "RunWithoutApiError error: MicroVMStopped without an error: Ok" in test_config_start_no_api_exit() so the test doesn't really test the fix however, it still makes sure that we get an Ok print for shutdown case.
- This is ok for now since we have a follow up PR where we won't see the "RunWithoutApiError error: MicroVMStopped without an error: Ok" print for the shutdown case and instead will get a proper shutdown ok message.
- We don't want to go ahead with this suggestion of adding error print in main() is because it will be redundant once the follow up PR is done.

sudanl0 (approved):

PR #4196, opened 2023-10-20
test(ab): artificially increase network latency test cases

roypat:
Splits the network latency test, which currently does 900 pings from guest to host into two tests each doing 450 pings. The rationale behind this is that we have very little network latency test cases (only 2, which is the number of guest kernels), which means that the A/B-testing framework does not have enough data to meaninfully correct for outliers (it has only 2 data points). This change increases the number of data points it can work with to 4, which should hopefully help with the high false-positive rate we have been seeing from this test.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/performance/test_network_ab.py
+++ b/tests/integration_tests/performance/test_network_ab.py
@@ -8,10 +8,9 @@ import pytest
 
 from framework.utils_iperf import IPerf3Test, emit_iperf3_metrics
 
-# each iteration is 30 * 0.2s = 6s
-# Thus 30 iterations are 6s * 30 = 5min
-REQUEST_PER_ITERATION = 30
-ITERATIONS = 30
+# each iteration is 15 * 30 * 0.2s = 90s
+ROUNDS = 15
+REQUEST_PER_ROUND = 30
 DELAY = 0.2
 
 #  MicroVM configuration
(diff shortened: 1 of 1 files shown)


tests/integration_tests/performance/test_network_ab.py
```diff
@@ -67,7 +66,8 @@ def network_microvm(request, microvm_factory, guest_kernel, rootfs):
 
 @pytest.mark.nonci
 @pytest.mark.parametrize("network_microvm", [1], indirect=True)
-def test_network_latency(network_microvm, metrics):
+@pytest.mark.parametrize("iteration", [1, 2])
```
sudanl0, line 69: would this be enough or should we increase the iterations to more now?

pb8o, line 69: May want to copy the explanation in the commit to the top of this file so that we don't switch it back to 1 iteration with 900 pings.

pb8o (approved):

tests/integration_tests/performance/test_network_ab.py
sudanl0: should we consider testing with more vcpus to the test to get more datapoints?

roypat: Done :)

sudanl0 (approved):

pb8o (approved):

PR #4200, opened 2023-10-23
Update Getting started docs

sudanl0:
## Change

- Firecracker CI artifacts were updated but the links were not so update the link.
- Unify the way firecracker binary is made available in a directory so that it is easier for the users to copy pase the commands with less modifications.

## Reason

A broken link to CI artifact was reported.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

- ~~[ ] This functionality cannot be added in [`rust-vmm`][1].~~

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -76,7 +76,7 @@ with a Ubuntu 22.04 rootfs from our CI:
 ARCH="$(uname -m)"
 
 # Download a linux kernel binary
-wget https://s3.amazonaws.com/spec.ccfc.min/firecracker-ci/v1.6/${ARCH}/vmlinux-5.10.197
+wget https://s3.amazonaws.com/spec.ccfc.min/firecracker-ci/v1.6/${ARCH}/vmlinux-5.10.198
 
 # Download a rootfs
 wget https://s3.amazonaws.com/spec.ccfc.min/firecracker-ci/v1.6/${ARCH}/ubuntu-22.04.ext4
@@ -191,7 +191,7 @@ curl -X PUT --unix-socket "${API_SOCKET}" \
     }" \
     "http://localhost/logger"
 
-KERNEL="./vmlinux-5.10.186"
+KERNEL="./vmlinux-5.10.198"
 KERNEL_BOOT_ARGS="console=ttyS0 reboot=k panic=1 pci=off"
 
 ARCH=$(uname -m)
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -127,10 +127,10 @@ sudo systemctl start docker
 # This will produce the firecracker and jailer binaries under
 # `./firecracker/build/cargo_target/${toolchain}/debug`.
 #
-sudo ./firecracker/tools/devtool build
+sudo ./firecracker_src/tools/devtool build
 
 # Rename the binary to "firecracker"
-mv ./firecracker/build/cargo_target/${ARCH}-unknown-linux-musl/debug/firecracker firecracker
+sudo cp ./firecracker_src/build/cargo_target/${ARCH}-unknown-linux-musl/debug/firecracker firecracker
```
kalyazin, line 133: to be honest, I don't quite see why we need this step. Can we run the binary from where it was produced?

docs/getting-started.md
```diff
@@ -143,11 +143,10 @@ process via HTTP requests:
 API_SOCKET="/tmp/firecracker.socket"
 
 # Remove API unix socket
-rm -f $API_SOCKET
+sudo rm -f $API_SOCKET
```
kalyazin, line 146: would it be a bad idea to remove all sudoes and let users manage their own permissions? Now it looks like it's mandatory, which may scare some of the users. We can leave a note saying something like "depending on your permissions management, you may need to run some of these with superuser rights". I'm not fully convinced either way at the moment.

docs/getting-started.md
sudanl0: there is an inconsistency between the steps. The below step downloads the release binary in the current folder
> mv release-${latest}-$(uname -m)/firecracker-${latest}-${ARCH} firecracker

so the below step to run firecracker from build doesn't work when we download the release binary.

># Run firecracker
>./firecracker/build/cargo_target/${ARCH}-unknown-linux-musl/debug/firecracker \
>    --api-sock "${API_SOCKET}"

docs/getting-started.md
sudanl0: the typical approach for the getting started page is to run the firecracker binary without the jailer and so we need to add sudo to run the commands. I usually copy paste the command to a notepad and add the sudo to it to run. This change would save that time.
And we already ask them to run a bunch of commands with sudo later in the doc.

docs/getting-started.md
kalyazin: I see, it's to unify the call to the binary that is either built or downloaded...

kalyazin (approved):

zulinx86 (approved):

PR #4188, opened 2023-10-19
Add preliminary support for a Linux 6.1 guest kernel, plus kvm_ptp

pb8o:
## Changes

- Add preliminary support for Linux 6.1 guest kernel. Only for c7g and Linux 6.1 host
- Add kvm_ptp support in aarch64 and Linux 6.1 

## Reason

...

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following
Developer Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

---

- [x] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #4149, opened 2023-10-05
A/B-ification of tests that can fail due to external influences

roypat:
## Changes

This PR introduces A/B-approaches to tests that can fail due to external factors that would otherwise block all PRs until they are resolved. Using the `cargo audit` test as an example, we want to differentiate between the following two cases:
- A PR introduces a new dependency for which a rustsec advisory exists
- A pre-existing dependency has a new rustsec advisory filed

We want to fail PR CI in the first scenario, but be able to deal with the second scenario out-of-band. A/B-Testing allows us to do exactly that.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following
Developer Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

---

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #4206, opened 2023-10-29
Fix nightly pr failures

pb8o:
## Changes

PRs #4149 and #4188 had a few regressions bugs that were not reachable from the normal PR tests.

In addition, add a change so we only compile example binaries once, instead of once per worker. This avoids downloading and compiling code.

## Reason

To make all pipelines pass.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


tests/integration_tests/security/test_vulnerabilities.py
```diff
@@ -180,6 +180,7 @@ def test_spectre_meltdown_checker_on_host(spectre_meltdown_checker):
         comparator=set_did_not_grow_comparator(
             spectre_meltdown_reported_vulnerablities
         ),
+        ignore_return_code=True,
```
sudanl0, line 183: is this a temporary fix till we find why checker returns 2?

sudanl0 (approved):

ShadowCurse (approved):

tests/integration_tests/security/test_vulnerabilities.py
bchalios: @zulinx86 do you have context around this?

bchalios (changes requested):
I think it's important to answer @sudanl0's question regarding the workaround we use.

tests/integration_tests/security/test_vulnerabilities.py
pb8o: I lack the context. In the A/B version of the test we ignore the return codes, so it would make sense to do it here too. I agree we need to figure out if ignoring is the right thing to do, but it could be done as a follow up, while we fix the pipelines.

zulinx86: What Pablo says is right. In A/B testing, the test would like just to test whether there is difference between A and B. The result of spectre/meltdown checker can fail due to AMI update. Not to be blocked by this type of failure, we introduced A/B testing. So we should ignore the return code here. But let me check whether it ignores the return code if it is not a PR.

zulinx86: I've caught up the discussion here now. Before introduction of A/B testing, this test was run as part of optional PR since it can fail due to AMI updates. My suggestion is to run this test mandatorily on PR but run this separately on nightly test (e.g. run in `.buildkite/pipeline_security.py`) to distinguish from functional test failures.

bchalios: Ok, so IIUC it makes sense to merge this PR to unblock the nightly pipeline and then we can do the suggested change on a follow up?

bchalios (approved):

PR #4216, opened 2023-11-06
feat: add a debug log for vm build time

pb8o:
Add a couple of events to the debug logs. With this information we can extract information of how long did those events take.

Added as debug messages to make it clear to customers that these are unstable and may change format or disappear at any moment, at should not be depended upon.

## Changes

...

## Reason

To have better visibility on where time goes during boottime.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -52,7 +52,7 @@ use crate::devices::virtio::vsock::{Vsock, VsockUnixBackend};
 use crate::devices::BusDevice;
 #[cfg(target_arch = "aarch64")]
 use crate::logger;
-use crate::logger::error;
+use crate::logger::{debug, error};
 use crate::persist::{MicrovmState, MicrovmStateError};
 use crate::resources::VmResources;
 use crate::vmm_config::boot_source::BootConfig;
@@ -380,14 +380,16 @@ pub fn build_and_boot_microvm(
     event_manager: &mut EventManager,
     seccomp_filters: &BpfThreadMap,
 ) -> Result<Arc<Mutex<Vmm>>, StartMicrovmError> {
+    debug!("event_start: build microvm for boot");
     let vmm = build_microvm_for_boot(instance_info, vm_resources, event_manager, seccomp_filters)?;
-
+    debug!("event_end: build microvm for boot");
     // The vcpus start off in the `Paused` state, let them run.
+    debug!("event_start: boot microvm");
     vmm.lock()
         .unwrap()
         .resume_vm()
         .map_err(StartMicrovmError::Internal)?;
-
+    debug!("event_end: boot microvm");
     Ok(vmm)
 }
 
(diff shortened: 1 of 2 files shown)


src/vmm/src/builder.rs
```diff
@@ -444,6 +446,7 @@ pub fn build_microvm_from_snapshot(
     })?;
 
     // Build Vmm.
+    debug!("event_start: build microvm");
```
kalyazin, line 449: would it make sense to have separate metrics for boot and restore cases?

tests/integration_tests/performance/test_boottime.py
```diff
@@ -184,3 +197,11 @@ def test_boottime(
         vm.start()
         boottime_us = _get_microvm_boottime(vm)
         metrics.put_metric("boot_time", boottime_us, unit="Microseconds")
+        timestamps = find_events(vm.log_data)
+        build_time = timestamps["build microvm"]["duration"]
+        metrics.put_metric("build_time", build_time.microseconds, unit="Microseconds")
+        metrics.put_metric(
+            "guest_boot_time",
```
kalyazin, line 204: is the notation different on purpose (`build_time` and `guest_boot_time`), not `guest_build_time` or `boot_time`?

kalyazin: yeah, that would be at least consistent

pb8o: What I wanted to convey is that guest_boot_time depends on the guest OS, while build time is Firecracker/host.

kalyazin: ok, that makes sense indeed.

kalyazin: we don't need a changelog entry for this, do we?

pb8o: Yeah I don't think we do. I don't want to publicize it too much yet :)

kalyazin (approved):

src/vmm/src/builder.rs
```diff
@@ -448,6 +450,7 @@ pub fn build_microvm_from_snapshot(
     })?;
 
     // Build Vmm.
+    debug!("event_start: build microvm from snapshot");
```
kalyazin, line 453: wouldn't we like a metric for this too?

pb8o: Hmm good point. But I am not sure if we have a good test candidate to generate those from. Should I remove those since they are unused.

sudanl0 (approved):

src/vmm/src/builder.rs
sudanl0: I think we can add the metrics as a follow up PR but this change would be good to have right now.

Issue #2699, opened 2021-08-13
Add codegen-units=1 


gbionescu:
During the `0.25` release we've noticed a significant variation in the binary size, where simply changing the firecracker version would add 15kb to the binary size.

Later investigation showed that LTO was at fault for this change in size - simply disabling the LTO would produce consistent binaries.
Further reading the [cargo reference manual on lto](https://doc.rust-lang.org/cargo/reference/profiles.html#lto) shows that `No LTO is performed if codegen units is 1 or opt-level is 0.`. Testing without LTO indeed produces consistent binaries and as a bonus, smaller ones - the `v0.25` one is 0.5 mb smaller.

Also, in the context of a previously opened PR by @KarthikNedunchezhiyan [here](https://github.com/firecracker-microvm/firecracker/pull/2148), I propose that we add `codegen-units=1` to the release profile to prevent variations such as this one.

alindima:
This PR is affected by the same thing. Even though it gets rid of a ton of code, the binary gets larger: https://github.com/firecracker-microvm/firecracker/pull/2661

JonathanWoollett-Light:
PR: https://github.com/firecracker-microvm/firecracker/pull/4218

PR #4218, opened 2023-11-07
build: `-Ccodegen-units=1`

JonathanWoollett-Light:
Issue: https://github.com/firecracker-microvm/firecracker/issues/2699

## Changes

Yields a 180224 bytes decrease in the binary size.

## Reason

Adds `-Ccodegen-units=1` to improve performance and decrease binary size.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


kalyazin (approved):

roypat:
Did you measure the impact on compile time? The [docs](https://doc.rust-lang.org/nightly/rustc/codegen-options/index.html#codegen-units say
> When a crate is split into multiple codegen units, LLVM is able to process them in parallel. Increasing parallelism may speed up compile times, but may also produce slower code. Setting this to 1 may improve the performance of generated code, but may be slower to compile.

If its bad we might only want to enable it for release builds so that local testing builds remain fast

kalyazin:
>If its bad we might only want to enable it for release builds so that local testing builds remain fast

I would prefer for both builds to use the same setting to avoid surprises.

JonathanWoollett-Light:
> Did you measure the impact on compile time?

The compile time is fast with and without the option, there is no noticeable difference.

roypat:
> > If its bad we might only want to enable it for release builds so that local testing builds remain fast
> 
> I would prefer for both builds to use the same setting to avoid surprises.

Our CI only uses release builds anyway, so I don't think there would be any surprises anywhere. But since the impact seems to be minimal, ship it

roypat (approved):

PR #4073, opened 2023-08-24
Final PVH fixups

roypat:
Bring the feature branch up-to-date with all changes that happened in `main` since it was last updated, and which didn't show up as a merge conflict. Also do some minor doc fixes, and add integration testing.

Since these days all our CI kernels are compiled with `CONFIG_PVH=y`, integration testing is just a matter of adding an assert to verify that PVH boot mode is indeed being used. We are not adding FreeBSD tests to our CI, as we have no easy means for maintaining FreeBSD artifacts, and last time I checked (which admittedly was August 23), it also didn't boot on AMD instances. 

See also https://github.com/firecracker-microvm/firecracker/issues/3041 

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following
Developer Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

---

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #4228, opened 2023-11-09
cargo QoL improvements

roypat:
## Changes

Turns example binaries in log-instrument into actual examples, and allows us to use `cargo run` to execute our helper tools

## Reason

I couldn't do `cargo run --bin cpu-template-helper` before + compile time improvements.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


JonathanWoollett-Light:
This has `-96.88%` coverage on `src/log-instrument/src/lib.rs`. Do we care about coverage here?

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
 members = ["src/clippy-tracing", "src/cpu-template-helper", "src/firecracker", "src/jailer", "src/log-instrument", "src/log-instrument-macros", "src/rebase-snap", "src/seccompiler", "src/snapshot-editor"]
-default-members = ["src/firecracker"]
+default-members = ["src/clippy-tracing", "src/cpu-template-helper", "src/firecracker", "src/rebase-snap", "src/seccompiler", "src/snapshot-editor"]
```
JonathanWoollett-Light, line 3: Can we not just remove `default-members`? It seems simplest to just get rid of this complexity. Do we lose any value removing it?

roypat: Yeah, that was my first approach, but it seems that originally, the jailer was intentionally excluded from this (see the commit linked in my commit), so I decided to keep it like that while adding the binaries that we actually want to be able to execute.

roypat:
> This has `-96.88%` coverage on `src/log-instrument/src/lib.rs`. Should we care about coverage here?

I dont think this matters much, the tests are still there after all

JonathanWoollett-Light:
> > This has `-96.88%` coverage on `src/log-instrument/src/lib.rs`. Should we care about coverage here?
> 
> I dont think this matters much, the tests are still there after all

It seems this is most frequently how we approach this. If we functionality ignore it by never taking action as a result of the metric. Why do we track it?

Even if outside this specific PR, although it can act as a good example, we should clarify the story here.

pb8o (approved):

Cargo.toml
JonathanWoollett-Light: I think we should remove `default-members` or we should document why specific items are in `default-members`.

roypat:
> > > This has `-96.88%` coverage on `src/log-instrument/src/lib.rs`. Should we care about coverage here?
> > 
> > 
> > I dont think this matters much, the tests are still there after all
> 
> It seems this is always how we approach this. If we functionally ignore it by never taking action as a result of the metric. Why do we track it? Maybe we should remove it if it is just noise?
> 
> This PR can act as a good example, we should clarify the story here.

To me, we track code coverage to identify gaps in our testing. E.g. a low coverage percentage can be a hint that some PR is missing some unit tests, or the annotations that codecov adds to the PR diff shows some uncovered lines which then leads us to add more testing (this is how I discovered the initrd issues in #4073). 

I agree that looking at the metric in isolation is meaningless. To me its a case-by-case decision on every PR. In this specific case, the coverage drops, but there is no gap in testing, so I ignore it.

sudanl0:
> > > > This has `-96.88%` coverage on `src/log-instrument/src/lib.rs`. Should we care about coverage here?
> > > 
> > > 
> > > I dont think this matters much, the tests are still there after all
> > 
> > 
> > It seems this is always how we approach this. If we functionally ignore it by never taking action as a result of the metric. Why do we track it? Maybe we should remove it if it is just noise?
> > This PR can act as a good example, we should clarify the story here.
> 
> To me, we track code coverage to identify gaps in our testing. E.g. a low coverage percentage can be a hint that some PR is missing some unit tests, or the annotations that codecov adds to the PR diff shows some uncovered lines which then leads us to add more testing (this is how I discovered the initrd issues in #4073).
> 
> I agree that looking at the metric in isolation is meaningless. To me its a case-by-case decision on every PR. In this specific case, the coverage drops, but there is no gap in testing, so I ignore it.

Could you please add more details on why there is no gap in testing?
Also if improving the build time is the only reason why the tests were turned to examples then, could we add the details on by how much did the build improve? I don't mind an additional build time of few minutes if codecov can show me a 100%coverage.

sudanl0 (changes requested):
I think the files were designed to be tests so changing them to examples doesn't add much value to it.
If improving the build time is the only reason why the tests were turned to examples then, could we add the details on by how much did the build improve? I don't mind an additional build time of few minutes if codecov can show me a 100%coverage.

roypat:
> I think the files were designed to be tests so changing them to examples doesn't add much value to it. If improving the build time is the only reason why the tests were turned to examples then, could we add the details on by how much did the build improve? I don't mind an additional build time of few minutes if codecov can show me a 100%coverage.

The compile time impact to `tools/devtool build --release` is 1m 45s for main vs 1m 27s on this PR branch. 

But independently of that, I feel fairly strongly about not exposing testing details as binary targets. Currently for example `./tools/devtool build --release` will point the user to the `release` directory, and if they look there, they see six very confusing binaries just named "one" through "six" (in addition to the expected firecracker binaries). They probabl also show up in various other `cargo` outputs (for example, if we go with Jonathan's suggestions to remove `default-members` altogether from Cargo.toml, they will show up in the output of `cargo run --bin` when it lists possible binaries to run). This really shouldnt be the case.

sudanl0:
> > I think the files were designed to be tests so changing them to examples doesn't add much value to it. If improving the build time is the only reason why the tests were turned to examples then, could we add the details on by how much did the build improve? I don't mind an additional build time of few minutes if codecov can show me a 100%coverage.
> 
> The compile time impact to `tools/devtool build --release` is 1m 45s for main vs 1m 27s on this PR branch (so we see an improvement of ~14%).
> 
> But independently of that, I feel fairly strongly about not exposing testing details as binary targets. Currently for example `./tools/devtool build --release` will point the user to the `release` directory, and if they look there, they see six very confusing binaries just named "one" through "six" (in addition to the expected firecracker binaries). This is just a single bug away from us accidentally uploading them as official release artifacts, too. They probably also show up in various other `cargo` outputs (for example, if we go with Jonathan's suggestions to remove `default-members` altogether from Cargo.toml, they will show up in the output of `cargo run --bin` when it lists possible binaries to run). This really shouldnt be the case.

The build improvement is ~20secs which is too small to justify removing the tests however, I understand why having those weird binaries could cause a confusion and this concern was also [raised](https://github.com/firecracker-microvm/firecracker/pull/4063#discussion_r1373228480) when they were first added.
If we want to remove the test bins I would suggest moving them to better place instead of completely removing them.
There is an alternative solution proposed [here](https://github.com/firecracker-microvm/firecracker/pull/4063#discussion_r1373272044).

I agree that the bins could cause confusion where they are right now but, replacing them with a python integration test goes against our code coverage policy. I suggest we create an issue to do the right thing (moving to correct place while maintaining coverage) but until then I don't think we should remove them.

roypat:
> The build improvement is ~20secs which is too small to justify removing the tests however, I understand why having those weird binaries could cause a confusion and this concern was also [raised](https://github.com/firecracker-microvm/firecracker/pull/4063#discussion_r1373228480) when they were first added. If we want to remove the test bins I would suggest moving them to better place instead of completely removing them. There is an alternative solution proposed [here](https://github.com/firecracker-microvm/firecracker/pull/4063#discussion_r1373272044).
> 
> I agree that the bins could cause confusion where they are right now but, replacing them with a python integration test goes against our code coverage policy. I suggest we create an issue to do the right thing (moving to correct place while maintaining coverage) but until then I don't think we should remove them.

I want to stress that no tests are being removed in this PR. We are still exercising exactly the same code paths and checking exactly the same assertions. Here, we have a case of code being very hard to unit tests (we could mock the logger, which would also solve this problem, but even then we cannot properly check any assertions because the log messages contain thread ids, and if the logs arent generated by a separate process, these become non-deterministic). Whenever code is very hard to unit tests, and workarounds introduce drawbacks (such as accidentally vending dummy binaries to users, or adding significant complexity to the testing code, as Jonathan mentions his alternative would do), we should use integration testing for it. It is also a matter of consistency with the rest of our testing framework: UFFD and seccomp tests are cargo example that are driven by python integration tests (and when we did this, even promoting them to examples from crates hidden inside of our testing framework went through a lot of discussion because of the same point: It puts testing code into a more prominent place alongside production code). And we have also in the past decided that we value sensibility of our testing over the code coverage number, when we decided to remove "coverage generating" tests for debug implementations, because they added no functional value. Because the code coverage number is a proxy for the property that we actually care about: Code being tested. And if we only focus on the proxy and try to game the number, we lose track of that goal, and _that_ is when it becomes useless.

> [...] goes against our code coverage policy.

I am not aware of us having any policy regarding code coverage (I know we decided against introducing static checks in our CI for it, because it was impossible to meaningfully phrase failure conditions that do not rule our valid usecases). We only have a team-set goal to reach 85% coverage, but even here I could argue that the 81.77% reported on the HEAD of this PR is a more accurate representation of our progress towards this goal:  We currently include the source code of `one.rs`, `two.rs`,... `six.rs` and so on as covered lines (which add up to ~100 lines, or 0.34% of tracked lines), but they are non-production code. By including them in our progress towards the goal, we are "cheating", in a way.

kalyazin:
I would agree that tests that run the entire binary tool with files as inputs are integration tests by nature and should not be contributing into our unit test coverage metric. In some cases (like this one) unit testing is hard and it is fine to resort to integration tests only while having clear visibility of gaps in unit test coverage (which we are fine with).

sudanl0:
We might not have a written policy but from what I see we concluded with using unit test instead of integration test for code coverage in multiple decision discussions we had internally.

The problem was not with this particular unit test being less efficient and more complex vs the integration test, it was which approach provides better visibility. I didn't see codecov as a proxy but as a tool which provides us a visibility and metric on coverage. Even if integration test in this case might provide us better testing we don't have a metric to know if the testing or coverage was reduced if someone modifies the test. I imagine it would be hard for someone to review all the corner cases if the interaction test was modified.
I am not saying we should let the binaries get into production or our release process, but with moving it to integration test we lose visibility if all cases are covered.

Having said that, I also agree that we can never achieve 100% code coverage and with a good enough explanation as you provided it seems logical to go ahead with this change.

BTW, thanks for being patient with me and explaining this in this detail. I really appreciate that you didn't take the easy path to just remove the commit and fought for the right thing.
And I hope if in future if some PR reduces unit test coverage they put same amount of effort in providing a valid argument for it such as this.

sudanl0 (approved):

PR #4231, opened 2023-11-09
Disable CPU idle states when running performance tests in CI

roypat:
## Changes

- Disable idle states when running performance tests in CI
- Limit P-states to the highest p-state at which the processor can continuously operate safely.
- Disable turbo boost also on non-intel instances (if they support turbo boost in the first place)

## Reason

Consolidate performance tweaks in one location

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/.buildkite/pipeline_ab.py
+++ b/.buildkite/pipeline_ab.py
@@ -62,7 +62,7 @@ def build_group(test):
     test_path = test.pop("test_path")
     return group(
         label=test.pop("label"),
-        command=f"./tools/devtool -y test --ab {devtool_opts} -- {REVISION_A} {REVISION_B} --test {test_path}",
+        command=f"./tools/devtool -y test --performance --ab {devtool_opts} -- {REVISION_A} {REVISION_B} --test {test_path}",
         artifacts=["./test_results/*"],
         instances=test.pop("instances"),
         platforms=test.pop("platforms"),
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -456,6 +456,24 @@ class Microvm:
             return True
         return False
 
+    def pin_threads(self, first_cpu):
+        """
+        Pins all microvm threads (VMM, API and vCPUs) to consecutive physical cpu core, starting with "first_cpu"
+        """
+        for vcpu, pcpu in enumerate(range(first_cpu, first_cpu + self.vcpus_count)):
+            assert self.pin_vcpu(
+                vcpu, pcpu
+            ), f"Failed to pin fc_vcpu {vcpu} thread to core {pcpu}."
+        # The cores first_cpu,...,first_cpu + self.vcpus_count - 1 are assigned to the individual vCPU threads,
+        # So the remaining two threads (VMM and API) get first_cpu + self.vcpus_count
+        # and first_cpu + self.vcpus_count + 1
+        assert self.pin_vmm(
+            first_cpu + self.vcpus_count
+        ), "Failed to pin firecracker thread."
+        assert self.pin_api(
+            first_cpu + self.vcpus_count + 1
+        ), "Failed to pin fc_api thread."
+
     def spawn(
         self,
         log_file="fc.log",
--- a/tools/devtool
+++ b/tools/devtool
@@ -582,9 +582,22 @@ apply_linux_61_tweaks() {
 
 
 apply_performance_tweaks() {
-  # Disable turbo boost. Some of our tests are performance tests, and we want minimum variability wrt processor frequency
-  # See also https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/processor_state_control.html
-  echo 1 |sudo tee /sys/devices/system/cpu/intel_pstate/no_turbo &> /dev/null
+  # m6a instances do not support the amd_pstate driver (yet), so nothing we can do there
+  if [[ -d /sys/devices/system/cpu/intel_pstate ]]; then
+    # Disable turbo boost. Some of our tests are performance tests, and we want minimum variability wrt processor frequency
+    # See also https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/processor_state_control.html
+    echo 1 |sudo tee /sys/devices/system/cpu/intel_pstate/no_turbo &> /dev/null
+
+    # Save old values to restore later
+    MIN_PERF_PCT=$(cat /sys/devices/system/cpu/intel_pstate/min_perf_pct)
+    MAX_PERF_PCT=$(cat /sys/devices/system/cpu/intel_pstate/max_perf_pct)
+
+    # Force the CPU to continuously stay in the highest, non-turbo P-state. The P-state will determine the
+    # CPU's clock frequency.
+    # https://www.kernel.org/doc/html/v4.12/admin-guide/pm/intel_pstate.html
+    echo 100 |sudo tee /sys/devices/system/cpu/intel_pstate/min_perf_pct
+    echo 100 |sudo tee /sys/devices/system/cpu/intel_pstate/max_perf_pct
+  fi
 
   # The governor is a linux component that can adjust CPU frequency. "performance" tells it to always run CPUs at
   # their maximum safe frequency. It seems to be the default for Amazon Linux, but it doesn't hurt to make this explicit.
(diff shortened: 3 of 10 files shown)


roypat:
welp, seems like there is yet another difference between my testing instance and the CI instances somewhere...

tools/devtool
```diff
@@ -621,27 +653,20 @@ cmd_test() {
...
     say "RPM microcode_ctl version: $(rpm -q microcode_ctl)"
-    say "Starting test run ..."
 
     # Testing (running Firecracker via the jailer) needs root access,
     # in order to set-up the Firecracker jail (manipulating cgroups, net
     # namespaces, etc).
     # We need to run a privileged container to get that kind of access.
     env |grep -P "^(AWS_EMF_|BUILDKITE|CODECOV_)" > env.list
+    if [[ "$BUILDKITE" == "true" ]] && [[ "$(uname --machine)" == "x86_64" ]]; then
+      say "Detected CI and performance tests, tuning CPU frequency scaling and idle states for reduced variability"
```
bchalios: I remember there was a problem with this condition, did you find out what it was?

tools/devtool
```diff
@@ -581,6 +581,38 @@ apply_linux_61_tweaks() {
 }
 
 
+apply_performance_tweaks() {
```
bchalios, line 584: Can you add messages for the configurations we are touch here, so that we can verify that they have been applied during a run?

tools/devtool
```diff
@@ -582,9 +582,18 @@ apply_linux_61_tweaks() {
 
 
 apply_performance_tweaks() {
-  # Disable turbo boost. Some of our tests are performance tests, and we want minimum variability wrt processor frequency
-  # See also https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/processor_state_control.html
-  sudo sh -c "echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo" &> /dev/null
+  # m6a instances do not support the amd_pstate driver (yet), so nothing we can do there
+  if [[ -d /sys/devices/system/cpu/intel_pstate ]]; then
```
bchalios, line 586: Should we try to do that for other drivers as well?

tools/devtool
```diff
@@ -603,13 +612,15 @@ apply_performance_tweaks() {
 
 unapply_performance_tweaks() {
   # reenable turbo boost
-  sudo sh -c "echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo" &> /dev/null
-
-  # We do not reset the governor, as keeping track of each CPUs configured governor is not trivial here. On our CI
-  # instances, the performance governor is current the default anyway (2023/11/14)
+  if [[ -d /sys/devices/system/cpu/intel_pstate ]]; then
+    sudo sh -c "echo 0 > /sys/devices/system/cpu/intel_pstate/no_turbo" &> /dev/null
+  fi
```
bchalios, line 617: Here, we should unapply the `min_perf_pct` and `max_perf_pct` configurations as well.

tools/devtool
```diff
@@ -641,6 +641,9 @@ cmd_test() {
                 shift
                 local cpuset_mems="$1"
                 ;;
+            "--performance")
+                local performance_tweaks=1;
+                ;;
```
bchalios, line 646: Should we add this flag to the help message as well?

tools/devtool
roypat: Ahhh, right, forgot about that, thanks!

Seems that now its even broken in the PR CI, let me investigate what's happening here :thinking:

tools/devtool
roypat: I'd prefer not to, I'd really like to avoid this being used outside of the CI

tools/devtool
roypat: seems that for some reason the "BUILDKITE" environment variable is unset inside of `cmd_test` when running things in my private pipeline.Not quite sure why, because I dump `env | BUILDKITE` right before running devtool and that contains `BUILDKITE=true`

tools/devtool
roypat: Done!

tools/devtool
roypat: Afaik for gravitron there's nothing to do (cpu frequency is constant anyway), and on AMD we don't have the pstate driver yet (https://wiki.archlinux.org/title/CPU_frequency_scaling#amd_pstate). I can have a look at acpi_pstate for AMD instances real quick if that allows us to do anything!

tools/devtool
roypat: This was because in my own pipeline I was running devtool as `sudo`, which reset environment variables. So its not a problem with the condition itself, just a problem with me haha

tools/devtool
roypat: Seems like we can at least disable turbo boost for m6a.metal, so done that :)

bchalios: awesome, thanks :)

tools/devtool
bchalios: unless something else runs afterwards. Right now, I can't think of anything that would be affected by this, other than these tests, but I feel a bit nervous about us adding something in the future and that gets affected in a way we don't understand because we will have forgotten this

tools/devtool
bchalios: ok, agreed

tools/devtool
roypat: Per offline discussion: The only thing stopping me from doing this was my lack of understanding of bash variable scopes. Fixed now :)

bchalios: thanks for that Patrick

bchalios (approved):

bchalios (approved):

tools/devtool
```diff
@@ -714,7 +722,7 @@ cmd_test() {
     cmd_fix_perms
 
     # undo performance tweaks (in case the instance gets recycled for a non-perf test)
-    if [[ "$BUILDKITE" == "true" ]] && [[ "$(uname --machine)" == "x86_64" ]]; then
+    if [[ $performance_tweaks -eq 1 ]] && [[ "$BUILDKITE" == "true" ]] && [[ "$(uname --machine)" == "x86_64" ]]; then
```
pb8o, line 731: hmm I think if we have the option then we would want to apply it regardless of the `BUILDKITE` value. For example if we are running it by hand.

tests/framework/microvm.py
```diff
@@ -456,6 +456,17 @@ def pin_api(self, cpu_id: int):
...
 
+    def pin_threads(self, first_cpu):
+        """
+        Pins all microvm threads (VMM, API and vCPUs) to consecutive physical cpu core, starting with "first_cpu"
+        """
+        assert self.pin_vmm(first_cpu), "Failed to pin firecracker thread."
+        assert self.pin_api(first_cpu + 1), "Failed to pin fc_api thread."
+        for i in range(self.vcpus_count):
+            assert self.pin_vcpu(
+                i, first_cpu + i + 2
```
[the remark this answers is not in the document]
roypat: If we start `i` at `first_cpu + 2` we would have to do `self.pin_vcpu(i - first_cpu - 2, i)` instead. I can move the `+2` before the `+i` and add a comment to clarify though

.buildkite/pipeline_ab.py
```diff
@@ -18,32 +18,32 @@
     "virtio-block": {
         "label": "🖴 Virtio Block Performance",
         "test_path": "integration_tests/performance/test_block_ab.py::test_block_performance",
-        "devtool_opts": "-c 1-10 -m 0",
+        "devtool_opts": "-c 0-10 -m 0",
```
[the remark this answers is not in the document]
roypat: Ah, okay, I'll happily add a comment and revert this. This commit was mostly a result of me playing around and trying to fix any sort of perceived weirdness

tools/devtool
roypat: Yeah, originally I wanted to have the `$BUILDKITE` check here as a failsafe against running this on a laptop and then leaving the power configuration messed up, but it does also rule out EC2 instances... Now that we also have the unapply working correctly, I feel a bit better about not having this check, so can remove it :)

tests/framework/microvm.py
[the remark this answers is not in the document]
roypat: Ah, I see what you mean now. Yes, that's indeed less confusing, let me update!

.buildkite/pipeline_ab.py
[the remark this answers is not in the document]
roypat: Done

tools/devtool
roypat: Done

tests/framework/microvm.py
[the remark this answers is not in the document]
roypat: Done

tools/devtool
```diff
@@ -581,6 +581,38 @@ apply_linux_61_tweaks() {
 }
 
 
+apply_performance_tweaks() {
+  # Disable turbo boost. Some of our tests are performance tests, and we want minimum variability wrt processor frequency
+  # See also https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/processor_state_control.html
+  sudo sh -c "echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo" &> /dev/null
```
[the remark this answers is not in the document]
roypat: Done! As synced offline, this allows us to save the "sudo sh -c" part :)

tests/framework/microvm.py
```diff
@@ -456,6 +456,22 @@ def pin_api(self, cpu_id: int):
             return True
         return False
 
+    def pin_threads(self, first_cpu):
+        """
+        Pins all microvm threads (VMM, API and vCPUs) to consecutive physical cpu core, starting with "first_cpu"
+        """
+        for i in range(self.vcpus_count):
```
pb8o, line 463: Shouldn't this depend on first_cpu? Since you have to keep track of both vcpus and pcpus, we may need sthg like:

```python
first_cpu = 8
vcpu_count = 4

for vcpu, pcpu in enumerate(range(first_cpu, first_cpu + vcpu_count)): 
    print(vcpu, pcpu)
0 8
1 9
2 10
3 11
```

pb8o (dismissed):

tests/framework/microvm.py
roypat: ahhh, yes, good catch! Fixed!

pb8o (approved):

bchalios (approved):

PR #4170, opened 2023-10-12
Vhost user block

ShadowCurse:
## Changes
This PR adds vhost-user-bock device implementation with unit tests and initial integration test.
In order to configure new device type, old `/drive` endpoint is used with new parameter `socket`.
Depending if certain fields are present in the request body, different types of device can be configured. More details in the swagger file.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following
Developer Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

---

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #4233, opened 2023-11-11
refactor(metrics): Use consistent naming for devmetric mod

sudanl0:
## Changes

- Rename block/block_metrics.rs to block/metrics.rs
- Remove module prefix from per device METRICS pool.

## Reason

- To remove redundant module names in metrics.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


roypat:
Would it make more sense to canonicalize the other way around maybe? The device name is already in the module path (`vmm::devices::net::metrics`), so by calling it `net_metrics` we'll have it in there twice (`vmm::devices::net::net_metrics`). I guess that'd mean we'd have lots of use-site renames due to all the metrics modules having the same name, though?

kalyazin (approved):

bchalios:
Actually, I do agree with @roypat on this one. I would be in favour of relying on the namespacing provided by modules to differentiate, since that's the common practice.

sudanl0:
> Would it make more sense to canonicalize the other way around maybe? The device name is already in the module path (`vmm::devices::net::metrics`), so by calling it `net_metrics` we'll have it in there twice (`vmm::devices::net::net_metrics`). I guess that'd mean we'd have lots of use-site renames due to all the metrics modules having the same name, though?

`vmm::devices::net::metrics` is what I had in mind as well but when we moving metrics to other devices we get compiler error of names being ambiguous. This was also the reason I had to rename `block::metrics` as `block::block_metrics` because compiler gave ambiguous name error.

sudanl0:
> > Would it make more sense to canonicalize the other way around maybe? The device name is already in the module path (`vmm::devices::net::metrics`), so by calling it `net_metrics` we'll have it in there twice (`vmm::devices::net::net_metrics`). I guess that'd mean we'd have lots of use-site renames due to all the metrics modules having the same name, though?
> 
> `vmm::devices::net::metrics` is what I had in mind as well but when we moving metrics to other devices we get compiler error of names being ambiguous. This was also the reason I had to rename `block::metrics` as `block::block_metrics` because compiler gave ambiguous name error.

actually let me have a look at this again.

roypat:
> > Would it make more sense to canonicalize the other way around maybe? The device name is already in the module path (`vmm::devices::net::metrics`), so by calling it `net_metrics` we'll have it in there twice (`vmm::devices::net::net_metrics`). I guess that'd mean we'd have lots of use-site renames due to all the metrics modules having the same name, though?
> 
> `vmm::devices::net::metrics` is what I had in mind as well but when we moving metrics to other devices we get compiler error of names being ambiguous. This was also the reason I had to rename `block::metrics` as `block::block_metrics` because compiler gave ambiguous name error.

Mhhh, I just used my IDE to rename `block_metrics` to just `metrics`, and everything compiles :thinking: 
I think the issue you encountered was that we used to have a bunch of glob-imports in `devices/virtio/mod.rs` that tried to dump everything from all the individual device submodules into the `virtio` namespace, which would indeed cause breakage with multiple `metrics` modules, but @ShadowCurse fixed that in https://github.com/firecracker-microvm/firecracker/pull/4170, so it shouldn't be a problem anymore

sudanl0:
> > > Would it make more sense to canonicalize the other way around maybe? The device name is already in the module path (`vmm::devices::net::metrics`), so by calling it `net_metrics` we'll have it in there twice (`vmm::devices::net::net_metrics`). I guess that'd mean we'd have lots of use-site renames due to all the metrics modules having the same name, though?
> > 
> > 
> > `vmm::devices::net::metrics` is what I had in mind as well but when we moving metrics to other devices we get compiler error of names being ambiguous. This was also the reason I had to rename `block::metrics` as `block::block_metrics` because compiler gave ambiguous name error.
> 
> Mhhh, I just used my IDE to rename `block_metrics` to just `metrics`, and everything compiles 🤔 I think the issue you encountered was that we used to have a bunch of glob-imports in `devices/virtio/mod.rs` that tried to dump everything from all the individual device submodules into the `virtio` namespace, which would indeed cause breakage with multiple `metrics` modules, but @ShadowCurse fixed that in #4170, so it shouldn't be a problem anymore

I remembered the issue but didn't know what fixed it, thanks for sharing this!
I have updated the code and also added 1 more change to be consistent with https://github.com/firecracker-microvm/firecracker/pull/4235#discussion_r1391031345

roypat (dismissed):

roypat (approved):

kalyazin (approved):

PR #4246, opened 2023-11-17
test(metrics): collect Firecracker vhost-user metrics

sudanl0:
## Changes

Add a test to collect vhost-user metrics from Firecracker.

## Reason

To monitor any regressions in vhost-user metrics over time.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- ~~[ ] User-facing changes are mentioned in `CHANGELOG.md`.

---

### DIFF
--- a/tests/integration_tests/performance/test_vhost_user_metrics.py
+++ b/tests/integration_tests/performance/test_vhost_user_metrics.py
@@ -1,0 +1,87 @@
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Tests to collect Firecracker metrics for vhost-user devices."""
+
+import time
+
+import pytest
+
+import host_tools.drive as drive_tools
+from framework.utils_drive import resize_vhost_user_drive, spawn_vhost_user_backend
+
+
+@pytest.mark.parametrize("vcpu_count", [1, 2], ids=["1vcpu", "2vcpu"])
+def test_vhost_user_block_metrics(
+    microvm_factory, guest_kernel, rootfs_ubuntu_22, vcpu_count, metrics
+):
+    """
+    This test tries to boot a VM with vhost-user-block
+    as a scratch device, resize the vhost-user scratch drive to have
+    config change notifications, collects and then uploads the related
+    vhost-user FirecrackerMetrics to Cloudwatch.
+    Having vhost-user as root device vs a scratch should not impact metrics, however,
+    we choose to have it as a scratch device because we are interested in config change
+    metrics which we cannot extract when vhost-user is root device
+    (read only rootfs won't have a config change).
+    """
+    orig_size = 10  # MB
+    # Picked from test_config_change assuming that the intention is to change size from
+    # low->high->low->high and so the numbers are not in monotonic sequence.
+    new_sizes = [20, 10, 30]  # MB
+    vhost_user_socket = "/vub.socket"
+
+    vm = microvm_factory.build(guest_kernel, rootfs_ubuntu_22, monitor_memory=False)
+    vm.spawn(log_level="Info")
+    vm.basic_config(vcpu_count=vcpu_count)
+
+    # Add a block device to test resizing.
+    fs = drive_tools.FilesystemFile(size=orig_size)
+    _backend = spawn_vhost_user_backend(vm, fs.path, vhost_user_socket)
+    vm.add_vhost_user_drive("scratch", vhost_user_socket)
+    vm.start()
+
+    # vhost-user-block is activated during boot but it takes a while so we wait.
+    # 300msec picked by the limited number of experiments tried to see how long
+    # it takes to get the activate_time_us metrics.
+    time.sleep(0.3)
+
+    metrics.set_dimensions(
+        {
+            "performance_test": "vhost_user_block_metrics",
+            "io_engine": "vhost-user",
+            **vm.dimensions,
+        }
+    )
+    fc_metrics = vm.flush_metrics()
+    assert 0 == fc_metrics["vhost_user_block_scratch"]["activate_fails"]
+    assert fc_metrics["vhost_user_block_scratch"]["init_time_us"]
+    assert fc_metrics["vhost_user_block_scratch"]["activate_time_us"]
+
+    metrics.put_metric(
+        "init_time_us",
+        fc_metrics["vhost_user_block_scratch"]["init_time_us"],
+        unit="Microseconds",
+    )
+    metrics.put_metric(
+        "activate_time_us",
+        fc_metrics["vhost_user_block_scratch"]["activate_time_us"],
+        unit="Microseconds",
+    )
+
+    for new_size in new_sizes:
+        # Instruct the backend to resize the device.
+        # It will both resize the file and update its device config.
+        resize_vhost_user_drive(vm, new_size)
+
+        # Instruct Firecracker to reread device config and notify
+        # the guest of a config change.
+        vm.patch_drive("scratch")
+
+        fc_metrics = vm.flush_metrics()
+        assert 0 == fc_metrics["vhost_user_block_scratch"]["cfg_fails"]
+        assert fc_metrics["vhost_user_block_scratch"]["config_change_time_us"]
+        metrics.put_metric(
+            "config_change_time_us",
+            fc_metrics["vhost_user_block_scratch"]["config_change_time_us"],
+            unit="Microseconds",
+        )


sudanl0:
Blocked on https://github.com/firecracker-microvm/firecracker/pull/4223 to add 1 more metric

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -393,3 +395,81 @@ def test_block_vhost_user_performance(
...
+
+    # Magic number 5 picked referring to test_memory_overhead.py
+    # because it helps in ABfying the test
+    for _ in range(5):
+        # logic to test vhost-use picked from test_vhost_user_block()
+        vm = microvm_factory.build(guest_kernel, None, monitor_memory=False)
+
+        # Converting path from tmpfs ("./srv/..") to local
+        # path on the host ("../build/..")
+        rootfs_path = to_local_dir_path(str(rootfs))
```
kalyazin, line 418: this is being reworked in https://github.com/firecracker-microvm/firecracker/pull/4247

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -393,3 +395,81 @@ def test_block_vhost_user_performance(
         fio_block_size,
         vm,
     )
+
+
+@pytest.mark.nonci
+@pytest.mark.parametrize("vcpu_count", [1, 2], ids=["1vcpu", "2vcpu"])
+def test_vhost_user_block(microvm_factory, guest_kernel, rootfs, vcpu_count, metrics):
+    """
+    This test simply tries to boot a VM with
```
kalyazin, line 404: this is no longer the purpose of the test, right?

tests/integration_tests/performance/test_block_performance.py
```diff
@@ -393,3 +395,81 @@ def test_block_vhost_user_performance(
         fio_block_size,
         vm,
     )
+
+
+@pytest.mark.nonci
+@pytest.mark.parametrize("vcpu_count", [1, 2], ids=["1vcpu", "2vcpu"])
```
kalyazin, line 401: do you think we need both cases here?

tests/integration_tests/performance/test_block_performance.py
sudanl0: removed.

tests/integration_tests/performance/test_block_performance.py
sudanl0: I took the reference from the other performance tests but initial run with this shows there is a small difference between metrics with 1 and 2 vcpus.

tests/integration_tests/performance/test_block_performance.py
sudanl0: updated the comment.

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -240,3 +247,104 @@ def test_block_vhost_user_performance(
...
+    assert stderr == ""
+    assert stdout.strip() == str(size)
+
+
+@pytest.mark.nonci
+@pytest.mark.parametrize("vcpu_count", [1, 2], ids=["1vcpu", "2vcpu"])
+def test_vhost_user_block(microvm_factory, guest_kernel, rootfs, vcpu_count, metrics):
+    """
+    This test simply tries to boot a VM with vhost-user-block
+    as a root device and upload FirecrackerMetrics emitted by
```
kalyazin, line 266: we don't use vhost-user-blk as a root here, we use it as a scratch drive instead.

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -240,3 +247,104 @@ def test_block_vhost_user_performance(
...
+    Checks the size of the block device.
+    """
+    _, stdout, stderr = ssh_connection.run("blockdev --getsize64 {}".format(dev_path))
+    assert stderr == ""
+    assert stdout.strip() == str(size)
+
+
+@pytest.mark.nonci
+@pytest.mark.parametrize("vcpu_count", [1, 2], ids=["1vcpu", "2vcpu"])
+def test_vhost_user_block(microvm_factory, guest_kernel, rootfs, vcpu_count, metrics):
```
kalyazin, line 263: is this a good place for this test? does it do any A/B testing or does it only emit metrics for our tracking?

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -240,3 +247,104 @@ def test_block_vhost_user_performance(
...
+
+        # Check that guest reports correct new size.
+        _check_block_size(vm.ssh, "/dev/vdb", new_size * 1024 * 1024)
+
+        # Check that we can create a filesystem and mount it
+        ret, out, err = vm.ssh.run(mkfs_mount_cmd)
+        assert ret == 0, f"{new_size=}, {ret=}, {out=}, {err=}"
+
+        fc_metrics = vm.flush_metrics()
+        metrics.put_metric(
```
kalyazin, line 326: we don't expect for most of the metrics to change on every iteration (I guess only `cfg_fails` and `config_change_time_us` would), do we?

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -240,3 +247,104 @@ def test_block_vhost_user_performance(
...
+        # Instruct the backend to resize the device.
+        # It will both resize the file and update its device config.
+        resize_vhost_user_drive(vm, new_size)
+
+        # Instruct Firecracker to reread device config and notify
+        # the guest of a config change.
+        vm.patch_drive("scratch")
+
+        # Check that guest reports correct new size.
+        _check_block_size(vm.ssh, "/dev/vdb", new_size * 1024 * 1024)
```
kalyazin, line 319: do we actually need this duplicated check here if it is not the purpose of the test?

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -240,3 +247,104 @@ def test_block_vhost_user_performance(
...
+        "guest_kernel": guest_kernel.name,
+        "vcpus": str(vcpu_count),
+    }
+
+    metrics.set_dimensions(
+        {
+            "performance_test": "test_vhost_user_block",
+            **DIMENSIONS,
+        }
+    )
```
roypat, line 307:
```suggestion
    metrics.set_dimensions(
        {
            "performance_test": "test_vhost_user_block_metrics",
            "io_engine": "vhost-user",
            **microvm.dimensions,
        }
    )
```

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -240,3 +247,104 @@ def test_block_vhost_user_performance(
...
+@pytest.mark.parametrize("vcpu_count", [1, 2], ids=["1vcpu", "2vcpu"])
+def test_vhost_user_block(microvm_factory, guest_kernel, rootfs, vcpu_count, metrics):
+    """
+    This test simply tries to boot a VM with vhost-user-block
+    as a root device and upload FirecrackerMetrics emitted by
+    the test to Cloudwatch.
+    """
+
+    orig_size = 10  # MB
+    new_sizes = [20, 10, 30]  # MB
```
roypat, line 271: Should these be a parametrization of this test?

Also, nit: Can we sort them? e.g. [10, 20, 30]

tests/integration_tests/performance/test_block_ab.py
roypat: +1, if we don't A/B-test this, and only want metrics, this should be in the PR CI (so the metrics get emitted by our nightly runs). Whether we want to A/B-test I suppose depends on how much influence firecracker has on the metrics. If they are mostly backend metrics, then we probably shouldnt

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -240,3 +247,104 @@ def test_block_vhost_user_performance(
...
+        metrics.put_metric(
+            "activate_fails",
+            fc_metrics["vhost_user_block_scratch"]["activate_fails"],
+            unit="Count",
+        )
+        metrics.put_metric(
+            "cfg_fails",
+            fc_metrics["vhost_user_block_scratch"]["cfg_fails"],
+            unit="Count",
+        )
```
roypat, line 335: These shouldn't happen at all, right?

tests/integration_tests/performance/test_block_ab.py
sudanl0: Moved the PR to draft to avoid the confusion. I am still looking for the right place. There was some code here from the last PR so this code is here no other reason.
For the A/B question, I don't think we need to have A/B, we just need to submit metrics to CW.

tests/integration_tests/performance/test_block_ab.py
sudanl0: I just copy pasted them from another test to experiment with them and see something in CW.
But good point, I will arrange this when I clean up.

sudanl0: on second thought I think the sequence is intentional to test the effect of resize for low->high->low->high configs. I'll double confirm this.

tests/integration_tests/performance/test_block_ab.py
sudanl0: do you mean it shouldn't happen in this test or shouldn't happen at all?

roypat: sorry, I meant that this will always be zero

roypat: I think in that case, it makes more sense to have a functional test that `assert`s them to be zero

sudanl0: updated the code to assert on the failure metrics

tests/integration_tests/functional/test_drive_vhost_user.py
```diff
@@ -381,3 +384,83 @@ def test_config_change(microvm_factory, guest_kernel, rootfs):
         # Check that we can create a filesystem and mount it
         ret, out, err = vm.ssh.run(mkfs_mount_cmd)
         assert ret == 0, f"{new_size=}, {ret=}, {out=}, {err=}"
+
+
+@pytest.mark.parametrize("vcpu_count", [1, 2], ids=["1vcpu", "2vcpu"])
+def test_vhost_user_block_metrics(
```
pb8o, line 390: Should the test be in performance instead of functional? Otherwise it will run in parallel with a lot of other stuff and the metrics may be noisy.

tests/integration_tests/functional/test_drive_vhost_user.py
```diff
@@ -381,3 +384,83 @@ def test_config_change(microvm_factory, guest_kernel, rootfs):
...
+    vm.start()
+
+    dimensions = {
+        "vcpus": str(vcpu_count),
+        "cpu_model": global_props.cpu_model,
+        "host_os": global_props.host_os,
+        "host_kernel": "linux-" + global_props.host_linux_version,
+        "guest_kernel": guest_kernel.name,
+        "instance": global_props.instance,
+    }
```
pb8o, line 425: There's a `vm.dimensions()` helper that does most of this.

tests/integration_tests/functional/test_drive_vhost_user.py
sudanl0: moved the test to performance and added more details in the commit message. Thanks !

tests/integration_tests/performance/test_vhost_user_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+import host_tools.drive as drive_tools
+from framework.utils_drive import resize_vhost_user_drive, spawn_vhost_user_backend
+
+
+@pytest.mark.parametrize("vcpu_count", [1, 2], ids=["1vcpu", "2vcpu"])
+def test_vhost_user_block_metrics(
+    microvm_factory, guest_kernel, rootfs_ubuntu_22, vcpu_count, metrics
+):
+    """
+    This test simply tries to boot a VM with vhost-user-block
```
kalyazin, line 16: not only does it boot a VM, it also triggers a sequence of drive resizes.

tests/integration_tests/performance/test_vhost_user_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+    as a scratch device and upload FirecrackerMetrics emitted by
+    the test to Cloudwatch.
+    Having vhost-user as root device vs a scratch should not impact metrics however,
+    we choose to have it as a scratch device because we are interested in config change
+    metrics which we cannot extract when vhost-user is root device
+    (read only rootfs won't have a config change).
+    """
+    orig_size = 10  # MB
+    # Picked from test_config_change assuming that the intention is to change size from
+    # low ->high->low->high and so the numbers are not in sequence.
```
kalyazin, line 26: nit:
```diff
-low ->high->low->high
+low->high->low->high
```

nit
```diff
-not in sequence
+not in a monotonic order (or something like that)
```

tests/integration_tests/performance/test_vhost_user_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+
+@pytest.mark.parametrize("vcpu_count", [1, 2], ids=["1vcpu", "2vcpu"])
+def test_vhost_user_block_metrics(
+    microvm_factory, guest_kernel, rootfs_ubuntu_22, vcpu_count, metrics
+):
+    """
+    This test simply tries to boot a VM with vhost-user-block
+    as a scratch device and upload FirecrackerMetrics emitted by
+    the test to Cloudwatch.
+    Having vhost-user as root device vs a scratch should not impact metrics however,
```
kalyazin, line 19: I would put a comma before "however".

tests/integration_tests/performance/test_vhost_user_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+    # Add a block device to test resizing.
+    fs = drive_tools.FilesystemFile(size=orig_size)
+    _backend = spawn_vhost_user_backend(vm, fs.path, vhost_user_socket)
+    vm.add_vhost_user_drive("scratch", vhost_user_socket)
+    vm.add_net_iface()
+    vm.start()
+
+    # Attempt to connect to the VM.
+    # Verify if guest can run commands.
+    # this also activates the scratch device.
```
kalyazin, line 43: no, the driver activates the device while probing. It occurs during boot.

tests/integration_tests/performance/test_vhost_user_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+        resize_vhost_user_drive(vm, new_size)
+
+        # Instruct Firecracker to reread device config and notify
+        # the guest of a config change.
+        vm.patch_drive("scratch")
+
+        fc_metrics = vm.flush_metrics()
+        assert 0 == fc_metrics["vhost_user_block_scratch"]["cfg_fails"]
+        assert fc_metrics["vhost_user_block_scratch"]["config_change_time_us"]
+        metrics.put_metric(
```
kalyazin, line 83: do we publish multiple datapoints per test (one per iteration)?

sudanl0: I confirmed that CW has data for metric we exported in each iteration. Its easy to check min and max by selecting the filter with the same name but checking all the value takes a bit of more effort.
We can try to make it more visible by experimenting with storage_resolution of put_metric but, since we are only interested in the average of the data and know that min/max/sum datapoints are available whenever we need to cross reference, I think its ok to leave the code as is right now.

kalyazin (approved):

pb8o (approved):

PR #4247, opened 2023-11-17
test(vhost-user-blk): switch to crosvm backend in functional tests

kalyazin:
This is to address intermittent failures related to vhost-user-blk.
Unfortunately this change cannot be split into multiple commits due to high level of entanglement of the changes.

## Changes

Switching from the Qemu to crosvm backend comes with the challenge that crosvm locks the disk file exclusively when opening it as rw (which makes sense). We have to make copies of the disk files before we give them to crosvm if we don't configure crosvm as readonly.

One of the ripples of the above is an existing race condition in the `partuuid_and_disk_path_host` fixture over the disk file kicks in.  We are removing this fixture altogether because now we use per microVM copies of the disks.

Finally, this change explicitly configures performance tests to switch to the Qemu backend to preserve the status quo.  We may switch them to the crosvm backend too at a later stage.

## Reason

The Qemu backend uses direct IO when working with the disk file which makes it scale very poorly.  That results in intermittent test failures, because our functional integration tests run in parallel, and our ssh connection timeouts expire if the microVM does not boot in time.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


ShadowCurse (approved):

sudanl0 (approved):

tests/integration_tests/functional/test_drive_vhost_user.py
```diff
@@ -207,8 +188,12 @@ def test_device_ordering(microvm_factory, guest_kernel, rootfs_ubuntu_22):
     fs2 = drive_tools.FilesystemFile(os.path.join(vm.fsfiles, "scratch2"), size=512)
     vm.add_drive("scratch2", fs2.path)
 
+    # Create a rw rootfs file that is unique to the microVM
+    rootfs_rw = Path(vm.chroot()) / "rootfs"
+    shutil.copy(rootfs_ubuntu_22, rootfs_rw)
```
bchalios, line 193: nit: Would it make sense to carve these two commands out to their own function or even fixture?

bchalios (commented):
LGTM. just a single comment for code deduplication

tests/integration_tests/functional/test_drive_vhost_user.py
kalyazin: To be honest, I don't expect for this to be ubiquitous as using vhost-user-blk rootfs doesn't seem like a separate dimension. We should be able to test most of the features with vhost-user-blk as an additional drive, plus it should be more efficient from the space point of view. So I don't see much value in it at this point, but we can reconsider it later.

bchalios (approved):

PR #4261, opened 2023-11-23
Cleanup confusion arising from mixing of unix exit codes and Rust `Result` types

roypat:
An exit code is a kind of `Result`: Either everything is `Ok()` if the exit code = 0, or some `Err()` happened if exit code != 0. Therefore, nesting `FcExitCode` inside the `Err` variant of a Rust `Result` yields a construct of `Result<_, Result<_, _>>`. This is no well-defined (as it is unclear what `Err(Ok(...))` is supposed to mean). This was made worse by the existence of `ApiServerError::MicroVMStoppedWithoutError`, which turned `ApiServerError` into a kind of `Result` itself. Since `MicroVmStoppedWithoutError` has a field of type `FcExitCode`, we ended up with a construct of `Result<_, Result<Result<_, _>, _>`, or, in terms of user visible error messages: "RunWithApiError error: MicroVMStopped without an error: GenericError". The confusion around this lead to bugs such as #4176. 

This PR attempts to clear up this confusion by removing `ApiServer::Error::MicroVMSteppedWithoutError`, and eliminating the use of `FcExitCode` as an `Err` type wherever possible. We are only left with a single instance of this (`ApiServerError::MicroVmStoppedWithError`), which cannot be resolved without significantly refractoring (and eliminating the use of exit codes from the emulation code).

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -109,7 +109,7 @@ pub mod vstate;
 use std::collections::HashMap;
 use std::io;
 use std::os::unix::io::AsRawFd;
-use std::sync::mpsc::{RecvTimeoutError, TryRecvError};
+use std::sync::mpsc::RecvTimeoutError;
 use std::sync::{Arc, Barrier, Mutex};
 use std::time::Duration;
 
@@ -905,23 +905,27 @@ impl MutEventSubscriber for Vmm {
             // Exit event handling should never do anything more than call 'self.stop()'.
             let _ = self.vcpus_exit_evt.read();
 
-            let mut exit_code = None;
-            // Query each vcpu for their exit_code.
-            for handle in &self.vcpus_handles {
-                match handle.response_receiver().try_recv() {
-                    Ok(VcpuResponse::Exited(status)) => {
-                        exit_code = Some(status);
-                        // Just use the first encountered exit-code.
-                        break;
-                    }
-                    Ok(_response) => {} // Don't care about these, we are exiting.
-                    Err(TryRecvError::Empty) => {} // Nothing pending in channel
-                    Err(err) => {
-                        panic!("Error while looking for VCPU exit status: {}", err);
+            let exit_code = 'exit_code: {
+                // Query each vcpu for their exit_code.
+                for handle in &self.vcpus_handles {
+                    // Drain all vcpu responses that are pending from this vcpu until we find an
+                    // exit status.
+                    for response in handle.response_receiver().try_iter() {
+                        if let VcpuResponse::Exited(status) = response {
+                            // It could be that some vcpus exited successfully while others
+                            // errored out. Thus make sure that error exits from one vcpu always
+                            // takes precedence over "ok" exits
+                            if status != FcExitCode::Ok {
+                                break 'exit_code status;
+                            }
+                        }
                     }
                 }
-            }
-            self.stop(exit_code.unwrap_or(FcExitCode::Ok));
+
+                // No CPUs exited with error status code, report "Ok"
+                FcExitCode::Ok
+            };
+            self.stop(exit_code);
         } else {
             error!("Spurious EventManager event for handler: Vmm");
         }
--- a/src/vmm/src/rpc_interface.rs
+++ b/src/vmm/src/rpc_interface.rs
@@ -20,7 +20,7 @@ use super::{
 };
 use crate::builder::StartMicrovmError;
 use crate::cpu_config::templates::{CustomCpuTemplate, GuestConfigError};
-use crate::logger::{error, info, warn, LoggerConfig, *};
+use crate::logger::{info, warn, LoggerConfig, *};
 use crate::mmds::data_store::{self, Mmds};
 use crate::persist::{CreateSnapshotError, RestoreFromSnapshotError, VmInfo};
 use crate::resources::VmmConfig;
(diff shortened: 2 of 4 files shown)


src/vmm/src/rpc_interface.rs
```diff
@@ -333,16 +344,9 @@ impl<'a> PrebootApiController<'a> {
...
-                    serde_json::from_str(data)
-                        .expect("MMDS error: metadata provided not valid json"),
-                )
-                .map_err(|err| {
-                    error!("Populating MMDS from file failed: {:?}", err);
-                    crate::FcExitCode::GenericError
-                })?;
+            vm_resources.locked_mmds_or_default().put_data(
+                serde_json::from_str(data).expect("MMDS error: metadata provided not valid json"),
+            )?;
```
JonathanWoollett-Light, line 349: This change removes the `error!("Populating MMDS from file failed: {:?}", err);` log message.

This should be more explicitly documented.

src/vmm/src/lib.rs
```diff
@@ -908,17 +908,19 @@ impl MutEventSubscriber for Vmm {
...
+                // Drain all vcpu responses that are pending from this vcpu until we find an
+                // exit status.
+                for response in handle.response_receiver().try_iter() {
+                    if let VcpuResponse::Exited(status) = response {
+                        // It could be that some vcpus exited successfully while other's
+                        // errored out. Thus make sure that error exits from one vcpu always
+                        // takes precedence over "ok" exits exit_code.is_none()
+                        if exit_code.is_none() || exit_code == Some(FcExitCode::Ok) {
+                            exit_code = Some(status);
+                        }
```
JonathanWoollett-Light, line 920: nit
```suggestion
                        if matches!(exit_code, None | Some(FcExitCode::Ok) {
                            exit_code = Some(status);
                        }
```

src/vmm/src/rpc_interface.rs
roypat: The log message is still there, its just implicit now. Its part of the error types `Display` implementation, so we get it when when `main` logs the error. We have a test that verifies this exact log message's presence, which still passes

bchalios: But this doesn't appear as a log anymore, right? as in:

```
2023-11-23T14:55:39.305207904 [anonymous-instance:main] Populating MMDS from file failed ...
```

I'm not sure whether this is a problem, though.

roypat: It appears as an `error!` log, with slightly different formatting (because it also prints all the encapsulating errors now):
```
 2023-11-23T13:19:32.079476801 [0c631bdd-4617-46d8-a62f-c68da695bce7:main] RunWithApiError error: Failed to build MicroVM: Populating MMDS from file failed: ...
```

bchalios: oh ok! awesome. I thought `.expect` would bypass the logger.

roypat: The `.expect` was there before, rustfmt just moved it :)

JonathanWoollett-Light: If we have a test that covers this I am satisfied with this (also considering we have previously noted we do not uphold the same stringent stability guarantees for log messages).

JonathanWoollett-Light (approved):

src/vmm/src/lib.rs
```diff
@@ -908,17 +908,19 @@ impl MutEventSubscriber for Vmm {
...
+                        // We do not care about any other messages from this vcpu
                         break;
```
sudanl0, line 932: I think this breaks on the 1st iteration of the for loop so we don't get to apply the logic of capturing the errors from other vcpus.

src/vmm/src/lib.rs
```diff
@@ -908,17 +908,19 @@ impl MutEventSubscriber for Vmm {
...
-                        exit_code = Some(status);
-                        // Just use the first encountered exit-code.
+                // Drain all vcpu responses that are pending from this vcpu until we find an
+                // exit status.
+                for response in handle.response_receiver().try_iter() {
+                    if let VcpuResponse::Exited(status) = response {
+                        // It could be that some vcpus exited successfully while other's
+                        // errored out. Thus make sure that error exits from one vcpu always
+                        // takes precedence over "ok" exits exit_code.is_none()
+                        if exit_code.is_none() || exit_code == Some(FcExitCode::Ok) {
```
sudanl0, line 928: if the intention is to capture the 1st error exit code can we break the loop once we find it?
If the 1st vcpu reported an error then as per current logic we don't change the exit code so we might as well break the loop there.

roypat: Mh, yeah, that does make the code a bit simpler, thanks!

src/vmm/src/lib.rs
roypat: Not really sure what you mean here :/

src/vmm/src/lib.rs
roypat: This code no longer exists

src/vmm/src/lib.rs
roypat: Done :)

src/vmm/src/lib.rs
sudanl0: ah nvm, there was a comment which clouded my vision :P and I thought we break from the for loop.

sudanl0 (approved):

ShadowCurse (approved):

PR #4264, opened 2023-11-23
test(vhost-user-blk): move backend management to microvm class

kalyazin:
## Changes

This change creates separate classes for Qemu and crosvm backends and moves management of backend processes to the Microvm class.

## Reason

The following problems are being solved by this change:
- proper cleanup of sockets related to vhost-user-blk
- fixing name conflict for crosvm backend socket

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -38,6 +38,7 @@ from framework.http_api import Api
 from framework.jailer import JailerContext
 from framework.microvm_helpers import MicrovmHelpers
 from framework.properties import global_props
+from framework.utils_drive import VhostUserBlkBackend, VhostUserBlkBackendType
 from host_tools.memory import MemoryMonitor
 
 LOG = logging.getLogger("microvm")
@@ -209,6 +210,7 @@ class Microvm:
         # device dictionaries
         self.iface = {}
         self.disks = {}
+        self.disks_vhost_user = {}
         self.vcpus_count = None
         self.mem_size_bytes = None
 
@@ -228,6 +230,13 @@ class Microvm:
     def kill(self):
         """All clean up associated with this microVM should go here."""
         # pylint: disable=subprocess-run-check
+
+        # We start with vhost-user backends,
+        # because if we stop Firecracker first, the backend will want
+        # to exit as well and this will cause a race condition.
+        for backend in self.disks_vhost_user.values():
+            backend.kill()
+        self.disks_vhost_user.clear()
 
         if (
             self.expect_kill_by_signal is False
@@ -466,6 +475,8 @@ class Microvm:
     def pin_threads(self, first_cpu):
         """
         Pins all microvm threads (VMM, API and vCPUs) to consecutive physical cpu core, starting with "first_cpu"
+
+        Return next "free" cpu core.
         """
         for vcpu, pcpu in enumerate(range(first_cpu, first_cpu + self.vcpus_count)):
             assert self.pin_vcpu(
@@ -480,6 +491,8 @@ class Microvm:
         assert self.pin_api(
             first_cpu + self.vcpus_count + 1
         ), "Failed to pin fc_api thread."
+
+        return first_cpu + self.vcpus_count + 2
 
     def spawn(
         self,
@@ -704,12 +717,27 @@ class Microvm:
     def add_vhost_user_drive(
         self,
         drive_id,
-        socket,
+        path_on_host,
         partuuid=None,
         is_root_device=False,
+        is_read_only=False,
         cache_type=None,
+        backend_type=VhostUserBlkBackendType.CROSVM,
     ):
         """Add a vhost-user block device."""
+
+        # It is possible that the user adds another drive
+        # with the same ID. In that case, we should clean
+        # the previous backend up first.
+        prev = self.disks_vhost_user.pop(drive_id, None)
+        if prev:
+            prev.kill()
+
+        backend = VhostUserBlkBackend.with_backend(
+            backend_type, path_on_host, self.chroot(), drive_id, is_read_only
+        )
+
+        socket = backend.spawn(self.jailer.uid, self.jailer.gid)
 
         self.api.drive.put(
             drive_id=drive_id,
@@ -718,6 +746,8 @@ class Microvm:
             is_root_device=is_root_device,
             cache_type=cache_type,
         )
+
+        self.disks_vhost_user[drive_id] = backend
 
     def patch_drive(self, drive_id, file=None):
         """Modify/patch an existing block device."""
--- a/tests/framework/utils_drive.py
+++ b/tests/framework/utils_drive.py
@@ -6,6 +6,7 @@
 import os
 import subprocess
 import time
+from abc import ABC, abstractmethod
 from enum import Enum
 from pathlib import Path
 from subprocess import check_output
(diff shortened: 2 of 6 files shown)


tests/framework/microvm.py
```diff
@@ -460,9 +469,17 @@ def pin_api(self, cpu_id: int):
             return True
         return False
 
+    def pin_vhost_user_blk_backend(self, socket, cpu_id: int):
```
pb8o, line 472: Could this move to the `VhostUserBlkBackend` class?

tests/framework/microvm.py
```diff
@@ -852,6 +871,41 @@ def ssh(self):
...
+
+        return socket
+
+    def resize_vhost_user_drive(self, socket, new_size):
+        """Resize a vhost-user-blk drive."""
+        self.vhost_user_blk_backends[socket].resize(new_size)
+
+    def kill_vhost_user_blk_backend(self, socket):
+        """Kill a vhost-user-blk backend."""
+        del self.vhost_user_blk_backends[socket]
```
pb8o, line 907: I think both of those should be accessed as interface below:

```python
uvm.drive_vhost["drive-id"].kill()
uvm.drive_vhost["drive-id"].resize()
```

tests/integration_tests/functional/test_api.py
```diff
@@ -749,11 +749,9 @@ def test_drive_patch(test_microvm_with_api):
...
         os.path.join(test_microvm.fsfiles, "scratch_vub")
     )
-    vhost_user_socket = "/vub.socket"
+
     # Launching vhost-user-block backend
-    _backend = spawn_vhost_user_backend(
-        test_microvm, fs_vub.path, vhost_user_socket, False
-    )
+    vhost_user_socket = test_microvm.spawn_vhost_user_blk_backend(fs_vub.path, False)
     test_microvm.add_vhost_user_drive("scratch_vub", vhost_user_socket)
```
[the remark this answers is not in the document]
kalyazin: I had this idea too, but the following stopped me (perhaps unjustifiably):
- we have tests with vhost-user as a root device and it requires some special treatment
- we have a test where we want to stop the backend independently and see if FC is still alive. This may not be transparent with that API.
- we have a test with multiple puts that may require additional handling on the uvm side. If we use the drive_id as the socket name, this would require stopping the previous backend first, which is not the same as FC disconnects from one socket and connects to another.

tests/framework/microvm.py
```diff
@@ -852,6 +871,41 @@ def ssh(self):
...
+        if backend == VhostUserBlkBackendType.QEMU:
+            backend = QemuVhostUserBlkBackend(
+                host_mem_path, self.chroot(), backend_id, readonly
+            )
+        elif backend == VhostUserBlkBackendType.CROSVM:
+            backend = CrosvmVhostUserBlkBackend(
+                host_mem_path, self.chroot(), backend_id, readonly
+            )
+        else:
+            assert False, f"unknown vhost-user-blk backend type `{backend}`"
```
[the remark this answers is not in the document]
kalyazin: what would be the saving from this approach? The following 2 lines will be 1 or any other advantages?
```
        if backend == VhostUserBlkBackendType.QEMU:
            backend = QemuVhostUserBlkBackend(
```

tests/framework/utils_drive.py
```diff
@@ -46,78 +46,119 @@ def partuuid_and_disk_path(rootfs_ubuntu_22, disk_path):
...
+        return jailed_socket_path
+
+    def spawn_cmd(self):
+        """Return a spawn command for the backend"""
+        return ""
+
+    def resize(self, new_size):
+        """Resize the vhost-user-backed drive"""
+
+    def __del__(self):
```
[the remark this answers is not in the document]
kalyazin: what is the specific problem with `__del__`? Is the convention that it should do nothing?

tests/framework/microvm.py
kalyazin: but there should still be a method in `Microvm` to wrap it?

tests/framework/microvm.py
kalyazin: what do you mean by `drive_vhost`? Would it only contain these two methods? Would it be unimplemented by virtio-backed drives?

tests/integration_tests/functional/test_api.py
```diff
@@ -749,11 +749,9 @@ def test_drive_patch(test_microvm_with_api):
     fs_vub = drive_tools.FilesystemFile(
         os.path.join(test_microvm.fsfiles, "scratch_vub")
     )
-    vhost_user_socket = "/vub.socket"
+
```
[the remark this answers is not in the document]
kalyazin: yeah, that would require a substantial change and prior design

tests/framework/utils_drive.py
```diff
@@ -46,78 +46,119 @@ def partuuid_and_disk_path(rootfs_ubuntu_22, disk_path):
...
 
-    uid = vm.jailer.uid
-    gid = vm.jailer.gid
-    host_vhost_user_socket_path = Path(vm.chroot()) / socket_path.strip("/")
+        # Give the backend time to initialise.
+        time.sleep(1)
 
-    if backend == VhostUserBlkBackendType.QEMU:
+        assert proc is not None and proc.poll() is None, "backend is not up"
+        assert os.path.exists(self.socket_path)
```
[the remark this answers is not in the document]
kalyazin: done, thanks

tests/framework/utils_drive.py
```diff
@@ -46,78 +46,119 @@ def partuuid_and_disk_path(rootfs_ubuntu_22, disk_path):
...
+
+        jailed_socket_path = str(Path("/") / os.path.basename(self.socket_path))
+
+        with utils.chroot(self.chroot):
+            # The backend will create the socket path with root rights.
+            # Change rights to the jailer's.
+            os.chown(jailed_socket_path, uid, gid)
```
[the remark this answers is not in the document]
kalyazin: I wasn't able to find `chown()` method for PosixPath, but works without chrooting, thanks.

tests/framework/utils_drive.py
```diff
@@ -46,78 +46,119 @@ def partuuid_and_disk_path(rootfs_ubuntu_22, disk_path):
     return (partuuid, disk_path)
 
 
-CROSVM_CTR_SOCKET_NAME = "crosvm_ctr.socket"
+class VhostUserBlkBackend:
```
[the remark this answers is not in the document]
kalyazin: done

tests/framework/utils_drive.py
```diff
@@ -46,78 +46,119 @@ def partuuid_and_disk_path(rootfs_ubuntu_22, disk_path):
...
+        with utils.chroot(self.chroot):
+            # The backend will create the socket path with root rights.
+            # Change rights to the jailer's.
+            os.chown(jailed_socket_path, uid, gid)
+
+        self.proc = proc
+
+        return jailed_socket_path
+
+    def spawn_cmd(self):
```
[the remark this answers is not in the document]
kalyazin: added

tests/framework/utils_drive.py
```diff
@@ -46,78 +46,119 @@ def partuuid_and_disk_path(rootfs_ubuntu_22, disk_path):
...
+        """Return a spawn command for the backend"""
+        return ""
+
+    def resize(self, new_size):
```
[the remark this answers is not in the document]
kalyazin: added

tests/framework/microvm.py
ShadowCurse: Maybe we can expose `vhost_user_blk_backends` as a member? So pinning will be: `vm.vu_block_backends[socket].pin(...)` So `vm` will own the backends, but we can access them anyway to configure them.

tests/framework/microvm.py
```diff
@@ -261,6 +268,8 @@ def kill(self):
                 self.memory_monitor.join(timeout=1)
             self.memory_monitor.check_samples()
 
+        self.vhost_user_blk_backends.clear()
```
ShadowCurse, line 271: Will this call `kill` on all backends? Shouldn't this be a loop?

tests/framework/microvm.py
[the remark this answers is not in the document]
pb8o: it should just be:

```python
backend = VhostUserBlkBackend.with_backend(backend, host_mem_path, self.chroot(), backend_id, readonly)
```

It also separates concerns between Microvm and VhostUserBlkBackend. `Microvm` is too much of a god class already :).

tests/framework/microvm.py
kalyazin: I don't particularly like this for 2 reasons:
- this will strengthen utils_drive.py dependency on microvm.py
- this may invite accessing the member in other places and make it harder to keep track

What is it specifically that don't you like about `pin_vhost_user_blk_backend()` being in the microvm.py?

tests/framework/microvm.py
kalyazin: ok, that's a leftover from before I replaced `__del__` with `kill`. Going to fix, thanks for spotting.

tests/framework/utils_drive.py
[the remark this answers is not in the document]
kalyazin: done

tests/framework/microvm.py
kalyazin: fixed

tests/framework/microvm.py
```diff
@@ -852,6 +871,42 @@ def ssh(self):
...
+        return socket
+
+    def resize_vhost_user_drive(self, socket, new_size):
+        """Resize a vhost-user-blk drive."""
+        self.vhost_user_blk_backends[socket].resize(new_size)
+
+    def kill_vhost_user_blk_backend(self, socket):
+        """Kill a vhost-user-blk backend."""
+        self.vhost_user_blk_backends[socket].kill()
+        del self.vhost_user_blk_backends[socket]
```
pb8o, line 908: These look like just shortcuts and can just be referenced via:

```python
microvm.vhost_user_blk_backends[socket].resize(...)
```

tests/framework/microvm.py
pb8o: My 2c is we should avoid provide wrapper methods that don't add some convenience/value. In this case it just forwards the function call. Ditto for resize_vhost_user_drive and kill_vhost_user_blk_backend.

tests/framework/utils_drive.py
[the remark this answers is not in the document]
pb8o: heh, I keep forgetting some methods don't exist.

tests/framework/microvm.py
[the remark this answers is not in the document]
kalyazin: @pb8o 
> W0212: Access to a protected member _get_all_subclasses of a client class (protected-access)

Do you know how to overcome this?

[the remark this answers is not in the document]
kalyazin: done

tests/framework/microvm.py
kalyazin: done

tests/framework/microvm.py
kalyazin: done

tests/integration_tests/functional/test_api.py
```diff
@@ -931,7 +926,12 @@ def _drive_patch(test_microvm):
             "path_on_host": None,
             "rate_limiter": None,
             "io_engine": None,
-            "socket": "/vub.socket",
+            "socket": str(
+                Path("/")
+                / os.path.basename(
+                    test_microvm.disks_vhost_user["scratch_vub"].socket_path
```
pb8o, line 932: `socket_path` is a `Path`, no? It can then be `Path("/") / test_microvm.disks_vhost_user["scratch_vub"].socket_path.name`

pb8o (dismissed):
Looks great! Just a small nit, feel free to ignore.

tests/integration_tests/functional/test_api.py
kalyazin: the reason I use `basename` is because the `socket_path` is host-side view, not Firecracker/jailer view.

bchalios (approved):

tests/framework/microvm.py
```diff
@@ -229,6 +231,13 @@ def kill(self):
         """All clean up associated with this microVM should go here."""
         # pylint: disable=subprocess-run-check
 
+        # We start with vhost-user backends,
+        # because if we stop Firecracker first, the backend will want
+        # to exit as well and this will cause a race condition.
+        for backend in self.disks_vhost_user.values():
+            backend.kill()
+        self.disks_vhost_user.clear()
+
```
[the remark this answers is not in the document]
kalyazin: we are testing it here: https://github.com/kalyazin/firecracker/blob/vub_be_mgmt/tests/integration_tests/functional/test_drive_vhost_user.py#L114

tests/framework/microvm.py
```diff
@@ -704,13 +717,28 @@ def add_drive(
     def add_vhost_user_drive(
         self,
         drive_id,
-        socket,
+        host_mem_path,
```
ShadowCurse, line 720: nit: shouldn't this be just a `file_path`/`disk_path`? `host_mem_path` reminds me of the `memfd`.

tests/framework/utils_drive.py
```diff
@@ -46,78 +47,141 @@ def partuuid_and_disk_path(rootfs_ubuntu_22, disk_path):
...
-    assert os.path.exists(
-        crosvm_ctr_socket_path
-    ), "crosvm backend must be spawned first"
-
-    utils.run_cmd(
-        f"crosvm disk resize 0 {new_size * 1024 * 1024} {crosvm_ctr_socket_path}"
-    )
+    def kill(self):
+        super().kill()
+        assert self.proc.poll() is None
```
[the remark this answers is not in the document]
ShadowCurse: Maybe worth doing a small loop/timeout here to verify the backend is killed? I feel this can cause a lot of "just rerun the test" situations.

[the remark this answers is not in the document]
kalyazin: ok, it looks like I found a flaw in the logic. Thanks for raising.

tests/integration_tests/functional/test_api.py
pb8o: But `os.path.basename` and `Path.name` should return the same thing no? https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

kalyazin: I misunderstood you, done!

tests/framework/microvm.py
kalyazin: updated `path_on_host` like in `add_drive()`

tests/framework/utils_drive.py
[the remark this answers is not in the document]
kalyazin: so looks like `popen.wait()` is the right thing to do here. I also fixed an incorrect check for `poll()` result later on.

PR #4278, opened 2023-11-28
fix: intermittent vsock timeout in snapshot tests

sudanl0:
Instead of creating a new socat process on every guest resume, create socat process in the 1st guest created so that the snapshotted guests just use the same process through.
Suspend and resume socat process before taking snapshot and after resuming from snapshot respectively to make sure the process is in the right state for the snapshot.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tests/integration_tests/functional/test_snapshot_basic.py
+++ b/tests/integration_tests/functional/test_snapshot_basic.py
@@ -112,6 +112,7 @@ def test_5_snapshots(
 
     logger.info("Create %s #0.", snapshot_type)
     # Create a snapshot from a microvm.
+    start_guest_echo_server(vm)
     snapshot = vm.make_snapshot(snapshot_type)
     base_snapshot = snapshot
 
@@ -120,17 +121,23 @@ def test_5_snapshots(
         microvm = microvm_factory.build()
         microvm.spawn()
         microvm.restore_from_snapshot(snapshot, resume=True)
+        # TODO: SIGCONT here and SIGSTOP later before creating snapshot
+        # is a temporary fix to avoid vsock timeout in
+        # _vsock_connect_to_guest(). This will be removed once we
+        # find the right solution for the timeout.
+        vm.ssh.run("pkill -SIGCONT socat")
         # Test vsock guest-initiated connections.
         path = os.path.join(
             microvm.path, make_host_port_path(VSOCK_UDS_PATH, ECHO_SERVER_PORT)
         )
         check_guest_connections(microvm, path, vm_blob_path, blob_hash)
         # Test vsock host-initiated connections.
-        path = start_guest_echo_server(microvm)
+        path = os.path.join(microvm.jailer.chroot_path(), VSOCK_UDS_PATH)
         check_host_connections(path, blob_path, blob_hash)
 
         # Check that the root device is not corrupted.
         check_filesystem(microvm.ssh, "squashfs", "/dev/vda")
+        vm.ssh.run("pkill -SIGSTOP socat")
 
         logger.info("Create snapshot %s #%d.", snapshot_type, i + 1)
         snapshot = microvm.make_snapshot(snapshot_type)


tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -120,17 +121,19 @@ def test_5_snapshots(
         microvm = microvm_factory.build()
         microvm.spawn()
         microvm.restore_from_snapshot(snapshot, resume=True)
+        vm.ssh.run("pkill -SIGCONT socat")
```
bchalios, line 124: Do we really need this SIGSTOP/SIGCONT pair?

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -120,17 +121,19 @@ def test_5_snapshots(
...
         )
         check_guest_connections(microvm, path, vm_blob_path, blob_hash)
         # Test vsock host-initiated connections.
-        path = start_guest_echo_server(microvm)
+        path = os.path.join(microvm.jailer.chroot_path(), VSOCK_UDS_PATH)
         check_host_connections(path, blob_path, blob_hash)
 
         # Check that the root device is not corrupted.
         check_filesystem(microvm.ssh, "squashfs", "/dev/vda")
+        vm.ssh.run("pkill -SIGSTOP socat")
```
pb8o, line 136: Does `socat` do something special regarding `STOP`/`CONT`? AFAICS `socat` would be waiting on an `accept()` system call, so how are STOP/CONT making a difference?

sudanl0: the socket connection is reset by Firecracker while creating a snapshot so it seems to leave socat in a weird state and without stop/cont 'ng socat before/after snapshot we end up in a vsock timeout when host tries to connect to guest. The PR unblocks the pipeline but we still need to dig deeper separately.

tests/integration_tests/functional/test_snapshot_basic.py
sudanl0: https://github.com/firecracker-microvm/firecracker/pull/4278#discussion_r1407868707

tests/integration_tests/functional/test_snapshot_basic.py
bchalios: I can confirm what sudanl0@ says.

Here is a run with the patch of Sudan but without SIGSTOP/SIGCONT pair: https://buildkite.com/firecracker/bchalios-ad-hoc-tests/builds/38#_
And here is with original patch: https://buildkite.com/firecracker/bchalios-ad-hoc-tests/builds/37

It looks as if vsock gets completely broken if we snapshot while something it's using it. I agree with Sudan we should merge this to fix the test and look into the vsock issue(?) separately.

bchalios (approved):

tests/integration_tests/functional/test_snapshot_basic.py
pb8o: So is the issue lingering `socat` processes from previous runs? What happens if we kill those processes after the checks?

pb8o: I mean, it is a great find that the STOP/CONT makes it work. I am fine approving the PR, but we should understand why this works.

Could we add a comment saying why we do STOP/CONT, and that we don't know why it works? Maybe a `TODO`?

bchalios: I agree with adding a TODO and opening an issue for investigating

sudanl0: Added a TODO but I would like to use existing ticket to debug the issue further instead of creating a new issue.

bchalios (approved):

pb8o (approved):
Great find! It looks like this will reduce the number of flaky runs significantly.

PR #4282, opened 2023-11-29
Enable entropy device for boot tests

bchalios:
## Changes

Enable entropy device during boot time tests to check for entropy effects in the boot time of microVM.

## Reason

The presence of an extra source of entropy might help decrease the boot time in some platforms where hardware instructions for getting entropy are not available (like some ARM instances).

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


tests/framework/microvm.py
```diff
@@ -620,6 +620,7 @@ def basic_config(
         track_dirty_pages: bool = False,
         rootfs_io_engine=None,
         cpu_template: Optional[str] = None,
+        enable_entropy_device = False,
```
pb8o, line 623: Are there any downsides to having it enabled by default?

tests/integration_tests/performance/test_boottime.py
```diff
@@ -129,6 +129,7 @@ def _configure_and_run_vm(microvm, network=False, initrd=False):
         "vcpu_count": 1,
         "mem_size_mib": 128,
         "boot_args": DEFAULT_BOOT_ARGS + " init=/usr/local/bin/init",
+        "enable_entropy_device": True,
```
sudanl0, line 132: This could be done in a separate PR as well but I wanted to know if it relating the bootime with entropy events would add any value?
If yes, then would it be ok to call flush_metrics and collect some entropy metrics like entropy_event_count, entropy_bytes or host_rng_fails? We can also try to assert if there was any failure detected in the entropy metrics like entropy_event_fails or activate_fails.

tests/framework/microvm.py
bchalios: Not sure if any of the tests makes assumption about which devices are enabled or not. I was thinking snapshotting tests, etc.

pb8o: Ok... my thinking is: we would expect customers to enable this most of the time. Also as a nice side-effect, faster boots could mean faster test runs :)

bchalios: It makes sense, yes. But from the point of view of testing, I prefer limiting the scope of the test to what is needed for the purposes of the test. IMHO, it leads to more stable tests.

bchalios: not strong opinions though.

tests/integration_tests/performance/test_boottime.py
sudanl0: ok so does it mean that if for some reason entropy event increase/decrease during boot then relating it to a change in boot time won't make sense at all?

tests/framework/microvm.py
sudanl0: if we ask customers to enable this then it makes sense to keep it enabled by default and to check if this has any effect on performance we could dry run AB test on it. If AB test doesn't show any regression with it being on by default then I guess it would be good to have enabled by default.

bchalios: That's what this PR does. It enables it for boot-time performance tests. I just made it optional for any other test, since it is not strictly needed for testing, say block or net devices.

tests/integration_tests/performance/test_boottime.py
bchalios: I'm saying that they do not necessarily correlate. On x86, for example there shouldn't be any difference.

pb8o (approved):

sudanl0 (approved):

Issue #4215, opened 2023-11-06
Test instrumented binary


JonathanWoollett-Light:
Add a test that checks a simple integration tests works when Firecracker is instrumented with [src/clippy-tracing](https://github.com/firecracker-microvm/firecracker/tree/main/src/clippy-tracing) and trace level logs are enabled.

MuhtasimTanmoy:
Is it active?

JonathanWoollett-Light:
> Is it active?

Yes, any contribution on this issue would be great.

LDagnachew:
Hi, we're a group of students at UT Austin looking to contribute to virtualization-related open-source projects for our class. We came across this issue and would like to work on it @JonathanWoollett-Light, if you could assign this to us please. 

Thanks so much!

JonathanWoollett-Light:
@LDagnachew Hi, I'm no longer a maintainer on this project, and I expect the view on some of these issues may have changed. I think @xmarcalx or @bchalios can provide some more insight.

LDagnachew:
Got it, thanks so much!

LDagnachew:
Hi @xmarcalx sorry for another late reply, but just wanted to update that we are currently working on it! Also, not sure if i can assign myself to the ticket currently, but regardless I'll submit a PR when I'm done.

kalyazin:
Hi @LDagnachew . The test in the `tests/integration_tests/functional/test_log_instrument.py` verifies that the `log-instrument` crate is able to instrument _a_ program. Specifically it uses a number of example programs (`src/log-instrument/examples/one.rs` to `src/log-instrument/examples/five.rs`). What we a looking in this issue is a test that would verify that `log-instrument` is able to instrument _Firecracker_ in particular. The test can run an instrumented Firecracker binary and look for meaningful calls to appear in the trace (eg processing a block request or a network packet).

demoncoder-crypto:
Please check #5256 and see if its upto expectation

PR #4283, opened 2023-11-29
Tracing fix

JonathanWoollett-Light:
## Changes

Improves tracing documentation and fixes dependency issue.

## Reason

When running `cargo build --features tracing` with `src/utils`
instrumented errors are produced due to the optional dependency not
being enabled. This makes the dependency non-optional to fix this, when
unneeded the dependency will still be ommited at compile time however
this may very slightly increase compile time.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/utils/Cargo.toml
+++ b/src/utils/Cargo.toml
@@ -18,10 +18,10 @@ versionize = "0.1.10"
 versionize_derive = "0.1.6"
 vmm-sys-util = "0.11.2"
 vm-memory = { version = "0.13.0", features = ["backend-mmap", "backend-bitmap"] }
-log-instrument = { path = "../log-instrument", optional = true }
+log-instrument = { path = "../log-instrument" }
 
 [dev-dependencies]
 serde_json = "1.0.99"
 
 [features]
-tracing = ["log-instrument"]
+tracing = []


ShadowCurse (dismissed):

src/utils/Cargo.toml
```diff
@@ -18,10 +18,10 @@ versionize = "0.1.10"
 versionize_derive = "0.1.6"
 vmm-sys-util = "0.11.2"
 vm-memory = { version = "0.13.0", features = ["backend-mmap", "backend-bitmap"] }
-log-instrument = { path = "../log-instrument", optional = true }
+log-instrument = { path = "../log-instrument" }
```
kalyazin, line 21: can we have a regression test that verifies that?

docs/tracing.md
```diff
@@ -33,16 +33,35 @@ outputting a `Trace` level log when entering and exiting every function.
...
+You can run `clippy-tracing --help` for help.
+
 To enable tracing in Firecracker, add instrumentation with:
 
 ```
 clippy-tracing \
 --action fix \
 --path ./src \
---exclude vmm_config/logger.rs,virtio/gen,bindings.rs,net/gen,benches,logger/,signal_handler.rs,time.rs
+--exclude \
```
kalyazin, line 49: could add indentation for better readability like
```bash
clippy-tracing \
  --action fix \
  --path ./src \
  --exclude \
```

docs/tracing.md
```diff
@@ -33,16 +33,35 @@ outputting a `Trace` level log when entering and exiting every function.
...
+benches,\
+virtio/gen,bindings.rs,net/gen,\
+log-instrument-macros/,log-instrument/,clippy-tracing/,\
+vmm_config/logger.rs,logger/,signal_handler.rs,time.rs
 ```
 
-and re-compile with `--features tracing`:
+`--exclude` can be used to avoid adding instrumentation to specific
+files, here it is used to avoid adding instrumentation in:
```
kalyazin, line 57: looks like 2 separate sentences.

docs/tracing.md
```diff
@@ -33,16 +33,35 @@ outputting a `Trace` level log when entering and exiting every function.
...
+
+- tests.
```
kalyazin, line 59: do we need full stops after every item?

src/utils/Cargo.toml
```diff
@@ -18,10 +18,10 @@ versionize = "0.1.10"
...
 
 [dev-dependencies]
 serde_json = "1.0.99"
 
 [features]
-tracing = ["log-instrument"]
+tracing = []
```
kalyazin, line 27: commit message `fix: Fix optional dependency issue`: would be helpful to reference log-instrument somehow in the summary.

src/utils/Cargo.toml
JonathanWoollett-Light: This would be this https://github.com/firecracker-microvm/firecracker/issues/4215 since it is a non-trivial task to write this test.

docs/tracing.md
JonathanWoollett-Light: I personally prefer it without the indents since the additional spaces are not ignored. With 1 space if it's copied and placed into 1 line it looks normal, with 4 space it can look unusual.

docs/tracing.md
JonathanWoollett-Light: To split it into 2 sentences it would need to be reworded
```
`--exclude` can be used to avoid adding instrumentation to specific
files. `--exclude` is used here it is used to avoid adding instrumentation in:
```

docs/tracing.md
JonathanWoollett-Light: I use bullet points as if they could be separate sentences.

So:
```
it is used to avoid adding instrumentation in tests.
it is used to avoid adding instrumentation in bindings.
it is used to avoid adding instrumentation in the instrumentation tooling.
```
becomes:
```
it is used to avoid adding instrumentation in [tests|bindings|the instrumentation tooling].
```
which becomes:
```
it is used to avoid adding instrumentation in:
- tests.
- bindings.
- the instrumentation tooling.
```
This makes the most sense to me, although I do not know a reference for this.

docs/tracing.md
kalyazin: > additional spaces are not ignored

not sure I understand that. Do you mean if we add spaces they are not always rendered?

> with 4 space it can look unusual.

I think 2 spaces would serve both purposes?

src/utils/Cargo.toml
kalyazin: makes sense. Would be worth adding a comment in the issue?

src/utils/Cargo.toml
JonathanWoollett-Light: I've updated the commit message to reference it:
```text
When running `cargo build --features tracing` with `src/utils`
instrumented errors are produced due to the `log-instrument` optional
dependency not being enabled by the `utils/tracing` feature. This commit
makes the dependency non-optional to fix this, when unneeded the
dependency will still be ommited at compile time however this may very
slightly increase compile time.
```

docs/tracing.md
kalyazin: > is used here it is used

This is a bit hard to parse. Also, maybe replace "here" with some more explicit reference to the piece above?

JonathanWoollett-Light: That's my mistake, updated to remove that.

docs/tracing.md
kalyazin: > This in my experience is not always consistent using the terminal, some terminals will format it as:

It doesn't look like a big concern to me as long as the command is still syntactically valid.

kalyazin (approved):

PR #4285, opened 2023-11-29
Block refactor

ShadowCurse:
## Changes
Refactoring of the block devices architecture. Return of a single `Block` type which now will have virtio or vhost-user
implementation inside itself.

## Reason
Lowers code complexity.

Currently Firecracker is designed with only `virtio` devices types in mind (i.e. net, block, vsock...). When Firecracker introduced second 'type' of block device (`vhost-user-block`) it created inconsistency and boilerplate code to distinguish between 2 block types. In order to reduce this newly introduced complexity we move back to a single `Block` type which plays nice with the rest of the system.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/vmm/src/devices/virtio/block/device.rs
```diff
@@ -15,6 +14,7 @@ use crate::devices::virtio::device::VirtioDevice;
 use crate::devices::virtio::queue::Queue;
 use crate::devices::virtio::{ActivateError, TYPE_BLOCK};
 use crate::rate_limiter::BucketUpdate;
+use crate::snapshot::Persist;
```
kalyazin, line 18: commit `refactor(block): replace persist type aliases`: would be good to have more explanation of the motivation.

src/vmm/src/devices/virtio/block/device.rs
```diff
@@ -15,25 +15,28 @@ use crate::devices::virtio::queue::Queue;
 use crate::devices::virtio::{ActivateError, TYPE_BLOCK};
 use crate::rate_limiter::BucketUpdate;
 use crate::snapshot::Persist;
+use crate::vmm_config::drive::BlockDeviceConfig;
```
kalyazin, line 19: commit `refactor(block): construct Block from BlockDeviceConfig`: can add more info on why this is simpler.

src/vmm/src/devices/virtio/block/device.rs
```diff
@@ -1,7 +1,202 @@
...
+
+    pub fn update_disk_image(&mut self, disk_image_path: String) -> Result<(), BlockError> {
+        match self {
+            Self::Virtio(b) => b.update_disk_image(disk_image_path),
+        }
+    }
+
+    pub fn update_rate_limiter(&mut self, bytes: BucketUpdate, ops: BucketUpdate) {
+        match self {
+            Self::Virtio(b) => b.update_rate_limiter(bytes, ops),
```
kalyazin: Now `Block` is aware of differences between implementations (virtio vs vhost-user). Is that what we want or would we like to encapsulate everything inside its variants?

kalyazin (commented):
Please add more information in the PR description on how specifically introducing `Block` enum reduces code complexity.

src/vmm/src/devices/virtio/block/device.rs
ShadowCurse: updated commit message

src/vmm/src/devices/virtio/block/device.rs
ShadowCurse: updated commit message

src/vmm/src/devices/virtio/block/device.rs
ShadowCurse: How would 'encapsulate everything inside its variants' look like?

src/vmm/src/devices/virtio/block/common.rs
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/devices/virtio/block/device.rs
ShadowCurse: I never considered that `vhost-user` will implement functions that do not make sense for it. I think we should keep each block type with only methods needed for it and the `Block` can handle all nuances. Also it is easier to return `BlockError` from `Block` rather than from `virtio` or `vhost-user` blocks.

bchalios (commented):
In general LGTM, however, I am trying to think if there is a way to organize the commits in a way that does not require us to disable vhost-user, and then re-enable.

I need to think over this better but what came to mind was: add the new type (the one that encapsulates the two possible backends) without plugging this with the rest of the code (having `#[allow(unused)]` should be fine for one commit. Then in later commits, do move everything in the correct place and adjust the device_manager and rpc part of the code.

As I said, I want to think about it better, but if you see something fundamentally wrong with this approach let me know. I'll take a look again tomorrow and will share my thoughts.

Also, did you run A/B performance tests on this?

src/vmm/src/devices/virtio/block/device.rs
ShadowCurse: `update_rate_limiter` does return an error. It is added with vhost-user block addition in later commits.

src/vmm/src/lib.rs
```diff
@@ -644,12 +643,13 @@ impl Vmm {
...
-            .with_virtio_device_with_id(TYPE_BLOCK, drive_id, |block: &mut VhostUserBlock| {
-                block.config_update().map_err(|err| format!("{:?}", err))
-            })
-            .map_err(VmmError::DeviceManager)
+    pub fn update_vhost_user_block_config(&mut self, _drive_id: &str) -> Result<(), VmmError> {
+        // self.mmio_device_manager
+        //     .with_virtio_device_with_id(TYPE_BLOCK, drive_id, |block: &mut VhostUserBlock| {
+        //         block.config_update().map_err(|err| format!("{:?}", err))
+        //     })
+        //     .map_err(VmmError::DeviceManager)
```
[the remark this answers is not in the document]
ShadowCurse: I left it there so I would not forget to add it back. I guess now I can clean this commit.

src/vmm/src/device_manager/persist.rs
```diff
@@ -17,6 +17,15 @@ use super::mmio::*;
 use crate::arch::DeviceType;
 use crate::devices::virtio::balloon::persist::{BalloonConstructorArgs, BalloonState};
 use crate::devices::virtio::balloon::{Balloon, BalloonError};
+use crate::devices::virtio::block::vhost_user_block::device::VhostUserBlock;
+use crate::devices::virtio::block::vhost_user_block::persist::{
+    VhostUserBlockConstructorArgs, VhostUserBlockState,
+};
+use crate::devices::virtio::block::vhost_user_block::VhostUserBlockError;
+use crate::devices::virtio::block::virtio_block::persist::{
```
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/builder.rs
```diff
@@ -935,7 +935,7 @@ pub mod tests {
 
     use super::*;
     use crate::arch::DeviceType;
-    use crate::devices::virtio::block::common::CacheType;
```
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/devices/virtio/block/device.rs
bchalios: could you please bring it over to this commit?

src/vmm/src/lib.rs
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/devices/virtio/block/device.rs
ShadowCurse: done

kalyazin (dismissed):

src/vmm/src/device_manager/persist.rs
```diff
@@ -273,13 +273,20 @@ impl<'a> Persist<'a> for MMIODeviceManager {
...
-                    states.block_devices.push(ConnectedBlockState {
-                        device_id: devid.clone(),
-                        device_state: block.save(),
-                        transport_state,
-                        device_info: device_info.clone(),
-                    })
+                    if block.is_vhost_user() {
+                        warn!(
+                            "Skipping vhost-user-block device. VhostUserBlock does not support \
+                             snapshotting yet"
```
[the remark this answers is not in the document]
ShadowCurse: I looked through the testing for the rest of the devices and and the problem is that tests utilize methods from `builder.rs` that assume usage of `Block` type (which has `VhostUser` inside which is `VhostUserBlockImpl<Frontend>`). So basically if we want to test dummy vhost blocks anywhere, we need to alter all methods to be more generic do an ugly moves with `cfg(not(test))` to make `VhostUserBlock` alias to be with mock implementation during tests.

src/vmm/src/device_manager/persist.rs
```diff
@@ -17,7 +17,6 @@ use super::mmio::*;
 use crate::arch::DeviceType;
```
bchalios, line 17: I don't understand the point of the commit `refactor(block): remove indirection in imports`.

It looks like it does two things:
1. remove `VhostUserBlock` variant from error type `DevicePersistError`.
2. it re-exports types from `virtio::device`, `virtio::metrics` and `virtio::persist` as types of the `virtio::block` module.
3. It makes modules `io` and `event_handler` public

I do not understand how 2 and 3 are useful. Actually, I think we should avoid doing 2, because we should really be able to differentiate between virtio and vhost-user cases. Regarding 3, I don't see why we should do that.

Finally, I think 1. should not be part of this commit.

src/vmm/src/devices/virtio/block/device.rs
bchalios, line 1: Could you please squash the commit `refactor(block): block update methods return results` with the commit that introduces the `Block` type?

bchalios (commented):
Left a second round of comments. In general, I think I would prefer the flow of the commits to be the following:

1. Create the module structure under `vmm::devices::virtio::block` as you do now.
2. Create the new `Block` type that encapsulates the Virtio and Vhost devices
    * Maybe you need to mark this as `#[allow(unused)]`
3. Switch everything to use the new type.

My main motivation for this flow is that it makes reviewing a lot easier IMHO and it makes easier the process of following the git history of the code base as well.

src/vmm/src/device_manager/persist.rs
ShadowCurse: 1. indeed can be done in previous commits
2. This is done to create an initial platform where `Block` will exist. As I just looked, it seems `virtio::metrics` is not used in the end, so I will remove it.
3. Initially as I remember they needed to be made public due to the changes. Seems `io` does not need to be in the end. `metrics` needs to be thought for logger

ShadowCurse: done

src/vmm/src/devices/virtio/block/device.rs
ShadowCurse: done

bchalios (approved):
Ok LGTM now, so I'm fine to approve it. I will just try to clarify my previous comment regarding the PR structure in terms of commits, mainly for future reference:

The way the PR was structured made it difficult for me to review, because:
1. It was difficult to follow the purpose of the changes.
2. There was a big re-direction in the PR with removing vhost-user-block and re-introducing it.

What the PR wanted to do (and did in the end) was to introduce a new type for block devices which encapsulates the two variants we support today (virtio and vhost-user). So, in my eyes it would be easier to:

1. create the new type without breaking the rest of the code. To ease review this could be done in 2 steps:
    * make the necessary changes in the file structure, moving existing types around as a preparatory step. Personally, I would group changes in fewer commits. I don't mind big commits as long as they do the same thing and having everything in one commit makes it easier to follow the purpose of a change.
    * add the new type as effectively dead code.
2. Plug the new type in the rest of the system.

I know that adding dead-code is counter intuitive and that's probably why my first thought would have been to do the change the way Egor did it here, as well. However, I think that doing it like that helps reviewing and, in reality, it doesn't have any issues. If the dead code of the first part is wrong, we will catch it in the second part where we plug in everything, so we can go back and amend the first part.

In any case, thanks a lot for the PR. I think this is a great improvement in our code structure and "cleaniness".

As I mentioned earlier I just wanted to write down my thoughts so we have data points in the future when we discuss PRs and git organization.

PR #4286, opened 2023-11-29
Fix PATCH operation in virtio-block devices backed by asynchronous engine

bchalios:
## Changes

Changes the implementation along the PATCH code path for `virtio-block` devices so that, for the case of asynchronous engines (IO uring), it reuses the existing `EventFd` for getting completion callback notifications, rather than creating a knew one.

It also modifies the `virtio-block` tests to also test the asynchronous engine. It extends a test that verifies the PATCH operation works to also check that `mount`-ing an asynchronous device works.

## Reason

The asynchronous engine maintains an event file descriptor which passes to the IO uring interface when creating a new ring. IO uring uses this EventFd to notify us about completion of IO requests.

When we PATCH an async block device, we create a new asynchronous engine, including a new EventFd. However, we still monitor the old EventFd. This breaks the use of async drives post PATCH requests, because we never get notified about the results of requests we submit to the IO uring engine.

As a result, mounting a virtio-block device backed from an asynchronous file engine was blocking for ever.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/devices/virtio/virtio_block/device.rs
+++ b/src/vmm/src/devices/virtio/virtio_block/device.rs
@@ -71,19 +71,20 @@ pub struct DiskProperties {
 }
 
 impl DiskProperties {
-    pub fn new(
-        disk_image_path: String,
-        is_disk_read_only: bool,
-        file_engine_type: FileEngineType,
-    ) -> Result<Self, VirtioBlockError> {
-        let mut disk_image = OpenOptions::new()
+    // Helper function that opens the file with the proper access permissions
+    fn open_file(disk_image_path: &str, is_disk_read_only: bool) -> Result<File, VirtioBlockError> {
+        OpenOptions::new()
             .read(true)
             .write(!is_disk_read_only)
             .open(PathBuf::from(&disk_image_path))
-            .map_err(|x| VirtioBlockError::BackingFile(x, disk_image_path.clone()))?;
+            .map_err(|x| VirtioBlockError::BackingFile(x, disk_image_path.to_string()))
+    }
+
+    // Helper function that gets the size of the file
+    fn file_size(disk_image_path: &str, disk_image: &mut File) -> Result<u64, VirtioBlockError> {
         let disk_size = disk_image
             .seek(SeekFrom::End(0))
-            .map_err(|x| VirtioBlockError::BackingFile(x, disk_image_path.clone()))?;
+            .map_err(|x| VirtioBlockError::BackingFile(x, disk_image_path.to_string()))?;
 
         // We only support disk size, which uses the first two words of the configuration space.
         // If the image is not a multiple of the sector size, the tail bits are not exposed.
@@ -95,6 +96,17 @@ impl DiskProperties {
             );
         }
 
+        Ok(disk_size)
+    }
+
+    /// Create a new file for the block device using a FileEngine
+    pub fn new(
+        disk_image_path: String,
+        is_disk_read_only: bool,
+        file_engine_type: FileEngineType,
+    ) -> Result<Self, VirtioBlockError> {
+        let mut disk_image = Self::open_file(&disk_image_path, is_disk_read_only)?;
+        let disk_size = Self::file_size(&disk_image_path, &mut disk_image)?;
         let image_id = Self::build_disk_image_id(&disk_image);
 
         Ok(Self {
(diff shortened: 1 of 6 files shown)


src/vmm/src/devices/virtio/virtio_block/device.rs
```diff
@@ -66,24 +66,26 @@ impl FileEngineType {
 pub struct DiskProperties {
     pub file_path: String,
     pub file_engine: FileEngine<PendingRequest>,
+    pub is_disk_read_only: bool,
```
ShadowCurse, line 69: We already have `read_only` in the block itself. I think we don't need to duplicate it here.

src/vmm/src/devices/virtio/virtio_block/device.rs
```diff
@@ -95,17 +97,44 @@ impl DiskProperties {
...
     }
 
+    /// Update the path to the file backing the block device
+    pub fn update(&mut self, disk_image_path: String) -> Result<(), VirtioBlockError> {
+        let mut disk_image = Self::open_file(&disk_image_path, self.is_disk_read_only)?;
+        let disk_size = Self::file_size(&disk_image_path, &mut disk_image)?;
+
+        self.image_id = Self::build_disk_image_id(&disk_image);
+        self.file_engine
+            .update_file_path(disk_image)
```
ShadowCurse, line 130: What if we move `file_engine` out of `DiskProperties` into block struct itself? I feel `DiskProperties` is not really a place for it.

bchalios: I thought about it. But I think, this is refactoring :) I'd like to just fix the problem without restructuring the logic of the implementation.

src/vmm/src/devices/virtio/virtio_block/device.rs
ShadowCurse: I think this is actually does not belong here. I think this is not a big deal to pass it to `update` method with `disk_image_path`:
```rust
pub fn update(&mut self, disk_image_path: String, read_only: bool) -> Result<(), VirtioBlockError> {
  ...
}

self.disk.update(disk_image_path, self.read_only)?;
```

src/vmm/src/devices/virtio/virtio_block/device.rs
ShadowCurse: Ok, can you create an issue to refactor this so we will not forget?

src/vmm/src/devices/virtio/virtio_block/device.rs
bchalios: I disagree. I don't think it makes sense to pass this over, since the PATCH command does not allow changing the read/write semantics of the file.

tests/integration_tests/functional/test_drive_virtio.py
```diff
@@ -23,7 +23,7 @@ def partuuid_and_disk_path_tmpfs(rootfs_ubuntu_22, tmp_path):
     disk_path.unlink()
 
 
-def test_rescan_file(test_microvm_with_api):
+def test_rescan_file(test_microvm_with_api, io_engine):
```
kalyazin, line 26: can we extend patch-related tests in https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/functional/test_api.py with async too?

src/vmm/src/devices/virtio/virtio_block/io/async_io.rs
```diff
@@ -66,13 +67,10 @@ impl<T: Debug> WrappedUserData<T> {
 }
 
 impl<T: Debug> AsyncFileEngine<T> {
-    pub fn from_file(file: File) -> Result<AsyncFileEngine<T>, AsyncIoError> {
-        log_dev_preview_warning("Async file IO", Option::None);
-
-        let completion_evt = EventFd::new(libc::EFD_NONBLOCK).map_err(AsyncIoError::EventFd)?;
-        let ring = IoUring::new(
+    fn new_ring(file: &File, completion_fd: RawFd) -> Result<IoUring, io_uring::IoUringError> {
```
kalyazin, line 70: Looking at these codecov messages, it looks like we do have tests that cover these lines. Is that because the tests run on a 4.14 host?

src/vmm/src/devices/virtio/virtio_block/device.rs
ShadowCurse: This is for the `VirtioBlock` type to decide what can be changed. The `DiskProperties` is just a helper fro building `image_id` and `virtio_config_space`. It does not need to know if disk it `readonly` or not. The `readonly` property is only needed for `file_engine`, but we discussed that it can be moved out. 
Anyway, maybe this can be done as a part of the refactoring.

tests/integration_tests/functional/test_drive_virtio.py
bchalios: done!

src/vmm/src/devices/virtio/virtio_block/device.rs
bchalios: Ok, I moved the `is_read_only` thing out of `DiskProperties` ;) We can think of how to organize this better if/when we refactor.

src/vmm/src/devices/virtio/virtio_block/io/async_io.rs
bchalios: Looking at the changed files, I don't see codecov complaining(!)

bchalios: It is indeed weird, because these things end up being called by `virtio_block::device.rs` tests, so I think you might be right that it's about 4.14.

kalyazin (approved):

PR #4299, opened 2023-12-04
chore: update base64 dependency from 0.13.0 to 0.21.0

roypat:
It seems that dependabot is not monitoring our library crate .toml files, as our top-level Cargo.toml only specifies the binary crates as workspace members. Note that this is the only dependency update we've missed, as for minor and patch versions we get them through `Cargo.lock` updates (which dependabot does independently of how a dependency is pulled in)

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


zulinx86 (commented):
Do we have a long-term solution to avoid this type of dependency update issues?

JonathanWoollett-Light:
> > Do we have a long-term solution to avoid this type of dependency update issues?
> 
> We can either add all our crates to the `workspace` list in `Cargo.toml`, or just wait until we're done merging everything so that only binary crates remain

I agree with this. We should remove `default_members` and just have `members` with everything in it.

zulinx86:
> > > Do we have a long-term solution to avoid this type of dependency update issues?
> > 
> > 
> > We can either add all our crates to the `workspace` list in `Cargo.toml`, or just wait until we're done merging everything so that only binary crates remain
> 
> I agree with this. We should remove `default-members` and just have `members` with everything in it.

Can we do it in this PR?

zulinx86 (approved):

roypat:
> > > > Do we have a long-term solution to avoid this type of dependency update issues?
> > > 
> > > 
> > > We can either add all our crates to the `workspace` list in `Cargo.toml`, or just wait until we're done merging everything so that only binary crates remain
> > 
> > 
> > I agree with this. We should remove `default-members` and just have `members` with everything in it.
> 
> Can we do it in this PR?

Usually I'd be all over that, but I'm a bit apprehensious because I don't quite know _why_ these are set up the way they are. Let's not needlessly break things before a release :) I'll log an issue about it

PR #4305, opened 2023-12-07
[jailer] Fix `--parent-cpu-time-us`

pb8o:
## Changes

After da92bf6d, the CPU time we got from get_time_us was much shorter, and that caused an underflow when calculating the jailer_cpu_time_us. That resulted in Firecracker getting called with a huge value for `--parent-cpu-time`.

Before this fix:

        /firecracker [...] --parent-cpu-time-us 18446744073709551368

After this fix:

        /firecracker [...] --parent-cpu-time-us 4375

The new time is higher than before da92bf6d, because a) that commit introduced a fork, which carries some overhead, and b) we now account for CPU time during forks.

## Reason

Fix the regression

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


src/jailer/src/env.rs
```diff
@@ -674,6 +673,8 @@ impl Env {
 
         // Daemonize before exec, if so required (when the dev_null variable != None).
         if let Some(dev_null) = dev_null {
+            // Meter CPU usage before fork()
+            self.jailer_cpu_time_us = utils::time::get_time_us(utils::time::ClockType::ProcessCpu);
```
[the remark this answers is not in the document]
roypat: Mh, why would https://github.com/firecracker-microvm/firecracker/blob/4d665627e1477919641fa9d9536390493a97c4c4/src/jailer/src/main.rs#L372 always return 0?

roypat (approved):

roypat (dismissed):

src/jailer/src/env.rs
```diff
@@ -344,8 +344,7 @@ impl Env {
 
     fn exec_into_new_pid_ns(&mut self, chroot_exec_file: PathBuf) -> Result<(), JailerError> {
         // Compute jailer's total CPU time up to the current time.
-        self.jailer_cpu_time_us =
-            utils::time::get_time_us(utils::time::ClockType::ProcessCpu) - self.start_time_cpu_us;
+        self.jailer_cpu_time_us += utils::time::get_time_us(utils::time::ClockType::ProcessCpu);
```
sudanl0, line 347: There is a change of behaviour when --daemonize option is not passed but --new-pid-ns is passed, so I think we need to use start_time_cpu_us here and not in the daemonize condition:
```suggestion
        self.jailer_cpu_time_us += utils::time::get_time_us(utils::time::ClockType::ProcessCpu);
        self.jailer_cpu_time_us += utils::time::get_time_us(utils::time::ClockType::ProcessCpu) - self.start_time_cpu_us;
```

pb8o: Good catch! Initially I thought just removing that substraction altogether.

So it gets a bit confusing depending on the combination of options. I think we need to take the accounting out of the `--new-pid-ns` option, and add the time before a fork/clone.

pb8o: Fixed (hopefully) in latest PR.

roypat (approved):

sudanl0 (approved):

PR #4306, opened 2023-12-08
Fcmetrics to cw

sudanl0:
## Changes

- Validate process_cpu_time against large values to avoid issues with large process time mentioned in
https://github.com/firecracker-microvm/firecracker/pull/4305
- Refactor Firecracker metrics specific functions to its own file.
- Upload Firecracker metrics from performance tests to CloudWatch.

## Reason

- To gain better observability of Firecracker metrics.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


tests/host_tools/metrics.py
```diff
@@ -206,6 +206,72 @@ def format_with_reduced_unit(value, unit):
     return f"{reduced_value:.2f}{formatted_unit}"
 
 
+def get_emf_unit_for_fc_metrics(key):
```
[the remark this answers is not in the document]
sudanl0: I was using this same file for Firecracker metrics as well and had added some functions like validate_fc_metrics() in this file last time. I don't have a strong opinion but if it looks better then I can move all firecracker metrics stuff from this file to a new file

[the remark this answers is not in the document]
pb8o: Yeah I think it would make sense to have the CW stuff in this file and anything FC specific in the other

tests/framework/microvm.py
```diff
@@ -409,6 +410,14 @@ def get_all_metrics(self):
         """Return all metric data points written by FC."""
         return [json.loads(line) for line in self.metrics_file.read_text().splitlines()]
 
+    def flush_metrics_to_cw(self, metrics):
+        """Uploads all metric data points emitted by Firecracker to CloudWatch."""
+
+        self.api.actions.put(action_type="FlushMetrics")
+        fc_metrics = open(self.metrics_file, encoding="utf-8").read().replace("\n", "")
+        fc_metrics = fc_metrics.replace(" ", "").replace("}{", "}\n{")
+        flush_fc_metrics_to_cw(fc_metrics, metrics)
```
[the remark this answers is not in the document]
sudanl0: >It doesn't need to know that the metrics are getting sent to CloudWatch.

I get this part but I think instead of polling every X secs it would be simpler to dump all metrics at the end of each test. I'll move the fc metrics related stuff from metrics.py to a new file and add something to discuss.

tests/host_tools/fcmetrics.py
```diff
@@ -363,3 +367,126 @@ def validate_per_device_metrics(self, fc_metrics):
...
+    def _flush_metrics(self):
+        """
+        Since vm.flush_metrics provides only the latest metrics,
+        we call vm.get_all_metrics() instead to be able to collect
+        and upload all metrics data emitted by the microvm.
+        This utility function is created to keep common code in one
+        place and is called every 60sec once the daemon starts and then
+        once when the daemon stops.
+        """
+        metrics = self.vm.get_all_metrics()
```
pb8o, line 470: Since we poll every 60s, then flush_metrics should be enough? If we want to have a separate, one-shot method that's ok (though we don't have a use for it in the PR), but I think having both here makes this more complex.

```
def _flush_metrics(self):
    flush_fc_metrics_to_cw(self.vm.flush_metrics())

def flush_metrics_all(self):
    for metrics in self.vm.get_all_metrics():
        flush_fc_metrics_to_cw(metrics)
```

tests/host_tools/fcmetrics.py
```diff
@@ -363,3 +367,126 @@ def validate_per_device_metrics(self, fc_metrics):
...
+        for index in range(self.metrics_index, len(metrics)):
```
pb8o, line 471:
```python
for update in metrics[self.metrics_index:]:
```

tests/host_tools/fcmetrics.py
```diff
@@ -363,3 +367,126 @@ def validate_per_device_metrics(self, fc_metrics):
...
+
+    def stop(self):
+        """
+        Stop the daemon gracefully.
+        Since we depend on the vm to provide the metrics,
+        this method should be called just before killing the vm.
+        We collect final metrics here in stop instead of letting it
+        be collected from the "run" method because, "run" could be
+        in sleep when stop is called and once it wakes out of sleep
+        the "vm" might not be avaiable to provide the metrics.
```
pb8o, line 483: I think there could be a race here if both threads are running, since stop will be called from a different thread. Maybe Python's GIL will avoid that case, but I think there's still a chance we submit the last metric twice.

Could we just kill the thread?

tests/host_tools/fcmetrics.py
```diff
@@ -363,3 +367,126 @@ def validate_per_device_metrics(self, fc_metrics):
...
+        Thread.__init__(self, daemon=True)
+        self.vm = vm
+        self.dimensions = dimensions
+        self.assert_on_err = assert_on_err
+
+        self.metrics_index = 0
+        self.running = True
+
+        self.metrics_logger = get_metrics_logger()
+        self.metrics_logger.set_dimensions(dimensions)
```
pb8o, line 459: I'm thinking the metrics_logger could just get passed to this class, so it would also avoid having to deal with dimensions. Since it looks like in the rest of the tests we just use the same dimensions, this would be simpler.

tests/host_tools/fcmetrics.py
```diff
@@ -363,3 +367,126 @@ def validate_per_device_metrics(self, fc_metrics):
...
+    }
+    skip = set()
+    for group, keys in fc_metrics.items():
+        if group == "utc_timestamp_ms":
+            continue
+        for key, value in keys.items():
+            full_key = f"{group}.{key}"
+            # values are 0 when:
+            # - there is no update
+            # - device is not used
```
pb8o, line 416: I think 0s for "no update" should still be sent (i.e. this interface received 0 bytes in this period is still good info), the other ones we can skip. However in this function we cannot tell those two apart. If we also passed the Microvm object we could tell the difference, i.e. we could see if a device is configured in the VM and then send the metrics.

tests/integration_tests/performance/test_process_startup_time.py
```diff
@@ -75,6 +75,11 @@ def _test_startup_time(microvm):
     )
 
     assert cpu_startup_time_us > 0
+    # No specific reason to compare against 150ms or 10ms
+    # its just to catch issues like the ones introduced in PR
+    # https://github.com/firecracker-microvm/firecracker/pull/4305
+    assert startup_time_us < 150000
+    assert cpu_startup_time_us < 10000
```
roypat, line 82: Instead of hardcoding values, can we do something like `initial = time.time() * 1000 * 1000` at the start of this function, then `end = time.time() * 1000 * 1000` after the `microvm.get_all_metrics()` call, and then 
 ```suggestion
    # No specific reason to compare against 150ms or 10ms
    # its just to catch issues like the ones introduced in PR
    # https://github.com/firecracker-microvm/firecracker/pull/4305
    assert startup_time_us < end - begin
    assert cpu_startup_time_us < end - begin
```
instead? This test the same regression (e.g. start time is not longer than the process has been alive in the first place), but without any magic numbers that might later on accidentally become SLAs

File: tests/host_tools/fcmetrics.py (hunk 5 above)
Comment (sudanl0) [reply tests/host_tools/fcmetrics.py:483]: I thought so too but couldn't find a better way to do it, I'll try the kill approach locally to see how it works.

File: tests/host_tools/fcmetrics.py (hunk 3 above)
Comment (sudanl0) [reply tests/host_tools/fcmetrics.py]: I find get_all_metrics() more safe because flush_metrics() there is a chance to miss some metrics entries.
e.g. suppose there are some incmetrics like block read_bytes just flushed by FC and they are the last entry in fcmetrics ndjson file, now if we call vm.flush_metrics() it will send a new api request to flush FC metrics and latest entry of fcmetrics ndjson file would now point to some different read_bytes data. We lose the previously filled read_bytes metrics in these cases so calling get_all_metrics() and iterating over all it seemed safe to capture all metrics.

File: tests/host_tools/fcmetrics.py (hunk 5 above)
Comment (sudanl0) [reply tests/host_tools/fcmetrics.py:479]: instead of killing the thread I tried to work around it by making the stop wait for 1 sec before collecting metrics and shortened the sleep of run() to 1 sec so that it can wake up sooner and break the loop

File: tests/host_tools/fcmetrics.py (hunk 7 above)
Comment (sudanl0) [reply tests/host_tools/fcmetrics.py:421]: I think microvm can provide list of disk and net devices attached but not other devices like vsock/balloon/entropy so uploading metrics based on devices seems difficult (unless I am missing something and there is a way to get the list ?). I agree that skipping metrics based on 0 doesn't look good either so removed that check. Now we'll upload metrics even if it is 0.

Comment (pb8o) [reply tests/host_tools/fcmetrics.py:421]: Yeah that's probably simpler. 

For vsock and balloon I think we could probably get from the `/vm/config` api. But that's probably overkill.

File: tests/host_tools/fcmetrics.py [hunk 9]
```diff
@@ -363,3 +365,135 @@ def validate_per_device_metrics(self, fc_metrics):
...
+    all metrics together in the end, to retain timestamp of the metrics.
+    """
+
+    def __init__(self, vm, metrics_logger, timer=60):
+        Thread.__init__(self, daemon=True)
+        self.vm = vm
+        self.timer = timer
+
+        self.metrics_index = 0
+        self.running = True
```
pb8o, line 453: nit: hmm should this be False here and then set to True as the 1st line in `run`?

tests/host_tools/fcmetrics.py
```diff
@@ -363,3 +365,135 @@ def validate_per_device_metrics(self, fc_metrics):
...
+        We collect final metrics here in stop instead of letting it
+        be collected from the "run" method because, "run" could be
+        in sleep when stop is called and once it wakes out of sleep
+        the "vm" might not be avaiable to provide the metrics.
+        """
+        self.running = False
+        # wait for the running thread to finish
+        # this should also avoid any race condition leading to
+        # uploading the same metrics twice
+        time.sleep(1)
```
pb8o, line 485: Would it make sense to wait for the thread here instead of waiting? `self.join()`

tests/host_tools/fcmetrics.py
pb8o: Makes sense, thanks!

tests/host_tools/fcmetrics.py
pb8o: Thanks for the explanation. I guess we can always optimize it later if need be.

tests/host_tools/fcmetrics.py
sudanl0: yeah that looks much better, thanks!

pb8o (approved):

tests/host_tools/fcmetrics.py
```diff
@@ -0,0 +1,365 @@
...
+        for vhost_user_dev in vhost_user_devices:
+            temp_pop_metrics = metrics[vhost_user_dev].pop("activate_time_us")
+            try:
+                jsonschema.validate(instance=metrics, schema=firecracker_metrics_schema)
+            except jsonschema.exceptions.ValidationError as error:
+                if error.message.strip() == "'activate_time_us' is a required property":
+                    pass
+                else:
+                    raise error
+            metrics[vhost_user_dev]["activate_time_us"] = temp_pop_metrics
```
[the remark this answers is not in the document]
roypat: Synced offline, yes this is indeed the intended purpose of this code. Since this PR is just moving things between files, it can stay this way.

roypat (approved):

Issue #4284, opened 2023-11-29
Vsock does not work on guest linux kernel version 6.1


kayoub-akamai:
# Describe the bug

When trying to use vsock on a guest with linux kernel version 6.1.63 there are infinite errors. It works with guest kernel version 5.10.202.

## To Reproduce
Used the vsock instructions at https://github.com/firecracker-microvm/firecracker/blob/main/docs/vsock.md#connecting-from-guest-to-host 
Config:
```
	"vsock": {
		"uds_path": "v.sock",
		"guest_cid": 3
	}
```
Command on host:
```
socat - UNIX-LISTEN:./v.sock_52
```
Command on vm:
```
socat - VSOCK-CONNECT:2:52
```
this results in a stream of errors from the vm that never ends:
```
2023-11-27T13:32:43.044305511 [anonymous-instance:main] vsock: RX queue error: BufDescMissing. Discarding the package.
2023-11-27T13:32:43.044308301 [anonymous-instance:main] vsock: RX queue error: BufDescMissing. Discarding the package.
2023-11-27T13:32:43.044310994 [anonymous-instance:main] vsock: RX queue error: BufDescMissing. Discarding the package.
2023-11-27T13:32:43.044313682 [anonymous-instance:main] vsock: RX queue error: BufDescMissing. Discarding the package.
2023-11-27T13:32:43.044316370 [anonymous-instance:main] vsock: RX queue error: BufDescMissing. Discarding the package.
2023-11-27T13:32:43.044319156 [anonymous-instance:main] vsock: RX queue error: BufDescMissing. Discarding the package.
and when I terminate the VM, the host socat is closed - which shows there was some interaction at least 
```
## Expected behaviour

It shouldn't have those errors.

## Environment

Firecracker versions tried 1.4 and 1.5
Host Kernel 5.15.0-88
Guest kernels that failed 6.4.7 and 6.1.63
Guest kernel that worked 5.10.202

## Additional context

Can't use vsock.

## Checks

Brought this up in Slack, was asked to make a ticket for it.

roypat:
Hello @kayoub-akamai,

Thank you for your bug report. It seems that this is related to changes in the linux kernel's vsock driver originally merged for 6.2, but recently backported to 6.1 as well. We will look into fixing this issue as soon as possible.

roypat:
[...]

PR #4310, opened 2023-12-11
Allow vsock device to process arbitrarily shaped descriptor chains, in anticipation of guest-side Linux 6.1 changes

roypat:
## Changes

Allow the vsock device to process arbitrarily shaped descriptor chains.

## Reason

As reported in #4284, with Linux 6.1 a linux guest will start using one virtio descriptor per vsock packet. Up until 6.0, it uses a two-descriptor descriptor chain, with the first descriptor containing the vsock packet header, and the second descriptor the payload (if it exists). 

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/devices/virtio/iovec.rs
+++ b/src/vmm/src/devices/virtio/iovec.rs
@@ -4,6 +4,7 @@
 use std::io::ErrorKind;
 
 use libc::{c_void, iovec, size_t};
+use smallvec::SmallVec;
 use vm_memory::{
     GuestMemoryError, ReadVolatile, VolatileMemoryError, VolatileSlice, WriteVolatile,
 };
@@ -29,7 +30,7 @@ pub enum IoVecError {
 #[derive(Debug)]
 pub struct IoVecBuffer {
     // container of the memory regions included in this IO vector
-    vecs: Vec<iovec>,
+    vecs: SmallVec<[iovec; 4]>,
     // Total length of the IoVecBuffer
     len: usize,
 }
@@ -37,7 +38,7 @@ pub struct IoVecBuffer {
 impl IoVecBuffer {
     /// Create an `IoVecBuffer` from a `DescriptorChain`
     pub fn from_descriptor_chain(head: DescriptorChain) -> Result<Self, IoVecError> {
-        let mut vecs = vec![];
+        let mut vecs = SmallVec::new();
         let mut len = 0usize;
 
         let mut next_descriptor = Some(head);
@@ -164,7 +165,7 @@ impl IoVecBuffer {
 #[derive(Debug)]
 pub struct IoVecBufferMut {
     // container of the memory regions included in this IO vector
-    vecs: Vec<iovec>,
+    vecs: SmallVec<[iovec; 4]>,
     // Total length of the IoVecBufferMut
     len: usize,
 }
@@ -172,7 +173,7 @@ pub struct IoVecBufferMut {
 impl IoVecBufferMut {
     /// Create an `IoVecBufferMut` from a `DescriptorChain`
     pub fn from_descriptor_chain(head: DescriptorChain) -> Result<Self, IoVecError> {
-        let mut vecs = vec![];
+        let mut vecs = SmallVec::new();
         let mut len = 0usize;
 
         for desc in head {
@@ -295,7 +296,8 @@ mod tests {
                 vecs: vec![iovec {
                     iov_base: buf.as_ptr() as *mut c_void,
                     iov_len: buf.len(),
-                }],
+                }]
+                .into(),
                 len: buf.len(),
             }
         }
@@ -325,7 +327,8 @@ mod tests {
                 vecs: vec![iovec {
                     iov_base: buf.as_mut_ptr().cast::<c_void>(),
                     iov_len: buf.len(),
-                }],
+                }]
+                .into(),
                 len: buf.len(),
             }
         }
@@ -544,6 +547,7 @@ mod verification {
     use std::mem::ManuallyDrop;
 
     use libc::{c_void, iovec};
+    use smallvec::SmallVec;
 
     use super::{IoVecBuffer, IoVecBufferMut};
 
@@ -557,7 +561,7 @@ mod verification {
     // >= 1.
     const MAX_DESC_LENGTH: usize = 4;
 
-    fn create_iovecs(mem: *mut u8, size: usize) -> (Vec<iovec>, usize) {
+    fn create_iovecs(mem: *mut u8, size: usize) -> (SmallVec<[iovec; 4]>, usize) {
         let nr_descs: usize = kani::any_where(|&n| n <= MAX_DESC_LENGTH);
         let mut vecs: Vec<iovec> = Vec::with_capacity(nr_descs);
         let mut len = 0usize;
@@ -575,7 +579,7 @@ mod verification {
             len += iov_len;
         }
 
-        (vecs, len)
+        (vecs.into(), len)
     }
 
     impl kani::Arbitrary for IoVecBuffer {
--- a/src/vmm/src/devices/virtio/iovec.rs
+++ b/src/vmm/src/devices/virtio/iovec.rs
@@ -5,7 +5,7 @@ use libc::{c_void, iovec, size_t};
 use vm_memory::GuestMemoryError;
 
 use crate::devices::virtio::queue::DescriptorChain;
-use crate::vstate::memory::{Bitmap, GuestMemory, GuestMemoryMmap};
+use crate::vstate::memory::{Bitmap, GuestMemory};
 
 #[derive(Debug, thiserror::Error, displaydoc::Display)]
 pub enum IoVecError {
@@ -32,10 +32,7 @@ pub(crate) struct IoVecBuffer {
 
 impl IoVecBuffer {
     /// Create an `IoVecBuffer` from a `DescriptorChain`
-    pub fn from_descriptor_chain(
-        mem: &GuestMemoryMmap,
-        head: DescriptorChain,
-    ) -> Result<Self, IoVecError> {
+    pub fn from_descriptor_chain(head: DescriptorChain) -> Result<Self, IoVecError> {
         let mut vecs = vec![];
         let mut len = 0usize;
 
@@ -48,7 +45,8 @@ impl IoVecBuffer {
             // We use get_slice instead of `get_host_address` here in order to have the whole
             // range of the descriptor chain checked, i.e. [addr, addr + len) is a valid memory
             // region in the GuestMemoryMmap.
-            let iov_base = mem
+            let iov_base = desc
+                .mem
                 .get_slice(desc.addr, desc.len as usize)?
                 .ptr_guard_mut()
                 .as_ptr()
@@ -162,10 +160,7 @@ pub(crate) struct IoVecBufferMut {
 
 impl IoVecBufferMut {
     /// Create an `IoVecBufferMut` from a `DescriptorChain`
-    pub fn from_descriptor_chain(
-        mem: &GuestMemoryMmap,
-        head: DescriptorChain,
-    ) -> Result<Self, IoVecError> {
+    pub fn from_descriptor_chain(head: DescriptorChain) -> Result<Self, IoVecError> {
         let mut vecs = vec![];
         let mut len = 0usize;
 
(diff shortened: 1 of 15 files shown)


roypat:
I separated the unit test changes from the actual device implementation changes to make review easier and keep the commits small-ish

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -124,8 +125,8 @@ impl IoVecBuffer {
             }
 
             let mut slice =
-            // SAFETY: the constructor IoVecBufferMut::from_descriptor_chain ensures that
-            // all iovecs contained point towards valid ranges of guest memory
+                // SAFETY: the constructor IoVecBufferMut::from_descriptor_chain ensures that
```
bchalios, line 135: nit: Would you mind fixing this indentation issue in the commit that introduces the comment?

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -248,8 +249,8 @@ impl IoVecBufferMut {
...
+                // all iovecs contained point towards valid ranges of guest memory
```
bchalios, line 260: nit: same comment here

src/vmm/src/devices/virtio/vsock/csm/connection.rs
```diff
@@ -160,7 +160,11 @@ where
     /// - `Err(VsockError::NoData)`: there was no data available with which to fill in the packet;
     /// - `Err(VsockError::PktBufMissing)`: the packet would've been filled in with data, but it is
     ///   missing the data buffer.
-    fn recv_pkt(&mut self, pkt: &mut VsockPacket, mem: &GuestMemoryMmap) -> Result<(), VsockError> {
+    fn recv_pkt(
+        &mut self,
+        pkt: &mut VsockPacket,
+        _mem: &GuestMemoryMmap,
```
bchalios: Should we just remove this parameter now?

src/vmm/src/devices/virtio/vsock/csm/connection.rs
```diff
@@ -604,7 +608,7 @@ where
     ///
     /// Raw data can either be sent straight to the host stream, or to our TX buffer, if the
     /// former fails.
-    fn send_bytes(&mut self, mem: &GuestMemoryMmap, pkt: &VsockPacket) -> Result<(), VsockError> {
+    fn send_bytes(&mut self, _mem: &GuestMemoryMmap, pkt: &VsockPacket) -> Result<(), VsockError> {
```
bchalios: same comment for `_mem` here.

bchalios (changes requested):
Generally LGTM. Just few comments.

I like the `SmallVec` solution for the performance issue, but we have been discussing for a while to fix this issue by pre-allocating an `IoVecBuffer(Mut)` in the device state and re-using it. It sounds like this would be the correct thing to do. WDYT?

Some per-commit comments:

commit: `fix: Allow vsock device to handle arbitrarily shaped DescriptorChains`
* Commit message seems malformated.

commit: `kani: Update the iovec harnesses`
* Can we please add a TODO in the code and remember to open an issue to remind us to move back to `SmallVec` once `kani` folks fix the issue?

commit: `refractor: remove unneeded GuestMemoryMmap parameters`
* I think I prefer doing this in place, at the commit that makes the `mem` parameter useless (I also left inline comments), but I could be convinced otherwise.

commit: `chore: Remove unused VsockError::BufDescMissing`
* Same as above. Can we squash this with the commit that makes `BufDescMissing` unused?

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -557,7 +560,7 @@ mod verification {
     // >= 1.
     const MAX_DESC_LENGTH: usize = 4;
 
-    fn create_iovecs(mem: *mut u8, size: usize) -> (Vec<iovec>, usize) {
+    fn create_iovecs(mem: *mut u8, size: usize) -> (smallvec::SmallVec<[iovec; 4]>, usize) {
```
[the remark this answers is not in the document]
roypat: Fixed

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -64,7 +68,7 @@ impl IoVecBuffer {
     }
 
     /// Get the total length of the memory regions covered by this `IoVecBuffer`
-    pub fn len(&self) -> usize {
+    pub(crate) fn len(&self) -> usize {
```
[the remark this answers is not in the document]
roypat: That's because otherwise, some clippy lint about "len without is_empty" gets triggered (god knows why that is warn-by-default)

src/vmm/src/devices/virtio/iovec.rs
roypat: fixed

src/vmm/src/devices/virtio/iovec.rs
roypat: fixed

roypat:
> Generally LGTM. Just few comments.
> 
> I like the `SmallVec` solution for the performance issue, but we have been discussing for a while to fix this issue by pre-allocating an `IoVecBuffer(Mut)` in the device state and re-using it. It sounds like this would be the correct thing to do. WDYT?

Yeah, I do still think that this would be the optimal solution here. Originally I wasn't gonna touch on it at all here, but using Vec-based IoVecBuffer caused some slightly changes in the numbers reported by our vsock throughput test (now, some of these were improvements and others were degradation, so I think its less that the allocations hurt us and more that they change the ratio of when vcpu and vmm threads do things just enough for us to get a different performance profile). Reusing the buffers still sounds quite complicated to me though (because its stateful, so lots of ways to mess things up - e.g. will need more elaborate testing and everything), definitely not something I'd feel comfortable doing in 2 days before a release. The `SmallVec` heuristic gets us there most of the way (because even with the kernel changes, vsock packets are 1 or 2 descriptors). Do you know anything about how long the descriptor chains are for the net device? This PR doesnt trigger an A/B-test failure, so I assume its more than 4.

> Some per-commit comments:
> 
> commit: `fix: Allow vsock device to handle arbitrarily shaped DescriptorChains`
> 
>     * Commit message seems malformated.

That's github thinking the line that happens to start with "6.1" starts a numbered list.
 
> commit: `kani: Update the iovec harnesses`
> 
>     * Can we please add a TODO in the code and remember to open an issue to remind us to move back to `SmallVec` once `kani` folks fix the issue?

Added the comment, will open issues after PR is merged.

> commit: `refractor: remove unneeded GuestMemoryMmap parameters`
> 
>     * I think I prefer doing this in place, at the commit that makes the `mem` parameter useless (I also left inline comments), but I could be convinced otherwise.
> 
> 
> commit: `chore: Remove unused VsockError::BufDescMissing`
> 
>     * Same as above. Can we squash this with the commit that makes `BufDescMissing` unused?

About these, I'd prefer to keep them separate, mainly to keep the refactoring from polluting the actual functional changes, so I tried to keep the changes in that commit to the minimum required to appease our linters (I realize I broke my own rule here with changing the order of parameters in the read and write functions on `VsockPacket`, but I'd like to not make it worse at least)

bchalios:
> About these, I'd prefer to keep them separate, mainly to keep the refactoring from polluting the actual functional changes, so I tried to keep the changes in that commit to the minimum required to appease our linters (I realize I broke my own rule here with changing the order of parameters in the read and write functions on VsockPacket, but I'd like to not make it worse at least)

That's fine for the Error types, but the `_mem` things I 'd like to have them fixed. That part doesn't look like a refactoring to me. You are changing code to not use the `mem` argument, so just remove it, instead of making it `_mem`.

roypat:
> > About these, I'd prefer to keep them separate, mainly to keep the refactoring from polluting the actual functional changes, so I tried to keep the changes in that commit to the minimum required to appease our linters (I realize I broke my own rule here with changing the order of parameters in the read and write functions on VsockPacket, but I'd like to not make it worse at least)
> 
> That's fine for the Error types, but the `_mem` things I 'd like to have them fixed. That part doesn't look like a refactoring to me. You are changing code to not use the `mem` argument, so just remove it, instead of making it `_mem`.

The problem here is that the `mem`->`_mem` change in 7f280150eecc4400984d10b9fdd7c259cd0973e3 is in a downstream implementation of a trait, so it cannot be removed, without removing it from the trait (which is what 36752b91eddc6b1894e75e4202e36a2cbb88ea5b does). However the trait also has other downstream implementations, so it kinda bubbles up and down into a larger refractor

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -251,16 +215,69 @@ impl IoVecBufferMut {
...
+
+            if offset >= iov.iov_len {
+                offset -= iov.iov_len;
+                continue;
+            }
+
+            let mut slice =
+                // SAFETY: the constructor IoVecBufferMut::from_descriptor_chain ensures that
+                // all iovecs contained point towards valid ranges of guest memory
+                unsafe { VolatileSlice::new(iov.iov_base.cast(), iov.iov_len).offset(offset)? };
```
bchalios, line 262: I _think_ we should create this with a bitmap, no?

roypat: In this case, no, because it would create a new bitmap that is completely detached from the bitmap object associated with the GuestMemoryMmap into which these pointers lead. We already mark the entire IoVecBufferMut region as dirty in `from_tx_descriptor_chain`, before we "forget" about the bitmap by only storing pointers.

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -133,16 +90,70 @@ impl IoVecBuffer {
...
+                    }
+                    Ok(bytes_read) => break bytes_read,
+                    Err(volatile_memory_error) => return Err(volatile_memory_error),
+                }
+            };
+            total_bytes_read += bytes_read;
+
+            if bytes_read < slice.len() {
+                break;
+            }
```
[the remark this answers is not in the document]
roypat: Mh, the old code only used u8-slices, and for u8 slices if we call this method through `read_at`, we cannot return anything "in-between" (because of the Read/Write semantics of u8 slices). The kani harnesses actually verify this property. The vsock code on the other hand uses UnixStreams as the Read/Write backend, and thus depends on this function not being "exact". Maybe we can add a check for "buffer was filled" to `read_at/write_at` and return `None` otherwise?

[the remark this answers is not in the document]
bchalios: > Mh, the old code only used u8-slices, and for u8 slices if we call this method through read_at, we cannot return anything "in-between" (because of the Read/Write semantics of u8 slices). 

Yup, that's my point. Before we new that for our use-case we cared/expected only 0 or `length`. That's why we ended up with the awkward API semantics.

Since this is now not the case any more, I think we should fix the API. But maybe it can happen in a follow up PR (?)

bchalios (approved):

src/vmm/src/devices/virtio/iovec.rs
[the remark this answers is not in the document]
roypat: We've synced on this off-line, I've added a commit at the end of this series that fixes up the API semantics slightly :)

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -133,16 +90,70 @@ impl IoVecBuffer {
...
     }
+
+    /// Reads up to `limit` bytes from the `IoVecBuffer` starting at the given offset.
+    ///
+    /// This will try to write to the given [`WriteVolatile`].
+    pub fn read_volatile_at<W: WriteVolatile>(
+        &self,
+        dst: &mut W,
+        mut offset: usize,
+        mut limit: usize,
```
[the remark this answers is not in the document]
roypat: changed :)

bchalios (dismissed):

ShadowCurse (approved):

ShadowCurse (approved):

bchalios (approved):

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -22,6 +22,14 @@ pub enum IoVecError {
     GuestMemory(#[from] GuestMemoryError),
 }
 
+// Using SmallVec in the kani proofs causes kani to use unbounded amounts of memory
+// during post-processing, and then crash.
+// TODO: remove new-type once kani performance regression are resolved
```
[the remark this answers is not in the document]
roypat: Ah, yes, thanks for reminding me, now that this is merged I need to cut them the ticket!

Issue #3707, opened 2023-05-19
[Feature Request] Adding vhost-net support to firecracker


majek:
Hi,

Firecracker doesn't have fast enough networking for many users. Firecracker networking benchmarks claiming tens of Gbps use TCP and use network offloads (TSO). However, often for real users these assumptions can't be met. For example, QUIC uses UDP, and when using XDP in the guest, the offloads must be disabled.

Under such conditions - UDP and no offloads - firecracker network performance is dismal. We can get only 1-2Gbps. This is really understandable once you realize how the tap code works.

Other kvm-based virtualization systems can do much better, they typically do this by utilizing vhost-net host kernel acceleration.

I'm considering working on adding vhost-net support to firecracker. Vhost-net is a host kernel API, exposing tap device in a format compatible with virtio-net guest driver. With shared memory and ioctl's dance it's possible to greatly speed up the data path for networking. Host vhost-net kernel thread would copy network data directly into the kvm guest memory. It's possible to set it up in such a way that firecracker process is not touching neither the data nor interrupts when doing networking. This saves loads of context switches, greatly improves latency and reduces CPU usage.

It doesn't solve the scalability issue - a naive vhost-net implementation would still only be single-queue and use one vCPU in the guest. This is a separate issue.

Vhost-net is a well established API. From the previous discussions it feels the firecracker community was not in favor of vhost-net. Vhost-net exposes the host kernel, so it extends the attack surface against the host. However, over the years vhost-net matured, the implementation strengthened and given its benefits it might be a good time to reconsider it.

I think with vhost-net we should be able to increase the network performance for UDP + disabled offloads from 1-2Gbps to 7-8Gbps. Users of vhost-net would see a reduced CPU usage for networking-heavy virtual machines and improved latency.

When looking at adding vhost-net to firecracker I had some concerns:

(1) It's unclear how to support MMDS with vhost-net. Is that a problem?

(2) It's unclear how to deal with networking rate limits.

I think both of these features could be supported by doing ebpf on the host side. I'm actually surprised by software rate-limit implementation done in firecracker since it could be done much more efficiently in bpf.  I'm not looking at vsock at this stage.

I would like to understand if:

* the maintainers are generally open for adding optional vhost-net support

* there are any extra conditions that must be met to consider the vhost-net code

[...]

PR #4312, opened 2023-12-12
Improve vqueue encapsulation - prerequisite for vhost work

majek:
## Changes

This PR refactors `virtio.Queue` and `VirtioDevice.queues()` interfaces in order to better encapsulate the vqueue.

## Reason

This PR is part of #3707 issue, where we're trying to speed up Firecracker networking with vhost. From Firecracker point of view, Linux vhost interface is basically another implementation of virtqueue (or Virtual Queue) from Virtio spec. In order to integrate vhost into firecracker in the future we need more flexibility around vqueue implementation.

However, this is not easy. Firecracker `virtio.Queue` is deeply integrated into the code. It's imported and directly used from all the device implementations. In order to make it more plugable, we need to put some abstractions around it.

This PR creates a foundation for that. We don't change any functionality of Firecracker, just refactor the `virtio.Queue` interface a bit. These changes are necessary to create decent vhost code in the future, otherwise the spaghetti explodes fast.

We believe these changes are not controversial and generally improve the abstractions and code cleanliness.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/vmm/src/device_manager/mmio.rs
+++ b/src/vmm/src/device_manager/mmio.rs
@@ -467,7 +467,7 @@ mod tests {
 
     use super::*;
     use crate::devices::virtio::device::VirtioDevice;
-    use crate::devices::virtio::queue::Queue;
+    use crate::devices::virtio::queue::{Queue, QueueIter, QueueIterMut};
     use crate::devices::virtio::ActivateError;
     use crate::vstate::memory::{GuestAddress, GuestMemoryExtension, GuestMemoryMmap};
     use crate::{builder, Vm};
@@ -535,12 +535,12 @@ mod tests {
             0
         }
 
-        fn queues(&self) -> &[Queue] {
-            &self.queues
+        fn queues(&self) -> QueueIter {
+            self.queues.iter()
         }
 
-        fn queues_mut(&mut self) -> &mut [Queue] {
-            &mut self.queues
+        fn queues_mut(&mut self) -> QueueIterMut {
+            self.queues.iter_mut()
         }
 
         fn queue_events(&self) -> &[EventFd] {
--- a/src/vmm/src/devices/virtio/balloon/device.rs
+++ b/src/vmm/src/devices/virtio/balloon/device.rs
@@ -14,7 +14,6 @@ use utils::eventfd::EventFd;
 use utils::u64_to_usize;
 
 use super::super::device::{DeviceState, VirtioDevice};
-use super::super::queue::Queue;
 use super::super::{ActivateError, TYPE_BALLOON};
 use super::metrics::METRICS;
 use super::util::{compact_page_frame_numbers, remove_range};
@@ -30,6 +29,7 @@ use super::{
 use crate::devices::virtio::balloon::BalloonError;
 use crate::devices::virtio::device::{IrqTrigger, IrqType};
 use crate::devices::virtio::gen::virtio_blk::VIRTIO_F_VERSION_1;
+use crate::devices::virtio::queue::{Queue, QueueIter, QueueIterMut};
 use crate::logger::IncMetric;
 use crate::vstate::memory::{Address, ByteValued, Bytes, GuestAddress, GuestMemoryMmap};
 
@@ -573,12 +573,12 @@ impl VirtioDevice for Balloon {
         TYPE_BALLOON
     }
 
-    fn queues(&self) -> &[Queue] {
-        &self.queues
+    fn queues(&self) -> QueueIter {
+        self.queues.iter()
     }
 
-    fn queues_mut(&mut self) -> &mut [Queue] {
-        &mut self.queues
+    fn queues_mut(&mut self) -> QueueIterMut {
+        self.queues.iter_mut()
     }
 
     fn queue_events(&self) -> &[EventFd] {
--- a/src/vmm/src/devices/virtio/balloon/persist.rs
+++ b/src/vmm/src/devices/virtio/balloon/persist.rs
@@ -209,7 +209,7 @@ mod tests {
         assert_eq!(restored_balloon.acked_features, balloon.acked_features);
         assert_eq!(restored_balloon.avail_features, balloon.avail_features);
         assert_eq!(restored_balloon.config_space, balloon.config_space);
-        assert_eq!(restored_balloon.queues(), balloon.queues());
+        assert!(restored_balloon.queues().eq(balloon.queues()));
         assert_eq!(
             restored_balloon.interrupt_status().load(Ordering::Relaxed),
             balloon.interrupt_status().load(Ordering::Relaxed)
--- a/src/vmm/src/devices/virtio/device.rs
+++ b/src/vmm/src/devices/virtio/device.rs
@@ -12,8 +12,8 @@ use std::sync::Arc;
 use utils::eventfd::EventFd;
 
 use super::mmio::{VIRTIO_MMIO_INT_CONFIG, VIRTIO_MMIO_INT_VRING};
-use super::queue::Queue;
 use super::ActivateError;
+use crate::devices::virtio::queue::{QueueIter, QueueIterMut};
 use crate::devices::virtio::AsAny;
 use crate::logger::{error, warn};
 use crate::vstate::memory::GuestMemoryMmap;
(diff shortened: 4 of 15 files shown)


majek:
Pulling in @bchalios @kalyazin. Let me know what you think.

bchalios:
Thanks a lot for the contribution @majek.

My main comment has to do with this:

> From Firecracker point of view, Linux vhost interface is basically another implementation of virtqueue (or Virtual Queue) from Virtio spec

I 'm interested to understand why you think that is the case. It is true that with `vhost-net` we will rely on Linux's implementation for the virtqueues, but IMHO the current Firecracker implementation provides enough interfaces to setup the Linux machinery. It is true that after that we won't need to use the Firecracker's queue implementation but I just can't understand how the changes in this PR make it easier to drive `vhost-net`.

majek:
@bchalios Absolutely. The essence of the change is:

```.patch
pub trait VirtioDevice: AsAny {
...
    /// Returns the device queues.
-    fn queues(&self) -> &[Queue];
+    fn queues(&self) -> QueueIter;
```

VirtioDevice is implemented by all the devices firecracker supports. There are at least two "customers" for that interface. First is virtio / firecracker device stack. This is the part we want to keep on using in vhost case. We want device setup, management, mmio registers, interrupts, all of these things VirtioDevice / mmio.rs do well and there is no point in changing this.

Then there is the second user of VirtioDevice, which is device driver internally. The device driver wants to, roughly, queue and dequeue descriptors on the vqueue. For example net driver does this - it pops/pushes data onto vqueue. For vhost we don't want that. In vhost case the data path of the queue is transparent and there is just no implementation of `.pop()`.  

Therefore we can't just return `&[Queue]` in vhost case. We need vhost-specific implementation there.

In next PR we'll propose an enum over QueueIter, that is able to expose the two implementations of vqueue to the users of VirtioDevice like mmio.rs:

```.patch
-pub type QueueIter<'a> = std::slice::Iter<'a, Queue>;
-pub type QueueIterMut<'a> = std::slice::IterMut<'a, Queue>;

+#[derive(Debug)]
+pub struct QueueIter<'a> {
 +   inner: QueueIterInner<'a>,
+}
+
+#[derive(Debug)]
+enum QueueIterInner<'a> {
+    Virtio(std::slice::Iter<'a, Queue>),
+    Vhost(std::slice::Iter<'a, VhostQueue>),
+}
```

Does this make more sense now? 

In summary: we're going to propose an enum over `QueueIter` in a subsequent PR. Even if the enum is controversial, change in this PR is still beneficial, as it draws clearer boundary between Queue and  mmio.rs, consider this:

```.patch
-                    0x44 => self.with_queue(0, |q| u32::from(q.ready)),
+                    0x44 => self.with_queue(0, |q| u32::from(q.ready())),
```

Pretty trivial, but allows Queue to be abstracted away in the future. Without this abstraction doing any changes to Queue implementation is impossible.

roypat:
Hey all, 
If I'm understanding correctly, the problem we're trying to solve here is that Firecracker's current `struct Queue` contains both the virtio protocol configuration fields that are common to the existing devices and future vhost devices, and then also bits related to actually processing virtio requests that will not be needed by vhost devices (because the processing happens outside of Firecracker there). I think we can solve this problem in a slightly simpler way that does not require getters/setters (I'll get into why I don't think these should happen at the end), enums, or a `struct VhostQueue`. Essentially, my proposal would be

<details>
<summary> Diff to queue.rs/device.rs </summary>

(please note that this diff is minimal to bring across the idea, it doesnt include all the related bits needed to get a compiling build)

```diff
diff --git a/src/vmm/src/devices/virtio/device.rs b/src/vmm/src/devices/virtio/device.rs
index 80135576e..2ca50ce1f 100644
--- a/src/vmm/src/devices/virtio/device.rs
+++ b/src/vmm/src/devices/virtio/device.rs
@@ -12,7 +12,7 @@ use std::sync::Arc;
 use utils::eventfd::EventFd;
 
 use super::mmio::{VIRTIO_MMIO_INT_CONFIG, VIRTIO_MMIO_INT_VRING};
-use super::queue::Queue;
+use super::queue::{Queue, QueueConfig};
 use super::ActivateError;
 use crate::devices::virtio::AsAny;
 use crate::logger::{error, warn};
@@ -110,11 +110,8 @@ pub trait VirtioDevice: AsAny + Send {
     /// The virtio device type.
     fn device_type(&self) -> u32;
 
-    /// Returns the device queues.
-    fn queues(&self) -> &[Queue];
-
-    /// Returns a mutable reference to the device queues.
-    fn queues_mut(&mut self) -> &mut [Queue];
+    fn nth_queue_config(&self, n: u32) -> Option<&QueueConfig>;
+    fn nth_queue_config_mut(&mut self, n: u32) -> Option<&mut QueueConfig>;
 
     /// Returns the device queues event fds.
     fn queue_events(&self) -> &[EventFd];
diff --git a/src/vmm/src/devices/virtio/queue.rs b/src/vmm/src/devices/virtio/queue.rs
index 45cf6349f..6cb9da1a4 100644
--- a/src/vmm/src/devices/virtio/queue.rs
+++ b/src/vmm/src/devices/virtio/queue.rs
@@ -176,12 +176,7 @@ impl<'a> Iterator for DescriptorIterator<'a> {
     }
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
-/// A virtio queue's parameters.
-pub struct Queue {
-    /// The maximal size in elements offered by the device
-    pub(crate) max_size: u16,
-
+pub struct QueueConfig {
     /// The queue size in elements the driver selected
     pub size: u16,
 
@@ -196,6 +191,15 @@ pub struct Queue {
 
     /// Guest physical address of the used ring
     pub used_ring: GuestAddress,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+/// A virtio queue's parameters.
+pub struct Queue {
+    pub config: QueueConfig,
+
+    /// The maximal size in elements offered by the device
+    pub(crate) max_size: u16,
 
     pub(crate) next_avail: Wrapping<u16>,
     pub(crate) next_used: Wrapping<u16>,
```

</details>

The idea here is that, similar to a vhost device, the mmio code doesnt need to know about the Firecracker specific bits of our queue implementation. By factoring this part out into a common struct, and have the abstract iterface of `VirtioDevice` only know about `VirtioConfig`, we achieve the same goal of encapsulating Firecracker-devices-specific queue fields away from mmio code and vhost devices (e.g. in this world, a vhost device would not store a `Vec<Queue>`, like Firecracker's traditional virtio devices to, but rather a `Vec<VirtioConfig>`).

The change in interface from `fn queue_configs() -> &[QueueConfig]` to `fn nth_queue_config(&self, n: usize) -> Option<&QueueConfig>` is a bit unfortunate, but we cannot get a `&[QueueConfig]` from a `Vec<Queue>`. There will probably be a need for some sort of iterator API, á la 

```rs
trait VirtioDevice {
    ...

    fn queue_iterator(&self) -> QueueIterator<'_> {
        QueueIterator(self, 0)
    }
}

struct QueueIterator<'a>(&'a dyn VirtioDevice, usize);
impl<'a> Iterator for QueueIterator<'a> {
    type Item = &'a QueueConfig;
    fn next(&mut self) -> Option<Self::Item> {
        let ret = self.0.nth_queue_config(self.1);
        self.1 += 1;
        ret
    }
}
```

but in general, I think it'd be a cleaner way to solve this problem. 

As for getters and setters, I think they are generally considered unidiomatic in Rust, as unlike in most other languages, they are not a zero-cost abstraction. Only exposing a getter/setter looses expressiveness, as they prohibit pattern matching, and taking ownership of multiple struct fields at once. This is because in Rust, there are 3(ish) ways to "get" some field: Taking ownership and moving out through some `fn take_x(self) -> X { self.x }`, by immutable reference `fn get_x(&self) -> &X {&self.x}` and by mutable reference `fn get_x_mut(&mut self) -> &mut X {&mut self.x}` (but this one blurs the line to "setter"). Yet none of them allow me to pattern match multiple fields, or to move multiple fields out of the struct at once. I'm not saying that we'd run into these restrictions here, but its why I would generally choose direct field access over setters and getters in Rust, especially for "data only" types like `struct VirtioConfig`.

kornelski:
We've actually initially implemented the "get nth queue" approach first, but eventually moved away from it, because it turned out to be less usable. 

This is because it's not possible to build an `Iterator` on top of it without use of `unsafe`. Repeated calls to a getter don't give a guarantee that items' lifetimes are disjoint, and such method works as a *streaming* iterator abstraction, but not Rust `Iterator`. OTOH the `Iterator` is more flexible, and has an `nth()` method too.

----

I agree that getters in Rust are not ideal, but once we have multiple queue implementations we'd use `dyn Queue` as the iterator item, which *requires* getters, because Rust doesn't support fields in traits. Fortunately fields on the `Queue` are small and copyable, so problems with simultaneous access to multiple fields don't happen.

majek:
Ok, I think we can avoid this abstractions if we reuse the block-vhost-user design.

majek:
Followup https://github.com/firecracker-microvm/firecracker/pull/4461

Issue #4162, opened 2023-10-10
[Feature Request] Are there plans to publish the snapshot crate?


bergwolf:
Hello firecracker maintainers,

First of all, thank you for creating such a wonderful project and sharing it with the rest of the world with open source. There are a lot of rust projects benefiting from firecracker and we are one of them :)

I'm writing in the hat of maintainers of the [`fuse-backend-rs`](https://github.com/cloud-hypervisor/fuse-backend-rs/pull/149) project. It is a library used by the [Nydus image service](https://nydus.dev/) to provide FUSE file system support on both Linux and MacOS. Recently we've been adding [hot-upgrade ability](https://github.com/cloud-hypervisor/fuse-backend-rs/pull/149) to the library, which heavily depends on firecracker's [`snapshot`](https://github.com/firecracker-microvm/firecracker/tree/main/src/snapshot) crate.

Since we do publish `fuse-backend-rs` on crates.io (https://crates.io/crates/fuse-backend-rs), we cannot really have a dependency on unpublished crates. Therefore, we'd love to seek advice. Are there plans to publish the snapshot crate on crate.io at some point? Or do you think it is better for us to just fork that port of code as it is not supposed to be maintained as a public library at the moment? If you have any other suggestions/thoughts, please do share them. Thanks a lot!

Cheers,
Tao

bchalios:
Hi @bergwolf,

Thanks for getting in touch, it is always nice to see people using (parts of) Firecracker :)

We discussed this internally and we decided that we would not like to publish the `snapshot` crate in crates.io at the moment. Doing so, would mean that we need to take a commitment to maintain it in a way that's more complicated for us. We would need for example to be careful with changes in its public API so that we do not break downstream projects like yours. At the moment, this would slow us down considerably.

The snapshot feature is still in Developer preview and we are still trying to consolidate API and functionality in something that makes sense for the particular case of Firecracker. For example, we have announced changes in https://github.com/firecracker-microvm/firecracker/discussions/4137 that are related with the versioning and parts of the functionality. 
TLDR; the biggest changes are:
1. We will decouple Firecracker versions from snapshot versions
2. We will stop using the `versionize` crate for (de)serializing state
3. We will not support creating snapshots for older versions of the snapshot format.

Given that, I believe that the best way forward for you is that you fork the crate and continue development in that fork.

I hope that makes sense for you.

bergwolf:
Hi @bchalios, 

[...]

PR #4319, opened 2023-12-13
chore: support ppc64le and riscv64

Desiki-high:
## Changes

Add `BASE_MAGIC_ID` in `ppc64le` and `riscv64` arches to support using `snapshot` in `ppc64le` and `riscv64`.

## Reason
Ref:https://github.com/firecracker-microvm/firecracker/issues/4162 https://github.com/kata-containers/dbs-snapshot/pull/5 https://github.com/dragonflyoss/nydus/pull/1510 

Our project depends on firecracker's [snapshot](https://github.com/firecracker-microvm/firecracker/tree/main/src/snapshot) crate, we need to work on different arches like `ppc64le` and `riscv64`. I think many other projects must also work on `ppc64le` and `riscv64`.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/snapshot/src/lib.rs
+++ b/src/snapshot/src/lib.rs
@@ -43,6 +43,12 @@ const BASE_MAGIC_ID: u64 = 0x0710_1984_8664_0000u64;
 #[cfg(target_arch = "aarch64")]
 const BASE_MAGIC_ID: u64 = 0x0710_1984_AAAA_0000u64;
 
+#[cfg(target_arch = "powerpc64")]
+const BASE_MAGIC_ID: u64 = 0x0710_1984_CC64_0000u64;
+
+#[cfg(target_arch = "riscv64")]
+const BASE_MAGIC_ID: u64 = 0x0710_1984_C564_0000u64;
+
 /// Error definitions for the Snapshot API.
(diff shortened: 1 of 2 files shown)


roypat:
Hi @Desiki-high, 
thank you for your contribution. Firecracker does not support ppc64le and riscv64, so I don't think having these two constants here is a very good idea, as it might send a wrong message about which platforms Firecracker supports. From my understanding of #4162, you were already planning on having a fork of this crate, is there a specific reason why this change cannot live in that fork?

Desiki-high:
> Hi @Desiki-high, thank you for your contribution. Firecracker does not support ppc64le and riscv64, so I don't think having these two constants here is a very good idea, as it might send a wrong message about which platforms Firecracker supports. From my understanding of #4162, you were already planning on having a fork of this crate, is there a specific reason why this change cannot live in that fork?

Thanks for the review @roypat. 
We only consider if the Firecracker needs this contribution, because this is useful for many projects which dependts on firecracker's [snapshot](https://github.com/firecracker-microvm/firecracker/tree/main/src/snapshot) crate. However it does lead to make a wrong message about which platforms Firecracker supports, so we will only use it in our fork crate. Thanks for your review again.

PR #4338, opened 2023-12-20
chore: Add x86_64 and aarch64 targets to devctr 

roypat:
Unconditionally add x86_64-unknown-linux-musl and
aarch64-unknown-linux-musl targets to devctr. This is because otherwise,
any cargo invocation would try to install the missing one anyway (due to
being listed in rust-toolchain.toml, which needs them listed to
on-the-fly download them for A/B-tests across toolchain versions), but
this installation can only succeed in a privileged docker container
(e.g. `./tools/devtool build` would always fail).
## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


.buildkite/pipeline_pr.py
```diff
@@ -110,11 +110,14 @@
 steps = [step_style]
 changed_files = get_changed_files("main")
 
-# run sanity build of devtool if Dockerfile is changed
-if any(x.name == "Dockerfile" for x in changed_files):
+# run sanity build of devtool if Dockerfile or devtool are changed
+if any(x.name in "Dockerfile" for x in changed_files):
```
zulinx86, line 114: It looks missing "devtool" in the list. It runs if "Dockerfile" or "devtool" are changed, doesn't it?

roypat: ooops, yeah, originally I changed this `if` because I thought it did the `make_release` test run, and apparently I messed up reverting it. Let me fix

roypat: fixed

tools/devctr/Dockerfile
```diff
@@ -107,8 +107,10 @@ RUN cd /tmp/poetry && \
 # - Install the Rust toolchain. Kani only work on x86, so only try to install it there
 # - Build and install crosvm (used as vhost-user-blk backend)
 # - Clean up cargo compilation directories
+# - Always install both x86_64 and aarch64 musl targets, as our rust-toolchain.toml would force on-the-fly installation of both anyway
 RUN curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain "$RUST_TOOLCHAIN" \
-    && rustup target add $ARCH-unknown-linux-musl \
+    && rustup target add x86_64-unknown-linux-musl \
```
pb8o, line 112: How large is the delta? Does it make a difference if we specify the minimal profile here too?

tools/devctr/Dockerfile
```diff
@@ -107,8 +107,10 @@ RUN cd /tmp/poetry && \
 # - Install the Rust toolchain. Kani only work on x86, so only try to install it there
 # - Build and install crosvm (used as vhost-user-blk backend)
 # - Clean up cargo compilation directories
+# - Always install both x86_64 and aarch64 musl targets, as our rust-toolchain.toml would force on-the-fly installation of both anyway
```
pb8o, line 110: Just occurred to me.. could we move devtool to always specify --privileged? Would we lose anything?

tools/devctr/Dockerfile
roypat: Its not a toolchain, just a cross-compilation target, so its essentially just the standard library (a few megabytes)

tools/devctr/Dockerfile
roypat: Mh, probably not a problem, but I kinda already got annoyed by it always downloading the target when I run any docker command, so I'd like to just have both in the container anyway

bchalios:
It's probably me not understanding something, but when we do A/B testing where A is on an older toolchain, we'll still be using the new container image which, IIUC, will have the newer toolchain only, so we'd still need to download the older toolchain for being able to test A. Is my understanding correct? If the answer is "yes", how does this PR solve this scenario?

roypat:
> It's probably me not understanding something, but when we do A/B testing where A is on an older toolchain, we'll still be using the new container image which, IIUC, will have the newer toolchain only, so we'd still need to download the older toolchain for being able to test A. Is my understanding correct? If the answer is "yes", how does this PR solve this scenario?

Yes, but that step happens in a privileged container, so it does not run into any problems

bchalios:
Thanks for the explanation @roypat!

zulinx86 (approved):

pb8o (approved):

PR #4340, opened 2023-12-21
Add option to run A/B-test on two arbitrary test-report.json files

roypat:
This allows comparisons of two test runs that were not orchestrated by the A/B-testing framework, for example if the two runs were done on different instances. Use this if the data you want to compare is from anything other than "two firecracker revisions tested on the same machine inside the same docker image".

## Changes

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tools/ab_test.py
+++ b/tools/ab_test.py
@@ -78,16 +78,11 @@ def extract_dimensions(emf):
     return {key: emf[key] for key in emf if key in dimension_list}
 
 
-def reemit_emf_and_get_data(log_entry: str, revision: str):
-    """Parses the given EMF log entry, and reemits it, overwriting the attached "git_commit_id" field
-    with the given revision
+def process_log_entry(emf: dict):
+    """Parses the given EMF log entry
 
     Returns the entries dimensions and its list-valued properties/metrics, together with their units
     """
-    emf = json.loads(log_entry)
-    emf["git_commit_id"] = revision
-    emit_raw_emf(emf)
-
     result = {
         key: (value, find_unit(emf, key))
         for key, value in emf.items()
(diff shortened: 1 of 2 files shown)


tools/ab_test.py
```diff
@@ -78,16 +78,12 @@ def extract_dimensions(emf):
     return {key: emf[key] for key in emf if key in dimension_list}
 
 
-def reemit_emf_and_get_data(log_entry: str, revision: str):
+def process_log_entry(emf: dict):
     """Parses the given EMF log entry, and reemits it, overwriting the attached "git_commit_id" field
     with the given revision
```
pb8o, line 83: I guess the comment does not match the code now?

tools/ab_test.py
```diff
@@ -109,22 +105,30 @@ def find_unit(emf: dict, metric: str):
...
 
-    Also reemits all EMF logs."""
+    If `reemit` is True, it also reemits all EMF logs to a local EMF agent."""
     # Dictionary mapping EMF dimensions to A/B-testable metrics/properties
     processed_emf = {}
 
-    report = json.loads(Path("test_results/test-report.json").read_text("UTF-8"))
+    report = json.loads(report_path.read_text("UTF-8"))
     for test in report["tests"]:
         for line in test["teardown"]["stdout"].splitlines():
```
pb8o, line 119: Do we have any tests that emit metrics during "call" phase? Could we collect both phases?

tools/ab_test.py
roypat: ahh, true, fixed!

tools/ab_test.py
roypat: We do not currently, because all metrics are emitted by the `metrics` fixture, which only flushes during teardown.

roypat: I've had a quick look at whether it'd be an easy change to consider both "teardown" and "call", but it seems that either the "call" key or the "stdout" subkey are not always present :(

pb8o (approved):

sudanl0 (approved):

PR #4341, opened 2023-12-21
doc(metrics): Make metrics related info easily accessible

sudanl0:
Present a list of metrics emitted by Firecracker and to make it easier to understand the metrics, provide links to the definition of each metrics.

Added a misc commit not related to the documentation to fix the validation of metrics.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

~- [ ] All added/changed functionality is tested.~

---

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -77,3 +77,56 @@ Otherwise, if the path points to a normal file, you can simply do:
 ```shell script
 cat metrics.file
 ```
+
+## Metrics emitted by Firecracker
+
+The metrics emitted by Firecracker are in JSON format.
+Below are the keys present in each metrics json object emitted by Firecracker:
+
+```
+"api_server"
+"balloon"
+"block"
+"deprecated_api"
+"entropy"
+"get_api_requests"
+"i8042"
+"latencies_us"
+"logger"
+"mmds"
+"net"
+"patch_api_requests"
+"put_api_requests"
+"rtc"
+"seccomp"
+"signals"
+"uart"
+"vcpu"
+"vhost_user_block"
+"vmm"
+"vsock"
+```
+
+Below table explains where Firecracker metrics are defined :
+
+| Metrics key | Device  | Additional comments |
+|-------|-------|-------|
+|   balloon                     |    [BalloonDeviceMetrics](../src/vmm/src/devices/virtio/balloon/metrics.rs)       | Respresent metrics for the Balloon device.|
+|   block                       |    [BlockDeviceMetrics](../src/vmm/src/devices/virtio/virtio_block/metrics.rs)    | Respresent aggregate metrics for Virtio Block device.|
+|   block_{block_drive_id}      |    [BlockDeviceMetrics](../src/vmm/src/devices/virtio/virtio_block/metrics.rs)    | Respresent Virtio Block device metrics for the endpoint `"/drives/{drive_id}"` e.g. `"block_rootfs":` represent metrics for the endpoint `"/drives/rootfs"`|
+|   i8042                       |    [I8042DeviceMetrics](../src/vmm/src/devices/legacy/i8042.rs)                   | Respresent Metrics specific to the i8042 device.|
+|   net                         |    [NetDeviceMetrics](../src/vmm/src/devices/virtio/net/metrics.rs)               | Respresent aggregate metrics for Virtio Net device.|
+|   net_{iface_id}              |    [NetDeviceMetrics](../src/vmm/src/devices/virtio/net/metrics.rs)               | Respresent Virtio Net device metrics for the endpoint `"/network-interfaces/{iface_id}"` e.g. `net_eth0` represent metrics for the endpoint `"/network-interfaces/eth0"`|
+|   rtc                         |    [RTCDeviceMetrics](../src/vmm/src/devices/legacy/serial.rs)                    | Respresent Metrics specific to the RTC device. `Note`: this is emitted only on `aarch64`.|
+|   uart                        |    [SerialDeviceMetrics](../src/vmm/src/devices/legacy/serial.rs)                 | Respresent Metrics specific to the serial device.|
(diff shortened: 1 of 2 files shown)


tests/host_tools/fcmetrics.py
```diff
@@ -304,6 +308,10 @@ def validate_missing_metrics(metrics):
                     pass
                 else:
                     raise error
+            else:
+                raise Exception(
+                    "Test expected an exception on removing rts:error_count but none was raised"
+                )
```
pb8o, line 314: Couldn't we just use `pytest.raises`? So the whole thing becomes:

```python
with pytest.raises(jsonschema.ValidationError, match="'error_count' is a required property"):
    jsonschema.validate(instance=metrics, schema=firecracker_metrics_schema)		                
```

docs/metrics.md
```diff
@@ -77,3 +77,55 @@ Otherwise, if the path points to a normal file, you can simply do:
...
+"vhost_user_block"
+"vmm"
+"vsock"
+```
+
+Below table explains where Firecracker metrics are defined :
+
+| Metrics key | Device  | Additional comments |
+|-------|-------|-------|
+|   balloon                     |    [BalloonDeviceMetrics](https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/devices/virtio/balloon/metrics.rs#L53)         | Respresent metrics for the Balloon device.                                                                                                                                          |
```
pb8o, line 113: How will we maintain that those links stay valid? I'm not sure it's worth it when one can just find all the metrics with grep: `git grep -P "struct \w+Metrics" src`

Also the links could be internal links rather than absolute links to github.

docs/metrics.md
```diff
@@ -77,3 +77,55 @@ Otherwise, if the path points to a normal file, you can simply do:
 ```shell script
 cat metrics.file
 ```
+
+## Metrics emitted by Firecracker
+
+Below are the metrics emitted by Firecracker:
```
bchalios, line 83: These are not the metrics. These are the Firecracker sub-systems for which we emit metrics

docs/metrics.md
bchalios: I think it's better if these are internal links, too. Maintaining the links is a problem indeed but we have this issue in various parts of the documentation, so I think we just need to live with it. Maybe we could check for a tool that crawls the documentation and checks if the links are still working.

bchalios (changes requested):

docs/metrics.md
bchalios: For example: https://stackoverflow.com/questions/32176318/github-tool-to-validate-links-in-markdown#34599982

docs/metrics.md
sudanl0: The 1st column is actually the key for the metric in the json object so changing this line to says sub-system would be confusing.

docs/metrics.md
roypat: We _could_ have an integration test that greps our markdown files for URLs and gives them all a friendly `HEAD` request, but not sure if that wouldn't be interpreted as a DoS by whichever websites the URLs lead to

sudanl0: [mlc](https://crates.io/crates/mlc) works perfectly for this purpose and running it locally shows we have quite a few broken links.
I have added a workflow which can install and run mlc on the new md files. Maybe we can add it to devctr and avoid installing it every time but I would like to know what you all think about it first.

sudanl0: there are still links like https://github.com/firecracker-microvm/firecracker/blob/main/docs/getting-started.md?plain=1#L79 which are hard to cover.

roypat: I like the idea, I'm just a bit worried about pinging so many websites in our CI. We do tend to have quite a lot of CI runs on each PR, so this could be quite noisy (especially because I think github actions dont get cancelled on new pushes). Another problem would be that now if some website that we link to goes dark, it would block all our PRs until it gets fixed (although I guess we can make it optional?). Local output also seems to contain quite a few false-positives (one where it fails to do a TLS handshake for some reason, and then a few where it cannot resolve things in `docs/images`, yet the images render fine).

Anyway, if you do end up building a new docker container before #4338 gets merged, feel free to overwrite the v69 container I built, I can redo my PR when I get back from PTO

sudanl0: the workflow already limits the validation to the file which is modified, I think we can further modify the workflow to search for links only in lines modified in the md files so that instead of pining for all links in the file it validates only the ones that are modified.

roypat: Mh, can we limit mlc to only verify github links (e.g. those starting with https://github.com and the special intra-repository links)? I don't think pinging GitHub from our CI is a big problem, we we no longer have to worry about external websites dying on us and blocking PRs (we might still want to check that, but lets not strive for perfection from the get-go when we can at least get a sizable chunk of the problem fixed easily).

roypat: As for #4338, that'll mean waiting until Jan 2nd, since I'll be on PTO until then. I don't particularly mind either way, I just don't want to block you from working on this between the years

sudanl0: I modified the code to validate only the links we modify in the PR and not the entire file. I tried running the code locally for an md file and verified that the code now only checks the 1 modified link out of the 3 links present in the file.

roypat: I don't think that's what we want, because our problems is not that we commit invalid links, our problem is that the links later on become invalid because of unrelated changes in the repository (e.g. the metrics links become invalid because we move the source files around). For mlc to solve that problem, it needs to run on the entire repository, not just the diff

sudanl0: ok yeah then I think it goes out of scope for the current PR. A CI test which validates all the links including embedded links like this https://github.com/firecracker-microvm/firecracker/blob/main/docs/getting-started.md?plain=1#L79 would be good to have. I'll remove the workflow changes and keep a backup just in case. I will create a github issue to track it separately.

roypat: Sounds good to me!

sudanl0: created https://github.com/firecracker-microvm/firecracker/issues/4342

bchalios: Yes, this is not something strictly related with this PR. To avoid the issues Patrick mentioned we can run this as a test of the nightly pipeline only.

bchalios (approved):

pb8o (approved):

PR #4346, opened 2023-12-28
Aggregate latency metrics for vm exits handling IO

sudanl0:
## Changes

- Add a new struct to record aggregate (min/max/sum) of time
    difference as a metric.
    Use the Aggregate metrics structure to record aggregate of
    IO and MMIO vm exits.
- Firecracker till now had a "group_metrics:key_metrics:value" pair and
    with new metrics for kvm exits this relation changes to "group_metrics:
    key_metrics:sub_key_metrics:value".
    Update the flush_fc_metrics_to_cw() to be able to parse this and make
    it future proof to support further sub_key_metrics levels.

## Reason

- latency of kvm exits will improve observability.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

~- [ ] Any required documentation changes (code and docs) are included in this PR.~

---

### DIFF
--- a/src/vmm/src/logger/metrics.rs
+++ b/src/vmm/src/logger/metrics.rs
@@ -692,7 +692,84 @@ impl SignalMetrics {
     }
 }
 
-/// Metrics specific to VCPUs' mode of functioning.
+/// Provides efficient way to record LatencyAggregateMetrics
+#[derive(Debug)]
+pub struct LatencyMetricsRecorder<'a> {
+    start_time: u64,
+    metric: &'a LatencyAggregateMetrics,
+}
+
+impl<'a> LatencyMetricsRecorder<'a> {
+    /// Const default construction.
+    fn new(metric: &'a LatencyAggregateMetrics) -> Self {
+        Self {
+            start_time: utils::time::get_time_us(utils::time::ClockType::Monotonic),
+            metric,
+        }
+    }
+}
+impl<'a> Drop for LatencyMetricsRecorder<'a> {
+    /// records aggregate (min/max/sum) for the given metric
+    /// This captures delta between self.start_time and current time
+    /// and updates min/max/sum metrics.
+    ///  self.start_time is recorded in new() and metrics are updated in drop
+    fn drop(&mut self) {
+        let delta_us =
+            utils::time::get_time_us(utils::time::ClockType::Monotonic) - self.start_time;
+        self.metric.sum_us.add(delta_us);
+        let min_us = self.metric.min_us.fetch();
+        let max_us = self.metric.max_us.fetch();
+        if (0 == min_us) || (min_us > delta_us) {
+            self.metric.min_us.store(delta_us);
+        }
+        if (0 == max_us) || (max_us < delta_us) {
+            self.metric.max_us.store(delta_us);
+        }
+    }
+}
+
+/// Used to record Aggregate (min/max/sum) of latency metrics
+#[derive(Debug, Default, Serialize)]
+pub struct LatencyAggregateMetrics {
+    /// represents minimum value of the metrics in microseconds
+    pub min_us: SharedStoreMetric,
+    /// represents maximum value of the metrics in microseconds
+    pub max_us: SharedStoreMetric,
+    /// represents sum of the metrics in microseconds
+    pub sum_us: SharedIncMetric,
+}
+impl LatencyAggregateMetrics {
+    /// Const default construction.
+    pub const fn new() -> Self {
+        Self {
+            min_us: SharedStoreMetric::new(),
+            max_us: SharedStoreMetric::new(),
+            sum_us: SharedIncMetric::new(),
+        }
+    }
+
+    /// returns a latency recorder which captures stores start_time
+    /// and updates the actual metrics at the end of recorders lifetime.
+    /// in short instead of below 2 lines :
+    /// 1st for start_time_us = get_time_us()
+    /// 2nd for delta_time_us = get_time_us() - start_time; and metrics.store(delta_time_us)
+    /// we have just `_m = metrics.record_latency_metrics()`
+    pub fn record_latency_metrics(&self) -> LatencyMetricsRecorder {
+        LatencyMetricsRecorder::new(self)
+    }
+}
+
+/// Structure provides Metrics specific to VCPUs' mode of functioning.
+/// Sample_count or number of kvm exits for IO and MMIO VM exits are covered by:
+/// `exit_io_in`, `exit_io_out`, `exit_mmio_read` and , `exit_mmio_write`.
+/// Count of other vm exits for events like shutdown/hlt/errors are
+/// covered by existing "failures" metric.
+/// The only vm exit for which sample_count is not covered is system
+/// event reset/shutdown but that should be fine since they are not
+/// failures and the vm is terminated anyways.
+/// LatencyAggregateMetrics only covers minimum, maximum and sum
+/// because average can be deduced from available metrics. e.g.
+/// dividing `exit_io_in_agg.sum_us` by exit_io_in` gives average of KVM exits handling input IO.
 #[derive(Debug, Default, Serialize)]
 pub struct VcpuMetrics {
     /// Number of KVM exits for handling input IO.
@@ -705,6 +782,14 @@ pub struct VcpuMetrics {
     pub exit_mmio_write: SharedIncMetric,
     /// Number of errors during this VCPU's run.
     pub failures: SharedIncMetric,
+    /// Provides Min/max/sum for KVM exits handling input IO.
+    pub exit_io_in_agg: LatencyAggregateMetrics,
+    /// Provides Min/max/sum for KVM exits handling output IO.
+    pub exit_io_out_agg: LatencyAggregateMetrics,
+    /// Provides Min/max/sum for KVM exits handling MMIO reads.
+    pub exit_mmio_read_agg: LatencyAggregateMetrics,
+    /// Provides Min/max/sum for KVM exits handling MMIO writes.
+    pub exit_mmio_write_agg: LatencyAggregateMetrics,
 }
 impl VcpuMetrics {
     /// Const default construction.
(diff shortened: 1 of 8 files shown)


src/vmm/src/logger/metrics.rs
```diff
@@ -78,6 +78,26 @@ use crate::devices::virtio::vhost_user_metrics;
 use crate::devices::virtio::virtio_block::metrics as block_metrics;
 use crate::devices::virtio::vsock::metrics as vsock_metrics;
 
+#[macro_export]
+/// records aggregate (min/max/sum) for the given $metrics
```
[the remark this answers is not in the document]
sudanl0: sample_count/number of kvm exits for IO and MMIO vm exits is covered by:
```
"exit_io_in",
"exit_io_out",
"exit_mmio_read",
"exit_mmio_write",
```
Count of other vm exits for events like shutdown/hlt/errors are covered by existing "failures" metric.
The only vm exit for which sample_count is not covered is system event reset/shutdown but that should be fine since they are not failures and the vm is terminated anyways.

src/vmm/src/vstate/vcpu/mod.rs
```diff
@@ -450,15 +450,21 @@ impl Vcpu {
...
                     }
                     Ok(VcpuEmulation::Handled)
                 }
                 VcpuExit::MmioWrite(addr, data) => {
                     if let Some(mmio_bus) = &self.kvm_vcpu.mmio_bus {
+                        let start_time =
+                            utils::time::get_time_us(utils::time::ClockType::Monotonic);
                         mmio_bus.write(addr, data);
                         METRICS.vcpu.exit_mmio_write.inc();
+                        record_aggregate_metrics!(METRICS.vcpu.exit_mmio_write_agg, start_time);
```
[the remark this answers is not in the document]
sudanl0: As per the list mentioned in https://github.com/rust-vmm/kvm-ioctls/blob/8f68bf6f6d974346c1fa725282deeafadace457b/src/ioctls/vcpu.rs#L85 we cover the relevant kvm exits and the rest are considered as failures. So it adds value to capture latency of only the below kvm exits on which the guest is waiting :
```
"exit_io_in",
"exit_io_out",
"exit_mmio_read",
"exit_mmio_write",
```

src/vmm/src/logger/metrics.rs
```diff
@@ -692,6 +712,27 @@ impl SignalMetrics {
     }
 }
 
+/// Used to record Aggregate metrics
+#[derive(Debug, Default, Serialize)]
+pub struct AggregateMetrics {
```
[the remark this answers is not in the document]
sudanl0: renamed to LatencyAggregateMetrics

src/vmm/src/logger/metrics.rs
```diff
@@ -78,6 +78,26 @@ use crate::devices::virtio::vhost_user_metrics;
 use crate::devices::virtio::virtio_block::metrics as block_metrics;
 use crate::devices::virtio::vsock::metrics as vsock_metrics;
 
+#[macro_export]
+/// records aggregate (min/max/sum) for the given $metrics
+/// This macro captures delta between start_time and current time
+/// and updates min/max/sum metrics so it expects that start_time
+/// is captured at the appropriate location.
+macro_rules! record_aggregate_metrics {
```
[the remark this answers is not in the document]
sudanl0: renamed to record_latency_summary

src/vmm/src/logger/metrics.rs
```diff
@@ -78,6 +78,26 @@ use crate::devices::virtio::vhost_user_metrics;
 use crate::devices::virtio::virtio_block::metrics as block_metrics;
 use crate::devices::virtio::vsock::metrics as vsock_metrics;
 
+#[macro_export]
```
ShadowCurse, line 81: `#[macro_export]` exports macro for the whole crate. Instead it would be better to use `pub(crate) use <macro_name>`. This way macro will be a part of the module and can be imported using normal `used mod::macro` syntax.

src/vmm/src/vstate/vcpu/mod.rs
```diff
@@ -450,15 +450,21 @@ impl Vcpu {
             Ok(run) => match run {
                 VcpuExit::MmioRead(addr, data) => {
                     if let Some(mmio_bus) = &self.kvm_vcpu.mmio_bus {
+                        let start_time =
+                            utils::time::get_time_us(utils::time::ClockType::Monotonic);
```
ShadowCurse, line 454: Maybe we can move this (and `record_latency_summary` line) into `LatencyAggregateMetricWriter`(naming is hard) so that it gets the current time on creation and writes to the metric on drop?

src/vmm/src/logger/metrics.rs
```diff
@@ -78,6 +78,26 @@ use crate::devices::virtio::vhost_user_metrics;
...
+        let min_us = $metrics.min_us.fetch();
+        let max_us = $metrics.max_us.fetch();
+        if (0 == min_us) || (min_us > delta_us) {
+            $metrics.min_us.store(delta_us);
+        }
+        if (0 == max_us) || (max_us < delta_us) {
+            $metrics.max_us.store(delta_us);
+        }
+    };
+}
```
roypat, line 99: Could this be
```suggestion
impl LatencyAggregateMetrics {
    pub fn record_latency_summary(&mut self, start_time: u64) {
        let delta_us = utils::time::get_time_us(utils::time::ClockType::Monotonic) - start_time;
        self.sum_us.add(delta_us);
        let min_us = self.min_us.fetch();
        let max_us = self.max_us.fetch();
        if (0 == min_us) || (min_us > delta_us) {
            self.min_us.store(delta_us);
        }
        if (0 == max_us) || (max_us < delta_us) {
            self.max_us.store(delta_us);
        }
    }
}
```

src/vmm/src/logger/metrics.rs
sudanl0: removed the macro and updated code as per other comments.

src/vmm/src/vstate/vcpu/mod.rs
sudanl0: Thanks for the suggestion, updated the code accordingly

src/vmm/src/logger/metrics.rs
sudanl0: Thanks for the suggestion, updated the code accordingly

ShadowCurse (dismissed):

src/vmm/src/logger/metrics.rs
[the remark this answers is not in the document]
pb8o: Thanks for the explanation... could we document that in the code, or should that info be in metrics.md?

ShadowCurse (approved):

src/vmm/src/logger/metrics.rs
```diff
@@ -692,7 +692,84 @@ impl SignalMetrics {
...
+            metric,
+        }
+    }
+}
+impl<'a> Drop for LatencyMetricsRecorder<'a> {
+    /// records aggregate (min/max/sum) for the given metric
+    /// This captures delta between self.start_time and current time
+    /// and updates min/max/sum metrics.
+    ///  self.start_time is recorded in new() and metrics are updated in drop
+    fn drop(&mut self) {
```
bchalios, line 716: It seems like the assumption here is that `drop()` will be called once we exit the function that created `LatencyMetricsRecorder` variable. Is there any type of optimization that might cause the `drop` method to be executed at a different time? Maybe @roypat or @JonathanWoollett-Light know something about it?

src/vmm/src/vstate/vcpu/mod.rs
```diff
@@ -450,13 +450,15 @@ impl Vcpu {
             Ok(run) => match run {
                 VcpuExit::MmioRead(addr, data) => {
                     if let Some(mmio_bus) = &self.kvm_vcpu.mmio_bus {
+                        let _metric = METRICS.vcpu.exit_mmio_read_agg.record_latency_metrics();
```
bchalios, line 453: Why only x86?

tests/integration_tests/performance/test_network_ab.py
```diff
@@ -96,10 +96,10 @@ def test_network_latency(network_microvm, metrics, iteration):
         assert rc == 0, stderr
 
         samples.extend(consume_ping_output(ping_output))
+    fcmetrics.stop()
```
bchalios, line 99: It is not clear to me from the commit message why this change fixes the CI issue. I'm probably missing something, but could you try to explain it a bit better?

bchalios (commented):
LGTM, Main question is why we do this only for x86?

src/vmm/src/logger/metrics.rs
sudanl0: drop will be called when the lifetime of the variable end i.e. when the brackets {} end.
I tried to add some comments on this [here](https://github.com/firecracker-microvm/firecracker/pull/4346/files#diff-e13e5029c4b049ea5100ce89b82d716a8f1d71241aa29344b57c793441509722R751)

src/vmm/src/vstate/vcpu/mod.rs
sudanl0: because we don't have these cases for aarch64 https://github.com/sudanl0/firecracker/blob/main/src/vmm/src/vstate/vcpu/aarch64.rs#L231

src/vmm/src/logger/metrics.rs
roypat: The drop scopes are specified in https://doc.rust-lang.org/reference/destructors.html, so I'd say anything that differs from that would be a rustc bug

tests/integration_tests/performance/test_network_ab.py
sudanl0: Unfortunately, since the issue was intermittent, it was very hard to reproduce and debug so, I was not able to find a rational explanation why this fixed the issue. I changed the sequence because I was working with metrics and that is the only difference I saw in the ci framework at that time; fcmetrics was a new thread added to the ci so I suspected it could have some kind of affect on the test but from the multiple tests I tried the thread didn't seem to cause any issue but after the change in sequence I didn't see the issue in the tests so I thought it would be useful to add it.
Worst case scenario is that the issue is still there and I was just not able to reproduce it, in this case I think its still a CI issue and not Firecracker so I feel the change to rearrange the sequence should be ok to have.

src/vmm/src/vstate/vcpu/mod.rs
bchalios: Oh my bad, I thought this file was x86 specific. Ignore me!

bchalios (approved):

pb8o (approved):

PR #4358, opened 2024-01-11
chore(docs): deprecate --start-time-cpu-us parameter

pb8o:
## Changes

Deprecate the `--start-time-cpu-us` parameter.

## Reason

It is always 0 so it does not provide any useful information.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,14 @@ and this project adheres to
   handler. Each memory region object now contains a `page_size_kib` field. See
   also the [hugepages documentation](docs/hugepages.md).
 
+### Deprecated
+
+- Firecracker's `--start-time-cpu-us` parameter is deprecated and will be
+  removed in v2.0 or later. It is used by the jailer to pass the value that
+  should be subtracted from the CPU time, but in practice it is always 0. The
+  parameter was never meant to be used by end customers, and we recommend doing
+  any such time adjustments outside Firecracker.
+
 ### Fixed
 
 - [#4409](https://github.com/firecracker-microvm/firecracker/pull/4409): Fixed a
--- a/src/firecracker/src/main.rs
+++ b/src/firecracker/src/main.rs
@@ -22,7 +22,7 @@ use utils::terminal::Terminal;
 use utils::validators::validate_instance_id;
 use vmm::builder::StartMicrovmError;
 use vmm::logger::{
-    debug, error, info, LoggerConfig, ProcessTimeReporter, StoreMetric, LOGGER, METRICS,
+    debug, error, info, warn, LoggerConfig, ProcessTimeReporter, StoreMetric, LOGGER, METRICS,
 };
 use vmm::persist::SNAPSHOT_VERSION;
 use vmm::resources::VmResources;
@@ -397,6 +397,7 @@ fn main_exec() -> Result<(), MainError> {
         });
 
         let start_time_cpu_us = arguments.single_value("start-time-cpu-us").map(|s| {
+            warn!("The --start-time-cpu-us argument is deprecated");
             s.parse::<u64>()
                 .expect("'start-time-cpu-us' parameter expected to be of 'u64' type.")
         });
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `VcpuExit::IoOut`. The average for these VM exits is not emitted since
   it can be deduced from the available emitted metrics.
 
+### Deprecated
+
+- Firecracker's `--start-time-cpu-us` parameter is deprecated and will be
```
sudanl0, line 20: should we also add a deprecation print to highlight when someone uses this parameter?

pb8o: It is always use by the jailer so it would always get printed but customers wouldn't have a way to supress it.

sudanl0: I mean if customers use Firecracker without our jailer then they can notice the new deprecation print.
For customers using our jailer, yes they wouldn't be able to suppress the print but would having an extra print at the start be an issue for them?

roypat: > It is always use by the jailer so it would always get printed but customers wouldn't have a way to suppress it.

If the jailer stopped using it, would that be a customer visible change?

pb8o: Not as far as I can see. The only thing this affects is the `api_server.process_startup_time_cpu_us` metric and it will continue to have the same value.

roypat: Could we then not use it anymore from the jailer, and add the deprecation warning? That way, customers would only see the deprecation warning if they explicitly use the flag.

sudanl0: it makes sense to to stop using it in the jailer as well since we are deprecating it.

src/firecracker/src/main.rs
```diff
@@ -397,6 +398,7 @@ fn main_exec() -> Result<(), MainError> {
         });
 
         let start_time_cpu_us = arguments.single_value("start-time-cpu-us").map(|s| {
+            log_dev_preview_warning("start-time-cpu-us", "start");
```
roypat, line 401: mh, not sure if "dev preview" is the correct message to print here :thinking: In Firecracker we just print deprecations as `warn!` logs, e.g. 
```suggestion
            warn("The --start-time-cpu-us argument is deprecated");
```

pb8o: doh! ok coffee is not making effect

roypat (dismissed):

roypat (dismissed):

CHANGELOG.md
pb8o: Done

CHANGELOG.md
```diff
@@ -28,6 +28,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Firecracker snapshots (up to Firecracker version v1.6.0) incompatible with
   the current Firecracker version.
 
+### Deprecated
+
+- Firecracker's `--start-time-cpu-us` parameter is deprecated and will be
+  removed in v2.0. It is used by the jailer to pass the value that should be
```
kalyazin, line 34: nit: not sure it's necessary to promise to remove it in v2.0.

CHANGELOG.md
```diff
@@ -28,6 +28,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  subtracted from the CPU time, but in practice it is always 0. The parameter
+  was never meant to be used by end customers, but we recommend doing any such
```
kalyazin, line 36: nit: the `but` is a little bit confusing. Is it supposed to urge customers to take an action that they haven't been taking so far?

CHANGELOG.md
pb8o: How about `will be removed in v2.0 or later`?

CHANGELOG.md
kalyazin: yes, sounds good

roypat (approved):

kalyazin (approved):

PR #4373, opened 2024-01-18
Uffd protocol update

ShadowCurse:
## Changes
Update to uffd handler used in tests to handle multiple uffd objects send through the UDS. This upgrade is needed for supporting snapshot restore with vhost-used devices. During restoration process FC will ask vhost-user devices to give it their uffds and then send them to global uffd handler.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/firecracker/examples/uffd/uffd_utils.rs
+++ b/src/firecracker/examples/uffd/uffd_utils.rs
@@ -55,7 +55,7 @@ pub enum MemPageState {
 }
 
 impl UffdPfHandler {
-    pub fn from_unix_stream(stream: UnixStream, backing_buffer: *const u8, size: usize) -> Self {
+    pub fn from_unix_stream(stream: &UnixStream, backing_buffer: *const u8, size: usize) -> Self {
         let mut message_buf = vec![0u8; 1024];
         let (bytes_read, file) = stream
             .recv_with_fd(&mut message_buf[..])
@@ -213,5 +213,5 @@ pub fn create_pf_handler() -> UffdPfHandler {
 
     let (stream, _) = listener.accept().expect("Cannot listen on UDS socket");
 
-    UffdPfHandler::from_unix_stream(stream, memfile_buffer, size)
+    UffdPfHandler::from_unix_stream(&stream, memfile_buffer, size)
 }
--- a/src/firecracker/examples/uffd/uffd_utils.rs
+++ b/src/firecracker/examples/uffd/uffd_utils.rs
@@ -11,7 +11,7 @@ use std::os::unix::net::UnixStream;
 use std::ptr;
 
 use serde::Deserialize;
-use userfaultfd::Uffd;
+use userfaultfd::{Error, Event, Uffd};
 use utils::get_page_size;
 use utils::sock_ctrl_msg::ScmSocket;
 
@@ -33,19 +33,6 @@ pub struct GuestRegionUffdMapping {
     pub offset: u64,
 }
 
-#[derive(Debug)]
-struct MemRegion {
-    mapping: GuestRegionUffdMapping,
-    page_states: HashMap<u64, MemPageState>,
-}
-
-#[derive(Debug)]
-pub struct UffdPfHandler {
-    mem_regions: Vec<MemRegion>,
-    backing_buffer: *const u8,
-    pub uffd: Uffd,
-}
-
 #[derive(Debug, Clone)]
 pub enum MemPageState {
     Uninitialized,
@@ -54,7 +41,20 @@ pub enum MemPageState {
     Anonymous,
 }
 
-impl UffdPfHandler {
+#[derive(Debug)]
+struct MemRegion {
+    mapping: GuestRegionUffdMapping,
+    page_states: HashMap<u64, MemPageState>,
+}
+
+#[derive(Debug)]
+pub struct UffdHandler {
+    mem_regions: Vec<MemRegion>,
+    backing_buffer: *const u8,
+    uffd: Uffd,
+}
+
+impl UffdHandler {
     pub fn from_unix_stream(stream: &UnixStream, backing_buffer: *const u8, size: usize) -> Self {
         let mut message_buf = vec![0u8; 1024];
         let (bytes_read, file) = stream
@@ -83,6 +83,10 @@ impl UffdPfHandler {
         }
     }
 
+    pub fn read_event(&mut self) -> Result<Option<Event>, Error> {
+        self.uffd.read_event()
+    }
+
     pub fn update_mem_state_mappings(&mut self, start: u64, end: u64, state: &MemPageState) {
         for region in self.mem_regions.iter_mut() {
             for (key, value) in region.page_states.iter_mut() {
@@ -91,36 +95,6 @@ impl UffdPfHandler {
                 }
             }
         }
-    }
-
-    fn populate_from_file(&self, region: &MemRegion, dst: u64, len: usize) -> (u64, u64) {
-        let offset = dst - region.mapping.base_host_virt_addr;
-        let src = self.backing_buffer as u64 + region.mapping.offset + offset;
-
-        let ret = unsafe {
-            self.uffd
-                .copy(src as *const _, dst as *mut _, len, true)
-                .expect("Uffd copy failed")
-        };
-
-        // Make sure the UFFD copied some bytes.
-        assert!(ret > 0);
-
-        (dst, dst + len as u64)
-    }
-
-    fn zero_out(&mut self, addr: u64) -> (u64, u64) {
-        let page_size = get_page_size().unwrap();
-
-        let ret = unsafe {
-            self.uffd
-                .zeropage(addr as *mut _, page_size, true)
-                .expect("Uffd zeropage failed")
-        };
-        // Make sure the UFFD zeroed out some bytes.
-        assert!(ret > 0);
-
-        (addr, addr + page_size as u64)
     }
 
     pub fn serve_pf(&mut self, addr: *mut u8, len: usize) {
@@ -160,6 +134,36 @@ impl UffdPfHandler {
             addr
         );
     }
+
+    fn populate_from_file(&self, region: &MemRegion, dst: u64, len: usize) -> (u64, u64) {
+        let offset = dst - region.mapping.base_host_virt_addr;
+        let src = self.backing_buffer as u64 + region.mapping.offset + offset;
+
+        let ret = unsafe {
+            self.uffd
+                .copy(src as *const _, dst as *mut _, len, true)
+                .expect("Uffd copy failed")
+        };
+
+        // Make sure the UFFD copied some bytes.
+        assert!(ret > 0);
+
+        (dst, dst + len as u64)
+    }
+
+    fn zero_out(&mut self, addr: u64) -> (u64, u64) {
+        let page_size = get_page_size().unwrap();
+
+        let ret = unsafe {
+            self.uffd
+                .zeropage(addr as *mut _, page_size, true)
+                .expect("Uffd zeropage failed")
+        };
+        // Make sure the UFFD zeroed out some bytes.
+        assert!(ret > 0);
+
+        (addr, addr + page_size as u64)
+    }
 }
 
 #[derive(Debug)]
@@ -168,7 +172,7 @@ pub struct Runtime {
     backing_file: File,
     backing_memory: *mut u8,
     backing_memory_size: usize,
-    uffds: HashMap<i32, UffdPfHandler>,
+    uffds: HashMap<i32, UffdHandler>,
 }
 
 impl Runtime {
@@ -207,7 +211,7 @@ impl Runtime {
     /// When uffd is polled, page fault is handled by
     /// calling `pf_event_dispatch` with corresponding
     /// uffd object passed in.
-    pub fn run(&mut self, pf_event_dispatch: impl Fn(&mut UffdPfHandler)) {
+    pub fn run(&mut self, pf_event_dispatch: impl Fn(&mut UffdHandler)) {
         let mut pollfds = vec![];
 
         // Poll the stream for incoming uffds
@@ -240,7 +244,7 @@ impl Runtime {
                     nready -= 1;
                     if pollfds[i].fd == self.stream.as_raw_fd() {
                         // Handle new uffd from stream
-                        let handler = UffdPfHandler::from_unix_stream(
+                        let handler = UffdHandler::from_unix_stream(
                             &self.stream,
                             self.backing_memory,
                             self.backing_memory_size,
--- a/src/firecracker/examples/uffd/uffd_utils.rs
+++ b/src/firecracker/examples/uffd/uffd_utils.rs
@@ -55,7 +55,7 @@ pub enum MemPageState {
 }
 
 impl UffdPfHandler {
-    pub fn from_unix_stream(stream: UnixStream, data: *const u8, size: usize) -> Self {
+    pub fn from_unix_stream(stream: UnixStream, backing_buffer: *const u8, size: usize) -> Self {
         let mut message_buf = vec![0u8; 1024];
         let (bytes_read, file) = stream
             .recv_with_fd(&mut message_buf[..])
@@ -78,7 +78,7 @@ impl UffdPfHandler {
 
         Self {
             mem_regions,
-            backing_buffer: data,
+            backing_buffer,
             uffd,
         }
     }
@@ -120,7 +120,7 @@ impl UffdPfHandler {
         // Make sure the UFFD zeroed out some bytes.
         assert!(ret > 0);
 
-        return (addr, addr + page_size as u64);
+        (addr, addr + page_size as u64)
     }
 
     pub fn serve_pf(&mut self, addr: *mut u8, len: usize) {
@@ -151,9 +151,7 @@ impl UffdPfHandler {
                     self.update_mem_state_mappings(start, end, &MemPageState::Anonymous);
                     return;
                 }
-                None => {
-                    ();
-                }
+                None => {}
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -162,6 +162,110 @@ impl UffdPfHandler {
...
+        Self {
+            stream,
+            backing_file,
+            backing_memory: ret.cast(),
+            backing_memory_size,
+            uffds: HashMap::default(),
+        }
+    }
+
+    /// Runs a poll on the fd of the stream and uffds.
```
roypat, line 209: nit: 
```suggestion
    /// Polls the `UnixStream` and UFFD fds in a loop.
```

roypat (commented):
The implementation makes sense to me, I'm just wondering if we maybe want to have some sort of test for the "multiple handlers" functionality? 

If its very difficult to do anything here without the actual vhost implementation, then I'm fine to merge this as is, and just resolve any issues that might show up later when we actually start making use of this in vhost test (it's not production code), but if a test is simple I think we should add one.

ShadowCurse:
So the issue with testing is that uffd is inside the examples. And it seems tests from there can not be run. Adding test to `firecracker/tests` is also not possible as there is no way to import the module from examples. I think all of this can be fixed if we move all examples out of `src/firecracker` and into `tests` into separate crate.

roypat:
> So the issue with testing is that uffd is inside the examples. And it seems tests from there can not be run. Adding test to `firecracker/tests` is also not possible as there is no way to import the module from examples. I think all of this can be fixed if we move all examples out of `src/firecracker` and into `tests` into separate crate.

Mh, they used to be inside a separate test crate, but we moved them to examples to avoid having multiple `Cargo.lock`s across the repository (because these kept going stale because only the top level locks got thought about when doing PRs), and also because it simplified dependabot a bit (although that's less of a problem now that we have grouped updates). 

What do you mean by "tests cannot be run"? When I try

```diff
diff --git a/src/firecracker/examples/uffd/valid_handler.rs b/src/firecracker/examples/uffd/valid_handler.rs
index 28332d7b3..70a648232 100644
--- a/src/firecracker/examples/uffd/valid_handler.rs
+++ b/src/firecracker/examples/uffd/valid_handler.rs
@@ -65,3 +65,11 @@ fn main() {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn oh_no() {
+        panic!("the disco");
+    }
+}
```

and run

```
cargo test --examples
```

I get 

```
 Finished test [unoptimized + debuginfo] target(s) in 28.78s
     Running unittests examples/seccomp/harmless.rs (build/cargo_target/debug/examples/seccomp_harmless-57505c8eedc9c033)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests examples/seccomp/jailer.rs (build/cargo_target/debug/examples/seccomp_jailer-314de9b614652e42)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests examples/seccomp/malicious.rs (build/cargo_target/debug/examples/seccomp_malicious-852052f59dfb42b3)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests examples/seccomp/panic.rs (build/cargo_target/debug/examples/seccomp_panic-89198cef3c67514a)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests examples/uffd/malicious_handler.rs (build/cargo_target/debug/examples/uffd_malicious_handler-961c5e5e005a1dfb)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests examples/uffd/valid_handler.rs (build/cargo_target/debug/examples/uffd_valid_handler-df9be2ff715b0e49)

running 1 test
test tests::oh_no ... FAILED

failures:

---- tests::oh_no stdout ----
thread 'tests::oh_no' panicked at src/firecracker/examples/uffd/valid_handler.rs:73:9:
the disco


failures:
    tests::oh_no

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: test failed, to rerun pass `-p firecracker --example uffd_valid_handler`

```

ShadowCurse:
oh, I didn't try with `--examples`

src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -55,7 +55,7 @@ pub enum MemPageState {
 }
 
 impl UffdPfHandler {
-    pub fn from_unix_stream(stream: UnixStream, backing_buffer: *const u8, size: usize) -> Self {
+    pub fn from_unix_stream(stream: &UnixStream, backing_buffer: *const u8, size: usize) -> Self {
```
JonathanWoollett-Light, line 58:
```suggestion
    pub fn from_unix_stream<S: AsRef<UnixStream>>(stream: S, backing_buffer: *const u8, size: usize) -> Self {
```
this means the later change
```diff
-     UffdPfHandler::from_unix_stream(stream, memfile_buffer, size)
+    UffdPfHandler::from_unix_stream(&stream, memfile_buffer, size)
```
is not required.

src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -292,3 +292,92 @@ fn create_mem_regions(mappings: &Vec<GuestRegionUffdMapping>) -> Vec<MemRegion>
 
     mem_regions
 }
+
+#[cfg(test)]
```
roypat, line 296: I think you also need to do something like 

```diff
diff --git a/tests/integration_tests/build/test_unittests.py b/tests/integration_tests/build/test_unittests.py
index 2919fb90a..312235715 100644
--- a/tests/integration_tests/build/test_unittests.py
+++ b/tests/integration_tests/build/test_unittests.py
@@ -23,6 +23,7 @@ def test_unittests(test_fc_session_root_path):
 
     extra_args = f"--target {TARGET}"
     host.cargo_test(test_fc_session_root_path, extra_args=extra_args)
+    host.cargo_test(test_fc_session_root_path, extra_args=extra_args + "--examples")

```

to get the CI to run the tests

src/firecracker/examples/uffd/uffd_utils.rs
ShadowCurse: I would prefer to stick with `&UnixStream`. There is not point of adding additional abstractions.

src/firecracker/examples/uffd/uffd_utils.rs
ShadowCurse: added

JonathanWoollett-Light (commented):
Slight grammar fixes in the [feat(uffd): new Runtime](https://github.com/firecracker-microvm/firecracker/pull/4373/commits/78fbf901e3c19a8dc09d3be76d350824215a1cad) commit message
```diff
- New `Runtime` struct can contain multiple
- uffd objects received from the UDS socket during
- its runtime.
- 
- This is done by polling on both socket and uffd
- file descriptors. If socket event is polled, new
- uffd object is retrieved. If uffd fd is polled,
- page fault for corresponding uffd is handled.
- 
- The behaviour for page faults can be configured
- with a parameter passed to the `Runtime::run`.
+ The new `Runtime` struct can contain multiple
+ uffd objects received from the UDS socket during
+ its runtime.
+ 
+ This is done by polling the socket and uffd
+ file descriptors. When the socket event is 
+ polled, a new uffd object is retrieved. When
+ the uffd fd is polled, a page fault for 
+ corresponding uffd is handled.
+ 
+ The behaviour for page faults can be configured
+ with a parameter passed to the `Runtime::run`.
```

src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -162,6 +162,110 @@ impl UffdPfHandler {
...
+                backing_memory_size,
+                libc::PROT_READ,
+                libc::MAP_PRIVATE,
+                backing_file.as_raw_fd(),
+                0,
+            )
+        };
+        if ret == libc::MAP_FAILED {
+            panic!("mmap on backing file failed");
+        }
```
JonathanWoollett-Light, line 198: Nit: A definite nit, feel free to ignore.

We can add some additional safety here by using https://docs.rs/nix/latest/nix/sys/mman/fn.mmap.html (nix is already in our dependency tree).

```suggestion
        let ret = unsafe { mmap(
            None,
            NonZeroUsize::new(backing_memory_size)?,
            ProtFlags::PROT_READ,
            MapFlags::MAP_PRIVATE,
            Some(backing_file),
            0
        )? };
```

roypat (dismissed):

src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -162,6 +162,110 @@ impl UffdPfHandler {
...
+pub struct Runtime {
+    stream: UnixStream,
+    backing_file: File,
+    backing_memory: *mut u8,
+    backing_memory_size: usize,
+    uffds: HashMap<i32, UffdPfHandler>,
+}
+
+impl Runtime {
+    pub fn new(stream: UnixStream, backing_file: File) -> Self {
```
JonathanWoollett-Light, line 179: NIt: It would be preferable to propagate the error but given this is an example its okay either way.
```suggestion
    pub fn new(stream: UnixStream, backing_file: File) -> Result<Self, NewRuntimeError> {
```

src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -162,6 +162,110 @@ impl UffdPfHandler {
...
+        // We can skip polling on stream fd if
+        // the connection is closed.
+        let mut skip_stream: usize = 0;
+        loop {
+            let pollfd_ptr = pollfds[skip_stream..].as_mut_ptr();
+            let pollfd_size = pollfds[skip_stream..].len() as u64;
+
+            // # Safety:
+            // Pollfds vector is valid
+            let mut nready = unsafe { libc::poll(pollfd_ptr, pollfd_size, -1) };
```
JonathanWoollett-Light, line 233: Nit: It is always preferable to use `epoll` over `poll`.

From https://man7.org/linux/man-pages/man7/epoll.7.html

> By contrast, when used as a level-triggered interface (the
> default, when EPOLLET is not specified), epoll is simply a faster
> [poll(2)](https://man7.org/linux/man-pages/man2/poll.2.html), and can be used wherever the latter is used since it
> shares the same semantics.

There is also a nice nix interface for epoll (I'm certainly not biased since I wrote it 😊) https://docs.rs/nix/latest/nix/sys/epoll/struct.Epoll.html

src/firecracker/examples/uffd/valid_handler.rs
```diff
@@ -7,42 +7,29 @@
...
+use std::os::unix::net::UnixListener;
 
-use uffd_utils::{create_pf_handler, MemPageState};
+use uffd_utils::{MemPageState, Runtime, UffdPfHandler};
 use utils::get_page_size;
 
 fn main() {
-    let mut uffd_handler = create_pf_handler();
+    let uffd_sock_path = std::env::args().nth(1).expect("No socket path given");
+    let mem_file_path = std::env::args().nth(2).expect("No memory file given");
```
JonathanWoollett-Light, line 18: Nit
```suggestion
    let mut args = std::env::args();
    let uffd_sock_path = args.next().expect("No socket path given");
    let mem_file_path = args.next().expect("No memory file given");
```

src/firecracker/examples/uffd/valid_handler.rs
```diff
@@ -53,15 +40,13 @@ fn main() {
...
         match event {
-            userfaultfd::Event::Pagefault { addr, .. } => {
-                uffd_handler.serve_pf(addr as *mut u8, len)
-            }
+            userfaultfd::Event::Pagefault { addr, .. } => uffd_handler.serve_pf(addr.cast(), len),
             userfaultfd::Event::Remove { start, end } => uffd_handler.update_mem_state_mappings(
-                start as *mut u8 as u64,
-                end as *mut u8 as u64,
+                start.cast::<u8>() as u64,
+                end.cast::<u8>() as u64,
```
JonathanWoollett-Light, line 45: Nit: I don't think you need to cast the pointer type.
```
assert_eq!(x.cast::<A>() as u64, x.cast::<B>() as u64);
```
will always be true regardless of the types of `A` or `B`.

(unless for some reason the compiler doesn't allow `as u64` for certain types)

src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -160,6 +134,36 @@ impl UffdPfHandler {
...
+            self.uffd
+                .copy(src as *const _, dst as *mut _, len, true)
+                .expect("Uffd copy failed")
+        };
+
+        // Make sure the UFFD copied some bytes.
+        assert!(ret > 0);
+
+        (dst, dst + len as u64)
+    }
```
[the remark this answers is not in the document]
ShadowCurse: `offset` requires `unsafe` block and casing `region.mapping.offset` and `offset` to `isize`. So in total it will be not much better than it is now.

src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -292,3 +292,92 @@ fn create_mem_regions(mappings: &Vec<GuestRegionUffdMapping>) -> Vec<MemRegion>
...
+            size: 0,
+            offset: 0,
+        }];
+        let error_memory_region_json = serde_json::to_string(&error_memory_region).unwrap();
+        stream
+            .send_with_fd(error_memory_region_json.as_bytes(), dummy_fd_2)
+            .unwrap();
+
+        runtime_thread.join().unwrap_err();
+    }
```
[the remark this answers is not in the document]
ShadowCurse: The issue is that test will stop at `let (stream, _) = listener.accept().expect("Cannot listen on UDS socket");` line which will wait forever, because the connection is only established later:  `let stream =
            UnixStream::connect(dummy_socket_path_clone).expect("Cannot connect to the socket");`

src/firecracker/examples/uffd/uffd_utils.rs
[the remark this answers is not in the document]
JonathanWoollett-Light: It is equally unsafe but communicates this explicitly, I think this is the better approach.

src/firecracker/examples/uffd/uffd_utils.rs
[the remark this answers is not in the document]
JonathanWoollett-Light:
```rust
#[test]
fn test_runtime() {
    let tmp_dir = TempDir::new().unwrap();
    let dummy_socket_path = tmp_dir.as_path().join("dummy_socket");
    let dummy_socket_path_clone = dummy_socket_path.clone();

    let pair = Arc::new(Mutex::new(OnceLock::new()), Condvar::new());
    let pair2 = pair.clone();

    // Open file for socket listener
    let tmp_file = TempFile::new().unwrap();
    tmp_file.as_file().set_len(0x1000).unwrap();
    let dummy_mem_path = tmp_file.as_path();
    let file = File::open(dummy_mem_path).unwrap();
    let listener = UnixListener::bind(dummy_socket_path).unwrap();
    
    // Start runtime thread
    let runtime_thread = std::thread::spawn(move || {
        let (lock, cvar) = &*pair2;
        // Accept connection
        let (stream, _) = listener.accept().unwrap();
        // Initalize runtime
        lock.lock().unwrap().set(Runtime::new(stream, file)).unwrap();
        // Notify test thread
        cvar.notify_one();
        // Run runtime
        lock.lock().unwrap().run(|_: &mut UffdHandler| {});
    });
    let (lock, cvar) = &*pair;

    // COnnect to socket
    let stream = UnixStream::connect(dummy_socket_path_clone).unwrap();

    let dummy_memory_region = vec![GuestRegionUffdMapping {
        base_host_virt_addr: 0,
        size: 0x1000,
        offset: 0,
    }];
    let dummy_memory_region_json = serde_json::to_string(&dummy_memory_region).unwrap();

    let dummy_file_1 = TempFile::new().unwrap();
    let dummy_fd_1 = dummy_file_1.as_file().as_raw_fd();
    stream
        .send_with_fd(dummy_memory_region_json.as_bytes(), dummy_fd_1)
        .unwrap();

    // Wait for runtime to be initialized.
    let _guard = cvar.wait_while(lock.lock().unwrap(), |pending| { pending.get().is_none() }).unwrap():

    // wait for the runtime thread to process message
    std::thread::sleep(std::time::Duration::from_millis(100));

    assert_eq!(lock.lock().unwrap().get().unwrap().uffds.len(), 1);

    let dummy_file_2 = TempFile::new().unwrap();
    let dummy_fd_2 = dummy_file_2.as_file().as_raw_fd();
    stream
        .send_with_fd(dummy_memory_region_json.as_bytes(), dummy_fd_2)
        .unwrap();
    // wait for the runtime thread to process message
    std::thread::sleep(std::time::Duration::from_millis(100));
    assert_eq!(lock.lock().unwrap().get().unwrap().uffds.len(), 2);

    // there is no way to properly stop runtime, so
    // we send a message with an incorrect memory region
    // to cause runtime thread to panic
    let error_memory_region = vec![GuestRegionUffdMapping {
        base_host_virt_addr: 0,
        size: 0,
        offset: 0,
    }];
    let error_memory_region_json = serde_json::to_string(&error_memory_region).unwrap();
    stream
        .send_with_fd(error_memory_region_json.as_bytes(), dummy_fd_2)
        .unwrap();

    runtime_thread.join().unwrap_err();
}
```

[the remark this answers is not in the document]
ShadowCurse: I think you can open a PR with this change later, after your changes compile and run.

src/firecracker/examples/uffd/valid_handler.rs
ShadowCurse: done

src/firecracker/examples/uffd/valid_handler.rs
ShadowCurse: done

src/firecracker/examples/uffd/uffd_utils.rs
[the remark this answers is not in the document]
JonathanWoollett-Light: Resolving this appears much more complex than I initially thought, leaving the test with an unlikely race condition is okay for now.

roypat (approved):

JonathanWoollett-Light (approved):

PR #4376, opened 2024-01-19
test: Add unit test for installing resource limits

georgesitoniv:
This change adds unit test for verifying that the resource limits for file size and the number of file descriptors has been set properly.

## Changes

Minor change to add a unit test for installing the ResourceLimits for file size and the number of file descriptors.

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- [N/A] If a specific issue led to this PR, this PR closes the issue.
- [N/A] Any required documentation changes (code and docs) are included in this PR.
- [N/A] New `TODO`s link to an issue.

---

### DIFF
--- a/src/jailer/src/resource_limits.rs
+++ b/src/jailer/src/resource_limits.rs
@@ -187,4 +187,42 @@ mod tests {
         assert_eq!(rlim.rlim_cur, new_limit);
         assert_eq!(rlim.rlim_max, new_limit);
     }
+
+    #[test]
+    fn test_install() {
+        // Setup the resource limits
+        let mut rlimits = ResourceLimits::default();
+        let new_limit = 100;
+        rlimits.set_file_size(new_limit);
+        rlimits.set_no_file(new_limit);
+        
+        // Install the new limits to file size and 
+        // the number of file descriptors
+        let result = rlimits.install();
+        assert_eq!(result.unwrap(), ());
+    
+        // Verify the new limit for file size
+        let file_size_resource = Resource::RlimitFsize;
+        let mut file_size_limit: libc::rlimit = libc::rlimit {
+            rlim_cur: 0,
+            rlim_max: 0,
+        };
+        unsafe { 
+            libc::getrlimit(file_size_resource.into(), &mut file_size_limit) 
+        };
+        assert_eq!(file_size_limit.rlim_cur, new_limit);
+        assert_eq!(file_size_limit.rlim_max, new_limit);
+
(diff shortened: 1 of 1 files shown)


src/jailer/src/resource_limits.rs
```diff
@@ -187,4 +187,42 @@ mod tests {
...
+        // Setup the resource limits
+        let mut rlimits = ResourceLimits::default();
+        let new_limit = 100;
+        rlimits.set_file_size(new_limit);
+        rlimits.set_no_file(new_limit);
+        
+        // Install the new limits to file size and 
+        // the number of file descriptors
+        let result = rlimits.install();
+        assert_eq!(result.unwrap(), ());
```
roypat, line 202: nit: This assertion cannot fail, because the unit type `()` only allows a single value, `()`. Therefore, if we have a value of type unit, we know it must equal `()`. The important part here is the `.unwrap()`, which checks that we indeed did not get an error during execution:

```suggestion
        rlimits.install().unwrap();
```

georgesitoniv: ah understood thank you Patrick. I assume that unwrap will also throw an error in failure cases?

roypat: Yes, `.unwrap()` will `panic!` if the returned result contains an `Err` variant

georgesitoniv: Thanks Patrick! pushed a change to fix this and fixed the formatting too.

src/jailer/src/resource_limits.rs
```diff
@@ -187,4 +187,37 @@ mod tests {
         assert_eq!(rlim.rlim_cur, new_limit);
         assert_eq!(rlim.rlim_max, new_limit);
     }
+
+    #[test]
+    fn test_install() {
+        // Setup the resource limits
+        let mut rlimits = ResourceLimits::default();
+        let new_limit = 100;
```
roypat, line 195: You might have to choose a bigger value here, the tests fail with `SIGXFSZ`, which seems to be caused by some file being larger than the allowed limit: https://en.wikipedia.org/wiki/Signal_(IPC)#SIGXFSZ

georgesitoniv: ah I see. do you know the largest file size used by those tests? also I can I run those locally?

roypat (approved):

sudanl0 (approved):

PR #4386, opened 2024-01-19
fix(net): Apply only supported TAP offloading features

git-Zink:
## Changes
Checking what offload features are supported by virtio driver and apply only them to TAP device
...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- [x ] If a specific issue led to this PR, this PR closes the issue.
- [x ] The description of changes is clear and encompassing.
- [ x] Any required documentation changes (code and docs) are included in this PR.
- [ x] All added/changed functionality is tested.

---

### DIFF
--- a/src/vmm/src/devices/virtio/mod.rs
+++ b/src/vmm/src/devices/virtio/mod.rs
@@ -10,6 +10,8 @@
 use std::any::Any;
 use std::io::Error as IOError;
 
+use crate::devices::virtio::net::TapError;
+
 pub mod balloon;
 pub mod block_common;
 pub mod device;
@@ -68,6 +70,8 @@ pub enum ActivateError {
     BadActivate,
     /// Vhost user: {0}
     VhostUser(vhost_user::VhostUserError),
+    /// Setting tap interface offload flags failed: {0}
+    TapSetOffload(TapError),
 }
 
 /// Trait that helps in upcasting an object to Any
--- a/src/vmm/src/devices/virtio/net/device.rs
+++ b/src/vmm/src/devices/virtio/net/device.rs
@@ -23,8 +23,10 @@ use vm_memory::GuestMemoryError;
 use crate::devices::virtio::device::{DeviceState, IrqTrigger, IrqType, VirtioDevice};
 use crate::devices::virtio::gen::virtio_blk::VIRTIO_F_VERSION_1;
 use crate::devices::virtio::gen::virtio_net::{
-    virtio_net_hdr_v1, VIRTIO_NET_F_CSUM, VIRTIO_NET_F_GUEST_CSUM, VIRTIO_NET_F_GUEST_TSO4,
-    VIRTIO_NET_F_GUEST_UFO, VIRTIO_NET_F_HOST_TSO4, VIRTIO_NET_F_HOST_UFO, VIRTIO_NET_F_MAC,
+    virtio_net_hdr_v1, VIRTIO_NET_F_CSUM, VIRTIO_NET_F_GUEST_CSUM, VIRTIO_NET_F_GUEST_ECN,
+    VIRTIO_NET_F_GUEST_TSO4, VIRTIO_NET_F_GUEST_TSO6, VIRTIO_NET_F_GUEST_UFO,
+    VIRTIO_NET_F_HOST_ECN, VIRTIO_NET_F_HOST_TSO4, VIRTIO_NET_F_HOST_TSO6, VIRTIO_NET_F_HOST_UFO,
+    VIRTIO_NET_F_MAC,
 };
 use crate::devices::virtio::gen::virtio_ring::VIRTIO_RING_F_EVENT_IDX;
 use crate::devices::virtio::iovec::IoVecBuffer;
@@ -160,7 +162,11 @@ impl Net {
             | 1 << VIRTIO_NET_F_HOST_TSO4
             | 1 << VIRTIO_NET_F_HOST_UFO
             | 1 << VIRTIO_F_VERSION_1
-            | 1 << VIRTIO_RING_F_EVENT_IDX;
+            | 1 << VIRTIO_RING_F_EVENT_IDX
+            | 1 << VIRTIO_NET_F_GUEST_TSO6
+            | 1 << VIRTIO_NET_F_HOST_TSO6
+            | 1 << VIRTIO_NET_F_GUEST_ECN
+            | 1 << VIRTIO_NET_F_HOST_ECN;
 
         let mut config_space = ConfigSpace::default();
         if let Some(mac) = guest_mac {
@@ -209,10 +215,6 @@ impl Net {
         tx_rate_limiter: RateLimiter,
     ) -> Result<Self, NetError> {
         let tap = Tap::open_named(tap_if_name).map_err(NetError::TapOpen)?;
-
-        // Set offload flags to match the virtio features below.
-        tap.set_offload(gen::TUN_F_CSUM | gen::TUN_F_UFO | gen::TUN_F_TSO4 | gen::TUN_F_TSO6)
-            .map_err(NetError::TapSetOffload)?;
 
         let vnet_hdr_size = i32::try_from(vnet_hdr_len()).unwrap();
         tap.set_vnet_hdr_size(vnet_hdr_size)
@@ -644,6 +646,44 @@ impl Net {
         }
     }
 
+    fn build_tap_supported_features(virtio_supported_features: u64) -> u32 {
+        let add_if_supported =
+            |tap_features: &mut u32, virtio_features: u64, tap_flag: u32, virtio_flag: u32| {
+                if virtio_features & (1 << virtio_flag) != 0 {
+                    *tap_features |= tap_flag;
+                }
+            };
+
+        let mut tap_features: u32 = 0;
+
+        add_if_supported(
+            &mut tap_features,
+            virtio_supported_features,
+            gen::TUN_F_CSUM,
+            VIRTIO_NET_F_CSUM,
+        );
+        add_if_supported(
+            &mut tap_features,
+            virtio_supported_features,
+            gen::TUN_F_UFO,
+            VIRTIO_NET_F_GUEST_UFO,
+        );
+        add_if_supported(
+            &mut tap_features,
+            virtio_supported_features,
+            gen::TUN_F_TSO4,
+            VIRTIO_NET_F_GUEST_TSO4,
+        );
+        add_if_supported(
+            &mut tap_features,
+            virtio_supported_features,
+            gen::TUN_F_TSO6,
+            VIRTIO_NET_F_GUEST_TSO6,
+        );
+
+        tap_features
+    }
+
     /// Updates the parameters for the rate limiters
     pub fn patch_rate_limiters(
         &mut self,
@@ -859,6 +899,11 @@ impl VirtioDevice for Net {
             }
         }
 
+        let supported_flags: u32 = Net::build_tap_supported_features(self.acked_features);
+        self.tap
+            .set_offload(supported_flags)
+            .map_err(super::super::ActivateError::TapSetOffload)?;
+
         if self.activate_evt.write(1).is_err() {
             error!("Net: Cannot write to activate_evt");
             return Err(super::super::ActivateError::BadActivate);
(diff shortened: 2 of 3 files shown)


src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -644,6 +641,18 @@ impl Net {
...
     }
 
+    fn build_tap_supported_features(desired_flags: Vec<(u32, u32)>, supported_features: u64) -> u32 {
+        let mut answer : u32 = 0;
+
+        for i in 0..desired_flags.len() {
+            if (supported_features & 1 << desired_flags[i].1) != 0 {
+                answer |= desired_flags[i].0
+            }
+        }
```
ShadowCurse, line 651: This can be:
```rust
fn build_tap_supported_features(desired_flags: &[(u32, u32)], supported_features: u64) -> u32 {
     let mut answer: u32 = 0;

   for (tap_flag, net_flag) in desired_flags.iter() {
       if (supported_features & 1 << net_flag) != 0 {
           answer |= tap_flag;
       }
   }

   answer
}
```
Basically if you pass `Vec<(u32, u32)>` then the function takes ownership of the `desired_flags`. This means if you call this function multiple times you will need to create new instance if `desired_flags` each time.
To convert from `Vec<(u32, u32)>` to `&[(u32, u32)]` you can simply use `&` before the vector on the calling side.

Also you can simply iterate over a vector of tuples, instead of accessing them through indexes.

git-Zink: thx. done

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -996,6 +1020,74 @@ pub mod tests {
...
+        let mut tap_features: u32 = 0;
+        let mut ack_features: u32 = 0;
+
+        for i in 0..desired_features.len() {
+            tap_features |= desired_features[i].0;
+            ack_features |= 1 << desired_features[i].1;
+        }
+
+        // extra supported feature
+        ack_features |= 1 << VIRTIO_NET_F_GUEST_TSO6;
```
bchalios, line 1083: Good that you added this version but I think this cannot happen. The MMIO layer will make sure that we will never see more flags than asked for in the acknowledged features.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -859,6 +869,19 @@ impl VirtioDevice for Net {
             }
         }
 
+        let desired_flags = [
+            (gen::TUN_F_CSUM, VIRTIO_NET_F_CSUM),
+            (gen::TUN_F_UFO, VIRTIO_NET_F_GUEST_UFO),
+            (gen::TUN_F_TSO4, VIRTIO_NET_F_GUEST_TSO4),
+            (gen::TUN_F_TSO6, VIRTIO_NET_F_GUEST_TSO6),
+        ];
```
ShadowCurse, line 877: I guess because this 'map' is used here and in all tests, let's make it a `const` at the top of the file.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -160,7 +161,8 @@ impl Net {
             | 1 << VIRTIO_NET_F_HOST_TSO4
             | 1 << VIRTIO_NET_F_HOST_UFO
             | 1 << VIRTIO_F_VERSION_1
-            | 1 << VIRTIO_RING_F_EVENT_IDX;
+            | 1 << VIRTIO_RING_F_EVENT_IDX
+            | 1 << VIRTIO_NET_F_GUEST_TSO6;
```
bchalios, line 165: nit: since we're adding IPv6 flags we might as well add `VIRTIO_NET_F_HOST_TSO6` and the `ECN` pair

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -996,6 +1020,74 @@ pub mod tests {
...
+            (gen::TUN_F_TSO6, VIRTIO_NET_F_GUEST_TSO6),
+        ];
+
+        let mut tap_features: u32 = 0;
+        let mut ack_features: u32 = 0;
+
+        for i in 0..desired_features.len() {
+            tap_features |= desired_features[i].0;
+            ack_features |= 1 << desired_features[i].1;
+        }
```
ShadowCurse, line 1040: And here you also can use `for (tap_feature, virtio_feature) in desired_features.iter()`

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -644,6 +642,18 @@ impl Net {
...
+
+        for (tap_flag, net_flag) in desired_flags.iter() {
+            if (supported_features & 1 << net_flag) != 0 {
+                answer |= tap_flag;
+            }
+        }
+
+        return answer;
+    }
+
```
bchalios, line 656: personally, I would keep this much simpler. Something like:

```Rust
fn build_tap_supported_features(&self, supported_features: u64) -> u32 {
    let mut tap_features: u32 = 0;

    if self.has_features(VIRTIO_NET_F_GUEST_TSO4) {
        tap_features |= gen::TUN_F_TSO4;
    }
    if self.has_feature(....) {

    }
}
```

and then later during activate, you don't have to build the array, you just:

```Rust
        let supported_flags =
            Net::build_tap_supported_features(self.acked_features());
```

src/vmm/src/devices/virtio/net/device.rs
git-Zink: but it's different in some tests, isn't it?

src/vmm/src/devices/virtio/net/device.rs
git-Zink: Tbh, I'm not sure it's better in term of testing. In this case we won't be able to write unit tests with extra / less flags we want to set up.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: In that test you can take a slice (because it only lacks last element): `let desired_flags = FLAGS_MAP[..FLAGS_MAP.len() - 1]` where `FLAGS_MAP` is the name of the constant.

src/vmm/src/devices/virtio/net/device.rs
bchalios: I think it would. The only purpose the `desired_flags` parameter serves is to "build" the correlation between VirtIO and TAP flags. You can do exactly the same test you are doing in this PR, without the need to pass around the `desired_flags` array. This is simpler, since you only need to encode knowledge for the TAP features in this function here. For example:

```rust
    #[test]
    fn test_build_tap_supported_features_ack_les_than_want() {
        let tap_features: u32 = gen::TUN_F_CSUM | gen::TUN_F_UFO;
        let ack_features: u32 = 1 << VIRTIO_NET_F_CSUM | 1 << VIRTIO_NET_F_GUEST_UFO;

        let supported_flags =
            Net::build_tap_supported_features(ack_features as u64);

        assert_eq!(supported_flags, tap_features);
    }
```

ShadowCurse: The issue with `if self.has_features(VIRTIO_NET_F_GUEST_TSO4) {` approach is that the method does not have `self` parameter. I think approach with map/array as a constant will be ok. Also I don't think we really need to test what happens if this `desired_flags` has incorrect/incomplete  entries.

bchalios: > The issue with if self.has_features(VIRTIO_NET_F_GUEST_TSO4) { approach is that the method does not have self parameter

Let's just add it then :)

ShadowCurse: Then in the tests we will need to create `net` devices. And all of this just to not write `if (supported_features & 1 << net_flag) != 0`.

bchalios: alternatively, it doesn't have to make use of the `has_feature` it can just be:

```rust
if (flags & (1 << VIRTIO_NET_F_CSUM)) {
    tap_flags |= gen::TUN_F_CSUM;
}
```

bchalios: Just to clarify, I'm just making this suggestion because I'd like to avoid the for loop and keeping track of this constant array. I think having an array is an overkill and looping might hurt performance

ShadowCurse: It seems we cycled back to

Testing each flag in separate `if`:
```rust
if (flags & (1 << VIRTIO_NET_F_CSUM)) {
    tap_flags |= gen::TUN_F_CSUM;
}
...
```
vs

Testing all in one loop:
```rust
const FEATURE_MAP = [
  (VIRTIO_NET_F_CSUM, TUN_F_CSUM),
  ...
  ]

...
for (tap_flag, net_flag) in desired_flags.iter() {
        if (supported_features & 1 << net_flag) != 0 {
            answer |= tap_flag;
        }
}
```

I am fine with any solution here.

bchalios: Yup, I'd rather go for the first approach.

src/vmm/src/devices/virtio/net/device.rs
git-Zink: I my be wrong, but it sounds to me like we rely on 3rd party thing to prevent such behaviour. Since we add flags here and now it looks reasonable to me to test it in the same place.

But I have no idea about FC and maybe virtio device and MMIO layer are inseparable things indeed.

bchalios: MMIO layer is implemented by Firecracker. Here: https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/devices/virtio/device.rs#L155 is where we check that we will not accept acked features that we did not offer.

src/vmm/src/devices/virtio/net/device.rs
git-Zink: It seems out of scope of this CR, but if you are fine with that I will add in the next rev.

src/vmm/src/devices/virtio/net/device.rs
git-Zink: Oh, sorry, I didn't mean it's 3rd party lib. I was on about relying on logic in a different place. And if it's changed once we may loose this coverage. Or we should leave a comment that we need this check there, cos we expect it in the different place.

It's more about global pattern, rather than this case in particular. If you are ok with that - I can delete this test.

src/vmm/src/devices/virtio/net/device.rs
git-Zink: Oh, well. Here are my 50 cents:

1. If we don't care about testing it at all we can inline `build_tap_supported_features()` into `activate()` function itself. With or without the array.
2. In general I'd suggest to have a  `convert_tap_to_vitio_future(tap_feat: u32) -> u32` function, which will own converting logic. Cos as it was fairly mentioned - we use the same kind of arrays in different places. I'm not sure where should it be though (in `if_tun.rs` or `virtio_net.rs`.
3. It's nit, but I'd like to see function `convert_vitio_feat_to_bitflag()`, which makes `1 << flag`, cos it's a little bit mess to have it everywhere around the code. Moreover, tap features' consts are bitflags and virtio flags are not. (which can be fixed instead of having the function). But I understand it may affect a lot of code and may be undesirable. 

In general I would vote for array implementation, cos:
1. For now we have few features. But if it was 30? 50? after what number you would convert it into an array?
2. It would be nice to have all tap feature we'd like to setup in one place for readability (maybe as a static field of the class or as const array in the top of the file). To not seek it in the code, when we need to change flags we try to set up.

ShadowCurse: I would prefer to keep complexity and amount of indirection to minimum. The whole conversion is a simple bit checking and we don't need to create a huge machinery to do that.
The amount of features it not really a concern here. It does not really matter if it will be 50 lines of array definition with all mappings or 50 lines of bit checks. Both will be +- same size and complexity. In the end it is to to the tests and person doing review to check that all mapping as correct.

I think smth simple as this should be enough. No additional arrays, statics... just one pure function.
```rust
fn virtio_to_tap_features(virtio_features: u64) -> u32 {
    let check_set = |virtio_features, virtio_feature, &mut tap_features, tap_feature| {
        if (virtio_features & (1 << virtio_feature)) {
            tap_features |= tap_feature;
        }
    };

    let mut tap_features = 0;

    check_set(virtio_features, VIRTIO_NET_F_CSUM, &mut tap_features, TUN_F_CSUM);
    check_set(virtio_features, VIRTIO_NET_F_GUEST_UFO, &mut tap_features, TUN_F_UFO);
    check_set(virtio_features, VIRTIO_NET_F_GUEST_TSO4, &mut tap_features, TUN_F_TSO4);
    check_set(virtio_features, VIRTIO_NET_F_GUEST_TSO6, &mut tap_features, TUN_F_TSO6);

    tap_features
}
```

git-Zink: as you wish

src/vmm/src/devices/virtio/mod.rs
```diff
@@ -68,6 +70,8 @@ pub enum ActivateError {
     BadActivate,
     /// Vhost user: {0}
     VhostUser(vhost_user::VhostUserError),
+    /// Setting tap interface offload flags failed
+    TapSetOffload(TapError),
```
ShadowCurse, line 74: Use `{0}` (or `{0:?}` if `TapError` does not implement `Display` trait) at the end of the comment. We used special magic to make comments act as Display implementations for errors.

bchalios:
Closing this one. I cherry-picked Nikita's commit in https://github.com/firecracker-microvm/firecracker/pull/4680.

PR #4389, opened 2024-01-22
Support reset for `net` device

### LINKED ISSUE #3074: [v.1.17.0] support for resetting virtio devices
# Feature Request

During the FreeBSD boot process, it resets network devices.  (Why?  Good question.)  When running in Firecrackers, this means resetting the vtnet device.

When a status of `0` is written to the virtio device, it should reset.

It looks like the MMIO code in Firecracker gets this right, but it calls a `_reset` routine for the device and none of the devices have implemented it.  As a result, the device is currently entering `FAILED` state instead.

FreeBSD's virtio driver doesn't actually *notice* that the device is in FAILED state, and Firecracker's devices continue to work despite being marked as FAILED -- so everything ends up working except for Firecracker logging warnings of `ack virtio features in invalid state 0x8f` and `update virtio queue in invalid state 0x8f`.  But this (arguably buggy) behaviour of ignoring that the device has failed could change in the future, so it would be good to avoid having it marked as failed in the first place.

## Checks

acj:
## Changes

This PR implements support for resetting the `net` device. Resolves #3074.

## Reason

The VirtIO spec says in [section 3.1.1](https://docs.oasis-open.org/virtio/virtio/v1.1/cs01/virtio-v1.1-cs01.html#x1-930001):

> The driver MUST follow this sequence to initialize a device: 
> 1. Reset the device. 

Operating systems like FreeBSD and NetBSD reset the device as part of initializing it, which currently causes Firecracker to put the device into a `FAILED` state. As noted in #3074, some guest OSes continue to have functional networking because they're unaware that the device is in a failed state, but we shouldn't rely on this behavior.

## Notes

- This is partly an RFC because I'm not familiar enough with the vmm code to be confident that the device is being reset properly. Still, this does work in my Linux and FreeBSD test cases. Linux doesn't seem to reset the net device. FreeBSD does reset it, and then the device is activated and becomes usable.
- `reset()` returns a copy of the interrupt FD and queue event FDs, but the MMIO driver does not use them. It's unclear to me how they should be used. Is that handshake only needed for specific drivers? If there were a way to signal that reset had succeeded without needing to clone the FDs, then I could remove the seccomp changes. Perhaps the return type could be `Result<Option<...>>` instead.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


acj:
> Add two reviewers to your pull request (a maintainer will do that for you if you're new).

@kalyazin @sudanl0 👋

acj:
Rebased. At least some of the build failures were from the branch being stale, I think. The rest ("failed to run custom build command") look like they might be related to seccomp, but there's not much output.

roypat:
Hi @acj,
thank you for your contribution, and sorry for taking a while to respond.

> Rebased. At least some of the build failures were from the branch being stale, I think. The rest ("failed to run custom build command") look like they might be related to seccomp, but there's not much output.

I think this is because there is no `readlink` syscall on ARM, instead the seccomp entry needs to be `readlinkat`. 

However, my team talked about this internally and I think we can avoid adding new syscalls. The callsite of `reset` does not make use of the returned file descriptors (and I'm also not sure what it _could_ use them for), so I think we can simply change the signature of `reset` to return nothing, saving us the need to dup the fds (and thus the syscalls). cc @bchalios because he has more context on this.

acj:
@roypat Hi! No worries, and thanks for the feedback. Changing the reset signature sounds good.

I went with `Result` as the return type so that the existing reset behavior doesn't change too much, i.e. `set_device_status` will still mark the device as failed if reset isn't implemented.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -870,6 +870,15 @@ impl VirtioDevice for Net {
     fn is_activated(&self) -> bool {
         self.device_state.is_activated()
     }
+
+    fn reset(&mut self) -> Result<(), ResetError> {
+        self.device_state = DeviceState::Inactive;
+        self.rx_bytes_read = 0;
+        self.rx_deferred_frame = false;
+        self.rx_frame_buf = [0u8; MAX_BUFFER_SIZE];
+        self.metrics = NetMetricsPerDevice::alloc(self.id.clone());
```
bchalios, line 879: We definitely need to reset the `acked_features` here, too.

Also, @sudanl0 do you think it is ok to reset the metrics here? I think it's more of a philosophical question, on whether we want our metrics to reflect the fact that the device has been reset by zeroing them out. Maybe, it would make sense to add a reset counter metric as well.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -2015,17 +2024,29 @@ pub mod tests {
...
+            // Test interrupts.
+            assert!(!&net.irq_trigger.has_pending_irq(IrqType::Vring));
+        };
+
+        validate(&net);
 
-        // Test corresponding queues events.
-        assert_eq!(net.queue_events().len(), NET_QUEUE_SIZES.len());
+        // Test reset.
+        let mut net = net;
```
bchalios, line 2044: I guess you can just mark as `mut` the `net` device at line 2025?

src/vmm/src/devices/virtio/device.rs
```diff
@@ -174,10 +174,9 @@ pub trait VirtioDevice: AsAny + Send {
     /// Checks if the resources of this device are activated.
     fn is_activated(&self) -> bool;
 
-    /// Optionally deactivates this device and returns ownership of the guest memory map, interrupt
-    /// event, and queue events.
-    fn reset(&mut self) -> Option<(EventFd, Vec<EventFd>)> {
-        None
+    /// Optionally deactivates this device.
+    fn reset(&mut self) -> Result<(), ResetError> {
+        Err(ResetError::NotImplemented)
```
bchalios, line 179: Yes, that's what I had in mind honestly. I _assume_ the initial idea of giving back the `EventFd`s was to "ensure" through the type system that they can't be used, but that is a very dubious choice, IMHO.

bchalios (commented):
This looks a lot like what I had in mind. One thing that I'd like to understand better is what we should do with the TAP device. I've considered two main alternatives:

1. Leave everything as it is. It _should_ _probably_ work.
2. Close the open end we are holding during reset and re-create it once we get activated. In fact, we could always defer the opening of the TAP device during activation. This would express better our intentions I think, but there are three problems that I see:
    * We will need to change the TAP type we are holding to be an `Option`, which is not ideal
    * We would need to handle the (de)registration of the TAP device with the event loop which might not be trivial
    * Opening the TAP device now, adds an extra system call to the boot time which is in the hot path.

One final thing that bothers me is testing this through integration tests. We can't use FreeBSD to do this. It is not feasible for us to maintain FreeBSD artifacts for testing, so it would need to be done using Linux drivers. However, I haven't found a way to trigger a reset through Linux.
    *

acj:
@bchalios Thanks, that's very helpful. I'll take a closer look at the TAP handling.

> It is not feasible for us to maintain FreeBSD artifacts for testing, so it would need to be done using Linux drivers. However, I haven't found a way to trigger a reset through Linux.

I think this might be an issue on the Firecracker side. My understanding of Linux's virtio implementation is crude at best, but it looks like it should reset the device [during registration](https://github.com/torvalds/linux/blob/6764c317b6bb91bd806ef79adf6d9c0e428b191e/drivers/virtio/virtio.c#L459-L461). I added logging code and verified that the Linux driver sets status to 0 during boot, prior to activation. But Firecracker ignores this because it only calls `reset` [if the device is already activated](https://github.com/firecracker-microvm/firecracker/blob/b56d201186a7928e0e303c66fc475a89520d6d97/src/vmm/src/devices/virtio/mmio.rs#L200). Do you think it would be okay to handle a reset even if the device is inactive?

bchalios:
Hey @acj. Really sorry for silence on this. We got context switched in other topics.

> I think this might be an issue on the Firecracker side. My understanding of Linux's virtio implementation is crude at best, but it looks like it should reset the device [during registration](https://github.com/torvalds/linux/blob/6764c317b6bb91bd806ef79adf6d9c0e428b191e/drivers/virtio/virtio.c#L459-L461). I added logging code and verified that the Linux driver sets status to 0 during boot, prior to activation. But Firecracker ignores this because it only calls reset [if the device is already activated](https://github.com/firecracker-microvm/firecracker/blob/b56d201186a7928e0e303c66fc475a89520d6d97/src/vmm/src/devices/virtio/mmio.rs#L200). Do you think it would be okay to handle a reset even if the device is inactive?

It is. We don't support reset :) I had seen this reset there when I looked at resetting some time ago, but I'm not sure it is enough for testing that reset works (I might be convinced otherwise :)). Ideally, what we _could_ do is load/unload the virtio-net module in the guest. But we explicitly don't want to use modules in the guest for safety reasons.

bchalios:
Hi @acj, just wanted to touch base in order to see if you're still working on this or have plans to do in the near future.

acj:
@bchalios :wave: Yes, sorry for the delay on my end. I've been busy with other projects but would still like to land this. I'll find time soon to rebase and understand what's changed. If your team needs this sooner, though, don't let me hold you up. :)

Manciukic:
Hey @acj, are you still planning to take this to the finish line?

acj:
@Manciukic Yes, I'd like to. Sorry again for the long delay.

Manciukic:
Hey @acj, we are reviewing old stale PRs. Are you still planning to work on this?

acj:
@Manciukic Yes, I've been doing Firecracker-related work again recently and have been thinking about this.

acj:
@bchalios I've finally made some progress on this.

> One final thing that bothers me is testing this through integration tests. We can't use FreeBSD to do this. It is not feasible for us to maintain FreeBSD artifacts for testing, so it would need to be done using Linux drivers. However, I haven't found a way to trigger a reset through Linux.

I experimented a bit and can reliably trigger a reset of the net device from inside a Linux guest and then restore it to a functional state:

```
# Reset happens here
echo virtio1 > /sys/bus/virtio/devices/virtio1/driver/unbind

# Then restore the device
echo virtio1 > /sys/bus/virtio/drivers/virtio_net/bind
ip addr add 172.16.0.2/30 dev eth0
ip link set eth0 up
ip route add default via 172.16.0.1
```

Do you think this is enough for an integration test?

bchalios:
> @bchalios I've finally made some progress on this.
> 
> > One final thing that bothers me is testing this through integration tests. We can't use FreeBSD to do this. It is not feasible for us to maintain FreeBSD artifacts for testing, so it would need to be done using Linux drivers. However, I haven't found a way to trigger a reset through Linux.
> 
> I experimented a bit and can reliably trigger a reset of the net device from inside a Linux guest and then restore it to a functional state:
> 
> ```
> # Reset happens here
> echo virtio1 > /sys/bus/virtio/devices/virtio1/driver/unbind
> 
> # Then restore the device
> echo virtio1 > /sys/bus/virtio/drivers/virtio_net/bind
> ip addr add 172.16.0.2/30 dev eth0
> ip link set eth0 up
> ip route add default via 172.16.0.1
> ```
> 
> Do you think this is enough for an integration test?

Hey @acj, that looks promising yes!

In case you didn't notice, we've added support for PCI devices (which is why you are having conflicts in your PR). Your approach will work fine with MMIO devices. Not sure whether it'll work with PCI. Would you mind testing that as well?

Manciukic:
Hey @acj, did you have time to take a look at babis suggestion and test this with PCI as well?

acj:
@Manciukic Not yet, but I'm still planning to test it. Thanks for being patient with me.

src/vmm/src/devices/virtio/mod.rs
```diff
@@ -76,6 +76,13 @@ pub enum ActivateError {
     QueueMemoryError(QueueError),
 }
 
+// Errors triggered when resetting a VirtioDevice.
+#[derive(Debug, thiserror::Error, displaydoc::Display)]
+pub enum ResetError {
```
ilstam, line 81: I'd merge this with the second patch where the error code is actually used

src/vmm/src/devices/virtio/mmio.rs
```diff
@@ -204,8 +204,12 @@ impl MmioTransport {
                     let mut device_status = self.device_status;
                     let reset_result = self.locked_device().reset();
                     match reset_result {
-                        Some((_interrupt_evt, mut _queue_evts)) => {}
-                        None => {
+                        Ok(_) => {
+                            // The device MUST initialize device status to 0 upon reset.
+                            device_status = INIT;
```
ilstam, line 209: nitpick: I'd actually use 0 here as per the spec. The `device_status` mod definition is a bit weird actually. Most values are supposed to be individual bits, but 0 is a value for the entire register, and 'INIT' is not a name that appears anywhere on the spec. But that doesn't have to do with your changes, just what I noticed by looking at that code, it might be better to refactor it eventually.

src/vmm/src/devices/virtio/mmio.rs
```diff
@@ -471,7 +475,7 @@ pub(crate) mod tests {
         let m = single_region_mem(0x1000);
         let mut dummy = DummyDevice::new();
         // Validate reset is no-op.
-        assert!(dummy.reset().is_none());
+        assert!(dummy.reset().is_err());
```
ilstam, line 478: I'd check for the exact error code here

ilstam:
Hi Adam. Thanks for your contribution. After discussing with the team, we believe it's better to implement this in a way that completely destroys and recreates the device object. This way a) the code becomes generic (no need for specific support for each device model) and b) we ensure no state is leaked after a reset. Since we are planning to add hot(un)pugging support for virtio devices this year, we will bundle this functionality with the rest of the code. However, I think we should be able to re-use your test commit as is. I'll let you know once that is merged. Thank you very much!

acj:
Sounds good @ilstam. Thanks!

PR #4417, opened 2024-01-31
ab: Do not fail with summary listing values below noise threshold

roypat:
In the past, it was possible for a performance A/B-Test to fail with the message listing all the metrics that had a significant change (p<0.01) having only entries of changes where the magnitude is below the noise threshold.

This is because to determine whether we want to print the message, try to correct for false-positives by considering the average change reported across all data points for the metric, and if the average of their relative changes is below the noise threshold, we ignore it. However, this average includes the metrics not deemed statistically significant (p>=0.01), so if we have large variation that is not deemed significant, this can still skew this average, and thus fail the overall test. Since the failure message only contains significant results, it will seem like the test failed with data below the noise threshold.

This commit fixes the above by adding a check that ensures that the average change of a metric across all statistically significant data points is above the noise threshold before failing the test.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/tools/ab_test.py
+++ b/tools/ab_test.py
@@ -23,6 +23,7 @@ import argparse
 import json
 import statistics
 import sys
+from collections import defaultdict
 from pathlib import Path
 
 # Hack to be able to use our test framework code
@@ -265,7 +266,8 @@ def ab_performance_test(
     # Note that with this approach, for performance changes to "cancel out", we would need essentially a perfect split
     # between scenarios that improve performance and scenarios that degrade performance, something we have not
     # ever observed to actually happen.
-    relative_changes_by_metric = {}
+    relative_changes_by_metric = defaultdict(list)
+    relative_changes_significant = defaultdict(list)
 
     failures = []
     for (dimension_set, metric), (result, unit) in results.items():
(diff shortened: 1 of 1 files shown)


tools/ab_test.py
```diff
@@ -282,10 +283,24 @@ def ab_performance_test(
         if result.pvalue < p_thresh and abs(result.statistic) > strength_abs_thresh:
             failures.append((dimension_set, metric, result, unit))
 
+            if metric not in relative_changes_significant:
+                relative_changes_significant[metric] = []
+            relative_changes_significant[metric].append(
+                result.statistic / baseline_mean
+            )
```
pb8o, line 290: `relative_changes_significant.setdefault(metric, []).append(result.statistic / baseline_mean)`

Or maybe use `defaultdict(list)`

pb8o (dismissed):

bchalios (dismissed):

tools/ab_test.py
roypat: oh, neat! done :)

bchalios (approved):

pb8o (approved):

PR #4418, opened 2024-01-31
jailer: check if exec file exists in the jail dir before copy

valighita:
## Changes

Check whether the exec file already exists in the jail dir before copying it.

## Reason

While trying to spawn multiple instances in parallel (more than 500), we're seeing contention while the jailer tries to copy the exec file to each of the instances jail dirs (some VMs end up starting in more than 20 seconds). Hard-linking the exec file in the jail dirs before starting the VMs avoids the issue, but the copy operation in the jailer needs to be skipped.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---

### DIFF
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -472,15 +472,19 @@ impl Env {
         // a new PathBuf, with something like chroot_dir.join(exec_file_name) ?!
         self.chroot_dir.push(exec_file_name);
 
-        // We do a copy instead of a hard-link for 2 reasons
-        // 1. hard-linking is not possible if the file is in another device
-        // 2. while hardlinking would save up disk space and also memory by sharing parts of the
-        //    Firecracker binary (like the executable .text section), this latter part is not
-        //    desirable in Firecracker's threat model. Copying prevents 2 Firecracker processes from
-        //    sharing memory.
-        fs::copy(&self.exec_file_path, &self.chroot_dir).map_err(|err| {
-            JailerError::Copy(self.exec_file_path.clone(), self.chroot_dir.clone(), err)
-        })?;
+        // Copying the exec file is a costly operation. Check if it exists before actually copying
+        // it, to avoid wasting unnecessary time on repeated runs in the same jail.
+        if !self.chroot_dir.is_file() {
+            // We do a copy instead of a hard-link for 2 reasons
+            // 1. hard-linking is not possible if the file is in another device
+            // 2. while hardlinking would save up disk space and also memory by sharing parts of the
+            //    Firecracker binary (like the executable .text section), this latter part is not
+            //    desirable in Firecracker's threat model. Copying prevents 2 Firecracker processes from
+            //    sharing memory.
+            fs::copy(&self.exec_file_path, &self.chroot_dir).map_err(|err| {
+                JailerError::Copy(self.exec_file_path.clone(), self.chroot_dir.clone(), err)
+            })?;
+        }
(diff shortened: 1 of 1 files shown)


zulinx86:
@valighita Thanks for your reply!

> From the security perspective, it's ok if the Firecracker executable is shared between multiple processes.

So does it mean your use case is single-tenant and only running trusted code?

> We're seeing long start times when running multiple instances in parallel and we've measured the time of various steps. We've noticed that the copy operation was the issue.

That makes sense. So sounds like disk I/O throughput contention or something.

> To make it optional, do you think it's better to use an env variable or a command line argument?

A command line argument would be preferable!
It would be better to:
- print a warning message to stderr when using the CLI argument.
- explain its usage and risk in docs and CHANGELOG. (like, When sharing a Firecracker binary file across multiple protected environments using hardlink, it does not ensure that the new process will not share memory with any other Firecracker process. It also does not ensure that the specified source file and the file already existing in the jailed environment are same. It is highly recommended not to use it when running either multi-tenant or untrusted code, or both.)

Thanks,

zulinx86:
And it would be super nice if you were able to share your performance comparison between with and without this change!!

valighita:
Ok. I will submit the changes and the performance comparison when they are ready.

But it's not really clear for us why this can cause a security concern. In theory, only the RO/exec sections of the executable would be shared between different firecracker instances, and any attempt to change the protections to allow writes would lead to a COW operation. This is also true for any shared library (libc for example). Or is this not necessarily a security concern but a policy to not have shared memory?

Thank you,
Valentin,

zulinx86:
> But it's not really clear for us why this can cause a security concern. In theory, only the RO/exec sections of the executable would be shared between different firecracker instances, and any attempt to change the protections to allow writes would lead to a COW operation. This is also true for any shared library (libc for example). Or is this not necessarily a security concern but a policy to not have shared memory?

Sorry for my poor explanation. Yes, that is our threat model, a kind of policy. In theory, if everything is working as intended without bugs, that couldn't be a security issue. We're in a stance of defense in depth, since Firecracker is purpose-built for multi-tenant. You may already know, but Firecracker is statically linked and built with musl toolchain by default. I appreciate your understanding.

zulinx86:
Having another discussion with other maintainers today, we re-confirmed that we wouldn't like to share any memory. This is because it could be vulnerable to some micro-architectural side-channel attacks (e.g. cache-timing attacks like FLUSH+RELOAD attack: https://www.usenix.org/conference/usenixsecurity14/technical-sessions/presentation/yarom). Thanks.

pb8o (commented):
Hi, as @zulinx86 mentioned, the Firecracker jailer is very opinionated on how it does things and was developed with a secure defaults. 

While the change is backwards compatible if the caller does not copy the file in advance, it would silently change behavior. As such we would prefer some mechanism to make this opt-in.

I think we could accept this change if we put this functionality behind an option (`--unsafe-no-copy` is the best name I can come up with), and document the tradeoff.

Another way would be to use a different jailer like `systemd-nspawn` or `runc`, which is more flexible but would require a bit more wiring up.

zulinx86:
Hello @valighita ,

We hope to find you well.
Have you had a chance to look at the previous update?

We don't  receive updates for about 5 months and we think it may not be relevant to you anymore.
We are closing this PR, but please feel free to reopen it or reach us out if you have any questions or updates.

Thanks for your understanding.

PR #4420, opened 2024-01-31
Replace markdownlint with mdformat

roypat:
## Changes

Replaces markdownlint with mdformat, and applies the new tool's formatting to our markdown files. 

Additionally, fixes various visual bugs of our documentation that became apparently from reviewing the diff mdformat generated.

Adds `iptables` to the docker container ahead of time so that @pb8o won't have to build another version in a few weeks.

## Reason

Markdownlint did not offer an automated way to fix its warnings, which made writing CI-conformant markdown a time-intensive manual workflow (especially when it came to reflowing lines to match the line-length constraints). mdformat allows us to automatically reformat our .md files via `mdformat .`, and additionally understand's github's extensions to the markdown specification. Additionally, the fact that we still had various visual bugs (which are fixed by this PR) shows that markdownlint did not actually prevent us from writing bad markdown.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

---


JonathanWoollett-Light (approved):

tools/devctr/poetry.lock
```diff
@@ -2,87 +2,87 @@
 
 [[package]]
 name = "aiohttp"
-version = "3.9.2"
+version = "3.9.1"
```
xmarcalx, line 5: why this version is getting downgraded?

tools/devctr/Dockerfile
```diff
@@ -73,14 +73,13 @@ RUN apt-get update \
         # needed for integration tests
         net-tools iproute2 iperf3 socat fdisk \
         numactl \
+        iptables \
```
xmarcalx, line 76: why we need iptables now?

.mdformat.toml
```diff
@@ -0,0 +1 @@
+wrap = 80
```
xmarcalx, line 1: we use this tool only to enforce 80 character limit? 
how we chose 80? for historical reasons? If for some reason we are trying to be aligned with other more popular project I think the linux kernel increased this number to 100?

xmarcalx: if the tools has default somehow do we want to spell out to avoid that the tool decide to change defaults and then we are forced to do another huge patch like this that migrates all our doc to whatever new rule?

xmarcalx: basically we removed mdl_style.rb and i do not see in this PR any equivalent and i am curious to understand why because maybe that would have prevented also the need to touch all the doc if was already formatted per markdown rules

xmarcalx (changes requested):
I have posted some questions to better understand the change

tools/devctr/poetry.lock
roypat: Mh, good catch, I might have messed up a merge conflict in the lock file :thinking:

roypat: Fixed (its now showing an update because the first step of building a new docker container is "update poetry dependencies")!

tools/devctr/Dockerfile
roypat: @pb8o has more context here. Generally, whenever someone builds a new docker container, we usually perform other changes that other team members will need in the near future, so avoid having to repeatedly build docker containers.

.mdformat.toml
roypat: > we use this tool only to enforce 80 character limit?

We use 80 for historical reasons (mdl enforced 80). I don't have opinions on what the exact number here should be, happy to change it to 100. 

> if the tools has default somehow do we want to spell out to avoid that the tool decide to change defaults and then we are forced to do another huge patch like this that migrates all our doc to whatever new rule?

The tool essentially just ensures that our markdown files are following the "GitHub Flavored Markdown" specification, so apart from the line length thing, it does not offer any configuration.

> basically we removed mdl_style.rb 

We used this file to mostly disable features of mdl. Since those are not present in mdformat in the first place, there's no need (or a possibility) to have an equivalent.

xmarcalx: > The tool essentially just ensures that our markdown files are following the "GitHub Flavored Markdown" specification, so apart from the line length thing, it does not offer any configuration.

I do not know the tool so i am not an expert of it. I quickly googled and find out that at least seems there are 3 options https://mdformat.readthedocs.io/en/stable/users/style.html so, i am interpret it correctly, my suggestion is to spell them out so it is clear which rules we are using and we are immune from potential changes dictated by that project to default values. Thanks.

tools/devctr/poetry.lock
xmarcalx: thanks for the change

tools/devctr/Dockerfile
xmarcalx: ok understood, so it is unrelated to this PR, thanks (maybe next time we can call out in the description of the PR), thanks.

roypat: Added to PR description

.mdformat.toml
roypat: Added that link to the body of the first commit. mdformat does not allow any configuration for these, so we cannot spell them out anywhere.

xmarcalx: sorry wrong link, i meant https://mdformat.readthedocs.io/en/stable/users/configuration_file.html

roypat: Ah, yeah, added those two options

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -14,8 +13,7 @@ assignees: ''
 ## To Reproduce
 
 `[Author TODO: Steps to reproduce the behaviour:]`
-`[ 1. Start Firecracker via....]`
-`[ 2. Configure Firecracker via....]`
+`[ 1. Start Firecracker via....]` `[ 2. Configure Firecracker via....]`
```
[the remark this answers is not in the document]
pb8o: Nevermind I see you update it later

tools/devctr/Dockerfile
xmarcalx: thanks

.mdformat.toml
xmarcalx: thanks for the change

tests/integration_tests/style/test_markdown.py
```diff
@@ -18,8 +18,6 @@ def test_markdown_style():
     assert len(md_files) != 0
 
     # Run commands
-    cmd = "mdl -c ../.mdlrc "
-    for fname in md_files:
-        cmd += fname + " "
+    cmd = "mdformat --check " + " ".join(md_files)
```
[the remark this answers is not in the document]
roypat: This is changed in a later commit to print diffs of the changes needed, but `mdformat` does not have a `--diff` option, so each individual file needs to be passed to the `diff` tool :(

CHANGELOG.md
```diff
@@ -3,175 +3,174 @@
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
-and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+and this project adheres to
+[Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## \[Unreleased\]
```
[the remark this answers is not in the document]
roypat: ahh, good catch, fixed in new commit (I think at least, I never used `sed`, so not sure if I got the syntax right >.<)

tests/README.md
```diff
@@ -162,37 +162,35 @@ tools/devtool test
 
 ## FAQ
 
-`Q1:` *I have a shell script that runs my tests and I don't want to rewrite it.*
-`A1:` Insofar as it makes sense, you should write it as a python test function.
-However, you can always call the script from a shim python test function. You
-can also add it as a microvm image resource in the s3 bucket (and it will be
-made available under `microvm.slot.path`) or copy it over to a guest filesystem
-as part of your test.
+`Q1:` *I have a shell script that runs my tests and I don't want to rewrite
```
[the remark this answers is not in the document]
roypat: I tried lists, and it wouldnt let me do spacing between the bullet points. I found a way to make the raw markdown look a bit nicer by using `\` instead of `<br>` (which won't get reflowed into a single paragraph).

CHANGELOG.md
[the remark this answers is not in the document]
roypat: alright, I updated all the scripts and ran them locally, and everything seems to work. The exception is `release-notes.sh`, which prints nothing, but it already prints nothing on the `main` branch, so I'm not sure what needs to be done here :(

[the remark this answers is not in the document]
roypat: Synced offline, added a commit that replaces this script with a python version. It seems that different versions of awk interpret the regex used in the bash script differently, so whether the script works depends on the environment.

tools/release_notes.py
```diff
@@ -0,0 +1,40 @@
...
+
+import sys
+from pathlib import Path
+
+if __name__ == "__main__":
+    cur_version = sys.argv[1]
+
+    with open(Path(__file__).parent.parent / "CHANGELOG.md", encoding="UTF-8") as f:
+        changelog_lines = f.readlines()
+
```
pb8o, line 17: nit:

```python
changelog_lines = Path(__file__).joinpath("../../CHANGELOG.md").open(encoding="UTF-8").readlines()
```

tools/release_notes.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
```
pb8o, line 1: I would prefer to keep the name of `release-notes.py`, also it doesn't match the references above.

pb8o (changes requested):
`mv release_notes.py release-notes.py`

tools/release_notes.py
roypat: This doesn't close the file I think, but I can do `with Path(...).joinpath(...).open() as f`!

tools/release_notes.py
roypat: I originally renamed it because pylint was complaining about the dash, but now it doesnt complain anymore, so no idea what that was about. Changed it back :)

tools/release_notes.py
roypat: Mh, it needs one `.parent` because otherwise it tries to construct the path `tools/release-notes.py/../../CHANGEOG.md`

CHANGELOG.md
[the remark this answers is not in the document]
pb8o: Thanks!

tools/release_notes.py
pb8o: Oh you can do:

```python
changelog_lines = Path(__file__).joinpath("../../CHANGELOG.md").read_text(encoding="UTF-8").splitlines()
```

tools/release_notes.py
pb8o: Thanks!

pb8o (approved):

xmarcalx (approved):
LGTM thanks for the changes

tools/release_notes.py
roypat: Ah, cool, didn't know about `splitlines`. Will keep in mind for next time :)

PR #4428, opened 2024-02-06
Add ACPI support for x86_64 

bchalios:
## Changes

Add minimal ACPI support for x86. We implement ACPI in Hardware Reduced mode. For the time being we support FADT, MADT for interrupt controllers and DSDT for passing info about VirtIO and legacy devices. This changes the way that we expose vCPUs, interrupt controllers and devices to the guest.

We pass information about vCPUs and interrupt controllers via MADT table. As a result, we do not need MPTable any more. We deprecate support for MPTable; it will be removed in a future release.

We pass information about VirtIO devices as AML bytecode in the DSDT table. As a result, we do not need to pass VirtIO devices as command line parameters any more.

## Reason

We want to implement ACPI in order to support the Virtual Machine Generation ID device [1]. Support for this device will come as a follow up PR. 

[1] https://learn.microsoft.com/en-us/windows/win32/hyperv_v2/virtual-machine-generation-identifier

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet


bchalios:
This is in a much better state now. Things that I still want to do is:
1. Add unit-tests for the newly introduced code.
2. Check if we really need to add AML code for legacy devices.

src/vmm/src/device_manager/resources.rs
```diff
@@ -0,0 +1,127 @@
...
+/// * GSIs for legacy x86_64 devices
+/// * GSIs for MMIO devicecs
+/// * Memory allocations in the MMIO address space
+#[derive(Debug)]
+pub struct ResourceAllocator {
+    // Allocator for device interrupt lines
+    gsi_allocator: RefCell<IdAllocator>,
+    // Allocator for memory in the MMIO address space
+    mmio_memory: RefCell<AddressAllocator>,
+}
```
ShadowCurse: Note: I highlighted this code segment, but this comment is about the set of changes in this PR.

I don't like this design and here is why:
- Combining allocators for different resources into 1 struct seems unnecessary, especially considerring that in the end this sturct has different methods to allocate resources from each allocator separately.
- Sharing allocators between `Vmm` and `MMIODeviceManager` is counter intuitive. `Vmm` should not allocate anything, it has `MMIODeviceManager` for this.
- Making `Vmm` as an entity to create acpi tables is also unnecessary mixing of responsibilities. As I can see to create acpi tables, only `guest_memory` and `mmio_infos` are needed. It is simper to just pass them to some acpi table function that will do all the work and leave `Vmm` out of this.
- Spreading `Vmm` impementation between files (not just in `acpi/mod.rs` but also in `acpi/x86.rs`) degrades the code quality. It makes it harder to understand what is `Vmm` actually should do because you need to jump though the hoops to find all `imps` for it. Also as I can see, only `setup_arch_dsdt` method actually uses `self`, but then `pio_device_manager` impl for `append_aml_bytes` retuns a constant data, so this whole implementation for `Vmm` actually does not need `Vmm`.

I would propose something like this instead:
```rust
/// leave all allocators as they are now

/// Make use of the `MMIODeviceInfo` struct that is returned 
/// by the `vmm.mmio_device_manager.register_mmio_virtio_for_boot`
/// call. (We currently just drop the return value)
pub fn attach_virtio_device(...) -> Result<MMIODeviceInfo, ...> {
   ...
    vmm.mmio_device_manager
        .register_mmio_virtio_for_boot(vmm.vm.fd(), id, device)
        .map_err(RegisterMmioDevice)
        /// .map(|_| ()) <- just remove this line
}
/// Update all `attach_boot_timer_device`, `attach_entropy_device` and so on to also
/// return `MMIODeviceInfo`

pub fn build_microvm_for_boot(...)  {
   ...
    let mut mmio_device_infos = vec![];

    if vm_resources.boot_timer {
        let mmio_info = attach_boot_timer_device(&mut vmm, request_ts)?;
        mmio_device_infos.append(mmio_info);
    }

    if let Some(balloon) = vm_resources.balloon.get() {
        let mmio_info = attach_balloon_device(&mut vmm, balloon, event_manager)?;
        mmio_device_infos.append(mmio_info);
    }

    // Maybe worth unwrapping this into a loop which attaches
   // each block separately
    let mmio_infos = attach_block_devices(
        &mut vmm,
        &mut boot_cmdline,
        vm_resources.block.devices.iter(),
        event_manager,
    )?;
    mmio_device_infos.extend(&mmio_infos);
 
    // Same here
    let mmio_infos = attach_net_devices(&mut vmm, vm_resources.net_builder.iter(), event_manager)?;
    mmio_device_infos.extend(&mmio_infos);

    if let Some(unix_vsock) = vm_resources.vsock.get() {
        let mmio_info = attach_unixsock_vsock_device(&mut vmm, unix_vsock, event_manager)?;
        mmio_device_infos.append(mmio_info);
    }

    if let Some(entropy) = vm_resources.entropy.get() {
        let mmio_info = attach_entropy_device(&mut vmm, entropy, event_manager)?;
        mmio_device_infos.append(mmio_info);
    }

   ...

     configure_system_for_boot(
        &mut vmm,
        ...
        mmio_device_infos
    )?;
   ...
}

pub fn configure_system_for_boot(..., mmio_device_infos) {
  ...
  acpi_table::x86::create_acpi_table(&vmm.guest_memory, mmio_device_infos, vcpus.len())?;
  ...
}

// in x86/acpi module
pub fn create_acpi_table(guest_memory: &GuestMemory, mmio_device_infos: ..., vcpu_num: u32) -> Result<...> {
  let acpi_memory = AddressAllocator::new(...);
  let dsdt_addr = acpi_memory.allocate(...);
  let dsdt_table = Dsdt::new(&mmio_device_infos);
  dsdt_table.write_to_guest(&guest_memory, dsdt_addr);
  // repeat for all other tables
  ...
}

```

src/vmm/src/device_manager/legacy.rs
```diff
@@ -169,6 +170,73 @@ impl PortIODeviceManager {
...
+            Self::COM_EVT_1_3_GSI,
+            Self::COM_EVT_2_4_GSI,
+            Self::COM_EVT_1_3_GSI,
+            Self::COM_EVT_2_4_GSI,
+        ];
+        for com in 0u8..4 {
+            // COM1
+            aml::Device::new(
+                format!("_SB_.COM{}", com + 1).as_str().into(),
+                vec![
```
ShadowCurse, line 186: Do we need to allocate vectors here? I seems the data is always the same. Overall can we make acpi tables no to do any unnecessary allocations?

ShadowCurse (changes requested):

bchalios:
Thanks for your review Egor! I think your comments make sense. Let me write what my thinking was behind the design choices.

> * Combining allocators for different resources into 1 struct seems unnecessary, especially considerring that in the end this sturct has different methods to allocate resources from each allocator separately.

I would _probably_ agree if it wasn't for the GSI allocations. The next PR, will add support for the VMGenID device. Both MMIO and ACPI/system devices need IRQs. Unless we have a common allocator we would need to statically divide the IRQ numbers between the two types of devices. We _could_ do that, after all we only have one such device now (VMGenID) but I think it would be much more clanky to do so.

> * Sharing allocators between Vmm and MMIODeviceManager is counter intuitive. Vmm should not allocate anything, it has MMIODeviceManager for this

`Vmm` doesn't, `ResourceAllocator` does. It just lives inside the `Vmm` struct (same as `MMIODeviceManager`). My intention was to create a component that handles guest resources, i.e. MMIO address space, System address space and IRQs, in one place. The main motivation was the fact that we need to share IRQs between VirtIO and ACPI devices.

> * Making Vmm as an entity to create acpi tables is also unnecessary mixing of responsibilities. As I can see to create acpi tables, only guest_memory and mmio_infos are needed. It is simper to just pass them to some acpi table function that will do all the work and leave Vmm out of this.

On this I initially had the exact same though as you. I'm not sure I like it 100% either. The reason why I went with this approach is the separation between x86 and aarch64 (eventually, we will want to add support for ARM as well). In x86 we need guest memory, VirtIO devices, number of vCPUS, PIO devices and (in the next PR) VMGenID and Generic Event Device info (that's 6 arguments). On ARM we won't need PIO devices but we might need to pass over the interrupt controller information.

I really dislike the following pattern:

```
call_some_function(put, 6, parameters, in, here, but)
call_some_function(put, 7, other, parameters, in, here, instead)
```

I believe that architecture specific code should be isolated in architecture specific code files. It's partly an aesthetic thing for me, but I do believe it makes the code more readable.

If you have an alternative that avoids this pattern, I'm happy to follow.

> * Spreading Vmm impementation between files (not just in acpi/mod.rs but also in acpi/x86.rs) degrades the code quality. It makes it harder to understand what is Vmm actually should do because you need to jump though the hoops to find all imps for it. Also as I can see, only setup_arch_dsdt method actually uses self, but then pio_device_manager impl for append_aml_bytes retuns a constant data, so this whole implementation for Vmm actually does not need Vmm.

On spreading `impl Vmm` I don't have strong feelings. I typically like to organize code in files based on functionality rather than the name of the thing, but once we agree on the previous comments, we can revisit that. I think it's less important.

src/vmm/src/acpi/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+        assert!(matches!(vmm.write_acpi_table(&mut sdt), Ok(ACPI_MEM_START)));
+
+        // Let's try to write two 4K pages plus one byte
+        let mut sdt = MockSdt(vec![0; 8193]);
+        assert!(matches!(
+            vmm.write_acpi_table(&mut sdt),
+            Err(AcpiError::VmAllocator(
+                vm_allocator::Error::ResourceNotAvailable
+            ))
+        ));
```
roypat, line 179: nit(applies in a few other places as well, but only commenting here): Just doing `assert!(matches!())` won't print a very nice failure message should the assert fail (it'll just dump some ill-formatted rust syntax and say "assertion failed").

```suggestion
        let err = vmm.write_acpi_table(&mut sdt).unwrap_err();
        assert!(matches!(
            err,
            Err(AcpiError::VmAllocator(
                vm_allocator::Error::ResourceNotAvailable
            ))
        ), "{:?}", err);
```

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [workspace]
-members = ["src/clippy-tracing", "src/cpu-template-helper", "src/firecracker", "src/jailer", "src/log-instrument", "src/log-instrument-macros", "src/rebase-snap", "src/seccompiler", "src/snapshot-editor"]
-default-members = ["src/clippy-tracing", "src/cpu-template-helper", "src/firecracker", "src/rebase-snap", "src/seccompiler", "src/snapshot-editor"]
+members = ["src/acpi-tables", "src/clippy-tracing", "src/cpu-template-helper", "src/firecracker", "src/jailer", "src/log-instrument", "src/log-instrument-macros", "src/rebase-snap", "src/seccompiler", "src/snapshot-editor"]
```
roypat, line 2: Could this just be a module in `vmm`?

src/acpi-tables/src/aml.rs
```diff
@@ -0,0 +1,1868 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#![allow(missing_debug_implementations)]
+
+use std::marker::PhantomData;
+
+pub trait Aml {
+    fn append_aml_bytes(&self, _v: &mut Vec<u8>) {
+        unimplemented!()
+    }
```
roypat, line 14:
```suggestion
    fn append_aml_bytes(&self, _v: &mut Vec<u8>);
```

src/acpi-tables/src/aml.rs
```diff
@@ -0,0 +1,1868 @@
...
+            }
+            n => {
+                bytes.push(0x2f); // MultiNamePrefix
+                bytes.push(n.try_into().unwrap());
+            }
+        };
+
+        for part in self.name_parts.clone().iter_mut() {
+            bytes.extend_from_slice(part.as_ref());
+        }
```
roypat, line 75:
```suggestion
        for part in &self.name_parts {
            bytes.extend_from_slice(part);
        }
```

src/vmm/src/builder.rs
```diff
@@ -1511,14 +1468,7 @@ pub mod tests {
...
 
-        let mut cmdline = default_kernel_cmdline();
-        insert_balloon_device(&mut vmm, &mut cmdline, &mut event_manager, balloon_config);
-        // Check if the vsock device is described in kernel_cmdline.
-        #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-        assert!(cmdline_contains(
-            &cmdline,
-            "virtio_mmio.device=4K@0xd0000000:5"
-        ));
+        insert_balloon_device(&mut vmm, &mut event_manager, balloon_config);
```
roypat, line 1471: This test does not actually assert anything anymore now, so I think we can remove it altogether

src/vmm/src/builder.rs
```diff
@@ -1528,14 +1478,7 @@ pub mod tests {
...
 
-        let mut cmdline = default_kernel_cmdline();
-        insert_entropy_device(&mut vmm, &mut cmdline, &mut event_manager, entropy_config);
-        // Check if the vsock device is described in kernel_cmdline.
-        #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-        assert!(cmdline_contains(
-            &cmdline,
-            "virtio_mmio.device=4K@0xd0000000:5"
-        ));
+        insert_entropy_device(&mut vmm, &mut event_manager, entropy_config);
```
roypat, line 1481: This test does not actually assert anything anymore now, so I think we can remove it altogether

src/vmm/src/builder.rs
```diff
@@ -1547,13 +1490,6 @@ pub mod tests {
...
 
-        let mut cmdline = default_kernel_cmdline();
-        insert_vsock_device(&mut vmm, &mut cmdline, &mut event_manager, vsock_config);
-        // Check if the vsock device is described in kernel_cmdline.
-        #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
-        assert!(cmdline_contains(
-            &cmdline,
-            "virtio_mmio.device=4K@0xd0000000:5"
-        ));
+        insert_vsock_device(&mut vmm, &mut event_manager, vsock_config);
```
roypat, line 1493: This test does not actually assert anything anymore now, so I think we can remove it altogether

roypat (commented):
Note that this is just a first somewhat superficial pass of the PR, as I havent actually read up on ACPI yet. As such, its mostly nits, feel free to just say "no" to them

src/vmm/src/device_manager/resources.rs
```diff
@@ -0,0 +1,158 @@
...
+#[derive(Debug)]
+pub struct ResourceAllocator {
+    // Allocator for device interrupt lines
+    gsi_allocator: RefCell<IdAllocator>,
+    // Allocator for memory in the MMIO address space
+    mmio_memory: RefCell<AddressAllocator>,
+    // Memory allocator for ACPI data
+    #[cfg(target_arch = "x86_64")]
+    acpi_memory: RefCell<AddressAllocator>,
+}
```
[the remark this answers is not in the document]
bchalios: I kind of agree with you that it's not ideal to have the `Rc` separated than the `RefCell`. The main reason why I went with this design was that I wanted to avoid to have to deal with lifetimes. `MMIODeviceManager` and `ACPIDeviceManager` (in the next PR) get to store a reference to the allocator. I am willing to re-consider this, although then I'd probably prefer to break down this change in a separate PR before sending the ACPI PR in.

Speaking about "unidiomatism", do you see a practical problem with the way things are? The one I see is that, in theory, someone could create a `ResourceAllocator` without putting it inside an `Rc` and I don't know how would that play out. But since this is not a library that arbitrary consumers might use `ResourceAllocator`, I considered it ok.

src/vmm/src/acpi/x86_64.rs
```diff
@@ -0,0 +1,53 @@
...
+        fadt.setup_iapc_flags(
+            1 << IAPC_BOOT_ARG_FLAGS_VGA_NOT_PRESENT
+                | 1 << IAPC_BOOT_ARG_FLAGS_PCI_ASPM
+                | 1 << IAPC_BOOT_ARG_FLAGS_MSI_NOT_PRESENT,
+        );
+    }
+
+    pub(crate) fn setup_arch_dsdt(&self, dsdt_data: &mut Vec<u8>) {
+        self.pio_device_manager.append_aml_bytes(dsdt_data);
+    }
```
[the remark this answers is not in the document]
bchalios: because we have an MMIO device manager both for x86 and Aarch64 and I kinda dislike having `#[cfg(target_arch = ...")] all over the place.

src/vmm/src/acpi/x86_64.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    pub(crate) fn get_apic_addr(&self) -> u32 {
+        layout::APIC_ADDR
+    }
+
+    pub(crate) fn rsdp_addr(&self) -> GuestAddress {
+        GuestAddress(layout::RSDP_ADDR)
+    }
```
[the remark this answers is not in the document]
bchalios: This again has to do with my preference for having architecture specific code in architecture specific modules/files. It is true that I can re-export from `vmm::arch`, so that I can just use `vmm::arch::RSDP_ADDR`, but atm we don't have Aarch64 support (so the `vmm::arch::aarch64::layout::RSDP_ADDR` doesn't yet exist), so I chose to hide this behind a function in the architecture specific module.

src/vmm/src/device_manager/mmio.rs
```diff
@@ -74,43 +76,65 @@ pub struct MMIODeviceInfo {
...
+
 /// Manages the complexities of registering a MMIO device.
 #[derive(Debug)]
 pub struct MMIODeviceManager {
     pub(crate) bus: crate::devices::Bus,
-    pub(crate) irq_allocator: IdAllocator,
-    pub(crate) address_allocator: AddressAllocator,
+    pub(crate) resource_allocator: Rc<ResourceAllocator>,
     pub(crate) id_to_dev_info: HashMap<(DeviceType, String), MMIODeviceInfo>,
+    pub(crate) dsdt_data: Vec<u8>,
```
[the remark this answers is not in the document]
bchalios: That's how I had it initially. The problem with this, is that iterating over the bus afterwards won't give me the devices in the order they were added in the bus. That is problem, because the root file system disk needs to appear first in the AML, so that the guest sees it as `/dev/vda`.

src/vmm/src/lib.rs
```diff
@@ -307,12 +309,19 @@ pub struct Vmm {
...
     // Guest VM devices.
     mmio_device_manager: MMIODeviceManager,
     #[cfg(target_arch = "x86_64")]
     pio_device_manager: PortIODeviceManager,
 }
 
+/// SAFETY: safe to implement Sync
+unsafe impl Sync for Vmm {}
+/// SAFETY: safe to implement Send
+unsafe impl Send for Vmm {}
```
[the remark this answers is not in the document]
bchalios: Damn! I forgot about this. Thanks for catching it.

So, it builds, but clippy complains about it. I assume that's exactly because I sticked these `Rc` inside it. Need to re-think about it. I'm open to suggestions.

Cargo.toml
bchalios: My ultimate goal is to merge these with the rust-vmm functionality. That's why I was thinking of it more in terms of a crate external to `vmm`, but I don't have a strong preference.

src/vmm/src/lib.rs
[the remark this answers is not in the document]
roypat: Oh, interesting, yeah I didn't run clippy! Clippy is complaining because if you put a non `Send + Sync` type into an `Arc<Mutex<...>>`, then that's kinda useless because the `Arc<Mutex<...>>` won't implement `Send + Sync` either in this case, meaning it cannot be shared between threads. But if an `Arc<Mutex<...>>` cannot be shared between threads, then it really does not need to be an `Arc<Mutex<...>>`, it can just be a `Rc<RefCell<...>>`, which will be more performant. But its not _wrong_ or unsafe to use `Arc<Mutex<...>>` anyway. Just kinda... useless. So it's fine to just suppress the clippy warning here.

Now, in our case we need `Arc<Mutex<...>>` because our event manager expects everything passed into it to be `Arc<Mutex<...>>`'d. Now the interesting point is that the event-manager is single threaded, so I don't really see _why_ we should require everything to be `Arc<Mutex<...>>`. A `Rc<RefCell<...>>` should be just as good (and in fact, we _know_ that `Rc<RefCell<...>>` would work, because a `Arc<Mutex<!Send + Sync>>` is just an `Rc<RefCell<...>>` with inefficient `borrow_mut()` method. If the event-manager actually required the additional functionality provided by `Arc<Mutex<...>>` then we would have a compilation failure as soon as we pass in something that's not `Send + Sync`). 

Going one step further, I wonder if we could even get rid of the `Rc<RefCell<...>>` and just have our event loop be directly `dyn MutEventSubscriber` (or even static dispatch through an enum of everything via register onto the loop). I know at the moment we want access to some fields of for example the `Vmm` struct from outside the event loop, but that should be solvable using channels or similar instead :thinking: 

I'm gonna play around with this for a bit

src/vmm/src/device_manager/resources.rs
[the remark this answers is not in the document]
roypat: > The main reason why I went with this design was that I wanted to avoid to have to deal with lifetimes. MMIODeviceManager and ACPIDeviceManager (in the next PR) get to store a reference to the allocator. 

Mh, I don't see where moving the `Rc` into the `ResourceAllocator` would introduce lifetimes :thinking: (in fact, you cannot have lifetimes here, because you'd get a self-referential struct - if, say, `MmioDeviceManager` gets a reference to the `ResourceAllocator`, then the resource allocator cannot be owned by the `Vmm` struct anymore, as `Vmm` also owns `MmioDeviceManager`. That's why the RC is needed here). You'd simply have multiple `ResourceAllocator`s that point to the same `{Id, Address}Allocator` through the internal `Rc`s. 

> Speaking about "unidiomatism", do you see a practical problem with the way things are? The one I see is that, in theory, someone could create a ResourceAllocator without putting it inside an Rc and I don't know how would that play out.

I don't really see any problems :). A RefCell without an Rc altogether is still a valid usecase. Am happy to keep it the way it is, just had me do a double-take when I started reviewing haha

src/vmm/src/acpi/x86_64.rs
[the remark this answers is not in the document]
roypat: Ahhhh, I see, yes makes sense!

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
roypat: Ah, I see, that's unfortunate. I guess we could use an `indexmap` instead of a `HashMap`, but that'd mean another dependency, so probably not worth it. Could we leave a comment on this field explaining what you explained in this comment? :o

src/acpi-tables/src/fadt.rs
```diff
@@ -0,0 +1,149 @@
...
+            std::mem::size_of::<Self>().try_into().unwrap(),
+            6, // revision 6
+            oem_id,
+            oem_table_id,
+            oem_revision,
+        );
+
+        Fadt {
+            _header: header,
+            _fadt_minor_version: 5,
```
roypat, line 111: nit: no underscore on these

src/acpi-tables/src/fadt.rs
```diff
@@ -0,0 +1,149 @@
...
+            ..Default::default()
+        }
+    }
+
+    /// Set the address of the DSDT table
+    ///
+    /// This sets the 64bit variant, X_DSDT field of the FADT table
+    pub fn set_x_dsdt(&mut self, addr: u64) {
+        self._x_dsdt = U64::new(addr);
```
roypat, line 120: nit: no underscore on `x_dsdt`

src/acpi-tables/src/fadt.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /// Set the FADT flags
+    pub fn set_flags(&mut self, flags: u32) {
+        self._flags = U32::new(flags);
```
roypat, line 125: nit: no underscore on `flags`

src/acpi-tables/src/fadt.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /// Set the IA-PC specific flags
+    pub fn setup_iapc_flags(&mut self, flags: u16) {
+        self._iapc_boot_arch = U16::new(flags);
```
roypat, line 130: nit: no underscore on `iapc_boot_arch`

src/acpi-tables/src/fadt.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /// Set the hypervisor vendor ID
+    pub fn set_hypervisor_vendor_id(&mut self, hypervisor_vendor_id: [u8; 8]) {
+        self._hypervisor_vendor_id = hypervisor_vendor_id;
```
roypat, line 135: nit: no underscore on `hypervisor_vendor_id`

src/acpi-tables/src/madt.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+impl LocalAPIC {
+    pub fn new(cpu_id: u8) -> Self {
+        Self {
+            _type: 0,
+            _length: 8,
+            _processor_uid: cpu_id,
+            _apic_id: cpu_id,
+            _flags: U32::new(1u32 << MADT_CPU_ENABLE_FLAG),
```
roypat, line 31: nit: no underscores on any of these

src/acpi-tables/src/madt.rs
```diff
@@ -0,0 +1,123 @@
...
+
+impl IoAPIC {
+    pub fn new(ioapic_id: u8, apic_address: u32) -> Self {
+        IoAPIC {
+            _type: 1,
+            _length: 12,
+            _ioapic_id: ioapic_id,
+            _reserved: 0,
+            _apic_address: U32::new(apic_address),
+            _gsi_base: U32::ZERO,
```
roypat, line 55: nit: no underscores on any of these

src/acpi-tables/src/madt.rs
```diff
@@ -0,0 +1,123 @@
...
+            6,
+            oem_id,
+            oem_table_id,
+            oem_revision,
+        );
+
+        let mut header = MadtHeader {
+            sdt: sdt_header,
+            _base_address: U32::new(base_address),
+            _flags: U32::ZERO,
```
roypat, line 97: nit: no underscore on `flags`

src/acpi-tables/src/rsdp.rs
```diff
@@ -0,0 +1,59 @@
...
+            // Space in the end of string is needed!
+            _signature: *b"RSD PTR ",
+            checksum: 0,
+            _oem_id: oem_id,
+            _revision: 2,
+            _rsdt_addr: U32::ZERO,
+            _length: U32::new(std::mem::size_of::<Rsdp>().try_into().unwrap()),
+            _xsdt_addr: U64::new(xsdt_addr),
+            extended_checksum: 0,
+            _reserved: [0u8; 3],
```
roypat, line 41: nit: no underscores on any of these

roypat (commented):
As discussed offline, the big dump of underscore related comments xP

okay now I'll really be off reading about ACPI

src/vmm/src/device_manager/resources.rs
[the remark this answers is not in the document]
bchalios: You are right! I hadn't really seen the self-referential issue there. In that case, I would rather go with my approach since it is just one `Rc` shared across all entities that need it, rather than having them share three of those. If you see a functional issue with this, I'm glad to re-consider.

ShadowCurse:
I haven't explored the idea throughout, but what if we combine all "device managers" (`pio_device_manager`, `mmio_device_manager`, `acpi_device_manager`) into 1 `DeviceManager`? This way it can contain all resource allocators and will contain all needed info for acpi tables.
```Rust
pub struct Vmm {
  ...
  device_manager: DeviceManager,
}

pub fn configure_system_for_boot(...) {
  ...
  // I don't like attaching this to `Vmm`, but it has all needed data for it.
  vmm.create_acpi_tables(vcpus.len())?;
  ...
}

impl Vmm {
  pub fn create_acpi_tables(&self, vcpus: usize) -> ... {
    acpi_table::x86::create_acpi_tables(&vmm.guest_memory, ..., vcpus.len())?;
  }
}

File: src/vmm/src/lib.rs (hunk 14 above)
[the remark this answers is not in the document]
Comment (bchalios) [reply src/vmm/src/lib.rs:327]: Thanks a lot for the dive deep into this. I will adjust.

File: src/vmm/src/acpi/x86_64.rs [hunk 24]
```diff
@@ -0,0 +1,53 @@
...
+        // Let the guest kernel know that there is not VGA hardware present
+        // neither do we support ASPM, or MSI type of interrupts.
+        // More info here:
+        // https://uefi.org/specs/ACPI/6.5/05_ACPI_Software_Programming_Model.html?highlight=0a06#ia-pc-boot-architecture-flags
+        fadt.setup_iapc_flags(
+            1 << IAPC_BOOT_ARG_FLAGS_VGA_NOT_PRESENT
+                | 1 << IAPC_BOOT_ARG_FLAGS_PCI_ASPM
+                | 1 << IAPC_BOOT_ARG_FLAGS_MSI_NOT_PRESENT,
+        );
+    }
```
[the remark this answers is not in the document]
bchalios: So, the first one doesn't use `self` in x86, but it will in aarch64 because IRQ handler info is stored inside `Vmm` (in the `vm: Vm` field. I'd like to have the same structure in both architectures. I don't think there's any drawback in terms of performance: these things will most likely be inlined by the compiler (we can always add `#[inline(always)]` to be sure of it). The `setup_arch_fadt` indeed won't touch `self` but it just feels awkwrd to add some of the in the `impl Vmm` block and one not.

So, if we are to keep this code organization structure, I'd prefer keeping them like that.

src/vmm/src/builder.rs
bchalios: same as above

src/vmm/src/builder.rs
roypat: Oh, fair, didnt see that. Nvm then, sorry!

bchalios: no worries! Hadn't seen it myself up until 15' ago :P

src/acpi-tables/src/madt.rs
bchalios: same as before. clippy complains if these are only accessed in `new`

src/acpi-tables/src/madt.rs
bchalios: same as before.

src/acpi-tables/src/fadt.rs
roypat: I feel relatively strongly about disabling the lint for this module/the structs then. I suppose clippy does not understand that we're converting the structs directly to bytes (which is not a "use" of the fields). My reasoning is two-fold:

1. By not having underscores on the fields that we assign in `new`, we have a clear separation between "we do not touch this field at all" and "we write to this field at some point". 
2. (this is more personal preference) I believe we should err on the side of disabling lints when clippy is not understanding the code, instead of working around malfunctioning lints by changing the way we write code.

bchalios:
> I haven't explored the idea throughout, but what if we combine all "device managers" (`pio_device_manager`, `mmio_device_manager`, `acpi_device_manager`) into 1 `DeviceManager`? This way it can contain all resource allocators and will contain all needed info for acpi tables.
> 
> ```rust
> pub struct Vmm {
>   ...
>   device_manager: DeviceManager,
> }
> 
> pub fn configure_system_for_boot(...) {
>   ...
>   // I don't like attaching this to `Vmm`, but it has all needed data for it.
>   vmm.create_acpi_tables(vcpus.len())?;
>   ...
> }
> 
> impl Vmm {
>   pub fn create_acpi_tables(&self, vcpus: usize) -> ... {
>     #[cfg(target_arch = "x86_64")]
>     acpi_table::x86::create_acpi_tables(&self.guest_memory, &self.device_manager, ..., vcpus.len())?;
>   }
> }
> ```

You have raised two issues in your initial comment, essentially:
1. The unification of different allocators in a single container and the fact that Vmm holds a reference to this allocator
2. The fact that ACPI table creation logic holds a reference to the `Vmm` object.

I don't see how the `DeviceManager` container changes the code structure much with either problem. Here's why:

Regarding the `ResourceAllocator` aspect, ACPI and MMIO devices will still need to share the IRQ allocator. With your proposed design, `ResourceAllocator` will not live in the `Vmm` object, but in the `DeviceManager`, but I don't see any practical difference between the two, especially since the latter will still live in the `Vmm` object.

We still need a single IRQ allocatorfor both types of devices. My argument is that we can have a single point where we allocate resources for the guest in a way that some sort of firmware would (that's what we do when we assign IRQs and allocate system memory (for ACPI or MMIO)).

Regarding the second problem, you don't need a `DeviceManager` to do that:

```
impl Vmm {
  pub fn create_acpi_tables(&self, vcpus: usize) -> ... {
    acpi_table::x86::crate_acpi_tables(&self.guest_memory, &self.mmio_device_manager, &self.pio_device_manager, ...)?;
    acpi_table::aarch64::create_acpi_tables(&self.guest_memory, &self.mmio_device_manager, ...)?;
  }
}
```

We can already do that. The reason why I didn't want to do that, is that 90% of what `create_acpi_tables` does is common between x86 and aarch64, so it would be just us replicating code. I wanted to be able to pass to `create_acpi_tables` the context which is common between the two architectures, so we can de-duplicate code and diverge only when it is needed.

If you feel so strongly about the `impl Vmm` thing, I think I can get rid of it (your comment about the PIO AML being constant helped me think of that). I will post a v2 based on this idea and let's discuss based on that.

src/vmm/src/acpi/x86_64.rs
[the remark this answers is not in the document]
roypat: Yeah, I can't really think of functional reasons here, so happy to defer to your preference :)

bchalios:
Changes from v1 to v2:

* Re-worked the logic that writes the tables to guest memory. The functions taking care of that are not members of the `Vmm` type any more
* Removed the `Send` and `Sync` marker traits for `Vmm`. This was a left over. Silenced the related clippy warnings, adding comments why.
* Re-worked `assert!(matches!(...))` blocks to print better messages in case of an error.
* Removed underscores from struct field names in ACPI types. Silenced the clippy for unused code for these structs.
* Added missing copyrights in `acpi-tables` files
* Small fixes in `acpi-tables/src/aml.rs`; making code rust idiomatic.

bchalios:
Changes from v2 to v3:

* Fixed aarch64 build error which was introduced in v2 due to importing of `acpi` module. This needs to be behind a feature gate (x86 only)

src/vmm/src/lib.rs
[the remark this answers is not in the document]
bchalios: Fixed. Removed the marker traits and disabled the lint on which clippy was complaining, adding a comment explaining why.

src/vmm/src/acpi/x86_64.rs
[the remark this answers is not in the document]
bchalios: In v2 these are not part of an `impl Vmm` anymore.

src/acpi-tables/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "acpi_tables"
+version = "0.1.0"
+authors = ["Amazon Firecracker team <firecracker-devel@amazon.com>"]
```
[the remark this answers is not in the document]
bchalios: Added Cloud hypervisor here, although I think it's not needed (?) I think the copyright in the respective files is enough.

src/acpi-tables/src/dsdt.rs
```diff
@@ -0,0 +1,60 @@
+// Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
bchalios: Added copyrights

src/vmm/src/acpi/mod.rs
bchalios: Fixed these wherever I found them. Please let me know if I missed any.

src/acpi-tables/src/aml.rs
bchalios: fixed!

src/acpi-tables/src/aml.rs
bchalios: Fixed

src/acpi-tables/src/fadt.rs
bchalios: I ignore the lint and everything is without an underscore now.

src/acpi-tables/src/fadt.rs
bchalios: I ignore the lint and everything is without an underscore now.

src/acpi-tables/src/fadt.rs
bchalios: I ignore the lint and everything is without an underscore now.

src/acpi-tables/src/fadt.rs
bchalios: I ignore the lint and everything is without an underscore now.

src/acpi-tables/src/fadt.rs
bchalios: I ignore the lint and everything is without an underscore now.

src/acpi-tables/src/madt.rs
bchalios: I ignore the lint and everything is without an underscore now.

src/acpi-tables/src/madt.rs
bchalios: I ignore the lint and everything is without an underscore now.

src/acpi-tables/src/rsdp.rs
bchalios: I ignore the lint and everything is without an underscore now.

src/vmm/src/device_manager/legacy.rs
bchalios: Unfortunately, at the moment yes. This is how the `aml` module handles collections of other objects. Most likely, we can improve this, but I don't think this is a priority right now. My measurements say that building everything need for ACPI takes a few tens of microseconds.

bchalios:
Changes v3 to v4:

* Added back support for MPTable and passing VirtIO devices via kernel command line
* MPTable uses the memory allocator to get space in guest address memory
* Renamed ACPI memory allocator to "System" memory allocator, to reflect the fact that we do not only allocate memory for ACPI.

Cargo.toml
roypat: I think this is fine, lets just keep track of the upstreaming endeavor somewhere :)

src/vmm/src/arch/x86_64/layout.rs
```diff
@@ -29,3 +29,41 @@ pub const KVM_TSS_ADDRESS: u64 = 0xfffb_d000;
...
+
+/// IOAPIC address
+pub const IOAPIC_ADDR: u32 = 0xfec0_0000;
+
+/// Location of RSDP pointer in x86 machines
+pub const RSDP_ADDR: u64 = 0x000e_0000;
+
+/// Start of memory region we will use for system data (MPTable, ACPI, etc). We are putting it
+/// where EBDA normally is, i.e. in the last 1 KiB of the first 640KiB of memory, i.e in the range:
+/// [0x9FC00, 0x9FFFF]
```
roypat, line 44: I'm a bit confused by this comment. I know it originates from the MPTable code, so I think its simply outdated (below we say that we resert 257KiB for ACPI and friends, so the "we are putting it in the [1KiB] range" seems wrong now), but then below you say that even for MPTables we can use up to 5KB, so how was this comment ever true? D:

bchalios: You are right. It should say: "We are putting _starting from_ where EBDA normally is, i.e. ...", as in, the beginning of this area is there but now it extends to more than that. Will remove the range as well.

What I did here is reserve the EBDA area plus a bit more now, all the way up to (but not included) the `HIMEM_START`. We do not need all of this, but I'm just reserving the space for "future use".

> but then below you say that even for MPTables we can use up to 5KB, so how was this comment ever true?

That was true because we limit the number of vCPUs to 32. Both specs use as apic id a `u8` though, which implies up to 256 vCPUs. With this limit we would need ~5000 bytes for describing vCPUS in MPTable :)

src/vmm/src/acpi/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+use crate::device_manager::resources::ResourceAllocator;
+use crate::vstate::memory::{GuestAddress, GuestMemoryMmap};
+use crate::Vcpu;
+
+mod x86_64;
+
+// Our (Original Equipment Manufacturer" (OEM) name. OEM is how ACPI names the manufacturer of the
+// hardware that is exposed to the OS, through ACPI tables. The OEM name is passed in every ACPI
+// table, to let the OS know that we are the owner of the table.
+// More information here:
```
roypat, line 24: nit: where? xP

src/vmm/src/arch/x86_64/mod.rs
```diff
@@ -143,8 +146,16 @@ pub fn configure_system(
         params.hdr.ramdisk_size = u32::try_from(initrd_config.size).unwrap();
     }
 
-    add_e820_entry(&mut params, 0, EBDA_START, E820_RAM)?;
-    add_e820_entry(&mut params, EBDA_START, EBDA_SIZE, E820_RESERVED)?;
+    // We mark first [0x0, EBDA_START) region as usable RAM
+    // and the subsequet [EBDA_START, (EBDA_START + EBDA_SIZE)) and
```
roypat, line 150: we only mark the `[SYSTEM_MEM_START, SYSTEM_MEM_START + SYSTEM_MEM_SIZE]` as reserved (which I guess implicitly also means the EBDA range is marked as reserved, since that's just the first KiB of the 257KiB system range, but the comment is a bit confusing imo)

src/vmm/src/arch/x86_64/layout.rs
roypat: Ahh, right the 32 vcpu limit, thanks for the explaination! :D

src/vmm/src/device_manager/mmio.rs
```diff
@@ -74,43 +78,72 @@ pub struct MMIODeviceInfo {
...
+                &aml::Name::new("_CCA".into(), &aml::ONE),
+                &aml::Name::new(
+                    "_CRS".into(),
+                    &aml::ResourceTemplate::new(vec![
+                        &aml::Memory32Fixed::new(
+                            true,
+                            self.addr.try_into().unwrap(),
+                            self.len.try_into().unwrap(),
+                        ),
+                        &aml::Interrupt::new(true, true, false, false, self.irqs[0]),
```
roypat, line 102: Mh, for these we assume that `self.irqs` is non-empty (and panic otherwise) by accessing index 0, but in line 320 (`register_mmio_boot_timer`), we describe the boot timer as an MMIO device with 0 irqs. How's that all fit together? I would expect the boottime will never show up in the ACPI tables, because its a fake device anyway (and also because all the tests pass). Does it make sense to handle the "no irqs" case here and emit a definition block without `_CRS`? Or should we add a check somewhere that prohibits MMIO devices with zero irqs (and then figure out what to do with the boot timer)?

Cargo.toml
ShadowCurse: I would strongly advice to move this to `vmm`. Otherwise it will become yet another "move crates into vmm" endeavor.

src/firecracker/src/api_server_adapter.rs
```diff
@@ -55,6 +55,12 @@ impl ApiServerAdapter {
         vmm: Arc<Mutex<Vmm>>,
         event_manager: &mut EventManager,
     ) -> Result<(), ApiServerError> {
+        // This wraps ApiServerAdapter inside an Arc<Mutex<...>> when it is not
+        // Send + Sync because the Vmm object (to which we hold a reference to) is not
+        // Send + Sync. The EventManager expects Arc<Mutex<dyn MutEventSubscriber>>
+        // kinds of things though, so here we are. This is *NOT* a problem, since we
+        // do not share the object among threads.
+        #[allow(clippy::arc_with_non_send_sync)]
```
ShadowCurse, line 63: This is quite big red flag to me. We rely on Rust types for safety/sanity reasons. Breaking it now (even if it is "safe") gives us an excuse to do it again in the future. I am sure there are alternative designs that do not require this workaround.

src/vmm/src/builder.rs
```diff
@@ -1087,6 +1098,7 @@ pub mod tests {
             uffd: None,
             vcpus_handles: Vec::new(),
             vcpus_exit_evt,
+            resource_allocator: Rc::new(ResourceAllocator::new().unwrap()),
```
ShadowCurse, line 1116: Even thought it is just a test, you already break your own requirement for `ResourceAllocator` to be shared between `Vmm` and `MMIODeviceManager`.

src/vmm/src/device_manager/mmio.rs
```diff
@@ -578,19 +570,18 @@ mod tests {
         }
     }
 
+    fn default_mmio_manager() -> MMIODeviceManager {
```
ShadowCurse, line 623: This function is same as one in `src/vmm/src/builder.1026`. Why do we even need them if they are 1 line anyway?

src/vmm/src/device_manager/mmio.rs
bchalios: > Don't think we will hit a limit on 18 irq we have access to

What do you mean? We do not control what the users are doing. They might want to have 18 devices and we must have a very good reason to decrease that number (for example adding a device we really care about, such as VMGenID).

src/vmm/src/device_manager/mmio.rs
```diff
@@ -76,12 +78,50 @@ pub struct MMIODeviceInfo {
...
     pub(crate) resource_allocator: Rc<ResourceAllocator>,
     pub(crate) id_to_dev_info: HashMap<(DeviceType, String), MMIODeviceInfo>,
+    // We create the AML byte code for every VirtIO device in the order we build
+    // it, so that we ensure the root block device is appears first in the DSDT.
+    // This is needed, so that the root device appears as `/dev/sda` in the guest
+    // filesystem.
+    // The alternative would be that we iterate the bus to get the data after all
+    // of the devices are build. However, iterating the bus won't give us the
+    // devices in the order they were added.
+    pub(crate) dsdt_data: Vec<u8>,
```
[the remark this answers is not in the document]
bchalios: We already store `MMIODeviceInfo` structs in the hash map of the MMIODeviceManager. Keeping them her would mean that we need to clone them. What's the added value of do so?

src/vmm/src/device_manager/mmio.rs
```diff
@@ -435,6 +479,12 @@ impl MMIODeviceManager {
     }
 }
 
+impl Aml for MMIODeviceManager {
```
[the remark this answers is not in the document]
bchalios: We need _some_ method to give access to DSDT data of `MMIODeviceManager`.

src/firecracker/src/api_server_adapter.rs
bchalios: > We rely on Rust types for safety/sanity reasons

Sure, this thing does not revert you statement. What exactly are you worried about?

src/vmm/src/device_manager/mmio.rs
bchalios: I agree however that it is fine for `MMIODeviceInfo` to have up to 1 IRQ, something like `irq: Option<u32>`.

ShadowCurse: `irq: Option<NonZeroU32>` will work even better, as we starts irq count from 5 (x86) and 32(aarch64).

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
ShadowCurse: The idea here is to store MMIO related info in the `MMIOManager`. By storing ACPI info here we mix things that should not know about each other.

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
ShadowCurse: Why not just access `dsdt_data` directly as it is `pub`?
```rust 
// this
dsdt_data.extend_from_slice(&mmio_device_manager.dsdt_data);

// instead of this
mmio_device_manager.append_aml_bytes(&mut dsdt_data);
```

src/vmm/src/arch/x86_64/mod.rs
bchalios: Fixed the comment. Thanks for catching this.

src/vmm/src/acpi/mod.rs
bchalios: Left over! Removed!

src/vmm/src/device_manager/mmio.rs
bchalios: Removed the `Aml` trait implementation. Now using a dedicated function for VirtIO devices that receives as arguments the memory range and IRQ. Like that we can be sure at compile time that we did not accidentally created DSDT data for something other than a VirtIO device.

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
bchalios: Can do that!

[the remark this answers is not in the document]
ShadowCurse: I think it is.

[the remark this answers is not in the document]
bchalios: Again, I'm fine to do the change, but it's really useful for the reviews if you explain "why" you think this is important.

src/vmm/src/device_manager/mmio.rs
ShadowCurse: The fact that we ended up with 2 `default_mmio_manager` kinda speaks for itself. `MMIODeviceManager::new(Rc::new(ResourceAllocator::new().unwrap())).unwrap()` might be ugly to look at, but it still has no "logic" inside, so I see no reason to shove it into separate "helper" method. I any case, at least let's use  function from `mmio.rs` and delete one from `bulder.rs`.

CHANGELOG.md
```diff
@@ -93,6 +98,10 @@ and this project adheres to
   the `start_time_us` and `start_time_cpu_us` metrics. These parameters were
   never meant to be used by end customers, and we recommend doing any such time
   adjustments outside Firecracker.
+- Booting with microVM kernels that rely on MPTable is deprecated and will be
+  removed in v2.0 or later. We suggest to users of Firecracker to use guest
+  kernels with ACPI support. ACPI will be the only way Firecracker passes
+  hardware information to the guest once MPTable is deprecated.
```
roypat, line 104:
```suggestion
  hardware information to the guest once MPTable is removed.
```
(it'll still work during the deprecation period)

CHANGELOG.md
```diff
@@ -24,6 +24,11 @@ and this project adheres to
   and underlying layers. Queue backlog length metrics are flushed periodically.
   They can be used to esimtate an average queue length by request by dividing
   its value by the number of requests served.
+- [#4428](https://github.com/firecracker-microvm/firecracker/pull/4428): Added
+  basic ACPI support to Firecracker for x86_64 microVMs. This allows booting
+  kernels with ACPI support. Currently, we pass to the guest ACPI tables with
```
roypat, line 29: nit: to remove repetition of "support", maybe

```suggestion
  basic ACPI support to Firecracker for x86_64 microVMs. This allows booting
  ACPI-enabled kernels. Currently, we pass to the guest ACPI tables with
```

CHANGELOG.md
```diff
@@ -93,6 +98,10 @@ and this project adheres to
   the `start_time_us` and `start_time_cpu_us` metrics. These parameters were
   never meant to be used by end customers, and we recommend doing any such time
   adjustments outside Firecracker.
+- Booting with microVM kernels that rely on MPTable is deprecated and will be
+  removed in v2.0 or later. We suggest to users of Firecracker to use guest
```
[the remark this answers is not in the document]
bchalios: If I remove "support" here, the subject of the sentence becomes "Booting with microVM kernels that rely on MPTable", so to me it reads as: "Booting with microVM kernels that rely on MPTable will be removed in v2.0" which doesn't sound right.

bchalios:
Changes from v5 to v6:

* Dropped implementation of `Aml` trait for `MMIODeviceManager`. We now access the AML data directly from `MMIODeviceManager::dsdt_data`.
* Fixes in CHANGELOG wording.

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
bchalios: > that should not know about each other

Why? And why would it be better to store the same object twice?

src/vmm/src/device_manager/mmio.rs
bchalios: > The fact that we ended up with 2 default_mmio_manager kinda speaks for itself

I don't understand what this means. I really don't think it's that important, but I am happy to use the other one.

CHANGELOG.md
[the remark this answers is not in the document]
roypat: My comment is about adding the word "support" :D

src/vmm/src/device_manager/mmio.rs
roypat: maybe we can just implement `Default` here?

CHANGELOG.md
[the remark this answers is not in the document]
bchalios: omg. sorry about that. Need to grab much more coffee.

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
ShadowCurse: The reasons are:
- This trait is not used as a trait bound anywhere
- We can access `dsdt_data` directly, without need for any traits

src/vmm/src/device_manager/mmio.rs
ShadowCurse: > The fact that we ended up with 2 default_mmio_manager kinda speaks for itself

You created second version of a function that already existed, which means that the original function was not known to you. If instead we would have used usual `MMIODeviceManager` constructor everywhere, this problem would have never existed. In the end you still needed to update multiple places where `MMIODeviceManager` is created, so the argument for "change in one place" is also questionable.

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
ShadowCurse: Because MMIO object will only have MMIO info, not ACPI info.

bchalios:
Changes from v6 to v7:

* Do not store the `ResourceAllocator` inside the `MMIODeviceManager`. Pass it only when we register devices. This allows us to only keep the `ResourceAllocator` in `Vmm` and not have to store it inside an `Rc`.

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
ShadowCurse: > The ACPI info is MMIO info.

Isn't it the way around? We convert `MMIODeviceInfo` into ACPI aml bytes.

CHANGELOG.md
```diff
@@ -93,6 +98,10 @@ and this project adheres to
   the `start_time_us` and `start_time_cpu_us` metrics. These parameters were
   never meant to be used by end customers, and we recommend doing any such time
   adjustments outside Firecracker.
+- Booting with microVM kernels that rely on MPTable is deprecated and support
```
ShadowCurse, line 101: I would add a note that this is an x86 only deprecation just in case.

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
bchalios: Yes, we 're saying the same thing. `dsdt_data` here is the `MMIODeviceInfo` translated into AML bytecode, it is not "random ACPI data". The reason why we store it there is described in a previous comment. It's an (unfortunate) way for us to keep this info stored in the order it should appear in the guest. Otherwise, I would just create them on the fly iterating the hash table.

I don't see any reason why we should store a copy of `MMIODeviceInfo`.

CHANGELOG.md
bchalios: agreed, I will add it.

src/vmm/src/device_manager/mmio.rs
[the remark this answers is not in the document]
ShadowCurse: Ok, so this can be removed once `MMIODeviceManager` will be able to remember the order of device insertions. Let's create a TODO/issue to remember this.

ShadowCurse (approved):

roypat (approved):

bchalios:
Changes from v7 to v8:
- Fixed `test_net_change_mac_address` to work both with and without ACPI support.
- Dropped the temporary switch to ACPI-enabled CI artifacts.
- Added documentation about kernel configurations needed for booting guest kernels with ACPI.

CHANGELOG.md
```diff
@@ -25,6 +31,10 @@ and this project adheres to
   the `start_time_us` and `start_time_cpu_us` metrics. These parameters were
   never meant to be used by end customers, and we recommend doing any such time
   adjustments outside Firecracker.
+- Booting with microVM kernels that rely on MPTable on x86_64 is deprecated and
+  support will be removed in v2.0 or later. We suggest to users of Firecracker
+  to use guest kernels with ACPI support. ACPI will be the only way Firecracker
+  passes hardware information to the guest once MPTable support is removed.
```
roypat, line 37: Let's also add this to `DEPRECATED.md`

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ and this project adheres to
 
 ### Added
 
+- [#4428](https://github.com/firecracker-microvm/firecracker/pull/4428): Added
```
sudanl0, line 13: What would be a user noticeable impact from a non-developer point of view?

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ and this project adheres to
...
+  basic ACPI support to Firecracker for x86_64 microVMs. This allows booting
+  ACPI-enabled kernels. Currently, we pass ACPI tables with information about
```
sudanl0, line 15: `ACPI-enabled kernels` might be confusing since IIUC mptables also allowed booting kernels that had ACPI configs enabled.

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ and this project adheres to
 
 ### Added
 
+- [#4428](https://github.com/firecracker-microvm/firecracker/pull/4428): Added
+  basic ACPI support to Firecracker for x86_64 microVMs. This allows booting
```
sudanl0, line 14: `basic ACPI support` creates more questions. Instead it would be better to just say something like "Added ACPI tables to pass information of available vCPUs, interrupt controllers, etc.".

CHANGELOG.md
```diff
@@ -25,6 +31,10 @@ and this project adheres to
   the `start_time_us` and `start_time_cpu_us` metrics. These parameters were
   never meant to be used by end customers, and we recommend doing any such time
   adjustments outside Firecracker.
+- Booting with microVM kernels that rely on MPTable on x86_64 is deprecated and
+  support will be removed in v2.0 or later. We suggest to users of Firecracker
+  to use guest kernels with ACPI support. ACPI will be the only way Firecracker
```
sudanl0, line 36: nit: ACPI will be the only way `on x86_64`
Also, can we add a link to the doc/section which points to the configs they'll need to enable for ACPI?

docs/kernel-policy.md
```diff
@@ -84,6 +85,24 @@ the minimal configuration for a guest kernel for a successful microVM boot is:
...
+Please note that Firecracker does not support PCI devices. The `CONFIG_PCI`
+option is needed for ACPI initialization inside the guest.
+
+If users want to disable the old way of booting (via MPTable and passing VirtIO
+devices as kernel command line parameters) they should disable the following two
+options:
+
+- `CONFIG_X86_MPPARSE`
+- `CONFIG_VIRTIO_MMIO_CMDLINE_DEVICES`
+
```
sudanl0, line 105: We should also mention the 2nd device initialisation failures which is ignored by kernel when ACPI is enabled and Firecracker passes mptable+acpi tables

CHANGELOG.md
bchalios: Done!

src/vmm/src/device_manager/resources.rs
```diff
@@ -0,0 +1,117 @@
...
+impl ResourceAllocator {
+    /// Create a new resource allocator for Firecracker devices
+    pub fn new() -> Result<Self, vm_allocator::Error> {
+        Ok(Self {
+            gsi_allocator: IdAllocator::new(arch::IRQ_BASE, arch::IRQ_MAX)?,
+            mmio_memory: AddressAllocator::new(arch::MMIO_MEM_START, arch::MMIO_MEM_SIZE)?,
+        })
+    }
+
+    /// Allocate a number of GSIs
```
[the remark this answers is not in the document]
bchalios: Done

src/vmm/src/device_manager/resources.rs
```diff
@@ -0,0 +1,117 @@
...
+        // And now we ran out of GSIs
+        assert_eq!(
+            allocator.allocate_gsi(1),
+            Err(vm_allocator::Error::ResourceNotAvailable)
+        );
+        // But we should be able to ask for 0 GSIs
+        assert_eq!(allocator.allocate_gsi(0), Ok(vec![]));
+
+        let mut allocator = ResourceAllocator::new().unwrap();
+        // We should be able to allocat 1 GSI
```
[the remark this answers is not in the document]
bchalios: Fixed

src/acpi-tables/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+displaydoc = "0.2.4"
+thiserror = "1.0.32"
+vm-memory = { version = "0.14.1", features = ["backend-mmap", "backend-bitmap"] }
+zerocopy = { version = "0.7.32", features = ["derive"] }
+
+[lib]
+bench = false
```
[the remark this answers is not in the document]
bchalios: Fixed

src/acpi-tables/src/dsdt.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use std::mem::size_of;
+
+use vm_memory::{Address, Bytes, GuestAddress, GuestMemory};
+use zerocopy::AsBytes;
+
+use crate::{checksum, AcpiError, Result, Sdt, SdtHeader};
+
+#[derive(Debug, Clone)]
+pub struct Dsdt {
```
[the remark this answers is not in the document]
bchalios: Added it here and all the other types (though I point to the 6.5 specification).

src/acpi-tables/src/fadt.rs
```diff
@@ -0,0 +1,155 @@
...
+/// fixed features.
+pub const FADT_F_HW_REDUCED_ACPI: u8 = 20;
+
+// clippy doesn't understand that we actually "use" the fields of this struct when we serialize
+// them as bytes in guest memory, so here we just ignore dead code to avoid having to name
+// everything with an underscore prefix
+#[allow(dead_code)]
+#[repr(packed)]
+#[derive(Debug, Copy, Clone, Default, AsBytes)]
+pub struct Fadt {
```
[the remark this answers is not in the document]
bchalios: Added it here and all the other types (though I point to the 6.5 specification).

CHANGELOG.md
bchalios: I don't think so. Did you have something particular in mind?

CHANGELOG.md
bchalios: Re-worded it to mention that we can boot kernels without MPTable support.

CHANGELOG.md
bchalios: Reworded as suggested

CHANGELOG.md
bchalios: Fixed both.

docs/kernel-policy.md
bchalios: Added some text regarding what exactly is supported and mention the warning messages.

bchalios:
Changes from v8 to v9:

- Pinned 5.10 guest kernel to 5.10.210 for our CI artifacts. Kernels after that don't build with our configuration. Will open issue to track the resolution of this
- Added a kernel configuration with ACPI disabled, so that we can test MPTables as well on x86.
- Mention MPTable deprecation in `DEPRECATED.md`
- Small fixes in resource manager documentation
- Add descriptions about the various ACPI data structures in `acpi-tables` crate pointing also to the latest ACPI specification
- Fixes in CHANGELOG wording
- Added a bit more info in `docs/kernel-policy.md` regarding our supported configurations during the MPTable deprecation period.

CHANGELOG.md
sudanl0: nope, I was wondering if we could add something that highlight what acpi support bring to the customers.

docs/kernel-policy.md
```diff
@@ -81,12 +82,49 @@ the minimal configuration for a guest kernel for a successful microVM boot is:
...
+During the deprecation period Firecracker will continue to support the legacy
+way of booting a microVM. Firecracker will be able to boot kernels with the
+following configurations:
+
+- Only ACPI
+- Only legacy mechanisms
+- Both ACPI and legacy mechanisms
+
+When booting with kernels that support both ACPI and legacy boot mechanisms
```
sudanl0, line 122: I think this can go in the Caveats section starting with something like : "When booting with kernels that support both .. users may see some initialisation failures which they can ignore for the following reasons:" and this follows the explanation provided.

bchalios: Moved it! Good catch.

src/vmm/src/acpi/mod.rs
```diff
@@ -0,0 +1,282 @@
...
+            table.len(),
+            addr
+        );
+
+        Ok(addr)
+    }
+
+    /// Build the DSDT table for the guest
+    fn build_dsdt(&mut self) -> Result<u64, AcpiError> {
+        debug!("acpi: Building DSDT table");
```
ShadowCurse, line 78: Do we really need `debug` logs? They only mention the function they are called from, but don't provide any additional info. Maybe lower them to `trace` level if you think they are still valuable?

bchalios: dropped the debug macros that just print the function we're in. I don't believe it's even worth having them as tracepoints. These functions are very very small.

resources/guest_configs/microvm-kernel-ci-x86_64-4.14.config
```diff
@@ -1,10 +1,10 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/x86 4.14.174 Kernel Configuration
+# Linux/x86 4.14.336 Kernel Configuration
```
pb8o, line 3: Could you add how much the kernel size changes in the commit body?

bchalios: yup, will do!

bchalios: Done!

pb8o (dismissed):

roypat (approved):

pb8o (approved):

sudanl0 (approved):

sudanl0:
Thanks for posting CHANGELOG for each version!

PR #4432, opened 2024-02-07
fix: broken license check for dependencies

sudanl0:
`test_licenses` does not warn or fail if we
accidentally add a dependency having GPL2.0 license. `cargo deny` with current deny.toml prints a warning if we a dependency has GPL2.0 license but the test doesn't consider the cargo output.

We update the deny.toml to deny copyleft (includes GPL) and unlicensed crates but there could be other licenses which we don't want to add accidentally so, along with it assert that `cargo deny` output does not have any warning at all.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/tests/integration_tests/build/test_dependencies.py
+++ b/tests/integration_tests/build/test_dependencies.py
@@ -24,4 +24,17 @@ def test_licenses():
         os.path.join(os.path.dirname(os.path.realpath(__file__)), "../../../Cargo.toml")
     )
 
-    cargo("deny", f"--manifest-path {toml_file} check licenses bans")
+    _, stdout, stderr = cargo("deny", f"--manifest-path {toml_file} check licenses bans")
+    assert "licenses ok" in stdout
+
+    # "cargo deny" should deny licenses by default but for some reason copyleft is allowed
+    # by it and if we add a dependency which has copyleft licenses "cargo deny" won't report
+    # it unless it is explicitly told to do so from the deny.toml.
+    # Our current deny.toml seems to cover all the cases we need but,
+    # if there is an exception like copyleft (where we don't want and don't deny
+    # in deny.toml and is allowed by cardo deny), we don't want to be left in the dark.
+    # For such cases check "cargo deny" output, make sure that there are no warnings reported
+    # related to the license and take appropriate actions i.e. either add them to allow list
+    # or remove them if they are incompatible with our licenses.
+    license_res = [line for line in stderr.split("\n") if "license" in line]
+    assert not license_res
--- a/deny.toml
+++ b/deny.toml
@@ -7,6 +7,12 @@ allow = [
     "Unicode-DFS-2016"
 ]
 
+# Lint level for licenses considered copyleft
+copyleft = "deny"
+
+# The lint level for crates which do not have a detectable license
+unlicensed = "deny"
+
 [[bans.deny]]
 name = "serde_derive"
 version = ">1.0.171, < 1.0.184"


tests/integration_tests/build/test_dependencies.py
```diff
@@ -24,4 +24,10 @@ def test_licenses():
...
     )
 
-    cargo("deny", f"--manifest-path {toml_file} check licenses bans")
+    res = cargo("deny", f"--manifest-path {toml_file} check licenses bans")
+    # if there is an incompatible license which is not copyleft and not added
+    # explicitly in the deny list of deny.toml then, it will be shown as part of
+    # warnings. Make sure we review all warnings related to licenses and take
+    # appropriate actions (i.e. either add them to allow list of create an exception here)
+    license_res = [line for line in res.stderr.split("\n") if "licenses" in line]
+    assert not license_res
```
JonathanWoollett-Light, line 33: My understanding is that if an invalid license is given it will fail the test.

JonathanWoollett-Light (commented):
Can we not use https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html#the-allow-field-optional to achieve the same affect?

JonathanWoollett-Light (approved):

tests/integration_tests/build/test_dependencies.py
roypat, line 33:
```suggestion
    res = cargo("deny", f"--manifest-path {toml_file} check licenses bans")
    # if there is an incompatible license which is not copyleft and not added
    # explicitly in the deny list of deny.toml then, it will be shown as part of
    # warnings. Make sure we review all warnings related to licenses and take
    # appropriate actions (i.e. either add them to allow list of create an exception here)
    _, stdout, stderr = [line for line in res.stderr.split("\n") if "licenses" in line]
    assert not stderr
    assert stdout == "licenses ok"  # not sure if the coloring might mess things up here
```

roypat (changes requested):
type in list comprehension, should be "license" instead of "licenses". I think just asserting stderr is empty is a more robust check here.

sudanl0:
> typo in list comprehension, should be "license" instead of "licenses". I think just asserting stderr is empty is a more robust check here.

yeah I think it was a last minute spell check, just found out while validating it again.

JonathanWoollett-Light (approved):

tests/integration_tests/build/test_dependencies.py
```diff
@@ -24,4 +24,17 @@ def test_licenses():
...
+    # "cargo deny" should deny licenses by default but for some reason copyleft is allowed
+    # by it and if we add a dependency which has copyleft licenses "cargo deny" won't report
+    # it unless it is explicitly told to do so from the deny.toml.
+    # Our current deny.toml seems to cover all the cases we need but,
+    # if there is an exception like copyleft (where we don't want and don't deny
+    # in deny.toml and is allowed by cardo deny), we don't want to be left in the dark.
+    # For such cases check "cargo deny" output, make sure that there are no warnings reported
+    # related to the license and take appropriate actions i.e. either add them to allow list
+    # or remove them if they are incompatible with our licenses.
+    license_res = [line for line in stderr.split("\n") if "license" in line]
```
roypat, line 39: was there a problem with just asserting stderr is empty? I still think that would be a more robust solution, say if the exact output of the tool ever changes.

sudanl0: we have a lot of other warning related to duplicate entries in Cargo.lock e.g. 
```
E   warning[duplicate]: found 2 duplicate entries for crate 'windows_x86_64_msvc'
E       ┌─ /firecracker/Cargo.lock:194:1
E       │  
E   194 │ ╭ windows_x86_64_msvc 0.48.5 registry+https://github.com/rust-lang/crates.io-index
E   195 │ │ windows_x86_64_msvc 0.52.0 registry+https://github.com/rust-lang/crates.io-index
E       │ ╰────────────────────────────────────────────────────────────────────────────────^ lock entries
```

roypat: Ah, thanks for the explanation! Didnt see those because I only ran `deny check license`, not `license bans`

roypat (approved):

PR #4436, opened 2024-02-08
chore: test to check if killing microvm works as expected

sudanl0:
## Changes:
Add a test to make sure there is no change in behaviour in killing firecracker_pid.

## Reason:
In our CI we pass `--new-pid-ns` along with `--daemonize` and so firecracker pid can be read from the file `firecracker.pid` present in the jailers root directory.
However, if there is a change in behaviour that leads to this file not having the actual firecracker pid then it could cause a regression.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/docs/jailer.md
+++ b/docs/jailer.md
@@ -277,6 +277,14 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
   associated with `--daemonize` runs towards the end, instead of the very
   beginning. We are working on adding better logging capabilities.
 
+### Known limitations
+
+- When passing the --daemonize option to Firecracker without the --new-ns-pid
+  option, the Firecracker process will have a different pid than the Jailer
+  process. The suggested workaround to get Firecracker process's pid in this
+  case is using `--new-pid-ns` flag and read Firecracker's pid from the
+  `firecracker.pid` file present in the jailer's root directory.
+
 ## Caveats
 
 - If all the cgroup controllers are bunched up on a single mount point using the
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -255,6 +255,20 @@ class Microvm:
                 LOG.exception("Process not found: %d", self.firecracker_pid)
             except FileNotFoundError:
                 LOG.exception("PID file not found")
+
+            # if microvm was spawned then check if it gets killed
+            if self._spawned:
+                # it is observed that we need to wait some time before
+                #  checking if the process is killed.
+                time.sleep(1)
+                # filter ps results for the jailer's unique id
+                rc, stdout, stderr = utils.run_cmd(
+                    f"ps aux | grep {self.jailer.jailer_id}"
+                )
+                # make sure firecracker was killed
+                assert (
+                    rc == 0 and stderr == "" and stdout.find("firecracker") == -1
+                ), f"Firecracker pid {self.firecracker_pid} was not killed as expected"
         else:
             # Killing screen will send SIGHUP to underlying Firecracker.
             # Needed to avoid false positives in case kill() is called again.


tests/framework/microvm.py
```diff
@@ -255,6 +255,14 @@ def kill(self):
...
             except FileNotFoundError:
                 LOG.exception("PID file not found")
+
+            # filter ps results for the jailer's unique id
+            rc, stdout, stderr = utils.run_cmd(f"ps aux | grep {self.jailer.jailer_id}")
+            assert rc == 0 and stderr == ""
+            # make sure firecracker was killed
+            assert (
+                stdout.find("firecracker") == -1
+            ), f"Firecracker was not killed as expected"
```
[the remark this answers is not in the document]
sudanl0: The string is intentionally added to highlight what the issue is immediately.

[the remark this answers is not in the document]
sudanl0: ah I see, done!

tests/framework/microvm.py
```diff
@@ -255,6 +255,18 @@ def kill(self):
...
+
+            # if microvm was spawned then check if it gets killed
+            if self._spawned:
+                # filter ps results for the jailer's unique id
+                rc, stdout, stderr = utils.run_cmd(
+                    f"ps aux | grep {self.jailer.jailer_id}"
+                )
+                assert rc == 0 and stderr == ""
+                # make sure firecracker was killed
+                assert (
```
kalyazin, line 267: if this is killed in a destructor, will the error propagate to the test run result? Eg marking the test as passed and printing an error out of band isn't very helpful.

docs/jailer.md
```diff
@@ -277,6 +277,13 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
   associated with `--daemonize` runs towards the end, instead of the very
   beginning. We are working on adding better logging capabilities.
 
+### Notes
+
+- The recommended method to get Firecracker process's pid is to use
+  `--new-pid-ns` flag and read Firecracker's pid from the `firecracker.pid` file
+  present in the jailer's root directory. This is also the recommended method to
```
kalyazin, line 284: It feels like there is some repetition in the second sentence. It looks like we want to convey 2 things:
- use `--new-pid-ns`
- use the pidfile

Not sure we need to mention the latter twice.

tests/framework/microvm.py
```diff
@@ -255,6 +255,18 @@ def kill(self):
...
             except FileNotFoundError:
                 LOG.exception("PID file not found")
+
+            # if microvm was spawned then check if it gets killed
+            if self._spawned:
+                # filter ps results for the jailer's unique id
+                rc, stdout, stderr = utils.run_cmd(
+                    f"ps aux | grep {self.jailer.jailer_id}"
+                )
+                assert rc == 0 and stderr == ""
```
kalyazin, line 265: would it make sense to combine both assertions under the same error message? It looks like if grep finds nothing, it won't return 0.

docs/jailer.md
```diff
@@ -277,6 +277,13 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
...
+  fetch Firecracker pid if the pid is needed to kill Firecracker process.
+
```
roypat, line 286: Mhh, I'm not entirely comfortable with saying that `--new-pid-ns` is the recommended way to get the pid, because that implies that the functionality of `--new-pid-ns` is "write a pid file" (but its not: it spawns firecracker into a new pid namespace, which might not be desired by the customer). So I would instead put it as a known issue (both here and on the 1.6 branch) to say something like "When passing the `--daemonize` option to Firecracker without the `--new-ns-pid` option, the Firecracker process will have a different PID than the jailer process, yet fail to write a pid file. To work around this, please see if using the `--new-ns-pid` flag is an option". 

And then when we fix the issue on main by having Firecracker actually write the PID file, remove the note again (or just fix the issue straightaway and never add a note to main, only to 1.6 xP)

tests/framework/microvm.py
sudanl0: yes, the error will propagate to the test run result and I think it would be good to know if it passes in few cases and fails in others.

docs/jailer.md
sudanl0: pid could be used for other stuff as well so in 1st line I just mentioned the way to get it and 2nd line focused on the specific use case we wanted to highlight.

docs/jailer.md
sudanl0: > because that implies that the functionality of --new-pid-ns is "write a pid file"
I don't think it does because we explain what `--new-pid-ns` does in usage and observation section twice. If someone sees this note they will check the other 2 to understand what it is.

> So I would instead put it as a known issue (both here and on the 1.6 branch)
That was the intention but I named the section incorrectly. Yes, the plan is to push on 1.6 as well once this is done.

> or just fix the issue straightaway and never add a note to main, only to 1.6 xP
I didn't want to rush into doing this because this is not a blocked and we have other priority work. But we'll pick this up soon next week :) .

I'll rename the section to known limitation and add the line to mention the change in pids, thank !

tests/framework/microvm.py
sudanl0: When I tested locally I see run_cmd always returns 0 and stdout has at least 1 grep command
```
ipdb> rc, stdout, stderr = utils.run_cmd(f"ps aux | grep something")
ipdb> rc
0
ipdb> print(stdout)
root          93  0.0  0.0   2892   984 pts/1    S+   10:27   0:00 /bin/sh -c ps aux | grep something
root          95  0.0  0.0   4024  2036 pts/1    S+   10:27   0:00 grep something
```

only if the jailer id is empty, the run_cmd throws an exception as below:
```
ipdb> rc, stdout, stderr = utils.run_cmd(f"ps aux | grep ")
*** ChildProcessError: 
[96] Command:
ps aux | grep 
[96] stdout:

[96] stderr:
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.

Returned error code: 2
stderr:
Usage: grep [OPTION]... PATTERNS [FILE]...
Try 'grep --help' for more information.
```

sudanl0: done.

docs/jailer.md
sudanl0: renamed the section and removed the recommendation to kill firecracker. Since we mention the change in behaviour I don't think we need to explicitly suggest what should be done to kill firecracker.
If we work on making the firecracker pid always available in the file then I think we can have a separate section in Notes to suggest a way to kill firecracker.

docs/jailer.md
sudanl0: done, more details https://github.com/firecracker-microvm/firecracker/pull/4436#discussion_r1484150881 
Thanks!

tests/framework/microvm.py
roypat: This should be fine, we dont do cleanup in destructors (because that caused issues in the past because python do not give strong guarantees about when and how often `__del__` is run). Instead it happens during fixture teardown, and pytest will associate these failures with the correct test.

docs/jailer.md
roypat: thanks! :)

zulinx86 (approved):

roypat (approved):

tests/framework/microvm.py
```diff
@@ -255,6 +255,20 @@ def kill(self):
                 LOG.exception("Process not found: %d", self.firecracker_pid)
             except FileNotFoundError:
                 LOG.exception("PID file not found")
+
+            # if microvm was spawned then check if it gets killed
+            if self._spawned:
+                # it is observde that we need to wait some time before
```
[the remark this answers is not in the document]
sudanl0: thanks fixed it.

roypat (approved):

zulinx86 (approved):

PR #4438, opened 2024-02-09
test(jailer): enable multi-threaded test

### LINKED ISSUE #4412: Fix running cargo test in parallel for Jailer
**Description:**

The unit tests in jailer fail when run in parallel, the result of `cargo test --package jailer` is as below:

```
failures:

---- env::tests::test_copy_cache_info stdout ----
thread 'env::tests::test_copy_cache_info' panicked at src/jailer/src/env.rs:860:31:
called `Result::unwrap()` on an `Err` value: CgroupHierarchyMissing("No hierarchy found for this cgroup version.")

---- cgroup::tests::test_cgroup_builder_v1 stdout ----
thread 'cgroup::tests::test_cgroup_builder_v1' panicked at src/jailer/src/cgroup.rs:580:17:
called `Result::unwrap()` on an `Err` value: CgroupHierarchyMissing("No hierarchy found for this cgroup version.")

---- cgroup::tests::test_cgroup_builder_v2 stdout ----
thread 'cgroup::tests::test_cgroup_builder_v2' panicked at src/jailer/src/cgroup.rs:588:17:
called `Result::unwrap()` on an `Err` value: CgroupHierarchyMissing("No hierarchy found for this cgroup version.")

---- cgroup::tests::test_cgroup_v2_write_value stdout ----
thread 'cgroup::tests::test_cgroup_v2_write_value' panicked at src/jailer/src/cgroup.rs:649:17:
called `Result::unwrap()` on an `Err` value: CgroupHierarchyMissing("No hierarchy found for this cgroup version.")

---- cgroup::tests::test_cgroup_build stdout ----
thread 'cgroup::tests::test_cgroup_build' panicked at src/jailer/src/cgroup.rs:614:54:
called `Result::unwrap()` on an `Err` value: CgroupHierarchyMissing("No hierarchy found for this cgroup version.")

---- env::tests::test_new_env stdout ----
thread 'env::tests::test_new_env' panicked at src/jailer/src/env.rs:874:35:
This new environment should be created successfully.: FileOpen("/tmp/firecracker/test/jailer/proc/mounts", Os { code: 2, kind: NotFound, message: "No such file or directory" })

---- cgroup::tests::test_cgroup_build_invalid stdout ----
thread 'cgroup::tests::test_cgroup_build_invalid' panicked at src/jailer/src/cgroup.rs:633:54:
called `Result::unwrap()` on an `Err` value: FileOpen("/tmp/firecracker/test/jailer/proc/mounts", Os { code: 2, kind: NotFound, message: "No such file or directory" })

---- env::tests::test_save_exec_file_pid stdout ----
thread 'env::tests::test_save_exec_file_pid' panicked at src/jailer/src/env.rs:860:31:
called `Result::unwrap()` on an `Err` value: FileOpen("/tmp/firecracker/test/jailer/proc/mounts", Os { code: 2, kind: NotFound, message: "No such file or directory" })

---- env::tests::test_userfaultfd_dev stdout ----
thread 'env::tests::test_userfaultfd_dev' panicked at src/jailer/src/env.rs:860:31:
called `Result::unwrap()` on an `Err` value: FileOpen("/tmp/firecracker/test/jailer/proc/mounts", Os { code: 2, kind: NotFound, message: "No such file or directory" })

---- env::tests::test_mknod_and_own_dev stdout ----
thread 'env::tests::test_mknod_and_own_dev' panicked at src/jailer/src/env.rs:860:31:
called `Result::unwrap()` on an `Err` value: FileOpen("/tmp/firecracker/test/jailer/proc/mounts", Os { code: 2, kind: NotFound, message: "No such file or directory" })

---- env::tests::test_setup_jailed_folder stdout ----
thread 'env::tests::test_setup_jailed_folder' panicked at src/jailer/src/env.rs:860:31:
called `Result::unwrap()` on an `Err` value: FileOpen("/tmp/firecracker/test/jailer/proc/mounts", Os { code: 2, kind: NotFound, message: "No such file or directory" })

---- env::tests::test_copy_exec_to_chroot stdout ----
thread 'env::tests::test_copy_exec_to_chroot' panicked at src/jailer/src/env.rs:1215:9:
assertion `left == right` failed
  left: []
 right: [115, 111, 109, 101, 95, 99, 111, 110, 116, 101, 110, 116]

---- env::tests::test_cgroups_parsing stdout ----
thread 'env::tests::test_cgroups_parsing' panicked at src/jailer/src/env.rs:1298:31:
called `Result::unwrap()` on an `Err` value: FileOpen("/tmp/firecracker/test/jailer/proc/mounts", Os { code: 2, kind: NotFound, message: "No such file or directory" })

---- cgroup::tests::test_cgroup_builder_v2_with_v1_mounts stdout ----
thread 'cgroup::tests::test_cgroup_builder_v2_with_v1_mounts' panicked at src/jailer/src/cgroup.rs:596:17:
called `Result::unwrap_err()` on an `Ok` value: CgroupBuilder { version: 2, hierarchies: {"unified": "/tmp/firecracker/test/jailer/sys_cgroup/unified"}, mount_points: [] }


failures:
    cgroup::tests::test_cgroup_build
    cgroup::tests::test_cgroup_build_invalid
    cgroup::tests::test_cgroup_builder_v1
    cgroup::tests::test_cgroup_builder_v2
    cgroup::tests::test_cgroup_builder_v2_with_v1_mounts
    cgroup::tests::test_cgroup_v2_write_value
    env::tests::test_cgroups_parsing
    env::tests::test_copy_cache_info
    env::tests::test_copy_exec_to_chroot
    env::tests::test_mknod_and_own_dev
    env::tests::test_new_env
    env::tests::test_save_exec_file_pid
    env::tests::test_setup_jailed_folder
    env::tests::test_userfaultfd_dev

test result: FAILED. 17 passed; 14 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.26s

error: test failed, to rerun pass `-p jailer --bin jailer`
root@cff916670245:/firecracker# 
```

The tests run successfully when run serially with this command : `cargo test --package jailer -- --test-threads=1`

### Action item

To fix the unit tests so they can run in parallel.

### Acceptance criteria

PR with the fix to run `cargo test --package jailer` command without any errors.

cm-iwata:
## Changes

Previously, all tests shared same temporary file/directory, causing concurrency conflicts when running tests in multi-threaded.
Resolved test concurrency issues by incorporating random strings into file/directory names.

## Reason

Fixes  #4412

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet


zulinx86:
Hello @cm-iwata ,

Thank you for your contribution!
Could you please fix clippy test errors?
https://buildkite.com/firecracker/firecracker-pr/builds/8932#018d8d15-d42f-4863-969e-4df250266426/55-402

cm-iwata:
@zulinx86 
Sorry. fixed it.

zulinx86:
@cm-iwata 

Could you please do the following?:
- Add a commit message explaining the context like why this commit is needed ( https://buildkite.com/firecracker/firecracker-pr/builds/8976#018d9cd9-1a22-4508-b4bb-f1da0279335a/55-182 )
- Fix all the unit test failures ( https://buildkite.com/firecracker/firecracker-pr/builds/8976#018d9cd9-1a2e-407f-b159-6eca360663da )
- Post the result here when running it in parallel as instructed in issue #4412 since we cannot still run all the rust unit tests in parallel at the moment.

Thanks!

zulinx86:
It's a matter of taste, but it might be better to use "test(jailer)" rather than "fix(jailer)". "fix" may look like fixing a bug of jailer. I believe it's like improvement of jailer unit tests.

cm-iwata:
@zulinx86 
Thank you for the review!

> Add a commit message explaining the context like why this commit is needed

fixed.

> Fix all the unit test failures

I believe I've probably fixed the relevant parts.
Since the tests were passing in my dev environment before the fix, I haven't been able to confirm correctly that the fix made the tests pass.
Also, since some unit tests are failing in my dev environment, I haven't been able to confirm that all unit tests are passing.
Could you please run CI?

※I use [GCE](https://github.com/firecracker-microvm/firecracker/blob/main/docs/dev-machine-setup.md#gcp)for dev environment.

>Post the result here when running it in parallel as instructed in Fix running cargo test in parallel for vmm #4412 since we cannot still run all the rust unit tests in parallel at the moment.

here is result.


```
root@dbaa0c959c45:/firecracker# cargo test --package jailer
warning: patch for `kvm-bindings` uses the features mechanism. default-features and features will not take effect because the patch dependency does not support this mechanism
    Finished test [unoptimized + debuginfo] target(s) in 0.10s
     Running unittests src/main.rs (build/cargo_target/debug/deps/jailer-2ccd13efbd56ca77)

running 32 tests
test cgroup::tests::test_cgroup_builder_no_mounts ... ok
test cgroup::tests::test_get_controller ... ok
test env::tests::test_dup2 ... ok
test env::tests::test_join_netns ... ok
test env::tests::test_parse_resource_limits ... ok
test cgroup::tests::test_inherit_from_parent ... ok
test env::tests::test_copy_exec_to_chroot ... ok
test resource_limits::tests::test_default_resource_limits ... ok
test env::tests::test_validate_exec_file ... ok
test cgroup::tests::test_cgroup_builder_v2 ... ok
test resource_limits::tests::test_display_resource ... ok
test cgroup::tests::test_cgroup_builder_v1_with_v2_mounts ... ok
test resource_limits::tests::test_from_resource ... ok
test resource_limits::tests::test_install ... ok
test resource_limits::tests::test_set_resource_limits ... ok
test tests::test_clean_env_vars ... ok
test env::tests::test_save_exec_file_pid ... ok
test cgroup::tests::test_cgroup_builder_v1 ... ok
test cgroup::tests::test_cgroup_builder_v2_with_v1_mounts ... ok
test tests::test_to_cstring ... ok
test env::tests::test_copy_cache_info ... ok
test env::tests::test_cgroups_parsing ... ok
test env::tests::test_setup_jailed_folder ... ok
test env::tests::test_userfaultfd_dev ... ok
test env::tests::test_mknod_and_own_dev ... ok
test cgroup::tests::test_cgroup_build ... ok
test cgroup::tests::test_cgroup_build_invalid ... ok
test cgroup::tests::test_cgroup_v2_write_value ... ok
test env::tests::test_new_env ... ok
test tests::test_fds_close_range ... ok
test tests::test_sanitize_process ... ok
test tests::test_fds_proc ... ok

test result: ok. 32 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.12s
```

src/jailer/src/cgroup.rs
```diff
@@ -28,18 +22,19 @@ struct CgroupMountPoint {
 
 // Allows creation of cgroups on the system for both versions
 #[derive(Debug)]
-pub struct CgroupBuilder {
+pub struct CgroupBuilder<'a> {
     version: u8,
     hierarchies: HashMap<String, PathBuf>,
     mount_points: Vec<CgroupMountPoint>,
+    proc_mounts_path: &'a str,
 }
```
JonathanWoollett-Light, line 30: This may use more memory than before.

It would be better to do:
```suggestion
#[derive(Debug)]
pub struct CgroupBuilder {
    version: u8,
    hierarchies: HashMap<String, PathBuf>,
    mount_points: Vec<CgroupMountPoint>,
    proc_mounts_path: &'static str,
}
```

cm-iwata: @JonathanWoollett-Light 
Thank you for your suggestion. However, since the field requires dynamic values. 
Because it's necessary to isolate the environment between multiple threads running tests.
If static lifetime were used, multiple threads running tests would access the same directory, potentially causing fail tests.

zulinx86:
@cm-iwata Could you please squash 3134792ac78cd256b9b3b0eeebc4dfb6b44d5f87 and 33ced7453eb12edc785cb4c3a9e5d58a0cf1ed04 into one commit? Two identical commits are shown in https://github.com/firecracker-microvm/firecracker/pull/4438/commits.

src/jailer/src/cgroup.rs
```diff
@@ -449,20 +446,25 @@ pub mod test_util {
...
 
-    use super::PROC_MOUNTS;
+    use utils::rand;
+
+    pub fn get_mock_proc_mounts() -> String {
+        format!(
+            "/tmp/firecracker/test/{}/jailer/proc/mounts",
+            rand::rand_alphanumerics(4).into_string().unwrap()
+        )
+    }
```
zulinx86, line 456: `get_mock_proc_mounts()` is always called just before `MockCgroupFs::new()`. How about putting `get_mock_proc_monts()` inside the `MockCgroupFs::new()`? By doing this, the lifetime parameter `<'a>` of `MockCgroupFs` can be removed as follows:
```
-    pub fn get_mock_proc_mounts() -> String {
-        format!(
-            "/tmp/firecracker/test/{}/jailer/proc/mounts",
-            rand::rand_alphanumerics(4).into_string().unwrap()
-        )
-    }
-
     #[derive(Debug)]
-    pub struct MockCgroupFs<'a> {
+    pub struct MockCgroupFs {
         mounts_file: File,
-        pub proc_mount_path: &'a str,
+        pub proc_mount_path: String,
     }
 
     // Helper object that simulates the layout of the cgroup file system
     // This can be used for testing regardless of the availability of a particular
     // version of cgroups on the system
-    impl<'a> MockCgroupFs<'a> {
+    impl MockCgroupFs {
         pub fn create_file_with_contents<P: AsRef<Path> + Debug>(
             filename: P,
             contents: &str,
@@ -480,10 +473,12 @@ pub mod test_util {
             Ok(())
         }
 
-        pub fn new(
-            mock_proc_mount: &'a str,
-        ) -> std::result::Result<MockCgroupFs<'a>, std::io::Error> {
-            let mock_proc_dir = Path::new(mock_proc_mount).parent().unwrap();
+        pub fn new() -> std::result::Result<MockCgroupFs, std::io::Error> {
+            let mock_proc_mount = format!(
+                "/tmp/firecracker/test/{}/jailer/proc/mounts",
+                rand::rand_alphanumerics(4).into_string().unwrap()
+            );
+            let mock_proc_dir = Path::new(&mock_proc_mount).parent().unwrap();
```

src/jailer/src/cgroup.rs
```diff
@@ -48,11 +43,12 @@ impl CgroupBuilder {
...
             hierarchies: HashMap::new(),
             mount_points: Vec::new(),
+            proc_mounts_path,
         };
 
         // search PROC_MOUNTS for cgroup mount points
-        let f = File::open(PROC_MOUNTS)
-            .map_err(|err| JailerError::FileOpen(PathBuf::from(PROC_MOUNTS), err))?;
+        let f = File::open(b.proc_mounts_path)
+            .map_err(|err| JailerError::FileOpen(PathBuf::from(&b.proc_mounts_path), err))?;
```
zulinx86, line 51: nitpicking, but `&` can be removed.

```suggestion
            .map_err(|err| JailerError::FileOpen(PathBuf::from(b.proc_mounts_path), err))?;
```

src/jailer/src/cgroup.rs
```diff
@@ -72,7 +68,8 @@ impl CgroupBuilder {
         ).map_err(JailerError::RegEx)?;
 
         for l in BufReader::new(f).lines() {
-            let l = l.map_err(|err| JailerError::ReadLine(PathBuf::from(PROC_MOUNTS), err))?;
+            let l =
+                l.map_err(|err| JailerError::ReadLine(PathBuf::from(&b.proc_mounts_path), err))?;
```
zulinx86, line 72: nitpicking, but & can be removed.
```suggestion
                l.map_err(|err| JailerError::ReadLine(PathBuf::from(b.proc_mounts_path), err))?;
```

src/jailer/src/cgroup.rs
```diff
@@ -449,20 +446,25 @@ pub mod test_util {
...
+            "/tmp/firecracker/test/{}/jailer/proc/mounts",
+            rand::rand_alphanumerics(4).into_string().unwrap()
+        )
+    }
 
     #[derive(Debug)]
-    pub struct MockCgroupFs {
+    pub struct MockCgroupFs<'a> {
         mounts_file: File,
+        pub proc_mount_path: &'a str,
```
zulinx86, line 461: nitpicking: the name `proc_mounts_path` is used for `CgroupBuilder` and the file path `/proc/mounts`, so `proc_mounts_path` would be preferable here as well.
```suggestion
        pub proc_mounts_path: &'a str,
```

src/jailer/src/cgroup.rs
```diff
@@ -478,28 +480,45 @@ pub mod test_util {
             Ok(())
         }
 
-        pub fn new() -> std::result::Result<MockCgroupFs, std::io::Error> {
+        pub fn new(
+            mock_proc_mount: &'a str,
```
zulinx86, line 484: nitpicking
```suggestion
            mock_proc_mounts: &'a str,
```

src/jailer/src/cgroup.rs
cm-iwata: fixed in e77496a63070784ab99d69f438c221555d99b990

src/jailer/src/cgroup.rs
cm-iwata: fixed in https://github.com/firecracker-microvm/firecracker/commit/e77496a63070784ab99d69f438c221555d99b990

src/jailer/src/cgroup.rs
cm-iwata: fixed in a9979eda67064f70c1b9f249692b9c0f485d36a1

src/jailer/src/cgroup.rs
cm-iwata: fixed in a9979eda67064f70c1b9f249692b9c0f485d36a1

src/jailer/src/cgroup.rs
cm-iwata: fixed in 540eea53557bdbeef44068e5db60f1c705bd339f and 14f8f38b9819514956eb5b9103ec7cd9cccdd6a1

cm-iwata:
> Could you please squash [3134792](https://github.com/firecracker-microvm/firecracker/commit/3134792ac78cd256b9b3b0eeebc4dfb6b44d5f87) and [33ced74](https://github.com/firecracker-microvm/firecracker/commit/33ced7453eb12edc785cb4c3a9e5d58a0cf1ed04) into one commit?

Sorry. I made a mistake while rebasing.
I squashed tow commits.

src/jailer/src/cgroup.rs
cm-iwata: Although only for testing, dynamic values are used in the following places:

https://github.com/firecracker-microvm/firecracker/blob/14f8f38b9819514956eb5b9103ec7cd9cccdd6a1/src/jailer/src/cgroup.rs#L477-L479

https://github.com/firecracker-microvm/firecracker/blob/14f8f38b9819514956eb5b9103ec7cd9cccdd6a1/src/jailer/src/cgroup.rs#L608

How about changing the type of proc_mounts_path to Cow<'static, str>? With this change, the production code can accept the constant PROC_MOUNTS, while the test code can dynamically generate a String. What do you think about this approach?

cm-iwata:
I thought the code was functioning correctly after the fix, but I've noticed occasional test failures.
Could you please advise on the following issues?



## 1.Too many open files

Sometimes tests  fail as follows.

```
---- tests::test_fds_proc stdout ----
thread 'tests::test_fds_proc' panicked at src/jailer/src/main.rs:409:33:
called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Uncategorized, message: "Too many open files" }

---- tests::test_fds_close_range stdout ----
thread 'tests::test_fds_close_range' panicked at src/jailer/src/main.rs:409:33:
called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Uncategorized, message: "Too many open files" }

---- tests::test_sanitize_process stdout ----
thread 'tests::test_sanitize_process' panicked at src/jailer/src/main.rs:409:33:
called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
```



The code causing this issue can be found here.

https://github.com/firecracker-microvm/firecracker/blob/b2eee1924637d659f551bb397cc4f427598d71b4/src/jailer/src/main.rs#L405-L408

When running the tests, 100 files are create per thread, it causing the file descriptor limit to be exceeded when running tests in multithread. Changing the loop condition `n` from 100 to 10 resolved the issue. Is there any specific reason for using the number 100? If not, would it be acceptable to change `n` to 10?





## 2. Bad file descriptor

Even after fixing 1, tests may still fail.

```
---- env::tests::test_new_env stdout ----
thread 'env::tests::test_new_env' panicked at src/jailer/src/env.rs:911:14:
This another new environment should be created successfully.: ReadLine("/tmp/firecracker/test/MCzQ/jailer/proc/mounts", Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" })
```



I believe the cause is the following code.

https://github.com/firecracker-microvm/firecracker/blob/b2eee1924637d659f551bb397cc4f427598d71b4/src/jailer/src/main.rs#L442-L445

The error should occur in the following sequence:

- thread1 opens fd4.
- thread2 opens fd5.
- tread1 closes fd4 and fd5.
- thread 2 access to fd5, then Bad file descriptor error will occur

Commenting out the above test cases allows the tests to run successfully. Do you have any suggestions for how to fix the code? I couldn't think of any good ideas, so I would appreciate any advice.

zulinx86:
Hi @cm-iwata ,

I really apologize for my very late response :(
I was overwhelmed with other prioritized tasks...

I'll definitely come back to this PR soon (at least during this week).
I appreciate your understanding and patience.

Please let me know if you came up with something for the bad file descriptor issue.

Thanks,

zulinx86:
Hi @cm-iwata,

The change itself looks good to me. Could you squash all the commits into one, because the later 4 commits are refactoring for the first one?

Regarding the "Too many open files" issue, I found a pretty weird behavior of Rust test. I inserted a code running `ulimit -n` at the beginning of `run_close_fds_test()` like this.
```
        let output = std::process::Command::new("sh")
            .arg("-c")
            .arg(format!("ulimit -n"))
            .output()
            .expect("Failed to run shell command");
        let stdout = String::from_utf8_lossy(&output.stdout);
        println!("ulimit -n: {stdout}");
```

When it succeeds, the value is 2047. On the other hand, when failing, it is 100.
```
# cargo test -p jailer -- --nocapture
...
ulimit -n: 2047

ulimit -n: 2047
...
test result: ok. 30 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.27s
```

```
# cargo test -p jailer -- --nocapture
...
ulimit -n: 100

ulimit -n: 100
...
thread 'thread 'tests::test_sanitize_processtests::test_fds_close_range' panicked at ' panicked at src/jailer/src/main.rssrc/jailer/src/main.rs::418418::3333:
:
called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Uncategorized, message: "Too many open files" }called `Result::unwrap()` on an `Err` value: Os { code: 24, kind: Uncategorized, message: "Too many open files" }

test tests::test_sanitize_process ... FAILED
test tests::test_fds_close_range ... FAILED
...
```

I'm investigating the root cause of it...

Regarding the second issue (Bad file descriptor), I'm not able to reproduce the issue on my side. Could you provide me reproduction steps?

Thanks,

zulinx86:
Now I found the root cause of different open file limit numbers between runs. Jailer has another test (`test_install()`) that imposes a new limit (100). If the test is concurrently running with them, those tests will fail because they open 100 files. We can increase the limit of `test_install()` to 1000 since its default value is `2048 - 1`.

https://github.com/firecracker-microvm/firecracker/blob/a364da806f8093e8d8ab1a8287be4a0efd4e4658/src/jailer/src/resource_limits.rs#L196

cm-iwata:
@zulinx86 
I squashed all commits and resolve conflicts.

>We can increase the limit of test_install() to 1000 since its default value is 2048 - 1.

Thank you for your research.
In my opinion, It is not good for test case affect other test cases.
Instead of increasing `new_no_file_limit`, how about restoring the resource limits at the completion of the `test_install()`?


>Regarding the second issue (Bad file descriptor), I'm not able to reproduce the issue on my side. Could you provide me reproduction steps?

Reproduction step is only execute `cargo test --package jailer`.
But, I believe this issue is fixed in c839707dfb78f68d24d02bb607e2288f20763630.
Because this commit drop `close_fds_by_reading_proc` function.

zulinx86:
Thank you for your quick action :D
I'll take another look at your change soon, but before that let me respond to your question first.

> > We can increase the limit of test_install() to 1000 since its default value is 2048 - 1.
> 
> Thank you for your research. In my opinion, It is not good for test case affect other test cases. Instead of increasing `new_no_file_limit`, how about restoring the resource limits at the completion of the `test_install()`?

Yes, that would be better. But unfortunately, even if we restore the limit, if they run concurrently, they will fail. This is mainly because Rust runs them in the same process as different threads.

> > Regarding the second issue (Bad file descriptor), I'm not able to reproduce the issue on my side. Could you provide me reproduction steps?
> 
> Reproduction step is only execute `cargo test --package jailer`. But, I believe this issue is fixed in [c839707](https://github.com/firecracker-microvm/firecracker/commit/c839707dfb78f68d24d02bb607e2288f20763630). Because this commit drop `close_fds_by_reading_proc` function.

OK, we got lucky XD

zulinx86 (changes requested):
It seems `proc_mounts_path` of `CgroupHierarchies` is not used from anywhere, so I believe we can remove it. Thanks to it, we can remove the lifetime annotation `'a` as well.
```
diff --git a/src/jailer/src/cgroup.rs b/src/jailer/src/cgroup.rs
index 05501ad0..57c5b60d 100644
--- a/src/jailer/src/cgroup.rs
+++ b/src/jailer/src/cgroup.rs
@@ -22,27 +22,25 @@ struct CgroupMountPoint {
 
 // Holds a cache of discovered mount points and cgroup hierarchies
 #[derive(Debug)]
-struct CgroupHierarchies<'a> {
+struct CgroupHierarchies {
     hierarchies: HashMap<String, PathBuf>,
     mount_points: Vec<CgroupMountPoint>,
-    proc_mounts_path: &'a str,
 }
 
-impl<'a> CgroupHierarchies<'a> {
+impl CgroupHierarchies {
     // Constructs a new cache of hierarchies and mount points
     // It will discover cgroup mount points and hierarchies configured
     // on the system and cache the info required to create cgroups later
     // within this hierarchies
-    fn new(ver: u8, proc_mounts_path: &'a str) -> Result<Self, JailerError> {
+    fn new(ver: u8, proc_mounts_path: &str) -> Result<Self, JailerError> {
         let mut h = CgroupHierarchies {
             hierarchies: HashMap::new(),
             mount_points: Vec::new(),
-            proc_mounts_path,
         };
 
         // search PROC_MOUNTS for cgroup mount points
-        let f = File::open(h.proc_mounts_path)
-            .map_err(|err| JailerError::FileOpen(PathBuf::from(h.proc_mounts_path), err))?;
+        let f = File::open(proc_mounts_path)
+            .map_err(|err| JailerError::FileOpen(PathBuf::from(proc_mounts_path), err))?;
 
         // Regex courtesy of Filippo.
         // This will match on each line from /proc/mounts for both v1 and v2 mount points.
@@ -62,8 +60,7 @@ impl<'a> CgroupHierarchies<'a> {
         ).map_err(JailerError::RegEx)?;
 
         for l in BufReader::new(f).lines() {
-            let l =
-                l.map_err(|err| JailerError::ReadLine(PathBuf::from(h.proc_mounts_path), err))?;
+            let l = l.map_err(|err| JailerError::ReadLine(PathBuf::from(proc_mounts_path), err))?;
             if let Some(capture) = re.captures(&l) {
                 if ver == 2 && capture["ver"].len() == 1 {
                     // Found the cgroupv2 unified mountpoint; with cgroupsv2 there is only one
@@ -135,15 +132,15 @@ impl<'a> CgroupHierarchies<'a> {
 
 // Allows creation of cgroups on the system for both versions
 #[derive(Debug)]
-pub struct CgroupConfigurationBuilder<'a> {
-    hierarchies: CgroupHierarchies<'a>,
+pub struct CgroupConfigurationBuilder {
+    hierarchies: CgroupHierarchies,
     cgroup_conf: CgroupConfiguration,
 }
 
-impl<'a> CgroupConfigurationBuilder<'a> {
+impl CgroupConfigurationBuilder {
     // Creates the builder object
     // It will initialize the CgroupHierarchy cache.
-    pub fn new(ver: u8, proc_mounts_path: &'a str) -> Result<Self, JailerError> {
+    pub fn new(ver: u8, proc_mounts_path: &str) -> Result<Self, JailerError> {
         Ok(CgroupConfigurationBuilder {
             hierarchies: CgroupHierarchies::new(ver, proc_mounts_path)?,
             cgroup_conf: match ver {
```

src/jailer/src/cgroup.rs
```diff
@@ -540,16 +533,28 @@ pub mod test_util {
         }
 
         pub fn new() -> std::result::Result<MockCgroupFs, std::io::Error> {
+            let mock_jailer_dir = format!(
+                "/tmp/firecracker/test/{}/jailer",
+                rand::rand_alphanumerics(4).into_string().unwrap()
```
pb8o, line 538: This would be 62^4, roughly 16M permutations, so the chance of a collision is very low about 1/1M using 16 threads. If we wanted to avoid further we could concatenate the PID with this string.

src/jailer/src/cgroup.rs
```diff
@@ -596,8 +599,14 @@ pub mod test_util {
...
         fn drop(&mut self) {
-            let _ = fs::remove_file(PROC_MOUNTS);
-            let _ = fs::remove_dir_all("/tmp/firecracker/test");
+            let tmp_dir = Path::new(self.proc_mounts_path.as_str())
+                .parent()
+                .unwrap()
+                .parent()
+                .unwrap()
+                .parent()
+                .unwrap();
```
pb8o, line 608: nit: Not sure if this would work: `.ancestors().nth(3)`

pb8o (approved):

zulinx86:
I'm merging this for now. Although leaving some nits, we can do it later.

PR #4442, opened 2024-02-11
Save Firecracker PID in Jailer's root directory

sudanl0:
## Changes

- Always save Firecracker PID in Jailer root directory regardless of whether `new_pid-ns` flag was set.
- Modify the CI to kill Firecracker using the PID saved in the file in Jailer's root directory.
- Add new tests to try all verify the Microvm.kill behaviour is same for all combinations of daemonize/new_pid_ns flags.

## Reason

- There was a change in behaviour introduced in Jailer because of which Firecracker pid will always be different from the Jailer PID. So we need to suggested a reliable workaround to fetch Firecracker PID.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,19 @@ and this project adheres to
   This fixes a bug where a microVM with incompatible balloon and guest memory
   size could be booted, due to the check for this condition happening after
   Firecracker's configuration was updated.
+- [#4259](https://github.com/firecracker-microvm/firecracker/pull/4259): Added a
+  double fork mechanism in the Jailer to avoid setsid() failures occurred while
+  running Jailer as the process group leader. However, this changed the
+  behaviour of Jailer and now the Firecracker process will always have a
+  different PID than the Jailer process.
+  [#4436](https://github.com/firecracker-microvm/firecracker/pull/4436): Added a
+  "Known Limitations" section in the Jailer docs to highlight the above change
+  in behaviour introduced in PR#4259.
+  [#4442](https://github.com/firecracker-microvm/firecracker/pull/4442): As a
+  solution to the change in behaviour introduced in PR#4259, provided a
+  mechanism to reliably fetch Firecracker PID. With this change, Firecracker
+  process's PID will always be available in the Jailer's root directory
+  regardless of whether new_pid_ns was set.
 
 ## \[1.6.0\]
 
--- a/docs/jailer.md
+++ b/docs/jailer.md
@@ -280,10 +280,13 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
 ### Known limitations
 
 - When passing the --daemonize option to Firecracker without the --new-ns-pid
-  option, the Firecracker process will have a different pid than the Jailer
-  process. The suggested workaround to get Firecracker process's pid in this
-  case is using `--new-pid-ns` flag and read Firecracker's pid from the
-  `firecracker.pid` file present in the jailer's root directory.
+  option, the Firecracker process will have a different PID than the Jailer
+  process and killing the Jailer will not kill the Firecracker process. As a
+  workaround to get Firecracker PID, the Jailer stores the PID of the child
+  process in the jail root directory inside `<exec_file_name>.pid` for all cases
+  regardless of whether `--new-pid-ns` was provided. The suggested way to fetch
+  Firecracker's PID when using the Jailer is to read the `firecracker.pid` file
+  present in the Jailer's root directory.
 
 ## Caveats
 
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -246,34 +246,31 @@ class Microvm:
             # as well as an intentional eye-sore in the test report.
             LOG.error(self.log_data)
 
-        if self.jailer.daemonize:
-            try:
-                if self.firecracker_pid:
-                    os.kill(self.firecracker_pid, signal.SIGKILL)
-            except ProcessLookupError:
-                LOG.exception("Process not found: %d", self.firecracker_pid)
-            except FileNotFoundError:
-                LOG.exception("PID file not found")
-
-            # if microvm was spawned then check if it gets killed
-            if self._spawned:
-                # it is observed that we need to wait some time before
-                #  checking if the process is killed.
-                time.sleep(1)
-                # filter ps results for the jailer's unique id
-                rc, stdout, stderr = utils.run_cmd(
-                    f"ps aux | grep {self.jailer.jailer_id}"
-                )
-                # make sure firecracker was killed
-                assert (
-                    rc == 0 and stderr == "" and stdout.find("firecracker") == -1
-                ), f"Firecracker pid {self.firecracker_pid} was not killed as expected"
+        try:
+            if self.firecracker_pid:
+                os.kill(self.firecracker_pid, signal.SIGKILL)
+        except ProcessLookupError:
+            LOG.exception("Process not found: %d", self.firecracker_pid)
+        except FileNotFoundError:
+            LOG.exception("PID file not found")
 
         if self.screen_pid:
             # Killing screen will send SIGHUP to underlying Firecracker.
             # Needed to avoid false positives in case kill() is called again.
             self.expect_kill_by_signal = True
             utils.run_cmd("kill -9 {} || true".format(self.screen_pid))
+
+        # if microvm was spawned then check if it gets killed
+        if self._spawned:
+            # it is observed that we need to wait some time before
+            #  checking if the process is killed.
+            time.sleep(1)
+            # filter ps results for the jailer's unique id
+            rc, stdout, stderr = utils.run_cmd(f"ps aux | grep {self.jailer.jailer_id}")
+            # make sure firecracker was killed
+            assert (
+                rc == 0 and stderr == "" and stdout.find("firecracker") == -1
+            ), f"Firecracker pid {self.firecracker_pid} was not killed as expected"
 
         # Mark the microVM as not spawned, so we avoid trying to kill twice.
         self._spawned = False
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -543,9 +543,6 @@ class Microvm:
             # Checking the timings requires DEBUG level log messages
             self.time_api_requests = False
 
-        if not self.jailer.daemonize:
-            self.jailer.new_pid_ns = False
-
         cmd = [str(self.jailer_binary_path)] + self.jailer.construct_param_list()
         if self._numa_node is not None:
             node = str(self._numa_node)
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -372,22 +372,15 @@ class Microvm:
         return self.api.describe.get().json()["state"]
 
     @property
-    def pid_in_new_ns(self):
-        """Get the pid of the Firecracker process in the new namespace.
-
-        Reads the pid from a file created by jailer with `--new-pid-ns` flag.
-        """
-        # Read the PID stored inside the file.
-        return int(self.jailer.pid_file.read_text(encoding="ascii"))
-
-    @property
     def firecracker_pid(self):
-        """Return Firecracker's PID"""
+        """Return Firecracker's PID
+
+        Reads the pid from a file created by jailer.
+        """
         if not self._spawned:
             return None
-        if self.jailer.new_pid_ns:
-            return self.pid_in_new_ns
-        return self._screen_firecracker_pid
+        # Read the PID stored inside the file.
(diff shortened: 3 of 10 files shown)


tests/framework/utils.py
```diff
@@ -715,14 +717,17 @@ def start_screen_process(screen_log, session_name, binary_path, binary_params):
...
-        )
+    if not ignore_binary_pid:
+        children_count = len(screen_ps.children())
+        if children_count != 1:
+            raise RuntimeError(
+                f"Failed to retrieve child process id for binary '{binary_path}' "
+                f"screen session process had [{children_count}]"
+            )
+
+        return screen_pid, screen_ps.children()[0].pid
```
pb8o, line 728: If there will be no users of this, it would be better to just remove this

tests/framework/microvm.py
```diff
@@ -505,6 +505,7 @@ def spawn(
         log_show_level=False,
         log_show_origin=False,
         metrics_path="fc.ndjson",
+        force_new_pid_ns=False,
```
pb8o, line 508: There is no reason to have this here since it is hardcoded later in the test right?

tests/integration_tests/security/test_jail.py
```diff
@@ -581,3 +581,34 @@ def test_new_pid_namespace(uvm_plain):
...
+):
+    """
+    Test that Firecracker is spawned in a new PID namespace if requested.
+    """
+    microvm = microvm_factory.build(
+        guest_kernel_linux_5_10,
+        rootfs_ubuntu_22,
+    )
+    microvm.jailer.daemonize = daemonize
+    microvm.jailer.new_pid_ns = new_pid_ns
```
pb8o, line 605: Just set this after `spawn` so it does not get overridden

CHANGELOG.md
```diff
@@ -46,6 +46,19 @@ and this project adheres to
...
   Firecracker's configuration was updated.
+- [#4259](https://github.com/firecracker-microvm/firecracker/pull/4259): Added a
+  double fork mechanism in the Jailer to avoid setsid failures occurred while
+  running Jailer as the process group leader. However, this changed the
+  behaviour of Jailer and now the Firecracker process will always have a
+  different PID than the Jailer process.
+  [#4440](https://github.com/firecracker-microvm/firecracker/pull/4440): Added a
+  "Known Limitations" section in the Jailer docs to highlight the above change
+  in behaviour introduced in PR#4259.
+  \[#4442\]https://github.com/firecracker-microvm/firecracker/pull/4442: As a
```
pb8o, line 57: The link needs to be wrapped in ()

tests/framework/microvm.py
sudanl0: This is juts to have less code changes for tests like [serial](https://github.com/sudanl0/firecracker/blob/3a7d32ca42b9d6fe668c5874ea018ca7e4d964a3/tests/integration_tests/functional/test_serial_io.py#L142) where the daemonize is set to false just before we spawn the microvm.

We cannot change the behaviour of Jailer after the spawn() and we need to test the condition "daemonize=False but new_pid_ns=True" so this flag was needed.

docs/jailer.md
```diff
@@ -280,10 +280,13 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
 ### Known limitations
 
 - When passing the --daemonize option to Firecracker without the --new-ns-pid
-  option, the Firecracker process will have a different pid than the Jailer
-  process. The suggested workaround to get Firecracker process's pid in this
-  case is using `--new-pid-ns` flag and read Firecracker's pid from the
-  `firecracker.pid` file present in the jailer's root directory.
+  option, the Firecracker process will have a different PID than the Jailer
+  process. As a workaround to get Firecracker PID, the Jailer now stores the PID
```
zulinx86, line 284: Using "now" sounds CHANGELOG to me (like "we didn't do A, but now we do"). This is a doc and it should always the current behavior.
```suggestion
  process. As a workaround to get Firecracker PID, the Jailer stores the PID
```

CHANGELOG.md
```diff
@@ -46,6 +46,19 @@ and this project adheres to
   This fixes a bug where a microVM with incompatible balloon and guest memory
   size could be booted, due to the check for this condition happening after
   Firecracker's configuration was updated.
+- [#4259](https://github.com/firecracker-microvm/firecracker/pull/4259): Added a
+  double fork mechanism in the Jailer to avoid setsid failures occurred while
```
zulinx86, line 50: nit: it's a matter of taste, but I'd prefer having `()` after a syscall name to make it clear that it is a function like man page does (e.g. https://man7.org/linux/man-pages/man2/setsid.2.html ).
```suggestion
  double fork mechanism in the Jailer to avoid setsid() failures occurred while
```

docs/jailer.md
```diff
@@ -280,10 +280,13 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
...
+  of the child process in the jail root directory inside `<exec_file_name>.pid`
+  for all cases regardless of whether the new_pid_ns flag was set. The suggested
+  way to fetch Firecracker process's pid when using the Jailer is to read
```
zulinx86, line 287: for consistency
```suggestion
  way to fetch Firecracker PID when using the Jailer is to read
```

docs/jailer.md
```diff
@@ -280,10 +280,13 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
...
+  Firecracker's PID from the `firecracker.pid` file present in the Jailer's root
```
zulinx86, line 288: nit: It looks a bit duplicated.
```suggestion
  `firecracker.pid` file present in the Jailer's root
```

tests/framework/utils.py
pb8o: Yeah if there are no users it's better to remove. The caller can implement that if needed.

tests/framework/microvm.py
pb8o: Is that a combination that works? Could we just remove the check in line 547 then?

tests/integration_tests/security/test_jail.py
```diff
@@ -581,3 +581,34 @@ def test_new_pid_namespace(uvm_plain):
...
+def test_firecracker_kill_by_pid(
+    microvm_factory, guest_kernel_linux_5_10, rootfs_ubuntu_22, daemonize, new_pid_ns
+):
+    """
+    Test that Firecracker is spawned in a new PID namespace if requested.
+    """
+    microvm = microvm_factory.build(
+        guest_kernel_linux_5_10,
+        rootfs_ubuntu_22,
+    )
```
pb8o, line 603: Could just use `uvm_plain`.

tests/integration_tests/security/test_jail.py
pb8o: Ah right. I feel there should be no need to add new options. There is `jailer_kwargs` to specify exactly what options the jailer should use.

tests/framework/microvm.py
sudanl0: yes the combination only works if we have line 547.
We have all these places where test explicitly sets the daemonize to False before spawn() https://github.com/search?q=repo%3Afirecracker-microvm%2Ffirecracker+%22jailer.daemonize+%3D+False%22&type=code 
If we remove the force_new_pid_ns flag and line 547 then we'll have to update all these tests to explicitly set new_pid_ns as False and I think it would only make the code look more ugly and increase chances of human error.

Using the force_new_pid_ns flag we keep the behaviour of all the other tests same and make sure that the new test `test_firecracker_kill_by_pid` also works correctly.

tests/integration_tests/security/test_jail.py
sudanl0: We cannot pass new_pid_ns via jailer_kwargs because we call JailerContext with new_pid_ns=True [here](https://github.com/firecracker-microvm/firecracker/blob/fc674a38817c044cde6b2ec8a90fbe15caf27a4f/tests/framework/microvm.py#L186) 
so it will return `multiple values for keyword argument 'new_pid_ns'` error.

For daemonize I chose this way so that its consistent with the other tests code we have in other places plus I think now that I changed code to use uvm_plain we don't have the option to use daemonize via the jailer_kwargs

tests/framework/microvm.py
pb8o: I disagree there. Tests should be explicit about what they want than depend on some (arbitrary) policy done within the Microvm class. If the combination works, there should be a clean way to ask Microvm to do that. We may need to change surrounding code and tests to match the new behavior, but that is more maintainable than adding flags to force behavior.

tests/integration_tests/security/test_jail.py
pb8o: > We cannot pass new_pid_ns via jailer_kwargs because we call JailerContext with new_pid_ns=True [here](https://github.com/firecracker-microvm/firecracker/blob/fc674a38817c044cde6b2ec8a90fbe15caf27a4f/tests/framework/microvm.py#L186)
so it will return multiple values for keyword argument 'new_pid_ns' error.

That can easily be solved like so:

```
    ...
    new_pid_ns=jailer_kwargs.pop("new_pid_ns", True)
```

alternatively, change the JailerContext default to `True` and then remove the True in Microvm.

tests/framework/microvm.py
sudanl0: I think there is some misunderstanding here, the tests **already** depended on some policy that was done in the Microvm class. The existing assumption in the tests is that if daemonize is False then new_pid_ns should be False as well. If we want to make this explicit then we'll have to remove this assumption and update all the tests that depended on it. Updating all the other tests is what I was trying to avoid since it looked like an unrelated change but I have added another commit to make that explicit.
I didn't use jailer_kwargs in the new commit because, some tests used the fixture uvm_plain, some used uvm_plain_initrd etc. so using jailer_kwargs means that we'll need to modify all these fixtures and I would like to avoid more code change at this point since it falls into cleanup category and out of scope for the current task.

docs/jailer.md
```diff
@@ -280,10 +280,12 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
 ### Known limitations
 
 - When passing the --daemonize option to Firecracker without the --new-ns-pid
-  option, the Firecracker process will have a different pid than the Jailer
-  process. The suggested workaround to get Firecracker process's pid in this
-  case is using `--new-pid-ns` flag and read Firecracker's pid from the
-  `firecracker.pid` file present in the jailer's root directory.
+  option, the Firecracker process will have a different PID than the Jailer
```
pb8o, line 283: Isn't this always true when using the daemonize option, regardless of the `--new-pid-ns`?

tests/integration_tests/security/test_jail.py
```diff
@@ -581,3 +581,29 @@ def test_new_pid_namespace(uvm_plain):
...
+def test_firecracker_kill_by_pid(uvm_plain, daemonize, new_pid_ns):
+    """
+    Test that Firecracker is spawned in a new PID namespace if requested.
+    """
+    microvm = uvm_plain
+    microvm.jailer.daemonize = daemonize
+    microvm.jailer.new_pid_ns = new_pid_ns
+    microvm.spawn()
+    microvm.basic_config()
+    microvm.start()
```
pb8o, line 603: Should we check that the microvm is running? Like a `.run.ssh("true")` to see that it is up and running

sudanl0: done

docs/jailer.md
sudanl0: rephrased the comments to make the issue more clear.

pb8o (dismissed):

docs/jailer.md
zulinx86: +1 to remove "without --new-ns-pid option".

docs/jailer.md
```diff
@@ -280,10 +280,13 @@ Note: default value for `<api-sock>` is `/run/firecracker.socket`.
...
+  process and killing the Jailer will not kill the Firecracker process. As a
+  workaround to get Firecracker PID, the Jailer stores the PID of the child
+  process in the jail root directory inside `<exec_file_name>.pid` for all cases
+  regardless of whether the new_pid_ns flag was set. The suggested way to fetch
```
zulinx86, line 287: nit: as users shouldn't care the internal flag, it would be better to use the user-facing parameter name here.
```suggestion
  regardless of whether `--new_pid_ns` was provided. The suggested way to fetch
```

zulinx86 (approved):
LGTM! Thanks for addressing the series of fixes!!

pb8o (approved):

PR #4453, opened 2024-02-15
feat(vmm): Make SerialOut into Write+Send+Debug

DavidVentura:
## Changes

Remove the SerialOut specialized enum and replace it with a generic Trait.

## Reason

Now other `Write` types can be used, such as `std::fs::File`, sockets, etc.
I'm using `vmm` as a crate to launch multiple VMs on the same process, and they all have un-synchronized access to stdout.
## Other

I've decided to not make changes to `stdin` without first having these reviewed first

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- [X] Any required documentation changes (code and docs) are included in this
  PR.
- [X] User-facing changes are mentioned in `CHANGELOG.md`.
- [X] New `TODO`s link to an issue.
- [X] Commits meet

### DIFF
--- a/src/cpu-template-helper/src/utils/mod.rs
+++ b/src/cpu-template-helper/src/utils/mod.rs
@@ -87,6 +87,7 @@ pub fn build_microvm_from_config(
         &vm_resources,
         &mut event_manager,
         &seccomp_filters,
+        Box::new(std::io::stdout()),
     )?;
 
     Ok((vmm, vm_resources))
--- a/src/firecracker/src/main.rs
+++ b/src/firecracker/src/main.rs
@@ -568,11 +568,13 @@ fn build_microvm_from_json(
         VmResources::from_json(&config_json, &instance_info, mmds_size_limit, metadata_json)
             .map_err(BuildFromJsonError::ParseFromJson)?;
     vm_resources.boot_timer = boot_timer_enabled;
+    set_stdout_nonblocking();
     let vmm = vmm::builder::build_and_boot_microvm(
         &instance_info,
         &vm_resources,
         event_manager,
         seccomp_filters,
+        Box::new(io::stdout()),
     )
     .map_err(BuildFromJsonError::StartMicroVM)?;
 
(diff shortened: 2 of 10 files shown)


DavidVentura:
My pre-commit hook was broken :cry:

JonathanWoollett-Light (changes requested):
We want to avoid dynamic dispatch, this is why I previously introduced an enum based approach to remove dynamic dispatch.

I view dynamic dispatch as an anti-pattern that introduces a myriad of other problems. This is the view behind the existence of a popular crate https://crates.io/crates/enum_dispatch.

I think it would be valuable if you can implement static dispatch, but I would agree with adding dynamic dispatch.

DavidVentura:
I wasn't aware :sweat_smile: I'll try to go back to a static dispatch implementation

DavidVentura:
@JonathanWoollett-Light I don't know how to change to be statically dispatched while allowing anything that is Write to be used.

I only could come up with one option; an `Other` variant of the enum which itself is dynamically dispatched:
```rust
pub enum SerialOut {
    Sink(std::io::Sink),
    Stdout(std::io::Stdout),
    Other(Box<dyn Write>),
}
```

Alternatively, we could embed my two use-cases into the enum: a `File` variant and a `log` variant (which needs to keep its own buffer, as `write` is called per character)

bchalios:
Hi @DavidVentura,

I wanted to comment on this:

>  I'm using vmm as a crate to launch multiple VMs on the same process, and they all have un-synchronized access to stdout.

Firstly, even though `vmm` is a library crate, it's not designed to be a generic library. It is tightly coupled with the `firecracker` binary crate (and other utility crates we maintain). That is to say, you cannot rely on the public interface of the crate to be stable in any way, i.e. any commit we do to it might break your tool.

Second (and most important), it seems you imply that you are launching multiple microVMs from the same process without forking. If that is the case, we would strongly advise against that. KVM is the strongest security barrier we rely on, but the process barrier is important as well. In fact we suggest that users launch Firecracker microVMs using the `jailer` (or similar technology), to ensure workload isolation.

If that is not the case, and indeed you launch each microVM in a separate process, I think you could achieve what you want (redirecting stdout to file descriptors other than `stdout`) by other means.

Looking forward to hear your thoughts.

DavidVentura:
> you cannot rely on the public interface of the crate to be stable in any way

That's clear to me

The workload isolation is not so important for my current experiments, so indeed I'm launching multiple VMs from the same process -- the workloads are trusted and only get started inside virtual machines to get dedicated kernels for integration testing.


I understand if this is far outside what you consider reasonable for firecracker to support

bchalios:
Hey @DavidVentura, really sorry for the late response. Thanks for confirming my understanding. Given that your use-case is not something we want to support (quite the contrary), I don't think it would make sense to merge this PR, especially since in order to support this using static dispatch we need to add code that we would not be testing. We typically don't integrate code that we don't plan to test in any way.

I hope that makes sense to you. Glad to hear your thoughts.

bchalios:
@DavidVentura I will close this PR now. Feel free to re-open this if you think there is more to discuss in the future

PR #4456, opened 2024-02-16
[tests] Network and cpu utilization improvements

ShadowCurse:
## Changes
Minor clean up of network and cpu utilization testing code.
Now cpu utilization is tracked for all threads.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/tests/framework/utils_iperf.py
+++ b/tests/framework/utils_iperf.py
@@ -9,9 +9,6 @@ import time
 from framework import utils
 from framework.utils import CmdBuilder, CpuMap, get_cpu_percent
 
-# Number of seconds to wait for the iperf3 server to start
-SERVER_STARTUP_TIME_SEC = 2
-
 
 class IPerf3Test:
     """Class abstracting away the setup and execution of an iperf3-based performance test"""
@@ -54,7 +51,8 @@ class IPerf3Test:
             utils.run_cmd(f"{self._microvm.netns.cmd_prefix()} {cmd}")
             first_free_cpu += 1
 
-        time.sleep(SERVER_STARTUP_TIME_SEC)
+        # Wait for the iperf3 server to start
+        time.sleep(2)
 
         with concurrent.futures.ThreadPoolExecutor() as executor:
             cpu_load_future = executor.submit(
--- a/tests/framework/utils_iperf.py
+++ b/tests/framework/utils_iperf.py
@@ -9,12 +9,6 @@ import time
 from framework import utils
 from framework.utils import CmdBuilder, CpuMap, get_cpu_percent
 
-# Dictionary mapping modes (guest-to-host, host-to-guest, bidirectional) to arguments passed to the iperf3 clients spawned
-MODE_MAP = {"g2h": [""], "h2g": ["-R"], "bd": ["", "-R"]}
-
-# Dictionary doing the reserve of the above, for pretty-printing
-REV_MODE_MAP = {"": "g2h", "-R": "h2g"}
-
 # Number of seconds to wait for the iperf3 server to start
 SERVER_STARTUP_TIME_SEC = 2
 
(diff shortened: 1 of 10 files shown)


tests/framework/utils_iperf.py
```diff
@@ -152,8 +152,11 @@ def guest_command(self, port_offset):
 
 def emit_iperf3_metrics(metrics, iperf_result, omit):
     """Consume the iperf3 data produced by the tcp/vsock throughput performance tests"""
-    for cpu_util in iperf_result["cpu_load_raw"]["firecracker"]:
-        metrics.put_metric("cpu_utilization_vmm", cpu_util, "Percent")
+    cpu_util = iperf_result["cpu_load_raw"]
+    for thread_name in cpu_util:
+        values = cpu_util[thread_name]
+        for value in values:
+            metrics.put_metric(f"cpu_utilization_{thread_name}", value, "Percent")
```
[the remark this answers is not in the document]
ShadowCurse: I run A/B tests several times and the block and net tests pass. I think the cpu usage should be not change much, because we usually have number of tasks equal to number of vcpus.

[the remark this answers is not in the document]
roypat: alright, thanks for checking!

roypat (approved):

tests/integration_tests/performance/test_network_ab.py
```diff
@@ -62,25 +62,15 @@ def network_microvm(request, microvm_factory, guest_kernel, rootfs):
 
 @pytest.mark.nonci
 @pytest.mark.parametrize("network_microvm", [1], indirect=True)
-@pytest.mark.parametrize("iteration", [1, 2])
-def test_network_latency(network_microvm, metrics, iteration):
```
bchalios, line 66: I understand that with automatic re-tries we reduce the failures but does it really hurt to have more iterations? It feels like the more data we have the better.

ShadowCurse: I can increase number of iterations in the test so that it test gives same amount of data as it was with 2 iterations. Considering each iteration gives 450 pings and I think this is enough data.

roypat (approved):

roypat (approved):

tests/integration_tests/performance/test_network_ab.py
bchalios: my point was, why do we need to change this? What do we gain?

ShadowCurse: If there are not benefits of having this test run several iterations, we should not run it several times. The original reason for this implementation is not longer relevant, so we should remove this workaround.

bchalios: Why? It does not cost us anything to have it run more times. Quite the opposite, the more iterations we run of an experiment the more statistical confidence we get for its result. We are not bound by time here, so I do not see why this is a problem.

ShadowCurse: As I said previously, this was a workaround for A/B tests. If 450 data points is not enough for statistical confidence for a ping test, then the problem is not in the number of data points, but somewhere else. I think we should not use the logic of "the more data, the better" in this case, because this leads to the unbound upper limit on number of data we think is enough. Following this logic + as you said "we are not bound by time", why don't we increase number of iterations from 2 to 2000?

ShadowCurse: Why leave code that brings 0 benefit and adds more confusion?

bchalios: Because it works. And it solves a problem, that of having less false positives. And it actually does it with less effort, because most likely avoids needless retries.

ShadowCurse: It avoid needless retries by always doing the retry, just in another form. If this workaround so useful, why don't we add it to all tests. This will maybe save us even more retries.

bchalios: My main point here is that I don't like us getting in the habit of changing things that work, without causing us any problem, just for the sake of it.

ShadowCurse: Well.. I am changing things because I work with this code at this point in time and I think it is worth to fix old hacks on the way.

Issue #1859, opened 2020-05-04
Look into notifying the guest when vcpus are paused


alxiord:
According to the [documentation](https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt):

```
KVM_KVMCLOCK_CTRL

This signals to the host kernel that the specified guest is being paused by
userspace.  The host will set a flag in the pvclock structure that is checked
from the soft lockup watchdog.  [...] This ioctl can be called any time
after pausing the vcpu, but before it is resumed.
```

KVM uses this ioctl to notify the guest that it's being paused([code](https://elixir.bootlin.com/linux/latest/source/arch/x86/kvm/x86.c#L4493)):

```c
case KVM_KVMCLOCK_CTRL: {
    r = kvm_set_guest_paused(vcpu);
    goto out;
}
```

```c
/*
 * kvm_set_guest_paused() indicates to the guest kernel that it has been
 * stopped by the hypervisor.  This function will be called from the host only.
 * EINVAL is returned when the host attempts to set the flag for a guest that
 * does not support pv clocks.
 */
static int kvm_set_guest_paused(struct kvm_vcpu *vcpu)
{
	if (!vcpu->arch.pv_time_enabled)
		return -EINVAL;
	vcpu->arch.pvclock_set_guest_stopped_request = true;
	kvm_make_request(KVM_REQ_CLOCK_UPDATE, vcpu);
	return 0;
}
```

In the context of snapshotting a microVM and resuming it later, this ioctl needs to be evaluated to assess its benefits & costs and whether it should be called immediately after pausing or between deserialization and resuming.

acatangiu:
Adding a bit more detail here:

In a case where the guest is paused by the hypervisor for a non-negligible period of time, that blackout period is seen by the guest as frozen/non-ticking CPUs. On resuming the VM, the guest soft-lockup watchdog might panic.

This ioctl makes KVM on the host set a special flag on the emulated/paravirtualized pv-clock. This flag is seen by the guest and tells the guest that the vCPUs haven't ticked in a while on purpose.

alindima:
> On resuming the VM, the guest soft-lockup watchdog might panic.

Remember seeing any such panics?

sudanl0:
Reference to this in our code https://github.com/firecracker-microvm/firecracker/blob/cb2ea9203da579c2a473f8538815a1e4eb4420d4/src/vmm/src/vstate/vcpu/mod.rs#L299

[...]

PR #4460, opened 2024-02-19
fix(vmm): call KVMCLOCK_CTRL when pausing

kalyazin:
## Changes

Call `KVM_KVMCLOCK_CTRL` when pausing vCPUs. This allows guest kernel's soft lockup watchdog to know that it was the hypervisor that paused the vCPUs and don't trigger an exception.

Fixes: https://github.com/firecracker-microvm/firecracker/issues/1859

## Reason

This is to avoid guest kernel panic on resume path due to soft lockup detection.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ and this project adheres to
   `VIRTIO_NET_F_RX_MRGBUF` support to the `virtio-net` device. When this feature
   is negotiated, guest `virtio-net` driver can perform more efficient memory
   management which in turn improves RX and TX performance.
+- [#4460](https://github.com/firecracker-microvm/firecracker/pull/4460): Add a
+  call to
+  [`KVM_KVMCLOCK_CTRL`](https://docs.kernel.org/virt/kvm/api.html#kvm-kvmclock-ctrl)
+  after pausing vCPUs on x86_64 architectures. This ioctl sets a flag in the KVM
+  state of the vCPU indicating that it has been paused by the host userspace. In
+  guests that use kvmclock, the soft lockup watchdog checks this flag. If it is
+  set, it won't trigger the lockup condition. Calling the ioctl for guests that
+  don't use kvmclock will fail. These failures are not fatal. We log the failure
+  and increase the `vcpu.kvmclock_ctrl_fails` metric.
 
 ### Changed
 
--- a/resources/seccomp/x86_64-unknown-linux-musl.json
+++ b/resources/seccomp/x86_64-unknown-linux-musl.json
@@ -1255,6 +1255,18 @@
                 ]
             },
             {
+                "syscall": "ioctl",
+                "args": [
+                    {
+                        "index": 1,
+                        "type": "dword",
+                        "op": "eq",
+                        "val": 44717,
+                        "comment": "KVM_KVMCLOCK_CTRL. We call this after pausing vCPUs to avoid soft lockups in the guest."
+                    }
+                ]
+            },
+            {
                 "syscall": "sched_yield",
                 "comment": "Used by the rust standard library in std::sync::mpmc. Firecracker uses mpsc channels from this module for inter-thread communication"
             },
--- a/src/vmm/src/logger/metrics.rs
+++ b/src/vmm/src/logger/metrics.rs
@@ -779,6 +779,8 @@ pub struct VcpuMetrics {
     pub exit_mmio_write: SharedIncMetric,
     /// Number of errors during this VCPU's run.
     pub failures: SharedIncMetric,
+    /// Number of times that the `KVM_KVMCLOCK_CTRL` ioctl failed.
+    pub kvmclock_ctrl_fails: SharedIncMetric,
     /// Provides Min/max/sum for KVM exits handling input IO.
     pub exit_io_in_agg: LatencyAggregateMetrics,
     /// Provides Min/max/sum for KVM exits handling output IO.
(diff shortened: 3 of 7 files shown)


bchalios:
I created a test that should trigger the guest kernel to detect lockups. I will let the CI run once, so we can make sure that the test works on our CI and then reapply the commits that set the KVM_KVMCLOCK_CTRL bit when we pause vCPUs. These should make the failure go away.

bchalios:
The test does trigger the lockup: https://buildkite.com/firecracker/firecracker-pr/builds/11546.

bchalios:
I've pushed again Nikita's commits that add the call the `KVM_KVMCLOCK_CTRL` ioctl. Tests should be fixed now.

bchalios:
The last commit is not related with PR per se. However, it fixes an intermittent issue in the CI which I was hitting in this PR's pipelines runs and I was too lazy to open a separate PR.

tests/framework/utils.py
```diff
@@ -462,6 +465,11 @@ def get_process_pidfd(pid):
         pidfd = os.pidfd_open(pid)
     except ProcessLookupError:
         return None
+    except OSError as err:
+        if err.errno == errno.EINVAL:
+            return None
```
roypat, line 470: I think we do want to raise exceptions in this case, because it might be the only thing that tells us that the firecracker process died unexpectedly (e.g. when we call .kill() we expect there to be a process, and if it's already dead that would indicate a bug). Is the intermittent failure just because of race conditions? In that case, can we just create the pidfd before sending the SIGKILL to fix it? :o

resources/seccomp/x86_64-unknown-linux-musl.json
```diff
@@ -1254,6 +1254,18 @@
                     }
                 ]
             },
+            {
+                "syscall": "ioctl",
+                "args": [
```
kalyazin, line 1259: we can actually add a comment here on why we need it.

src/vmm/src/logger/metrics.rs
```diff
@@ -779,6 +779,8 @@ pub struct VcpuMetrics {
     pub exit_mmio_write: SharedIncMetric,
     /// Number of errors during this VCPU's run.
     pub failures: SharedIncMetric,
+    /// Number of times that the `KVM_KVMCLOCK_CTRL` ioctl failed.
```
kalyazin, line 782: My parser doesn't approve the commit message: "but we still it doesn't hurt tracking this."

CHANGELOG.md
```diff
@@ -14,6 +14,15 @@ and this project adheres to
...
   is negotiated, guest `virtio-net` driver can perform more efficient memory
   management which in turn improves RX and TX performance.
+- [#4460](https://github.com/firecracker-microvm/firecracker/pull/4460): Add a
+  call to
+  [`KVM_KVMCLOCK_CTRL`](https://docs.kernel.org/virt/kvm/api.html#kvm-kvmclock-ctrl)
+  after pausing vCPUs on x86_64 architectures. This ioctl sets a flag in the KVM
+  state of the vCPU indicating that it has been paused by the host userspace. In
+  guests that use kvmclock, the soft lockup watchdog checks this flag. If it is
+  set, it won't trigger the lockup condition. Calling the ioctl for guests that
+  don't use kvmclock will fail. These failures are not fatal; we log the failure
```
kalyazin, line 24: nit: s/;/:/

tests/framework/utils.py
roypat: synced offline, I got this wrong. The check is in `wait_process_termination`, e.g. at a point in time where we only wait for the process to go away. We don't assert it's still alive here, that happens in .kill()

resources/seccomp/x86_64-unknown-linux-musl.json
bchalios: Extended the comment with the use-case.

src/vmm/src/logger/metrics.rs
bchalios: I agree with your parser. (I think) I fixed it.

CHANGELOG.md
bchalios: I intentionally used a semicolon here, but I don't my English that much, so just changed this to a full stop.

kalyazin (commented):
Unfortunately I can't approve the change, apparently because the PR originated from me, but it does LGTM.

tests/integration_tests/functional/test_pause_resume.py
```diff
@@ -127,3 +130,38 @@ def test_pause_resume_preboot(uvm_nano):
...
+    # that we will pause the execution inside the kernel and cause a lock up. Setting KVM_CLOCK_CTRL
+    # bit that informs the guest we're pausing the vCPUs, should avoid that lock up.
+    microvm.ssh.check_output(
+        "timeout 60 sh -c 'while true; do ls -R /; done' > /dev/ttyS0 2>&1 < /dev/null &"
+    )
+
+    for _ in range(12):
+        microvm.api.vm.patch(state="Paused")
+        time.sleep(5)
+        microvm.api.vm.patch(state="Resumed")
```
roypat, line 162: nit:
```suggestion
        microvm.pause()
        time.sleep(5)
        microvm.resume()
```

roypat (approved):

PR #4467, opened 2024-02-20
fix: do not limit pipeline_perf.py to nonci tests in metrics mode

roypat:
When running A/B-Tests, the pipeline generated by pipeline_perf.py does not care about pytest markers, it will run all tests specified and do the A/B-comparision. However, when running the pipeline in metrics collection mode, it passes `-m nonci` to pytest, which causes it to filter out tests that do not have the `nonci` marker. Some of our A/B-Tests (such as memory overhead) indeed do not have this marker, leading to confusing errors (where everything works in A/B-mode, but if just collecting metrics pytest fails with "no tests collected"). This commit makes the behavior consistent, but also triggering non-nonci tests in metrics collection mode.

## Changes

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/.buildkite/pipeline_perf.py
+++ b/.buildkite/pipeline_perf.py
@@ -73,7 +73,8 @@ def build_group(test):
         devtool_opts += " --ab"
         pytest_opts = f" {ab_opts} run {REVISION_A} {REVISION_B} --test {test_path}"
     else:
-        pytest_opts += f" -m nonci {test_path}"
+        # Passing `-m ''` below instructs pytest to collect tests regardless of their markers (e.g. it will collect both tests marked as nonci, and tests without any markers).
+        pytest_opts += f" -m '' {test_path}"
     binary_dir = test.pop("binary_dir")
     return group(
         label=test.pop("label"),


.buildkite/pipeline_perf.py
```diff
@@ -73,7 +73,7 @@ def build_group(test):
         devtool_opts += " --ab"
         pytest_opts = f" {ab_opts} run {REVISION_A} {REVISION_B} --test {test_path}"
     else:
-        pytest_opts += f" -m nonci {test_path}"
+        pytest_opts += f" -m '' {test_path}"
```
pb8o, line 76: Doesn't this mean that nonci tests will not be collected?

roypat: Passing an empty string to `-m` (somewhat unintuitively) means "collect all tests, ignoring markers". We already use this trick in `tools/ab_test.py` to get it to run both nonci and non-nonci tests.

roypat: True, added a comment

pb8o (approved):

ShadowCurse (approved):

PR #4475, opened 2024-02-28
Add microvm helper for enabling console data

bchalios:
## Changes

Add a property in microvm class for printing console data. If console is disabled this will print an empty string.

## Reason

I found it useful, when debugging some guest kernel changes. Essentially, if we call that method before launching a microVM we can later print it's console output in the same way we can print Firecracker logs.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -371,6 +371,13 @@ class Microvm:
         return self.log_file.read_text()
 
     @property
+    def console_data(self):
+        """Return the output of microVM's console"""
+        if self.screen_log is None:
+            return None
+        return Path(self.screen_log).read_text(encoding="utf-8")
+
+    @property
     def state(self):
         """Get the InstanceInfo property and return the state field."""
         return self.api.describe.get().json()["state"]


pb8o (approved):

tests/framework/microvm.py
```diff
@@ -370,6 +370,13 @@ def log_data(self):
             return ""
         return self.log_file.read_text()
 
+    @property
+    def console_data(self):
+        """Return the output of microVM's console"""
+        if self.screen_log is None:
+            return ""
```
pb8o, line 377: nit: would it be worth to differentiate it from the case where it exists but the output is empty (for example if the boot is hanging).

zulinx86 (approved):

tests/framework/microvm.py
bchalios: It definitely would be nice to be able to do that, but I can't think of a way. Is there a way to check at this point if the microVM is actually running?

bchalios: oh, I see what you mean. Is there an Option type in python? :P

bchalios: Ok, I have it return `None` if the screen log is not there.

zulinx86 (approved):

pb8o (approved):

PR #4477, opened 2024-02-28
[doc] Fix getting started permission problems

XiangpengHao:
## Changes
Remove `sudo` when starting firecracker
...

## Reason
Don't run firecracker with root or the curl will fail, specifically, the following operations will fail due to permission error
```bash
curl -X PUT --unix-socket "${API_SOCKET}" ...
```

This is because when starting with sudo, the $API_SOCKET is created with root, which is not accessible from subsequent curl requests
...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


roypat:
Hi @XiangpengHao, 
Thank you for your contribution! 

For this getting started guide, we omit a lot of permission management, with the aim to be able to start a Firecracker VM quickly. When starting Firecracker without `sudo`, the user starting the Firecracker process needs to be able to access `/dev/kvm`, as well as permissions to create and manage TAP devices. We do mention at the end of the getting started guide, that if Firecracker is started with `sudo`privileges, then `curl` also needs to be executed as superuser, but I agree that this is fairly suboptimal as just a footnote. 

Could you maybe amend your PR so that instead of removing the `sudo` from the Firecracker invocation, we add it to all the `curl` invocations?

Thanks!

XiangpengHao:
Hi @roypat , thanks for the information, I followed the guide sequentially so didn't see the footnote :-)

I've updated the doc to use sudo consistently.

roypat:
> Hi @roypat , thanks for the information, I followed the guide sequentially so didn't see the footnote :-)
> 
> I've updated the doc to use sudo consistently.

Thanks! As a last thing, could you reformat your commit message to follow the 50/72 rule, and add your sign-off (e.g. via `git commit --amend -s`)? Both of these are required for your CI to pass

XiangpengHao:
> > Hi @roypat , thanks for the information, I followed the guide sequentially so didn't see the footnote :-)
> > I've updated the doc to use sudo consistently.
> 
> Thanks! As a last thing, could you reformat your commit message to follow the 50/72 rule, and add your sign-off (e.g. via `git commit --amend -s`)? Both of these are required for your CI to pass

Thanks for the guide! I've updated the commit message!

roypat (approved):

docs/getting-started.md
```diff
@@ -192,7 +192,7 @@ LOGFILE="./firecracker.log"
 touch $LOGFILE
 
 # Set log file
-curl -X PUT --unix-socket "${API_SOCKET}" \
+sudo curl -X PUT --unix-socket "${API_SOCKET}" \
```
pb8o, line 195: Alternatively, could we change the permissions of the API_SOCKET? i.e. `chmod a+rw $API_SOCKET`?

XiangpengHao: it could work, `$API_SOCKET` requires root because we started firecracker with root.

For consistency, I think we can either use sudo everywhere (so it's easy to setup etc), or we don't use sudo at all, which requires user to setup the access to `/dev/kvm`. 

what do you think?

pb8o: I think the `sudo` starting Firecracker is OK, as it does need some special permissions. But the ones for curl are not strictly required, as we could just allow writing to the socket.

Another option is to just run everything after `sudo su`, but it's equivalent to running all with `sudo`.

pb8o (approved):

PR #4486, opened 2024-03-05
feat(metrics): add IO queue len and backend latency metrics

kalyazin:
## Changes

This adds the following accumulated metrics:
- block IO: read and write backend latencies (`read_agg`, `write_agg`)
- block IO: length of the queue backlog (`remaining_reqs_count`)
- net: tap write backend latencies (`tap_agg`)
- net: length of the Tx queue backlog (`tx_remaining_reqs_count`)

The latency metrics show how long the VMM thread spends waiting for the requests being processed by the block/net backend. The queue backlog metrics show how many requests are still waiting to be processed by the VMM thread. They are updated during processing of every request, so dividing it by the total number of requests would show the average queue length per request processing.

## Reason

We would like to have more visibility on how block and network backends are saturated.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,13 @@ and this project adheres to
   be deduced from the available emitted metrics.
 - [#4360](https://github.com/firecracker-microvm/firecracker/pull/4360): Added
   dev-preview support for backing a VM's guest memory by 2M hugetlbfs pages.
-  Please see the [documentation](docs/hugepages.md) for more information.
+  Please see the [documentation](docs/hugepages.md) for more information
+- [#4486](https://github.com/firecracker-microvm/firecracker/pull/4486): Added
+  block and net device metrics for file/tap access latencies and queue backlog
+  lengths, which can be used to analyse saturation of the Firecracker VMM thread
+  and underlying layers. Queue backlog length metrics are flushed periodically.
+  They can be used to esimtate an average queue length by request by dividing
+  its value by the number of requests served.
 
 ### Changed
 
--- a/src/vmm/src/devices/virtio/block/virtio/metrics.rs
+++ b/src/vmm/src/devices/virtio/block/virtio/metrics.rs
@@ -187,6 +187,8 @@ pub struct BlockDeviceMetrics {
     /// Number of virtio events throttled because of the IO engine.
     /// This happens when the io_uring submission queue is full.
     pub io_engine_throttled_events: SharedIncMetric,
+    /// Number of remaining requests in the queue.
+    pub remaining_reqs_count: SharedIncMetric,
 }
 
 impl BlockDeviceMetrics {
@@ -231,6 +233,8 @@ impl BlockDeviceMetrics {
             .add(other.rate_limiter_throttled_events.fetch_diff());
         self.io_engine_throttled_events
             .add(other.io_engine_throttled_events.fetch_diff());
+        self.remaining_reqs_count
+            .add(other.remaining_reqs_count.fetch_diff());
     }
 }
 
--- a/src/vmm/src/devices/virtio/block/virtio/metrics.rs
+++ b/src/vmm/src/devices/virtio/block/virtio/metrics.rs
@@ -83,7 +83,7 @@ use std::sync::{Arc, RwLock};
 use serde::ser::SerializeMap;
 use serde::{Serialize, Serializer};
 
-use crate::logger::{IncMetric, SharedIncMetric};
+use crate::logger::{IncMetric, LatencyAggregateMetrics, SharedIncMetric};
 
 /// map of block drive id and metrics
 /// this should be protected by a lock before accessing.
@@ -105,7 +105,7 @@ impl BlockMetricsPerDevice {
                 .write()
                 .unwrap()
                 .metrics
-                .insert(drive_id.clone(), Arc::new(BlockDeviceMetrics::default()));
+                .insert(drive_id.clone(), Arc::new(BlockDeviceMetrics::new()));
         }
         METRICS
             .read()
@@ -178,6 +178,10 @@ pub struct BlockDeviceMetrics {
     pub read_count: SharedIncMetric,
     /// Number of successful write operations.
     pub write_count: SharedIncMetric,
+    /// Duration of all read operations.
+    pub read_agg: LatencyAggregateMetrics,
+    /// Duration of all write operations.
+    pub write_agg: LatencyAggregateMetrics,
     /// Number of rate limiter throttling events.
     pub rate_limiter_throttled_events: SharedIncMetric,
     /// Number of virtio events throttled because of the IO engine.
@@ -186,6 +190,15 @@ pub struct BlockDeviceMetrics {
 }
 
 impl BlockDeviceMetrics {
+    /// Const default construction.
+    pub fn new() -> Self {
+        Self {
+            read_agg: LatencyAggregateMetrics::new(),
+            write_agg: LatencyAggregateMetrics::new(),
+            ..Default::default()
+        }
+    }
+
     /// block metrics are SharedIncMetric where the diff of current vs
     /// old is serialized i.e. serialize_u64(current-old).
     /// So to have the aggregate serialized in same way we need to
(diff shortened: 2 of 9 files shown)


bchalios (commented):
Generally LGTM.

I would like it if there was more information in the commit messages about what the commit does and mainly why. Right now, the descriptions make sense reading the PR description, but by themselves are a bit cryptic.

Other than that, right now, the new latency metrics only measure the backend. I think it would make sense to also measure the time we spend in Firecracker to handle each request. Ideally, this should give us a sense of the "overhead" of Firecracker's processing and could point to low-hanging fruit optimizations in the short term, as well as protect us from doing stupid things in the long term. WDYT?

CHANGELOG.md
```diff
@@ -17,7 +17,11 @@ and this project adheres to
   be deduced from the available emitted metrics.
 - [#4360](https://github.com/firecracker-microvm/firecracker/pull/4360): Added
   dev-preview support for backing a VM's guest memory by 2M hugetlbfs pages.
-  Please see the [documentation](docs/hugepages.md) for more information.
+  Please see the [documentation](docs/hugepages.md) for more information
+- [#4486](https://github.com/firecracker-microvm/firecracker/pull/4486): Added
+  sync block and net device metrics for backend latencies and queue backlog
+  lengths, which can be used to analyse saturation of the Firecracker VMM thread
+  and device backends.
```
ShadowCurse, line 24: nit: for me the `backend` word is associated with `vhost-user-backed`. Maybe a bit rephrased version like this is better:
`Added metrics for processing latency and queue length in (sync)block and net devices. These metrics can be used to analyze saturation of the Firecracker VMM thread when these devices are in use.`

src/vmm/src/devices/virtio/block/virtio/metrics.rs
```diff
@@ -178,11 +178,17 @@ pub struct BlockDeviceMetrics {
     pub read_count: SharedIncMetric,
     /// Number of successful write operations.
     pub write_count: SharedIncMetric,
+    /// Duration of all read operations.
```
sudanl0, line 181: If we use [LatencyAggregateMetrics](https://github.com/sudanl0/firecracker/blob/be25214c564812babbc614ad2e4077afd0bba4a3/src/vmm/src/logger/metrics.rs#L731) instead of SharedIncMetrics we can capture min/max and sum of the read_duration_us. Do you think there is any advantage of exporting main/max in addition to the sum that the code currently exports?

src/vmm/src/devices/virtio/block/virtio/request.rs
```diff
@@ -371,14 +371,35 @@ impl Request {
         block_metrics: &BlockDeviceMetrics,
     ) -> ProcessingResult {
         let pending = self.to_pending_request(desc_idx);
+        let start_request_us = utils::time::get_time_us(utils::time::ClockType::Monotonic);
         let res = match self.r#type {
             RequestType::In => {
-                disk.file_engine
-                    .read(self.offset(), mem, self.data_addr, self.data_len, pending)
+                let res = disk.file_engine.read(
```
sudanl0, line 377: if we implement record_latency_metrics  for ShareIncMetrics like we did for LatencyAggregateMetrics [here](https://github.com/sudanl0/firecracker/blob/be25214c564812babbc614ad2e4077afd0bba4a3/src/vmm/src/logger/metrics.rs#L755) we can avoid the new function update_inc_metric_with_elapsed_time.
We also don't need to remember to call the 2nd lines since drop will take care of that.
Code would look something like below:
```
let pending = self.to_pending_request(desc_idx);
        let res = match self.r#type {
            RequestType::In => {
                let _metric = block_metrics.read_duration_us.record_latency_metrics(); >> this takes care of updating the read_duration_us metric as part of dropping _metrics
                disk.file_engine
                    .read(self.offset(), mem, self.data_addr, self.data_len, pending)
            }
            RequestType::Out => {
                let _metric = block_metrics.write_duration_us.record_latency_metrics(); >> this takes care of updating the write_duration_us metric as part of dropping _metrics
                disk.file_engine
                    .write(self.offset(), mem, self.data_addr, self.data_len, pending)
            }
```

src/vmm/src/logger/mod.rs
```diff
@@ -44,3 +44,11 @@ pub fn update_metric_with_elapsed_time(metric: &SharedStoreMetric, start_time_us
     metric.store(delta_us);
     delta_us
 }
+
+/// Helper function for updating the value of an inc metric with elapsed time since some time in a
+/// past.
+pub fn update_inc_metric_with_elapsed_time(metric: &SharedIncMetric, start_time_us: u64) -> u64 {
```
sudanl0, line 50: If we don't want min/max for the metrics then I would prefer to modify [LatencyMetricsRecorder](https://github.com/sudanl0/firecracker/blob/be25214c564812babbc614ad2e4077afd0bba4a3/src/vmm/src/logger/metrics.rs#L695) to use generic data types so that it can work for LatencyAggregateMetrics and SharedIncMetrics.

sudanl0:
>The queue backlog metrics show how many requests are still waiting to be processed by the VMM thread. They are updated during processing of every request, 
remaining_reqs_count and tx_remaining_reqs_count could be bit confusing since they are reset every time we flush. So technically there could be requests still pending in the queue but metrics won't report them because they have already been reported. I think its fine the way they are right now but we should explicitly call out that these 2 metrics being 0 does not mean that there are no requests in queue. This could be confusing if we are debugging a crash where the last metrics showed that there are no requests but there actually were.

>so dividing it by the total number of requests would show the average queue length per request processing.
Can we document it in CHANGELOG or in code saying divide by the metric `queue_event_count` to get avg queue lenght?

src/vmm/src/devices/virtio/block/virtio/metrics.rs
kalyazin: yes, LatencyAggregateMetrics looks like a good fit to that, thanks for proposing, made use of them. The only thing is it was not easy to use it with tap read due to borrowing rules, so I decided to skip that metric for now, especially I only add queue len for Tx, not for Rx.

CHANGELOG.md
kalyazin: rephrased to avoid work backend.

src/vmm/src/devices/virtio/block/virtio/request.rs
kalyazin: made use of LatencyAggregateMetrics, thanks.

src/vmm/src/logger/mod.rs
kalyazin: removed.

kalyazin:
> > The queue backlog metrics show how many requests are still waiting to be processed by the VMM thread. They are updated during processing of every request,
> 
> remaining_reqs_count and tx_remaining_reqs_count could be bit confusing since they are reset every time we flush. So technically there could be requests still pending in the queue but metrics won't report them because they have already been reported. I think its fine the way they are right now but we should explicitly call out that these 2 metrics being 0 does not mean that there are no requests in queue. This could be confusing if we are debugging a crash where the last metrics showed that there are no requests but there actually were.
> 
> > so dividing it by the total number of requests would show the average queue length per request processing.
> 
> Can we document it in CHANGELOG or in code saying divide by the metric `queue_event_count` to get avg queue lenght? Maybe worth adding a small section in https://github.com/firecracker-microvm/firecracker/blob/main/docs/metrics.md to highlight how these metrics can help with analysing saturation.

added a note in the changelog, but not in doc for now.

kalyazin:
> Generally LGTM.
> 
> I would like it if there was more information in the commit messages about what the commit does and mainly why. Right now, the descriptions make sense reading the PR description, but by themselves are a bit cryptic.
> 
> Other than that, right now, the new latency metrics only measure the backend. I think it would make sense to also measure the time we spend in Firecracker to handle each request. Ideally, this should give us a sense of the "overhead" of Firecracker's processing and could point to low-hanging fruit optimizations in the short term, as well as protect us from doing stupid things in the long term. WDYT?

Added more info in commit messages. I agree, it would be useful to have a pie chart of VMM time split by idling, being busy and blocked on lower layers. That could be the next step.

bchalios (approved):

PR #4487, opened 2024-03-06
Support for VMGenID device on x86 microVMs

bchalios:
## Changes

Add support for Virtual Machine Generation ID for x86 microVMs. This also adds support for the ACPI Generic Event Device which routes interrupts for ACPI devices. VMGenID provides to the guest a 16-byte cryptographically random number which changes every time the microVM resumes from a snapshot.

## Reason

VMGenID is used by the Linux kernel to re-seed its internal PRNG whenever a VM resumes from a snapshot. With Linux 6.8, Linux VMGenID driver will also emit a uevent to guest user-space every time the generation ID changes (i.e. the VM resumes from a snapshot).

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


bchalios:
This is still a draft, since it depends on #4428

tests/integration_tests/security/test_vulnerabilities.py
```diff
@@ -59,6 +59,7 @@ def configure_microvm(
     else:
         microvm = factory.build(kernel, rootfs)
 
+    microvm.help.enable_console()
```
ShadowCurse, line 62: is this needed?

docs/snapshotting/snapshot-support.md
```diff
@@ -571,15 +575,32 @@ we also consider microVM A insecure if it resumes execution.
...
-We've started a discussion on how the Linux operating system might securely
-handle being snapshotted [here](https://lkml.org/lkml/2020/10/16/629).
+[Virtual Machine Generation Identifier](https://learn.microsoft.com/en-us/windows/win32/hyperv_v2/virtual-machine-generation-identifier)
+(VMGenID) is a virtual device that allows VM guests to detect when they have
+resumed from a snapshot. It works by exposing a cryptographically random
+16-bytes identifier to the guest. The VMM ensures that the value of the
+indentifier changes every time the VM a time shift happens in the lifecycle of
+the VM, e.g. when it resumes from a snapshot.
+
+Linux supports VMGenID since version 6.1. When Linux detects a change in the
```
ShadowCurse, line 585: Did you mention that VmGenId is supported from 5.18?

docs/snapshotting/snapshot-support.md
```diff
@@ -571,15 +575,32 @@ we also consider microVM A insecure if it resumes execution.
...
+identifier, it uses its value to reseed its internal PRNG. Moreover,
+[since version 6.8](https://lkml.org/lkml/2023/5/31/414) Linux VMGenID driver
+also emits to userspace a uevent. User space processes can monitor this uevent
+for detecting snapshot resume events.
+
+Firecracker supports VMGenID device on x86 platforms. Firecracker will always
+enable the device. During snapshot resume, Firecracker will update the 16-byte
+generation ID and inject a notification in the guest before resuming its vCPUs.
+
+As a result, guests that run Linux versions >= 6.1 will re-seed their in-kernel
```
ShadowCurse, line 595: Again I think this should be 5.18

docs/snapshotting/random-for-clones.md
```diff
@@ -41,6 +43,8 @@ and there’s also an input pool which gathers entropy from various sources
 available on the system, and is used to feed into or seed the other two
 components. A very detailed description is available [here][3].
 
+### Linux kernels from 4.8 until 5.17 (included)
```
ShadowCurse, line 46: Would be nice to mention what happens with VmGenId device on older kernels.

src/vmm/src/builder.rs
```diff
@@ -498,7 +526,7 @@ pub fn build_microvm_from_snapshot(
 
     // Restore devices states.
     let mmio_ctor_args = MMIODevManagerConstructorArgs {
-        mem: guest_memory,
+        mem: guest_memory.clone(),
```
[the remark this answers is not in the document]
bchalios: It could, but I think there's a lot of things that need to change all over the `persist.rs` files of various components. I wouldn't want to do this in this particular PR.

src/vmm/src/builder.rs
```diff
@@ -520,6 +560,13 @@ pub fn build_microvm_from_snapshot(
             .clone(),
     )?;
 
+    // Inject the notification to VMGenID that we have resumed from a snapshot
+    // This needs to happen after we have started the vCPUs so that the interrupt is injected
+    // correctly, and before we resume them, so that we minimize the time between vCPUs resuming
```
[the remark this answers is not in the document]
bchalios: Hmmm, I just tried that and didn't work at the time. Looking at the code of `start_vcpus` it does work with KVM state but not sure if it's related. I can try it out to double-check and adjust the comment if it is incorrect.

src/vmm/src/device_manager/acpi.rs
```diff
@@ -0,0 +1,123 @@
...
+    ///
+    /// This will create the device and register the interrupt line with KVM.
+    pub fn register_vmgenid(
+        &mut self,
+        gsi: u32,
+        guest_address: GuestAddress,
+        mem: &GuestMemoryMmap,
+        vm_fd: &VmFd,
+    ) -> Result<(), ACPIDeviceManagerError> {
+        let vmgenid = VmGenId::new(guest_address, gsi, mem)?;
```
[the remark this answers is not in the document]
bchalios: We don't do that for other devices. With MMIO it's the device manager that allocates resources for the particular device and then does the registration. What if we rethink this logic (if there is a strong reason for that) when we do the device manager refactoring

tests/integration_tests/security/test_vulnerabilities.py
bchalios: Nope, but it is extremely useful for debugging.

docs/snapshotting/snapshot-support.md
bchalios: You are right. I initially thought it was 6.1, later found it was 5.18 and apparently missed substituting in some places.

docs/snapshotting/random-for-clones.md
bchalios: will do

sudanl0 (commented):
still going through the code but here are some initial comments

src/vmm/src/builder.rs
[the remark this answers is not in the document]
ShadowCurse: From what I can see, only changes needed are:
```rust
  // in MMIODevManagerConstructorArgs
  -pub mem: GuestMemoryMmap,
  +pub mem: &GuestMemoryMmap,

  // in MMIODeviceManager::restore
  -let mem = &constructor_args.mem;
  +let mem = constructor_args.mem;
```

src/vmm/src/device_manager/acpi.rs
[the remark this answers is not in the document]
ShadowCurse: For other devices we do smth similar: we create a device first and then we put them in the device manager. The only reason I made `VmGenId` to allocate it's resources because there is no `MMIODeviceInfo` alternative for `acpi` devices. The `VmGenId` contains this info in itself, so I think it make sense for it to know what it needs.

tests/integration_tests/security/test_vulnerabilities.py
ShadowCurse: Sorry, what I meant: do we still need it? I assume it is left from when you were debugging, but not it can be removed.

src/vmm/src/builder.rs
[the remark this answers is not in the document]
bchalios: oh, you are (almost) right! I thought it would be more. Will do.

src/vmm/src/device_manager/acpi.rs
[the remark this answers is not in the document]
bchalios: We do have the equivalent of `MMIODeviceInfo` for ACPI devices, it's the memory address and IRQ. For MMIO devices, we do allocate the memory and IRQ in the equivalent device manager. IMO, VMGenID needs just to know that it has an interrupt and an address, it doesn't need to know how to allocate them.

CHANGELOG.md
```diff
@@ -17,6 +17,15 @@ and this project adheres to
   without MPTable support. Please see our
   [kernel policy documentation](docs/kernel-policy.md) for more information
   regarding relevant kernel configurations.
+- [#4487](https://github.com/firecracker-microvm/firecracker/pull/4487): Added
+  support for the Virtual Machine Generation Identifier (VMGenID) device on
+  x86_64 platforms. VMGenID is a virtual device that allows VMMs to notify
+  guests when they are resumed from a snapshot. Linux includes VMGenID support
+  since version 5.18. It uses notifications from the device to reseed its
+  internal CSPRNG. Please refer to
```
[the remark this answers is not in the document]
bchalios: I would rather have in the CHANGELOG what is the effect of using VMGenID in the guest kernel.

docs/snapshotting/snapshot-support.md
```diff
@@ -605,6 +626,16 @@ section 5.10.6.6 Device Events.
 Firecracker handles sending the `reset` event to the vsock driver, thus the
 customers are no longer responsible for closing active connections.
 
+## VMGenID device limitation
```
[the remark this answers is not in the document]
bchalios: > this is still a breaking change for customers which makes things complicated

Although I agree in principle, I think it is not really practical to take a snapshot before the guest kernel even initializes the CPUs. Snapshotting is a way to speed-up the boot process. It seems counter intuitive to take a snapshot so early (the first few milliseconds) of the boot process. Even if someone wanted to do this, there's not a practical way to know that the VM is in this state.

> there is no way around this?

I don't think so. This depends on which point in guest kernel initialization we are. We don't have any visibility in that.

> we have all the details of what "proper interrupt handling" is and exactly how long should customer wait before taking the snapshot?

The interesting moment in time is when Linux initializes its CPUs. After that, everything should be working fine. I think for all practical reasons, users would want to snapshot the microVM after the guest has fully booted. Probably, even later than that; after they have initialized their own logic inside the microVM.

CHANGELOG.md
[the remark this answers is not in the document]
sudanl0: >`VMGenID is a virtual device that allows VMMs to notify
  guests when they are resumed from a snapshot.` 

What I meant was that I think above line is enough to explain what vmgenid does in the guest. CSPRNG goes in bit more detail which may not be needed here.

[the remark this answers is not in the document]
bchalios: I added it because the reseeding of CSPRNG is the main reason why we want VMGenID support

docs/snapshotting/snapshot-support.md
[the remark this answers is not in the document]
sudanl0: Yeah I agree that it doesn't make sense to take snapshot that early but just wanted to confirm how long we should ask customers to wait before taking a snapshot and the current wording looks fine for now.
I would however suggest to update the below line in the overview section:
`The original guest which the snapshot is created from, should see no side effects from this process (other than the latency introduced by the snapshot creation process).`
to something like 
`The original **linux booted** guest which the snapshot is created from, should see no side effects from this process  (other than the latency introduced by the snapshot creation process).`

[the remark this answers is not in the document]
bchalios: The original guest should see no side effects regardless of whether linux has booted or not.

src/vmm/src/devices/acpi/vmgenid.rs
```diff
@@ -0,0 +1,115 @@
...
+            gsi,
+        })
+    }
+
+    // Create a 16-bytes random number
+    fn make_genid() -> Result<u128, RandError> {
+        let mut gen_id_bytes = [0u8; 16];
+        rand::fill(&mut gen_id_bytes)
+            .inspect_err(|err| error!("vmgenid: could not create new generation ID: {err}"))?;
+        Ok(u128::from_le_bytes(gen_id_bytes))
```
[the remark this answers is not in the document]
bchalios: Hmmm, I don't see the `gen` module in aws_lc_rs: https://docs.rs/aws-lc-rs/latest/aws_lc_rs/rand/index.html

src/vmm/src/device_manager/acpi.rs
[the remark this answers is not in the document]
ShadowCurse: In this case maybe there should be a `ACPIDeviceInfo` struct that has the resources and `VmGenId` can just contain this struct?
```rust
let acpi_info = vmm.acpi_device_manager.alloc_resource(&mut vmm.resource_allocator);
let vmgenid = VmGenId::new(&vmm.guest_memory, acpi_info);
vmm.acpi_device_manager.attach_vmgenid(&vmm.vm, vmgenid);

// in restore case the `acpi_info` can be constructed from pieces.
let acpi_info = AcpiInfo::new(irq, mem_region);
```
Just to rephrase my overall concern here: the way devices work is that they are created outside of device managers and only are inserted into device managers when it is needed. I don't mind device manager to allocate resources(irqs, mem regions) when device is being inserted into the device manager, but I don't like when device manager actually created a device and then adds it into itself in one function call. If all acpi devices will have same resources requirements (same amout if irqs and memory region size) then maybe smth link `AcpiInfo::new(&mut vmm.resource_allocator)` will work nicely. This way the interface for `ACPIDeviceManager` can be more generic like `ACPIDeviceManager::add_device<D: AcpiDevice>(device: D, acpi_info: AcpiInfo)`. 
Anyway I think I am getting into the depth of how devices should work. Let me open an issue so we can gather all our thoughts there.
For now I think what you have here does it's job.

src/vmm/src/builder.rs
```diff
@@ -520,6 +560,12 @@ pub fn build_microvm_from_snapshot(
             .clone(),
     )?;
 
+    // Inject the notification to VMGenID that we have resumed from a snapshot.
+    // This needs to happen before we resume vCPUs, so that we minimize the time between vCPUs
+    // resuming and notification being handled by the driver.
+    #[cfg(target_arch = "x86_64")]
+    vmm.acpi_device_manager.notify_vmgenid()?;
```
ShadowCurse, line 567: 1. Comment still here
2. If this does not need to be after `start_vcpus` can you move up to the acpi section just before `start_vcpus`?

docs/snapshotting/snapshot-support.md
```diff
@@ -571,15 +575,32 @@ we also consider microVM A insecure if it resumes execution.
...
+identifier, it uses its value to reseed its internal PRNG. Moreover,
+[since version 6.8](https://lkml.org/lkml/2023/5/31/414) Linux VMGenID driver
+also emits to userspace a uevent. User space processes can monitor this uevent
+for detecting snapshot resume events.
+
+Firecracker supports VMGenID device on x86 platforms. Firecracker will always
+enable the device. During snapshot resume, Firecracker will update the 16-byte
+generation ID and inject a notification in the guest before resuming its vCPUs.
+
+As a result, guests that run Linux versions >= will re-seed their in-kernel PRNG
```
ShadowCurse, line 595: `>=` than what?

src/vmm/src/device_manager/acpi.rs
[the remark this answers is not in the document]
bchalios: I don't see much value in creating a struct to store two variables inside. In MMIO this makes sense because later we store that struct somewhere (`MMIODeviceManager::id_to_dev_info`).

But I think you convinced me to do actually perform the allocation inside `Vmgenid`, because not all ACPI devices will need to perform the same allocations. I will do it like that.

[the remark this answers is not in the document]
sudanl0: Don't want to hijack the conversation but my 2 cents on the device info:
`ACPIDeviceManager->VmGenId->guest_address` and `ACPIDeviceManager->VmGenId->gsi` though provide device info, they are still specific to vmgenid and not the acpi device manager itself which in the future can have more devices.
Some context on the next problem: vmgenid will also be used with acpi device manager but with devicetree and not with acpi support. To create devicetree nodes we will need vmgenid info so we need to pass the vmgenid addr/gsi form builder.rs all the way to fdt.rs.
Problem:
1. With current approach if we pass vmgenid to fdt.rs we pass more than just addr/gsi which fdt.rs doesn't need to know.
2. If we have more devices in the future which follows same approach of using ACPI device manager for devicetree, the hashmap `pub(crate) id_to_dev_info: HashMap<(DeviceType, String), MMIODeviceInfo>,` in MMIODeviceManager looks more reasonable for passing info to fdt than passing individual device info.
The 2nd issue, though revolves around hypothetical devices, shouldn't be hard to avoid with an ACPIDeviceInfo.

docs/snapshotting/snapshot-support.md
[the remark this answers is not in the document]
sudanl0: ah sorry I misread, I am just trying to see if the 1 line in vmgenid limitation is enough to highlight the behaviour. nvm

src/vmm/src/devices/acpi/vmgenid.rs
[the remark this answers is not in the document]
roypat: Ah, my bad, I forgot we're using aws_lc_rs. I was looking at the docs for the `rand` crate.

src/vmm/src/builder.rs
bchalios: The comment about injecting the notification before resuming the vCPUs is valid. Before or after the `start_vcpus` doesn't make any difference, so why move it?

docs/snapshotting/snapshot-support.md
bchalios: Fixed

src/vmm/src/device_manager/persist.rs
```diff
@@ -226,6 +230,70 @@ impl fmt::Debug for MMIODevManagerConstructorArgs<'_> {
...
+
+#[cfg(target_arch = "x86_64")]
+#[derive(Default, Debug, Clone, Serialize, Deserialize)]
+pub struct ACPIDeviceManagerState {
+    vmgenid: Option<VMGenIDState>,
+}
+
+#[cfg(target_arch = "x86_64")]
+pub struct ACPIDeviceManagerConstructorArgs<'a> {
+    pub mem: GuestMemoryMmap,
```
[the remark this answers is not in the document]
bchalios: done!

bchalios:
Changes from v2 to v3:

* Re-worked the logic to create VMGenID devices. Now, the resource allocation happens inside the `VmGenId` "constructor". Like that, we hide the logic of allocating resources which are specific to VMGenID devices.
* Re-worked the logic for (de)serializing the VMGenID state, by moving VMGenID specific stuff under the VMGenID logic (similar to what we do with VirtIO devices).

src/vmm/src/builder.rs
ShadowCurse: Sorry, did not read comment carefully enough.
The move is mainly to combine 2 `#[cfg(target_arch = "x86_64")]` sections into 1 that deals with all acpi related things.

bchalios: Ok, that makes sense. Moved it.

src/vmm/src/lib.rs
```diff
@@ -522,13 +531,17 @@ impl Vmm {
...
+        let acpi_dev_state = self.acpi_device_manager.save();
 
         Ok(MicrovmState {
             vm_info: vm_info.clone(),
             memory_state,
             vm_state,
             vcpu_states,
             device_states,
+            #[cfg(target_arch = "x86_64")]
+            acpi_dev_state,
```
ShadowCurse, line 544: As I recall when we change the state of the snapshot we should increase the snapshot version.

tests/framework/artifacts.py
```diff
@@ -46,6 +46,11 @@ def select_supported_kernels():
     return supported_kernels
 
 
+def select_all_kernels():
+    """Select all guest kernels available in our CI artifacts folder"""
+    return [r"vmlinux-\d.\d+.\d+", r"vmlinux-5.10-no-sve-bin"]
```
ShadowCurse, line 51: I think this should be a part of `kernels_unfiltered`.

ShadowCurse (commented):
Overall PR seems fine to me. The only thing to make it a bit better would be to split first commit into several commits smth like:
- `update mmio_restore_args`
- `add acpi_device_manager`
- `add vmgenid`
- ...

src/vmm/src/lib.rs
bchalios: good catch!

bchalios:
@ShadowCurse regarding splitting the first commit in parts, I believe that splitting out the MMIO restore argument changes it makes sense. I don't see much value in splitting out the ACPI device manager though. Without the VMGenID changes, this would be just adding an empty struct (without any implementation at all). I would rather keep them together.

src/vmm/src/lib.rs
ShadowCurse: We should think about some sort of test for this. Maybe A/B test with "before PR" binary and "after PR" binary where we just try to restore snap from one in another if reported snap versions are same?
Obviously this should not be a part of this PR.

ShadowCurse (dismissed):

src/vmm/src/devices/acpi/vmgenid.rs
```diff
@@ -0,0 +1,179 @@
...
+///
+/// The device specification can be found here: https://go.microsoft.com/fwlink/?LinkId=260709
+#[derive(Debug)]
+pub struct VmGenId {
+    /// Current generation ID of guest VM
+    pub gen_id: u128,
+    /// Interrupt line for notifying the device about generation ID changes
+    pub interrupt_evt: EventFdTrigger,
+    /// Guest physical address where VMGenID data lives.
+    pub guest_address: GuestAddress,
```
sudanl0, line 33: instead of guest_address and gsi can we have VMGenIDState here instead?
This will help to pass vmgenid.VMGenIDState from builder.rs for creating vmgenid devicetree node

bchalios: Why can't we just pass  guest_address and gsi, or just a reference to the VMGenID device?

bchalios: I still don't understand why we need a separate object for this, we could just do:

```Rust
        crate::arch::aarch64::configure_system(
            &vmm.guest_memory,
            cmdline,
            vcpu_mpidr,
            vmm.mmio_device_manager.get_device_info(),
            vmm.acpi_manager.vmgenid.as_ref(),
            vmm.vm.get_irqchip(),
            initrd,
        )
```

so, inside the `configure_system` we have an `Option<VmGenID>`, where we can get `gsi` and `guest_addr` from. What problem does a `VMGenIDState` type solve?

sudanl0: Thats true, in the example [here](https://github.com/firecracker-microvm/firecracker/commit/f382e09a9b1e4453d2b231232d89bcb85296ac87#diff-a4f542f22c2bab2421c33bb265388e32dddbf0e7bd5d49eb6a4d54c2d17005b4R280) I have done the same. But my point was that along with gsi and guest_addr we are also sending gen_id and interrupt_evt which the file doesn't need. I just feel if the information is not needed by the fdt.rs then it shouldn't receive it.

bchalios: > I just feel if the information is not needed by the fdt.rs then it shouldn't receive it.

What is your concern with that? We are passing a reference to the enclosing type (so we do not have time overhead).

IMHO, it looks an unnecessary complication create a new type for the sole purpose of passing it to the FDT generation code.

ShadowCurse: I think this discussion can be held here: #4542. As of right now, we don't pass acpi info to the fdt. And when we will need to do it, we can think about necessary changes.

bchalios: I agree with Egor. I find it atypical to create a new type just to facilitate "hiding" info for the "potential" case of some coding error. If we really think that's a problem we could just make the fields of VMGenID we are worried about private. But again, I personally do not see any risk about it.

sudanl0: discussed offline on the next steps. The VMGenIDState is used for serializing the state during snapshot. Moving it inside vmgenid doesn't provide much benefit. That being said, if we want to protect other members of vmgenid from being passed to fdt we can move them to private but the work is out of scope for this PR.

tests/conftest.py
```diff
@@ -352,6 +352,11 @@ def rootfs_fxt(request, record_property):
 guest_kernel_linux_5_10 = pytest.fixture(
     guest_kernel_fxt, params=kernel_params("vmlinux-5.10*")
 )
+# Use the unfiltered selector, since we don't officially support 6.1 yet.
```
sudanl0, line 355: nit: the "yet" sounds to me like we should add a TODO to use kernel_params when we officially start supporting 6.1.

sudanl0 (approved):
LGTM, I would however like to get a 2nd opinion on the `kernels_unfiltered` from @pb8o

tests/conftest.py
bchalios: added the TODO

ShadowCurse (approved):

sudanl0 (approved):

PR #4491, opened 2024-03-08
Fixes for 6.5 kernel

ShadowCurse:
## Changes
Couple fixes for the CI for both aarch64 and x86. 

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


tests/integration_tests/functional/test_cpu_template_helper.py
```diff
@@ -446,10 +449,21 @@ def test_consecutive_cpu_config_consistency(uvm_plain, cpu_template_helper, tmp_
             "msr_modifiers": [],
         }
     elif PLATFORM == "aarch64":
+        # 0x603000000013df01 -> CNTPCT_EL0
+        ignore_registers = ["0x603000000013df01"]
```
zulinx86, line 453: How about excluding it from CPU config?
https://github.com/firecracker-microvm/firecracker/blob/5d3a13d225b452f36e23f5a616cc190f3b30a5cf/src/vmm/src/vstate/vcpu/aarch64.rs#L239

ShadowCurse: I really don't want FC to have any filters for cpu registers. If we will, I think It will become an issue by itself over time. Same as cpu normalization for x86.

zulinx86: We already have it both on x86 and aarch64. Otherwise, fingerprint verification will keep false alarming and users need to filter them as you do in the test. I guess moving the filter from Firecracker to CPU template helper tool is outside of this PR. So let's filter it in Firecracker as of now.

ShadowCurse: This is exactly what I don't want for aarch64 is to repeat filtering as it is done in x86. Considering that this register does not cause any problems by itself (snapshot tests pass with it not filtered out), I think it is our tests should have a knowledge what to test. Also fingerprint verification will be alarming in both cases when new register suddenly will become available. I can add it to the `cpu-template-helper-tool` if you want.

zulinx86: > Considering that this register does not cause any problems by itself (snapshot tests pass with it not filtered out), I think it is our tests which should have a knowledge of what to test and what not to test.

I don't think so. I believe the CPU template helper should filter them out for users.

> I can add it to the cpu-template-helper-tool if you want.

I'll take the task because I'm working on another refactoring stuff.
OK. Please leave this PR as it is. I'll remove the filters in test in the refactor.

zulinx86 (dismissed):

zulinx86 (commented):
https://github.com/firecracker-microvm/firecracker/blob/main/docs/cpu_templates/cpu-template-helper.md#appendix

Could you please add them to the ignore list of the doc.

tests/integration_tests/functional/test_cpu_template_helper.py
ShadowCurse: > I don't think so. I believe the CPU template helper should filter them out for users. As maintainer, we should decide which registers should be cared by users in the fingerprint verification process. Otherwise, the helper tool is not helpful despite its name.

I am totally fine with template helper filtering out things. Just not FC itself.

ShadowCurse:
Added `HV_X64_MSR_TSC_INVARIANT_CONTROL` to the doc. Did not include aarch64 register, because it is filtered out in tests and not in FC.

tests/integration_tests/functional/test_cpu_template_helper.py
zulinx86: yep, I'm planning to move them all from FC to the helper tool.

zulinx86 (approved):

sudanl0 (approved):

tests/integration_tests/functional/test_cpu_template_helper.py
zulinx86: While doing the refactoring, I found that KVM_GET_MSRS call fails for some of MSR index list returned by KVM_GET_MSR_INDEX_LIST call (e.g., MSR_ARCH_PERFMON_FIXED_CTR0). That means in any cases, we need to have such a list on Firecracker side.

ShadowCurse: This is an x86 part of code right? For aarch64 you should be able to move this one filtered out reg into the helper tool.

zulinx86: Why do we want to handle it differently between x86 and aarch64?

PR #4492, opened 2024-03-08
Enhance usability of CPU template helper tool

zulinx86:
## Reasons

This PR mainly aims to enhance the usability of the CPU template helper tool.

The most inconvenient thing is that users need to prepare kernel, rootfs and Firecracker config to run the tool, although it doesn't actually start the microVM and those resources don't lead to meaningful changes to guest CPU config. But unfortunately, it cannot run without those resources because the CPU template helper tool should use the Firecracker's microVM initialization process (`build_microvm_for_boot()` of `vmm` crate) which requires such resources.

## Changes

The main change is: CPU template helper tool prepares some required resources (kernel, rootfs and Firecracker config) on the behalf of users if `--config` parameter is not provided.

There are some refactoring commits in the first 6 commits.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet


pb8o (commented):
- chore(cpu-template-helper): Add mock kernels
  - Could we keep the sources in the repo? Looks like a simple .c and Makefile can create those kernel files?

zulinx86:
> * chore(cpu-template-helper): Add mock kernels
>   
>   * Could we keep the sources in the repo? Looks like a simple .c and Makefile can create those kernel files?

Yes, we can keep the source. I'm still trying to find a way to create a PE binary instead of copying it from linux-loader binary in order to make it under our control.

src/vmm/src/arch/aarch64/regs.rs
```diff
@@ -115,6 +115,9 @@ arm64_sys_reg!(KVM_REG_ARM_TIMER_CNT, 3, 3, 14, 3, 2);
 pub const KVM_REG_ARM64_SVE_VLS: u64 =
     KVM_REG_ARM64 | KVM_REG_ARM64_SVE as u64 | KVM_REG_SIZE_U512 | 0xffff;
 
+/// Program Counter
+pub const PC: u64 = arm64_core_reg_id!(KVM_REG_SIZE_U64, 256);
```
ShadowCurse, line 126: I think keeping the calculation of `PC` is valuable (otherwise where does 256 comes from). Also maybe worth removing the custom `offset_of` macro from `regs.rs` as there is same macro in std. You can do:
```rust
use std::mem::offest_of;
pub const PC: u64 = {
  let kreg_off = offset_of!(kvm_regs, regs);
  let pc_off = offset_of!(user_pt_regs, pc);
  arm64_core_reg_id!(KVM_REG_SIZE_U64, kreg_off + pc_off)
};
```

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -222,13 +222,15 @@ impl KvmVcpu {
...
+        // SYS_CNTV_CVAL_EL0 and SYS_CNTPCT_EL0 are timer registers and depend on the elapsed time.
+        // This type of registers are not useful as guest CPU config dump.
+        //
         // The value of program counter (PC) is determined by the given kernel image. It should not
         // be overwritten by a custom CPU template and does not need to be tracked in a fingerprint
         // file.
-        reg_list.retain(|&reg_id| reg_id != SYS_CNTV_CVAL_EL0 && reg_id != PC);
+        reg_list.retain(|&reg_id| {
+            reg_id != SYS_CNTV_CVAL_EL0 && reg_id != SYS_CNTPCT_EL0 && reg_id != PC
+        });
```
ShadowCurse, line 233: We had a discussion offline and I thought we were aligned to move this filtering to the `cpu-template-helper` tool.

src/cpu-template-helper/src/utils/mod.rs
```diff
@@ -64,6 +66,40 @@ pub enum UtilsError {
...
+
+// Utility function to prepare scratch kernel image and rootfs and build mock Firecracker config.
+fn build_mock_config() -> Result<(TempFile, TempFile, String), UtilsError> {
+    let kernel = TempFile::new()?;
+    kernel.as_file().write_all(
+        #[cfg(target_arch = "aarch64")]
+        include_bytes!("mock_kernel/aarch64.bin"),
+        #[cfg(target_arch = "x86_64")]
+        include_bytes!("mock_kernel/x86_64.bin"),
+    )?;
```
pb8o, line 83: Nice trick!

pb8o (approved):

src/cpu-template-helper/src/utils/mod.rs
zulinx86: Yeees! I run into this feature by chance in another place :)

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ and this project adheres to
   and underlying layers. Queue backlog length metrics are flushed periodically.
   They can be used to esimtate an average queue length by request by dividing
   its value by the number of requests served.
+- [#4492](https://github.com/firecracker-microvm/firecracker/pull/4492): Added
+  `cpu-template-helper fingerprint verify` command that functions as a
+  combination of the fingerprint dump and compare commands. This is useful when
+  ensuring that a fingerprint has not changed since the baseline was taken.
```
ShadowCurse, line 30: I this command really necessary? At least for me having 2 separate commands, each doing their own small portion of the work seems like more intuitive and simple design. Now with `verify` you are combining them just for the sake of running 1 command instead of 2. Also looking at the implementation the existence of `--output` parameter is confusing. As I can see it is used to specify path to which current fingerprint will be written, but does anyone expect something called `verify` to output anything except some true/false statement? Maybe if you rename it to something like `DumpCompare` it would be more understandable, but still I don't see much benefit from this.

src/cpu-template-helper/src/utils/mock_kernel/Makefile
```diff
@@ -0,0 +1,16 @@
+ARCH := $(shell uname -m)
+NPROC := $(shell nproc)
+
+.PHONY: mock_kernel clean
```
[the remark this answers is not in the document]
zulinx86: Good catch! I forgot to delete it!

src/vmm/src/vstate/vcpu/aarch64.rs
zulinx86: I'm still thinking the x86 part yet and will do the migration in a follow-up :)

CHANGELOG.md
zulinx86: not necessary, but nice-to have :)

src/vmm/src/arch/aarch64/regs.rs
zulinx86: `offset_of!()` seems to be an unstable feature.
```
error[E0658]: use of unstable library feature 'offset_of'
   --> src/vmm/src/arch/aarch64/regs.rs:122:20
    |
122 |     let kreg_off = offset_of!(kvm_regs, regs);
    |                    ^^^^^^^^^
    |
    = note: see issue #106655 <https://github.com/rust-lang/rust/issues/106655> for more information
```

To avoid this, we have our own `offset__of!()`, but it doesn't work during const eval at compile time :(
```
error: pointers cannot be cast to integers during const eval
   --> src/vmm/src/arch/aarch64/regs.rs:35:13
    |
35  |             std::ptr::addr_of!((*ptr).$field) as usize - ptr as usize
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
120 |     let kreg_off = offset__of!(kvm_regs, regs);
    |                    --------------------------- in this macro invocation
    |
    = note: at compile-time, pointers do not have an integer value
    = note: avoiding this restriction via `transmute`, `union`, or raw pointers leads to compile-time undefined behavior
    = note: this error originates in the macro `offset__of` (in Nightly builds, run with -Z macro-backtrace for more info)

```

zulinx86: I can add a comment where it comes from :)

https://docs.kernel.org/virt/kvm/api.html
> | 0x6030 0000 0010 0040 | PC | 64 | regs.pc |

zulinx86: It seems it becomes stable since 1.77.0: https://doc.rust-lang.org/nightly/src/core/mem/mod.rs.html#1398-1401

zulinx86: I'll add TODO comment there.

src/cpu-template-helper/src/utils/mock_kernel/Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+$(ARCH).bin: main.c
+ifeq ($(ARCH),x86_64)
+	gcc -nostdlib -fno-asynchronous-unwind-tables -s -o x86_64.bin main.c
+else
+	git clone --depth=1 git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git
+	cd linux; make allnoconfig; make -j$(NPROC)
+	head -c 4096 linux/arch/arm64/boot/Image > aarch64.bin
```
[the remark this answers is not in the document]
zulinx86: Yeah, I wasn't able to find an easy way to generate PE formatted mock kernel image, because it requires to have a special header in front of the actual PE image. I'll add that there.

CHANGELOG.md
zulinx86: Rethinking what you pointed out and doubting myself, I came back to the "minimalistic" style of Firecracker. The new command may be just my prediction of customers' need and we can defer this task. I'll remove it for now.

src/cpu-template-helper/src/utils/mock_kernel/Makefile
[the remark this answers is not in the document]
zulinx86: I managed to find a much easier way to generate a file acceptable. added it with some comments.

ShadowCurse (approved):

pb8o (approved):

PR #4505, opened 2024-03-15
fix: use release branch if available to detect changed files

sudanl0:
## Changes:
In nightly tests, use $BUILDKITE_BRANCH instead of main, to get list of changed files on release branch.

## Reason:
Comparing changed files with main leads to running unnecessary tests like buid_devctr even when there is no Dockerfile change in the release branch.
With this change nightly pipeline for release branches will not show unrelated failures.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/.buildkite/common.py
+++ b/.buildkite/common.py
@@ -92,8 +92,18 @@ def pipeline_to_json(pipeline):
 def get_changed_files():
     """
     Get all files changed since `branch`
+    For branch other than `main`:
+        Use BUILDKITE_BRANCH for nightly tests
+        Use BUILDKITE_PULL_REQUEST_BASE_BRANCH for PR tests
     """
-    branch = os.environ.get("BUILDKITE_PULL_REQUEST_BASE_BRANCH", "main")
+    branch = os.environ.get(
+        "BUILDKITE_PULL_REQUEST_BASE_BRANCH", os.environ.get("BUILDKITE_BRANCH", "main")
+    )
+    # This is to allow running buildkite jobs on PRs where
+    # BUILDKITE_BRANCH=refs/pull/123/head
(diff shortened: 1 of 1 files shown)


.buildkite/common.py
```diff
@@ -92,8 +92,13 @@ def pipeline_to_json(pipeline):
 def get_changed_files():
     """
     Get all files changed since `branch`
+    For branch other than `main`:
+        Use BUILDKITE_BRANCH for nightly tests
+        Use BUILDKITE_PULL_REQUEST_BASE_BRANCH for PR tests
     """
-    branch = os.environ.get("BUILDKITE_PULL_REQUEST_BASE_BRANCH", "main")
+    branch = os.environ.get(
+        "BUILDKITE_PULL_REQUEST_BASE_BRANCH", os.environ.get("BUILDKITE_BRANCH", "main")
```
pb8o, line 100: I think this will always evaluate to `""`. It should be sthg like:

```
  branch = os.environ.get("BUILDKITE_PULL_REQUEST_BASE_BRANCH") or os.environ.get("BUILDKITE_BRANCH") or "main"
```

roypat: huh, why does it evaluate to `""`?

pb8o: If the value of the environment variable is an empty string (which is what BK does IIRC), that will be returned as an empty string. That helps differentiate the case when the env var is not present.

```python
% BLA="" ipython3
In [1]: os.environ.get("BLA")
Out[1]: ''
```

roypat: ohhh, interesting! I didn't realize buildkite would set it to empty, instead of not setting it at all. Thanks for the explanation!

PR #4526, opened 2024-03-27
Fail early in case of too big a frame in the TX path

bchalios:
## Changes

Introduce a check that the size of a frame in the TX queue is not bigger than the one the device can handle.

## Reason

We do not handle malformed packets that the driver might send to us. Also, we avoid big allocations when copying from guest memory to Firecracker memory when handling MMDS frames.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,15 @@ and this project adheres to
 
 ### Fixed
 
+- [4526](https://github.com/firecracker-microvm/firecracker/pull/4526): Added a
+  check in the network TX path that the size of the network frames the guest
+  passes to us is not bigger than the maximum frame the device expects to
+  handle. On the TX path, we copy frames destined to MMDS from guest memory to
+  Firecracker memory. Without the check, a mis-behaving virtio-net driver could
+  cause an increase in the memory footprint of the Firecracker process. Now, if
+  we receive such a frame, we ignore it and increase `Net::tx_malformed_frames`
+  metric.
+
 ## \[1.7.0\]
 
 ### Added
--- a/src/vmm/src/devices/virtio/net/device.rs
+++ b/src/vmm/src/devices/virtio/net/device.rs
@@ -607,6 +607,17 @@ impl Net {
                     continue;
                 }
             };
+
+            // We only handle frames that are up to MAX_BUFFER_SIZE
+            if buffer.len() > MAX_BUFFER_SIZE {
+                error!("net: received too big frame from driver");
+                self.metrics.tx_malformed_frames.inc();
+                tx_queue
+                    .add_used(mem, head_index, 0)
+                    .map_err(DeviceError::QueueError)?;
+                continue;
+            }
+
             if !Self::rate_limiter_consume_op(&mut self.tx_rate_limiter, buffer.len() as u64) {
                 tx_queue.undo_pop();
                 self.metrics.tx_rate_limiter_throttled.inc();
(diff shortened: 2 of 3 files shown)


src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -607,6 +607,17 @@ impl Net {
...
                 }
             };
+
+            // We only handle frames that are up to MAX_BUFFER_SIZE
+            if buffer.len() > MAX_BUFFER_SIZE {
+                error!("net: received too big frame from driver");
+                self.metrics.tx_malformed_frames.inc();
+                tx_queue
+                    .add_used(mem, head_index, 0)
+                    .map_err(DeviceError::QueueError)?;
```
kalyazin, line 617: out of curiousity, why `QueueError`, not `MalformedPayload`?

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -1323,6 +1334,28 @@ pub mod tests {
...
     }
 
+    #[test]
+    fn test_tx_big_frame() {
+        let mut th = TestHelper::get_default();
+        th.activate_net();
+        let tap_traffic_simulator = TapTrafficSimulator::new(if_index(&th.net().tap));
+
+        // Send an invalid frame (too big, maximum buffer is MAX_BUFFER_SIZE).
+        th.add_desc_chain(NetQueue::Tx, 0, &[(0, 100000, 0)]);
```
kalyazin, line 1344: nit: can we use a number relative to `MAX_BUFFER_SIZE` (such as 1.5x) instead of a literal, to be explicit?

CHANGELOG.md
```diff
@@ -41,6 +41,15 @@ and this project adheres to
 
 ### Fixed
 
+- [4526](https://github.com/firecracker-microvm/firecracker/pull/4526): Added a
```
kalyazin, line 44: nit: what if we rearrange sentences to tell the following story like:
- we introduced a check (summary)
- this is what happens if it's not there (if guest sends a malformed request, we copy and use host memory)
- that's how it fixes the problem (we check for max frame, drop packet and update metrics)

?

src/vmm/src/devices/virtio/net/device.rs
bchalios: Because we don't fail, we just ignore that frame. The failure which we return is when we try to add the frame in the used queue and that fails

kalyazin: ah, got it, thanks.

src/vmm/src/devices/virtio/net/device.rs
bchalios: Did `MAX_BUFFER_SIZE + 1`

CHANGELOG.md
bchalios: Done!

kalyazin (approved):

src/vmm/src/devices/virtio/net/test_utils.rs
```diff
@@ -395,7 +395,7 @@ pub mod test {
         pub fn get_default() -> TestHelper<'a> {
             let mut event_manager = EventManager::new().unwrap();
             let mut net = default_net();
-            let mem = single_region_mem(MAX_BUFFER_SIZE);
+            let mem = single_region_mem(2 * MAX_BUFFER_SIZE);
```
zulinx86, line 398: Just out of curiosity, why do we need to double it?

zulinx86 (approved):

pb8o (approved):

src/vmm/src/devices/virtio/net/test_utils.rs
bchalios: because in the `test_tx_big_frame` unit test we are trying to add a descriptor in that is bigger than `MAX_BUFFER_SIZE`. The queue logic will catch it if the guest tried to pass us a descriptor that points outside of the allocated guest memory, so we won't hit the control we're checking against here.

PR #4553, opened 2024-04-10
test(cmdline): add delay for serial output file to appear

kalyazin:
## Changes

Add a delay before opening the VM serial output file to avoid a race condition.

## Reason

It is possible that by the time we open the screen output file, guest Linux has not yet started booting, so we may fail to open the file.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- ~~[ ] Any required documentation changes (code and docs) are included in this~~
  PR.
- [x] Commits meet

### DIFF
--- a/tests/framework/microvm.py
+++ b/tests/framework/microvm.py
@@ -993,6 +993,11 @@ class Serial:
             # serial already opened
             return
 
+        attempt = 0
+        while not Path(self._vm.screen_log).exists() and attempt < 5:
+            time.sleep(0.2)
+            attempt += 1
+
         screen_log_fd = os.open(self._vm.screen_log, os.O_RDONLY)
         self._poller = select.poll()
         self._poller.register(screen_log_fd, select.POLLIN | select.POLLHUP)


roypat (approved):

tests/integration_tests/functional/test_kernel_cmdline.py
```diff
@@ -26,6 +28,9 @@ def test_init_params(uvm_plain):
     )
 
     vm.start()
+    # Let the guest start booting for the serial output file to appear
+    time.sleep(1)
```
pb8o, line 32: In microvm.py, could we wait until the file exists? I'm not sure how long that takes and it would only affect microvms launched under screen, which are just a few.

kalyazin: yeah, I was thinking about that too, let me see.

kalyazin: done

tests/framework/microvm.py
```diff
@@ -993,6 +993,11 @@ def open(self):
             # serial already opened
             return
 
+        attempt = 0
+        while not Path(self._vm.screen_log).exists() and attempt < 5:
+            time.sleep(0.2)
+            attempt += 1
+
```
pb8o, line 1000: Should we continue knowing there is no log still? or do a final `assert self._vm.screen_log.exists()`?

kalyazin: I can't really see a difference, to be honest. It will fail either way. I used a pattern from https://github.com/firecracker-microvm/firecracker/blob/main/tests/framework/utils_vsock.py#L150 .

tests/framework/microvm.py
```diff
@@ -993,6 +993,11 @@ def open(self):
             # serial already opened
             return
 
+        attempt = 0
+        while not Path(self._vm.screen_log).exists() and attempt < 5:
+            time.sleep(0.2)
+            attempt += 1
```
bchalios, line 999: Is it only enough for solving the intermittent issues that we see to have the `Serial` opened, or do we need to wait for some output at it? Would it make sense to _also_ wait for the microVM to boot (by running something like `vm.ssh("true")`?

kalyazin: The problem with that is the cmdline test uses `init=/bin/cat -- /etc/issue`, which I doubt will let us connect via ssh.

bchalios: then what about a `sleep` after booting the microVM (additional to the sleep for the Serial output)?

kalyazin: It feels like it's already covered by the pattern matching code: https://github.com/firecracker-microvm/firecracker/blob/main/tests/framework/microvm.py#L982 . If the file exists, it should take care of that.

bchalios (approved):

pb8o (approved):

PR #4567, opened 2024-04-19
fix: power off vcpus only during startup

ShadowCurse:
## Changes
Do not power off vcpus during snapshot restore.

## Reason
Vcpus should be in power off state during kernel boot, but during snapshot restore we do not need to power them off even thought kvm does not complain.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/src/vmm/src/vstate/vcpu/aarch64.rs
+++ b/src/vmm/src/vstate/vcpu/aarch64.rs
@@ -129,15 +129,12 @@ impl KvmVcpu {
         vm_fd: &VmFd,
         vcpu_features: &[VcpuFeatures],
     ) -> Result<(), KvmVcpuError> {
-        let mut kvi = Self::default_kvi(vm_fd, self.index)?;
+        let mut kvi = Self::default_kvi(vm_fd)?;
 
         for feature in vcpu_features.iter() {
             let index = feature.index as usize;
             kvi.features[index] = feature.bitmap.apply(kvi.features[index]);
         }
-
-        self.init_vcpu(&kvi)?;
-        self.finalize_vcpu(&kvi)?;
 
         self.kvi = if !vcpu_features.is_empty() {
             Some(kvi)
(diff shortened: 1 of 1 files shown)


bchalios (dismissed):

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -161,11 +163,6 @@ impl KvmVcpu {
         // We already checked that the capability is supported.
         kvi.features[0] |= 1 << kvm_bindings::KVM_ARM_VCPU_PSCI_0_2;
 
-        // Non-boot cpus are powered off initially.
-        if index > 0 {
-            kvi.features[0] |= 1 << kvm_bindings::KVM_ARM_VCPU_POWER_OFF;
-        }
-
```
roypat, line 168: Mh, I was wondering how this works currently. Surely if we just turn off VCPUs under the guest's nose during a restore, that would break the guest? But from my understanding of the restore code, in `restore_state`, we don't call `default_kvi`, as we expect the kvi to be part of the snapshot. Am I missing something here? What is the consequence of this change for the `match state.kvi { None => ... }` path in `restore_state` below?

ShadowCurse: >  Surely if we just turn off VCPUs under the guest's nose during a restore, that would break the guest?

We were doing this the whole time and it seems to work.

During restore we only call `default_kvi` if snapshot does not have one stored inside of it. And only way the snapshot has one stored is if we use cpu templates.
Actually I looked at the `init` method once again and found that the `kvi` was saved after `KVM_ARM_VCPU_POWER_OFF` was applied to the kvi. So during snap restore of a vm that was using cpu templates, `kvi` stored in the snapshot had `KVM_ARM_VCPU_POWER_OFF` set so secondary vcpus were still powered off. Now I fixed it by saving `kvi` before `KVM_ARM_VCPU_POWER_OFF` is applied.

Also I think we don't need to store kvi optionally. This will make things a bit simpler.

bchalios (approved):

src/vmm/src/vstate/vcpu/aarch64.rs
roypat: Makes sense, thanks for the explanation!

roypat (approved):

PR #4581, opened 2024-04-25
chore: Update to kvm-bindings 0.8.0

roypat:
## Changes

The goal of this PR is to update kvm-bindings from our custom fork to upstream version 0.8.0, as since https://github.com/rust-vmm/kvm-bindings/pull/101 the serialization support we had our fork for is merged upstream. 

Additionally, refactor our emulation code to be more easily unit-testable.

## Reason

The refactors are necessary due to a change in kvm-ioctls. The `self` parameter to `VcpuFd::run` is now a mutable reference. Please see the final commit for a detailed breakdown of why this was incompatible with our existing mocking-based approach to testing vmexits.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


zulinx86 (changes requested):
looks good to me overall other than tiny nitpicking!

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -67,26 +66,33 @@ impl KvmVcpu {
...
+    pub fn run_arch_emulation(&self, exit: VcpuExit) -> Result<VcpuEmulation, VcpuError> {
+        METRICS.vcpu.failures.inc();
+        // TODO: Are we sure we want to finish running a vcpu upon
+        // receiving a vm exit that is not necessarily an error?
+        error!("Unexpected exit reason on vcpu run: {:?}", exit);
+        Err(VcpuError::UnhandledKvmExit(format!("{:?}", exit)))
+    }
+}
+
+impl Vcpu {
```
ShadowCurse, line 95: I can't say I am in favor of this. With these changes `Vcpu` implementation is spread across `mod.rs` and specific arch files. Same with `KvmVcpu` impl now exists in the `mod.rs`.  So we are basically cooking spaghetti  here at this point.
I did some experimentation and I think simply moving all handling of KVM exits to the arch files will be cleaner.
Example is here (only x86): https://github.com/firecracker-microvm/firecracker/compare/main...ShadowCurse:firecracker:vcpu_exit_process_update?expand=1

This actually makes more sense in my mind as all `KVM` specific implementation will be in the arch specific files, while `mod.rs` and `Vcpu` will only act as a communication wrappers.
The change leaves `fd` in the `KvmVcpu`s, so it leaves these annoying accesses to the fds, but I think we can do a follow up that cleans this with additional methods on `KvmVcpu` side.
The only downside of my approach is that there will be duplicated handling of many `KVMEXIT`s. But at the same time, majority of them have no logic, just set metric and return error, so maybe it is not that bad.

roypat: I think duplicating the code is strictly worse, because then we'd have the danger of diverging kvm exit handling if we happen to forget to update one of the copies. At this point, I would greatly prefer to just have everything be stand-alone functions (we actually have a ton of those already in arch/aarch64/vcpu.rs, which now makes the wonder where we actually draw the boundary between the vcpu moduel and the arch module?). That said, I think the splitting of the `Vcpu` impls across multiple modules to account for differences due to arch specifics is fine.

For the KvmVcpu impl in `mod.rs` I kinda agree that it's not nice. I think the problem here is that `KvmVcpu` is not actually "kvm-arch specific functionality" like the comment says - mmio_bus is present on both architectures, and not actually used at all in the arch modules! So we could fix this by hoisting `mmio_bus` out of `KvmVcpu`, either via introducing something like 

```rs
struct VcpuEmulationState {
    mmio_bus: Bus,
    arch_vcpu: KvmVcpu
}

impl VcpuEmulationState {
    fn handle_kvm_exit(&mut self, result: Result<...,...>) -> ... { ... }
}
```

inside of `mod.rs`. Or by just moving `mmio_bus` into `Vcpu` and making `handle_kvm_exit` freestanding, e.g. `fn handle_kvm_exit(kvm_vcpu: &mut KvmVcpu, mmio_bus: &mut Bus, result: ...) -> ...`. Happy to do these in this PR, but larger refactors are probably out of scope here, so maybe we can just track those in a follow up?

src/cpu-template-helper/Cargo.toml
```diff
@@ -19,7 +19,7 @@ serde_json = "1.0.116"
 thiserror = "1.0.59"
 
 vmm = { path = "../vmm" }
-vmm-sys-util = "0.12.1"
+vmm-sys-util = { version = "0.12.1", features = ["with-serde"] }
```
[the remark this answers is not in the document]
roypat: fixed, thanks :)

src/vmm/src/vstate/vcpu/aarch64.rs
ShadowCurse: Ok, how about sharing processing function, like this: https://github.com/firecracker-microvm/firecracker/compare/main...ShadowCurse:firecracker:vcpu_exit_process_update
The only nuance is that we will have `#[cfg(target_arch = "x86_64")]` and `#[cfg(target_arch = "aarch64")]` for `IoOut` and `IoIn` exits, but don't think this is a big deal. Technically for `aarch64` these kvm exits should never fire, because `aarch64` does not support pio.
I think the moving of `fd` or `mmio_bus` and other things can be in a separate PR.

roypat: Okay, I've dropped the VcpuFd moving commit and done something similar to what you proposed (free standing processing function, but still calls down into architecture specific modules)

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -141,11 +141,18 @@ pub struct KvmVcpu {
     pub index: u8,
     /// KVM vcpu fd.
     pub fd: VcpuFd,
+    /// Vcpu peripherals, such as buses
+    pub(super) peripherals: Peripherals,
```
ShadowCurse, line 133: Do we need a `(super)` here?

src/vmm/src/vstate/vcpu/mod.rs
```diff
@@ -458,108 +441,115 @@ impl Vcpu {
...
+                Ok(VcpuEmulation::Interrupted)
+            }
+            emulation_result => handle_kvm_exit(&mut self.kvm_vcpu.peripherals, emulation_result),
+        }
+    }
+}
+
+/// Handle the return value of a call to [`VcpuFd::run`] and update our emulation accordingly
+fn handle_kvm_exit(
+    emulation_state: &mut Peripherals,
```
ShadowCurse, line 458: maybe rename this to `peripherals: &mut Peripherals` for consistency?

src/vmm/src/vstate/vcpu/aarch64.rs
```diff
@@ -54,12 +54,19 @@ pub struct KvmVcpu {
     pub index: u8,
     /// KVM vcpu fd.
     pub fd: VcpuFd,
-    /// Mmio bus.
-    pub mmio_bus: Option<crate::devices::Bus>,
+    /// Vcpu peripherals, such as buses
+    pub(super) peripherals: Peripherals,
```
ShadowCurse, line 58: Same here `(super)` is not really needed.

src/vmm/src/vstate/vcpu/x86_64.rs
roypat: Well, we do not need it to be accessible beyond the vcpu module. Although I'd also be fine with `pub(crate)`, just as long as we don't export it outside of the crate. We have a _ton_ of dead code that rustc doesnt report to us as dead because its marked as `pub` instead of something more restrictive, so I try to not make that problem worse.

src/vmm/src/vstate/vcpu/mod.rs
roypat: done :)

src/vmm/src/vstate/vcpu/x86_64.rs
ShadowCurse: Ok, let's leave this for now.

ShadowCurse (approved):

zulinx86 (approved):

Issue #4549, opened 2024-04-09
Avoid IoVecBuffer allocations in the TX network data path


bchalios:
### Description

When the guest wants to transmit an ethernet frame over a virtio-net device it places the frame somewhere in guest memory and sends us an interrupt. A single frame might be placed in multiple buffers in memory.
So, our virtio-net implementation, does scatter-gather writes to the backing TAP device (call writev once with mutliple buffers, instead of write many times) to increase throughput on the TX path. We do that with
the help of a type we call IoVecBuffer. This is, essentially an abstraction layer on top of a buffer scattered in guest memory.

The current implementation creates a new IoVecBuffer item for every ethernet frame we forward from the guest into the TAP device. This implies memory allocations across a hot-path that we would like to avoid.

We can avoid the multiple allocations if we change the implementation of IoVecBuffer to allow for re-using, i.e. we create the IoVecBuffer once during device initialization and then re-use it for all TX frames.

### Solution

* Make the change in the API of IoVecBuffer and integrate it with virtio-net device.
* Ensure everything is working as expected (all network-related tests keep passing).
* For bonus points: check for potential performance benefits on the TX path.

JackThomson2:
Hey, I'll take a look at this! I'll keep you posted on how I get on

roypat:
Fixed by https://github.com/firecracker-microvm/firecracker/pull/4589

PR #4589, opened 2024-05-01
Reuse the IoVecBuffer on TX

JackThomson2:
## Changes

Removes the allocation of the `IoVecBuffer` in the virtio net device during tx by reusing the same instance for each message.

To do this we have to implement Send on the type for it to work, I have tried to cover the unsafe sections the best I can.
## Reason

Reduce the number of allocation

Ticket here: https://github.com/firecracker-microvm/firecracker/issues/4549

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/src/vmm/src/devices/virtio/iovec.rs
+++ b/src/vmm/src/devices/virtio/iovec.rs
@@ -37,7 +37,7 @@ type IoVecVec = SmallVec<[iovec; 4]>;
 /// It describes a buffer passed to us by the guest that is scattered across multiple
 /// memory regions. Additionally, this wrapper provides methods that allow reading arbitrary ranges
 /// of data from that buffer.
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct IoVecBuffer {
     // container of the memory regions included in this IO vector
     vecs: IoVecVec,
@@ -45,11 +45,21 @@ pub struct IoVecBuffer {
     len: u32,
 }
 
+// SAFETY: `IoVecBuffer` doesn't allow for interior mutability and no shared ownership is possible
+// as it doesn't implement clone
+unsafe impl Send for IoVecBuffer {}
+
 impl IoVecBuffer {
     /// Create an `IoVecBuffer` from a `DescriptorChain`
-    pub fn from_descriptor_chain(head: DescriptorChain) -> Result<Self, IoVecError> {
-        let mut vecs = IoVecVec::new();
-        let mut len = 0u32;
+    ///
+    /// # Safety
+    ///
+    /// The descriptor chain cannot be referencing the same memory location as another chain
+    pub unsafe fn load_descriptor_chain(
+        &mut self,
+        head: DescriptorChain,
+    ) -> Result<(), IoVecError> {
+        self.clear();
 
         let mut next_descriptor = Some(head);
         while let Some(desc) = next_descriptor {
@@ -66,18 +76,32 @@ impl IoVecBuffer {
                 .ptr_guard_mut()
                 .as_ptr()
                 .cast::<c_void>();
-            vecs.push(iovec {
+            self.vecs.push(iovec {
                 iov_base,
                 iov_len: desc.len as size_t,
             });
-            len = len
+            self.len = self
+                .len
                 .checked_add(desc.len)
                 .ok_or(IoVecError::OverflowedDescriptor)?;
 
             next_descriptor = desc.next_descriptor();
         }
 
-        Ok(Self { vecs, len })
+        Ok(())
+    }
+
+    /// Create an `IoVecBuffer` from a `DescriptorChain`
+    ///
+    /// # Safety
+    ///
+    /// The descriptor chain cannot be referencing the same memory location as another chain
+    pub unsafe fn from_descriptor_chain(head: DescriptorChain) -> Result<Self, IoVecError> {
+        let mut new_buffer: Self = Default::default();
+
+        new_buffer.load_descriptor_chain(head)?;
+
+        Ok(new_buffer)
     }
 
     /// Get the total length of the memory regions covered by this `IoVecBuffer`
@@ -93,6 +117,12 @@ impl IoVecBuffer {
     /// Returns the length of the `iovec` array.
     pub fn iovec_count(&self) -> usize {
         self.vecs.len()
+    }
+
+    /// Clears the `iovec` array
+    pub fn clear(&mut self) {
+        self.vecs.clear();
+        self.len = 0u32;
     }
 
     /// Reads a number of bytes from the `IoVecBuffer` starting at a given offset.
@@ -428,11 +458,13 @@ mod tests {
         let mem = default_mem();
         let (mut q, _) = read_only_chain(&mem);
         let head = q.pop(&mem).unwrap();
-        IoVecBuffer::from_descriptor_chain(head).unwrap();
+        // SAFETY: This descriptor chain is only loaded into one buffer
+        unsafe { IoVecBuffer::from_descriptor_chain(head).unwrap() };
 
         let (mut q, _) = write_only_chain(&mem);
         let head = q.pop(&mem).unwrap();
-        IoVecBuffer::from_descriptor_chain(head).unwrap_err();
+        // SAFETY: This descriptor chain is only loaded into one buffer
+        unsafe { IoVecBuffer::from_descriptor_chain(head).unwrap_err() };
 
         let (mut q, _) = read_only_chain(&mem);
         let head = q.pop(&mem).unwrap();
@@ -449,7 +481,8 @@ mod tests {
         let (mut q, _) = read_only_chain(&mem);
         let head = q.pop(&mem).unwrap();
 
-        let iovec = IoVecBuffer::from_descriptor_chain(head).unwrap();
+        // SAFETY: This descriptor chain is only loaded once in this test
+        let iovec = unsafe { IoVecBuffer::from_descriptor_chain(head).unwrap() };
         assert_eq!(iovec.len(), 4 * 64);
     }
 
@@ -459,6 +492,7 @@ mod tests {
         let (mut q, _) = write_only_chain(&mem);
         let head = q.pop(&mem).unwrap();
 
+        // SAFETY: This descriptor chain is only loaded once in this test
         let iovec = IoVecBufferMut::from_descriptor_chain(head).unwrap();
         assert_eq!(iovec.len(), 4 * 64);
     }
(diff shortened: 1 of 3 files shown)


src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -64,16 +72,30 @@ impl IoVecBuffer {
...
+        Ok(())
+    }
+
+    /// Create an `IoVecBuffer` from a `DescriptorChain`
+    ///
+    /// # Safety
+    ///
+    /// The descriptor chain cannot be referencing the same memory location as another chain
+    pub unsafe fn from_descriptor_chain(head: DescriptorChain) -> Result<Self, IoVecError> {
+        // SAFETY: New buffer is created from the DescriptorChain which doesnt implement clone
```
bchalios, line 93: Why is that useful here?

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -197,6 +199,8 @@ impl Net {
             activate_evt: EventFd::new(libc::EFD_NONBLOCK).map_err(NetError::EventFd)?,
             mmds_ns: None,
             metrics: NetMetricsPerDevice::alloc(id),
+            // SAFETY: Only constructed in the VMM thread so no concurrent buffers
```
bchalios, line 202: I don't think we need to have a `SAFETY` comment for this. calling `IoVecBuffer::default()` is not `unsafe`

bchalios: same as before

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -35,19 +35,27 @@ type IoVecVec = SmallVec<[iovec; 4]>;
 /// It describes a buffer passed to us by the guest that is scattered across multiple
 /// memory regions. Additionally, this wrapper provides methods that allow reading arbitrary ranges
 /// of data from that buffer.
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct IoVecBuffer {
     // container of the memory regions included in this IO vector
     vecs: IoVecVec,
```
roypat, line 41: Mh, now that we're reusing iovec buffers, I was hoping that the `IoVecVec`/`smallvec` stuff won't be needed anymore. Can we revert https://github.com/firecracker-microvm/firecracker/commit/e5a5d0e02b3d5941ada4f48891b836507cce1dea on top of this and run some performance tests to validate that hope? It'd simplify the code nicely

src/vmm/src/devices/virtio/net/device.rs
roypat: Yup, I think these might be left-overs from when `new` was `unsafe`?

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -597,19 +601,17 @@ impl Net {
...
-                Ok(buffer) => buffer,
-                Err(_) => {
-                    self.metrics.tx_fails.inc();
-                    tx_queue
-                        .add_used(mem, head_index, 0)
-                        .map_err(DeviceError::QueueError)?;
-                    continue;
-                }
-            };
+            // SAFETY: This descriptor chain is only loaded into this buffer
```
roypat, line 604: Let's elaborate on these - virtio requests are handled sequentially, so no two iovecbuffers are ever "live" at the same time, meaning this one really has exclusive ownership over the memory (well, from rust's side.. the guest can do as it pleases)

roypat: We should `clear()` this buffer as soon as we're done using it (e.g. at the end of this function)

src/vmm/src/devices/virtio/vsock/packet.rs
```diff
@@ -124,7 +124,8 @@ impl VsockPacket {
     /// - [`VsockError::DescChainTooShortForPacket`] if the contained vsock header describes a vsock
     ///   packet whose length exceeds the descriptor chain's actual total buffer length.
     pub fn from_tx_virtq_head(chain: DescriptorChain) -> Result<Self, VsockError> {
-        let buffer = IoVecBuffer::from_descriptor_chain(chain)?;
+        // SAFETY: chain is only loaded into a single buffer
```
roypat, line 127: Same comment as above :)

src/vmm/src/devices/virtio/net/device.rs
JackThomson2: Ah yes must of been left over my bad

src/vmm/src/devices/virtio/iovec.rs
JackThomson2: Oh sure I can give that a try that would be nice.

src/vmm/src/devices/virtio/net/device.rs
JackThomson2: Sounds great thanks I'll update the comments and clear at the end thanks

src/vmm/src/devices/virtio/iovec.rs
roypat: As per offline discussion: I mixed up some things here, to avoid the smallvec stuff we'd need to use this in the vsock device, not the net device.

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -37,19 +37,23 @@ type IoVecVec = SmallVec<[iovec; 4]>;
...
+// as it doesn't implement clone
+unsafe impl Send for IoVecBuffer {}
+
 impl IoVecBuffer {
     /// Create an `IoVecBuffer` from a `DescriptorChain`
-    pub fn from_descriptor_chain(head: DescriptorChain) -> Result<Self, IoVecError> {
-        let mut vecs = IoVecVec::new();
-        let mut len = 0u32;
+    pub unsafe fn load_descriptor_chain(&mut self, head: DescriptorChain) -> Result<(), IoVecError> {
+        self.vecs.clear();
```
roypat, line 55: I think we really need an additional `pub fn clear(&mut self) { self.vecs.clear() }` function in these iovecbuffer types that call inside `process_tx` at the very end of the function. That's because for safety, we rely on no two buffers pointing towards the same slices of guest memory. We know this will be true because all buffers across all devices are processed sequentially, but we need to make sure that some cached buffer (such as the one in the net device) does not contain "stale" points while we're processing another buffer. Tbh, this is mainly just me being paranoid, I don't think anything bad can happen from it (because the only thing we do with those stale pointers is overwrite them later), but let's be precise here :)

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -142,6 +142,8 @@ pub struct Net {
     /// Only if MMDS transport has been associated with it.
     pub mmds_ns: Option<MmdsNetworkStack>,
     pub(crate) metrics: Arc<NetDeviceMetrics>,
+
+    io_vec_buffer: IoVecBuffer,
```
roypat, line 146: Maybe the name here can indicate that the buffer is used on the tx path?

roypat (commented):
Looks pretty good! Could you just have a look at my two comments and the test failures in CI? They're related to clippy complaining about missing safety comments. For the ones in `mod test`, you can just add a `#![allow(clippy::undocumented_unsafe_blocks)]` at the top of the module. No point in annotating these in unittests.

src/vmm/src/devices/virtio/iovec.rs
roypat: Yeah, sorry, I meant that in addition to clearing the buffer here, it should also get cleared at the end `process_tx`. Because right now what can happen is 
1. we process a TX packet and load the descriptor chain into the cached buffer
2. guest reuses memory location for a vsock package
3. vsock code loads the same memory location into its iovecbuffer, meaning we now have two buffers pointing to the same memory
4. cached buffer in net device is only cleared when the next TX packet arrives.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -596,19 +599,16 @@ impl Net {
...
+                self.metrics.tx_fails.inc();
+                tx_queue
+                    .add_used(mem, head_index, 0)
+                    .map_err(DeviceError::QueueError)?;
+                continue;
             };
 
             // We only handle frames that are up to MAX_BUFFER_SIZE
-            if buffer.len() as usize > MAX_BUFFER_SIZE {
+            if self.io_vec_buffer.len() as usize > MAX_BUFFER_SIZE {
```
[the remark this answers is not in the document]
roypat: Actually, can we avoid it when clippy isn't forcing us to? u32/u64 -> usize is always safe, and I really dislike the clutter that `try_into().unwrap()` unnecessarily adds :sweat_smile:

src/vmm/src/devices/virtio/iovec.rs
bchalios: oh, you are right! I had not thought of this scenario. That makes perfect sense. I wonder if there is a way to structure the code so that this is guaranteed at compile time.

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
bchalios: I know it's very unlikely to happen, but I like the idea of eliminating the possibility of silent lossy conversions. Again, I know a lossy conversion is not possible here, but I'd like us to avoid the habit.

Ideally, I'd like us to have something like that: https://github.com/materializeinc/materialize/blob/main/src/ore/src/cast.rs.

But anyways, this is a nit, no need to apply.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -142,6 +142,8 @@ pub struct Net {
     /// Only if MMDS transport has been associated with it.
     pub mmds_ns: Option<MmdsNetworkStack>,
     pub(crate) metrics: Arc<NetDeviceMetrics>,
+
+    rx_buffer: IoVecBuffer,
```
roypat, line 146:
```suggestion
    tx_buffer: IoVecBuffer,
```
?

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -37,19 +37,30 @@ type IoVecVec = SmallVec<[iovec; 4]>;
...
+    ///
+    /// # Safety
+    ///
+    /// The descriptor chain cannot be referencing the same memory location as another chain
+    pub unsafe fn load_descriptor_chain(
+        &mut self,
+        head: DescriptorChain,
+    ) -> Result<(), IoVecError> {
+        self.clear();
+        self.len = 0u32;
```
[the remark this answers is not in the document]
JackThomson2: Yeah good catch thanks I'll update in next revision

src/vmm/src/devices/virtio/net/device.rs
JackThomson2: My bad thanks I'll update in next revision

roypat (approved):
:tada:

bchalios (approved):
:shipit:

PR #4606, opened 2024-05-07
fix(snapshot): prevent an integer overflow when calculating TSC scaling

pb8o:
Because we are using u32 integer math when checking the TSC, it could happen that the value can overflow if the TSC value is very high, around 18GHz. While there is no processor on the market near that frequency, it's worth fixing this in case the snapshot gets corrupted somehow.

## Changes

...

## Reason

To make the snapshot implementation more robust

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/src/vmm/src/vstate/vcpu/x86_64.rs
+++ b/src/vmm/src/vstate/vcpu/x86_64.rs
@@ -29,8 +29,8 @@ use crate::vstate::vm::Vm;
 // The value of 250 parts per million is based on
 // the QEMU approach, more details here:
 // https://bugzilla.redhat.com/show_bug.cgi?id=1839095
-const TSC_KHZ_TOL_NUMERATOR: u32 = 250;
-const TSC_KHZ_TOL_DENOMINATOR: u32 = 1_000_000;
+const TSC_KHZ_TOL_NUMERATOR: i64 = 250;
+const TSC_KHZ_TOL_DENOMINATOR: i64 = 1_000_000;
 
 /// Errors associated with the wrappers over KVM ioctls.
 #[derive(Debug, PartialEq, Eq, thiserror::Error, displaydoc::Display)]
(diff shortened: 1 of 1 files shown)


src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -433,7 +433,7 @@ impl KvmVcpu {
         // per million beacuse it is common for TSC frequency
         // to differ due to calibration at boot time.
         let diff = (i64::from(self.get_tsc_khz()?) - i64::from(state_tsc_freq)).abs();
-        Ok(diff > i64::from(state_tsc_freq * TSC_KHZ_TOL_NUMERATOR / TSC_KHZ_TOL_DENOMINATOR))
+        Ok(diff > i64::from(state_tsc_freq) * TSC_KHZ_TOL_NUMERATOR / TSC_KHZ_TOL_DENOMINATOR)
```
roypat, line 436: Mh, this kinda just punts the issue to instead happen at `300000000000GHz`. Maybe we can use `checked_mul` and propagate the error?

bchalios: does it? `state_tsc_freq` remains a `u32`. So the max value of it will be `u32::MAX`. I think that `i64::from(u32::MAX)` * 250 should not overflow, right?

That said, doing the `checked_mul` thing is probably a better way to handle this anyways.

pb8o: Good point, I will try and fix all those

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -433,7 +433,7 @@ impl KvmVcpu {
         // per million beacuse it is common for TSC frequency
         // to differ due to calibration at boot time.
         let diff = (i64::from(self.get_tsc_khz()?) - i64::from(state_tsc_freq)).abs();
-        Ok(diff > i64::from(state_tsc_freq * TSC_KHZ_TOL_NUMERATOR / TSC_KHZ_TOL_DENOMINATOR))
+        Ok(diff > i64::from(state_tsc_freq).checked_mul(TSC_KHZ_TOL_NUMERATOR).unwrap() / TSC_KHZ_TOL_DENOMINATOR)
```
pb8o, line 436: This should be an ok_or but I guess I need to introduce a new error for this to work... or are we happy enough that this should not overflow given that it's `u32 * 250`?

roypat: Yeah, I'm perfectly happy to keep it with `*` since u32 * 250 cannot overflow!

roypat (approved):

PR #4617, opened 2024-05-16
Revert "feat: use host network for docker"

ShadowCurse:
Reverts firecracker-microvm/firecracker#4614 because this configuration change appeared to be unnecessary.
Additionally it introduce unwanted changes to the way CI was working.


bchalios (changes requested):
Could you please add context both in the commit message and the PR description as to why we are reverting it?

bchalios (approved):

roypat (approved):

PR #4618, opened 2024-05-17
fix(snapshot/x86_64): make sure TSC_DEADLINE MSR is non-zero

kalyazin:
## Changes

This change introduces a workaround. If when taking a snapshot, we see a zero MSR_IA32_TSC_DEADLINE, we replace its value with the MSR_IA32_TSC value from the same vCPU to make sure the vCPU will continue to receive TSC interrupts.

## Reason

On x86_64, we observed that when restoring from a snapshot, one of the vCPUs had MSR_IA32_TSC_DEADLINE cleared and never received TSC interrupts until the MSR is updated externally (eg by setting the system time).

We believe this happens because the TSC interrupt is lost during snapshot taking process: the MSR is cleared, but the interrupt is not delivered to the guest, so the guest does not rearm the timer.

A visible effect of that is failure to connect to a restored VM via SSH, similar to https://buildkite.com/firecracker/firecracker-pr-nightly/builds/1403#018f83db-5395-4656-8d9c-83b6fcfcfd54/50-1994 .

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet


docs/snapshotting/snapshot-support.md
```diff
@@ -171,6 +171,11 @@ The snapshot functionality is still in developer preview due to the following:
   the data store is not persisted across snapshots.
 - Configuration information for metrics and logs are not saved to the snapshot.
   These need to be reconfigured on the restored microVM.
+- On x86_64, if a vCPU has MSR_IA32_TSC_DEADLINE set to 0 when a snapshot it
```
zulinx86, line 174: typo?
```suggestion
- On x86_64, if a vCPU has MSR_IA32_TSC_DEADLINE set to 0 when a snapshot is
```

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -143,6 +144,40 @@ pub(super) struct Peripherals {
...
+/// If any of per-vCPU IA32_TSC_DEADLINE MSRs are zero, update them
+/// with the maximum value observed across all vCPUs.
+///
+/// We observed that sometimes when taking a snapshot,
+/// the IA32_TSC_DEADLINE MSR is cleared, but the interrupt is not
+/// delivered to the guest, leading to a situation where one
+/// of the vCPUs never receives TSC interrupts after restoring,
+/// until the MSR is updated externally, eg by setting the system time.
+pub fn fix_zero_tsc_deadline_msr(vcpu_states: &mut [VcpuState]) {
+    let max_tsc_deadline_val = vcpu_states
```
zulinx86, line 156: Is there any reason to use the max value? Doesn't setting it to 1 (the minimum number larger than 0) work? If the interrupt just gets ignored, it makes sense to me to set vCPU interrupted as soon as possible. wdyt?

kalyazin: That's quite right, I didn't think about it. Looks like a better option!

docs/snapshotting/snapshot-support.md
kalyazin: Yes, thanks! Fixed.

src/vmm/src/vstate/vcpu/x86_64.rs
kalyazin: done

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -282,6 +283,36 @@ impl KvmVcpu {
...
+            for msr in msrs.as_mut_slice().iter_mut() {
+                if msr.index == MSR_IA32_TSC_DEADLINE && msr.data == 0 {
+                    warn!(
+                        "MSR_IA32_TSC_DEADLINE is 0, replacing with {:x}.",
+                        tsc_value
+                    );
+                    msr.data = tsc_value;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
kalyazin: I don't think it takes long, but I decided to put it in the snapshot path (which is less latency-sensitive than restore) for this reason, just in case.

pb8o (dismissed):

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -949,4 +984,73 @@ mod tests {
...
+
+        // We expect for the MSR_IA32_TSC_DEADLINE to get updated with the MSR_IA32_TSC value, while
+        // other MSRs should remain unchanged.
+        assert_msrs!(
+            msr_chunks,
+            &[
+                (MSR_TSC_AUX, 1),
+                (MSR_IA32_TSC_DEADLINE, 42),
+                (MSR_IA32_TSC, 42)
+            ]
```
ShadowCurse, line 1039: Considering `msr_chunks` is a `Vec<Vec<...>>` to do an easier comparison you can flatten the vector into just `Vec<...>` and do a simple loop over 2 vectors.
```Rust
let flat_msr_chunks = msr_chunks.iter().flatten().cloned().collect();
let correct_msrs = [
                (MSR_TSC_AUX, 1),
                (MSR_IA32_TSC_DEADLINE, 42),
                (MSR_IA32_TSC, 42)
            ];
for (a, b) in flat_msr_chunks.iter().zip(correct_msrs.iter()) {
  assert_eq!(a.index, b.index);
  assert_eq!(a.data, b.data);
}
```

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -282,6 +283,36 @@ impl KvmVcpu {
...
+
+            for msr in msrs.as_mut_slice().iter_mut() {
+                if msr.index == MSR_IA32_TSC_DEADLINE && msr.data == 0 {
+                    warn!(
+                        "MSR_IA32_TSC_DEADLINE is 0, replacing with {:x}.",
+                        tsc_value
+                    );
+                    msr.data = tsc_value;
+                }
+            }
```
ShadowCurse, line 312: As I understand both `TSC` and `TSC_DEADLINE` need to be present in the msr list for this to work right?
The ` .unwrap_or(0)` line is a bit confusing. If `TSC_DEADLINE` is 0 then there is no point of setting it to 0 again.
Think this version will be more clear:
```Rust
// If no TSC found we cannot do anything
let Some(tsc_msr) = msrs
                .as_slice()
                .iter()
                .find(|msr| msr.index == MSR_IA32_TSC).cloned() else { continue; };
// If no tsc_deadline we cannot do anything
let Some(tsc_deadline_msr) = msrs
                .as_mut_slice()
                .iter_mut()
                .find(|msr| msr.index == MSR_IA32_TSC) else { continue; };
if tsc_deadline_msr.data == 0 {
  warn!(
    "MSR_IA32_TSC_DEADLINE is 0, replacing with {:x}.",
    tsc_value
  );
  tsc_deadline_msr.data = tsc_msr.data;
}
```

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -949,4 +984,73 @@ mod tests {
...
+                    }
+                }
+            }
+        };
+    }
+
+    #[test]
+    fn test_fix_zero_tsc_deadline_msr_zero() {
+        let mut msr_chunks = vec![
+            msrs_from_entries!(&[(MSR_TSC_AUX, 1)]),
```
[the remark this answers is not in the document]
kalyazin: this is to make sure we don't touch other MSRs unexpectedly. This is a weak check, so I'm not overly attached to it. If you feel it's confusing, I'm happy to remove.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -949,4 +984,73 @@ mod tests {
...
+                        .iter()
+                        .find(|msr| msr.index == index)
+                        .map(|msr| msr.data)
+                    {
+                        assert_eq!(msr_data, data);
+                    }
+                }
+            }
+        };
+    }
```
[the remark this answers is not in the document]
kalyazin: good point, done, thanks.

src/vmm/src/vstate/vcpu/x86_64.rs
kalyazin: done.

src/vmm/src/vstate/vcpu/x86_64.rs
kalyazin: While I was thinking about this, I realised that the original code was not properly handling the case where TSC and TSC_DEADLINE MSRs were in separate chunks. I fixed that part and added a test.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -949,4 +986,78 @@ mod tests {
...
+                    ..Default::default()
+                })
+                .collect::<Vec<_>>(),
+        )
+        .unwrap()
+    }
+
+    fn assert_msrs(msr_chunks: &[Msrs], expected_msr_entries: &[(u32, u64)]) {
+        let flattened_msrs = msr_chunks.iter().flat_map(|msrs| msrs.as_slice());
+        let expected_msrs: Vec<(u32, u64)> = expected_msr_entries.to_vec();
```
ShadowCurse, line 1006: You don't really need to copy here. Just used `expected_msr_entries` for iteration.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -282,6 +283,39 @@ impl KvmVcpu {
...
+                    if msr.index == MSR_IA32_TSC_DEADLINE && msr.data == 0 {
+                        warn!(
+                            "MSR_IA32_TSC_DEADLINE is 0, replacing with {:x}.",
+                            tsc_value
+                        );
+                        msr.data = tsc_value;
+                    }
+                }
+            }
+        }
```
ShadowCurse, line 316: Here is a version with a bit less indentation.
```Rust
let Some(tsc_value) = msr_chunks
            .iter()
            .flat_map(|msrs| msrs.as_slice())
            .filter(|msr| msr.index == MSR_IA32_TSC)
            .map(|msr| msr.data) else { continue; };

for msr in msr_chunks
            .iter_mut()
            .flat_map(|msrs| msrs.as_mut_slice())
            .filter(|msr| msr.index == MSR_IA32_TSC_DEADLINE && msr.data == 0) {
  warn!(
    "MSR_IA32_TSC_DEADLINE is 0, replacing with {:x}.",
    tsc_value
  );
  msr.data = tsc_msr.data;
}
```

src/vmm/src/vstate/vcpu/x86_64.rs
kalyazin: true, done

src/vmm/src/vstate/vcpu/x86_64.rs
kalyazin: Not sure how `continue` can be used outside a loop. Could you elaborate?
I got rid of the nested loop, let me know if it looks better to you.

ShadowCurse: My mistake, there should have been `return` instead of `continue`.
I think your current version is fine.

ShadowCurse (approved):

pb8o (approved):

PR #4638, opened 2024-06-11
Add vCPUs to VMM via new endpoint 

jxcurtis:
## Changes
* Created new API endpoint `hotplug` used to request resources on demand (currently only vCPU implemented)
* Currently, when a valid request is made to `hotplug`, the specified number of vCPU threads are added directly to the guest VMM
* For now, in order to allow hotplugging Firecracker must be launched with `--no-seccomp` option
* Unit tests added to test functionality of both endpoint and adding of vCPUs

## Reason
This is the first step in implementing resource hotplugging via ACPI.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/src/firecracker/src/api_server/parsed_request.rs
+++ b/src/firecracker/src/api_server/parsed_request.rs
@@ -15,6 +15,8 @@ use super::request::boot_source::parse_put_boot_source;
 use super::request::cpu_configuration::parse_put_cpu_config;
 use super::request::drive::{parse_patch_drive, parse_put_drive};
 use super::request::entropy::parse_put_entropy;
+#[cfg(target_arch = "x86_64")]
+use super::request::hotplug::parse_put_hotplug;
 use super::request::instance_info::parse_get_instance_info;
 use super::request::logger::parse_put_logger;
 use super::request::machine_configuration::{
@@ -89,6 +91,8 @@ impl TryFrom<&Request> for ParsedRequest {
             (Method::Put, "boot-source", Some(body)) => parse_put_boot_source(body),
             (Method::Put, "cpu-config", Some(body)) => parse_put_cpu_config(body),
             (Method::Put, "drives", Some(body)) => parse_put_drive(body, path_tokens.next()),
+            #[cfg(target_arch = "x86_64")]
+            (Method::Put, "hotplug", Some(body)) => parse_put_hotplug(body),
             (Method::Put, "logger", Some(body)) => parse_put_logger(body),
             (Method::Put, "machine-config", Some(body)) => parse_put_machine_config(body),
             (Method::Put, "metrics", Some(body)) => parse_put_metrics(body),
@@ -740,6 +744,22 @@ pub mod tests {
                     \"size\": 0, \"one_time_burst\": 0, \"refill_time\": 0 } } }";
         sender
             .write_all(http_request("PUT", "/drives/string", Some(body)).as_bytes())
+            .unwrap();
+        connection.try_read().unwrap();
+        let req = connection.pop_parsed_request().unwrap();
+        ParsedRequest::try_from(&req).unwrap();
+    }
+
+    #[test]
+    #[cfg(target_arch = "x86_64")]
+    fn test_try_from_put_hotplug_vcpu() {
+        let (mut sender, receiver) = UnixStream::pair().unwrap();
+        let mut connection = HttpConnection::new(receiver);
+        let body = r#"{
+            "Vcpu": { "add": 1 }
+        }"#;
+        sender
+            .write_all(http_request("PUT", "/hotplug", Some(body)).as_bytes())
             .unwrap();
         connection.try_read().unwrap();
         let req = connection.pop_parsed_request().unwrap();
--- a/src/firecracker/src/api_server/request/hotplug.rs
+++ b/src/firecracker/src/api_server/request/hotplug.rs
@@ -1,0 +1,47 @@
+// Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+use vmm::logger::{IncMetric, METRICS};
+use vmm::rpc_interface::VmmAction;
+use vmm::vmm_config::hotplug::HotplugRequestConfig;
+
+use super::super::parsed_request::{ParsedRequest, RequestError};
+use super::Body;
+
+pub(crate) fn parse_put_hotplug(body: &Body) -> Result<ParsedRequest, RequestError> {
+    METRICS.put_api_requests.hotplug.inc();
+    METRICS.hotplug.hotplug_request_count.inc();
+    let config = serde_json::from_slice::<HotplugRequestConfig>(body.raw()).map_err(|err| {
+        METRICS.hotplug.hotplug_request_fails.inc();
+        err
+    })?;
+    Ok(ParsedRequest::new_sync(VmmAction::HotplugRequest(config)))
+}
+
+#[cfg(test)]
+mod tests {
+
+    use hotplug::parse_put_hotplug;
+    use vmm::rpc_interface::VmmAction;
+    use vmm::vmm_config::hotplug::{HotplugRequestConfig, HotplugVcpuConfig};
+
+    use super::super::*;
+    use crate::api_server::parsed_request::tests::vmm_action_from_request;
+
+    #[test]
+    fn test_parse_put_hotplug() {
+        // Case 1. Invalid body
+        parse_put_hotplug(&Body::new("invalid body")).unwrap_err();
+
+        // Case 2. vCPU Resource
+        let body = r#"{
+            "Vcpu" : { "add": 4 }
+        }"#;
+
+        let expected_config = HotplugVcpuConfig { add: 4 };
+
+        assert_eq!(
+            vmm_action_from_request(parse_put_hotplug(&Body::new(body)).unwrap()),
+            VmmAction::HotplugRequest(HotplugRequestConfig::Vcpu(expected_config))
+        );
+    }
+}
--- a/src/firecracker/swagger/firecracker.yaml
+++ b/src/firecracker/swagger/firecracker.yaml
@@ -282,6 +282,30 @@ paths:
           schema:
             $ref: "#/definitions/Error"
 
+  /hotplug:
+    put:
+      summary: Hotplugs resources to the microVM. Only available for x86_64 hosts. Post-boot only.
+      description:
+        Adds new resources to the VM.
+      parameters:
+        - name: body
+          in: body
+          description: The resource type and amount to be hotplugged.
+          required: true
+          schema: 
+            $ref: "#/definitions/Hotplug"
+      responses:
+        204:
+          description: Resources hotplugged successfully.
+        400: 
+          description: Resources cannot be hotplugged due to bad input or a VMM-related error.
+          schema:
+            $ref: "#/definitions/Error"
+        default:
+          description: Internal server error.
+          schema:
+            $ref: "#/definitions/Error"
+
   /logger:
     put:
       summary: Initializes the logger by specifying a named pipe or a file for the logs output.
@@ -942,6 +966,21 @@ definitions:
           $ref: "#/definitions/NetworkInterface"
       vsock:
         $ref: "#/definitions/Vsock"
+
+  Hotplug:
+    type: object
+    description:
+      Defines the configuration for hot plugging.
+    properties:
+      Vcpu:
+        type: object
+        description: Hotplug requests relating to vCPU cores.
+        properties:
+          add:
+            description: Number of vCPUs cores to be added to the VM
+            type: integer
+            minimum: 1
+
 
   InstanceActionInfo:
     type: object
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -145,6 +145,7 @@ impl std::convert::From<linux_loader::cmdline::Error> for StartMicrovmError {
 }
 
 #[cfg_attr(target_arch = "aarch64", allow(unused))]
+#[allow(clippy::too_many_arguments)]
 fn create_vmm_and_vcpus(
     instance_info: &InstanceInfo,
     event_manager: &mut EventManager,
@@ -152,6 +153,7 @@ fn create_vmm_and_vcpus(
     uffd: Option<Uffd>,
     track_dirty_pages: bool,
     vcpu_count: u8,
+    #[cfg(target_arch = "x86_64")] seccomp_filters: BpfThreadMap,
     kvm_capabilities: Vec<KvmCapability>,
 ) -> Result<(Vmm, Vec<Vcpu>), StartMicrovmError> {
     use self::StartMicrovmError::*;
@@ -229,6 +231,8 @@ fn create_vmm_and_vcpus(
         uffd,
         vcpus_handles: Vec::new(),
         vcpus_exit_evt,
+        #[cfg(target_arch = "x86_64")]
+        seccomp_filters,
         resource_allocator,
         mmio_device_manager,
         #[cfg(target_arch = "x86_64")]
@@ -309,6 +313,8 @@ pub fn build_microvm_for_boot(
         None,
         track_dirty_pages,
         vm_resources.vm_config.vcpu_count,
+        #[cfg(target_arch = "x86_64")]
+        seccomp_filters.clone(),
         cpu_template.kvm_capabilities.clone(),
     )?;
 
@@ -478,6 +484,8 @@ pub fn build_microvm_from_snapshot(
         uffd,
         vm_resources.vm_config.track_dirty_pages,
         vm_resources.vm_config.vcpu_count,
+        #[cfg(target_arch = "x86_64")]
+        seccomp_filters.clone(),
         microvm_state.vm_state.kvm_cap_modifiers.clone(),
     )?;
 
@@ -1155,6 +1163,8 @@ pub mod tests {
             uffd: None,
             vcpus_handles: Vec::new(),
             vcpus_exit_evt,
+            #[cfg(target_arch = "x86_64")]
+            seccomp_filters: crate::seccomp_filters::get_empty_filters(),
             resource_allocator: ResourceAllocator::new().unwrap(),
             mmio_device_manager,
             #[cfg(target_arch = "x86_64")]
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -124,11 +124,17 @@ use device_manager::resources::ResourceAllocator;
 use devices::acpi::vmgenid::VmGenIdError;
 use event_manager::{EventManager as BaseEventManager, EventOps, Events, MutEventSubscriber};
 use seccompiler::BpfProgram;
+#[cfg(target_arch = "x86_64")]
+use seccompiler::BpfThreadMap;
 use userfaultfd::Uffd;
 use utils::epoll::EventSet;
 use utils::eventfd::EventFd;
 use utils::terminal::Terminal;
 use utils::u64_to_usize;
+#[cfg(target_arch = "x86_64")]
+use vmm_config::hotplug::{HotplugVcpuConfig, HotplugVcpuError};
+#[cfg(target_arch = "x86_64")]
+use vmm_config::machine_config::{MachineConfigUpdate, MAX_SUPPORTED_VCPUS};
 use vstate::vcpu::{self, KvmVcpuConfigureError, StartThreadedError, VcpuSendEventError};
 
 use crate::arch::DeviceType;
@@ -317,6 +323,9 @@ pub struct Vmm {
     vcpus_handles: Vec<VcpuHandle>,
     // Used by Vcpus and devices to initiate teardown; Vmm should never write here.
     vcpus_exit_evt: EventFd,
+    // seccomp_filters are only needed in VMM for hotplugging vCPUS.
+    #[cfg(target_arch = "x86_64")]
+    seccomp_filters: BpfThreadMap,
 
     // Allocator for guest resrouces
     resource_allocator: ResourceAllocator,
@@ -598,6 +607,66 @@ impl Vmm {
             .collect::<Result<Vec<CpuConfiguration>, DumpCpuConfigError>>()?;
 
         Ok(cpu_configs)
+    }
+
+    /// Adds new vCPUs to VMM.
+    #[cfg(target_arch = "x86_64")]
+    pub fn hotplug_vcpus(
+        &mut self,
+        config: HotplugVcpuConfig,
+    ) -> Result<MachineConfigUpdate, HotplugVcpuError> {
+        use crate::logger::IncMetric;
+        if config.vcpu_count < 1 {
+            return Err(HotplugVcpuError::VcpuCountTooLow);
+        } else if self
+            .vcpus_handles
+            .len()
+            .checked_add(config.vcpu_count.into())
+            .ok_or(HotplugVcpuError::VcpuCountTooHigh)?
+            > MAX_SUPPORTED_VCPUS.into()
+        {
+            return Err(HotplugVcpuError::VcpuCountTooHigh);
+        }
+
+        // Create and start new vcpus
+        let mut vcpus = Vec::with_capacity(config.vcpu_count.into());
+
+        #[allow(clippy::cast_possible_truncation)]
+        let start_idx = self.vcpus_handles.len().try_into().unwrap();
+        for cpu_idx in start_idx..(start_idx + config.vcpu_count) {
+            let exit_evt = self
+                .vcpus_exit_evt
+                .try_clone()
+                .map_err(HotplugVcpuError::EventFd)?;
+            let vcpu =
+                Vcpu::new(cpu_idx, &self.vm, exit_evt).map_err(HotplugVcpuError::VcpuCreate)?;
+            vcpus.push(vcpu);
+        }
+
+        self.start_vcpus(
+            vcpus,
+            self.seccomp_filters
+                .get("vcpu")
+                .ok_or_else(|| HotplugVcpuError::MissingSeccompFilters("vcpu".to_string()))?
+                .clone(),
+        )
+        .map_err(HotplugVcpuError::VcpuStart)?;
+
+        #[allow(clippy::cast_lossless)]
+        METRICS.hotplug.vcpus_added.add(config.vcpu_count.into());
+
+        // Update VM config to reflect new CPUs added
+        #[allow(clippy::cast_possible_truncation)]
+        let new_machine_config = MachineConfigUpdate {
+            vcpu_count: Some(self.vcpus_handles.len() as u8),
+            mem_size_mib: None,
+            smt: None,
+            cpu_template: None,
+            track_dirty_pages: None,
+            huge_pages: None,
+        };
+
+        Ok(new_machine_config)
     }
 
     /// Retrieves the KVM dirty bitmap for each of the guest's memory regions.
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -616,12 +616,12 @@ impl Vmm {
         config: HotplugVcpuConfig,
     ) -> Result<MachineConfigUpdate, HotplugVcpuError> {
         use crate::logger::IncMetric;
-        if config.vcpu_count < 1 {
+        if config.add < 1 {
             return Err(HotplugVcpuError::VcpuCountTooLow);
         } else if self
             .vcpus_handles
             .len()
-            .checked_add(config.vcpu_count.into())
+            .checked_add(config.add.into())
             .ok_or(HotplugVcpuError::VcpuCountTooHigh)?
             > MAX_SUPPORTED_VCPUS.into()
         {
@@ -629,11 +629,11 @@ impl Vmm {
         }
 
         // Create and start new vcpus
-        let mut vcpus = Vec::with_capacity(config.vcpu_count.into());
+        let mut vcpus = Vec::with_capacity(config.add.into());
 
         #[allow(clippy::cast_possible_truncation)]
         let start_idx = self.vcpus_handles.len().try_into().unwrap();
-        for cpu_idx in start_idx..(start_idx + config.vcpu_count) {
+        for cpu_idx in start_idx..(start_idx + config.add) {
             let exit_evt = self
                 .vcpus_exit_evt
                 .try_clone()
@@ -653,7 +653,7 @@ impl Vmm {
         .map_err(HotplugVcpuError::VcpuStart)?;
 
         #[allow(clippy::cast_lossless)]
-        METRICS.hotplug.vcpus_added.add(config.vcpu_count.into());
+        METRICS.hotplug.vcpus_added.add(config.add.into());
 
         // Update VM config to reflect new CPUs added
         #[allow(clippy::cast_possible_truncation)]
--- a/src/vmm/src/logger/metrics.rs
+++ b/src/vmm/src/logger/metrics.rs
@@ -410,6 +410,10 @@ pub struct PutRequestsMetrics {
     pub drive_count: SharedIncMetric,
     /// Number of failures in attaching a block device.
     pub drive_fails: SharedIncMetric,
+    /// Number of PUTs for hotplugging
+    pub hotplug: SharedIncMetric,
+    /// Number of failures for hotplugging.
+    pub hotplug_fails: SharedIncMetric,
     /// Number of PUTs for initializing the logging system.
     pub logger_count: SharedIncMetric,
     /// Number of failures in initializing the logging system.
@@ -449,6 +453,8 @@ impl PutRequestsMetrics {
             boot_source_fails: SharedIncMetric::new(),
             drive_count: SharedIncMetric::new(),
             drive_fails: SharedIncMetric::new(),
+            hotplug: SharedIncMetric::new(),
+            hotplug_fails: SharedIncMetric::new(),
             logger_count: SharedIncMetric::new(),
             logger_fails: SharedIncMetric::new(),
             machine_cfg_count: SharedIncMetric::new(),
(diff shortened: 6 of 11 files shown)


src/vmm/src/lib.rs
```diff
@@ -600,6 +602,50 @@ impl Vmm {
...
+                .clone(),
+        )
+        .map_err(HotplugVcpuError::VcpuStart)?;
+
+        println!("Num vCPUs after hotplug: {}", self.vcpus_handles.len());
+
+        // Update VM config to reflect new CPUs added
+
+        // Resume VM
+        self.resume_vm().unwrap();
```
roypat, line 645: You don't actually need to resume the VM here. vCPU hotplug can happen on both paused and running virtual machines. If the VM is running, then the interrupt that Firecracker will inject will simply be immediately handled by the guest (by re-reading specific ACPI table entries). If its paused, then KVM will "buffer" the interrupt and insert it whenever we next call `KVM_RUN`. The vcpu hotplug handler here should be completely agnostic to whether the VM is paused or running, and should not try to modify its state.

src/vmm/src/lib.rs
```diff
@@ -600,6 +602,50 @@ impl Vmm {
...
+            vcpus,
+            seccomp_filters
+                .get("vcpu")
+                .unwrap()
+                // .ok_or_else(|| VmmError::SeccompFilters(()))?
+                .clone(),
+        )
+        .map_err(HotplugVcpuError::VcpuStart)?;
+
+        println!("Num vCPUs after hotplug: {}", self.vcpus_handles.len());
```
roypat, line 640: Here, we could have a metric that counts the actual number of hotplugged vCPUs. You could then use that metric as a first, simple integration test (e.g. have an integration test that calls the the hotplug endpoint and then does an assertion on the metric)

src/vmm/src/lib.rs
```diff
@@ -600,6 +602,50 @@ impl Vmm {
...
+            let exit_evt = self
+                .vcpus_exit_evt
+                .try_clone()
+                .map_err(HotplugVcpuError::EventFd)?;
+            let vcpu =
+                Vcpu::new(cpu_idx, &self.vm, exit_evt).map_err(HotplugVcpuError::VcpuCreate)?;
+            vcpus.push(vcpu);
+        }
+
+        let seccomp_filters = get_empty_filters();
```
roypat, line 629: Here, we should use the same seccomp filters as on the other vcpu threads. The problem here is Firecracker currently doesn't store the `BpfThreadMap` object post-boot, I think. In that case, we'll need to do some slight refactoring to store it and use it here

src/vmm/src/lib.rs
```diff
@@ -600,6 +602,50 @@ impl Vmm {
...
+        self.start_vcpus(
+            vcpus,
+            seccomp_filters
+                .get("vcpu")
+                .unwrap()
+                // .ok_or_else(|| VmmError::SeccompFilters(()))?
```
roypat, line 635: Let's avoid this unwrap and adjust the commented-out error handling

roypat (commented):
I think this looks very good already, nice job! I left some in-line comments about some specific part that can be improved. Some high level comments:

Every project has some slightly different rules and preferences about commit histories in pull requests, but in Firecracker we to prefer not have commits that fix/reformat code introduced in earlier commits. For example, the formatting bf738f59283d585eaf9c959aea4ed0384fdac3bc should be squashed into the commit that introduces the code you're reformatting here. As a rule of thumb, whenever you're just fixing some CI failure, squash the change into the commit that introduced the issue you're fixing. 
 
Beyond that, in this PR I would structure the commits as follows:
1. a commit that refactors the seccomp filter handling so that they are stored in `Vmm`
2. a commit that introduces the API endpoint, including a stub handler in `rpc_interface.rs` (if you want, you can include the unit tests in this one, or add a separate commit for them)
3. a commit that implements `Vmm::hotplug_vcpus` (with or without the unit tests, again up to you)

I don't really have any fixed rules as for how I came up with that, but generally I try to have my commits be a single logical unit of code that is still independently useful. That's why I would put the seccomp refactor into a commit separate from the `hotplug_vcpus` function (as they're two units of code to me, a refactor + a feature), but I would combine your first two commits that introduce the API endpoints in two steps (because having the endpoint without parsing the json would feel a bit too fine-grained to me). But it's not an exact science, so feel free to deviate from whatever I'm suggesting if you disagree with it!

Lastly, could you drop Pablo's commits from this PR? We can just rebase the feature branch onto main to pick them up :)

jxcurtis:
Thanks for the feedback, still trying to fix some issues Buildkite is picking up on but will work on this as soon as I'm done with it :)

src/firecracker/src/api_server/request/hotplug.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    use super::super::*;
+
+    #[test]
+    fn test_parse_put_hotplug() {
+        // Case 1. Invalid resource type
+        parse_put_hotplug(&Body::new("invalid body"), Some("invalid")).unwrap_err();
+
+        // Case 2. No resource type
+        parse_put_hotplug(&Body::new("invalid body"), None).unwrap_err();
```
roypat, line 57: nit: Since these are explicitly testing incorrect values for the "resource type" parameter, we should use valid bodies, to ensure that the error we are receiving is actually related to the resource type (and not some other invalid configuration we happen to also pass). Alternatively, you could do some assertion on the returned error (e.g. by checking the error variant is, say, `InvalidPathMethod` in the first case, and looking at the method/error string)

tools/create_snapshot_artifact/main.py
```diff
@@ -24,7 +24,6 @@
 from framework.utils_cpuid import CpuVendor, get_cpu_vendor
 from host_tools.cargo_build import get_firecracker_binaries
 
-
```
roypat, line 27: nit: accidental change?

src/vmm/src/builder.rs
```diff
@@ -1155,6 +1160,7 @@ pub mod tests {
             uffd: None,
             vcpus_handles: Vec::new(),
             vcpus_exit_evt,
+            seccomp_filters: crate::seccomp_filters::get_empty_filters(),
```
roypat, line 1163: nit: even if its just a handful of lines, the new seccomp argument could still be in its own refactoring commit :)

src/vmm/src/lib.rs
```diff
@@ -600,6 +603,50 @@ impl Vmm {
...
+        )
+        .map_err(HotplugVcpuError::VcpuStart)?;
+
+        #[allow(clippy::cast_lossless)]
+        METRICS
+            .hotplug
+            .vcpu_hotplug_count
+            .add(config.vcpu_count as u64);
+
+        // Update VM config to reflect new CPUs added
```
roypat, line 645: This part seems to be missing? :/

src/vmm/src/lib.rs
```diff
@@ -600,6 +603,50 @@ impl Vmm {
         Ok(cpu_configs)
     }
 
+    /// Adds new vCPUs to VMM.
+    pub fn hotplug_vcpus(&mut self, config: HotplugVcpuConfig) -> Result<(), HotplugVcpuError> {
+        #[allow(clippy::cast_possible_truncation)]
+        if config.vcpu_count < 1 {
+            return Err(HotplugVcpuError::VcpuCountTooLow);
+        } else if self.vcpus_handles.len() as u8 + config.vcpu_count > MAX_SUPPORTED_VCPUS {
```
roypat, line 611: This _could_ overflow and wrap around, e.g. if `self.vcpu_handles.len() == 2` and `config.vcpu_count == 255`! Let's add a test case for this and fix it :)

src/vmm/src/rpc_interface.rs
```diff
@@ -2052,6 +2074,27 @@ mod tests {
...
+            VmmActionError::HotplugVcpu(HotplugVcpuError::VcpuCountTooLow),
+        );
+
+        let req = VmmAction::HotplugVcpu(HotplugVcpuConfig { vcpu_count: 33 });
+        check_runtime_request_err(
+            req,
+            VmmActionError::HotplugVcpu(HotplugVcpuError::VcpuCountTooHigh),
+        );
+    }
+
```
roypat, line 2097: As we discussed offline, this test actually only tests the big `match` statement at the top of this file, but never actually calls your `vcpu_hotplug` method. That's a preexisting issue with the code, so obviously not something you'll need to (or should!) fix here. However, it means that we can probably remove this entire test (and any interaction you've written wit the `rpc_interface.rs` mocking stuff).

Instead, lets write a unit test in `lib.rs` that uses `default_vmm` and then calls the actual `Vmm::hotplug_vcpu`.

For posterity, because James and me were wondering for a while: How does this test even pass, if `VcpuCountTooLow` is only returned in one location, and that location is reported by codecov as uncovered? The answer is that the mocks consider two `VmmActionError`s to be equal if they are the same variant, without actually taking into consideration what the variant contains (so `VmmActionError::HotplugVcpu(HotPlugVcpuError::VcpuCountTooLow)` compares equal to `VmmActionError::HotplugVcpu(HotPlugVcpuError::VcpuCountTooHigh)`)

src/vmm/src/vmm_config/hotplug.rs
```diff
@@ -1,19 +1,20 @@
 // Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-use serde::{Deserialize, Serialize};
 use std::io;
+
+use serde::{Deserialize, Serialize};
+
```
roypat, line 7: unintentional change?

src/vmm/src/vmm_config/hotplug.rs
```diff
@@ -1,19 +1,20 @@
...
 use crate::vstate::vcpu::VcpuError;
 use crate::StartVcpusError;
 
 /// Firecracker aims to support small scale workloads only, so limit the maximum
 /// vCPUs supported.
-pub const MAX_SUPPORTED_VCPUS: u8 = 32;
```
roypat: nit: this should already not have been introduced in the first commit :) Try to avoid adding code in one commit and then deleting it again in a follow up (unless necessary of course!)

roypat (commented):
Nice work! I've left some minor comments, mainly the one about testing that we talked about offline.

tools/create_snapshot_artifact/main.py
jxcurtis: Nope, CI was failing due to this extra line. Separated change to its own commit for clarity

src/vmm/src/builder.rs
jxcurtis: Seccomp argument given its own refactoring commit

src/vmm/src/lib.rs
jxcurtis: Added now :)

src/vmm/src/vmm_config/hotplug.rs
```diff
@@ -0,0 +1,32 @@
...
+    VcpuStart(StartVcpusError),
+}
+
+/// Config for hotplugging vCPUS
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct HotplugVcpuConfig {
+    /// Number of vcpus to start.
+    pub vcpu_count: u8,
+}
```
ShadowCurse, line 36: If we will have 1 endpoint for all hot plugging I would prefer to have 1 enum that represents all options. Smth like this:
```rust
enum HotplugRequest {
  Vcpu(HotplugVcpu),
  // Memory(HotplugMemory),
}
```
This better represents available options. Additionally way we don't need to do any additional checks like you do in the `request/hotplug.rs`.

src/vmm/src/builder.rs
```diff
@@ -309,6 +312,7 @@ pub fn build_microvm_for_boot(
         None,
         track_dirty_pages,
         vm_resources.vm_config.vcpu_count,
+        seccomp_filters.clone(),
```
ShadowCurse, line 315: I would check the reason why `seccomp_filters` is passed here by reference. Most likely it can be passed by value, so you don't need to clone the thing.

jxcurtis: Yes, Patrick and I looked at this and we agree that ownership could be transferred to the VMM. It just requires changes across more of the codebase and has a minimal impact either way, so we decided to clone for the time being.

ShadowCurse: I think the time investment here is quite small and at the same time, if this is not fixed now, it most likely will never be fixed.

src/vmm/src/vmm_config/hotplug.rs
jxcurtis: As you suggested, I've added enums above the vCPU specific ones in order to unify vCPU hotplugging with other resources (should they be implemented). Let me know if it looks okay

ShadowCurse: I don't see any changes

src/vmm/src/builder.rs
jxcurtis: Created a low priority issue to track this: #4647

src/vmm/src/vmm_config/hotplug.rs
jxcurtis: Can you clarify what you were looking for? I've left the existing enums as is, but have added an extra enum above `HotplugRequestConfig`, which would unify the two different types of request into one enum. The result of this is that there is a single VmmAction for hotplugging, `VmmAction::HotplugRequest(HotplugRequestConfig)`, and at this point the handling of the different hotplug requests is handled.

ShadowCurse: Ok, I see this now. For some reason if I look at the commits 1 by 1 I don't see the change, but if I look at all changed files it is there.

src/firecracker/src/api_server/request/hotplug.rs
```diff
@@ -0,0 +1,100 @@
...
+    METRICS.hotplug.vcpu_hotplug_request_count.inc();
+    let config = serde_json::from_slice::<HotplugVcpuConfig>(body.raw()).map_err(|err| {
+        METRICS.hotplug.vcpu_hotplug_request_fails.inc();
+        err
+    })?;
+
+    Ok(ParsedRequest::new_sync(VmmAction::HotplugRequest(
+        HotplugRequestConfig::Vcpu(config),
+    )))
+}
```
ShadowCurse, line 41: And now when we have `HotplugRequestConfig` we don't need to look at the `resource_type_from_path`. We can just deserialize the request directly `serde_json::from_slice::<HotplugRequestConfig>(body.raw())` and send it as a `VmmAction`.

src/vmm/src/lib.rs
```diff
@@ -601,6 +609,66 @@ impl Vmm {
...
+    pub fn hotplug_vcpus(
+        &mut self,
+        config: HotplugVcpuConfig,
+    ) -> Result<MachineConfigUpdate, HotplugVcpuError> {
+        use crate::logger::IncMetric;
+        #[allow(clippy::cast_possible_truncation)]
+        if config.vcpu_count < 1 {
+            return Err(HotplugVcpuError::VcpuCountTooLow);
+        } else if config.vcpu_count > MAX_SUPPORTED_VCPUS
+            || self.vcpus_handles.len() as u8 + config.vcpu_count > MAX_SUPPORTED_VCPUS
```
ShadowCurse, line 623: `self.vcpus_handles.len() as u8 + config.vcpu_count > MAX_SUPPORTED_VCPUS` check should be enough here.

ShadowCurse (commented):
Overall LGTM

src/vmm/src/lib.rs
roypat: No, that can overflow, see https://github.com/firecracker-microvm/firecracker/pull/4638#discussion_r1644272678

src/vmm/src/rpc_interface.rs
```diff
@@ -2045,6 +2148,47 @@ mod tests {
         );
     }
 
+    #[test]
+    #[cfg(target_arch = "x86_64")]
+    fn test_runtime_hotplug_vcpu() {
```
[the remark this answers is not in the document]
jxcurtis: True, I initially put it into `rpc_interface.rs` as that's where the other VMM actions are also tested and there are no tests currently residing in `lib.rs`. Happy to make the change if it makes more sense though

src/vmm/src/lib.rs
ShadowCurse: Maybe in this case we just need to cast not to `u8` but to smth bigger?

src/vmm/src/vmm_config/hotplug.rs
```diff
@@ -0,0 +1,48 @@
...
+
+use crate::vstate::vcpu::VcpuError;
+use crate::StartVcpusError;
+/// Unifying enum for all types of hotplug request configs.
+/// Currently only Vcpus may be hotplugged.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum HotplugRequestConfig {
+    /// Vcpu hotplug request
+    Vcpu(HotplugVcpuConfig),
```
bchalios, line 16: Do we need the extra `HotplugVcpuConfig` type here? Wouldn't a `u8` be enough?

src/vmm/src/vmm_config/hotplug.rs
```diff
@@ -0,0 +1,48 @@
...
+    VcpuCreate(VcpuError),
+    /// Failed to start vCPUs
+    VcpuStart(StartVcpusError),
+}
+
+/// Config for hotplugging vCPUS
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct HotplugVcpuConfig {
+    /// Number of vcpus to start.
```
bchalios, line 48: This comment needs a bit of clarification. Does `vcpu_count` represent the amount of vCPUs the VM will end up with after (successfully) handling this request? If yes, then we might not "start" any vCPUs. We might actually shut down some.

src/vmm/src/vmm_config/mod.rs
```diff
@@ -20,6 +20,8 @@ pub mod boot_source;
 pub mod drive;
 /// Wrapper for configuring the entropy device attached to the microVM.
 pub mod entropy;
+/// Wrapper over hotplug configuration.
+pub mod hotplug;
```
bchalios, line 25: I assume this needs to be `[cfg(target_arch = "x86_64")]`. Please bring it over to the first commit.

src/firecracker/src/api_server/request/hotplug.rs
```diff
@@ -22,7 +22,7 @@ mod tests {
 
     use hotplug::parse_put_hotplug;
     use vmm::rpc_interface::VmmAction;
-    use vmm::vmm_config::hotplug::{HotplugVcpuConfig, HotplugRequestConfig};
+    use vmm::vmm_config::hotplug::{HotplugRequestConfig, HotplugVcpuConfig};
```
bchalios, line 25: I assume you fixed here the ordering of the types (there appears a change in the third commit). Please fix the errors in the commits that introduced them.

src/vmm/src/lib.rs
```diff
@@ -317,6 +321,10 @@ pub struct Vmm {
     vcpus_handles: Vec<VcpuHandle>,
     // Used by Vcpus and devices to initiate teardown; Vmm should never write here.
     vcpus_exit_evt: EventFd,
+    // seccomp_filters are only needed in VMM for hotplugging vCPUS.
+    // Since this is something that can only occur on x86_64, we need to allow
+    // dead code for aarch64.
+    #[allow(dead_code)]
```
bchalios, line 327: Does it make sense to instead use `[cfg(target_arch = "x86_64")]`? I think I would prefer it, unless it pollutes the code with too many conditional compilation flags.

src/vmm/src/lib.rs
bchalios: Or just do. `self.vcpus_handles.len().checked_add(config.vcpu_count)? > MAX_SUPPORTED_VCPUS`.

src/vmm/src/lib.rs
```diff
@@ -601,6 +609,66 @@ impl Vmm {
...
+            || self.vcpus_handles.len() as u8 + config.vcpu_count > MAX_SUPPORTED_VCPUS
+        {
+            return Err(HotplugVcpuError::VcpuCountTooHigh);
+        }
+
+        // Create and start new vcpus
+        let mut vcpus = Vec::with_capacity(config.vcpu_count as usize);
+
+        #[allow(clippy::cast_possible_truncation)]
+        let start_idx = self.vcpus_handles.len() as u8;
```
bchalios, line 632: let's avoid `as` in favour of `try_into().unwrap()`.

src/vmm/src/lib.rs
```diff
@@ -601,6 +609,66 @@ impl Vmm {
...
+        if config.vcpu_count < 1 {
+            return Err(HotplugVcpuError::VcpuCountTooLow);
+        } else if config.vcpu_count > MAX_SUPPORTED_VCPUS
+            || self.vcpus_handles.len() as u8 + config.vcpu_count > MAX_SUPPORTED_VCPUS
+        {
+            return Err(HotplugVcpuError::VcpuCountTooHigh);
+        }
+
+        // Create and start new vcpus
+        let mut vcpus = Vec::with_capacity(config.vcpu_count as usize);
```
bchalios, line 629: `config.vcpu_count.into()` instead of `as`

src/vmm/src/lib.rs
```diff
@@ -601,6 +609,66 @@ impl Vmm {
...
+                .ok_or_else(|| HotplugVcpuError::MissingSeccompFilters("vcpu".to_string()))?
+                .clone(),
+        )
+        .map_err(HotplugVcpuError::VcpuStart)?;
+
+        #[allow(clippy::cast_lossless)]
+        METRICS
+            .hotplug
+            .vcpu_hotplug_count
+            .add(config.vcpu_count as u64);
```
bchalios, line 656: again, can we please avoid `as`?

bchalios (changes requested):
Hey James, thanks for the PR and apologies for the delayed review. Overall, great work! 

I 've left some comments inline. Some of them are open for discussion, e.g. the actual API we should use for this. These things can be changed in subsequent PRs. But I'd love to hear your thoughts.

Others, I 'd like them fixed in this PR. Mainly this includes:
* Fix errors introduced by a commit in the commit that introduced them, rather than later on
* I'd like as to avoid using `as` in favour of explicit `into()` or `try_into().unwrap()`, where applicable.

Also, could you please update the swagger file with the definition of the new API whichever we decide it is.

src/vmm/src/vmm_config/hotplug.rs
bchalios: I see now that we only support adding new vCPUs. It would be nice to at least document that somewhere, ideally within the code itself. I would be in favour of having the request type look like this:

```rust
pub enum HotplugRequestConfig {
    /// Vcpu hotplug request
    AddVcpus(u8),
}
```

src/vmm/src/vmm_config/hotplug.rs
jxcurtis: I agree that it feels a bit OTT. The reason I did it this way was to utilise the serde_json `Deserialize` attribute, in-line with what all the other endpoints have done, even those with one attribute (see `MetricsConfig`, `BalloonUpdateConfig`, `BalloonUpdateStatsConfig`). Happy to change it if you think it makes more sense done alternatively.

src/vmm/src/vmm_config/hotplug.rs
jxcurtis: At the moment, the value supplied is the number of vCPUs to be added. I'm not sure whether to change this to setting the total number of vCPUs, as this could be used to down VMs as well.

src/firecracker/src/api_server/parsed_request.rs
```diff
@@ -89,6 +90,8 @@ impl TryFrom<&Request> for ParsedRequest {
             (Method::Put, "boot-source", Some(body)) => parse_put_boot_source(body),
             (Method::Put, "cpu-config", Some(body)) => parse_put_cpu_config(body),
             (Method::Put, "drives", Some(body)) => parse_put_drive(body, path_tokens.next()),
+            #[cfg(target_arch = "x86_64")]
+            (Method::Put, "hotplug", Some(body)) => parse_put_hotplug(body),
```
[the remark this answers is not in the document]
jxcurtis: I also think that this could make sense. It would require some change, as there already exists PATCH for machine config, and therefore hotplug code would need to be changed to fit into what already exists. My main initial concern is that, currently, the logic of hotplugging is part of the `Vmm` struct, whereas the `machine-config` PATCH is handled in the VmResources struct. The current hotplugging logic uses both the seccomp filters and vcpu_exit_evt attributes of the Vmm, and so if it were to be moved into VmResources there would need to be some refactoring in order to achieve the same effect. I haven't scoped out how much change would be required to do this, but I feel as if this is something I can revisit later, as it likely won't change the performance of hotplugging, which is the main interest.

src/vmm/src/logger/metrics.rs
```diff
@@ -410,6 +410,10 @@ pub struct PutRequestsMetrics {
     pub drive_count: SharedIncMetric,
     /// Number of failures in attaching a block device.
     pub drive_fails: SharedIncMetric,
+    /// Number of PUTs for hotplugging
+    pub hotplug: SharedIncMetric,
+    /// Number of failures for hotplugging.
+    pub hotplug_fails: SharedIncMetric,
```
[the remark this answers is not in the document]
jxcurtis: This is merely for the API call, and is in-line with metrics for other calls. I've also added metrics which track the number of vCPUs hotplugged (see `vcpu_hotplug_request_count`), and so if new resources are introduced, similar metrics can be introduced for them.

Some of the hotplug metrics I've added also seem rather useless (i.e. `vcpu_hotplug_request_fails`), and I haven't actually made use of them. I'll push a change that removes that shortly.

src/firecracker/src/api_server/parsed_request.rs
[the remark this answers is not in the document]
jxcurtis: This change would also clean up the updating of the machine config after hotplugging. Right now it's a bit a messy, as it's essentially another API call's worth of work without actually using the API. Not the neatest bit of code.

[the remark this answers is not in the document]
ShadowCurse: I would disagree. The `PATCH` for `machine-config` seems worse, because not all fields can be patched (we cannot turn off smt or change cpu template on the fly). Separate API endpoint feels more correct to me. I don't want to have same story as with block API where we need to decide which fields can be present, which cannot and so on.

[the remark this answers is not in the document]
bchalios: My motivation is "cleaniness" of the API rather than easiness of the implementation, but I don't see the problem with adding a check on the fields that are being updated. However, if you think this is a problem we could even have a different body for the request, which will include only the fields that can be updated. We do that with various other devices, e.g. [block](https://github.com/firecracker-microvm/firecracker/blob/f5906eeb47f71f1ba412b24786d6762fc06dcbf4/src/firecracker/swagger/firecracker.yaml#L1141), [net](https://github.com/firecracker-microvm/firecracker/blob/f5906eeb47f71f1ba412b24786d6762fc06dcbf4/src/firecracker/swagger/firecracker.yaml#L1156).
 
In any case, as I said, I don't see this as a blocker for this PR.

src/vmm/src/vmm_config/hotplug.rs
bchalios: Ok, speaking with @roypat indeed the `u8` field won't fly due to how serde parses enums from JSON. You were right James.

src/vmm/src/vmm_config/hotplug.rs
bchalios: Since using the `u8` thing doesn't work (see previous comment), this is fine. Resolving.

src/vmm/src/logger/metrics.rs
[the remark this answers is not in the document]
jxcurtis: @bchalios is it worth making hotplug metrics compile only for x86?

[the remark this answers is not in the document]
roypat: That might make updating the metrics integration test quite difficult :sweat_smile:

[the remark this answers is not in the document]
jxcurtis: Fair point. They don't cause any trouble as they are now, so perhaps just leave them?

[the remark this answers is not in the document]
bchalios: yup, I don't think there's a problem with having the metrics there on Aarch64. There will just be 0.

src/firecracker/swagger/firecracker.yaml
```diff
@@ -282,6 +282,30 @@ paths:
...
+          in: body
+          description: The resource type and amount to be hotplugged.
+          required: true
+          schema: 
+            $ref: "#/definitions/Hotplug"
+      responses:
+        204:
+          description: Resources hotplugged successfully.
+        400: 
+          description: Resources cannot be hotplugged due to bad input.
```
bchalios, line 301: Can the errors be only due to bad input?

src/firecracker/swagger/firecracker.yaml
```diff
@@ -282,6 +282,30 @@ paths:
           schema:
             $ref: "#/definitions/Error"
 
+  /hotplug:
+    put:
+      summary: Used to hot plug resources into the microVM
```
bchalios, line 287: Please, use active voice here, similarly to the other endpoints.

Suggestion:

```suggestion
      summary: Hotplugs resources to the microVM.
```

Also, is this endpoint post-boot only? I assume it should be, right?

src/firecracker/swagger/firecracker.yaml
```diff
@@ -943,6 +967,17 @@ definitions:
       vsock:
         $ref: "#/definitions/Vsock"
 
+  Hotplug:
+    type: object
+    description:
+      Defines the configuration for hot plugging.
```
bchalios, line 973: shouldn't `vcpu_count` be required here?

```suggestion
      Defines the configuration for hot plugging.
    required:
      - vcpu_count
```

src/firecracker/swagger/firecracker.yaml
```diff
@@ -943,6 +967,17 @@ definitions:
...
+    properties:
+      vcpu_count:
+        type: integer
+        minimum: 1
```
bchalios, line 977: what is the purpose of `minimum` here?

src/vmm/src/builder.rs
```diff
@@ -145,13 +145,15 @@ impl std::convert::From<linux_loader::cmdline::Error> for StartMicrovmError {
...
 #[cfg_attr(target_arch = "aarch64", allow(unused))]
+#[allow(clippy::too_many_arguments)]
 fn create_vmm_and_vcpus(
     instance_info: &InstanceInfo,
     event_manager: &mut EventManager,
     guest_memory: GuestMemoryMmap,
     uffd: Option<Uffd>,
     track_dirty_pages: bool,
     vcpu_count: u8,
+    #[cfg(target_arch = "x86_64")] seccomp_filters: BpfThreadMap,
```
bchalios, line 156:
```suggestion
    #[cfg(target_arch = "x86_64")]
    seccomp_filters: BpfThreadMap,
```

bchalios (changes requested):

src/firecracker/swagger/firecracker.yaml
jxcurtis: Not necessarily. The way the JSON is being deserialized is 'untagged', so that the resource that is being requested does not have to be identified, it merely matches the provided fields with one of the `HotplugResourceConfig` structs that are part of the `HotplugRequestConfig` enum. This means that if `HotplugMemoryConfig` struct is created, with field `mem_size_mib`, a model request would be:
```
{ "mem_size_mib": 128}
```
Perhaps this should be changed for clarity, to include the type in the request:
```
{ "type": "vcpu", "vcpu_count": 8 }
```
Opinions?

src/firecracker/swagger/firecracker.yaml
jxcurtis: Just identifies that `vcpu_count` can't be 0 or less.

src/firecracker/swagger/firecracker.yaml
jxcurtis: Yes, post-boot only. I'll add that to the summary, in line with what other endpoints have.

src/vmm/src/builder.rs
jxcurtis: Rust's formatter seems to disagree with this one. Rust-analyzer kept moving it whenever I tried to save, and even when I turned it off, the pre-commit hook won't allow me to commit without it on the same line.

src/firecracker/swagger/firecracker.yaml
ShadowCurse: I think the way it work now is fine. Adding additional `type` only adds redundant info. But actually I thought deserialization will require request to look like this: 
```json
{ "Vcpu": { "vcpu_count": 1 } }
```

jxcurtis: There are a few different ways to handle it, see: https://serde.rs/enum-representations.html

ShadowCurse: Ok, nice to know. I would leave this with default settings as we don't use any `tag` modifications in FC. In any case this can be changed in the future.

jxcurtis: Just to clarify, when you say default settings, are you referring to how you have it above? With Vcpu externally tagged?

ShadowCurse: yes. Also in this case maybe we can rename `vcpu_count` to just `add`, but it is just a thought, as I said, all of this can be changed later if needed.

src/firecracker/swagger/firecracker.yaml
bchalios: In that case, I suggest we update the description here.

src/firecracker/swagger/firecracker.yaml
bchalios: > Not necessarily. The way the JSON is being deserialized is 'untagged', so that the resource that is being requested does not have to be identified, it merely matches the provided fields with one of the HotplugResourceConfig structs that are part of the HotplugRequestConfig enum.

So what happens if the user does not provide anything here? Will it get translated to a `vcpu_count` equal to 0?

src/firecracker/swagger/firecracker.yaml
jxcurtis: Will do

src/firecracker/swagger/firecracker.yaml
jxcurtis: > > Not necessarily. The way the JSON is being deserialized is 'untagged', so that the resource that is being requested does not have to be identified, it merely matches the provided fields with one of the HotplugResourceConfig structs that are part of the HotplugRequestConfig enum.
> 
> So what happens if the user does not provide anything here? Will it get translated to a `vcpu_count` equal to 0?

I believe serde would throw an error, as it would not be able to match to any of the configs. However, I've just pushed a change to update it to use default settings, as suggested by @ShadowCurse.

src/firecracker/swagger/firecracker.yaml
jxcurtis: Updated now. Decided to leave it just as "bad input or VMM error", as I imagine users are not interested in the specific VMM-related errors that could occur.

bchalios (approved):

roypat (approved):

roypat:
:tada:

PR #4640, opened 2024-06-11
fix(test_rng): do not test for virtio-rng performance

bchalios:
## Changes

Modify `test_rng.py::test_rng_bw_rate_limiter` to not check if the device achieves the performance defined by the devices rate limiter limit.

## Reason

Currently, our test for virtio-rng rate limiter tests if the performance obtained by the device is within 10% of the rate limiter limit. This does not only test whether the rate limiter works (achieved bandwidth > 10% of limit), but also if the device performs at that same limit.

This test is supposed to be a functional test, not a performance one. In fact, it does not run in isolation. Moreover, we have not thoroughly profiled the performance we can achieve from the device, so we cannot be sure that the limits in the functional test correspond to the capabilities of the device. If we want, we should do that as part of a performance test.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/tests/integration_tests/functional/test_rng.py
+++ b/tests/integration_tests/functional/test_rng.py
@@ -149,14 +149,29 @@ def _check_entropy_rate_limited(ssh, random_bytes, expected_kbps):
     """
     Ask for `random_bytes` from `/dev/hwrng` in the guest and check
     that achieved throughput does not exceed the expected throughput by
-    more than 10%.
+    more than 2%.
 
-    NOTE: 10% is an arbitrarily selected limit which should be safe enough,
-    so that we don't run into many intermittent CI failures.
+    NOTE: 2% is accounting for the initial credits available in the buckets
+    which can be consumed immediately. In the `dd` command we read `size * 100`
+    bytes, where `size` is the size of the bucket. As a result, the first
+    `size` bytes will be read "immediately" and the remaining `99 * size` bytes
+    will be read at a rate of `size / refill_time`. So, the total test runtime
+    will be `99 * refill_time`. That helps us calculate the expected throughput
+    allowed from our rate limiter like this:
+
+    size * 100 / (99 * refill_time) =
+    (100 / 99) * (size / refill_time) =
+    (100 / 99) * expected_throughput_rate =
+    1.01 * expected_throughput_rate
+
+    (kudos to @roypat for this analysis)
+
+    So, we should expect a 1% margin from the expected throughput. We use 2%
+    for accounting for rounding/measurements errors.
     """
     measured_kbps = _get_throughput(ssh, random_bytes)
     assert (
-        _get_percentage_difference(measured_kbps, expected_kbps) <= 10
+        _get_percentage_difference(measured_kbps, expected_kbps) <= 2
     ), "Expected {} KB/s, measured {} KB/s".format(expected_kbps, measured_kbps)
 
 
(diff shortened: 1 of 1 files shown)


tests/integration_tests/functional/test_rng.py
```diff
@@ -84,7 +84,7 @@ def _get_percentage_difference(measured, base):
     if measured == base:
         return 0
     try:
-        return (abs(measured - base) / base) * 100.0
+        return ((measured - base) / base) * 100.0
```
kalyazin, line 87: why do we need the 10% margin actually? Should the measured value not be less or equal to the desired?

bchalios: you are right! So, my experiments show that some times we overshoot. See here for example where I reduced the limit to 1%: https://buildkite.com/firecracker/firecracker-pr/builds/10042

We seem to be getting 102KB/sec when the limit is 100KB/sec. So there is some margin there, which is obviously smaller than 10%. I think, I initially had 10% there because that's the kind of failures I was seeing on the other way of the comparison (90KB/sec), so I added the same range both ways.

In fact, now I changed this to 5% just to avoid intermittent issues.

Does that make sense to you?

roypat (dismissed):

pb8o (dismissed):

tests/integration_tests/functional/test_rng.py
kalyazin: What is the intuition behind this excess? Does it look like we genuinely allow for slightly more traffic than the limiter is supposed to enforce? Or is this some kind of a rounding error?
I think it would be nice to find a feasible explanation and leave a comment in the test about it.
I don't want to block the PR due to this, because the change is an improvement regardless.

kalyazin (approved):

tests/integration_tests/functional/test_rng.py
roypat: I think one source is that the bucket starts off as full. So within the first time delta ε, we get an instantaneous theoretical throughput of (bucket size) / ε, and only after that do we hit (bucket size)/(refill time). The effect of this will be larger the smaller ε is, and is then amortized over the runtime of the test. 

We can test this theory by increasing the runtime of the test - the longer the test, the more diluted this initial effect should be.

kalyazin: That sounds plausible indeed. Would that be sensible to obtain more than one datapoint in the test (a long and a short one) to confirm that? Also since we know the duration of the test and the depth of the bucket, can we calculate the margin based on that instead of picking it randomly/empirically?

roypat: So we do `size * 100` many random bytes, which means `size` random bytes are read immediately, and then `size * 99` random bytes are read at a rate of `size / refill_time`. So in total the test runtime should be `99 * refill_time`, so the throughput will be `size * 100 / (99 * refill_time) = (100 / 99) * (size / refill_time) = (100/99) * expected_throughput_rate`. 100 / 99 is 1.01(01), so 1.01(01)% excess. So I'd say 2% should be a safe enough threshold to account for rounding/measurement errors

bchalios: 20 subsequent runs on all instance types and kernel combinations pass successfully, for a limit of 2%: https://buildkite.com/firecracker/bchalios-ad-hoc-tests/builds/42#_.

I pushed a version which adds @roypat's math in the comment. @kalyazin @pb8o could you take a look when you have time?

kalyazin (approved):

roypat (approved):

PR #4590, opened 2024-05-01
feat: store snapshot files on test fail

ShadowCurse:
## Changes
If test uses snapshots and fails save most files in `root` to the `test_results`.

## Reason
If CI test fails due to snapshot problems it will be easier to reproduce with original snapshot files.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following Developer
Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this
  PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet
  [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

______________________________________________________________________

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #4641, opened 2024-06-11
Delete intermediary snapshots created by tests, to avoid out-of-disk-space errors during artifact generation

roypat:
## Changes

Delete intermediary snapshot while when tests create snapshots in loops.

## Reason

Tests such as `test_snapshot_ab.py` or `test_5_snapshots` create snapshot files in a loop. This interacts badly with some debugging infrastructure added in #4590: If a test fails, then all files from all microVM chroots from that test are copied to `test_results`, and then uploaded as run artifacts. This means that if one of the 30 microVM restorations in `test_snapshot_ab.py` fails, we copy all, potentially 30, snapshot files to `test_results`. If these are large (which they can be: the test goes up to 12GB microVMs), this can cause the buildkite agent to run out of disk space.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/tests/integration_tests/performance/test_snapshot_ab.py
+++ b/tests/integration_tests/performance/test_snapshot_ab.py
@@ -89,7 +89,7 @@ class SnapshotRestoreTest:
                 monitor_memory=False,
             )
             microvm.spawn()
-            microvm.restore_from_snapshot(snapshot, resume=True)
+            snapshot_copy = microvm.restore_from_snapshot(snapshot, resume=True)
 
             fcmetrics = FCMetricsMonitor(microvm)
             fcmetrics.start()
@@ -108,6 +108,7 @@ class SnapshotRestoreTest:
             values.append(value)
             fcmetrics.stop()
             microvm.kill()
+            snapshot_copy.delete()
 
         snapshot.delete()
         return values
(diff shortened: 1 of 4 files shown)


tests/integration_tests/performance/test_snapshot_ab.py
```diff
@@ -108,6 +108,7 @@ def sample_latency(
             values.append(value)
             fcmetrics.stop()
             microvm.kill()
```
pb8o, line 110: Should we also remove the chroot in addition or does this cover the majority of the disk space?

pb8o (approved):

tests/integration_tests/performance/test_snapshot_ab.py
roypat: I went back in our commit history, and apparently we used to delete the chroot, but it was causing problems where other parts of the framework were still expecting the chroot itself to exist. So I went ahead and only deleted the snapshot for this, which is indeed the largest contributor (everything else is at most a handful of MB)

pb8o (approved):

kalyazin (approved):

PR #4642, opened 2024-06-14
Improve observability into failing tests

roypat:
## Changes

The spiritual successor to "Misc Changes 4". The meat of this PR are two small changes in `utils.run_cmd_sync` that aim to improve output if a command fails (e.g. `ChildProcessError` is raised), or times out (e.g. `TimeoutExpired` is raised). 

Before this patch series, if a test failed due to `TimeoutExpired`, we would only be provided this fact, along with a backtrace. We change this to include the stdout and stderr of the command before it timed out in the pytest log.

Second, previously if a command returned a non-zero return code, and the `ChildProcessError` raised resulted in termination of the test run, pytest would truncate the exception message, which often lead to stdout/stderr of the failing command to become truncated or even omitted. Similarly to the above case, we now always log stdout and stderr in full. 

After making `run_cmd_sync` nicely provide a lot of debugging data, we could change a lot of ad-hoc assertions on error code (especially when it comes to ssh commands) to just rely on `run_cmd_sync` to handle exception raising and logging. This is where the PR gets its large-ish diff from.

Lastly, we update the whole `run_cmd` API in `utils` by inverting the default behavior of raising exceptions to align it with the APIs offered by `SSHConnection` and the `subprocess` module.

## Reason

I got a bit annoyed when I couldn't read some critical stderr output from a rarely reproducing intermittent CI failure because pytest truncated it.

... and then it kinda just spiraled from there into fixing up other things that have been bothering me for the better part of a year

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/tests/framework/utils.py
+++ b/tests/framework/utils.py
@@ -374,7 +374,7 @@ def _format_output_message(proc, stdout, stderr):
     return output_message
 
 
-def run_cmd(cmd, check=False, no_shell=False, cwd=None, timeout=None) -> CommandReturn:
+def run_cmd(cmd, check=False, shell=True, cwd=None, timeout=None) -> CommandReturn:
     """
     Execute a given command.
 
@@ -385,7 +385,7 @@ def run_cmd(cmd, check=False, no_shell=False, cwd=None, timeout=None) -> Command
     :param timeout: Time before command execution should be aborted with a `TimeoutExpired` exception
     :return: return code, stdout, stderr
     """
-    if isinstance(cmd, list) or no_shell:
+    if isinstance(cmd, list) or not shell:
         # Create the async process
         proc = subprocess.Popen(
             cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=cwd
@@ -423,9 +423,9 @@ def run_cmd(cmd, check=False, no_shell=False, cwd=None, timeout=None) -> Command
     return CommandReturn(proc.returncode, stdout.decode(), stderr.decode())
 
 
-def check_output(cmd, no_shell=False, cwd=None, timeout=None) -> CommandReturn:
+def check_output(cmd, shell=True, cwd=None, timeout=None) -> CommandReturn:
     """Identical to `run_cmd`, but always sets `check_output` to `True`."""
-    return run_cmd(cmd, True, no_shell, cwd, timeout)
+    return run_cmd(cmd, True, shell, cwd, timeout)
 
 
 def assert_seccomp_level(pid, seccomp_level):
--- a/tests/framework/utils.py
+++ b/tests/framework/utils.py
@@ -57,7 +57,7 @@ def set_cpu_affinity(pid: int, cpulist: list) -> list:
 
 def get_cpu_utilization(pid: int) -> Dict[str, float]:
     """Return current process per thread CPU utilization."""
-    _, stdout, _ = run_cmd(GET_CPU_LOAD.format(pid))
+    _, stdout, _ = check_output(GET_CPU_LOAD.format(pid))
     cpu_utilization = {}
 
     # Take all except the last line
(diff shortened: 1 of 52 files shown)


tests/host_tools/network.py
```diff
@@ -60,8 +60,7 @@ def remote_path(self, path):
 
     def _scp(self, path1, path2, options):
         """Copy files to/from the VM using scp."""
-        ecode, _, stderr = self._exec(["scp", *options, path1, path2])
-        assert ecode == 0, stderr
+        self._exec(["scp", *options, path1, path2], check_output=True)
```
pb8o, line 63: use `check_output` here

tests/host_tools/network.py
```diff
@@ -95,7 +94,7 @@ def _init_connection(self):
         if ecode != 0:
             raise ConnectionError
 
-    def run(self, cmd_string, timeout=None):
+    def run(self, cmd_string, timeout=None, *, check_output=False):
```
pb8o, line 95: I'd rather try to mimic the `subprocess` module here. So if we want it to raise we call `check_output`, otherwise we call `run`. 

That would involve converting all current uses of run to check_output, so maybe that ship has sailed.

pb8o (approved):

tests/host_tools/network.py
pb8o: Ok I see that you did that at the end of the PR :)

tests/framework/utils.py
```diff
@@ -374,14 +374,14 @@ def _format_output_message(proc, stdout, stderr):
     return output_message
 
 
-def check_output(
-    cmd, ignore_return_code=False, no_shell=False, cwd=None, timeout=None
+def run_cmd(
+    cmd, check_output=False, no_shell=False, cwd=None, timeout=None
```
pb8o, line 378: Can this be `check`, same as `subprocess.run`? While we are at it, can we also use `shell` instead of `no_shell`?

tests/host_tools/network.py
roypat: We can't, because `checck_output` delegates to `run`, which prepends the `ssh -i ... root@192.168.0.2` stuff, which we do not want for `scp`

tests/framework/utils.py
```diff
@@ -374,14 +374,17 @@ def _format_output_message(proc, stdout, stderr):
     return output_message
 
 
-def run_cmd_sync(cmd, ignore_return_code=False, no_shell=False, cwd=None, timeout=None):
```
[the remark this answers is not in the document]
roypat: Ohhh, that makes sense, thanks for digging up that information!

tests/framework/utils.py
roypat: Done :)

tests/host_tools/network.py
pb8o: Ah right!

pb8o (approved):

zulinx86 (approved):
I love this PR! Thank you!

tests/framework/utils.py
```diff
@@ -409,11 +409,15 @@ def run_cmd_sync(cmd, ignore_return_code=False, no_shell=False, cwd=None, timeou
 
         raise
 
+    output_message = _format_output_message(proc, stdout, stderr)
+
     # If a non-zero return code was thrown, raise an exception
     if not ignore_return_code and proc.returncode != 0:
-        raise ChildProcessError(_format_output_message(proc, stdout, stderr))
+        CMDLOG.warning("Command failed: %s\n", output_message)
```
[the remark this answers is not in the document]
roypat: I tried using f-strings here at first, but actually ran into a pylint warning against it: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-format-interpolation.html

PR #4649, opened 2024-06-20
refactor: simplify get_msr_chunks

ShadowCurse:
## Changes
Move logic to create single msr chunk into separate function. Also use iterators for msr indexes to remove some unneeded allocations.

## Reason
Minor simplification and removal of unneeded allocations.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/src/vmm/src/cpu_config/x86_64/custom_cpu_template.rs
+++ b/src/vmm/src/cpu_config/x86_64/custom_cpu_template.rs
@@ -144,12 +144,9 @@ pub struct CustomCpuTemplate {
 }
 
 impl CustomCpuTemplate {
-    /// Get a list of MSR indices that are modified by the CPU template.
-    pub fn get_msr_index_list(&self) -> Vec<u32> {
-        self.msr_modifiers
-            .iter()
-            .map(|modifier| modifier.addr)
-            .collect()
+    /// Get an iterator of MSR indices that are modified by the CPU template.
+    pub fn msr_index_iter(&self) -> impl ExactSizeIterator<Item = u32> + '_ {
+        self.msr_modifiers.iter().map(|modifier| modifier.addr)
     }
 
     /// Validate the correctness of the template.
--- a/src/vmm/src/vstate/vcpu/x86_64.rs
+++ b/src/vmm/src/vstate/vcpu/x86_64.rs
@@ -324,7 +324,7 @@ impl KvmVcpu {
     ///
     /// # Arguments
     ///
-    /// * `msr_index_list`: List of MSR indices.
+    /// * `msr_index_iter`: Iterator over MSR indices.
     ///
     /// # Errors
     ///
@@ -332,32 +332,67 @@ impl KvmVcpu {
     /// * When [`kvm_ioctls::VcpuFd::get_msrs`] returns errors.
     /// * When the return value of [`kvm_ioctls::VcpuFd::get_msrs`] (the number of entries that
     ///   could be gotten) is less than expected.
-    fn get_msr_chunks(&self, msr_index_list: &[u32]) -> Result<Vec<Msrs>, KvmVcpuError> {
-        let mut msr_chunks: Vec<Msrs> = Vec::new();
+    fn get_msr_chunks(
+        &self,
+        mut msr_index_iter: impl ExactSizeIterator<Item = u32>,
+    ) -> Result<Vec<Msrs>, KvmVcpuError> {
+        let num_chunks = msr_index_iter.len().div_ceil(KVM_MAX_MSR_ENTRIES);
 
-        for msr_index_chunk in msr_index_list.chunks(KVM_MAX_MSR_ENTRIES) {
-            let mut msrs = Msrs::new(msr_index_chunk.len())?;
-            let msr_entries = msrs.as_mut_slice();
-            assert_eq!(msr_index_chunk.len(), msr_entries.len());
-            for (pos, index) in msr_index_chunk.iter().enumerate() {
-                msr_entries[pos].index = *index;
-            }
+        let mut msr_chunks: Vec<Msrs> = Vec::with_capacity(num_chunks);
 
-            let expected_nmsrs = msrs.as_slice().len();
-            let nmsrs = self
-                .fd
-                .get_msrs(&mut msrs)
-                .map_err(KvmVcpuError::VcpuGetMsrs)?;
-            if nmsrs != expected_nmsrs {
-                return Err(KvmVcpuError::VcpuGetMsr(msr_index_chunk[nmsrs]));
-            }
-
-            msr_chunks.push(msrs);
+        for _ in 0..num_chunks {
+            let chunk_len = msr_index_iter.len().min(KVM_MAX_MSR_ENTRIES);
+            let chunk = self.get_msr_chunk(&mut msr_index_iter, chunk_len)?;
+            msr_chunks.push(chunk);
         }
 
         Self::fix_zero_tsc_deadline_msr(&mut msr_chunks);
 
         Ok(msr_chunks)
+    }
+
+    /// Get single MSR chunk for the given MSR index iterator with
+    /// specified length. Iterator should have enough elements
+    /// to fill the chunk with indices, otherwise KVM will
+    /// return an error when processing half filled chunk.
+    ///
+    /// # Arguments
+    ///
+    /// * `msr_index_iter`: Iterator over MSR indices.
+    /// * `chunk_size`: Lenght of a chunk.
+    ///
+    /// # Errors
+    ///
+    /// * When [`kvm_bindings::Msrs::new`] returns errors.
+    /// * When [`kvm_ioctls::VcpuFd::get_msrs`] returns errors.
+    /// * When the return value of [`kvm_ioctls::VcpuFd::get_msrs`] (the number of entries that
+    ///   could be gotten) is less than expected.
+    pub fn get_msr_chunk(
+        &self,
+        msr_index_iter: impl Iterator<Item = u32>,
+        chunk_size: usize,
+    ) -> Result<Msrs, KvmVcpuError> {
+        let chunk_iter = msr_index_iter.take(chunk_size);
+
+        let mut msrs = Msrs::new(chunk_size)?;
+        let msr_entries = msrs.as_mut_slice();
+        for (pos, msr_index) in chunk_iter.enumerate() {
+            msr_entries[pos].index = msr_index;
+        }
+
+        let nmsrs = self
+            .fd
+            .get_msrs(&mut msrs)
+            .map_err(KvmVcpuError::VcpuGetMsrs)?;
+        // GET_MSRS returns a number of successfully set msrs.
+        // If number of set msrs is not equal to the length of
+        // `msrs`, then the value retuned by GET_MSRS can act
+        // as an index to the problematic msr.
+        if nmsrs != chunk_size {
+            Err(KvmVcpuError::VcpuGetMsr(msrs.as_slice()[nmsrs].index))
+        } else {
+            Ok(msrs)
+        }
     }
 
     /// Get MSRs for the given MSR index list.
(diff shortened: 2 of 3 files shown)


src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -324,42 +324,76 @@ impl KvmVcpu {
...
+        let mut msr_chunks: Vec<Msrs> = Vec::with_capacity(num_chunks);
+
+        for _ in 0..num_chunks {
+            let chunk = self.get_msr_chunk(&mut msr_index_iter, KVM_MAX_MSR_ENTRIES)?;
+            msr_chunks.push(chunk);
+        }
+        if excess != 0 {
+            let chunk = self.get_msr_chunk(&mut msr_index_iter, excess)?;
+            msr_chunks.push(chunk);
         }
```
zulinx86, line 351: The `num_chunks` is not the correct length if `excess != 0`. How about this?
```suggestion
        let num_chunks = (msr_index_iter.len() + KVM_MAX_MSR_ENTRIES - 1) / KVM_MAX_MSR_ENTRIES;
        let mut msr_chunks: Vec<Msrs> = Vec::with_capacity(num_chunks);
        for _ in 0..num_chunks - 1 {
            let chunk = self.get_msr_chunk(&mut msr_index_iter, KVM_MAX_MSR_ENTRIES)?;
            msr_chunks.push(chunk);
        }
        let last = msr_index_iter.len() - KVM_MAX_MSR_ENTRIES * (num_chunks - 1);
        let chunk = self.get_msr_chunk(&mut msr_index_iter, last)?;
        msr_chunks.push(chunk);
```

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -324,42 +324,76 @@ impl KvmVcpu {
...
+    /// * When [`kvm_ioctls::VcpuFd::get_msrs`] returns errors.
+    /// * When the return value of [`kvm_ioctls::VcpuFd::get_msrs`] (the number of entries that
+    ///   could be gotten) is less than expected.
+    pub fn get_msr_chunk(
+        &self,
+        msr_index_iter: impl ExactSizeIterator<Item = u32>,
+        max_chunk_len: usize,
+    ) -> Result<Msrs, KvmVcpuError> {
+        let chunk_size = msr_index_iter.len().min(max_chunk_len);
+        let chunk_iter = msr_index_iter.take(chunk_size);
```
zulinx86, line 378: why does it have to be `max_chunk_len` rather than just `chunk_len`?
Isn't the exact length of chunk calculated in `get_msr_chunks()`?

```suggestion
        chunk_len: usize,
    ) -> Result<Msrs, KvmVcpuError> {
        let chunk_iter = msr_index_iter.take(chunk_len);
```

src/vmm/src/cpu_config/x86_64/custom_cpu_template.rs
```diff
@@ -144,12 +144,9 @@ pub struct CustomCpuTemplate {
...
 
 impl CustomCpuTemplate {
-    /// Get a list of MSR indices that are modified by the CPU template.
-    pub fn get_msr_index_list(&self) -> Vec<u32> {
-        self.msr_modifiers
-            .iter()
-            .map(|modifier| modifier.addr)
-            .collect()
+    /// Get an iterator of MSR indices that are modified by the CPU template.
+    pub fn msr_index_iter(&self) -> impl ExactSizeIterator<Item = u32> + '_ {
```
zulinx86, line 148: I don't have strong preference, but if you want to reduce unneeded allocation, how about making it return `impl ExactSizeIterator<Item = &u32>`? By this we can remove some `.copied()` (e.g., `vcpu.msrs_to_save.iter().copied()`, `self.msrs_to_save.iter().copied()`, `msrs_to_dump.as_slice().iter().copied()`).

src/vmm/src/vstate/vcpu/x86_64.rs
ShadowCurse: I was thinking about the case when iterator has less elements than the chunk size. In this case the `Msrc::new` will be given a bigger size than iterator has indexes. And I assume that KVM will not be happy to receive this half populated object. But I guess you are right, the `get_msr_chunks` should give correct size all the time. I will update this.

src/vmm/src/cpu_config/x86_64/custom_cpu_template.rs
ShadowCurse: I don't like `.copied()` business as well, but this does not do any allocations. Here we are talking about `&u32` and `u32` which is actually cheaper to copy than to have references (`u32` is 32 bits, `&u32` is `usize` so 64 bits).

src/vmm/src/vstate/vcpu/x86_64.rs
ShadowCurse: As far as my math goes, both approaches are correct. 
- In case the number of indexes is not divisible by `KVM_MAX_MSR_ENTRIES` both approaches do get first `n` full chunks in the first loop and then get last chunk as a second step.
- In case where  the number of indexes is divisible by `KVM_MAX_MSR_ENTRIES` my approach gets all chunks in the first loop and `excess` is 0, last chunk logic is ignored. In your case it will do `n-1` iterations in the first loop and will get last full chunk in the last part.

IMO my approach is simpler just because there is less math.

ShadowCurse: I wish I could write a unit test for this, but the amount of msrs returned by `get_msrs_to_dump` or `vcpu.msrs_to_save` is smaller than `KVM_MAX_MSR_ENTRIES`

zulinx86: What I meant is its capacity (`Vec::with_capacity(num_chunks);`) is not correct when the number of indices is not divisible by `KVM_MAX_MSR_ENTRIES`. The last `push()` requires it to gain another space.

src/vmm/src/vstate/vcpu/x86_64.rs
zulinx86: indeed. you're also right. it might be better to have the note in the function's comment.

src/vmm/src/vstate/vcpu/x86_64.rs
ShadowCurse: oh, yeah, I knew about this, but didn't give it a lot of attention. I guess I can just do `Vec::with_capacity(num_chunks + 1);` to be sure it has enough space.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -324,42 +324,78 @@ impl KvmVcpu {
...
 
+    /// Get single MSR chunk for the given MSR index iterator with
+    /// specified length. Iterator should have enough elements
+    /// to fill the chunk with indices, otherwise KVM will
+    /// return an error when processing half filled chunk.
+    ///
+    /// # Arguments
+    ///
+    /// * `msr_index_iter`: Iterator over MSR indices.
+    /// * `max_chunk_len`: Maximum lenght of a chunk to get.
```
zulinx86, line 367: should be `chunk_size` with a right comment?

ShadowCurse: fixed

src/vmm/src/vstate/vcpu/x86_64.rs
zulinx86: I'd like to hear @roypat 's opinion for this.

roypat: How about just something like (pseudo-code)

```rs
self.msr_modifiers
     .chunks(KVM_MAX_MSR_ENTRIES)
     .map(|chunk| chunk.iter().map(|modifier| modifier.addr))
     .map(|chunk| self.get_msr_chunk(chunk))  // we have an exactsize iterator here, so can drop the "how many" parameter
     .collect::<Vec<_>>();
```

ShadowCurse: unfortunately `get_msr_chunks` is used in `get_msrs`, so it needs to accept something that contains indexes.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -324,42 +324,78 @@ impl KvmVcpu {
...
+        for (pos, msr_index) in chunk_iter.enumerate() {
+            msr_entries[pos].index = msr_index;
+        }
+
+        let nmsrs = self
+            .fd
+            .get_msrs(&mut msrs)
+            .map_err(KvmVcpuError::VcpuGetMsrs)?;
+        if nmsrs != chunk_size {
+            Err(KvmVcpuError::VcpuGetMsr(msrs.as_slice()[nmsrs].index))
```
roypat, line 393: Mh, I'm assuming `msrs.as_slice()[nmsrs]` is supposed to be the first MSR from the request that failed to be retrieved? Can we add a comment about that? I had to scratch my head a bit about why we're looking at `nmsrs`, because at first I thought we were trying to get the last MSR that _was_ successfully retrieved (in which case this would've needed to be `nmsrs - 1` and would break if `nmsrs == 0`).

src/vmm/src/vstate/vcpu/x86_64.rs
roypat: Alright, let's use https://doc.rust-lang.org/std/primitive.u32.html#method.div_ceil then :)
```rs
    fn get_msr_chunks(
        &self,
        mut msr_index_iter: impl ExactSizeIterator<Item = u32>,
    ) -> Result<Vec<Msrs>, KvmVcpuError> {
        let num_chunks = msr_index_iter.len().div_ceil(KVM_MAX_MSR_ENTRIES);

        // Allocate 1 more element in case excess is not 0
        let mut msr_chunks: Vec<Msrs> = Vec::with_capacity(num_chunks);

        for _ in 0..num_chunks {
            let len = msr_index_iter.len();
            let chunk = self.get_msr_chunk(&mut msr_index_iter, KVM_MAX_MSR_ENTRIES.min(len))?;
            msr_chunks.push(chunk);
        }

        Self::fix_zero_tsc_deadline_msr(&mut msr_chunks);

        Ok(msr_chunks)
    }
```

src/vmm/src/vstate/vcpu/x86_64.rs
ShadowCurse: I had the same question at first, so I asked @zulinx86 about it. It seems the `get_msrs` does return a number of msrs set successfully. In this case the `nmsrs != chunk_size` check is correct. The `msrs.as_slice()[nmsrs]` is also correct because if some msrs could not be retrieved `nmsrs` will be an index of the failed `msr`. Example: you ask for 2 msrs, if KVM returns 1 because it could only retrieve mrs with index 0, then 1 points to the problematic msr. If KVM returns 2, than everything is fine.

Added a comment about this.

src/vmm/src/vstate/vcpu/x86_64.rs
ShadowCurse: done

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -324,42 +324,78 @@ impl KvmVcpu {
...
-            }
-
-            msr_chunks.push(msrs);
+    fn get_msr_chunks(
+        &self,
+        mut msr_index_iter: impl ExactSizeIterator<Item = u32>,
+    ) -> Result<Vec<Msrs>, KvmVcpuError> {
+        let num_chunks = msr_index_iter.len().div_ceil(KVM_MAX_MSR_ENTRIES);
+
+        // Allocate 1 more element in case excess is not 0
```
zulinx86, line 341: No longer allocate an unnecessary element, isn't it?

src/vmm/src/vstate/vcpu/x86_64.rs
roypat: Thank you (both for the comment and explanation)!

src/vmm/src/vstate/vcpu/x86_64.rs
ShadowCurse: It does. For example 3 / 2 will return 2 in this case

zulinx86 (approved):

roypat (approved):

Issue #746, opened 2018-12-05
 [:ERROR:devices/src/virtio/net.rs:305] Failed to write to tap


rmzg:
I've been attempting to perform some basic firecracker tests and this is my setup so far:

`firecracker-v0.11.0` pre-built binary on `Linux dl380 4.9.0-8-amd64 #1 SMP Debian 4.9.110-3+deb9u6 (2018-10-08) x86_64 GNU/Linux`

firecracker running as root via sudo.

Setup commands:
```
curl --unix-socket /tmp/firecracker.socket -i     -X PUT 'http://localhost/boot-source'       -H 'Accept: application/json'               -H 'Content-Type: application/json'         -d '{
        "kernel_image_path": "./hello-vmlinux.bin",
        "boot_args": "console=ttyS0 reboot=k panic=1 pci=off"
    }'

curl --unix-socket /tmp/firecracker.socket -i     -X PUT 'http://localhost/drives/rootfs'     -H 'Accept: application/json'               -H 'Content-Type: application/json'         -d '{
        "drive_id": "rootfs",
        "path_on_host": "./hello-rootfs.ext4",
        "is_root_device": true,
        "is_read_only": false
    }'

curl --unix-socket /tmp/firecracker.socket -i     -X PUT 'http://localhost/network-interfaces/0' -H 'Accept: application/json' -H 'Content-Type: application/json' -d '{
 "host_dev_name":"tap1","iface_id":"0", "state":"Attached"}'

curl --unix-socket /tmp/firecracker.socket -i     -X PUT 'http://localhost/actions'           -H  'Accept: application/json'              -H  'Content-Type: application/json'        -d '{
        "action_type": "InstanceStart"
     }'
```

All of which reported the usual 204 responses.

`ip link ls` on the host:
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: enp3s0f0: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether 44:1e:a1:07:51:c2 brd ff:ff:ff:ff:ff:ff
3: enp3s0f1: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
[...]

Issue #754, opened 2018-12-10
Return error when attempting to open a non-existing tap device


aghecenco:
Currently, if Firecracker has the `CAP_NET_ADMIN` capability set and is asked to open a tap device that has not been previously created on the host, it creates it first and opens it. The name of the new tap device is `vmtap${INTERFACE_ID}`. This can lead to confusion therefore it would make sense for Firecracker to simply not allow opening non-existing tap devices.

andreeaflorescu:
Not sure if this is going to be possible with our current implementation. I can't find any official documentation, but looking at the kernel code it looks like the ioctl TUNSETIFF is going to create the tap device if it does not exist.

cmannett85:
I have written an [implementation](https://github.com/cmannett85/firecracker/commit/08df645fc928275b8e9b7e77cff300ce1f058a3c) of this that works as expected with basic manual testing.  However I don't know how to unit test it as all the tests relied on `Tap::open_named(..)` to create the TAP if it doesn't exist.

I [started an attempt](https://github.com/cmannett85/firecracker/commit/8cae250eee3898e99e7367d77b15fbfc8e48b905) by changing the TAP to a multi-queue type so I could create the TAP in the relevant unit tests before having the API open it - but this broke other unit tests that relied on a TAP device not being able to be opened twice.  This also required that a method be exposed that still created a TAP as other crates' unit tests needed this ability.

The only alternative I can think of is using a system call `ip tuntap add ...` inside the tests that require it, the downside is that the device is created with the `IFF_PERSIST` flag set and so must be manually deleted with `ip tuntap del ...`.  Does anyone have any better suggestions?

aghecenco:
@cmannett85 your solution relies on `/sys` which is unavailable in the jail, which Firecracker should always be running in per the security recommendations. Outside of `/sys`, I don't know of any native (ioctl-only) way to find out if a tap already exists. And I don't think `/sys` should be mounted in the jail - only the basic requirements should be present there, and I don't believe this qualifies as one. If you can't think of any other way to check this (I can't), I think the best thing to do is to close the issue.

dianpopa:
Double checked in the kernel code if there is any native way we could use to detect whether a tap device already exists. Could not find one. Closing this since no other viable way to implement the proposed enhancement was found.

raduiliescu:
You can also use netlink sockets to list the netdevices. But I am not sure if that is allowed through our Seccomp and if we want to allow it.

[...]

PR #4650, opened 2024-06-22
Don't spam console with "failed to write to tap" messages

larsks:
A common first experience with firecracker (see e.g. #746) is to start a
new virtual machine and immediately start getting spammed with messages
like:

    2024-06-21T22:10:18.189726925 [anonymous-instance:main] Failed to write
    to tap: Os { code: 5, kind: Uncategorized, message: "Input/output
    error" }

This happens because firecracker will create the tap device if it doesn't
already exist, and unlike most vmms it will not configure the device `up`
in this case.

All that means that this is both a common and an expected situation and
does not require this level of notification.

Signed-off-by: Lars Kellogg-Stedman <lars@oddbit.com>

### DIFF
--- a/src/vmm/src/devices/virtio/net/device.rs
+++ b/src/vmm/src/devices/virtio/net/device.rs
@@ -498,8 +498,7 @@ impl Net {
                 net_metrics.tx_packets_count.inc();
                 net_metrics.tx_count.inc();
             }
-            Err(err) => {
-                error!("Failed to write to tap: {:?}", err);
+            Err(_) => {
                 net_metrics.tap_write_fails.inc();
             }
         };


zulinx86:
Hi @larsks ,
Thank you for your PR.

As explained in our network setup doc, the first step to use network is to create a tap device:  https://github.com/firecracker-microvm/firecracker/blob/main/docs/network-setup.md

Also, as mentioned in the issue #764, we recommend to provide an already-created tap interface.

> The recommended method of running firecracker though is unprivileged and providing an already-created tap interface.

To prevent users from misusing Firecracker, we attempted to make it exit with error when the tap has not been created in issue #754, but we weren't able to find the way to detect that unfortunately.

> Double checked in the kernel code if there is any native way we could use to detect whether a tap device already exists. Could not find one. Closing this since no other viable way to implement the proposed enhancement was found.

We think just suppressing the error message doesn't solve the fundamental problem now, but we might get chance of reconsidering it! So could you please provide more information?
- Are you seeing the error message before microVM starts or after?
- How do you set up the tap device created by Firecracker to make it work?
- Are there any reasons why you set up the tap device before running Firecracker?

Thanks,

zulinx86:
Hi @larsks ,

We hope this finds you well. 
Had you have a chance to look at the previous update?

We are closing this issue because there is no updates for more than 1 month and we think this may not be relevant to you anymore. 
Please feel free to reopen it or reach us out if you have any questions or updates.

Thanks for your understanding.

Issue #2609, opened 2021-06-03
[Feature Request] Hot-plug vCPUs


jeromegn:
# Feature Request

We'd like to be able to add (and remove) vCPUs from running firecracker microvms. It appears to be possible to do that with KVM. Examples I've seen define how many vCPUs at most a VM might use and then the actual number it will be using at boot. Then you can add vCPUs via `virsh`.

https://www.unixarena.com/2015/12/linux-kvm-how-to-add-remove-vcpu-to-guest-on-fly.html/

This would allow use to add a "burst" feature when CPU usage spikes.

## Describe the desired solution

An API to modify a running microvm's vCPUs count.

This should notify the guest VM of the change:

```
Dec 16 12:48:28 UA-KVM1 kernel: CPU1 has been hot-added
Dec 16 12:48:28 UA-KVM1 kernel: SMP alternatives: switching to SMP code
Dec 16 12:48:57 UA-KVM1 kernel: smpboot: Booting Node 0 Processor 1 APIC 0x1
Dec 16 12:48:57 UA-KVM1 kernel: kvm-clock: cpu 1, msr 0:3ff87041, secondary cpu clock
Dec 16 12:48:57 UA-KVM1 kernel: TSC synchronization [CPU#0 -> CPU#1]:
Dec 16 12:48:57 UA-KVM1 kernel: Measured 906183720569 cycles TSC warp between CPUs, turning off TSC clock.
Dec 16 12:48:57 UA-KVM1 kernel: tsc: Marking TSC unstable due to check_tsc_sync_source failed
Dec 16 12:48:57 UA-KVM1 kernel: KVM setup async PF for cpu 1
Dec 16 12:48:57 UA-KVM1 kernel: kvm-stealtime: cpu 1, msr 3fd0d240
Dec 16 12:48:57 UA-KVM1 kernel: microcode: CPU1 sig=0x206c1, pf=0x1, revision=0x1
Dec 16 12:48:57 UA-KVM1 kernel: Will online and init hotplugged CPU: 1
```

## Describe possible alternatives

We could give every firecracker microvm access to all cores and only use cgroups to limit actual scheduling time. This is not great though as it might create a lot of CPU steal. We prefer to give full cores when possible.

## Checks

- [x] Have you searched the Firecracker Issues database for similar requests?
- [x] Have you read all the existing relevant Firecracker documentation?
- [x] Have you read and understood Firecracker's core tenets?

[...]

PR #4657, opened 2024-06-28
Create ACPI Devices for Hotplugging

jxcurtis:
## Changes
* ACPI AML methods and devices for hotplugging defined.
* `Bus` refactored, previously contained `ArcMutexes` of `BusDevice`, now stores `ArcMutexes` of the devices inside the `BusDevice`. This allows the individual devices to be shared.
* x86 CPUID updated to account for the possibility of hotplugged CPUs.

## Reason
After new vCPU threads have been added to the VMM, ACPI device is used to notify the guest that the CPUs are available to them. Related to feature request #2609.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet


src/vmm/src/builder.rs
```diff
@@ -711,7 +711,8 @@ pub fn setup_serial_device(
         ),
         input: Some(input),
     })));
-    event_manager.add_subscriber(serial.clone());
+    event_manager.add_subscriber(Arc::new(Mutex::new(serial.clone()))); // TODO: Check this actually
```
roypat, line 736: should just be 

```suggestion
    event_manager.add_subscriber(Arc::clone(&serial)); 
```

I think

src/vmm/src/devices/bus.rs
```diff
@@ -102,97 +102,65 @@ impl ConstantDevice {
...
     }
-    pub fn mmio_transport_mut(&mut self) -> Option<&mut MmioTransport> {
+    pub fn mmio_transport_ref(&self) -> Option<&Arc<Mutex<MmioTransport>>> {
         match self {
             Self::MmioTransport(x) => Some(x),
             _ => None,
         }
     }
-    pub fn serial_mut(&mut self) -> Option<&mut SerialDevice<std::io::Stdin>> {
+    pub fn serial_ref(&self) -> Option<&Arc<Mutex<SerialDevice<std::io::Stdin>>>> {
```
roypat, line 134: These could have signatures along the lines of 
```suggestion
    pub fn serial_ref(&self) -> Option<MutexGuard<SerialDevice<std::io::Stdin>>> {
```
by moving the `.lock().expect("Poisoned mutex")` that all call-sites do into the these functions.

src/vmm/src/device_manager/legacy.rs
```diff
@@ -35,9 +35,9 @@ pub enum LegacyDeviceError {
 pub struct PortIODeviceManager {
     pub io_bus: crate::devices::Bus,
     // BusDevice::Serial
-    pub stdio_serial: Arc<Mutex<BusDevice>>,
+    pub stdio_serial: BusDevice,
```
roypat, line 38: Ah, lol, so we did face the problem of not being able to share devices properly before already! We just relied on the awkward matches apparently. Well, now that we have this refactor, we can clean this up and just store a `Arc<Mutex<Serial>>` here and a `Arc<Mutex<I8042Device>>` below :)

src/vmm/src/device_manager/acpi.rs
```diff
@@ -1,21 +1,28 @@
...
 use kvm_ioctls::VmFd;
 
+use crate::devices::acpi::cpu_container::CpuContainer;
 use crate::devices::acpi::vmgenid::VmGenId;
 
 #[derive(Debug)]
 pub struct ACPIDeviceManager {
     /// VMGenID device
     pub vmgenid: Option<VmGenId>,
+    pub cpu_container: Option<Arc<Mutex<CpuContainer>>>,
```
roypat, line 16: mh, will we ever _not_ have a CpuContainer?

src/vmm/src/lib.rs
```diff
@@ -393,6 +393,15 @@ impl Vmm {
...
 
+        #[cfg(target_arch = "x86_64")]
+        if let Some(devices::BusDevice::CpuContainer(cont)) =
+            self.get_bus_device(DeviceType::CpuContainer, "CpuContainer")
+        {
+            let mut locked_container = cont.lock().expect("Poisoned lock");
+            for i in self.vcpus_handles.len()..(self.vcpus_handles.len() + vcpu_count) {
+                locked_container.cpu_devices[i].inserting = true;
+            }
+        }
```
roypat, line 407: Mhh, is this correct to do for the boot cpus? I guess when we construct the boot CPUs, we haven't attached the cpucontainer to the bus yet, and thus we don't set the "inserting" bit, but that sounds quite brittle. Can we explicitly prevent this from happening on the boot cpus?

src/vmm/src/lib.rs
```diff
@@ -677,6 +686,8 @@ impl Vmm {
             huge_pages: None,
         };
 
+        self.resume_vm()?;
```
roypat, line 689: While this get's the job done, it's not really what we should do here. `resume_vm()`
1. sends notifications to all devices on the MMIO bus
2. sends a "Resume" notification to _all_ vcpu threads
3. waits for the vcpu threads to get back to us with a "resumed" response

Here, we only need 2. and 3. for the newly added vcpus, but we probably don't need the device notifications, and we definitely don't want to send notifications to all other vcpus, because that just needlessly causes them to KVM_EXIT.

roypat (commented):
Great work!! Amazing to see this progressing so quickly :)

src/vmm/src/lib.rs
jxcurtis: Hmm, I've been thinking about the possible effects of this even if the cpucontainer is attached. What will happen is that when `CSCN` AML method is called, AML will notify the device with 0x1, device check. From ACPI spec:

> Device Check. Used to notify OSPM that the device either appeared or disappeared. If the device has appeared, OSPM will re-enumerate from the parent. If the device has disappeared, OSPM will invalidate the state of the device. OSPM may optimize out re-enumeration.

So this may cause a reenumeration of the Processor container object `PNP0A05`/`ACPI0010`. This may also be optimised out, I can't tell. But I think it's probably best not to send the notification out anyway. There is also possibility that this could cause initiate an undock, if a `_DCK` object is ever present within the `CpuDevice` AML (not sure it would be, but it's a possibility I suppose). From the same paragraph of the spec:

> If _DCK is present, then Notify(object,1) is assumed to indicate an undock request. 

For this reason, I think it's best to perhaps move this block out of `start_vcpus` and put it into the `hotplug_vcpus` function, so that `inserting` is only set for the newly added vCPUs. I will just need to set `CpuDevices` as `active` differently, but this should be easy enough to do, just pass in the boot `cpu_count` as a parameter to the CpuContainer constructor. That should be fairly easy.

src/vmm/src/device_manager/acpi.rs
jxcurtis: Hmm. Well in the case that you don't have ACPI enabled, I suppose not? But you also wouldn't have the whole ACPIDevMgr, so that's a moot point I suppose. I'd need to look at how snapshot restore works with this, I imagine it's a fairly simple change, but I need to have a look

src/vmm/src/lib.rs
jxcurtis: Yes, I had noticed that, it was more just a convenient way to see if starting the threads would work as solution. I can probably create a new fn to do steps 2. and 3. and use that instead.

src/vmm/src/devices/bus.rs
jxcurtis: All done. Thanks for the suggestion, I missed that somehow.

src/vmm/src/builder.rs
jxcurtis: Is there any semantic difference between `Arc::clone(&serial)` and `serial.clone()`? The latter seems to be accepted here, but the former is not.

roypat: Mhhh, wow, okay, my turn to be absolutely stumped by what the rust compiler wants. The following works:

```diff
diff --git a/src/vmm/src/builder.rs b/src/vmm/src/builder.rs
index 27d986251..42055b592 100644
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -723,7 +723,8 @@ pub fn setup_serial_device(
     let interrupt_evt = EventFdTrigger::new(EventFd::new(EFD_NONBLOCK).map_err(VmmError::EventFd)?);
     let kick_stdin_read_evt =
         EventFdTrigger::new(EventFd::new(EFD_NONBLOCK).map_err(VmmError::EventFd)?);
-    let serial = BusDevice::Serial(Arc::new(Mutex::new(SerialWrapper {
+
+    let serial = Arc::new(Mutex::new(SerialWrapper {
         serial: Serial::with_events(
             interrupt_evt,
             SerialEventsWrapper {
@@ -732,10 +733,13 @@ pub fn setup_serial_device(
             SerialOut::Stdout(out),
         ),
         input: Some(input),
-    })));
-    event_manager.add_subscriber(Arc::new(Mutex::new(serial.clone()))); // TODO: Check this actually
-                                                                        // works?
-    Ok(serial)
+    }));
+
+    let to_return = Arc::clone(&serial);
+
+    event_manager.add_subscriber(serial);
+
+    Ok(BusDevice::Serial(to_return))
 }
```

but changing the last 3 lines to be 

```
event_manager.add_subscriber(Arc::clone(&serial));
Ok(BusDevice::Serial(to_return))
```

doesn't compile, with it complaining that `serial` doesnt implement `MutEventSubscriber` (even though it does). 

Anyway, in my original comment, I missed that `serial` is a `BusDevice` instead of a `SerialWrapper`, so your solution is okay (as you're still just cloning an Arc instead of the actual device). However, it does end up registering something like `Arc<Mutex<Arc<Mutex<SerialWrapper>>>>` to the event bus.

roypat: Ah, and you've pretty much already done exactly what I suggested, lol. Should've checked the diff before commenting, sorry haha. Just having the `.clone()` is fine, although for these smart pointer types that `Deref` into their pointed-to value, writing `Arc::clone()` clarifies that we're cloning the `Arc` and not what's inside the `Arc`. Definitely more of a nit though, no need to fight rustc over this :)

src/vmm/src/device_manager/acpi.rs
roypat: Do we support disabling ACPI (I genuinely don't remember)?

jxcurtis: Well for the time being we have the no-acpi config for 5.10

roypat: But does using that actually cause Firecracker to disable ACPI features, or does it mean Firecracker writes the ACPI tables anyway, but the guest simply ignores them?

jxcurtis: I wasn't sure about this to be honest, so I just tested it. It looks like the latter is right. After running the API call there are no new CPUs present to the guest. However, it does lead to an interesting edge case: the vCPU threads are made on the VMM side, but the CPUs are never usable. I'm not sure the best way to handle this edge case, I can potentially look and see if there is a way to query the kernel about ACPI support? Then an error can be returned.

src/vmm/src/lib.rs
jxcurtis: Okay cool, I've now refactored it slightly, adding a new function `resume_vcpu_threads()`. It takes an int index as an argument and starts the threads starting after that. This should solve this problem

src/vmm/src/lib.rs
jxcurtis: Update: I've made the changes I mentioned.

src/vmm/src/device_manager/acpi.rs
roypat: > Would it maybe be better just to assume there is ACPI support, and ignore cases where there isn't?

Yeah, I think so. Booting without ACPI is deprecated already, so in the (near) future there won't even be the possibility of not having ACPI support anymore.

jxcurtis: Okay cool. In that case, I don't see any reason why the CpuContainer shouldn't exist. I'll modify it as such.

src/vmm/src/devices/acpi/cpu_container.rs
```diff
@@ -0,0 +1,375 @@
...
+                &aml::Name::new("_UID".into(), &self.cpu_id),
+                // Currently, AArch64 cannot support following fields.
+                // _STA return value:
+                // Bit [0] – Set if the device is present.
+                // Bit [1] – Set if the device is enabled and decoding its resources.
+                // Bit [2] – Set if the device should be shown in the UI.
+                // Bit [3] – Set if the device is functioning properly (cleared if device failed
+                // its diagnostics). Bit [4] – Set if the battery is present.
+                // Bits [31:5] – Reserved (must be cleared).
+                #[cfg(target_arch = "x86_64")]
```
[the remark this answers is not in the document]
jxcurtis: Yeah, and perhaps intel as well, a lot of it seems to have originally come from NEMU: https://github.com/intel/nemu/wiki/CPU-hotplug#from-acpi

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -40,7 +40,7 @@ def clean_and_mkdir(dir_path):
 
 def _check_cpuid_x86(test_microvm, expected_cpu_count, expected_htt):
     expected_cpu_features = {
-        "maximum IDs for CPUs in pkg": f"{expected_cpu_count:#x} ({expected_cpu_count})",
+        "maximum IDs for CPUs in pkg": f"{32:#x} ({32})",
```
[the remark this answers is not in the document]
jxcurtis: I'm thinking that instead of just using the number 32, this should be extracted into a constant that can be used across tests, so that it can be easily updated should we support more vCPUs.

src/vmm/src/devices/bus.rs
```diff
@@ -102,97 +106,69 @@ impl ConstantDevice {
...
         }
     }
-    pub fn serial_mut(&mut self) -> Option<&mut SerialDevice<std::io::Stdin>> {
+    pub fn serial_ref(&self) -> Option<MutexGuard<SerialDevice<std::io::Stdin>>> {
         match self {
-            Self::Serial(x) => Some(x),
+            Self::Serial(x) => Some(x.lock().expect("Poisoned lock")),
             _ => None,
         }
     }
```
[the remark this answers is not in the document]
jxcurtis: Hmm. It looks like `serial_ref` is still in use for some aarch64 code, see `lib.rs:488`. The legacy device manager is x86 specific no? So although currently unused, should we not keep these functions available so that ARM machines can still access these devices?

src/vmm/src/devices/acpi/cpu_container.rs
```diff
@@ -0,0 +1,375 @@
...
+                    active: true,
+                    inserting: false,
+                })
+            } else {
+                cpu_devices.push(CpuDevice {
+                    cpu_id: i,
+                    active: false,
+                    inserting: false,
+                })
+            }
```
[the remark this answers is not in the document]
jxcurtis: Neat! I hadn't considered I could do it like this

src/vmm/src/devices/acpi/cpu_container.rs
```diff
@@ -0,0 +1,375 @@
...
+            CPU_STATUS_OFFSET => {
+                if self.selected_cpu < MAX_SUPPORTED_VCPUS {
+                    let cpu_device = &self.cpu_devices[self.selected_cpu as usize];
+                    if cpu_device.active {
+                        data[0] |= 1 << CPU_ENABLE_FLAG;
+                    }
+                    if cpu_device.inserting {
+                        data[0] |= 1 << CPU_INSERTING_FLAG;
+                    }
```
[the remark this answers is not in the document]
jxcurtis: Hmm. I'm not sure what to think about this one. It's very nice and concise but it might take away some of the readability.

src/vmm/src/devices/acpi/cpu_container.rs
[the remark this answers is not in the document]
jxcurtis: Well it looks like NEMU was the precursor to Cloud Hypervisor. Not too sure about the proper way to format the copyright.

src/vmm/src/lib.rs
```diff
@@ -624,22 +636,28 @@ impl Vmm {
...
-            for cpu_idx in start_idx..(start_idx + config.add) {
+            for cpu_idx in start_idx..(config.add + start_idx) {
                 let exit_evt = self
                     .vcpus_exit_evt
                     .try_clone()
                     .map_err(HotplugVcpuError::EventFd)?;
-                let vcpu =
+                let mut vcpu =
                     Vcpu::new(cpu_idx, &self.vm, exit_evt).map_err(HotplugVcpuError::VcpuCreate)?;
+                if let Some(kvm_config) = self.vcpu_config.as_ref() {
```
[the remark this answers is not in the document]
jxcurtis: Potentially, I did it this way to preserve the order of events as much as possible. Right now, the `vcpu_config` isn't made until `configure_system_for_boot` runs, it then uses the CPU template to create the `cpu_config` and in turn the `vcpu_config`.

It's possible to move the creation of the vcpu_config to the creation of the VMM, passing the CPU template into `create_vmm_and_vcpus`. I'll investigate this, and see if there are any adverse effects.

src/vmm/src/device_manager/acpi.rs
```diff
@@ -42,44 +57,90 @@ impl ACPIDeviceManager {
...
-                "_SB_.GED_".into(),
-                vec![
-                    &aml::Name::new("_HID".into(), &"ACPI0013"),
-                    &aml::Name::new(
-                        "_CRS".into(),
-                        &aml::ResourceTemplate::new(vec![&aml::Interrupt::new(
+        // Depending on what devices are available, generate AML for GED.
+
+        let locked_container = self.cpu_container.lock().expect("Poisoned lock");
+        match self.vmgenid.as_ref() {
```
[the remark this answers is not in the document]
jxcurtis: I agree, it is very similar. I tried to prevent this whilst I was working on it, but I'm yet to find a suitable solution. I keep running into issues with lifetimes. Looking into it.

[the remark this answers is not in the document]
jxcurtis: I'm not sure there is a way to do this. The way the AML is implemented, using `&dyn Aml`, we cannot meaningfully store any AML implementing structs. If you try to store one and later try and convert it to AML bytes later, you receive a `temporary value dropped while borrowed` error. Instead we must fully specify the device and convert it to AML bytes immediately.

This is what I've understood from looking into it, I've tried many different permutations of this and they all lead to the same error above. Obviously though, my Rust knowledge is limited, so if there is a way to do this that I'm missing, please let me know :)

src/vmm/src/device_manager/acpi.rs
bchalios: Confirming that Firecracker always writes ACPI tables in guest memory. The `no-acpi` kernel is a guest kernel that does not support ACPI. In these cases the guest just boots with the "legacy" way of MPTABLE and command line arguments. As Patrick correctly said, we will eventually drop support for these, i.e. we will not write the MPTABLE in guest memory, neither pass VirtIO devices via the command line.

src/vmm/src/device_manager/acpi.rs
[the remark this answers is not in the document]
bchalios: James is right here. Most of the `aml` module methods get as an argument a `&'a dyn Aml` arguments for the child objects. (I assume the authors did that for some sort of optimization(?)). As a result, you can use this in a sort of builder pattern which needs to end to a call to `append_aml_bytes`:

```Rust
aml::Device::new(.... vec[ &aml::SomeOtherObject(...), ...]).append_aml_bytes(v);l
```

which is unfortunate, because it makes it extremely hard to write nice/readable code.

src/vmm/src/device_manager/acpi.rs
jxcurtis: I'm thinking of undoing this change, it's causing issues that aren't immediately obvious. When `CpuContainer` is not wrapped in an option, it must be constructed and passed as an argument into the constructor of the `ACPIDeviceManager`. The issue with this is that `ACPIDeviceManager` must be constructed in the `create_vmm_and_vcpus` method, this means that `CpuContainer::new` must also occur in this method. The issue with this is that `CpuContainer::new` allocates MMIO memory and a GSI for ACPI. 

To make this possible `setup_interrupt_controller` must be moved inside `create_vmm_and_vcpus` (so GSIs can be allocated), and the `BootTimer` also has to be moved inside the same method, as it has to occupy the first MMIO address in order to work properly. This has the potential to cause confusion, as now devices are being registered in a place separate from the rest of them, and it leads to `create_vmm_and_vcpus` having unexpected side-effects. By making `CpuContainer` optional again, both issues can be sorted. The `BootTimer` device can be registered first, and `setup_interrupt_controller` can be moved back to it's original location, in `build_microvm_for_boot`.

src/vmm/src/lib.rs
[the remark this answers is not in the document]
jxcurtis: Update: delayed this work until a later point. Snapshot restore is a stretch goal for this project

src/vmm/src/devices/bus.rs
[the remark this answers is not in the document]
roypat: > Hmm. It looks like serial_ref is still in use for some aarch64 code, see lib.rs:488.

Yes, `serial_ref` and `mmio_transport_ref` are still in use, but the 3 functions above them (`boot_timer_ref`, `i8042_device_ref` and `rtc_device_ref`) aren't, so we can remove them to avoid carrying around dead code.

> So although currently unused, should we not keep these functions available so that ARM machines can still access these devices? They could be marked as aarch64 only.

I'm not sure I understand what you mean by this :(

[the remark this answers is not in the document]
jxcurtis: > Yes, serial_ref and mmio_transport_ref are still in use, but the 3 functions above them (boot_timer_ref, i8042_device_ref and rtc_device_ref) aren't, so we can remove them to avoid carrying around dead code.

* To me, it looks like we can get rid of `boot_timer_ref`. There doesn't seem to be a case where we would need a reference to the device after it has been added to the MMIO bus. It writes directly to the logs after boot and then we never need it again. 
* It's a similar story for the `RTCDevice` and `rtc_device_ref`.
* Right now, the I8042 device only seems to be used on x86_64, where it can be accessed through the `PortIODeviceManager`. However, whilst not in use, the device is also available to ARM64 devices. The way that this would be accessed would be using the MMIO bus, and so surely if we support it on ARM, we should keep a method around to access the device?

I am of the opinion that if the `I8042Device` is only going to be used on x86_64, we should tag it as such so that it is clear that this feature is not available to ARM. If there is a chance it is used on ARM systems in the future, and the way that it would be accessed is through the MMIO bus, then I don't know if it makes sense to remove the small function used to access it. Instead we could just tag it as `aarch64` only, so that it is clear that this function is only used to access the i8042 device on ARM systems.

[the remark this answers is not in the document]
roypat: Ah, right. You can tag the `i8042` device as `#[cfg(target_arch = "x86_64")]` if you want, it's only used for the CTRL+ALT+ESC = restart functionality on x86 (which is not supported on ARM). So I don't see why we would ever need this decide on ARM (and if we ever do, we can just readd these function - however, keep in mind that on x86_64, the device can be used without these functions, so why would we suddenly need them on ARM?)

src/vmm/src/device_manager/acpi.rs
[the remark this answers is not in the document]
roypat: Ouch, alright :(

bchalios (commented):
I did a first pass. Have some points that I'd like to clarify, but it generally looks good. I would like to take a better look at the KVM part of the PR, though. Also, I 'd like someone who's more in touch with the normalization code to look at the relevant parts. Maybe @zulinx86 or @ShadowCurse (?)

src/vmm/src/devices/acpi/cpu_container.rs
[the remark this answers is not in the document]
roypat: Let's definitely add the cloudhypervisor copyright. The page you link is written by Rob and Sebastian from CloudHypervisor (from the time they were working at intel), so I'm hoping its fine? Probably a question we can ignore for now, as we're not merging anything into main.

src/vmm/src/devices/bus.rs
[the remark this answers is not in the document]
roypat: > So boot_timer_ref and i8042_device_ref are actually dead code. We should remove them from here as Patrick said. We could also mark the I8042Device as x86 only all over the place. We should actually mark the vmm::devices::legacy::i8042 module x86 only.

Yeah, its not even too big of a change:

<details>

```diff
diff --git a/src/vmm/src/devices/bus.rs b/src/vmm/src/devices/bus.rs
index 2b016d730..e22fccfdf 100644
--- a/src/vmm/src/devices/bus.rs
+++ b/src/vmm/src/devices/bus.rs
@@ -51,15 +51,19 @@ pub struct Bus {
 }
 
 use event_manager::{EventOps, Events, MutEventSubscriber};
+use crate::devices::legacy::SerialDevice;
 
 #[cfg(target_arch = "aarch64")]
 use super::legacy::RTCDevice;
-use super::legacy::{I8042Device, SerialDevice};
+
+#[cfg(target_arch = "x86_64")]
+use super::legacy::{I8042Device};
 use super::pseudo::BootTimer;
 use super::virtio::mmio::MmioTransport;
 
 #[derive(Debug)]
 pub enum BusDevice {
+    #[cfg(target_arch = "x86_64")]
     I8042Device(I8042Device),
     #[cfg(target_arch = "aarch64")]
     RTCDevice(RTCDevice),
@@ -102,12 +106,6 @@ impl ConstantDevice {
 }
 
 impl BusDevice {
-    pub fn i8042_device_ref(&self) -> Option<&I8042Device> {
-        match self {
-            Self::I8042Device(x) => Some(x),
-            _ => None,
-        }
-    }
     #[cfg(target_arch = "aarch64")]
     pub fn rtc_device_ref(&self) -> Option<&RTCDevice> {
         match self {
@@ -133,13 +131,6 @@ impl BusDevice {
             _ => None,
         }
     }
-
-    pub fn i8042_device_mut(&mut self) -> Option<&mut I8042Device> {
-        match self {
-            Self::I8042Device(x) => Some(x),
-            _ => None,
-        }
-    }
     #[cfg(target_arch = "aarch64")]
     pub fn rtc_device_mut(&mut self) -> Option<&mut RTCDevice> {
         match self {
@@ -168,6 +159,7 @@ impl BusDevice {
 
     pub fn read(&mut self, offset: u64, data: &mut [u8]) {
         match self {
+            #[cfg(target_arch = "x86_64")]
             Self::I8042Device(x) => x.bus_read(offset, data),
             #[cfg(target_arch = "aarch64")]
             Self::RTCDevice(x) => x.bus_read(offset, data),
@@ -183,6 +175,7 @@ impl BusDevice {
 
     pub fn write(&mut self, offset: u64, data: &[u8]) {
         match self {
+            #[cfg(target_arch = "x86_64")]
             Self::I8042Device(x) => x.bus_write(offset, data),
             #[cfg(target_arch = "aarch64")]
             Self::RTCDevice(x) => x.bus_write(offset, data),
diff --git a/src/vmm/src/devices/legacy/mod.rs b/src/vmm/src/devices/legacy/mod.rs
index 224f77c7b..4642029d1 100644
--- a/src/vmm/src/devices/legacy/mod.rs
+++ b/src/vmm/src/devices/legacy/mod.rs
@@ -6,6 +6,7 @@
 // found in the THIRD-PARTY file.
 
 //! Implements legacy devices (UART, RTC etc).
+#[cfg(target_arch = "x86_64")]
 mod i8042;
 #[cfg(target_arch = "aarch64")]
 pub mod rtc_pl031;
@@ -19,6 +20,7 @@ use serde::Serializer;
 use utils::eventfd::EventFd;
 use vm_superio::Trigger;
 
+#[cfg(target_arch = "x86_64")]
 pub use self::i8042::{I8042Device, I8042Error as I8042DeviceError};
 #[cfg(target_arch = "aarch64")]
 pub use self::rtc_pl031::RTCDevice;
@@ -67,6 +69,7 @@ impl EventFdTrigger {
 /// Called by METRICS.flush(), this function facilitates serialization of aggregated metrics.
 pub fn flush_metrics<S: Serializer>(serializer: S) -> Result<S::Ok, S::Error> {
     let mut seq = serializer.serialize_map(Some(1))?;
+    #[cfg(target_arch = "x86_64")]
     seq.serialize_entry("i8042", &i8042::METRICS)?;
     #[cfg(target_arch = "aarch64")]
     seq.serialize_entry("rtc", &rtc_pl031::METRICS)?;
diff --git a/src/vmm/src/lib.rs b/src/vmm/src/lib.rs
index ee27194e4..ae1ef1bb1 100644
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -213,6 +213,7 @@ pub enum VmmError {
     /// Event fd error: {0}
     EventFd(io::Error),
     /// I8042 error: {0}
+    #[cfg(target_arch = "x86_64")]
     I8042Error(devices::legacy::I8042DeviceError),
     /// Cannot access kernel file: {0}
     KernelFile(io::Error),
```

</details>

> However, maybe not related exactly with this PR, but could we actually get rid off all this mess, by turning BusDevice type to something like:

It actually used to be a trait: https://github.com/firecracker-microvm/firecracker/commit/0aecb443b8fd23673e37cb9dba82fe2ba5c290b6#diff-a3fcbea0daddf47c324fcbb2ac6738976593ec8f1961a9d7205088b434f1c537

src/vmm/src/builder.rs
```diff
@@ -172,14 +180,28 @@ fn create_vmm_and_vcpus(
...
     // Instantiate ACPI device manager.
     #[cfg(target_arch = "x86_64")]
-    let acpi_device_manager = ACPIDeviceManager::new();
+    let acpi_device_manager = {
+        let cpu_container = Arc::new(Mutex::new(CpuContainer::new(
+            &mut resource_allocator,
+            vcpu_count,
+        )?));
+        ACPIDeviceManager::new(cpu_container)
+    };
```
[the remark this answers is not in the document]
roypat: That'd be my fault - I suggested that we don't use an option, because unlike vmgenid, we always have a cpu container, so having it be an option looked weird to me (and using Option only to delay initialization "smells" wrong). It seemed to cause a bunch of troubles though, so I think our current code architecture forces us down the Option path :(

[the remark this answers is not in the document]
bchalios: Yes, I agree that is not ideal. I don't like it much either (we also always have a VMGenID). We might be able to get away with it if we ever get to restructuring the whole builder part.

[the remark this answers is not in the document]
jxcurtis: I'll revert this change then, putting it back to being optional.

src/vmm/src/acpi/mod.rs
```diff
@@ -272,15 +272,15 @@ pub mod tests {
...
         let err = writer.write_acpi_table(&mut sdt).unwrap_err();
         assert!(
             matches!(
                 err,
                 AcpiError::AcpiTables(acpi_tables::AcpiError::GuestMemory(
                     vm_memory::GuestMemoryError::PartialBuffer {
-                        expected: 263168,  // SYSTEM_MEM_SIZE
-                        completed: 259072  // SYSTEM_MEM_SIZE - 4096
+                        expected: 259072,  // SYSTEM_MEM_SIZE - 4096
+                        completed: 254976  // SYSTEM_MEM_SIZE - 8192
```
ShadowCurse, line 283: Maybe use `SYSTEM_MEM_SIZE` directly? + instead of 4096 can use `MMIO_LEN`.

src/vmm/src/acpi/mod.rs
```diff
@@ -263,7 +263,7 @@ pub mod tests {
     fn test_write_acpi_table_small_memory() {
         let mut vmm = default_vmm();
         vmm.guest_memory = arch_mem(
-            (SYSTEM_MEM_START + SYSTEM_MEM_SIZE - 4096)
+            (SYSTEM_MEM_START + SYSTEM_MEM_SIZE - 8192)
```
ShadowCurse, line 266: General comment to the whole commit: to these changes fix smith broken in previous ones? In that case would be nice to move necessary changes to commits which need them.

jxcurtis: Yes, ACPI tables are larger because of the extra CPUs and the hotplug device, and MMIO addresses have changed because the CPU container is assigned an MMIO address first. I'll move them to their respective commits

src/vmm/src/builder.rs
```diff
@@ -38,6 +38,8 @@ use crate::cpu_config::templates::{
     CpuConfiguration, CustomCpuTemplate, GetCpuTemplate, GetCpuTemplateError, GuestConfigError,
     KvmCapability,
 };
+use crate::cpu_config::x86_64::cpuid::{self, Cpuid};
```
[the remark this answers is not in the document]
jxcurtis: It's actually removed in the next commit, so I'll just remove it here instead.

src/vmm/src/device_manager/acpi.rs
```diff
@@ -48,8 +48,10 @@ impl ACPIDeviceManager {
         cpu_container: Arc<Mutex<CpuContainer>>,
         vm_fd: &VmFd,
     ) -> Result<(), kvm_ioctls::Error> {
-        let locked_container = cpu_container.lock().expect("Poisoned lock");
-        vm_fd.register_irqfd(&locked_container.acpi_interrupt_evt, locked_container.gsi)?;
+        {
+            let locked_container = cpu_container.lock().expect("Poisoned lock");
+            vm_fd.register_irqfd(&locked_container.acpi_interrupt_evt, locked_container.gsi)?;
+        }
```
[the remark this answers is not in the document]
jxcurtis: It is needed, we need to unlock it so we can assign it to the `ACPIDeviceManager`. Will add it where it was introduced.

src/vmm/src/acpi/mod.rs
jxcurtis: Looks like I can't do the subtraction needed during the instantiation of the struct. Would you prefer if I shifted it to a variable above the assert?

jxcurtis: Hmm, when I tried extracting it to a variable, it didn't compare the values inside, so I've left it as is for now.

src/vmm/src/cpu_config/x86_64/cpuid/amd/normalize.rs
```diff
@@ -107,7 +108,7 @@ impl super::AmdCpuid {
         self.update_largest_extended_fn_entry()?;
         self.update_extended_feature_fn_entry()?;
         self.update_amd_feature_entry(cpu_count)?;
-        self.update_extended_cache_topology_entry(cpu_count, cpus_per_core)?;
+        self.update_extended_cache_topology_entry(cpus_per_core)?;
```
zulinx86, line 111: I only looked at the commit "Edit CPUID Normalization to account for possibility of hotplugging", so this might be a basic question, but why can't we update it when hotplugging vcpus instead of boot time?
Software running inside guests (like guest kernel) only checks it at boot time and allocates memory to hold some data that is not extensible afterwards? Or we just don't want to update CPUID/MSR of the existing vcpus?

jxcurtis: The changes that I've made are to the "maximum values". I.e. maximum APIC ID sharing L3 cache, and maximum APIC address per physical package. I'm not sure if this reserves memory at boot time or not. 

From my understanding, the effect of this is that any APIC address that falls within that range is assigned to the cache/package. So all that results from this change is we say that if more vCPUs are added, they will all fall under the current cache/package.  For this feature branch at least, hot-plugging is always available, so I think it makes sense for these fields to be set from boot, so that they don't have to changed later

src/acpi-tables/src/madt.rs
```diff
@@ -28,13 +29,23 @@ pub struct LocalAPIC {
...
+                flags: U32::new(1u32 << MADT_CPU_ONLINE_CAPABLE_FLAG),
+            }
+        } else {
+            Self {
+                r#type: 0,
+                length: 8,
+                processor_uid: cpu_id,
+                apic_id: cpu_id,
+                flags: U32::new(1u32 << MADT_CPU_ENABLE_FLAG),
+            }
```
bchalios, line 48: I think it's more natural to say whether a CPU is enabled or not and make everything else online capable. Also, you can write this more concisely:

```Rust
pub fn new(cpu_id: u8, online: bool) -> Self {
    let flags = if online {
          U32::new(1u32 << MADT_CPU_ENABLE_FLAG)
    } else {
          U32::new(1u32 << MADT_CPU_ONLINE_CAPABLE_FLAG)
    };
    Self {
        r#type: 0,
        length: 8,
        processor_uid: cpu_id,
        apic_id: cpu_id,
        flags
    }
}
```

src/vmm/src/acpi/mod.rs
```diff
@@ -272,15 +273,18 @@ pub mod tests {
...
         let err = writer.write_acpi_table(&mut sdt).unwrap_err();
         assert!(
             matches!(
                 err,
                 AcpiError::AcpiTables(acpi_tables::AcpiError::GuestMemory(
                     vm_memory::GuestMemoryError::PartialBuffer {
-                        expected: 263168,  // SYSTEM_MEM_SIZE
-                        completed: 259072  // SYSTEM_MEM_SIZE - 4096
+                        expected: 259072,  // SYSTEM_MEM_SIZE - MMIO_LEN
+                        completed: 254976, // SYSTEM_MEM_SIZE - MMIO_LEN * 2
```
bchalios, line 287: not related with your PR, but I should have probably used the expressions in the comments rather than the hardcoded numerics. If you feel like, could you please fix it in the next iteration?

bchalios (commented):
Looks good to me apart from a couple of minor comments.

src/vmm/src/cpu_config/x86_64/cpuid/amd/normalize.rs
zulinx86: Fair enough. Sounds good to me :)

src/vmm/src/acpi/mod.rs
jxcurtis: I don't think you can use the expression here, I think we have to use the actual values because we're inside a pattern match. We have to convert the constant values into `usize`, at which point the compiler no longer views them as constant, and therefore pattern matching won't actually match on the exact values, just the structure.

zulinx86 (approved):
At least to me, I didn't find anything to block this to be merged into the feature branch.

src/vmm/src/acpi/mod.rs
bchalios: oh ok, that explains why it was like that in the first place. Thanks for looking into it.

bchalios (approved):

PR #4665, opened 2024-07-03
fix: do not panic if virtio device activation return Err(...)

roypat:
When the guest driver sets a virtio devices status to `DRIVER_OK`, we
proceed with calling `VirtioDevice::activate`. However, our MMIO
transport layer assumes that this activation cannot go wrong, and calls
`.expect(...)` on the result. For most devices, this is fine, as the
activate method doesn't do much besides writing to an event_fd (and I
can't think of a scenario in which this could go wrong). However, our
vhost-user-blk device has some non-trivial logic inside of its
`activate` method, which includes communication with the
vhost-user-backend via a unix socket. If this unix socket gets closed
early, this causes `activate` to return an error, and thus consequently
a panic in the MMIO code.

The virtio spec, in Section 2.2, has the following to say [[1]]:

> The device SHOULD set DEVICE_NEEDS_RESET when it enters an error state
  that a reset is needed. If DRIVER_OK is set, after it sets
  DEVICE_NEEDS_RESET, the device MUST send a device configuration
  change notification to the driver.

So the spec-conform way of handling an activation error is setting
the `DEVICE_NEEDS_RESET` flag in the device_status field (which is what
this commit does).

This will fix the panic, however it will most certainly still not result
in correct device operations (e.g. a vhost-user-backend dying will still
be unrecoverable). This is because Firecracker does not actually
implement device reset, see also https://github.com/firecracker-microvm/firecracker/issues/3074. Thus, the device will simply be
"dead" to the guest. But at least Firecracker won't crash anymore.

[1]: https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.pdf

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/src/vmm/src/devices/virtio/mmio.rs
+++ b/src/vmm/src/devices/virtio/mmio.rs
@@ -363,6 +363,7 @@ pub(crate) mod tests {
     use utils::u64_to_usize;
 
     use super::*;
+    use crate::devices::virtio::device::IrqTrigger;
     use crate::devices::virtio::ActivateError;
     use crate::utilities::test_utils::single_region_mem;
     use crate::vstate::memory::GuestMemoryMmap;
@@ -371,8 +372,7 @@ pub(crate) mod tests {
     pub(crate) struct DummyDevice {
         acked_features: u64,
         avail_features: u64,
-        interrupt_evt: EventFd,
-        interrupt_status: Arc<AtomicU32>,
+        interrupt_trigger: IrqTrigger,
         queue_evts: Vec<EventFd>,
         queues: Vec<Queue>,
         device_activated: bool,
@@ -384,8 +384,7 @@ pub(crate) mod tests {
             DummyDevice {
                 acked_features: 0,
                 avail_features: 0,
-                interrupt_evt: EventFd::new(libc::EFD_NONBLOCK).unwrap(),
-                interrupt_status: Arc::new(AtomicU32::new(0)),
+                interrupt_trigger: IrqTrigger::new().unwrap(),
                 queue_evts: vec![
                     EventFd::new(libc::EFD_NONBLOCK).unwrap(),
                     EventFd::new(libc::EFD_NONBLOCK).unwrap(),
@@ -430,12 +429,8 @@ pub(crate) mod tests {
             &self.queue_evts
         }
 
-        fn interrupt_evt(&self) -> &EventFd {
-            &self.interrupt_evt
-        }
-
-        fn interrupt_status(&self) -> Arc<AtomicU32> {
-            self.interrupt_status.clone()
+        fn interrupt_trigger(&self) -> &IrqTrigger {
+            &self.interrupt_trigger
         }
 
         fn read_config(&self, offset: u64, data: &mut [u8]) {
--- a/src/vmm/src/devices/virtio/mmio.rs
+++ b/src/vmm/src/devices/virtio/mmio.rs
@@ -14,7 +14,7 @@ use utils::byte_order;
 use crate::devices::virtio::device::{IrqType, VirtioDevice};
 use crate::devices::virtio::device_status;
 use crate::devices::virtio::queue::Queue;
-use crate::logger::warn;
+use crate::logger::{error, warn};
 use crate::vstate::memory::{GuestAddress, GuestMemoryMmap};
 
 // TODO crosvm uses 0 here, but IIRC virtio specified some other vendor id that should be used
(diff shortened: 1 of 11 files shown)


src/vmm/src/devices/virtio/mmio.rs
```diff
@@ -186,10 +186,18 @@ impl MmioTransport {
             DRIVER_OK if self.device_status == (ACKNOWLEDGE | DRIVER | FEATURES_OK) => {
                 self.device_status = status;
                 let device_activated = self.locked_device().is_activated();
-                if !device_activated && self.are_queues_valid() {
-                    self.locked_device()
-                        .activate(self.mem.clone())
-                        .expect("Failed to activate device");
+                if !device_activated
+                    && self.are_queues_valid()
+                    && self.locked_device().activate(self.mem.clone()).is_err()
```
kalyazin, line 191: nit: I don't quite like having a stateful call (activate) inside an `if` condition. I guess it reduces nesting, but may be a little harder to read.

src/vmm/src/devices/virtio/mmio.rs
```diff
@@ -186,10 +186,18 @@ impl MmioTransport {
...
+                {
+                    self.device_status |= DEVICE_NEEDS_RESET;
+
+                    // Section 2.1.2 of the specification states that we need to send a device
+                    // configuration change interrupt
+                    let _ = self
+                        .locked_device()
+                        .interrupt_trigger()
+                        .trigger_irq(IrqType::Config);
```
kalyazin, line 200: would we log the occurrence to aid debugging?

kalyazin (commented):
>however it will most certainly still not result
in correct device operations (e.g. a vhost-user-backend dying will still
be unrecoverable)

When you speak about recovery, do you mean Firecracker will need to try to reconnect to the backend when the device is reset? This isn't going to solve the problem for `test_vhost_user_block_disconnect` where the backend never comes back though?

src/vmm/src/devices/virtio/mmio.rs
roypat: Me neither, but clippy insists :(

src/vmm/src/devices/virtio/mmio.rs
roypat: Mh, when I wrote this I decided against it because there are usually log statements inside of `activate`, but I just realized that's not the case for vhost - let me add some logging here :)

roypat:
> > however it will most certainly still not result
> > in correct device operations (e.g. a vhost-user-backend dying will still
> > be unrecoverable)
> 
> When you speak about recovery, do you mean Firecracker will need to try to reconnect to the backend when the device is reset? This isn't going to solve the problem for `test_vhost_user_block_disconnect` where the backend never comes back though?

Yeah, it won't solve the problem where the backend never comes online, I would expect in this case the device will just remain in either `DEVICE_NEEDS_RESET` or `FAILED` indefinitely. It's hard to tell, because Linux actually also does not implement this part of the virtio spec, lol. I'd expect a well-behaving guest to not retry indefinitely, and just mark the device as FAILED after, say, 3 failed resets.

src/vmm/src/devices/virtio/mmio.rs
roypat: It's our lucky day, the logging means clippy no longer insists on things being this way

src/vmm/src/devices/virtio/mmio.rs
roypat: done

src/vmm/src/devices/virtio/mmio.rs
```diff
@@ -186,10 +186,18 @@ impl MmioTransport {
...
                 let device_activated = self.locked_device().is_activated();
-                if !device_activated && self.are_queues_valid() {
-                    self.locked_device()
-                        .activate(self.mem.clone())
-                        .expect("Failed to activate device");
+                if !device_activated
+                    && self.are_queues_valid()
+                    && self.locked_device().activate(self.mem.clone()).is_err()
+                {
+                    self.device_status |= DEVICE_NEEDS_RESET;
```
ShadowCurse: I think for vhost-user-block we should also set this when socket connection is closed, i.e. vhost-user-backend crashed. Maybe in later PR.

src/vmm/src/devices/virtio/mmio.rs
```diff
@@ -186,18 +186,21 @@ impl MmioTransport {
...
+                        self.device_status |= DEVICE_NEEDS_RESET;
+
+                        // Section 2.1.2 of the specification states that we need to send a device
+                        // configuration change interrupt
+                        let _ = self
+                            .locked_device()
+                            .interrupt_trigger()
+                            .trigger_irq(IrqType::Config);
+
+                        log::error!("Failed to activate virtio device: {}", err)
```
ShadowCurse, line 202: nit: just import `log::error` (or `logger::error`, I think we should use this everywhere for consistency)

roypat: done

src/vmm/src/devices/virtio/mmio.rs
roypat: Yeah, let's keep that for another PR, that sounds a bit more complicated :sweat_smile:

ShadowCurse (approved):

kalyazin (approved):

PR #4666, opened 2024-07-05
fix: Always restore MSR_IA32_TSC_DEADLINE after MSR_IA32_TSC

### LINKED ISSUE #4099: [Bug] Processes get stuck after resuming VM from snapshot
# Describe the bug

After resuming a VM from snapshot, processes occasionally get stuck. A minimal example is an `init` binary that just runs `while (true) { sleep 100ms ; print 'hello' }` - after resuming from snapshot, it only sometimes is able to resume the loop, and other times it gets stuck and does not print anything at all.

## To Reproduce

1. Clone the following repo, which has the minimal code needed to reproduce: https://github.com/bduffany/firecracker-sleep-issue
2. Make sure you have `make`, `cc` (C compiler, from `gcc` package), `glibc-static` (for statically linking the init binary), and `jq` (for parsing firecracker API output)
3. Run `make test`

`make test` is doing the following:
- Fetches the firecracker v1.4.0 release binary from GitHub as well as vmlinux 4.14 from the https://github.com/firecracker-microvm/firecracker-demo repo
- Runs firecracker with an initrd where the init binary just loops infinitely, printing "running" then sleeping for 100ms. 
- Streams the VM logs to the kernel (using `tail -f` in the background)
- Every 1s, pauses the VM, takes a snapshot, then kills firecracker (SIGTERM), then restarts firecracker, resuming the VM from snapshot.

## Expected behaviour

When running `make test` in that repo, the init binary should print `running` several times after each resume. But on some resumes, it appears stuck, and does not print anything until the next resume.

Interesting details:
- The issue does NOT reproduce if I replace the `nanosleep` syscalls with an NOP loop of 1e9 iterations (`for (int i=0; i<1e9; i++) continue;`)
- The issue seems specific to snapshotting, not just pausing and resuming the VM. I tried just doing pause/resume without taking a snapshot and then restarting the `firecracker` binary in between, but could not reproduce.
- I could not reproduce this on an Intel CPU so far, only AMD (have tried 2 different Intel machines and 2 different AMD machines).

## Environment

- Firecracker v1.4.0
- Host kernel v6.2.0 (Ubuntu 22.04)
  - **UPDATE:** also reproduced on host kernel v5.10 - `m6a.metal` instance
- Guest kernel 4.14.55-84.37.amzn2.x86_64, from https://github.com/firecracker-microvm/firecracker-demo
  - Also tried compiling 5.10 using the recommended guest config - the issue still reproduces.
- Rootfs: none (initrd only)
- Architecture: x86_64 (AMD Ryzen CPU)
  - **UPDATE:** Also reproduced with AMD EPYC (`m6a.metal` instance)
- GLIBC 2.35 (for init binary)

## Additional context

This repro above is a minimal example of a much more troublesome issue where we are having trouble reconnecting to microVMs after resuming them from snapshots. We are running a server inside the VM and are having trouble connecting to over vsock. I suspected that the guest process was "stuck" somehow since when running `sleep 1 && print('hello')` in a background loop, it sometimes doesn't print anything. I came up with this minimal reproducer for this behavior.

## Checks

- Have you searched the Firecracker Issues database for similar problems?
  - https://github.com/firecracker-microvm/firecracker/issues/3020 - not sure if this is related, but it was closed without resolution. The final comment said to re-open if it could be reproduced with a supported kernel version
- Have you read the existing relevant Firecracker documentation?
  - I have read the FAQ about guest clock drift / NTP, but this appears more significant than just clock drift since I think nanosleep should work based on relative timing? I could be wrong, though.
- Are you certain the bug being reported is a Firecracker issue?
  - Not 100% certain, but given that it happens only when loading a snapshot, it seems like it could be Firecracker-related

roypat:
When restoring MSRs, we currently do not enforce any sort of relative ordering. However, some MSRs have implicit dependencies on other MSRs being restored before them, and failing to fulfill these dependencies can result in incorrect VM behavior after resuming.

One example of such an implicit dependency between MSRs is the pair (`MSR_IA32_TSC_DEADLINE`, `MSR_IA32_TSC`). When restoring `MSR_IA32_TSC_DEADLINE`, KVM internally checks whether the value of this restoration implies that the guest was waiting for the tsc_deadline timer to expire at the time of being paused for snapshotting. If yes, it primes a (either hardware or software depending on support) timer on the host to make sure the guest will receive the expected interrupt after restoration. To determine whether this is needed, KVM looks at the guest's timestamp counter (TSC) and compares it with the requested tsc_deadline value. The value KVM reads for the guest's TSC depends on the value of `MSR_IA32_TSC` [[6]][[7]]. Thus, if `MSR_IA32_TSC_DEADLINE` is set before `MSR_IA32_TSC` is restored, this comparison will yield a wrong result (as the deadline value is compared with something uninitialized) [[5]]. This can either result in KVM determining the guest wasn't waiting for a timing expiry at the time of snapshotting, or cause it to schedule the timer interrupt too far into the future.

Note that the former is only a problem on AMD platforms, which do not support the TSC_DEADLINE feature at the hardware level. Here, KVM falls back to a software timer [[1]], which explicitly does not notify the vCPU if the deadline value is "in the past" [[2]][[3]]. The hardware timer used on other x86 platforms on the other hand always fires (potentially firing immediately if the deadline value is in the past) [[4]].

This commit fixes the above by ensuring we always restore `MSR_IA32_TSC` before restoring `MSR_IA32_TSC_DEADLINE`. We realize this by splitting the lists of MSRs that KVM gives us into one additional chunk containing all "deferred" MSRs that needs to be restored "as late as possible". This splitting happens at snapshot creation time, to get it off the hot-path.

Fixes #4099 (validated by running the reproducer script for 50 iterations on an m6a.metal and not observing the issue anymore)

[1]: https://elixir.bootlin.com/linux/v6.1.97/source/arch/x86/kvm/lapic.c#L1995
[2]: https://elixir.bootlin.com/linux/v6.1.97/source/arch/x86/kvm/lapic.c#L1759
[3]: https://elixir.bootlin.com/linux/v6.1.97/source/arch/x86/kvm/lapic.c#L1790
[4]: https://elixir.bootlin.com/linux/v6.1.97/source/arch/x86/kvm/lapic.c#L1928
[5]: https://elixir.bootlin.com/linux/v6.1.97/source/arch/x86/kvm/lapic.c#L1784
[6]: https://elixir.bootlin.com/linux/v6.1.97/source/arch/x86/kvm/x86.c#L2560
[7]: https://elixir.bootlin.com/linux/v6.1.97/source/arch/x86/kvm/x86.c#L2596

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -316,6 +327,39 @@ impl KvmVcpu {
...
+        // KVM will later yell at us about invalid MSRs.
+        let mut deferred_msrs = Msrs::new(0)?;
+
+        for msrs in msr_chunks {
+            msrs.retain(|msr| {
+                if DEFERRED_MSRS.contains(&msr.index) {
+                    deferred_msrs
+                        .push(*msr)
+                        .inspect_err(|err| {
+                            error!(
```
kalyazin, line 348: what could be the cause of this? Shall we have a test to cover this case?

kalyazin (commented):
1. Well done resolving it!
2. We should mention this in the changelog.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -33,6 +34,16 @@ use crate::vstate::vm::Vm;
 const TSC_KHZ_TOL_NUMERATOR: i64 = 250;
 const TSC_KHZ_TOL_DENOMINATOR: i64 = 1_000_000;
 
+/// A set of MSRs that should be restored separately after all other MSRs have already been restored
+const DEFERRED_MSRS: [u32; 1] = [
+    // MSR_IA32_TSC_DEADLINE must be restored after MSR_IA32_TSC, otherwise we risk "loosing" timer
```
pb8o, line 39: typo: losing

src/vmm/src/vstate/vcpu/x86_64.rs
roypat: The vmm-sys-util methods can only fail because of various allocation failures (like "fam struct length exceeds isize::MAX"), so I don't think we can test this

src/vmm/src/vstate/vcpu/x86_64.rs
roypat: This is one of those words I always get wrong, thanks!

roypat: fixed

roypat:
> 1. Well done resolving it!

thanks!!

>     2. We should mention this in the changelog.

True, done :)

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -348,6 +393,9 @@ impl KvmVcpu {
 
         Self::fix_zero_tsc_deadline_msr(&mut msr_chunks);
 
+        let deferred = Self::defer_msrs(&mut msr_chunks)?;
+        msr_chunks.push(deferred);
```
ShadowCurse, line 397: The name `defer_msrs` in my mind sounds like we are differing them in the method, but we actually just remove them. So maybe either rename `defer_msrs` to `remove_defer_msrs` or move `msr_chunks.push(deferred);` into the `defer_msrs` as well?

kalyazin (dismissed):

zulinx86 (dismissed):
Well done, Patrick!!!

src/vmm/src/vstate/vcpu/x86_64.rs
roypat: Good point, originally I wanted to have the `push` inside the function, but couldn't because of it only taking a slice. But then I never updated the function name to reflect what is actually going on. Now its called `extract_deferred_msrs`.

zulinx86 (dismissed):

src/vmm/src/cpu_config/x86_64/mod.rs
```diff
@@ -187,14 +187,14 @@ mod tests {
...
-            msrs: HashMap::from([(0x8000, 0b1000), (0x9999, 0b1010)]),
+            msrs: BTreeMap::from([(0x8000, 0b1000), (0x9999, 0b1010)]),
         }
     }
 
     fn unsupported_cpu_config() -> CpuConfiguration {
         CpuConfiguration {
             cpuid: build_supported_cpuid(),
-            msrs: HashMap::from([(0x8000, 0b1000), (0x8001, 0b1010)]),
+            msrs: BTreeMap::from([(0x8000, 0b1000), (0x8001, 0b1010)]),
```
pb8o, line 197: Could we add any more tests to validate that insertion order is preserved?

roypat: Good point, I gave it a try in the newest revision!

ShadowCurse (approved):

pb8o (approved):

PR #4658, opened 2024-06-28
Add VIRTIO_NET_F_MRG_RXBUF to virtio-net 

ShadowCurse:
## Changes
Add `VIRTIO_NET_F_MRG_RXBUF` to virtio-net.

Now virtio-net device can split incoming packets across multiple descriptor chains if `VIRTIO_NET_F_MRG_RXBUF` is enabled by the guest. The amount of descriptor chains (also known as heads) is written into the `virtio_net_hdr_v1` structure which is located at the very begging of the packet. Virtio spec states that the number of heads used should always be correct:
- 1 - if `VIRTIO_NET_F_MRG_RXBUF` is not negotiated
- N - if `VIRTIO_NET_F_MRG_RXBUF` is negotiated
Prior to this commit Firecracker never set the number of used heads to 1, but Linux was fine with it. Now we always set correct number of heads. Because of this some changes were introduced into the unit test code that was generating testing frames.

Fixes: #1314

## Reason
Better performance and guest memory utilization.

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following Developer
Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this
  PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet
  [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

______________________________________________________________________

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #4683, opened 2024-07-16
Virtio tweaks

ShadowCurse:
## Changes
Minor tweaks to the virtio queue methods:
- more explicit offset calculations
- more descriptive names and methods

## Reason
These changes were part of #4658, but are quite independent of it, so moved to separate PR.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/src/vmm/src/devices/virtio/net/device.rs
+++ b/src/vmm/src/devices/virtio/net/device.rs
@@ -264,7 +264,11 @@ impl Net {
         &self.tx_rate_limiter
     }
 
-    fn signal_used_queue(&mut self, queue_type: NetQueue) -> Result<(), DeviceError> {
+    /// Trigger queue notification for the guest if we used enough descriptors
+    /// for the notification to be enabled.
+    /// https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-320005
+    /// 2.6.7.1 Driver Requirements: Used Buffer Notification Suppression
+    fn try_signal_queue(&mut self, queue_type: NetQueue) -> Result<(), DeviceError> {
         // This is safe since we checked in the event handler that the device is activated.
         let mem = self.device_state.mem().unwrap();
 
@@ -552,9 +556,7 @@ impl Net {
             }
         }
 
-        // At this point we processed as many Rx frames as possible.
-        // We have to wake the guest if at least one descriptor chain has been used.
-        self.signal_used_queue(NetQueue::Rx)
+        self.try_signal_queue(NetQueue::Rx)
     }
 
     // Process the deferred frame first, then continue reading from tap.
@@ -566,7 +568,7 @@ impl Net {
             return self.process_rx();
         }
 
-        self.signal_used_queue(NetQueue::Rx)
+        self.try_signal_queue(NetQueue::Rx)
     }
 
     fn resume_rx(&mut self) -> Result<(), DeviceError> {
@@ -647,7 +649,7 @@ impl Net {
             self.metrics.no_tx_avail_buffer.inc();
         }
 
-        self.signal_used_queue(NetQueue::Tx)?;
+        self.try_signal_queue(NetQueue::Tx)?;
 
         // An incoming frame for the MMDS may trigger the transmission of a new message.
         if process_rx_for_mmds {
--- a/src/vmm/src/devices/virtio/queue.rs
+++ b/src/vmm/src/devices/virtio/queue.rs
@@ -462,7 +462,7 @@ impl Queue {
         // This fence ensures all descriptor writes are visible before the index update is.
         fence(Ordering::Release);
 
-        self.set_next_used(self.next_used.0, mem);
+        self.set_used_ring_idx(self.next_used.0, mem);
         Ok(())
     }
 
@@ -525,7 +525,7 @@ impl Queue {
 
     /// Helper method that writes to the `avail_event` field of the used ring.
     #[inline(always)]
-    fn set_avail_event<M: GuestMemory>(&mut self, avail_event: u16, mem: &M) {
+    fn set_used_ring_avail_event<M: GuestMemory>(&mut self, avail_event: u16, mem: &M) {
         debug_assert!(self.is_layout_valid(mem));
 
         // Used ring has layout:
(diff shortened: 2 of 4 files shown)


src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -553,9 +556,11 @@ impl Net {
             }
         }
 
-        // At this point we processed as many Rx frames as possible.
-        // We have to wake the guest if at least one descriptor chain has been used.
-        self.signal_used_queue(NetQueue::Rx)
+        // We only notify guest if there is at least one packet processed successfully.
+        if signal_queue {
+            self.signal_used_queue(NetQueue::Rx)?;
```
roypat, line 561: We still only signal once, no matter how many descriptors we put into the used ring, right? So we're still not completely spec compliant:

> After the device writes a descriptor index into the used ring:
> – If flags is 1, the device SHOULD NOT send a notification.
> – If flags is 0, the device MUST send a notification

(from Section 2.6.7.2: https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.pdf)

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -49,6 +51,17 @@ struct Descriptor {
 // SAFETY: `Descriptor` is a POD and contains no padding.
 unsafe impl ByteValued for Descriptor {}
 
+/// A virtio used element in the used ring.
+#[repr(C)]
+#[derive(Default, Clone, Copy)]
+struct UsedElement {
+    id: u32,
```
roypat, line 58: nit:
```suggestion
    used_idx: u32,
```

roypat (dismissed):
Seems reasonable to me, and the offset calculations do indeed look clearer now

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: This is from `If the VIRTIO_F_EVENT_IDX feature bit is not negotiated: ` section. But for us, the `VIRTIO_F_EVENT_IDX` is basically always enabled.
According to this:
```
After the device writes a descriptor index into the used ring:

    If the idx field in the used ring (which determined where that descriptor index was placed) was equal to used_event, the device MUST send a notification.
    Otherwise the device SHOULD NOT send a notification.
```
I think we are quite complaint.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -430,31 +443,52 @@ impl Queue {
...
+        self.set_next_used(self.next_used.0, mem);
+        Ok(())
+    }
+
+    fn write_used_ring<M: GuestMemory>(
+        &self,
+        mem: &M,
+        index: u16,
+        id: u16,
+        len: u32,
```
bchalios, line 464: I would pass here a `UsedElement` type instead of the fields of it.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -488,13 +522,40 @@ impl Queue {
...
+            + std::mem::size_of::<UsedElement>() * usize::from(self.actual_size());
         let avail_event_addr = self
             .used_ring
-            .unchecked_add(u64::from(4 + 8 * self.actual_size()));
+            .unchecked_add(usize_to_u64(avail_event_offset));
 
         mem.write_obj(val, avail_event_addr).unwrap();
     }
 
+    fn set_next_used<M: GuestMemory>(&mut self, val: u16, mem: &M) {
```
bchalios: nit: Instead of `val: u16`, use `next_used_idx: u16` to make the name a bit more descriptive.

bchalios: Can you please add here `#[inline(always)]`? We only use it in one place.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -553,9 +556,11 @@ impl Net {
             }
         }
 
-        // At this point we processed as many Rx frames as possible.
-        // We have to wake the guest if at least one descriptor chain has been used.
-        self.signal_used_queue(NetQueue::Rx)
+        // We only notify guest if there is at least one packet processed successfully.
```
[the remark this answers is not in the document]
ShadowCurse: the `signal_used_queue` will only signal queue if we used more or equal number of descriptors the guest told us to use before sending the notification. 
I guess the confusion I had with `signal_used_queue` is that it actually does not guarantee  that the guest will be notified at there is an internal logic that can prevent it. I can additionally rename the method to smth like `try_signal_queue` to make it more clear. But still I think the logic: `if we did not use any descriptors -> we don't need to even try to notify guest` is still valid and I would prefer for it to be easily visible.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -49,6 +51,17 @@ struct Descriptor {
 // SAFETY: `Descriptor` is a POD and contains no padding.
 unsafe impl ByteValued for Descriptor {}
 
+/// A virtio used element in the used ring.
```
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: done

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
bchalios: > the signal_used_queue will only signal queue if we used more or equal number of descriptors the guest told us to use before sending the notification.

Exactly! My point was: if we have used one descriptor, with your logic, we will call `signal_used_queue` , but if we haven't reached the limit the guest told us, we won't send the notification. The logic inside `signal_used_queue` is there to calculate when we should actually raise an interrupt. So, it supersedes the logic you added here.

Your logic makes sense only as an optimization, but with notification suppression this optimization is redundant. Actually, if anything, I'd bet it adds (probably negligible) overhead due to all the `if` branches all over the place and it makes the code (IMHO) less readable. So I don't want us to add it.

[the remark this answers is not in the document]
ShadowCurse: Ok, I would argue about what is more expensive: syscal or one if statement, but I guess I can drop this change.
But at least I will rename the method to `try_signal_queue` as it is a more fitting name here.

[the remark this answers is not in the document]
bchalios: What syscall? In any case, almost 100% of the times we actually do process a descriptor from the queue. So almost 100% of the times the if statement is `true`.

bchalios (dismissed):
Cool changes. I do agree the offset calculations are much cleaner and safer now.

kalyazin (approved):

bchalios (approved):

PR #4716, opened 2024-08-01
fix(ab): do not leave around tmp branches

roypat:
Delete temporary branches created during checkout in the A/B process again, to avoid polluting checkouts with weirdly named branches.

In the shared build step, this fixes a bug where we couldn't use CI to do an A/B-Test across an empty commit range (e.g. with REVISION_A and REVISION_B set to the same commit). This can be useful to sanity test the A/B-test framework (e.g. we know that the null hypothesis "performance did not change" is true, so if A/B-Tests fail in this scenario, something is wrong). Note that we won't compile the same Firecracker revision twice, as we don't delete the build artifacts.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


kalyazin (approved):

.buildkite/common.py
```diff
@@ -186,8 +186,11 @@ def ab_revision_build(revision):
         f"if ! git cat-file -t $$commitish; then commitish=origin/{revision}; fi",
         "branch_name=tmp-$$commitish",
         "git branch $$branch_name $$commitish",
+        f"if [ ! -d build/{revision} ]; then",
         f"git clone -b $$branch_name . build/{revision}",
         f"cd build/{revision} && ./tools/devtool -y build --release && cd -",
+        "git branch -D $$branch_name",
+        "fi",
```
bchalios, line 193: Hmmm, does this mean that if (for whatever reason) the branch folder already exists we will not build? If this happens, does it mean we will use an old checkout? It looks like it.

I guess this cannot really happen since we later delete it, but (unless I 'm missing something) the code like that does not reflect our intention

roypat: Actually, mh, we can't delete the repository because the upload step needs it. I guess we can just have a simple step that checks if "revision a == revision b" and in this case only does the clone + build thing once

bchalios (approved):

kalyazin (approved):

PR #4723, opened 2024-08-06
virtio: skip redundant memory check

ihciah:
## Changes

Remove redundant memory check for `DescriptorChain::new`.

## Reason

Since we already validated the queue layout and desc index, there's no need to check it again.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/src/vmm/src/devices/virtio/queue.rs
+++ b/src/vmm/src/devices/virtio/queue.rs
@@ -96,6 +96,9 @@ pub struct DescriptorChain<'a, M: GuestMemory = GuestMemoryMmap> {
 }
 
 impl<'a, M: GuestMemory> DescriptorChain<'a, M> {
+    /// Creates a new `DescriptorChain` from the given memory and descriptor table.
+    ///
+    /// Note that the desc_table and queue_size are assumed to be validated by the caller.
     fn checked_new(
         mem: &'a M,
         desc_table: GuestAddress,
@@ -106,8 +109,9 @@ impl<'a, M: GuestMemory> DescriptorChain<'a, M> {
             return None;
         }
 
-        let desc_head = mem.checked_offset(desc_table, (index as usize) * 16)?;
-        mem.checked_offset(desc_head, 16)?;
+        // There's no need for checking as we already validated the descriptor table and index
+        // bounds.
+        let desc_head = desc_table.unchecked_add(u64::from(index) * 16);
 
         // These reads can't fail unless Guest memory is hopelessly broken.
         let desc = match mem.read_obj::<Descriptor>(desc_head) {


src/vmm/src/devices/virtio/queue.rs
```diff
@@ -106,8 +106,8 @@ impl<'a, M: GuestMemory> DescriptorChain<'a, M> {
             return None;
         }
 
-        let desc_head = mem.checked_offset(desc_table, (index as usize) * 16)?;
-        mem.checked_offset(desc_head, 16)?;
+        // There's no need for checking as we already validated the descriptor table and index bounds.
+        let desc_head = desc_table.unchecked_add((index as usize) * 16);
```
roypat, line 110: think it needs to be
```suggestion
        let desc_head = desc_table.unchecked_add((index as u64) * 16);
```
for the compiler to be happy here

roypat (commented):
You're right, this check is superfluous and can be omitted without marking the function as unsafe. The same range checks on guest memory are also performed by `mem.read_obj` a few lines down, and as long as the caller validated `index` to be less than `queue_size` (happens in `DescriptorChain::has_next`), and that the in-memory layout of the virtio queues is valid (happens in `Queue::is_valid`), we are even guaranteed to read from inside the virtio queue structure. However, reading from inside the queue structure is just a functional requirement, not a safety one in Rust's sense. 

I think we should still add a doc comment to the function that notes this additional requirement (e.g. callers need to make sure that index and queue layout are valid). All current callers do indeed guarantee this, so no further code changes are needed.

src/vmm/src/devices/virtio/queue.rs
ihciah: Sorry it was my mistake. I've fixed it and added doc for `DescriptorChain::checked_new`.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -106,8 +109,8 @@ impl<'a, M: GuestMemory> DescriptorChain<'a, M> {
             return None;
         }
 
-        let desc_head = mem.checked_offset(desc_table, (index as usize) * 16)?;
-        mem.checked_offset(desc_head, 16)?;
+        // There's no need for checking as we already validated the descriptor table and index bounds.
+        let desc_head = desc_table.unchecked_add((index as u64) * 16);
```
ShadowCurse, line 113: We don't allow `as` conversions, so you need to replace `index as u64` -> `u64::from(index)`

ShadowCurse (commented):
Good change. Just couple of things to fix:
- Add body to the commit. Smth like: `Remove redundant memory check in descriptor chain creation`.
- Run `cargo fmt` or just manually fix: https://buildkite.com/firecracker/firecracker-pr/builds/10597#0191367a-6c71-4977-90f5-3df557e5a8a4/62-194.

roypat (approved):

src/vmm/src/devices/virtio/queue.rs
ihciah: Thank you! The commit body is added as you suggested.

However, my `cargo fmt` didn't find the comment width issue. I believe it is because I didn't corrently setup git hook.

But to make it work for others like me, I'm wondering if `/tests/fmt.toml` can be moved or copied to `/rustfmt.toml`? It will be used automatically when `cargo fmt`. I can raise a PR if this acceptable.

roypat: Sadly, specifying the option in rustfmt.toml would require a nightly compiler (at least last time I checked). However, we do have `./tools/devtool fmt`, which does use `fmt.toml` and should format everything for you :)

PR #4731, opened 2024-08-15
Allow snapshot tap changes

andrewla:
## Changes

Allow renaming of tap devices on snapshot restore

## Reason

In some scenarios it is not possible to use the jailer, especially in limited privilege environments where the security is external to firecracker itself. But in these cases a snapshot may have to use a different tap device than the one that it was using when it was snapshotted.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [X] Commits meet


pb8o:
Hi @andrewla thank you for your contribution! We would like to understand the use case better in case it can be resolved through other means first. We recommend using a network namespace where you can create TAP devices with the same name, but that probably requires `CAP_SYS_ADMIN`, which I understand is what you mean with "limited privilege environments".

Could you elaborate on your use case? Is there a way you could create the namespace in a privileged setting and then use something like `nsenter firecracker ...`?

andrewla:
That assessment is correct -- basically to run the jailer in a network namespace you need the setns syscall which requires CAP_SYS_ADMIN. So nsenter is not an option.

Our particular case is running in a containerized environment where our privileges are limited by the nature of the general environment. Once we're in our particular container we have lost all relevant privileges.

pb8o:
Hi again @andrewla, we have been talking internally about this PR and we may need to spend some time to decide on the API aspects of it to make sure it doesn't conflict with other efforts. 

In the meantime, we thought of another workaround. The `snapshot-editor` could be enhanced to rename the tap devices in an snapshot file. That would be an easier decision for us, but we want to make sure it would handle your use case.

For example we imagine the tool would work like this:

```sh
snapshot-editor edit-vmstate rename-network eth0 tap1
```

Would this work within your environment?

andrewla:
This was our initial approach as it required minimal changes. But we found that the performance cost of making the copy (as opposed to hardlinking) during the operation (plus serde costs) were more expensive than we were willing to tolerate in our environment.

andrewla:
Hi @pb8o -- is there anything we can do to help move this forward?

pb8o:
Hi @andrewla I haven't had time to look at this, but this is next on my list now. Thanks for your patience!

kanpov:
On a related note, another reason why renaming the tap device is a better approach than namespaced NAT from the "Network for Clones" guide is that the namespaced NAT imposes measurable overhead onto the host kernel due to the addition of about 5 more `iptables`/`nft` rules, plus an RTNETLINK route for forwarding the guest IP out of the netns.

Even though I made an effort to support namespaced NAT in [fcnet](https://github.com/kanpov/fcnet), it increased complexity by a factor of 4-5x in comparison to regular NAT only to support one usecase: two simultaneous microVM clones. So I'd be in favor of this change, or a `snapshot-editor` equivalent.

pb8o:
Hello @andrewla ! I apologize for the long time between updates, but some other stuff came up. So we have decided to go ahead with this. I gave a first initial review and I only have some minor comments, but mostly looks good to me. I just have a question if the `netowrk_overrides` field also works when starting from a JSON config file.

docs/snapshotting/network-for-clones.md
```diff
@@ -146,6 +146,20 @@ Otherwise, packets originating from the guest might be using old Link Layer
...
 connectivity will work both ways even without an explicit flush.
 
+### Renaming host device names
+
+In some environments where the jailer is not being used, restoring a snapshot
+may be tricky because the tap device on the host will not be the same as the tap
+device that the original VM was mapped to when it was snapshotted, as when the
+tap device come from a pool of such devices.
+
+In this case you can use the network_overrides parameter to snapshot restore to
```
pb8o, line 156: nit: quote it `network_overrides`

tests/framework/microvm.py
```diff
@@ -992,13 +1001,13 @@ def thread_backtraces(self):
...
 
-    def wait_for_up(self, timeout=10):
+    def wait_for_up(self, timeout=10, iface=0):
         """Wait for guest running inside the microVM to come up and respond.
 
         :param timeout: seconds to wait.
         """
         try:
-            rc, stdout, stderr = self.ssh.run("true", timeout)
+            rc, stdout, stderr = self.ssh_iface(iface).run("true", timeout)
```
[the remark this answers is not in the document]
andrewla: Will remove. This was for a testing approach I later backed out of, as was the addition of the `ssh_iface` method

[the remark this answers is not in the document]
andrewla: No, I guess the ssh_iface function was already there

andrewla:
Re: config -- currently there is no config support for snapshots (https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/resources.rs) -- the snapshot configuration and restore has to be done with a running firecracker instance

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -577,3 +579,31 @@ def test_vmgenid(guest_kernel_linux_6_1, rootfs, microvm_factory, snapshot_type)
 
         # Update the base for next iteration
         base_snapshot = snapshot
+
+
+def test_snapshot_rename_interface(uvm_nano, microvm_factory):
```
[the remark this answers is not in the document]
andrewla: Snapshots and config have a strange relationship because much of the config is loaded from the snapshot rather than from the command line or config file, so I don't think there's a lot we can do here.

src/firecracker/swagger/firecracker.yaml
```diff
@@ -1216,6 +1216,24 @@ definitions:
           Type of snapshot to create. It is optional and by default, a full
           snapshot is created.
 
+  NetworkOverride:
+    type: object
+    description:
+      Allows for changing the mapping between tap devices and host devices
+      during snapshot restore
```
bchalios, line 1223: Tap devices _are_ host devices. Suggestion below:

```
Allows for changing the backing TAP device of a network interface during snapshot restore.
```

src/vmm/src/builder.rs
```diff
@@ -448,6 +448,8 @@ pub enum BuildMicrovmFromSnapshotError {
     ACPIDeviManager(#[from] ACPIDeviceManagerRestoreError),
     /// VMGenID update failed: {0}
     VMGenIDUpdate(std::io::Error),
+    /// Unknown Network Device.
+    UnknownNetworkDevice,
```
bchalios, line 452: I don't believe this error variant belongs here. I admit that our error handling structure is not the most intuitive, but I believe this error belongs in the `SnasphotStateFromFileError` inside `src/vmm/src/persist.rs`, which is where you emit the error from as well. I _think_ that if you plug it in there everything should propagate upwards nicely.

tests/framework/microvm.py
```diff
@@ -996,11 +997,19 @@ def restore_from_snapshot(
         # Adjust things just in case
         self.kernel_file = Path(self.kernel_file)
 
+        iface_overrides = []
+        if rename_interfaces:
```
bchalios, line 1001: nit: `if rename_interfaces is not None`

tests/framework/microvm.py
```diff
@@ -1060,6 +1069,27 @@ def _dump_debug_information(self, exc: Exception):
...
+                f"Thread backtraces:\n{self.thread_backtraces}"
+            )
+            raise
+        assert rc == 0, (
+            f"Remote command exited with non-0 status code\n\n"
+            f"{rc=}\n{stdout=}\n{stderr=}\n\n"
+            f"Firecracker logs:\n{self.log_data}\n"
+            f"Thread backtraces:\n{self.thread_backtraces}"
+        )
+
```
bchalios: Why not just wait for SSH up?

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -540,3 +542,31 @@ def test_vmgenid(guest_kernel_linux_6_1, rootfs, microvm_factory, snapshot_type)
...
+    vm.start()
+    vm.wait_for_up()
+
+    snapshot = vm.snapshot_full()
+
+    restored_vm = microvm_factory.build()
+    restored_vm.spawn()
+    iface2 = dataclasses.replace(base_iface, tap_name="tap2")
+    snapshot.net_ifaces.clear()
+    snapshot.net_ifaces.append(iface2)
```
bchalios, line 566: I don't follow this logic very well. Why do we need to do that? (Not saying it's wrong, maybe it's just me not understanding very well what the snapshot object does exactly.)

CHANGELOG.md
```diff
@@ -111,6 +111,9 @@ and this project adheres to
   support for `-h` help flag to the Jailer. The Jailer will now print the help
   message with either `--help` or `-h`.
 
+- [#4731](https://github.com/firecracker-microvm/firecracker/pull/4731): Added
+  support for modifying the host device name during snapshot restore.
```
bchalios: nit: `s/the host device name/the host TAP device name`

docs/snapshotting/network-for-clones.md
```diff
@@ -142,6 +142,20 @@ Otherwise, packets originating from the guest might be using old Link Layer
...
+specify which network device (based on the name inside the VM, such as "eth0")
+maps to which host tap device (e.g. "vmtap01").
+
+This may require reconfiguration of the networking inside the VM so that it is
+still routable externally.
+
```
bchalios, line 158: Could you please add an example here of how this can be used.

CHANGELOG.md
```diff
@@ -112,7 +112,7 @@ and this project adheres to
   message with either `--help` or `-h`.
 
 - [#4731](https://github.com/firecracker-microvm/firecracker/pull/4731): Added
-  support for modifying the host device name during snapshot restore.
+  support for modifying the host TAP device name during snapshot restore.
```
bchalios, line 115: Oh I see you did that already in a later commit (I am reviewing commit by commit). Could you squash all the commits that update documentation into a single commit?

bchalios (changes requested):
It generally looks good and thanks for the contribution Andrew.

A few comments/questions from me.

Also, I've commented this for the documentation changes, but could you please squash as well the commits for the test changes into a single commit?

src/firecracker/swagger/firecracker.yaml
andrewla: Done

tests/framework/microvm.py
andrewla: Done

tests/framework/microvm.py
andrewla: This was from an earlier version where this function existed; it has been replaced by the ssh_up function and the tests have been fixed to remove this.

CHANGELOG.md
andrewla: Done

CHANGELOG.md
andrewla: Done

src/vmm/src/builder.rs
andrewla: Done

docs/snapshotting/network-for-clones.md
andrewla: Added some sample code and more detail.

bchalios: It would be great if the example include the steps to start a microVM with a certain network config (it could be the one from "Getting started") then take a snapshot and, finally, load this snapshot in a different host with a different TAP device (show the snapshot load command with the override).

src/vmm/src/vmm_config/snapshot.rs
```diff
@@ -60,6 +70,8 @@ pub struct LoadSnapshotParams {
     /// When set to true, the vm is also resumed if the snapshot load
     /// is successful.
     pub resume_vm: bool,
+    /// The network devices to override on load.
+    pub network_overrides: Vec<NetworkOverride>,
```
pb8o, line 74: According to the integration tests failures, `LoadSnapshotConfig` just below also needs to be updated with this.

docs/snapshotting/network-for-clones.md
```diff
@@ -142,6 +142,38 @@ Otherwise, packets originating from the guest might be using old Link Layer
 Address for up to arp cache timeout seconds. After said timeout period,
 connectivity will work both ways even without an explicit flush.
 
+### Renaming host device names
```
pb8o, line 145: The markdown linter [seems to complain](https://buildkite.com/firecracker/firecracker-pr/builds/12273#01945105-f732-4a16-812e-07e0fdf0e5d2/345). You can fix it locally with `./tools/devtool fmt` and check with `./tools/devtool checkstyle`.

pb8o (changes requested):

docs/snapshotting/network-for-clones.md
andrewla: The problem with this in principle is that it does require that the VM reconfigure its networking to match the expectations of the external routing. This complexity is easily managed in the real world where you will have some channel of communication from host to guest (vsocks or swapping block devices or mmds or vmgenid) but hard to capture in a small example.

docs/snapshotting/network-for-clones.md
andrewla: Fixed

src/vmm/src/vmm_config/snapshot.rs
andrewla: Any ideas on what the problem is here? The network_overrides field is present in both structs, and for the life of me I can't see where this error message is originating -- the field is in the yaml, it's in both the config and params struct. I don't have an easy callstack for the firecracker side of the failure, but that's where I'll start looking next, but any insights here would be appreciated

pb8o: Hi @andrewla I think what is happening is that the test that fails: `test_check_vulnerability_files_ab` is an A/B test, and is running the new tests with the Firecracker from current `main`. A workaround would be to not pass `network_overrides` if it's empty. I will write a comment where I think it should be done.

tests/framework/microvm.py
```diff
@@ -996,11 +997,19 @@ def restore_from_snapshot(
...
+                {"iface_id": k, "host_dev_name": v}
+                for k, v in rename_interfaces.items()
+            ]
+
         self.api.snapshot_load.put(
             mem_backend=mem_backend,
             snapshot_path=str(jailed_vmstate),
             enable_diff_snapshots=snapshot.is_diff,
             resume_vm=resume,
+            network_overrides=iface_overrides,
```
pb8o, line 1012: To avoid sending `network_overrides` if it's empty do something like this.

```
kwargs = {}
if network_overrides:
    kwargs["network_overrides"] = network_overrides

self.api.snapshot_load.put(
    ...
    **kwargs,
)
```

src/vmm/src/vmm_config/snapshot.rs
andrewla: This does not match what I'm seeing in CI, I see a number of tests failing with the messages

```
E   RuntimeError: ('An error occurred when deserializing the json body of a request: unknown field `network_overrides`, expected one of `snapshot_path`, `mem_file_path`, `mem_backend`, `enable_diff_snapshots`, `resume_vm` at line 1 column 173.', {'fault_message': 'An error occurred when deserializing the json body of a request: unknown field `network_overrides`, expected one of `snapshot_path`, `mem_file_path`, `mem_backend`, `enable_diff_snapshots`, `resume_vm` at line 1 column 173.'}, <Response [400]>)
```

Mostly these are in `integration_tests/security/test_vulnerabilities.py`, but not all tests in that file fail consistently. The `_ab` tests fail but not consistently.

I am unable to reproduce any of this locally -- the tests pass when I run them specifically.

pb8o: Try to reproduce it like this:

```
BUILDKITE_PULL_REQUEST=true BUILDKITE_PULL_REQUEST_BASE_BRANCH=main ./tools/devtool test -- -n8 --dist worksteal integration_tests/security/test_vulnerabilities.py::test_check_vulnerability_files_ab
```

andrewla: I'll try applying the patch to verify. Trying the command you give has the tests fail with a different error:

```
framework/microvm.py:206: in __init__
    assert fc_binary_path.exists()
E   AssertionError
```

I don't know exactly where I've gone wrong that the firecracker binary is not in the right place in the test framework. I'll try from scratch, and in the meantime I'll submit the suggested patch to see if it works in CI

andrewla: A clean build does not fix this; I get this error when running from scratch. All of the old firecracker binaries are present under build/img/x86_64/firecracker; I have no idea why this doesn't work, or why even in CI some of the ab tests work.

The other failing test is the spectre/meltdown test which also appears to have an `ab` mode based on whether it is a PR or not, so likely failing for the same reason.

Is there anything I have to do to get buildkite to run the CI? After pushing changes it goes into a "blocked" mode; a 24 hr turnaround on CI is useless when I can repro the failure locally.

roypat: Hi, this is because `devtool test` apparently only tried to build the "B" binary, but never the "A" binary. In CI, it works because there's a separate step that builds the binary so that they can be shared across all runners. Can you try running `./tools/devtool build --rev main` and then rerun the test command pablo posted? That should create the firecracker binaries compiled from main in the locations where the test script looks for them.

andrewla: Now it can find the binaries (under build/main/..., I see now) but it fails elsewhere still. Error looks like

```
framework/microvm.py:678: in _wait_create
  os.stat(self.jailer.api_socket_path())
E   FileNotFoundError: [Errno 2] No such file or directory: '/srv/jailer/firecracker/594eec23-7064-4306-9acd-6a6e137c3d30/root/run/firecracker.socket'
```

roypat: Ah, it needs to be `./tools/devtool build --rev main --release` (although it should also work without `--release`, wonder why that breaks :thinking:), sorry!

I've also opened a PR to make devtool do that automatically + add some documentation: https://github.com/firecracker-microvm/firecracker/pull/4998

docs/snapshotting/network-for-clones.md
bchalios: That is fine. I think we can still add something useful here. The guest configuration you have already is fine. I would add something like that:

```
For example, if we have a network interface named `eth0` in the snapshotted microVM. We can override it during snapshot resume, like this:


curl --unix-socket /tmp/firecracker.socket -i \
    -X PUT 'http://localhost/snapshot/load' \
    -H  'Accept: application/json' \
    -H  'Content-Type: application/json' \
    -d '{
            "snapshot_path": "./snapshot_file",
            "mem_backend": {
                "backend_path": "./mem_file",
                "backend_type": "File"
            },
            "enable_diff_snapshots": true,
            "resume_vm": false,
            "network_overrides": [
                 {
                     iface_id: "eth0",
                     host_dev_name": "vmtap01"
                 }
            ]
    }'
```

after this bit:

> In this case you can use the `network_overrides` parameter to snapshot restore
to specify which network device (based on the name inside the VM, such as
"eth0") maps to which host tap device (e.g. "vmtap01").

src/vmm/src/persist.rs
```diff
@@ -415,7 +415,21 @@ pub fn restore_from_snapshot(
...
+        if let Some(device) = net_devices
+            .iter_mut()
+            .find(|x| x.device_state.id == entry.iface_id)
+        {
+            device
+                .device_state
+                .tap_if_name
+                .clone_from(&entry.host_dev_name);
+        } else {
+            return Err(SnapshotStateFromFileError::UnknownNetworkDevice.into());
```
bchalios, line 430: These lines are uncovered from unit tests? Can we add one (or extend an existing) to try to override a non-existing interface?

andrewla: I'm investigating this (once I get the tests to pass) because this is really the core of the change, and if the tests are actually running then this code should be exercised. I'm concerned that my test is accidentally a noop but I need to validate this.

andrewla:
It turns out that the test for renaming devices was failing when run with other tests that used network devices. After some experimentation, it seems that we are not cleaning up network devices from other tests, and modifying a network device results in an incompatible network configuration, rendering the VM unreachable.

For now I've patched this by having the new test use an unallocated network device, but I'm not sure if we're comfortable with this or if we want to try to figure out why the test passes when run alone but not when run in tandem with other tests.

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -606,10 +606,11 @@ def test_snapshot_rename_interface(uvm_nano, microvm_factory):
     Test that we can restore a snapshot and point its interface to a
     different host interface.
     """
-    base_iface = net_tools.NetIfaceConfig.with_id(0)
+    # We start at an unused tap index to avoid conflicts with other tests.
+    base_iface = net_tools.NetIfaceConfig.with_id(7)
 
     vm = uvm_nano
-    iface1 = dataclasses.replace(base_iface, tap_name="tap1")
+    iface1 = dataclasses.replace(base_iface, tap_name="tap8")
```
pb8o, line 613: oh this may be because since #4966 we re-use network namespaces and devices during the integration tests.

src/vmm/src/persist.rs
andrewla: Now that the tests are passing this correctly shows up as covered.

docs/snapshotting/network-for-clones.md
andrewla: Done

tests/integration_tests/functional/test_snapshot_basic.py
andrewla: Interesting -- might be worth making the tap_name non-optional, or disabling this behavior if it differs, to avoid this issue cropping up again

andrewla: I tried some debugging but couldn't get anywhere -- it doesn't appear when you run the test in isolation, even the earlier repro only worked when run in tandem with other tests.

I upped the id of the interface; hopefully that will let this squeeze by while I try to root cause it

pb8o: So I was looking at this today in c5n. I _think_ it's because the network namespace re-used conflicts with a previous TAP device.

So basically re-using network namespaces is just not worth it in this test and then we can simplify it to this:

```
def test_snapshot_rename_interface(uvm_nano, microvm_factory):
    """
    Test that we can restore a snapshot and point its interface to a
    different host interface.
    """

    vm = uvm_nano
    base_iface = vm.add_net_iface()
    vm.start()
    snapshot = vm.snapshot_full()

    # We don't re-use the network namespace as it may conflict with
    # previous/future devices
    restored_vm = microvm_factory.build(netns=NetNs(str(uuid.uuid4())))
    # Override the tap name, but keep the same IP configuration
    iface_override = dataclasses.replace(base_iface, tap_name="tap_override")

    restored_vm.spawn()
    snapshot.net_ifaces.clear()
    snapshot.net_ifaces.append(iface_override)
    restored_vm.restore_from_snapshot(
        snapshot,
        rename_interfaces={iface_override.dev_name: iface_override.tap_name},
        resume=True
    )
```

(I don't think it's worth it to do the negative test actually so I removed it here)

pb8o: Hi @andrewla, do you plan to take this again in the near future. If not, I may be able to do the rebases in the coming weeks. Thanks!

bchalios: I can take over this as well @pb8o if you want.

bchalios:
I have applied the changes suggested by @pb8o. Also, I squashed all test commits to a single commit and I moved some code around in the appropriate commits.

tests/integration_tests/functional/test_snapshot_basic.py
andrewla: Reviewed and pushed the changes, sorry for the delay. Thanks for looking at this

andrewla: Whoops, missing context; will revert my changes. Thanks @bchalios for digging into this

andrewla: I fixed a typo and an unnecessary line (base_iface is assigned twice) and rebased against main

bchalios (dismissed):

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -570,3 +573,30 @@ def test_physical_counter_reset_aarch64(uvm_nano):
...
+
+    restored_vm.spawn()
+    snapshot.net_ifaces.clear()
+    snapshot.net_ifaces.append(iface_override)
+    restored_vm.restore_from_snapshot(
+        snapshot,
+        rename_interfaces={iface_override.dev_name: iface_override.tap_name},
+        resume=True,
+    )
+    restored_vm.wait_for_ssh_up()
```
roypat, line 602: nit: not needed, as `restore_from_snapshot` already does this iirc

docs/snapshotting/network-for-clones.md
```diff
@@ -142,6 +142,65 @@ Otherwise, packets originating from the guest might be using old Link Layer
...
+
+In some environments where the jailer is not being used, restoring a snapshot
+may be tricky because the tap device on the host will not be the same as the tap
+device that the original VM was mapped to when it was snapshotted, as when the
+tap device come from a pool of such devices.
```
roypat, line 150: I don't quite understand the last part of this sentence. Should it be

```suggestion
In some environments where the jailer is not being used, restoring a snapshot
may be tricky because the tap device on the host will not be the same as the tap
device that the original VM was mapped to when it was snapshotted, for example when the
tap device comes from a pool of such devices.
```

?

docs/snapshotting/network-for-clones.md
```diff
@@ -142,6 +142,65 @@ Otherwise, packets originating from the guest might be using old Link Layer
...
+
+In this case you can use the `network_overrides` parameter to snapshot restore
```
roypat, line 152:
```suggestion
In this case you can use the `network_overrides` parameter during snapshot restore
```

of

```suggestion
In this case you can use the `network_overrides` parameter of the snapshot restore API
```

docs/snapshotting/network-for-clones.md
```diff
@@ -142,6 +142,65 @@ Otherwise, packets originating from the guest might be using old Link Layer
...
+to specify which guest network device maps to which host tap device.
+
+For example, if we have a network interface named `eth0` in the snapshotted
+microVM. We can override it to point to the host device `vmtap01` during
+snapshot resume, like this:
```
roypat, line 157:
```suggestion
For example, if we have a network interface named `eth0` in the snapshotted
microVM, we can override it to point to the host device `vmtap01` during
snapshot resume like this:
```

docs/snapshotting/network-for-clones.md
```diff
@@ -142,6 +142,65 @@ Otherwise, packets originating from the guest might be using old Link Layer
...
+    -H  'Accept: application/json' \
+    -H  'Content-Type: application/json' \
+    -d '{
+            "snapshot_path": "./snapshot_file",
+            "mem_backend": {
+                "backend_path": "./mem_file",
+                "backend_type": "File"
+            },
+            "enable_diff_snapshots": true,
+            "resume_vm": false,
```
roypat, line 171: nit: maybe this example should focus only on the mandatory arguments + `network_overrides`. E.g. drop the `enable_diff_snapshots` and `resume_vm` and just leave them at Firecracker's defaults, because they're not important for this example

docs/snapshotting/network-for-clones.md
```diff
@@ -142,6 +142,65 @@ Otherwise, packets originating from the guest might be using old Link Layer
...
+                     iface_id: "eth0",
+                     host_dev_name": "vmtap01"
+                 }
+            ]
+    }'
+```
+
+This may require reconfiguration of the networking inside the VM so that it is
+still routable externally. The
+[network setup documentation](../network-setup.md) in the "In The Guest" section
```
roypat, line 183: nit: could be a direct link to the subsection

tests/integration_tests/functional/test_snapshot_basic.py
bchalios: done

docs/snapshotting/network-for-clones.md
bchalios: I think so. I've amended it with your suggested changes.

docs/snapshotting/network-for-clones.md
bchalios: done

docs/snapshotting/network-for-clones.md
bchalios: done

docs/snapshotting/network-for-clones.md
bchalios: done

docs/snapshotting/network-for-clones.md
```diff
@@ -142,6 +142,63 @@ Otherwise, packets originating from the guest might be using old Link Layer
...
+    -d '{
+            "snapshot_path": "./snapshot_file",
+            "mem_backend": {
+                "backend_path": "./mem_file",
+                "backend_type": "File"
+            },
+            "network_overrides": [
+                 {
+                     iface_id: "eth0",
+                     host_dev_name": "vmtap01"
```
roypat, line 173: Sorry, missed this earlier, but this needs to be
```suggestion
                     "iface_id": "eth0",
                     "host_dev_name": "vmtap01"
```

bchalios: :facepalm:

bchalios: fixed

roypat (approved):

pb8o (approved):

PR #4738, opened 2024-08-20
test(jailer): do not kill screen with new_pid_ns

kalyazin:
## Changes

This change lets the screen terminate on its own without attempting to kill it.

## Reason

In our setup, if Firecracker is run by the screen with --new-pid-ns passed to the jailer, the Firecracker process detaches from the screen causing the screen process to exit as well.
We attempt to kill the screen process by pid, which causes a race condition.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- ~~[ ] Any required documentation changes (code and docs) are included in this
  PR.~~
- [x] Commits meet


bchalios (approved):

tests/framework/microvm.py
```diff
@@ -632,6 +632,12 @@ def spawn(
             )
             self._screen_pid = screen_pid
 
+        # If `--new-pid-ns` is used, the Firecracker process will detach from
+        # the screen and the screen process will exit. We do not want to
+        # attempt to kill it in that case to avoid a race condition.
```
pb8o, line 637: In that case we just shouldn't use screen. Which test is doing this?

We may also want to raise `if screen and new-pid-ns`

pb8o: In the microvm class. That combination of options cannot be handled by screen, so we should raise an exception when we detect it. I just don't think it's a valid use case.

kalyazin: My point is we still use screen to start FC in this combination in `Microvm::spawn`, even though the screen itself exits soon after, so screen is still running for some time and we can't reliable assert its absence.
In order to get rid of screen usage for this use case we need to create a new helper that would allow to run a FC command that does not complete immediately (unlike in screen-detached or daemonised cases), which I'm not sure is warranted just by this case.
What kind of check do you mean? In the `Microvm::kill`?

pb8o (approved):

tests/framework/microvm.py
pb8o: Synced offline. Doing this with the current codebase would be very complicated, so we will defer this until we extract the screen part as a separate "Jailer"

PR #4742, opened 2024-08-21
refactor(virtio-net): avoid copy on rx

ihciah:
Improve virtio-net performance by directly read into desc chain; introduce Readiness management to avoid redundant readv and make code more readable.

## Changes

1. Change virtio-net from copy-read to direct-read: Previously, the net device implementation would read data from the TAP into a buffer in VMM, and then copy it into the descriptor chain. The new implementation converts the descriptor chain into iovec first and then reads the data directly into it.
2. Optimize the reading performance of the descriptor chain: Reduce stack copying and the overhead of guest memory reads.
3. Refactor the code related to reading data in the net device: Add management of readiness and centralize the main logic in the process_rx function, making the code simpler and clearer.

## Reason

1. Performance
In the current VMM simulated net device read implementation, data is copied to a VMM-owned buffer before being copied to the descriptor chain. The new method converts the descriptor chain directly into an iovec and then uses tap readv for direct input.

Each implementation has its advantages and disadvantages. The current method avoids traversing the entire descriptor chain for iovec conversion when handling small packets but incurs copy overhead for larger packets.

Previously, https://github.com/firecracker-microvm/firecracker/pull/2958 switched from copy+write to writev, resulting in performance improvement for the write side. The author of that PR mentioned that modifying it to readv was tried but faced iovec conversion overhead.

However, I believe this issue is hard to avoid completely but isn't unsolvable. We can try to minimize the read overhead of the descriptor chain to reduce conversion costs:
   1. In https://github.com/firecracker-microvm/firecracker/pull/4723, I removed an unnecessary check which reduced certain memory read costs in my benchmarks.
   2. I introduced a `load_next_descriptor` method that reduces the stack copying overhead of the DescriptorChain structure.
   3. I added a `MemBytesExt`, which after passing certain safety checks (inevitably passed by a non-malicious guest, if not pass, it fallbacks to read_obj), uses `ptr::read_volatile` to replace the more complex `read_obj`, achieving significant performance improvements in benchmarks.
   4. For future optimization, focusing on GuestMemory look-up could be effective. Since memory mappings are static, if the entire virtio queue is within a single contiguous mapping region, we could save this specific GuestMemoryRegion instead of the entire GuestMemory upon device activation. This would substantially reduce the lookup overhead for each memory access. Although there might typically be only two regions, the performance data show significant overhead that should not be overlooked. I have attempted to make changes to facilitate this optimization in the vm-memory repository: https://github.com/rust-vmm/vm-memory/pull/293.

Performance comparison data under these optimizations(the unit is Gbps):
```markdown
| Bench  | Old   | New   | Change   |
|--------|-------|-------|----------|
| F      | 37.74 | 42.4  | +12.35%  |
| R      | 40.8  | 40.4  | -0.98%   |
| F-96   | 37.6  | 42.8  | +13.83%  |
| F-128  | 37.81 | 42.5  | +12.40%  |
| F-256  | 37.43 | 42.8  | +14.35%  |
| F-1460 | 37.89 | 42.4  | +11.90%  |
| R-96   | 38.42 | 39.82 | +3.64%   |
| R-128  | 38.22 | 38.29 | +0.18%   |
| R-256  | 38.55 | 38.3  | -0.65%   |
| R-1460 | 40.8  | 40.3  | -1.23%   |
```

Attention, because this is a home-use CPU and Turbo Boost has not been disabled, the data may be somewhat inaccurate. If needed, I can upload some perf data.

Testing environment:
- CPU: AMD Ryzen 7 7840HS w/ Radeon Graphics
- VM: 4C2G, Host: 16C32G
- Affinity: 4 vCPUs and 1 main thread bound to 5 different cores; 4 host iperf processes bound to 4 different cores.
- iperf3: 4 host clients, 4 guest servers.
    - F: `iperf3 -A 1 --client 172.16.0.2 -T 5201 -p 5201 --omit 5 --time 30 -l 512K`
    - R: `iperf3 -A 1 --client 172.16.0.2 -T 5201 -p 5201 --omit 5 --time 30 -l 512K -R`
    - F-num: `iperf3 -A 1 --client 172.16.0.2 -T 5201 -p 5201 --omit 5 --time 30 -l 512K -M "$1"`
    - R-num: `iperf3 -A 1 --client 172.16.0.2 -T 5201 -p 5201 --omit 5 --time 30 -l 512K -R -M "$1"`

The new implementation shows significant performance improvements when reading large packets, with over a 10% maximum throughput increase in my benchmarks; however, it has a minor performance degradation with small packets.

2. Code Refactoring
The current implementation on the read link is quite complex as below:
image
In the new implementation, the main logic is consolidated in `process_rx` and `process_tx`, with states managed using readiness tracking. Recording states with readiness also avoids potential overhead from queue and tap reads, making the code easier to understand.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/src/vmm/src/devices/virtio/net/device.rs
+++ b/src/vmm/src/devices/virtio/net/device.rs
@@ -5,13 +5,11 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 
-#[cfg(not(test))]
-use std::io::Read;
 use std::mem;
 use std::net::Ipv4Addr;
+use std::num::NonZeroUsize;
 use std::sync::{Arc, Mutex};
 
-use libc::EAGAIN;
 use log::{error, warn};
 use utils::eventfd::EventFd;
 use utils::net::mac::MacAddr;
@@ -26,7 +24,7 @@ use crate::devices::virtio::gen::virtio_net::{
     VIRTIO_NET_F_HOST_TSO6, VIRTIO_NET_F_HOST_UFO, VIRTIO_NET_F_MAC,
 };
 use crate::devices::virtio::gen::virtio_ring::VIRTIO_RING_F_EVENT_IDX;
-use crate::devices::virtio::iovec::IoVecBuffer;
+use crate::devices::virtio::iovec::{IoVecBuffer, IoVecBufferMut};
 use crate::devices::virtio::net::metrics::{NetDeviceMetrics, NetMetricsPerDevice};
 use crate::devices::virtio::net::tap::Tap;
 use crate::devices::virtio::net::{
@@ -47,12 +45,8 @@ const FRAME_HEADER_MAX_LEN: usize = PAYLOAD_OFFSET + ETH_IPV4_FRAME_LEN;
 
 #[derive(Debug, thiserror::Error, displaydoc::Display)]
 enum FrontendError {
-    /// Add user.
-    AddUsed,
     /// Descriptor chain too mall.
     DescriptorChainTooSmall,
-    /// Empty queue.
-    EmptyQueue,
     /// Guest memory error: {0}
     GuestMemory(GuestMemoryError),
     /// Read only descriptor.
@@ -103,6 +97,60 @@ pub struct ConfigSpace {
 // SAFETY: `ConfigSpace` contains only PODs in `repr(C)` or `repr(transparent)`, without padding.
 unsafe impl ByteValued for ConfigSpace {}
 
+// Ideally, we should put all readiness inside the Readiness struct.
+// But let rate limiter maintain its own readiness is simpler.
+#[derive(Debug)]
+struct Readiness(u8);
+
+impl Default for Readiness {
+    fn default() -> Self {
+        Readiness(Self::FULL)
+    }
+}
+
+macro_rules! define_mark_clear {
+    ($mark:ident, $clear:ident, $get:ident, $mask:ident) => {
+        #[allow(unused)]
+        fn $mark(&mut self) {
+            self.0 |= Self::$mask;
+        }
+
+        #[allow(unused)]
+        fn $clear(&mut self) {
+            self.0 &= !Self::$mask;
+        }
+
+        #[allow(unused)]
+        fn $get(&self) -> bool {
+            self.0 & Self::$mask != 0
+        }
+    };
+}
+
+impl Readiness {
+    const TAP_READY: u8 = 1 << 0;
+    const RX_READY: u8 = 1 << 1;
+    const TX_READY: u8 = 1 << 2;
+    const MMDS_READY: u8 = 1 << 3;
+
+    const FULL: u8 = Self::TAP_READY | Self::RX_READY | Self::TX_READY | Self::MMDS_READY;
+    const RX_MMDS_CHECK: u8 = Self::MMDS_READY | Self::RX_READY;
+    const RX_TAP_CHECK: u8 = Self::TAP_READY | Self::RX_READY;
+
+    define_mark_clear!(mark_tap_ready, clear_tap_ready, tap_ready, TAP_READY);
+    define_mark_clear!(mark_rx_ready, clear_rx_ready, rx_ready, RX_READY);
+    define_mark_clear!(mark_tx_ready, clear_tx_ready, tx_ready, TX_READY);
+    define_mark_clear!(mark_mmds_ready, clear_mmds_ready, mmds_ready, MMDS_READY);
+
+    fn rx_mmds_pre_check(&self) -> bool {
+        Self::RX_MMDS_CHECK & self.0 == Self::RX_MMDS_CHECK
+    }
+
+    fn rx_tap_pre_check(&self) -> bool {
+        Self::RX_TAP_CHECK & self.0 == Self::RX_TAP_CHECK
+    }
+}
+
 /// VirtIO network device.
 ///
 /// It emulates a network device able to exchange L2 frames between the guest
@@ -122,12 +170,10 @@ pub struct Net {
 
     pub(crate) rx_rate_limiter: RateLimiter,
     pub(crate) tx_rate_limiter: RateLimiter,
+    readiness: Readiness,
 
-    pub(crate) rx_deferred_frame: bool,
-
-    rx_bytes_read: usize,
-    rx_frame_buf: [u8; MAX_BUFFER_SIZE],
-
+    rx_mmds_read: Option<NonZeroUsize>,
+    rx_mmds_buf: [u8; MAX_BUFFER_SIZE],
     tx_frame_headers: [u8; frame_hdr_len()],
 
     pub(crate) irq_trigger: IrqTrigger,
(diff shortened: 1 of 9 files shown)


ShadowCurse (commented):
Hi @ihciah, thx for the PR. Really nice description and from the first glance changes are good as well. But can you please split commit into multiple? It would be easier for people to review it.

ihciah:
> Hi @ihciah, thx for the PR. Really nice description and from the first glance changes are good as well. But can you please split commit into multiple? It would be easier for people to review it.

Thank you. It is splitted now, and also more testing are added.

ihciah:
@ShadowCurse Hi! I've splitted the commits, could you re-review this PR? Thanks.

bchalios:
Hi @ihciah and thanks a lot for the PR. It looks very promising. Strangely enough, I was working as well in a similar approach and have a draft branch here: https://github.com/bchalios/firecracker/tree/pre-process-rx-bufs.

I went through a different path for eliminating the overhead of processing `DescriptorChain` objects which is I am pre-processing them outside the hot path (when the TAP device is ready). I am processing the buffers when the guest sends us a notification that it added RX buffers in the queue. The idea is that once we have TAP ready we will already have an IoVecBuffer read for handling the RX.

We are currently in the process of evaluating the performance of both PRs. In our setup we do see performance improvements, but some test cases regress and we want to understand why and if we could fix them.

I will keep you posted with my investigation.

ihciah:
@bchalios
> I went through a different path for eliminating the overhead of processing DescriptorChain objects which is I am pre-processing them outside the hot path (when the TAP device is ready).

I think this may good for latency but cannot eliminate the cost of desc chain converting. In my testing, I emit metrics about the cvt op and the chain length, it is really surprising that all read desc chains' length are 19(every one is exact 19), which is just the same as I speculated here https://github.com/cloud-hypervisor/cloud-hypervisor/pull/6636#issuecomment-2276197274.

19 is big enough for a performance regression. What we can do is to make the loop cheaper. The cost which can be avoided here is the checking and copying.

1. I compared the diff before and after the https://github.com/firecracker-microvm/firecracker/pull/4723 , it shows very positive optimization.
2. `read_obj` is heavy. With the quick `load_obj` in this PR, it reduces a lot memcpy.
    1. before: https://share.firefox.dev/3MmTQru
    2. after: https://share.firefox.dev/3MovNrY
    3. Note: select firecracker process only.
4. Update DescriptorChain inplace also benefits in a small amount. After all, DescriptorChain itself is 40byte, and 40*19 is a unnegligible stack copy size.

ShadowCurse:
>Update DescriptorChain inplace also benefits in a small amount. After all, DescriptorChain itself is 40byte, and 40*19 is a unnegligible stack copy size.

Did you validate this? I would assume because the usage pattern is:
```rust
let mut next_descriptor = Some(head);
while let Some(desc) = next_descriptor {
  ...
  next_descriptor = desc.next_descriptor();
}
```
the `next_descriptor = desc.next_descriptor();`  would just reuse already existing `next_descriptor`.

Also regarding `read_obj` optimization: I am working on an overall improvement for the `Queue` here: #4748, and would prefer to go that direction (e.g. address all unnecessary calls to `GuestMemoryMmap` at once) rather than optimizing single `read_obj` call.

bchalios:
> I think this may good for latency but cannot eliminate the cost of desc chain converting. In my testing, I emit metrics about the cvt op and the chain length, it is really surprising that all read desc chains' length are 19(every one is exact 19), which is just the same as I speculated here https://github.com/cloud-hypervisor/cloud-hypervisor/pull/6636#issuecomment-2276197274.

I know and I agree. The two optimizations (reducing the cost of parsing and moving the parsing out of the hot path) are largely orthogonal. I suggest we co-ordinate as well with Egor in order to understand which optimizations make sense and bundle them together. I will be working to measure performance of the various pieces and post updates here.

ShadowCurse:
Hi @ihciah , we have looked through your PR and we appreciate the effort you put in, but the scope of your changes is bigger than it is needed for a `readv` implementation. As @bchalios mentioned, we were working on similar changes recently and we think it is better for us to take over the implementation of `readv` in this case.
The change we are happy to accept though is the `load_obj` change. If you think it is indeed faster than the default `read_obj`, can you please open a new PR for it?

ihciah:
> The change we are happy to accept though is the load_obj change. If you think it is indeed faster than the default read_obj, can you please open a new PR for it?

Appreciated it. I'll work on it recently.

> the scope of your changes is bigger than it is needed for a readv implementation. As @bchalios mentioned, we were working on similar changes recently and we think it is better for us to take over the implementation of readv in this case.

This PR indeed contains a lot things to make the performance better to avoid performance degradation on certain cases. If the PR you think can be splitted, could you tell me your expected way? If you think there's anything needed to change, please comment it and I can adjust it as your preferred way. I'm willing to work with your team to get it done.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -312,25 +363,381 @@ impl Net {
...
+                let head_index = head.index;
+
+                macro_rules! push_desc {
+                    ($n:expr) => {
+                        rx_queue
+                            .add_used(mem, head_index, $n)
+                            .map_err(DeviceError::QueueError)?;
+                        notify_guest = true;
+                    };
+                }
```
ShadowCurse, line 531: I don't think we really need a macros for this. Also if you address comment above, I think this will go away naturally.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -312,25 +363,381 @@ impl Net {
...
+            }
+
+            let frame_consumed_by_mmds = Self::write_to_mmds_or_tap(
+                self.mmds_ns.as_mut(),
+                &mut self.tx_rate_limiter,
+                &mut self.tx_frame_headers,
+                &self.tx_buffer,
+                &mut self.tap,
+                self.guest_mac,
+                &self.metrics,
```
ShadowCurse, line 717: Would be nice if you could make this function work on `self` (so `self.write_to_mmds_or_tap`). It only takes `self.*` arguments anyway.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -312,25 +363,381 @@ impl Net {
...
+            let head_index = head.index;
+
+            macro_rules! push_desc {
+                () => {
+                    tx_queue
+                        .add_used(mem, head_index, 0)
+                        .map_err(DeviceError::QueueError)?;
+                    used_any = true;
+                };
+            }
```
ShadowCurse, line 681: Here we can also skip macros and repeat these 4 line a couple of time.

ShadowCurse (commented):
Ok, let's see what we can do about this. As I previously mentioned, the `feat: add load_obj which is a faster read_obj` commit can be move into it's own PR. Also as I mentioned, @bchalios  will take care of `readv`, so `improve: add load_next_descriptor`, `feat: impl basic iovec ability for IoVecBufferMut`, `feat: support read_iovec for tap device`,  and small portion of `refactor(virtio-net): avoid copy on rx` commits are not needed in this PR. So only the remaining part of `refactor(virtio-net): avoid copy on rx` remains. I will leave some comments on it, but in general it would be nice if you could somehow split it into smaller commits (like `add readiness`, `update rx path`, ...). Otherwise it basically touches the whole `device.rs`.

ShadowCurse:
Hi @ihciah, it has bee some time and since then we updated virtio-net with readv (and MRG_RXBUF) support. Because of those changes, this PR seems to be redundant now, so I will close it. Feel free to open follow up PR if you think there is space for additional improvements.

PR #4743, opened 2024-08-22
Jailer: added the `-h` flag as a shorthand for `--help`

seafoodfry:
## Changes

Jailer: added the `-h` flag as a shorthand for `--help`.

## Reason

This is a common convention - thought it would be useful.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

---

# Additional Details

Tried to replicate the workings of the ArgParser by starting from
https://github.com/firecracker-microvm/firecracker/blob/e67b2bedbe01a958cc3140f0e951bd5ffa98ff0a/src/jailer/src/main.rs#L318-L320

Which leads to 
https://github.com/firecracker-microvm/firecracker/blob/e67b2bedbe01a958cc3140f0e951bd5ffa98ff0a/src/utils/src/arg_parser.rs#L320-L323

From there I tried to replicate the logic with this small script to see if this patch would work:

```rust
use std::env;                                                           
                                                                        
const ARG_SEPARATOR: &str = "--";                                       
const HELP_ARG: &str = "--help";                                        
const SHORT_HELP_ARG: &str = "-h";                                      
                                                                        
fn split_args(args: &[String]) -> (&[String], &[String]) {              
    if let Some(index) = args.iter().position(|arg| arg == ARG_SEPARATOR) {
        return (&args[..index], &args[index + 1..]);                    
    }                                                                   
                                                                        
    (args, &[])                                                         
}                                                                       
                                                                        
fn parse(args: &[String]) {                                             
    // Skipping the first element of `args` as it is the name of the binary.
    let (args, _) = split_args(&args[1..]);                             
                                                                        
    if args.contains(&HELP_ARG.to_string()) || args.contains(&SHORT_HELP_ARG.to_string()) {
        println!("found help args: {args:?}");                          
    }                                                                   
}                                                                       
                                                                        
fn main() {                                                             
    let args: Vec<String> = env::args().collect();                      
                                                                        
    parse(&args)                                                        
}
```

I tried it out and saw the following:

```
➜  ~/src/temp/args_help git:(master) ✗ ./target/debug/args_help --help 
found help args: ["--help"]
➜  ~/src/temp/args_help git:(master) ✗ ./target/debug/args_help -h    
found help args: ["-h"]
➜  ~/src/temp/args_help git:(master) ✗ ./target/debug/args_help --helpo
➜  ~/src/temp/args_help git:(master) ✗ ./target/debug/args_help --h    
➜  ~/src/temp/args_help git:(master) ✗ ./target/debug/args_help -hello
```
So it seems like this "should" work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ and this project adheres to
   [#4741](https://github.com/firecracker-microvm/firecracker/pull/4741),
   [#4746](https://github.com/firecracker-microvm/firecracker/pull/4746): Added
   official support for 6.1 microVM guest kernels.
+- [#4743](https://github.com/firecracker-microvm/firecracker/pull/4743): Added
+  support for `-h` help flag to the Jailer. The Jailer will now print the help
+  message with either `--help` or `-h`.
 
 ### Changed
 
--- a/src/utils/src/arg_parser.rs
+++ b/src/utils/src/arg_parser.rs
@@ -9,6 +9,7 @@ pub type Result<T> = result::Result<T, UtilsArgParserError>;
 const ARG_PREFIX: &str = "--";
 const ARG_SEPARATOR: &str = "--";
 const HELP_ARG: &str = "--help";
+const SHORT_HELP_ARG: &str = "-h";
 const VERSION_ARG: &str = "--version";
 
 /// Errors associated with parsing and validating arguments.
@@ -330,10 +331,10 @@ impl<'a> Arguments<'a> {
         let (args, extra_args) = Arguments::split_args(&args[1..]);
         self.extra_args = extra_args.to_vec();
 
-        // If `--help` is provided as a parameter, we artificially skip the parsing of other
+        // If `--help` or `-h`is provided as a parameter, we artificially skip the parsing of other
         // command line arguments by adding just the help argument to the parsed list and
         // returning.
-        if args.contains(&HELP_ARG.to_string()) {
+        if args.contains(&HELP_ARG.to_string()) || args.contains(&SHORT_HELP_ARG.to_string()) {
             let mut help_arg = Argument::new("help").help("Show the help message.");
             help_arg.user_value = Some(Value::Flag);
             self.insert_arg(help_arg);
@@ -643,6 +644,16 @@ mod tests {
         let mut arguments = arg_parser.arguments().clone();
 
         let args = vec!["binary-name", "--exec-file", "foo", "--help"]
+            .into_iter()
+            .map(String::from)
+            .collect::<Vec<String>>();
+
+        arguments.parse(&args).unwrap();
+        assert!(arguments.args.contains_key("help"));
+
+        arguments = arg_parser.arguments().clone();
+
+        let args = vec!["binary-name", "--exec-file", "foo", "-h"]
             .into_iter()
             .map(String::from)
             .collect::<Vec<String>>();


roypat (dismissed):
Hi @seafoodfry, 
thanks for the suggestions :) I think if we want to go with this, we should mention it in CHANGELOG.md. More importantly though is figuring out how to tighten up the parsing to avoid false-positives.

src/utils/src/arg_parser.rs
```diff
@@ -330,10 +331,10 @@ impl<'a> Arguments<'a> {
         let (args, extra_args) = Arguments::split_args(&args[1..]);
         self.extra_args = extra_args.to_vec();
 
-        // If `--help` is provided as a parameter, we artificially skip the parsing of other
+        // If `--help` or `-h`is provided as a parameter, we artificially skip the parsing of other
         // command line arguments by adding just the help argument to the parsed list and
         // returning.
-        if args.contains(&HELP_ARG.to_string()) {
+        if args.contains(&HELP_ARG.to_string() || &SHORT_HELP_ARG.to_string()) {
```
roypat, line 337: I think what you want to express here is

```suggestion
        if args.contains(&HELP_ARG.to_string()) || args.contains(&SHORT_HELP_ARG.to_string()) {
```

but that wouldn't really work because then even something like `./jailer --hello` would end up printing the help text :/

seafoodfry: excuse the very dumb mistake 😅

seafoodfry: @roypat thank you for the PR review!
Based on your comment here I attempted to do some more due dilligence and tried to replicate the logic of the ArgParser to see if this would work as intended.
I added the program I used and its output in the PR description. ptal.

I also added this to the changelog.

roypat: ah, my bad, I was misreading the code. `args` is a vec of strings, not a string itself. Yes, then this should work! 

Could you add a unit test as well? :o

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ and this project adheres to
   [#4746](https://github.com/firecracker-microvm/firecracker/pull/4746): Added
   official support for 6.1 microVM guest kernels.
 
+- [#4743](https://github.com/firecracker-microvm/firecracker/pull/4743): Added
+  support for `-h` help flag to the Jailer. The Jailer will now print the help
+  message with either `--help` or `-h`.
+
```
roypat, line 27:
```suggestion
- [#4743](https://github.com/firecracker-microvm/firecracker/pull/4743): Added
  support for `-h` help flag to the Jailer. The Jailer will now print the help
  message with either `--help` or `-h`.

```

roypat (commented):
Looks good, I guess the only thing CI is complaining about is the newline in the changelog file

roypat (commented):
Could you squash your two commits into one? That's also no need to list me as a co-author, I was just repeating what the CI was saying :)

seafoodfry:
> Could you squash your two commits into one? That's also no need to list me as a co-author, I was just repeating what the CI was saying :)

Deal and done. ptal!

And I think i'm getting the hang of contributing here (when to squash commits and when not to).
But, again, thank you for the reviews and for your patience @roypat !

roypat (approved):

roypat:
> > Could you squash your two commits into one? That's also no need to list me as a co-author, I was just repeating what the CI was saying :)
> 
> Deal and done. ptal!

Thanks! Looks good now :)

> And I think i'm getting the hang of contributing here (when to squash commits and when not to). But, again, thank you for the reviews and for your patience @roypat !

Based on your experience, do you think there's anything we could clarify in our documentation about it, to make it easier for the next contributor? :o

roypat:
(also sorry for the PR cross-ref to rust-vmm, I messed up while copy-pasting >.>)

seafoodfry:
> Based on your experience, do you think there's anything we could clarify in our documentation about it, to make it easier for the next contributor? :o

Right now the only thing that comes to mind is something along the lines of "be diligent about squashing commits when implementing suggestions/fedbacks from PR reviews".
But i'm keeping this on my notes while I learn the cod. Hopefully after some more nontrivial PRs I'll get a better idea.

kalyazin (approved):

PR #4744, opened 2024-08-22
Updated CONTRIBUTING and getting-started docs

seafoodfry:
## Changes

1. Added some extra commands to ensure the kvm linux group is configured.
2. Added documentation detailing a way to make use of the `./tools/devctr` poetry project for the pre-commit.
3. Added a `.python-version` file to help the installation of the poetry project which supports python versions 3.10.*

## Reason

Had to figure out these details while getting started.
Figured I should try contributing this info back into the project.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet


CONTRIBUTING.md
```diff
@@ -65,6 +65,15 @@ cargo install rusty-hook
...
 ```
 
+If you are updating docs, you will also need to install
+[github.com/executablebooks/mdformat](https://github.com/executablebooks/mdformat):
+
+```bash
+python3 -m venv .venv
+source .venv/bin/activate
+python3 -m pip install mdformat
+```
```
pb8o, line 75: Alternatively you can run `./tools/devtool checkstyle`

seafoodfry: ah so I suggested this because of

https://github.com/firecracker-microvm/firecracker/blob/45a1a1be3efde06f06034fbe1798c3747c1424cd/pre-commit#L37-L39
(the pre-commit explicitly calls to `mdformat`.)

Now that I look at it, the pre-commit also has these other dependencoes explicitly called out

https://github.com/firecracker-microvm/firecracker/blob/45a1a1be3efde06f06034fbe1798c3747c1424cd/pre-commit#L32-L36


And for completeness, this is what I get when running `./tools/devtool checkstyle` (i spin up EC2s for actually running the code, as recommended in the getting started guide, but otherwise keep my fork on a mac):
```
/Users/user/go/src/github.com/firecracker/tools/functions: line 11: BASH_SOURCE: bad array subscript
[Firecracker ] /dev/kvm not found. Aborting.
```

So I think this PR in short is more of a question of: would it be ok to have a list of all depencies listed along with instructions for how to install them? or should the pre-commit and other such bits rely on `./tools/devtool` (and explicitly state that people should do code commiting on a linux machine that is configured in a manner similar to the machine we use for tests)? or some other option I'm not thinking of at the moment?

wdyt?

pb8o (dismissed):

CONTRIBUTING.md
seafoodfry: @pb8o should I make an issue to use `devtool checkstyle` in the pre-commit ?

roypat: The difference between `checkstyle` and the precommit hook is that the precommit hook in its current form will fix the formatting issues for you, while `checkstyle` is only going to yell at you :(

Nevertheless, having the style thingy fail on the kvm checks is still a bit silly, maybe we can have an issue to fix that instead? :o

CONTRIBUTING.md
```diff
@@ -216,5 +225,5 @@ message automatically.
 
 Forgot to add DCO to a commit? Amend it with `git commit --amend -s`.
 
-[^1]: Performance improvements in non-hot paths are unlikely to be considered
-    valuable.
+\[^1\]: Performance improvements in non-hot paths are unlikely to be considered
```
roypat, line 228: Mh, this mdformat broke this footnote. I think you also need to install the extensions that are installed in the docker container for it to correctly deal with github-style footnotes.

CONTRIBUTING.md
roypat: Actually, after leaving my other comment, I think its probably better to have a `./tools/devtool fixstyle` command that runs all the style-fixing stuff that the pre-commit hook tries to do inside of docker (so that it picks up all the correct versions of the tools and such), and then the pre-commit hook can just point to that?

pb8o: Looks good! Thanks for doing that!

CONTRIBUTING.md
seafoodfry: @roypat thank you again for the review!
This one comment of yours made me realize that a better way forward would be to actually make use of what existed so I dove a bit more and realized that `./tools/devctr` is the thing keeping track of the linting deps.
Since that project is a poetry project, I figured out a way to actually make use of it (and included pyenv because python 3.10 is somewhat old).

Thank you for catching this!

CONTRIBUTING.md
seafoodfry: Thank you @roypat and @pb8o for the reviews!
I rewrote to capture the latest round of ideas https://github.com/firecracker-microvm/firecracker/issues/4747 ptal

and also updated the PR to actually make use of the `./tools/devctr` poetry project - instead fo doing a weird rando thing with a python virtualenv.

seafoodfry:
Forced push to rebase with the latest from upstream/main (and because I forgot to cleanup one of the github-style footnotes).

CONTRIBUTING.md
```diff
@@ -106,7 +123,7 @@ Your contribution needs to meet the following standards:
   should be accompanied by a comment detailing its...
 
   - Justification, potentially including quantifiable reasons why safe
-    alternatives were not used (e.g. via a benchmark showing a valuable\[^1\]
+    alternatives were not used (e.g. via a benchmark showing a valuable[^1]
```
pb8o, line 126: this and below could be squashed in the previous commit.

pb8o (approved):

roypat (approved):
Looks good to me! Thanks :)

CONTRIBUTING.md
seafoodfry: Done!

seafoodfry:
Thank you again for the reviews @roypat and @pb8o !
I squashed the commits.

PR #4746, opened 2024-08-23
Use 6.1 guest kernel configs from AL

bchalios:
## Changes

Use AmazonLinux tailored guest configs for our 6.1 guest microVM kernels. Also, split the Kconfig for our CI kernels in a base part which will always be reflecting the AmazonLinux config and a set of patches that we apply on top of it before rebuilding our CI kernels. The patches live under `resources/guest_configs/patches/`.

## Reason

AmazonLinux now ships a microVM kernels for 6.1. We want to build with the corresponding Kconfigs and keep track of our CI-specific changes in a form of patches.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet


resources/guest_configs/patches/0002-ci-adjust-6.1-guest-kernel-configs-for-our-CI.patch
```diff
@@ -0,0 +1,76 @@
+From 911b916fdc53acfc7229302b029fd615f1e670cf Mon Sep 17 00:00:00 2001
```
pb8o, line 1: [commenting here since GH does not let me comment on commit title/body]

typo in commit title: updsate

resources/guest_configs/patches/0002-ci-adjust-6.1-guest-kernel-configs-for-our-CI.patch
```diff
@@ -0,0 +1,76 @@
...
+From: Babis Chalios <bchalios@amazon.es>
+Date: Fri, 23 Aug 2024 12:54:26 +0200
+Subject: [PATCH 2/2] ci: adjust 6.1 guest kernel configs for our CI
```
pb8o, line 4: I was thinking this could be combined with the other patch, but actually I prefer like you have done it here, it may be better to split this per gues kernel version (4.14, 5.10, 6.1) so we can drop them as we deprecate them. Though editing does not seem that hard.

pb8o (approved):

resources/rebuild.sh
```diff
@@ -179,16 +179,6 @@ function build_al_kernel {
         format="pe"
         target="Image"
         binary_path="arch/arm64/boot/$target"
-
-        # Patch 6.1 kernels on ARM with 6.10 patches for supporting VMGenID
-        # via DeviceTree bindings.
-        # TODO: drop this (and remove the patches from the repo) when AL backports the
```
roypat, line 179: If the patches we have here are the same as the upstream 6.10 patches, could we just point people to the upstream and say "backport these", instead of keeping them in our repo? Because I guess otherwise we might run into danger of our copy of the patches going stale

resources/guest_configs/patches/0002-ci-adjust-6.1-guest-kernel-configs-for-our-CI.patch
bchalios: I actually did it like that because the changes happened in different points in time (not from the same initial commit), and I thought it'd be nicer to have them in different patches (easier to revert, etc.).

resources/rebuild.sh
bchalios: If our patches go stale, the commit we will be pointing to will be stale too, right?

Manciukic (approved):

resources/rebuild.sh
bchalios: They won't backport it. They only back-port fixes, not new features

roypat: No, I mean they backport something _else_ (e.g. some unrelated bugfix), and that causes our patches to no longer apply cleanly

bchalios: oh ok, got it. I _think_ it's unlikely because these patches basically change only VMGenID things, but I see what you mean (it took me a bit of time :P).

I think it makes sense, but at the same time the same would apply for the upstream commits, right?

roypat: Yeah, but at that point, if we only say "backport these", then it's implied that the whoever follows the guide will have to work out potential conflicts themselves, instead of us updating the .patch files.

I guess long story short is I'm not a fan of keeping the patches around, although if others are fine with it, feel free to just merge with the approvals you have :)

Manciukic (dismissed):

resources/rebuild.sh
bchalios: Nope, I think you're right. I've changed the docs to point to the PR from Linus. @Manciukic @pb8o could you please take another look to see if you're fine with that

roypat (approved):

Manciukic (approved):

pb8o (approved):

PR #4748, opened 2024-08-23
Less memory checks in the queue

ShadowCurse:
## Changes
Replace accesses to the queue objects reside in the guest memory from always using `GuestMemoryMmap` to storing pointers to them directly. The reason we can do this is that these objects do not move in the guest memory, so storing direct pointers to them is safe.

## Reason
This optimization avoid many memory check when accessing guest memory and thus can improves performance of all virtio devices.

# Becnes
Added benchmarks for `Queue::pop/add_used`, `DescriptorChain::next_descriptor` and `Request::parse`(from virtio block) and these are results for x86:
```
queue_pop_1             time:   [5.2791 ns 5.2791 ns 5.2792 ns]
                        change: [-90.046% -90.033% -90.021%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 19 outliers among 200 measurements (9.50%)
  3 (1.50%) low mild
  5 (2.50%) high mild
  11 (5.50%) high severe

queue_pop_4             time:   [18.621 ns 18.621 ns 18.622 ns]
                        change: [-89.364% -89.361% -89.358%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 200 measurements (4.00%)
  4 (2.00%) low mild
  2 (1.00%) high mild
  2 (1.00%) high severe

queue_pop_16            time:   [71.963 ns 71.964 ns 71.966 ns]
                        change: [-89.351% -89.347% -89.343%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 200 measurements (6.00%)
  5 (2.50%) high mild
  7 (3.50%) high severe

queue_add_used_1        time:   [1.9462 ns 1.9463 ns 1.9465 ns]
                        change: [-92.597% -92.595% -92.593%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 200 measurements (4.50%)
  8 (4.00%) high mild
  1 (0.50%) high severe

queue_add_used_16       time:   [55.649 ns 55.656 ns 55.666 ns]
                        change: [-86.365% -86.349% -86.327%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 200 measurements (3.50%)
  7 (3.50%) high severe

queue_add_used_256      time:   [720.92 ns 720.93 ns 720.94 ns]
                        change: [-87.653% -87.649% -87.646%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 200 measurements (4.50%)
  4 (2.00%) high mild
  5 (2.50%) high severe

next_descriptor_1       time:   [10.002 ns 10.003 ns 10.003 ns]
                        change: [-74.067% -74.046% -74.026%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 23 outliers among 200 measurements (11.50%)
  8 (4.00%) low mild
  9 (4.50%) high mild
  6 (3.00%) high severe

next_descriptor_2       time:   [10.021 ns 10.023 ns 10.024 ns]
                        change: [-80.125% -80.113% -80.103%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 200 measurements (0.50%)
  1 (0.50%) high severe

next_descriptor_4       time:   [12.431 ns 12.459 ns 12.489 ns]
                        change: [-83.198% -83.167% -83.133%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 200 measurements (7.50%)
  12 (6.00%) high mild
  3 (1.50%) high severe

next_descriptor_16      time:   [60.834 ns 61.115 ns 61.393 ns]
                        change: [-73.133% -72.990% -72.853%] (p = 0.00 < 0.05)
                        Performance has improved.

request_parse           time:   [833.56 ps 833.57 ps 833.58 ps]
                        change: [-39.999% -39.997% -39.995%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 23 outliers among 200 measurements (11.50%)
  10 (5.00%) low mild
  8 (4.00%) high mild
  5 (2.50%) high severe

```

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


src/vmm/src/devices/virtio/queue.rs
```diff
@@ -522,11 +522,10 @@ impl Queue {
     pub fn is_valid<M: GuestMemory>(&self, mem: &M) -> bool {
         if !self.is_layout_valid(mem) {
             false
-        } else if self.len(mem) > self.max_size {
+        } else if self.size > self.max_size {
```
roypat, line 500: `self.size <= self.max_size` is already checked in `is_layout_valid`. But this is also a fundamentally different check. Checking `size` against `max_size` checks whether the queue configuration the guest passed us via MMIO before activation makes sense at all. Checking that the `len` does not exceed `max_size` (well, I guess really this should be `size`?) is a runtime check that ensures the guest is using the queue correctly, e.g. it doesn't tell us there are more available descriptors than is mathematically possible. I'm not sure why you want to drop that check

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -242,13 +290,170 @@ impl Queue {
...
+                .cast::<u16>()
+                .write_volatile(val)
+        }
+    }
+
+    /// Get element from UsedRing.ring at index
+    /// # Safety
+    /// The `index` parameter should be in 0..queue_size bounds
+    #[inline(always)]
+    pub unsafe fn used_ring_ring_set(&mut self, index: usize, val: UsedElement) {
```
roypat, line 391: Can you please add kani proofs covering these new functions? E.g. just something simple that `kani::assumes` the safety invariants, to verify that they are indeed sufficient to make the function safe.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -242,13 +290,170 @@ impl Queue {
...
+    fn used_ring_size(&self) -> usize {
+        std::mem::size_of::<u16>()
+            + std::mem::size_of::<u16>()
+            + std::mem::size_of::<UsedElement>() * usize::from(self.size)
+            + std::mem::size_of::<u16>()
+    }
+
+    /// Set up pointers to the queue objects in the guest memory
+    /// and mark memory dirty for those objects
+    pub fn set_pointers<M: GuestMemory>(&mut self, mem: &M) -> Result<(), QueueError> {
```
roypat, line 300: nit: Maybe this function would be better named `initialize` or `activate` of something?

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -899,21 +958,6 @@ mod verification {
         }
     }
 
-    mod stubs {
```
roypat, line 902: this is a note to myself, but now that the functions under test are a lot simpler, I want to have a look at whether we can generalize the proofs in this module.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -239,16 +262,173 @@ impl Queue {
...
+                .add(std::mem::size_of::<u16>().unchecked_mul(2))
+                .cast::<UsedElement>()
+                .add(index)
+                .write_volatile(val)
+        }
+    }
+
+    #[cfg(test)]
+    #[inline(always)]
+    pub fn used_ring_avail_event_get(&mut self) -> u16 {
```
roypat, line 404: Why the `_get`/`_set` suffixes? I don't think I've seen that convention anywhere before

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: The reason why I need to remove the `len(mem)`  from here is that the validation of queue happen before device activation. So with new changes the `len` method will use null pointers to try to calculate len, so it will fail.
Changing the logic of testing if a queue is valid seems beyond this PR, even though it would totally make sense to do the validation check in `activate` method and not before it. Also the `is_valid` is not be called during an actual runtime because all `is_valid` checks are in the `debug_assert!` macros, so at runtime guest can easily "make up" queue len bigger than it should be.
I see now that `is_layout_valid` is indeed check the `self.size <= self.max_size`, so I can remove the `is_valid` all together.

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: I guess `initialize` makes sense. I'll rename it later if no one has better ideas.

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: Because I could not use types to abstract the queue objects (technically I could, but I would need to store the queue len in those types and why would I duplicate same info several times...).
The name came from an imaginary semantic as "if" I used types to abstract things, so `used_ring_avail_event_get` can be interpreted as `used_ring.avail_event.get()`

src/vmm/src/devices/virtio/queue.rs
roypat: Mh, that makes sense, thanks for the explanation! I think the `is_valid` thing used to also do the check after snapshot restore, but that's of questionable value, too, I guess, because `pop` already has this "guest doesn't set avail_idx to nonsense" check.

src/vmm/src/devices/virtio/queue.rs
roypat: I see. I guess changing it is more of a nit, so feel free to resolve if you disagree

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: done in the last commit.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -686,26 +686,6 @@ impl Queue {
         mem.write_obj(avail_event, avail_event_addr).unwrap();
     }
 
-    /// Helper method that writes to the `idx` field of the used ring.
-    #[inline(always)]
-    fn set_used_ring_idx<M: GuestMemory>(&mut self, next_used: u16, mem: &M) {
-        debug_assert!(self.is_valid(mem));
```
bchalios: can we keep these `debug_assert!`s in the new functions?

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -625,14 +625,7 @@ impl Queue {
...
 
     /// Puts an available descriptor head into the used ring for use by the guest.
-    pub fn add_used<M: GuestMemory>(
-        &mut self,
-        mem: &M,
-        desc_index: u16,
-        len: u32,
-    ) -> Result<(), QueueError> {
-        debug_assert!(self.is_valid(mem));
-
```
bchalios: Can we merge the commit that does this change with the one that switches from `write_used_ring` to `used_ring_ring_set`?

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -95,38 +89,20 @@ pub struct DescriptorChain<'a, M: GuestMemory = GuestMemoryMmap> {
...
     /// Note that the desc_table and queue_size are assumed to be validated by the caller.
-    fn checked_new(
-        mem: &'a M,
-        desc_table: GuestAddress,
-        queue_size: u16,
-        index: u16,
-    ) -> Option<Self> {
-        if index >= queue_size {
+    fn checked_new(desc_table_ptr: *const Descriptor, queue_size: u16, index: u16) -> Option<Self> {
+        if queue_size <= index {
```
bchalios, line 97: :) What's with flipping the comparison?

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -618,9 +618,7 @@ impl Queue {
     /// successfully enabled. Otherwise it means that one or more descriptors can still be consumed
     /// from the available ring and we can't guarantee that there will be a notification. In this
     /// case the caller might want to consume the mentioned descriptors and call this method again.
-    pub fn try_enable_notification<M: GuestMemory>(&mut self, mem: &M) -> bool {
-        debug_assert!(self.is_valid(mem));
-
```
bchalios: Please, merge this commit with the change that makes with the commit that makes `mem` unused

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -1073,10 +1071,52 @@ mod verification {
 
     #[kani::proof]
     #[kani::unwind(0)]
-    fn verify_set_used_ring_avail_event() {
+    fn verify_avail_ring_idx_get() {
+        let ProofContext(queue, _) = ProofContext::bounded_queue();
```
roypat, line 1075: If I replace this one with `kani::any()`, I get a failure about a null pointer dereference. It's because `impl kani::Arbitrary for ProofContext` is missing `queue.set_pointers`. If I add that, then I can apply the following diff

```suggestion
        let ProofContext(queue, _) = kani::any();
```

which passes in 25 seconds, and is a waaaay more general proof

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -504,9 +504,7 @@ impl Queue {
     }
 
     /// Pop the first available descriptor chain from the avail ring.
-    pub fn pop<M: GuestMemory>(&mut self, mem: &M) -> Option<DescriptorChain> {
-        debug_assert!(self.is_valid(mem));
```
bchalios: I was about to comment here as well, to "merge the commit that makes this change to the commit that removed `mem`", but then I noticed that this `debug_assert!`s are used by `kani`. Are we making the kani proofs "useless" by dropping these?

bchalios (commented):
Ok, did a first round. It generally seems good. A few things in comments

src/vmm/benches/queue.rs
```diff
@@ -0,0 +1,241 @@
...
+use vmm::utilities::test_utils::single_region_mem;
+
+pub fn queue_benchmark(c: &mut Criterion) {
+    let mem = single_region_mem(2 * 65562);
+    let rxq = VirtQueue::new(GuestAddress(0), &mem, 256);
+    let mut queue = rxq.create_queue();
+
+    rxq.dtable[0].set(2048, 65562, 0x2, 0);
+    rxq.avail.ring[0].set(0);
+    rxq.avail.idx.set(1);
```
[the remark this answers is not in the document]
ShadowCurse: I moved benches to another PR: #4783

src/vmm/benches/queue.rs
```diff
@@ -0,0 +1,241 @@
...
+            while let Some(d) = head {
+                head = d.next_descriptor();
+            }
+        })
+    });
+
+    rxq.dtable[0].set(2048, 65562 / 2, 0x3, 1);
+    rxq.dtable[1].set(2048 + 65562 / 2, 65562 / 2, 0x2, 0);
+    rxq.avail.ring[0].set(0);
+    rxq.avail.idx.set(2);
```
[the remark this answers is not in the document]
ShadowCurse: I moved benches to another PR: #4783

src/vmm/benches/queue.rs
```diff
@@ -0,0 +1,241 @@
...
+            }
+        })
+    });
+
+    rxq.dtable[0].set(2048, 65562 / 4, 0x3, 1);
+    rxq.dtable[1].set(2048 + (65562 / 4), 65562 / 4, 0x3, 2);
+    rxq.dtable[2].set(2048 + (65562 / 4) * 2, 65562 / 4, 0x3, 3);
+    rxq.dtable[3].set(2048 + (65562 / 4) * 3, 65562 / 4, 0x2, 0);
+    rxq.avail.ring[0].set(0);
+    rxq.avail.idx.set(4);
```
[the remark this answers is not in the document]
ShadowCurse: I moved benches to another PR: #4783

src/vmm/benches/block_request.rs
```diff
@@ -0,0 +1,46 @@
...
+    let mem = single_region_mem(2 * 65562);
+    let virt_queue = VirtQueue::new(GuestAddress(0), &mem, 256);
+
+    let chain = RequestDescriptorChain::new(&virt_queue);
+    let request_header = RequestHeader::new(VIRTIO_BLK_T_IN, 99);
+    chain.set_header(request_header);
+
+    let mut queue = virt_queue.create_queue();
+
+    // SAFETY: queue has 1 desc chanin
```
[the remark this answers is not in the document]
ShadowCurse: I moved benches to another PR: #4783

src/vmm/src/devices/virtio/queue.rs
roypat: Thanks! Could you also do it for the pre-existing ones?

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -231,6 +276,9 @@ pub struct Queue {
     pub num_added: Wrapping<u16>,
 }
 
+/// SAFETY: Only one thread will access this at a time
```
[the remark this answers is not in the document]
ShadowCurse: 1. Queues are only accessed by vmm thread. You might be thinking about mmio region that can be accessed by vcpus and vmm threads.
2. Pointers are not `Send` by default.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -222,6 +223,50 @@ pub struct Queue {
     /// Guest physical address of the used ring
     pub used_ring_address: GuestAddress,
 
+    /// This is a user space pointer
+    /// to the available ring, set up by the
+    /// guest. Getting access to the underling
+    /// data structure should only occur after the
```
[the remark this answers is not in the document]
ShadowCurse: I am not sure there is ambiguity. This is a 'pointer'. It would be strange for it to be a guest pointer. I will update this to `host pointer`.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -242,13 +290,170 @@ impl Queue {
...
+    }
+
+    /// Set up pointers to the queue objects in the guest memory
+    /// and mark memory dirty for those objects
+    pub fn set_pointers<M: GuestMemory>(&mut self, mem: &M) -> Result<(), QueueError> {
+        let size = self.desc_table_size();
+        let slice = mem
+            .get_slice(self.desc_table_address, size)
+            .map_err(QueueError::MemoryError)?;
+        slice.bitmap().mark_dirty(0, size);
```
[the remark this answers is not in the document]
ShadowCurse: I don't see a problem here. This pattern only used in 1 function. Moving it into another function will just create unnecessary indirection.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -242,13 +290,170 @@ impl Queue {
...
+        self.used_ring_ptr = slice.ptr_guard_mut().as_ptr().cast();
+
+        Ok(())
+    }
+
+    /// Mark memory used for queue objects as dirty.
+    pub fn mark_memory_dirty<M: GuestMemory>(&self, mem: &M) -> Result<(), QueueError> {
+        let size = self.desc_table_size();
+        let slice = mem
```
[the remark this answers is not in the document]
ShadowCurse: Same here.

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: new functions do not have access to guest memory object

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: Just a preference. Small items on the left, big on the right.

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: If we do this check on `pop` each time, do we really need to do it one more time on activation/initialization? I will adjust the commit message about this. Also as far as I can think, even if guest tells us to pop more descriptors than the queue should hold, we will just go around the queue and start poping same valid descriptors again.

ShadowCurse: but if you strongly want this check to be in `initialize`, I can add it back.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: The terminology "host virtual address", "guest virtual address", "guest physical address" is very common in the virtualization world. It would be nice if we were accurate.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: I wanted all these functions to be as small as possible and without any logic inside. I think `unsafe` is enough to attract attention to the requirements for `index`. Also as I can see, I can just make them private as they are only used inside the queue. Will this suffice?

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: There is no problem. It's just less typing, more readable code and if in the future someone needs to change something they only need to do it in one place. Also, there's no indirection in terms of actually executable code, we can just inline this (actually, I'd be surprised if the compiler wouldn't do it by itself).

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: Hmmm, I saw in later commits that some times it's handy to have this `unsafe` because we "anyways" check before calling this function.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -214,23 +190,70 @@ pub struct Queue {
...
+    /// Representation of in memory struct layout.
+    // struct UsedRing {
+    //     flags: u16,
+    //     idx: u16,
+    //     ring: [UsedElement; <queue size>],
+    //     avail_event: u16,
+    // }
+    /// Because types in the UsedRing are different (u16 and u32)
+    /// store pointer as *mut u8.
+    pub used_ring_ptr: *mut u8,
```
roypat, line 243: maybe

```suggestion
    pub used_ring_ptr: *mut libc::c_void,
```

?

(for all of these, really)

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: ok

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: `Send` just means we can send instance of the type to another thread (and we do it with `Arc<Mutex<Device..>>` thing inside the `Bus`). But if we "send" the object, it means that we should not have write access to it (until type is also `Sync`). Before we did not need `Send` because all types were Rust verifiable (with lifetimes). But pointers are not. By saying that we only access queue in one thread, we say that sending these pointers is fine, as no one will use them except vmm thread.
At least this is my understanding. It is hard for my C++ brain to understand all intricacies of Rust memory model.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: By indirection I mean the human will look at 1 function and then need to look at another one. The whole function is 20 line of code. We don't need to do this abstraction for such small thing.

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: Why? It is nice to know what types are behind the pointers. And it makes math a bit easier.

roypat: well, for this specific one I had the idea because of the comment. u8 is already not the type of the stuff we point at, and `void *` is the canonical type for these opaque kinda things.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: The proofs I added use `kany::any` for `ProofContext`. And there we have `is_valid` call. So the only purpose these assertions in normal functions had, were for test code to verify that the memory layout that the test set up is valid.

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: why use type from `libc` when Rust has u8 builtin? Technically `*void` in Rust will be `*()`. But this feel very wrong. With `u8`, `u16` we at least see the underlying type. So when doing arithmetic we know that `*u8 + 1` just increases pointer by 1, while `*u16 + 1` actually moves it 2 bytes forward. With `libc::void` it is more confusing (even though technically it is same as `*u8`).

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: Actually as I already had `desc_table_size` and friends, so I updated them to do smth like you suggested.

src/vmm/src/devices/virtio/queue.rs
bchalios: > So the only purpose these assertions in normal functions had, were for test code to verify that the memory layout that the test set up is valid.

Right, so what is the problem with that?

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: > By saying that we only access queue in one thread, we say that sending these pointers is fine, as no one will use them except vmm thread.

If I understand the semantics correctly, by saying we only access the type from one thread we say: "We don't care if this is `Send` because we won't "Send" it anyways, but we mark it `Send` so that the compiler stops complaining. If that is the case, it sounds wrong. Because, the moment the assumption (access from one thread only) changes (e.g. because someone comes and adds multi-threading IO handling) this will be fundamentally broken.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: That's not what I suggested. And it's actually 3 times more indirection (by your definition above).

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: Yes, please just mark them private and it's fine.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: Ok, what is your suggestion for this then?

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: I don't understand and I don't want to guess what you exactly want. If you have better ideas just write the whole thing down here.

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: There is no problem.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: I suggest you understand if the code here is `Send` and if it is document why. If not, mention clearly in the comment why it is not.

FWIW, I think the type is indeed `Send` (and `Sync`).
`Send` because this `Queue` is the only object holding these pointers, and `Sync` because it doesn't allow for interior mutability.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
bchalios: I wrote the thing in the first comment of this thread.

src/vmm/src/devices/virtio/queue.rs
bchalios: So, is there a way to keep these `debug_assert`s? If the problem is the fact we don't have the `mem` object now, Can we re-write `is_valid` in a way that it matches with the current state of the code?

ShadowCurse: `is_valid` needs access to guest memory in order to check bounds on the queue objects. None of the methods now have access to the guest memory, so they cannot call `is_valid`.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: done

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: added it back as an assert, so it will have same behavior as `pop`

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: why? these are 2 different changes?

ShadowCurse: Also can you give commits names? It is very annoying trying to find commit where you left the comment.

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: I would prefer to keep these separate. Code without this commit is technically valid. And squashing it with `refactor(queue): remove set_used_ring_avail_event method` will only make it more complicated.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: I think the fact that only VMM thread will access `Queue`s is enough to mark it `Send`. Even thought we actually do not "Send" queues as they are inside devices which are allocated on the heap. The only thing we "Send" are the `Arc<Mutex<....>>`.

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: Alright, I reworked this one more time, see if this fits you

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: @roypat as an Rust safety and correctness ambassador, do you have a better resolution for this issue?

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -839,20 +861,10 @@ mod verification {
...
-            ProofContext(queue, mem)
-        }
-
-        /// Creates a [`ProofContext`] where the queue layout is fixed to a valid one and where
-        /// guest memory is initialized to all zeros.
-        pub fn bounded() -> Self {
-            let mem = setup_zeroed_guest_memory();
-            let queue = less_arbitrary_queue();
+            let mut queue = less_arbitrary_queue();
+            queue.initialize(&mem).unwrap();
```
roypat, line 865: I think you'll have to inline `initialize` here if you want to keep the assertion. Otherwise kani is gonna keep hitting it (and we don't want to `kani::assume` that we don't hit it outside of the actual proof for `pop` because technically all those other functions can be called with the queue in a "guest messed it up" state).

ShadowCurse: What about marking assert as `cfg(not(kani))`?

roypat: not a huge fan, but if you strongly prefer it, sure

src/vmm/src/devices/virtio/queue.rs
roypat: wohoo, we could

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
roypat: shouldnt the same logic as with the iovec stuff apply here?

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: done

src/vmm/src/devices/virtio/queue.rs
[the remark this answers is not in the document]
ShadowCurse: `IoVecBuffer` does not allow for interior mutability. `Queue` does. (not really `interior`, but close to it)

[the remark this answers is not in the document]
bchalios: So, we can hold a `&Queue` reference and change the pointer? Also, IIUC the interior mutability thing is a requirement for `Sync` not `Send`. My understanding coming from the nomicon: https://doc.rust-lang.org/nomicon/send-and-sync.html

> Something can safely be Send unless it shares mutable state with something else without enforcing exclusive access to it. Each Carton has a unique pointer, so we're good.

and

> What about Sync? For Carton to be Sync we have to enforce that you can't write to something stored in a &Carton while that same something could be read or written to from another &Carton. Since you need an &mut Carton to write to the pointer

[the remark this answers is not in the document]
ShadowCurse: > So, we can hold a &Queue reference and change the pointer?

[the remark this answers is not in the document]
ShadowCurse: But because all methods that use pointers and mutate memory need `&mut Queue` maybe everything is fine.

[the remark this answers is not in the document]
roypat: > IoVecBuffer does not allow for interior mutability.
IoVecBufferMut then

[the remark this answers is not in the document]
ShadowCurse: Settled on new safety comment.

bchalios (approved):

roypat (approved):
(for the queue.rs and kani parts)

roypat (approved):

PR #4780, opened 2024-09-03
feat: update block perf tests to use fio json

ShadowCurse:
## Changes
Block tests were using fio logs to calculate bandwidth. The issue is that the values in logs are in KB/sec and we were just summing them up, so getting a sum of averages. But what we want is just an average bandwidth for the test.
Now we use json output of fio which contains much more info in a nice format. This info already includes all bandwidth calculations. This change will also allow us to output other metrics in the future if we want.

## Reason
More correct block performance tests

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/tests/integration_tests/performance/test_block_ab.py
+++ b/tests/integration_tests/performance/test_block_ab.py
@@ -3,6 +3,7 @@
 """Performance benchmark for block device emulation."""
 
 import concurrent
+import json
 import os
 import shutil
 from pathlib import Path
@@ -67,18 +68,17 @@ def run_fio(microvm, mode, block_size):
         .with_arg("--cpus_allowed_policy=split")
         .with_arg("--randrepeat=0")
         .with_arg(f"--runtime={RUNTIME_SEC}")
-        .with_arg(f"--write_bw_log={mode}")
-        .with_arg("--log_avg_msec=1000")
         .with_arg("--output-format=json+")
+        .with_arg("--output=fio.json")
         .build()
     )
 
-    logs_path = Path(microvm.jailer.chroot_base_with_id()) / "fio_output"
+    output_path = Path(microvm.jailer.chroot_base_with_id()) / "fio_output"
 
-    if logs_path.is_dir():
-        shutil.rmtree(logs_path)
+    if output_path.is_dir():
+        shutil.rmtree(output_path)
 
-    logs_path.mkdir()
+    output_path.mkdir()
 
     prepare_microvm_for_test(microvm)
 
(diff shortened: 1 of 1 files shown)


tests/integration_tests/performance/test_block_ab.py
```diff
@@ -96,38 +96,23 @@ def run_fio(microvm, mode, block_size):
         assert rc == 0, stderr
         assert stderr == ""
 
-        microvm.ssh.scp_get("/tmp/*.log", logs_path)
-        microvm.ssh.check_output("rm /tmp/*.log")
+        microvm.ssh.scp_get("/tmp/fio.json", output_path)
+        microvm.ssh.check_output("rm /tmp/fio.json")
 
-        return logs_path, cpu_load_future.result()
+        return output_path, cpu_load_future.result()
```
pb8o, line 102: `return output_path / "fio.json", ...`

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -96,38 +96,23 @@ def run_fio(microvm, mode, block_size):
...
 
 
-def process_fio_logs(vm, fio_mode, logs_dir, metrics):
-    """Parses the fio logs in `{logs_dir}/{fio_mode}_bw.*.log and emits their contents as CloudWatch metrics"""
+def process_fio_json(json_path, metrics):
+    """Parse fio json ouput"""
+    with open(os.path.join(json_path, "fio.json"), "r", encoding="utf-8") as file:
+        j = json.load(file)
```
pb8o, line 108: Combined with above comment:

```python
fio = json.loads(json_path.read_text(encoding="utf-8") 
```

tests/integration_tests/performance/test_block_ab.py
ShadowCurse: done

tests/integration_tests/performance/test_block_ab.py
ShadowCurse: done

ShadowCurse:
Each line in the log contains `value seen over the specified period of time` (I assume this is an average over 1s because we specify to log every second). So each line contains KB/s (not just KB). We were summing these KB/s and were getting a `sum of averages`.
And if they were KB, then our tests were completely wrong, as were calculating the total amount of data send instead of the bandwidth.

roypat (changes requested):
Ah, I see now. This won't work with our A/B-tests, because they require a data series to be emitted, not just a single data point.

I admittedly still don't quite understand what was wrong with the old approach. It gives you a data series where each entry is "number of bytes written/read since the last entry", with entries being emitted once per second. That's different from "kb/s", because its not "average bandwidth since the test started". So it's fine to sum up the data points at each second (and if it weren't, I don't see why your summing up of _all_ datapoints wouldn't be just as problematic).

ShadowCurse:
I am pretty sure A/B works with 1 data point as this is what is happening now with VMs with 1 vcpu (1 fio job). We sum all values and emit 1 metric.

PR #4782, opened 2024-09-03
First intermittent failures of test_balloon.py::test_size_reduction

roypat:
## Changes

Adds a bunch of missing `wait_for_up`s, and clears up some excessive logging

## Reason

Missing `wait_for_up` in various balloon tests caused RSS measurements to be taken during the boot process, which later results in intermittent failures. For example, in `test_size_reduction`, we try to inflate the balloon based on an RSS reading taken during boot, which leaves too little memory available for a booted guest.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


tests/framework/microvm.py
```diff
@@ -1001,7 +1001,7 @@ def wait_for_up(self, timeout=10):
         :param timeout: seconds to wait.
         """
         try:
-            rc, stdout, stderr = self.ssh.run("true", timeout)
+            rc, stdout, stderr = self.ssh.run("true", timeout, warn_on_fail=False)
```
pb8o, line 1004: TBH I think those messages are fine. We can print a message saying "timeout messages are expected while the microvm boots".

tests/framework/utils.py
```diff
@@ -401,20 +405,22 @@ def run_cmd(cmd, check=False, shell=True, cwd=None, timeout=None) -> CommandRetu
...
-        )
+        if warn_on_fail:
+            # Log the message with one call so that multiple statuses
+            # don't get mixed up
+            CMDLOG.warning(
+                "Timeout executing command: %s\n",
+                _format_output_message(proc, stdout, stderr),
+            )
 
         raise
```
pb8o, line 416: Maybe instead of warning we should attach proc, stdout and stderr to the exception and raise that. The caller can choose to suppress it or display it.

tests/framework/utils.py
```diff
@@ -401,20 +405,22 @@ def run_cmd(cmd, check=False, shell=True, cwd=None, timeout=None) -> CommandRetu
...
 
     output_message = _format_output_message(proc, stdout, stderr)
 
     # If a non-zero return code was thrown, raise an exception
     if check and proc.returncode != 0:
-        CMDLOG.warning("Command failed: %s\n", output_message)
+        if warn_on_fail:
+            CMDLOG.warning("Command failed: %s\n", output_message)
 
         raise ChildProcessError(output_message)
```
pb8o, line 425: Same this seems redundant since we attachh output_message to the exception.

tests/framework/utils.py
roypat: I like the explicit logs better because they always get printed in full, while exception messages get truncated by pytest

pb8o: Could we change the logging level in the caller? 

```
prev_level = CMDLOG.getLevel()
CMDLOG.getLevel(logging.ERROR)
...
CMDLOG.setLevel(prev_level)
```

I imagine another option is to use the `warnings.warn`, but that would not get printed the same as with `logging`.

roypat: Would that work with the parallel tests? But it's a lot more code than this, what advantage do you see in having the caller do it like this instead of just passing a flag? :o

tests/framework/microvm.py
roypat: They're quite spammy tbh. The one in `init_connection` has 20 retries, and if it does need all or most of them, then scrolling through them gets old quite fast :sweat_smile:

tests/integration_tests/functional/test_cmd_line_start.py
```diff
@@ -432,7 +432,7 @@ def test_config_start_and_mmds_with_api(uvm_plain, vm_config_file):
     assert response.json() == {}
 
     # Populate MMDS with data.
-    response = test_microvm.api.mmds.put(**data_store)
```
bchalios, line 437: I wonder why our Python linters don't catch these things

tests/integration_tests/functional/test_balloon.py
```diff
@@ -341,17 +341,28 @@ def test_stats(uvm_plain_any):
...
+
     # Get an initial reading of the stats.
     initial_stats = test_microvm.api.balloon_stats.get().json()
 
+    # Major faults happen when a page fault has to be satisfied from disk. They are not
+    # triggered by our `make_guest_dirty_memory` workload, as it uses MAP_ANONYMOUS, which
+    # only triggers minor faults. However, during the boot process, things are read from the
+    # rootfs, so we should at least see a non-zero number of major faults.
+    assert initial_stats["major_faults"] > 0
+
```
bchalios, line 365: Do we even need that check? Major PFs happening during boot is not a phenomenon related with the balloon test and how can this thing even be non-zero if we don't boot from initrd?

bchalios (dismissed):
Some minor comments

tests/integration_tests/functional/test_cmd_line_start.py
roypat: I honestly have no idea. My IDE caught it, so it should be possible

tests/integration_tests/functional/test_balloon.py
roypat: I think the point of this test is more that the reporting of stats from the guest works. So its less about testing that major page faults happen (they definitely do), but more about that the statistics reporting by the balloon device works.

tests/framework/microvm.py
pb8o: For example we could change `_init_connection` to:

```
        wait=wait_chain(wait_fixed(.5), wait_fixed(.3)),
        stop=stop_after_delay(3),
```

This would wait .5s on the 1st attempt, then .3s until we reach 3s.

roypat: Alright, I've dropped the commit, since it is kinda orthogonal to the main goal of this PR anyway, maybe can revisit the logging in another context and solve it properly :)

pb8o (approved):

bchalios (approved):
:shipit:

PR #4790, opened 2024-09-09
fix(release): ensure debuginfo file contains debugging info

pb8o:
Seems recently Cargo defaulted to use `strip=debuginfo`. This inadvertently made our debuginfo files much smaller and that much less useful.

Fix the issue by using `strip=none` and add a test so that it breaks if this somehow changes again.

## Changes

- Fix `debuginfo` files
- Move the test together with the static link check as they are similar.

## Reason

`debuginfo` files can be used to troubleshoot issues, so we want to keep them functional.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet


roypat (approved):

tests/integration_tests/build/test_build.py
```diff
@@ -0,0 +1,32 @@
...
+        ".debug_aranges",
+        ".debug_info",
+        ".debug_abbrev",
+        ".debug_line",
+        ".debug_frame",
+        ".debug_str",
+        ".debug_ranges",
+    }
+    missing_sections = needed_sections - matches
+    assert missing_sections == set()
```
roypat, line 32: if you do another revision for any reason, maybe nit:

```suggestion
    assert not missing_sections, missing_sections
```

so that pytest prints out the full set (because the condition part gets trimmed when printing failures). But not very important given that for this test we don't really care _what's_ inside the `missing_sections`.

pb8o: This is what I get now.

```
>>>>>>>>>>>>>>>>>>>>>>>>>>>> traceback >>>>>>>>>>>>>>>>>>>>>>>>>>>>>
integration_tests/build/test_build.py:33: in test_release_debuginfo
    assert missing_sections == set()
E   AssertionError: assert {'.debug_foo'} == set()
E     Extra items in the left set:
E     '.debug_foo'
E     Full diff:
E     - set()
E     + {'.debug_foo'}
```

Do you mean if the diff is too large, we don't see everything? Maybe we should enable `-v` in pytest options.

roypat: Mh, maybe pytest actually does it more cleverly for sets, but I remember that for dictionaries that was indeed the case (diff too large, and pytest at some point just goes `...`)

pb8o: Yeah I think with `-v` or `-vv` it will not omit info.

roypat (dismissed):

Cargo.toml
```diff
@@ -27,3 +27,7 @@ panic = "abort"
 [profile.release]
 panic = "abort"
 lto = true
+strip = "none"
```
kalyazin, line 30: Just leaving a note here that we probably don't want to merge it just yet as the optional tests (benches) are failing, because they now use unstripped binaries and take longer. We're still looking into that.

roypat: They also fail on https://github.com/firecracker-microvm/firecracker/pull/4798 for example, which only touched python code, so I don't think it has anything to do with stripping (in fact, `cargo bench` doesn't use the `release` profile: https://doc.rust-lang.org/cargo/commands/cargo-bench.html#description). The new queue benchmarks simply seem very volatile

tests/integration_tests/build/test_build.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
pb8o, line 1: hmm I should move this to functional so as to not interfere with #4803

pb8o: done

roypat (dismissed):

Cargo.toml
kalyazin: Indeed, it looks like that.
Shall we add an entry in the changelog for that?

pb8o: Added changelog entry.

roypat (approved):

ShadowCurse (approved):

kalyazin (approved):

PR #4799, opened 2024-09-12
Use `readv` for the RX path of the network device to avoid one memory copy per frame

bchalios:
## Changes

Change the network device emulation to avoid the intermediate memory copy from the TAP device to Firecracker memory buffer. To achieve this, we use `readv` system call to read from the TAP device directly in guest memory.

## Reason

Currently, on the RX path of our network device we perform two memory copies for receiving a network frame from the TAP device to the guest driver. We first read the frame from the TAP device inside a memory buffer maintained by Firecracker. Next, we copy the frame from the buffer inside guest memory. We need the intermediate buffer because, typically, the guest side buffer is scattered across the guest memory.

We had tried to implement this optimization [before](https://github.com/firecracker-microvm/firecracker/pull/2958) but we observed performance regression which was due to overhead related with parsing of guest memory buffers (aka `DescriptorChain` objects).

Recently, we merged a few improvements in the logic that parses `DescriptorChain` objects from VirtIO queues (#4723, #4748) which allow us to parse descriptor chains faster, so the overheads we experienced before reduce significantly.
Moreover, we apply a further optimization, where we move the parsing of `DescriptorChain` objects outside the hot-path.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/src/vmm/src/devices/virtio/iov_deque.rs
+++ b/src/vmm/src/devices/virtio/iov_deque.rs
@@ -71,9 +71,9 @@ pub enum IovDequeError {
 // so making a slice out of them does not require any copies.
 #[derive(Debug)]
 pub struct IovDeque {
-    iov: *mut libc::iovec,
-    start: u16,
-    len: u16,
+    pub iov: *mut libc::iovec,
+    pub start: u16,
+    pub len: u16,
 }
 
 // SAFETY: This is `Send`. We hold sole ownership of the underlying buffer.
@@ -215,7 +215,7 @@ impl IovDeque {
 
     /// Returns `true` if the [`IovDeque`] is full, `false` otherwise
     #[inline(always)]
-    fn is_full(&self) -> bool {
+    pub fn is_full(&self) -> bool {
         self.len() == FIRECRACKER_MAX_QUEUE_SIZE
     }
 
--- a/src/vmm/src/devices/virtio/iov_deque.rs
+++ b/src/vmm/src/devices/virtio/iov_deque.rs
@@ -1,0 +1,449 @@
+// Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::os::fd::AsRawFd;
+
+use libc::{c_int, c_void, iovec, off_t, size_t};
+use memfd;
+
+use super::queue::FIRECRACKER_MAX_QUEUE_SIZE;
+use crate::arch::PAGE_SIZE;
+
+#[derive(Debug, thiserror::Error, displaydoc::Display)]
+pub enum IovDequeError {
+    /// Error with memfd: {0}
+    Memfd(#[from] memfd::Error),
+    /// Error while resizing memfd: {0}
+    MemfdResize(std::io::Error),
+    /// Error calling mmap: {0}
+    Mmap(std::io::Error),
+}
+
+/// ['IovDeque'] is a ring buffer tailored for `struct iovec` objects.
+///
+/// From the point of view of API, [`IovDeque`] is a typical ring buffer that allows us to push
+/// `struct iovec` objects at the end of the buffer and pop them from its beginning.
+///
+/// It is tailored to store `struct iovec` objects that described memory that was passed to us from
+/// the guest via a VirtIO queue. This allows us to assume the maximum size of a ring buffer (the
+/// negotiated size of the queue).
+// An important feature of the data structure is that it can give us a slice of all `struct iovec`
+// objects in the queue, so that we can use this `&mut [iovec]` to perform operations such as
+// `readv`. A typical implementation of a ring buffer allows for entries to wrap around the end of
+// the underlying buffer. For example, a ring buffer with a capacity of 10 elements which
+// currently holds 4 elements can look like this:
+//
+//                      tail                        head
+//                       |                           |
+//                       v                           v
+//                 +---+---+---+---+---+---+---+---+---+---+
+// ring buffer:    | C | D |   |   |   |   |   |   | A | B |
+//                 +---+---+---+---+---+---+---+---+---+---+
+//
+// When getting a slice for this data we should get something like that: &[A, B, C, D], which
+// would require copies in order to make the elements continuous in memory.
+//
+// In order to avoid that and make the operation of getting a slice more efficient, we implement
+// the optimization described in the "Optimization" section of the "Circular buffer" wikipedia
+// entry: https://en.wikipedia.org/wiki/Circular_buffer. The optimization consists of allocating
+// double the size of the virtual memory required for the buffer and map both parts on the same
+// physical address. Looking at the same example as before, we should get, this picture:
+//
+//                                    head   |    tail
+//                                     |     |     |
+//                                     v     |     v
+//   +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+//   | C | D |   |   |   |   |   |   | A | B | C | D |   |   |   |   |   |   | A | B |
+//   +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+//            First virtual page             |       Second virtual page
+//                                           |
+//                                           |
+//
+//                                     Virtual memory
+// ---------------------------------------------------------------------------------------
+//                                    Physical memory
+//
+//                      +---+---+---+---+---+---+---+---+---+---+
+//                      | C | D |   |   |   |   |   |   | A | B |
+//                      +---+---+---+---+---+---+---+---+---+---+
+//
+// Like that, the elements stored in the buffer are always laid out in contiguous virtual memory,
+// so making a slice out of them does not require any copies.
+#[derive(Debug)]
+pub struct IovDeque {
+    iov: *mut libc::iovec,
+    start: u16,
+    len: u16,
+}
+
+// SAFETY: This is `Send`. We hold sole ownership of the underlying buffer.
+unsafe impl Send for IovDeque {}
+
+impl IovDeque {
+    /// Create a [`memfd`] object that represents a single physical page
+    fn create_memfd() -> Result<memfd::Memfd, IovDequeError> {
+        // Create a sealable memfd.
+        let opts = memfd::MemfdOptions::default().allow_sealing(true);
+        let mfd = opts.create("sized-1K")?;
+
+        // Resize to system page size.
+        mfd.as_file()
+            .set_len(PAGE_SIZE.try_into().unwrap())
+            .map_err(IovDequeError::MemfdResize)?;
+
+        // Add seals to prevent further resizing.
+        mfd.add_seals(&[memfd::FileSeal::SealShrink, memfd::FileSeal::SealGrow])?;
+
+        // Prevent further sealing changes.
+        mfd.add_seal(memfd::FileSeal::SealSeal)?;
+
+        Ok(mfd)
+    }
+
+    /// A safe wrapper on top of libc's `mmap` system call
+    ///
+    /// # Safety: Callers need to make sure that the arguments to `mmap` are valid
+    unsafe fn mmap(
+        addr: *mut c_void,
+        len: size_t,
+        prot: c_int,
+        flags: c_int,
+        fd: c_int,
+        offset: off_t,
+    ) -> Result<*mut c_void, IovDequeError> {
+        let ptr = libc::mmap(addr, len, prot, flags, fd, offset);
+        if ptr == libc::MAP_FAILED {
+            return Err(IovDequeError::Mmap(std::io::Error::last_os_error()));
+        }
+
+        Ok(ptr)
+    }
+
+    /// Allocate memory for our ring buffer
+    ///
+    /// This will allocate exactly two pages of virtual memory. In order to implement the
+    /// optimization that allows us to always have elements in contiguous memory we need
+    /// allocations at the granularity of `PAGE_SIZE`. Now, our queues are at maximum 256
+    /// descriptors long and `struct iovec` looks like this:
+    ///
+    /// ```Rust
+    /// pub struct iovec {
+    ///    pub iov_base: *mut ::c_void,
+    ///    pub iov_len: ::size_t,
+    /// }
+    /// ```
+    ///
+    /// so, it's 16 bytes long. As a result, we need a single page for holding the actual data of
+    /// our buffer.
+    fn allocate_ring_buffer_memory() -> Result<*mut c_void, IovDequeError> {
+        // The fact that we allocate two pages is due to the size of `struct iovec` times our queue
+        // size equals the page size. Add here a debug assertion to reflect that and ensure that we
+        // will adapt our logic if the assumption changes in the future.
+        const {
+            assert!(
+                std::mem::size_of::<iovec>() * FIRECRACKER_MAX_QUEUE_SIZE as usize == PAGE_SIZE
+            );
+        }
+
+        // SAFETY: We are calling the system call with valid arguments
+        unsafe {
+            Self::mmap(
+                std::ptr::null_mut(),
+                PAGE_SIZE * 2,
+                libc::PROT_NONE,
+                libc::MAP_PRIVATE | libc::MAP_ANONYMOUS,
+                -1,
+                0,
+            )
+        }
+    }
+
+    /// Create a new [`IovDeque`] that can hold memory described by a single VirtIO queue.
+    pub fn new() -> Result<Self, IovDequeError> {
+        let memfd = Self::create_memfd()?;
+        let raw_memfd = memfd.as_file().as_raw_fd();
+        let buffer = Self::allocate_ring_buffer_memory()?;
+
+        // Map the first page of virtual memory to the physical page described by the memfd object
+        // SAFETY: We are calling the system call with valid arguments
+        let _ = unsafe {
+            Self::mmap(
+                buffer,
+                PAGE_SIZE,
+                libc::PROT_READ | libc::PROT_WRITE,
+                libc::MAP_SHARED | libc::MAP_FIXED,
+                raw_memfd,
+                0,
+            )
+        }?;
+
+        // Map the second page of virtual memory to the physical page described by the memfd object
+        //
+        // SAFETY: This is safe because:
+        // * Both `buffer` and the result of `buffer.add(PAGE_SIZE)` are within bounds of the
+        //   allocation we got from `Self::allocate_ring_buffer_memory`.
+        // * The computed offset is `PAGE_SIZE * size_of::<c_void>() == PAGE_SIZE bytes` which fits
+        //   in `isize`
+        // * The resulting pointer is the beginning of the second page of our allocation, so it
+        //   doesn't wrap around the address space.
+        let next_page = unsafe { buffer.add(PAGE_SIZE) };
+
+        // SAFETY: We are calling the system call with valid arguments
+        let _ = unsafe {
+            Self::mmap(
+                next_page,
+                PAGE_SIZE,
+                libc::PROT_READ | libc::PROT_WRITE,
+                libc::MAP_SHARED | libc::MAP_FIXED,
+                raw_memfd,
+                0,
+            )
+        }?;
+
+        Ok(Self {
+            iov: buffer.cast(),
+            start: 0,
+            len: 0,
+        })
+    }
+
+    /// Returns the number of `iovec` objects currently in the [`IovDeque`]
+    #[inline(always)]
+    pub fn len(&self) -> u16 {
+        self.len
+    }
+
+    /// Returns `true` if the [`IovDeque`] is full, `false` otherwise
+    #[inline(always)]
+    fn is_full(&self) -> bool {
+        self.len() == FIRECRACKER_MAX_QUEUE_SIZE
+    }
+
+    /// Resets the queue, dropping all its elements.
+    #[inline(always)]
+    pub fn clear(&mut self) {
+        self.start = 0;
+        self.len = 0;
+    }
+
+    /// Adds an `iovec` in the ring buffer.
+    ///
+    /// Returns an `IovDequeError::Full` error if the buffer is full.
+    pub fn push_back(&mut self, iov: iovec) {
+        // This should NEVER happen, since our ring buffer is as big as the maximum queue size.
+        // We also check for the sanity of the VirtIO queues, in queue.rs, which means that if we
+        // ever try to add something in a full ring buffer, there is an internal bug in the device
+        // emulation logic. Panic here because the device is hopelessly broken.
+        assert!(
+            !self.is_full(),
+            "The number of `iovec` objects is bigger than the available space"
+        );
+
+        // SAFETY: self.iov is a valid pointer and `self.start + self.len` is within range (we
+        // asserted before that the buffer is not full).
+        unsafe {
+            self.iov
+                .add((self.start + self.len) as usize)
+                .write_volatile(iov)
+        };
+        self.len += 1;
+    }
+
+    /// Pops the first `nr_iovecs` iovecs from the buffer.
+    ///
+    /// Returns the total number of bytes of all the popped iovecs. This will panic if we are asked
+    /// to pop more iovecs than what is currently available in the buffer.
+    pub fn pop_front(&mut self, nr_iovecs: u16) {
+        assert!(
+            self.len() >= nr_iovecs,
+            "Internal bug! Trying to drop more iovec objects than what is available"
+        );
+
+        self.start += nr_iovecs;
+        self.len -= nr_iovecs;
+        if self.start >= FIRECRACKER_MAX_QUEUE_SIZE {
+            self.start -= FIRECRACKER_MAX_QUEUE_SIZE;
+        }
+    }
+
+    /// Get a slice of the iovec objects currently in the buffer.
+    pub fn as_slice(&self) -> &[iovec] {
+        // SAFETY: Here we create a slice out of the existing elements in the buffer (not the whole
+        // allocated memory). That means that we can:
+        // * We can read `self.len * mem::size_of::<iovec>()` bytes out of the memory range we are
+        //   returning.
+        // * `self.iov.add(self.start.into())` is a non-null pointer and aligned.
+        // * The underlying memory comes from a single allocation.
+        // * The returning pointer points to `self.len` consecutive initialized `iovec` objects.
+        // * We are only accessing the underlying memory through the returned slice. Since we are
+        //   returning a slice of only the existing pushed elements the slice does not contain any
+        //   aliasing references.
+        // * The slice can be up to 1 page long which is smaller than `isize::MAX`.
+        unsafe {
+            let slice_start = self.iov.add(self.start.into());
+            std::slice::from_raw_parts(slice_start, self.len.into())
+        }
+    }
+
+    /// Get a mutable slice of the iovec objects currently in the buffer.
+    pub fn as_mut_slice(&mut self) -> &mut [iovec] {
+        // SAFETY: Here we create a slice out of the existing elements in the buffer (not the whole
+        // allocated memory). That means that we can:
+        // * We can read/write `self.len * mem::size_of::<iovec>()` bytes out of the memory range we
+        //   are returning.
+        // * The underlying memory comes from a single allocation.
+        // * `self.iov.add(self.start.into())` is a non-null pointer and aligned
+        // * The returning pointer points to `self.len` consecutive initialized `iovec` objects.
+        // * We are only accessing the underlying memory through the returned slice. Since we are
+        //   returning a slice of only the existing pushed elements the slice does not contain any
+        //   aliasing references.
+        // * The slice can be up to 1 page long which is smaller than `isize::MAX`.
+        unsafe {
+            let slice_start = self.iov.add(self.start.into());
+            std::slice::from_raw_parts_mut(slice_start, self.len.into())
+        }
+    }
+}
+
+impl Drop for IovDeque {
+    fn drop(&mut self) {
+        // SAFETY: We are passing an address that we got from a previous allocation of `2 *
+        // PAGE_SIZE` bytes by calling mmap
+        let _ = unsafe { libc::munmap(self.iov.cast(), PAGE_SIZE * 2) };
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use libc::iovec;
+
+    use super::IovDeque;
+
+    #[test]
+    fn test_new() {
+        let deque = IovDeque::new().unwrap();
+        assert_eq!(deque.len(), 0);
+    }
+
+    fn make_iovec(id: u16, len: u16) -> iovec {
+        iovec {
+            iov_base: id as *mut libc::c_void,
+            iov_len: len as usize,
+        }
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_push_back_too_many() {
+        let mut deque = IovDeque::new().unwrap();
+        assert_eq!(deque.len(), 0);
+
+        for i in 0u16..256 {
+            deque.push_back(make_iovec(i, i));
+            assert_eq!(deque.len(), i + 1);
+        }
+
+        deque.push_back(make_iovec(0, 0));
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_pop_front_from_empty() {
+        let mut deque = IovDeque::new().unwrap();
+        deque.pop_front(1);
+    }
+
+    #[test]
+    #[should_panic]
+    fn test_pop_front_too_many() {
+        let mut deque = IovDeque::new().unwrap();
+        deque.push_back(make_iovec(42, 42));
+        deque.pop_front(2);
+    }
+
+    #[test]
+    fn test_pop() {
+        let mut deque = IovDeque::new().unwrap();
+        assert_eq!(deque.len(), 0);
+        assert!(!deque.is_full());
+        deque.pop_front(0);
+
+        for i in 0u16..256 {
+            deque.push_back(make_iovec(i, i));
+            assert_eq!(deque.len(), i + 1);
+        }
+
+        assert!(deque.is_full());
+        assert!(deque.len() != 0);
+
+        for i in 0u16..256 {
+            deque.pop_front(1);
+            assert_eq!(deque.len(), 256 - i - 1);
+        }
+    }
+
+    #[test]
+    fn test_pop_many() {
+        let mut deque = IovDeque::new().unwrap();
+
+        for i in 0u16..256 {
+            deque.push_back(make_iovec(i, i));
+        }
+
+        deque.pop_front(1);
+        assert_eq!(deque.len(), 255);
+        deque.pop_front(2);
+        assert_eq!(deque.len(), 253);
+        deque.pop_front(4);
+        assert_eq!(deque.len(), 249);
+        deque.pop_front(8);
+        assert_eq!(deque.len(), 241);
+        deque.pop_front(16);
+        assert_eq!(deque.len(), 225);
+        deque.pop_front(32);
+        assert_eq!(deque.len(), 193);
+        deque.pop_front(64);
+        assert_eq!(deque.len(), 129);
+        deque.pop_front(128);
+        assert_eq!(deque.len(), 1);
+    }
+
+    #[test]
+    fn test_as_slice() {
+        let mut deque = IovDeque::new().unwrap();
+        assert!(deque.as_slice().is_empty());
+
+        for i in 0..256 {
+            deque.push_back(make_iovec(i, 100));
+            assert_eq!(deque.as_slice().len(), (i + 1) as usize);
+        }
+        let copy: Vec<iovec> = deque.as_slice().to_vec();
+
+        assert_eq!(copy.len(), deque.len() as usize);
+        for (i, iov) in deque.as_slice().iter().enumerate() {
+            assert_eq!(iov.iov_len, copy[i].iov_len);
+        }
+    }
+
+    #[test]
+    fn test_as_mut_slice() {
+        let mut deque = IovDeque::new().unwrap();
+        assert!(deque.as_mut_slice().is_empty());
+
+        for i in 0..256 {
+            deque.push_back(make_iovec(i, 100));
+            assert_eq!(deque.as_mut_slice().len(), (i + 1) as usize);
+        }
+
+        let copy: Vec<iovec> = deque.as_mut_slice().to_vec();
+        deque
+            .as_mut_slice()
+            .iter_mut()
+            .for_each(|iov| iov.iov_len *= 2);
+
+        assert_eq!(copy.len(), deque.len() as usize);
+        for (i, iov) in deque.as_slice().iter().enumerate() {
+            assert_eq!(iov.iov_len, 2 * copy[i].iov_len);
+        }
+    }
+}
--- a/src/vmm/src/devices/virtio/iovec.rs
+++ b/src/vmm/src/devices/virtio/iovec.rs
@@ -4,6 +4,7 @@
 use std::io::ErrorKind;
 
 use libc::{c_void, iovec, size_t};
+use serde::{Deserialize, Serialize};
 #[cfg(not(kani))]
 use smallvec::SmallVec;
 use vm_memory::bitmap::Bitmap;
@@ -218,7 +219,7 @@ impl IoVecBuffer {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ParsedDescriptorChain {
     pub head_index: u16,
     pub length: u32,
@@ -233,9 +234,9 @@ pub struct ParsedDescriptorChain {
 #[derive(Debug)]
 pub struct IoVecBufferMut {
     // container of the memory regions included in this IO vector
-    vecs: IovDeque,
+    pub vecs: IovDeque,
     // Total length of the IoVecBufferMut
-    len: usize,
+    pub len: usize,
 }
 
 impl IoVecBufferMut {
--- a/src/vmm/src/devices/virtio/iovec.rs
+++ b/src/vmm/src/devices/virtio/iovec.rs
@@ -4,12 +4,14 @@
 use std::io::ErrorKind;
 
 use libc::{c_void, iovec, size_t};
+#[cfg(not(kani))]
 use smallvec::SmallVec;
 use vm_memory::bitmap::Bitmap;
 use vm_memory::{
     GuestMemory, GuestMemoryError, ReadVolatile, VolatileMemoryError, VolatileSlice, WriteVolatile,
 };
 
+use super::iov_deque::{IovDeque, IovDequeError};
 use crate::devices::virtio::queue::DescriptorChain;
 use crate::vstate::memory::GuestMemoryMmap;
 
@@ -23,6 +25,8 @@ pub enum IoVecError {
     OverflowedDescriptor,
     /// Guest memory error: {0}
     GuestMemory(#[from] GuestMemoryError),
+    /// Error with underlying `IovDeque`: {0}
+    IovDeque(#[from] IovDequeError),
 }
 
 // Using SmallVec in the kani proofs causes kani to use unbounded amounts of memory
@@ -214,21 +218,100 @@ impl IoVecBuffer {
     }
 }
 
+#[derive(Debug)]
+pub struct ParsedDescriptorChain {
+    pub head_index: u16,
+    pub length: u32,
+    pub nr_iovecs: u16,
+}
+
 /// This is essentially a wrapper of a `Vec<libc::iovec>` which can be passed to `libc::readv`.
 ///
 /// It describes a write-only buffer passed to us by the guest that is scattered across multiple
 /// memory regions. Additionally, this wrapper provides methods that allow reading arbitrary ranges
 /// of data from that buffer.
-#[derive(Debug, Default, Clone)]
+#[derive(Debug)]
 pub struct IoVecBufferMut {
     // container of the memory regions included in this IO vector
-    vecs: IoVecVec,
+    vecs: IovDeque,
     // Total length of the IoVecBufferMut
-    len: u32,
+    len: usize,
 }
 
 impl IoVecBufferMut {
-    /// Create an `IoVecBuffer` from a `DescriptorChain`
+    /// Parse a `DescriptorChain` object and append the memory regions it describes in the
+    /// underlying ring buffer.
+    ///
+    /// # Safety
+    ///
+    /// The descriptor chain cannot be referencing the same memory location as another chain
+    unsafe fn parse_descriptor(
+        &mut self,
+        mem: &GuestMemoryMmap,
+        head: DescriptorChain,
+    ) -> Result<ParsedDescriptorChain, IoVecError> {
+        let head_index = head.index;
+        let mut next_descriptor = Some(head);
+        let mut length = 0u32;
+        let mut nr_iovecs = 0u16;
+        while let Some(desc) = next_descriptor {
+            if !desc.is_write_only() {
+                self.vecs.pop_front(nr_iovecs);
+                return Err(IoVecError::ReadOnlyDescriptor);
+            }
+
+            // We use get_slice instead of `get_host_address` here in order to have the whole
+            // range of the descriptor chain checked, i.e. [addr, addr + len) is a valid memory
+            // region in the GuestMemoryMmap.
+            let slice = mem.get_slice(desc.addr, desc.len as usize).map_err(|err| {
+                self.vecs.pop_front(nr_iovecs);
+                err
+            })?;
+
+            // We need to mark the area of guest memory that will be mutated through this
+            // IoVecBufferMut as dirty ahead of time, as we loose access to all
+            // vm-memory related information after converting down to iovecs.
+            slice.bitmap().mark_dirty(0, desc.len as usize);
+
+            let iov_base = slice.ptr_guard_mut().as_ptr().cast::<c_void>();
+            self.vecs.push_back(iovec {
+                iov_base,
+                iov_len: desc.len as size_t,
+            });
+            nr_iovecs += 1;
+            length = length
+                .checked_add(desc.len)
+                .ok_or(IoVecError::OverflowedDescriptor)
+                .map_err(|err| {
+                    self.vecs.pop_front(nr_iovecs);
+                    err
+                })?;
+
+            next_descriptor = desc.next_descriptor();
+        }
+
+        self.len = self.len.checked_add(length as usize).ok_or_else(|| {
+            self.vecs.pop_front(nr_iovecs);
+            IoVecError::OverflowedDescriptor
+        })?;
+
+        Ok(ParsedDescriptorChain {
+            head_index,
+            length,
+            nr_iovecs,
+        })
+    }
+
+    /// Create an empty `IoVecBufferMut`.
+    pub fn new() -> Result<Self, IovDequeError> {
+        let vecs = IovDeque::new()?;
+        Ok(Self { vecs, len: 0 })
+    }
+
+    /// Create an `IoVecBufferMut` from a `DescriptorChain`
+    ///
+    /// This will clear any previous `iovec` objects in the buffer and load the new
+    /// [`DescriptorChain`].
     ///
     /// # Safety
     ///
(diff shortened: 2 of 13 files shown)


src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -0,0 +1,412 @@
...
+    fn mmap(
+        addr: *mut c_void,
+        len: size_t,
+        prot: c_int,
+        flags: c_int,
+        fd: c_int,
+        offset: off_t,
+    ) -> Result<*mut c_void, IovDequeError> {
+        // SAFETY: We are calling the system call with valid arguments and properly checking its
+        // return value
```
roypat, line 114: To be able to claim that we are calling the syscall with valid arguments, this function should be `unsafe`, with the invariant being "arguments are valid for `mmap`"

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -0,0 +1,412 @@
...
+        // * `buffer` is valid both for reads and writes (allocated with `libc::PROT_READ |
+        //   libc::PROT_WRITE`. `
+        // * `buffer` is aligned at `PAGE_SIZE`
+        // * `buffer` points to memory allocated with a single system call to `libc::mmap`
+        let iov = unsafe {
+            std::slice::from_raw_parts_mut(
+                buffer.cast(),
+                2 * PAGE_SIZE / std::mem::size_of::<libc::iovec>(),
+            )
+        };
```
roypat, line 197: Mhhh, I don't think we can do this. The two pages contain the same physical memory, so we're creating aliasing mutable references here. But if we just store everything as `*mut libc::iovec`, I think it'll be fine :)

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -0,0 +1,412 @@
...
+        if self.is_empty() {
+            return None;
+        }
+
+        let iov = self.iov[self.head];
+        self.head += 1;
+        if self.head > usize::from(FIRECRACKER_MAX_QUEUE_SIZE) {
+            self.head -= usize::from(FIRECRACKER_MAX_QUEUE_SIZE);
+            self.tail -= usize::from(FIRECRACKER_MAX_QUEUE_SIZE);
+        }
```
roypat, line 260: I wonder if we could do the trick that the virtio spec does here, and simply store these indices as `Wrapping<usize>`, let them count up indefinitely, and just do modulo when we want to access an element? Although I also don't really see _why_ we would want to do that :joy:

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -0,0 +1,412 @@
...
+
+        Some(iov)
+    }
+
+    /// Drop the first `iovec` objects that describe guest memory of `size` length.
+    ///
+    /// This will drop `iovec` buffers which describe a total memory size of up to `size`. This
```
roypat, line 267:
```suggestion
    /// This will drop `iovec` buffers which describe a total memory size of at least `size`. This
```
no? Because if `size == 1` and the first iovec contains 10 bytes, then it drop iovecs describing 10 bytes

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -0,0 +1,412 @@
...
+            }
+        }
+
+        dropped
+    }
+
+    /// Gets a mutable slice of the `iovec` objects currently in the buffer.
+    pub(crate) fn as_mut_slice(&mut self) -> &mut [iovec] {
+        &mut self.iov[self.head..self.tail]
+    }
```
roypat, line 286: Also need to work with pointers here

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -257,18 +257,78 @@ impl IoVecBufferMut {
...
+        self.parse_descriptor(mem, head)
+    }
+
+    /// Drop memory from the `IoVecBufferMut`
+    ///
+    /// This will drop memory described by the `IoVecBufferMut` starting from the beginning.
+    pub fn drop_iovecs(&mut self, size: u32) -> Result<(), IoVecError> {
+        let dropped = self.vecs.drop_iovs(size as usize);
+
+        // Users should ask us to drop a `size` of memory that is not exactly covered by `iovec`
```
roypat, line 322: nit: s/should/should not ?

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -257,18 +257,78 @@ impl IoVecBufferMut {
...
+        // objects. In other words, the sum of the lengths of all dropped `iovec` objects should be
+        // equal to the `size` we were asked to drop. If it isn't, something is seriously wrong
+        // with the VirtIO queue or the emulation logic, so fail at this point.
+        assert_eq!(u32::try_from(dropped).unwrap(), size);
```
roypat, line 326:
```suggestion
        assert_eq!(dropped, usize::from(size));
```

do we really want to panic here though? The function returns `Result`, so could we just return an `Err`?

src/vmm/src/devices/virtio/iov_deque.rs
roypat: And if we use pointers, we also get rid of the lifetime parameter (which is kinda wrong to have anyway, because the `iov` field is self referential, and that's not something rust's lifetimes can express)

src/vmm/src/devices/virtio/rng/device.rs
```diff
@@ -112,15 +112,15 @@ impl Entropy {
...
+        let mut rand_bytes = vec![0; iovec.len()];
         rand::fill(&mut rand_bytes).map_err(|err| {
             METRICS.host_rng_fails.inc();
             err
         })?;
 
         // It is ok to unwrap here. We are writing `iovec.len()` bytes at offset 0.
         iovec.write_all_volatile_at(&rand_bytes, 0).unwrap();
-        Ok(iovec.len())
+        Ok(u32::try_from(iovec.len()).unwrap())
```
roypat, line 123: this makes me sad :(

I don't think its strictly necessary, but it'd be cool to differentiate between the new "multi descriptor chain" iovecbuffermuts and the old "one descriptor chain" ones. We could add a `L: Add` type parameter that is either `u32` or `usize` and specifies the length of the `IoVecBufferMut`. Then only the `impl IoVecBufferMut<usize>` block would have all the appending stuff, and in the rng device and everywhere else can continue using `IoVecBufferMut<u32>` without these ugly `try_froms`.

src/vmm/src/devices/virtio/rng/device.rs
```diff
@@ -145,7 +145,10 @@ impl Entropy {
                     // Check for available rate limiting budget.
                     // If not enough budget is available, leave the request descriptor in the queue
                     // to handle once we do have budget.
-                    if !Self::rate_limit_request(&mut self.rate_limiter, u64::from(iovec.len())) {
+                    if !Self::rate_limit_request(
+                        &mut self.rate_limiter,
+                        u64::try_from(iovec.len()).unwrap(),
```
roypat, line 150: `usize_to_u64` here

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -517,17 +547,33 @@ impl Net {
...
+            .mark_used(len.try_into().unwrap(), &self.metrics);
+        Ok(len)
     }
 
+    /// Read as many frames as possible.
     fn process_rx(&mut self) -> Result<(), DeviceError> {
-        // Read as many frames as possible.
+        if self.rx_buffer.iovec.len() == 0 {
+            self.parse_rx_descriptors();
+        }
```
ShadowCurse, line 569: We can always call `self.parse_rx_descriptors();`. This is way there is a chance for us to read more packets (as we can parse additional chains). Otherwise we always rely on `RX` event handling to do preprocessing.

src/vmm/src/devices/virtio/net/tap.rs
```diff
@@ -190,6 +190,20 @@ impl Tap {
         }
         Ok(usize::try_from(ret).unwrap())
     }
+
+    /// Read from tap to an `IoVecBufferMut`
+    pub(crate) fn read_iovec(&mut self, buffer: &mut IoVecBufferMut) -> Result<usize, IoError> {
+        let iovcnt = buffer.iovec_count() as i32;
+        let iov = buffer.as_iovec_ptr();
```
ShadowCurse, line 197: Here is the issue. When we want to read from TAP, we want to read onto a single chain, but here you give it all `iovec`s, so `readv` can write to all the chains it wants.
To fix this we just need smth like this: https://github.com/firecracker-microvm/firecracker/pull/4813/files#diff-205f89a698c2fafa14dcbd3c76dfee31dd597294c8334bc6b82d8c92814a497fR108

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -102,6 +89,113 @@ pub struct ConfigSpace {
...
+        if size > parsed_dc.length {
+            parsed_dc.length = 0;
+            metrics.rx_fails.inc();
+        } else {
+            parsed_dc.length = size;
+            metrics.rx_bytes_count.add(size as u64);
+            metrics.rx_packets_count.inc();
+        }
+        self.iovec.drop_descriptor_chain(&parsed_dc);
+        self.deferred_descriptor = Some(parsed_dc);
```
ShadowCurse, line 176: I would prefer if we would write used descriptor into chain right here, but do actual notification in `finish_frame` (like it is done here https://github.com/firecracker-microvm/firecracker/pull/4813/files#diff-205f89a698c2fafa14dcbd3c76dfee31dd597294c8334bc6b82d8c92814a497fR226)
This way it will require less changes for MRG_RXBUF support and it will reduce the size of the `RxBuffers` struct as we will need to store `u16` instead of `Option<ParsedDescriptorChain>`.

src/vmm/src/devices/virtio/iov_deque.rs
bchalios: I see what you mean, but on the other hand is there a value that we can pass here as an argument that causes the call to the system call to break any safety guarantees? IIUC, if we pass "bad" arguments this will fail and we check for the failure.

src/vmm/src/devices/virtio/iov_deque.rs
bchalios: That was a good catch. I updated the code to only deal with pointers.

src/vmm/src/devices/virtio/iov_deque.rs
bchalios: Yes. we could do that. If you see a strong reason why we should, I 'm happy to do it.

src/vmm/src/devices/virtio/iov_deque.rs
bchalios: I changed this one. Now I'm dropping by number of descriptors rather than "total size of covered bytes"

src/vmm/src/devices/virtio/iov_deque.rs
roypat: Nah, I don't really see a reason, was more of a random thought. Feel free to resolve :)

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -102,6 +89,113 @@ pub struct ConfigSpace {
...
+        // descriptor back to the guest without claiming any written bytes
+        if size > parsed_dc.length {
+            parsed_dc.length = 0;
+            metrics.rx_fails.inc();
+        } else {
+            parsed_dc.length = size;
+            metrics.rx_bytes_count.add(size as u64);
+            metrics.rx_packets_count.inc();
+        }
+        self.iovec.drop_descriptor_chain(&parsed_dc);
```
ShadowCurse, line 175: Can you also add an update of the packet header with correct number of buffers like it is done here: https://github.com/firecracker-microvm/firecracker/pull/4813/files#diff-205f89a698c2fafa14dcbd3c76dfee31dd597294c8334bc6b82d8c92814a497fR260
This is part of the spec and without MRG_RXBUF the `num_buffers` should always be 1.

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -214,64 +218,131 @@ impl IoVecBuffer {
     }
 }
 
+#[derive(Debug)]
+pub struct ParsedDescriptorChain {
+    pub head_index: u16,
+    pub length: u32,
+    pub nr_iovecs: u16,
```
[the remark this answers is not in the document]
bchalios: number

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -0,0 +1,401 @@
...
+//                      +---+---+---+---+---+---+---+---+---+---+
+//
+// Like that, the elements stored in the buffer are always laid out in contiguous virtual memory,
+// so making a slice out of them does not require any copies.
+#[derive(Debug)]
+pub struct IovDeque {
+    iov: *mut libc::iovec,
+    start: u16,
+    len: u16,
+}
```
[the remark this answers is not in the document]
bchalios: We could. What's that useful for though?

src/vmm/src/devices/virtio/iov_deque.rs
roypat: MAP_FIXED could probably break quite a few things :D

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -102,6 +89,113 @@ pub struct ConfigSpace {
...
+    Parsing(#[from] IoVecError),
+    /// RX buffer is too small
+    BufferTooSmall,
+}
+
+/// A map of all the memory the guest has provided us with for performing RX
+#[derive(Debug)]
+pub(crate) struct RxBuffers {
+    // minimum size of a usable buffer for doing RX
+    min_buffer_size: u32,
```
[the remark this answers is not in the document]
bchalios: > But in that case nothing critically bad happens

If we make sure that the guest is VirtIO spec compliant, this will never happen. Once we parse a descriptor that is not of the correct size will we discard it. So we will avoid re-transmissions.

src/vmm/src/devices/virtio/net/device.rs
bchalios: I want to avoid parsing descriptors in the hot path, when possible.

In my previous attempt to implement `readv` parsing descriptors the moment we want to actually read from the TAP was leading to performance degradation. It is true that parsing is faster now but chances are that we will see _some_ degradation again.

Typically, the guest provides us with 13 descriptors. So imagine a scenario where we currently have 12 parsed descriptors in the queue. This means that we can receive up to 12 frames and pay the cost of parsing only for the 13th one. If we always parse, we will be paying the cost of parsing for all 13 frames.

src/vmm/src/devices/virtio/net/tap.rs
bchalios: Sure, but this won't happen if we make sure that we only keep the descriptor chains that are of the appropriate size. However, I like the extra guarantee, so I'll do this.

src/vmm/src/devices/virtio/iov_deque.rs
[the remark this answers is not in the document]
ShadowCurse: In this case we can remove `len` from `IoVecBufferMut`. And the whole `IoVecBufferMut` will become just a type alias to `IovDeque` with some additional methods.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: This parsing will only happens once per TAP event. So in your example we will parse 13th descriptor here, but then we go into actual loop that reads from the TAP, so we will potentially read all 13 packets. Otherwise we will only be able to read 12 packets (from your example only 12 desc are parsed) because buffer will not be empty.

bchalios: > Otherwise we will only be able to read 12 packets (from your example only 12 desc are parsed) because buffer will not be empty.

I get that. What I'm saying is that I would like us to avoid reading if we don't need to. I will another (hopefully better) example. If we currently have 1 descriptor chain parsed and there is a single frame to read from the tap device, we will pay the cost of parsing 12 descriptors for reading a single packet.

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
bchalios: > I just think this logic is an additional complexity which will never trigger if guest does everything correct

This complexity you will need to add it with mergeable buffers. You need to make sure that the guest provides buffers of at least the virtio header size. It _might_ never trigger with a correctly operating guest. But we do not assume correctly operating guests.

> , it should be expected that something is going wrong and retransmission is acceptable.

I agree. With my logic, what goes wrong is that the buffer is discarded. This way we have much better visibility on what went actually wrong. In the other way around, the event that tells us that something went wrong is external to Firecracker (TCP re-transmissions). I would like to keep it that way.

src/vmm/src/devices/virtio/iov_deque.rs
bchalios: Ok, I am convinced.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: yeah, this is valid. But I think if TAP has small amount of packets to read, this means network is not under high load, so doing parsing here will not cause major delay. Last argument I have: we will need to always parse for mrg_ rxbuf support, otherwise there will be a lot of retransmissions.

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
ShadowCurse: > This complexity you will need to add it with mergeable buffers.

In mrg_rxbuf case we only need a head descriptor to point to a buffer of at least 12 bytes (or whatever the size of the header struct). And we can just assert this.

> But we do not assume correctly operating guests.

Yes, we do not assume that. But at the same time we don't need to check this assumption. Our job is to create a system that works regardless of buffer sizes. If guest does smth that causes retransmission, it is not our problem and we should not try to prevent it on our side.

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -214,64 +218,131 @@ impl IoVecBuffer {
...
+
+        Ok(ParsedDescriptorChain {
+            head_index,
+            length,
+            nr_iovecs,
+        })
+    }
+
+    /// Create an empty `IoVecBufferMut`.
+    pub(crate) fn new() -> Result<Self, IovDequeError> {
```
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/devices/virtio/net/device.rs
bchalios: I am not sure I understand what you mean here. We need the number of the bytes that were actually written as well, right?

ShadowCurse: You have `size` argument which is the number of bytes written into chain. Currently you do this: `parsed_dc.length = size;` and save the `parsed_dc`. But you can use this `size` argument to write a used descriptor into queue at this point. And in the `finish_frame` just bump queue by 1.

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -342,11 +427,14 @@ impl IoVecBufferMut {
...
 
-        for iov in &self.vecs {
+        for i in 0..self.vecs.len() {
             if len == 0 {
                 break;
             }
 
+            // It's ok to unwrap here because `i` is within range.
+            let iov = self.vecs.peek(i).unwrap();
+
```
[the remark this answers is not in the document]
ShadowCurse: I can still see:
```Rust
        for i in 0..self.vecs.len() {
            if len == 0 {
                break;
            }

            let iov = self.vecs.as_slice()[i as usize];
```

src/vmm/src/devices/virtio/net/device.rs
bchalios: Ok, I see now. You want to break down `add_used` in two parts. What do we gain with that?

src/vmm/src/devices/virtio/iovec.rs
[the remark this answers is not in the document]
bchalios: oh god. Sorry, I missed the outer loop. Thanks. Will update.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: This:

>This way it will require less changes for MRG_RXBUF support and it will reduce the size of the RxBuffers struct as we will need to store u16 instead of Option<ParsedDescriptorChain>.

`MRG_RXBUF` will need to know all chains it used for packet. With current approach we will need `Vec<ParsedDescriptorChain>` to store them. But we can avoid this by splitting `add_used` into 2 parts and only store the number of heads we used in the `RxBuffers`.

src/vmm/src/devices/virtio/iov_deque.rs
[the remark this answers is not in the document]
bchalios: I think that knowledge for the amount of memory we are holding belongs to `IoVecBufferMut`. `IovDeque` is meant to be a container for `iovec` objects. It is in `IoVecBufferMut` that we have operations such as `write_volatile`, et al. In any case, I think this is a nit, so I'd rather keep things as are now.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -102,6 +89,128 @@ pub struct ConfigSpace {
...
+    fn header_set_num_buffers(&mut self, nr_descs: u16) {
+        // We can unwrap here, because we have checked before that the `IoVecBufferMut` holds at
+        // least one buffer with the proper size, depending on the feature negotiation. In any
+        // case, the buffer holds memory of at least `std::mem::size_of::<virtio_net_hdr_v1>()`
+        // bytes.
+        self.iovec
+            .write_all_volatile_at(
+                &nr_descs.to_le_bytes(),
+                std::mem::offset_of!(virtio_net_hdr_v1, num_buffers),
+            )
```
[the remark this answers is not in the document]
bchalios: If that is true we should add a check that the first descriptor is at least size_of(virtio_net_hdr_v1) in the emulation code. Then, this will be true. However, I didn't find in the spec this requirement. Could you point me to it?

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: Well.. in my opinion it is better than current version anyway, but ok, lest's leave this for mrg_rxbuf impl.

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
ShadowCurse: Ok, is seems I am incorrect here. I was thinking the size requirements were for actual buffers descriptors point to (so the minimum is size is for the header alone), but it seems it refers to the whole descriptor chain capacity.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -715,8 +777,10 @@ impl Net {
         self.tx_rate_limiter.update_buckets(tx_bytes, tx_ops);
     }
 
-    fn read_tap(&mut self) -> std::io::Result<usize> {
-        self.tap.read(&mut self.rx_frame_buf)
+    pub(crate) fn read_tap(&mut self) -> std::io::Result<usize> {
+        let nr_iovecs = self.rx_buffer.parsed_descriptors[0].nr_iovecs as usize;
```
ShadowCurse, line 781: We assume here that there are parsed descriptors. We do check that `read_from_mmds_or_tap` executes only with non empty `rx_buffers`. Maybe this should be `unsafe` with requirement of `rx_buffers` is not empty?
Or we can use `VecDeque::front` method to get optionally first element. And return some error if it is `None`.

src/vmm/src/devices/virtio/iov_deque.rs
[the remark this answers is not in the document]
ShadowCurse: Yes, `IovDeque` is designed to hold only `iovec`s, so it would make sense if it will know the total size of them it holds. We can still avoid adding `write_volatile` and other things with `IoVecBufferMut` by just wrapping `IovDeque`:
```Rust
struct IoVecBufferMut(IovDeque);
impl IoVecBufferMut {
...
```
Considering `IoVecBufferMut` will be used very frequently, I still think saving 4 bytes is worth it.

src/vmm/src/devices/virtio/net/device.rs
bchalios: Yes, I agree. I'll go with the second approach and `unwrap` on in `read_from_mmds_or_tap` since at this point we do check that `rx_buffers` is not empty

src/vmm/src/devices/virtio/net/device.rs
bchalios: Ok, I launched an A/B test to compare the version with conditional parsing against one we are always parsing, and I didn't see a noticeable difference: https://buildkite.com/firecracker/bchalios-a-b-testing/builds/102#_. So I will implement this, but I will do it inside `read_from_mmds_or_tap` so that I will only parse extra descriptors when I need them (`rx_buffer.len() == 0`). This way we can read from the TAP immediately and only spend time to parse descriptors for the frames we actually need them for.

src/vmm/src/devices/virtio/net/device.rs
bchalios: added this.

src/vmm/src/devices/virtio/net/device.rs
bchalios: I went with the `unsafe` part of this, so that I avoid having to add yet another Error type.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -562,6 +630,7 @@ impl Net {
     fn handle_deferred_frame(&mut self) -> Result<(), DeviceError> {
         if self.rate_limited_rx_single_frame() {
             self.rx_deferred_frame = false;
+            self.rx_buffer.deferred_descriptor = None;
```
ShadowCurse, line 633: I think this is incorrect. If we defer packet because of the rate limiter and then go into this function, we lose deferred descriptor and quest is never notified about it. 
I guess we just need to call `finish_frame` here.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -102,6 +89,128 @@ pub struct ConfigSpace {
...
+            .write_all_volatile_at(
+                &nr_descs.to_le_bytes(),
+                std::mem::offset_of!(virtio_net_hdr_v1, num_buffers),
+            )
+            .unwrap()
+    }
+
+    /// This will let the guest know that about all the `DescriptorChain` object that has been
+    /// used to receive a frame from the TAP.
+    fn finish_frame(&mut self, rx_queue: &mut Queue) -> bool {
```
ShadowCurse, line 196: Regarding returning `bool` here. The only way to return `false` is to error out on `add_used`. But `add_used` can only error out if `head_index` is out of range. And this is not possible as we got it from `DescriptorChain` which checks all indexes. So basically this function should never return `false`. So the whole return statement seems redundant.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -102,6 +89,128 @@ pub struct ConfigSpace {
...
+        // fundamentally broken.
+        let mut parsed_dc = self.parsed_descriptors.pop_front().expect(
+            "net: internal bug. Mismatch between written frame size and available descriptors",
+        );
+
+        // We are trying to mark as used more bytes than a single descriptor. However without
+        // VIRTIO_NET_F_MRG_RXBUF negotiated (we don't support it currently) the guest can't handle
+        // frames spanning multiple DescriptorChains. There is nothing we can do now, just pass the
+        // descriptor back to the guest without claiming any written bytes
+        if size > parsed_dc.length {
```
ShadowCurse, line 167: Is there a situation when `size` is bigger than `parsed_dc.length`? When we write for mmds, the `write_all_..` will error out if we try to write more than the chain has and `readv` will always return value smaller equal to the chain capacity.

ShadowCurse: Removing this `if` will also remove a need for `pop_front` inside the `drop_descriptor_chain` to count total capacity and it can be simplified to:
```Rust
    pub fn drop_descriptor_chain(&mut self, parse_descriptor: &ParsedDescriptorChain) {
        self.vecs.pop_front(parse_descriptor.nr_iovecs);
        self.len -= parse_descriptor.length as usize;
    }

```

src/vmm/src/devices/virtio/net/device.rs
roypat: `finish_frame` is called inside `rate_limited_rx_single_frame`, no?

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -102,6 +89,128 @@ pub struct ConfigSpace {
...
+            )
+            .unwrap()
+    }
+
+    /// This will let the guest know that about all the `DescriptorChain` object that has been
+    /// used to receive a frame from the TAP.
+    fn finish_frame(&mut self, rx_queue: &mut Queue) -> bool {
+        match self.deferred_descriptor.take() {
+            Some(used_dc) => {
+                if let Err(err) = rx_queue.add_used(used_dc.head_index, used_dc.length) {
```
roypat, line 199: Mh, I don't think this is completely right. The VirtIO spec, section 2.6.8 says

>  `len` [indicates] the total of bytes written into the buffer

so this shouldn't be `used_dc.length`, but `self.rx_bytes_read`, right?

roypat: Ah, I see now, in `mark_used` we adjust `used_dc.length` to match the number of bytes we wrote. Nice! I do like that

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -102,6 +89,128 @@ pub struct ConfigSpace {
...
+pub(crate) struct RxBuffers {
+    // minimum size of a usable buffer for doing RX
+    min_buffer_size: u32,
+    // An [`IoVecBufferMut`] covering all the memory we have available for receiving network
+    // frames.
+    iovec: IoVecBufferMut,
+    // A map of which part of the memory belongs to which `DescriptorChain` object
+    parsed_descriptors: VecDeque<ParsedDescriptorChain>,
+    // Buffers that we have used and they are ready to be given back to the guest.
+    deferred_descriptor: Option<ParsedDescriptorChain>,
```
roypat: I take a bit of a philosophical issue with this field being used for non-deferred descriptors (e.g. we assign each descriptor we use to this field temporarily, but they only become "deferred" if we hit the ratelimiter in a check that happens after the assignment to this variable). 

I take a second issue with the `rx_bytes_read` variable in `Net`. That variable I think should move into `RxBuffers`, because really, it now means `rx_bytes_deferred`. BUT, I don't think we need this `rx_bytes_*` variable at all, because it has the same value as the `length` of the deferred descriptor. Additionally, if we only set `deferred_descriptor` while we actually are deferring a descriptor, then the `rx_deferred_descriptor` bool in `Net` also becomes useless, as it will just be `rx_buffers.deferred_descriptor.is_some()`.

So my proposal is the following: Change `.add_used` and `.read_from_mmds_or_tap` to returned the `ParsedDescriptor` that was written to. Change `rate_limited_rx_single_frame` and `finish_frame` to take a `ParsedDescriptor` parameter. If the ratelimits are insufficient, then set `deferred_descriptor` only in this function.

Essentially, the following diff implements this idea (compile-tested only, and unittests blissfully ignored):

<details>

```diff
diff --git a/src/vmm/src/devices/virtio/net/device.rs b/src/vmm/src/devices/virtio/net/device.rs
index f42c34c0d..adeec3422 100755
--- a/src/vmm/src/devices/virtio/net/device.rs
+++ b/src/vmm/src/devices/virtio/net/device.rs
@@ -152,7 +152,7 @@ impl RxBuffers {
     /// The `size` should be the number of bytes previously written in the [`RxBuffers`]. This
     /// means that `size` cannot be more than the currently available bytes in the underlying
     /// `IoVecBufferMut`.
-    fn mark_used(&mut self, size: u32, metrics: &NetDeviceMetrics) {
+    fn mark_used(&mut self, size: u32, metrics: &NetDeviceMetrics) -> ParsedDescriptorChain {
         // Since we were able to write a frame in guest memory, we should have at least one
         // descriptor chain here. If not, we have a bug, so fail fast, since the device is
         // fundamentally broken.
@@ -174,7 +174,7 @@ impl RxBuffers {
         }
         self.header_set_num_buffers(1);
         self.iovec.drop_descriptor_chain(&parsed_dc);
-        self.deferred_descriptor = Some(parsed_dc);
+        parsed_dc
     }
 
     /// Write the number of descriptors used in VirtIO header
@@ -193,20 +193,15 @@ impl RxBuffers {
 
     /// This will let the guest know that about all the `DescriptorChain` object that has been
     /// used to receive a frame from the TAP.
-    fn finish_frame(&mut self, rx_queue: &mut Queue) -> bool {
-        match self.deferred_descriptor.take() {
-            Some(used_dc) => {
-                if let Err(err) = rx_queue.add_used(used_dc.head_index, used_dc.length) {
-                    error!(
-                        "net: Failed to add used descriptor {} of length {} to RX queue: {err}",
-                        used_dc.head_index, used_dc.length
-                    );
-                    false
-                } else {
-                    true
-                }
-            }
-            None => false,
+    fn finish_frame(&mut self, rx_queue: &mut Queue, frame: &ParsedDescriptorChain) -> bool {
+        if let Err(err) = rx_queue.add_used(frame.head_index, frame.length) {
+            error!(
+                "net: Failed to add used descriptor {} of length {} to RX queue: {err}",
+                frame.head_index, frame.length
+            );
+            false
+        } else {
+            true
         }
     }
 }
@@ -231,9 +226,6 @@ pub struct Net {
     pub(crate) rx_rate_limiter: RateLimiter,
     pub(crate) tx_rate_limiter: RateLimiter,
 
-    pub(crate) rx_deferred_frame: bool,
-
-    rx_bytes_read: usize,
     rx_frame_buf: [u8; MAX_BUFFER_SIZE],
 
     tx_frame_headers: [u8; frame_hdr_len()],
@@ -299,8 +291,6 @@ impl Net {
             queue_evts,
             rx_rate_limiter,
             tx_rate_limiter,
-            rx_deferred_frame: false,
-            rx_bytes_read: 0,
             rx_frame_buf: [0u8; MAX_BUFFER_SIZE],
             tx_frame_headers: [0u8; frame_hdr_len()],
             irq_trigger: IrqTrigger::new().map_err(NetError::EventFd)?,
@@ -422,15 +412,18 @@ impl Net {
     // Attempts to copy a single frame into the guest if there is enough
     // rate limiting budget.
     // Returns true on successful frame delivery.
-    fn rate_limited_rx_single_frame(&mut self) -> bool {
+    fn rate_limited_rx_single_frame(&mut self, frame: ParsedDescriptorChain) -> bool {
         let rx_queue = &mut self.queues[RX_INDEX];
-        if !Self::rate_limiter_consume_op(&mut self.rx_rate_limiter, self.rx_bytes_read as u64) {
+        if !Self::rate_limiter_consume_op(&mut self.rx_rate_limiter, frame.length as u64) {
             self.metrics.rx_rate_limiter_throttled.inc();
+            self.rx_buffer.deferred_descriptor = Some(frame);
             return false;
         }
 
-        if !self.rx_buffer.finish_frame(rx_queue) {
-            Self::rate_limiter_replenish_op(&mut self.rx_rate_limiter, self.rx_bytes_read as u64);
+        if !self.rx_buffer.finish_frame(rx_queue, &frame) {
+            Self::rate_limiter_replenish_op(&mut self.rx_rate_limiter, frame.length as u64);
+            self.rx_buffer.deferred_descriptor = Some(frame);
+            return false;
         }
         true
     }
@@ -550,7 +543,7 @@ impl Net {
     }
 
     // We currently prioritize packets from the MMDS over regular network packets.
-    fn read_from_mmds_or_tap(&mut self) -> Result<usize, NetError> {
+    fn read_from_mmds_or_tap(&mut self) -> Result<ParsedDescriptorChain, NetError> {
         // If we don't have any buffers available try to parse more from the RX queue. There might
         // be some buffers we didn't get the chance to process, because we got to handle the TAP
         // event before the RX queue event.
@@ -560,7 +553,7 @@ impl Net {
             // If after parsing the RX queue we still don't have any buffers stop processing RX
             // frames.
             if self.rx_buffer.len() == 0 {
-                return Ok(0);
+                return Err(NetError::NoRxDescriptors)
             }
         }
 
@@ -575,33 +568,29 @@ impl Net {
                 self.rx_buffer
                     .iovec
                     .write_all_volatile_at(&self.rx_frame_buf[..vnet_hdr_len() + len], 0)?;
-                self.rx_buffer
-                    .mark_used((vnet_hdr_len() + len).try_into().unwrap(), &self.metrics);
-                return Ok(vnet_hdr_len() + len);
+                return Ok(self.rx_buffer
+                    .mark_used((vnet_hdr_len() + len).try_into().unwrap(), &self.metrics))
             }
         }
 
         // SAFETY: this is safe because we ensured that `self.rx_buffer` has at least one
         // DescriptorChain parsed in it.
         let len = unsafe { self.read_tap().map_err(NetError::IO) }?;
-        self.rx_buffer
-            .mark_used(len.try_into().unwrap(), &self.metrics);
-        Ok(len)
+        Ok(self.rx_buffer
+            .mark_used(len.try_into().unwrap(), &self.metrics))
     }
 
     /// Read as many frames as possible.
     fn process_rx(&mut self) -> Result<(), DeviceError> {
         loop {
             match self.read_from_mmds_or_tap() {
-                Ok(0) => {
+                Err(NetError::NoRxDescriptors) => {
                     self.metrics.no_rx_avail_buffer.inc();
                     break;
                 }
-                Ok(count) => {
-                    self.rx_bytes_read = count;
+                Ok(parsed_chain) => {
                     self.metrics.rx_count.inc();
-                    if !self.rate_limited_rx_single_frame() {
-                        self.rx_deferred_frame = true;
+                    if !self.rate_limited_rx_single_frame(parsed_chain) {
                         break;
                     }
                 }
@@ -627,25 +616,15 @@ impl Net {
         self.try_signal_queue(NetQueue::Rx)
     }
 
-    // Process the deferred frame first, then continue reading from tap.
-    fn handle_deferred_frame(&mut self) -> Result<(), DeviceError> {
-        if self.rate_limited_rx_single_frame() {
-            self.rx_deferred_frame = false;
-            self.rx_buffer.deferred_descriptor = None;
-            // process_rx() was interrupted possibly before consuming all
-            // packets in the tap; try continuing now.
-            return self.process_rx();
-        }
-
-        self.try_signal_queue(NetQueue::Rx)
-    }
-
     fn resume_rx(&mut self) -> Result<(), DeviceError> {
-        if self.rx_deferred_frame {
-            self.handle_deferred_frame()
-        } else {
-            self.process_rx()
+        if let Some(frame) = self.rx_buffer.deferred_descriptor.take() {
+            if !self.rate_limited_rx_single_frame(frame)
+            {
+                return self.try_signal_queue(NetQueue::Rx)
+            }
         }
+
+        self.process_rx()
     }
 
     fn process_tx(&mut self) -> Result<(), DeviceError> {
@@ -706,7 +685,7 @@ impl Net {
                 &self.metrics,
             )
             .unwrap_or(false);
-            if frame_consumed_by_mmds && !self.rx_deferred_frame {
+            if frame_consumed_by_mmds && !self.rx_buffer.deferred_descriptor.is_some() {
                 // MMDS consumed this frame/request, let's also try to process the response.
                 process_rx_for_mmds = true;
             }
@@ -836,16 +815,7 @@ impl Net {
             return;
         }
 
-        if self.rx_deferred_frame
-        // Process a deferred frame first if available. Don't read from tap again
-        // until we manage to receive this deferred frame.
-        {
-            self.handle_deferred_frame()
-                .unwrap_or_else(|err| report_net_event_fail(&self.metrics, err));
-        } else {
-            self.process_rx()
-                .unwrap_or_else(|err| report_net_event_fail(&self.metrics, err));
-        }
+        self.resume_rx().unwrap_or_else(|err| report_net_event_fail(&self.metrics, err));
     }
 
     /// Process a single TX queue event.
diff --git a/src/vmm/src/devices/virtio/net/mod.rs b/src/vmm/src/devices/virtio/net/mod.rs
index e8a3f86ac..cefccc6cf 100644
--- a/src/vmm/src/devices/virtio/net/mod.rs
+++ b/src/vmm/src/devices/virtio/net/mod.rs
@@ -58,4 +58,6 @@ pub enum NetError {
     VnetHeaderMissing,
     /// IoVecBuffer(Mut) error: {0}
     IoVecError(#[from] IoVecError),
+    /// No rx descriptors available
+    NoRxDescriptors,
 }

```

</details>

Now, admittedly, I don't know if this idea is compatible with future extentions for mergable buffers

roypat (commented):
I only made it half-way through commit 2 before my lunch break, but dumping these comments already in case they are helpful

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: you are correct. In this case we don't need `self.rx_buffer.deferred_descriptor = None;` at all.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -312,125 +423,55 @@ impl Net {
     // rate limiting budget.
     // Returns true on successful frame delivery.
     fn rate_limited_rx_single_frame(&mut self) -> bool {
```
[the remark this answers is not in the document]
ShadowCurse: This will not work well. If we limit amount of buffer space we give to `readv` with what we have rate limit for, it will make a lot of packets to be split into multiple `readv` calls. Example: we have 32K allowance from rate limiter. We give it to the `readv`. TAP has 64K packet ready to be read. `readv` happily writes first 32K into provided buffer. We have not idea that this is not the whole packet, so we process it as a whole packet. Next time we call `readv` it will write remaining part of the packet, but again, we have no idea that it is not a new packet.
Basically because we have not way of asking TAP about how big is the pending packet, we cannot rate limit before `readv`.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -312,125 +423,55 @@ impl Net {
...
-            // If chunk is empty we are done here.
-            if chunk.is_empty() {
-                let len = data.len() as u64;
-                net_metrics.rx_bytes_count.add(len);
-                net_metrics.rx_packets_count.inc();
-                return Ok(());
+    /// Returns the minimum size of buffer we expect the guest to provide us depending on the
+    /// features we have negotiated with it
+    fn minimum_rx_buffer_size(&self) -> u32 {
+        if !self.has_feature(VIRTIO_NET_F_MRG_RXBUF as u64) {
```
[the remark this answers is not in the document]
bchalios: So, the reason why I added this here is that if I make sure that all the buffers we have parsed are of the correct size we can't have a `readv` from the TAP that leaks to a second `DescriptorChain`, which makes all of the error handling much easier. It also, allows to write the `num_buffers` in the header without having to check for the length.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -517,17 +571,31 @@ impl Net {
...
+        let len = unsafe { self.read_tap().map_err(NetError::IO) }?;
+        self.rx_buffer
+            .mark_used(len.try_into().unwrap(), &self.metrics);
+        Ok(len)
     }
 
+    /// Read as many frames as possible.
     fn process_rx(&mut self) -> Result<(), DeviceError> {
-        // Read as many frames as possible.
         loop {
```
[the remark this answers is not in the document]
bchalios: I agree that this error handling is much more meaningful. I have a branch somewhere that does something similar to this. Would you mind if I do this in a separate PR?

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
bchalios: The problem with rate limiting before is that we don't know how much we should rate limit. We only know that after we read from the TAP. The next frame in the TAP might be 52 bytes, so if we were to rate limit by the size of the available buffer we would penalize the guest.

Unfortunately, we can't know how many bytes we can read out of the TAP device (If we did, this whole PR would be significantly simpler :))

Now a ping is not the usual case here, but (1) right now the buffers we have are ~72K big and (2) even with mergeable buffers merged, we would be rate-limiting in chunks of an MTU which might penalize traffic that consists of many small frames. I've not idea whether such a scenario exists realistically, but I would be unwilling to change things before understanding whether that's a valid use-case.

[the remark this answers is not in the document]
ShadowCurse: without mrg_rxbuf we check the capacity of the desc chain to be 64K minimum. So we always pass 64K to readv. With mrg_rxbuf we will do approximately the same thing. So we should never have 1 packet split between 2 `readv` calls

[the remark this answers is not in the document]
roypat: Ack, makes sense, thanks for the explanation!

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
roypat: Works for me, yeah, this isn't really related to this PR

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
bchalios: oh, right. Yes, I can do that.

src/vmm/src/devices/virtio/net/device.rs
bchalios: you are right. I will drop it.

src/vmm/src/devices/virtio/net/device.rs
bchalios: I remember trying to do that and having some issues with error handling and error propagation, etc., but some time has passed since, so let me try to do that again now. If nothing breaks, I'm happy to commit it.

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -846,8 +893,37 @@ mod verification {
...
+    pub fn drop_iov_deque(deque: &mut IovDeque) {
+        // We previously allocated this memory with a call to `std::alloc::alloc` with the same
+        // allocator
+        unsafe {
+            std::alloc::dealloc(
+                deque.iov.cast(),
+                std::alloc::Layout::from_size_align_unchecked(2 * PAGE_SIZE, PAGE_SIZE),
+            )
+        }
+    }
```
roypat, line 923: You don't actually need to do this, kani has no concept of memory leaks

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: mrg_rxbuf will not like this as there we will have to return a `Vec<ParsedDescriptor>` and this is not the way we should go.
I have a commit in my previous PR that deals with deferred variable: https://github.com/firecracker-microvm/firecracker/pull/4813/commits/a267a0ce3ab8790f1ad397e83faf2fdbbf199b96

src/vmm/src/devices/virtio/net/tap.rs
```diff
@@ -190,6 +190,20 @@ impl Tap {
...
+        // is provide by the `IoVecBufferMut` implementation and we check the return value.
+        let ret = unsafe { libc::readv(self.tap_file.as_raw_fd(), iov, iovcnt) };
+        if ret == -1 {
+            return Err(IoError::last_os_error());
+        }
+        Ok(usize::try_from(ret).unwrap())
+    }
 }
 
 impl Read for Tap {
```
roypat, line 209: Can remove this `Read` impl now

src/vmm/src/devices/virtio/net/device.rs
roypat: > mrg_rxbuf will not like this as there we will have to return a `Vec<ParsedDescriptor>` and this is not the way we should go. I have a commit in my previous PR that deals with deferred variable: [a267a0c](https://github.com/firecracker-microvm/firecracker/commit/a267a0ce3ab8790f1ad397e83faf2fdbbf199b96)

Mh, so in https://github.com/firecracker-microvm/firecracker/commit/06d97cdd92bd40cdf8c485bc097c5e7bd8c75646 you also move the writing of the used ring entries to `mark_used` (instead of only doing it in `finish_packet`), and in that case we can just use a `(bytes_written, nr_descriptors_used)` tuple instead of a `ParsedDescriptorChain`, and the rest of my diff would look pretty much the same, I think

ShadowCurse: ok, in any case, seems like this can be a follow up PR.

roypat: > ok, in any case, seems like this can be a follow up PR.

I disagree, this is mostly about code introduced in this PR, and imo it makes the logic a lot easier to follow because it becomes less stateful.

src/vmm/src/devices/virtio/net/device.rs
bchalios: done

src/vmm/src/devices/virtio/net/device.rs
bchalios: done

src/vmm/src/devices/virtio/net/device.rs
bchalios: done

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -0,0 +1,447 @@
...
+    /// so, it's 16 bytes long. As a result, we need a single page for holding the actual data of
+    /// our buffer.
+    fn allocate_ring_buffer_memory() -> Result<*mut c_void, IovDequeError> {
+        // The fact that we allocate two pages is due to the size of `struct iovec` times our queue
+        // size equals the page size. Add here a debug assertion to reflect that and ensure that we
+        // will adapt our logic if the assumption changes in the future.
+        debug_assert_eq!(
+            std::mem::size_of::<iovec>() * usize::from(FIRECRACKER_MAX_QUEUE_SIZE),
+            PAGE_SIZE
+        );
```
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -0,0 +1,447 @@
...
+                libc::MAP_SHARED | libc::MAP_FIXED,
+                raw_memfd,
+                0,
+            )
+        }?;
+
+        // Map the second page of virtual memory to the physical page described by the memfd object
+        // SAFETY: safe because `Self::allocate_ring_buffer_memory` allocates exactly two pages for
+        // us
+        let next_page = unsafe { buffer.add(PAGE_SIZE) };
```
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -214,64 +218,131 @@ impl IoVecBuffer {
...
 impl IoVecBufferMut {
-    /// Create an `IoVecBuffer` from a `DescriptorChain`
-    ///
-    /// # Safety
-    ///
-    /// The descriptor chain cannot be referencing the same memory location as another chain
-    pub unsafe fn load_descriptor_chain(
+    /// Parse a `DescriptorChain` object and append the memory regions it describes in the
+    /// underlying ring buffer.
+    fn parse_descriptor(
```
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/devices/virtio/iovec.rs
```diff
@@ -429,17 +515,36 @@ mod tests {
...
+            let mut len = 0usize;
+            let mut vecs = IovDeque::new().unwrap();
+            buffer.drain(..).for_each(|slice| {
+                len += slice.len();
+
+                vecs.push_back(iovec {
+                    iov_base: slice.as_ptr() as *mut c_void,
+                    iov_len: slice.len(),
+                });
+            });
```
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/devices/virtio/net/device.rs
bchalios: Ok, I implemented a version of this suggestion.

I renamed `deferred_frame` to `used_descriptor`. It does not exactly address you philosophical issue :) but my rationale here is that `used_descriptor` will hold that last used `DescriptorChain`, until someone calls `finish_frame`. Then, we can simply drop the `rx_deferred_descriptor` `bool` and substitute it with the `is_some()` condition you mentioned.

I did not have `mark_used` to return anything. What we need is the number of bytes written in guest memory, however that we already know at the call site; it is the `size` argument of `mark_used` itself. This way we can also, trivially drop the `rx_bytes_read` as well, i.e. now, `rate_limited_rx_single_frame` takes the same `size` argument.

I think this is trivially extended for mergeable buffers, WDYT?

src/vmm/src/devices/virtio/iovec.rs
bchalios: I need to do something for the `munmap` failure, so I did this. Would you prefer that I just `std::mem::forget` `deque.iov`?

src/vmm/src/devices/virtio/net/tap.rs
bchalios: done

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -510,6 +495,10 @@ impl Net {
 
     // We currently prioritize packets from the MMDS over regular network packets.
     fn read_from_mmds_or_tap(&mut self) -> Result<usize, NetError> {
+        if self.rx_buffer.len() == 0 {
```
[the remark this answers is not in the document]
bchalios: It doesn't if the method is not `pub`, but I added it anyways.

ShadowCurse (dismissed):

src/vmm/src/devices/virtio/net/device.rs
roypat: Yeah, I think resolving the duplication of the fields in Net/RxBuffers addresses my main concern, so the rest of this is a nit. I will try to motivate my proposed changes again though xP: Right now, what does it mean to call `finish_frame` if `used_descriptor` is `None`? The type system doesn't stop us from doing that, but it would be a logic bug in our code. If we instead have `finish_frame` take all the information it needs as (non-optional) parameters, then we eliminate this class of bugs at compile time.

Also, if we do a follow up, can we untangle `resume_rx`, `handle_deferred_frame` and `process_rx_events` like I was forced to do in my diff above? Right now they're quite hard to follow imo because they all contain very similar logic to check for deferred frames, while also calling each other.

src/vmm/src/devices/virtio/iovec.rs
roypat: Ah, yeah, just doing `std::mem::forget` is fine :)

src/vmm/src/devices/virtio/net/device.rs
bchalios: Ok, let me have another go with it :)

src/vmm/src/devices/virtio/iovec.rs
bchalios: done

roypat (approved):
:tada:

ShadowCurse (approved):

PR #4803, opened 2024-09-13
Improve dependency management between buildkite steps

roypat:
## Changes

- Make the `build` group not block on the shared compilation step (by moving the 2 tests that require the shared artifacts to the functional group)
- Make x86 functional tests not wait on arm compilation finishing

## Reason

my ocd, mostly

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


.buildkite/common.py
```diff
@@ -13,14 +13,14 @@
 import subprocess
 from pathlib import Path
 
-DEFAULT_INSTANCES = [
```
pb8o, line 16: Could use a StrEnum, but 3.11 only :/

```
class InstanceType(StrEnum):
    m5d = "m5d.metal"
    m6g = "m6g.metal"
    m7g = "m7g.metal"

    @property
    def arch(self):
        if self.value[2] == 'g':
            return "aarch64"
        return "x86_64"
```

it may be too magic anyway, nevermind, we can go there if this becomes more complex.

.buildkite/common.py
```diff
@@ -258,7 +258,9 @@ def __init__(self, with_build_step=True, **kwargs):
         # Build sharing
         if with_build_step:
             build_cmds, self.shared_build = shared_build()
-            self.build_group_per_arch("🏗️ Build", build_cmds, decorate=False)
+            self.build_group_per_arch(
+                "🏗️ Build", build_cmds, decorate=False, key="build"
```
pb8o, line 262: `key_prefix` may be a better name as the key is then derived from it.

.buildkite/common.py
```diff
@@ -307,13 +309,18 @@ def build_group(self, *args, **kwargs):
...
 
-    def build_group_per_arch(self, *args, **kwargs):
+    def build_group_per_arch(self, label, *args, **kwargs):
         """
         Build a group, parametrizing over the architectures only.
         """
         decorate = kwargs.pop("decorate", True)
         combined = overlay_dict(self.per_arch, kwargs)
-        return self.add_step(group(*args, **combined), decorate=decorate)
+        key = kwargs.pop("key", None)
```
pb8o, line 321: may want to move it one line above so the `combined` dict does not have they `key` key :)

.buildkite/pipeline_pr.py
```diff
@@ -21,15 +21,18 @@
...
     with_build_step=not DOC_ONLY_CHANGE,
 )
 
+pipeline.add_step(
+    {
+        "command": "./tools/devtool -y checkstyle",
+        "label": "🪶 Style",
+        "depends_on": None,
+    },
+    decorate=False,
```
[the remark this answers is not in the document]
roypat: Added on a commit with a rename :)

.buildkite/common.py
roypat: done!

.buildkite/common.py
roypat: Ah, yes, good catch. I've also added a comment about which kwargs are popped before the call to `group`

pb8o (approved):

Issue #4811, opened 2024-09-18
[Bug] Ongoing `read` Syscalls on Vsocks Don't Get Interrupted after the Second Snapshot Resume/Restore


pojntfx:
# Describe the bug

Recently, we [reported a bug](https://github.com/firecracker-microvm/firecracker/issues/4736) that lead to ongoing `read` syscalls not being interrupted after a snapshot resume. @ShadowCurse Then published a [fix](
https://github.com/firecracker-microvm/firecracker/pull/4796/files) for this. We then verified that their fix worked by using our reproducer repo:

https://github.com/firecracker-microvm/firecracker/issues/4736#issuecomment-2347815017

In the reproducer repo, we're essentially doing this:

1. Start Firecracker
2. Create a VM with a VSock
3. Start a VSock-over-UDS listener on the host with `socat`
4. In the guest VM, connect to the listener on the host through VSock with `socat`
5. Pause the VM and create a snapshot
6. Stop the listener on the host
7. Resume the VM

And this now all works fine now! At step 7, any in-progress reads from the VSock in the guest VM now return RST errors (as they should).

But if we add the following steps:

8. Start a VSock-over-UDS listener on the host with `socat` again
9.  In the guest VM, connect to the listener on the host through VSock with `socat`
10. Pause the VM and create a snapshot
11. Stop the listener on the host
12. Resume the VM

The issue appears again! Essentially, if we create & resume two or more subsequent snapshots (already verified this with upstream Firecracker using our existing reproducer repo), reads from the VSocks don't get any RST errors anymore but instead hang forever.

## To Reproduce

See the [original reproduction steps](https://github.com/firecracker-microvm/firecracker/issues/4736#issue-2474280231), but add the additional steps described above (snapshot & restore more than one time)

## Expected behaviour

See the [original expected behavior](https://github.com/firecracker-microvm/firecracker/issues/4736#issue-2474280231) - from our reading, this behavior should continue to work even if it's the 2nd, 3rd etc. snapshot/restore cycle.

## Potential Fix

We noticed that in `src/vmm/src/device_manager/persist.rs`, the VSock state is serialized _before_ the reset event is sent. If this is swapped, the issue goes away, and the 2nd, 3rd etc. resume still correctly resets any connected VSocks. We're not sure if this is a proper solution though, if there are any better workarounds or if there is a better way to fix the issue. To test it, apply the following patch:

```patch
[...]

PR #4796, opened 2024-09-11
Kick vsock during VM resume

ShadowCurse:
## Changes
Kick vsock when resuming VM. This ensures that if VM is restored from a snapshot, guest gets a notification to read a queue with `TRANSPORT_RESET_EVENT` inside. This will trigger guest to close all vsock connections. This is expected behavior as vsock devices does not support bringing connections over the snapshot boundary.

## Reason
Fixes #4736

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following Developer
Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] The description of changes is clear and encompassing.
- [ ] Any required documentation changes (code and docs) are included in this
  PR.
- [ ] API changes follow the [Runbook for Firecracker API changes][2].
- [ ] User-facing changes are mentioned in `CHANGELOG.md`.
- [ ] All added/changed functionality is tested.
- [ ] New `TODO`s link to an issue.
- [ ] Commits meet
  [contribution quality standards](https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md#contribution-quality-standards).

______________________________________________________________________

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #4812, opened 2024-09-18
fix(vsock): save state after sending a notification

ShadowCurse:
## Changes
This is a fix for a fix introduced in #4796
The issue was in vsock device hanging after snapshot restoration due to the guest not being notified about the termination packet. But there was bug in the fix, maily we saved the vsock state before the notification was sent, thus discarding all modifications made to sent the notification.

The reason original fix worked, is because we were only testing with 1 iteration of snap/restore. This way even though we lost synchronization with the guest in the event queue state, it worked fine once. But doing more iterations causes vsock to hang as before.

This PR fixes the issue by storing vsock state after the notification is sent and modifies the vsock test to run multiple iterations of snap/restore.

## Reason
Fixes: #4811

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/tests/integration_tests/functional/test_vsock.py
+++ b/tests/integration_tests/functional/test_vsock.py
@@ -225,54 +225,61 @@ def test_vsock_transport_reset_g2h(uvm_nano, microvm_factory):
     test_vm.api.vsock.put(vsock_id="vsock0", guest_cid=3, uds_path=f"/{VSOCK_UDS_PATH}")
     test_vm.start()
 
-    host_socket_path = os.path.join(
-        test_vm.path, f"{VSOCK_UDS_PATH}_{ECHO_SERVER_PORT}"
-    )
-    host_socat_commmand = [
-        "socat",
-        "-dddd",
-        f"UNIX-LISTEN:{host_socket_path},fork",
-        "STDOUT",
-    ]
-    host_socat = subprocess.Popen(
-        host_socat_commmand, stdout=subprocess.PIPE, stderr=subprocess.PIPE
-    )
-
-    # Give some time for host socat to create socket
-    time.sleep(0.5)
-    assert Path(host_socket_path).exists()
-    test_vm.create_jailed_resource(host_socket_path)
-
-    # Create a socat process in the guest which will connect to the host socat
-    guest_socat_commmand = f"tmux new -d 'socat - vsock-connect:2:{ECHO_SERVER_PORT}'"
-    test_vm.ssh.run(guest_socat_commmand)
-
-    # socat should be running in the guest now
-    code, _, _ = test_vm.ssh.run("pidof socat")
-    assert code == 0
-
-    # Create snapshot.
+    # Create snapshot and terminate a VM.
     snapshot = test_vm.snapshot_full()
-    test_vm.resume()
-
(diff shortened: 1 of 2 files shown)


tests/integration_tests/functional/test_vsock.py
```diff
@@ -225,54 +225,64 @@ def test_vsock_transport_reset_g2h(uvm_nano, microvm_factory):
...
+
+        # Create a socat process in the guest which will connect to the host socat
+        guest_socat_commmand = (
+            f"tmux new -d 'socat - vsock-connect:2:{ECHO_SERVER_PORT}'"
+        )
+        new_vm.ssh.run(guest_socat_commmand)
+
+        # socat should be running in the guest now
+        code, _, _ = new_vm.ssh.run("pidof socat")
+        assert code == 0
```
pb8o, line 273: Can be `new_vm.ssh.check_output()`

tests/integration_tests/functional/test_vsock.py
```diff
@@ -225,54 +225,64 @@ def test_vsock_transport_reset_g2h(uvm_nano, microvm_factory):
...
+        snapshot = new_vm.snapshot_full()
+        new_vm.resume()
+
+        # After `create_snapshot` + 'restore' calls, connection should be dropped
+        code, _, _ = new_vm.ssh.run("pidof socat")
+        assert code == 1
+
+        # Kill host socat as it is not useful anymore
+        host_socat.kill()
+        host_socat.communicate()
```
pb8o, line 285: we could re-use the socat echo server between runs, but for that we would need a path outside the VM root.

pb8o (approved):

tests/integration_tests/functional/test_vsock.py
```diff
@@ -225,54 +225,64 @@ def test_vsock_transport_reset_g2h(uvm_nano, microvm_factory):
...
-    guest_socat_commmand = f"tmux new -d 'socat - vsock-connect:2:{ECHO_SERVER_PORT}'"
-    test_vm.ssh.run(guest_socat_commmand)
-
-    # socat should be running in the guest now
-    code, _, _ = test_vm.ssh.run("pidof socat")
-    assert code == 0
-
     # Create snapshot.
     snapshot = test_vm.snapshot_full()
     test_vm.resume()
```
zulinx86, line 230: why do we need to resume it here, although it'll be killed soon?

tests/integration_tests/functional/test_vsock.py
ShadowCurse: The reason I explicitly check the code is to make it more apparent when it is 1 and when it is 0. (so when it should fail and when it should not)

tests/integration_tests/functional/test_vsock.py
ShadowCurse: Don't think there will be meaningful difference compared to added complexity.

tests/integration_tests/functional/test_vsock.py
ShadowCurse: Actually I need to resume it to check that after a snapshot socat is killed.

zulinx86 (approved):

pb8o (approved):

PR #4820, opened 2024-09-25
bench: add benchmark for page fault latencies

roypat:
Add a benchmark that attempts to capture the latency of a single page
fault in a guest memory configuration that we also use in production.
This should catch regressions such as the 1.6 memfd/vhost post
snapshot-restore latency regression, and does indeed do so:

MAP_ANONYMOUS:
```
     Running benches/memory_access.rs
page_fault    time:   [2.8611 µs 2.9327 µs 3.0170 µs]
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe
```

memfd_create:
```
     Running benches/memory_access.rs
page_fault    time:   [5.7449 µs 5.8779 µs 6.0450 µs]
              change: [+85.649% +92.231% +98.074%] (p = 0.00 < 0.05)
              Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe
```

Add the same benchmark for huge pages, just in case. Funnily enough,
there is no vhost-tax to be paid for huge pages.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/src/vmm/benches/memory_access.rs
+++ b/src/vmm/benches/memory_access.rs
@@ -1,0 +1,70 @@
+// Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
+use vm_memory::{GuestAddress, GuestMemory};
+use vmm::resources::VmResources;
+use vmm::vmm_config::machine_config::{HugePageConfig, VmConfig};
+
+fn bench_single_page_fault(c: &mut Criterion, configuration: VmResources) {
+    c.bench_function("page_fault", |b| {
+        b.iter_batched(
+            || {
+                let memory = configuration.allocate_guest_memory().unwrap();
+                let ptr = memory
+                    .get_slice(GuestAddress(0), 1)
+                    .unwrap()
+                    .ptr_guard_mut()
+                    .as_ptr();
+
+                // fine to return both here, because ptr is not a reference into `memory` (e.g. no
+                // self-referential structs are happening here)
+                (memory, ptr)
+            },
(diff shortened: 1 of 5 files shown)


kalyazin (approved):

src/vmm/benches/memory_access.rs
```diff
@@ -0,0 +1,70 @@
...
+            },
+            ..Default::default()
+        },
+    )
+}
+
+criterion_group! {
+    name = memory_access_benches;
+    config = Criterion::default().noise_threshold(0.05);
+    targets = bench_4k_page_fault, bench_2m_page_fault
```
Manciukic, line 65: is this also testing the case with memfd on/off? it not, should we add also that to the benchmarks?

roypat: The idea Nikita and me had was that the benchmark should be oblivious to the actual memory allocation mechanism being used, and notify us if a change to the memory allocation mechanism (e.g. the change from MAP_ANONYMOUS to memfd_create in 1.6, or a potential future change to guest_memfd) incurs a higher page fault overhead.

Manciukic: but we're only benchmarking one of the two branches of the function, while it would cost nothing to benchmark also the other part.
what if we change the backing in vhost-enabled vms?

Manciukic: If we really wanted to, then we could change the function to accept a parameter for how the memory has to be allocated, so that we can easily test both paths.
But I wouldn't delay merging this PR on that tbh

Manciukic (approved):

PR #4834, opened 2024-10-04
Add VIRTIO_NET_F_MRG_RXBUF to virtio-net

ShadowCurse:
## Changes
Add VIRTIO_NET_F_MRG_RXBUF to virtio-net.

Now virtio-net device can split incoming packets across multiple descriptor chains if VIRTIO_NET_F_MRG_RXBUF is enabled by the guest. 

## Reason
Improves performance of virtio-net device.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -141,37 +144,56 @@ impl RxBuffers {
...
-        parsed_dc.length = size;
-        parsed_dc
+        self.header_set_num_buffers(used_heads);
+        for _ in 0..used_heads {
+            let parsed_dc = self
+                .parsed_descriptors
+                .pop_front()
+                .expect("This should never happen if write to the buffer succeded.");
+            self.iovec.drop_descriptor_chain(&parsed_dc);
+        }
```
roypat, line 196: Can we avoid the closure and second loop by doing

```rs
if bytes_written == 0 {
  return Ok(());
}

let mut used_heads = 0u16;
while let Some(parsed_dc) = self.parsed_descriptors.pop_front() {
  let used_bytes = bytes_written.min(parsed_dc.length);
  rx_queue.write_used_element(rx_queue.next_used + Wrapping(used_heads)).0, parsed_dc.head_index, used_bytes).unwrap(); // Safe because we know head_index isn't out of bounds
  bytes_written -= used_bytes;
  self.iovec.drop_descriptor_chain(&parsed_dc);
  used_heads += 1;

  if bytes_written == 0 {
    break;
  }
}

self.header_set_num_buffers(used_heads);
self.used_descriptors += used_heads;
```

?

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -190,14 +212,24 @@ impl RxBuffers {
...
+    }
+
+    /// Return a slice of iovecs for the first slice in the buffer.
+    ///
+    /// # Safety
+    /// Buffer needs to have at least one descriptor chain parsed.
+    unsafe fn single_chain_slice_mut(&mut self) -> &mut [iovec] {
+        let nr_iovecs = self.parsed_descriptors[0].nr_iovecs as usize;
+        &mut self.iovec.as_iovec_mut_slice()[..nr_iovecs]
+    }
```
roypat, line 228: I don't think this needs to be `unsafe` (same for `read_tap` actually), because it just panics if there's no descriptors, right?

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: The second loop is only avoidable if we store the pointer to the first `iovec` before doing any `drop_descriptor_chain`. Otherwise `header_set_num_buffers` will set `used_heads` into incorrect place (as the chains we used for packet are already dropped).

ShadowCurse: Removed the closure.

roypat: Ah, missed that. Thanks!

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: If we follow this logic, we can make any function `safe` if we just assert all invariants in the function itself. I don't have strong opinions here. I can remove this if you want.

roypat: > If we follow this logic, we can make any function safe if we just assert all invariants in the function itself.
Yes, we could :D That's pretty much the difference between `std::str::from_utf8` and `std::str::from_utf8_unchecked` - omitting assertions/checks for performance, at the cost of an `unsafe`. Its more difficult with functions like `VolatileSlice::new(ptr: *mut u8, len: usize)` - there's no assertion you can write that would guarantee that `ptr` points to valid memory of length `len`, so the function must be unsafe in every incarnation.

ShadowCurse: So, do we leave this as it is or remove `unsafe` qualifier? Actually if we keep them `unsafe` we can do:
```Rust
let nr_iovecs = self.parsed_descriptors.get_unchecked(0).nr_iovecs as usize;
```

roypat: `VecDeque` doesn't have a `get_unchecked`, so let's go with removing `unsafe` here

ShadowCurse: done

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -574,8 +574,13 @@ impl Queue {
         self.next_avail -= Wrapping(1);
```
bchalios, line 574: Could you please re-word the first commit into something like:

```
feat(virtio): split `add_used()` into 2 separate methods

`Queue::add_used()` method first writes a Descriptor head in the used descriptor
ring buffer and then advances the index of this buffer to let the guest know we
used one or more Descriptors.

Carve out each one of these steps in their own function so that we can add multiple
descriptors in the used ring and advance the index only once we finish handling
descriptors in one step.

This will be useful when, in later commits, we will implement mergeable buffers for
the RX queue of the network device

File: src/vmm/src/devices/virtio/net/device.rs [hunk 4]
```diff
@@ -6,11 +6,12 @@
 // found in the THIRD-PARTY file.
```
bchalios, line 6: The commit message body needs reworking here as well. Something like this maybe:

```
Implement VIRTIO_NET_F_MRG_RXBUF feature for VirtIO network device. It allows
receiving a single network frame into multiple descriptor chains. The amount of
descriptor chains (also known as heads) that were used is written into the
`virtio_net_hdr_v1` structure which is located a the beginning of the guest buffer.
```

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -154,18 +157,37 @@ impl RxBuffers {
...
+    unsafe fn mark_used(&mut self, mut bytes_written: u32, rx_queue: &mut Queue) {
+        self.used_bytes = bytes_written;
+
+        let mut used_heads: u16 = 0;
+        for parsed_dc in self.parsed_descriptors.iter() {
+            let used_bytes = bytes_written.min(parsed_dc.length);
+            // Safe because we know head_index isn't out of bounds
+            rx_queue
+                .write_used_element(
+                    (rx_queue.next_used + Wrapping(used_heads)).0,
```
bchalios, line 169: I would prefer it if we only passed `used_heads` to `write_used_element` and the calculation happens inside there. Like that, we don't need to expose `Queue` internals here.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -445,6 +488,9 @@ impl Net {
                 }
 
                 error!("net: Could not parse an RX descriptor: {err}");
+                // Notify queue about ready frames. We need this
+                // to bring queue into up to date state.
+                self.rx_buffer.finish_frame(queue);
```
bchalios, line 493: Could you explain to me why we call `finish_frame` here?

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -556,29 +604,44 @@ impl Net {
                 self.rx_buffer
                     .iovec
                     .write_all_volatile_at(&self.rx_frame_buf[..vnet_hdr_len() + len], 0)?;
+
+                // SAFETY:
+                // * len is never 0
+                // * Maximum size of the packet is 65562 which fits into u32
+                #[allow(clippy::cast_possible_truncation)]
+                let len = unsafe { NonZeroU32::new_unchecked((vnet_hdr_len() + len) as u32) };
```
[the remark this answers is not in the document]
ShadowCurse: I used `NonZeroU32` to keep the:
```Rust
            match self.read_from_mmds_or_tap() {
                Ok(None) => {
                     ...
                }
                Ok(Some(bytes)) => {
```
match. Otherwise it would be `Ok(0)` and `Ok(n)`. I don't really mind this option, but I guess `NonZeroU32` is a bit more explicit here.
Also `readv` should never return 0 as we only call it with some buffers passed. If `readv` has nothing to read, it will error with `EAGAIN`.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -556,29 +604,44 @@ impl Net {
...
+                return Ok(Some(len));
             }
         }
 
         // SAFETY: this is safe because we ensured that `self.rx_buffer` has at least one
         // DescriptorChain parsed in it.
         let len = unsafe { self.read_tap().map_err(NetError::IO) }?;
 
+        // SAFETY:
+        // * len is never 0
```
bchalios, line 631: I am not sure that this is true. `read` and `readv` might read less bytes than what we asked for, without returning an Error. Looking at the man pages, there doesn't seem to be any guarantees that these system calls cannot return 0 bytes.

In any case, how does the non-zero count guarantee help us, at all? What's the problem if `readv` returns 0?

src/vmm/src/devices/virtio/net/persist.rs
```diff
@@ -37,14 +36,15 @@ pub struct NetConfigSpaceState {
...
+    // Number of deferred_descriptors
+    deferred_descriptors: u16,
 }
 
 impl RxBufferState {
     fn from_rx_buffers(rx_buffer: &RxBuffers) -> Self {
         RxBufferState {
             parsed_descriptor_chains_nr: rx_buffer.parsed_descriptors.len().try_into().unwrap(),
-            deferred_descriptor: rx_buffer.deferred_descriptor.clone(),
+            deferred_descriptors: rx_buffer.used_descriptors,
```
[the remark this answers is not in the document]
ShadowCurse: added `used_bytes` as well to store/restore them.
But in general this would not have been a big problem to skip this field, as `used_bytes` are only used for rate limiter, and the worst thing which could have happened is if guest was throttled when snapshot was taken, it would be not be throttled for 1 packet after restore.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -531,16 +577,18 @@ impl Net {
...
-        // If we don't have any buffers available try to parse more from the RX queue. There might
-        // be some buffers we didn't get the chance to process, because we got to handle the TAP
-        // event before the RX queue event.
-        if self.rx_buffer.is_empty() {
+    fn read_from_mmds_or_tap(&mut self) -> Result<Option<NonZeroU32>, NetError> {
+        // We only want to read from TAP (or mmds) if we have at least 64K of available capacity as
+        // this is the max size of 1 packet.
+        // SAFETY:
+        // * MAX_BUFFER_SIZE is constant and fits into u32
+        #[allow(clippy::cast_possible_truncation)]
```
bchalios, line 585: I really dislike `try_into().unwrap()` (after Patrick brainwashing me :P) but I dislike this more :) I know that there are places that the truncation cannot happen today (like this one) but 1. you use them all over the place and 2. I'm afraid of what might happen when/if the code changes in a way that makes it possible to actually truncate. This is like leaving an Easter egg for someone in the future.

bchalios (changes requested):

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -141,37 +144,56 @@ impl RxBuffers {
...
+        }
+
+        self.header_set_num_buffers(used_heads);
+        for _ in 0..used_heads {
+            let parsed_dc = self
+                .parsed_descriptors
+                .pop_front()
+                .expect("This should never happen if write to the buffer succeded.");
+            self.iovec.drop_descriptor_chain(&parsed_dc);
+        }
```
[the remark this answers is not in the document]
ShadowCurse: I need to call `header_set_num_buffers` before doing any `self.iovec.drop...` because otherwise the header will be written into incorrect buffers.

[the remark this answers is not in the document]
roypat: I guess since both Babis and me tripped over this, maybe we should leave a comment

src/vmm/src/devices/virtio/net/device.rs
roypat: I think for some of these, we can sidestep the issue by upcasting the u32 to usize for the comparison, instead of coercing the constant (oh how I wish clippy was clever enough to understand that constants will not truncate)

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -141,75 +144,114 @@ impl RxBuffers {
...
         // descriptor chain here. If not, we have a bug, so fail fast, since the device is
         // fundamentally broken.
-        let mut parsed_dc = self.parsed_descriptors.pop_front().expect(
-            "net: internal bug. Mismatch between written frame size and available descriptors",
+        debug_assert!(!self.iovec.is_empty());
+        let header_ptr: *mut virtio_net_hdr_v1 = self.iovec.vecs.as_mut_slice()[0].iov_base.cast();
+        let header_buff_len = self.iovec.vecs.as_mut_slice()[0].iov_len;
+        assert!(
+            vnet_hdr_len() <= header_buff_len,
+            "Network buffer should be big enough for virtio_net_hdr_v1 object"
```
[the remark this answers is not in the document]
ShadowCurse: I think this is old comment. This code does not exist anymore.

src/vmm/src/devices/virtio/queue.rs
ShadowCurse: done

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: done

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: done

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: Added a better comment. But in short, because now we write to the used_ring separately from bumping queue, if we have pending packet, `add_used` call will mess everything up.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: Added a comment. `readv` will return 0 if we pass it array of iovecs with 0 capacity. But we never do this.
The main thing with `NonZeroU32` was to keep existing `match` in the `process_rx`. I can swap it back to u32 if you and @roypat really want.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: swapped to use `try_into().unwrap()`.
In general it is not a clippy issue, but Rust issue that it cannot understand compile time integers.

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
ShadowCurse: Added a comment.

src/vmm/src/devices/virtio/net/device.rs
bchalios: `readv` does not return 0 only when we pass a buffer of 0. According to the `man read(2):

> It is not an error if this number is smaller than the number of bytes requested; this may happen for example because fewer bytes are actually available right now (maybe because we were close to end-of-file, or because we are reading from a pipe, or from a terminal)  __or because read() was interrupted by a signal__.

So, it seems to imply that reading might be interrupted without returning an `EINTR`.

But again, I don't think that `NonZeroU32` adds any value here.

src/vmm/src/devices/virtio/net/device.rs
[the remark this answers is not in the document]
ShadowCurse: Replaced with `u32`.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -154,18 +156,36 @@ impl RxBuffers {
...
-        parsed_dc
+        // We need to set num_buffers before dropping chains from `self.iovec`. Otherwise
+        // when we set headers, we will iterate over new, yet unused chains instead of the ones
+        // we need.
+        self.header_set_num_buffers(used_heads);
+        for _ in 0..used_heads {
+            let parsed_dc = self
+                .parsed_descriptors
+                .pop_front()
+                .expect("This should never happen if write to the buffer succeded.");
```
bchalios, line 186: `s/succeded/succeeded/`

ShadowCurse: fixed

src/vmm/src/devices/virtio/net/device.rs
bchalios: So, I think this might be a problem. If there is a frame to "finish" here, that's because we have rate limited it. So, running `finish_frame` on it, doesn't seem right.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -445,6 +478,12 @@ impl Net {
                 }
 
                 error!("net: Could not parse an RX descriptor: {err}");
+                // Notify queue about ready frames. We need this
+                // to bring queue into up to date state. Otherwise
+                // if there is a throttled packet, next call to `add_used`
+                // will owerwrite already written entries in the used_ring
```
bchalios, line 484: `s/owerwrite/overwrite/`

src/vmm/src/devices/virtio/net/device.rs
roypat: Well, even if Rust could at compile time emit an error if the cast of the constant would cause a truncation (which would be neat), there's still the problem that clippy's `cast_possible_truncation` lint is "stupid" in the sense that it doesnt try to do any const evaluation. So even if rustc determines "yes this is fine", clippy would also need to be updated to specifically ignore const-casts that rustc determined are fine.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: we can have a check if there is a rate limited frame and if so, we do `undo_pop` and `break`. This way we will have to parse same broken chain 2 or more times though.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: fixed

src/vmm/src/devices/virtio/net/device.rs
bchalios: I'm not sure what's best here. @roypat WDYT?

ShadowCurse: The other approach is to add this broken chain into the queue same way we do with valid chains with `write_used_element` (we can use `rx_buffer.used_descriptors` to calculate offset.) and add 1 to the `rx_buffer.used_descriptors`.

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ and this project adheres to
 
 ### Added
 
+- [#4834](https://github.com/firecracker-microvm/firecracker/pull/4834): Add
+  RX_MRGBUF support to the `virtio-net` device. When this flag is enabled
+  guest network driver can perform better buffer management which in turn
+  improves RX and TX performance.
```
bchalios, line 16:
```suggestion
- [#4834](https://github.com/firecracker-microvm/firecracker/pull/4834): Add
  support for `VIRTIO_NET_F_MRG_RXBUF` for the `virtio-net` device. When this
  feature is negotiated, guest `virtio-net` driver can perform better memory
  management which in turn improves RX and TX performance.
```



`RX_MRGBUF` is not a valid identifier. Either use, `VIRTIO_NET_F_MRG_RXBUF` or "mergeable buffers". I just saw you used it in the commit message bodies as well. Please fix it there, too.

src/vmm/src/devices/virtio/net/device.rs
bchalios: So, we would do:

```rust
self.write_used_element(self.rx_buffer.used_descriptors, index, 0);
self.rx_buffer_used_descriptors += 1;
```

right? I think this works.

bchalios: maybe worth making the above snippet a function in `RxBuffers`

bchalios: doesn't seem to be working. Things are massively failing in Buildkite

ShadowCurse: now should be fixed

bchalios (approved):

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -556,29 +593,36 @@ impl Net {
                 self.rx_buffer
                     .iovec
                     .write_all_volatile_at(&self.rx_frame_buf[..vnet_hdr_len() + len], 0)?;
-                // SAFETY: This is safe:
+                // SAFETY:
+                // * len will never be bigger that u32::MAX
```
roypat, line 597: nit: this one could use a bit more explanation, e.g. it wont be bigger than 65562 because mmds returns the number of bytes written into `self.rx_frame_buf`, which is an array of size 65562.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -556,29 +593,36 @@ impl Net {
...
+                    self.rx_buffer.mark_used(len, &mut self.queues[RX_INDEX]);
+                }
+                return Ok(Some(len));
             }
         }
 
-        // SAFETY: this is safe because we ensured that `self.rx_buffer` has at least one
-        // DescriptorChain parsed in it.
+        // SAFETY:
+        // * We ensured that `self.rx_buffer` has at least one DescriptorChain parsed in it.
```
roypat, line 612: nit (because unrelated to this PR): I don't think this needs to be unsafe. If we dont have any descriptor chains, we simply panic, or we pass an empty slice to `readv`. Both are fine and well-defined.

roypat (commented):
Do we have any unittests where we explicitly test functionality that's special to mergeable buffers? E.g I see that we extended a bunch of the existing tests to make sure those cases still work for mergeable buffers, but do we have anything where we construct a chain that would be invalid without mergbuf, but can be processed with the feature?

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -184,14 +204,22 @@ impl RxBuffers {
...
-        // instead of logging an error and continuing.
-        rx_queue.add_used(dc.head_index, dc.length).unwrap();
+    fn finish_frame(&mut self, rx_queue: &mut Queue) {
+        rx_queue.advance_used_ring(self.used_descriptors);
+        self.used_descriptors = 0;
+        self.used_bytes = 0;
+    }
+
+    /// Return a slice of iovecs for the first slice in the buffer.
+    /// Panics if there are no parsed descriptor.
```
kalyazin, line 214: s/descriptor/descriptors/

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ and this project adheres to
 
 ### Added
 
+- [#4834](https://github.com/firecracker-microvm/firecracker/pull/4834): Add
+  `VIRTIO_NET_F_RX_MRGBUF` support to the `virtio-net` device. When this feature
+  is negotiated, guest `virtio-net` driver can perform better memory management
```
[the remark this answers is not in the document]
bchalios: What do you have in mind? The performance improvement will depend highly on the machine and the kernel versions.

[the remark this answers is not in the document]
kalyazin: I mean instead of "better" we could expand what the guest driver will do differently if the feature is negotiated. I didn't mean exact perf numbers.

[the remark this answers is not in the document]
bchalios: oh I see. Sorry, I misunderstood the initial comment. Yes that sounds reasonable.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: updated comment

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: fixed

CHANGELOG.md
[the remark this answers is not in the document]
ShadowCurse: updated message with `more efficient` instead of `better`

kalyazin (approved):
doc LGTM

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -1394,14 +1518,86 @@ pub mod tests {
...
+        let mut th = TestHelper::get_default(&mem);
+        // VIRTIO_NET_F_MRG_RXBUF is not enabled by default
+        th.net().acked_features = 1 << VIRTIO_NET_F_MRG_RXBUF;
+        rx_multiple_frames(th);
+    }
+
+    fn rx_mrg_rxbuf_only(mut th: TestHelper) {
+        th.activate_net();
+
+        // Create 2 valid Rx avail descriptor chains. Each one is 32K so
```
bchalios, line 1550: They are not 32K.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -1394,14 +1518,86 @@ pub mod tests {
...
+        // network without VIRTIO_NET_F_MRG_RXBUF enabled should not
+        // process 64K packet, but version with it enabled should.
```
bchalios, line 1552: You are not passing a 64K packet.

bchalios (changes requested):

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: this were remains at my initial attempts. Apparently, trying to sent big packet with `inject_tap_tx_frame` will fail.
Fixed now.

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: same

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -1394,14 +1518,85 @@ pub mod tests {
...
+        // be at least 64K to pass the capacity check for rx_buffers.
+        // First chain is intentionally small, so non VIRTIO_NET_F_MRG_RXBUF
+        // version will skip it.
+        th.add_desc_chain(NetQueue::Rx, 0, &[(0, 500, VIRTQ_DESC_F_WRITE)]);
+        th.add_desc_chain(
+            NetQueue::Rx,
+            1000,
+            &[(1, MAX_BUFFER_SIZE as u32, VIRTQ_DESC_F_WRITE)],
+        );
+        // Inject 2 frames to tap and run epoll.
```
bchalios, line 1560: 1 frame

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -1394,14 +1518,85 @@ pub mod tests {
...
+        let mut frame = inject_tap_tx_frame(&th.net(), 1000);
+        check_metric_after_block!(
+            th.net().metrics.rx_packets_count,
+            1,
+            th.event_manager.run_with_timeout(100).unwrap()
+        );
+
+        // Check that the frames weren't deferred.
```
bchalios, line 1568: s/the frames/the frame/

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -1394,14 +1518,85 @@ pub mod tests {
...
+        assert!(th.net().rx_buffer.used_bytes == 0);
+        // Check that the used queue has advanced.
+        assert_eq!(th.rxq.used.idx.get(), 2);
+        assert!(&th.net().irq_trigger.has_pending_irq(IrqType::Vring));
+        // Check that the 1st frame was written successfully to the 1st Rx descriptor chain.
```
bchalios, line 1573: s/the 1st frame/the frame/

bchalios (approved):

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: fixed

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: fixed

src/vmm/src/devices/virtio/net/device.rs
ShadowCurse: fixed

kalyazin (approved):
doc LGTM

PR #4847, opened 2024-10-11
fix(ci): epoll on pidfd to wait for Firecracker exit

bchalios:
## Changes

Substitute this mechanism with calling epoll() on the pidfd of the process instead. This should deterministically block until the process exits. If there's something else wrong, we will hit the pytest timeout.

## Reason

Currently, we use psutil.pid_exists in a loop with a timeout of 10 seconds. This is racy and indeed some times we hit it in our CI.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [x] Commits meet

### DIFF
--- a/tests/framework/utils.py
+++ b/tests/framework/utils.py
@@ -7,6 +7,7 @@ import logging
 import os
 import platform
 import re
+import select
 import signal
 import stat
 import subprocess
@@ -450,16 +451,39 @@ def run_guest_cmd(ssh_connection, cmd, expected, use_json=False):
     assert stdout == expected
 
 
-@retry(wait=wait_fixed(1), stop=stop_after_attempt(10), reraise=True)
+def get_process_pidfd(pid):
+    """Get a pidfd file descriptor for the process with PID `pid`
+
+    Will return a pid file descriptor for the process with PID `pid` if it is
+    still alive. If the process has already exited it will return `None`.
+
+    Any other error while calling the system call, will raise an OSError
+    exception.
+    """
+    try:
+        pidfd = os.pidfd_open(pid)
+    except ProcessLookupError:
+        return None
+
+    return pidfd
+
+
 def wait_process_termination(p_pid):
     """Wait for a process to terminate.
 
-    Will return sucessfully if the process
+    Will return successfully if the process
     got indeed killed or raises an exception if the process
     is still alive after retrying several times.
     """
-    if psutil.pid_exists(p_pid):
-        raise Exception(f"[{p_pid}] process is still alive")
+    pidfd = get_process_pidfd(p_pid)
+
+    # If pidfd is None the process has already terminated
+    if pidfd is not None:
+        epoll = select.epoll()
+        epoll.register(pidfd, select.EPOLLIN)
+        # This will return once the process exits
+        epoll.poll()
+        os.close(pidfd)
 
 
 def get_firecracker_version_from_toml():


tests/framework/utils.py
```diff
@@ -450,16 +451,38 @@ def run_guest_cmd(ssh_connection, cmd, expected, use_json=False):
...
     is still alive after retrying several times.
     """
-    if psutil.pid_exists(p_pid):
-        raise Exception(f"[{p_pid}] process is still alive")
+    pidfd = get_process_pidfd(p_pid)
+
+    # If pidfd is None the process has already terminated
+    if pidfd is not None:
+        epoll = select.epoll()
+        epoll.register(pidfd, select.EPOLLIN)
```
kalyazin, line 483: if the process goes between `get_process_pidfd` and `register`, will we still receive the `EPOLLIN`?

bchalios: I _think_ so. IIUC, if `pidfd_open` succeeds it opens a pidfd file and installs the file descriptor in the caller's file descriptor table. This means that we can poll on it until we close it.

Also, when we `poll` on it, we end up doing this: https://elixir.bootlin.com/linux/v6.11.3/source/fs/pidfs.c#L109, which checks for the `struct task` that corresponds to the `pid` that corresponds to the `pidfd`. If this is NULL, the system call returns `EPOLLIN` directly.

bchalios: I've changed this to use `select.select` directly

ShadowCurse (dismissed):

bchalios:
@ShadowCurse I had to revert back to using `epoll()`. `select()` hits the `FD_SETSIZE` issue (1024 descriptors only).

ShadowCurse:
> @ShadowCurse I had to revert back to using `epoll()`. `select()` hits the `FD_SETSIZE` issue (1024 descriptors only).

How can it hit the limit if you are waiting for 1 fd?

bchalios:
> How can it hit the limit if you are waiting for 1 fd?

It's not about the numbers of PIDs you're waiting on. It's about the maximum PID value it can handle. Reading from `man select(2)`:

>        select() can monitor only file descriptors numbers that are less than FD_SETSIZE; poll(2) and epoll(7) do not have this limitation.  See BUGS.

roypat:
Since we only use `utils.wait_for_termination` in two locations, and both of those are in `Microvm`, can we move this function into `Microvm` and create the `pidfd` already when we spawn the Firecracker process? Otherwise, how does this behave if the Firecracker process dies unexpectedly and the pid gets freed significantly before we get to the `wait_for_termination` call? :thinking:

bchalios:
> Otherwise, how does this behave if the Firecracker process dies unexpectedly and the pid gets freed significantly before we get to the wait_for_termination call? 🤔

If the process with pid `PID` is already dead when creating the `pidfd`, then the system call will raise a `ProcessLookupError` exception and `get_process_pidfd` will return `None`, so `wait_for_termination` will return immediately.

> can we move this function into Microvm and create the pidfd already when we spawn the Firecracker process?

I thought of doing that, but I think it makes the error handling much more complicated.

ShadowCurse:
Ok, but to avoid issue with 2 calls: `register` and `epoll`, maybe we should use simple `poll`?

bchalios:
`select.poll()` returns you a "polling" object to which you register file descriptors.

roypat (approved):

ShadowCurse (approved):

kalyazin (approved):

PR #4862, opened 2024-10-21
Minor test fixes to improve debuggability and readability

roypat:
## Changes

- more debug output if running SSH commands in the guest fails
- relax timeouts
- minor code cleanup

## Reason

- more debug information when SSH failures, and less spurious timeout failures

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/tests/framework/utils_vsock.py
+++ b/tests/framework/utils_vsock.py
@@ -180,7 +180,7 @@ def check_guest_connections(vm, server_port_path, blob_path, blob_hash):
     cmd += "  ({})& ".format(worker_cmd)
     cmd += '  workers="$workers $!";'
     cmd += "done;"
-    cmd += "for w in $workers; do wait $w || exit -1; done"
+    cmd += "for w in $workers; do wait $w || exit 1; done"
 
     ecode, _, stderr = vm.ssh.run(cmd)
     echo_server.terminate()
--- a/tests/framework/utils_vsock.py
+++ b/tests/framework/utils_vsock.py
@@ -180,7 +180,7 @@ def check_guest_connections(vm, server_port_path, blob_path, blob_hash):
     cmd += "  ({})& ".format(worker_cmd)
     cmd += '  workers="$workers $!";'
     cmd += "done;"
-    cmd += "for w in $workers; do wait $w || exit 1; done"
+    cmd += "for w in $workers; do wait $w || (wait; exit 1); done"
 
     ecode, _, stderr = vm.ssh.run(cmd)
     echo_server.terminate()
--- a/tests/host_tools/network.py
+++ b/tests/host_tools/network.py
@@ -103,7 +103,7 @@ class SSHConnection:
         We'll keep trying to execute a remote command that can't fail
         (`/bin/true`), until we get a successful (0) exit code.
         """
-        self.check_output("true", timeout=10, debug=True)
+        self.check_output("true", timeout=100, debug=True)
 
     def run(self, cmd_string, timeout=None, *, check=False, debug=False):
         """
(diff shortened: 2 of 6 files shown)


tests/host_tools/network.py
```diff
@@ -52,7 +54,17 @@ def __init__(self, netns, ssh_key: Path, host, user):
...
         ]
 
-        self._init_connection()
+        # _init_connection loops until it can connect to the guest
+        # dumping debug state on every iteration is not useful or wanted, so
+        # only dump it once if _all_ iterations fail.
+        try:
+            self._init_connection()
+        except Exception as exc:
+            on_error(exc)
```
pb8o, line 63: what if `on_error` is `None`?

We could define `on_error` as `lambda x: print(x)`

pb8o, line 63: Shouldn't this be already be handled in `_exec`, so we are handled it here a 2nd time?

tests/framework/utils_vsock.py
```diff
@@ -180,7 +180,7 @@ def check_guest_connections(vm, server_port_path, blob_path, blob_hash):
     cmd += "  ({})& ".format(worker_cmd)
     cmd += '  workers="$workers $!";'
     cmd += "done;"
-    cmd += "for w in $workers; do wait $w || exit 1; done"
+    cmd += "for w in $workers; do wait $w || (wait; exit 1); done"
```
pb8o, line 183: could we just replace this with a single `wait || exit 1`

pb8o: Actually it looks like `wait` just returns 0 with no indication if any failed.

An alternative may be https://unix.stackexchange.com/questions/742036/using-wait-bash-posix-and-fail-if-one-process-fails-in-a-script

tests/host_tools/network.py
roypat: argh, I handled the none case below, but forgot here. Thanks!

If we define it as `lambda x: print(x)` we'll end up just printing the exception twice, without attaching any new information to it, so I'd prefer to not do that (although `on_error` is always set in current usecases anyway, so probably it does not matter).

roypat: For `_init_connection`, inside of `_exec` the `_on_error` field is still `None` because the constructor only sets it after `_init_connection`. This is to avoid logging a ton of stuff on every retry.

tests/framework/utils_vsock.py
roypat: Yeah, `wait` on its own just returns the status code of the last child process that exits, or sth like that. 

> An alternative may be https://unix.stackexchange.com/questions/742036/using-wait-bash-posix-and-fail-if-one-process-fails-in-a-script

tests/integration_tests/functional/test_snapshot_basic.py
```diff
@@ -183,20 +182,20 @@ def test_5_snapshots(
...
-        snapshot = microvm.make_snapshot(snapshot_type)
+        new_snapshot = microvm.make_snapshot(snapshot_type)
 
         # If we are testing incremental snapshots we must merge the base with
         # current layer.
         if snapshot.is_diff:
-            logger.info("Base: %s, Layer: %s", base_snapshot.mem, snapshot.mem)
-            snapshot = snapshot.rebase_snapshot(
-                base_snapshot, use_snapshot_editor=use_snapshot_editor
+            logger.info("Base: %s, Layer: %s", snapshot.mem, snapshot.mem)
```
kalyazin, line 190: did we mean to print the same value twice?

roypat: absolutely not, thanks for catching!

kalyazin (approved):

pb8o (approved):

PR #4864, opened 2024-10-22
Start GDB session with API

JackThomson2:
Allow a GDB debugging session to be started over API

## Changes

Moved the GDB address config inside the machine-config section.

This allows a user to start a GDB debugging session using the API and are not limited to just using the config file

## Reason

Allow for GDB to be setup when using config file is not possible

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ and this project adheres to
 - [#4869](https://github.com/firecracker-microvm/firecracker/pull/4869): Added
   support for Aarch64 systems which feature CPU caches with a number of sets
   higher than `u16::MAX`.
+- [#4797](https://github.com/firecracker-microvm/firecracker/pull/4797),
+  [#4854](https://github.com/firecracker-microvm/firecracker/pull/4854): Added
+  GDB debugging support for a microVM guest kernel. Please see our
+  [GDB debugging documentation](docs/gdb-debugging.md) for more information.
 
 ### Changed
 
--- a/docs/gdb-debugging.md
+++ b/docs/gdb-debugging.md
@@ -6,8 +6,6 @@ Firecracker supports debugging the guest kernel via GDB remote serial protocol.
 This allows us to connect GDB to the firecracker process and step through debug
 the guest kernel.
 
-The GDB feature requires Firecracker to be booted with a config file.
-
 ## Prerequisites
 
 Firstly, to enable GDB debugging we need to compile Firecracker with the `gdb`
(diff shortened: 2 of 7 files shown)


src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -111,6 +111,10 @@ pub struct MachineConfig {
     /// Configures what page size Firecracker should use to back guest memory.
     #[serde(default)]
     pub huge_pages: HugePageConfig,
+    /// GDB socket address.
+    #[cfg(feature = "gdb")]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub gdb_socket_address: Option<String>,
```
kalyazin, line 117: I think we can call this `path` rather than `address` for consistency with other APIs. Also, we need to update https://github.com/firecracker-microvm/firecracker/blob/main/src/firecracker/swagger/firecracker.yaml to reflect this change.

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to
   `VIRTIO_NET_F_RX_MRGBUF` support to the `virtio-net` device. When this feature
   is negotiated, guest `virtio-net` driver can perform more efficient memory
   management which in turn improves RX and TX performance.
+- [#4797](https://github.com/firecracker-microvm/firecracker/pull/4797),
```
kalyazin, line 17: can provide a link to the doc here

src/vmm/src/vmm_config/machine_config.rs
JackThomson2: Much better thanks, I've updated

docs/gdb-debugging.md
```diff
@@ -25,22 +23,35 @@ debugging to work. The key config options to enable are:
 
 ```
 CONFIG_FRAME_POINTER=y
-CONFIG_KGDB=y
-CONFIG_KGDB_SERIAL_CONSOLE=y
```
pb8o, line 29: are these really not needed anymore?

JackThomson2: Yeah not required with this setup

src/firecracker/swagger/firecracker.yaml
```diff
@@ -1028,6 +1028,9 @@ definitions:
     properties:
       cpu_template:
         $ref: "#/definitions/CpuTemplate"
+      gdb_socket_path:
+        type: string
+        description: Path to the GDB socket. Requires the gdb feature to be enabled.
```
pb8o, line 1033: Can this break customers that use Swagger to autogenerate the client? I.e. I guess if it's not specified it won't be included in the payload, but I wonder if adding it here is more risky than it's worth.

pb8o: Yeah the scenario I see is, some auto-generator could start including this (optional) field in all request (maybe with value `null`?) and then Firecracker fails because it does not recognize the option.

I think it's farfetched as that would not be a very sane client, but at the same time, I'm not sure there's a lot of value exposing it in the Swagger definition? Then again, a customer could just remove this snippet if it causes issues.

kalyazin: yeah, I can see that happening. Ok, maybe we can just add this as a comment  in yaml instead?

pb8o: Yeah, I am good with either solution. Worst case we have the workaround of commenting a few lines when generating the client.

pb8o (approved):

kalyazin (approved):

src/firecracker/swagger/firecracker.yaml
JackThomson2: I've commented it out for now to be on the safer side

pb8o (approved):

kalyazin (approved):

PR #4884, opened 2024-10-30
Guest vs host features check

ShadowCurse:
## Changes
Add tests comparing host and guest default cpu features.

## Reason
We want to know what is the difference between cpu features on the host and in the guest.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  PR.
- [ ] Commits meet


tests/integration_tests/functional/test_cpu_features_aarch64.py
```diff
@@ -63,6 +64,37 @@ def get_cpu_template_dir(cpu_template):
...
+    """Check CPU features host vs guest"""
+
+    vm = uvm_nano
+    vm.add_net_iface()
+    vm.start()
+    host_feats = set(utils.check_output(CPU_FEATURES_CMD).stdout.strip().split(" "))
+    guest_feats = set(vm.ssh.check_output(CPU_FEATURES_CMD).stdout.strip().split(" "))
+
+    cpu_model = cpuid_utils.get_cpu_model_name()
+    match cpu_model:
```
pb8o, line 81: nit: I think using match here is overkill, but it could grow more complex in the future.

tests/integration_tests/functional/test_cpu_features_aarch64.py
```diff
@@ -63,6 +64,37 @@ def get_cpu_template_dir(cpu_template):
...
+        case CpuModel.ARM_NEOVERSE_N1:
+            assert host_feats - guest_feats == set()
+            assert guest_feats - host_feats == {"ssbs"}
+        case CpuModel.ARM_NEOVERSE_V1:
+            assert host_feats - guest_feats == {
+                "pacg",
+                "sve",
+                "paca",
+                "svebf16",
+                "svei8mm",
```
pb8o, line 91: I'd sort this for visual grouping.

tests/integration_tests/functional/test_cpu_features_aarch64.py
```diff
@@ -63,6 +64,37 @@ def get_cpu_template_dir(cpu_template):
     return cpu_template if cpu_template else "none"
 
 
+@pytest.mark.skipif(
+    PLATFORM != "aarch64",
+    reason="This is aarch64 specific test.",
+)
```
pb8o, line 70: We could make this test more generic and support x86_64. Maybe it belongs in a generic `test_cpu_features.py` (and move the current `test_cpu_features.py` to `test_cpu_features_x86.py`)

tests/framework/utils_cpuid.py
```diff
@@ -10,6 +10,8 @@
 from framework.utils import check_output
 from framework.utils_imdsv2 import imdsv2_get
 
+CPU_FEATURES_CMD = r"lscpu |grep -oP '^Flags:\s+\K.+'"
```
pb8o, line 13: Only used in one test, so I don't think it needs to be here.

tests/integration_tests/functional/test_cpu_features_aarch64.py
ShadowCurse: I would prefer to keep our current separation where `test_cpu_feautres` is for x86 and `test_cpu_features_aarch64` is for aarch64. No reason for creating some new file for a singe test.
I don't mind renaming `test_cpu_features` to `test_cpu_features_x86` though.

tests/framework/utils_cpuid.py
ShadowCurse: Idea is that it might be used in x86 version of this test. I did not implement it here though.

tests/integration_tests/functional/test_cpu_features_aarch64.py
ShadowCurse: I think match on an enum is better than if/else combos

tests/integration_tests/functional/test_cpu_features_aarch64.py
ShadowCurse: done

tests/integration_tests/functional/test_cpu_features_aarch64.py
```diff
@@ -63,6 +64,37 @@ def get_cpu_template_dir(cpu_template):
...
+            assert host_feats - guest_feats == {
+                "pacg",
+                "sve",
+                "paca",
+                "svebf16",
+                "svei8mm",
+            }
+            assert guest_feats - host_feats == {"ssbs"}
+        case _:
+            assert False, f"Cpu model {cpu_model} is not supported"
```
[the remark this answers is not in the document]
ShadowCurse: what is the problem with this test failing on an instance for which we don't have cpu feature diff?

tests/framework/utils_cpuid.py
pb8o: In that case we can wait until that happens, but in general this test is simple enough that could work for both architectures.

tests/integration_tests/functional/test_cpu_features_aarch64.py
pb8o: On the other hand, we should look for any opportunities to consolidate and unify those tests. I think this test could be the first step in that direction.

ShadowCurse: These are 2 different architectures, we don't need to unify tests for them.

tests/framework/utils_cpuid.py
ShadowCurse: I added x86_64 test as well, so this global constant is now used in 2 places.

tests/integration_tests/functional/test_cpu_features_aarch64.py
```diff
@@ -63,6 +64,37 @@ def get_cpu_template_dir(cpu_template):
...
+    vm.add_net_iface()
+    vm.start()
+    host_feats = set(utils.check_output(CPU_FEATURES_CMD).stdout.strip().split(" "))
+    guest_feats = set(vm.ssh.check_output(CPU_FEATURES_CMD).stdout.strip().split(" "))
+
+    cpu_model = cpuid_utils.get_cpu_model_name()
+    match cpu_model:
+        case CpuModel.ARM_NEOVERSE_N1:
+            assert host_feats - guest_feats == set()
+            assert guest_feats - host_feats == {"ssbs"}
```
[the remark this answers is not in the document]
ShadowCurse: added notes to the aarch64 test. Not sure how to describe the diffs for x86_64 test though.

tests/integration_tests/functional/test_cpu_features_aarch64.py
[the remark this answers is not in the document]
roypat: I don't particularly care about it failing in CI on an instance that we don't support, but I don't think it's very nice for external contributors if an integration test always fails on them if they want to run it locally.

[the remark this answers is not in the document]
ShadowCurse: ok, fair, added this check

tests/integration_tests/functional/test_cpu_features_aarch64.py
pb8o: I disagree. If the test is the same, we don't need to separate tests just because of different architecture. It's less effort to maintain and fix one test than two. Some CPU tests may be very specific and apply to only one architecture, but that's not a good argument to duplicate some tests.

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -203,6 +205,272 @@ def test_brand_string(uvm_plain_any):
...
+                "pausefilter",
+                "pfthreshold",
+                "sme",
+                "tsc_scale",
+                "v_vmsave_vmload",
+                "vgif",
+                "vmcb_clean",
+            } | {"brs", "rapl", "v_spec_ctrl"}
+
+            if version.parse(utils.get_kernel_version()) < version.parse("6.1"):
```
pb8o, line 280: Use `global_props.host_linux_version_tpl < (6, 1)`

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -203,6 +205,272 @@ def test_brand_string(uvm_plain_any):
...
+                "tsc_known_freq",
+                "umip",
+            }
+        case CpuModel.INTEL_ICELAKE:
+            host_guest_diff_5_10 = {
+                "dtes64",
+                "hwp_act_window",
+                "pdcm",
+                "acpi",
+                "aperfmperf",
```
pb8o, line 407: SL/CL/IL have some commonalities. Can we extract those to make the diff smaller?

tests/integration_tests/functional/test_cpu_features.py
```diff
@@ -16,12 +16,14 @@
 from pathlib import Path
```
pb8o, line 16: Thanks for doing x86 too! The commit message has a typo and probably should be labelled `test`

`test: x86_64: add host vs guest cpu feature test`

tests/integration_tests/functional/test_cpu_features_aarch64.py
ShadowCurse: > It's less effort to maintain and fix one test than two.

tests/integration_tests/functional/test_cpu_features.py
ShadowCurse: fixed

tests/integration_tests/functional/test_cpu_features.py
ShadowCurse: I think this will only make it harder to maintain as features will be split into common groups spread across the file. Also optimizing for the diff size is not very useful.

tests/integration_tests/functional/test_cpu_features_x86_64.py
pb8o, line 1: Thanks! The rename needs to be reflected in `.buildkite/pipeline_cpu_template.py`

tests/integration_tests/functional/test_cpu_features_aarch64.py
pb8o: To me number of files is not that important. As I see it, it's better to have 1 new file with only one test that handles all architectures, than 2 tests in existing files.

But anyway, I do have some unrelated changes in these files so I can do this follow up work then.

tests/integration_tests/functional/test_cpu_features.py
ShadowCurse: done

tests/integration_tests/functional/test_cpu_features_aarch64.py
```diff
@@ -63,6 +64,43 @@ def get_cpu_template_dir(cpu_template):
...
+            # and Firecracker does not enable them either.
+            assert host_feats - guest_feats == {
+                "paca",
+                "pacg",
+                "sve",
+                "svebf16",
+                "svei8mm",
+            }
+            # kernel should hide this feature, but our guest kernel
+            # is not recent enough for this.
```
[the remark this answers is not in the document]
ShadowCurse: added a link to the patch

tests/integration_tests/functional/test_cpu_features_x86_64.py
ShadowCurse: updated

kalyazin (approved):

tests/integration_tests/functional/test_cpu_features_aarch64.py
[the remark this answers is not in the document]
pb8o: So v6.10+. Would be nice to add that

pb8o (approved):

PR #4913, opened 2024-11-13
snapshot: Remove max_connections and max_pending_resets fields

zulinx86:
## Changes

Remove `max_connections` and `max_pending_resets` fields from snapshot.

**Note that I only focused on removing them in this PR although I might open another PR to refactor more later.**

## Reason

`TcpIPv4Handler` for MMDS network stack preallocates several buffers
whose sizes are saved into a snapshot as `max_connections` and
`max_pending_resets` in `MmdsNetworkStackState`. But they are always the
same constant hardcoded values (`DEFAULT_MAX_CONNECTIONS` and
`DEFAULT_MAX_PENDING_RESETS`) as of today, which means there is no need
to save them into a snapshot. Even if we change the hardcoded sizes
across Firecracker versions, that should not be a problem. This is
because the snapshot feature does not support migration of network
connections and those buffers are initialized with empty on snapshot
restoration. When migrating from a Firecracker version with larger
buffers to another version with smaller ones, guest workloads that
worked previously might start to fail due to the less buffer spaces.
However, the issue is not a problem of the snapshot feature and it
should also occur even on a purely booted microVM (not restored from a
snapshot). Thus, it is fine to remove those fields from a snapshot.

Since this is a breaking change of the snapshot format, bumps the major
version.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.~~
  integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to
 
 ### Changed
 
+- [#4913](https://github.com/firecracker-microvm/firecracker/pull/4913): Removed
+  unnecessary fields (`max_connections` and `max_pending_resets`) from the
+  snapshot format, bumping the snapshot version to 5.0.0.
+
 ### Deprecated
 
 ### Removed
--- a/src/vmm/src/mmds/ns.rs
+++ b/src/vmm/src/mmds/ns.rs
@@ -553,14 +553,8 @@ mod tests {
         let ip = Ipv4Addr::from(DEFAULT_IPV4_ADDR);
         let other_ip = Ipv4Addr::new(5, 6, 7, 8);
         let mac = MacAddr::from_bytes_unchecked(&[0; 6]);
-        let mut ns = MmdsNetworkStack::new(
-            mac,
-            ip,
-            DEFAULT_TCP_PORT,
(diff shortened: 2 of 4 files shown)


roypat (approved):

roypat:
>  might start to fail due to the less buffer spaces.

Not even this can happen I think, because its just the capacity of the data structures. If they reach that capacity, then they'll simply grow :D

zulinx86:
> > might start to fail due to the less buffer spaces.
> 
> Not even this can happen I think, because its just the capacity of the data structures. If they reach that capacity, then they'll simply grow :D

yeah, I'm on the same page. I tried to make excuses at best I could :P

pb8o (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to
 
 ### Changed
 
+- [#4913](https://github.com/firecracker-microvm/firecracker/pull/4913): Removed
```
kalyazin, line 15: Not opposing the change, but what are customers supposed to do in response to this?

zulinx86: Good point, they need to regenerate a snapshot. We might want to apply the same fix for the previous change.

PR #4916, opened 2024-11-15
Fix 16K host page support

ShadowCurse:
## Changes
Separate `PAGE_SIZE` constant into 2: `GUEST_PAGE_SIZE` and `HOST_PAGE_SIZE`.
`HOST_PAGE_SIZE` has a default of 4K, but is also queried at runtime to support other page sizes.

The only structure which did not work with non default 4K pages is `IovDeque`. This PR updates it
and removes restrictions on the queue size and on the host page size.

Tested16K pages support on rp5 with 16K page kernel.

## Reason
Previously Firecracker was working on hosts with non 4K pages, but with 1.10 version we introduced `IovDeque` type
which assumed the host page size to be 4K.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ and this project adheres to
 
 - [#4921](https://github.com/firecracker-microvm/firecracker/pull/4921): Fixed
   swagger `CpuConfig` definition to include missing aarch64-specific fields.
+- [#4916](https://github.com/firecracker-microvm/firecracker/pull/4916): Fixed
+  `IovDeque` implementation to work with any host page size. This fixes
+  virtio-net device on non 4K host kernels.
 
 ## [1.10.1]
 
--- a/src/vmm/src/arch/aarch64/mod.rs
+++ b/src/vmm/src/arch/aarch64/mod.rs
@@ -94,7 +94,8 @@ pub fn initrd_load_addr(
     guest_mem: &GuestMemoryMmap,
     initrd_size: usize,
 ) -> Result<u64, ConfigurationError> {
-    let round_to_pagesize = |size| (size + (super::PAGE_SIZE - 1)) & !(super::PAGE_SIZE - 1);
+    let round_to_pagesize =
+        |size| (size + (super::GUEST_PAGE_SIZE - 1)) & !(super::GUEST_PAGE_SIZE - 1);
     match GuestAddress(get_fdt_addr(guest_mem)).checked_sub(round_to_pagesize(initrd_size) as u64) {
         Some(offset) => {
             if guest_mem.address_in_range(offset) {
--- a/src/vmm/src/arch/mod.rs
+++ b/src/vmm/src/arch/mod.rs
@@ -2,7 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::fmt;
+use std::sync::LazyLock;
 
+use log::warn;
 use serde::{Deserialize, Serialize};
 
 /// Module for aarch64 related functionality.
@@ -55,8 +57,20 @@ pub struct InitrdConfig {
 /// Default page size for the guest OS.
 pub const GUEST_PAGE_SIZE: usize = 4096;
 
-/// Default page size for the host OS.
-pub const HOST_PAGE_SIZE: usize = 4096;
+/// Get the size of the host page size.
+pub fn host_page_size() -> usize {
+    /// Default page size for the host OS.
+    static PAGE_SIZE: LazyLock<usize> = LazyLock::new(|| {
+        // # Safety: Value always valid
+        let r = unsafe { libc::sysconf(libc::_SC_PAGESIZE) };
+        usize::try_from(r).unwrap_or_else(|_| {
+            warn!("Could not get host page size with sysconf, assuming default 4K host pages");
+            4096
+        })
+    });
+
+    *PAGE_SIZE
+}
 
 impl fmt::Display for DeviceType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
--- a/src/vmm/src/arch/mod.rs
+++ b/src/vmm/src/arch/mod.rs
@@ -52,8 +52,11 @@ pub struct InitrdConfig {
     pub size: usize,
 }
 
-/// Default (smallest) memory page size for the supported architectures.
-pub const PAGE_SIZE: usize = 4096;
+/// Default page size for the guest OS.
+pub const GUEST_PAGE_SIZE: usize = 4096;
+
+/// Default page size for the host OS.
+pub const HOST_PAGE_SIZE: usize = 4096;
 
 impl fmt::Display for DeviceType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
--- a/src/vmm/src/devices/virtio/iov_deque.rs
+++ b/src/vmm/src/devices/virtio/iov_deque.rs
@@ -77,10 +77,7 @@ pub enum IovDequeError {
 //    pub iov_len: ::size_t,
 // }
 // ```
-//
-// This value must be a multiple of 256 because this is the maximum number of `iovec` can fit into
-// 1 memory page: 256 * sizeof(iovec) == 4096 == HOST_PAGE_SIZE. IovDeque only operates with
-// `HOST_PAGE_SIZE` granularity.
+
 #[derive(Debug)]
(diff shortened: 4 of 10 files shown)


src/vmm/src/arch/mod.rs
```diff
@@ -56,7 +57,23 @@ pub struct InitrdConfig {
 pub const GUEST_PAGE_SIZE: usize = 4096;
 
 /// Default page size for the host OS.
-pub const HOST_PAGE_SIZE: usize = 4096;
+pub static mut HOST_PAGE_SIZE: usize = 4096;
+
+/// Updates the HOST_PAGE_SIZE global variable to the output of
```
pb8o, line 62: Instead of modifying a global, could we make it a function that returns the pagesize? Or use `lazy_static!`.

ShadowCurse: added a method to return the host page size.

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -77,13 +77,11 @@ pub enum IovDequeError {
...
 // ```
-//
-// This value must be a multiple of 256 because this is the maximum number of `iovec` can fit into
-// 1 memory page: 256 * sizeof(iovec) == 4096 == HOST_PAGE_SIZE. IovDeque only operates with
-// `HOST_PAGE_SIZE` granularity.
+
 #[derive(Debug)]
 pub struct IovDeque<const L: u16> {
     pub iov: *mut libc::iovec,
+    pub bytes: u32,
```
kalyazin, line 84: I suppose this should be mentioned in the changelog.

ShadowCurse: added a note in the Fixed section

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -79,8 +79,8 @@ pub enum IovDequeError {
 // ```
 //
 // This value must be a multiple of 256 because this is the maximum number of `iovec` can fit into
-// 1 memory page: 256 * sizeof(iovec) == 4096 == PAGE_SIZE. IovDeque only operates with `PAGE_SIZE`
```
kalyazin, line 82: we don't claim PAGE_SIZE == 4096 anymore, do we?

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to
 
 - [#4921](https://github.com/firecracker-microvm/firecracker/pull/4921): Fixed
   swagger `CpuConfig` definition to include missing aarch64-specific fields.
+- [#4916](https://github.com/firecracker-microvm/firecracker/pull/4916): Fix
```
kalyazin, line 28: we seem to use past tense in those

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to
...
+  `IovDeque` implementation to work with any host page size.
```
kalyazin, line 29: Could mention which devices were affected.

src/vmm/src/arch/aarch64/mod.rs
```diff
@@ -94,7 +94,8 @@ pub fn initrd_load_addr(
     guest_mem: &GuestMemoryMmap,
     initrd_size: usize,
 ) -> Result<u64, ConfigurationError> {
-    let round_to_pagesize = |size| (size + (super::PAGE_SIZE - 1)) & !(super::PAGE_SIZE - 1);
+    let round_to_pagesize =
```
kalyazin, line 97: commit message: s/repalce/replace/

src/vmm/src/arch/mod.rs
```diff
@@ -56,7 +57,29 @@ pub struct InitrdConfig {
 pub const GUEST_PAGE_SIZE: usize = 4096;
 
 /// Default page size for the host OS.
-pub const HOST_PAGE_SIZE: usize = 4096;
+static mut HOST_PAGE_SIZE: usize = 4096;
+
+/// Updates the HOST_PAGE_SIZE global variable to the output of
+/// sysconf(_SC_PAGESIZE). If call is unsuccessfull the default
```
kalyazin, line 63: s/unsuccessfull/unsuccessful/

src/vmm/src/devices/virtio/iov_deque.rs
ShadowCurse: The type now can handle any host page size, so we don't need to mention 4096 pages anymore.

CHANGELOG.md
ShadowCurse: fixed

CHANGELOG.md
ShadowCurse: added

src/vmm/src/arch/aarch64/mod.rs
ShadowCurse: fixed

src/vmm/src/arch/mod.rs
ShadowCurse: fixed

kalyazin (dismissed):

src/vmm/src/arch/mod.rs
roypat: Seems like we never added this method. Why? If you don't want to do a syscall in the initializer for IoVecDeque, you can just use a `LazyLock` or some other variant of `OnceCell` here

src/vmm/src/devices/virtio/iov_deque.rs
```diff
@@ -153,16 +147,21 @@ impl<const L: u16> IovDeque<L> {
 
     /// Create a new [`IovDeque`] that can hold memory described by a single VirtIO queue.
     pub fn new() -> Result<Self, IovDequeError> {
-        let memfd = Self::create_memfd()?;
+        let host_page_size = host_page_size();
+        let bytes = L as usize * std::mem::size_of::<iovec>();
+        let num_host_pages = bytes.div_ceil(host_page_size);
+        let pages_bytes = num_host_pages * host_page_size;
```
roypat, line 153: nit: this computation could be moved into a private function, so we dont duplicate it in `drop`

src/vmm/src/arch/mod.rs
ShadowCurse: There is an `update_host_page_size` method just bellow.

CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ and this project adheres to
 
 ### Fixed
 
+- [#4916](https://github.com/firecracker-microvm/firecracker/pull/4916): Fix
+  `IovDeque` implementation to work with any host page size.
```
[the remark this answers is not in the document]
ShadowCurse: Not sure it can be called a bug when we explicitly designed for 4K pages at the implementation time. Also as I understand, if no net devices were used, 16K pages should work fine.

src/vmm/src/devices/virtio/iov_deque.rs
ShadowCurse: done

src/vmm/src/arch/mod.rs
roypat: I think Pablo's point was to use a function that returns the host page size via a call to sysconf each time, eliminating the mutable static (at least that's how I understood it. If not, then it is very much my point though xP). If that's not acceptable for performance reasons, then please use one of the stdlib types that safely abstract this pattern away, such as `LazyLock`:

```
pub static HOST_PAGE_SIZE: LazyLock<usize> = LazyLock::new(|| {
    let r = unsafe { libc::sysconf(libc::_SC_PAGESIZE) };
    if r < 0 {
        warn!("Could not get host page size with sysconf, assuming default 4K host pages");
        4096
    } else {
        r as usize
    }
});
```

ShadowCurse: LazyLock needs unstable `lazy_cell` feature. Updated to use `Once` instead.

ShadowCurse: now with new Rust version `OnceLock` is available and I adjusted the implementation to use it.

roypat (approved):

src/vmm/src/arch/mod.rs
roypat: ah right. forgot that's what we were waiting for :rofl: thanks!

kalyazin (approved):

PR #4926, opened 2024-11-25
Update seccompiler to use libseccomp

ShadowCurse:
## Changes
Replace our custom implementation of a seccompiler with [libseccomp](https://github.com/seccomp/libseccomp).
By out test, `libseccomp` produces ~65% smaller binaries which is very beneficial as we embed the combination of those into the Firecracker.

In order to interact with `libseccomp` we add custom bindings which contain only the needed set of methods, constants needed for our use case. This simplifies the maintenance and avoids adding dependencies.

## Reason
`libseccomp` provides better quality compiler for BPF seccomp programs than our current implementation.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.


Cargo.toml
```diff
@@ -22,6 +22,14 @@ tests_outside_test_module = "warn"
 assertions_on_result_states = "warn"
 error_impl_error = "warn"
 
+[profile.test]
+# Some unit tests create files from raw fds and
+# this triggers UB checks to panic. There is no
+# separate flag to disable those as of right now,
+# so we have to disable debug-assertions.
+# Tracking issue: https://github.com/rust-lang/rust/issues/123499
+debug-assertions = false
```
roypat, line 31: In rust-vmm we fixed these double-close cases by just std::mem::forget-ing the already closed File objects (or just doing into_raw_fd() on them). I don't think we should disable debug assertions.

alindima:
> libseccomp provides better quality compiler for
bpf seccomp programs than our current implementation.

I'm curious, what does this mean? and how was this evaluated?

src/seccompiler/Cargo.toml
```diff
@@ -12,25 +12,19 @@ bench = false
 
 [[bin]]
 name = "seccompiler-bin"
-path = "src/seccompiler_bin.rs"
+path = "src/bin.rs"
```
roypat, line 15: If nothing depends on this as a library component anymore, then we should probably this file to `main.rs` and remove the library target from the crate (and maybe rename the entire crate `seccompiler_bin`)

src/seccompiler/src/types.rs
```diff
@@ -0,0 +1,119 @@
...
+impl TryInto<TargetArch> for &str {
+    type Error = String;
+    fn try_into(self) -> std::result::Result<TargetArch, String> {
+        match self.to_lowercase().as_str() {
+            "x86_64" => Ok(TargetArch::X86_64),
+            "aarch64" => Ok(TargetArch::Aarch64),
+            _ => Err(self.to_string()),
+        }
+    }
+}
```
roypat, line 108: nit: Why not `FromStr`?

src/vmm/src/seccomp.rs
```diff
@@ -0,0 +1,241 @@
...
+    map.insert("vcpu".to_string(), Arc::new(vec![]));
+    map
+}
+
+/// Binary filter deserialization errors.
+#[derive(Debug, thiserror::Error, displaydoc::Display)]
+pub enum DeserializationError {
+    /// Bincode deserialization failed: {0}
+    Bincode(BincodeError),
+}
```
roypat, line 38: nit: do we really need an enum with just one variant here? Can we just use `BincodeError` directly?

src/vmm/src/seccomp.rs
```diff
@@ -0,0 +1,241 @@
...
+            .deserialize_from::<R, HashMap<String, BpfProgram>>(reader),
+        // No limit is the default.
+        None => bincode::deserialize_from::<R, HashMap<String, BpfProgram>>(reader),
+    }
+    .map_err(DeserializationError::Bincode)?;
+
+    Ok(result
+        .into_iter()
+        .map(|(k, v)| {
+            println!("bpf len: {}", v.len());
```
roypat, line 59: leftover from testing?

src/vmm/src/seccomp.rs
```diff
@@ -0,0 +1,241 @@
...
+#[derive(Debug, PartialEq, Eq, thiserror::Error, displaydoc::Display)]
+pub enum InstallationError {
+    /// Filter length exceeds the maximum size of {BPF_MAX_LEN:} instructions
+    FilterTooLarge,
+    /// prctl` syscall failed with error code: {0}
+    Prctl(i32),
+}
+
+/// The maximum seccomp-BPF program length allowed by the linux kernel.
+pub const BPF_MAX_LEN: usize = 4096;
```
roypat, line 75: ah, so ever worrying about whether our seccomp rules were larger than a page was moot anyway cuz the kernel would've rejected then :rofl:

src/vmm/src/seccomp.rs
```diff
@@ -0,0 +1,241 @@
...
+        };
+        let bpf_prog_ptr = &bpf_prog as *const SockFprog;
+        {
+            let rc = libc::prctl(
+                libc::PR_SET_SECCOMP,
+                libc::SECCOMP_MODE_FILTER,
+                bpf_prog_ptr,
+            );
+            if rc != 0 {
+                return Err(InstallationError::Prctl(*libc::__errno_location()));
```
roypat, line 124: `std::io::Error::last_os_error()`, or `vmm_sys_util::errno::errno_result(())`

src/seccompiler/Cargo.toml
```diff
@@ -20,7 +20,6 @@ bincode = "1.2.1"
 clap = { version = "4.5.21", features = ["derive", "string"] }
 displaydoc = "0.2.5"
 libc = "0.2.167"
-libseccomp = "0.3.0"
```
roypat: let's merge this commit into the first one, so we don't add a dependency and then remove it again 2 commits later in the same PR.

src/seccompiler/src/bindings.rs
```diff
@@ -0,0 +1,169 @@
...
+#[link(name = "seccomp")]
+extern "C" {
+    /// Initialize the filter state
+    ///
+    /// - `def_action`: the default filter action
+    ///
+    /// This function initializes the internal seccomp filter state and should
+    /// be called before any other functions in this library to ensure the filter
+    /// state is initialized.  Returns a filter context on success, `ptr::null()` on failure.
+    pub fn seccomp_init(def_action: u32) -> scmp_filter_ctx;
```
roypat, line 90: `safe`? :pleading_face:

src/seccompiler/src/lib.rs
```diff
@@ -110,32 +127,97 @@ pub fn compile_bpf(
...
-                        } else {
-                            rule.op.to_scmp_type()
-                        };
-                        ScmpArgCompare::new(rule.index as u32, op, rule.val)
+                            SeccompCmpOp::Ge => scmp_arg_cmp {
+                                arg: rule.index as u32,
+                                op: scmp_compare::SCMP_CMP_GE,
+                                datum_a: rule.val,
+                                datum_b: 0,
+                            },
```
roypat, line 153: nit: this construction can probably go into a utility function? no strong opinions though

src/seccompiler/src/lib.rs
```diff
@@ -146,9 +228,9 @@ pub fn compile_bpf(
         let instructions = size / std::mem::size_of::<u64>();
         let mut bpf = vec![0_u64; instructions];
 
-        memfd
+        _ = memfd
             .read_exact(bpf.as_mut_bytes())
-            .map_err(CompilationError::MemfdRead)?;
+            .map_err(CompilationError::MemfdRead);
```
roypat, line 233:
```suggestion
            .map_err(CompilationError::MemfdRead)?;
```

src/seccompiler/Cargo.toml
ShadowCurse: The Firecracker build script uses this as a lib.

src/vmm/src/seccomp.rs
ShadowCurse: Firecracker seccomp.rs needs this because FC does not depend on bincode directly.

roypat: Ah, annoying. Could solve that by using a public reexport of BincodeError inside this file (e.g. `pub use bincode::BincodeError as DeserializationError`)

src/vmm/src/seccomp.rs
roypat: ah, makes sense

src/seccompiler/Cargo.toml
roypat: ah right. nvm then. maybe we can deprecate seccompiler-bin altogether, and remove it in 2.0. but that's neither here nor there.

src/seccompiler/src/lib.rs
ShadowCurse: done

src/seccompiler/src/lib.rs
ShadowCurse: done

src/seccompiler/src/types.rs
ShadowCurse: done

src/vmm/src/seccomp.rs
ShadowCurse: removed

src/seccompiler/src/bindings.rs
ShadowCurse: done

src/vmm/src/seccomp.rs
ShadowCurse: done

src/vmm/src/seccomp.rs
ShadowCurse: yup, this work

pb8o:
>> libseccomp provides better quality compiler for bpf seccomp programs than our current implementation.

> I'm curious, what does this mean? and how was this evaluated?

Hi @alindima! We found this while revisiting seccomp recently:
- libseccomp produces shorter cBPF programs, around ~65% fewer instructions. The difference is small and probably jitted away, but the bytecode is much easier to read and audit.
- libseccomp has a few interesting features implemented like syscall prioritization. While we don't use this presently, we have had situations where we would like control the order of the syscall rules. For example we can put the most called syscalls first.
- Finally libseccomp is installed by default in many distros and is a dependency of systemd for example, so we assume it is well-maintained and robust. By doing that we can reduce our maintenance effort and focus more on Firecracker.

pb8o (changes requested):
We should fix the ioctl syscall id getting resolved at build time instead of at runtime.

src/seccompiler/src/types.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub op: SeccompCmpOp,
+    pub val: u64,
+}
+
+impl SeccompCondition {
+    pub fn to_scmp_type(&self, syscall: i32) -> scmp_arg_cmp {
+        #[cfg(target_arch = "x86_64")]
+        const IOCTL: i32 = 16;
+        #[cfg(target_arch = "aarch64")]
+        const IOCTL: i32 = 29;
```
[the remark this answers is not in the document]
ShadowCurse: done

src/seccompiler/src/bindings.rs
```diff
@@ -0,0 +1,169 @@
+// Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
ShadowCurse: These are some mix of https://github.com/libseccomp-rs/libseccomp-rs/blob/main/libseccomp-sys/src/lib.rs and new rust syntax (the `unsafe extern` and `safe` keyword). I did not auto generate them.

src/seccompiler/Cargo.toml
```diff
@@ -12,25 +12,18 @@ bench = false
...
+path = "src/bin.rs"
 bench = false
 
 [dependencies]
 bincode = "1.2.1"
+clap = { version = "4.5.21", features = ["derive", "string"] }
 displaydoc = "0.2.5"
-libc = "0.2.168"
-log-instrument = { path = "../log-instrument", optional = true }
+libc = "0.2.167"
```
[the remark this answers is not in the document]
ShadowCurse: fixed

src/seccompiler/src/types.rs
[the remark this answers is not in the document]
roypat: doesnt libseccomp have this cool function for converting between syscall name and number? or is that only available in python? :o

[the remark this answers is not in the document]
pb8o: yeah in C there's two: https://github.com/seccomp/libseccomp/blob/main/include/seccomp.h.in#L609-L631

src/seccompiler/Cargo.toml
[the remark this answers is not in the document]
pb8o: There's still one. Not a blocker, dependabot will fix it in the next run.

src/seccompiler/src/types.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+impl SeccompCondition {
+    pub fn to_scmp_type(&self, arch: &TargetArch, syscall: i32) -> scmp_arg_cmp {
+        // For `ioctls` we need to mask upper bits as musl
+        // sets them to 1, but libseccomp explicitly checks that they are 0.
+        // with 0x00000000FFFFFFFF mask upper bits are always 0.
+        let ioctl = match arch {
+            TargetArch::X86_64 => 16,
+            TargetArch::Aarch64 => 29,
```
pb8o, line 41: if we passed syscall as a string, then we can resolve it here and not have to hardcode these values.

pb8o (dismissed):

src/seccompiler/src/types.rs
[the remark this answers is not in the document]
pb8o: But we already have the string version, we get it from the JSON file. So I think we could just pass the syscall string instead of the number and then we don't need to do any re-translation.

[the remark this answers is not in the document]
roypat: You okay if I commit this diff?

```
diff --git a/src/seccompiler/src/bindings.rs b/src/seccompiler/src/bindings.rs
index c13797ef31a4..bf6214f55473 100644
--- a/src/seccompiler/src/bindings.rs
+++ b/src/seccompiler/src/bindings.rs
@@ -163,6 +163,18 @@ unsafe extern "C" {
     /// This function generates seccomp Berkeley Packer Filter (BPF) code and writes
     /// it to the given fd.  Returns zero on success, negative values on failure.
     pub fn seccomp_export_bpf(ctx: const_scmp_filter_ctx, fd: c_int) -> c_int;
+
+    ///
+    /// Resolve a syscall name to a number
+    /// @param arch_token the architecture token, e.g. SCMP_ARCH_*
+    /// @param name the syscall name
+    ///
+    /// Resolve the given syscall name to the syscall number for the given
+    /// architecture.  Returns the syscall number on success, including negative
+    /// pseudo syscall numbers (e.g. __PNR_///); returns __NR_SCMP_ERROR on failure.
+    ///
+    ///
+    pub fn seccomp_syscall_resolve_name_arch(arch_token: u32, name: *const libc::c_char) -> c_int;
 }
 
 /// Negative pseudo syscall number returned by some functions in case of an error
diff --git a/src/seccompiler/src/types.rs b/src/seccompiler/src/types.rs
index 449b0438fa86..d0cddfa55952 100644
--- a/src/seccompiler/src/types.rs
+++ b/src/seccompiler/src/types.rs
@@ -36,10 +36,11 @@ impl SeccompCondition {
         // For `ioctls` we need to mask upper bits as musl
         // sets them to 1, but libseccomp explicitly checks that they are 0.
         // with 0x00000000FFFFFFFF mask upper bits are always 0.
-        let ioctl = match arch {
-            TargetArch::X86_64 => 16,
-            TargetArch::Aarch64 => 29,
+        // SAFETY: c"ioctl".as_ptr() is a pointer to a nul-terminated C-string.
+        let ioctl = unsafe {
+            seccomp_syscall_resolve_name_arch(arch.to_scmp_type(), c"ioctl".as_ptr())
         };
+
         match self.op {
             SeccompCmpOp::Eq => {
                 if syscall == ioctl {

```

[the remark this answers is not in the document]
roypat: Ah, you're right... we first translate from string to number, only to then kinda need the string here again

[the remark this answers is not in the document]
roypat: Actually, instead of hotpatching these in seccompiler-bin, maybe we should just update our seccomp filters for ioctls to use masked_eq? :sweat_smile:

roypat (approved):

roypat (dismissed):

roypat (dismissed):

pb8o (approved):

roypat (approved):

pb8o (approved):

PR #4937, opened 2024-12-02
devtool: add download_ci_artifacts command

roypat:
Every now and then, I managed to end up with an empty CI artifacts
folder, for whatever reason (mostly me being silly). The solution is
generally to remove whatever messed up stuff I ended up with in
build/img, and run some arbitrary devtool command that I know will
redownload everything. Similar workflows must happen whenever we do an
in-place update of our CI artifacts.

Abstract this away into a command that deletes the build/img folder, and
then redownloads it.

## Changes

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tools/devtool
+++ b/tools/devtool
@@ -878,7 +878,6 @@ cmd_shell() {
 
 cmd_sh() {
     ensure_build_dir
-    ensure_ci_artifacts
     run_devctr \
         --privileged \
         --ulimit nofile=4096:4096 \
@@ -916,6 +915,7 @@ cmd_sandbox_native() {
 }
 
 cmd_test_debug() {
+    ensure_ci_artifacts
     cmd_sh "tmux new ./tools/test.sh --pdb $@"
 }
 
--- a/tools/devtool
+++ b/tools/devtool
@@ -562,7 +562,8 @@ ensure_ci_artifacts() {
 
     # Fetch all the artifacts so they are local
     say "Fetching CI artifacts from S3"
-    S3_URL=s3://spec.ccfc.min/firecracker-ci/v1.11/$(uname -m)
+    FC_VERSION=$(cmd_sh "cd src/firecracker/src; cargo pkgid | cut -d# -f2 | cut -d. -f1-2")
+    S3_URL=s3://spec.ccfc.min/firecracker-ci/v$FC_VERSION/$(uname -m)
     ARTIFACTS=$MICROVM_IMAGES_DIR/$(uname -m)
     if [ ! -d "$ARTIFACTS" ]; then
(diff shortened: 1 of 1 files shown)


tools/devtool
```diff
@@ -562,7 +562,10 @@ ensure_ci_artifacts() {
 
     # Fetch all the artifacts so they are local
     say "Fetching CI artifacts from S3"
-    S3_URL=s3://spec.ccfc.min/firecracker-ci/v1.11/$(uname -m)
+    pushd $FC_ROOT_DIR/src/firecracker/src
+    FC_VERSION=$(cargo pkgid | cut -d# -f2 | rev | cut -d. -f2- | rev)
```
pb8o, line 576: But this is running using cargo from outside devctr, which we cannot guarantee will be available (for example in the CI metals). I think a `cmd_sh cargo ...` could work, maybe:

```sh
cmd_sh "cd src/firecracker; cargo pkgid | cut -d# -f2 | cut -d- -f1"
```

Also I think the `cut` is enough?

```
% echo 1.11.0 |cut -d- -f1
1.11.0
% echo 1.11.0-dev |cut -d- -f1
1.11.0
```

roypat: Ah, missed that, thanks!

We need the second cut to transform `1.11.0` -> `1.11` because of how our artifacts bucket is named

bchalios (approved):

tools/devtool
pb8o: Ah right, would this work without the revs `cut -d. -f1-2`?

tools/devtool
```diff
@@ -562,7 +562,14 @@ ensure_ci_artifacts() {
 
     # Fetch all the artifacts so they are local
     say "Fetching CI artifacts from S3"
-    S3_URL=s3://spec.ccfc.min/firecracker-ci/v1.11/$(uname -m)
+
+    FC_VERSION=$(run_devctr \
```
pb8o, line 578: Can we re-use `cmd_sh` here?

roypat: we cannot because it calls ensure_ci_artifacts :(

pb8o: Can we decouple it so that every existing call site of `cmd_sh`, we call `cmd_download_artifacts` first (if applicable)? I think in most cases it was unneeded but we had no way of controlling it.

tools/devtool
roypat: O, neat, stackoverflow seemingly didnt know about that option :joy: fixed :)

bchalios (approved):

PR #4943, opened 2024-12-04
refactor(tests): move return code check for ssh into ssh.run

ShadowCurse:
## Changes
Move return code check for ssh into ssh.run.

## Reason
Ssh connection should be considered an error by default. In order to unify handling of this case, move check for non zero return code into the function itself.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tests/host_tools/network.py
+++ b/tests/host_tools/network.py
@@ -79,7 +79,7 @@ class SSHConnection:
 
     def _scp(self, path1, path2, options):
         """Copy files to/from the VM using scp."""
-        self._exec(["scp", *options, path1, path2], check=True)
+        self._exec(["scp", *options, path1, path2])
 
     def scp_put(self, local_path, remote_path, recursive=False):
         """Copy files to the VM using scp."""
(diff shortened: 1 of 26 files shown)


roypat:
Ah, we've kept the APIs as run/check_output to be consistent with the subprocess APIs. If we change this, then we should also change `utils.run_cmd`/`utils.check_output` the same way, otherwise things will be confusing.

tests/host_tools/network.py
```diff
@@ -124,11 +125,7 @@ def run(self, cmd_string, timeout=None, *, check=False, debug=False):
 
         return self._exec(command, timeout, check=check)
 
-    def check_output(self, cmd_string, timeout=None, *, debug=False):
-        """Same as `run`, but raises an exception on non-zero return code of remote command"""
-        return self.run(cmd_string, timeout, check=True, debug=debug)
```
pb8o, line 129: We modeled `check_output/run` on `subprocess.check_output/run`. I think we should keep them as they are, as they are both useful. Most cases can be handled by check_output, but run is useful if the caller wants to handle the error.

Why not move more `run` calls to explicitly use `check_output` instead?

pb8o (changes requested):

tests/host_tools/network.py
ShadowCurse: Wouldn't it be easier if there was only 1 method to call? By default `run` will check output and if we need some exception, we will explicitly set `check=False`.
I did not touch `supbprocess.run` yet here, but was going to do same treatment there.

Manciukic: I agree with Pablo. Ideally out `run`/`check_output` would behave exactly like `subprocess.run/check_output`.

PR #4944, opened 2024-12-05
test: Enable to run CI with custom CPU templates for testing / debugging purpose

zulinx86:
## Changes

Enable us to run CI with custom CPU templates for testing / debugging purpose.

- Added `tests/data/custom_cpu_templates` dir and pick JSON files under the dir as custom CPU templates unconditionally to run the CI.
- Added the two new options to the buildkite steps generation scripts:
  - `--additional-prepend`: allows us to inject or download custom CPU templates before running tests.
  - `--keywords` / `-k`: allows us to filter tests to run
- Documented how to run CI with custom CPU templates in the test README.

## Reason

To speed up the testing / debugging cycle of custom CPU templates.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

- ~~[ ] I have read and understand [CONTRIBUTING.md][3].~~
  automated style checks.
  in the PR.
  integration tests.


pb8o:
I think the issue with this approach is we will only run the tests that explicitly use a CPU template. I think what we want there is to run all tests with the specified template. Many will fail because they don't apply, but we will have to investigate and see if it's expected or not. The idea roughly is:

```
./tools/devtool test -- --cpu-template my-template.json functional
```

That will apply `my-template.json` for all tests that use a microvm. In addition it should probably also override the `cpu_templates` fixture to only return that template, or disable it completely maybe.

zulinx86:
> I think the issue with this approach is we will only run the tests that explicitly use a CPU template. I think what we want there is to run all tests with the specified template. Many will fail because they don't apply, but we will have to investigate and see if it's expected or not. The idea roughly is:
> 
> ```
> ./tools/devtool test -- --cpu-template my-template.json functional
> ```
> 
> That will apply `my-template.json` for all tests that use a microvm. In addition it should probably also override the `cpu_templates` fixture to only return that template, or disable it completely maybe.

Let me confirm what we want first. Do we want to run all tests (not only tests normally running with CPU templates) with the given CPU template?

If no, I think making `custom_cpu_template_params()` return the custom CPU template looks good. Implementing `--cpu-template` option was what I tried first, but I didn't come up with a good way to make it return the custom CPU template. Maybe because I'm not good at pytest :/ Do you have any idea to implement it in a neat way?

zulinx86:
Close this in favor of #4975.

PR #4948, opened 2024-12-06
fix: rename page_size_kib to page_size

acsmyth:
## Changes
* update `page_size_kib` name to the more accurate `page_size`, since it's actually in bytes, not KiB.
...

We leave the old page_size_kib field in there, but mark it as deprecated to avoid any Firecracker code from accidentally using it (and update everything on the Firecracker side to be called page_size instead of page_size_kib). This way we're backward compatible (and can remove the wrongly named field in 2.0), but have correctly named variables in the code base at least.
...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@ and this project adheres to
 
 ### Deprecated
 
+- [#4948](https://github.com/firecracker-microvm/firecracker/pull/4948):
+  Deprecated the `page_size_kib` field in the
+  [UFFD handshake](docs/snapshotting/handling-page-faults-on-snapshot-resume.md#registering-memory-to-be-handled-via-userfault-file-descriptors),
+  and replaced it with a `page_size` field. The `page_size_kib` field is
+  misnamed, as the value Firecracker sets it to is actually the page size in
+  _bytes_, not KiB. It will be removed in Firecracker 2.0.
+
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


tests/integration_tests/performance/test_huge_pages.py
```diff
@@ -51,8 +51,8 @@ def check_hugetlbfs_in_use(pid: int, allocation_name: str):
     cmd = f"cat /proc/{pid}/smaps | grep {allocation_name} -A 23 | grep KernelPageSize"
     _, stdout, _ = utils.check_output(cmd)
 
-    kernel_page_size_kib = int(stdout.split()[1])
-    assert kernel_page_size_kib > 4
+    kernel_page_size = int(stdout.split()[1])
+    assert kernel_page_size > 4096
```
pb8o, line 55: Is this intended? the `smaps` file does report this value in kB:

```
KernelPageSize:        4 kB
```

I think this should be left as-is.

pb8o (changes requested):

tests/integration_tests/performance/test_huge_pages.py
pb8o: Other than that, it looks good to me. Thanks for your contribution! :)

roypat:
Hi @acsmyth, 
thanks for catching this! The team's discussed this, and we agreed that although huge pages support is somewhat new, this would still be a breaking change, and so there should be some sort of transition period for users of Firecracker. So for that reason, I think we should just carry two fields for a while (e.g. both `page_size` and `page_size_kib` with the broken semantics). That being said, I think most of your PR can go through unchanged, we'll only need to readd `page_size_kib` to `GuestRegionUffdMapping` and set it to the same value as `page_size` in `create_guest_memory`. We'd also need to add something to `DEPRECATED.md` warning that we'll remove `page_size_kib` in an upcoming version.

pb8o (approved):

pb8o (approved):

PR #4953, opened 2024-12-10
test: use single SSH connection for lifetime of microvm

roypat:
## Changes

In the `SSHConnection` class in our integration test framework, open a single SSH connection and reuse that for all communication with a specific microvm.

## Reason

Currently, we were recreating SSH connections for each command we want to run inside a microvm. This was leading to spurious failures because only the first connection attempt (which we also use to wait for microvm boot) was using retries.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.


tests/host_tools/network.py
```diff
@@ -68,35 +69,19 @@ def __init__(self, netns, ssh_key: Path, host, user, *, on_error=None):
...
-        """remote address for in SSH format <user>@<IP>"""
-        return f"{self.user}@{self.host}"
-
-    def remote_path(self, path):
-        """Convert a path to remote"""
-        return f"{self.user_host}:{path}"
-
     def _scp(self, path1, path2, options):
         """Copy files to/from the VM using scp."""
         self._exec(["scp", *options, path1, path2], check=True)
```
zulinx86, line 74: nit: Do we still need `_scp()`? If no, we can also remove the internal function `_exec()` and hardcode the logic in `run()`.

zulinx86 (approved):

tests/host_tools/network.py
roypat: Ah, yeah, good catch! Removed :)

roypat:
I'm closing this PR since @pb8o has shown me an approach using ControlMaster/ControlPersist that doesnt require new dependencies. Will open a separate PR with that.

PR #4955, opened 2024-12-11
test: use single SSH connection for lifetime of microvm

roypat:
Instead of creating new SSH connections every time we want to run a
command inside the microvm, open a single daemonized ssh connection in
the constructor of `SSHConnection`, and reuse it until we kill the
microvm.

Realize this using openssh's ControlMaster/ControlPersist functionality,
which allows us to persist the first connection opened to a microvm and
multiplex all subsequent commands over this one connection.

We need some slight changes in test_pause_restore.py and test_net.py. In
the former, since we no longer reconnect to the VM on every ssh command,
we will now observe a timeout instead of a connection failure. For the
latter, it turns out that @lru_cache treats .ssh_iface() and
.ssh_iface(0) as distinct invokations, despire the iface_idx argument of
ssh_iface defaulting to 0 (meaning they are actually the same function
call). This caused a re-intialization of the SSHConnection object, which
then triggered the assertion in _init_connection about the socket file
not existing.

Lastly, fix up some other small issues that were causing intermittent failures I observed while testing this.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.


tests/integration_tests/functional/test_net.py
```diff
@@ -43,7 +43,7 @@ def test_high_ingress_traffic(uvm_plain_any):
     # Start iperf3 client on the host. Send 1Gbps UDP traffic.
     # If the net device breaks, iperf will freeze. We have to use a timeout.
     utils.check_output(
-        "timeout 31 {} {} -c {} -u -V -b 1000000000 -t 30".format(
+        "{} {} -c {} -u -V -b 1000000000 -t 30".format(
```
ShadowCurse, line 46: Need to remove the comment as well.

tests/framework/microvm.py
```diff
@@ -1007,13 +1013,16 @@ def ssh_iface(self, iface_idx=0):
         """Return a cached SSH connection on a given interface id."""
         guest_ip = list(self.iface.values())[iface_idx]["iface"].guest_ip
         self.ssh_key = Path(self.ssh_key)
-        return net_tools.SSHConnection(
+        connection = net_tools.SSHConnection(
             netns=self.netns.id,
             ssh_key=self.ssh_key,
             user="root",
             host=guest_ip,
+            control_path=Path(self.chroot()) / f"ssh-{iface_idx}.sock",
```
ShadowCurse, line 1021: Do I understand correctly that now instead of openning new host<->guest ssh connections, we open 1 master ssh on first `ssh_iface` call and then create new ssh processes that just communicate with guest through the master ssh?

roypat: yep!

tests/integration_tests/functional/test_net.py
roypat: fascinating how I completely manage to fail to read that comment when writing the commit :joy: 

oh well, pytest timeout will do the job better anyway lol

tests/framework/microvm.py
ShadowCurse: simplification...

tests/framework/microvm.py
```diff
@@ -282,6 +284,10 @@ def kill(self):
         for monitor in self.monitors:
             monitor.stop()
 
+        # Kill all background SSH connections
+        for connection in self._connections:
+            connection.close()
+
```
pb8o, line 290: There may be a way of introspecting the `ssh_iface` cache. Or we could do `self.ssh_iface.cache_clear()` and close the connection on `SSHConnection.__del__`  though I don't like desctructors that much..

tests/host_tools/network.py
```diff
@@ -141,27 +201,6 @@ def _exec(self, cmd, timeout=100, check=False):
 
             raise
 
-    # pylint:disable=invalid-name
```
pb8o, line 144: I know this is not used in the integ tests, but it doesn't work anymore with this approach?

tests/framework/microvm.py
pb8o: Nvm, just read the commit body about @lru_cache treating.. Though it may be worth removing the default iface_idx = 0. I don't think anything depends on it?

pb8o (approved):

tests/integration_tests/functional/test_net.py
```diff
@@ -41,9 +41,9 @@ def test_high_ingress_traffic(uvm_plain_any):
     time.sleep(1)
 
     # Start iperf3 client on the host. Send 1Gbps UDP traffic.
-    # If the net device breaks, iperf will freeze. We have to use a timeout.
+    # If the net device breaks, iperf will freeze, and we'll hit the pytest timeout
     utils.check_output(
-        "timeout 31 {} {} -c {} -u -V -b 1000000000 -t 30".format(
+        "{} {} -c {} -u -V -b 1000000000 -t 30".format(
```
zulinx86, line 46: Cool! It was also seen last week in the nightly pipeline.

zulinx86 (approved):

tests/host_tools/network.py
roypat: Ah, I think this is mostly an artifact of the paramiko approach, where this wasn't supported anymore. Should still work here? But admittedly didnt look at it, since as you say its dead code :sweat_smile:

tests/framework/microvm.py
roypat: Yeah, I tried inspecting the lru_cache and it didn't seem to provide a way to get access to the cached objects. Can probably remove the default, but I guess now that test_net.py is updated its also not causing any problems anymore haha

roypat: Btw destructors also wouldn't work because the connections need to be closed before the microvm is killed, because further down in microvm.kill() there's the check about "something using the jailer id is still alive"

PR #4961, opened 2024-12-16
test: use fio instead of dd for performance tests

roypat:
In test_drive_rate_limiter, use `fio` for measuring latencies of writing a fixed number of bytes to a block device, instead of `dd` (which the test itself notes is unreliable). Should fix intermitted failures we're been seeing in this test, hopefully.

## Changes

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tests/integration_tests/performance/test_drive_rate_limiter.py
+++ b/tests/integration_tests/performance/test_drive_rate_limiter.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 """Tests for checking the rate limiter on /drives resources."""
-
+import json
 import os
 
 import host_tools.drive as drive_tools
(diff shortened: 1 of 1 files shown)


pb8o (approved):

Manciukic:
> Should fix intermitted failures we're been seeing in this test, hopefully

Do you have any link to the intermittent failures? How far off are they from the target? Are we sure it's a `dd` issue? I doubt that `dd` is doing weird things, it just copies bytes from one place to another with a single thread, so it might be impacted by other things running at the same time.

roypat:
> > Should fix intermitted failures we're been seeing in this test, hopefully
> 
> Do you have any link to the intermittent failures? How far off are they from the target? Are we sure it's a `dd` issue? I doubt that `dd` is doing weird things, it just copies bytes from one place to another with a single thread, so it might be impacted by other things running at the same time.

e.g. https://buildkite.com/firecracker/firecracker-pr/builds/12155#0193cfc3-5704-40ba-9ebb-9e671d9e7dbc. It seems to just be slightly above the upper target by a few ms. There shouldn't be anything running at the same time, the perf tests aren't parallelized at the pytest level, and are assigned a whole .metal (e.g. only one buildkite agent on the metal). My main point here is that `dd` is not a performance testing tool, its a tool for copying data around that happens to print some statistics, while `fio` is a proper I/O benchmark tool. E.g. on my laptop, `dd` has a range of around 200ms, while `fio` only has a range of around 100ms, while also being ~100ms faster on average.

Manciukic:
Makes sense, thanks for the context!

I don't know if it's karma but this same test just failed on my PR as well lol

Manciukic (approved):

PR #4962, opened 2024-12-16
fix(test): do not use absolute path to true when opening ssh connection

Manciukic:
## Changes
This change fixes the `test-popular-containers` suite by using the binary name `true` when estabilishing the ssh connection instead of the absolute path, and cleaning up the ssh control socket on error before the next retry.

## Reason
Fixes #4955 

In #4955 the executable to check the ssh connection liveliness was changed from `true` to `/usr/bin/true`, but that is not its path in all rootfs, causing failures in the `test-popular-containers` suite.

Also, since the error is retried but the control socket is not cleaned up, subsequent retries would fail for the assertion.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
- [NA] I have updated any relevant documentation (both in code and in the docs)
  in the PR.
- [NA] I have mentioned all user-facing changes in `CHANGELOG.md`.
- [NA] If a specific issue led to this PR, this PR closes the issue.
- [NA] When making API changes, I have followed the
- [NA] I have tested all new and changed functionalities in unit tests and/or
  integration tests.
- [NA] I have linked an issue to every new `TODO`.

### DIFF
--- a/tests/host_tools/network.py
+++ b/tests/host_tools/network.py
@@ -129,14 +129,21 @@ class SSHConnection:
             "ControlPersist=yes",
             *self.options,
             self.user_host,
-            "/usr/bin/true",
+            "true",
         ]
 
-        # don't set a low timeout here, because otherwise we might get into a race condition
-        # where ssh already forked off the persisted connection daemon, but gets killed here
-        # before exiting itself. In that case, self._control_path will exist, and the retry
-        # will hit the assert at the start of this function.
-        self._exec(establish_cmd, check=True)
+        try:
+            # don't set a low timeout here, because otherwise we might get into a race condition
+            # where ssh already forked off the persisted connection daemon, but gets killed here
+            # before exiting itself. In that case, self._control_path will exist, and the retry
+            # will hit the assert at the start of this function.
+            self._exec(establish_cmd, check=True)
+        except Exception:
+            # if the control socket is present, then the daemon is running, and we should stop it
+            # before retrying again
+            if self._control_path.exists():
+                self.close()
+            raise
 
     def _check_liveness(self) -> int:
         """Checks whether the ControlPersist connection is still alive"""


tests/host_tools/network.py
```diff
@@ -129,14 +129,20 @@ def _init_connection(self):
...
+            # don't set a low timeout here, because otherwise we might get into a race condition
+            # where ssh already forked off the persisted connection daemon, but gets killed here
+            # before exiting itself. In that case, self._control_path will exist, and the retry
+            # will hit the assert at the start of this function.
+            self._exec(establish_cmd, check=True)
+        except Exception:
+            # cleanup the control socket to allow for retries
+            if self._control_path.exists():
+                self._control_path.unlink()
+            raise
```
roypat, line 145: Mh, if we do this then we can probably remove the assert altogether (and the comment, too). Although I think I slightly prefer not explicitly removing the path here. The idea behind this was that if the socket file ends up getting created, then we definitely have an SSH daemon process, and just retrying in this case will simply create more daemon processes. The assert prevent this from happening. Another option here would be to catch the exception and instead of unlinking, we do `self.close()`.

Manciukic: fair point. I tested with just `self.close()` and it works as well. `close()` may fail, but in that case it would be retried anyways

roypat: Yeah, but if `self._exec` fails and then `self.close()` as well, I'd honestly prefer us to error out completely, because something has fundamentally gone wrong then.

Mhhh... thinking about it now, maybe we should simply not retry in the case "exception and self._control_path.exists()". E.g. inside the `if` instead of unlink/close, we do `raise AbortRetriesAndFailException() from e` (name tbd), and in the retry decorator we make it not retry when it encounters this exception type (or potentially, only do that if _close_ also fails). That's pretty much the same behavior we have today, with the only difference being that instead of the assertion causing the abort (and thus hiding the actual cause), we get a better error message

pb8o (approved):

tests/host_tools/network.py
Manciukic: It starts to sound complicated for a fix that's a couple of lines. We cannot predict every possible failure mode, and know whether it's recoverable or not. If it ever fails like you describe, then we'd see it on the final retry and it would have the same error message. It would just take 20s more.

roypat: that's fair enough

roypat (approved):

tests/host_tools/network.py
```diff
@@ -129,14 +129,21 @@ def _init_connection(self):
             "ControlPersist=yes",
             *self.options,
             self.user_host,
-            "/usr/bin/true",
+            "true",
```
pb8o, line 132: nit: I wonder if we need to run a command at all... would `-N` work?

pb8o (approved):

PR #4978, opened 2025-01-02
chore(vmm): Add pointer alignment check for Queue

zulinx86:
## Changes

Add pointer alignment check and exit with an error early instead of panic.

## Reason

Raw pointers of `struct Queue` are assumed to be aligned properly;
otherwise some methods (e.g. `read_volatile()`) will panic. Such an
alignment is possible when restored from a broken/fuzzed snapshot.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.~~
  integration tests.

### DIFF
--- a/src/vmm/src/devices/virtio/queue.rs
+++ b/src/vmm/src/devices/virtio/queue.rs
@@ -34,6 +34,8 @@ pub enum QueueError {
     DescIndexOutOfBounds(u16),
     /// Failed to write value into the virtio queue used ring: {0}
     MemoryError(#[from] vm_memory::GuestMemoryError),
+    /// Pointer is not aligned properly: {0:#x} not {1}-byte aligned.
+    PointerNotAligned(usize, u8),
 }
 
 /// A virtio descriptor constraints with C representative.
@@ -323,11 +325,36 @@ impl Queue {
             .get_slice_ptr(mem, self.used_ring_address, self.used_ring_size())?
             .cast();
 
-        // Disable it for kani tests, otherwise it will hit this assertion
-        // and fail.
-        #[cfg(not(kani))]
-        if self.actual_size() < self.len() {
-            return Err(QueueError::InvalidQueueSize(self.len(), self.actual_size()));
+        // All the above pointers are expected to be aligned properly; otherwise some methods (e.g.
+        // `read_volatile()`) will panic. Such an unalignment is possible when restored from a
+        // broken/fuzzed snapshot.
+        //
+        // Specification of those pointers' alignments
+        // https://docs.oasis-open.org/virtio/virtio/v1.2/csd01/virtio-v1.2-csd01.html#x1-350007
+        // > ================ ==========
+        // > Virtqueue Part    Alignment
+        // > ================ ==========
+        // > Descriptor Table 16
+        // > Available Ring   2
+        // > Used Ring        4
+        // > ================ ==========
+        if !self.desc_table_ptr.cast::<u128>().is_aligned() {
+            return Err(QueueError::PointerNotAligned(
+                self.desc_table_ptr as usize,
+                16,
+            ));
+        }
+        if !self.avail_ring_ptr.is_aligned() {
+            return Err(QueueError::PointerNotAligned(
+                self.avail_ring_ptr as usize,
+                2,
+            ));
+        }
+        if !self.used_ring_ptr.cast::<u32>().is_aligned() {
+            return Err(QueueError::PointerNotAligned(
+                self.used_ring_ptr as usize,
+                4,
+            ));
         }
 
         Ok(())
@@ -1234,7 +1261,6 @@ mod verification {
 
 #[cfg(test)]
 mod tests {
-
     use vm_memory::Bytes;
 
     pub use super::*;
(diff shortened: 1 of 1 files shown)


src/vmm/src/devices/virtio/queue.rs
```diff
@@ -323,6 +325,20 @@ impl Queue {
...
+        // a pointer of interest is aligned properly; otherwise results in a panic. Such an
+        // unalignment is possible when restored from a broken/fuzzed snapshot. Both
+        // `avail_ring_ptr` and `used_ring_ptr` must be aligned at least for u16 (2-byte alignment)
+        // as documented in `struct Queue`.
+        if !self.avail_ring_ptr.is_aligned() {
+            return Err(QueueError::PointerNotAligned(
+                "Queue::avail_ring_ptr".into(),
+            ));
+        }
+        if !self.used_ring_ptr.cast::<u16>().is_aligned() {
```
pb8o, line 338: why is the cast needed here but not above?

zulinx86: as documented [here](https://github.com/firecracker-microvm/firecracker/blob/11d8a01365bfa7d472e75e1dde0ac05e00fe19eb/src/vmm/src/devices/virtio/queue.rs#L235-L243), the `used_ring_ptr` is a mix of u16 and u32 so its type is `*mut u8`. but it has to be aligned at least for u16.

pb8o (dismissed):

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -323,6 +325,20 @@ impl Queue {
             .get_slice_ptr(mem, self.used_ring_address, self.used_ring_size())?
             .cast();
 
+        // read_volatile() method is called on avail_ring_ptr and used_ring_ptr and it assumes that
+        // a pointer of interest is aligned properly; otherwise results in a panic. Such an
+        // unalignment is possible when restored from a broken/fuzzed snapshot. Both
+        // `avail_ring_ptr` and `used_ring_ptr` must be aligned at least for u16 (2-byte alignment)
+        // as documented in `struct Queue`.
+        if !self.avail_ring_ptr.is_aligned() {
```
ShadowCurse, line 333: Let's add check for `desc_table_ptr` as well. Also these alignment requirements should hold regardless of `read_volatile` usage e.g they always must be aligned, so the comment can be simplified to that requirement.

zulinx86: I'm not 100% sure if it is always aligned, but yeah let's check it and see if the CI passes.

ShadowCurse: In the virtio spec in 2.7 Split Virtqueues:
```
The memory alignment and size requirements, in bytes, of each part of the virtqueue
are summarized in the following table:

Virtqueue Part   | Alignment   | Size
Descriptor Table | 16          | 16∗(Queue Size)
Available Ring   | 2           | 6 + 2∗(Queue Size)
Used Ring.       | 4           | 6 + 8∗(Queue Size)
```

zulinx86: done.

zulinx86: Cool! Thank you for pointing me to it! Will add the spec link!

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -34,6 +34,8 @@ pub enum QueueError {
     DescIndexOutOfBounds(u16),
     /// Failed to write value into the virtio queue used ring: {0}
     MemoryError(#[from] vm_memory::GuestMemoryError),
+    /// Pointer is not aligned properly: {0}
+    PointerNotAligned(String),
```
ShadowCurse, line 38: Maybe we should create 3 enum values instead of using strings here because strings are quite big in size (24 bytes). In this particular enum this is not really an issue since `GuestMemoryError` is already 24 bytes, so the total size of the error enum is 32 bytes (8 bytes for tag because `GuestMemoryError` is 8 bytes aligned) and adding string will not change things. But still having separate values seems a bit cleaner solution here.

zulinx86: Do we really care about the size of error type? To be honest, I'd prefer to omit it at all rather than having an enum just for this error message.

ShadowCurse: Well, the error size influences all `Result<..., QueueError>` types and the bigger the types, the more space they take in cache and the less of them can fit there. Considering this one is used in hot place in the code (the networking) I think the size matters. What do you mean by `omit` the error? Just replace errors with asserts? Personally I am in favor of this, but it is not really a thing we do in the code base.

ShadowCurse: In any case, this is not critical, but maybe worth a `TODO` note. Maybe some contributor will be willing to go through all error enums and minimize them.

zulinx86: > What do you mean by omit the error? Just replace errors with asserts?

What I meant is removing the `String` arg (like `PointerNotAligned` rather than `PointerNotAligned(String)`).

Another alternative is like:
```
    /// Pointer is not aligned properly: {0:p} not aligned for u{1}
    PointerNotAligned(*const u8, u8),
```
It will print `Pointer is not aligned properly: 0x0001 not aligned for u16`.

ShadowCurse: Ok, these look nice. But alignment is usually in bytes, so instead of 128, 16, 32 it is just 16, 2, 4 (as in the table)
And message can be a simple: `Pointer {0:p} is not {1} byte aligned`.

zulinx86: done

ShadowCurse (approved):

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -34,6 +34,8 @@ pub enum QueueError {
     DescIndexOutOfBounds(u16),
     /// Failed to write value into the virtio queue used ring: {0}
     MemoryError(#[from] vm_memory::GuestMemoryError),
+    /// Pointer is not aligned properly: {0:p} not {1}-byte aligned.
+    PointerNotAligned(*const u8, u8),
```
ShadowCurse, line 38: Rust really does not like passing pointers, so let's just convert them to `usize` and print them in hex `{:#x}`.

pb8o (approved):

ShadowCurse (dismissed):

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -323,11 +325,36 @@ impl Queue {
...
+        // > ================ ==========
+        // > Descriptor Table 16
+        // > Available Ring   2
+        // > Used Ring        4
+        // > ================ ==========
+        if !self.desc_table_ptr.cast::<u128>().is_aligned() {
+            return Err(QueueError::PointerNotAligned(
+                self.desc_table_ptr as usize,
+                16,
+            ));
```
bchalios, line 345: Could we add a unit test for these cases, where we emulate a broken snapshot case and expect the error? Same for following cases.

zulinx86: sure, of course! thanks for pointing it out!

zulinx86: added.

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -1644,6 +1670,63 @@ mod tests {
...
+        q.avail_ring_address = GuestAddress(0x1);
+        match q.initialize(&mem) {
+            Ok(_) => panic!("Unexpected success"),
+            Err(QueueError::PointerNotAligned(addr, alignment)) => {
+                assert_eq!(addr % 2, 0x1);
+                assert_eq!(alignment, 2);
+            }
+            Err(e) => panic!("Unexpected error {e:#?}"),
+        }
+        q.avail_ring_address = GuestAddress(0x0);
```
bchalios, line 1716: Shouldn't having a 0 address cause issues?

zulinx86: No, it doesn't. `initialize()` method only populates `*_ptr` fields from `*_address` fields (not access them at all).

bchalios (approved):

ShadowCurse (approved):

PR #4981, opened 2025-01-07
fix: stop messing with C-states in performance tests

roypat:
When running performance tests, the devtool script was disabling lower C-states, leaving only C0 (e.g. "CPU actively polling for new instructions") enabled, with the goal to reduce latencies of switching between C-states, and therefore increase test stability. However, we found out that on m6i (Icelake) this was causing performance to become erratic, with most performance metrics taking on bi-modal characteristics (funnily enough, this seems to be tied to something in the host kernel, as this volatility appears and disappears with host kernel updates, but we weren't able to narrow it down to _what_).

Since back when we originally disabled sleep states, we did not see any impact on metrics, the theory that it helps with test stability doesn't have any experimental support, so removing the C-state configuration to fix the volatility on m6i is not expected to have any drawbacks.

## Changes

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tools/devtool
+++ b/tools/devtool
@@ -630,11 +630,10 @@ apply_linux_61_tweaks() {
 }
 
 
-# Modifies the processors C- and P-state configuration (x86_64 only) for consistent performance. This means
+# Modifies the processors CPU governor and P-state configuration (x86_64 only) for consistent performance. This means
 # - Disable turbo boost (Intel only) by writing 1 to /sys/devices/system/cpu/intel_pstate/no_turbo
 # - Disable turbo boost (AMD only) by writing 0 to /sys/devices/system/cpu/cpufreq/boost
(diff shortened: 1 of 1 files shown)


tools/devtool
```diff
@@ -660,15 +660,6 @@ apply_performance_tweaks() {
   # their maximum safe frequency. It seems to be the default for Amazon Linux, but it doesn't hurt to make this explicit.
   # See also https://wiki.archlinux.org/title/CPU_frequency_scaling
   echo performance | sudo tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor &> /dev/null
```
kalyazin, line 662: should probably also update the comment on line 637

roypat: ah, yeah, good point. Fixed!

pb8o (approved):

zulinx86 (approved):

kalyazin (approved):

PR #5011, opened 2025-01-24
fix: Rare failure in pause resume test

JackThomson2:
## Changes

Test with 1 vCPU when running the KVMCLOCK_CTRL tests so we don't fail on intermittent issue with clock.

Confirmed with the KVMCLOCK_CTR call removed the test still fails after the update.

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tests/integration_tests/functional/test_pause_resume.py
+++ b/tests/integration_tests/functional/test_pause_resume.py
@@ -141,7 +141,12 @@ def test_kvmclock_ctrl(uvm_plain_any):
     microvm = uvm_plain_any
     microvm.help.enable_console()
     microvm.spawn()
-    microvm.basic_config()
+
+    # With 2 vCPUs under certain conditions soft lockup warnings can rarely be in dmesg causing this test to fail.
+    # Example of the warning: `watchdog: BUG: soft lockup - CPU#0 stuck for (x)s! [(udev-worker):758]`
+    # With 1 vCPU this intermittent issue doesn't occur. If the KVM_CLOCK_CTRL IOCTL is not made
+    # the test will fail with 1 vCPU, so we can assert the call to the IOCTL is made.
+    microvm.basic_config(vcpu_count=1)
     microvm.add_net_iface()
     microvm.start()
 


Manciukic (approved):

tests/integration_tests/functional/test_pause_resume.py
```diff
@@ -141,7 +141,9 @@ def test_kvmclock_ctrl(uvm_plain_any):
     microvm = uvm_plain_any
     microvm.help.enable_console()
     microvm.spawn()
-    microvm.basic_config()
+    # Testing with 1 vcpu due to rare kernel bug which raises soft lockup warning when we have
+    # multiple vcpus
```
pb8o, line 145: Is it worth keeping the "soft lockup" assert if we are avoiding it this way? Looking at the history, that assert was used to see if we had an RCU soft lockup, maybe we should be more specific in which kind of soft lockup we are testing for?

JackThomson2: Probably didn't explain it well, with the KVMCLOCK_CTRL removed we do see the `soft lockup` warning. So it does assert that the call is working. This just saves the issue with it in with multiple vcpus

kalyazin: Is there a link for the kernel bug that we can provide? Ie, what makes use believe it's a kernel bug?

bchalios: I agree with Nikita. Jack, what makes you believe that this is a kernel bug? Have you found something relevant in the LKML? If so, please could you update both the PR description, commit message and comment here to point to that and explain what the issue is?

JackThomson2: I haven't been able to find anything related to this specifically. I am assuming it is, as the ioctl has worked for the rcu_shed self-detected stall warning which doesn't print the warning, it's only the watchdog which throws this.

roypat (dismissed):

tests/integration_tests/functional/test_pause_resume.py
```diff
@@ -141,7 +141,12 @@ def test_kvmclock_ctrl(uvm_plain_any):
     microvm = uvm_plain_any
     microvm.help.enable_console()
     microvm.spawn()
-    microvm.basic_config()
+
+    # With 2 vCPUs under certain conditions soft lockup warnings can rarely be in dmesg causing this test to fail.
+    # Example of the warning: `watchdog: BUG: soft lockup - CPU#0 stuck for (x)s! [(udev-worker):758]`
+    # With 1 vCPU this intermittent issue doesn't occur. If the the KVM_CLOCK_CTRL IOCTL is not made
```
kalyazin, line 147: double "the"

JackThomson2: Fixed, thanks!

kalyazin (approved):

roypat (approved):

PR #5021, opened 2025-01-30
Fix UFFD handler never receiving `remove` messages when balloon device inflates

roypat:
When the balloon inflates, and the guest gives us back some pages of memory, we need to free those pages. In booted VMs, we do this with madvise(MADV_DONTNEED), and in restored VMs we do it by MAP_FIXED-ing a new VMA on top of the range-to-free. This is because if guest memory is a MAP_PRIVATE of a memory file, madvise has no effect.

However, we also do this MAP_FIXED trick if the snapshot is restored with UFFD. In this case, its not needed (madvise works perfectly fine), and in fact, its wrong: If we map over the memory range, UFFD will not receive Remove events for the specified range.

Fix this by only using the mmap trick for file-based restored.

After doing this, fix our test uffd handler to actually handle remove events correctly (since it never received any before, we never noticed that its was completely broken - but now that we actually exercise this in the test, it needed fixing).

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ and this project adheres to
 - [#5007](https://github.com/firecracker-microvm/firecracker/pull/5007): Fixed
   watchdog softlockup warning on x86_64 guests when a vCPU is paused during GDB
   debugging.
+- [#5021](https://github.com/firecracker-microvm/firecracker/pull/5021) If a
+  balloon device is inflated post UFFD-backed snapshot restore, Firecracker now
+  causes `remove` UFFD messages to be sent to the UFFD handler. Previously, no
+  such message would be sent.
 
 ## [1.10.1]
 
--- a/src/firecracker/examples/uffd/fault_all_handler.rs
+++ b/src/firecracker/examples/uffd/fault_all_handler.rs
@@ -36,7 +36,7 @@ fn main() {
             userfaultfd::Event::Pagefault { .. } => {
                 for region in uffd_handler.mem_regions.clone() {
                     uffd_handler
-                        .serve_pf(region.mapping.base_host_virt_addr as _, region.mapping.size)
+                        .serve_pf(region.mapping.base_host_virt_addr as _, region.mapping.size);
                 }
             }
             _ => panic!("Unexpected event on userfaultfd"),
--- a/src/firecracker/examples/uffd/uffd_utils.rs
+++ b/src/firecracker/examples/uffd/uffd_utils.rs
@@ -116,7 +116,7 @@ impl UffdHandler {
         }
     }
 
-    pub fn serve_pf(&mut self, addr: *mut u8, len: usize) {
+    pub fn serve_pf(&mut self, addr: *mut u8, len: usize) -> bool {
         // Find the start of the page that the current faulting address belongs to.
         let dst = (addr as usize & !(self.page_size - 1)) as *mut libc::c_void;
         let fault_page_addr = dst as u64;
@@ -133,14 +133,18 @@ impl UffdHandler {
                 //    event was received. This can be a consequence of guest reclaiming back its
                 //    memory from the host (through balloon device)
                 Some(MemPageState::Uninitialized) | Some(MemPageState::FromFile) => {
-                    let (start, end) = self.populate_from_file(region, fault_page_addr, len);
-                    self.update_mem_state_mappings(start, end, MemPageState::FromFile);
-                    return;
+                    match self.populate_from_file(region, fault_page_addr, len) {
+                        Some((start, end)) => {
+                            self.update_mem_state_mappings(start, end, MemPageState::FromFile)
+                        }
+                        None => return false,
+                    }
+                    return true;
                 }
                 Some(MemPageState::Removed) | Some(MemPageState::Anonymous) => {
                     let (start, end) = self.zero_out(fault_page_addr);
                     self.update_mem_state_mappings(start, end, MemPageState::Anonymous);
-                    return;
+                    return true;
                 }
                 None => {}
             }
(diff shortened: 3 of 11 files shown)


src/firecracker/examples/uffd/valid_handler.rs
```diff
@@ -26,24 +26,78 @@ fn main() {
...
-        // We expect to receive either a Page Fault or Removed
-        // event (if the balloon device is enabled).
-        match event {
-            userfaultfd::Event::Pagefault { addr, .. } => {
-                uffd_handler.serve_pf(addr.cast(), uffd_handler.page_size)
+        // !DISCLAIMER!
+        // When using UFFD together with the balloon device, this handler needs to deal with Removed
+        // and Pagefault events. There are multiple things to keep in mind in such setups:
+        //
+        // As long as any Removed event in pending in the UFFD queue, all ioctls return EAGAIN
```
kalyazin, line 33: s/in pending/is pending/

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ and this project adheres to
 - [#4926](https://github.com/firecracker-microvm/firecracker/pull/4926): Replace
   underlying implementation for seccompiler from in house one in favor of
   `libseccomp` which produces smaller and more optimized BPF code.
+- [#5021](https://github.com/firecracker-microvm/firecracker/pull/5021) Fix UFFD
```
kalyazin, line 29: Nit: This reads a bit like "we fixed UFFD handlers". We can rephrase it to make it clear we fixed Firecracker behaviour that led to that. We can also start with "If a balloon device is used" which will make some readers stop reading.

src/firecracker/examples/uffd/uffd_utils.rs
```diff
@@ -152,20 +156,39 @@ impl UffdHandler {
...
+            match self.uffd.copy(src as *const _, dst as *mut _, len, true) {
+                Ok(value) => value,
+                // Catch EAGAIN errors, which occur when a `remove` event lands in the UFFD
+                // queue while we're processing `pagefault` events.
+                // The weird cast is because the `bytes_copied` field is based on the
+                // `uffdio_copy->copy` field, which is a signed 64 bit integer, and if something
+                // goes wrong, it gets set to a -errno code. However, uffd-rs always casts this
+                // value to an unsigned `usize`, which scrambled the errno.
+                Err(Error::PartiallyCopied(bytes_copied))
+                    if bytes_copied == 0 || bytes_copied == (-libc::EAGAIN) as usize =>
```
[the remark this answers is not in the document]
roypat: Yeah, there's two paths in the kernel that can lead to EAGAIN
- the uffd is already blocked before starting the copy, in which case it short-circuits and returns -EAGAIN without touching the bytes_copied field (e.g. it remains 0 as initialized)
- the uffd gets interrupted during the actual copy operation, in which case copy is set to -EAGAIN

sadly, userfaultfd-rs mangles these return values into the above.

src/firecracker/examples/uffd/fault_all_handler.rs
```diff
@@ -36,7 +36,7 @@ fn main() {
             userfaultfd::Event::Pagefault { .. } => {
                 for region in uffd_handler.mem_regions.clone() {
                     uffd_handler
-                        .serve_pf(region.mapping.base_host_virt_addr as _, region.mapping.size)
+                        .serve_pf(region.mapping.base_host_virt_addr as _, region.mapping.size);
```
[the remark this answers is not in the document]
roypat: yep

src/firecracker/examples/uffd/valid_handler.rs
roypat: fixed

CHANGELOG.md
roypat: changed!

src/firecracker/examples/uffd/valid_handler.rs
```diff
@@ -26,24 +26,79 @@ fn main() {
...
-        // event (if the balloon device is enabled).
-        match event {
-            userfaultfd::Event::Pagefault { addr, .. } => {
-                uffd_handler.serve_pf(addr.cast(), uffd_handler.page_size)
+        // !DISCLAIMER!
+        // When using UFFD together with the balloon device, this handler needs to deal with
+        // `remove` and `pagefault` events. There are multiple things to keep in mind in
+        // such setups:
+        //
+        // As long as any `remove`` event is pending in the UFFD queue, all ioctls return EAGAIN
```
kalyazin, line 34: double backtick after `remove`

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ and this project adheres to
 - [#4926](https://github.com/firecracker-microvm/firecracker/pull/4926): Replace
   underlying implementation for seccompiler from in house one in favor of
   `libseccomp` which produces smaller and more optimized BPF code.
+- [#5021](https://github.com/firecracker-microvm/firecracker/pull/5021) If a
+  balloon device is inflated post UFFD-backed snapshot restore, Firecracker now
+  causes `remove` UFFD messages to be sent to the UFFD handler.
```
kalyazin, line 31: Now I'm thinking it should go to the "fixed" section and we can also add that those "remove" messages were previously missed (it isn't that we're adding them as an extra feature).

roypat: oops, yeah, I meant to put it into "fixed"

src/firecracker/examples/uffd/valid_handler.rs
roypat: fixed

CHANGELOG.md
roypat: fixed (lol)

kalyazin (approved):

bchalios (approved):

Issue #4522, opened 2024-03-25
Implement a solution for unmapping guest memory from kernel address space


roypat:
Currently when using anonymous memory for KVM guest RAM, the memory all remains mapped into the kernel direct map. We are looking at options to get KVM guest memory out of the kernel’s direct map as a principled approach to mitigating speculative execution issues in the host kernel. Our goal is to more completely address the class of issues whose leak origin is categorized as "Mapped memory" [[1]].

As part oft his initiative, we plan to work with the upstream Linux kernel community [[2]] to design a solution that allows us to remove a microVMs guest memory from the host kernel's address space, which we will then consume in Firecracker.

[1]: https://download.vusec.net/papers/quarantine_raid23.pdf
[2]: https://lore.kernel.org/linux-mm/cc1bb8e9bc3e1ab637700a4d3defeec95b55060a.camel@amazon.com/

PR #5031, opened 2025-02-07
refactor: move arch specific code from vm.rs into modules

roypat:
Follow a similar approach we already take for the vcpu module, where all
the architecture specific stuff is inside aarch64/x86_64 modules,
keeping only the common code in the `vm` module.

No functional change intended.

---

Some more unmotivated-at-face-value code I had lying around, which I wrote while messing around for #4522, but which constitutes an improvement standalone anyway (at least imo it makes the code a bit easier to read due to less cfg-ing)

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

Follow a similar approach we already take for the vcpu module, where all
the architecture specific stuff is inside aarch64/x86_64 modules,
keeping only the common code in the `vm` module.

No functional change intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,9 @@ and this project adheres to
   Firecracker's `--parent-cpu-time-us` values, which caused development builds
   of Firecracker to crash (but production builds were unaffected as underflows
   do not panic in release mode).
+- [#5045](https://github.com/firecracker-microvm/firecracker/pull/5045): Fixed
+  an issue where firecracker intermittently receives SIGHUP when using jailer
+  with `--new-pid-ns` but without `--daemonize`.
 
 ## [1.10.1]
 
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -330,12 +330,52 @@ impl Env {
     }
 
     fn exec_into_new_pid_ns(&mut self, chroot_exec_file: PathBuf) -> Result<(), JailerError> {
+        // https://man7.org/linux/man-pages/man7/pid_namespaces.7.html
+        // > a process in an ancestor namespace can send signals to the "init" process of a child
+        // > PID namespace only if the "init" process has established a handler for that signal.
+        //
+        // Firecracker (i.e. the "init" process of the new PID namespace) sets up handlers for some
+        // signals including SIGHUP and jailer exits soon after spawning firecracker into a new PID
+        // namespace. If the jailer process is a session leader and its exit happens after
+        // firecracker configures the signal handlers, SIGHUP will be sent to firecracker and be
+        // caught by the handler unexpectedly.
+        //
+        // In order to avoid the above issue, if jailer is a session leader, creates a new session
+        // and makes the child process (i.e. firecracker) become the leader of the new session to
+        // not get SIGHUP on the exit of jailer.
+
+        // Check whether jailer is a session leader or not before clone().
+        // Note that, if `--daemonize` is passed, jailer is always not a session leader. This is
+        // because we use the double fork method, making itself not a session leader.
+        let is_session_leader = match self.daemonize {
+            true => false,
+            false => {
+                // SAFETY: Safe because it doesn't take any input parameters.
+                let sid = SyscallReturnCode(unsafe { libc::getsid(0) })
+                    .into_result()
+                    .map_err(JailerError::GetSid)?;
+                // SAFETY: Safe because it doesn't take any input parameters.
+                let ppid = SyscallReturnCode(unsafe { libc::getpid() })
+                    .into_result()
+                    .map_err(JailerError::GetPid)?;
+                sid == ppid
+            }
+        };
+
         // Duplicate the current process. The child process will belong to the previously created
         // PID namespace. The current process will not be moved into the newly created namespace,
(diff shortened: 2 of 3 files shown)


src/vmm/src/vstate/vm/mod.rs
```diff
@@ -39,43 +38,25 @@ pub enum VmError {
...
-        let vm_fd = kvm.fd.create_vm().map_err(VmError::VmFd)?;
-
-        #[cfg(target_arch = "aarch64")]
-        {
-            Ok(Vm {
-                fd: vm_fd,
-                irqchip_handle: None,
-            })
-        }
+        let fd = kvm.fd.create_vm().map_err(VmError::VmFd)?;
```
ShadowCurse, line 56: As discussed offline, let's remove this `Vm` type and move this one line of fd creation into each `ArchVm`.

src/vmm/src/vstate/vm/mod.rs
```diff
@@ -39,43 +38,25 @@ pub enum VmError {
...
-        }
+        Ok(Vm { fd, arch })
     }
 
     /// Initializes the guest memory.
-    pub fn memory_init(
-        &self,
-        guest_mem: &GuestMemoryMmap,
-    ) -> Result<(), VmError> {
+    pub fn memory_init(&self, guest_mem: &GuestMemoryMmap) -> Result<(), VmError> {
```
ShadowCurse, line 83: Once `Vm` type is done, these impls should just extend the `ArchVm` imported from either x86 or aarch64 modules.

src/vmm/src/vstate/vm/x86_64.rs
```diff
@@ -134,6 +142,11 @@ impl Vm {
             ioapic,
         })
     }
+
+    /// Gets the list of MSRs to save when creating snapshots
+    pub fn msrs_to_save(&self) -> &[u32] {
+        self.arch.msrs_to_save.as_slice()
+    }
```
ShadowCurse, line 161: Once `arch` files will be gone, this one can be gone as well. Don't see much value here.

src/vmm/src/vstate/vm/mod.rs
```diff
@@ -55,6 +63,26 @@ impl Vm {
...
+        for cpu_idx in 0..vcpu_count {
+            let exit_evt = exit_evt.try_clone().map_err(VmError::EventFd)?;
+            let vcpu = Vcpu::new(cpu_idx, self, exit_evt).map_err(VmError::CreateVcpu)?;
+            vcpus.push(vcpu);
+        }
+
+        self.arch_post_create_vcpus(vcpu_count)?;
+
+        Ok((vcpus, exit_evt))
+    }
```
ShadowCurse, line 80: Maybe just copy it for each arch. This one loop is not worth doing this `pre_*`/`post_*` shenanigans.

src/vmm/src/vstate/vm/mod.rs
roypat: Yeah, with just the `fd` shared between architectures there's not much point in having this. Originally I had also moved the `GuestMemoryMmap` stuff into `Vm`, but since that never made the PR there's no point in keeping around `struct Vm` in preparation (of something which might never happen - and if it does we can always change it again)

src/vmm/src/vstate/vm/mod.rs
roypat: Yup, done

src/vmm/src/vstate/vm/x86_64.rs
roypat: This one contains the `.as_slice()` call, so it is slightly different than just accessing the field. I'd prefer to keep it tbh

src/vmm/src/vstate/vm/mod.rs
roypat: This one I prefer the pre_*/post_* thingy over duplicating the function. It's how the kernel implements these things, and it seems to work pretty well over there

src/vmm/src/vstate/kvm.rs
```diff
@@ -82,19 +76,22 @@ impl Kvm {
...
             })
         }
     }
 
+    /// Msrs needed to be saved on snapshot creation.
+    #[cfg(target_arch = "x86_64")]
+    pub fn msrs_to_save(&self) -> Result<MsrList, crate::arch::x86_64::msr::MsrError> {
+        crate::arch::x86_64::msr::get_msrs_to_save(&self.fd)
+    }
+
```
bchalios, line 94: Don't have to do it in this PR. But do you think it would make sense to move this arch-specific logic in its own dedicated modules as well? I can do it as a follow up

bchalios (dismissed):

src/vmm/src/vstate/kvm.rs
roypat: Yeah, I was thinking about that too, but wanted to keep this PR from getting too big. Definitely agree on it being a good idea though

ShadowCurse: ok

ShadowCurse (approved):

ShadowCurse (approved):

bchalios (approved):

Issue #5033, opened 2025-02-09
[Bug] Fail to launch daemonize jailer v1.10.1


xuehaonan27:
# Describe the bug

`jailer`(v1.10.1) fail to run with flag `--daemonize`, and could run without flag `--daemonize`.

## To Reproduce

1. Built `jailer` binary from tag `v1.10.1`.
2. Run following command:
```shell
/root/firecracker_src/build/cargo_target/x86_64-unknown-linux-musl/debug/jailer \
    --exec-file /root/firecracker_src/build/cargo_target/x86_64-unknown-linux-musl/debug/firecracker \
    --id "test-instance" \
    --gid 100 \
    --uid 123 \
    --daemonize
```
3. The process exits immediately.
4. Run following command (without `--daemonize` this time)
```shell
/root/firecracker_src/build/cargo_target/x86_64-unknown-linux-musl/debug/jailer \
--exec-file /root/firecracker_src/build/cargo_target/x86_64-unknown-linux-musl/debug/firecracker \
--id "test-instance" \
--gid 100 \
--uid 123
```
5. The console hang with `[test-instance:main] Running Firecracker v1.10.1` which means `jailer` is running well.
6. Open another console, run `tree /srv/jailer/firecracker/test-instance/` and could see output like this:
```shell
/srv/jailer/firecracker/test-instance/
└── root
    ├── dev
    │   ├── kvm
    │   ├── net
    │   │   └── tun
    │   ├── urandom
    │   └── userfaultfd
    ├── firecracker
    ├── firecracker.pid
    └── run
        └── firecracker.socket
```
There's a `firecracker.socket` which means it's running well too.
 
## Expected behaviour

When running `jailer` with `--daemonize`, the process could be started and expose the API socket for communication.

## Environment

- Firecracker version: v.1.10.1
- Host and guest kernel versions: Host kernel `6.8.0-51-generic`; Guest kernel not used (bare jailer process).
- Rootfs used: Not used (bare jailer process).
- Architecture: x86_64 (x86_64 x86_64 x86_64 GNU/Linux)

## Additional context

`[Author TODO: How has this bug affected you?]`
I am developing a firecracker SDK in Rust, this bug prevents me to test functionality with daemonized instances.

`[Author TODO: What are you trying to achieve?]`
Start the daemonized jailer and communicate with it via API socket.

`[Author TODO: Do you have any idea of what the solution might be?]`
[...]

PR #5034, opened 2025-02-10
fix: compute jailer cpu time without underflow 

roypat:
Oh, here we go again. 

The jailer wishes to compute the CPU time it spents before exec-ing into
Firecracker, so that it can be passed to Firecracker's `parent-cpu-us`
flag and emitted as a metric.

The CLOCK_PROCESS_CPUTIME_ID is reset on every fork() call, so we need
to stitch together the different values from before the first fork, and
then between each consecutive fork. However, we also need to account for
the scenarios that we do not fork() at all (no daemonization).

Now, this all goes slightly wrong because we want to exclude the time
spent the call to Env::run() from the reported time (note that inside
the jailer, this is only the time spent parsing arguments, but if a user
ended up exec()-ing into the jailer, it would also include the time from
prior to calling exec(), which makes sense to exclude). So we grab the
value of CLOCK_PROCESS_CPUTIME_ID right after parsing arguments, and
subtract it at the end. Sadly, in the case of daemonization, we subtract
it not from the total amount of time spent in the jailer, but rather
from the time spent since the final fork(). This works out if the time
spent since the last fork was greater than the time spent parsing
arguments, but if not, we have an integer underflow on our hands. Now,
this is only a problem in debug builds, and in release builds, we
underflow, and then _over_flow again during the next addition of the
underflowed value to jailer_cpu_time_us, and Firecracker even seens a
sane value for parent_time_cpu_us. But in debug builds, we panic.

Fix this by breaking the subtraction and addition into two separate
statements, to avoid relying on the double over-/underflow

Closes https://github.com/firecracker-microvm/firecracker/issues/5033

Fixes: https://github.com/firecracker-microvm/firecracker/commit/68ab56eb242ae2bca3b2918759d524aa26a66c27
## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -3,13 +3,13 @@
 
 use std::ffi::{CString, OsString};
 use std::fs::{self, canonicalize, read_to_string, File, OpenOptions, Permissions};
+use std::io;
 use std::io::Write;
 use std::os::unix::fs::PermissionsExt;
 use std::os::unix::io::AsRawFd;
 use std::os::unix::process::CommandExt;
 use std::path::{Component, Path, PathBuf};
 use std::process::{exit, id, Command, Stdio};
-use std::{fmt, io};
 
 use utils::arg_parser::UtilsArgParserError::MissingValue;
 use utils::time::{get_time_us, ClockType};
@@ -114,6 +114,7 @@ enum UserfaultfdParseError {
     NotFound,
 }
 
+#[derive(Debug)]
 pub struct Env {
     id: String,
     chroot_dir: PathBuf,
@@ -130,26 +131,6 @@ pub struct Env {
     cgroup_conf: Option<CgroupConfiguration>,
     resource_limits: ResourceLimits,
     uffd_dev_minor: Option<u32>,
-}
-
-impl fmt::Debug for Env {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("Env")
-            .field("id", &self.id)
-            .field("chroot_dir", &self.chroot_dir)
-            .field("exec_file_path", &self.exec_file_path)
-            .field("uid", &self.uid)
-            .field("gid", &self.gid)
-            .field("netns", &self.netns)
-            .field("daemonize", &self.daemonize)
-            .field("new_pid_ns", &self.new_pid_ns)
-            .field("start_time_us", &self.start_time_us)
-            .field("jailer_cpu_time_us", &self.jailer_cpu_time_us)
-            .field("extra_args", &self.extra_args)
-            .field("cgroups", &self.cgroup_conf)
-            .field("resource_limits", &self.resource_limits)
-            .finish()
-    }
 }
 
 impl Env {
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -696,7 +696,8 @@ impl Env {
         }
 
         // Compute jailer's total CPU time up to the current time.
-        self.jailer_cpu_time_us += get_time_us(ClockType::ProcessCpu) - self.start_time_cpu_us;
+        self.jailer_cpu_time_us += get_time_us(ClockType::ProcessCpu);
+        self.jailer_cpu_time_us -= self.start_time_cpu_us;
         // Reset process start time.
         self.start_time_cpu_us = 0;
 
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -645,11 +645,10 @@ impl Env {
             self.mknod_and_own_dev(DEV_UFFD_PATH, DEV_UFFD_MAJOR, minor)?;
         }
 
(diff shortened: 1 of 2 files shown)


src/jailer/src/env.rs
```diff
@@ -715,9 +696,8 @@ impl Env {
         }
 
         // Compute jailer's total CPU time up to the current time.
-        self.jailer_cpu_time_us += get_time_us(ClockType::ProcessCpu) - self.start_time_cpu_us;
-        // Reset process start time.
-        self.start_time_cpu_us = 0;
+        self.jailer_cpu_time_us += get_time_us(ClockType::ProcessCpu);
+        self.jailer_cpu_time_us -= self.start_time_cpu_us;
```
Manciukic, line 700: I didn't quite yet wrap my head around why this calculation is what we want the `jailer_cpu_time_us` to be. It's definitely measuring different values between the daemonize and non-daemonize case. Which is the right one? Should the calculation be different in case of daemonize?

roypat: In the daemonize case, the time spent doing the triple-fork() dance is added. If we daemonize, then the addition in ln 699 accounts for the time since the last fork(). In the non-daemonize case, it accounts for the overall time spent in the jailer (in the daemonize case, the counterpart of this is ln 648) and in this case it shouldn't be read as an addition, but rather as an initialization (the previous value will have been zero).

roypat: admittedly can probably be clearer to follow by hoisting lines 699/700 and 648 above the if statement... let me try something in an additional commit

roypat: does 5d2da197b418ac4a34e57b5e9eb4ee2d7ccf3fc7 clarify things?

Manciukic: ok, I missed the assignment in line 648 (this code is so confusing lol). So what happens is:
 - Non daemonize: `jailer_cpu_time_us`  is `0` at line 699, to which we add the current process cpu time and subtract the `start_time_cpu_us`.
 - Daemonize: `jailer_cpu_time_us`  is set to the parent cpu time at line 648, to which the first child is added at line 673, and the second at line 699. Finally we subtract the `start_time_cpu_us`.

Would it be simpler to follow if we rewrote it as follows?
```
        // line 644
        // jailer overhead is the difference between the current (jailer) cpu time and start_time_cpu_us
        self.jailer_cpu_time_us = get_time_us(ClockType::ProcessCpu) - self.start_time_cpu_us;
        
        // Daemonize before exec, if so required (when the dev_null variable != None).
        if let Some(dev_null) = dev_null {
            fork()
            // add the cpu time added by the first child
            self.jailer_cpu_time_us += get_time_us(ClockType::ProcessCpu);
            fork()
            dup() x3
            // add the cpu time added by the second child
            self.jailer_cpu_time_us += get_time_us(ClockType::ProcessCpu);
        }
        // proceed to do the exec
```

It would work for both cases as the only part that wouldn't be measured in the non-daemonize case would be the if itself.

roypat: Yes, that's pretty much exactly what I committed above :)

zulinx86 (approved):

src/jailer/src/env.rs
```diff
@@ -696,7 +696,8 @@ impl Env {
         }
 
         // Compute jailer's total CPU time up to the current time.
-        self.jailer_cpu_time_us += get_time_us(ClockType::ProcessCpu) - self.start_time_cpu_us;
+        self.jailer_cpu_time_us += get_time_us(ClockType::ProcessCpu);
+        self.jailer_cpu_time_us -= self.start_time_cpu_us;
```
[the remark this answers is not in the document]
roypat: This code doesnt exist anymore in the latest version :/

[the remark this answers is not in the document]
zulinx86: aaaah, right! sorry, nvm :P

src/jailer/src/env.rs
```diff
@@ -495,7 +476,10 @@ impl Env {
         Command::new(chroot_exec_file)
             .args(["--id", &self.id])
             .args(["--start-time-us", &self.start_time_us.to_string()])
-            .args(["--start-time-cpu-us", &self.start_time_cpu_us.to_string()])
+            .args([
+                "--start-time-cpu-us",
+                &get_time_us(ClockType::ProcessCpu).to_string(),
```
Manciukic, line 481: did you double check that now the metric reported by firecracker makes sense and doesn't double count?

roypat: Yeah, before: 
```
Process startup time is: 4841 us (9554 CPU us)
```

after:
```
Process startup time is: 3456 us (3371 CPU us, 0 in parent)
```

(turns out the reason I couldn't get it to reproduce at all initially was because our test framework sneaks in a `--new-pid-ns`, which again causes a fork())

Manciukic (approved):

Issue #1761, opened 2020-04-08
Have MMDS as a plug in component of Virtio Network


andreeaflorescu:
Right now the MMDS functionality is backed into the virtio net implementation. We should find a way of separating this or making it pluggable so that we can consume Virtio Network from rust-vmm when it is going to be available.

iulianbarbu:
This is somehow the opposite of https://github.com/firecracker-microvm/firecracker/issues/1768. From a separation of concerns standpoint, both will achieve the separation. However, this separation is similar to the current dependencies present in Firecracker, and might work with less fuss than #1768, which needs a new design, or at least thoughts/discussions about what can be done differently to support it. Mentioned this here to keep these issues close, and have both suggestions at hand when resolving them.

andreeaflorescu:
Hmm, why do you say they're opposite?

I think of them as separate issues. This is related to the VirtioNetwork interface and how we make the interface/implementation not depend on Mmds. This issue doesn't really care where the MmdsNetworkStack is defined. It is just about separating the core functionality of the virtio network device from mmds in such a way that mmds can easily become a feature for example.

iulianbarbu:
> Hmm, why do you say they're opposite?
> 
> I think of them as separate issues. This is related to the VirtioNetwork interface and how we make the interface/implementation not depend on Mmds. This issue doesn't really care where the MmdsNetworkStack is defined. It is just about separating the core functionality of the virtio network device from mmds in such a way that mmds can easily become a feature for example.

I said that their opposite because of the the fact that with pluggable MMDS, Firecracker virtio net will still depend on the MMDS network stack, while #1768 suggests to reverse this dependency. However, reading again what you're thinking at, it looks like the changes might be very similar, and separate, as you say, or sequential maybe. I take back what I've said. :D

gc-plp:
Related to https://github.com/firecracker-microvm/firecracker/issues/518

roypat:
Marking this as parked, as untangling MMDS from the networking path is something we are still interesting in, yet do not have any immediate plans to work on

jyliu02:
Hello, I'm interested in this issue. My preliminary idea is to use a eBPF filter that redirects MMDS packets to a dedicated MMDS handling component (in this way I can reuse a lot of existing code). I started looking into this while exploring ways to boost network performance. It seems that introducing new features like vhost-user-net would be very tricky without first decoupling MMDS from the current network stack.

[...]

PR #5036, opened 2025-02-10
[RFC] Decouple MMDS service from network stack

jyliu02:
## Changes

- Implemented an vsock-based HTTP server for MMDS.
- Optional MMDS server API and configuration.

## Reason

- MMDS traffic is much less frequent than standard network I/O, so it benefits networking-intense application by separating MMDS from the fast data path. And it is acceptable for MMDS to be a bit slower.
- This design simplifies future attempts at `vhost-user-net` or other data-plane offloading by preventing MMDS from interfering with typical network traffic.
- Modularity & Maintainability: allows independent development of MMDS logic and eases the burden of maintaining the complex Dumbo stack.
- The guest can issue requests via `socat` directly or using `curl` by bridging an IP to vsock via `socat`.

Related to #1761

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/src/firecracker/src/api_server/parsed_request.rs
+++ b/src/firecracker/src/api_server/parsed_request.rs
@@ -21,7 +21,9 @@ use super::request::machine_configuration::{
     parse_get_machine_config, parse_patch_machine_config, parse_put_machine_config,
 };
 use super::request::metrics::parse_put_metrics;
-use super::request::mmds::{parse_get_mmds, parse_patch_mmds, parse_put_mmds};
+use super::request::mmds::{
+    parse_get_mmds, parse_patch_mmds, parse_put_mmds, parse_put_mmds_server_config,
+};
 use super::request::net::{parse_patch_net, parse_put_net};
 use super::request::snapshot::{parse_patch_vm_state, parse_put_snapshot};
 use super::request::version::parse_get_version;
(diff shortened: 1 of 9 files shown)


bchalios:
Hi @jyliu02! Thanks a lot for your contribution! We'd also like to apologise for not responding to [your comment](https://github.com/firecracker-microvm/firecracker/issues/1761#issuecomment-2614134834). We somehow missed your comment when reviewing open source activity.

That said, this topic is an interesting one for us. We would like to decouple MMDS from the emulation of the network device. However, we would like to think a bit more carefully the available options for doing so.

One thing that we definitely need to consider is backwards compatibility. We have users that already consume the existing MMDS API, that is the one where clients in the guest speak to link-local addresses to interact with MMDS. We simply cannot drop the existing support. In that sense your changes make perfect sense.

However, we still want to discuss a bit internally our options before committing to any given solution. We will discuss this internally and let you know, how we would like to proceed.

jyliu02:
Thanks for the comment. Backwards compatibility is also my first design consideration. The current design is still sketchy, but I’ll commit to a polished version once we can settle on a direction. Looking forward to further feedback.

bchalios:
Hey @jyliu02 :wave: 

Sorry for taking long to come back to you on this. We finally found time to go over this issue and discuss it a bit more internally. So, our initial motivation for opening issue https://github.com/firecracker-microvm/firecracker/issues/1761 was that, back then, we were contemplating to drop our own implementation of virtio-net and start using VirtIO net emulation from rust-vmm crates (which did not support MMDS).

Since then, we have invested a lot of work into improving our own virtio-net implementation and we don't plan to migrate to the rust-vmm implementation any more. That means that our main motivation for decoupling has gone.

That said, decoupling the implementation of MMDS and virtio-net is a nice-to-have, engineering wise, but it would most likely require _some_ change from users inside the guest (consuming a new API, or using `socat` to direct traffic to link-local addresses all involve the users of changing their software stack).

We do not believe that the value we get from doing this decoupling justifies us asking users to change their software. As a result, we don't believe it makes to invest more time in this currently. I will go ahead and close this PR for now. Please feel free to open a new PR or get in touch with us in the comment section here if you believe there are ways to switch the implementation without breaking existing users of IMDS.

Again, thanks a lot for your contribution!

PR #5041, opened 2025-02-14
test: catch correct exception in memory.py

roypat:
The access to the `firecracker_pid` property causes a read to the vm's
pid file. This means that in the scenario where the VM is already dead
by the time the monitor starts running, we won't get a `NoSuchProcess`
error, we potentially get a `FileNotFoundError` (if the jail is already
cleaned up). Thus, to avoid spurious failures, also catch
`FileNotFoundError`.

We leave the catch for `NoSuchProcess` error in case a VM exits between
reading the pid file and the call to the`psutils.Process` constructor
(race condition).

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -336,11 +336,12 @@ def microvm_factory(request, record_property, results_dir, netns_factory):
     # if the test failed, save important files from the root of the uVM into `test_results` for troubleshooting
     report = request.node.stash[PHASE_REPORT_KEY]
     if "call" in report and report["call"].failed:
-        dmesg = utils.run_cmd(["dmesg", "-dPx"])
         for uvm in uvm_factory.vms:
             uvm_data = results_dir / uvm.id
             uvm_data.mkdir()
-            uvm_data.joinpath("host-dmesg.log").write_text(dmesg.stdout)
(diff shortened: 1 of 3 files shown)


pb8o (approved):

tests/conftest.py
```diff
@@ -336,11 +336,12 @@ def microvm_factory(request, record_property, results_dir, netns_factory):
     # if the test failed, save important files from the root of the uVM into `test_results` for troubleshooting
     report = request.node.stash[PHASE_REPORT_KEY]
     if "call" in report and report["call"].failed:
-        dmesg = utils.run_cmd(["dmesg", "-dPx"])
         for uvm in uvm_factory.vms:
             uvm_data = results_dir / uvm.id
             uvm_data.mkdir()
-            uvm_data.joinpath("host-dmesg.log").write_text(dmesg.stdout)
+            uvm_data.joinpath("host-dmesg.log").write_text(
```
pb8o, line 342: Thanks! This was indeed very annoying.

kalyazin (approved):

PR #5045, opened 2025-02-20
fix(jailer): Make "init" in new PID NS a session leader

zulinx86:
## Changes & Reasons

https://man7.org/linux/man-pages/man7/pid_namespaces.7.html
> A process in an ancestor namespace can send signals to the "init"
> process of a child PID namespace only if the "init" process has
> established a handelr for that signal.

Firecracker (i.e. the "init" process of the new PID namespace) sets up handlers for some signals including SIGHUP and jailer exits soon after spawning firecracker into the new PID namespace. If the jailer process is a session leader and its exit happens after firecracker configures the signal handlers, SIGHUP will be sent to firecracker and be caught by the handler unexpectedly.

In order to avoid the above issue, if jailer is a session leader, creates a new session and makes the child process (i.e. firecracker) become the leader of the new session to not get SIGHUP on the exit of jailer.

Note that this is the case only if `--daemonize` is not passed. This is because we use the double fork method to daemonize, making itself not a session leader.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.


roypat (dismissed):
Wohooo, nice find on the "init is special" thing! 

I wonder if its worth it to skip the two syscalls (getpid, getsid) on the `--daemonize` path (because there we statically know we dont run into session leader issues), but probably not worth it.

zulinx86:
> sudo screen -L -Logfile screen.log -dm build/cargo_target/x86_64-unknown-linux-musl/release/jailer --new-pid-ns --exec-file build/cargo_target/x86_64-unknown-linux-musl/release/firecracker --gid 123 --uid 123 --id test --chroot-base-dir root/

Good question. Inserting a single branch (whether `--daemonize` is passed or not) should not impact performance, even if the branch predictor mis-speculates to the path of calling those syscalls. Let's leverage the domain knowledge!

roypat (approved):

kalyazin (approved):

Manciukic:
should we add an integration test to check all the combinations of `--daemonize` and `--new-pid-ns` to verify the fix?

PR #5048, opened 2025-02-24
Support for PVH boot protocol in Firecracker

roypat:
This is a rebased version of the feature/pvh feature branch contributed by @cperciva a while ago. 

In Linux guests supporting the PVH protocol (which is all linux guests compiled with `CONFIG_PVH=y`, a default), Firecracker will now boot guests using the PVH boot protocol. Our performance testing has shown there to be no difference in performance to the old way of booting using the Linux boot protocol.

Functionally, this means Firecracker can now boot FreeBSD guests (which only support the PVH boot protocol).

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,11 @@ and this project adheres to
   kernels. For older kernels physical counter will still be passed to the guest
   unmodified. See more info
   [here](https://github.com/firecracker-microvm/firecracker/blob/main/docs/prod-host-setup.md#arm-only-vm-physical-counter-behaviour)
+- [#5048](https://github.com/firecracker-microvm/firecracker/pull/5048): Added
+  support for [PVH boot mode](docs/pvh.md). This is used when an x86 kernel
+  provides the appropriate ELF Note to indicate that PVH boot mode is supported.
+  Linux kernels newer than 5.0 compiled with `CONFIG_PVH=y` set this ELF Note,
+  as do FreeBSD kernels.
 
 ### Changed
 
--- a/docs/pvh.md
+++ b/docs/pvh.md
@@ -1,0 +1,15 @@
+# PVH boot mode
+
+Firecracker supports booting x86 kernels in "PVH direct boot" mode
+[as specified by the Xen project](https://github.com/xen-project/xen/blob/master/docs/misc/pvh.pandoc).
+If a kernel is provided which contains the XEN_ELFNOTE_PHYS32_ENTRY ELF Note
+then this boot mode will be used. This boot mode was designed for virtualized
+environments which load the kernel directly, and is simpler than the "Linux
+boot" mode which is designed to be launched from a legacy boot loader.
+
+PVH boot mode can be enabled for Linux by setting `CONFIG_PVH=y` in the kernel
+configuration. (This is not the default setting.)
+
+PVH boot mode is enabled by default in FreeBSD, which has support for
+Firecracker starting with FreeBSD 14.0. Instructions on building a FreeBSD
+kernel and root filesystem are available [here](rootfs-and-kernel-setup.md).
--- a/docs/rootfs-and-kernel-setup.md
+++ b/docs/rootfs-and-kernel-setup.md
@@ -1,6 +1,6 @@
 # Creating Custom rootfs and kernel Images
 
-## Creating a kernel Image
+## Creating a Linux kernel Image
 
 ### Manual compilation
 
@@ -79,7 +79,7 @@ but without ACPI support) and `6.1`.
 After the command finishes, the kernels along with the corresponding KConfig
 used will be stored under `resources/$(uname -m)`.
 
-## Creating a rootfs Image
+## Creating a Linux rootfs Image
 
 A rootfs image is just a file system image, that hosts at least an init system.
 For instance, our getting started guide uses an ext4 filesystem image. Note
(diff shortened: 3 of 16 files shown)


src/vmm/src/arch/x86_64/mod.rs
```diff
@@ -110,13 +125,146 @@ pub fn initrd_load_addr(
...
+    // Add the table entry to the vector
+    memmap.push(hvm_memmap_table_entry {
+        addr,
+        size,
+        type_: mem_type,
+        reserved: 0,
+    });
+
+    Ok(())
+}
```
ShadowCurse, line 262: just inline this

docs/rootfs-and-kernel-setup.md
```diff
@@ -185,3 +185,45 @@ adjust the script(s) to suit your use case.
...
+
+1. Install git and check out the FreeBSD src tree:
+
+   ```sh
+   pkg install -y git
+   git clone https://git.freebsd.org/src.git /usr/src
+   ```
+
+   At present (July 2023) Firecracker support is only present in the `main`
+   branch.
```
ShadowCurse, line 214: Not very present anymore.

docs/pvh.md
```diff
@@ -0,0 +1,15 @@
...
+If a kernel is provided which contains the XEN_ELFNOTE_PHYS32_ENTRY ELF Note
+then this boot mode will be used.  This boot mode was designed for virtualized
+environments which load the kernel directly, and is simpler than the "Linux
+boot" mode which is designed to be launched from a legacy boot loader.
+
+PVH boot mode can be enabled for Linux by setting CONFIG_XEN_PVH=y in the
+kernel configuration.  (This is not the default setting.)
+
+PVH boot mode is enabled by default in FreeBSD, which has support for
+Firecracker starting with FreeBSD 14.0.  Instructions on building a FreeBSD
```
ShadowCurse: Is FreeBSD support for Firecracker correlated to the Firecracker support of booting FreeBSD?

tests/integration_tests/style/test_gitlint.py
```diff
@@ -4,10 +4,16 @@
...
+import pytest
+
 from framework import utils
 from framework.ab_test import DEFAULT_A_REVISION
 
 
+@pytest.mark.skipif(
+    os.environ.get("BUILDKITE_PULL_REQUEST") == "5048",
+    reason="PR of a feature branch from before this test was modified to follow Linux sign-off rules for co-authors.",
+)
```
ShadowCurse, line 16: need to add a TODO to remove it after merge

ShadowCurse (commented):
How hard would it be to merge some commits and give them more standard names(`feat:`, ...)?

docs/pvh.md
cperciva: Yes.  FreeBSD cannot boot on non-PVH Firecracker, but with these changes FreeBSD is able to boot on Firecracker.

ShadowCurse: yes, with this patch FC can boot FreeBSD. It is just the part `which(FreeBSD I assume) has support for Firecracker starting with FreeBSD 14.0` is confusing. Is PVH enabled by default in FreeBSD only from 14.0 version?

cperciva: Oh, FreeBSD has had PVH support for longer than that, but wouldn't boot in Firecracker/PVH due to other issues (assuming that PVH booting = support for Xen hypercalls, lacking virtio mmio cmdline support, etc).  So it really is specifically Firecracker support which landed in FreeBSD 14.0.

roypat:
> How hard would it be to merge some commits and give them more standard names(`feat:`, ...)?

Well, it's the merge of a feature branch, they're ought to be a little bit different from what we're used to. We never rewrote feature branches on merge to make the commit history more like a normal PR. I guess we could here because its small, but I don't really think we need/should.

cperciva:
> > How hard would it be to merge some commits and give them more standard names(`feat:`, ...)?
> 
> Well, it's the merge of a feature branch, they're ought to be a little bit different from what we're used to. We never rewrote feature branches on merge to make the commit history more like a normal PR. I guess we could here because its small, but I don't really think we need/should.

For the record: I'm totally fine with having my commit messages rewritten to match house style, if the Firecracker team wants to do this.  I kept the original author's style for his commits that I reworked, and then just used FreeBSD commit message style for my own commits simply because that's what I'm used to.

I know I'm a guest here. :-)

docs/pvh.md
ShadowCurse: Ok now it makes sense, thx for the explanation.

roypat:
> Well, ok, let commit message be as they are. But I feel [5d37abd](https://github.com/firecracker-microvm/firecracker/commit/5d37abd2f338191c157b7625f0400dcb780fb31b) [b95ea58](https://github.com/firecracker-microvm/firecracker/commit/b95ea58f8e276e33914bf085d123c95aabcc71c0) and [aaf9609](https://github.com/firecracker-microvm/firecracker/commit/aaf96094fb415801fc0de0216fff064945072568) commits should be merged with original commits. We have a unofficial rule that all commits should compile, so let's at least keep this.

okay I've folded all the lint fixes/etc into their respective commits. probably should've just done that in the first place whenever I rebased this branch anyway.

bchalios (approved):

docs/rootfs-and-kernel-setup.md
ShadowCurse: I think here we can say to clone 14.0 version since it is one with FC support

roypat: changed

CHANGELOG.md
```diff
@@ -192,6 +192,10 @@ and this project adheres to
   [random for clones](docs/snapshotting/random-for-clones.md) documention for
   more info on VMGenID. VMGenID state is part of the snapshot format of
   Firecracker. As a result, Firecracker snapshot version is now 2.0.0.
+- Added support for PVH boot mode.  This is used when an x86 kernel provides
+  the appropriate ELF Note to indicate that PVH boot mode is supported.
+  Linux kernels compiled with CONFIG_XEN_PVH=y set this ELF Note, as do
+  FreeBSD kernels.
```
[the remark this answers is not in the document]
roypat: done

src/vmm/src/arch/x86_64/mod.rs
roypat: I don't really see any strong argument for doing it either way. I kinda like this admittedly because calling there's less visual clutter. Inlining it would also mean doing the constructing the struct at all 8 call sites, which involves explicitly writing out all fields, because outside of this functions the local variable names don't match up with the struct names.I

It doesnt need to return `Result` though, I'll give you that.

roypat: @bchalios heya, could you weigh in here to resolve our deadlock? :D

bchalios: Personally, I would just call `memap.push` directly and get rid of this function all together. Wrapping this in a function doesn't add any value whatsoever. There's nothing to abstract and it doesn't really remove that much of clutter :)

roypat: Okay, done

bchalios (approved):

ShadowCurse (approved):

Manciukic (approved):

PR #5062, opened 2025-03-05
fix: Allow jailer to continue with partial CPU cache info

gudmundur:
## Changes

Allows the `jailer` to continue when files in `/sys/devices/system/cpu/cpuX/cache/indexY/` are missing.

## Reason

Nested virtualization support was added on macOS Sequoia for Apple M3/M4 chips. When using `vz` with [Lima](https://lima-vm.io) a few of the CPU cache information files are missing, namely `size`, `coherency_line_size`, and `number_of_sets`. `firecracker` itself [handles if these files are missing](https://github.com/firecracker-microvm/firecracker/blob/b4cb9d311d026c5e45f13bd56089d2db024ec983/src/vmm/src/arch/aarch64/cache_info.rs#L101-L124), while the `jailer` doesn't . This patch passes whatever is present to the copied cache information, and allows `firecracker` to determine if it is valid or not.

By making this change, we are able to simplify our development environments drastically by not requiring either GitHub Codespaces or AWS EC2 metal instances.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
~- [ ] When making API changes, I have followed the
  [Runbook for Firecracker API changes][2].~
  integration tests.
~- [ ] I have linked an issue to every new `TODO`.~

### DIFF
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -575,17 +575,20 @@ impl Env {
                 let host_cache_file = host_path.join(entry);
                 let jailer_cache_file = jailer_path.join(entry);
 
-                let line = readln_special(&host_cache_file)?;
-                writeln_special(&jailer_cache_file, line)?;
+                if let Ok(line) = readln_special(&host_cache_file) {
+                    writeln_special(&jailer_cache_file, line)?;
 
-                // We now change the permissions.
-                let dest_path_cstr = to_cstring(&jailer_cache_file)?;
-                // SAFETY: Safe because dest_path_cstr is null-terminated.
-                SyscallReturnCode(unsafe {
-                    libc::chown(dest_path_cstr.as_ptr(), self.uid(), self.gid())
-                })
-                .into_empty_result()
-                .map_err(|err| JailerError::ChangeFileOwner(jailer_cache_file.to_owned(), err))?;
+                    // We now change the permissions.
+                    let dest_path_cstr = to_cstring(&jailer_cache_file)?;
+                    // SAFETY: Safe because dest_path_cstr is null-terminated.
+                    SyscallReturnCode(unsafe {
+                        libc::chown(dest_path_cstr.as_ptr(), self.uid(), self.gid())
+                    })
+                    .into_empty_result()
+                    .map_err(|err| {
+                        JailerError::ChangeFileOwner(jailer_cache_file.to_owned(), err)
+                    })?;
+                }
             }
         }
         Ok(())


gudmundur:
With this change, the `jailer` will start `firecracker` after printing out these lines:

```
Could not read the size, coherency_line_size, number_of_sets for cache level 0.
Could not read the size, coherency_line_size, number_of_sets for cache level 1.
Could not read the size, coherency_line_size, number_of_sets for cache level 2.
```

src/jailer/src/env.rs
```diff
@@ -575,17 +577,35 @@ impl Env {
...
+                    SyscallReturnCode(unsafe {
+                        libc::chown(dest_path_cstr.as_ptr(), self.uid(), self.gid())
+                    })
+                    .into_empty_result()
+                    .map_err(|err| {
+                        JailerError::ChangeFileOwner(jailer_cache_file.to_owned(), err)
+                    })?;
+                } else {
+                    err_str += entry;
+                    err_str += ", ";
```
ShadowCurse, line 595: Instead of building an error string you can just print the error inline:
```rust
eprintln!("Cache level {index}: {entry} entry not found, skipping");
```
This is a simpler approach, event thought it will produce more output.

ShadowCurse (commented):
Hi @gudmundur, thank you for the PR. Even though Firecracker officially does not support nested virtualization usage, it will be nice to have a consistent behavior between Jailer and Firecracker.

src/jailer/src/env.rs
roypat: Does the jailer even have to print this at all if Firecracker is gonna print it later anyway?

gudmundur: @ShadowCurse @roypat Happy to update for either case, I'll let you decide which you'd prefer. 

I'm assuming that if I updating for @roypat's version, the code should be updated to continue in the absence of a file, rather than blanket copying anything in `/sys/devices/system/cpu/cpuX/cache/indexY/` over?

roypat: I was thinking we just drop all of the string manipulations and the `else` branch on the `if let`, and leave the rest as is

gudmundur: Addressed in [426b027](https://github.com/firecracker-microvm/firecracker/pull/5062/commits/426b027d0d979fa238ce09214ae3f2571d48a899). The commit also contains a body, so it should pass the previously failing check.

roypat (approved):

src/jailer/src/env.rs
roypat: Thanks, looks good to me!

Manciukic (approved):

PR #5065, opened 2025-03-05
feat: Intel AMX support

zulinx86:
This is the first PR for Intel Sapphire Rapids (EC2 7th-gen Intel instance type) support.

**Note that this PR focuses on Intel AMX support and any integration tests for Intel Sapphire Rapids will be added in the upcoming PR.**

## Changes

- Request permission for Intel AMX via `arch_prctl()` before `KVM_GET_SUPPORTED_CPUID` to boot guests successfully
- Change XSAVE type from `kvm_xsave` to support snapshot restore of Intel AMX

## Reason

Intel AMX (Advanced Matrix Extensions) is introduced in Intel Sapphire Rapids and a new instruction set for deep learning / AI workloads. Intel AMX is supported in `XSAVE`/`XRSTOR` that are instructions to save/restore extensional CPU features' states into memory (e.g. for context switch). Which states to be saved/restored is configured by writing a bit vector to XCR0 via `XSETBV` instruction. Intel AMX introduces two new bits (TILECFG and TILEDATA) in (1) the bit vector (`XCR0.TILECFG[bit 17]` and `XCR0.TILEDATA[bit 18]`) as well as (2) CPUID to enumerate their supports (`CPUID.(EAX=0DH,ECX=0):EAX.TILECFG[bit 17]` and `CPUID.(EAX=0DH,ECX=0):EAX.TILEDATA[bit 18]`).

Since the memory size required to save TILEDATA state is 8KB and it is larger than previously statically allocated memory size (4KB), Linux kernel decided to disable TILEDATA by default and allows userspace applications to enable it dynamically via [`arch_prctl()` syscall][kernel_doc_xstate]. This default disabling behavior is also the case with KVM. To enable TILEDATA for guests, VMM has to call `arch_prctl(ARCH_REQ_XCOM_GUEST_PERM, ...)`, which makes `KVM_GET_SUPPORTED_CPUID` return a value with `CPUID.(EAX=0DH,ECX=0):EAX.TILEDATA[bit 18]` set. Conversely, without `arch_prctl()` prior to `KVM_GET_SUPPORTED_CPUID`, it returns an inconsistent state where `CPUID.(EAX=0DH,ECX=0):EAX.TILECFG[bit 17]` is set but `CPUID.(EAX=0DH,ECX=0):EAX.TILEDATA[bit 18]` is not set. If such a AMX-half-enabled CPUID is passed to `KVM_SET_CPUID2` as it is, guests will crash with general protection fault during boot (See Appendix). This is because Linux kernel (guest) attempts to execute `XSETBV` instruction with all XSAVE feature bits enumerated on CPUID during boot and `XSETBV` only accepts either of both Intel AMX bits enabled or disabled. This bug of `KVM_GET_SUPPORTED_CPUID` returning such a half-enabled state was fixed in [kernel v6.4][bug_fix]. But in any case, Firecracker supports the CPU template feature that enables users to mask CPU features (including Intel AMX), so we enable TILEDATA by default to make it work even on earlier kernels.

To support a dynamically-sized XSTATE buffer, the Linux kernel extended the existing `kvm_xsave` by adding a flexible array member (FAM) in the end. Along with it, `KVM_GET_XSAVE2` API was added and `KVM_SET_XSAVE` API was extended. To support these changes, rust-vmm added (1) `kvm_xsave2` that holds `kvm_xsave` and the length of the FAM, (2) `Xsave` as `FamStructWrapper` of `kvm_xsave2`, (3) `get_xsave2()` for `KVM_GET_XSAVE2` and (4) `set_xsave2()` to take `Xsave` and call `KVM_SET_XSAVE`. Accordingly, use these methods and structs to support Intel AMX in Firecracker.

[kernel_doc_xstate]: https://github.com/torvalds/linux/blob/master/Documentation/arch/x86/xstate.rst
[bug_fix]: https://lore.kernel.org/all/20230405004520.421768-1-seanjc@google.com/

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.~~
  integration tests.

## Appendix: GP fault on guest boot without Intel AMX enablement

```
[    0.048097] general protection fault, maybe for address 0x0: 0000 [#1] PREEMPT SMP NOPTI
[    0.048097] CPU: 0 PID: 0 Comm: swapper/0 Not tainted 6.1.102 #1
[    0.048097] RIP: 0010:fpu__init_cpu_xstate+0x69/0xb0
[    0.048097] Code: 35 2d fd 00 b9 c4 01 00 00 48 89 c2 48 c1 ea 20 0f 30 65 48 89 05 07 36 ff 7e 48 8b 05 48 3d fd 00 31 c9 48 89 c2 48 c1 ea 20 <0f> 01 d1 48 8b 05 5d 7b 2d 01 a8 08 75 0a 48 8b 5d f8 c9 e9 1f 8e
[    0.048097] RSP: 0000:ffffffff82203e68 EFLAGS: 00010246
[    0.048097] RAX: 00000000000202e7 RBX: 0000000000000206 RCX: 0000000000000000
[    0.048097] RDX: 0000000000000000 RSI: 0000000000000000 RDI: 0000000000770eb0
[    0.048097] RBP: ffffffff82203e70 R08: 0000000000000000 R09: 00000000ffffefff
[    0.048097] R10: ffffffff82203d28 R11: ffffffff822b0568 R12: 0000000000000000
[    0.048097] R13: 00000000000202e7 R14: 0000000000000200 R15: 0000000000000000
[    0.048097] FS:  0000000000000000(0000) GS:ffff888007a00000(0000) knlGS:0000000000000000
[    0.048097] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[    0.048097] CR2: ffff888002800000 CR3: 0000000002210001 CR4: 0000000000770eb0
[    0.048097] DR0: 0000000000000000 DR1: 0000000000000000 DR2: 0000000000000000
[    0.048097] DR3: 0000000000000000 DR6: 00000000fffe07f0 DR7: 0000000000000400
[    0.048097] PKRU: 55555554
[    0.048097] Call Trace:
[    0.048097]  <TASK>
[    0.048097]  ? show_trace_log_lvl+0x1c5/0x28b
[    0.048097]  ? show_trace_log_lvl+0x1c5/0x28b
[    0.048097]  ? fpu__init_system_xstate+0x16d/0x3bb
[    0.048097]  ? show_regs.part.0+0x1e/0x24
[    0.048097]  ? die_addr.cold+0x8/0xd
[    0.048097]  ? exc_general_protection+0x1a7/0x3f0
[    0.048097]  ? asm_exc_general_protection+0x27/0x30
[    0.048097]  ? fpu__init_cpu_xstate+0x69/0xb0
[    0.048097]  ? fpu__init_cpu_xstate+0x31/0xb0
[    0.048097]  fpu__init_system_xstate+0x16d/0x3bb
[    0.048097]  fpu__init_system+0x136/0x161
[    0.048097]  arch_cpu_finalize_init+0x22/0x4c
[    0.048097]  start_kernel+0x400/0x499
[    0.048097]  x86_64_start_reservations+0x24/0x2a
[    0.048097]  x86_64_start_kernel+0x72/0x7c
[    0.048097]  secondary_startup_64_no_verify+0xcd/0xdb
[    0.048097]  </TASK>
[    0.048097] ---[ end trace 0000000000000000 ]---
```

With this PR, the GP fault doesn't happen. (Note that a functional test for CPU feature set will be added in an upcoming PR for functional integration tests support.)
```
[    0.048697] x86/fpu: Supporting XSAVE feature 0x001: 'x87 floating point registers'
[    0.048697] x86/fpu: Supporting XSAVE feature 0x002: 'SSE registers'
[    0.048697] x86/fpu: Supporting XSAVE feature 0x004: 'AVX registers'
[    0.048697] x86/fpu: Supporting XSAVE feature 0x020: 'AVX-512 opmask'
[    0.048697] x86/fpu: Supporting XSAVE feature 0x040: 'AVX-512 Hi256'
[    0.048697] x86/fpu: Supporting XSAVE feature 0x080: 'AVX-512 ZMM_Hi256'
[    0.048697] x86/fpu: Supporting XSAVE feature 0x200: 'Protection Keys User registers'
[    0.048697] x86/fpu: Supporting XSAVE feature 0x20000: 'AMX Tile config'
[    0.048697] x86/fpu: Supporting XSAVE feature 0x40000: 'AMX Tile data'
[    0.048697] x86/fpu: xstate_offset[2]:  576, xstate_sizes[2]:  256
[    0.048697] x86/fpu: xstate_offset[5]:  832, xstate_sizes[5]:   64
[    0.048697] x86/fpu: xstate_offset[6]:  896, xstate_sizes[6]:  512
[    0.048697] x86/fpu: xstate_offset[7]: 1408, xstate_sizes[7]: 1024
[    0.048697] x86/fpu: xstate_offset[9]: 2432, xstate_sizes[9]:    8
[    0.048697] x86/fpu: xstate_offset[17]: 2496, xstate_sizes[17]:   64
[    0.048697] x86/fpu: xstate_offset[18]: 2560, xstate_sizes[18]: 8192
...
Ubuntu 24.04.1 LTS ubuntu-fc-uvm ttyS0
...
root@ubuntu-fc-uvm:~#
```


src/vmm/src/vstate/kvm.rs
```diff
@@ -86,6 +98,74 @@ impl Kvm {
...
+    ///
+    /// Intel AMX (Advanced Matrix Extensions) is an instruction set for AI workloads that was
+    /// introduced in Intel Sapphire Rapids (*7i.metal). Since it requires larger area to save the
+    /// state, it is disabled by default.
+    /// https://github.com/torvalds/linux/blob/master/Documentation/arch/x86/xstate.rst
+    ///
+    /// We enable it by default but can be disabled by CPU template; otherwise,
+    /// KVM_GET_SUPPORTED_CPUID returns a inconsistent state where TILECFG is enabled but TILEDATA
+    /// is disabled, causing guest's #GP fault on xsetbv due to the lack of sanity check.
+    /// https://lore.kernel.org/all/20230405004520.421768-1-seanjc@google.com/
```
pb8o, line 116: Will this fix get backported to 6.1? 

Also do I understood it right that this is then only safe to disable via CPU template if we are using a v6.4+ kernel?

zulinx86: Regardless of whether backporting or not, we have to request permission of XTILEDATA.

CPU template can disable it safely on any kernels supporting Intel AMX.

The issue that was fixed in v6.4 is that we cannot directly pass the value returned from KVM_GET_SUPPORTED_CPUID to KVM_SET_CPUID2 without the arch_prctl() call. As long as we call arch_prctl() before KVM_GET_SUPPORTED_CPUID, the issue doesn't affect us.

src/vmm/src/vstate/kvm.rs
```diff
@@ -215,4 +295,99 @@ pub(crate) mod tests {
...
+                            .unwrap();
+                    }
+                }
+                panic!("`vendor_id` not found in /proc/cpuinfo");
+            }
+        }
+
+        #[cfg(target_arch = "x86_64")]
+        #[test]
+        fn test_enable_intel_amx() {
```
Manciukic, line 353: Can we simplify the test by not requiring to assert the cpu model and kernel version?
```
fn test_enable_intel_amx() {
    Kvm::enable_intel_amx().unwrap;

    let mut xfeatures: libc::c_ulong = 0;
    arch_prctl(GET_XCOMP_SUPP, xfeatures)
        .and_then(|_| {
            // If the cpu supports AMX, then assert that it was set
            if xfeatures & INTEL_AMX_XCOMP_MASK != 0 {
                arch_prctl(GET_XCOMP_GUEST_PERM, xfeatures).unwrap()
                assert!(xfeatures & INTEL_AMX_XCOMP_MASK != 0);
            }
            // otherwise the cpu doesn't support AMX and enable_intel_amx is a no-op
        });
     // if the GET_XCOMP_SUPP failed, then the kernel doesn't support setting AMX so 
     // the enable_intel_amx function is a noop and we have nothing to assert
}
```

zulinx86: you're right, we can :)

zulinx86: fixed. thanks!

roypat (commented):
Do we already want to add a changelog entry for this? I know the title says "make perf tests work", but really what we're doing is making firecracker support AMX xP

zulinx86:
> Do we already want to add a changelog entry for this? I know the title says "make perf tests work", but really what we're doing is making firecracker support AMX xP

I'd like to declare official support of Intel AMX when snapshot restore is supported :)

src/vmm/src/vstate/kvm.rs
```diff
@@ -73,6 +83,15 @@ impl Kvm {
 
         #[cfg(target_arch = "x86_64")]
         {
+            // Request permission for Intel AMX (Advanced Matrix Extensions) TILEDATA.
+            //
+            // Unless requested, on kernels prior to v6.4, KVM_GET_SUPPORTED_CPUID returns an
+            // inconsistent state where TILECFG is set but TILEDATA isn't. Such a half-enabled state
```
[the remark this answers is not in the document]
zulinx86: yes, you're right.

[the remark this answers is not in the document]
zulinx86: this part actually does filter out TILECFG if TILEDATA is not permitted: https://elixir.bootlin.com/linux/v6.13.5/source/arch/x86/kvm/x86.h#L412-L419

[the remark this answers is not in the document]
roypat: Thanks!!

src/vmm/src/vstate/kvm.rs
```diff
@@ -215,4 +293,54 @@ pub(crate) mod tests {
             .iter()
             .any(|c| *c == kvm_bindings::KVM_CAP_IOEVENTFD));
     }
+
+    #[cfg(target_arch = "x86_64")]
+    #[test]
+    fn test_request_xstate_feature_permission() {
```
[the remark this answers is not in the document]
zulinx86: yeah, lots duplication :( previously I tested in a different way here ( https://github.com/firecracker-microvm/firecracker/compare/2ae659a5728d8cc00e4530f7fdaa58d3a7650f2e..9ed7e6cd550aa8db605dfa3980311fe547dd0081 ). Based on Riccardo's suggestion, I removed some test utility (kernel version and CPU check).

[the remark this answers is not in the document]
zulinx86: So the main part of the test is that the requested features are actually permitted by calling `ARCH_GET_XCOMP_GUEST_PERM` (note that this is not `ARCH_REQ_XCOMP_GUEST_PERM`).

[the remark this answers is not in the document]
zulinx86: yeah, separating it into `get_supported_xsave_features()` and `request_xsave_feature()` makes sense to me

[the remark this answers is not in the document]
zulinx86: Do you think it's better to move them under vmm/src/arch/x86_64/ ?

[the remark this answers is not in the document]
zulinx86: OK, now it should look much better!

Manciukic (approved):

src/vmm/src/vstate/kvm.rs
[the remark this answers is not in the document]
roypat: Yes, this reads a lot better to me, thanks!! 

I was thinking about this over breakfast, and wondered, shouldn't we also do a `KVM_GET_SUPPORTED_CPUID` in a test and assert that the configuration returned from that "makes sense"? E.g. assert either both bits are 0 or both bits are 1.

> But now I feel like it's just testing the syscall itself 🤔

Yup, that was also my feeling

> Do you think it's better to move them under vmm/src/arch/x86_64/ ?

I think there's some long term ideas by @bchalios and me about splitting kvm.rs into kvm/x86_64 and kvm/aarch, similar to the vm and vcpu modules, and then this could live in `kvm/x86_64.rs`, but that's out of scope here. Maybe one day I'll have an excuse for doing that refactor xP so putting in in `src/arch` is fine by me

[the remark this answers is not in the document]
zulinx86: > I was thinking about this over breakfast, and wondered, shouldn't we also do a KVM_GET_SUPPORTED_CPUID in a test and assert that the configuration returned from that "makes sense"? E.g. assert either both bits are 0 or both bits are 1.

Actually I implemented the test for KVM_GET_SUPPORTED_CPUID but I thought it might be too much as a unit test, then I decided to add CPU feature test inside guest for end-to-end integration test in a next round.

src/vmm/src/vstate/vm/x86_64.rs
```diff
@@ -57,10 +63,32 @@ impl ArchVm {
...
         let msrs_to_save = kvm.msrs_to_save().map_err(ArchVmError::GetMsrsToSave)?;
 
+        // `KVM_CAP_XSAVE2` was introduced to support dynamically-sized XSTATE buffer in kernel
+        // v5.17. `KVM_GET_EXTENSION(KVM_CAP_XSAVE2)` returns the required size in byte if
+        // supported; otherwise returns 0.
+        // https://github.com/torvalds/linux/commit/be50b2065dfa3d88428fdfdc340d154d96bf6848
+        //
+        // Cache the value in order not to call it at each vCPU creation.
+        let xsave2_size = match fd.check_extension_int(Cap::Xsave2) {
+            -1 => {
```
Manciukic, line 73: should this branch catch any negative result?

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -496,7 +503,27 @@ impl KvmVcpu {
...
+        let xsave = match self.xsave2_size {
+            Some(size) => {
+                // SAFETY: The subtraction never underflows because
+                // `KVM_CHECK_EXTENSION(KVM_CAP_XSAVE2)` returns at least 4096 bytes which equals
+                // the size of `kvm_xsave`.
+                // https://docs.kernel.org/virt/kvm/api.html#kvm-get-xsave2
+                //
+                // Note that the FAM entry type is `__u32` although `xsave2_xize` is in bytes.
+                let fam_len = (size - std::mem::size_of::<kvm_xsave>())
+                    .div_ceil(std::mem::size_of::<<kvm_xsave2 as FamStruct>::Entry>());
```
Manciukic, line 515: I'm wondering if we can avoid this "magic" or at least hide it inside the `Xsave` struct. It's so confusing at the moment that you have to compute "how many members of size A are exceeding the size B". I had my head spin for a good 5m on this.

src/vmm/src/vstate/vcpu/x86_64.rs
```diff
@@ -496,7 +503,27 @@ impl KvmVcpu {
...
+                let mut xsave = Xsave::new(fam_len).map_err(KvmVcpuError::Fam)?;
+                // SAFETY: Safe because `xsave` is allocated above with enough size to save XSTATE.
+                unsafe { self.fd.get_xsave2(&mut xsave) }.map_err(KvmVcpuError::VcpuGetXsave2)?;
+                xsave
+            }
+            // `KVM_CAP_XSAVE2` not supported
+            None => Xsave::from(vec![kvm_xsave2 {
+                len: 0,
```
Manciukic, line 523: I think we should explain clearly why this is `0`. It's all so confusing that the `len` is only the extra bytes.

Manciukic (commented):
Overall lgtm, but I think that `fam_len` thing is so confusing right now. If we don't want/can't change the api, I think we should better explain in the comment what is happening because it makes my head spin every time I look at it

tools/bindgen.sh
```diff
@@ -154,6 +154,15 @@ fc-bindgen \
     "amazonlinux-v5.10.y/include/uapi/linux/io_uring.h" \
     >src/vmm/src/io_uring/gen.rs
 
+# Latest upstream kernel
+KERNEL_SRC_DIR="linux"
+[ -d ${KERNEL_SRC_DIR} ] || git clone --depth 1 https://github.com/amazonlinux/linux ${KERNEL_SRC_DIR}
```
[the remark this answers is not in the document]
zulinx86: yeah, I agree. let's do it separately.

src/vmm/src/vstate/vm/x86_64.rs
zulinx86: ioctl() should not return any other negative value, but that would be more robust! will fix.

src/vmm/src/vstate/vcpu/x86_64.rs
zulinx86: indeed. thanks.

src/vmm/src/vstate/vcpu/x86_64.rs
zulinx86: `Xsave` itself is defined in rust-vmm. So, to hide it inside `Xsave` struct, we have to make another release of kvm-bindings. I can put the calculate into another function in `xstate.rs`.

zulinx86: I put them into another method of `KvmVcpu` with more info about the structs.

src/vmm/src/vstate/vm/x86_64.rs
```diff
@@ -57,10 +63,34 @@ impl ArchVm {
...
+        // `KVM_CAP_XSAVE2` was introduced to support dynamically-sized XSTATE buffer in kernel
+        // v5.17. `KVM_GET_EXTENSION(KVM_CAP_XSAVE2)` returns the required size in byte if
+        // supported; otherwise returns 0.
+        // https://github.com/torvalds/linux/commit/be50b2065dfa3d88428fdfdc340d154d96bf6848
+        //
+        // Cache the value in order not to call it at each vCPU creation.
+        let xsave2_size = match fd.check_extension_int(Cap::Xsave2) {
+            // Catch all negative values just in case although the possible negative retuned value
+            // of ioctl() is only -1.
+            i32::MIN..=-1 => {
```
roypat, line 75: nit: I think this can just be
```suggestion
            ..=-1 => {
```

src/vmm/src/vstate/vm/x86_64.rs
```diff
@@ -57,10 +63,34 @@ impl ArchVm {
...
         let msrs_to_save = kvm.msrs_to_save().map_err(ArchVmError::GetMsrsToSave)?;
 
+        // `KVM_CAP_XSAVE2` was introduced to support dynamically-sized XSTATE buffer in kernel
+        // v5.17. `KVM_GET_EXTENSION(KVM_CAP_XSAVE2)` returns the required size in byte if
+        // supported; otherwise returns 0.
+        // https://github.com/torvalds/linux/commit/be50b2065dfa3d88428fdfdc340d154d96bf6848
+        //
+        // Cache the value in order not to call it at each vCPU creation.
+        let xsave2_size = match fd.check_extension_int(Cap::Xsave2) {
+            // Catch all negative values just in case although the possible negative retuned value
```
roypat, line 73:
```suggestion
            // Catch all negative values just in case although the possible negative return value
```

roypat (dismissed):

src/vmm/src/vstate/vm/x86_64.rs
zulinx86: fixed!

roypat (approved):

PR #5069, opened 2025-03-06
Kani tweaks

ShadowCurse:
## Changes
- Add option to skip kani step in PR pipeline
- Increase timeout for kani tests

## Reason
There are some cases where kani step is not needed, so add an option to skip it.
Also kani runtime was getting very close to the timeout, so increase it.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tests/integration_tests/test_kani.py
+++ b/tests/integration_tests/test_kani.py
@@ -13,10 +13,12 @@ from framework import utils
 
 PLATFORM = platform.machine()
 
+TIMEOUT = 3600
+
 
 # The `check_output` timeout will always fire before this one, but we need to
 # set a timeout here to override the default pytest timeout of 180s.
-@pytest.mark.timeout(2420)
+@pytest.mark.timeout(TIMEOUT)
 @pytest.mark.skipif(
     os.environ.get("BUILDKITE") != "true",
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline_pr.py
```diff
@@ -48,8 +48,10 @@
         "./tools/devtool -y make_release",
     )
 
-if not changed_files or any(
-    x.suffix in [".rs", ".toml", ".lock"] for x in changed_files
+if (
+    not pipeline.args.no_kani
+    and not changed_files
+    or any(x.suffix in [".rs", ".toml", ".lock"] for x in changed_files)
```
roypat, line 54: This needs to be
```suggestion
if (
    not pipeline.args.no_kani
    and (not changed_files
    or any(x.suffix in [".rs", ".toml", ".lock"] for x in changed_files))
```

as `and` takes precedence over `or`. As you've written it, the --no-kani flag will be ignored if a .rs/.toml/.lock file is touched.

ShadowCurse: good catch, fixed

roypat (approved):

zulinx86 (approved):

PR #5050, opened 2025-02-25
Tiny integration test fixes to fill accidental gaps in coverage

roypat:
Fixes a handful of small issues discovered in integration tests, such as
- specific tests that were supposed to cover all io engines only testing the async engine
- tests not asserting error messages even though that was clearly the intention
- test duplicating functionality for which we have helper functions

## License Acceptance

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license. For more information on following Developer
Certificate of Origin and signing off your commits, please check
[`CONTRIBUTING.md`][3].

## PR Checklist

- [ ] I have read and understand [CONTRIBUTING.md][3].
- [ ] I have run `tools/devtool checkstyle` to verify that the PR passes the
  automated style checks.
- [ ] I have described what is done in these changes, why they are needed, and
  how they are solving the problem in a clear and encompassing way.
- [ ] I have updated any relevant documentation (both in code and in the docs)
  in the PR.
- [ ] I have mentioned all user-facing changes in `CHANGELOG.md`.
- [ ] If a specific issue led to this PR, this PR closes the issue.
- [ ] When making API changes, I have followed the
  [Runbook for Firecracker API changes][2].
- [ ] I have tested all new and changed functionalities in unit tests and/or
  integration tests.
- [ ] I have linked an issue to every new `TODO`.

______________________________________________________________________

- [ ] This functionality cannot be added in [`rust-vmm`][1].

[1]: https://github.com/rust-vmm
[2]: https://github.com/firecracker-microvm/firecracker/blob/main/docs/api-change-runbook.md
[3]: https://github.com/firecracker-microvm/firecracker/blob/main/CONTRIBUTING.md

PR #5074, opened 2025-03-10
test: Modernize test_send_ctrl_alt_del

roypat:
Instead of sleeping for 2 seconds to wait for booting, add a network device and wait for ssh availability (happens automatically in .start()). Use mark_killed() to assert that Firecracker exited instead of "waitpid". Using `waitpid(2)` here is wrong, because it only works on child processes, and Firecracker is no child of the pytest process (it is daemonized, for starters). The reason we never caught this being broken is because the test also simply ignores all exceptions raised.

But there's more issues here: The test never caused _killed to be set to `True`, so if Firecracker really did exit, then we would have been seeing intermittent failures during microvm teardown (as Microvm.kill() asserts that the process is still alive if _killed is False). In fact, our guest kernels do not have the required drivers compiled in to support CTRL+ALT+DEL (thanks Riccardo for figuring this part out).

## Changes

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ and this project adheres to
 
 ### Fixed
 
+- #\[[5074](https://github.com/firecracker-microvm/firecracker/pull/5074)\] Fix
+  the `SendCtrlAltDel` command not working for ACPI-enabled guest kernels, by
+  dropping the i8042.nopnp argument from the default kernel command line
+  Firecracker constructs.
+
 ## [1.11.0]
 
 ### Added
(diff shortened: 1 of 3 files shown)


roypat:
Split out from #5050 so that the other fixes in there can get merged

Manciukic:
tests are failing because the kernel is not setting up the virtual keyboard used to deliver the ctrl+alt+del. The default command line disables ACPI PnP (`i8042.nopnp`), so we should override it in this test.
```
diff --git a/tests/integration_tests/functional/test_api.py b/tests/integration_tests/functional/test_api.py
index eb2cd1b60..0f68f6719 100644
--- a/tests/integration_tests/functional/test_api.py
+++ b/tests/integration_tests/functional/test_api.py
@@ -766,7 +766,8 @@ def test_send_ctrl_alt_del(uvm_plain_any):
     test_microvm = uvm_plain_any
     test_microvm.spawn()
 
-    test_microvm.basic_config()
+    # The default command line disables ACPI PnP (i8042.nopnp), so override it
+    test_microvm.basic_config(boot_args="reboot=k panic=1 pci=off")
     test_microvm.add_net_iface()
     test_microvm.start()
 ```

Manciukic (approved):

zulinx86 (approved):

PR #5097, opened 2025-03-19
Try to make UFFD handlers more robust

bchalios:
## Changes

Add logic to our UFFD handlers to retry the negotiation with Firecracker up to 5 times before giving up. This helps making them (slightly) more robust. Also, we add some logging in the receive logic so that we can inspect failures post-mortem.

## Reason

According to our UFFD protocol, UFFD handlers negotiate with Firecracker during initialization and wait for it to send over a UDS the UFFD file descriptor along with the memory mappings that are being handled over the UFFD.

During this handshake, our (testing only/not production grade) UFFD handlers issue what essentially is a `recvmsg` that should return with the UFFD fd and the mappings. Some times instead of the file descriptor, the `recvmsg` wrapper returns a `None` value for the file descriptor. When this happens, the UFFD handler crashes and Firecracker process hangs.

According to `man recv(2)`:

```
Datagram sockets in various domains (e.g., the UNIX and Internet
domains) permit zero-length datagrams.  When such a datagram is
received, the return value is 0.
```

which means it is possible to receive a zero-length message (we are communicating with Firecracker over a UDS).

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/src/firecracker/examples/uffd/uffd_utils.rs
+++ b/src/firecracker/examples/uffd/uffd_utils.rs
@@ -52,21 +52,45 @@ pub struct UffdHandler {
 }
 
 impl UffdHandler {
+    fn get_mappings_and_file(stream: &UnixStream) -> (String, File) {
+        // Sometimes, reading from the stream succeeds but we don't receive any
+        // UFFD descriptor. We don't really have a good understanding why this is
+        // happening, but let's try to be a bit more robust and retry a few times
+        // before we declare defeat.
+        for _ in 1..=5 {
+            let mut message_buf = vec![0u8; 1024];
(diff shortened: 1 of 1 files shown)


ShadowCurse (dismissed):

roypat (approved):
There's no chance that the body and the fds simply arrive in two distinct packets?

bchalios:
> There's no chance that the body and the fds simply arrive in two distinct packets?

I would assume that `recvmsg` would return us a complete message regardless of how packets have been arriving. In any case, if something like this happens we would now see it in the logs.

ShadowCurse (approved):

PR #5101, opened 2025-03-20
[1.10 Backport] Revert "fix: test_sec_audit: ignore cargo audit output"

roypat:
This reverts commit https://github.com/firecracker-microvm/firecracker/commit/d761b013db1caeec046b187927133b7fdac1375d. The output
to stdout was fixed in cargo audit 0.21.2, so if we rebuild the docker
container the grep is no longer necessary.

In fact, the grep has broken this test in our nightly pipeline because
it overwrites the return code of cargo audit itself, meaning the non-PR
version of this test (which is supposed to fail if there exist any cargo
audit warnings) was never failing.

(cherry picked from commit https://github.com/firecracker-microvm/firecracker/commit/1d98a21e087aa4f188f2d6e665b54c6962569199)

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tests/integration_tests/security/test_sec_audit.py
+++ b/tests/integration_tests/security/test_sec_audit.py
@@ -35,6 +35,6 @@ def test_cargo_audit():
         )
 
     git_ab_test_host_command_if_pr(
-        "cargo audit --deny warnings -q --json |grep -Po '{.*}'",
+        "bash -o pipefail -c \"cargo audit --deny warnings -q --json |grep -Po '{.*}'\"",
         comparator=set_did_not_grow_comparator(set_of_vulnerabilities),
     )


Manciukic:
This is a lot of changes for a one line fix. Is there a way we can fix the audit without bringing in alll these other unrelated changes?

roypat:
> This is a lot of changes for a one line fix. Is there a way we can fix the audit without bringing in alll these other unrelated changes?

I tried building a sort of `v75.1` of the docker container specifically for the 1.10 branch that only pulls in the cargo audit update, but that didn't easily work because we don't pin versions in the docker container, and a bunch of our `cargo install`ed tools dont compile on the toolchain specified in the container anymore. And writing even more commits on top of that specifically for the release branch seemed like more tech debt than a series of large, but cleanly applying cherry-picks from main.

Manciukic:
> But this won't work in the scenario where we want to run the grep, because it'll just abort without filtering out the unwanted part of the output in case of failures, no?

Isn't that the same behavior of running without the grep on the fixed `cargo-audit`?

roypat:
> > But this won't work in the scenario where we want to run the grep, because it'll just abort without filtering out the unwanted part of the output in case of failures, no?
> 
> Isn't that the same behavior of running without the grep on the fixed `cargo-audit`?

When we run this as an A/B-Test, in python we ignore the return code and only parse the json it outputs to stderr. But for that to work, all non-json parts actually need to be stripped from stderr, e.g. we need to run the grep part even if cargo audit fails.

Now yes, there's way we can hack around all of this by running different commands for A/B and non-A/B, but I admittedly don't see the problem with backporting a handful (or two) of commits :sweat_smile:

Manciukic:
Oh, I see what you mean now. Bash will not abort `grep`, it will wait for the entire pipeline to complete.

       The return status of a pipeline is the exit status of the last command,
       unless  the  pipefail  option  is enabled.  If pipefail is enabled, the
       pipeline's return status is the value of the last  (rightmost)  command
       to  exit  with a non-zero status, or zero if all commands exit success‐
       fully.  If the reserved word !  precedes a pipeline, the exit status of
       that  pipeline  is the logical negation of the exit status as described
       above.  The shell waits for all commands in the pipeline  to  terminate
       before returning a value.

roypat:
> Oh, I see what you mean now. Bash will not abort `grep`, it will wait for the entire pipeline to complete.
> 
> ```
>    The return status of a pipeline is the exit status of the last command,
>    unless  the  pipefail  option  is enabled.  If pipefail is enabled, the
>    pipeline's return status is the value of the last  (rightmost)  command
>    to  exit  with a non-zero status, or zero if all commands exit success‐
>    fully.  If the reserved word !  precedes a pipeline, the exit status of
>    that  pipeline  is the logical negation of the exit status as described
>    above.  The shell waits for all commands in the pipeline  to  terminate
>    before returning a value.
> ```

oh! neat, didn't now that. I thought it would just abort before the grep without running it if cargo audit fails.

>      We're currently standing at 13 commits and 93 files changed unless you pushed the wrong thing.

I still don't see any problem :joy: 

but let me update to your proposed on-line fix. Kinda wished we'd done it that way in main as well now, so that all the branches would be in sync, but oh well :(

Manciukic:
> but let me update to your proposed on-line fix. Kinda wished we'd done it that way in main as well now, so that all the branches would be in sync, but oh well :(

I wouldn't worry too much about that. We'll likely never touch this part of the code again and in a few months we'll forget this branch even existed.

Manciukic (approved):

PR #5104, opened 2025-03-21
doc: clarify kernel configuration usage in our repo

### LINKED ISSUE #4881: [Bug] Latest 6.1 guest kernel config results in kernel panic when booting VM
# Describe the bug

The **x86_64 6.1 guest kernel config** as it is currently on master after the https://github.com/firecracker-microvm/firecracker/commit/9157a0cf7188706cef3bbe01316171964f569a76 commit, runs seemingly okay with the latest 6.1 kernel (6.1.114), but produces an unusually larger vmlinux (38MB instead of 29MB), and when booting a VM with it the following happens:

```
[   12.489510] /dev/root: Can't open blockdev
[   12.489784] VFS: Cannot open root device "vda" or unknown-block(0,0): error -6
[   12.490205] Please append a correct "root=" boot option; here are the available partitions:
[   12.490717] Kernel panic - not syncing: VFS: Unable to mount root fs on unknown-block(0,0)
```

## To Reproduce

1. Download 6.1.114 Linux kernel source via tarball from [kernel.org](https://cdn.kernel.org/pub/linux/kernel/v6.x/linux-6.1.114.tar.xz)
2. Enter the extracted kernel directory
3. Copy the full contents (much larger after that commit) of https://github.com/firecracker-microvm/firecracker/blob/main/resources/guest_configs/microvm-kernel-ci-x86_64-6.1.config into `.config`
4. Run `make -j N vmlinux` in the dir with N being the number of cores, I personally used 12 as I hit this bug but I doubt this matters
5. Copy the produced `vmlinux` anywhere used to start Firecracker VMs
6. Start a VM configured with: **one ext4 block device that is a root device, sync io engine, and no initramfs is used** (and this kernel, of course)
7. Receive the aforementioned error when booting the VM

## Expected behaviour

Everything would build correctly and the VM would start.

## Environment

- Firecracker version: 1.9.1, tried 1.7.0 with the same thing happening
- Host and guest kernel versions: 6.1.99 host and 6.1.114 guest
- Rootfs used: trimmed down Debian ext4 with systemd, created with buildfs as per [my docs about it](https://github.com/kanpov/buildfs)
- Architecture: x86_64
- Any other relevant software versions:

## Additional context

Even though the output is the same as https://github.com/firecracker-microvm/firecracker/issues/4816, I'm quite sure this isn't the same issue as I'm not using `noapic` as a kernel boot arg (mine are `console=ttyS0 reboot=k panic=1`) and adding/removing it doesn't change anything.

**HOWEVER!** the way I pinned this issue down to that specific commit is that if you take the commit right before that problematic one and open the relevant guest kernel config in it: https://github.com/firecracker-microvm/firecracker/blob/86a2559b26a4b9a05405aeaa58bab0f7261d71bc/resources/guest_configs/microvm-kernel-ci-x86_64-6.1.config

And do the same steps with that config, everything works perfectly and a 29MB working vmlinux is produced.

## Checks

bchalios:
## Changes

Clarify in the documentation that the kernel configurations we keep in our repo are meant to be used with kernels provided by Amazon Linux and that we don't guarantee that they can be used to build kernels from upstream Linux.

## Reason

We keep in the repo the kernel configurations from which we build the kernel images we use in our CI pipelines. However, we don't mention what kernel sources we use to build these kernel images.

Fixes #4881

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/docs/kernel-policy.md
+++ b/docs/kernel-policy.md
@@ -37,6 +37,19 @@ The guest kernel configs used in our validation pipelines can be found
 [here](../resources/guest_configs/) while a breakdown of the relevant guest
 kernel modules can be found in the next section.
 
+We use these configurations to build microVM-specific kernels vended by Amazon
+Linux. microVM kernel source code is published in the Amazon Linux
+[linux repo](https://github.com/amazonlinux/linux) under tags in the form of
+`microvm-kernel-*`, e.g. 6.1.128-3.201.amazn2023 kernel can be found
+[here](https://github.com/amazonlinux/linux/tree/microvm-kernel-6.1.128-3.201.amzn2023).
+These kernels may have diverged from the equivalent mainline versions, as we
+often backport patches that we require for supporting Firecracker features not
+present in the kernel versions we officially support. As a result, kernel
+configurations found in this repo should be used to build exclusively the
+aforementioned Amazon Linux kernels. We do not guarantee that using these
+configurations to build upstream kernels, will work or produce usable kernel
+images.
+
 ## Guest kernel configuration items
 
 The configuration items that may be relevant for Firecracker are:


docs/kernel-policy.md
```diff
@@ -21,22 +21,33 @@ and removed in a following release, after its minimum end of support date.
 
 ### Host Kernel
 
-| Host kernel | Min. version | Min. end of support |
-| ----------: | -----------: | ------------------: |
-|       v5.10 |       v1.0.0 |          2024-01-31 |
-|        v6.1 |       v1.5.0 |          2025-10-12 |
+| Host kernel | Min. version | Min. end of support | | ----------: |
+-----------: | ------------------: | | v5.10 | v1.0.0 | 2024-01-31 | | v6.1 |
+v1.5.0 | 2025-10-12 |
```
Manciukic, line 26: I know some mdformatter prefers this format but it makes it difficult to read as text.

ShadowCurse, line 26: These are looking broken. Seems you missed `\n`.

Manciukic (commented):
the new paragraph lgtm, but I'd fix those tables above.

docs/kernel-policy.md
bchalios: sorry about that, that's my mdformatter changing things without me asking it to. Will fix

bchalios: autoformatter kicked in. Should be fixed now

Manciukic (dismissed):

docs/kernel-policy.md
```diff
@@ -37,6 +37,19 @@ The guest kernel configs used in our validation pipelines can be found
 [here](../resources/guest_configs/) while a breakdown of the relevant guest
 kernel modules can be found in the next section.
 
+We use these configurations to build microVM-specific kernels vended by Amazon
+Linux. microVM kernels source code are published in Amazon Linux
```
roypat, line 41:
```suggestion
Linux. microVM kernel source code is published in the Amazon Linux
```

roypat (approved):

PR #5108, opened 2025-03-24
Add option to force guests to use swiotlb in dedicated memory region for all I/O

roypat:
## Changes

Add a new API field to /machine-config, `initial_swiotlb_size`, which causes Firecracker to carve out a region of guest memory of the specified size, mark is as dma-restricted in FDT, and force all virtio devices to use this memory for their vrings and buffers.

## Reason

With secret hiding, I/O needs to be handled by swiotlb, as we cannot do I/O into guest memory that is direct map removed

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.


tests/integration_tests/functional/test_swiotlb.py
```diff
@@ -0,0 +1,65 @@
...
+        component, start, end = match.group(1, 2, 3)
+
+        assert (
+            start >= swiotlb_start and end <= swiotlb_end
+        ), f"Guest placed virtio queue component {component} outside of swiotlb region!"
+
+    assert found_any, "Did not find any virtio devices in Firecracker logs"
+
+
+def test_swiotlb_snapshot(microvm_factory, guest_kernel_linux_6_1, rootfs):
```
kalyazin, line 52: nit: can this be done via the uvm_any fixture somehow?

src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -97,6 +111,10 @@ pub struct MachineConfig {
     pub vcpu_count: u8,
     /// The memory size in MiB.
     pub mem_size_mib: usize,
+    /// Additional configuration options for guest memory
+    #[serde(default)]
+    #[serde(skip_serializing_if = "is_default")]
+    pub mem_config: MemoryConfig,
```
kalyazin, line 117: would it be worth updating the swagger file, lest we forget later?

src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -178,6 +201,9 @@ pub struct MachineConfigUpdate {
     /// The memory size in MiB.
     #[serde(default)]
     pub mem_size_mib: Option<usize>,
+    /// The memory configuration
+    #[serde(default)]
+    pub mem_config: Option<MemoryConfig>,
```
kalyazin, line 206: is it possible for us to update it post boot?

src/vmm/src/resources.rs
```diff
@@ -473,6 +489,47 @@ impl VmResources {
...
+    }
+
+    /// Allocates the dedicated I/O region for swiotlb use, if one was requested.
+    pub fn allocate_swiotlb_region(&self) -> Result<Option<GuestRegionMmap>, MemoryError> {
+        if !self.swiotlb_used() {
+            return Ok(None);
+        }
+
+        let start = self.memory_size();
+        let start = start.max(crate::arch::bytes_before_last_gap());
```
kalyazin, line 524: I'm not sure I fully follow this logic. We wouldn't like to place the swiotlb region past the point where the gap starts, but how do we makes sure we place it after it ends?

src/vmm/src/arch/aarch64/fdt.rs
```diff
@@ -233,6 +246,19 @@ fn create_memory_node(fdt: &mut FdtWriter, guest_mem: &GuestMemoryMmap) -> Resul
     fdt.property_array_u64("reg", mem_reg_prop)?;
     fdt.end_node(mem)?;
 
+    if let Some(region) = swiotlb_region {
```
kalyazin, line 249: can link relevant documentation

src/vmm/src/devices/virtio/queue.rs
```diff
@@ -308,6 +308,15 @@ impl Queue {
         Ok(slice.ptr_guard_mut().as_ptr())
     }
 
+    fn log_queue_component_address(component: &'static str, addr: GuestAddress, size: usize) {
+        log::info!(
+            "Placed virt queue {} at [{}, {})",
```
kalyazin, line 313: nit: I thought we call the combination of desc table and avail/used rings a virt queue. A desc table on its own doesn't look like a queue to me.

tests/integration_tests/functional/test_swiotlb.py
```diff
@@ -0,0 +1,65 @@
...
+    assert match is not None, "Firecracker did not print swiotlb region placement"
+
+    swiotlb_start, swiotlb_end = match.group(1, 2)
+
+    found_any = False
+
+    for match in re.finditer(
+        r"Placed virt queue ([a-zA-Z ]+) at \[(\d+), (\d+)\)", vm.log_data
+    ):
+        found_any = True
```
kalyazin, line 42: do we want "any" or "all"?

src/vmm/src/vmm_config/machine_config.rs
```diff
@@ -89,6 +92,17 @@ impl From<HugePageConfig> for Option<memfd::HugetlbSize> {
...
 
+/// Structure containing options for tweaking guest memory configuration.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Deserialize, Serialize)]
+pub struct MemoryConfig {
+    /// The initial size of the swiotlb region. If 0, no swiotlb region will be created.
+    /// If non-zero, all device will be forced to bounce buffers through a swiotlb region
+    /// of the specified size that will have been placed into a dedicated kvm memslot.
+    #[cfg(target_arch = "aarch64")]
+    #[serde(default)]
+    pub initial_swiotlb_size: usize,
```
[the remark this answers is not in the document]
roypat: The only lower bound I can think of is "all vqueues fit into swiotlb", but since a vqueue fits into a single page, and this only allows specifying swiotlb at mb granularity, that's trivially met because even the smallest possible swiotlb region can contain more queues than we have interrupt lines available.

src/vmm/src/vmm_config/machine_config.rs
roypat: no

tests/integration_tests/functional/test_swiotlb.py
roypat: any. This is just to check the loop isn't a no-op

tests/conftest.py
```diff
@@ -376,6 +376,16 @@ def io_engine(request):
     return request.param
 
 
+@pytest.fixture(
+    params=[None, 16, 32, 64] if platform.machine() == "aarch64" else [None]
```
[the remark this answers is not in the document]
roypat: Not that I'm aware :/ Probably best we can do is figure out the point at which performance starts to suffer. These sizes are just some random guesses on my part.

src/vmm/src/vmm_config/machine_config.rs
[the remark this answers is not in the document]
kalyazin: are we absolutely sure that sub-mb swiotlb isn't practically useful?

tests/integration_tests/functional/test_swiotlb.py
kalyazin: what I mean is wouldn't it be useful to check that all devices we attach allocate queues in the swiotlb?

src/vmm/src/devices/virtio/queue.rs
kalyazin: ok, got it now, thx.

tests/conftest.py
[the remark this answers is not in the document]
kalyazin: could we also add, say, a 1M to verify this thinking?

tests/integration_tests/functional/test_swiotlb.py
roypat: yeah, that's what the assert in the loop does. This boolean is only to verify that firecracker actually printed the log message we're expecting.

kalyazin: ah, see it now, thx.

tests/framework/microvm.py
```diff
@@ -747,15 +750,23 @@ def basic_config(
 
         Reference: file:../../src/vmm/src/vmm_config/boot_source.rs::DEFAULT_KERNEL_CMDLINE
         """
+        # Have to do it this way as otherwise A/B-tests fail if the 'A' revision
+        # of Firecracker doesn't knwo about the mem_config parameter.
```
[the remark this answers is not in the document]
roypat: fixed, thanks!

tests/conftest.py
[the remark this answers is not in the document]
roypat: yup, 1MiB works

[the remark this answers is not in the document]
kalyazin: Thanks. What I meant was if we'd like to empirically find the point where the swiotlb is sufficiently low to reduce the performance, we can add 1M in the perf tests as well to confirm it is indeed the case.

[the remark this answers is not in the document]
roypat: Ah, yeah, let me do that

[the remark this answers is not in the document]
roypat: So our performance tests didn't show a point at which performance started getting worse - just that at some point (2MiB for vsock, 1MiB for net; block always works) the test just doesn't pass anymore, presumably because the swiotlb area is too small. Performance between the "working" versions seems fairly consistent though, except 64MiB sometimes being worse, probably due to the tests using 128MB VMs, and a 64MiB swiotlb region not leaving much space for the guest kernel and userspace.

kalyazin (approved):

JackThomson2 (approved):

PR #5125, opened 2025-03-31
Refactor: Move `GuestMemoryMmap` into `struct Vm`

roypat:
## Changes

Move the `GuestMemoryMmap` that keeps track of a Firecracker VM's memory regions into `struct Vm`, from `struct Vmm`, and then update some related functionality (such as various snapshot helper functions) to operate on `struct Vm` instead of `struct Vmm`.

## Reason

So far, Firecracker always dealt with a single collection of homogeneous guest memory regions. However, #5108 introduces a second type of memory regions, which makes the set of memory regions heterogeneous, and other features, such as virtio-pmem support, will come with similar "special" memory regions. By moving the guest memory management into `Vm`, we encapsule this heterogeneity and allow incremental building of guest memory configuration.

Admittedly, there's nothing _right now_ that would benefit from this refactor (it doesn't even reduce LoC :cry:), but it will most certainly make life on the feature branch easier to not have to continuously rebase this :sweat_smile: 

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/src/vmm/src/arch/aarch64/mod.rs
+++ b/src/vmm/src/arch/aarch64/mod.rs
@@ -89,7 +89,7 @@ pub fn configure_system_for_boot(
     // Configure vCPUs with normalizing and setting the generated CPU configuration.
     for vcpu in vcpus.iter_mut() {
         vcpu.kvm_vcpu.configure(
-            &vmm.guest_memory,
+            vmm.vm.guest_memory(),
             entry_point,
             &vcpu_config,
             &optional_capabilities,
@@ -105,7 +105,7 @@ pub fn configure_system_for_boot(
         .expect("Cannot create cstring from cmdline string");
 
     let fdt = fdt::create_fdt(
-        &vmm.guest_memory,
+        vmm.vm.guest_memory(),
         vcpu_mpidr,
         cmdline,
         vmm.mmio_device_manager.get_device_info(),
@@ -114,8 +114,10 @@ pub fn configure_system_for_boot(
         initrd,
     )?;
 
-    let fdt_address = GuestAddress(get_fdt_addr(&vmm.guest_memory));
-    vmm.guest_memory.write_slice(fdt.as_slice(), fdt_address)?;
+    let fdt_address = GuestAddress(get_fdt_addr(vmm.vm.guest_memory()));
+    vmm.vm
+        .guest_memory()
+        .write_slice(fdt.as_slice(), fdt_address)?;
 
     Ok(())
 }
--- a/src/vmm/src/arch/aarch64/mod.rs
+++ b/src/vmm/src/arch/aarch64/mod.rs
@@ -32,7 +32,7 @@ use crate::utils::{align_up, usize_to_u64};
 use crate::vmm_config::machine_config::MachineConfig;
 use crate::vstate::memory::{Address, Bytes, GuestAddress, GuestMemory, GuestMemoryMmap};
 use crate::vstate::vcpu::KvmVcpuError;
-use crate::{Vcpu, VcpuConfig, Vmm};
+use crate::{Vcpu, VcpuConfig, Vmm, logger};
 
 /// Errors thrown while configuring aarch64 system.
 #[derive(Debug, thiserror::Error, displaydoc::Display)]
@@ -58,9 +58,35 @@ pub const MMIO_MEM_SIZE: u64 = layout::DRAM_MEM_START - layout::MAPPED_IO_START;
 
 /// Returns a Vec of the valid memory addresses for aarch64.
 /// See [`layout`](layout) module for a drawing of the specific memory model for this platform.
-pub fn arch_memory_regions(size: usize) -> Vec<(GuestAddress, usize)> {
-    let dram_size = min(size, layout::DRAM_MEM_MAX_SIZE);
-    vec![(GuestAddress(layout::DRAM_MEM_START), dram_size)]
+///
+/// The `offset` parameter specified the offset from [`layout::DRAM_MEM_START`].
+pub fn arch_memory_regions(offset: usize, size: usize) -> Vec<(GuestAddress, usize)> {
+    assert!(size > 0, "Attempt to allocate guest memory of length 0");
+    assert!(
+        offset.checked_add(size).is_some(),
+        "Attempt to allocate guest memory such that the address space would wrap around"
+    );
+    assert!(
+        offset < layout::DRAM_MEM_MAX_SIZE,
+        "offset outside allowed DRAM range"
+    );
+
+    let dram_size = min(size, layout::DRAM_MEM_MAX_SIZE - offset);
+
+    if dram_size != size {
+        logger::warn!(
+            "Requested offset/memory size {}/{} exceeds architectural maximum (1022GiB). Size has \
+             been truncated to {}",
+            offset,
+            size,
+            dram_size
+        );
+    }
+
+    vec![(
+        GuestAddress(layout::DRAM_MEM_START + offset as u64),
+        dram_size,
+    )]
 }
 
 /// Configures the system for booting Linux.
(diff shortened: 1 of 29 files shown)


src/vmm/src/arch/x86_64/mod.rs
```diff
@@ -110,17 +110,33 @@ pub const MMIO_MEM_SIZE: u64 = MEM_32BIT_GAP_SIZE;
...
+    assert!(size > 0, "Attempt to allocate guest memory of length 0");
+    assert!(
+        offset.checked_add(size).is_some(),
+        "Attempt to allocate guest memory such that the address space would wrap around"
+    );
+
     // It's safe to cast MMIO_MEM_START to usize because it fits in a u32 variable
     // (It points to an address in the 32 bit space).
-    match size.checked_sub(usize::try_from(MMIO_MEM_START).unwrap()) {
+    match (size + offset).checked_sub(u64_to_usize(MMIO_MEM_START)) {
```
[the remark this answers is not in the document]
roypat: If you look like 3 lines above this, there's an assert that ensures the overflow cannot happen. Just doing checked_add here is wrong because then we'll incorrectly enter the `None` branch and construct wrong memory regions.

kalyazin (dismissed):

JackThomson2 (dismissed):

src/vmm/src/arch/aarch64/mod.rs
```diff
@@ -58,9 +58,12 @@ pub const MMIO_MEM_SIZE: u64 = layout::DRAM_MEM_START - layout::MAPPED_IO_START;
 
 /// Returns a Vec of the valid memory addresses for aarch64.
 /// See [`layout`](layout) module for a drawing of the specific memory model for this platform.
-pub fn arch_memory_regions(size: usize) -> Vec<(GuestAddress, usize)> {
-    let dram_size = min(size, layout::DRAM_MEM_MAX_SIZE);
-    vec![(GuestAddress(layout::DRAM_MEM_START), dram_size)]
+pub fn arch_memory_regions(offset: usize, size: usize) -> Vec<(GuestAddress, usize)> {
+    let dram_size = min(size, layout::DRAM_MEM_MAX_SIZE - offset);
```
bchalios, line 62: I hit this as well when writing the pmem PoC. It's mostly a nit, but what should the semantics be if someone passes (`offset`, `size`) that ends up being past `DRAM_MEM_MAX_SIZE`? Should we actually error out (return an error or `panic`) or at the very least print a warning?

src/vmm/src/arch/x86_64/mod.rs
```diff
@@ -110,17 +110,33 @@ pub const MMIO_MEM_SIZE: u64 = MEM_32BIT_GAP_SIZE;
...
 /// carve out at the end of 32bit address space.
-pub fn arch_memory_regions(size: usize) -> Vec<(GuestAddress, usize)> {
+pub fn arch_memory_regions(offset: usize, size: usize) -> Vec<(GuestAddress, usize)> {
+    // If we get here with size == 0 something has seriously gone wrong. Firecracker should never
+    // try to allocate guest memory of size 0
+    assert!(size > 0, "Attempt to allocate guest memory of length 0");
+    assert!(
+        offset.checked_add(size).is_some(),
+        "Attempt to allocate guest memory such that the address space would wrap around"
+    );
```
bchalios, line 120: Look at previous comment.

src/vmm/src/arch/x86_64/mod.rs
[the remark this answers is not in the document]
bchalios: why are you speaking to machines? xD

src/vmm/src/arch/x86_64/mod.rs
```diff
@@ -456,6 +472,56 @@ pub fn load_kernel(
...
+        let offset: u64 = kani::any::<u64>();
+        let len: u64 = kani::any::<u64>();
+
+        kani::assume(len > 0);
+        kani::assume(offset.checked_add(len).is_some());
+
+        let regions = arch_memory_regions(offset as usize, len as usize);
+
+        // There's only one MMIO gap, so we can get either 1 or 2 regions
+        assert!(regions.len() <= 2);
```
bchalios, line 492: mainly a note to myself: PCIe might change this

src/vmm/src/persist.rs
```diff
@@ -479,19 +476,14 @@ fn guest_memory_from_uffd(
...
         create_guest_memory(mem_state, track_dirty_pages, huge_pages)?;
 
     let mut uffd_builder = UffdBuilder::new();
 
-    if enable_balloon {
-        // We enable this so that the page fault handler can add logic
-        // for treating madvise(MADV_DONTNEED) events triggerd by balloon inflation.
-        uffd_builder.require_features(FeatureFlags::EVENT_REMOVE);
-    }
+    uffd_builder.require_features(FeatureFlags::EVENT_REMOVE);
```
bchalios, line 486: A comment here, why we always enable it, would be nice.

src/vmm/src/vstate/memory.rs
```diff
@@ -67,15 +66,15 @@ where
 
     /// Creates a GuestMemoryMmap with `size` in MiB backed by a memfd.
     fn memfd_backed(
-        mem_size_mib: usize,
+        regions: &[(GuestAddress, usize)],
```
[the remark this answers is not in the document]
roypat: It's because we need to sum up all the usizes to determine the size for the memfd, and we cannot iterate over the same `impl Iterator` twice.

src/vmm/src/arch/aarch64/mod.rs
roypat: Ah, yeah... I focussed a lot on the x86 version of this function and kinda forgot about the ARM one. Let me do the same things here I did over on x86 :)

src/vmm/src/vstate/memory.rs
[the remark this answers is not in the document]
bchalios: We are copying it right? Can't we just copy it first and then iterate it to get the total size? (Sorry, I'm probably missing some Rust nuance here).

bchalios:
All in all, this is a great change I think. To answer your initial question, I definitely helps massively in case we want to build `virtio-pmem` at some point.

src/vmm/src/arch/aarch64/mod.rs
```diff
@@ -58,9 +58,24 @@ pub const MMIO_MEM_SIZE: u64 = layout::DRAM_MEM_START - layout::MAPPED_IO_START;
...
-    let dram_size = min(size, layout::DRAM_MEM_MAX_SIZE);
-    vec![(GuestAddress(layout::DRAM_MEM_START), dram_size)]
+///
+/// The `offset` parameter specified the offset from [`layout::DRAM_MEM_START`].
+pub fn arch_memory_regions(offset: usize, size: usize) -> Vec<(GuestAddress, usize)> {
+    assert!(size > 0, "Attempt to allocate guest memory of length 0");
+    assert!(
+        offset.checked_add(size).is_some(),
+        "Attempt to allocate guest memory such that the address space would wrap around"
+    );
```
bchalios, line 68: I would also add a check that `offset + size <= layout::DRAM_MEX_MAX_SIZE`

bchalios (changes requested):

src/vmm/src/arch/aarch64/mod.rs
roypat: I deliberately didn't add this because the existing behavior of this function before this commit was "if size > DRAM_MAX_MEM_SIZE then truncate to DRAM_MAX_MEM_SIZE", so that's why I only added the assert for "offset < DRAM_MAX_MEM_SIZE" (which is needed to ensure that after truncation we still have a non-zero memory region)

src/vmm/src/arch/x86_64/mod.rs
roypat: Just out of curiosity, how so? :o

src/vmm/src/arch/aarch64/mod.rs
bchalios: My point is that maybe the behaviour of this function should not be that. Or, at the very least it would need to warn users that truncation is happening, so we can help them debugging issues.

bchalios (approved):

src/vmm/src/arch/x86_64/mod.rs
roypat: Ah, that makes sense. Will probably end up having to take that into account when it comes to allocating the swiotlb region over in #5108, since I guess we'll get PCI before secret hiding, but shouldn't be a problem tbh.

kalyazin (approved):

Issue #4866, opened 2024-10-22
[Feature Request] Steal Time support in ARM


Manciukic:
# Feature Request

PV Steal Time reporting on ARM is available in KVM since 5.10 at least but Firecracker doesn't enable it.

## Describe the desired solution

We could enable the PVTIME device in Firecracker on ARM to allow for stolen time reporting.

## Describe possible alternatives
- check if `rust-vmm` has some off-the-shelf implementation we could use

## Additional context

I found that this is the KVM patch that introduced it: https://lwn.net/Articles/797954/
And this is how QEMU integrated with it: https://patchew.org/QEMU/20200711101033.47371-1-drjones@redhat.com/20200711101033.47371-4-drjones@redhat.com/

It needs to be enabled via ioctl to the vcpu fd, there's an example in the kernel self tests: https://github.com/amazonlinux/linux/blob/kernel-5.10.215-203.850.amzn2/tools/testing/selftests/kvm/steal_time.c#L172

```
static void steal_time_init(struct kvm_vm *vm)
{
        struct kvm_device_attr dev = {
                .group = KVM_ARM_VCPU_PVTIME_CTRL,
                .attr = KVM_ARM_VCPU_PVTIME_IPA,
        };
        int i, ret;

        ret = _vcpu_ioctl(vm, 0, KVM_HAS_DEVICE_ATTR, &dev);
        if (ret != 0 && errno == ENXIO) {
                print_skip("steal-time not supported");
                exit(KSFT_SKIP);
        }

        for (i = 0; i < NR_VCPUS; ++i) {
                uint64_t st_ipa;

                vcpu_ioctl(vm, i, KVM_HAS_DEVICE_ATTR, &dev);

                dev.addr = (uint64_t)&st_ipa;

                /* ST_GPA_BASE is identity mapped */
                st_gva[i] = (void *)(ST_GPA_BASE + i * STEAL_TIME_SIZE);
                sync_global_to_guest(vm, st_gva[i]);

                st_ipa = (ulong)st_gva[i] | 1;
                ret = _vcpu_ioctl(vm, i, KVM_SET_DEVICE_ATTR, &dev);
                TEST_ASSERT(ret == -1 && errno == EINVAL, "Bad IPA didn't report EINVAL");

                st_ipa = (ulong)st_gva[i];
                vcpu_ioctl(vm, i, KVM_SET_DEVICE_ATTR, &dev);

[...]

PR #5139, opened 2025-04-04
Add Steal Time support in ARM

DakshinD:
Opening this PR as a draft as we continue working on this issue to add steal time support for ARM. Please see below for the changes we have implemented so far.

## Changes

- Added PVTime struct in `src/vmm/src/arch/aarch64/pvtime/mod.rs` that allocates and stores per-vCPU stolen_time regions
  - It offers a `new` constructor to create the device (allocating system memory and storing regions per vcpu), and `register_vcpu` which attempts to use `kvm_set_device_attr` to register the corresponding IPA for a vcpu
- Created a function `setup_pv_time` in `src/vmm/src/builder.rs` which is invoked in `build_microvm_for_boot`
  - Checks if PVTime is supported using `kvm_has_device_attr`
  - Creates a PVTime device
  - Iterates over vcpu fd's to register their IPAs for steal_time struct

- Started using debug/info logs while booting microvm for testing

### Future Progress
- We hope to continue making progress on this PR by:
  - Getting feedback on our current design of the PVTime device and `setup_pv_time`
  - Support snapshotting
  - Any other things we need to address

## Reason

Attempting to resolve issue #4866 

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to
 
 ### Added
 
+- [#5139](https://github.com/firecracker-microvm/firecracker/pull/5139): Added
+  support for [PVTime](https://docs.kernel.org/virt/kvm/arm/pvtime.html). This
+  is used to support steal time on ARM machines.
 - [#5048](https://github.com/firecracker-microvm/firecracker/pull/5048): Added
   support for [PVH boot mode](docs/pvh.md). This is used when an x86 kernel
   provides the appropriate ELF Note to indicate that PVH boot mode is supported.
--- a/src/vmm/src/builder.rs
+++ b/src/vmm/src/builder.rs
@@ -15,6 +15,8 @@ use linux_loader::cmdline::Cmdline as LoaderKernelCmdline;
 use userfaultfd::Uffd;
 use utils::time::TimestampUs;
 #[cfg(target_arch = "aarch64")]
+use vm_memory::GuestAddress;
+#[cfg(target_arch = "aarch64")]
 use vm_superio::Rtc;
 use vm_superio::Serial;
 use vmm_sys_util::eventfd::EventFd;
@@ -82,6 +84,9 @@ pub enum StartMicrovmError {
     CreateLegacyDevice(device_manager::legacy::LegacyDeviceError),
     /// Error creating VMGenID device: {0}
     CreateVMGenID(VmGenIdError),
+    /// Error enabling pvtime on vcpu: {0}
+    #[cfg(target_arch = "aarch64")]
+    EnablePVTime(crate::arch::VcpuArchError),
     /// Invalid Memory Configuration: {0}
     GuestMemory(crate::vstate::memory::MemoryError),
     /// Error with initrd initialization: {0}.
@@ -289,6 +294,13 @@ pub fn build_microvm_for_boot(
 
     attach_vmgenid_device(&mut vmm)?;
 
+    #[cfg(target_arch = "aarch64")]
+    if vcpus[0].kvm_vcpu.supports_pvtime() {
+        setup_pvtime(&mut vmm, &mut vcpus)?;
+    } else {
+        log::warn!("Vcpus do not support pvtime, steal time will not be reported to guest");
+    }
+
     configure_system_for_boot(
         &mut vmm,
         vcpus.as_mut(),
@@ -449,6 +461,16 @@ pub fn build_microvm_from_snapshot(
         }
     }
 
+    // Restore allocator state
+    #[cfg(target_arch = "aarch64")]
+    if let Some(pvtime_ipa) = vcpus[0].kvm_vcpu.pvtime_ipa {
+        allocate_pvtime_region(
+            &mut vmm,
+            vcpus.len(),
+            vm_allocator::AllocPolicy::ExactMatch(pvtime_ipa.0),
+        )?;
+    }
+
     // Restore vcpus kvm state.
     for (vcpu, state) in vcpus.iter_mut().zip(microvm_state.vcpu_states.iter()) {
         vcpu.kvm_vcpu
@@ -550,6 +572,44 @@ pub fn setup_serial_device(
     })));
     event_manager.add_subscriber(serial.clone());
     Ok(serial)
+}
+
+/// 64 bytes due to alignment requirement in 3.1 of https://www.kernel.org/doc/html/v5.8/virt/kvm/devices/vcpu.html#attribute-kvm-arm-vcpu-pvtime-ipa
+#[cfg(target_arch = "aarch64")]
+const STEALTIME_STRUCT_MEM_SIZE: u64 = 64;
+
+/// Helper method to allocate steal time region
+#[cfg(target_arch = "aarch64")]
+fn allocate_pvtime_region(
+    vmm: &mut Vmm,
+    vcpu_count: usize,
+    policy: vm_allocator::AllocPolicy,
+) -> Result<GuestAddress, StartMicrovmError> {
+    let size = STEALTIME_STRUCT_MEM_SIZE * vcpu_count as u64;
+    let addr = vmm
+        .resource_allocator
+        .allocate_system_memory(size, STEALTIME_STRUCT_MEM_SIZE, policy)
+        .map_err(StartMicrovmError::AllocateResources)?;
+    Ok(GuestAddress(addr))
+}
+
+/// Sets up pvtime for all vcpus
+#[cfg(target_arch = "aarch64")]
+fn setup_pvtime(vmm: &mut Vmm, vcpus: &mut [Vcpu]) -> Result<(), StartMicrovmError> {
+    // Alloc sys mem for steal time region
+    let pvtime_mem: GuestAddress =
+        allocate_pvtime_region(vmm, vcpus.len(), vm_allocator::AllocPolicy::LastMatch)?;
+
+    // Register all vcpus with pvtime device
+    for (i, vcpu) in vcpus.iter_mut().enumerate() {
+        vcpu.kvm_vcpu
+            .enable_pvtime(GuestAddress(
+                pvtime_mem.0 + i as u64 * STEALTIME_STRUCT_MEM_SIZE,
+            ))
+            .map_err(StartMicrovmError::EnablePVTime)?;
+    }
+
+    Ok(())
 }
 
 #[cfg(target_arch = "aarch64")]
--- a/src/vmm/src/persist.rs
+++ b/src/vmm/src/persist.rs
@@ -148,7 +148,7 @@ pub enum CreateSnapshotError {
 }
 
 /// Snapshot version
-pub const SNAPSHOT_VERSION: Version = Version::new(6, 0, 0);
+pub const SNAPSHOT_VERSION: Version = Version::new(7, 0, 0);
 
 /// Creates a Microvm snapshot.
 pub fn create_snapshot(
--- a/tests/integration_tests/functional/test_steal_time.py
+++ b/tests/integration_tests/functional/test_steal_time.py
@@ -1,0 +1,121 @@
+# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for verifying the PVTime device behavior under contention and across snapshots."""
+
+import time
+
+import pytest
+
+from framework.properties import global_props
+
+
+def get_steal_time_ms(vm):
+    """Returns total steal time of vCPUs in VM in milliseconds"""
+    _, out, _ = vm.ssh.run("grep -w '^cpu' /proc/stat")
+    steal_time_tck = int(out.strip().split()[8])
+    clk_tck = int(vm.ssh.run("getconf CLK_TCK").stdout)
+    return steal_time_tck / clk_tck * 1000
+
+
+@pytest.mark.skipif(
+    global_props.cpu_architecture != "aarch64", reason="Only run in aarch64"
+)
+def test_guest_has_pvtime_enabled(uvm_plain):
+    """
+    Check that the guest kernel has enabled PV steal time.
+    """
+    vm = uvm_plain
+    vm.spawn()
+    vm.basic_config()
+    vm.add_net_iface()
+    vm.start()
+
+    _, stdout, _ = vm.ssh.run("dmesg | grep 'stolen time PV'")
+    assert (
+        "stolen time PV" in stdout
+    ), "Guest kernel did not report PV steal time enabled"
+
+
+def test_pvtime_steal_time_increases(uvm_plain):
+    """
+    Test that PVTime steal time increases when both vCPUs are contended on the same pCPU.
+    """
+    vm = uvm_plain
+    vm.spawn()
+    vm.basic_config()
+    vm.add_net_iface()
+    vm.start()
+
+    # Pin both vCPUs to the same physical CPU to induce contention
+    vm.pin_vcpu(0, 0)
+    vm.pin_vcpu(1, 0)
+
+    # Start two infinite loops to hog CPU time
+    hog_cmd = "nohup bash -c 'while true; do :; done' >/dev/null 2>&1 &"
+    vm.ssh.run(hog_cmd)
+    vm.ssh.run(hog_cmd)
+
+    # Measure before and after steal time
+    steal_before = get_steal_time_ms(vm)
+    time.sleep(2)
+    steal_after = get_steal_time_ms(vm)
+
+    # Require increase in steal time
+    assert (
+        steal_after > steal_before
+    ), f"Steal time did not increase as expected. Before: {steal_before}, After: {steal_after}"
+
+
+def test_pvtime_snapshot(uvm_plain, microvm_factory):
+    """
+    Test that PVTime steal time is preserved across snapshot/restore
+    and continues increasing post-resume.
+    """
+    vm = uvm_plain
+    vm.spawn()
+    vm.basic_config()
+    vm.add_net_iface()
+    vm.start()
+
+    vm.pin_vcpu(0, 0)
+    vm.pin_vcpu(1, 0)
+
+    hog_cmd = "nohup bash -c 'while true; do :; done' >/dev/null 2>&1 &"
+    vm.ssh.run(hog_cmd)
+    vm.ssh.run(hog_cmd)
+
+    # Snapshot pre-steal time
+    steal_before = get_steal_time_ms(vm)
+
+    snapshot = vm.snapshot_full()
+    vm.kill()
+
+    # Restore microVM from snapshot and resume
+    restored_vm = microvm_factory.build()
+    restored_vm.spawn()
+    restored_vm.restore_from_snapshot(snapshot, resume=False)
+    snapshot.delete()
+
+    restored_vm.pin_vcpu(0, 0)
+    restored_vm.pin_vcpu(1, 0)
+    restored_vm.resume()
+
+    # Steal time just after restoring
+    steal_after_snap = get_steal_time_ms(restored_vm)
+
+    time.sleep(2)
+
+    # Steal time after running resumed VM
+    steal_after_resume = get_steal_time_ms(restored_vm)
+
+    # Ensure steal time persisted and continued increasing
+    tolerance = 2000  # 2.0 seconds tolerance for persistence check
+    persisted = (
+        steal_before < steal_after_snap and steal_after_snap - steal_before < tolerance
+    )
+    increased = steal_after_resume > steal_after_snap
+
+    assert (
+        persisted and increased
+    ), "Steal time did not persist through snapshot or failed to increase after resume"
--- a/src/vmm/src/arch/aarch64/vcpu.rs
+++ b/src/vmm/src/arch/aarch64/vcpu.rs
@@ -11,6 +11,7 @@ use std::mem::offset_of;
 use kvm_bindings::*;
 use kvm_ioctls::{VcpuExit, VcpuFd, VmFd};
 use serde::{Deserialize, Serialize};
+use vm_memory::GuestAddress;
 
 use super::get_fdt_addr;
 use super::regs::*;
@@ -42,6 +43,8 @@ pub enum VcpuArchError {
     Fam(vmm_sys_util::fam::Error),
     /// {0}
     GetMidrEl1(String),
+    /// Failed to set/get device attributes for vCPU: {0}
+    DeviceAttribute(kvm_ioctls::Error),
 }
 
 /// Extract the Manufacturer ID from the host.
@@ -115,6 +118,8 @@ pub struct KvmVcpu {
     /// Vcpu peripherals, such as buses
     pub peripherals: Peripherals,
     kvi: kvm_vcpu_init,
+    /// IPA of steal_time region
+    pub pvtime_ipa: Option<GuestAddress>,
 }
 
 /// Vcpu peripherals
@@ -148,6 +153,7 @@ impl KvmVcpu {
             fd: kvm_vcpu,
             peripherals: Default::default(),
             kvi,
+            pvtime_ipa: None,
         })
     }
 
@@ -243,6 +249,8 @@ impl KvmVcpu {
         // the boot state and turned secondary vcpus on.
         state.kvi.features[0] &= !(1 << KVM_ARM_VCPU_POWER_OFF);
 
+        state.pvtime_ipa = self.pvtime_ipa.map(|guest_addr| guest_addr.0);
+
         Ok(state)
     }
 
@@ -276,6 +284,13 @@ impl KvmVcpu {
         }
         self.set_mpstate(state.mp_state)
             .map_err(KvmVcpuError::RestoreState)?;
+
+        // Assumes that steal time memory region was set up already
+        if let Some(pvtime_ipa) = state.pvtime_ipa {
+            self.enable_pvtime(GuestAddress(pvtime_ipa))
+                .map_err(KvmVcpuError::RestoreState)?;
+        }
+
         Ok(())
     }
 
@@ -439,6 +454,38 @@ impl KvmVcpu {
     pub fn set_mpstate(&self, state: kvm_mp_state) -> Result<(), VcpuArchError> {
         self.fd.set_mp_state(state).map_err(VcpuArchError::SetMp)
     }
+
+    /// Check if pvtime (steal time on ARM) is supported for vcpu
+    pub fn supports_pvtime(&self) -> bool {
+        let pvtime_device_attr = kvm_bindings::kvm_device_attr {
+            group: kvm_bindings::KVM_ARM_VCPU_PVTIME_CTRL,
+            attr: kvm_bindings::KVM_ARM_VCPU_PVTIME_IPA as u64,
+            addr: 0,
+            flags: 0,
+        };
+
+        // Use kvm_has_device_attr to check if PVTime is supported
+        self.fd.has_device_attr(&pvtime_device_attr).is_ok()
+    }
+
+    /// Enables pvtime for vcpu
+    pub fn enable_pvtime(&mut self, ipa: GuestAddress) -> Result<(), VcpuArchError> {
+        self.pvtime_ipa = Some(ipa);
+
+        // Use KVM syscall (kvm_set_device_attr) to register the vCPU with the steal_time region
+        let vcpu_device_attr = kvm_bindings::kvm_device_attr {
+            group: KVM_ARM_VCPU_PVTIME_CTRL,
+            attr: KVM_ARM_VCPU_PVTIME_IPA as u64,
+            addr: &ipa.0 as *const u64 as u64, // userspace address of attr data
+            flags: 0,
+        };
+
+        self.fd
+            .set_device_attr(&vcpu_device_attr)
+            .map_err(VcpuArchError::DeviceAttribute)?;
+
+        Ok(())
+    }
 }
 
 impl Peripherals {
@@ -467,6 +514,8 @@ pub struct VcpuState {
     pub mpidr: u64,
     /// kvi states for vcpu initialization.
     pub kvi: kvm_vcpu_init,
+    /// ipa for steal_time region
+    pub pvtime_ipa: Option<u64>,
 }
 
 impl Debug for VcpuState {


src/vmm/src/arch/aarch64/pvtime/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+impl PVTime {
+    /// Creates a new PVTime device by allocating system memory for all vCPUs
+    pub fn new(
+        resource_allocator: &mut ResourceAllocator,
+        vcpu_count: u8,
+    ) -> Result<Self, PVTimeError> {
+        info!("Creating PVTime for {vcpu_count} vCPUs...");
+        // This returns the IPA(?) of the start of our shared memory region for all vCPUs.
+        // Q: Confirm that allocate_system_memory returns an IPA?
```
Manciukic, line 41: yes, the return of this call is an IPA, iow a guest physical address. You may use the type `GuestAddress` (a wrapper around `u64`) to make this clear.

src/vmm/src/arch/aarch64/pvtime/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+        debug!(
+            "Registering vCPU {} with stolen_time IPA = 0x{:x} (at userspace addr 0x{:x})",
+            vcpu_index, ipa_val, ipa_ptr as u64
+        );
+
+        // Use KVM syscall (kvm_set_device_attr) to register the vCPU with the steal_time region
+        let vcpu_device_attr = kvm_bindings::kvm_device_attr {
+            group: KVM_ARM_VCPU_PVTIME_CTRL,
+            attr: KVM_ARM_VCPU_PVTIME_IPA as u64,
+            addr: ipa_ptr as u64, // userspace address of attr data
```
Manciukic, line 98: The "IPA" ("Intermediate Physical Address" or simply Guest Physical Address, gpa) is already an address in the guest physical memory address space, so it should be passed directly.

This may be confusing as, when dealing with virtualization, there's not just one view of the memory (address space), but (at least) 4: virtual and physical addresses from guest and host. In this case 2 are important:
 - GPA: guest physical address, ie the physical view of the system from the point of view of the guest (VM). The `ipa` is an address in this space.
 - HVA: host virtual address, ie the address space of the hypervisor process (Firecracker). The `ipa_ptr` is an address in this space as it's a pointer to a memory location in FC virtual memory.

For a more detailed explanation:
 - https://www.alibabacloud.com/blog/599058
 - https://kernel.org/doc/Documentation/virtual/kvm/mmu.txt

Manciukic (changes requested):
Let's try making the proposed change to pass the right address and see if it works. 
I haven't left any comment regarding style or design to focus on functionality first. Once we get something working we can see how to make it "look better".

src/vmm/src/arch/aarch64/pvtime/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+
+use displaydoc::Display;
+use kvm_bindings::{KVM_ARM_VCPU_PVTIME_CTRL, KVM_ARM_VCPU_PVTIME_IPA};
+use log::{debug, info, warn};
+use thiserror::Error;
+
+use crate::device_manager::resources::ResourceAllocator;
+
+/// Size of the stolen_time struct in bytes, see 3.2.2 in DEN0057A
+pub const STEALTIME_STRUCT_MEM_SIZE: u64 = 16;
```
Manciukic, line 13: although it requires just 16 bytes, it needs to be 64B aligned. So we need to allocate 64B for each of them.

src/vmm/src/arch/aarch64/pvtime/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+        resource_allocator: &mut ResourceAllocator,
+        vcpu_count: u8,
+    ) -> Result<Self, PVTimeError> {
+        info!("Creating PVTime for {vcpu_count} vCPUs...");
+        // This returns the IPA(?) of the start of our shared memory region for all vCPUs.
+        // Q: Confirm that allocate_system_memory returns an IPA?
+        let base_addr = resource_allocator
+            .allocate_system_memory(
+                STEALTIME_STRUCT_MEM_SIZE * vcpu_count as u64,
+                16, // Q: We believe this to be 16, need confirmation.
```
Manciukic, line 45: > The base address must be 64 byte aligned and exist within a valid guest memory region.

https://www.kernel.org/doc/html/v5.8/virt/kvm/devices/vcpu.html#attribute-kvm-arm-vcpu-pvtime-ipa

Manciukic (commented):
I got confused by the KVM API as well! Passing a hva pointer to kvm is ok, but it needs to be 64B aligned.

The hint was in the KVM documentation:

> -EINVAL 	Base address not 64 byte aligned

https://www.kernel.org/doc/html/v5.8/virt/kvm/devices/vcpu.html#attribute-kvm-arm-vcpu-pvtime-ipa

DakshinD:
Thank you so much for your fast response! Totally missed the docs for 64 byte alignment, but that ended up fixing the bugs and I can successfully boot a microVM manually. If you think its a good time now, we can look to flesh out the implementation and "make it better".

Other than that, we believe the next step is adding unit/integration tests to make sure our device is working properly, as well as implementing persistence across snapshots. Any leads on how to do both of these things would be really helpful.

The lead we have been looking at for snapshotting is the `Persist` interface. 
- We store the `MicrovmState` in `save_snapshot`, so we need to first store our PVTime device somewhere in the vmm
- Then we implement `Persist` for PVTime, and store the `PVTimeState` we get from `save()` somewhere in the `MicrovmState` (could we extend MicrovmState to include this? or add this state to one of the other states already existing inside of it?)
- Finally, would need to write comprehensive tests to make sure this works

DakshinD:
Hi @Manciukic, I was just about to comment :), thanks for your response. We've started a bare bones implementation of snapshot persistence in the below commit. So if you also have the time to take a look at it, that would be awesome.

We are now trying to change the existing tests to include PVTime. Any input on our design choices we made would be appreciated again, specifically:
- Using either ONLY the base_addr or the hashmap in PVTime struct (we believe only the base_addr is necessary if the order of the vCPUs does not change across snapshots)
- Our restoration of the device inside `build_microvm_from_snapshot`
- Choosing to make the device an `Option<PVTime` inside the `struct Vmm`, and `Option<PVTimeState>` inside `struct MicrovmState`

src/vmm/src/builder.rs
```diff
@@ -289,6 +296,17 @@ pub fn build_microvm_for_boot(
...
+    #[cfg(target_arch = "aarch64")]
+    if let Err(e) = setup_pv_time(&mut vmm, vcpus.as_mut()) {
+        match e {
+            StartMicrovmError::PVTimeNotSupported(e) => {
+                warn!("PVTime not supported: {}", e);
+            }
+            other => return Err(other),
+        }
+    }
+
```
Manciukic, line 309: With the changes proposed below, this would become
```
vmm.pvtime = if PVTime::is_supported(&vcpus[0].kvm_vcpu.fd) {
        Some(setup_pv_time(&mut vmm, vcpus.as_mut())?)
    } else {
        warn!("PVTime is not supported by KVM. Steal time will not be reported to the guest.");
        None
};
```

src/vmm/src/arch/aarch64/pvtime/mod.rs
```diff
@@ -0,0 +1,148 @@
+// Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Manciukic, line 1: nit: the path could be `src/vmm/src/arch/aarch64/pvtime.rs`

Manciukic (commented):
Overall looks good. I'd just move most of the pv_time specific logic into the new module to avoid cluttering the builder.rs even more.

We also need to bump the snapshot version and add a changelog entry, other than the integration tests.

Manciukic:
I tried to run the branch through our CI and there are a couple of issues: https://buildkite.com/firecracker/firecracker-pr/builds/13103#
 -  x86 doesn't compile, for example:
```
 error[E0433]: failed to resolve: could not find `aarch64` in `arch`
```
 - arm  has a few failures
```
integration_tests/build/test_clippy.py::test_rust_clippy[aarch64-unknown-linux-gnu] FAILED                                                               [ 12%]
integration_tests/build/test_clippy.py::test_rust_clippy[aarch64-unknown-linux-musl] FAILED                                                              [ 25%]
integration_tests/build/test_coverage.py::test_coverage FAILED                                                                                           [ 37%]
integration_tests/build/test_dependencies.py::test_unused_dependencies PASSED                                                                            [ 50%]
integration_tests/build/test_gdb.py::test_gdb_compiles PASSED                                                                                            [ 62%]
integration_tests/build/test_seccomp_no_redundant_rules.py::test_redundant_seccomp_rules PASSED                                                          [ 75%]
integration_tests/build/test_unittests.py::test_unittests FAILED                                                                                         [ 87%]
integration_tests/build/test_unittests.py::test_benchmarks_compile PASSED                                                                                [100%]
```
 - unit tests need to be updated with the new field
 - clippy is unhappy

But the good news is functional and performance tests passed :tada: 

To run these tests locally you can run:
```
./tools/devtool test -y -- integration_tests/build
```

DakshinD:
Managed to fix the errors reported in the build as well as clippy errors, would be awesome if it could be re-run.

One small question was if there was a way to view the code coverage tests. I see that a `.profraw` file gets generated, and I managed to find a way to convert it to `.profdata` and view it on a html file, but was wondering if there was an easier way to view which files/lines were covered.

We also have 2 random tests failing during integration tests, both having to do with `gicv2`
-`arch::aarch64::gic::gicv2::regs::dist_regs::tests::test_access_dist_regs stdout`
-`arch::aarch64::gic::gicv2::regs::tests::test_vm_save_restore_state stdout`

We started looking at unit/integration tests, as per your suggestions, and were hoping to get a small lead as to going about it. 

- For the unit tests in `src/vmm/src/builder.rs` as well as `test_microvm_state_snapshot`, we couldn't find a way to get the `Vec<Vcpus>` in context to setup the PVTime device. Is it necessary to mimic these existing tests for the pvtime device?
- I think we have a handle on the integration test with 2 cpus, so will go ahead writing that python test.

Manciukic:
> For the unit tests in `src/vmm/src/builder.rs` as well as `test_microvm_state_snapshot`, we couldn't find a way to get the `Vec<Vcpus>` in context to setup the PVTime device. Is it necessary to mimic these existing tests for the pvtime device?

It's difficult to unit test code that mostly performs syscalls (ioctls on vcpu fd in this case), so the policy to not over-complicate the codebase with mocks is to test these features through end-to-end integration tests only.
I think it would be enough to have unit tests inside `PVTime` to ensure the logic (excluding the `ioctl`s) is correct.
`builder.rs` is basically untesteable and is tested mostly

DakshinD:
Thanks for the clarification on that. We've added some very basic unit tests to PVTime and also 2 integration tests for basic checking and snapshotting. If there's any changes/next steps we need to take, please let us know. 

A re-run of the test suite would also be very helpful so we can nail down any bugs, as it was a bit buggy locally.

Manciukic:
> Thanks for the clarification on that. We've added some very basic unit tests to PVTime and also 2 integration tests for basic checking and snapshotting. If there's any changes/next steps we need to take, please let us know.

Can you also add a basic test that checks that the guest has enabled steal PV by looking at the kernel logs?
```
uvm.ssh.run("dmesg | grep 'stolen time PV'")
```
Make sure that only runs on aarch64.

> A re-run of the test suite would also be very helpful so we can nail down any bugs, as it was a bit buggy locally.

Done. x86 has the same issues as above: https://buildkite.com/firecracker/firecracker-pr/builds/13138#0196241a-77dd-4fe4-9023-e053f8004c2d

DakshinD:
Added that extra test and should've completely fixed the x86 errors, should be good for test suite re-run.

tests/integration_tests/functional/test_pvtime.py
```diff
@@ -0,0 +1,146 @@
...
+    # Steal time after running resumed VM
+    _, out_after_resume, _ = restored_vm.ssh.run("grep '^cpu[0-9]' /proc/stat")
+    steal_after_resume = [
+        int(re.split(r"\s+", line.strip())[8])
+        for line in out_after_resume.strip().splitlines()
+    ]
+
+    # Ensure steal time persisted and continued increasing
+    persisted = sum(steal_before) + 100 <= sum(steal_after_snap)
+    increased = sum(steal_after_resume) > sum(steal_after_snap)
```
Manciukic, line 142: As above, this may be flaky (I'm even surprised it's working at the moment without any race!). For the purpose of our functional test we want to know that:
 - the value on restore is the same (approx) as the value at the time of snapshot
 - the steal time keeps increasing after the snapshot is restored

So the first condition needs to check that after_snap is between before and before+tolerance (we can remove the sleep above and keep the tolerance as 1s maybe). The second condition is ok.

tests/integration_tests/functional/test_pvtime.py
Manciukic, line 1: as some of these tests also run on x86 (and I don't see any reason not to), maybe it's best to rename the test suite to `test_steal_time.py`

src/vmm/src/arch/aarch64/pvtime.rs
```diff
@@ -0,0 +1,209 @@
...
+        Ok(())
+    }
+
+    /// Register all vCPUs with their pre-allocated steal time regions
+    pub fn register_all_vcpus(&self, vcpus: &mut [Vcpu]) -> Result<(), PVTimeError> {
+        // Register the vcpu with the pvtime device to map its steal time region
+        for (i, vcpu) in vcpus.iter().enumerate() {
+            #[allow(clippy::cast_possible_truncation)]
+            // We know vcpu_count is u8 according to VcpuConfig
+            self.register_vcpu(i as u8, &vcpu.kvm_vcpu.fd)?;
```
Manciukic, line 119: nit: I think a `try_into` would be nicer
```suggestion
            self.register_vcpu(i.try_into().expect("cpu index should be less than 256"), &vcpu.kvm_vcpu.fd)?;
```

src/vmm/src/builder.rs
```diff
@@ -12,13 +11,17 @@ use std::sync::{Arc, Mutex};
 use event_manager::{MutEventSubscriber, SubscriberOps};
 use libc::EFD_NONBLOCK;
 use linux_loader::cmdline::Cmdline as LoaderKernelCmdline;
+#[cfg(target_arch = "aarch64")]
+use log::warn;
```
Manciukic, line 15: let's avoid conditionally importing non-arch-specific dependencies. We can probably just avoid the `use` and call `log::warn` directly.

src/vmm/src/arch/aarch64/pvtime.rs
```diff
@@ -0,0 +1,209 @@
...
+use crate::snapshot::Persist;
+
+/// 64 bytes due to alignment requirement in 3.1 of https://www.kernel.org/doc/html/v5.8/virt/kvm/devices/vcpu.html#attribute-kvm-arm-vcpu-pvtime-ipa
+pub const STEALTIME_STRUCT_MEM_SIZE: u64 = 64;
+
+/// Represent PVTime device for ARM
+#[derive(Debug)]
+pub struct PVTime {
+    /// Number of vCPUs
+    vcpu_count: u8,
```
Manciukic, line 21: I see down below that having this as `u8` adds some complication in handling the downcasts. As there's no tangible difference in using a `u64` I'd suggest to just use the bigger type and avoid all those downcast mess.

tests/integration_tests/functional/test_pvtime.py
```diff
@@ -0,0 +1,146 @@
...
+    # Measure steal time after
+    _, out_after, _ = vm.ssh.run("grep '^cpu[0-9]' /proc/stat")
+    steal_after = sum(
+        int(re.split(r"\s+", line.strip())[8])
+        for line in out_after.strip().splitlines()
+    )
+
+    # Require increase in steal time
+    assert (
+        steal_after - steal_before >= 200
```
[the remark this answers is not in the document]
DakshinD: I also observed this was flaky, due to various other interrupts I believe. It was prevalent when we compared steal time on a per CPU basis as well. Due to the fact that "relaxing the constraint" or creating some tolerance could be finicky, I think checking a simple increase could be better.

src/vmm/src/arch/aarch64/pvtime.rs
DakshinD: Avoiding this by using a u64 for vcpu_index

src/vmm/src/arch/aarch64/pvtime.rs
DakshinD: I will go this route and use a u64 as the vcpu_count. I just assume that this places some sort of trust in the fact that vcpus.len() (during creation of pvtime) will never be above 256 and create a invalid vcpu index. Is any assert/check necessary here or is that situation not possible.

DakshinD:
Thanks for the revisions @Manciukic. I've gone ahead and:
- changed `vcpu_index` to a u64 as per your suggestion and I think it cleans up the code nicely. It just assumes that `vcpus.len()` will not exceed 256 which I believe is a valid assumption?
- Fixed the integration tests to check for increase only, and also cleaned it up with helper function.

Please let me know if there is anything else on that end, and looking towards a merge, what would need to be done such as:
- changing snapshot version?
- changelog
- squashing my commits into ~3 substantial ones for clarity?

Manciukic:
> Thanks for the revisions @Manciukic. I've gone ahead and:
> 
>     * changed `vcpu_index` to a u64 as per your suggestion and I think it cleans up the code nicely. It just assumes that `vcpus.len()` will not exceed 256 which I believe is a valid assumption?

It's the assumption we already make in other parts of the code. I think it's fine.

> 
>     * Fixed the integration tests to check for increase only, and also cleaned it up with helper function.

Thanks!

> 
> 
> Please let me know if there is anything else on that end, and looking towards a merge, what would need to be done such as:
> 
>     * changing snapshot version?
> 
>     * changelog
> 
>     * squashing my commits into ~3 substantial ones for clarity?

Yup, all three are required for merging. I can't think of anything else at the moment, other than keeping the branch up-to-date with main.

I will take a deeper look next week after the Easter holidays.

DakshinD:
Just updated the snapshot version by minor version, added changelog, and attempted to rebase the commits but git freaked out so hopefully I did it right. Also apologies it seemed to auto-request reviews after I force pushed.

src/vmm/src/persist.rs
```diff
@@ -152,7 +157,7 @@ pub enum CreateSnapshotError {
 }
 
 /// Snapshot version
-pub const SNAPSHOT_VERSION: Version = Version::new(5, 0, 0);
+pub const SNAPSHOT_VERSION: Version = Version::new(6, 1, 0);
```
Manciukic, line 160: We only update the major version at every change.

```suggestion
pub const SNAPSHOT_VERSION: Version = Version::new(6, 0, 0);
```

tests/integration_tests/functional/test_steal_time.py
```diff
@@ -0,0 +1,119 @@
...
+    steal_after_snap = get_steal_time_ms(restored_vm)
+
+    time.sleep(2)
+
+    # Steal time after running resumed VM
+    steal_after_resume = get_steal_time_ms(restored_vm)
+
+    # Ensure steal time persisted and continued increasing
+    tolerance = 1500  # 1.5 seconds tolerance for persistence check
+    persisted = steal_after_snap - steal_before < tolerance
```
Manciukic, line 114: We also want to assert it didn't go back.

```suggestion
    persisted = steal_before < steal_after_snap and steal_after_snap - steal_before < tolerance
```

Manciukic (commented):
mostly lgtm except a few minor things. Another small nit: please clean up the commit messages a little (make sure to mention what, why, and how).

tests/integration_tests/functional/test_pvtime.py
Manciukic: We're expecting steal time to increase by at least 1s, and we're checking exactly one second after. Any small measurement error may make it fail, but I don't think we need this kind of precision. We're not trying to test how accurate is the kernel, we just want to make sure that the steal time reporting is working after a snapshot resume.

src/vmm/src/arch/aarch64/pvtime.rs
Manciukic: I don't think it's necessary to assert the vcpu count as for what the pvtime it's concerned, as long as the vcpu fd is valid and the index is less than the total count, everything should work.

src/vmm/src/builder.rs
```diff
@@ -548,6 +590,24 @@ pub fn setup_serial_device(
     Ok(serial)
 }
 
+/// Sets up the pvtime device.
+#[cfg(target_arch = "aarch64")]
+fn setup_pv_time(vmm: &mut Vmm, vcpus: &mut [Vcpu]) -> Result<PVTime, StartMicrovmError> {
+    use crate::arch::aarch64::pvtime::PVTime;
```
ShadowCurse, line 597: This import seems unnecessary

src/vmm/src/builder.rs
```diff
@@ -289,6 +295,17 @@ pub fn build_microvm_for_boot(
...
+    // Attempt to setup PVTime, continue if not supported
+    #[cfg(target_arch = "aarch64")]
+    {
+        vmm.pv_time = if PVTime::is_supported(&vcpus[0].kvm_vcpu.fd) {
+            Some(setup_pv_time(&mut vmm, vcpus.as_mut())?)
+        } else {
+            warn!("PVTime is not supported by KVM. Steal time will not be reported to the guest.");
+            None
+        };
+    }
```
ShadowCurse: This can be moved into `create_vmm_and_vcpus`

src/vmm/src/arch/aarch64/pvtime.rs
```diff
@@ -18,7 +18,7 @@ pub const STEALTIME_STRUCT_MEM_SIZE: u64 = 64;
 #[derive(Debug)]
 pub struct PVTime {
     /// Number of vCPUs
-    vcpu_count: u8,
+    vcpu_count: u64,
```
ShadowCurse, line 21: Please move rust related fixes/changes into the first one, so they are not scattered. Let second commit be just a python integration test.

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to
 
 ### Added
 
+- [#5139](https://github.com/firecracker-microvm/firecracker/pull/5139): Added
```
ShadowCurse, line 13: Please move `CHANGELOG` update into a separate commit.

src/vmm/src/arch/aarch64/pvtime.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// Error while setting or getting device attributes for vCPU: {0}
+    DeviceAttribute(kvm_ioctls::Error),
+}
+
+impl PVTime {
+    /// Helper function to get the IPA of the steal_time region for a given vCPU
+    fn get_steal_time_region_addr(&self, vcpu_index: u64) -> Result<GuestAddress, PVTimeError> {
+        if vcpu_index >= self.vcpu_count {
+            return Err(PVTimeError::InvalidVcpuIndex(vcpu_index));
+        }
```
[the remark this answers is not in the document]
ShadowCurse: This will also change return type to just `GuestAddress`

ShadowCurse (commented):
Overall looks reasonable. Couple minor fixes and it should be good to go.

src/vmm/src/arch/aarch64/pvtime.rs
roypat, line 1: Since we have one pvtime memory area per vcpu, I think it makes more have a single field `pvtime_ipa: Option<GuestAddress>` inside each `struct KvmVcpu`, instead of one struct that manages PV time for all vcpus. Then a lot of the functions that are currently defined on struct PvTime but take vcpu[_index] as arguments will just become `&self` functions on struct KvmVcpu, and the whole setup would essentially become

```
let stealtime_mem = ResourceAllocator.allocate(...);

for vcpu in vcpus {
	if vcpu.supports_stealtime() { // does .has_device_attr()
        vcpu.enable_stealtime(stealtime_mem + i * STEALTIME_STRUCT_MEM_SIZE)?;  // does .set_device_attr()
    }
}
```

roypat (changes requested):

src/vmm/src/arch/aarch64/pvtime.rs
DakshinD: Hi, thanks for looking over the PR. So from my understanding, you want to move methods such as `is_supported` and `register_vcpu` inside `KvmVcpu`? 

This does make sense to me, but then I believe some code might become redundant, as if we add `pvtime_ipa` to `KvmVcpu` and persist that, we would also need to persist the `base_ipa` inside `PVTime` to restore the system memory. Unless we do one or the other, maybe get rid of `PVTime` entirely and handle the restore logic somewhere else?

roypat: I think we can persist the `pvtime_ipa` inside each `struct VcpuState` as you say. I don't think there is a need to persist the `base_ipa` at all in this case, so `PVTime` would be gone entirely

DakshinD: Thanks for the clarification, as this is a large change to my existing PR that I've been working on, wanted to confirm the "go-ahead" on this by listing out my thought process.

- Move methods such as support checking and registering ipa into `KvmVcpu`
- `setup_pv_time` will change to allocating system memory itself and iterating over all vcpus and registering them as you proposed in your code snippet above
- The logic to restore the pvtime device in `build_microvm_from_snapshot` disappears as we will move the `ipa` per vcpu into the `VcpuState` that gets saved in `Vcpu::paused` after the vcpu is sent an event to save in `Vmm::save_vcpu_states`. This requires me to change the `save_state` and `restore_state` (re-register ipa in here) methods of `KvmVcpu`

If these changes are ok, I will go ahead and get them done ASAP. One question I did have was what would happen to the `PVTime` unit tests. Would there be no unit tests, just integration tests now? Or would I move them into `KvmVcpu`?

roypat: Yes, that sounds good to me, thanks! As for the unit tests, yeah, it does seems like the code they're currently testing would just go away, so just having integration tests seems fine to me.

DakshinD:
Thanks for the PR review earlier. 

I have gone ahead and:
- Addressed @roypat design changes. I have gone ahead and got rid of `PVTime` device, moved the logic into `KvmVcpu`, and just left some setup logic inside builder.rs. 
- I have cleaned up the commit messages, as well as fixed the small nit on the integration test as @Manciukic said.
- I have separated the commits into rust and integration test commits, as well as a separate one for the Changelog as @ShadowCurse  said.

I've double checked that style, clippy, steal time tests pass, etc. Any feedback on this new design would be welcome, as well as a run of the test-suite if possible. I believe (hopefully) that it should build fine on x86 as well. 

Thank you!

src/vmm/src/builder.rs
```diff
@@ -552,6 +573,44 @@ pub fn setup_serial_device(
...
+/// 64 bytes due to alignment requirement in 3.1 of https://www.kernel.org/doc/html/v5.8/virt/kvm/devices/vcpu.html#attribute-kvm-arm-vcpu-pvtime-ipa
+#[cfg(target_arch = "aarch64")]
+const STEALTIME_STRUCT_MEM_SIZE: u64 = 64;
+
+/// Helper method to allocate steal time region
+#[cfg(target_arch = "aarch64")]
+fn allocate_pvtime_region(
+    vmm: &mut Vmm,
+    vcpu_count: usize,
+    policy: vm_allocator::AllocPolicy,
```
roypat, line 585: If you drop the call to this on the restore path, you can also drop this `policy` argument (and just inline this into `setup_pvtime` probably).

roypat (commented):
Awesome, thanks for doing that change! Everything looks a lot simpler now :D 

Just a small comment about the restore code, but then this looks good to go from me too!

roypat (dismissed):

src/vmm/src/builder.rs
```diff
@@ -449,6 +460,16 @@ pub fn build_microvm_from_snapshot(
...
 
+    // Need to alloc sys mem for steal time regions before vcpu setup
+    #[cfg(target_arch = "aarch64")]
+    if let Some(pvtime_ipa) = vcpus[0].kvm_vcpu.pvtime_ipa {
+        allocate_pvtime_region(
+            &mut vmm,
+            vcpus.len(),
+            vm_allocator::AllocPolicy::ExactMatch(pvtime_ipa.0),
+        )?;
+    }
```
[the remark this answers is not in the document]
Manciukic: Yeah, that seems to be the pattern followed for vmgenid and mmio devices. Checking again, this allocator is used 3 times (`mptable`, `acpi`, `vmgenid`), but only `vmgenid` restores its state, while the other 2 places don't, which means it's already out-of-sync :(
Restoring the allocator state seems like a better option than remembering to go back and allocate again all places (but it's also out of scope). I'll open an issue for that.

Manciukic:
One more issue on x86 :( 

```
error[E0432]: unresolved import `crate::arch::VcpuArchError`
  --> src/vmm/src/builder.rs:23:39
   |
23 | use crate::arch::{ConfigurationError, VcpuArchError, configure_system_for_boot, load_kernel};
   |                                       ^^^^^^^^^^^^^
   |                                       |
   |                                       no `VcpuArchError` in `arch`
   |                                       help: a similar name exists in the module: `KvmArchError`
warning: unused import: `vm_memory::GuestAddress`
  --> src/vmm/src/builder.rs:17:5
   |
17 | use vm_memory::GuestAddress;
   |     ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```

src/vmm/src/builder.rs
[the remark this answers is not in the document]
roypat: hahahaha, if we already aren't doing this consistently, I have a weak opinion in favor of not doing it here either xP

DakshinD:
@Manciukic Just pushed that fix for x86, sorry about that. Gonna try getting the cross-compilation working locally so I can test that soon.

DakshinD:
Thanks for pointing that out. It was telling me `the `x86_64-unknown-linux-gnu` target may not be installed`, so I would do `rustup target add x86_64-unknown-linux-gnu` and it would tell me everything was fine. But I had to run it with sudo for it to actually download the needed things :( Oversight on my part

roypat (approved):

Manciukic (approved):

PR #5142, opened 2025-04-07
Clarify supported CPU models for CPU templates + CPU template-related tests cleanups

zulinx86:
## Changes

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.


tests/data/msr/msr_list_C3_INTEL_CASCADELAKE_5.10host_5.10guest.csv
pb8o, line 1: can we squash this one with the one that removes it? are they that different?

zulinx86: I'd like to do so, but I don't treat it in this PR. This is a preliminary work for an upcoming custom CPU template for Intel Sapphire Rapids. There are still many quirks, I'll clean it up more after the Intel Sapphire Rapids tasks.

zulinx86: OK, I dropped those changes to make this PR focus on clarification of CPU models supported by CPU templates.

.buildkite/pipeline_cpu_template.py
```diff
@@ -70,7 +70,6 @@ class BkStep(str, Enum):
             "m5n.metal",
             "m6i.metal",
             "m6a.metal",
-            "m7a.metal-48xl",
```
zulinx86, line 73: Note that we remove this because m7a.metal does not support any existing CPU templates. Without this removal, we'll get an error in the artifact upload step.

src/vmm/src/arch/x86_64/cpu_model.rs
```diff
@@ -64,49 +87,14 @@ impl From<&CpuModel> for u32 {
     }
 }
```
roypat, line 88: nit: while we're at it, this `From<&CpuModel> for u32` impl is dead code

src/vmm/src/arch/x86_64/vcpu.rs
```diff
@@ -915,24 +906,37 @@ mod tests {
...
+                );
                 assert!(!t2a_res);
             }
             cpuid::VENDOR_ID_AMD => {
                 assert!(!t2_res);
                 assert!(!c3_res);
                 assert!(!t2s_res);
                 assert!(!t2cl_res);
-                assert!(t2a_res);
+                assert_eq!(t2a_res, cpu_model == MILAN_FMS);
```
roypat, line 939: Mhh, I think we don't need the `match` here? As in, this thing could just be

```
assert_eq!(t2_res, [SKYLAKE_FMS, CASCADE_LAKE_FMS, ICE_LAKE_FMS].contains(&cpu_model));
assert_eq!(c3_res, [SKYLAKE_FMS, CASCADE_LAKE_FMS, ICE_LAKE_FMS].contains(&cpu_model));
...
assert_eq!(t2a_res, [MILAN_FMS].contains(&cpu_model));
```

no?

src/vmm/src/cpu_config/x86_64/custom_cpu_template.rs
```diff
@@ -27,12 +29,18 @@ impl GetCpuTemplate for Option<CpuTemplateType> {
...
+                    let cpu_model = CpuModel::get_cpu_model();
                     match template {
                         StaticCpuTemplate::C3 => {
                             if &vendor_id != VENDOR_ID_INTEL {
                                 return Err(CpuVendorMismatched);
+                            } else if cpu_model != SKYLAKE_FMS
+                                && cpu_model != CASCADE_LAKE_FMS
+                                && cpu_model != ICE_LAKE_FMS
+                            {
+                                return Err(InvalidCpuModel);
```
roypat, line 41: Mh, we're repeating these in quite a few places. Maybe we could have `StaticCpuTemplate::supported_cpu_models(&self) -> &'static [CpuModel]`, and then just a single check at the top of this huge match that goes `if !template.supported_cpu_models().contains(CpuModel::get_cpu_model()) { return Err(InvalidCpuModel) }`?

that could then also be used in the tests below, to avoid having to repeat the set of supported hardware all the time.

src/vmm/src/cpu_config/x86_64/custom_cpu_template.rs
```diff
@@ -27,12 +29,18 @@ impl GetCpuTemplate for Option<CpuTemplateType> {
                 CpuTemplateType::Custom(template) => Ok(Cow::Borrowed(template)),
                 CpuTemplateType::Static(template) => {
                     let vendor_id = get_vendor_id_from_host().map_err(GetCpuVendor)?;
+                    let cpu_model = CpuModel::get_cpu_model();
                     match template {
                         StaticCpuTemplate::C3 => {
                             if &vendor_id != VENDOR_ID_INTEL {
                                 return Err(CpuVendorMismatched);
```
roypat, line 36: This is not really needed anymore - all of the 3 cpu models below are Intel

goes for every branch actually

tests/data/msr/msr_list_T2A_AMD_GENOA_5.10host_5.10guest.csv
Manciukic, line 1: This is probably due to my ignorance, but do we have a test that would check the msr changing on Genoa?

src/vmm/src/cpu_config/x86_64/custom_cpu_template.rs
zulinx86: isn't the vendor check required to error AMD?

docs/cpu_templates/cpu-templates.md
```diff
@@ -48,14 +48,14 @@ Firecracker supports two types of CPU templates:
...
-| T2CL         | Intel      | Cascade Lake or newer |
-| T2S          | Intel      | any                   |
-| V1N1         | ARM        | Neoverse V1           |
+| CPU template | CPU vendor | CPU model                       |
+| ------------ | ---------- | ------------------------------- |
+| C3           | Intel      | Skylake, Cascade Lake, Ice Lake |
+| T2           | Intel      | Skylake, Cascade Lake, Ice Lake |
+| T2A          | AMD        | Milan                           |
+| T2CL         | Intel      | Cascade Lake, Ice Lake          |
+| T2S          | Intel      | Skylake, Cascade Lake           |
```
Manciukic, line 57: why not icelake?

src/vmm/src/cpu_config/x86_64/custom_cpu_template.rs
roypat: We would reject any AMD CPU in the check that verifies the exact CPU model anyway, because the only 3 CPU models we allow are Intel ones.

docs/cpu_templates/cpu-templates.md
```diff
@@ -72,7 +72,7 @@ Firecracker expects the host to always be running the latest version of the
 microcode.
 
 The T2CL template is mapped to be close to Intel Cascade Lake. It is not safe to
-use it on Intel CPUs older than Cascade Lake (such as Skylake).
+use it on Intel CPUs other than Cascade Lake and Ice Lake.
```
Manciukic, line 75: nit: I'd use a more assertive form, rather than a double negation: "It is only safe to use on Cascade Lake and Ice Lake"

src/vmm/src/cpu_config/x86_64/custom_cpu_template.rs
zulinx86: Although I don't think FMS value conflicts between vendors, I'd like to keep it just in case.

roypat: Ah, didn't realize we dont have guarantees about these not being reused. Sorry!

In that case, let's also hoist this check above the match though? E.g. have some sort of `StaticCpuTemplate::supported_vendor(&self) -> vendor_t` and just check that once against the vendor_id, instead of doing it in every branch

zulinx86: yeah, makes sense to me!

roypat (approved):

zulinx86:
@roypat sorry. I needed to fix an integration test failure that sets C3 template unconditionally on Intel Sapphire Rapids.

roypat (approved):

docs/cpu_templates/cpu-templates.md
Manciukic: yeah, I guess it was done before Icelake. Then we dropped the idea of the common migration pool, so Icelake got left out I believe.
As it's deprecated, I'm fine leaving it as is, although I think it'd work on Icelake as well.

tests/data/msr/msr_list_T2A_AMD_GENOA_5.10host_5.10guest.csv
Manciukic: ack

Manciukic (approved):

docs/cpu_templates/cpu-templates.md
zulinx86: actually I don't think it works properly. For example, Ice Lake enumerates PSFD (CPUID.(EAX=07H,ECX=2):EDX[1]) and it's a security feature bit, but T2S doesn't mask it. That means a guest booted on Ice Lake may use PSFD bits on IA32_SPEC_CTRL MSR but it is not available when restored on older generations.

PR #5160, opened 2025-04-17
Refactor VirtIO devices, preparing for a PCI transport layer

bchalios:
## Changes

This does a few changes around interrupts for VirtIO devices and code organization for being able to handle multiple transports for VirtIO devices:

1. Creates a `transport` module within `vmm::devices::virtio` and moves `mmio` module inside it
2. Adds a `VirtioInterrupt` trait that abstracts away the functionality of interrupts used by VirtIO devices. At the moment, only `IrqTrigger` implements `VirtioInterrupt`.
3. Changes the moment in which an `IrqTrigger` object is assigned to a VirtIO device. Up until now we created this object when creating the device itself. This PR changes this logic so that it is the transport layer that creates this interrupt and passes it over to the device during activation time. This is how things are done with PCIe in the Cloud Hypervisor implementation. It also, doesn't change anything for us, semantically; devices only trigger interrupts after having been activated.

## Reason

Preparing for adding PCIe support

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.


src/vmm/src/devices/virtio/device.rs
```diff
@@ -36,7 +36,7 @@ impl DeviceState {
     }
 
     /// Gets the memory and interrupt attached to the device if it is activated.
-    pub fn active_state(&self) -> Option<(&GuestMemoryMmap, Arc<IrqTrigger>)> {
+    pub fn active_state(&self) -> Option<(&GuestMemoryMmap, Arc<dyn VirtioInterrupt>)> {
```
ShadowCurse, line 39: I see why you use dynamic dispatch here, and maybe it is fine from a perf perspective, but I wonder if we can somehow avoid it. Technically any device will be either `MMIO` or `PCIe`. So maybe we can make them generic over `VirtioInterrupt`? This will be a bigger change, but should not be a difficult one. Eventually all devices are abstracted as `VirtioDevice` anyway.

src/vmm/src/devices/virtio/balloon/device.rs
```diff
@@ -401,9 +401,12 @@ impl Balloon {
         Ok(())
     }
 
-    pub(crate) fn signal_used_queue(&self) -> Result<(), BalloonError> {
+    pub(crate) fn signal_used_queue(&self, qidx: usize) -> Result<(), BalloonError> {
         self.interrupt_trigger()
-            .trigger_irq(IrqType::Vring)
+            .trigger(VirtioInterruptType::Queue(
+                qidx.try_into()
+                    .unwrap_or_else(|_| panic!("balloon: invalid queue id: {qidx}")),
```
ShadowCurse, line 408: Let's change `qidx` to be `u16` and remove this ugliness. Or add a the conversion as a first line:
```rust
let idx: u16 = qidx.try_into().expect("Invalid queue index");
```

src/vmm/src/devices/virtio/block/virtio/device.rs
```diff
@@ -525,10 +527,12 @@ impl VirtioBlock {
         self.disk.update(disk_image_path, self.read_only)?;
         self.config_space.capacity = self.disk.nsectors.to_le(); // virtio_block_config_space();
 
-        // Kick the driver to pick up the changes.
-        self.interrupt_trigger()
-            .trigger_irq(IrqType::Config)
-            .unwrap();
+        // Kick the driver to pick up the changes. (But only if the device is already activated).
+        if self.is_activated() {
```
ShadowCurse, line 531: Considering this should never be called if device is not activated, this should be an assert.

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -390,20 +390,22 @@ impl Net {
...
             let (_, interrupt) = self
                 .device_state
                 .active_state()
                 .expect("Device is not initialized");
-            interrupt.trigger_irq(IrqType::Vring).map_err(|err| {
-                self.metrics.event_fails.inc();
-                DeviceError::FailedSignalingIrq(err)
-            })?;
+            interrupt
+                .trigger(VirtioInterruptType::Queue(qidx.try_into().unwrap()))
```
ShadowCurse, line 404: We really need to make queue indexes `u16`. Otherwise it is painful to watch comptime constants needing this treatment.

src/vmm/src/devices/virtio/vhost_user.rs
```diff
@@ -442,7 +443,17 @@ impl<T: VhostUserHandleBackend> VhostUserHandleImpl<T> {
             // No matter the queue, we set irq_evt for signaling the guest that buffers were
             // consumed.
             self.vu
-                .set_vring_call(*queue_index, &irq_trigger.irq_evt)
+                .set_vring_call(
+                    *queue_index,
+                    interrupt
+                        .notifier(VirtioInterruptType::Queue(
+                            (*queue_index).try_into().unwrap_or_else(|_| {
+                                panic!("vhost-user: invalid queue index: {}", *queue_index)
```
ShadowCurse, line 451: Same here, please fix this insane conversion code.

src/vmm/src/devices/virtio/vsock/device.rs
```diff
@@ -136,14 +136,16 @@ where
...
+    pub fn signal_used_queue(&self, qidx: usize) -> Result<(), DeviceError> {
         let interrupt = self
             .device_state
             .active_state()
             .expect("Device is not initialized")
             .1;
         interrupt
-            .trigger_irq(IrqType::Vring)
+            .trigger(VirtioInterruptType::Queue(qidx.try_into().unwrap_or_else(
+                |_| panic!("vsock: invalid queue index: {qidx}"),
```
ShadowCurse, line 147: And here

src/vmm/src/devices/virtio/block/virtio/device.rs
Manciukic: this is due to how the guests behave today. If it changes or we allow for deactivating devices, then we may patch a deactivated device, in theory.

src/vmm/src/device_manager/mmio.rs
```diff
@@ -548,7 +552,8 @@ mod tests {
             cmdline: &mut kernel_cmdline::Cmdline,
             dev_id: &str,
         ) -> Result<u64, MmioError> {
-            let mmio_device = MmioTransport::new(guest_mem, device, false);
+            let interrupt = Arc::new(IrqTrigger::new().unwrap());
```
Manciukic, line 555: I was wondering if it made sense to create this directly in the MmioTransport

Manciukic (commented):
I had just a quick look but it looks good to me overall.

src/vmm/src/device_manager/mmio.rs
bchalios: I tried that. But restoring from a snapshot was a nightmare, because we first restore the device state and then we restore the transport. So, I needed the `IrqTrigger` before restoring the device itself.

src/vmm/src/devices/virtio/device.rs
```diff
@@ -35,10 +35,10 @@ impl DeviceState {
...
-    /// Gets the memory attached to the device if it is activated.
-    pub fn mem(&self) -> Option<&GuestMemoryMmap> {
+    /// Gets the memory and interrupt attached to the device if it is activated.
+    pub fn active_state(&self) -> Option<(&GuestMemoryMmap, Arc<IrqTrigger>)> {
         match self {
-            DeviceState::Activated(mem) => Some(mem),
+            DeviceState::Activated((mem, interrupt)) => Some((mem, interrupt.clone())),
             DeviceState::Inactive => None,
         }
     }
```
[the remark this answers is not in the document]
bchalios: How would that work? We would need to "take" the interrupt away from the device.

src/vmm/src/devices/virtio/device.rs
```diff
@@ -23,7 +23,7 @@ use crate::vstate::memory::GuestMemoryMmap;
 #[derive(Debug)]
 pub enum DeviceState {
     Inactive,
-    Activated(GuestMemoryMmap),
+    Activated((GuestMemoryMmap, Arc<IrqTrigger>)),
```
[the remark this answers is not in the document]
bchalios: Yes, I thought about that, don't really have a preference honestly. Can do.

src/vmm/src/devices/virtio/balloon/device.rs
bchalios: If I change `qidx` to be `u16` here I'd need to do the conversion in all calling sites.

src/vmm/src/devices/virtio/block/virtio/device.rs
bchalios: Also, we can update configuration (apparently) even before starting a microVM. I hit that case in one of the integration tests.

src/vmm/src/devices/virtio/device.rs
bchalios: Yup, that's not a bad idea. However, all this re-working is not going to go away. We will need to pass the `VirtioInterrupt` during activation (that's how PCIe devices do it).

src/vmm/src/devices/virtio/net/device.rs
bchalios: Yup, agreed. But I don't think that's part of this PR

src/vmm/src/devices/virtio/transport/mmio.rs
```diff
@@ -375,13 +376,56 @@ pub enum IrqType {
...
+
 /// Helper struct that is responsible for triggering guest IRQs
 #[derive(Debug)]
 pub struct IrqTrigger {
     pub(crate) irq_status: Arc<AtomicU32>,
     pub(crate) irq_evt: EventFd,
 }
 
+impl VirtioInterrupt for IrqTrigger {
+    fn trigger(&self, interrupt_type: super::VirtioInterruptType) -> Result<(), std::io::Error> {
```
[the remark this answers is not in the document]
bchalios: oops. Sorry, vim autocomplete.

src/vmm/src/devices/virtio/transport/mmio.rs
```diff
@@ -993,6 +1037,25 @@ pub(crate) mod tests {
         assert!(d.locked_device().is_activated());
     }
 
+    impl IrqTrigger {
+        pub fn has_pending_irq(&self, irq_type: IrqType) -> bool {
```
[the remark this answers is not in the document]
bchalios: Yup, left over. Will drop it.

src/vmm/src/devices/virtio/transport/mod.rs
```diff
@@ -1,5 +1,39 @@
...
+
+    /// Get the `EventFd` (if any) that backs the underlying interrupt.
+    fn notifier(&self, _interrupt_type: VirtioInterruptType) -> Option<EventFd> {
+        None
+    }
+
+    /// Get the current device interrupt status.
+    fn status(&self) -> Arc<AtomicU32> {
+        Arc::new(AtomicU32::new(0))
+    }
```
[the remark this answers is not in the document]
bchalios: Yup, could do that. I don't see any reason why not.

src/vmm/src/devices/virtio/vhost_user.rs
bchalios: same as before, I think that the solution is fixing the queue indexes at their definitions. If we need to pass them somewhere that expects a `u16` at _some_ point we will need to do a conversion.

src/vmm/src/devices/virtio/balloon/device.rs
ShadowCurse: Ok, what about at least doing a conversion on a separate line and replacing `.unwrap_or_else(|_| panic!(...))` with `expect(&format!("..."))`?

src/vmm/src/devices/virtio/device.rs
ShadowCurse: I am fine with passing it during activation. I am mainly concerned with dynamic dispatch. If we were to pass `&impl VirtioInterrupt` it would be better I think. In any case, this can be discussed later, when first PR with PCIe has arrived.

src/vmm/src/devices/virtio/balloon/device.rs
bchalios: the `expect(&format!("..."))` triggers clippy. That's how I had it at the beginning.

src/vmm/src/device_manager/persist.rs
```diff
@@ -530,14 +535,20 @@ impl<'a> Persist<'a> for MMIODeviceManager {
...
                 &balloon_state.transport_state,
+                interrupt,
                 &balloon_state.device_info,
                 constructor_args.event_manager,
             )?;
         }
 
         for block_state in &state.block_devices {
+            let interrupt =
+                Arc::new(IrqTrigger::new().expect("Could not create interrupt for MMIO device"));
```
roypat, line 546: We unwrap on `IrqTrigger::new()` literally everywhere. Can we instead just unwrap() inside `new()` and have it not return `Result`?

src/vmm/src/devices/virtio/transport/mmio.rs
```diff
@@ -375,13 +376,56 @@ pub enum IrqType {
...
+
+            let irq_status = self.irq_status.load(Ordering::SeqCst);
+            return matches!(
+                (irq_status, interrupt_type.into()),
+                (VIRTIO_MMIO_INT_CONFIG, IrqType::Config) | (VIRTIO_MMIO_INT_VRING, IrqType::Vring)
+            );
+        }
+
+        false
+    }
```
roypat, line 426: Can this just be a free-standing function `has_pending_interrupt<T: VirtioInterrupt>(&T, VirtioInterruptType) -> bool` somewhere inside a test module, so we avoid the `#[cfg(test)]`?

src/vmm/src/devices/virtio/transport/mod.rs
```diff
@@ -1,5 +1,37 @@
...
+#[derive(Debug, Clone)]
+pub enum VirtioInterruptType {
+    /// Interrupt for VirtIO configuration changes
+    Config,
+    /// Interrupts for new events in a queue.
+    Queue(u16),
+}
+
+/// API of interrupt types used by VirtIO devices
+pub trait VirtioInterrupt: std::fmt::Debug + Send + Sync {
```
roypat, line 22: Mh, what is the advantage of this being a trait over it being an enum?

Cargo.lock
```diff
@@ -1618,6 +1624,16 @@ dependencies = [
...
 
+[[package]]
+name = "vm-device"
+version = "0.1.0"
+dependencies = [
+ "anyhow",
+ "serde",
+ "thiserror 2.0.12",
+ "vmm-sys-util",
+]
```
roypat, line 1635: Leftovers from earlier experimentation? :o

src/vmm/src/devices/virtio/block/vhost_user/device.rs
```diff
@@ -311,7 +311,7 @@ impl<T: VhostUserHandleBackend + Send + 'static> VirtioDevice for VhostUserBlock
         &self.queue_evts
     }
 
-    fn interrupt_trigger(&self) -> Arc<IrqTrigger> {
+    fn interrupt_trigger(&self) -> Arc<dyn VirtioInterrupt> {
```
roypat, line 314: I suppose all of these could just return `&dyn VirtioInterrupt`, as none of the callers make use of the `Arc`s as far as I can see. But also out of scope here.

src/vmm/src/devices/virtio/vsock/event_handler.rs
```diff
@@ -182,17 +182,46 @@ where
...
+                        None
+                    }
+                }
+                _ => {
+                    warn!("Unexpected vsock event received: {:?}", source);
+                    None
+                }
+            };
+            if let Some(qidx) = raise_irq {
+                self.signal_used_queue(qidx as usize).unwrap_or_default();
```
roypat, line 224: I don't think this call to unwrap_or_default() is needed

src/vmm/src/devices/virtio/device.rs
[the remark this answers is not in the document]
ShadowCurse: With `as_ref`? Or will you need to mutate the interrupt trigger for PCIe case?

src/vmm/src/devices/virtio/transport/mod.rs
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/devices/virtio/transport/mmio.rs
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/devices/virtio/transport/mmio.rs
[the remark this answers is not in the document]
bchalios: done

src/vmm/src/devices/virtio/device.rs
bchalios: > I don't think we necessarily can, because then we'd have to decide at compile time whether we use MMIO or PCIe, right?

But I think we have two different execution paths. One through a MMIO device manager and one through a PCIe device manager. So both variants would be compiled in. Am I getting this wrong?

src/vmm/src/devices/virtio/device.rs
```diff
@@ -7,23 +7,23 @@
...
+use crate::logger::warn;
 use crate::vstate::memory::GuestMemoryMmap;
 
 /// Enum that indicates if a VirtioDevice is inactive or has been activated
 /// and memory attached to it.
 #[derive(Debug)]
 pub enum DeviceState {
     Inactive,
-    Activated(GuestMemoryMmap),
+    Activated((GuestMemoryMmap, Arc<dyn VirtioInterrupt>)),
```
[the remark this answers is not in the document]
bchalios: Added a struct type that includes them

src/vmm/src/devices/virtio/transport/mod.rs
bchalios: Don't necessarily see an advantage. That's how it's defined on the CH side, so got it as is.
I do have a sort of personal preference to dynamic dispatch when it's not on a performance critical path, because I (visually) hate this:

```Rust
impl VirtioInterrupt {
    pub fn foo(&self) -> Bar {
        match self {
             Irq(irq) => irq.foo()
             Msi(msi) => msi.foo()
        }
    }
}
```

Cargo.lock
bchalios: oh, yes. sorry.

src/vmm/src/devices/virtio/block/vhost_user/device.rs
bchalios: Yup, shall we do this when we bring the MSI code in?

src/vmm/src/devices/virtio/vsock/event_handler.rs
bchalios: the `_or_default()` variant was used already, I assume, to ignore errors. I am happy to change this to something like `expect()`. Is that what you mean?

Cargo.lock
bchalios: hmmmm. Rebuilding seems to bring this in. I've no idea who's pulling it in.

bchalios: Ok, figured it.

src/vmm/src/devices/virtio/vsock/event_handler.rs
roypat: Ah, no I was thinking of just doing
```suggestion
                self.signal_used_queue(qidx as usize);
```
but you're right, clippy would complain about the unused results, so it'd have to be
```suggestion
                _ = self.signal_used_queue(qidx as usize).unwrap_or_default();
```
but iirc we do unwrap() on these in other devices, so that should be fine too?

src/vmm/src/devices/virtio/transport/mod.rs
roypat: Ah fair, I personally actually prefer the enum version, but that's indeed personal preference. I mainly left the comment because I found the explicit Debug bound on a trait that functionally has nothing to do with debugging a bit ugly, but can also stay like this if you prefer it

src/vmm/src/device_manager/persist.rs
bchalios: done

src/vmm/src/devices/virtio/vsock/event_handler.rs
bchalios: Changed it to `expect()`.

src/vmm/src/devices/virtio/transport/mmio.rs
bchalios: hmmm. The problem is that for as long as we support both MMIO and PCIe we will not have an easy way to create an type independent implementation. WDYT?

src/vmm/src/devices/virtio/device.rs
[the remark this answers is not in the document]
bchalios: Changed this to be a struct. Do you think it's still relevant?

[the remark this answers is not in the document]
ShadowCurse: Now it is better. Thx.

src/vmm/src/devices/virtio/block/virtio/device.rs
ShadowCurse: ok

src/vmm/src/devices/virtio/balloon/device.rs
ShadowCurse: As discussed offline, we can fix this later.

bchalios:
I checked the codecov failures. All reported unchecked lines of code were already existing. It's not new code.

src/vmm/src/devices/virtio/vsock/event_handler.rs
```diff
@@ -182,17 +182,47 @@ where
...
+                    }
+                }
+                _ => {
+                    warn!("Unexpected vsock event received: {:?}", source);
+                    None
+                }
+            };
+            if let Some(qidx) = raise_irq {
+                self.signal_used_queue(qidx as usize)
+                    .expect("vsock: Could not trigger device interrupt");
```
roypat, line 225: In the `Self::PROCESS_EVQ` case, why are we settig `raise_irq` to `PROCESS_RXQ`? And if this indeed should be `PROCESS_RXQ`, I think the entire thing could just stay as it is, with the only exception that the call `self.signal_used_queue()` becomes`self.signal_used_queue(source)`, no?

Otherwise, if there's not supposed to be a 1:1 correspondence between `source` and `qidx`, then nit: these match arms can at least be a little bit shorter:

```suggestion
            let raise_irq = match source {
                Self::PROCESS_ACTIVATE => {
                    self.handle_activate_event(ops);
                    None
                }
                Self::PROCESS_RXQ => {
                    self.handle_rxq_event(evset)
                        .then_some(Self::PROCESS_RXQ)
                }
                Self::PROCESS_TXQ => {
                    self.handle_txq_event(evset)
                        .then_some(Self::PROCESS_TXQ)
                }
                Self::PROCESS_EVQ => {
                    self.handle_evq_event(evset)
                        .then_some(Self::PROCESS_RXQ)
                }
                Self::PROCESS_NOTIFY_BACKEND => {
                    self.notify_backend(evset)
                        .then_some(Self::PROCESS_NOTIFY_BACKEND)
                }
                _ => {
                    warn!("Unexpected vsock event received: {:?}", source);
                    None
                }
            };
            if let Some(qidx) = raise_irq {
                self.signal_used_queue(qidx as usize)
                    .expect("vsock: Could not trigger device interrupt");
```

bchalios: super good catch. This was just a typo. We should be passing `Self::PROCESS_EVQ`. The reason why we didn't catch it, is that with MMIO transport the queue ID doesn't really matter.

src/vmm/src/device_manager/persist.rs
```diff
@@ -573,9 +585,12 @@ impl<'a> Persist<'a> for MMIODeviceManager {
         }
 
         for net_state in &state.net_devices {
+            let interrupt =
+                Arc::new(IrqTrigger::new().expect("Could not create interrupt for MMIO device"));
```
ShadowCurse: I saw comment from @roypat above about removing result from an `IrqTrigger::new`. Did you decide not to do it? Or maybe do in a separate PR?

ShadowCurse (approved):

roypat (approved):

PR #5166, opened 2025-04-23
test(perf): add block latency test 

roypat:
fio emits latency metrics regarding how much time was spent inside the
guest operating system (submission latency, slat) or how much time was
spent in the device (clat). For firecracker, the latter could be
relevant, so emit them from the block performance tests.

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

this option only affects the output to stdout, but we are ignoring fio's
stdout (we only work with the log files, which are separate). So drop
this parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@ and this project adheres to
   Clarified what CPU models are supported by each existing CPU template.
   Firecracker exits with an error if a CPU template is used on an unsupported
   CPU model.
+- [#5165](https://github.com/firecracker-microvm/firecracker/pull/5165): Changed
+  Firecracker snapshot feature from developer preview to generally available.
+  Incremental snapshots remain in developer preview.
 
 ### Deprecated
 
--- a/docs/snapshotting/snapshot-support.md
+++ b/docs/snapshotting/snapshot-support.md
@@ -39,12 +39,13 @@ workload at that particular point in time.
 
 ### Supported platforms
 
-> [!WARNING]
->
-> The Firecracker snapshot feature is in
-> [developer preview](../RELEASE_POLICY.md) on all CPU micro-architectures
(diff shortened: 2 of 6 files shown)


kalyazin (dismissed):

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -101,37 +103,51 @@ def run_fio(microvm, mode, block_size):
...
+            metrics.put_metric("bw_read", sum(bw_read), "Kilobytes/Second")
         if bw_write:
-            metrics.put_metric("bw_write", bw_write, "Kilobytes/Second")
+            metrics.put_metric("bw_write", sum(bw_write), "Kilobytes/Second")
+
+    for lat_read, lat_write in process_fio_log_files(f"{logs_dir}/*_clat.*.log"):
+        # latency values in fio logs are in nanosecons, but cloudwatch only supports
+        # microseconds as the more granular unit, so need to divide by 1000.
+        for value in lat_read:
+            metrics.put_metric("clat_read", value / 1000, "Microseconds")
```
Manciukic, line 148: What I've noticed when looking into PCI is that, when using the async fio engine, the latency is much higher than the actual device latency (as we're pushing many requests at the same time asynchronously), and it's also quite unstable, with one fio thread having much higher latency than the other (if you look at bandwidth, one accounts for 80% of the throughput, while the other just 20% as it also has to handle the driver processing and interrupts).
If the goal is to measure the actual device latency, it's best to use the sync fio engine, which is also more representative of the performance of real world readers/writers.

Manciukic (dismissed):

kalyazin (approved):

Manciukic (approved):

PR #5176, opened 2025-04-29
More boot time metrics

ShadowCurse:
## Changes
- Emit `systemd-analyze` output as metrics.
- Emit already collected, but not emitted metrics

## Reason
Better boot time test visibility.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tests/integration_tests/performance/test_boottime.py
+++ b/tests/integration_tests/performance/test_boottime.py
@@ -150,9 +150,11 @@ def test_boottime(
             unit="Microseconds",
         )
 
-        timestamps = find_events(vm.log_data)
-        build_time = timestamps["build microvm for boot"]["duration"]
+        events = find_events(vm.log_data)
+        build_time = events["build microvm for boot"]["duration"]
         metrics.put_metric("build_time", build_time.microseconds, unit="Microseconds")
+        resume_time = events["boot microvm"]["duration"]
+        metrics.put_metric("resume_time", resume_time.microseconds, unit="Microseconds")
 
         kernel, userspace, total = get_systemd_analyze_times(vm)
         metrics.put_metric("systemd_kernel", kernel, unit="Milliseconds")
--- a/tests/integration_tests/performance/test_boottime.py
+++ b/tests/integration_tests/performance/test_boottime.py
@@ -69,6 +69,37 @@ def find_events(log_data):
     return timestamps
 
 
+def get_systemd_analyze_times(microvm):
+    """
+    Parse systemd-analyze output
+    """
+    rc, stdout, stderr = microvm.ssh.run("systemd-analyze")
+    assert rc == 0, stderr
+    assert stderr == ""
+
+    boot_line = stdout.splitlines()[0]
+    # The line will look like this:
+    # Startup finished in 79ms (kernel) + 231ms (userspace) = 310ms
+    # In the regex we capture the time and the unit for kernel, userspace and total values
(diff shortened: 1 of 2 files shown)


tests/integration_tests/performance/test_boottime.py
```diff
@@ -102,14 +120,28 @@ def test_boottime(
...
+        events = find_events(vm.log_data)
+        build_time = events["build microvm for boot"]["duration"]
+        metrics.put_metric("build_time", build_time.microseconds, unit="Microseconds")
+        resume_time = events["boot microvm"]["duration"]
+        metrics.put_metric("resume_time", resume_time.microseconds, unit="Microseconds")
+
+        kernel, userspace, total = get_systemd_analyze_times(vm)
+        metrics.put_metric("systemd_kernel", kernel, unit="Milliseconds")
+        metrics.put_metric("systemd_userspace", userspace, unit="Milliseconds")
+        metrics.put_metric("systemd_total", total, unit="Milliseconds")
```
roypat, line 145: how stable are these wrt A/B-testing? (because they'll be automatically picked up in the A/B pipeline, so if these are volatile, we'll want to put them onto the ignorelist)

tests/integration_tests/performance/test_boottime.py
```diff
@@ -69,6 +69,31 @@ def find_events(log_data):
...
+        if "ms" not in v:
+            return int(v[:-1]) * 1000
+        else:
+            return int(v[:-2])
+
+    kernel = to_ms(kernel)
+    userspace = to_ms(userspace)
+    total = to_ms(total)
+
+    return kernel, userspace, total
```
pb8o: Since these are the only numbers in the line, this works a is a bit more succinct:

```python
return [int(val) * (1000 if unit == 's' else 1) for val, unit in re.findall(r"(\d+)(s|ms)", boot_line)]
```

tests/integration_tests/performance/test_boottime.py
```diff
@@ -69,6 +69,31 @@ def find_events(log_data):
...
+        if "ms" not in v:
+            return int(v.strip()[:-1]) * 1000
+        else:
+            return int(v.strip()[:-2])
+
+    kernel = to_ms(kernel)
+    userspace = to_ms(userspace)
+    total = to_ms(total)
+
+    return kernel, userspace, total
```
pb8o, line 98: Since these are the only numbers in the line, this works a is a bit more succinct:

```python
return [int(val) * (1000 if unit == 's' else 1) for val, unit in re.findall(r"(\d+)(s|ms)", boot_line)]
```

pb8o (dismissed):
Nice!

tests/integration_tests/performance/test_boottime.py
ShadowCurse: I run A/B with last commit and it seems it is stable enough.

tests/integration_tests/performance/test_boottime.py
ShadowCurse: I think current version is a bit more understandable for us Rust heads.

tests/integration_tests/performance/test_boottime.py
ShadowCurse: Actually that was just one lucky run. `resume_time` appears to be more volatile than needed, so added it to ignore list.

roypat (approved):

kalyazin (approved):

PR #5177, opened 2025-04-29
Miscellaneous flaky test fixes

Manciukic:
## Changes
Fixes:
 - signal unit test that was failing due to the async signal delivery (supposedly) by adding a sleep
 - jailer unit test that was crashing because the node was already existing by using a tmp directory
 - tap offload unit test that was failing because the file read was written async by adding a sleep
 - balloon unit test that was failing because the stats weren't refreshed by adding a sleep to ensure they get refreshed

## Reason

I've gone through the recent flaky test failures and this is a proposed fix. In most cases, the fix is to add a time interval to break an async race condition for which we have no better way to resolve.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/src/jailer/src/env.rs
+++ b/src/jailer/src/env.rs
@@ -854,7 +854,7 @@ mod tests {
         arg_vec
     }
 
-    fn create_env(mock_proc_mounts: &str) -> Env {
+    fn create_env(mock_proc_mounts: &Path) -> Env {
         // Create a standard environment.
         let arg_parser = build_arg_parser();
         let mut args = arg_parser.arguments().clone();
@@ -862,7 +862,7 @@ mod tests {
         let pseudo_exec_file_path = get_pseudo_exec_file_path();
         args.parse(&make_args(&ArgVals::new(pseudo_exec_file_path.as_str())))
             .unwrap();
-        Env::new(&args, 0, 0, mock_proc_mounts).unwrap()
+        Env::new(&args, 0, 0, mock_proc_mounts.to_str().unwrap()).unwrap()
     }
 
     #[test]
@@ -876,7 +876,7 @@ mod tests {
         let mut args = arg_parser.arguments().clone();
         args.parse(&make_args(&good_arg_vals)).unwrap();
         // This should be fine.
-        let good_env = Env::new(&args, 0, 0, mock_cgroups.proc_mounts_path.as_str())
+        let good_env = Env::new(&args, 0, 0, mock_cgroups.proc_mounts_path.to_str().unwrap())
             .expect("This new environment should be created successfully.");
 
         let mut chroot_dir = PathBuf::from(good_arg_vals.chroot_base);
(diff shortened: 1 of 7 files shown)


src/vmm/src/signal_handler.rs
```diff
@@ -229,6 +230,9 @@ mod tests {
         });
         child.join().unwrap();
 
+        // signal delivery is asynchronous so wait a little to avoid race conditions
+        thread::sleep(Duration::from_millis(100));
```
roypat, line 234: whew. what a test. Doing non-async-signal safe stuff in a signal handler. Having a signal handler that compiles out the std::process::exit() using `cfg(not(test))` just so that this unit test can exist in the first place. pretty cursed :sob: 

btw, it does seem weird to me that we'd need to sleep after a Thread::join call, to see changes made by the just-joined thread. Did we happen to only see this failure on ARM? Are we sure we're using the correct memory barriers to capture the load/store dependencies?

And then lastly: Can we just replace this entire thing with an integration test?

src/jailer/src/env.rs
```diff
@@ -1138,27 +1138,38 @@ mod tests {
         mock_cgroups.add_v1_mounts().unwrap();
         let env = create_env(mock_cgroups.proc_mounts_path.as_str());
 
+        let mock_dev_dir_name = format!(
+            "/tmp/firecracker/test/{}/jailer/dev",
+            rand::rand_alphanumerics(4).into_string().unwrap()
+        );
+        let mock_dev_dir = Path::new(mock_dev_dir_name.as_str());
```
roypat, line 1145: `vmm_sys_util::TempDir` should do exactly this for you + handle the cleanup

tests/integration_tests/functional/test_net.py
```diff
@@ -118,5 +118,13 @@ def test_tap_offload(uvm_any):
...
     # Check that the server received the message
-    ret = vm.ssh.run(f"sync ; cat {out_filename}")
+    # Allow for some delay due to the asynchronous nature of the test
+    retry = 10
+    while retry > 0:
+        ret = vm.ssh.run(f"sync ; cat {out_filename}")
+        if ret.stdout.strip():
+            break
+        time.sleep(0.1)
+        retry -= 1
```
roypat, line 128:
```suggestion
    for attempt in Retrying(
        stop=stop_after_attempt(10),
        wait=wait_fixed(0.1),
        reraise=True,
    ):
        with attempt: 
            ret = vm.ssh.check_output(f"sync; cat {out_filename}")

src/vmm/src/signal_handler.rs
Manciukic: yeah, it took me a while to figure it out, and I'm not even sure we need the thread here.
We already have integration tests for this lol. I would be ok removing it, but I thought that sleeping might have been less contentious.
There's no guarantee that the signal handlers are completed after we join as the signals are handled asynchronously. Maybe it's the sleep syscall itself makes it check the signals lol

src/jailer/src/env.rs
Manciukic: true, I found out about it after making this commit. I copied this logic from another place in the same file lol
Will fix

tests/integration_tests/functional/test_net.py
Manciukic: love this

roypat (approved):

Issue #2901, opened 2022-02-14
Remove/hook unused metrics


alindima:
We have some metrics that are never incremented, for example: `sync_response_fails` and `sync_vmm_send_timeout_count` (I assume there are more).
We should assess whether it's worth hooking them in the specific Firecracker code or just remove them.

roypat:
There is some ongoing work to redesign our metrics subsystem to allow for per-device metrics by @sudanl0 at the moment, which will also tackle this issue.

LDagnachew:
Not sure if this is still needed, but I'm fine with taking a look at this issue if needed @roypat.

roypat:
> Not sure if this is still needed, but I'm fine with taking a look at this issue if needed [@roypat](https://github.com/roypat).

So there are definitely a ton of metrics that are never incremented/set anywhere in the code, and cleaning that up is still very much wanted I think! It's just a bit tricky to find them, because an IDE won't be able to find them as dead code (generally, they are still formally "used" in constructors and aggregation functions, or random tests), so it's literally a matter of going through each of them and checking whether any of the uses are actually incrementing/setting the metric from production code, and if none exist, deleting it.

mdhaduk:
Hi @roypat, I submitted a draft PR and would love some comments on my work. Feel free to take a look at your own convenience. Thanks!

Note: I'm aware I'm failing tools/devtool checkstyle; I will rebase and fix this before submitting my PR

PR #5182, opened 2025-05-05
[RFC] Remove/hook unused metrics #2901

mdhaduk:
Opening this as a Draft PR

## Changes
Removed unused metrics; used this [script](https://gist.github.com/mdhaduk/97636632806443c0623d77e9d199a831 "script") to generally identify which metrics **might** not be used, then manually removed them 

To identify unused metrics, I used the following convention

For **SharedIncMetric**:

- Not incremented/added to
- Not stored
- Not fetched

For **LatencyAggregateMetrics**:

- Not calling record_latency_metrics

**Metrics Removed**:
- sync_response_fails
- sync_vmm_send_timeout_count
- deprecated_cmd_line_api_calls
- log_fails
- device_events
- rx_partial_writes
- tx_partial_reads

## Reason

Resolving issue #2901 - Remove/hook unused metrics

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ test_results/*
 /resources/linux
 /resources/x86_64
 /resources/aarch64
+/src/check_metrics.py
--- a/src/vmm/src/devices/virtio/net/metrics.rs
+++ b/src/vmm/src/devices/virtio/net/metrics.rs
@@ -161,8 +161,6 @@ pub struct NetDeviceMetrics {
     pub rx_queue_event_count: SharedIncMetric,
     /// Number of events associated with the rate limiter installed on the receiving path.
     pub rx_event_rate_limiter_count: SharedIncMetric,
-    /// Number of RX partial writes to guest.
-    pub rx_partial_writes: SharedIncMetric,
     /// Number of RX rate limiter throttling events.
     pub rx_rate_limiter_throttled: SharedIncMetric,
     /// Number of events received on the associated tap.
@@ -233,8 +231,6 @@ impl NetDeviceMetrics {
             .add(other.rx_queue_event_count.fetch_diff());
         self.rx_event_rate_limiter_count
             .add(other.rx_event_rate_limiter_count.fetch_diff());
-        self.rx_partial_writes
-            .add(other.rx_partial_writes.fetch_diff());
         self.rx_rate_limiter_throttled
             .add(other.rx_rate_limiter_throttled.fetch_diff());
         self.rx_tap_event_count
--- a/src/vmm/src/devices/virtio/net/metrics.rs
+++ b/src/vmm/src/devices/virtio/net/metrics.rs
@@ -189,8 +189,6 @@ pub struct NetDeviceMetrics {
     pub tx_count: SharedIncMetric,
     /// Number of transmitted packets.
     pub tx_packets_count: SharedIncMetric,
-    /// Number of TX partial reads from guest.
-    pub tx_partial_reads: SharedIncMetric,
     /// Number of events associated with the transmitting queue.
     pub tx_queue_event_count: SharedIncMetric,
     /// Number of events associated with the rate limiter installed on the transmitting path.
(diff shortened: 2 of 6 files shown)


roypat:
This is great, thanks for tackling this! I think we can probably even include your script in this PR (somewhere under `tools`) and then run it as part of our regular test suite, so we don't again reintroduce unneeded metrics later on :)

Btw, I think you can catch a few more unused ones with a tiny tweak to your script: Add `metrics.rs` files to the check that excludes specific use-sites. For example, the `rx_partial_writes` metric in net/metrics.rs is unused, but there's a `.add` call due to metrics aggregation inside the metrics files itself that makes your script believe it is in use.

mdhaduk:
Great! Thank you for all the feedback, and I'm super stoked to potentially have my scripted integrated within the test suite!!

I went ahead and made the change to the script as you recommended which you can see above or [here](https://gist.github.com/mdhaduk/97636632806443c0623d77e9d199a831). Additionally, the following metrics may or may not be removed:
- metrics_count (SharedIncMetric)
- metrics_fails (SharedIncMetric)
- missed_metrics_count (SharedIncMetric)

I opted to ask you since removing them would require reworking src/metrics.rs and requests/metrics.rs; if you deem that is okay I will go ahead and make those changes as well!

Thanks!

roypat:
Ahh, I didn't realize we had files called metrics.rs that dont actually contain definition of metrics structs. How inconvenient haha. Those 3 metrics will need to stick around. 

In this case, we can change

```
        or "metrics.rs" in path
```

to

```
        or ("metrics.rs" in path and "vmm" in path)
```

since all metrics definitions are in the "vmm" crate, and requests/metrics.rs is in the "firecracker" crate, and then it won't get these false-positives anymore :)

btw, I think your last push maybe combined changes from multiple branches into this PR? :D

mdhaduk:
Ahhh, I see. Thanks for pointing that out; I made the change, and the script seems to be working as intended. 

And yes, sorry about that, I went ahead and cleaned up the history of this feature branch to only include changes related to the pr.

Let me know if any additional changes are needed, thank you!

roypat:
> Ahhh, I see. Thanks for pointing that out; I made the change, and the script seems to be working as intended.
> 
> And yes, sorry about that, I went ahead and cleaned up the history of this feature branch to only include changes related to the pr.
> 
> Let me know if any additional changes are needed, thank you!

This is great! I think we can squash all the commits about removing individual metrics into a single one. Then, could you include your script under `tests/host_tools/` as part of this PR, and add a very simple test in `tests/integration_tests/style` that just executes the script, failing if any unused metrics are found? :o

mdhaduk:
Sounds good! I will get to that as soon as I finish up some touches on the script! 

So to clarify, if a metric is defined and/or aggregated within the vmm crate, but not used anywhere else (excluding tests), then that metric should be marked unused, right?

I ask because I've added a modification to account for metrics like the following that are improperly being marked as unused: 

`Some((&METRICS.latencies_us.load_snapshot, "load snapshot"))`
            
The modification added `rf".+{field}"` as a part of the usage patterns to account for instances similar to the above example. This solves the issue, and the only remaining metrics are the following:
- process_startup_time_us
- process_startup_time_cpu_us
- min_us
- max_us
- sum_us

All of which are only defined within the "vmm" crate and not used elsewhere in the codebase, so they should be removed, correct? 

Let me know what you think! :D

roypat:
> So to clarify, if a metric is defined and/or aggregated within the vmm crate, but not used anywhere else (excluding tests), then that metric should be marked unused, right?

Yeah, if the only thing a metric is used for is aggregation, or some other logic contained within a metrics.rs file in the vmm crate, then it should be marked as unused, except...

>  This solves the issue, and the only remaining metrics are the following:
> 
>     * process_startup_time_us
> 
>     * process_startup_time_cpu_us
> 
>     * min_us
> 
>     * max_us
> 
>     * sum_us

...these are indeed special .-. The first two only get store()'d in logger/metrics.rs, but it's not actually an aggregation, it's because metrics.rs defines a helper function for setting the metric, and this utility function is then used outside of the metrics module. Similarly the min/max/sum thingies are set in the `Drop` impl for `LatencyMetricsRecorder`. I don't think there's a general rule that would catch these, so let's just have an "allowlist" that causes the script to ignore precisely these 5 metrics.

JackThomson2:
Hey @mdhaduk are you planning to continue working on this and taking it to the finish line?

mdhaduk:
Hey @JackThomson2, yes, I have been meaning to add the integration test to complete this pr, I've just been busy with my current internship; I was thinking of completing this over the weekend, is that alright?

JackThomson2:
Hey @mdhaduk of course, that's great thank you! If you need any help feel free to reach out :)

PR #5191, opened 2025-05-06
Save tests metrics and performance artifacts

ShadowCurse:
## Changes
Save metrics and performance artifacts(logs/json files) we get from test into `test_results` for further processing.
Allow to change the results directory for CI.

Previously the content of `test_results` looked like this:
```
test_results
  - test-report.json 
```
Now it will be:
```
test_results
  - test-report.json
  - test_boottime
    - 'test_boottime[vmlinux-5.10.233-1-128]'
      - metrics.json
      - other test related files
  - ...
```

If the CI is run in BK the `test_results` directory will be archived into `test_results/data.tar.gz`

## Reason
Currently the only output of tests we have are emitted metrics. This means we discard a lot of data we get during test runs (mainly perf tests). This is turn makes it more difficult to analyze performance of Firecracker to the point where we have to use custom scripts to do so. By simply saving all original data, we simplify data gathering and processing step.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -389,7 +389,7 @@ def io_engine(request):
 
 
 @pytest.fixture
-def results_dir(request):
+def results_dir(request, pytestconfig):
     """
     Fixture yielding the path to a directory into which the test can dump its results
 
@@ -411,9 +411,9 @@ def results_dir(request):
     during doc tests, it will return None.
     """
     try:
-        results_dir = (
-            defs.TEST_RESULTS_DIR / request.node.originalname / request.node.name
-        )
+        report_file = pytestconfig.getoption("--json-report-file")
+        parent = Path(report_file).parent.absolute()
+        results_dir = parent / request.node.originalname / request.node.name
     except AttributeError:
         return None
     results_dir.mkdir(parents=True, exist_ok=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -172,7 +172,7 @@ def pytest_runtest_logreport(report):
 
 
 @pytest.fixture()
-def metrics(request):
+def metrics(results_dir, request):
     """Fixture to pass the metrics scope
 
     We use a fixture instead of the @metrics_scope decorator as that conflicts
@@ -188,6 +188,8 @@ def metrics(request):
         metrics_logger.set_property(prop_name, prop_val)
     yield metrics_logger
     metrics_logger.flush()
+    if results_dir:
+        metrics_logger.store_data(results_dir)
 
 
 @pytest.fixture
@@ -391,17 +393,29 @@ def results_dir(request):
     """
     Fixture yielding the path to a directory into which the test can dump its results
 
-    Directories are unique per test, and named after the test name. Everything the tests puts
-    into its directory will to be uploaded to S3. Directory will be placed inside defs.TEST_RESULTS_DIR.
+    Directories are unique per test, and their names include test name and test parameters.
+    Everything the tests puts into its directory will to be uploaded to S3.
+    Directory will be placed inside defs.TEST_RESULTS_DIR.
 
     For example
     ```py
-    def test_my_file(results_dir):
+    @pytest.mark.parametrize("p", ["a", "b"])
+    def test_my_file(p, results_dir):
         (results_dir / "output.txt").write_text("Hello World")
     ```
-    will result in `defs.TEST_RESULTS_DIR`/test_my_file/output.txt.
+    will result in:
+    - `defs.TEST_RESULTS_DIR`/test_my_file/test_my_file[a]/output.txt.
+    - `defs.TEST_RESULTS_DIR`/test_my_file/test_my_file[b]/output.txt.
+
+    When this fixture is called with DoctestItem as a request.node
+    during doc tests, it will return None.
     """
-    results_dir = defs.TEST_RESULTS_DIR / request.node.originalname
+    try:
+        results_dir = (
+            defs.TEST_RESULTS_DIR / request.node.originalname / request.node.name
+        )
+    except AttributeError:
+        return None
     results_dir.mkdir(parents=True, exist_ok=True)
     return results_dir
 
--- a/tests/framework/ab_test.py
+++ b/tests/framework/ab_test.py
@@ -160,7 +160,7 @@ def git_ab_test_host_command(
 
 
 def set_did_not_grow_comparator(
-    set_generator: Callable[[CommandReturn], set]
+    set_generator: Callable[[CommandReturn], set],
 ) -> Callable[[CommandReturn, CommandReturn], bool]:
     """Factory function for comparators to use with git_ab_test_command that converts the command output to sets
     (using the given callable) and then checks that the "B" set is a subset of the "A" set
--- a/tests/host_tools/metrics.py
+++ b/tests/host_tools/metrics.py
@@ -46,52 +46,59 @@ import asyncio
 import json
 import os
 import socket
+from pathlib import Path
 from urllib.parse import urlparse
 
 from aws_embedded_metrics.constants import DEFAULT_NAMESPACE
 from aws_embedded_metrics.logger.metrics_logger_factory import create_metrics_logger
 
 
-class MetricsWrapperDummy:
-    """Send metrics to /dev/null"""
-
-    def set_dimensions(self, *args, **kwargs):
-        """Set dimensions"""
-
-    def put_metric(self, *args, **kwargs):
-        """Put a datapoint with given dimensions"""
-
-    def set_property(self, *args, **kwargs):
-        """Set a property"""
-
-    def flush(self):
-        """Flush any remaining metrics"""
-
-
 class MetricsWrapper:
     """A convenient metrics logger"""
 
     def __init__(self, logger):
+        self.data = {}
         self.logger = logger
 
-    def __getattr__(self, attr):
-        """Dispatch methods to logger instance"""
-        if attr not in self.__dict__:
-            return getattr(self.logger, attr)
-        return getattr(self, attr)
+    def set_dimensions(self, *args, **kwargs):
+        """Set dimensions"""
+        if self.logger:
+            self.logger.set_dimensions(*args, **kwargs)
+
+    def put_metric(self, name, data, unit):
+        """Put a datapoint with given dimensions"""
+        if name not in self.data:
+            self.data[name] = {"unit": unit, "values": []}
+        self.data[name]["values"].append(data)
+
+        if self.logger:
+            self.logger.put_metric(name, data, unit)
+
+    def set_property(self, *args, **kwargs):
+        """Set a property"""
+        if self.logger:
+            self.logger.set_property(*args, **kwargs)
 
     def flush(self):
         """Flush any remaining metrics"""
-        asyncio.run(self.logger.flush())
+        if self.logger:
+            asyncio.run(self.logger.flush())
+
+    def store_data(self, dir_path):
+        """Store data into a file"""
+        metrics_path = Path(dir_path / "metrics.json")
+        with open(metrics_path, "w", encoding="utf-8") as f:
+            json.dump(self.data, f)
 
 
 def get_metrics_logger():
     """Get a new metrics logger object"""
     # if no metrics namespace, don't output metrics
-    if "AWS_EMF_NAMESPACE" not in os.environ:
-        return MetricsWrapperDummy()
-    logger = create_metrics_logger()
-    logger.reset_dimensions(False)
+    if "AWS_EMF_NAMESPACE" in os.environ:
+        logger = create_metrics_logger()
+        logger.reset_dimensions(False)
+    else:
+        logger = None
     return MetricsWrapper(logger)
 
 
--- a/tests/integration_tests/performance/test_block_ab.py
+++ b/tests/integration_tests/performance/test_block_ab.py
@@ -5,7 +5,6 @@
 import concurrent
 import glob
 import os
-import shutil
 from pathlib import Path
 
 import pytest
@@ -45,7 +44,7 @@ def prepare_microvm_for_test(microvm):
     check_output("echo 3 > /proc/sys/vm/drop_caches")
 
 
-def run_fio(microvm, mode, block_size, fio_engine="libaio"):
+def run_fio(microvm, mode, block_size, test_output_dir, fio_engine="libaio"):
     """Run a fio test in the specified mode with block size bs."""
     cmd = (
         CmdBuilder("fio")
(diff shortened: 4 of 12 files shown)


tests/conftest.py
```diff
@@ -396,12 +411,18 @@ def results_dir(request):
 
     For example
     ```py
-    def test_my_file(results_dir):
+    def test_my_file(reggsults_dir):
```
pb8o, line 414: typo introduced?

pb8o (changes requested):

tests/conftest.py
```diff
@@ -171,8 +171,21 @@ def pytest_runtest_logreport(report):
     METRICS.flush()
 
 
+@pytest.fixture
+def test_output_dir(results_dir, request):
```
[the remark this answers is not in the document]
ShadowCurse: fixed

tests/conftest.py
roypat: :egg:

tests/framework/ab_test.py
roypat, line 1: huh, why didnt the CI black step catch this?

tests/host_tools/metrics.py
```diff
@@ -46,52 +46,59 @@
...
 from aws_embedded_metrics.logger.metrics_logger_factory import create_metrics_logger
 
 
-class MetricsWrapperDummy:
-    """Send metrics to /dev/null"""
+class MetricsWrapper:
+    """A convenient metrics logger"""
+
+    def __init__(self, logger):
+        self.data = {}
```
roypat, line 60: nit: could be `defaultdict(list)`, which'll save you some "if name not in self.data" statements

tests/host_tools/metrics.py
```diff
@@ -46,52 +46,59 @@
...
         """Set dimensions"""
+        if self.logger:
+            self.logger.set_dimensions(*args, **kwargs)
 
-    def put_metric(self, *args, **kwargs):
+    def put_metric(self, name, data, unit):
         """Put a datapoint with given dimensions"""
+        if name not in self.data:
+            self.data[name] = []
+        self.data[name].append(data)
```
roypat, line 72: mhh, would be really cool if we could save this in the same format that `tools/ab_test.py analyze` expects, so make it really easy to do A/B-tests across arbitrary runs.

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -93,21 +86,25 @@ def run_fio(microvm, mode, block_size, fio_engine="libaio"):
...
 
 
-def process_fio_log_files(logs_glob):
+def process_fio_log_files(root_dir, logs_glob):
     """Parses all fio log files matching the given glob and yields tuples of same-timestamp read and write metrics"""
     data = [
-        Path(pathname).read_text("UTF-8").splitlines()
-        for pathname in glob.glob(logs_glob)
+        Path(root_dir / pathname).read_text("UTF-8").splitlines()
+        for pathname in glob.glob(logs_glob, root_dir=root_dir)
```
roypat, line 107: I admittedly don't quite see why this change to this function is needed?

tests/integration_tests/performance/test_block_ab.py
```diff
@@ -213,6 +211,7 @@ def test_block_vhost_user_performance(
     fio_mode,
     fio_block_size,
     metrics,
+    test_output_dir,
```
roypat, line 214: wrong fixture name?

tests/integration_tests/performance/test_network_ab.py
```diff
@@ -133,4 +136,16 @@ def test_network_tcp_throughput(
     )
     data = test.run_test(network_microvm.vcpus_count + 2)
 
+    cpu_load_path = Path(results_dir / "cpu_load.json")
+    with open(cpu_load_path, "w", encoding="utf-8") as f:
```
roypat, line 140: nit(for all of these): `cpu_load_path.write_text(json.dumps(data["cpu_load_raw"])`

also, why split everything across so many files? This could just be a single json file into which all of `data` is dumped. That way if we ever change the structure of the `data` dict, we dont have to worry about also forgetting to update this. 

But actually, since we already updated the metrics thingy to always dump all emitted metrics into test_result, why do we need to put the same data into separate json files again in the first place?

tests/integration_tests/performance/test_vsock_ab.py
```diff
@@ -107,4 +116,16 @@ def test_vsock_throughput(
     test = VsockIPerf3Test(vm, mode, payload_length)
     data = test.run_test(vm.vcpus_count + 2)
 
+    cpu_load_path = Path(results_dir / "cpu_load.json")
```
roypat, line 119: same as above

tools/ab_test.py
```diff
@@ -195,9 +195,12 @@ def collect_data(binary_dir: Path, pytest_opts: str):
         check=True,
         shell=True,
     )
-    return load_data_series(
-        Path("test_results/test-report.json"), binary_dir, reemit=True
-    )
+
+    # move results into a tag directory
+    os.mkdir(tag)
+    os.rename("test_results", f"{tag}")
```
roypat, line 201:
```suggestion
    os.rename("test_results", tag)
```

tools/ab_test.py
```diff
@@ -180,7 +180,7 @@ def uninteresting_dimensions(processed_emf):
     return uninteresting
 
 
-def collect_data(binary_dir: Path, pytest_opts: str):
+def collect_data(tag: str, binary_dir: Path, pytest_opts: str):
```
roypat, line 183: iirc, the location of the json files can be modified via pytest's `--json-report-file` option. I think it'll be cleaner to use this, and modify our test framework to dynamically determine `TEST_ARTIFACTS_DIR` to be the parent of the file passed via --json-report-file.

tests/integration_tests/performance/test_block_ab.py
roypat: Ah, fair. Let's put that in the commit message/a comment, otherwise I can see it getting "simplified" again in the future by someone not aware of the context

roypat: (and let's also validate this assumption is actually true before changing the code)

tests/integration_tests/performance/test_network_ab.py
ShadowCurse: Will remove the cpu usage files for now as it indeed will copy what is in the metrics anyway.
Regarding other files, this is because I want to save all original iperf files we have, so we will have access to all data we currently do not emit as metrics.
Also in general I would prefer for us to remove all metric emitting from tests and do that as a separate post step which will look at the original data we saved and retrieve whatever metrics it needs. The metric retrieval can also be adopted by A/B framework later.

tests/host_tools/metrics.py
ShadowCurse: I forgot to add the `unit` to the metrics, so now the logic of append is a bit more comlex

tests/integration_tests/performance/test_block_ab.py
ShadowCurse: fixed

tests/host_tools/metrics.py
ShadowCurse: Currently it produces `frozenset` like this:
```
{
  ({('host_kernel', 'linux-6.1'), ('guest_kernel', 'linux-5.10'), ('cpu_model', 'ARM_NEOVERSE_V2'), ('instance', 'm8g.metal-24xl'), ('mem_size_mib', '128'), ('host_os', 'amzn2023'), ('performance_test', 'test_boottime'), ('vcpus', '1')}):
   {
     'guest_boot_time': ([58695, 58646], 'Microseconds'),
     'guest_cpu_boot_time': ([46399, 46562], 'Microseconds'),
     ...
   }
}
```
while `metrics.json` is:
```json
{"guest_boot_time": {"unit": "Microseconds", "values": [59117, 58180]}, "guest_cpu_boot_time": {"unit": "Microseconds", "values": [46792, 45805]}, "build_time": {"unit": "Microseconds", "values": [5568, 4685]}, "resume_time": {"unit": "Microseconds", "values": [41, 37]}, "systemd_kernel": {"unit": "Milliseconds", "values": [47.0, 47.0]}, "systemd_userspace": {"unit": "Milliseconds", "values": [373.0, 373.0]}, "systemd_total": {"unit": "Milliseconds", "values": [420.0, 420.0]}}
```

So I would prefer to change A/B script to read the instance info (host kernel/instane) from `test-report.json` and then read `metrics.json` for each directory/test and do a comparison like this.

tools/ab_test.py
ShadowCurse: fixed

tests/integration_tests/performance/test_vsock_ab.py
ShadowCurse: removed

tests/integration_tests/performance/test_block_ab.py
ShadowCurse: checked that this is indeed the case and added a comment

tests/host_tools/metrics.py
roypat: Well, currently the A/B script expects EMF messages that it parses from test-report.json, the frozenset stuff is just its internal representation into which it transforms the EMF stuff (which, yeah, I agree ain't great). We can change the EMF part and instead have the A/B-script consume a custom data format, but there's a couple problems I see with that: 
- Keeping track of the dimensions will become a bit of a pain (you mention that we'll need to grab the instance and such from test-report.json, sure, but we also additionally need to parse the directory name to extract the rest of them - so they'll be split across 3 different places, instead of just being a single file)
- We'll need to then have logic to convert from the custom metrics format to EMF to emit to cloudwatch now, and this need to happen for literally every pipeline separately if you want it to happen outside of the test framework

I think an much easier way to achieve what you want is to simply have two metrics loggers, one that is set with `AWS_EMF_ENVIRONMENT=local AWS_EMF_NAMESPACE=local` set, and one that always published to the UDP socket for the cloudwatch daemon. Our MetricsWrapper thingy then simply forwards all put_metrics calls to both of these. Then everything will automatically be contained in test-report.json always (which will also simplify the A/B-testing script, for one it won't need to re-emit metrics anymore, and it also won't need to set env variables).

tests/integration_tests/performance/test_block_ab.py
roypat: thanks!

tests/host_tools/metrics.py
ShadowCurse: > we also additionally need to parse the directory name to extract the rest of them

Do we really need all the dimensions though? What if we just use directory name as a unique test name (as it is already unique)? And A/B will compare data for same dir name?

roypat: because the A/B-testing script uses all the dimensions to do cross-test error correction, but looking at the correlation in regressions between tests with "similar" dimensions

tests/framework/ab_test.py
ShadowCurse: idk. these files were always formatted when I run black locally, and it annoyed me, so I added the change.

tests/host_tools/metrics.py
roypat: yeah fair

tests/framework/ab_test.py
```diff
@@ -113,8 +113,8 @@ def binary_ab_test(
     Similar to `git_ab_test`, but instead of locally checking out different revisions, it operates on
     directories containing firecracker/jailer binaries
     """
-    result_a = test_runner(a_directory, True)
-    result_b = test_runner(b_directory, False)
+    result_a = test_runner("A", a_directory, True)
+    result_b = test_runner("B", b_directory, False)
```
roypat, line 117: This additional parameter shouldn't be needed, since we already have the boolean

tools/ab_test.py
```diff
@@ -346,7 +350,7 @@ def ab_performance_test(
     """Does an A/B-test of the specified test with the given firecracker/jailer binaries"""
 
     return binary_ab_test(
-        lambda bin_dir, _: collect_data(bin_dir, pytest_opts),
+        lambda tag, bin_dir, _: collect_data(tag, bin_dir, pytest_opts),
```
roypat, line 353:
```suggestion
        lambda bin_dir, is_a: collect_data(is_a and 'A' or 'B', bin_dir, pytest_opts),
```

tests/framework/ab_test.py
ShadowCurse: fixef

tools/ab_test.py
ShadowCurse: fixed

roypat (approved):
validated everything still works by running a dummy A/B test

tests/conftest.py
```diff
@@ -391,17 +393,29 @@ def results_dir(request):
...
+    - `defs.TEST_RESULTS_DIR`/test_my_file/test_my_file[a]/output.txt.
+    - `defs.TEST_RESULTS_DIR`/test_my_file/test_my_file[b]/output.txt.
+
+    When this fixture is called with DoctestItem as a request.node
+    during doc tests, it will return None.
     """
-    results_dir = defs.TEST_RESULTS_DIR / request.node.originalname
+    try:
+        results_dir = (
+            defs.TEST_RESULTS_DIR / request.node.originalname / request.node.name
```
pb8o: Have you tried it with the [cross-snapshot restore test](https://github.com/firecracker-microvm/firecracker/blob/main/tests/integration_tests/functional/test_snapshot_restore_cross_kernel.py#L73-L84)? does it need some changes?

ShadowCurse: I tried to run this one, but the first phase fails with `find: ‘test_results/test_snapshot_phase1’: No such file or directory` which is quite strange since this dir does exist. If I run it locally, it all works just fine. Any ideas why?

ShadowCurse: ok, I fixed the issue. It was because now the `test_results` are compressed by default, so needed to add an option to not compress the dir for snapshot tests.

roypat (approved):

tools/devtool
```diff
@@ -802,6 +806,14 @@ cmd_test() {
     # do not leave behind env.list file
     rm env.list
 
+    # archive everything in the `test_result` to speed up upload/download
+    # to s3 if we are in CI
+    if [ $do_archive != 0 ] && [ -n "$BUILDKITE" ] && [ "$BUILDKITE" = "true" ]; then
+      tar -czf data.tar -C test_results .
```
kalyazin, line 812: I thought compressed tars were usually called ".tar.gz". Could less confusing for manual processing.

roypat (dismissed):

roypat (approved):

PR #5196, opened 2025-05-07
Fixing VMM & RNG per-device metrics. (Minor Tap fix)

### LINKED ISSUE #4709: Allow Running `vmm` unittests in parallel
Currently, the unittests in the `vmm` package have to be run sequentially due to reliance on global state in some virtio device tests. More specifically, various tests in the `vsock` and `rng` modules use a global `METRICS` object (see also https://github.com/firecracker-microvm/firecracker/issues/2899). This design using a global METRICS object is fine for production, because a VMM can have at most one vsock and rng device each, however multiple unit tests running in parallel can create multiple device objects, which then simultaneously update the global metrics. Since our tests contain assertions on the value of the metrics to validate functionality, this causes spurious failures. 

We can fix this by moving the vsock and rng device to a "per device metrics" model, as already adopted by net and block devices (see for example https://github.com/firecracker-microvm/firecracker/pull/4145). 

As part of this issue, 3 things will need to be done
- [ ] Update the vsock device to use a "per device metrics" model
- [ ] Update the rng device to use a "per device metrics" model
- [ ] Fix the `test_tap_name` test to no longer use a hardcoded value of `tap0` (if multiple tests run in parallel and create tap devices, it is no longer guaranteed that the tap device created in this test will be called "tap0", it could be "tap1" or "tap43". Thus we probably just want some sort of regex match on the tap device name).

These can all be addressed in individual PRs, or all in one, depending on how much of a task you want to tackle :)

bstrong04:
## Changes

Added a new device metric system for both RNG and VMM VIRTIO devices. This includes mapping them through a BTreeMap as mentioned in issue #4145 for the net devices. Handles multiple devices through ID matching while also adding an Arc to each metric to make them thread safe. Also adds some minor tests replicating the ones in the net devices folder to check for thread safety and aggregation functionality. 

Also resolves all former references to metrics within the vsock and rng objects to a reference to the "global" id set to arbitrarily point to the single required vsock and rng devices for their respective files.

Also adds a minor fix to the test_tap_name method allowing for regex matching.

## Reason

Done to allow unit tests to run in parallel to each other. In production one vsock device and one rng device suffice for use, but with this implementation device metrics will overwrite each other during testing and tests would have to be run sequentially. Therefore, support for a global metric and multiple other metrics during testing is needed to ensure functionality for tests and production. This behavior is simply replicated on both vsock and rng.

test_tap_name changed from prior hardcoded value of tap0 to handle any numerical value after 'tap' when passed in.

Closes #4709 

Worked with @gjkeller for this.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/src/vmm/src/devices/virtio/net/tap.rs
+++ b/src/vmm/src/devices/virtio/net/tap.rs
@@ -241,10 +241,17 @@ pub mod tests {
             generated::ifreq__bindgen_ty_1::default().ifrn_name.len()
         });
 
-        // Empty name - The tap should be named "tap0" by default
+        // Empty name - The tap should be named by the kernel (e.g., "tap0", "tap1", etc.)
         let tap = Tap::open_named("").unwrap();
-        assert_eq!(b"tap0\0\0\0\0\0\0\0\0\0\0\0\0", &tap.if_name);
-        assert_eq!("tap0", tap.if_name_as_str());
+        let tap_name_str = tap.if_name_as_str();
+
+        // Check that it starts with "tap" and the remainder is numeric.
+        assert!(
+            Regex::new(r"^tap\d+$").unwrap().is_match(tap_name_str),
+            "Generated tap name '{}' does not match expected pattern",
+            tap_name_str
+        );
+
 
         // Test using '%d' to have the kernel assign an unused name,
         // and that we correctly copy back that generated name
--- a/src/vmm/src/devices/virtio/rng/device.rs
+++ b/src/vmm/src/devices/virtio/rng/device.rs
@@ -9,7 +9,7 @@ use aws_lc_rs::rand;
 use vm_memory::GuestMemoryError;
 use vmm_sys_util::eventfd::EventFd;
 
-use super::metrics::METRICS;
+use super::metrics::EntropyMetricsPerDevice;
 use super::{RNG_NUM_QUEUES, RNG_QUEUE};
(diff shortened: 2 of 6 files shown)


kalyazin (commented):
Hi @bstrong04 . Thanks for your contribution!

I left some inline comments regarding the approach.
I also triggered a CI run: https://buildkite.com/firecracker/firecracker-pr/builds/13383#0196b489-4d01-4b39-b984-2e3a363e5745 . It reports some style errors. You can run those yourself locally (please see https://github.com/firecracker-microvm/firecracker/blob/main/tests/README.md).

> Worked with @gjkeller for this.

I suggest you split your change in multiple commits, at least the tap test change and rng/vsock changes. You could tag @gjkeller appropriately in those commits via Co-Developed-by, Tested-by, etc.

JackThomson2:
Hey @bstrong04 are you planning to continue working on this and taking it to the finish line?

kalyazin:
@bstrong04 we're closing this due to inactivity. Please feel free to reopen if you have time to work on it again.

shashforge:
Hi everyone! 🙏
If nobody is currently working on this, I’d be grateful for the opportunity to pick it up. Please let me know if it’s already spoken for—otherwise I’ll start drafting a plan. Thanks!

PR #5222, opened 2025-05-22
fix `IovDeque` for non 4K pages

### LINKED ISSUE #5217: [Bug] Regression v1.10.0 tap device unreliable and unresponsive
# Describe the bug

A recent Firecracker version seems to have introduced a regression.

The connection via tap device becomes quickly unreliable and unresponsive after boot (however, it works fine for a short time). I am just following the getting started guide, nothing fancy. I am on ARM64.

## To Reproduce

`[Author TODO: Steps to reproduce the behaviour:]`

1. Start Firecracker via `sudo ./firecracker-v1.11.0-aarch64 --config-file firecracker.conf --api-sock ./firecracker.socket`
1. Configure Firecracker via

```
{
  "boot-source": {
    "kernel_image_path": "vmlinux-6.1.128",
    "boot_args": "keep_bootcon console=ttyS0 reboot=k panic=1 pci=off"
  },
  "drives": [
    {
      "drive_id": "rootfs",
      "is_root_device": true,
      "is_read_only": false,
      "path_on_host": "ubuntu-24.04.ext4"
    }
  ],
  "machine-config": null,
  "cpu-config": null,
  "balloon": null,
  "network-interfaces": [
    {
      "iface_id": "net1",
      "guest_mac": "06:00:AC:10:00:02",
      "host_dev_name": "tap0"
    }
  ],
  "vsock": null,
  "logger": {
    "log_path": "./firecracker.log",
    "level": "Debug",
    "show_level": true,
    "show_log_origin": true
  },
  "metrics": null,
  "mmds-config": null,
  "entropy": null
}
```
3. Connect via SSH, it will fail after a few attempts

For some reason this config does not parse with v1.10.0. However I am pretty sure this version introduced the bug since it mentions changes with virtio-net.

The bug is not reproducible with v1.9.1

```
sudo ./firecracker-v1.9.1-aarch64 --config-file firecracker.conf --api-sock ./firecracker.socket
```

## Expected behaviour

It remains possible to ssh into the device.

```
ping 172.16.0.2
```

should keep pinging, instead packets get delayed, dropped, etc.

## Environment

- Firecracker version: v1.10.0
- Host and guest kernel versions: host: 6.14.6-400.asahi.fc41.aarch64+16k, guest: 6.1.128
- Rootfs used: the one from the getting started guide
- Architecture: arm64
- Any other relevant software versions: N/A

## Additional context

Not much just trying out Firecracker

Just run the thing

Probably the virtio changes are related

Note that Ashai Linux uses 16K pages, which may have something to do with it

## Checks

According to @Manciukic this has not been reported yet.

ShadowCurse:
## Changes
The `L` const generic was determining the maximum number of `iov`
elements in the `IovDeque`. This cases the issue when the host kernel
uses pages which can contain more entries than `L`. For example usual
4K pages can contain 256 `iov`s while 16K pages can contain 1024 `iov`s.
Current implementation on 16K (and any other bigger than 4K page size)
will continue wrap `IovDeque` when it reaches 256'th element. This
breaks the implementation since elements written past 256'th index will
not be 'duplicated' at the beginning of the queue.

Curren implementation expects this behavior:
```
 page 1 page 2
|ABCD|#|ABCD|
      ^ will wrap here
```

With big page sizes current impl will:
```
 page 1              page2
|ABCD|EFGD________|#|ABCDEFGD________|
     ^ sill wrap here
                   ^ but should wrap here
```

The solution is to calculate the maximum capacity the `IovDeque` can
hold, and use it for wrapping purposes. This capacity is allowed to be
bigger than `L`. The actual used number of entries in the queue will
still be guarded by the `L` parameter used in the `is_full` method.

## Reason
Fixes #5217

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/docs/kernel-policy.md
+++ b/docs/kernel-policy.md
@@ -9,29 +9,29 @@ We are continuously validating the currently supported Firecracker releases (as
 per [Firecracker’s release policy](../docs/RELEASE_POLICY.md)) using a
 combination of all supported host and guest kernel versions in the table below.
 
-While other versions and other kernel configs might work, they are not
-periodically validated in our test suite, and using them might result in
-unexpected behaviour. Starting with release `v1.0` each major and minor release
-will specify the supported kernel versions.
-
 Once a kernel version is officially added, it is supported for a **minimum of 2
 years**. At least 2 major guest and host versions will be supported at any time.
 When support is added for a third kernel version, the oldest will be deprecated
 and removed in a following release, after its minimum end of support date.
 
+**Note** While other versions and other kernel configs might work, they are not
+periodically validated in our test suite, and using them might result in
+unexpected behaviour. Starting with release `v1.0` each major and minor release
+will specify the supported kernel versions.
+
 ### Host Kernel
 
-| Host kernel | Min. version | Min. end of support |
-| ----------: | -----------: | ------------------: |
-|       v5.10 |       v1.0.0 |          2024-01-31 |
-|        v6.1 |       v1.5.0 |          2025-10-12 |
+| Page size | Host kernel | Min. version | Min. end of support |
+| --------: | ----------: | -----------: | ------------------: |
+|        4K |       v5.10 |       v1.0.0 |          2024-01-31 |
+|        4K |        v6.1 |       v1.5.0 |          2025-10-12 |
 
 ### Guest Kernel
 
-| Guest kernel | Min. version | Min. end of support |
-| -----------: | -----------: | ------------------: |
-|        v5.10 |       v1.0.0 |          2024-01-31 |
-|         v6.1 |       v1.9.0 |          2026-09-02 |
+| Page size | Guest kernel | Min. version | Min. end of support |
+| --------: | -----------: | -----------: | ------------------: |
+|        4K |        v5.10 |       v1.0.0 |          2024-01-31 |
+|        4K |         v6.1 |       v1.9.0 |          2026-09-02 |
 
 The guest kernel configs used in our validation pipelines can be found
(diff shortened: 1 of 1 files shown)


src/vmm/src/devices/virtio/iov_deque.rs
Manciukic, line 1: could you to add a unit test to verify the behaviour when `L < capacity`? For example, using `L=64` and testing a scenario where it was failing before this change

louwers:
Confirmed that this fix seems to solve the problem I reported.

Manciukic (dismissed):

Manciukic:
Build is complaining about some markdown formatting
```
FAILED integration_tests/style/test_markdown.py::test_markdown_style - AssertionError: Some markdown files need formatting. Either run `./tools/devtool sh mdformat .` in the repository root, or apply the above diffs manually.
```

Manciukic (dismissed):

CHANGELOG.md
```diff
@@ -29,6 +29,10 @@ and this project adheres to
 
 ### Fixed
 
+- [#5222](https://github.com/firecracker-microvm/firecracker/pull/5222): Fixed
+  `IovDeque` wraparound logic which prevented it from working properly on
+  systems with non 4K pages.
```
roypat, line 34: Might be more interesting to say how this bug would appear for a customer, instead of the technical detail we fixed. E.g. something like "fixed a bug affecting systems with non 4K pages where network devices stop working after 256 packets were exchanged"?

roypat (approved):

PR #5233, opened 2025-05-29
test: dump debug information if http request to firecracker fails

roypat:
Sometimes, we intermittently see `ConnectionRefused` errors when doing
http requests to the firecracker API in integration tests. Have the test
framework dump relevant logs in these cases.

...

## Reason

...

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/tests/framework/http_api.py
+++ b/tests/framework/http_api.py
@@ -58,7 +58,12 @@ class Resource:
     def get(self):
         """Make a GET request"""
         url = self._api.endpoint + self.resource
-        res = self._api.session.get(url)
+        try:
+            res = self._api.session.get(url)
+        except Exception as e:
+            if self._api.error_callback:
+                self._api.error_callback("GET", self.resource, str(e))
+            raise
         assert res.status_code == HTTPStatus.OK, res.json()
         return res
 
@@ -66,7 +71,12 @@ class Resource:
         """Make an HTTP request"""
         kwargs = {key: val for key, val in kwargs.items() if val is not None}
         url = self._api.endpoint + path
-        res = self._api.session.request(method, url, json=kwargs)
+        try:
+            res = self._api.session.request(method, url, json=kwargs)
+        except Exception as e:
+            if self._api.error_callback:
+                self._api.error_callback(method, path, str(e))
+            raise
         if res.status_code != HTTPStatus.NO_CONTENT:
             json = res.json()
             msg = res.content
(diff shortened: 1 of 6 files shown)


tests/framework/http_api.py
```diff
@@ -58,15 +58,25 @@ def __init__(self, api, resource, id_field=None):
     def get(self):
         """Make a GET request"""
         url = self._api.endpoint + self.resource
-        res = self._api.session.get(url)
+        try:
+            res = self._api.session.get(url)
+        except Exception as e:
+            if self._api.error_callback:
+                self._api.error_callback("GET", self.resource, str(e))
+            raise
```
ShadowCurse, line 66: maybe move `error_callback` into `Api` class and add exception catching there so it will apply to all api calls?

tests/framework/microvm.py
```diff
@@ -1101,10 +1101,12 @@ def thread_backtraces(self):
...
         for thread_name, thread_pids in utils.get_threads(self.firecracker_pid).items():
             for pid in thread_pids:
-                backtraces.append(
-                    f"{thread_name} ({pid=}):\n"
-                    f"{utils.check_output(f'cat /proc/{pid}/stack').stdout}"
-                )
+                try:
+                    stack = Path(f"/proc/{pid}/stack").read_text("UTF-8")
+                except FileNotFoundError:
+                    continue  # process might've gone away between get_threads() call and here
```
ShadowCurse, line 1119: Maybe do `.exists()` call to check instead of exception catching?

ShadowCurse (approved):

tests/framework/http_api.py
pb8o, line 66: You could avoid all the duplication by delegating to `request`:
```
res = self.request("GET", url)
```

This was not done initially because it saved nothing, but now you can save the error handling duplication.

roypat: So I wanted to do that at first, too, but request() and get() have different error conditions. request() asserts that we get 204 NO CONTENT (unless we have an error response), while get() asserts we get 200 OK with a json body. So it's not that straightforward I think?

pb8o: Oh right I did not see that. You could move that check to request and then dispatch on the method `if GET and not 200, else ...`

But I will leave that up to you.

pb8o (approved):

PR #5240, opened 2025-06-02
Add support for virtio-pci transport

bchalios:
## Changes

This PR adds support for using a PCI transport for VirtIO devices. The PR adds PCI support for x86 and Aarch64 systems both for booted and snapshot-restored microVMs. The VirtIO-PCI transport is used whenever Firecracker is launched  with PCI enabled. Otherwise, we fall back to the MMIO transport.

This PR also adds support for MSI-X interrupts for VirtIO devices. When using the PCI transport VirtIO devices always use MSI-X type of interrupts. MMIO devices still use the IRQ based ones.

Since we needed to change various things in order to support PCI along with MMIO transport the PR brings a few quality of life changes in the code base:

* PCI devices allow that the guest moves the MMIO ranges allocated to a PCI device. This operation requires access to the `ResourceAllocator`, so that we free old memory and allocate new one, the MMIO and Port IO buses, so that we register the device to its new allocated address range and the KVM Vm file descriptor so that we correctly set the new IO event fds. As a result, we move buses, and resource allocator inside the `Vm` object and make the latter responsible for the relocation when needed. This moves simplify the signatures of a lot of functions that were using a combination of `ResourceAllocator`, `Vm` and `Bus` objects. In the future, we might want to even move `DeviceManager` inside the `Vm` object. This should simplify things even further.
* Clean-up the MMIO restore logic. There was a lot of repetition for every device type we support. We now keep a generic type for (re)storing the state of VirtIO devices. In the future, we might even be able to further generalize, so that the restoration logic is common across MMIO and PCI devices if needed.
* We make use of the `serde` support in the new version of `vm-allocator` crate, so that we don't have to perform allocations with fixed addresses upon snapshot restore.

## Reason

Support VirtIO devices with PCI transport

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.


src/vmm/src/builder.rs
```diff
@@ -191,19 +165,26 @@ pub fn build_microvm_for_boot(
...
-        cpu_template.kvm_capabilities.clone(),
-    )?;
+    /*
+        let (mut vmm, mut vcpus) = create_vmm_and_vcpus(
+            instance_info,
+            event_manager,
+            vm_resources.machine_config.vcpu_count,
+            cpu_template.kvm_capabilities.clone(),
+        )?;
+    */
```
ShadowCurse: remaining comments

src/vmm/src/vstate/vm.rs
```diff
@@ -52,6 +56,151 @@ pub struct RoutingEntry {
...
+            event_fd: EventFd::new(libc::EFD_NONBLOCK).map_err(InterruptError::EventFd)?,
+            enabled: AtomicBool::new(enabled),
+        })
+    }
+}
+
+impl MsiVector {
+    /// Enable vector
+    fn enable(&self, vmfd: &VmFd) -> Result<(), errno::Error> {
+        if !self.enabled.load(Ordering::Acquire) {
```
roypat, line 89: should this be an `compare_exchange` (with a rollback if register_irqfd fails, although I guess we dont really expect it to fail) instead to prevent the function from racing with itself? Can it even race with itself?

src/vmm/src/vstate/vm.rs
```diff
@@ -52,6 +56,151 @@ pub struct RoutingEntry {
...
+impl MsiVectorGroup {
+    /// Returns the number of vectors in this group
+    pub fn num_vectors(&self) -> u16 {
+        u16::try_from(self.irq_routes.len()).unwrap()
+    }
+}
+
+impl InterruptSourceGroup for MsiVectorGroup {
+    fn enable(&self) -> vm_device::interrupt::Result<()> {
+        for (_, route) in self.irq_routes.iter() {
```
roypat: nit:

```suggestion
        for route in self.irq_routes.values() {
```

(there's a few more cases of this)

src/vmm/src/vstate/vm.rs
```diff
@@ -52,6 +56,151 @@ pub struct RoutingEntry {
...
+        }
+
+        Ok(())
+    }
+
+    fn trigger(
+        &self,
+        index: vm_device::interrupt::InterruptIndex,
+    ) -> vm_device::interrupt::Result<()> {
+        if let Some(route) = self.irq_routes.get(&index) {
```
roypat: nit: reuse the `notifier` function from below here maybe?

src/vmm/src/vstate/vm.rs
```diff
@@ -339,10 +487,97 @@ impl Vm {
...
+            type_: KVM_IRQ_ROUTING_MSI,
+            ..Default::default()
+        };
+        entry.u.msi.address_lo = config.low_addr;
+        entry.u.msi.address_hi = config.high_addr;
+        entry.u.msi.data = config.data;
+
+        if self.common.fd.check_extension(kvm_ioctls::Cap::MsiDevid) {
+            // On AArch64, there is limitation on the range of the 'devid',
+            // it cannot be greater than 65536 (the max of u16).
```
roypat: 65535?

src/vmm/src/vstate/vm.rs
```diff
@@ -339,10 +487,97 @@ impl Vm {
...
+        resource_allocator: &ResourceAllocator,
+        base: u32,
+        count: u16,
+    ) -> Result<MsiVectorGroup, InterruptError> {
+        debug!("Creating new MSI group with {count} vectors");
+        let mut irq_routes = HashMap::with_capacity(count as usize);
+        for (i, gsi) in resource_allocator
+            .allocate_gsi(count as u32)?
+            .iter()
+            .enumerate()
```
roypat: .zip(0u32..) here instead of enumerate() lets you avoid the cast below

src/vmm/src/vstate/vm.rs
```diff
@@ -52,6 +56,151 @@ pub struct RoutingEntry {
...
+            // (KVM: SVM: fix panic on out-of-bounds guest IRQ).
+            if !masked {
+                route.enable(&self.vm.common.fd)?;
+            }
+
+            return Ok(());
+        }
+
+        Err(std::io::Error::other(format!(
+            "MSI-X update: invalid vector index {index}"
```
roypat, line 227: as an aside (irrelevant to this PR because its vendored code): I really dislike this error handling using std::io::Error::other

src/vmm/src/devices/virtio/transport/pci/device.rs
```diff
@@ -0,0 +1,1236 @@
...
+            driver_status: 0,
+            config_generation: 0,
+            device_feature_select: 0,
+            driver_feature_select: 0,
+            queue_select: 0,
+            msix_config: VIRTQ_MSI_NO_VECTOR,
+            msix_queues: vec![VIRTQ_MSI_NO_VECTOR; num_queues],
+        });
+        let (device_activated, interrupt_status, cap_pci_cfg_info) =
+            (false, 0, VirtioPciCfgCapInfo::default());
```
roypat: nit: I dont quite get why we keep these 3 in local variables instead of just inlining them into the struct initializer below

src/vmm/src/vstate/vm.rs
```diff
@@ -34,6 +60,8 @@ pub struct VmCommon {
     max_memslots: usize,
     /// The guest memory of this Vm.
     pub guest_memory: GuestMemoryMmap,
+    /// Interrupts used by Vm's devices
+    interrupts: Arc<Mutex<HashMap<u32, RoutingEntry>>>,
```
Manciukic: do we actually need an `Arc` here if the `ArchVm` is already stored into a `Arc`?

src/vmm/src/vstate/vm.rs
```diff
@@ -52,6 +56,151 @@ pub struct RoutingEntry {
...
+            enabled: AtomicBool::new(enabled),
+        })
+    }
+}
+
+impl MsiVector {
+    /// Enable vector
+    fn enable(&self, vmfd: &VmFd) -> Result<(), errno::Error> {
+        if !self.enabled.load(Ordering::Acquire) {
+            vmfd.register_irqfd(&self.event_fd, self.gsi)?;
```
Manciukic, line 90: we don't need to track this through the `VmCommon` object? If the tracking is only for GSI, we should probably just rename that `register_gsi`?

(I'm still quite confused around the `gsi` vs `irq` nomenclature)

src/vmm/src/vstate/vm.rs
```diff
@@ -52,6 +56,151 @@ pub struct RoutingEntry {
...
+/// MSI interrupts created for a VirtIO device
+pub struct MsiVectorGroup {
+    vm: Arc<Vm>,
+    irq_routes: HashMap<u32, MsiVector>,
+}
+
+impl MsiVectorGroup {
+    /// Returns the number of vectors in this group
+    pub fn num_vectors(&self) -> u16 {
+        u16::try_from(self.irq_routes.len()).unwrap()
```
Manciukic, line 120: why safe? can we use `.expect("irq routes should be less than 16k")` or smth like that?

src/vmm/src/vstate/vm.rs
```diff
@@ -52,6 +56,151 @@ pub struct RoutingEntry {
...
+        for (_, route) in self.irq_routes.iter() {
+            route.disable(&self.vm.common.fd)?;
+        }
+
+        Ok(())
+    }
+
+    fn trigger(
+        &self,
+        index: vm_device::interrupt::InterruptIndex,
```
Manciukic: nit: can we just import `InterruptIndex`?

src/vmm/src/devices/virtio/transport/pci/device.rs
```diff
@@ -0,0 +1,1226 @@
...
+
+        // Dropping the MutexGuard to unlock the VirtioDevice. This is required
+        // in the context of a restore given the device might require some
+        // activation, meaning it will require locking. Dropping the lock
+        // prevents from a subtle deadlock.
+        std::mem::drop(locked_device);
```
Manciukic: nit: we can probably wrap the section we need the lock for inside a block to avoid this explicitly.

```
let (num_queues, device_type) = {
    let locked_device = device.lock().unwrap();
    (locked_device.queues().len(), locked_device.device_type())
};
let pci_device_id =
            VIRTIO_PCI_DEVICE_ID_BASE + u16::try_from(device_type).unwrap();
let (class, subclass) = match device_type {
[...]
}
```

src/vmm/src/devices/virtio/transport/pci/device.rs
```diff
@@ -0,0 +1,1226 @@
...
+            VirtioInterruptType::Queue(queue_index) => {
+                self.queues_vectors.lock().unwrap()[queue_index as usize]
+            }
+        };
+
+        self.interrupt_source_group
+            .notifier(vector as InterruptIndex)
+    }
+
+    fn status(&self) -> Arc<std::sync::atomic::AtomicU32> {
```
Manciukic: nit: it's imported already `s/std::sync::atomic::AtomicU32/AtomicU32/`

src/vmm/src/devices/virtio/block/virtio/device.rs
```diff
@@ -397,7 +397,7 @@ impl VirtioBlock {
 
         if queue.prepare_kick() {
             interrupt
-                .trigger(VirtioInterruptType::Queue(index))
```
Manciukic, line 400: what if we had multiple queues? is there any way we could get what's the queue index, or it's too big of a change as we don't support multiple queues yet?

src/vmm/src/devices/virtio/queue.rs
Manciukic, line 1: yeah, it's part of the virtio spec that size=0 means the queue is disabled.

src/vmm/src/device_manager/mod.rs
```diff
@@ -274,6 +281,106 @@ impl DeviceManager {
         self.pci_devices
             .attach_pci_segment(&self.resource_allocator)
     }
+
+    fn do_kick_device(virtio_device: Arc<Mutex<dyn VirtioDevice>>) {
```
Manciukic: since we're touching this, can we just add a function to the `VirtioDevice` trait to avoid this match? Either `is_activated` and `process_virtio_queues`, or just a `kick_device`.

src/vmm/src/device_manager/persist.rs
```diff
@@ -337,28 +289,31 @@ impl<'a> Persist<'a> for MMIODeviceManager {
...
                             transport_state,
-                            device_info: device.resources,
-                        })
+                            device_info,
+                        });
                     }
                 }
                 TYPE_NET => {
-                    let net = locked_device.as_any().downcast_ref::<Net>().unwrap();
+                    let net = locked_device.as_mut_any().downcast_mut::<Net>().unwrap();
```
Manciukic, line 299: do we need it as mut?

src/vmm/src/device_manager/persist.rs
```diff
@@ -311,20 +261,22 @@ impl<'a> Persist<'a> for MMIODeviceManager {
         let _: Result<(), ()> = self.for_each_virtio_device(|_, devid, device| {
             let mmio_transport_locked = device.inner.lock().expect("Poisoned lock");
             let transport_state = mmio_transport_locked.save();
+            let device_info = device.resources;
+            let device_id = devid.clone();
 
             let mut locked_device = mmio_transport_locked.locked_device();
             match locked_device.device_type() {
```
Manciukic, line 265: also here, can this be a trait function?

src/vmm/src/devices/virtio/net/device.rs
```diff
@@ -930,6 +931,26 @@ impl Net {
...
+    pub fn prepare_save(&mut self) {
+        // We shouldn't be messing with the queue if the device is not activated.
+        // Anyways, if it isn't there's nothing to prepare; we haven't parsed any
+        // descriptors yet from it and we can't have a deferred frame.
+        if !self.is_activated() {
+            return;
+        }
+
+        // Give potential deferred RX frame to guest
+        self.rx_buffer.finish_frame(&mut self.queues[RX_INDEX]);
```
Manciukic, line 945: what's the case in which there's some used descriptors but the `next_used` wasn't advanced? rate limiting?

src/vmm/src/vstate/vm.rs
```diff
@@ -622,23 +624,107 @@ impl Vm {
...
+                    .expect("Poisoned lock")
+                    .free(&old_range)
+                    .map_err(|_| {
+                        std::io::Error::other("pci: failed deallocating old MMIO range")
+                    })?;
+
+                allocator
+                    .lock()
+                    .unwrap()
+                    .allocate(len, len, vm_allocator::AllocPolicy::ExactMatch(new_base))
```
Manciukic: I think it's possible for the driver to relocate from 32b to 64b addresses (or at least that what happened to me when I had a conflict in the 32b address space iirc). So we may need to free an allocator and reallocate in another.

src/vmm/src/vstate/vm.rs
```diff
@@ -622,23 +624,107 @@ impl Vm {
...
+                    .enumerate()
+                {
+                    const NOTIFICATION_BAR_OFFSET: u64 = 0x6000;
+                    const NOTIFY_OFF_MULTIPLIER: u64 = 4;
+                    let notify_base = old_base + NOTIFICATION_BAR_OFFSET;
+                    let io_addr =
+                        IoEventAddress::Mmio(notify_base + i as u64 * NOTIFY_OFF_MULTIPLIER);
+                    self.common
+                        .fd
+                        .unregister_ioevent(queue_evt, &io_addr, NoDatamatch)?;
```
Manciukic: this requires some seccomp rules iirc
I'm not sure how we could test a relocation in an integ test, but we probably should.

src/vmm/src/arch/aarch64/gic/mod.rs
```diff
@@ -80,6 +86,14 @@ impl GICDevice {
         }
     }
 
+    /// Returns the file descriptor of the ITS device, if any
+    pub fn its_fd(&self) -> Option<&DeviceFd> {
+        match self {
+            Self::V2(_) => None,
```
Manciukic, line 92: do we still use GICv2? Can we just completely remove it?

src/vmm/src/vstate/resources.rs
```diff
@@ -19,19 +20,14 @@ use crate::arch;
...
     pub gsi_allocator: Arc<Mutex<IdAllocator>>,
-    // Allocator for memory in the 32-bit MMIO address space
+    /// Allocator for memory in the 32-bit MMIO address space
     pub mmio32_memory: Arc<Mutex<AddressAllocator>>,
-    // Allocator for memory in the 64-bit MMIO address space
+    /// Allocator for memory in the 64-bit MMIO address space
     pub mmio64_memory: Arc<Mutex<AddressAllocator>>,
-    // Memory allocator for system data
+    /// Memory allocator for system data
     pub system_memory: Arc<Mutex<AddressAllocator>>,
```
ShadowCurse: As far as I can see, the `Arc<Mutex>` is no longer needed here.

src/vmm/src/vstate/resources.rs
```diff
@@ -152,23 +145,74 @@ impl ResourceAllocator {
...
+pub struct ResourceAllocatorState {
+    /// Allocator for device interrupt lines
+    pub gsi_allocator: Arc<Mutex<IdAllocator>>,
+    /// Allocator for memory in the 32-bit MMIO address space
+    pub mmio32_memory: Arc<Mutex<AddressAllocator>>,
+    /// Allocator for memory in the 64-bit MMIO address space
+    pub mmio64_memory: Arc<Mutex<AddressAllocator>>,
+    /// Memory allocator for system data
+    pub system_memory: Arc<Mutex<AddressAllocator>>,
+}
```
ShadowCurse: This is just a copy of `ResourceAllocator`, so we don't even need the `State` object here. Just store `ResourceAllocator` directly.

src/vmm/src/vstate/vm.rs
```diff
@@ -276,14 +499,215 @@ impl Vm {
...
+            type_: KVM_IRQ_ROUTING_MSI,
+            ..Default::default()
+        };
+        entry.u.msi.address_lo = config.low_addr;
+        entry.u.msi.address_hi = config.high_addr;
+        entry.u.msi.data = config.data;
+
+        if self.common.fd.check_extension(kvm_ioctls::Cap::MsiDevid) {
+            // On AArch64, there is limitation on the range of the 'devid',
+            // it cannot be greater than 65536 (the max of u16).
```
ShadowCurse: Can you provide a link to the source?

src/vmm/src/vstate/vm.rs
```diff
@@ -276,14 +499,215 @@ impl Vm {
...
+            .expect("Poisoned lock")
+            .insert(route.gsi, RoutingEntry { entry, masked });
+
+        Ok(())
+    }
+
+    /// Create a group of MSI-X interrupts
+    pub fn create_msix_group(
+        vm: Arc<Vm>,
+        base: u32,
```
ShadowCurse: As I can see, the base is always 0. So maybe just remove this param?

src/vmm/src/device_manager/mod.rs
ShadowCurse: +1. This way this function will not even be needed. It will collapse into `kick_virtio_devices` into simple `for device in self.mmio_devices { device.kick() }`

src/vmm/src/vstate/vm.rs
```diff
@@ -34,6 +246,12 @@ pub struct VmCommon {
     max_memslots: usize,
     /// The guest memory of this Vm.
     pub guest_memory: GuestMemoryMmap,
+    /// Interrupts used by Vm's devices
+    pub interrupts: Arc<Mutex<HashMap<u32, RoutingEntry>>>,
+    /// Allocator for VM resources
+    pub resource_allocator: Arc<ResourceAllocator>,
+    /// MMIO bus
+    pub mmio_bus: Arc<vm_device::Bus>,
```
ShadowCurse: These fields should not be here. They all should be in the `DeviceManager` and it should be responsible for all interrupt/MSI/relocation handling. Since the `VmFd` will still be needed there, we can just copy it in the device manager (only `VmFd`, not the whole `Vm` struct)

src/vmm/src/builder.rs
bchalios: you sure you're not looking into an old version? I think these are dropped now

src/vmm/src/vstate/resources.rs
bchalios: We do need the `Mutex` if we want to keep these behind an `Arc`, right? Otherwise we can't get a `mut` reference for interacting with them. Let's resolve the discussion of the where should `ResourceAllocation` live and then revisit this one. WDYT?

src/vmm/src/vstate/resources.rs
bchalios: You are right. There's just a small complication that if we want to save this as `ResourceAllocator` it needs to be `Clone`. If we have the fields as `Arc` then that's fine, otherwise not.

src/vmm/src/vstate/vm.rs
bchalios: Can you please elaborate why this "should" be the case?

src/vmm/src/vstate/resources.rs
ShadowCurse: why do we need `Mutex`? As I can see, we always pass the ref to `ResourceAllocator` everywhere, not specific fields (but even if we need to pass fields, we don't need `Mutex`).

src/vmm/src/vstate/resources.rs
ShadowCurse: If we need it `Clone`, just make it `Clone`. I see that `IdAllocator` does not have it for some reason, so we need to update `vm-allocator` crate for this. Using `Arc`s as a workaround adds unnecessary complication.

src/vmm/src/vstate/resources.rs
bchalios: Because we need interior mutability. Because allocating memory or ids needs a `&mut` reference. If these live inside an object that it's behind an `Arc` which **must** be the case due to `DeviceRelocation`.

src/vmm/src/devices/virtio/transport/pci/device.rs
bchalios: I've refactored this bits a bit and dropped the `mem::drop`.

src/vmm/src/vstate/vm.rs
ShadowCurse: - Interrupt handling - device related activity
- Resource allocation - device related activity
- Mmio bus managment - device related activity

Everything you added to the `Vm` has to do with device management, so why not have it in `DeviceManager`?
Even in the commit message you mention in the commit that moves `ResourceAllocator` in `Vm` that you do this because it needs `VmFd`. But why not just copy the fd? There is nothing else in `Vm` needed for device management. By doing this we keep all device related activities in one place instead of spreading them all over the codebase.

src/vmm/src/devices/virtio/block/virtio/device.rs
bchalios: We could pass the queue index and `&self` in here and get it. But I think it's not worth the effort right now. WDYT?

src/vmm/src/devices/virtio/queue.rs
bchalios: Thanks for the hint.

src/vmm/src/device_manager/mod.rs
bchalios: yup, this should be doable

src/vmm/src/device_manager/persist.rs
bchalios: yep, `net.prepare_save()` takes a `&mut self`.

src/vmm/src/device_manager/persist.rs
bchalios: I've contemplated that for a while. This is a bit more convoluted, because we need to somehow abstract away as well the serialization of the transport layer. I _think_ it can be done, but I'd like to give it a go in a later PR. This one is too big already.

src/vmm/src/devices/virtio/net/device.rs
bchalios: Yes, that's the reason we need that.

src/vmm/src/vstate/vm.rs
bchalios: That's weird. The trait method here takes an argument that distinguishes between 32bit and 64bit addresses (and it takes only one of them). If what you say is true, what range does this argument refer to, the old or the new one?

src/vmm/src/vstate/vm.rs
bchalios: I have updated the seccomp for these. Yes, not sure how we could trigger those from an integration test. I have also tried to trigger a rescan, but I didn't manage to trigger a relocation. We could trigger from a unit test probably. I was hoping to somehow drive it through kani (or fuzzing)

src/vmm/src/arch/aarch64/gic/mod.rs
bchalios: We don't. Open source does (raspberry PIs :P). So, I wouldn't like to drop it, but we could definitely make this code cleaner.

src/vmm/src/vstate/resources.rs
bchalios: What about cloning performance. It's much cheaper to copy an Arc than a tree data structure.

src/vmm/src/vstate/vm.rs
bchalios: not "all over the codebase". Just in a single component that describes Vm resources. Like that we nicely abstract away things like "IO bus" only exists for x86 systems, which we already separate nicely withing `ArchVm`. TBH, I think that the `DeviceManager` itself should be as well part of the `Vm` object, but that's another discussion.

src/vmm/src/vstate/resources.rs
ShadowCurse: Well, there should be no cloning at all. After the allocator is `deserialized`, it is ready to be used.
Also, even though `Arc::clone` is cheaper than full deep copy, it is much more expensive to create `Arc` in first place , compared to no doing it.

src/vmm/src/vstate/resources.rs
ShadowCurse: oh, ok. Can we do with just `Mutex` on the whole `ResourceAllocator`? (no `Arc`s anywhere) As I can see now it is only in `Arc` in the `Vm` and not shared with anything.

src/vmm/src/vstate/resources.rs
bchalios: The cloning is needed during the snapshot part. We cannot "take" the `ResourceAllocator` object outside its container, we need to "clone" the state. So, if we use the `ResourceAllocator` as the state it needs to be cloned.

src/vmm/src/vstate/vm.rs
bchalios: You are right. I dropped the `Arc`.

src/vmm/src/vstate/vm.rs
bchalios: What do you mean "track it through the `VmCommon` object?

It should be named `register_gsi`, I agree with you. The naming here comes from the KVM nomenclature `KVM_IRQFD`.

So, generally speaking IRQ typically is associated with traditional IOAPIC interrupts, whereas GSI is a generic term referring to a generic interrupt source. So, I think that GSI is a better term when there are both traditional IRQs and MSI vectors.

src/vmm/src/vstate/vm.rs
bchalios: This is safe because the create `self.irq_routes` throught `Vm::create_msix_group` function for which `count` is `u16`. Added a comment mentioning that.

src/vmm/src/vstate/vm.rs
bchalios: yup. Did that, also for `InterruptSourceIndex`

src/vmm/src/devices/virtio/transport/pci/device.rs
bchalios: Changed it

src/vmm/src/vstate/resources.rs
ShadowCurse: doing one `clone` during snap creation of a some tree structure with ~0 elements will not affect perf in any way.

src/vmm/src/device_manager/mod.rs
bchalios: Done, added as an extra commit

src/vmm/src/vstate/resources.rs
bchalios: Yes, I can get rid of the `Arc`s. I think we can't get a single `Mutex` in the whole of `ResourceAllocator` because there are parts of the PCI crate that expect a `&mut` to both 32bit and 64bit allocators. If we get a `Mutex` on the whole object then we won't be able to do that, because we 'd need to get a mutable reference of the same object twice.

bchalios: Here: https://github.com/bchalios/firecracker/blob/8fefa5b44bd72a2b9c7d06bf362fe4134cab4a80/src/vmm/src/device_manager/pci_mngr.rs#L153.

src/vmm/src/vstate/resources.rs
bchalios: Yup, I agree on that. I'll see how easy it is to get `Clone` for `IdAllocator`.

src/vmm/src/vstate/resources.rs
ShadowCurse: well, how about just changing `allocate_bars` api to take `&mut ResourceAllocator` as an arg? As I can see, we will need to move `allocate_bars` from `PciDevice` trait into `VirtioPciDevice` impl for that.
In general, because the crate is in our possession, if it does not fit our needs, we can just change it.

src/vmm/src/vstate/vm.rs
ShadowCurse: After offline discussion, I was convinced that moving all `Vm` related things (like `DeviceManager`, `vcpus` etc) into `Vm` type is a good solution. Current approach, though not perfect, will be updated in later PR.

src/vmm/src/vstate/resources.rs
bchalios: For the time being, I removed the `Arc<Mutex<>>` containers from the fields. Now the whole `ResourceAllocator` is within a `Mutex`. We still need this for internal mutability as `Vm` is within an `Arc`.

src/vmm/src/vstate/resources.rs
bchalios: Done

src/vmm/src/vstate/vm.rs
bchalios: Thanks for spotting this. I looked again into the Linux code and couldn't find any limitation specific to Aarch64. The KVM documentation does says: " For PCI, this is usually a BDF identifier in the lower 16 bits.". in our case these are the only bits that are non zero so we can just pass `config.devid`.

src/vmm/src/vstate/vm.rs
bchalios: done

src/vmm/src/vstate/vm.rs
bchalios: I don't thing anything would do that, concurrently try to enable/disable the same vector of an MSI-X.

src/vmm/src/vstate/vm.rs
bchalios: I thought I had gotten rid of those. Weird that clippy did not complain.

src/vmm/src/vstate/vm.rs
bchalios: cool! thanks

src/vmm/src/vstate/vm.rs
bchalios: Yup. We should definitely get rid of those.

src/vmm/src/devices/virtio/transport/pci/device.rs
bchalios: I didn't either. These were removed in later iterations of the PR.

src/vmm/src/vstate/vm.rs
bchalios: This whole comment was inaccurate. I've reworked it.

src/vmm/src/vstate/vm.rs
bchalios: yup, nice catch. Thanks

src/vmm/src/vstate/vm.rs
Manciukic: I mean that we're tracking all the other GSI in `vm.register_irq`. Why is this function calling directly into the fd? Why don't we need to keep track of these GSIs?

src/vmm/src/devices/virtio/block/virtio/device.rs
Manciukic: fine by me. it seems like a small change anyways

src/vmm/src/vstate/vm.rs
Manciukic: I think I am wrong. I added the 32b->64b logic in a previous POC because I was having some issues with relocation but I don't think that extra logic was ever triggered (checking now the relocation was between two 32b addresses it seems). 

Looking at the spec it seems it's the device that decides whether to support 32b only or the full 64b. The driver should respect the advertised 32/64 "capability" of the device (quotes as it's not a real capability in the pci sense). The case where what I said would happen is if the device presents a 64b address that's contained in the first 4GB (so bar type 64b but upper 32b are 0). But that shouldn't happen in our case as we're marking devices in the 32b region as having a 32b BAR.

ShadowCurse (approved):

src/vmm/src/vstate/vm.rs
```diff
@@ -276,14 +506,200 @@ impl Vm {
...
+
+    /// Register a device IRQ
+    pub fn register_irq(&self, fd: &EventFd, gsi: u32) -> Result<(), errno::Error> {
+        self.common.fd.register_irqfd(fd, gsi)?;
+
+        let mut entry = kvm_irq_routing_entry {
+            gsi,
+            type_: KVM_IRQ_ROUTING_IRQCHIP,
+            ..Default::default()
+        };
```
ShadowCurse, line 516: Why do we need to store routing entries for the normal irqs? Shouldn't `set_gsi_routing` only update interrupts which are passed in the list?

bchalios: Nope. KVM needs to know all of the interrupts that are enabled. The "routing" part here essentially is what tells KVM where to route the interrupt, MSI or IOAPIC/GIC.

src/vmm/src/vstate/vm.rs
bchalios: Oh, that is because we are only enabling GSIs that correspond to MSI interrupts when they get enabled and not masked. If they're masked we don't need to register them with KVM. Is that what you are asking?

src/vmm/src/vstate/vm.rs
ShadowCurse: hmm.. sounds super strange, but fine, let it be for now

ShadowCurse (approved):

src/vmm/src/vstate/vm.rs
bchalios: That makes sense. I was looking at the code that handles the guest access that triggers relocation. It explicitly handles differently 64bit and 32bit cases. It only handles 64bit cases when both of the BARs are written.

src/vmm/src/vstate/vm.rs
Manciukic: Got it. IRQs are always enabled so `register_irq` also takes care of adding it into the `interrupts` hashmap, but for MSI we need to keep track whether they're masked or not, so `register_msi` only tracks it, while `enable/disable` take care of only (un)registering it with KVM.

src/vmm/src/vstate/vm.rs
Manciukic: I was double checking my PoC code and `KVM_IOEVENTFD` should be allowed in seccomp for this call to work. https://github.com/Manciukic/firecracker/commit/faa464031adabee8e88faff63e768e641dd33140

src/vmm/src/vstate/vm.rs
```diff
@@ -276,14 +506,200 @@ impl Vm {
...
+                // Update MMIO bus
+                self.common
+                    .mmio_bus
+                    .update_range(old_base, len, new_base, len)
+                    .map_err(std::io::Error::other)?;
+            }
+        }
+
+        if let Some(virtio_pci_dev) = pci_dev.as_any_mut().downcast_mut::<VirtioPciDevice>() {
+            if virtio_pci_dev.config_bar_addr() == new_base {
```
Manciukic: I was playing around with relocation and got an error:
```
ioctl(31, KVM_IOEVENTFD, 0x7128c45ff1f8) = -1 ENOENT (No such file or directory)
```

This fixes the issue:

```suggestion
            if virtio_pci_dev.config_bar_addr() == old_base {
```

bchalios: For the record, we decided to remove support for device relocation at the moment. In order to properly test this we need to be able to reset VirtIO devices which is something we don't support at the moment. We will evaluate whether to support this feature with this release at a later point.

Manciukic (approved):

PR #5245, opened 2025-06-03
Add more assertions

ShadowCurse:
## Changes
Add asserts in places where we have strong assumptions about inner workings of the code.

## Reason
Adding asserts improves security posture and allows assumptions to be stated more verbosely.
These asserts are only related to the parts of the code which do not depend on the user input.

## License Acceptance

[`CONTRIBUTING.md`][3].

## PR Checklist

  automated style checks.
  in the PR.
  integration tests.

### DIFF
--- a/src/vmm/src/vstate/vcpu.rs
+++ b/src/vmm/src/vstate/vcpu.rs
@@ -5,7 +5,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 
-use std::cell::Cell;
+use std::cell::RefCell;
 #[cfg(feature = "gdb")]
 use std::os::fd::AsRawFd;
 use std::sync::atomic::{Ordering, fence};
@@ -14,9 +14,9 @@ use std::sync::{Arc, Barrier};
 use std::{fmt, io, thread};
 
 use kvm_bindings::{KVM_SYSTEM_EVENT_RESET, KVM_SYSTEM_EVENT_SHUTDOWN};
-use kvm_ioctls::VcpuExit;
 #[cfg(feature = "gdb")]
 use kvm_ioctls::VcpuFd;
+use kvm_ioctls::{KvmRunWrapper, VcpuExit};
 use libc::{c_int, c_void, siginfo_t};
 use log::{error, info, warn};
 use vmm_sys_util::errno;
@@ -69,9 +69,6 @@ pub struct VcpuConfig {
     pub cpu_config: CpuConfiguration,
 }
 
-// Using this for easier explicit type-casting to help IDEs interpret the code.
-type VcpuCell = Cell<Option<*mut Vcpu>>;
-
 /// Error type for [`Vcpu::start_threaded`].
 #[derive(Debug, derive_more::From, thiserror::Error)]
 #[error("Failed to spawn vCPU thread: {0}")]
@@ -88,7 +85,10 @@ pub enum CopyKvmFdError {
     CreateVcpuError(#[from] kvm_ioctls::Error),
 }
 
-thread_local!(static TLS_VCPU_PTR: VcpuCell = const { Cell::new(None) });
+// Stores the mmap region of `kvm_run` struct for the current Vcpu. This allows for the
+// signal handler to safely access the `kvm_run` even when Vcpu is dropped and vcpu fd
+// is closed.
+thread_local!(static TLS_VCPU_PTR: RefCell<Option<KvmRunWrapper>> = const { RefCell::new(None) });
 
 /// A wrapper around creating and using a vcpu.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/vmm/src/vstate/vm.rs
```diff
@@ -362,9 +362,11 @@ pub(crate) mod tests {
 
             let res = vm.register_memory_region(region);
 
-            if i >= max_nr_regions {
+            // For some reason max_nr_regions is considered unused here
```
roypat: that's because `matches!` can only compare against constants. the assertion below is actually creating a new temporary binding for the `u32` stored in the `VmError` which is also called `max_nr_regions` and which shadows the variable

ShadowCurse: fixed by using normal `match` instead

tools/devtool
```diff
@@ -526,6 +526,22 @@ cmd_build() {
...
+        --privileged \
+        --workdir "$CTR_FC_ROOT_DIR" \
+        -- \
+        cargo $@
+
+    # Running as root would have created some root-owned files under the build
+    # dir. Let's fix that.
+    cmd_fix_perms
+}
+
```
pb8o: You can already do `./tools/devtool sh cargo`

ShadowCurse: fair, dropped the commit.

src/vmm/src/vstate/vm.rs
```diff
@@ -362,14 +362,11 @@ pub(crate) mod tests {
...
-                    "{:?} at iteration {} - max_nr_memslots: {}",
-                    res,
-                    i,
-                    max_nr_regions
-                );
+            if max_nr_regions <= i {
+                match res {
+                    Err(VmError::NotEnoughMemorySlots(v)) => assert_eq!(v, max_nr_regions),
+                    _ => unreachable!(),
+                }
```
roypat: nit:
```suggestion
                assert!(
                    matches!(res, Err(VmError::NotEnoughMemorySlots(v)) if v == max_nr_regions),
                    "{:?} at iteration {}",
                    res,
                    i
                );
```

roypat (approved):

src/vmm/src/vstate/vcpu.rs
```diff
@@ -611,7 +587,7 @@ impl Drop for Vcpu {
 
                 // Have to do this trick because `update` method is
                 // not stable on Cell.
-                Self::TLS_VCPU_PTR.with(|cell| cell.take());
+                TLS_VCPU_PTR.with(|cell| cell.take());
```
roypat: if we're touching this anyway, this can just be `Cell::take()`, as `Option::<T>::default()` is `None`

src/vmm/src/vstate/vm.rs
ShadowCurse: applied

src/vmm/src/vstate/vcpu.rs
ShadowCurse: done

roypat (dismissed):

src/vmm/src/vstate/vcpu.rs
```diff
@@ -615,7 +595,23 @@ fn handle_kvm_exit(
...
+            // put on their threads. This can happen if some error occurs during Vmm
+            // setup before `start_threaded` call.
+            if let Some(_vcpu_ptr) = cell.get() {
+                // During normal runtime there is a strong assumption that vcpus will be
+                // put on their own threads, thus TLS will be initialized with the
+                // correct pointer.
+                // In test we do not put vcpus on separate threads, so TLS will have a value
+                // of the last created vcpu.
+                #[cfg(not(test))]
+                assert!(std::ptr::eq(_vcpu_ptr, self));
```
Manciukic: I think this is one of those cases where having a separate function return a result and then the caller decide what to do with it is better than having a assert that depends on whether we're testing or not, as it makes the code actually testable without hacks like this.

All this change would be is unwrap the result of `self.reset_thread_local_data()`

src/vmm/src/vstate/vcpu.rs
```diff
@@ -110,61 +112,35 @@ pub struct Vcpu {
...
     /// This function will panic if there already is a `Vcpu` present in the TLS.
     fn init_thread_local_data(&mut self) {
-        Self::TLS_VCPU_PTR.with(|cell: &VcpuCell| {
+        TLS_VCPU_PTR.with(|cell: &VcpuCell| {
             assert!(cell.get().is_none());
             cell.set(Some(self as *mut Vcpu));
         })
     }
 
-    /// Runs `func` for the `Vcpu` associated with the current thread.
```
Manciukic, line 152: I think this change is really subjective. I see no issues with the current state and I actually prefer that. What is the benefit of doing like so?

src/vmm/src/vstate/vcpu.rs
roypat: On the other hand, why do we have a test that fails this assertion? Been a while since I looked at this PR, but maybe we should fix that test to actually set things up the same way we do in prod?

ShadowCurse: added another commit which drops the `VcpuDrop`

src/vmm/src/vstate/vcpu.rs
ShadowCurse: It was used only once, so inlining it made sense. It also makes code a bit shorter.

roypat (dismissed):

src/vmm/src/vstate/vcpu.rs
roypat: personally, a huge fan. Thanks!

Cargo.lock
```diff
@@ -120,9 +120,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
```
Manciukic: why are we updating the Cargo.lock?

ShadowCurse: removed

ShadowCurse:
@Manciukic added reasons to the PR description.

Manciukic (approved):

